PR #4824, opened 2019-05-02
Adds unit tests to all trigger methods in TriggerAttachment; minor other refactoring.

MelvinWM:
## Overview

The changes adds somewhat superficial unit tests to all the trigger methods in TriggerAttachment, partially to help increase test coverage, and partially to enable making further refactorings in TriggerAttachment in future pull requests, such as the minor amount of duplicated code across the 19 "trigger methods" in TriggerAttachment. Another potential refactoring might be to create an immutable parameter object that only contains the fields `beforeOrAfter, stepName, useUses, testUses, testChance, testWhen`, and then change the parameters of the trigger methods to use that parameter object instead, though that would require updating other parts of the codebase that uses those methods.

I have updated and changed some tests to use less mocking as I got more experience with the code base, mocking as well as Mockito.

## Functional Changes

Hopefully none.

## Manual Testing Performed

Played through a single game, no manual testing apart from that.

## Additional Review Notes

There are several questions and issues:
- Many of the unit tests are somewhat superficial, are the introduced tests valuable enough to keep as is?
- Both `triggerNotifications()` and `triggerVictory()` were changed to support passing in `NotificationMessages`, since it otherwise would not be easy to mock in the unit tests. However, partially that is less nice, and partially a duplicated test is currently used for both of them to prevent test failures in other tests. The relevant parts are currently covered with TODOs.
- There is also a very minor TODO reg. moving a test down such that its definition order fits with the corresponding methods under test order in TriggerAttachment.
- The usage of mocking has been decreased. I have only limited experience with Mockito as well as best practices reg. mocking, so I am in doubt about whether I have used mocking well reg. the tests.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/attachments/TriggerAttachment.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/TriggerAttachment.java
@@ -1441,6 +1441,22 @@ public class TriggerAttachment extends AbstractTriggerAttachment {
   public static void triggerNotifications(final Set<TriggerAttachment> satisfiedTriggers, final IDelegateBridge bridge,
       final String beforeOrAfter, final String stepName, final boolean useUses, final boolean testUses,
       final boolean testChance, final boolean testWhen) {
+
+    // TODO: This check is currently (2019-04-28) needed due to another test failing otherwise, namely
+    // MustFightBattleTest, reg. property files. That may or may not indicate a bug and/or other.
+    if (satisfiedTriggers.stream().anyMatch(t -> t.getNotification() != null)) {
+      triggerNotifications(
+          satisfiedTriggers, bridge, beforeOrAfter, stepName, useUses, testUses, testChance, testWhen,
+          NotificationMessages.getInstance());
+    }
+  }
+
+  @VisibleForTesting
+  static void triggerNotifications(final Set<TriggerAttachment> satisfiedTriggers, final IDelegateBridge bridge,
+      final String beforeOrAfter, final String stepName, final boolean useUses, final boolean testUses,
+      final boolean testChance, final boolean testWhen,
+      final NotificationMessages notificationMessages) {
+
     final GameData data = bridge.getData();
     Collection<TriggerAttachment> trigs = CollectionUtils.getMatches(satisfiedTriggers, notificationMatch());
     if (testWhen) {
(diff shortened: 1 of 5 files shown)


MelvinWM:
@RoiEXLab Thank you very much for the very kind words, and I am happy to contribute. I will say that I currently have more time at the moment than I will have in the future, but I do intend to contribute a little once in a while, mostly minor refactorings, bug fixes, and sometimes documentation updates and the like. You seem like nice and good developers, and you also seem to care about quality and the users, which I find very nice (having at least some quality can make development much more fun and enjoyable, and it also helps make it worth investing in the project). And judging from the state of the project, I get the impression that you have done a (very) good job, especially for an open-source game project. I also like to learn and share what I know, and I hope to be able to learn from you and share with you in the future.

Reg. the code review, I have pushed a commit. I still have some TODOs in the changes, two of them reg. the added checks, and the third one about moving a test method down to fit with the order of the corresponding methods that are tested. Should I look into the ones reg. the added checks and see if there is a way to avoid needing the checks?

RoiEXLab:
@MelwinWM My answer is simple: If you feel like it's worth your time then go for it.
Otherwise we won't force you to do something just because it might be a good idea.

MelvinWM:
@RoiEXLab Thank you, I will look into it a bit further then, and then come back and remove the 'WIP' part of the pull request once I think it might be ready :smiley: .

game-core/src/main/java/games/strategy/triplea/attachments/TriggerAttachment.java
```diff
@@ -1441,6 +1431,22 @@ private static void placeUnits(final TriggerAttachment t, final Territory terr,
   public static void triggerNotifications(final Set<TriggerAttachment> satisfiedTriggers, final IDelegateBridge bridge,
       final String beforeOrAfter, final String stepName, final boolean useUses, final boolean testUses,
       final boolean testChance, final boolean testWhen) {
+
+    // NOTE: The check is needed to ensure that UI-related code (namely 'NotificationMessages.getInstance()') is
+    // not executed when unit testing non-UI related code such as 'MustFightBattleTest'.
+    if (satisfiedTriggers.stream().anyMatch(notificationMatch())) {
```
RoiEXLab: Thanks for digging into this, looks like we're still a long way from a good modular design.
However the overloaded method is a good way to add unit tests nevertheless. 👍

PR #4828, opened 2019-05-04
Fix current player logic in battle simulator when history is open.

asvitkine:
## Overview
When using the battle simulator while history is open, the attacking player was not being properly selected.

## Functional Changes
The logic for choosing the attacking player has been adjusted to take into account the selected history node, when viewing a past battle.

## Manual Testing Performed
- In a game, play several turns.
- Open history viewer and select a past combat step.
- Mouse over a territory where battle is taking place and open the Battle Simulator.
- The correct active player for that history turn should be set.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/history/History.java
+++ b/game-core/src/main/java/games/strategy/engine/history/History.java
@@ -3,8 +3,8 @@ package games.strategy.engine.history;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.List;
+import java.util.Optional;
 
-import javax.annotation.Nullable;
 import javax.swing.SwingUtilities;
 import javax.swing.tree.DefaultMutableTreeNode;
 import javax.swing.tree.DefaultTreeModel;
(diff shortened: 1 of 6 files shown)


game-core/src/main/java/games/strategy/triplea/odds/calculator/OddsCalculatorPanel.java
```diff
@@ -500,6 +506,17 @@
     revalidate();
   }
 
+  private @Nullable PlayerId getCurrentPlayer(final History history) {
```
ron-murhammer: Would it be better to return an Optional rather than make it nullable here?

game-core/src/main/java/games/strategy/engine/history/History.java
```diff
@@ -59,6 +61,13 @@ public void goToEnd() {
     }
   }
 
+  public @Nullable PlayerId getActivePlayer() {
```
ron-murhammer: Would it be better to return an Optional rather than make it nullable here?

ron-murhammer:
@asvitkine Great fix! I've noticed this before as well and it was somewhere on my long list of things to fix. Just some small minor comments.

game-core/src/main/java/games/strategy/engine/history/History.java
asvitkine: Done.

game-core/src/main/java/games/strategy/triplea/odds/calculator/OddsCalculatorPanel.java
asvitkine: Done.

asvitkine:
Comments addressed. I added a helper function to PlayerId for the logic of converting a Nullable PlayerId to an Optional.

PR #4830, opened 2019-05-05
Refactors trigger attachment and related.

MelvinWM:
## Overview

The first commit refactors filtering in the different trigger methods, and the second commit refactors away the long parameter lists for the different trigger methods as well as elsewhere by using a parameter object.

## Functional Changes

It should only be refactoring.

## Manual Testing Performed

Played a simple game including Hard AI (Middle Earth).

## Additional Review Notes

I believe the second commit is valuable, though while it overall seems like an improvement to me, parameter objects are in some ways less nice.

I have some TODOs reg. the documentation of the parameter object, since I am not quite certain where to find all the relevant documentation or what exactly they do (I have added some references I found).

I tried to refactor away the part reg.:

```java
if (fireTriggerParams.testChance && !t.testChance(bridge)) {
  continue;
}
if (fireTriggerParams.useUses) {
  t.use(bridge);
}
```

, but the code snippet has a fair bit of side-effects, and I did not find an approach that would preserve the order of statement execution while making the code seem better.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/attachments/FireTriggerParams.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/FireTriggerParams.java
@@ -1,0 +1,75 @@
+package games.strategy.triplea.attachments;
+
+
+/**
+ * Parameter object for fire trigger methods, first and foremost in {@link TriggerAttachment}.
+ */
+public class FireTriggerParams {
+
+  /**
+   * Filter value for before or after a given step.
+   *
+   * <p>
+   * See {@link #testWhen}, documentation for trigger attachment option "when" as well as
+   * {@link AbstractTriggerAttachment#whenOrDefaultMatch} .
+   * </p>
+   */
+  public final String beforeOrAfter;
+
+  /**
+   * Filter value for a specific step/phase (like "russianPolitics" step).
+   *
+   * <p>
+   * See {@link #testWhen}, documentation for trigger attachment option "when" as well as
+   * {@link AbstractTriggerAttachment#whenOrDefaultMatch} .
+   * </p>
+   */
+  public final String stepName;
+
+  /**
+   * TODO: Document.
+   *
+   * <p>
+   * See documentation for trigger attachment option "uses".
+   * </p>
+   */
+  public final boolean useUses;
+
+  /**
+   * TODO: Document.
+   *
+   * <p>
+   * See documentation for trigger attachment option "uses".
+   * </p>
+   */
+  public final boolean testUses;
+
+  /**
+   * TODO: Document.
+   *
+   * <p>
+   * See documentation for trigger attachment option "chance".
+   * </p>
+   */
+  public final boolean testChance;
+
+  /**
+   * Whether to filter for "when".
+   *
+   * <p>
+   * See documentation for trigger attachment option "when" as well as
+   * {@link AbstractTriggerAttachment#whenOrDefaultMatch} .
+   * </p>
+   */
+  public final boolean testWhen;
+
+  public FireTriggerParams(final String beforeOrAfter, final String stepName,
+      final boolean useUses, final boolean testUses, final boolean testChance, final boolean testWhen) {
+    this.beforeOrAfter = beforeOrAfter;
+    this.stepName = stepName;
+    this.useUses = useUses;
+    this.testUses = testUses;
+    this.testChance = testChance;
+    this.testWhen = testWhen;
+  }
+}
--- a/game-core/src/main/java/games/strategy/triplea/attachments/FireTriggerParams.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/FireTriggerParams.java
@@ -1,9 +1,14 @@
 package games.strategy.triplea.attachments;
 
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.Value;
 
 /**
  * Parameter object for fire trigger methods, first and foremost in {@link TriggerAttachment}.
  */
+@Value
+@Getter(AccessLevel.NONE)
 public class FireTriggerParams {
 
   /**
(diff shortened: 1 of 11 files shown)


game-core/src/main/java/games/strategy/triplea/attachments/TriggerAttachment.java
```diff
@@ -1420,6 +1422,29 @@ private static void placeUnits(final TriggerAttachment t, final Territory terr,
...
+      final Set<TriggerAttachment> satisfiedTriggers, final Predicate<TriggerAttachment> customPredicate,
+      final FireTriggerParams fireTriggerParams) {
+    Predicate<TriggerAttachment> combinedPredicate = customPredicate;
+    if (fireTriggerParams.testWhen) {
+      combinedPredicate = combinedPredicate
+          .and(whenOrDefaultMatch(fireTriggerParams.beforeOrAfter, fireTriggerParams.stepName));
+    }
+    if (fireTriggerParams.testUses) {
+      combinedPredicate = combinedPredicate.and(availableUses);
+    }
```
RoiEXLab: We have a `PredicateBuilder` class that simplifies this process.

game-core/src/main/java/games/strategy/triplea/attachments/TriggerAttachment.java
```diff
@@ -2313,39 +2230,31 @@ public static void triggerActivateTriggerOther(final Map<ICondition, Boolean> te
...
-      if (testChance && !t.testChance(bridge)) {
+      if (fireTriggerParams.testChance && !t.testChance(bridge)) {
         continue;
       }
-      if (useUses) {
+      if (fireTriggerParams.useUses) {
         t.use(bridge);
       }
-      if (t.getVictory() == null || t.getPlayers() == null) {
+      if (t.getPlayers() == null) {
```
RoiEXLab: I think you need to clarify this change, is `getVictory()` guaranteed to be non-null?

game-core/src/main/java/games/strategy/triplea/attachments/FireTriggerParams.java
```diff
@@ -0,0 +1,75 @@
...
+
+  public FireTriggerParams(final String beforeOrAfter, final String stepName,
+      final boolean useUses, final boolean testUses, final boolean testChance, final boolean testWhen) {
+    this.beforeOrAfter = beforeOrAfter;
+    this.stepName = stepName;
+    this.useUses = useUses;
+    this.testUses = testUses;
+    this.testChance = testChance;
+    this.testWhen = testWhen;
+  }
```
RoiEXLab: I'm not entirely sure about this class. Using public non-static fields is not really common for java.
You _could_ use lombok to generate getters for you but I'm not entirely sure either if this actually is better.

RoiEXLab:
I like the change in general, just some minor things to consider 👍

game-core/src/main/java/games/strategy/triplea/attachments/TriggerAttachment.java
MelvinWM: Very good point. I removed it because that check is contained inside `victoryMatch`, which is used to filter those `TriggerAttachment`s before the `for`-loop. I believe that the code inside the `for`-loop does not at any point modify `victory`, though I must admit that I am not fully certain about it.

game-core/src/main/java/games/strategy/triplea/attachments/FireTriggerParams.java
MelvinWM: I agree. I did actually try to use Lombok, but the "getter" for the boolean fields ended up with `is` prepended, and that was verbose and did not look quite right (though I could probably make the variable names I used for `FireTriggerParams` in the various places more concise). I looked into renaming the "accessors", but it seems to be an experimental feature in Lombok: https://projectlombok.org/features/experimental/Accessors .

Thinking about it, I might be able to use Lombok, and then disable automatic generation of "getters" and create them myself, such that `equals` and `hashCode` is generated while still getting decently named "getters".

About a kind of data class like this, you are fully right that it is non-idiomatic for Java in multiple regards, though I believe it makes somewhat sense in this case, maybe if I improved it further. The trouble with not using "getters" in Java is that you then cannot modify how it is accessed later. In languages such as (I believe) C# and Scala (and Kotlin?), you can add "getters" later, maybe including reg. binary compatibility, which makes that aspect much less of an issue.

Likewise for the immutable, final access, that is non-typical for Java, for multiple reasons. In many cases, encapsulation and protecting access to members make a lot of sense, since in many ways the specific internal implementation is not meant to be a part of the interface or be used. In other cases, it may make sense to make the fields directly accessible (though maybe through a "getter"), generally if the fields are meant to be a part of the interface. This is still somewhat verbose in Java unless you use Lombok or similar. There are investigations in regards to maybe adding something in this direction to the Java language in the future by Brian Goetz and others: http://cr.openjdk.java.net/~briangoetz/amber/datum.html .

MelvinWM: I have just looked into creating "getters" named for instance `testChance`, since `isTestChance` seemed off. But that looks off in the following line, since there is already a method on `AbstractTriggerAttachment` also named `testChance`:

`if (fireTriggerParams.testChance() && !t.testChance(bridge)) {`.

I am not certain about the approach. Maybe keep the fields `public` and `final`, disable "getters", and use `Value` to auto-generate `equals`, `hashCode` and `toString`?

RoiEXLab:
@MelvinWM Could you do me a favour?
I just noticed 
```java
messageForRecord = messageForRecord.replaceAll("\\<br.*?>", " ");
messageForRecord = messageForRecord.replaceAll("\\<.*?>", "");
```
exists 2 times in TriggerAttachment.
Could you replace it with
```java
messageForRecord = messageForRecord.replaceAll("<br.*?>", " ");
messageForRecord = messageForRecord.replaceAll("<.*?>", "");
```
? The backslash is completely redundant for the regex.

I'd create a PR for it, but depending on what you intend to change as part of this PR (it's still labeled WIP) this could potentially introduce merge conflicts I don't want to risk.

MelvinWM:
@RoiEXLab I apologize, I should have followed up sooner. Apart from cleaning up the `\\` from the regexes, I have also made minor changes to the `FireTriggerParams` parameter class. If those changes are OK, the pull request should be ready.

PR #4831, opened 2019-05-08
Divide isSub into Attributes

ron-murhammer:
## Overview
- Second part of https://forums.triplea-game.org/topic/328/unit-option-can-submerge-hide-for-land-units-partisan-guerrilla-spy-diplomat-munition
- Breaking isSub and isDestroyer into smaller attributes so more fine grained control can be achieved (canHide, firstStrike, canTarget, canBeTargetedBy, etc)
- This should have no impact on existing maps and sea subs. Most functionality will now work for land/air subs but there are still some further changes need especially for cantTarget and cantBeTargetedBy as these are hardcoded to just check if empty or not.

## Functional Changes
**Separate isSub Functions**
1. **canEvade** - submerge
2. **isFirstStrike** - roll and take casualties before regular units
3. **canNotTarget** - can't target specific units like air units
4. **canNotBeTargetedBy** - can't be targeted by specific units like air units
5. **canMoveThroughEnemies** - unblockable, Treat Hostile Territories as Friendly
6. **canBeMovedThroughByEnemies** - enemies can move through territories if only these types of units are present (would also need blitz for enemy land territories)

## Manual Testing Performed
- Launched a few maps including revised, ww2v3, and global to ensure games still work properly

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
@@ -106,12 +106,36 @@ public final class Matches {
     return unit -> UnitAttachment.get(unit.getType()).getIsSea();
   }
 
-  public static Predicate<Unit> unitIsSub() {
-    return unit -> UnitAttachment.get(unit.getType()).getIsSub();
+  // public static Predicate<Unit> unitIsSub() {
+  // return unit -> UnitAttachment.get(unit.getType()).getIsSub();
+  // }
+
+  public static Predicate<Unit> unitHasSubBattleAbilities() {
+    return unitCanEvade().or(unitIsFirstStrike()).or(unitCantBeTargetedByAll());
   }
 
-  public static Predicate<Unit> unitIsNotSub() {
-    return unitIsSub().negate();
+  public static Predicate<Unit> unitCanEvade() {
+    return unit -> UnitAttachment.get(unit.getType()).getCanEvade();
+  }
+
+  public static Predicate<Unit> unitIsFirstStrike() {
+    return unit -> UnitAttachment.get(unit.getType()).getIsFirstStrike();
+  }
+
+  public static Predicate<Unit> unitCanMoveThroughEnemies() {
+    return unit -> UnitAttachment.get(unit.getType()).getCanMoveThroughEnemies();
+  }
+
+  public static Predicate<Unit> unitCanBeMovedThroughByEnemies() {
+    return unit -> UnitAttachment.get(unit.getType()).getCanBeMovedThroughByEnemies();
+  }
+
+  public static Predicate<Unit> unitCantTargetAll() {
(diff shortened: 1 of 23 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
```diff
@@ -106,12 +106,36 @@ private Matches() {}
     return unit -> UnitAttachment.get(unit.getType()).getIsSea();
   }
 
-  public static Predicate<Unit> unitIsSub() {
-    return unit -> UnitAttachment.get(unit.getType()).getIsSub();
+  // public static Predicate<Unit> unitIsSub() {
+  // return unit -> UnitAttachment.get(unit.getType()).getIsSub();
+  // }
```
RoiEXLab: This commented out code should be removed

game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
```diff
@@ -106,12 +106,36 @@ private Matches() {}
...
+
+  public static Predicate<Unit> unitCanMoveThroughEnemies() {
+    return unit -> UnitAttachment.get(unit.getType()).getCanMoveThroughEnemies();
+  }
+
+  public static Predicate<Unit> unitCanBeMovedThroughByEnemies() {
+    return unit -> UnitAttachment.get(unit.getType()).getCanBeMovedThroughByEnemies();
+  }
+
+  public static Predicate<Unit> unitCantTargetAll() {
```
MelvinWM: Is this a duplicate of `unitCantBeTargetedByAll`?

game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
```diff
@@ -106,12 +106,36 @@ private Matches() {}
...
+    return unit -> !UnitAttachment.get(unit.getType()).getCantBeTargetedBy().isEmpty();
```
MelvinWM: Should this be?:

    return unit -> !UnitAttachment.get(unit.getType()).getCantTarget().isEmpty();

MelvinWM:
I have only looked at the changes somewhat superficially, but they also look (very) good to me.

game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
ron-murhammer: Nice catch. You are correct. Its fixed now.

game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
ron-murhammer: Yep. Removed.

ron-murhammer:
Thanks for the quick reviews. Updated to fix the 2 issues that were pointed out.

PR #4833, opened 2019-05-09
Refactor MustFightBattle

ron-murhammer:
## Overview
- Do some refactoring and clean up of MustFightBattle as its probably the largest and most complex class in the code base

## Refactoring Changes
- Refactor FireAa to separate class
- Refactor sound logic to separate utility class
- Move some methods to TransportTracker
- Move some methods to BattleCalculator
- Lots of minor clean up and renamings
- Lots of reordering

## Manual Testing Performed
- Ran all unit tests
- Launched a few all AI games

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/Constants.java
+++ b/game-core/src/main/java/games/strategy/triplea/Constants.java
@@ -82,8 +82,7 @@ public interface Constants {
   String ALLIED_AIR_INDEPENDENT = "Allied Air Independent";
   String DEFENDING_SUBS_SNEAK_ATTACK = "Defending Subs Sneak Attack";
   String ATTACKER_RETREAT_PLANES = "Attacker Retreat Planes";
-  String NAVAL_BOMBARD_CASUALTIES_RETURN_FIRE_RESTRICTED =
-      "Naval Bombard Casualties Return Fire";
+  String NAVAL_BOMBARD_CASUALTIES_RETURN_FIRE = "Naval Bombard Casualties Return Fire";
   String SURVIVING_AIR_MOVE_TO_LAND = "Surviving Air Move To Land";
   String BLITZ_THROUGH_FACTORIES_AND_AA_RESTRICTED = "Blitz Through Factories And AA Restricted";
   String AIR_ATTACK_SUB_RESTRICTED = "Air Attack Sub Restricted";
--- a/game-core/src/main/java/games/strategy/triplea/delegate/MustFightBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/MustFightBattle.java
@@ -3,7 +3,6 @@ package games.strategy.triplea.delegate;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -89,7 +88,6 @@ public class MustFightBattle extends DependentBattle implements BattleStepString
 
   private static final long serialVersionUID = 5879502298361231540L;
 
-  // maps Territory-> units (stores a collection of who is attacking from where, needed for undoing moves)
   private final Collection<Unit> attackingWaitingToDie = new ArrayList<>();
   private final Collection<Unit> defendingWaitingToDie = new ArrayList<>();
   // keep track of all the units that die in the battle to show in the history window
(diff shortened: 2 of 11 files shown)


game-core/src/main/java/games/strategy/triplea/Constants.java
```diff
@@ -82,8 +82,7 @@
   String ALLIED_AIR_INDEPENDENT = "Allied Air Independent";
   String DEFENDING_SUBS_SNEAK_ATTACK = "Defending Subs Sneak Attack";
   String ATTACKER_RETREAT_PLANES = "Attacker Retreat Planes";
-  String NAVAL_BOMBARD_CASUALTIES_RETURN_FIRE_RESTRICTED =
-      "Naval Bombard Casualties Return Fire";
+  String NAVAL_BOMBARD_CASUALTIES_RETURN_FIRE = "Naval Bombard Casualties Return Fire";
```
ron-murhammer: This property was inverted a while ago but wasn't renamed in the code.

game-core/src/main/java/games/strategy/triplea/delegate/MustFightBattle.java
```diff
@@ -1911,13 +1867,7 @@ private void fireNavalBombardment(final IDelegateBridge bridge) {
...
       bridge.addChange(change);
     }
-    /*
-     * TODO This code is actually a bug- the property is intended to tell if the return fire is
-     * RESTRICTED- but it's used as if it's ALLOWED. The reason is the default values on the
-     * property definition. However, fixing this will entail a fix to the XML to reverse
-     * all values. We'll leave it as is for now and try to figure out a patch strategy later.
-     */
-    final boolean canReturnFire = isNavalBombardCasualtiesReturnFire();
+    final boolean canReturnFire = Properties.getNavalBombardCasualtiesReturnFire(gameData);
```
ron-murhammer: This was fixed a while ago.

MelvinWM:
I really like refactoring of large source files, especially since it can make maintenance and development easier, both now and later, and large source files can be more than linearly difficult to analyze and change depending on their size (I once worked in a codebase that included a decades-old C++ source file at about 23,000 lines and growing - smart and capable developers had in the past attempted and only been partially successful at seeking to refactor that part of the codebase). I will look at reviewing the changes this weekend or the next if it is OK. The changes are a bit large - I think I would have split them into two or more pull requests, merely to make it easier to review them - though refactoring is typically somewhat relatively easy to review. And I would guess that the changes include refactoring code out into files (with better separation of concerns among other aspects, such as the sound changes, which is a very good idea and very nice), and that kind of refactoring can further be one of the easier kinds of refactorings to review.

ron-murhammer:
@MelvinWM I'd suggest reviewing it commit by commit. A few of the commits are almost entirely just reordering methods to try to group them together into some logical ordering so that they can hopefully then be later extracted. This PR ended up being a bit larger than I intended but given its primarily in a single class and moving things around, its difficult to break it up.

RoiEXLab:
@ron-murhammer Converting IExecutables should be possible to a certain degree.
The only concern I have is regarding serialization: Lambdas don't have a serialVersionUID. However I'm more or less sure this doesn't matter because the classes are anonymous anyways, but we'd have to test it.

Regarding 2. I'd like to point out that I've seen terrible constructions of how different IExecutables push each other onto the stack, but have forgotten where I've seen this. So I'd appreciate if when separating IExecutables from each other the could become less entangled as well.

MelvinWM:
One thing that I like to do when creating commits is to use `git rebase -i` in order to help make the commits nice, typically only with the `fixup` or `squash` changes. While one has to be careful when using `git rebase -i` (for instance `git rebase -i HEAD~4` to rebase from the 5th-latest commit) it can be very nice to use. I also think your `verify` script is (very) nice, and I like to use it before committing (or else doing it after committing and then using `git rebase -i` to fix up any minor changes). Of course, the commits from a pull request being squashed together when merged into the project can help ensure that the commits are nice in various regards, which makes commands like `git rebase -i` somewhat less useful relatively speaking in that case. There is also git pre-commit hooks, though I have not investigated or thought about those a lot. Personally for me, running `verify` manually works very well in my workflow (only very minor comment I have is that it takes some time to run, though that is not an issue for me).

MelvinWM:
The sound-related refactorings seem somewhat independent from the others, so I wonder if they could be refactored separately.

MelvinWM:
Conditions have changed a bit for me, so if someone else wants to review the code instead of me, they are fully welcome.

ron-murhammer:
Merging this as its only refactoring and has been open 3 weeks now. Feel free to add any comments and they will be addressed in follow up PRs.

PR #4841, opened 2019-05-27
Add DB deployment to ansible and support scripts

DanVanAtta:
## Overview

Updates ansible deployment scripts to be more functional and can now
deploy a functioning lobby server with database.

Key changes:
- Adds database installation to ansible. Previously we had to install Postgres
by hand and configure the database user and create the database by hand. Now
ansible will do this.
- Adds flyway deployment to ansible.
- Database user is changed from 'postgres' to a new database user 'triplea-lobby'.
This is done to follow best-practices and not use the postgres DB super-user account
for application access. This update required changes to the flyway migrations.
When deploying to production we will need to do a fresh migration and a table data
copy to account for this. The intent is for the next production deploment will happen on a fresh
set of servers and fresh database for a less risky cut-over.
- Concept of infrastructure server is introduced, this a single server from where we will
run ansible. The setup on this server is one-time. Copies of scripts deployed to this server
are checked in under 'infrastructure-machine' and should be kept in-sync. A notable
TODO is to update travis to execute 'run_deployment' for prerelease on this server after
successful merges.

Notable Changes:
- A secrets file is now deployed to an infrastructure server. The secrets
file is a series of export statements that will set environment variables
from where we can grab password values in ansible scripts.
- Admin user is renamed in ansible scripts from 'triplea' to 'admin'. It was
confusing as who the admin user was, and currently we are using the user 'admin'.
With this change the 'triplea' user on systems would no longer be used. Each app
like lobby and bot have their own system users created already.
- Http server and lobby have been merged, this update reflects that and merges
http_server configuration into lobby.

## Functional Changes
- Ansible can deploy a DB with flyway migrations. Can be used to set up a lobby from scratch
- Flyway migrations are updated, migrations are no longer trivially compatible with existing DB deployments. A correction was also made in the migrations to reflect the new DB name of 'lobby'
- Various new support scripts are checked in, particularly one for use with local vagrant and secondly copies that will be used on the infrastructure server for executing deployments.

## Manual Testing Performed
- Deployed with ansible to a local vagrant
- Deployed from infrastructure server to prerelease.

## Additional Review Notes

This is a decent milestone but there is more to do:
- Prerelease lobby is available, but looks like there is more to do to get the bots to be fully configured.
- prerelease deployment should become automated and tied in with travis
- http server configuration is still to be done.
- production configuration needs prep, new servers are going to be needed for migrating to the next stable release.


DanVanAtta:
Looks like I could use help with the Docker piece that would create a user. I'm not sure where docker is even creating the DB, in fact this has been a persistence problem with it creating a 'ta_users' (incorrect) and I've had to drop and recreate to get a proper 'lobby' database.

Happy to get pointers or a PR into this branch to update docker to use a 'triplea_lobby' user and see if we can perhaps also get the DB name finally updated to 'lobby'.

DanVanAtta:
@ssoloff @RoiEXLab , ^ wondering if either of you can help.

RoiEXLab:
Hmm I have few to nothing experience with docker. 

@ssoloff Probably knows a lot more about this than I do

ssoloff:
Sorry, I'm unable to run a Docker environment at this time and so can't really investigate what changes are required to achieve the desired goal.

infrastructure/ansible/group_vars/production.yml
```diff
@@ -1,7 +1,2 @@
-# TODO: get these values from environment variables
-papertrail_token: ""
-lobby_postgres_user: ""
-lobby_postgres_password: ""
-lobby_postgres_port: ""
-http_server_error_report_github_token: ""
 http_server_error_report_github_repo: "triplea"
+bot_lobby_address: production.triplea-game.org
```
RoiEXLab: This DNS entry doesn't exist (yet?)
Do you plan on adding it?

DanVanAtta: Good point/catch. The production configuration is not yet fully baked, we'll need to spend some time to ensure it is right when we get closer to deployment time. I'm thinking/planning for us to deploy to different servers to help make sure we do not have any kinks/problems such as a non-existent DNS.

With that said, it's a bit of an open question of what the DNS name should be for prod. I'm wondering if we should bake in the lobby version number to the DNS name so we can distinguish between the multiple production versions, or if even that makes sense. We could use 'project' names to refer to the different production versions of TripleA.

RoiEXLab: I mean if we're really crazy we could also use the namecheap api to have the scripts auto-manage dns entries for us:
https://www.namecheap.com/support/api/methods/domains-dns/set-hosts.aspx

This way baking in the number wouldn't be that big of a deal for the system.

For the user that's another problem though

PR #4846, opened 2019-05-28
Split Http Server from Lobby and migrate Spark To DropWizard

DanVanAtta:
## Overview

- Split http server out from the lobby code to be an independent app.

  Lobby no longer launches an http server, the two will be different applications.
  This is done for several reasons, primarily for code separation, separation of
  concerns, and to better be able to debug/troubleshoot in case of for example
  memory leaks.

- Switch http server frameworks from Spark to DropWizard (https://www.dropwizard.io/)

  DropWizard has a few advantages at the cost of some additional complexity.
  Notably with drop wizard we get:
    - built-in way to inject configuration properties
    - automatic JSON serialization/deserialization
  The http server was *not* converted to SpringBoot for a few reasons:
    - we're avoiding the J2EE stack for now, it brings in a lot of complexity,
      perhaps not well suited for a project of volunteers and potentially
      many non-Java experts, let alone J2EE experts
    - avoiding spring stack for same reasons to avoid J2EE stack.

- Rename 'lobby-http-client' sub-project to 'http-clients'

  The subproject contains all http-clients, and not just the http client to the lobby.
  Furthermore, the http server has been split now from the lobby and the lobby
  no longer serves http connections.

- Some future TODOs remain:
  - [ ] Verify packaging and release artifact
  - [ ] See if we can do http client integration testing with a server that has a mocked backend.
  - [ ] See if we can fill any unit testing gaps
  - [ ] Error reporting needs to have a DB table of requests sent by IP and enforce a rate limit.

## Functional Changes
- Lobby no longer launches an http server. The infrastructure was not quite complete to configure the http server, it's one functionality, error report uploading, was not yet functional anyways.

## Manual Testing Performed
- Launched the 'new' http server with environment variable set with github API token, ran a curl request and successful created a github issue: https://github.com/triplea-game/test/issues/56


http-server/configuration-prerelease.yml
```diff
@@ -0,0 +1,2 @@
+githubRepo: test
+githubApiToken: ${GITHUB_ISSUES_API_TOKEN}
```
DanVanAtta: Note this is a syntax that will use environment variables for the configuration property. Any confiugration value that needs environment variable, or is different between prod and prerelease are defined in the YML. Any other configuration values that are constant between prod and prerelease are `static final String` constants in application configuration.

I've been debating if we should have all configuration values in YML and no `static final String` constants. I'm leaning towards the current setup so we have less to manage between the two YML files. Keeping the two files in-sync could become more of a challenge if we put all configuration here, even the configuration that is the same for all environments. On the other hand splitting configuration between YML and Java code is not quite ideal either.

IMO the biggest liability/risk is likely with configuration files that drift and we have a production file that is missing something or does not work, hence I'm leaning to keep the two configs small and restricted to just secrets that need to be env variable injected or for configuration that differs between prod and non-prod.

http-server/src/main/java/org/triplea/server/error/reporting/CreateIssueStrategy.java
```diff
@@ -18,15 +18,15 @@
   @Nonnull
   private final Function<CreateIssueResponse, ErrorUploadResponse> responseAdapter;
   @Nonnull
-  private final GithubIssueClient createIssueClient;
+  private final GithubIssueClient githubIssueClient;
```
DanVanAtta: simple variable rename here.

http-server/src/main/java/org/triplea/server/error/reporting/ErrorReportController.java
```diff
@@ -0,0 +1,52 @@
+package org.triplea.server.error.reporting;
```
DanVanAtta: This is technically not a new file, as we had a controller previously: https://github.com/triplea-game/triplea/pull/4846/files#diff-7e7834725765093021ce16251828b07c

Given drop-wizard will do the JSON serialization and deserialization for us automatically, it is a bit simpler. On the other hand the check for empty fields so we throw an illegal arg adds complexity as does the new way to grab IP address from the request.

Side-note, I'm pretty sure we could use javax validator annotations to do the same check as we do here for empty fields. For now keeping things simple and trying to focus on getting the http server up and running and functional.

http-server/src/main/java/org/triplea/server/http/AppConfig.java
```diff
@@ -0,0 +1,32 @@
+package org.triplea.server.http;
```
DanVanAtta: This is an important class to be aware of. DropWizard serializes the YML configuration file values to this class and then injects this class as a bean to the main application. The YML configuration file is specified as the second argument to the main method that launches drop wizard, see the application class for more details: https://github.com/triplea-game/triplea/pull/4846/files#diff-69b1b6a2651245b3f3d83460de5856fd

RoiEXLab:
@DanVanAtta I would even go as far as claiming that Java EE can be considered legacy. From my (limited) experience Spring is probably considered the "state of the art" framework, whereas Java EE is only used by mostly legacy applications.

But yeah, agreed that Spring isn't really beginner friendly either. I recall the times where this whole concept of Dependency injection was completely weird to me, and beans seemed to be returned by useless methods with a weird annotation.
Currently trying to figure out how to deal with Websockets + Spring, authentication is hard ^^

RoiEXLab:
btw. feel free to merge once master is deployed, it _should_ work now, but I just want to be sure.

DanVanAtta:
FWIW, spring (referring more to SpringBoot) is opinionated and is "convention over configuration". Dropwizard is "configuratoin over convention". As an example, when turning on spring security, it'll give you a full blown and working security/authentication system. The classic complaint is that often that is not what you want, and those parts need to be custom and a simple user authentication out of the box is not correct. It largely depends on how much the convention aligns with your application.

Dropwizard/Jersey on the other hand is a bit more minimal and modules and configuration are added for more features.

With that said, the J2EE stack merges/coincides a lot with various spring facilities. For example JPA is a J2EE animal.

Sometimes primitive and easier to figure out are worth more than fast prototyping and not having boilerplate. JPA can be very nice, but certainly adds to the learning curve, and issues like spring authentication/security further increase difficulty. #preachingToTheChoir

----

Thanks for looking this over @RoiEXLab , really appreciate it!

PR #4849, opened 2019-05-28
Fix IDEA launchers

DanVanAtta:
## Overview
Fixes working path of launchers, they should work again with the very latest IDEA that was released.


DanVanAtta:
Did you update to the latest IDEA @RoiEXLab ?  2019.1.3

Let's revert this if it causes you any issues, these updates seemed to have been required

RoiEXLab:
I just double checked, worked fine on both of my machines.
The same is true for the updated configuration 🤷‍♂

PR #4851, opened 2019-05-29
Deployable, tested HTTP Server & Finish Error Upload Feature

DanVanAtta:
## Overview
- Fix up gradle config for http-server to create a launchable artifact.
- Add an integration test for drop wizard http server, fires up the http server and verifies http client code can connect.
- Add database support to http server (per drop wizard recommendation, using JDBI framework)
- Add rate limiting for error report upload. Except for http server deployment, this completes the error report upload feature. For server side rate limiting we add a simple history table and update it on the server side when error reports are submitted. Of note, whenever we add to the history table we'll also purge any records older than two days to prevent the table from growing. An initial rate limit of 5 error reports per day has been configured.
- Update headed-client to have a 'fail-early' check if upload limit has been reached. If so we'll show a dialog explaining this to the user instead of showing the upload-error-report form. To enable this a new GET endpoint was added: '/can-submit-error-report'
- Set a default value for 'github API token' to be 'test'. This is used as a magic value to trigger a stubbed response when creating github issues. To safeguard against using a stub in production, a 'prod' flag is added which will crash the application if we are configured to use a stub response.

## Functional Changes
- Error report upload now has a server side rate limit, 5 per day
- Client side error report upload will tell the user when they hit the limit if they click "report to triplea" when seeing errors

## Manual Testing Performed
- fired up a local http server and tested out the endpoints with curl. (Note, the new integration tests really helped quite a lot, like most automated testing, that testing was more valuable and replaced any curl testing, but it's not bad to have curl testing nonetheless)
- fired up a local http server, lobby, and added a "generate error button" to the homepage that threw an exception. I clicked this to verify the DB data was as expected and that the rate limit dialog appeared. Also was able to verify the DB pruning manually in this fashion and saw old records drop away from the DB.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -98,7 +98,6 @@ subprojects {
 
         testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:$junitJupiterVersion"
         testRuntimeOnly 'org.junit.platform:junit-platform-launcher:1.4.0'
-        testRuntimeOnly 'org.slf4j:slf4j-nop:1.7.25'
     }
 
     tasks.withType(JavaCompile).configureEach {
--- a/game-core/src/main/java/games/strategy/debug/ErrorMessage.java
+++ b/game-core/src/main/java/games/strategy/debug/ErrorMessage.java
@@ -17,6 +17,7 @@ import org.triplea.http.client.error.report.ErrorUploadClient;
 import org.triplea.swing.JButtonBuilder;
 import org.triplea.swing.JLabelBuilder;
 import org.triplea.swing.JPanelBuilder;
+import org.triplea.swing.SwingComponents;
 
 import com.google.common.base.Preconditions;
 
@@ -128,7 +129,14 @@ public enum ErrorMessage {
       }
       INSTANCE.uploadButton.addActionListener(e -> {
         hide();
-        StackTraceReportView.showWindow(windowReference, serviceClient, record);
+        if (serviceClient.canSubmitErrorReport()) {
+          StackTraceReportView.showWindow(windowReference, serviceClient, record);
+        } else {
+          SwingComponents.showDialog(
+              "Error Report Limit Reached",
+              "You have reached a daily limit for error uploads. "
+                  + "Please wait a day before submitting additional error reports.");
+        }
       });
     }
   }
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/login/LobbyServerProperties.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/login/LobbyServerProperties.java
@@ -71,9 +71,10 @@ public final class LobbyServerProperties {
   public URI getHttpsServerUri() {
     try {
       return new URIBuilder()
+          // TODO: use https
           .setScheme(
               // allow env variable override of https so we can do local development with http
-              Optional.ofNullable(System.getenv("HTTP_SERVER_PROTOCOL")).orElse("https"))
+              Optional.ofNullable(System.getenv("HTTP_SERVER_PROTOCOL")).orElse("http"))
           .setHost(host)
           .setPort(httpsPort)
           .build();
--- a/http-clients/src/main/java/org/triplea/http/client/error/report/ErrorUploadClient.java
+++ b/http-clients/src/main/java/org/triplea/http/client/error/report/ErrorUploadClient.java
@@ -4,8 +4,6 @@ import java.net.URI;
 
 import org.triplea.http.client.HttpClient;
 
-import com.google.common.annotations.VisibleForTesting;
-
 import feign.FeignException;
 import feign.Headers;
 import feign.RequestLine;
@@ -16,8 +14,8 @@ import feign.RequestLine;
 @SuppressWarnings("InterfaceNeverImplemented")
 public interface ErrorUploadClient {
 
-  @VisibleForTesting
   String ERROR_REPORT_PATH = "/error-report";
+  String CAN_REPORT_PATH = "/can-submit-error-report";
 
   /**
    * API to upload an exception error report from a TripleA client to TripleA server.
@@ -30,6 +28,9 @@ public interface ErrorUploadClient {
       "Accept: application/json"
   })
   ErrorUploadResponse uploadErrorReport(ErrorUploadRequest request);
+
+  @RequestLine("GET " + CAN_REPORT_PATH)
+  boolean canSubmitErrorReport();
 
   /**
    * Creates an error report uploader clients, sends error reports and gets a response back.
--- a/http-clients/src/main/java/org/triplea/http/client/github/issues/GithubIssueClient.java
+++ b/http-clients/src/main/java/org/triplea/http/client/github/issues/GithubIssueClient.java
@@ -33,6 +33,13 @@ public class GithubIssueClient {
     this.authToken = authToken;
   }
 
+  /**
+   * For local or integration testing, we may want to have a fake that does not actually
+   * call github. This method returns true if we are doing a fake call to github.
+   */
+  public boolean isTest() {
+    return authToken.equalsIgnoreCase("test");
+  }
 
   /**
    * Invokes github web-API to create a github issue with the provided parameter data.
--- a/http-server/build.gradle
+++ b/http-server/build.gradle
@@ -37,8 +37,7 @@ dependencies {
 
     testImplementation project(':test-common')
     testImplementation 'com.github.database-rider:rider-junit5:1.5.2'
-
-    testCompile group: 'io.dropwizard', name: 'dropwizard-testing', version: '1.3.12'
+    testImplementation 'io.dropwizarddropwizard-testing:1.3.12'
 }
 
 jar {
--- a/http-server/build.gradle
+++ b/http-server/build.gradle
@@ -13,6 +13,8 @@ ext {
 
 configurations {
     testImplementation {
+        // database-rider brings in slf4j-simple as a transitive dependency
+        // DropWizard has logback baked in and cannot have multiple slf4j bindings.
         exclude group: 'org.slf4j', module: 'slf4j-simple'
     }
 }
--- a/http-server/build.gradle
+++ b/http-server/build.gradle
@@ -1,22 +1,42 @@
 plugins {
     id 'application'
     id 'com.github.johnrengelman.shadow' version '4.0.4'
-    id 'de.undercouch.download' version '3.4.3'
 }
 
 archivesBaseName = "$group-$name"
 description = 'TripleA DropWizard Http Server'
-mainClassName = '' // org.triplea.game.client.HeadedGameRunner'
-
+mainClassName = 'org.triplea.server.http.ServerApplication'
 ext {
     releasesDir = file("$buildDir/releases")
 }
 
+
+configurations {
+    testImplementation {
+        exclude group: 'org.slf4j', module: 'slf4j-simple'
+    }
+}
+
 dependencies {
-    implementation 'io.dropwizard:dropwizard-core:1.3.11'
     implementation project(':java-extras')
     implementation project(':http-clients')
+
+    implementation 'io.dropwizard:dropwizard-core:1.3.12'
+    implementation 'io.dropwizard:dropwizard-jdbi3:1.3.12'
+    implementation 'org.jdbi:jdbi3-core:3.8.2'
+    implementation 'org.jdbi:jdbi3-sqlobject:3.8.2'
+
+    implementation 'javax.xml.bind:jaxb-api:2.3.1'
+    implementation 'com.sun.xml.bind:jaxb-impl:2.3.2'
+    implementation 'com.sun.xml.bind:jaxb-core:2.3.0'
+    implementation 'javax.activation:activation:1.1.1'
+
+    runtimeOnly "org.postgresql:postgresql:$postgresqlVersion"
+
     testImplementation project(':test-common')
+    testImplementation 'com.github.database-rider:rider-junit5:1.5.2'
+
+    testCompile group: 'io.dropwizard', name: 'dropwizard-testing', version: '1.3.12'
 }
 
 jar {
@@ -26,6 +46,9 @@ jar {
 }
 
 task portableInstaller(type: Zip, group: 'release', dependsOn: shadowJar) {
+    from file('configuration-prerelease.yml')
+    from file('configuration-production.yml')
+
     from(shadowJar.outputs) {
         into 'bin'
     }
--- a/http-server/build.gradle
+++ b/http-server/build.gradle
@@ -37,7 +37,7 @@ dependencies {
 
     testImplementation project(':test-common')
     testImplementation 'com.github.database-rider:rider-junit5:1.5.2'
-    testCompile group: 'io.dropwizard', name: 'dropwizard-testing', version: '1.3.12'
+    testImplementation 'io.dropwizard:dropwizard-testing:1.3.12'
 }
 
 jar {
--- a/http-server/build.gradle
+++ b/http-server/build.gradle
@@ -37,7 +37,7 @@ dependencies {
 
     testImplementation project(':test-common')
     testImplementation 'com.github.database-rider:rider-junit5:1.5.2'
-    testImplementation 'io.dropwizarddropwizard-testing:1.3.12'
+    testCompile group: 'io.dropwizard', name: 'dropwizard-testing', version: '1.3.12'
 }
 
 jar {
--- a/http-server/configuration-prerelease.yml
+++ b/http-server/configuration-prerelease.yml
@@ -1,5 +1,5 @@
 githubRepo: test
-githubApiToken: ${GITHUB_ISSUES_API_TOKEN:-test}
+githubApiToken: ${GITHUB_API_TOKEN:-test}
 # if we are using any stubbing, we'll assert that this value is false.
 # this is to help guarantee that we do not accidently use test configuration in prod.
 prod: false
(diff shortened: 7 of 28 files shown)


build.gradle
```diff
@@ -98,7 +98,6 @@ subprojects {
 
         testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:$junitJupiterVersion"
         testRuntimeOnly 'org.junit.platform:junit-platform-launcher:1.4.0'
-        testRuntimeOnly 'org.slf4j:slf4j-nop:1.7.25'
```
DanVanAtta: I think we had this for Spark. It was VERY important to remove this as DropWizard fails on startup if there are multiple SLF4J bindings present.  I added a warning note to the http-server README about this.

game-core/src/main/java/games/strategy/debug/ErrorMessage.java
```diff
@@ -128,7 +129,14 @@ private void setUploadRecord(final LogRecord record) {
       }
       INSTANCE.uploadButton.addActionListener(e -> {
         hide();
-        StackTraceReportView.showWindow(windowReference, serviceClient, record);
+        if (serviceClient.canSubmitErrorReport()) {
+          StackTraceReportView.showWindow(windowReference, serviceClient, record);
+        } else {
+          SwingComponents.showDialog(
```
DanVanAtta: This is a user facing change here. THe idea is if a user has hit the rate limit, we do not want them to fill in the optional details dialog to then click submit and get rejected then. Instead we check with the server when user clicks "report to triplea" and if they have hit the limit will show them a message instead of the error report form.

game-core/src/main/java/games/strategy/engine/lobby/client/login/LobbyServerProperties.java
```diff
@@ -71,9 +71,10 @@
   public URI getHttpsServerUri() {
     try {
       return new URIBuilder()
+          // TODO: use https
           .setScheme(
               // allow env variable override of https so we can do local development with http
-              Optional.ofNullable(System.getenv("HTTP_SERVER_PROTOCOL")).orElse("https"))
+              Optional.ofNullable(System.getenv("HTTP_SERVER_PROTOCOL")).orElse("http"))
```
DanVanAtta: I'm not sure if having local development set a special environment variable is the best way for us to override to http. For now, the protocol is set to http to keep things simple and working.

http-clients/src/main/java/org/triplea/http/client/error/report/ErrorUploadClient.java
```diff
@@ -16,8 +14,8 @@
 @SuppressWarnings("InterfaceNeverImplemented")
 public interface ErrorUploadClient {
 
-  @VisibleForTesting
```
DanVanAtta: No long visible for testing, the endpoint path definition re-uses this same constant. This way we guarantee that the endpoint referenced here is the same one we use for the http path.

http-clients/src/main/java/org/triplea/http/client/github/issues/GithubIssueClient.java
```diff
@@ -33,6 +33,13 @@ public GithubIssueClient(
     this.authToken = authToken;
   }
 
+  /**
+   * For local or integration testing, we may want to have a fake that does not actually
+   * call github. This method returns true if we are doing a fake call to github.
+   */
+  public boolean isTest() {
+    return authToken.equalsIgnoreCase("test");
```
DanVanAtta: Here is where we detect a magic value, kinda dirty, but it does allow for us to have an out-of-the-box configuration that is working and can be used for local test. The production application configuration does not default the environment variable to 'test', and we have an 'prod' check that will doubly make sure we do not use a magic value that triggers stubbing prod.

http-server/build.gradle
```diff
@@ -1,22 +1,42 @@
...
-    implementation 'io.dropwizard:dropwizard-core:1.3.11'
     implementation project(':java-extras')
     implementation project(':http-clients')
+
+    implementation 'io.dropwizard:dropwizard-core:1.3.12'
+    implementation 'io.dropwizard:dropwizard-jdbi3:1.3.12'
+    implementation 'org.jdbi:jdbi3-core:3.8.2'
+    implementation 'org.jdbi:jdbi3-sqlobject:3.8.2'
+
+    implementation 'javax.xml.bind:jaxb-api:2.3.1'
```
DanVanAtta: there was a nasty startup warning displayed when drop wizard started. It comes down to jaxb being packaged with JDK8 and before, but no longer with JDK8+. The warnings did not necessarily stop the server from starting, regardless having these dependencies fixes the warnings and is one less thing to worry about.

http-server/build.gradle
```diff
@@ -26,6 +46,9 @@ jar {
 }
 
 task portableInstaller(type: Zip, group: 'release', dependsOn: shadowJar) {
+    from file('configuration-prerelease.yml')
+    from file('configuration-production.yml')
```
DanVanAtta: This bundles the configuration with the http-server zip. When extracting we'll need to drop these in the "working directory" from where we run the dropwizard fat jar. Which configuration to use is a main method arg.

EG:
```
java -jar bin/http-server-VERSION.jar server configuration-production.yml
```

http-server/configuration-prerelease.yml
```diff
@@ -1,2 +1,34 @@
 githubRepo: test
-githubApiToken: ${GITHUB_ISSUES_API_TOKEN}
+githubApiToken: ${GITHUB_API_TOKEN:-test}
```
DanVanAtta: The ':-test' syntax defaults a missing env variable value to 'test' here. We do not have that defaulting active in the prod configuration.

Otherwise the updates here are mostly a copy/paste of me googling what it took to configure database in drop wizard. The values for the most part looked reasonable.

Of note we still bake in the fact/notion that database must be running on the same machine, I believe we have this assumption for lobby and it's a bit of a long standing TODO to allow for lobby, http-server and DB to be running on different machines. For now we intend a single-machine deployment, so it's okay..

http-server/src/main/java/org/triplea/server/error/reporting/CreateIssueStrategy.java
```diff
@@ -9,27 +11,56 @@
...
+  private final ErrorReportingDao errorReportingDao;
+
 
   @Override
   public ErrorUploadResponse apply(final ErrorReportRequest errorReportRequest) {
-    if (allowErrorReport.test(errorReportRequest)) {
-      final CreateIssueResponse response =
-          githubIssueClient.newIssue(errorReportRequest.getErrorReport());
-      return responseAdapter.apply(response);
+    if (allowErrorReport.test(errorReportRequest.getClientIp())) {
```
DanVanAtta: This code is completing the error upload feature, it's the server side throttle that limits how many error uploads can be submitted (per IP address).

http-server/src/main/java/org/triplea/server/error/reporting/ErrorReportController.java
```diff
@@ -49,4 +55,16 @@ private String extractClientIp(final HttpServletRequest request) {
...
     return request.getRemoteHost();
   }
+
+  /**
+   * Checks if the user has hit their rate limit for submitting error reports.
+   *
+   * @return True if the user can submit an error report, false if they
+   *         have hit their limit.
+   */
+  @GET
```
DanVanAtta: This endpoint is used to before we show the upload dialog. The idea is it's bad to let the user do work, like fill in optional error report details to click submit and then get an error. Instead of that we'll show a "you are rate limited dialog" instead of showing the upload error report form.

http-server/src/main/java/org/triplea/server/error/reporting/ErrorReportingDao.java
```diff
@@ -0,0 +1,46 @@
...
+import java.time.Instant;
+
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+
+/**
+ * DAO class for error reporting functionality.
+ */
+public interface ErrorReportingDao {
```
DanVanAtta: This class uses JDBI, SQL Object syntax. 
Overall it seems pretty nice!

@ron-murhammer FYI, you may be interested to see this example. I was tempted to resurrect the conversation to use MyBatis and go for that. I was reading through drop wizard docs on how to configure DB and they strongly recommended to use JDBI (Sql Object). Specifically we're using JDBI3 here.

DanVanAtta: Of note also, database-rider is agnostic to what framework we use, it was pretty straight forward to test this code using DB rider. The DAO pattern also made it easy to mock the DB interaction and test the classes that used this DAO.

http-server/src/test/java/org/triplea/server/error/reporting/CreateIssueStrategyTest.java
```diff
@@ -0,0 +1,106 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.triplea.http.client.error.report.ErrorUploadRequest;
+import org.triplea.http.client.error.report.ErrorUploadResponse;
+import org.triplea.http.client.github.issues.GithubIssueClient;
+import org.triplea.http.client.github.issues.create.CreateIssueResponse;
+
+@ExtendWith(MockitoExtension.class)
+class CreateIssueStrategyTest {
```
DanVanAtta: The createIssueStrategy is not new code, but looked to have been missing tests. With the further addition of rate limiting logic, it was well past time to unit test it.

DanVanAtta:
Hmm, failure is:
```
      Detail: Key (id)=(1) already exists. [statement:"insert into error_report_history(user_ip) values(:ip)", rewritten:"insert into error_report_history(user_ip) values(:ip)", parsed:"ParsedSql{sql='insert into error_report_history(user_ip) values(?)', parameters=ParsedParameters{positional=false, parameterNames=[ip]}}", arguments:{positional:{0:second}, named:{ip:second}, finder:[]}]
```
Do not see this locally, will need to look into it. Will have some time hopefully tomorrow. I'm surprised by this and wondering if maybe a synthetic PK is perhaps better not used at all on the error_report_history. I'm also wondering/surprised that we hit this error at all.

The fix to this should be pretty small, I'd appreciate any review to try and cover the bulk of the other changes. I'm also happy to get any recommendations/suggestions about this error, if there is a better way to do an auto-incrementing sequence for a synthetic PK, or perhaps better to drop the synthetic PK column altogether.

DanVanAtta:
I think I figured out the problem. There is an explicit insert into the new table, so the sequence number would collide with that. I did not see that in testing as the sequence number for me had advanced past '1' when I first wrote the db-rider test.

http-clients/src/main/java/org/triplea/http/client/github/issues/GithubIssueClient.java
DanVanAtta: The isTest method is later used to detect if we should use stub behavior, or if we should use github webservices API to create an issue.

http-server/build.gradle
```diff
@@ -1,22 +1,43 @@
 plugins {
     id 'application'
     id 'com.github.johnrengelman.shadow' version '4.0.4'
-    id 'de.undercouch.download' version '3.4.3'
```
DanVanAtta: This was left in from previous commit, not needed in this sub-project AFAIK.

game-core/src/main/java/games/strategy/debug/ErrorMessage.java
RoiEXLab: This will probably break once a user (or 2 users with the same public IP) is getting 2 errors at once in 2 different windows when they're just about to hit the limit.

Not sure if it's worth dealing with this rare to unlikely case.

http-server/build.gradle
```diff
@@ -1,22 +1,43 @@
...
 ext {
     releasesDir = file("$buildDir/releases")
 }
 
+
+configurations {
+    testImplementation {
+        // database-rider brings in slf4j-simple as a transitive dependency
+        // DropWizard has logback baked in and cannot have multiple slf4j bindings.
+        exclude group: 'org.slf4j', module: 'slf4j-simple'
```
RoiEXLab: It's always great having to deal with conflicting dependencies ^^

http-server/build.gradle
```diff
@@ -1,22 +1,43 @@
...
+    implementation 'javax.xml.bind:jaxb-api:2.3.1'
+    implementation 'com.sun.xml.bind:jaxb-impl:2.3.2'
+    implementation 'com.sun.xml.bind:jaxb-core:2.3.0'
+    implementation 'javax.activation:activation:1.1.1'
+
+    runtimeOnly "org.postgresql:postgresql:$postgresqlVersion"
+
     testImplementation project(':test-common')
+    testImplementation 'com.github.database-rider:rider-junit5:1.5.2'
+    testCompile group: 'io.dropwizard', name: 'dropwizard-testing', version: '1.3.12'
```
RoiEXLab: For consistency I'd prefer the 'usual syntax'
```suggestion
    testCompile group: 'io.dropwizard:dropwizard-testing:1.3.12'
```

http-server/build.gradle
RoiEXLab: Is this documented somewhere?

I feel like this is an easy thing to miss.

http-server/src/main/java/org/triplea/server/error/reporting/CreateIssueStrategy.java
```diff
@@ -9,27 +11,56 @@
...
+      final ErrorUploadResponse errorUploadResponse = sendRequest(errorReportRequest);
+
+      errorReportingDao.insertHistoryRecord(errorReportRequest.getClientIp());
+      errorReportingDao.purgeOld(Clock.systemUTC().instant().minus(2, ChronoUnit.DAYS));
```
RoiEXLab: I'm wondering:
Is there any difference between `Clock.systemUTC().instant()` and `Instant.now()`?

http-server/src/main/java/org/triplea/server/error/reporting/ErrorReportController.java
```diff
@@ -1,42 +1,48 @@
...
 
-  static final String ERROR_REPORT_PATH = "/error-report";
-
+  @Nonnull
   private final Function<ErrorReportRequest, ErrorUploadResponse> errorReportIngestion;
+  @Nonnull
+  private final Predicate<String> errorReportRateChecker;
 
   @POST
+  @Path(ErrorUploadClient.ERROR_REPORT_PATH)
```
RoiEXLab: What's the difference between using a `@<REQUEST_TYPE>` + `@Path` annotation vs `@RequestLine`?
The latter one was used earlier in this diff.

http-server/src/main/java/org/triplea/server/error/reporting/ErrorReportingDao.java
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Inserts a new record indicating a user has submitted an error report
+   * at a given date.
+   *
+   * @param userIp Ip host address of the user submitting an error report.
+   */
+  @SqlUpdate("insert into error_report_history(user_ip) values(:ip)")
+  void insertHistoryRecord(@Bind("ip") String userIp);
```
RoiEXLab: I really like this approach to variable binding, it's super readable

http-server/build.gradle
DanVanAtta: Oh yeah, the no-op SLF4J logger was a nice icing as in some circumstances there was no warning about multiple SLF4J binding. #HeadSeeKeyboard :face_with_head_bandage:

http-server/build.gradle
DanVanAtta: Not sure what the problem I ran into there originally, regardless I was able to use the shorthand syntax in the end without issue, updated in: 624229878

http-server/build.gradle
DanVanAtta: It will be in the ansible deployment scripts. For dev/prerelease running with no-args will default to the right values.

http-server/src/main/java/org/triplea/server/error/reporting/ErrorReportController.java
DanVanAtta: Http client vs Server. RequestLine is Feign,  `@Path` is server side (drop-wizard):
- import feign.RequestLine;
- import javax.ws.rs.Path;

game-core/src/main/java/games/strategy/debug/ErrorMessage.java
DanVanAtta: Good consideration. It does seem like it should be pretty rare, I'd be a bit surprised if many people even hit the limit at all. We could increase the limit to 10 to see how it goes. On the plus side, as this is server side code, we can re-deploy with an updated limit pretty readily (so the risk of getting it wrong now is not that great).

DanVanAtta: TLDR; IMO we can see how it goes for now, I'd be surprised if we were to see that problem.

http-server/src/main/java/org/triplea/server/error/reporting/CreateIssueStrategy.java
DanVanAtta: Instant is essentially a representation of an epoch milli timestamp relative to zulu time. So no, there should not be any difference. I updated this to be the slightly simpler `Instant.now()` in: 3ce69994e

http-server/src/main/java/org/triplea/server/error/reporting/ErrorReportController.java
RoiEXLab: Ah, my bad. Thanks for the clarification

PR #4855, opened 2019-05-31
Rename ta_users table to 'lobby_user' and add moderator audit table

DanVanAtta:
## Overview

Change summary:
- Rename db table 'ta_users' to 'lobby_user'
- Rename column 'ta_users.lastLogin' to 'lobby_user.last_login' to conform with SQL column naming convention of lower_snake_case.
- Add an integer PK column to 'lobby_user'.
- Add a moderator audit history table to record moderator actions.
- Insert a record into the new moderator audit history table on mac bans or username bans

Motivation for changes:
- The 'ta_users' table name does not conform to convention of singular name and follows an old and needless naming pattern where all tables were prefixed 'ta'
- The new PK column added to 'lobby_user' lets us start adding new tables with numeric foreign keys. A first example is the moderator audit history.
- The moderator audit history will eventually be viewable from the moderator toolbox. It will also simplify life for moderators as they will no longer need to report all actions they take to the forum as there will be an audit trail. A secondary benefit is we will have a more reliable audit trail and perhaps increased accountability.

## Functional Changes
- No user facing changes
- New audit trail of moderator actions recorded in database

## Manual Testing Performed
- none, automated testing added

### DIFF
--- a/lobby-db/src/main/resources/db/migration/V1.10.00.20190330__lobby_user_table.sql
+++ b/lobby-db/src/main/resources/db/migration/V1.10.00.20190330__lobby_user_table.sql
@@ -32,6 +32,8 @@ insert into lobby_user (username, password, email, date_created, last_login, adm
 select username, password, email, joined, lastLogin, admin
 from ta_users;
 
+drop table ta_users;
+
 create table moderator_action_history
 (
     id            serial primary key,
--- a/lobby-db/src/main/resources/db/migration/V1.10.00.20190330__lobby_user_table.sql
+++ b/lobby-db/src/main/resources/db/migration/V1.10.00.20190330__lobby_user_table.sql
@@ -1,0 +1,54 @@
+create table lobby_user
+(
+    id              serial primary key,
+    username        character varying(40) not null unique,
+    password        character varying(60),
+    email           character varying(40) not null,
+    date_created    timestamptz           not null default current_timestamp,
+    last_login      timestamptz,
+    admin           boolean               not null default false,
+    bcrypt_password character(60) check (char_length(bcrypt_password) = 60)
+);
+
+-- alter table user owner to triplea_lobby;
+alter table lobby_user
+    owner to postgres;
+alter table lobby_user
+    add constraint lobby_user_pass_check check (password IS NOT NULL OR bcrypt_password IS NOT NULL);
+
+
+comment on table lobby_user is 'The table storing all the information about Lobby TripleA users.';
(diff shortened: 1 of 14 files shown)


lobby-db/src/main/resources/db/migration/V1.10.00.20190330__lobby_user_table.sql
```diff
@@ -0,0 +1,54 @@
...
+comment on column lobby_user.email is 'Email storage of every user. Large size to match the maximum email length. More information here: https://stackoverflow.com/a/574698.';
+comment on column lobby_user.date_created is 'The timestamp of the creation of the account.';
+comment on column lobby_user.last_login is 'The timestamp of the last successful login.';
+comment on column lobby_user.admin is 'The role of the user, controls privileges. If true the user is able to ban and mute other people.';
+comment on column lobby_user.bcrypt_password is 'The BCrypt-Hashed password of the user, should be the same as the md5 password but in another form. The length of the hash must always be 60 chars. Either password or bcrypt_password must be not null.';
+
+
+insert into lobby_user (username, password, email, date_created, last_login, admin)
+select username, password, email, joined, lastLogin, admin
+from ta_users;
```
RoiEXLab: Shouldn't the "old" table get deleted after this?

lobby/src/main/java/org/triplea/lobby/server/db/UserController.java
```diff
@@ -35,7 +35,8 @@ public HashedPassword getPassword(final String username) {
   private HashedPassword getPassword(final String username, final boolean legacy) {
     try (Connection con = connection.get();
         PreparedStatement ps = con
-            .prepareStatement("select password, coalesce(bcrypt_password, password) from ta_users where username=?")) {
+            .prepareStatement(
+                "select password, coalesce(bcrypt_password, password) from lobby_user where username=?")) {
```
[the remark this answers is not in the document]
DanVanAtta: I made it a bit more consistent: d3d160b8b
It does not bother me too much, with existing code it's probably better in a different PR to avoid adding too many different kinds of changes to this PR. Regardless, it's a bit more consistent now

lobby-db/src/main/resources/db/migration/V1.10.00.20190330__lobby_user_table.sql
DanVanAtta: I initially forgot, good catch! I was then waffling a bit yesterday after wondering about rollback. I'd agree best to drop it so the rename is complete: 2139de19f

Issue #3162, opened 2018-02-26
Transport loading bug (rules)


panther2:
This issue has been brought up on 
https://www.axisandallies.org/forums/index.php?topic=41580.0

### Engine version:
Tested with 1.9.0.0.9347,  also present in 7621 (so most likely introduced earlier)
### Map name and version:
Tested with WWII_Global_1940 and WWII_v5
### Can you describe how to trigger the error? (eg: what sequence of actions will recreate it?)
A transport that has been part of a battle incorrectly loads units during Noncombat Move Phase
### Instead of this error, what should have happened?
Loading units should be prohibited.

Rulebooks say (Noncombat Move Phase):

> Transports can move to friendly coastal territories and load or
> offload cargo, **unless** they loaded, moved, offloaded, or were
> involved in combat during the Combat Move or Conduct
> Combat phase.

### Any additional information that may help:
test3.tsvg illustrates the issue for Global
test5.tsvg illustrates the issue for v5

[savedGames.zip](https://github.com/triplea-game/triplea/files/1757622/savedGames.zip)

ron-murhammer:
@panther2 Any idea if this has ever been prohibited in any previous version? Just wondering if I need to see if something has changed in the past couple of months or if this was something that never was enforced.

DanVanAtta:
> Any idea if this has ever been prohibited in any previous version?

I suspect this. @panther2 are there not also ruleset differences, or was it determined this was the right rule for all versions?

panther2:
@redrum 
Having tested a bit I was able to reproduce this bug even with TripleA 1.8.0.9.
So I assume that this core rule has never been coded in.

@DanVanAtta 
Revised OOB (v2) was the last ruleset that allowed loading transports after combat.
Starting from and including Revised LHTR (v2 LHTR) this has changed!

ron-murhammer:
@panther2 This should be fixed with https://github.com/triplea-game/triplea/pull/4834

panther2:
@ron-murhammer I hope to find some time for testing soon.

Cernelius:
Is this fix supposed to cover v2 LHTR too?

panther2:
@ron-murhammer I have finally tested these scenarios using 1.10.14782 and can confirm, that they work correctly in wwII_Global and wwII_v5.

However in v2-LHTR the transport still incorrectly loads, despite having been part of the battle:

[...]

PR #4857, opened 2019-06-04
Don't allow transports to load after being in battle if LHTR

ron-murhammer:
## Overview
- Address LHTR portion of #3162

## Functional Changes
- If property "LHTR Carrier production rules" is true then don't allow transports to load after being in a battle

## Manual Testing Performed
- Tested revised LHTR


ron-murhammer:
Merging as minor fix.

PR #4859, opened 2019-06-05
Extract http accept and content type headers to constant

DanVanAtta:
## Overview
Simple refactor to extract constants set on HTTP clients for header values to a constants class.

## Functional Changes
- none

## Additional Review Notes
Note: simple text type will be used in future updates and is not yet used.

### DIFF
--- a/http-clients/src/main/java/org/triplea/http/client/HttpConstants.java
+++ b/http-clients/src/main/java/org/triplea/http/client/HttpConstants.java
@@ -1,0 +1,15 @@
+package org.triplea.http.client;
+
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+
+/**
+ * Utility class with constants for HTTP clients.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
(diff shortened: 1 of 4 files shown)


http-clients/src/main/java/org/triplea/http/client/HttpConstants.java
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Utility class with constants for HTTP clients.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class HttpConstants {
+  public static final String ACCEPT_JSON = "Accept: application/json";
+  public static final String ACCEPT_TEXT = "Accept: text/plain";
+  public static final String CONTENT_TYPE_JSON = "Content-Type: application/json";
+  public static final String CONTENT_TYPE_TEXT = "Content-Type: text/plain";
+}
```
RoiEXLab: I'm surprised that there isn't a built-in way to do this.
The apache HTTPComponents library has constants for the most headers and common values

PR #4872, opened 2019-06-06
Moderator Toolbox 2.0 Initial Add

DanVanAtta:
## Overview

- Adds support for moderators to enter their key, stored in client settings, and validated on http requests
- Adds a basic toolbox window UI with a 'bad-words' tab and 'events' tab. The first tab allows for viewing/adding/removing entries, the events tab is a paged view of the events table.

Note on API key rate limiting:
- there is extensive support to ensure validation and rate limiting of invalid keys. This is both out of concern for brute-force cracking attacks and DDOS. In essence we keep an in-memory cache of failed attempts and will go into a 'lock-out' mode if there are too many failed attempts. During lock-out any API keys presented are rejected with exception to a white-list of valid API keys. Valid API keys are cached for 24 hours.

## Functional Changes
- Adds a new menu item to launch the moderator toolbox to the moderator menu (will be behind beta feature flag).

## Manual Testing Performed
- The toolbox window still has a main method for testing to launch the toolbox window.
- docs/dev/README.md has some new notes with SQL commands that will load a moderator user with password == "password" and api-key == "password", the SQL commands also load a number of history events so that paging can be demonstrated.

### Invalid key when launching toolbox window
- API key is pulled from client settings and is verified when the  toolbox window is launched. If valid we proceed to the toolbox window, if not valid or not present then we are presented with the 'enter-api-key' window below. A valid key is entered then it will be stored in client settings for future usage.

Screenshot from 2019-06-06 02-03-58

### Wait spinner when validating a key (typically seen only on invalid keys)
- The background task runner can't be cancelled, no plan to fix that for this branch. Instead will focus on not retrying invalid keys.
Screenshot from 2019-06-06 02-04-12

### Bad Words Tab
- The toolbox will gain more tabs, for now this is a pretty simple tab that shows the contents of the `bad-word` table and provides controls to add/remove entries
Screenshot from 2019-06-06 02-05-24

### Event log Tab
- This is a read-only view showing the contents of the moderator audit history table. The results loaded into this tab are paged.
Screenshot from 2019-06-06 02-05-56


RoiEXLab:
Will have a look at this tomorrow. Thanks for investing all the time into this, this is a huge step forward 👍

RoiEXLab:
Also a test fails

DanVanAtta:
Yeah, sadly the test only fails on Travis, the output from server is not there and forcing errors on my local system is crashing my entire OS... I don't think any updates for the test are going to need very much, but am working on trying to get more debug info.

DanVanAtta:
Wow, that took a lot of debugging!

Problem was with postgres DB version. For the query with paging that was using 'offset', we need postgres version 10.

I had to make a number of changes/modifications though to get here:
- added code to .travis.yml to launch dropwizard from CLI, this meant creating a shadow jar and then launching via `java -jar`. Overall I think this is a more robust way to launch the server and is probably a better integration test in the end. In the end the net effect is the server output finally would land on the travis CI output, otherwise it was completely suppressed.
- added a try/catch in the server-client integration test to ignore startup failures, the idea being that the server may already be started. In general I also think I like this as it means we can launch the server out-of-band and use the integration test to verify.
- switched travis postgres version to '10' and had to switch distro to xenial to support this. I think this just means on the 1.10 release we'll need to be sure to have a latest postgres, which is not too inconvenient as we'll be launching a new DB and migrating DB data anyways
- Xenial does not have support for Java 8, had to drop it from the .travis config. I believe there is really well beyond critical mass to always bundle a java with the triplea distro, I think this might be a good final reason for us to commit to upgrading to java 11.

docs/dev/README.md
```diff
@@ -38,3 +38,86 @@ The following project-specific Docker images, which may be useful during develop
...
   - [Lobby database](https://github.com/triplea-game/triplea/tree/master/lobby-db/Dockerfile)
 
+
+
+## DB Data for Local Testing
+
+Connect to localhost database:
+```bash
+psql -h localhost -U postgres
+\c lobby
```
RoiEXLab: What about
```bash
psql -h localhost -U postgres lobby
```
as a single command?

game-core/src/main/java/games/strategy/engine/lobby/moderator/toolbox/ToolBoxWindow.java
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * This window shows a series of tabs that provide CRUD operations to a moderator. Each tab roughly maps
+ * to a DB table.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class ToolBoxWindow {
+
+  // TODO: remove this main method, just here for demo purposes.
+  public static void main(final String[] args) {
```
RoiEXLab: Is this still being used?

game-core/src/main/java/games/strategy/engine/lobby/moderator/toolbox/tabs/EventLogTabModel.java
```diff
@@ -0,0 +1,72 @@
...
+import lombok.RequiredArgsConstructor;
+
+
+/**
+ * Model to interact with the backend for fetching moderator audit history records. The data is paged
+ * so we can avoid fetching too much data all at once. This model keeps some state, specifically it
+ * keeps track of the last fetched.
+ */
+@RequiredArgsConstructor
+class EventLogTabModel {
```
RoiEXLab: Review stop

game-core/src/main/java/games/strategy/engine/lobby/moderator/toolbox/ToolBoxWindow.java
DanVanAtta: In the short-term, yes. It'll probably be removed when some of the last tabs are added.

I'm thinking before moderator toolbox can be considered 'done', we'll want:
- moderators tab for super-moderator to be able to issue API keys
- thinking we may want an API keys tab, not 100% sure yet
- a users tab for moderators to do bans
- a bans tab to view bans and remove bans

So there is still work to be done and the temp main method should remain useful for a little bit yet.

DanVanAtta:
@RoiEXLab Thanks for tackling, at 4k lines, I think this is the largest PR submitted to TripleA

docs/dev/README.md
DanVanAtta: Sure: f0408aae6

.travis.yml
```diff
@@ -1,14 +1,12 @@
-dist: trusty
+dist: xenial
```
DanVanAtta: postgres 10 is not available on trusty

.travis.yml
```diff
@@ -1,14 +1,12 @@
...
 matrix:
   include:
-  - jdk: oraclejdk8
-    env: TRIPLEA_RELEASE=true
   - jdk: openjdk11
-    env: TRIPLEA_RELEASE=false
+    env: TRIPLEA_RELEASE=true
 addons:
-  postgresql: "9.5"
+  postgresql: "10"
```
DanVanAtta: postgres 10 needed for "offset" (paging) query: https://github.com/triplea-game/triplea/pull/4872/files#diff-0886cbe4926f9b7699f3155e549d7d5eR86

game-core/src/main/java/games/strategy/engine/lobby/moderator/toolbox/tabs/ModeratorsTab.java
```diff
@@ -0,0 +1,55 @@
+package games.strategy.engine.lobby.moderator.toolbox.tabs;
```
DanVanAtta: Do not review this class, highly WIP

game-core/src/main/java/games/strategy/engine/lobby/moderator/toolbox/tabs/ModeratorsTabModel.java
```diff
@@ -0,0 +1,22 @@
+package games.strategy.engine.lobby.moderator.toolbox.tabs;
```
DanVanAtta: Do not review, WIP

http-clients/src/main/java/org/triplea/http/client/HttpClient.java
```diff
@@ -86,12 +85,12 @@ private static Exception errorDecoder(final String methodKey, final Response res
         .map(body -> {
           try (BufferedReader reader = new BufferedReader(body.asReader())) {
             final String errorMessageBody = reader.lines().collect(Collectors.joining("\n"));
-            return new HttpCommunicationException(response.status(), firstLine + "\n" + errorMessageBody);
+            return new HttpInteractionException(response.status(), firstLine + "\n" + errorMessageBody);
```
DanVanAtta: HttpCommunicationException is renamed to HttpInteractionException, we throw the exception in more cases than just communications errors.

http-clients/src/main/java/org/triplea/http/client/HttpClient.java
```diff
@@ -36,7 +35,7 @@
 @AllArgsConstructor
 public class HttpClient<ClientTypeT> implements Supplier<ClientTypeT> {
 
-  private static final Decoder gsonDecoder = new GsonDecoder();
+  private static final Decoder gsonDecoder = JsonDecoder.gsonDecoder();
```
DanVanAtta: A custom JSON decoder was needed (fun) for Instant deserialization. The server encodes instant into an epoch second plus epoch second, which seems to not be natively supported.

http-clients/src/main/java/org/triplea/http/client/HttpInteractionException.java
```diff
@@ -0,0 +1,14 @@
+package org.triplea.http.client;
```
DanVanAtta: This is renamed/moved, compare to HttpCommunicationException above.

http-clients/src/main/java/org/triplea/http/client/JsonEncoder.java
```diff
@@ -15,6 +15,7 @@
 public class JsonEncoder implements Encoder {
   private static final GsonEncoder gsonEncoder = new GsonEncoder();
 
+  @SuppressWarnings("deprecation")
```
DanVanAtta: The deprecated method was introduced in a recent PR. It's not clear what the alternative usage would be, perhaps I have not looked into it enough. Regardless, the encode logic is inspired/copy+pasted from the default feign encoder (thank you open source!) which uses a deprecated method. It seems to work perfectly fine for now, hence I'm not terribly concerned yet about the deprecated API usage and am just suppressing it for now.

http-clients/src/main/java/org/triplea/http/client/moderator/toolbox/ModeratorToolboxFeignClient.java
```diff
@@ -0,0 +1,53 @@
...
+ * This set of http methods should all require an API key passed via headers to authenticate
+ * the moderator issuing the request.
+ * </p>
+ * <p>
+ * Rate-limiting: of note, the backend implementation should be careful to apply rate limiting
+ * to any/all endpoints that take an API key so as to avoid brute-force attacks to try and crack
+ * an API key value.
+ * </p>
+ */
+interface ModeratorToolboxFeignClient {
```
DanVanAtta: Of some note, this class is package-protected and provides a "Raw" API. To avoid any user of this interface from knowing how to create headers, we have a public "ModeratorToolboxClient" which takes care of those details. So for example instead of providing a map parameter, we can request a String parameter to pass the api key that is then passed to this class as a proper map.

http-server/configuration-production.yml
```diff
@@ -3,6 +3,8 @@ githubApiToken: ${GITHUB_API_TOKEN}
 # if we are using any stubbing, we'll assert that this value is false.
 # this is to help guarantee that we do not accidently use test configuration in prod.
 prod: true
+logRequestAndResponses: false
+logSqlStatements: false
```
DanVanAtta: I'm a bit on the fence if SQL statements should be logged in prod. The parameters are not logged, so it would be 'safe', but I think in the end it would become very noisy.

http-server/src/main/java/org/triplea/server/error/reporting/CreateIssueStrategy.java
```diff
@@ -10,6 +10,7 @@
 import org.triplea.http.client.error.report.ErrorUploadResponse;
 import org.triplea.http.client.github.issues.GithubIssueClient;
 import org.triplea.http.client.github.issues.create.CreateIssueResponse;
+import org.triplea.lobby.server.db.ErrorReportingDao;
```
DanVanAtta: ErrorReportingDao was moved to the `lobby-db-dao` subproject, hence `http-server` subproject needs to import the dao

http-server/src/main/java/org/triplea/server/http/IllegalArgumentMapper.java
```diff
@@ -8,7 +8,7 @@
  * by http endpoint controllers to return HTTP status 400 codes.
  * Without this, those errors would be 500s.
  */
-class ClientExceptionMapper implements ExceptionMapper<IllegalArgumentException> {
+class IllegalArgumentMapper implements ExceptionMapper<IllegalArgumentException> {
```
DanVanAtta: Just a rename to be more specific.

http-server/src/main/java/org/triplea/server/http/IpAddressExtractor.java
```diff
@@ -0,0 +1,30 @@
...
+import javax.servlet.http.HttpServletRequest;
+
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+
+/**
+ * Utility class to extract the remote host machine IP address from an http servlet request.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class IpAddressExtractor {
```
DanVanAtta: This functionality is now used in multiple places and was extracted.

http-server/src/main/java/org/triplea/server/http/ServerApplication.java
```diff
@@ -59,50 +62,50 @@ public void initialize(final Bootstrap<AppConfig> bootstrap) {
...
-    // register all endpoint handlers here:
-    environment.jersey().register(errorReportController(configuration, jdbi));
+    exceptionMappers()
+        .forEach(mapper -> environment.jersey().register(mapper));
+
+    endPointControllers(configuration, environment)
+        .forEach(controller -> environment.jersey().register(controller));
   }
 
-  private static ErrorReportController errorReportController(
```
DanVanAtta: This code is moved to ErrorReportControllerFactory. It follows a pattern where there is a "*Factory" class that creates each controller, hence we do not have to construct each controller in the server application class.

http-server/src/main/java/org/triplea/server/moderator/toolbox/api/key/validation/ApiKeyLookup.java
```diff
@@ -0,0 +1,29 @@
+package org.triplea.server.moderator.toolbox.api.key.validation;
```
DanVanAtta: api/key/validation package has a number of helper (package-protected) classes to implement the overall functionality. To see the high level flow, see the service class: https://github.com/triplea-game/triplea/pull/4872/files#diff-8a0be339eef1a70416907b4c3ef2268e

http-server/src/test/java/org/triplea/server/error/reporting/Database.java
```diff
@@ -1,20 +0,0 @@
-package org.triplea.server.error.reporting;
```
DanVanAtta: This class is moved to lobby-db-dao and is renamed to JdbiDatabase

http-server/src/test/java/org/triplea/server/error/reporting/ErrorReportingDaoTest.java
```diff
@@ -1,93 +0,0 @@
-package org.triplea.server.error.reporting;
```
DanVanAtta: DAOs are moved to lobby-db-dao

http-server/src/test/java/org/triplea/server/http/ClientServerIntegrationTest.java
```diff
@@ -0,0 +1,145 @@
+package org.triplea.server.http;
```
DanVanAtta: This was existing, but renamed. This class is an integration test that sets up a server if not already running and exercises all http clients. I wanted to break it up per client, but the constant starting/stopping was slow and not reliable.

http-server/src/test/java/org/triplea/server/http/HttpServerIntegrationTest.java
```diff
@@ -1,57 +0,0 @@
-package org.triplea.server.http;
```
DanVanAtta: This was renamed, see class above.

http-server/src/test/resources/datasets/error_reporting/post-insert.yml
```diff
@@ -1,3 +0,0 @@
-error_report_history:
```
DanVanAtta: These test data files are moved to lobby-db-dao

http-server/src/test/resources/datasets/integration.yml
```diff
@@ -0,0 +1,13 @@
+error_report_history:
```
DanVanAtta: This is the dataset for the client-server integration test. Notably needed so we have a specific API key and can better exercise endpoints by passing a valid key.

lobby/src/main/java/org/triplea/lobby/server/EnvironmentVariable.java
```diff
@@ -7,17 +7,7 @@
 /** Class that represent OS environment variable keys with default values. */
 @AllArgsConstructor
 public enum EnvironmentVariable {
-  PORT("3304"),
```
DanVanAtta: DB environment specific variables are useful for http server, so there is a need to share it now. 
These values were essentially moved, see:  https://github.com/triplea-game/triplea/pull/4872/files#diff-2c4b408f0009d57a5b1dc98e642bd13a

swing-lib/src/main/java/org/triplea/swing/JTableBuilder.java
```diff
@@ -76,6 +76,12 @@ public static void addRows(final JTable table, final List<List<String>> rows) {
     rows.forEach(row -> model.addRow(row.toArray(new String[0])));
   }
 
+  public static void setData(final JTable table, final List<List<String>> rows) {
```
DanVanAtta: This is needed for when we click 'refresh' on the events tab. That is a case where we want to overwrite existing JTable data.

DanVanAtta:
@ron-murhammer @ssoloff drive-by reviews welcome, this PR is likely too large to be really reviewed and best-effort and iterating I think is all we can do

http-clients/src/main/java/org/triplea/http/client/JsonEncoder.java
```diff
@@ -15,6 +15,7 @@
...
   @Override
   public void encode(final Object object, final Type bodyType, final RequestTemplate template) {
     if (bodyType.getTypeName().equals(String.class.getName())) {
```
RoiEXLab: Just looked up the source:
https://github.com/OpenFeign/feign/blob/23ee09ef88a02ead7c8c925cdab7b094053b3d9e/core/src/main/java/feign/RequestTemplate.java#L754
It seems like creating a `Request.Body` instance and using that is the preffered way now.

Also the default code uses reference equality i.e. `bodyType == String.class` https://github.com/OpenFeign/feign/blob/23ee09ef88a02ead7c8c925cdab7b094053b3d9e/core/src/main/java/feign/codec/Encoder.java#L88

not sure if you changed that to meet a errorprone rule or something.

http-clients/src/test/java/org/triplea/http/client/moderator/toolbox/ModeratorToolboxClientTest.java
```diff
@@ -0,0 +1,158 @@
...
+  private static final int ROW_COUNT = 5;
+  private static final int ROW_START = 10;
+
+  private static final String API_KEY = "api-key";
+  private static final Map<String, Object> expectedHeader;
+
+  static {
+    expectedHeader = new HashMap<>();
+    expectedHeader.put(ModeratorToolboxClient.MODERATOR_API_KEY_HEADER, API_KEY);
+  }
```
RoiEXLab: Is there something speaking against `Collections#singletonMap`?

http-clients/src/test/java/org/triplea/http/client/moderator/toolbox/ModeratorToolboxFeignClientTest.java
```diff
@@ -0,0 +1,202 @@
...
+import com.github.tomakehurst.wiremock.client.WireMock;
+
+import lombok.Builder;
+import ru.lanwen.wiremock.ext.WiremockResolver;
+import ru.lanwen.wiremock.ext.WiremockUriResolver;
+
+@ExtendWith({
+    WiremockResolver.class,
+    WiremockUriResolver.class
+})
```
RoiEXLab: This is probably a design preference, but for me
```java
@ExtendWith(WiremockResolver.class)
@ExtendWith(WiremockUriResolver.class)
```
seems more logical?

Regardless I'm not sure if this is done in a consistent manner, so that might be a goal.

http-clients/src/test/java/org/triplea/http/client/moderator/toolbox/ModeratorToolboxFeignClientTest.java
```diff
@@ -0,0 +1,202 @@
...
+
+  private static final Map<String, Object> headerMap;
+
+  private static final String HEADER_KEY = "key";
+  private static final String HEADER_VALUE = "value";
+
+  static {
+    headerMap = new HashMap<>();
+    headerMap.put(HEADER_KEY, HEADER_VALUE);
+  }
```
RoiEXLab: Same here with `Collections#singletonMap`

http-server/README.md
```diff
@@ -71,3 +71,29 @@ wizard may fail to start with this error:
...
+
+## Stream is already closed
+
+This can happen when the server side does not fail fast on incorrect input.
+This can be if we use headers that are missing or do not check that parameters are present.
+
+The stack trace indicating this will look like this:
+```bash
+ERROR [2019-06-06 05:07:22,247] org.glassfish.jersey.server.ServerRuntime$Responder: An I/O error has occurred while writing a response message entity to the container output stream.
+! java.lang.IllegalStateException: The output stream has already been closed.
```
RoiEXLab: Just curious: Would this issue occur when shutting down?
Like when the stream gets closed but there are still processes running that are ready to write, whenever there's a response?

http-server/README.md
```diff
@@ -71,3 +71,29 @@ wizard may fail to start with this error:
...
+
+This is bad as it could be used in a DDOS attack.
+
+### Prevention
+
+Essentially fail-fast:
+- When looking for headers, verify headers exist or terminate the request
+- Verify that all needed GET parameters are present or terminate the request
+
+To terminate the request, just throw a IllegalArgumentException, it'l be mapped to a 400.
```
RoiEXLab: This reminds me of a terrible habit in a project a have been working on in the past, where essentially everything (even NullPointerException/internal or external Server errors) got mapped to 400.
So I think getting a little bit more specific with the error code will not hurt. There are so many status codes to choose from for client errors, my favourite still being [418](https://httpstatuses.com/418).
For example [422](https://httpstatuses.com/422) is much more specific and indicates a semantic error whereas [400](https://httpstatuses.com/400) indicates a syntax error.

http-server/configuration-production.yml
RoiEXLab: I believe logging makes sense when using something like Hibernate where we don't control the SQL that gets sent to the database.
But that's currently not the case, right?

http-server/src/main/java/org/triplea/server/http/IpAddressExtractor.java
RoiEXLab: Good ol' `X-Forwarded-For` header ^^
Just saying that this would in theory allow to "modify" the users origin address, we might want to check both

http-server/src/main/java/org/triplea/server/moderator/toolbox/api/key/validation/ApiKeyValidationServiceFactory.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class ApiKeyValidationServiceFactory {
+
+  static {
+    InvalidKeyCache.setCache(
+        CacheBuilder.newBuilder()
+            .expireAfterWrite(AppConfig.FAILED_API_KEY_CACHE_EXPIRATION, TimeUnit.MINUTES)
+            .build());
+  }
```
RoiEXLab: Why is the Cache injected this way?
I'm concerned that this might cause weird issues when InvalidKeyCache is loaded on its own, without this class.

http-server/src/main/java/org/triplea/server/moderator/toolbox/api/key/validation/InvalidKeyCache.java
```diff
@@ -0,0 +1,48 @@
+package org.triplea.server.moderator.toolbox.api.key.validation;
```
RoiEXLab: Review stop 2

http-clients/src/test/java/org/triplea/http/client/moderator/toolbox/ModeratorToolboxClientTest.java
DanVanAtta: Not initially, but it looks like 'API_KEY_PASSWORD' is going to be a second element soon.

http-clients/src/test/java/org/triplea/http/client/moderator/toolbox/ModeratorToolboxFeignClientTest.java
DanVanAtta: I forget that Java now allows the same annotation multiple times. IMO this is a good cleanup to do post-merge, the annotation for now is consistent with other client tests.

http-clients/src/test/java/org/triplea/http/client/moderator/toolbox/ModeratorToolboxFeignClientTest.java
DanVanAtta: Ditto response, API key password is coming along soon.

http-server/README.md
DanVanAtta: >  (even NullPointerException/internal or external Server errors) got mapped to 400.

Ouch!

There should be some code that throws IllegalArg if an argument is missing. For the most part the syntax is not checked beyond arguments being present. Hence 422 is not appropriate. I would be relunctant to create our `ServerPrecondition.checkArg` method to throw a specific RuntimeException just for the purpose of mapping that to a 400.

One does need to be careful though about IllegalArgumentException. I've seen it abused where that exception is thrown due to programmer error at low levels in the service call stack. So essentially 2 or 3 layers deep, some service dependency fails, returns no data, that is then passed to another layer which throws illegal arg. That scenario should be a 500 as repeating the same exact request could eventually return a 200.

I did check around on 400s, does not seem that there are too many that are even more specific.  For now I think 400 and IllegalArg handling is okay, it was a good consideration but let's not also over-think it too much.

DanVanAtta:
Not sure why github is not allowing me to respond to all conversations.. Regardless:

> I believe logging makes sense when using something like Hibernate where we don't control the SQL that gets sent to the database.
But that's currently not the case, right?


Once I was able to see server output in travis, the error was a syntax error. To dig into that, seeing the SQL statements was useful.

I'm not a fan of "copy/paste this line of code to one-off debug". IMO that falls into tribal knowledge and is just a lack of flagging/automation.

In prerelease seeing SQL statements along with input/output JSON is pretty verbose but has been useful and I think might continue to be just for sanity checking and to be really sure things are working as they should. I'm hoping that is not an excuse for lack of info logging (hopefully the info logging that exists is appropriate and adequate).


To answer directly, we do pretty much control the SQL that is sent to DB with the only exception that the bind parameters seem to be converted to positional parameters by JDBI (eg: :userName, is converted to $1). Hence an error like "syntax error near $2" was what I saw in the error message when debugging postgres version issues.

I think at the end of the day here that more information in pre-prod is better than less.


> Good ol' X-Forwarded-For header ^^
Just saying that this would in theory allow to "modify" the users origin address, we might want to check both

Good point, though is "origin address" also a header? If so that can be also modified. I'm not sure if there is a really good way to detect IP address.. :(

That is perhaps the biggest reason for a "global fail rate" lockout mechanism as modifying IP address could be done in a number of ways. My understanding is also that remote host address in the proxy situation is going to be not the actual remote host but an intermediary device. The two values mismatching does not mean the actual remote host is a better value.

http-clients/src/test/java/org/triplea/http/client/moderator/toolbox/ModeratorToolboxFeignClientTest.java
DanVanAtta: Hmm, trying out that cleanup:
```java
@ExtendWith(WiremockResolver.class)
@ExtendWith(WiremockUriResolver.class)
```
I'm actually not sure I like it. When duplicating the `@ExtendWith`, the two are not guaranteed to be together and they should/must be. I think unrelated `@ExtendWith` are probably better to be done as additional annotations.

Initially my preference seems to be for something like:

```java
@ExtendWith({
    WiremockResolver.class,
    WiremockUriResolver.class
})
@ExtendWith(MockitoExtension.class)
```

rather than:
```java
@ExtendWith(WiremockResolver.class)
@ExtendWith(WiremockUriResolver.class)
@ExtendWith(MockitoExtension.class)
```

I think we're really just getting down to personal preference here, probably best we revisit this in a follow-up PR if someone is really inclined to change it and/or add a note to the coding styling guide.

http-server/src/main/java/org/triplea/server/moderator/toolbox/api/key/validation/ApiKeyValidationServiceFactory.java
DanVanAtta: Out of puritanism, I was avoiding the 'service' from knowing about the `AppConfig` class, so that all configuration is injected. I don't think that really matters when it comes to a static block like this.

Looking at this, I can see how cohesion and/or localization of code suffers, that we're getting an "effect-at-distance". You're also right that we get an odd initialization dependency.

With those considerations I moved the static block to the cache class proper: 6f9ad994c, and also marked the setter as visible for testing.

http-server/README.md
DanVanAtta: The exception happens while processing the request, at which point everything hangs, the request thread on the client as well as the server side processing thread. There is no response sent, and hence why it's a potential DDOS vector.

The hung thread does not terminate properly on shutdown, and so shutdown hangs and requires a force-kill.

It seems like the stream is closed early due to an exception but the server tries to write a response anyways and then just fails..

RoiEXLab:
@DanVanAtta 
> Good point, though is "origin address" also a header? If so that can be also modified. I'm not sure if there is a really good way to detect IP address.. :(

No, `X-Forwarded-For` is a header used by transparent proxies to tell a server on whose behalf it's acting, `HttpServletRequest#getRemoteHost` should return the IP address being used by the underlying TCP Socket, so no way to fake that.

DanVanAtta:
Do I understand right that `#getRemoteHost` would return the IP address of a transparent proxy (if one is being used)? Perhaps that is just as well and we should always use that IP address value.

RoiEXLab:
@DanVanAtta Yes, this method should return the ip address of the device the server is communicating with, vpn endpoint, proxy or the user itself.
The X-Forwarded-For header is just a commonly used custom header that gets added by some proxies to tell servers they are just a proxy and acting on behalf of another device. Of course this isn't a guarantee that the device you're talking to is actually a proxy, but many "find my ip" services are checking the header as well.

RoiEXLab:
That's why I proposed to just check both, and if one of them is blacklisted, just refuse the request.

DanVanAtta:
Sounds like a good suggestion. I'll add that to my backlog, let's do a follow-up so that can get proper review/dedication.

DanVanAtta:
Re: IP, just thinking about it more:
- in my experimentation it might not always be set, will need to investigate further how to reliably get the IP from the underlying socket connection (perhaps there is another method to do it)
- using x-forwarded might not be good, could be an attack vector. If an attacker were able to obtain the IP address of a moderator, one could potentially spoof the x-forwarded and lock that IP address out. Perhaps being overly paranoid, but it's perhaps reason to only use the IP address, whether it is a proxy or not.

DanVanAtta:
@RoiEXLab Thank you for the review passes! I'd like to suggest we go ahead and merge this sooner, perhaps after one more review pass (if review is not yet done by then). I'd be happy to continue with updates based on comments to a merged PR.  The next round of updates are getting close to ready and I will also run out of "inter-vacation" time in three or four days and will be back to traveling.

RoiEXLab:
@DanVanAtta I'll do the rest of the review tomorrow during the day (national holiday, so I'll have plenty of time).
I hope that's sufficient time for you

DanVanAtta:
Sounds good, thanks! The next set of changes are already queued up  =  )

lobby-db-dao/src/main/java/org/triplea/lobby/server/db/ModeratorAuditHistoryDao.java
```diff
@@ -60,4 +61,32 @@ void insertAuditRecord(
...
+  @SqlQuery("select\n"
+      + "  h.date_created " + LookupHistoryItemsColumns.DATE_CREATED + ",\n"
+      + "  u.username " + LookupHistoryItemsColumns.USER_NAME + ",\n"
+      + "  h.action_name " + LookupHistoryItemsColumns.ACTION_NAME + ",\n"
+      + "  h.action_target " + LookupHistoryItemsColumns.ACTION_TARGET + "\n"
+      + "from moderator_action_history h \n"
+      + "join lobby_user u on u.id = h.lobby_user_id\n"
+      + "order by h.date_created desc\n"
+      + "offset :rowOffset rows\n"
+      + "fetch next :rowCount rows only")
```
RoiEXLab: Java needs template strings ^^

lobby-db-dao/src/main/java/org/triplea/lobby/server/db/ModeratorAuditHistoryItem.java
```diff
@@ -0,0 +1,45 @@
...
+  private String actionName;
+  private String actionTarget;
+
+  /**
+   * Returns a JDBI row mapper used to convert results into an instance of this bean object.
+   */
+  static RowMapper<ModeratorAuditHistoryItem> moderatorAuditHistoryItemMapper() {
+    return (rs, ctx) -> {
+      final Calendar cal = Calendar.getInstance();
+      cal.setTimeZone(TimeZone.getTimeZone("UTC"));
```
RoiEXLab: `java.util` time classes are evil 😛 
Is there anything speaking against just omitting the timezone and converting the Timestamp straight to an Instant?
The latter isn't even supposed to contain any Timezone information anyways

DanVanAtta:
@RoiEXLab 

> where new UI can be written using JavaFX exclusively

IMO we need to establish patterns where we use non-FXML code to generate the UI with JavaFx. XML-like code does not scale well to screens that have lots of conditional logic or for cases where very similar components are created in many places.

http-server/src/main/java/org/triplea/server/moderator/toolbox/api/key/validation/InvalidKeyCache.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  @Setter(onMethod_ = {@VisibleForTesting})
+  private static Cache<String, AtomicInteger> cache;
+
+  InvalidKeyCache() {
+    Preconditions.checkNotNull(cache);
+  }
+
+  synchronized void increment(final HttpServletRequest request) {
```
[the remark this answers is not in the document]
DanVanAtta: I was not trying to get the synchronization to be perfect, but instead good-enough for what we need. Not sure if that is a great reason for atomic int or even the sync.

I'm a bit hesitant to move the sync block to where we actually need it as it then wraps database calls. Blocking threads for that long could be worse than even correctly implementing locking.

The deciding factor is that it's probably better to have the locking where it's actually needed, the synchronization at this low of level is still broken (I find this kind of broken sync is a common pattern in TripleA, while we might access the variable/value in an atomic way, multi-threading at a higher level is still going to yield an incorrect answer)

6bada7fd2

I also remove the AtomicInteger, I suppose it's not worthwhile/necessary.

lobby-db-dao/src/main/java/org/triplea/lobby/server/db/ModeratorAuditHistoryItem.java
DanVanAtta: While Instant does not contain timezones, the source data for converting to an instant does matter.

Showing UTC timestamps IMO is a good thing, I don't think we should necessarily rely on DB implementation details to know that the timezone will be a UTC (I'm not sure if java.sql.ResultSet will guarantee that).

My thoughts here are:
- while java.util classes are evil, this code is 'safe' and is pretty guaranteed to give us a consistent time stamp with proper time zone. If that means 2 lines of ugly java.util code, I don't really care as long as it is more certain to be correct.

Some proper experimentation would be useful to really know the answer, but TBH i"m not willing to put that time in if the cost is a few lines of ugly java.util time, I'll trade certainly correct over the prettiest code and save time on this.

DanVanAtta:
@RoiEXLab thanks again for the time to do a review. Updates pushed. The timezone issue I'd like to leave for another day, coding around uncertainty is far cheaper here for me than digging through the docs and considering every case we could have (one case, I'm not sure if we should rely on the DB returning to us a UTC timezone).

DanVanAtta:
This checkstyle rule is taking too much time from failed builds:
```
[ant:checkstyle] [WARN] /home/travis/build/triplea-game/triplea/http-server/src/main/java/org/triplea/server/moderator/toolbox/api/key/validation/ApiKeyValidationService.java:50: <p> tag should be placed immediately before the first word, with no space after. [JavadocParagraph]
```

@RoiEXLab please do not wait for the build to be completed before jumping back in to verify the latest updates. Getting a 4k PR merged is very respectable, at the same time I don't think we appreciate the rate of progress we need to make compared to how much time we have available and how much needs to be done. I've maybe another 2 or 3 hours today, then my time for TripleA will drop close to zero while traveling for the next 2 months. After that, I assume my professional coding will take up almost all my time. I suspect the other devs are in similar situations. I think this is a recurring theme, it's important we keep the big picture in mind.

RoiEXLab:
@DanVanAtta I'll have a final look on the recent commits in a couple of hours (this evening), after that we should be ready to merge this after going with java 11

DanVanAtta:
AFAIK java 11 is independent of toolbox, I don't understand the relationship

RoiEXLab:
@DanVanAtta 
> Xenial does not have support for Java 8, had to drop it from the .travis config. I believe there is really well beyond critical mass to always bundle a java with the triplea distro, I think this might be a good final reason for us to commit to upgrading to java 11.

Ah I thought because of this that you were holding off with this PR until the Java 11 PR is merged. 🤔

DanVanAtta:
Oh, I see the dependency now, good recall, thanks!
Merging does more-or-less put us beyond a point of no return for Java8. The alternative, to support an old version of postgres for the sake of travis build on java8 is really ugly. It's not wise to do that IMO as we will be running a more advanced version of postgres in prod, running different versions means our testing will less reliable.

I think it's just as well to push forward, it seems like we will be moving forward with Java11.

DanVanAtta:
Thanks @RoiEXLab. Policy is to treat PRs as FIFO, that comes from a time when the queue was at 20+ large and important changes were simply put off in favor of reviewing smaller nitpick updates.  In those case the larger changes kept getting conflicts by updates that occurred afterwards and there was no predictability in what would come along and conflict.   I think the record for merge conflict resolution was 6 rounds.

A line-by-line review I don't think is good, if there are any class names that are not great, it's worth calling out, but we are free to continue iterating and it's not like a merge is going to be immutable and immediately land in production.

DanVanAtta:
*Caveat, line-by-line is appropriate to some level, but a review that is focused on nitpicks and not structure and maintenability IMO is missing the point.

PR #4877, opened 2019-06-07
Fix datasets to remove data from tables with FKs

DanVanAtta:
## Overview

One downside of database-rider datasets is that tables with FKs are not automatically cleared.
Any new table with a FK needs to be added to any existing dataset where the primary key exists.
To explain that, if we have table A, and then a new table B with FK to A, then any existing dataset
that uses table A will delete it. If we have run a test that populates table B, then the deletion of
table A will cause a FK constraint violation. This introduces a non-determinism where tests can
fail simply depending on the order in which they are run.

This update fixes this whooopsie and deletes tables that have FKs into lobby_user
(notably the moderator api key table).

Second, integration.yml is simply not used and is being dropped.

### DIFF
--- a/lobby-db-dao/src/test/resources/datasets/bad_words/integration.yml
+++ b/lobby-db-dao/src/test/resources/datasets/bad_words/integration.yml
@@ -1,4 +1,0 @@
-lobby_user:
-  - username: "moderator"
-    email: "email@email.com"
-    password: "1234567890"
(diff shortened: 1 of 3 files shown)


DanVanAtta:
Ah! This is actually not a problem. My DB data had the newest table, was never cleared out, and then switching to 'master' a branch without that table of course invoked this issue. This was a problem only with my local setup (grant it, better local-dev automation is perhaps called for. In the meantime going to move right along :running_man: )

PR #4878, opened 2019-06-07
Fix IDEA launcher module name on lobby runner

DanVanAtta:


RoiEXLab:
I just want to point out that the coverage report clearly shows that the decoder class has flaky coverage, probably indicating that it isn't well-tested

PR #4879, opened 2019-06-07
Fix some mixed mutability error prones

DanVanAtta:
## Overview
Fix a number of error prone warnings for mixed mutability on return types:
```
warning: [MixedMutabilityReturnType] This method returns both mutable and immutable collections or maps from different paths. This may be confusing for users of the method.
```

It looks like a number of these are from returning Collections.emptyList() and also returning array list. The former is immutable and the latter mutable.

## Functional Changes
- none

## Manual Testing Performed
- launched and connected to a local lobby
- started an all Hard-AI game on WWII revised and let the AIs duke it out for a few rounds, no errors.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/Route.java
+++ b/game-core/src/main/java/games/strategy/engine/data/Route.java
@@ -20,6 +20,8 @@ import javax.annotation.Nullable;
 import org.triplea.java.collections.CollectionUtils;
 import org.triplea.util.Tuple;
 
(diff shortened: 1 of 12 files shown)


RoiEXLab:
Seems like tests are saving the day

DanVanAtta:
Oddly, I thought the manual testing here was pretty thorough. Agree automated testing saved the day and IMO speaks to the difficulty of thorough coverage (and false-security) when  manual testing.

PR #4880, opened 2019-06-07
Reduce DBUser usage in APIs, move 'MakeAdmin' to test-only

DanVanAtta:
## Overview
- DBUser is used in a number of places in lobby code and not just in DAO.
To help move DAO to the 'lobby-db-dao' subproject, the DBUser parameter
is removed from DAO APIs. This begins to decouples the DAOs from DBUser
and eventually would allow the DAOs to be moved to the dedicated sub-project.

- Second, 'makeAdmin' is a test-only method, the method is moved to test code
so that production code can no longer 'see' it.

## Functional Changes
None, refactor

## Manual Testing Performed
None

### DIFF
--- a/lobby/src/main/java/org/triplea/lobby/server/UserManager.java
+++ b/lobby/src/main/java/org/triplea/lobby/server/UserManager.java
@@ -38,7 +38,9 @@ final class UserManager implements IUserManager {
     final HashedPassword password = new HashedPassword(hashedPassword);
 
     try {
-      database.getUserDao().updateUser(user,
+      database.getUserDao().updateUser(
+          user.getName(),
+          user.getEmail(),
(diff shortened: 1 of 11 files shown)


RoiEXLab:
It would be nice if this kind of functionality could soon find a way into an actual feature.
Currently @prastle has to manually alter the database to promote or demote someone

DanVanAtta:
Agree, FWIW I'm planning to incorporate that in moderator toolbox.

There have been a couple of big gaps that have always been there in terms of DB access, like the bad words table. Historically, the first rev of TripleA was launched after the lobby was set up, but the DB part was new to @veqryn (this was a really long time ago) and was one of the last pieces built. It was also pretty clear that he gained employment around then and that is where we see start to see some missing/unfinished features

PR #4885, opened 2019-06-10
Remove javadoc paragraph checkstyle.

DanVanAtta:
## Overview

It does not matter if `<p>` is same line or next line. This rule conflicts with the checked-in formatter and IDEA will mangle this. Checkstyle should not be a verification that hand-formatting was done but instead that automatic formatting is done.

Had two failed builds today due to this, that's 10 minutes lost and context switch time. For something that is so pedantic and not automated, the ROI of this rule is  negative.

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -231,7 +231,6 @@
         <module name="SummaryJavadoc">
             <property name="forbiddenSummaryFragments" value="^@return the *|^This method returns |^A [{]@code [a-zA-Z0-9]+[}]( is a )"/>
         </module>
-        <module name="JavadocParagraph"/>
         <module name="AtclauseOrder">
             <property name="tagOrder" value="@param, @return, @throws, @deprecated"/>
             <property name="target" value="CLASS_DEF, INTERFACE_DEF, ENUM_DEF, METHOD_DEF, CTOR_DEF, VARIABLE_DEF"/>


ssoloff:
I believe Spotless enforces an equivalent `<p>` tag formatting rule.  Therefore, the build may still fail (via Spotless instead of Checkstyle) if the same Javadoc formatting is introduced.

DanVanAtta:
I ran spotless apply on a recent failure, it did not fix the checkstyle violation : (

DanVanAtta:
Oh, i misinterpreted that @ssoloff ; that is good to know and will keep an eye out for that scenario.

DanVanAtta:
Responding too fast here.. spotless apply passing but still having a checkstyle violation would indicate that the rule did not kick in for the scenarios where this has been a problem, FWIW.

ssoloff:
I'm pretty sure `spotlessApply` always passes because it corrects the violations instead of failing when one is encountered; you'll need to run `spotlessCheck` to mimic the build.

I don't have a local environment to work with, but I managed to use the GitHub web interface to demonstrate this.  I introduced the same format violation that was fixed in [a2de371d5](https://github.com/triplea-game/triplea/pull/4872/commits/a2de371d50771e8f83ff7b471a965a2e332c073f).  The full Travis build log is [here](https://travis-ci.org/ssoloff/triplea-game-triplea/builds/543934117).  Here's the relevant bit:

```
> Task :http-server:spotlessJava FAILED
* What went wrong:
Execution failed for task ':http-server:spotlessJava'.
> The following files had format violations:
      http-server/src/main/java/org/triplea/server/moderator/toolbox/api/key/validation/ApiKeyValidationService.java
          @@ -47,7 +47,8 @@
           ···*·the·moderator·that·owns·that·key.·If·no·key·is·found,·or·if·verification·is·locked·down,·or·no·key
           ···*·is·present,·then·an·exception·is·thrown.
           ···*
          -···*·<p>Synchronized·so·that·the·check·for·locked·out·request·and·then·incrementing·the·failed
          +···*·<p>
          +···*·Synchronized·so·that·the·check·for·locked·out·request·and·then·incrementing·the·failed
           ···*·request·count·is·atomic.
           ···*
           ···*·@param·request·Server·request·object·expected·to·contain·an·API·key·header.
  Run 'gradlew spotlessApply' to fix these violations.
```

DanVanAtta:
IF one runs spotlessApply, in theory spotlessCheck should then pass. My point was I had run spotlessApply and still got a checkstyle violation anyways.

Issue #4886, opened 2019-06-10
GPG key checksum failing on xenial


DanVanAtta:
Builds are failing with:
```
gpg: key 5B474E6A/5B474E6A: error sending to agent: Operation cancelled
gpg: Total number processed: 2
gpg:               imported: 1
gpg:       secret keys read: 2
The command "./.travis/setup_gpg" failed and exited with 2 during .
```

(1) This is perhaps a difference between xenial and the previous ubuntu distro that was configured on travis

(2) It looks like the magic env variable is not being picked up by this version of PGP

(3) Looking at setup_gpg, it skips if the right environment variable is not set. That means branch builds do not execute this step and hence we can be surprised after merge with failures.

(4) reproducing/troubleshooting this is a PITA and takes a lot of time with travis builds needing to be on master

-------------

Short-term disabling the checksum seems to at least let us move forward. 

For reference, artifact signing was discussed/introduced in: https://github.com/triplea-game/triplea/issues/2466

It looks like this might be the way to go to fix the problem: https://stackoverflow.com/questions/45188811/how-to-gpg-sign-a-file-that-is-built-by-travis-ci

I'm not sure how we fix the issue so that branch builds can test the key signing.

To another extent, I'm not sure if we *really* should be fixing this or not. 
- TripleA has a small user base, we are well under 10k users, this is not the worlds biggest application
- Adware injection is unlikely, TripleA was hosted on sourceforge for a long time, was not a problem then
- It's not clear there has been a single user to use the keys to verify what they download
- It won't be clear if you get a hacked version of TripleA even if you download one. Nothing tells you or directs you to check, if you get a bogus copy, you'll probably be none the wiser and nothing points you in the direction of checking.

Beyond disabling gpg in the short term, long term my 2 cents is it's great if its easy and works. Now that it's not, I don't see the ROI being favorable on this .

RoiEXLab:
I don't have any strong opinion on this.
@ssoloff is probably the person to discuss this with as he was the one introducing it.

ssoloff:
I came across something, quite by accident, and it reminded me of this issue.

[...]

PR #4887, opened 2019-06-10
Temporary disable GPG artifact signing

DanVanAtta:
It is broken going to a new distro of travis ubuntu. This update disable GPG signing until it is fixed and/or decided if we really want to continue supporting the feature. Discussion in: https://github.com/triplea-game/triplea/issues/4886


DanVanAtta:
This is a temporary update, merging to restore the build until we figure out how to move forward or can fix gpg

PR #4888, opened 2019-06-10
Support for moderator single-use-key

DanVanAtta:
## Overview
Adds support and concept for a single-use-key that moderators can use to register a
password and be granted a permanent API key that is stored on their system.

This allows for keys to be distributed and then a new one to be regenerated that
is not known to anyone other than the moderator.

Change Summary:
- incorporate logic so that a key password is registered and hashed with the main moderator key
- update UI to prompt for single-use-key and password if no API key is stored in client settings
- Update toolbox flow to always at least ask for api key password
- add DB table for single-use-key
- update backend to do validation of api key salted with password

## Functional Changes
See change summary in overview

## Manual Testing Performed
- use the SQL posted in readme
- launch the main 'show toolbox controller'
- click 're-register' to clear out any stored API keys in client settings
- use api key value 'password' to register a new API key, password is of your choosing, I typically use 1234
- following registration enter 1234 in api key password prompt
- should then see the moderator toolbox

* server application needs to be running
* flyway needs to be run to get the latest DB code

To reset single use-key without re-running flyway, execute DB command:
```
update moderator_single_use_key set date_used = null;
```

One note, given the hashing and security of the API key, debugging is a bit of a PITA if things do not work right.

### After
Screenshot from 2019-06-10 14-31-28
Screenshot from 2019-06-10 14-31-37

## Additional Review Notes
- once we get the last of moderator toolbox merged up I intend to spend some time doing some polishing changes. Happy to have anything/everything get called out, though some updates will likely conflict with in-flight changes and I may opt to follow-up. 
- More changes on the way! There is another branch inflight that adds more tabs.
- Reasonable number of interesting logics introduced, notably to do key hashing.
- DB changes are listed at the end of the diff, it might be easiest to see those first and then jump back to the top.


game-core/src/main/java/games/strategy/engine/lobby/moderator/toolbox/tabs/BadWordsTabModel.java
```diff
@@ -4,9 +4,8 @@
 
 import javax.annotation.Nonnull;
 
-import org.triplea.http.client.moderator.toolbox.AddBadWordArgs;
 import org.triplea.http.client.moderator.toolbox.ModeratorToolboxClient;
-import org.triplea.http.client.moderator.toolbox.RemoveBadWordArgs;
+import org.triplea.http.client.moderator.toolbox.UpdateBadWordsArg;
```
DanVanAtta: RemoveBadWordArgs and AddBadWordArgs were essentially the same. Instead of needing to ever update both it was renamed to 'UpdateBadWordsArgs' and combined.

http-clients/src/main/java/org/triplea/http/client/moderator/toolbox/ModeratorToolboxClient.java
```diff
@@ -10,13 +11,17 @@
...
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
 
 /**
  * Wrapper around moderator toolbox feign client. This wrapper handles exceptions and returns
  * a success string result for most methods, otherwise the return value is any error message
  * returned back from the server.
  */
+@AllArgsConstructor(access = AccessLevel.PACKAGE)
 public class ModeratorToolboxClient {
```
DanVanAtta: Updates here at to account for having a password now. This class is constructed with a password that is then used with any of the http endpoints. This class also knows how to pull the API key from client settings, with those two values any users of this class need not worry about key and password.

http-server/src/main/java/org/triplea/server/http/ServerApplication.java
```diff
@@ -80,18 +81,17 @@ public void run(final AppConfig configuration, final Environment environment) {
         .forEach(controller -> environment.jersey().register(controller));
   }
 
-
   private List<Object> exceptionMappers() {
     return ImmutableList.of(
         new IllegalArgumentMapper(),
         new IncorrectApiKeyMapper(),
-        new ApiKeyVerificationLockOutMapper());
+        new ApiKeyLockOutMapper());
```
DanVanAtta: Rename here as lockout is not just on verification but can also be during registration

http-server/src/main/java/org/triplea/server/moderator/toolbox/api/key/KeyHasher.java
```diff
@@ -0,0 +1,27 @@
...
+import com.google.common.hash.Hashing;
+
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+
+/**
+ * Utility class to provide hashing functions for moderator API key.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class KeyHasher {
```
DanVanAtta: This logic used to live in a controller factory. It was extracted here and a the key+password and key only variants were created.

http-server/src/main/java/org/triplea/server/moderator/toolbox/api/key/registration/ApiKeyRegistrationService.java
```diff
@@ -0,0 +1,64 @@
...
+
+  private final BiFunction<String, String, String> keyHasher;
+
+  private final Supplier<String> newApiKeySupplier;
+
+  private final InvalidKeyLockOut invalidKeyLockOut;
+
+  private final ApiKeyRegistrationDao apiKeyRegistrationDao;
+
+  String registerKey(
```
DanVanAtta: Important logic flow here.

http-server/src/main/java/org/triplea/server/moderator/toolbox/api/key/validation/ApiKeyValidationService.java
```diff
@@ -54,34 +58,46 @@ public void verifyApiKey(final HttpServletRequest request) {
...
+   * @throws ApiKeyLockOutException Thrown if rate-limiting has kicked in and the API key verification
    *         was not attempted. This can be caused by too many attempts from a specific IP address, or too many failed
    *         attempts across all IP addresses. To avoid the latter case from locking all users out, valid API keys
    *         are cached and will by-pass the lock-out if we enter into that state. We do a lock-out for all
    *         IP addresses in case an attacker sets up many computers or is able to spoof their IP address.
    */
   public synchronized int lookupModeratorIdByApiKey(final HttpServletRequest request) {
-    final String apiKey = request.getHeader(ModeratorToolboxClient.MODERATOR_API_KEY_HEADER);
-    Preconditions.checkArgument(apiKey != null && !apiKey.isEmpty());
+    final String hashedKey = extractHashedKey(request);
```
DanVanAtta: This is another important location, it was where we previously did api key verification. Now we need to combine password for a salt. Some logging was also added.

game-core/src/main/java/games/strategy/engine/lobby/moderator/toolbox/ShowToolboxController.java
```diff
@@ -0,0 +1,68 @@
...
+      EnterApiKeyPasswordWindow.show(parent, httpServerUri.get());
+    } else {
+      CreateNewApiKeyWindow.show(parent, httpServerUri.get());
+    }
+  }
+
+  private static Optional<URI> getHttpServerUri() {
+    final Optional<URI> uriOverride = ClientSetting.testLobbyHost.getValue()
+        .map(host -> ClientSetting.testLobbyHttpsPort.getValue()
+            .map(port -> URI.create("http://" + host + ":" + port))
```
RoiEXLab: Is the reason that http (without s) is hardcoded here, because this case is intended for local development?

http-server/src/main/java/org/triplea/server/moderator/toolbox/api/key/KeyHasher.java
```diff
@@ -0,0 +1,27 @@
...
+  public static String applyHash(final String apiKey, final String password) {
+    Preconditions.checkNotNull(apiKey);
+    Preconditions.checkNotNull(password);
+    return applyHash(apiKey + password);
+  }
+
+  public static String applyHash(final String valueToHash) {
+    Preconditions.checkNotNull(valueToHash);
+    Preconditions.checkArgument(!valueToHash.isEmpty());
+    return Hashing.sha512().hashString(valueToHash, Charsets.UTF_8).toString();
```
RoiEXLab: I like the abstraction, but at least for the engine code SHA hashes should have inconsistent casing, so it might be worth using a consistent case to avoid confusion in case we have to deal with the actual hashes at some point.

http-server/src/main/java/org/triplea/server/moderator/toolbox/api/key/registration/ApiKeyRegistrationService.java
```diff
@@ -0,0 +1,64 @@
...
+      final HttpServletRequest request, final String singleUseKey, final String newPassword) {
+    Preconditions.checkNotNull(singleUseKey);
+    Preconditions.checkNotNull(newPassword);
+
+    if (invalidKeyLockOut.isLockedOut(request)) {
+      throw new ApiKeyLockOutException();
+    }
+
+    final String hashedKey = singleKeyHasher.apply(singleUseKey);
+    final Integer userId = apiKeyRegistrationDao.lookupModeratorBySingleUseKey(hashedKey)
```
RoiEXLab: Since this value is guaranteed to be non-null at this point, you might as well want to use the primitive type instead.

http-server/src/main/java/org/triplea/server/moderator/toolbox/api/key/validation/ApiKeyValidationService.java
RoiEXLab: I'm wondering: Is there any chance this code might be vulnerable to a timing attack?
In other words: Is user data always being hashed before being compared in memory or against the Database or something?
I haven't really looked into the algorithm of yours in-depth, so both might be possible.

DanVanAtta: I don't think so, notably because there is not a second piece of information being validated, just the key.

The primary defense is the lockout feature. Ignoring that, we always hash the key with password, or in the single-use case hash the single-use key and compare to what is in DB. The lookup time of bad key vs good key will be a function of database retrieval time. If a user has a good key, then they have a good key, bad key attempts will always take the same amount of time (assuming DB lookup time remains pretty speedy no matter what kind of key value is used).

http-server/src/main/java/org/triplea/server/moderator/toolbox/api/key/registration/ApiKeyRegistrationService.java
DanVanAtta: Updated: bd8ee5a14

http-server/src/main/java/org/triplea/server/moderator/toolbox/api/key/KeyHasher.java
DanVanAtta: Sorry, I don't quite follow "actual hashes" and consistent vs inconsistent casing.

game-core/src/main/java/games/strategy/engine/lobby/moderator/toolbox/ShowToolboxController.java
DanVanAtta: Local development IMO should try to use 'https'. The more things that are unique on pre-prod or prod make modification more difficult and increases potential for tribal knowledge.

Good callout for the hardcoded protocol, but I'm not sure how to fix this right now and if it's worth parameterizing.

game-core/src/main/java/games/strategy/engine/lobby/moderator/toolbox/CreateNewApiKeyWindow.java
```diff
@@ -0,0 +1,158 @@
...
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+final class CreateNewApiKeyWindow {
+
+  private static final String WINDOW_TITLE = "Enter API Key & Create Password";
+
+  private static final String DESCRIPTION_LABEL =
+      String.format(
+          "<html>Enter the API key provided to you below.<br/>"
+              + "Password field is a value of your choosing, make it easy to remember.<br/>"
+              + "This is a new password for your API key, it can be the same as your lobby login password.<br/>"
```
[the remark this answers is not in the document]
DanVanAtta: Fair enough, removed the text that mentions this: a3380e46d

game-core/src/main/java/games/strategy/engine/lobby/moderator/toolbox/CreateNewApiKeyActions.java
```diff
@@ -0,0 +1,51 @@
...
+      final URI serverUri,
+      final String apiKey,
+      final String newPassword) {
+
+    final ModeratorToolboxClient toolboxClient =
+        ModeratorToolboxClient.newClient(serverUri, newPassword);
+
+    final RegisterApiKeyResult result = toolboxClient.registerNewKey(apiKey);
+
+    if (result.getNewApiKey() != null) {
```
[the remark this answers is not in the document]
DanVanAtta: Fields should be marked nullable too: d3260bcc0
Converted return API to: 2398936c7

Took a bit to convince myself that latter update should be done and how to do it cleanly. The result is maybe marginally better. I was really hoping for an `@OptionalGetter` lombok, researching that it looks like the lombok team is religiously opposed to optional.

DanVanAtta:
@RoiEXLab updated. Didn't follow on the comment of consistent vs inconsistent casing on hashes.

http-server/src/main/java/org/triplea/server/moderator/toolbox/api/key/KeyHasher.java
RoiEXLab: IIRC,in the game engine, sha-512 always looks like this: aaabbeeff13477
Whereas the api you're using might be outputting this: AABBEEFF13477

DanVanAtta: Sorry @RoiEXLab , I don't understand either the concern or the suggested course of action.

> the engine code SHA hashes should have inconsistent casing

You desire an inconsistent casing?

>  it might be worth using a consistent case

Are we shooting for an inconsistent casing? I'm confused.
Do you have suggestions for how to get an inconsistent casing if so?

>  avoid confusion in case we have to deal with the actual hashes at some point.

I'm not sure which 'actual hashes' are being referenced here.

Perhaps there is missing context, we are hashing a brand new key, the hashed key does not interact with the game engine. In the case of a new API key, generated on the server, the unhashed value is sent back to the client and that same value is returned to the server. All hashing is done by the server.

--------------

FWIW, it looks like the `Hashing.sha512` function produces lower case output. If the suggestion is to `toLowerCase()` the value, I'm not sure if that is a good move:
- looks like it is redundant, be what it may
- constraining the output could increase the chance of hash collisions. It seems like the output hash is complex anyways to avoid that

RoiEXLab: Ok, note to myself: Do not reply in a hurry when just out of bed 😅 
What I was trying to say is:
I don't know the Hashing class. The java standard library does hashing all lowercase, where as the guava tooling IIRC does hashing all upper case.
I just wanted to be sure that we don't end up comparing the same hash with a different case at some point, just because of some random implementation difference like this. Might be worth documenting if the output does interact with user input at some point.

Off Topic:
It's interesting to see how language differences (german and english in particular) affect the desired intention of a sentence.
> the engine code SHA hashes should have inconsistent casing

while this sentence is clearly wrong (I don't even know why I wrote this, it makes 0 sense 😅), the "should" in this sentence was supposed to mean "should" as in "to my knowledge I expect X to happen" rather than "in an ideal world X is the desired outcome".

I honestly don't know if both meanings are equally valid in english, in german they definitely are.


Anyways I hope everything should be clear now ^^

DanVanAtta: FWIW:

> the engine code SHA hashes should have inconsistent casing

Rephrasing with more idiomatic english:
- "the engine code SHA hashes I believe have inconsistent casing"
- "the engine code SHA hashes are an inconsistent casing"

For expressing uncertainty and also conjecturing:
- "I thought the engine code SHA hashes would have an inconsistent casing"
- "I thought they would have an inconsistent casing"

DanVanAtta: Guava hashing sha-512 appears to be all lower case.

Doing some quick reading on this, we probably want to use bcrypt. A fast algorithm like sha-512 is not that great, if DB data were leaked then it would not take too long to start cracking the hashes.

DanVanAtta: I added a TODO comment to note that we should use bcrypt: 803ad49b8

DanVanAtta:
@RoiEXLab I created an issue to help keep track of outstanding TODO items: https://github.com/triplea-game/triplea/issues/4894

http-server/src/main/java/org/triplea/server/moderator/toolbox/api/key/KeyHasher.java
RoiEXLab: Fair point. The API key is somewhat of a password replacement, but hashing isn't as crucial because they are long enough completely random tokens in a ideal world.
Adding that extra bit of security is a good idea nevertheless.

RoiEXLab:
Not sure if there's anything else to say here. The changes you proposed look good so far, and while there's probably room for improvement in a lot of places it's hard to figure out those places when the code isn't completely final yet (if code can ever considered "completely final" that is)

DanVanAtta:
Thanks for the review. I'll note that 'good enough' is probably a better review bar than trying to find every improvement possible. I learned that I think the hard way, diminishing returns is easy to achieve while some issues are going to take real times while others are pedantic. With that said, some pedantic issues due add up and can be hard to change, for example naming, lack of test or overly coupled testing and/or the more important like brittle constructs, lack of documentation, etc..

DanVanAtta:
The next set of changes are almost queued up and more or less finish up the rest of the toolbox tabs. I dont' think I'll break the PR up, so regrettably it'll be quite a lot of code once again. Of some consolation, our release model does not go directly to prod and new features can and probably should get the benefit of multiple iterations.

PR #4749, opened 2019-03-08
Remove Access Log DAO

DanVanAtta:
## Overview
Keep the DB table for now to retain the data, but otherwise
the table is deprecated as we prepare to a new schema
that will give us the same info from a new table 'player_session'
table.

## Functional Changes
No longer write or read  'access_log' DB table 

## Manual Testing Performed
- none

## Additional Review Notes

- For future schema direction see: https://github.com/triplea-game/triplea/issues/4746, note we'll eventually get the access_log table from 'user_session'

PR #4896, opened 2019-06-13
Revert "Remove Access Log DAO (#4749)"

DanVanAtta:
This reverts commit 11d4daacf728263df8886e7961f755ae8d042582.

Turns out this table is still very useful and does not need to be re-created. Moderator Toolbox 2.0 can display this table and then eventually show buttons for adding username or player bans.


DanVanAtta:
Sigh.. perhaps breaking up the PRs is not going to work, this seems like it'll take weeks at this rate. This is a very low foundational piece and iterating to get this to work without all the other changes is painful, let alone the future merge conflicts.

PR #4897, opened 2019-06-13
Add JSON class PagingParams

DanVanAtta:
This will be wired between client and server for various endpoint interactions with Moderator toolbox tabs involving large tables.


DanVanAtta:
Closing, better to include this with the http-client updates.

PR #4901, opened 2019-06-13
Update lobby text, remove tournament of champions link, already started

DanVanAtta:


RoiEXLab:
Another PR that clearly shows a flaky coverage spot

Issue #4904, opened 2019-06-14
[Mitigated] Prod DNS Not resolving


DanVanAtta:
![Screenshot from 2019-06-14 16-38-30](https://user-images.githubusercontent.com/12397753/59543935-f4995580-8ec2-11e9-9038-2863f26aa38a.png)

```
dan@dan-desk:~$ nslookup lobby.triplea-game.org
Server:		127.0.0.53
Address:	127.0.0.53#53

** server can't find lobby.triplea-game.org: SERVFAIL


dan@dan-desk:~$ dig lobby.triplea-game.org

; <<>> DiG 9.11.3-1ubuntu1.7-Ubuntu <<>> lobby.triplea-game.org
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: SERVFAIL, id: 21284
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 65494
;; QUESTION SECTION:
;lobby.triplea-game.org.		IN	A

;; Query time: 0 msec
;; SERVER: 127.0.0.53#53(127.0.0.53)
;; WHEN: Fri Jun 14 16:34:30 PDT 2019
;; MSG SIZE  rcvd: 51
```

DanVanAtta:
Uptime robot shows lobby as up, and is using the DNS address. I wonder what the secret sauce is for it.

DanVanAtta:
https://forums.triplea-game.org/ is also not resolving.

Looks like the NGINX redirect in front of the forums is re-writing to forums.triplea-game.org when accessing that machine via IP address:45.33.80.67, so there does not seem to be a work-around for forums currently.

DanVanAtta:
Just saw a user join the lobby, located in USA . So this is not a global DNS problem, regional.

DanVanAtta:
Switch to use IP address is deployed, can access lobby again.
Keeping open as forums.triplea-game.org is still dead (regional problem)

prastle:
Ya suffering same problem on any triple-game.org
lobby/forum/or downloads

prastle:
Interesting that my cell works to join forum = rogers but my comp doesn't also = rogers but diff isp

DanVanAtta:
A gap in escalation process here is having phone numbers to contact people. 
[...]

PR #4905, opened 2019-06-14
Use IP address for lobby instead of host name

DanVanAtta:
Prod DNS is not resolving, switching to use IP address.
Prod issue tracked in: https://github.com/triplea-game/triplea/issues/4904


DanVanAtta:
This is a very long wait for the build to finish  -_-

RoiEXLab:
Noting it here as well:
Hardcoding the ip like this has 2 disadvantages:
1. No IPv6 support
2. No TLS Support using free certificates

PR #4921, opened 2019-07-31
Seamless zoom

RoiEXLab:
## Overview
Had a good idea this noon and wanted to try it out.
Didn't turn out exactly how I imagined it, but I guess it's good enough.

## Functional Changes
People can change the zoom on the fly by Ctrl+Scrolling 

## Manual Testing Performed
I verified that everything including screenshots scale correctly after my changes

### Before
grafik
grafik
grafik

### After
grafik
grafik
grafik

## Additional Review Notes
There are some downsides to this change though:
1. As you might have noticed from the screenshots, the borders aren't redrawn at a fixed pixel width. Instead the width of the borders is dependent on the actual map size, so when the map is downscaled, the border is downscaled as well.
This isn't easily fixable because this border outline is part of the actual map tiles, and we'd have to redraw the whole thing which is exactly the thing we wan't to avoid here. Let me know what you think.
2. Because the performance was so bad I enabled hardware acceleration. Why is this a bad thing? It's not available on all systems and I noticed a couple of minor graphic glitches. This is the only way to achieve 60 FPS though.

I hope the pros weigh out the cons


Cernelius:
I'm really looking forward to this. The current zoom is so cluncky that it is a substantially unusable feature (and I doubt most people even know they can zoom with the scroll, anyways).

Some considerations:

1) As long as this feature is now decent (and if you wish it to be known, thus have a value, for most users), please have it as the default for scrolling. Scrolling should be the best, and most used, way of zooming. This is what I believe most games do, and using the scroll wheel to drag the board up and down is hardly meaningful and very redundant (no matter if you can also go left and right by holding shift, that most people are not going to realize, anyways). Just out of curiosity, why did you move this function to Ctrl+Scrolling (as currently it is Alt+Scrolling)?

2) You should have an option for setting how many zoom points you move at once, when scrolling. I suggest the default value being 5 (currently it is 20, that is way too much).

3) Make clicking on the scroll wheel reset the zoom (to 100%).

4) Currently, there is the (hacky) behaviour of setting "Draw Borders On Top" to medium when the zoom goes under 100%, and to low when it goes to (exactly) 100%. This has been done to avoid the borders sometime disappearing, when zooming maps, with special reference to those maps having simple 1 pixel wide borders. This is very damaging to the zoomed visual of any maps having otherwise zoom friendly details, especially if not based on pure black lines.
I strongly disagree with this current visual behaviour (so, I agree with the change at this commit, if I'm understanding it correctly), on several levels.
Firstly, if you make a map in which the borders are just 1 pixel wide, you are making a map not meant to be zoomed that much, and, if it is needed to be zoomed, you are making it wrong, so it is not something that should be addressed (or hacked) at the engine level. For example, you can see that "World War II Global..." has 1 pixel thick borders, that will partially disappear if you set zoom at 50% and Draw Borders On Top at low, but this is a fault with the skin of the map, that has too thin borders. For an example of good mapmaking, instead, you can open "World War II v5 1942 Second Edition", that has fairly thick borders (2 pixels wide), that look good if you set zoom at 50% and Draw Borders On Top at low (you can still set it at medium, if you want to have thicker borders).
Secondly, there may be and, indeed, there are maps that have details with borders that are not pure black; so this behaviour destroys the intended view, upon zooming, for all such maps.
Thirdly, a good zoom should not have the borders disappearing if they are 100% black and 1 pixel thick, if you just zoom at 50%. They should just become less black, as it happens when you shrink an image using any decent interpolation. So, even for those maps, like Global, having exceedingly thin borders, if not by remaking the actual map in a better way, the matter should rather be rather addressed by having a better zoom, then by the hack of stroking black lines over the details upon going under 100%.
Finally, this behaviour (that I've never seen documented anywhere) is unthinkable, and may throw into confusion all but the most experienced power users, and indirectly damage those few users actually using the "Draw Borders On Top" setting, as the setting will be forcefully changed upon moving the zoom into or out of the 100% value.
So, please, get rid of this thing (or I agree if you are), and have "Draw Borders On Top" alway stable, no matter the zoom (of course, default at "Low") (users can still change the default, for maps that need it, or whatever).

>As you might have noticed from the screenshots, the borders aren't redrawn at a fixed pixel width. Instead the width of the borders is dependent on the actual map size, so when the map is downscaled, the border is downscaled as well.

This is correctly how it should work. If the map has been made with borders too thin, that become hardly visible when zooming, that is a problem with the map, that has been made zoom unfriendly, and I would not agree that the engine should be supposed to address it by hacking or distorting the relative dimensions of the borders. The fact that the pixel is the minimal dimension of a raster image doesn't mean that you have to take that dimension for the borders; doing so is dumb at a mapmaking level. Will some mapmakers keep having 1 pixel thick borders even into the 8k era?

About this matter, are you aware that the difference you see is because, currently, the borders are drawn (at a fixed pixel width) above the details, when you zoom under 100%? Also, users could still have that, if they want, by setting the "Draw Borders On Top" option on medium (in case anybody will complain about the new zoom on maps with thin borders).

p.s.: Why are not we discussing/annuncing this in forum?

Cernelius:
World At War (as in the images you posted) is a partially good example of my fourth point. The current engine forces the white sea borders to become black, damaging the intended look of the board details. The matter is even worse for those maps having non black borders also for land territories.

The only problem I see with the World At War image you posted, for the new behaviour, is that some borders become so thin they almost disappear, but I don't think anyone is going to play with the map zoomed so much as in that picture. For World At War, at very most it would be meaningful comparing images taken at 50% zoom, as I don't believe anyone can possibly want to go under that.

RoiEXLab:
@Cernelius Thanks for your feedback.

From my experience zooming now feels really good and intuitive.

Good point that there should be a setting. I'll implement it later this day.

We can discuss this topic in the forum but before your feedback I wasn't quite sure what to discuss. I think you have to try it on your own to see what I mean.

Cernelius:
I'll surely try once it is merged.

Not sure if you got what I said at point 1. What I meant is not having this as Ctrl+Scrolling or Alt+Scrolling, or anything requiring holding down buttons, but just as what happens when you are simply scrolling. If you want to keep the option, you can move the up-down board scrolling to Alt+Scrolling, instead. This is very important, as only very few power users bother with anything requiring a combination of keys, or with reading any information, that may communicate it. So, if you don't want this to stay restricted to the knowledge of very few users, and to the use of even less people, zooming needs to be the primary use of the scrolling wheel.

RoiEXLab:
Interesting standpoint. I was primarily using the touchpad of my laptop where this felt super intuitive. We might want to add a setting for this, but those are just minor details we can discuss afterwards when everyone got their hands on it.

RoiEXLab:
@Cernelius We could also restrict the maximum zoom level so that you can't "scroll smaller than the map". I'll quickly try how well that works and will push it when it turns out to work well.

Anyways here's a picture with the map zoomed out so that the borders are on the edges of my window:
grafik
As you can clearly see the thin borders are just about to vanish, so we should be fine there.

Cernelius:
Not much of an idea on this, but at least for the map that wrap that makes sense, as if you zoom too much, then the map cannot wrap anymore. From what I understand, the map would stop zooming down as soon as one of the axis (either X or Y) is fully on view (which means that the map may not be fully displayed, at that point, as in the case of the image you linked, where you have all the Y axis but not all of the X axis (a bit of America is cut)), correct? If this is how it works, I tend to agree with this behaviour, especially for the wrapping map.

The problem with the borders is essentially a combination of a (in my opinion wrong) TripleA tradition at having the borders 1 pixel thick and bad quality zoom (if you shrink a image with a good quality interpolation, the borders don't disappear like that).

Regarding Classic, other than the 1 pixel thickness, you can see the details are missing for a good portion of the borders on the sea (in which you only see the base tiles). This is certainly a display issue, though I haven't looked into it (whether portion of the relief tiles are missing or are not being displayed).

The borders being put on top of the details upon zooming is a (bad) hack that was added to address the problem of the borders sometimes disappearing when zooming, in the maps having 1 pixel thick vertical or horizontal borders, usually in the sea zones (while the proper way to solve this issue would be having a better zoom, and preferably all maps with at least 2 pixels thick borders).

RoiEXLab:
@Cernelius So I went ahead an limited the zoom by the size of the map.
Once @ron-murhammer reviews the code and a new installer is built you should definitely give it a try.
I made the zoom factor a setting and left the other stuff mostly the same. It's always stunning how redoing such a "trivial" feature like zooming allows us to drop like half the code that gets changed.
I also removed the custom scale logic with the borders than only applies with 100% zoom as this was broken anyways on this branch.

Cernelius:
>So I went ahead an limited the zoom by the size of the map.

What does that mean? Will it fit the width, the height, or the one that fit the first or last?

RoiEXLab:
@Cernelius You will no longer be able to zoom so far out that you can see "outside" of the map.

Cernelius:
I personally surely agree. It never made sense to be able to zoom so much as to leave part of your available display area unused, even only on one axis. Also this way, aside from it being so at 100% already, you should avoid getting to a point where you are unable to wrap around, right? Side note, a few maps (like 270BC) will be totally unable to use the zoom feature on 4k, then, but this makes sense, as well.

RoiEXLab:
Yeah I was thinking about the high-rez case as well.
I was wondering if we should ignore the 100% zoom cap for those cases and just drag the whole thing over the screen with a shitty resolution by comparison.
I'd guess I'd prefer a huge map with a bad resolution over a perfectly fine resolution that only uses half the screen and is therefore almost too small to select units on.

Cernelius:
No. I would never enlarge raster images as default (and hardly ever as an option).

RoiEXLab:
Hey @ron-murhammer 
would you mind reviewing this when you have the time?

ron-murhammer:
@RoiEXLab Love it. Definitely a huge step in the right direction and the zoom feels good on an array of maps (some have borders that are too thin at far zoom but that is really a map issue IMO). Glanced over the code and looks good (also great to see a lot of code removed). Here are some thoughts moving forward:
1. I think 5 is a better default than 10 for zoom step as it feels a bit smoother.
2. I'd like this behavior to just be the default scroll wheel (no need for CTRL). So essentially swapping the 2 behaviors as using the scroll wheel to move the screen up and down always seemed odd so can just be moved to CTRL.
3. I think @Cernelius mentions this but the whole "draw borders on top" is a weird feature/setting and IMO should just be removed.
4. The one thing this feature makes me want is being able to drag the map by holding down the left mouse button and moving it. That with this zoom feature then would make it very easy to move around the map without having to move your mouse a lot to scroll to the edges.

Merging this. Sorry for the slow review, needed some time to actually sit and test the functionality in this case.

RoiEXLab:
@ron-murhammer Thanks for the review. No problem it took a while ^^

Yeah I'm open for discussion regarding the exact controls. The only thing I want to ensure that touchpads aren't neglected, as I'm currently really happy about the map movement using my laptop touchpad (gestures etc. seem to work, but I'm pretty sure that's just java 11 translating horizontal scrolling to scrolling + shift and zoom gestures to scroll + ctrl, but I haven't really played around with that).

RoiEXLab:
@Cernelius You can try out this feature now using https://github.com/triplea-game/triplea/releases/tag/1.10.15051 or newer
Feel free to open a Thread in the forum where we can discuss controls with everyone.

Cernelius:
>The one thing this feature makes me want is being able to drag the map by holding down the left mouse button and moving it. That with this zoom feature then would make it very easy to move around the map without having to move your mouse a lot to scroll to the edges.

@ron-murhammer Stupid question, but I really cannot understand that in any other ways: are you aware that in TripleA you can drag the map by holding down the right mouse button?

ron-murhammer:
@Cernelius Yeah, I mean changing it to the left mouse button instead. Seems like that is more standard across other strategy games?

Cernelius:
I don't play many games, but I've never come across any application that uses the right button to drag. As far as I remember, that is always the duty of the left button. It is also a bit annoying, as, then, you tend to try to drag with the right button in other applications or vice versa. However, I'm not sure how not to get in the way of the select/target handling of the left button, unless flipping everything left and right from now. Maybe you are thinking about selecting upon releasing the button without moving, but that would sometimes cause you to fall selecting for moving a little bit (and more so with the 4k, as the pixels are smaller). I'm already quite annoyed about that for the right button for deselecting, that sometimes I have to click again to actually deselect.

PR #4924, opened 2019-08-04
Update triplea_maps.yaml

beelee1:
Not sure if this is correct. It's a little different than the ones above. Maybe that's because this is the initial commit ? Or should I add "/archive/master.zip" to the end of the url and replace "github.com" with "raw.githubusercontent.com" in the "img:" ?

Thanks

## Manual Testing Performed
- ..


ron-murhammer:
@beelee1 The URL and image link need to have the same pattern as the other entries.

PR #4927, opened 2019-08-04
Remove draw borders on top

RoiEXLab:
## Overview
As discussed in #4921 this PR removes the "draw borders on top feature"
Also it does some cleanup I missed during #4921 

## Functional Changes
There is no longer a menu option to draw borders on top.

## Manual Testing Performed
I verified the game still works and normal borders still exist.


game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -48,7 +48,7 @@
 /**
  * Pro combat move AI.
  */
-class ProCombatMoveAi {
+public class ProCombatMoveAi {
```
RoiEXLab: This change is just there so my IntelliJ doesn't complain that this class is referenced in a javadoc outside of this package

Cernelius:
I recommend against removing this feature. While I don't think you would ever use it in well made map, you cannot exclude that you might encounter maps in which, because of a number of factors (borders too thin, details too confusing, units all over the place) you may want to see the borders above the details, or even above the units. I believe the main reason of existence of this option is for being able to still see the borders well upon strongly zooming maps with extremely thin borders.

RoiEXLab:
@Cernelius Your response came a little bit late there. The feature is already removed.
While I agree that there are probably maps out there that benefit from this feature we can all agree that it's probably bad map design.
I think @ron-murhammer and I agree that this should be fixed on the map side instead of at the engine level.
Also maps with this problem are probably rarely ever played and so people probably won't even notice, which is a good reason to reduce the amount of code which is why we removed it.

Cernelius:
To be clear, I was not against the existence of the ability of drawing the borders on top, as long as merely optional. I was strongly against the behaviour of automatically drawing them over the details each time you zoomed under 100%. I guess this reduces the risk of somebody coming back with the idea of having that again...

Did the recent changes imply also a modification of the zoom quality itself? I'm confronting current prerelease and current stable, and it looks to me that the quality of World At War zoomed at 60% is much better in the perelease. I also see that in the prerelease you can zoom World War II Global till 50% before any sea borders become invisible, while in the stable the borders start disappearing even if you just go down to 95%.

In both cases, for the stable, I'm referring to the process of setting the borders to "low" after each time you zoom, to actually see the true zoomed image (so, I'm not talking about that stupid thing of sneaking the borders up; just referring to the pure zoom quality, to be clear).

RoiEXLab:
> Did the recent changes imply also a modification of the zoom quality itself?

Not directly. There are some changes that could affect that, but nothing intentional.
Another thing I noticed is that Java 11 seems to respect windows zoom settings, so on my laptop the map has a different size than on my desktop using the "same" zoom level I think

Cernelius:
>The one thing this feature makes me want is being able to drag the map by holding down the left mouse button and moving it. That with this zoom feature then would make it very easy to move around the map without having to move your mouse a lot to scroll to the edges.

@ron-murhammer Stupid question, but I really cannot understand that in any other ways: are you aware that in TripleA you can drag the map by holding down the right mouse button?

RoiEXLab:
@Cernelius I think you meant to post this in the other PR, just sayin' ^^

Cernelius:
@RoiEXLab Right, thanks; I don't think I would have noticed.

PR #4928, opened 2019-08-05
Fix build with JDK 12

asvitkine:
Adds a workaround for a bug in the errorprone plugin whose fix isn't yet released.

Also adds JDK12 to the travis build config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ matrix:
   include:
   - jdk: openjdk11
     env: TRIPLEA_RELEASE=true
+  - jdk: openjdk12
+    env: TRIPLEA_RELEASE=true
 addons:
   postgresql: "10"
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ matrix:
   - jdk: openjdk11
     env: TRIPLEA_RELEASE=true
   - jdk: openjdk12
-    env: TRIPLEA_RELEASE=true
+    env: TRIPLEA_RELEASE=false
 addons:
   postgresql: "10"
   apt:
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ plugins {
     id 'java'
     id 'com.github.ben-manes.versions' version '0.20.0'
     id 'io.franzbecker.gradle-lombok' version '2.0' apply false
-    id 'net.ltgt.errorprone' version '0.8' apply false
+    id 'net.ltgt.errorprone' version '0.8.1' apply false
     id 'com.diffplug.gradle.spotless' version '3.23.0' apply false
 }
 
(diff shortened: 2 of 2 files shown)


RoiEXLab:
Hmm we might want to add a openJDK build for version 12 then. 
Thoughts @ron-murhammer ?

ron-murhammer:
@RoiEXLab Probably a good idea.

RoiEXLab:
Hey @asvitkine thanks for your PR.
Would you mind adding a travis build using java 12 to verify java 12 will continue in the future?

This should give enough clues on how to achieve that:
https://github.com/triplea-game/triplea/pull/4876/files#diff-354f30a63fb0907d4ad57269548329e3
Best-case scenario is adding a couple of lines is sufficient.

asvitkine:
> Hey @asvitkine thanks for your PR.
> Would you mind adding a travis build using java 12 to verify java 12 will continue to work in the future?
> 
> This should give enough clues on how to achieve that:
> https://github.com/triplea-game/triplea/pull/4876/files#diff-354f30a63fb0907d4ad57269548329e3
> Best-case scenario is adding a couple of lines is sufficient.

Do you mean something like this? I don't actually know what I'm doing so please advise if this is what you meant.

.travis.yml
```diff
@@ -5,6 +5,8 @@ matrix:
   include:
   - jdk: openjdk11
     env: TRIPLEA_RELEASE=true
+  - jdk: openjdk12
+    env: TRIPLEA_RELEASE=true
```
DanVanAtta: Only one jdk should have the triplea_release flag set to true, so this setting value would want to be flipped to 'false'. 

Checking usages:
```
~/work/triplea$ grep -r "TRIPLEA_RELEASE"
.travis.yml:    env: TRIPLEA_RELEASE=true
.travis.yml:    condition: "$TRIPLEA_RELEASE = true"
```
Looking at that file, seems we would get artifacts built when that is set to true for the corresponding JDK. We'd only want one set to be generated. For now we're targeting JDK11. To jump to JDK12 we'd need to update the install4j bundled JREs to be version 12 and fix any build issues with java 12. The bunlded JRE is used with the game installer, hence the extra coordination before we can switch java versions. The jump to 11 is pretty recent and the latest release is still on java 8.

DanVanAtta:
> Do you mean something like this? I don't actually know what I'm doing so please advise if this is what you meant.

@asvitkine you got it right, though the release flag as commented should be flipped in value to false

asvitkine:
> > Do you mean something like this? I don't actually know what I'm doing so please advise if this is what you meant.
> 
> @asvitkine you got it right, though the release flag as commented should be flipped in value to false

Done.

asvitkine:
Looks like some tests don't pass with the JDK12 build.

RoiEXLab:
Updating the mockito version in `build.gradle` to at least  2.24.5 should do the trick. 
See https://github.com/mockito/mockito/issues/1615

asvitkine:
> Updating the mockito version in `build.gradle` to at least 2.24.5 should do the trick.
> See [mockito/mockito#1615](https://github.com/mockito/mockito/issues/1615)

Done.

PR #4931, opened 2019-08-06
Decouple launcher code from UI

RoiEXLab:
## Overview
This PR tries to separate the game launching code a little bit more from swing.
I'm planning to work again on the JavaFX code some time soon™ so this needs to be out of the way first.

Also during testing I noticed a bug: When leaving a headless host, the chatpanel stays visible even though it probably should be removed. Not sure when this bug was introduced but it is there.

## Functional Changes
None.

## Manual Testing Performed
I verified I could host a game and connect to it.


game-core/src/main/java/games/strategy/engine/framework/ServerGame.java
```diff
@@ -442,22 +442,23 @@ private void runStep(final boolean stepIsRestoredFromSavedGame) {
...
-      autoSaveAfter(currentDelegate, headless);
+      autoSaveAfter(currentDelegate);
     }
   }
 
-  private void autoSaveAfter(final String stepName, final boolean headless) {
-    saveGame(AutoSaveFileUtils.getAfterStepAutoSaveFile(stepName, headless));
+  private void autoSaveAfter(final String stepName) {
+    final var saveUtils = launchAction.getAutoSaveFileUtils();
+    saveGame(saveUtils.getAfterStepAutoSaveFile(saveUtils.getAutoSaveStepName(stepName)));
```
DanVanAtta: I think I see a code smell here where an object is called to compute a value, then that value is passed to another method in the same object. EG: `a.bar(a.foo(value))`, can be converted to `a.bar(value)`, ie: `saveUtils.getAfterStepAutoSaveFile(stepName)`

game-core/src/main/java/games/strategy/engine/framework/ServerGame.java
```diff
@@ -442,22 +442,23 @@ private void runStep(final boolean stepIsRestoredFromSavedGame) {
...
   }
 
-  private void autoSaveAfter(final IDelegate delegate, final boolean headless) {
+  private void autoSaveAfter(final IDelegate delegate) {
     final String typeName = delegate.getClass().getTypeName();
     final String stepName = typeName.substring(typeName.lastIndexOf('.') + 1).replaceFirst("Delegate$", "");
-    saveGame(AutoSaveFileUtils.getAfterStepAutoSaveFile(stepName, headless));
+    saveGame(launchAction.getAutoSaveFileUtils().getAfterStepAutoSaveFile(stepName));
```
DanVanAtta: See a law of demeter violation here, ie: `a.b().c()`, which can be converted to `a.d()`, ie: `launchAction.getAfterStepAutoSaveFile(stepName)`. I think perhaps we might actually want two objects, a `launchAction` and a `saveGameActions`. In that case we could call something like `saveFileUtils.getAfterStepAutoSaveFile(stepName)`

Then also, `ServerGame` knows far too much about type and step names to formulate a file name. Perhaps these three lines from `autoSaveAfter` should be more simply: `saveGame(saveFileUtils.generateAfterStepFileName(delegate))`

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LaunchAction.java
```diff
@@ -0,0 +1,38 @@
...
+import games.strategy.engine.framework.LocalPlayers;
+import games.strategy.engine.framework.ServerGame;
+import games.strategy.engine.framework.startup.mc.GameSelectorModel;
+import games.strategy.engine.framework.startup.mc.ServerModel;
+import games.strategy.engine.player.IGamePlayer;
+import games.strategy.sound.ISound;
+import games.strategy.triplea.ui.display.ITripleADisplay;
+
+/**
+ * Abstraction to allow decoupling the UI framework (namely swing)
```
DanVanAtta: Comment is a bit problematic IMO as it has a lot of context about why you are creating this interface and when.

A second issue I'd point out is `LaunchAction` does not quite describe the operations. For examples, `onEnd` I think is when a game is end, not the end of the launch sequence, or `handleGameInterruption` is any interruption and not an interruption in launch. `GameEventHandler` or `GameLifeCycleHandler` might be better names perhaps for this interface.

For comment, I'd suggest to reword to state this interface is called to handle game life cycle events, which could be to notify a user or in the case of headless to perform other operations.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/ServerLauncher.java
```diff
@@ -187,44 +175,25 @@ void launchInternal(final Component parent, @Nullable final Void none) {
...
     serverModel.setServerLauncher(null);
     serverModel.newGame();
     if (inGameLobbyWatcher != null) {
       inGameLobbyWatcher.setGameStatus(GameDescription.GameStatus.WAITING_FOR_PLAYERS, null);
     }
-    if (headless) {
-      // tell headless server to wait for new connections:
-      HeadlessGameServer.waitForUsersHeadlessInstance();
-    }
+    launchAction.onGameInterrupt();
```
DanVanAtta: `handleGameInterruption` and `onGameInterrupt` seem like they are basically the same method, they are always called together, can we combine the two? It does seem there are sequencing issues, perhaps those can be solved?

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/ServerLauncher.java
```diff
@@ -187,44 +175,25 @@ void launchInternal(final Component parent, @Nullable final Void none) {
       });
       stopGame();
     } catch (final RuntimeException e) {
-      final String errorMessage = "Unrecognized error occurred. If this is a repeatable error, "
-          + "please make a copy of this savegame and report to:\n" + UrlConstants.GITHUB_ISSUES;
-      log.log(Level.SEVERE, errorMessage, e);
-      stopGame();
+      // no-op, this is a simple player disconnect, no need to scare the user with some giant stack trace
+      if (!(e.getCause() != null && e.getCause() instanceof ConnectionLostException)) {
```
DanVanAtta: Is the null check needed?

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/ServerLauncher.java
```diff
@@ -300,11 +267,7 @@ private void saveAndEndGame(final INode node) {
     stopGame();
 
     final String message = "Connection lost to:" + node.getName() + " game is over.  Game saved to:" + f.getName();
-    if (headless) {
-      log.info(message);
-    } else {
-      SwingUtilities.invokeLater(() -> JOptionPane.showMessageDialog(JOptionPane.getFrameForComponent(ui), message));
-    }
+    launchAction.onEnd(message);
```
DanVanAtta: Seeing an abstraction problem here, SRP violation perhaps. In one case we use `launchAction` to tell us the save file name, then we call it later to be a handler for an event. Very different actions.

IN another comment I suggest breaking up the object to two different ones, that perhaps could help here.

game-core/src/main/java/org/triplea/game/server/HeadlessServerSetupPanelModel.java
```diff
@@ -21,7 +39,64 @@
 
   @Override
   public void showSelectType() {
-    new ServerModel(gameSelectorModel, this, null).createServerMessenger();
+    new ServerModel(gameSelectorModel, this, null, new LaunchAction() {
+      @Override
```
DanVanAtta: Large anonymous inner class here, perhaps it's better this be a standalone class? I notice as well the headed version is its own class, yet this one is defined as an anonymous inner. Perhaps it would be more consistent as a stand-alone class so then we have the headed and headless versions.

game-core/src/main/java/org/triplea/game/server/HeadlessServerSetupPanelModel.java
```diff
@@ -21,7 +39,64 @@
...
+      public void handleGameInterruption(final GameSelectorModel gameSelectorModel, final ServerModel serverModel) {
+        try {
+          log.info("Game ended, going back to waiting.");
+          // if we do not do this, we can get into an infinite loop of launching a game,
+          // then crashing out, then launching, etc.
+          serverModel.setAllPlayersToNullNodes();
+          final File f1 = getAutoSaveFileUtils().getHeadlessAutoSaveFile();
```
DanVanAtta: With the new scoping, `f1` is probably not needed as a variable name and this could be renamed to `autoSaveFile`

game-core/src/main/java/org/triplea/game/server/HeadlessServerSetupPanelModel.java
```diff
@@ -21,7 +39,64 @@
...
+          if (!f1.exists() || !gameSelectorModel.load(f1)) {
+            gameSelectorModel.resetGameDataToNull();
+          }
+        } catch (final Exception e1) {
```
DanVanAtta: Ditto on exception variable name, probably can be renamed to more expected name `e` than `e1`

game-core/src/main/java/org/triplea/game/server/HeadlessServerSetupPanelModel.java
```diff
@@ -21,7 +39,64 @@
...
+          log.log(Level.SEVERE, "Failed to load game", e1);
+          gameSelectorModel.resetGameDataToNull();
+        }
+      }
+
+      @Override
+      public void onGameInterrupt() {
+        // tell headless server to wait for new connections:
```
DanVanAtta: Comment is a bit of a code smell as it describes a single line that should perhaps describe itself. Can we reasonably rename the method `waitForNewConnections()` and remove the comment? If not, would the comment be better off as a TODO comment to rename the method?

game-core/src/main/java/games/strategy/engine/framework/HeadlessAutoSaveType.java
```diff
@@ -7,21 +7,21 @@
  * The types of auto-saves that can be loaded by a headless game server.
  */
 public enum HeadlessAutoSaveType {
-  DEFAULT(AutoSaveFileUtils.getHeadlessAutoSaveFile()),
+  DEFAULT(new HeadlessAutoSaveFileUtils().getHeadlessAutoSaveFile()),
```
DanVanAtta: It is a bit of a code smell for new objects to be created without state to call a no-param method, does seem like it could be a static method call, ie: `SaveFileUtils.getHeadlessAutoSaveFile()`

I think probably more the enum might need to be flipped. Instead of having  path variable, have the caller compute the path. EG: `HeadlessAutoSaveFileUtils.getSaveFilePath(HeadlessAutoSave.ODD_ROUND)`, in that case this enum would be a simple enum without the path variable.

DanVanAtta:
@RoiEXLab left some comments, this is a tough update to cludgy code. There's a lot of cleanup to be done, so it's difficult to go directly to the fix you would want as there are many fixes needed and this seems to be a cross-sectional slice of those problems. It perhaps would be worthwhile to split off the bug fix so it is not mixed in with the refactoring updates.

RoiEXLab:
@DanVanAtta Maybe I didn't make it obvious: The bug is still there, this _really is_ just a plain refactoring PR.
But yeah I agree on a lot of your Points. The existing code is really a pain to decouple, and after an hour or something I just wanted to be done with this and didn't put too much thought into naming or Object design.
I'll try to follow up on most stuff tomorrow.

game-core/src/main/java/games/strategy/engine/framework/AutoSaveFileUtils.java
```diff
@@ -17,57 +17,47 @@
...
-  }
-
-  public static File getLostConnectionAutoSaveFile(final LocalDateTime localDateTime) {
+  public File getLostConnectionAutoSaveFile(final LocalDateTime localDateTime) {
     checkNotNull(localDateTime);
 
     return getAutoSaveFile(
-        addExtension("connection_lost_on_" + DateTimeFormatter.ofPattern("MMM_dd_'at'_HH_mm").format(localDateTime)));
+        addExtension("connection_lost_on_" + DateTimeFormatter.ofPattern("MMM_dd_'at'_HH_mm",
+            Locale.ENGLISH).format(localDateTime)));
```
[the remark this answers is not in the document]
RoiEXLab: On my system the test cases fail because the filename doesn't contain "May" but rather "Mai" (german word for May).
Since the whole filename is in english anyways, the month name probably doesn't matter IMO.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/ServerLauncher.java
RoiEXLab: Probably not, but I wanted to be safe here.

Never underestimate null. ^^

game-core/src/main/java/org/triplea/game/server/HeadlessServerSetupPanelModel.java
RoiEXLab: Yeah originally both implementations were anonymous classes, then I noticed I actually needed the headed version more than once, so I had to create a separate class.

game-core/src/main/java/org/triplea/game/server/HeadlessServerSetupPanelModel.java
RoiEXLab: That's actually a thing I thought as well but forgot during the process.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/ServerLauncher.java
DanVanAtta: Likely indicates a missing test case. We should be more certain than not, the `!(!a && b)` conditional is a bit confusing compared to `a || !b`. Simplest way I see to resolve is to remove the null check.

game-core/src/main/java/org/triplea/game/server/HeadlessServerSetupPanelModel.java
DanVanAtta: Will you convert this to a stand-alone class? The lack of symmetry is bothersome not to mention the anonymous inner class length. AFAIK anonymous inner classes are best short, not for full blown implementations.

DanVanAtta:
@RoiEXLab yeah, my sympathies on working with the cludgy code. I think in part the changes need to do be done in more iterations to be more sane; TripleA is actually a pretty difficult code base to work with and stay efficient.

game-core/src/main/java/org/triplea/game/server/HeadlessServerSetupPanelModel.java
RoiEXLab: Yeah, I will

RoiEXLab:
@DanVanAtta I'm hoping the codebase gets far more maintainable by redoing the UI.

By using dependency injection for the JavaFX controllers and the fact that there's such a thing in JavaFX whereas in swing you'd just extend classes, the code has to be at least a little bit more modular by design.
Having JavaFX being far more strict on threading rules and our codebase being so large we need to replace components in smaller iterations is probably also helping to write better code.

Side note: I have recently played around with JavaFX rendering and was happy to find out that the possibilities are near endless. Once JavaFX can launch basic games I will probably end up completely rewriting the rendering engine because for one javafx isn't really designed to work with our current approach of having images on top of images that get all drawn to the next layer individually, but instead you can just add primitive shapes to the node graph, apply effects and transformations if necessary and everything gets rendered for you. I also came up with a good idea on how to implement the infinite scrolling mechanism without having to actually render everything twice or more times using a plane with a texture of the current map in 3D space, where the texture gets automatically repeated by opengl or directx or whatever.

I originally wanted to write a custom effect for this, but unfortunately it turns out that the whole magic is not public api. (They created a java shading language just for this)

DanVanAtta:
> I'm hoping the codebase gets far more maintainable by redoing the UI.

Interesting topic of conversation. The UI code is embedded with much of the other logic, meaning there is not a split of responsibilities and lack of testing.

My thoughts have been to focus on compatibility issues and reflection issues, so that changes can be made without needing an incompat release or adding support code for backward compat. We also see a lot of extra testing around this as well.

Beyond that, I'm generally of the opinion that the UI needs isolation, whether it is isolated and re-written or just isolated is still the same thing.

Overall I think the best focus of JavaFX for the short term is to get the loading and launch screen rewrite completed. It's been a long running project, divide and conquer IMO and get some milestones identified and completed. Re-writing all UI is a very risky project, "too big to succeed" perhaps.

In many ways I do not disagree though, JavaFX solves a number of problems. For us though not all problems are because we use swing, but more how things are structured (or lack of structure and testing).

RoiEXLab:
Yeah, rewriting all of the UI is just the long term goal. For now the goal is really just having a working launcher, so we get to the point where we can start dropping swing code.
I probably could create a prototype launcher tomorrow, but I wouldn't have to copy all the existing code and just change the swing lines to javafx, which really shouldn't be the goal. That's why I created this PR. Hepps actually created a really nice mockup design for the new launcher UI, so I'd just have to follow that.

RoiEXLab:
And testing in JavaFX is easy when using fxml. All the components can be mocked and passed to the controller class no problem.
The only tricky part is trying to mock some specific javafx objects that are final or implement final methods.
We might want to consider enabling the mockito feature that allows mocking even those classes so that all the boilerplate code I wrote for those cases becomes redundant

DanVanAtta:
Not the right place to discuss, and off topic, but: 
I am concerned by FXML sometimes as it is not programmatic. For example you can't just code up a parameterized component, but have to parameterize external XML code and use a tool to be able to sanely update it. For launcher screens, so long as we can avoid redundancies in XMLs, and even if we can't, I do not think it'll be that much of a problem. As a general solution I would be concerned about it being maintainable. For example, the sidebar panels are all very similar in the main game, if we had "n" versions of that, it could be an issue. I think just being careful about using it well is the key. Overall though I'm a much bigger proponent of programmatic UI compared to (F)XML generated via a WYSIWYG. Code that is generated and then checked in, IMO is a actually a really big anti-pattern as that regeneration needs to be repeated. That is in contrast a bit to code that is generated post-checkin automatically, in which case it stays maintainable since the generation part does not need to be repeated (eg: lombok getter/setter annotation is good, IDE generate getter/setter methods is bad).

DanVanAtta:
@RoiEXLab do you have updates in flight for this PR? I'm thinking to submit the reformat PR. If the answer is no, I'd suggest we merge these updates and then come back to fix/cleanups in a second iteration.

RoiEXLab:
No changes right now, go ahead 👍

DanVanAtta:
@RoiEXLab cool, I'll merge now; probably should have the reformat PR up I think in an hour or two.

PR #4932, opened 2019-08-07
Plumb error correctly when yaml decoding fails.

asvitkine:
## Overview
If YAML parsing fails when trying to download maps, the underlying IOException message was not shown because the error thrown was not actually an IOException but a YamlEngineException wrapping it. This change fixes this by explicitly catching the YamlEngineException and rethrowing it as an IOException so it can be caught by higher level code expecting an IOException.

## Functional Changes
If you click "Download Maps" with your client set up such that "game-headed" is a directory (e.g. in a developer checkout), the error message will actually show the correct error message containing the path in question and not just the cryptic "Is a directory".

## Manual Testing Performed
- Click "Download Maps" in a developer instance and look at the error message shown.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/map/download/ContentReader.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/map/download/ContentReader.java
@@ -41,7 +41,7 @@ public final class ContentReader {
    * @param uri The resource URI; must not be {@code null}.
    * @param action The action to perform using the give InputStream; must not be {@code null}.
    */
-  public <T> Optional<T> download(final String uri, final Function<InputStream, T> action) {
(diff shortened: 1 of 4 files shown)


RoiEXLab:
@asvitkine Looks like you forgot to adjust the test cases. 
The tests currently don't compile

asvitkine:
Tests should be fixed.

DanVanAtta:
@asvitkine checkstyle problem. Recommend running `./gradlew spotlessApply`, likely will fix the import ordering issue.

asvitkine:
Done.

DanVanAtta:
Thanks for the follow up and changes @asvitkine , LGTM

PR #4943, opened 2019-08-08
Fix misspelling in 'channel' method name

DanVanAtta:
## Overview
Small spelling fix to private method name.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
@@ -303,7 +303,7 @@ public class ServerModel extends Observable implements IMessengerErrorListener, 
       objectStreamFactory.setData(data);
       localPlayerTypes.clear();
     }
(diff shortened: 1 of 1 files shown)


DanVanAtta:
Killing `Tuple` I think was important for moving DB dedicated code to the DAO sub-project. I'd support removing it for that reason alone. It does need its own dedicated iterations though.

RoiEXLab:
@DanVanAtta I was actually referring to a real object in the GameData graph.
I don't recall where exactly it's located, but it's the object that caused the checkstyle warning "line length exceeded 120 chard" for a long time because the generic arguments weren't wrapped by the eclipse formatter automatically.
Just search for list or something of a tuple containing a tuple containing 2 player ids, and a tuple with 2 other types using some regex, or dig through the checkstyle related pr history.

When I checked the other day if it was still there, I realised that this construct should really be some sort of multimap. ^^

RoiEXLab:
I found the code I was reffering to in BattleTracker.java:
grafik
I also found this in UnitAttachment:
grafik

PR #4946, opened 2019-08-08
Disallow multiple names from same IP address

DanVanAtta:
## Overview
Prevents users from logging in from the same computer with different names. Intended to prevent malicious users from creating 'sock puppet' accounts where they can appear as if they are different people but are actually the same. The typical TripleA user should be logging in just once from a given IP address.

## Functional Changes
users can no longer log in from the same computer with multiple names.

## Manual Testing Performed
- launched a local hosted game and logged in multiple times. Note, the host is not considered as a connected node, so someone could host and self connect with a new name. Any subsequent connections would not be allowed to have new names.
- launched a lobby and verified that only the first name is allowed, subsequent names get numerals appended to them.

## Screenshots
Error message when logged in under a different name:
Screenshot from 2019-08-07 23-17-16

## Additional Review Notes
A good chunk of the code is existing. The code in `PlayerNamAssigner` is moved there and added the new method "findLoggedInName". 
There is then an update to use that method with an `Optional.ofNullable(..).orElse)` and second update to the existing code is to extract the "aa" dummy name to a constant.

Beyond that the existing code was extracted and then heavily tested.

There is in addition a bug fix as the stream to find an existing name found the first match which might not be the shortest match. For example if there is the name "dan" and "dan (1)", the code could find "dan (1)" and generate "dan (1) (1)" instead of "dan (2)". The 'sort' line added fixes this.

- The new thread to show the dialog is a bit suspect, but is actually needed. Removing that results in  a case where the dialog message cannot be closed.

### DIFF
--- a/game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
+++ b/game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
@@ -49,16 +49,9 @@ public final class PlayerNameAssigner {
     if (currentName.length() < 2) {
       currentName = "aa" + currentName;
     }
-    if (isNameTaken(currentName, loggedInNodes)) {
-      int i = 1;
-      while (true) {
-        final String newName = currentName + " (" + i + ")";
-        if (!isNameTaken(newName, loggedInNodes)) {
-          currentName = newName;
-          break;
-        }
-        i++;
-      }
+    final String originalName = currentName;
+    for (int i = 1; isNameTaken(currentName, loggedInNodes); i++) {
+      currentName = originalName + " (" + i + ")";
     }
     return currentName;
   }
--- a/game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
+++ b/game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
@@ -47,9 +47,6 @@ public final class PlayerNameAssigner {
     if (currentName.length() > 50) {
       currentName = currentName.substring(0, 50);
     }
-    if (currentName.length() < 2) {
-      currentName = "aa" + currentName;
-    }
     final String originalName = currentName;
     for (int i = 1; isNameTaken(currentName, loggedInNodes); i++) {
       currentName = originalName + " (" + i + ")";
--- a/game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
+++ b/game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
@@ -69,7 +69,6 @@ public final class PlayerNameAssigner {
   private static boolean isNameTaken(final String nodeName, final Collection<INode> nodes) {
     return nodes
         .stream()
-        .sorted()
         .map(INode::getName)
         .anyMatch(nodeName::equalsIgnoreCase);
   }
--- a/game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
+++ b/game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
@@ -1,0 +1,83 @@
+package games.strategy.net;
+
+import java.net.InetAddress;
+import java.util.Collection;
+import java.util.Optional;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+
+/**
+ * Utility class that will assign a name to a newly logging in player.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class PlayerNameAssigner {
+
+  /**
+   * Name string returned in the case of a hacked client sending us an invalid/unexpected name.
+   */
+  @VisibleForTesting
+  static final String DUMMY_NAME = "aa";
+
+  /**
+   * Returns a node name, based on the specified node name, that is unique across all nodes. The node name is made
+   * unique by adding a numbered suffix to the existing node name. For example, for the second node with the name "foo",
+   * this method will return "foo (1)".
+   *
+   * @param desiredName The name being requested by a new user joining.
+   * @param ipAddress The IP address of the user that is joining. Used to determine if the user
+   *        is already connected under a different name.
+   * @param loggedInNodes Collection of nodes that have already joined.
+   */
+  public static String assignName(
+      final String desiredName,
+      final InetAddress ipAddress,
+      final Collection<INode> loggedInNodes) {
+    Preconditions.checkNotNull(ipAddress);
+    Preconditions.checkNotNull(loggedInNodes);
+
+    String currentName =
+        (desiredName == null || desiredName.length() < 3)
+            ? DUMMY_NAME
+            : findLoggedInName(ipAddress, loggedInNodes).orElse(desiredName);
+    if (currentName.length() > 50) {
+      currentName = currentName.substring(0, 50);
+    }
+    if (currentName.length() < 2) {
+      currentName = "aa" + currentName;
+    }
+    if (isNameTaken(currentName, loggedInNodes)) {
+      int i = 1;
+      while (true) {
+        final String newName = currentName + " (" + i + ")";
+        if (!isNameTaken(newName, loggedInNodes)) {
+          currentName = newName;
+          break;
+        }
+        i++;
+      }
+    }
+    return currentName;
+  }
+
+  private static Optional<String> findLoggedInName(
+      final InetAddress socketAddress, final Collection<INode> nodes) {
+    return nodes
+        .stream()
+        .sorted()
+        .filter(node -> node.getAddress().equals(socketAddress))
+        .findFirst()
+        .map(INode::getName);
+  }
+
+  private static boolean isNameTaken(final String nodeName, final Collection<INode> nodes) {
+    return nodes
+        .stream()
+        .sorted()
+        .map(INode::getName)
+        .anyMatch(nodeName::equalsIgnoreCase);
+  }
+}
--- a/game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
+++ b/game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
@@ -2,6 +2,7 @@ package games.strategy.net;
 
 import java.net.InetAddress;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.Optional;
 
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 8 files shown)


DanVanAtta:
Looks like the pop-up warning text is triggered in an integ tests, should be reasonably fast to fix, will do so tomorrow.

game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
```diff
@@ -0,0 +1,83 @@
...
+      int i = 1;
+      while (true) {
+        final String newName = currentName + " (" + i + ")";
+        if (!isNameTaken(newName, loggedInNodes)) {
+          currentName = newName;
+          break;
+        }
+        i++;
+      }
+    }
```
RoiEXLab: I know this code is copied but what about changing it into this:
```java
final String originalName = currentName;
for (int i = 1; isNameTaken(currentName, loggedInNodes); i++) {
  currentName = originalName + " (" + i + ")";
}
```

game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
```diff
@@ -0,0 +1,83 @@
...
+    return currentName;
+  }
+
+  private static Optional<String> findLoggedInName(
+      final InetAddress socketAddress, final Collection<INode> nodes) {
+    return nodes
+        .stream()
+        .sorted()
+        .filter(node -> node.getAddress().equals(socketAddress))
+        .findFirst()
```
RoiEXLab: I'd recommend filtering before sorting, and using `min(Comparator.naturalOrder())` instead of `sorted().findFirst()`, because the former method only has a complexity of O(n) vs O(n log n)

game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
```diff
@@ -0,0 +1,83 @@
...
+        .map(INode::getName);
+  }
+
+  private static boolean isNameTaken(final String nodeName, final Collection<INode> nodes) {
+    return nodes
+        .stream()
+        .sorted()
```
RoiEXLab: This sorted call seems unecessary.

EDIT: I just saw that you added this call compared to the "original code". I assume it was used at some point?

game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
```diff
@@ -0,0 +1,83 @@
...
+
+    String currentName =
+        (desiredName == null || desiredName.length() < 3)
+            ? DUMMY_NAME
+            : findLoggedInName(ipAddress, loggedInNodes).orElse(desiredName);
+    if (currentName.length() > 50) {
+      currentName = currentName.substring(0, 50);
+    }
+    if (currentName.length() < 2) {
+      currentName = "aa" + currentName;
```
RoiEXLab: Does this prefix have something to do with `DUMMY_NAME`?
If not, could you document why?

game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
DanVanAtta: My initial reaction was to see if we can avoid using a "non-standard" for loop construct. At the same time the suggestion is pretty simple and any other while loop construct is not terribly clean. It's interesting to see this pattern of while loop as well, it's a Java 1.3 style loop :) 

Updated per suggestion: 63772f3fc0a690259fbed5dda4104c5d2c02c356

swing-lib/src/main/java/org/triplea/swing/DialogBuilder.java
```diff
@@ -15,8 +16,18 @@
...
+
   private DialogBuilder() {}
 
+  /**
+   * Method to turn off UI notifications for cases when we are running tests.
+   */
+  @VisibleForTesting
+  public static void disableUi() {
+    uiEnabled = false;
+  }
```
RoiEXLab: This is kinda ugly, but arguably swing is pretty annoying when it comes to writing tests

game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
DanVanAtta: We expect very small collections and we are iterating over this list repeatedly, so the runtime is actually squared! Perhaps the best way to add all names to a hash set once, then do a O(1) lookup. 

I took your initial suggestion here @RoiEXLab : d5229a042493ff267f4fc327144573ddeab6b08e

Turns out the conversion to a set is about as complicated and reduces the runtime from O(n^2) to O(n), secondly it can be parallelized more cleanly as well, updated in:  84e08e528

game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
DanVanAtta: Yeah, I was getting sort happy, it turns out the bug I was noting in the overview is actually only a bug in the new code and was not there to begin with. I removed the unnecessary sort here: 908e96daa9dddb01484c4adcde2cda7437bd9740

ON another update this method is removed entirely though and names are mapped to a set:  84e08e5281e2f5a5a1f93115a7c3cd15731c4610

game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
DanVanAtta: This if statement should have been just removed: b9513050d90623449b309747dd560c19f65ae93f

We are defending only against custom clients here. I'm not sure offhand what the best way to handle this is, maybe even generating a random name or probably more properly to disallow the name.

I went ahead and made that update here, to reject names that are too short and that removes the need for any 'dummy' name code: 765cf0d6a

Small nuance is that typically the too short name case should not happen, but we can have a custom client that forces the scenario, so server should deal with it.

swing-lib/src/main/java/org/triplea/swing/DialogBuilder.java
DanVanAtta: Agree, this is a place where the integ test should be injecting a mock or a fake implementation. Certainly a bit quick and dirty.

game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
```diff
@@ -0,0 +1,72 @@
...
+        (desiredName == null || desiredName.length() < 3)
+            ? DUMMY_NAME
+            : findLoggedInName(ipAddress, loggedInNodes).orElse(desiredName);
+    if (currentName.length() > 50) {
+      currentName = currentName.substring(0, 50);
+    }
+    final Collection<String> playerNames = new HashSet<>(
+        loggedInNodes.stream()
+            .map(INode::getName)
+            .parallel()
```
RoiEXLab: Is parallel really increasing performance here?

DanVanAtta: It should, right? Instead of iterating over the list and mapping and adding one at a time, we'll do "n" elements at a time. 

With a 'min' operation, a parallel may not improve performance as the min element needs to be held in memory, but mapping to a set it should be done faster

RoiEXLab: Well in this particular scenario the only thing this is parallelizing is the collection step which is probably not that much faster or even slower because the collections per thread have to be merged somewhere

DanVanAtta: Looks like the answer is 'no' for as-is, if "parallel" is put before the map then it is slightly slower.

I wrote a test script that did a similar map operation over 5M elements and got these results (elapsed time is in milliseconds):

```
Sequential elapsed: 3503, element count: 5000000
Parallel first: 4928, element count: 5000000
Parallel last: 3987, element count: 5000000
```
test script:
````
  private static class NameHolder {
    @Getter
    private final String name = String.valueOf(Math.random());
  }

public static void main(String[] args) {
    final int collectionSize = 5_000_000;
    List<NameHolder> names = new ArrayList<>(collectionSize);
    for(int i = 0; i < collectionSize; i ++) {
      names.add(new NameHolder());
    }

    long start = System.currentTimeMillis();
    final Collection<String> notParallel = new HashSet<>(names.stream()
        .map(NameHolder::getName)
        .collect(Collectors.toSet()));

    long elapsed = System.currentTimeMillis() - start;

    System.out.println("Sequential elapsed: " + elapsed + ", element count: " + notParallel.size());

    start = System.currentTimeMillis();
    final Collection<String> parallelFirst = new HashSet<>(names.stream()
        .parallel()
        .map(NameHolder::getName)
        .collect(Collectors.toSet()));

    elapsed = System.currentTimeMillis() - start;
    System.out.println("Parallel first: " + elapsed + ", element count: " + parallelFirst.size());

    start = System.currentTimeMillis();
    final Collection<String> parallelLast = new HashSet<>(names.stream()
        .map(NameHolder::getName)
        .parallel()
        .collect(Collectors.toSet()));

    elapsed = System.currentTimeMillis() - start;
    System.out.println("Parallel last: " + elapsed + ", element count: " + parallelLast.size());
}
```

DanVanAtta: > if "parallel" is put before the map then it is slightly slower.

I was expected that to be the result, I was wrong there; it's actually slowest when the mapping is done in parallel. That is interesting.

DanVanAtta: I also wonder a bit what happens when the resulting set is not then added back to a new hash set... :thinking:

DanVanAtta: If the results are not added to a hashset, it does get a bit faster (and sequential still wins):
```
Sequential elapsed: 2493, element count: 5000000
Parallel first: 3308, element count: 5000000
Parallel last: 3474, element count: 5000000
```

ssoloff: FYI, as stream execution is lazy, it shouldn't matter where `parallel()` is called, as long as the parallel flag is set before invoking the terminal operation (`collect()` in this case).  From the [Javadocs](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/package-summary.html#Parallelism):

> The stream pipeline is executed sequentially or in parallel depending on the mode of the stream on which the terminal operation is invoked. The sequential or parallel mode of a stream can be determined with the `BaseStream.isParallel()` method, and the stream's mode can be modified with the `BaseStream.sequential()` and `BaseStream.parallel()` operations. **The most recent sequential or parallel mode setting applies to the execution of the _entire_ stream pipeline.**

PR #4949, opened 2019-08-09
Dev documentation, updated IDE setup

DanVanAtta:
## Overview

 - Simplify/update how to build steps, add steps needed to run local testing.
 - Add new contribution guideline page with notable gotchyas when it comes to
  TripleA development, useful for new devs.
 - Remove IDEA analysis config, was not terribly well setup and could use redoing.

## Functional Changes
- none

## Review Notes

I split up the IDE setup instructions so they are on their own: #4951


docs/dev/local_build_and_testing.md
```diff
@@ -0,0 +1,37 @@
...
+Use docker to launch a local postgres DB, flyway used to install schema. Steps:
+```bash
+cd lobby-db
+# build docker only needs to be run once
+./build_docker
+./run_docker
+# wait about 3 or 5 seconds for database to start
+./run_flyway
+```
+Docker for Mac can be obtained at: https://store.docker.com/editions/community/docker-ce-desktop-mac
```
RoiEXLab: Actually this link doesn't help for users that don't have a docker account. I don't know if you had to deal with this before but I certainly had to:
As described in this issue: https://github.com/docker/docker.github.io/issues/6910 direct links still exists, but are sort of hidden in the release notes: https://docs.docker.com/docker-for-mac/release-notes/

But for simplicity I'd recommend installing docker via `brew cask install docker` if you have brew and brew cask installed. (Which as a developer you probably should)

docs/dev/updating_gradle.md
```diff
@@ -1,6 +1,7 @@
 ## How to upgrade Gradle
 
-- Execute: `./gradlew wrapper --gradle-version=<version>`, where `<version>` is replaced with the desired Gradle version (e.g. `4.8.1`, `4.9`, etc.).
+- Execute: `./gradlew wrapper --gradle-version=<version>`, where `<version>` is replaced with 
+the desired Gradle version (e.g. `4.8.1`, `4.9`, etc.).
```
RoiEXLab: Actually the latest version is 5.5.1, you might want to update the examples ^^

docs/dev/build_installers.md
```diff
@@ -0,0 +1,21 @@
+### Pre-Requisite
+- install47 license key (can get from maintainers or an open-source one from install4j)
+
+## Building installers
+ - Install [Install4j7](https://www.ej-technologies.com/download/install4j/files)
```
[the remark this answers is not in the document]
DanVanAtta: Probably even better to remove the version so that we don't have the point of maintenance.

docs/dev/local_build_and_testing.md
DanVanAtta: We did not have much information on the docker page, this seemed to have been the most useful. The update here is simply dropping that one line of info to the main page and then deleting the rest of the docker page. It's been some time since I installed docker on a mac, I can't speak to the best way of doing it for now. It could make sense to delete this link entirely if it is not useful and allow the next person to install on mac to update the notes.

docs/dev/updating_gradle.md
DanVanAtta: As long as it's a matter of replacing the numbers with the latest values, it's good enough IMO. Notably, even `5.5.1` would become out of date as soon as there is another version. We do not want documentation that we have to constantly maintain

PR #4950, opened 2019-08-09
DBUser / Name+email validation refactor  and validate name on server side

DanVanAtta:
## Overview
- Refactor: Remove DBUser class, extract and consolidate username and user email validation from DBUser to dedicated utility classes.
- Update: Perform username validation on server side. Bot names need to be ignored, some small logic added for server to ignore 'lobby_watcher' from name validation. Custom clients could remove name validation, this update enforces the consistent naming rule. Most notably prevents userse from 'spoofing' other users by choosing a user name like "Bob (1)"
- Fix: previous update to detect multiple logins from same IP address did not account for the case when a player is hosting a bot and then logs in. This caused a problem where username was incorrectly overriden to match the automated host lobby watcher name.

## Functional Changes
- Server (lobby) will do name validation to match name validation done on client side
- Fixes name override when a user is hosting a bot

## Manual Testing Performed
Launched a lobby, bot and headed game. Verified that:
- bot could connect
- could connect to same lobby and was allowed to choose a name, did not default back to a 'lobby_watcher' name.


RoiEXLab:
Hmm looks like your plan didn't work out...
Merge conflicts showed up

DanVanAtta:
@RoiEXLab merge conflicts fixed. Recreated the branch from a latest mast  and cherry-picked the new commits.

game-core/src/main/java/games/strategy/engine/lobby/PlayerEmailValidation.java
```diff
@@ -0,0 +1,26 @@
...
+    final String quotedString = "\"(?:[^\"\\\\]|\\\\\\p{ASCII})*\"";
+    final String atom = "[^()<>@,;:\\\\\".\\[\\] \\x28\\p{Cntrl}]+";
+    final String word = "(?:" + atom + "|" + quotedString + ")";
+    final String subdomain = "(?:" + atom + "|\\[(?:[^\\[\\]\\\\]|\\\\\\p{ASCII})*\\])";
+    final String domain = subdomain + "(?:\\." + subdomain + ")*";
+    final String localPart = word + "(?:\\." + word + ")*";
+    final String email = localPart + "@" + domain;
+    // String regex = "(\\s*[\\w\\.-]+@\\w+\\.[\\w\\.]+\\s*)*";
+    final String regex = "(\\s*" + email + "\\s*)*";
+    return emailAddress.matches(regex) ? null : "Invalid email address";
```
RoiEXLab: I haven't actually tried this, but is this regex actually "accurate"?
I mean this is debatable, but the best way to check if an email is actually valid is still to actually try to send one to the address.

game-core/src/main/java/games/strategy/engine/lobby/PlayerNameValidation.java
```diff
@@ -0,0 +1,49 @@
...
+        : serverSideValidate(userName);
+  }
+
+  /** Less restrictive validation intended to be done on server side. */
+  public static String serverSideValidate(final String userName) {
+    if ((userName == null) || (userName.length() < MIN_LENGTH)) {
+      return "Name is too short (minimum " + MIN_LENGTH + " characters)";
+    } else if (userName.length() > MAX_LENGTH) {
+      return "Name is too long (maximum " + MAX_LENGTH + " characters)";
+    } else if (!userName.matches("[0-9a-zA-Z_-]+")) {
```
RoiEXLab: I know certain unicode characters are problematic, but I don't think we should restrict users to just a subset of ascii. At least allowing common europe characters like äöü é è ê ñ etc. would be nice.

game-core/src/main/java/games/strategy/engine/lobby/server/userDB/DBUser.java
```diff
@@ -1,140 +0,0 @@
-// CHECKSTYLE-OFF: PackageName
-// move this class to lobby.common upon next lobby-incompatible release; it is shared between client
```
RoiEXLab: We should definitely search for "incompatible" before pushing a new release

game-core/src/test/java/games/strategy/engine/lobby/PlayerEmailValidationTest.java
```diff
@@ -0,0 +1,37 @@
...
+  @Test
+  void shouldReturnTrueWhenAddressIsValid() {
+    Arrays.asList(
+            "some@some.com",
+            "some.someMore@some.com",
+            "some@some.com some2@some2.com",
+            "some@some.com some2@some2.co.uk",
+            "some@some.com some2@some2.co.br",
+            "",
+            "some@some.some.some.com")
```
RoiEXLab: I know the tests are just copied but JUnit has a maven artifact that allows for parameterized tests.
It works actually really well and makes this probably more readable

game-core/src/test/java/games/strategy/engine/lobby/PlayerNameValidationTest.java
```diff
@@ -0,0 +1,84 @@
...
+            invalidName -> {
+              assertThat(
+                  "Expected name to be marked as invalid: " + invalidName,
+                  PlayerNameValidation.isValid(invalidName),
+                  is(false));
+              assertThat(
+                  "Expected name to have validation error messages: " + invalidName,
+                  PlayerNameValidation.validate(invalidName),
+                  not(emptyString()));
+            });
```
RoiEXLab: Ditto on parameterized tests

lobby/src/test/java/org/triplea/lobby/server/db/UserControllerIntegrationTest.java
```diff
@@ -37,52 +36,50 @@ void testCreate() throws Exception {
...
-    assertTrue(controller.login(user.getName(), new HashedPassword(password)));
-    assertTrue(controller.login(user.getName(), new HashedPassword(obfuscate(password))));
+        user, generateEmailAddress(user), new HashedPassword(bcrypt(obfuscate(password))));
+    assertTrue(controller.login(user, new HashedPassword(password)));
+    assertTrue(controller.login(user, new HashedPassword(obfuscate(password))));
   }
 
   @Test
   void testUpdate() throws Exception {
-    final DBUser user = newUserWithMd5CryptHash();
```
RoiEXLab: I love this change 👍 

Took me a while to figure out that this refactoring dropped this class entirely

game-core/src/main/java/games/strategy/engine/lobby/PlayerEmailValidation.java
DanVanAtta: Funny enough, I think we had this same debate about regex the last time this code was touched. For now, it's a more significant change to remove the previous validation. Please note this is moved code from StringUtils.

I do in general agree that regex is not a good way to go, in general checking for an `@` sign and min length is almost as good and the only way to really know is to send a confirmation link to the email.

I do not rightly know if it's really that worthwhile to downgrade the regex check since it is already in place and seemingly accurate enough. Ultimately I suspect we perhaps should focus our attention elsewhere until we start to do more with email.

game-core/src/main/java/games/strategy/engine/lobby/PlayerNameValidation.java
DanVanAtta: Cernel pointed out rendering problems when non-latin characters are entered into chat, we probably should fix that before opening it up to usernames. The test burden is also there to be sure DB can handle it and that we do not get other issues. I wonder if the original motivation for the restriction is to avoid 'leet-speak' names. I like the idea of keeping it simple and not having very fun '|p0VV|\|/\6E', but I think that can also be debated both ways.

I'm afraid to say the update to loosen that up is involved enough it really should not be done as part of this refactoring. The client side validation does need to be done on the server side though as the original code really has a vulnerability where it assumes a person is using an unmodified client, which is a bad assumption. That leads to a situation where you can spoof someone by using a custom client and then adding trailing whitespace to a name or a ' (1)'.

game-core/src/main/java/games/strategy/engine/lobby/server/userDB/DBUser.java
DanVanAtta: Yeah! My 2 cents, I think we probably should work on an incompatible very quickly after this upcoming release. The efficiency improvement and liberty to make change is very significant. @prastle points out that for players the incompatible update is not quite as impactful as one would think (it still hurts, but the people claiming to abandon TripleA after incompatible are few and mostly being loud; yet with that said we should try to make upgrades as painless as we can).

game-core/src/test/java/games/strategy/engine/lobby/PlayerEmailValidationTest.java
DanVanAtta: This is in part habit on my part; I've yet to really get used to parameterized tests and they have their limitations. Some early versions I really did not like very much and thought they made things less readable. AFAIK the latest does need some magic glue to work.

Is it worth fixing this as a one-off, or would we be better to convert many unit tests to be parameterized all in one go? I'm not sure i like the idea of further increasing scope and mixing more changes into this update, particularly when we are no worse off after the update.

RoiEXLab: I'm fine with the latter 🤷‍♂

game-core/src/main/java/games/strategy/engine/lobby/PlayerEmailValidation.java
RoiEXLab: I knew I discussed this piece of code some time in the past, I just couldn't recall whom I discussed this with ^^

For now I'd be fine if you just remove the comment there

game-core/src/main/java/games/strategy/engine/lobby/PlayerNameValidation.java
RoiEXLab: Agreed that this should be discussed in another place at another time.

Yes, swing is really stupid when it comes to non-ascii characters sadly. It doesn't even support emoji (I tried), JavaFX should be able to deal with this IIRC ^^

game-core/src/test/java/games/strategy/engine/lobby/PlayerEmailValidationTest.java
RoiEXLab: Well using the JUnit 5 variant you basically have a couple of options on how you want to retrieve the parameters for the test.
I believe there's an option for an external csv file, an option for "inline csv" which might be what you want, or an option to specify a source method, unfortunately via reflection because annotations don't allow lambdas or any other potentially mutable object.

RoiEXLab: BUT IntelliJ supports the reflection variant and actually recognizes that this parameterized test is feeded by the other method.

DanVanAtta: CSV file for this one test is not a good option IMO. It's just one test, I don't think it would scale well as we should wind up with thousands more unit tests. Inline as well, why go to the complication and the formatting headache?

We could have a 'feeder' method, it does have limitations, I'm not necessarily convinced it's the right thing to do and would not want to couple it with more changes. The main change here is to extract the validation code so it can be used elsewhere. Really the only update to all this is one new line in ServerQuarantine that does additional validation.

DanVanAtta: For brand new code, I would agree we should look at test parameterization, particularly 'feeder' methods. That does add a layer of indirection and non-locality, which has me a bit pressed to be 100% convinced it's actually better. I do think it's a valid thing to bring up, so I appreciate the consideration/conversation of it.

lobby/src/test/java/org/triplea/lobby/server/db/UserControllerIntegrationTest.java
DanVanAtta: I find it interesting that I added `DBUser` to begin with and have wanted to remove it now for some time. Just interesting to see my own style evolve over a couple of years and to see what I once thought was a good pattern for type safety to not be so good.

RoiEXLab: Well I think the biggest (and perhaps only) problem is that really a DB-entity as you'd expect from a framework like spring. So you had to constantly sync this object with the actual state of the database which added a lot of boilerplate.

TL;DR; If the object itself would've modified the database based on its state and vice versa it would've probably been better

PR #4955, opened 2019-08-09
Fix BattleWindow draw issue

### LINKED ISSUE #4953: Combat Window Territory Thumbnail Not Rendering Correctly
## Problem
Territory thumbnail in combat window is not rendering correctly

## Repro
Start a combat, notice the territory thumbnail rendering

Not able to repro with SHA: f55fd39fae124aaa7bbe2384544f29b5f617a180 (before seamless zoom, problem is introduced at seamless zoom or thereafter)

### Observed:
Screenshot from 2019-08-09 14-10-36

### Expected:
Screenshot from 2019-08-09 14-14-23

RoiEXLab:
## Overview
Fixes #4953
Looks like TreeSet considers 2 drawables equal when their level is the same. Not quite sure why this is because the javadocs state that the objects have to be equal according to Objects.equals, but whatever

## Functional Changes
Bugfix.

## Manual Testing Performed
I verified it is displayed correctly.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/screen/TileManager.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/screen/TileManager.java
@@ -499,11 +499,12 @@ public class TileManager {
       final Rectangle bounds,
       final Graphics2D graphics,
       final boolean drawOutline) {
-    final List<IDrawable> drawables = getTiles(bounds).stream()
-        .map(Tile::getDrawables)
-        .flatMap(Collection::stream)
-        .sorted()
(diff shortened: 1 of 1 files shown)


DanVanAtta:
Rubber stamping, I see the 'missing' call to `getTiles` was added, but it's not readily clear to me why that is a fix beyond perhaps it was there before... :man_shrugging:

RoiEXLab:
@DanVanAtta I mean I explained the fix in the description, but basically TreeSet seems to treat some drawables as duplicates and doesn't even add them to the tree

PR #4956, opened 2019-08-11
Simplify Unit.java, use lombok

DanVanAtta:
## Manual Testing Performed
- none

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/Unit.java
+++ b/game-core/src/main/java/games/strategy/engine/data/Unit.java
@@ -6,9 +6,11 @@ import com.google.common.collect.ImmutableMap;
 import games.strategy.net.GUID;
 import games.strategy.triplea.attachments.UnitAttachment;
 import java.util.Map;
-import java.util.Objects;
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/engine/data/Unit.java
```diff
@@ -86,22 +62,6 @@ public boolean isEquivalent(final Unit unit) {
...
-              + (type == null ? "Unit of UNKNOWN TYPE" : type.getName())
-              + " owned by "
-              + (owner == null ? "UNKNOWN OWNER" : owner.getName())
-              + " with id: "
-              + getId();
-      UnitDeserializationErrorLazyMessage.printError(text);
-      return 0;
-    }
-    return Objects.hashCode(id);
-  }
```
RoiEXLab: This reminds me of something. I believe some time ago we had this issue where something would silently fail due to serialization errors.
I don't remember who exactly dealt with this issue. I believe @ssoloff might be the person that added the code to fail-fast.
Not sure if this would be useful here as well.

PR #4961, opened 2019-08-12
Add: Lobby Login Throttle

DanVanAtta:
## Overview

Adds a throttle for too many failed login attempts to lobby. If a user fails too many times due to invalid password (or other), then they will be locked out for three minutes. This is to prevent brute-force password-cracking attacks.

## Functional Changes
- Adds new lobby login failure if user attempts to login too many times and fails repeatedly.
- 3 minute cache is the cooldown period, any renewed attempts will extend the cooldown by 3 minutes
- Lockout can happen for any login failure reason, bad version, bad name, or bad password.

## Manual Testing Performed
- Before update, verified bad passwords can be repeatedly tried
- After update, verified that after 'n' failed password attempts there is a lockout message.

## Screen Shots

After failing to login too many times:
Screenshot from 2019-08-11 20-34-18


RoiEXLab:
There's a checkstyle violation

PR #4964, opened 2019-08-13
Log message as well

RoiEXLab:
## Overview
To debug https://forums.triplea-game.org/topic/1510/ further

## Functional Changes
More logging.

## Manual Testing Performed
None.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/pbem/NodeBbForumPoster.java
+++ b/game-core/src/main/java/games/strategy/engine/pbem/NodeBbForumPoster.java
@@ -82,10 +82,12 @@ abstract class NodeBbForumPoster implements IForumPoster {
     try (CloseableHttpResponse response = client.execute(post)) {
       final int code = response.getStatusLine().getStatusCode();
       if (code != HttpURLConnection.HTTP_OK) {
+        final var message =
+            response.getEntity() == null
+                ? ""
+                : " and message '" + EntityUtils.toString(response.getEntity()) + '\'';
         throw new IllegalStateException(
-            String.format(
-                "Forum responded with code %s and message '%s'",
-                code, EntityUtils.toString(response.getEntity())));
+            String.format("Forum responded with code %s%s", code, message));
       }
     }
   }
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/engine/pbem/NodeBbForumPoster.java
```diff
@@ -82,7 +82,9 @@ private void post(
     try (CloseableHttpResponse response = client.execute(post)) {
       final int code = response.getStatusLine().getStatusCode();
       if (code != HttpURLConnection.HTTP_OK) {
-        throw new IllegalStateException("Forum responded with code " + code);
+        throw new IllegalStateException(String.format(
+            "Forum responded with code %s and message '%s'",
+            code, EntityUtils.toString(response.getEntity())));
```
DanVanAtta: Will entity always be non-null? Empty and error responses are certain to be not null?

DanVanAtta: Does `response.toString()` provide too much info?

DanVanAtta:
The 500 we are debugging, there is no info on the server about it?

DanVanAtta:
> The following files had format violations:
      game-core/src/main/java/games/strategy/engine/pbem/NodeBbForumPoster.java

game-core/src/main/java/games/strategy/engine/pbem/NodeBbForumPoster.java
RoiEXLab: I don't think `response.toString()` does provide any information. I might be wrong, but `EntityUtils` is what's used everywhere so I used the thing I know works.

The documentation is lacking a lot of details like if certain values can be null, but AFAIK `getEntity()` always returned a non-null value. If anything goes wrong with the request that might prevent creating an entity this will result in an exception and never reach this point of execution.

RoiEXLab:
> The 500 we are debugging, there is no info on the server about it?

The server logged that there's a 500 using a java client with java 8_144 and apache http components (user agent string), but obviously it doesn't log all the responses, otherwise the log wouldn't be lasting very long.
There isn't any stacktrace if that was your question.

game-core/src/main/java/games/strategy/engine/pbem/NodeBbForumPoster.java
DanVanAtta: Looking through source code implementations, it looked like at least one implementation of `Response` did have a `toString` implementation and the other delegated back to that same implementation.

I don't think we can assume a non-null response.

RoiEXLab:
Will do later this evening

DanVanAtta:
@RoiEXLab thanks, feel free to merge after you take either route of toString'ing response or adding a NPE check.

PR #4966, opened 2019-08-13
Move add/remove error listener from IMessenger to IClientMessenger

DanVanAtta:
## Overview
Goal is to remove calls that result in no-op. ServerMessenger and LocalNoOpMessenger are no-ops for add/remove error listener. The methods are moved to related interfacer IClientMessenger where is always an effect. Cases where those methods are called on IMessenger, and sometimes when we have an IClientMessenger, an 'instanceof' check is added to cast the IMessenger to an IClientMessenger. In cases when the IMessenger is always a ServerMessenger (and a no-op), those calls are removed.

## Functional Changes
None

## Manual Testing Performed
None

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
@@ -29,7 +29,6 @@ import games.strategy.engine.framework.startup.ui.ServerOptions;
 import games.strategy.engine.message.RemoteName;
 import games.strategy.io.IoUtils;
 import games.strategy.net.IConnectionChangeListener;
-import games.strategy.net.IMessengerErrorListener;
 import games.strategy.net.INode;
 import games.strategy.net.IServerMessenger;
 import games.strategy.net.Messengers;
@@ -68,8 +67,7 @@ import org.triplea.util.Version;
 
 /** Represents a network-aware game server to which multiple clients may connect. */
 @Log
-public class ServerModel extends Observable
-    implements IMessengerErrorListener, IConnectionChangeListener {
+public class ServerModel extends Observable implements IConnectionChangeListener {
   public static final RemoteName SERVER_REMOTE_NAME =
       new RemoteName(
           "games.strategy.engine.framework.ui.ServerStartup.SERVER_REMOTE",
@@ -273,7 +271,6 @@ public class ServerModel extends Observable
     if (messengers != null) {
       chatController.deactivate();
       messengers.shutDown();
-      messengers.removeErrorListener(this);
       chatModel.setChat(null);
     }
   }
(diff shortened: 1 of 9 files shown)


game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
```diff
@@ -273,7 +271,6 @@ public void cancel() {
     if (messengers != null) {
       chatController.deactivate();
       messengers.shutDown();
-      messengers.removeErrorListener(this);
```
RoiEXLab: Can't quite follow this change here. Are the messengers guaranteed to never be instances of IClientMessenger?

game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
```diff
@@ -389,7 +386,6 @@ private void createServerMessenger(@Nonnull final ServerConnectionProps props) {
       final ClientLoginValidator clientLoginValidator = new ClientLoginValidator(serverMessenger);
       clientLoginValidator.setGamePassword(props.getPassword());
       serverMessenger.setLoginValidator(clientLoginValidator);
-      serverMessenger.addErrorListener(this);
```
RoiEXLab: Ditto with this removal

game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
```diff
@@ -523,14 +519,6 @@ public IServerMessenger getMessenger() {
...
   }
 
-  @Override
-  public void messengerInvalid(final Throwable reason) {
-    if (ui != null) {
-      JOptionPane.showMessageDialog(ui, "Connection lost", "Error", JOptionPane.ERROR_MESSAGE);
-    }
-    serverSetupModel.showSelectType();
-  }
-
```
RoiEXLab: Great to see changes like this. Was ServerModel simply never used as IMessengerErrorListener?

game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
DanVanAtta: In short, yes, so it's always a no-op. This kind of update I think is the main reason for this kind of refactor.

Line 395: `      messengers = new Messengers(serverMessenger);`, the one place where Messengers is initialized. 

`serverMessenger` is defined line 83:  ` private IServerMessenger serverMessenger;`, and the only implementations are no-op.

game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
DanVanAtta: Same answer, but more clearly as `IServerMessenger` is always a no-op on error listeners.

game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
DanVanAtta: Whether for better or worse, yes, the operations were always a no-op and this was dead code.
It's a possible bug really that this stuff wound up being a no-op. I suspect there was more code thrown at the problem to make the system work properly.

PR #4973, opened 2019-08-15
Refactor: Immutable data class ClientProps.

DanVanAtta:
## Overview
- Convert ClientProps to be immutable
- Small simplification to combine ClientSettings set and flush to a single statement


game-core/src/main/java/games/strategy/engine/framework/startup/mc/ClientModel.java
```diff
@@ -470,11 +469,12 @@ public Action getHostBotGetGameSaveClientAction(final Component parent) {
...
 
+  /** Simple data object for which host we are connecting to and with which name. */
   @Getter
-  @Setter
+  @Builder
   private static class ClientProps {
-    private int port;
-    private String name;
-    private String host;
+    @Nonnull private Integer port;
```
RoiEXLab: Its a real bummer we can't use "normal" ints just to have lombok require them.

PR #4975, opened 2019-08-15
Refactor: rename variable names (InGameLobbyWatcher.java)

DanVanAtta:


game-core/src/main/java/games/strategy/engine/framework/startup/ui/InGameLobbyWatcher.java
```diff
@@ -218,11 +218,13 @@ public static InGameLobbyWatcher newInGameLobbyWatcher(
...
               mac,
               login);
-      final UnifiedMessenger um = new UnifiedMessenger(messenger);
-      final RemoteMessenger rm = new RemoteMessenger(um);
-      final RemoteHostUtils rhu = new RemoteHostUtils(messenger.getServerNode(), gameMessenger);
-      rm.registerRemote(rhu, IRemoteHostUtils.Companion.newRemoteNameForNode(um.getLocalNode()));
-      return new InGameLobbyWatcher(messenger, rm, gameMessenger, handler, oldWatcher);
+      final var unifiedMessenger = new UnifiedMessenger(messenger);
+      final var remoteMessenger = new RemoteMessenger(unifiedMessenger);
+      final var remoteHostUtils = new RemoteHostUtils(messenger.getServerNode(), gameMessenger);
```
RoiEXLab: Excellent use of the var keyword 👍

game-core/src/main/java/games/strategy/engine/framework/startup/ui/InGameLobbyWatcher.java
```diff
@@ -302,10 +304,10 @@ private void postUpdate(final GameDescription newDescription) {
 
   void shutDown() {
     isShutdown = true;
-    if (messenger instanceof IClientMessenger) {
-      ((IClientMessenger) this.messenger).removeErrorListener(messengerErrorListener);
+    if (lobbyMessenger instanceof IClientMessenger) {
+      ((IClientMessenger) this.lobbyMessenger).removeErrorListener(messengerErrorListener);
```
RoiEXLab: I wonder if we could move listener cleanup to shutdown. On IClientMessenger implementations

game-core/src/main/java/games/strategy/engine/framework/startup/ui/InGameLobbyWatcher.java
DanVanAtta: Thanks! It's not universally a good thing, but this does seem to be a really good example of where it's useful.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/InGameLobbyWatcher.java
DanVanAtta: A lot of the listener cleanup is very suspeicious IMO. I suspect we might be able to further and not need the cleanup and instead rely on tight/proper scoping. With that said, moving the cleanup code could be a good first step.

PR #4981, opened 2019-08-15
Add: Forgot Password DB Table and DAO

DanVanAtta:
## Overview
Adds table and DAO for storing temp passwords generated by upcoming 'forget password' feature. Code to use the DAO will be added in future commits.

### DIFF
--- a/lobby-db-dao/src/main/java/org/triplea/lobby/server/db/dao/TempPasswordDao.java
+++ b/lobby-db-dao/src/main/java/org/triplea/lobby/server/db/dao/TempPasswordDao.java
@@ -1,0 +1,49 @@
+package org.triplea.lobby.server.db.dao;
+
+import java.util.Optional;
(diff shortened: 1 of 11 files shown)


DanVanAtta:
DB is appealing:
- if we can ever stop reading password in DB, then we can combine the DB updates with password verification in one call.
- we have to access DB regardless to lookup user info, retrieve email and verify username exists
- DB is helpful for rate limiting, we keep seeing passwords generated we want to limit it.
- The temp password verification is essentially the same as password verification. Having the two be similar is fundamental consistency
- DB allows for multiple lobby servers if we ever want that, and as you mentioned survives restarts

RoiEXLab:
Your last argument convinced me 100%

We're probably not in the need to load balance the server, but for multiple (potentially different versions) of the lobby this seems like a good thing to have.

DanVanAtta:
@RoiEXLab Cool, glad to hear it was convincing. I do applaud the consideration

PR #4983, opened 2019-08-15
Refactor Remove lobby login challenge keys from RsaAuthenticator

DanVanAtta:
## Overview
Simplify RsaAuthenticator slightly by removing challenge map keys and moving them to callers. Adds a bit of redundancy with a trade-off of fixing a SRP violation (RsaAuthenticator should not know about challenge keys necessarily) and makes the logic of which keys are placed where more epxlicit.


DanVanAtta:
AFAIK this is fine in another feature branch where it came from; I'll submit it as part of the larger work.

PR #4985, opened 2019-08-15
Update: Allow user password update without updating email

DanVanAtta:
## Overview
- (1) Future looking update to allow for email to be passed as 'null' when updating user password. In this case the email is not updated. In this way we do not require user email when updating user password. This will be useful for future 'forget password' feature so we do not need to lookup user email nor ask user for email so we can call this method with a non-null email.

- (2) Small refactor to combine try-with resources lines. The open connection line can be moved lower and combined into a single try-with-resources line.

### DIFF
--- a/lobby/src/main/java/org/triplea/lobby/server/db/UserController.java
+++ b/lobby/src/main/java/org/triplea/lobby/server/db/UserController.java
@@ -57,11 +57,15 @@ final class UserController implements UserDao {
         PreparedStatement ps =
             con.prepareStatement(
                 String.format(
-                    "update lobby_user set %s=?, email=? where username=?",
-                    getPasswordColumn(hashedPassword)))) {
+                    "update lobby_user set %s=?%s where username=?",
+                    getPasswordColumn(hashedPassword), email == null ? "" : ", email=?"))) {
       ps.setString(1, hashedPassword.value);
-      ps.setString(2, email);
-      ps.setString(3, name);
+      int index = 2;
+      if (email != null) {
+        ps.setString(index, email);
+        index++;
+      }
+      ps.setString(index, name);
       ps.execute();
       if (!hashedPassword.isBcrypted()) {
         try (PreparedStatement ps2 =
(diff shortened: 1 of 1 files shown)


lobby/src/main/java/org/triplea/lobby/server/db/UserController.java
```diff
@@ -57,11 +57,15 @@ public void updateUser(
...
+                    getPasswordColumn(hashedPassword), email == null ? "" : ", email=?"))) {
       ps.setString(1, hashedPassword.value);
-      ps.setString(2, email);
-      ps.setString(3, name);
+      int index = 2;
+      if (email != null) {
+        ps.setString(index, email);
+        index++;
+      }
+      ps.setString(index, name);
```
RoiEXLab: I personally find this very hard to read tbh.
I'd rather go with something like this:
```java
if (email != null) {
  ps.setString(2, email);
}
ps.setString(email != null ? 3 : 2, name);
```
Because IMO this makes it more clear which indices are used

DanVanAtta: It's not terribly clean either way and the string format piece above is hacky. With just the two columns your suggestion is nicer, though I don't think it would be if there were more columns. 

I don't have a real preference on this either way, updated: 512746e0c

lobby/src/main/java/org/triplea/lobby/server/db/UserController.java
```diff
@@ -109,24 +113,23 @@ public void createUser(
...
-      }
+    final HashedPassword actualPassword = getPassword(username);
+    if (actualPassword == null) {
+      return false;
+    }
+    Preconditions.checkState(actualPassword.isBcrypted());
+    if (!BCrypt.checkpw(hashedPassword.value, actualPassword.value)) {
+      return false;
+    }
+    // update last login time
```
DanVanAtta: This is a place where it's a shame we have to read password. We could do password validation and update in one go with a where clause:

```
update lobby_user set last_login = now() where username = ? and password = ?
```

Benefits there is the update is fully atomic, and the encrypted password never leaves the database. Having functions to read passwords from DB is not the best. If an attacker is able to call arbitrary server commands somehow, that could be used to read all encrypted passwords from DB, while on the other hand if it is one-way only then the best they could do is a brute force attack against unknown values.

RoiEXLab: I agree that having to read a password before comparing it isn't optimal, but it's recommended security practice to store passwords this way.

DanVanAtta: Bcrypt is a good 'slow' hashing mechanism; I wonder if there is another API we can call to compute the expected hash and use that. Essentially avoid the comparison of using `Bcrypt.checkPw`, IMO it's more the APIs we are using and less the encryption mechanism.

RoiEXLab: I mean if you insist on comparing hashes directly, you could of course query the db, to give you the metadata information of the hash (amount of rounds, salt) and hash the "original password" using this information.
This will result in the same hash that can be compared with the db.

However if you're looking for a way to compare passwords without any previous query whatsoever, there isn't any possibility because the whole point of hashing passwords with salt is that they need to be compared individually so an attacker can't pre-compute all the hashes

DanVanAtta: Interesting, thanks for the respons @RoiEXLab ; something I'll need to think about.

PR #4987, opened 2019-08-15
Small code cleanup

RoiEXLab:
## Overview
There are a lot of minor changes that are suggested by IntelliJ.
Only applied the ones that are 100% safe to apply, but please double check so I didn't miss something.
Will open 2 PRs with changes that might alter behaviour.

## Functional Changes
None

## Manual Testing Performed
None

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/export/GameDataExporter.java
+++ b/game-core/src/main/java/games/strategy/engine/data/export/GameDataExporter.java
@@ -119,6 +119,7 @@ public class GameDataExporter {
             for (final String definedName : TechAdvance.ALL_PREDEFINED_TECHNOLOGY_NAMES) {
               if (definedName.equals(name)) {
                 name = cat;
+                break;
               }
             }
             returnValue.append("                <tech name=\"").append(name).append("\"/>\n");
@@ -144,6 +145,7 @@ public class GameDataExporter {
         for (final String definedName : TechAdvance.ALL_PREDEFINED_TECHNOLOGY_NAMES) {
           if (definedName.equals(name)) {
             name = cat;
+            break;
           }
         }
         returnValue.append("            <techname name=\"").append(name).append("\"");
--- a/game-core/src/main/java/games/strategy/engine/data/export/GameDataExporter.java
+++ b/game-core/src/main/java/games/strategy/engine/data/export/GameDataExporter.java
@@ -116,11 +116,8 @@ public class GameDataExporter {
           for (final TechAdvance tech : frontier.getTechs()) {
             String name = tech.getName();
             final String cat = tech.getProperty();
-            for (final String definedName : TechAdvance.ALL_PREDEFINED_TECHNOLOGY_NAMES) {
-              if (definedName.equals(name)) {
-                name = cat;
-                break;
-              }
+            if (TechAdvance.ALL_PREDEFINED_TECHNOLOGY_NAMES.contains(name)) {
+              name = cat;
             }
             returnValue.append("                <tech name=\"").append(name).append("\"/>\n");
           }
--- a/game-core/src/main/java/games/strategy/triplea/delegate/BattleCalculator.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/BattleCalculator.java
@@ -719,14 +719,14 @@ public class BattleCalculator {
     // check right number
     if (!isEditMode
         && !(numhits + damaged.size()
-            == (hitsRemaining > totalHitpoints ? totalHitpoints : hitsRemaining))) {
+            == (Math.min(hitsRemaining, totalHitpoints)))) {
       tripleaPlayer.reportError("Wrong number of casualties selected");
       if (headLess) {
         log.severe(
             "Possible Infinite Loop: Wrong number of casualties selected: number of hits on units "
                 + (numhits + damaged.size())
                 + " != number of hits to take "
-                + (hitsRemaining > totalHitpoints ? totalHitpoints : hitsRemaining)
+                + (Math.min(hitsRemaining, totalHitpoints))
                 + ", for "
                 + casualtySelection.toString());
       }
--- a/game-core/src/main/java/games/strategy/triplea/delegate/BattleCalculator.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/BattleCalculator.java
@@ -717,14 +717,14 @@ public class BattleCalculator {
       }
     }
     // check right number
-    if (!isEditMode && !(numhits + damaged.size() == (Math.min(hitsRemaining, totalHitpoints)))) {
+    if (!isEditMode && numhits + damaged.size() != Math.min(hitsRemaining, totalHitpoints)) {
       tripleaPlayer.reportError("Wrong number of casualties selected");
       if (headLess) {
         log.severe(
             "Possible Infinite Loop: Wrong number of casualties selected: number of hits on units "
                 + (numhits + damaged.size())
                 + " != number of hits to take "
-                + (Math.min(hitsRemaining, totalHitpoints))
+                + Math.min(hitsRemaining, totalHitpoints)
                 + ", for "
                 + casualtySelection.toString());
       }
--- a/game-core/src/main/java/games/strategy/triplea/delegate/BattleCalculator.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/BattleCalculator.java
@@ -717,9 +717,7 @@ public class BattleCalculator {
       }
     }
     // check right number
-    if (!isEditMode
-        && !(numhits + damaged.size()
-            == (Math.min(hitsRemaining, totalHitpoints)))) {
+    if (!isEditMode && !(numhits + damaged.size() == (Math.min(hitsRemaining, totalHitpoints)))) {
       tripleaPlayer.reportError("Wrong number of casualties selected");
       if (headLess) {
         log.severe(
--- a/game-core/src/main/java/games/strategy/triplea/delegate/BattleDelegate.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/BattleDelegate.java
@@ -806,8 +806,8 @@ public class BattleDelegate extends BaseTripleADelegate implements IBattleDelega
       if (defender == null || scramblers.isEmpty()) {
         continue;
       }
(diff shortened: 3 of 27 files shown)


game-core/src/main/java/games/strategy/engine/data/export/GameDataExporter.java
```diff
@@ -119,6 +119,7 @@ private static String playertechs(final GameData data) {
             for (final String definedName : TechAdvance.ALL_PREDEFINED_TECHNOLOGY_NAMES) {
               if (definedName.equals(name)) {
                 name = cat;
+                break;
```
DanVanAtta: Any thoughts to using java stream API here?

game-core/src/main/java/games/strategy/triplea/delegate/BattleCalculator.java
```diff
@@ -717,16 +717,14 @@ public static CasualtyDetails selectCasualties(
       }
     }
     // check right number
-    if (!isEditMode
-        && !(numhits + damaged.size()
-            == (hitsRemaining > totalHitpoints ? totalHitpoints : hitsRemaining))) {
+    if (!isEditMode && !(numhits + damaged.size() == (Math.min(hitsRemaining, totalHitpoints)))) {
```
DanVanAtta: This is really quite the conditional expression. The `Math.min` helps. Further refactoring would be good here, but I can certainly appreciate the improvement for now.

game-core/src/main/java/games/strategy/triplea/delegate/BattleDelegate.java
```diff
@@ -806,8 +806,8 @@ private void doScrambling() {
       if (defender == null || scramblers.isEmpty()) {
         continue;
       }
-      final Tuple<Territory, PlayerId> terrPlayer = Tuple.of(to, defender);
-      final Collection<Map<Territory, Tuple<Collection<Unit>, Collection<Unit>>>> tempScrambleList =
+      final var terrPlayer = Tuple.of(to, defender);
+      final var tempScrambleList =
```
DanVanAtta: Not critical - Any reason to not just inline this variable?

game-core/src/main/java/games/strategy/triplea/delegate/BattleDelegate.java
```diff
@@ -834,8 +833,7 @@ private void doScrambling() {
         if (scramblers.isEmpty()) {
           continue;
         }
-        final Map<Territory, Collection<Unit>> toScramble =
-            getRemotePlayer(defender).scrambleUnitsQuery(to, scramblers);
+        final var toScramble = getRemotePlayer(defender).scrambleUnitsQuery(to, scramblers);
```
DanVanAtta: There is a cost to losing type info, I'm not sure if the code neatness is worth the loss of the type information. Perhaps if the variable were named: `unitsToScrambleByTerritory`, then the type information would be more redundant. Any thoughts?

game-core/src/main/java/games/strategy/triplea/delegate/BattleDelegate.java
```diff
@@ -861,9 +859,9 @@ private void doScrambling() {
         // Validate players have enough fuel to move there and back
         final Map<PlayerId, ResourceCollection> playerFuelCost = new HashMap<>();
         for (final Entry<Territory, Collection<Unit>> entry : toScramble.entrySet()) {
-          final Map<PlayerId, ResourceCollection> map =
+          final var map =
```
DanVanAtta: Ditto with the variable name not being useful. The previous code relied more on the code knowing the type to make sense of what was happening. Perhaps renaming this here to `playerToResourcesMap` would make the type information unnecessary.

game-core/src/main/java/games/strategy/triplea/delegate/BattleDelegate.java
```diff
@@ -861,9 +859,9 @@ private void doScrambling() {
...
               Route.getScrambleFuelCostCharge(entry.getValue(), entry.getKey(), to, data);
-          for (final Entry<PlayerId, ResourceCollection> playerAndCost : map.entrySet()) {
+          for (final var playerAndCost : map.entrySet()) {
```
DanVanAtta: My first reaction is this would be better named: `playerAndCostEntry` so we can keep some information on the typing. I'm not sure if that is more what I'm used to or actually helpful, but perhaps worth considering.

game-core/src/main/java/games/strategy/triplea/delegate/BattleTracker.java
```diff
@@ -156,9 +156,8 @@ boolean didAllThesePlayersJustGoToWarThisTurn(
   private boolean didThesePlayersJustGoToWarThisTurn(final PlayerId p1, final PlayerId p2) {
     // check all relationship changes that are p1 and p2, to make sure that oldRelation is not war,
     // and newRelation is war
-    for (final Tuple<Tuple<PlayerId, PlayerId>, Tuple<RelationshipType, RelationshipType>> t :
-        relationshipChangesThisTurn) {
-      final Tuple<PlayerId, PlayerId> players = t.getFirst();
+    for (final var tuple : relationshipChangesThisTurn) {
```
DanVanAtta: `tuple` probably could use a rename; This is a scary bit of code, it might even be worth leaving alone and waiting for a more significant refactor to fix it up.
Failing that, a `var` in foreach loop I think raises question if the lambda foreach variant shoudl be used, eg:
`relationshipChangesThisTurn.forEach(tuple -> ...)`.

Overall it's a code smell for procedural code. The refactoring fix I think is to make `relationshipChangesThisTurn` a first class object. In which case the for loop is extracted entirely to the new object and we get something like this: `return relationshipChangesThis.arePlayersAtWar(p1, p2)`. In that case the entire private method is probably extracted to the new object since the method is essentially just this for loop.

game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
```diff
@@ -385,8 +385,9 @@ private void selectUnitsToMove(
                     TransportUtils.mapTransportsToLoad(unitsToLoad, airTransportsToLoad);
                 boolean ableToLoad = true;
                 for (final Unit unit : unitsToLoad) {
-                  if (!unitMap.keySet().contains(unit)) {
+                  if (!unitMap.containsKey(unit)) {
                     ableToLoad = false;
+                    break;
```
DanVanAtta: Thoughts to using stream API here?

game-core/src/main/java/games/strategy/engine/data/export/GameDataExporter.java
RoiEXLab: Found something even simpler

game-core/src/main/java/games/strategy/triplea/delegate/BattleDelegate.java
RoiEXLab: In this case I'm fine reverting this change.
I had this stash laying around somewhere and I remeber stopping because at some point the changes were no longer good.

game-core/src/main/java/games/strategy/triplea/delegate/BattleTracker.java
RoiEXLab: That's the piece of code I was refferring to in another PR.
And I agree with you, but for now the only goal is to make the code more readable, and I think you can agree that the existing code isn't really better.

game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
RoiEXLab: I found something better

game-core/src/main/java/games/strategy/triplea/delegate/BattleTracker.java
DanVanAtta: While the existing code was ugly, the type information is there. `var` is good IMO when the type information is redundant and obvious, eg:
- `List playerList = new ArrayList<>();`
- `RelationshipTracker relationshipTracker = new RelationshipTracker()`

Type information is helpful, so while harder to read with the nasty Tuple mess, it does add information of what data is there. Losing that requires a reader to do a lookup of the definition of the `relationshipChangesThisTurn` variable and cross-reference its type with the usage, which I think actually detracts from understandability (can't just understand this code by looking at it, a cross-reference to other locations is an undersatndability smell)

game-core/src/main/java/games/strategy/triplea/delegate/BattleTracker.java
```diff
@@ -156,9 +156,8 @@ boolean didAllThesePlayersJustGoToWarThisTurn(
...
+      final Tuple<PlayerId, PlayerId> players = tuple.getFirst();
       if (players.getFirst().equals(p1)) {
```
DanVanAtta: I see `players` is a new variable, is there variable shadowing happening here?

game-core/src/main/java/games/strategy/triplea/delegate/BattleTracker.java
DanVanAtta: I do notice now more clearly you added lines like: `  final Tuple<PlayerId, PlayerId> players = tuple.getFirst();`

Perhaps that is enough to help and mitigates the typing concern.

game-core/src/main/java/games/strategy/triplea/delegate/BattleTracker.java
RoiEXLab: The variable existed before as far as I can tell

game-core/src/main/java/games/strategy/engine/data/export/GameDataExporter.java
```diff
@@ -116,11 +116,8 @@ private static String playertechs(final GameData data) {
           for (final TechAdvance tech : frontier.getTechs()) {
             String name = tech.getName();
             final String cat = tech.getProperty();
-            for (final String definedName : TechAdvance.ALL_PREDEFINED_TECHNOLOGY_NAMES) {
-              if (definedName.equals(name)) {
-                name = cat;
-                break;
-              }
+            if (TechAdvance.ALL_PREDEFINED_TECHNOLOGY_NAMES.contains(name)) {
```
DanVanAtta: Nice :+1: 
I wonder/suspect this code might predate Java 1.5 and the collections API

PR #4989, opened 2019-08-15
Correct condition in for loop?

RoiEXLab:
## Overview
The static analysis found this either impossible condition, condition that executes until the int overflows.

## Functional Changes
BattleRecords will be added correctly?

## Manual Testing Performed
None

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/BattleRecordsList.java
+++ b/game-core/src/main/java/games/strategy/engine/data/BattleRecordsList.java
@@ -133,7 +133,7 @@ public class BattleRecordsList extends GameDataComponent {
       brs.add(brl.getCurrentRoundCopy());
     } else {
       final Map<Integer, BattleRecords> currentList = brl.getBattleRecordsMapCopy();
-      for (int i = beginningRound; i < endRound; i++) {
+      for (int i = beginningRound; i <= endRound; i++) {
         final BattleRecords currentRecords = currentList.get(i);
         if (currentRecords != null) {
           brs.add(currentRecords);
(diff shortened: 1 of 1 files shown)


DanVanAtta:
I think we may want to understand this code first. I'd be much more confident running through this in a debugger to understand better the impact and current behavior. Is this just dead code maybe and hence we never see the problem? Are battle records correctly added later downstream?

RoiEXLab:
Good Point. I might have some time during the weekend. My current guess is that this is just dead code. Maybe @ron-murhammer knows something about this code?

ron-murhammer:
@RoiEXLab This appears to only be used in one place and is used to check conditions that use the option `battle`. I did a quick search on the map repos and appears only POS2 has an example of it but no map actually uses it: https://github.com/search?q=org%3Atriplea-maps+%22option+name%3D%5C%22battle%5C%22%22&type=Code

So I think its just a bug that no one has ever encountered/reported for that parameter. I think the fix is actually to make it <= instead of <.

RoiEXLab:
If no map actually uses it, could it be worth to just drop the feature? 
Not sure how much code would be affected.

ron-murhammer:
@RoiEXLab Well no maps in the repo use it that I found with a quick search. I think some of the larger XMLs aren't included in that search and then you have all the maps outside the repo. I'd say without a forum discussion that its better to just leave it as it seems like a reasonable feature.

RoiEXLab:
No worries, wasn't actually planning something.
It's just that whenever I hear "unused" I'm wondering how much code this would make redundant.
Less code is always a good thing.

PR #4991, opened 2019-08-16
Add: Forgot Password Feature

DanVanAtta:
## Overview:
Update lobby login prompt to have a 'forgot password' button. Clicking that a user can enter a username (subject to rate limit of three times a day) to then request a temporary password be sent to them (by username) via email. The email contains a temporary password that can then be used to login. After logging in with a temporary password, user is prompted to change password.

## Changes:
- Add a 'Forgot Password' button to the lobby login pop-up
- Add table 'temp_password_request_history' to keep track of temporary passwords requested. If user reaches a limit then no new temporary passwords will be issued.
- Add lobby logic to generate a temporary password and store it in temp password table
- Add lobby logic to check password supplied against temporary password table
- Add magic return value to indicate a temorary password was used for login, this triggers a new 'change password' dialog on login.
- Add environment flag for lobby "local_dev" to skip sending email when running lobby locally.
- Refactor: extract creation of ClientMessenger to a factory class. This is to help consolidate the challenge/response and simplify creation of the client messenger through convenience methods.
- Add new ClientMessenger for forgot password. Server responds with a magic string if a temporary password is sent via email, otherwise returns an error message to client.
- Refactor: RSA authenticator code, extract challenge keys from the class and move creation of challenge keys to client classes of the authenticator.
- Update: rename table 'user_temp_password' to 'temp_password_request'

## Functional Changes
Adds 'forgot password' feature

## Manual Testing Performed
- Verifies main use-cases locally.
- Small bug note: after hitting the rate limit case, server socket can be closed out of sync from client and this displays a pop-up. It is not consistent, does not happen all the time, and overall is a relatively minor (skipping a fix for it, for now.)
- Deployed a test-app to prod server to send email and verified email sender code.

### Before
Screenshot from 2019-08-15 21-46-08

### After

Clicking "Play Online" button, notice the new "Forgot Password Button" (and the dots are removed from "Create Account.."):
Screenshot from 2019-08-15 21-44-38

Clicking forgot password, the following dialog is displayed:
Screenshot from 2019-08-15 21-20-00

Enter in a username in forgot password that exists, this dialog is displayed:
Screenshot from 2019-08-15 21-20-10

If entering in an invalid username, one that does not exist in DB, this error message is shown:
Screenshot from 2019-08-15 21-49-09

If user has requested too many temporary passwords, this error message is displayed:
Screenshot from 2019-08-15 21-24-23

If successfully requested new temporary password, the email sent looks like this:
Screenshot from 2019-08-15 21-43-55

The body of the email looks like this:
Screenshot from 2019-08-15 21-44-03

After logging in to the server with temporary password, this change password dialog is shown (okay button is enabled when password matches verification and is of min length):
Screenshot from 2019-08-15 21-20-32

After setting a new password, this dialog is displayed:
Screenshot from 2019-08-15 21-20-46

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/login/ChangePasswordPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/login/ChangePasswordPanel.java
@@ -130,8 +130,8 @@ public final class ChangePasswordPanel extends JPanel {
   }
 
   private void close() {
-    if (validatePasswords() && dialog != null) {
-      dialog.dispose();
+    if (dialog != null) {
+      dialog.setVisible(false);
     }
   }
 
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/login/ChangePasswordPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/login/ChangePasswordPanel.java
@@ -1,0 +1,161 @@
+package games.strategy.engine.lobby.client.login;
+
+import games.strategy.ui.Util;
+import java.awt.BorderLayout;
+import java.awt.FlowLayout;
+import java.awt.GridBagConstraints;
+import java.awt.GridBagLayout;
+import java.awt.Insets;
+import java.awt.Window;
+import java.util.Arrays;
+import javax.annotation.Nullable;
+import javax.swing.BorderFactory;
+import javax.swing.ImageIcon;
+import javax.swing.JButton;
+import javax.swing.JDialog;
+import javax.swing.JLabel;
+import javax.swing.JOptionPane;
+import javax.swing.JPanel;
+import javax.swing.JPasswordField;
+import org.triplea.swing.DocumentListenerBuilder;
+import org.triplea.swing.SwingComponents;
+
+/** Panel dedicated to changing password after user has logged in with a temporary password. */
+public final class ChangePasswordPanel extends JPanel {
+
+  private final String title = "Change Password";
+  private @Nullable JDialog dialog;
+  private final JPasswordField passwordField = new JPasswordField();
+  private final JPasswordField passwordConfirmField = new JPasswordField();
+  private final JButton okButton = new JButton("OK");
+
+  private ChangePasswordPanel() {
+    layoutComponents();
+    setupListeners();
+  }
+
+  /**
+   * Creates a new instance of the {@code CreateUpdateAccountPanel} class that is used to update the
+   * specified lobby account.
+   *
+   * @return A new {@code CreateUpdateAccountPanel}.
+   */
+  public static ChangePasswordPanel newChangePasswordPanel() {
+    return new ChangePasswordPanel();
+  }
+
+  private void layoutComponents() {
+    setLayout(new BorderLayout());
+    final JLabel label = new JLabel(new ImageIcon(Util.getBanner(title)));
+    add(label, BorderLayout.NORTH);
+
+    final JPanel main = new JPanel();
+    add(main, BorderLayout.CENTER);
+    main.setBorder(BorderFactory.createEmptyBorder(10, 10, 10, 10));
+    main.setLayout(new GridBagLayout());
+    main.add(
+        new JLabel("Password:"),
+        new GridBagConstraints(
+            0,
+            0,
+            1,
+            1,
+            0.0,
+            0.0,
+            GridBagConstraints.WEST,
+            GridBagConstraints.NONE,
+            new Insets(5, 0, 0, 0),
+            0,
+            0));
+    main.add(
+        passwordField,
+        new GridBagConstraints(
+            1,
+            0,
+            1,
+            1,
+            1.0,
+            0.0,
+            GridBagConstraints.WEST,
+            GridBagConstraints.HORIZONTAL,
+            new Insets(5, 5, 0, 0),
+            0,
+            0));
+    main.add(
+        new JLabel("Confirm Password:"),
+        new GridBagConstraints(
+            0,
+            1,
+            1,
+            1,
+            0.0,
+            0.0,
+            GridBagConstraints.WEST,
+            GridBagConstraints.NONE,
+            new Insets(5, 0, 0, 0),
+            0,
+            0));
+    main.add(
+        passwordConfirmField,
+        new GridBagConstraints(
+            1,
+            1,
+            1,
+            1,
+            1.0,
+            0.0,
+            GridBagConstraints.WEST,
+            GridBagConstraints.HORIZONTAL,
+            new Insets(5, 5, 0, 0),
+            0,
+            0));
+
+    final JPanel buttons = new JPanel();
+    add(buttons, BorderLayout.SOUTH);
+    buttons.setBorder(BorderFactory.createEmptyBorder(10, 5, 10, 5));
+    buttons.setLayout(new FlowLayout(FlowLayout.RIGHT, 5, 0));
+    buttons.add(okButton);
+    okButton.setEnabled(false);
+  }
+
+  private void setupListeners() {
+    okButton.addActionListener(e -> close());
+
+    SwingComponents.addEnterKeyListener(this, this::close);
+
+    DocumentListenerBuilder.attachDocumentListener(
+        passwordField, () -> okButton.setEnabled(validatePasswords()));
+    DocumentListenerBuilder.attachDocumentListener(
+        passwordConfirmField, () -> okButton.setEnabled(validatePasswords()));
+  }
+
+  private void close() {
+    if (validatePasswords() && dialog != null) {
+      dialog.dispose();
+    }
+  }
+
+  private boolean validatePasswords() {
+    return Arrays.equals(passwordField.getPassword(), passwordConfirmField.getPassword())
+        && passwordField.getPassword().length > 2;
+  }
+
+  /**
+   * Shows this panel in a modal dialog.
+   *
+   * @param parent The dialog parent window.
+   * @return New password entered by user, otherwise null if the window is closed.
+   */
+  @Nullable
+  public String show(final Window parent) {
+    dialog = new JDialog(JOptionPane.getFrameForComponent(parent), title, true);
+    dialog.getContentPane().add(this);
+    SwingComponents.addEscapeKeyListener(dialog, this::close);
+    dialog.pack();
+    dialog.setLocationRelativeTo(parent);
+    dialog.setVisible(true);
+    dialog.dispose();
+    dialog = null;
+    return validatePasswords() ? String.valueOf(passwordField.getPassword()) : null;
+  }
+}
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/login/ForgotPasswordPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/login/ForgotPasswordPanel.java
@@ -1,0 +1,151 @@
+package games.strategy.engine.lobby.client.login;
+
+import games.strategy.engine.lobby.PlayerNameValidation;
+import games.strategy.ui.Util;
+import java.awt.BorderLayout;
+import java.awt.FlowLayout;
+import java.awt.GridBagConstraints;
+import java.awt.GridBagLayout;
+import java.awt.Insets;
+import java.awt.Window;
+import javax.annotation.Nullable;
+import javax.swing.BorderFactory;
+import javax.swing.ImageIcon;
+import javax.swing.JButton;
+import javax.swing.JDialog;
+import javax.swing.JLabel;
+import javax.swing.JOptionPane;
+import javax.swing.JPanel;
+import javax.swing.JTextField;
+import org.triplea.swing.SwingComponents;
+
+/**
+ * Panel used to request a new temporary password. The panel displays a single text field for
+ * username.
+ */
+final class ForgotPasswordPanel extends JPanel {
+
+  /** Indicates how the user dismissed the dialog displaying the panel. */
+  public enum ReturnValue {
+    CANCEL,
+    OK
+  }
+
+  private final String title = "Forgot Password";
+  private @Nullable JDialog dialog;
+  private final JTextField userNameField = new JTextField();
+  private final JButton okButton = new JButton("OK");
+  private final JButton cancelButton = new JButton("Cancel");
+  private ReturnValue returnValue = ReturnValue.CANCEL;
+
+  private ForgotPasswordPanel() {
+    layoutComponents();
+    setupListeners();
+  }
+
+  /**
+   * Creates a new instance of the {@code CreateUpdateAccountPanel} class used to request a
+   * temporary password.
+   *
+   * @return A new {@code CreateUpdateAccountPanel}.
+   */
+  static ForgotPasswordPanel newForgotPasswordPanel() {
+    return new ForgotPasswordPanel();
+  }
+
+  private void layoutComponents() {
+    setLayout(new BorderLayout());
+    final JLabel label = new JLabel(new ImageIcon(Util.getBanner(title)));
+    add(label, BorderLayout.NORTH);
+
+    final JPanel main = new JPanel();
+    add(main, BorderLayout.CENTER);
+    main.setBorder(BorderFactory.createEmptyBorder(10, 10, 10, 10));
+    main.setLayout(new GridBagLayout());
+    main.add(
+        new JLabel("Username:"),
+        new GridBagConstraints(
+            0,
+            0,
+            1,
+            1,
+            0.0,
+            0.0,
+            GridBagConstraints.WEST,
+            GridBagConstraints.NONE,
+            new Insets(0, 0, 0, 0),
+            0,
+            0));
+    main.add(
+        userNameField,
+        new GridBagConstraints(
+            1,
+            0,
+            1,
+            1,
+            1.0,
+            0.0,
+            GridBagConstraints.WEST,
+            GridBagConstraints.HORIZONTAL,
+            new Insets(0, 5, 0, 0),
+            0,
+            0));
+
+    final JPanel buttons = new JPanel();
+    add(buttons, BorderLayout.SOUTH);
+    buttons.setBorder(BorderFactory.createEmptyBorder(10, 5, 10, 5));
+    buttons.setLayout(new FlowLayout(FlowLayout.RIGHT, 5, 0));
+    buttons.add(okButton);
+    buttons.add(cancelButton);
+  }
+
+  private void setupListeners() {
+    cancelButton.addActionListener(e -> close());
+    okButton.addActionListener(e -> okPressed());
+
+    SwingComponents.addEnterKeyListener(this, this::okPressed);
+  }
+
+  private void close() {
+    if (dialog != null) {
+      dialog.setVisible(false);
+    }
+  }
+
+  private void okPressed() {
+    if (!PlayerNameValidation.isValid(userNameField.getText())) {
+      JOptionPane.showMessageDialog(
+          this,
+          PlayerNameValidation.validate(userNameField.getText()),
+          "Invalid name",
+          JOptionPane.ERROR_MESSAGE);
+      return;
+    }
+
+    returnValue = ReturnValue.OK;
+    close();
+  }
+
+  /**
+   * Shows this panel in a modal dialog.
+   *
+   * @param parent The dialog parent window.
+   * @return {@link ReturnValue#OK} if the user clicks 'okay' button, otherwise {@link
+   *     ReturnValue#CANCEL}.
+   */
+  ReturnValue show(final Window parent) {
+    dialog = new JDialog(JOptionPane.getFrameForComponent(parent), title, true);
+    dialog.getContentPane().add(this);
+    SwingComponents.addEscapeKeyListener(dialog, this::close);
+    dialog.pack();
+    dialog.setLocationRelativeTo(parent);
+    dialog.setVisible(true);
+    dialog.dispose();
+    dialog = null;
+    return returnValue;
+  }
+
+  String getUserName() {
+    return userNameField.getText();
+  }
+}
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/login/LobbyLogin.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/login/LobbyLogin.java
@@ -2,19 +2,16 @@ package games.strategy.engine.lobby.client.login;
 
 import games.strategy.engine.framework.GameRunner;
 import games.strategy.engine.lobby.client.LobbyClient;
-import games.strategy.net.ClientMessenger;
+import games.strategy.net.ClientMessengerFactory;
 import games.strategy.net.CouldNotLogInException;
+import games.strategy.net.IClientMessenger;
 import games.strategy.net.IMessenger;
 import games.strategy.net.MacFinder;
 import java.awt.Window;
 import java.io.IOException;
-import java.util.HashMap;
-import java.util.Map;
 import javax.annotation.Nullable;
 import javax.swing.JOptionPane;
-import org.triplea.lobby.common.LobbyConstants;
-import org.triplea.lobby.common.login.LobbyLoginResponseKeys;
-import org.triplea.lobby.common.login.RsaAuthenticator;
+import org.triplea.swing.DialogBuilder;
 
 /**
  * The client side of the lobby authentication protocol.
@@ -47,11 +44,16 @@ public class LobbyLogin {
 
   private @Nullable LobbyClient login(final LoginPanel panel) {
     try {
-      final IMessenger messenger =
+      final IClientMessenger messenger =
           GameRunner.newBackgroundTaskRunner()
               .runInBackgroundAndReturn(
                   "Connecting to lobby...",
-                  () -> login(panel.getUserName(), panel.getPassword(), panel.isAnonymousLogin()),
+                  () ->
+                      panel.isAnonymousLogin()
+                          ? ClientMessengerFactory.newAnonymousUserMessenger(
+                              lobbyServerProperties, panel.getUserName())
+                          : ClientMessengerFactory.newRegisteredUserMessenger(
+                              lobbyServerProperties, panel.getUserName(), panel.getPassword()),
                   IOException.class);
       panel.getLobbyLoginPreferences().save();
       return new LobbyClient(messenger, panel.isAnonymousLogin());
@@ -65,27 +67,6 @@ public class LobbyLogin {
       Thread.currentThread().interrupt();
       return null;
     }
-  }
-
-  private IMessenger login(
-      final String username, final String password, final boolean anonymousLogin)
-      throws IOException {
-    return new ClientMessenger(
-        lobbyServerProperties.getHost(),
-        lobbyServerProperties.getPort(),
-        username,
-        MacFinder.getHashedMacAddress(),
-        challenge -> {
-          final Map<String, String> response = new HashMap<>();
-          if (anonymousLogin) {
-            response.put(LobbyLoginResponseKeys.ANONYMOUS_LOGIN, Boolean.TRUE.toString());
-          } else {
-            response.putAll(RsaAuthenticator.newResponse(challenge, password));
-          }
-          response.put(
-              LobbyLoginResponseKeys.LOBBY_VERSION, LobbyConstants.LOBBY_VERSION.toString());
-          return response;
-        });
   }
 
   private static String playerMacIdString() {
@@ -107,6 +88,9 @@ public class LobbyLogin {
         return null;
       case CREATE_ACCOUNT:
         return createAccount();
+      case FORGOT_PASSWORD:
+        forgotPassword();
+        return null;
       default:
         throw new AssertionError("unknown login panel return value: " + returnValue);
     }
@@ -131,7 +115,12 @@ public class LobbyLogin {
           GameRunner.newBackgroundTaskRunner()
               .runInBackgroundAndReturn(
                   "Connecting to lobby...",
-                  () -> createAccount(panel.getUserName(), panel.getPassword(), panel.getEmail()),
+                  () ->
+                      ClientMessengerFactory.newCreateAccountMessenger(
+                          lobbyServerProperties,
+                          panel.getUserName(),
+                          panel.getEmail(),
+                          panel.getPassword()),
                   IOException.class);
       panel.getLobbyLoginPreferences().save();
       return new LobbyClient(messenger, false);
@@ -147,21 +136,40 @@ public class LobbyLogin {
     }
   }
 
-  private IMessenger createAccount(final String username, final String password, final String email)
-      throws IOException {
-    return new ClientMessenger(
-        lobbyServerProperties.getHost(),
-        lobbyServerProperties.getPort(),
-        username,
-        MacFinder.getHashedMacAddress(),
-        challenge -> {
-          final Map<String, String> response = new HashMap<>();
-          response.put(LobbyLoginResponseKeys.REGISTER_NEW_USER, Boolean.TRUE.toString());
-          response.put(LobbyLoginResponseKeys.EMAIL, email);
-          response.putAll(RsaAuthenticator.newResponse(challenge, password));
-          response.put(
-              LobbyLoginResponseKeys.LOBBY_VERSION, LobbyConstants.LOBBY_VERSION.toString());
-          return response;
-        });
+  private void forgotPassword() {
+    final ForgotPasswordPanel forgotPasswordPanel = ForgotPasswordPanel.newForgotPasswordPanel();
+    final ForgotPasswordPanel.ReturnValue returnValue = forgotPasswordPanel.show(parentWindow);
+    switch (returnValue) {
+      case OK:
+        forgotPassword(forgotPasswordPanel);
+        return;
+      case CANCEL:
+        return;
+      default:
+        throw new AssertionError("unknown forgot password panel return value: " + returnValue);
+    }
+  }
+
+  private void forgotPassword(final ForgotPasswordPanel panel) {
+    try {
+
+      GameRunner.newBackgroundTaskRunner()
+          .runInBackgroundAndReturn(
+              "Sending forgot password request...",
+              () ->
+                  ClientMessengerFactory.newForgotPasswordMessenger(
+                      lobbyServerProperties, panel.getUserName()),
+              IOException.class);
+
+      DialogBuilder.builder()
+          .parent(parentWindow)
+          .title("Temporary Password Sent")
+          .infoMessage("A temporary password has been sent to your email.")
+          .showDialog();
+    } catch (final IOException e) {
+      showError("Error", "Failed to generate a temporary password: " + e.getMessage());
+    } catch (final InterruptedException e) {
+      Thread.currentThread().interrupt();
+    }
   }
 }
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/login/LoginPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/login/LoginPanel.java
@@ -27,7 +27,8 @@ final class LoginPanel extends JPanel {
   enum ReturnValue {
     CANCEL,
     LOGON,
-    CREATE_ACCOUNT
+    CREATE_ACCOUNT,
+    FORGOT_PASSWORD
   }
 
   private @Nullable JDialog dialog;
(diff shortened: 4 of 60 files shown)


game-core/src/main/java/games/strategy/engine/lobby/client/login/ChangePasswordPanel.java
```diff
@@ -0,0 +1,161 @@
...
+
+    DocumentListenerBuilder.attachDocumentListener(
+        passwordField, () -> okButton.setEnabled(validatePasswords()));
+    DocumentListenerBuilder.attachDocumentListener(
+        passwordConfirmField, () -> okButton.setEnabled(validatePasswords()));
+  }
+
+  private void close() {
+    if (validatePasswords() && dialog != null) {
+      dialog.dispose();
```
RoiEXLab: Hmm this seems like a smell to me.
What about registering listeners when the dialog is shown, and unregistering them afterwards?

game-core/src/main/java/games/strategy/engine/lobby/client/login/ForgotPasswordPanel.java
```diff
@@ -0,0 +1,151 @@
...
+  ReturnValue show(final Window parent) {
+    dialog = new JDialog(JOptionPane.getFrameForComponent(parent), title, true);
+    dialog.getContentPane().add(this);
+    SwingComponents.addEscapeKeyListener(dialog, this::close);
+    dialog.pack();
+    dialog.setLocationRelativeTo(parent);
+    dialog.setVisible(true);
+    dialog.dispose();
+    dialog = null;
+    return returnValue;
```
RoiEXLab: I'm a little confused I'd have assumed JOptionPane in combination with your swing helper library would reduce the amount of code?

game-core/src/main/java/games/strategy/engine/lobby/client/login/ForgotPasswordPanel.java
```diff
@@ -0,0 +1,151 @@
...
+  private void setupListeners() {
+    cancelButton.addActionListener(e -> close());
+    okButton.addActionListener(e -> okPressed());
+
+    SwingComponents.addEnterKeyListener(this, this::okPressed);
+  }
+
+  private void close() {
+    if (dialog != null) {
+      dialog.setVisible(false);
```
RoiEXLab: Same here with the dialog.
I feel like there must be a better way.
Interestingly enough here the dialog is only made invisible rather than being disposed

game-core/src/main/java/games/strategy/engine/lobby/client/login/LobbyLogin.java
```diff
@@ -147,21 +136,40 @@ private void showError(final String title, final String message) {
...
-          return response;
-        });
+  private void forgotPassword() {
+    final ForgotPasswordPanel forgotPasswordPanel = ForgotPasswordPanel.newForgotPasswordPanel();
+    final ForgotPasswordPanel.ReturnValue returnValue = forgotPasswordPanel.show(parentWindow);
+    switch (returnValue) {
+      case OK:
+        forgotPassword(forgotPasswordPanel);
+        return;
+      case CANCEL:
```
RoiEXLab: I get why you're using enums at this point, but I feel like just having a simple boolean would simplify this code a lot.

game-core/src/main/java/games/strategy/engine/lobby/client/login/LoginPanel.java
```diff
@@ -27,15 +27,17 @@
   enum ReturnValue {
     CANCEL,
     LOGON,
```
RoiEXLab: I know, not your fault but:
Shouldn't this be called LOGIN?

game-core/src/main/java/games/strategy/net/ClientMessenger.java
```diff
@@ -212,8 +228,9 @@ public boolean isServer() {
 
   @Override
   public void socketUnqaurantined(
-      final SocketChannel channel, final QuarantineConversation converstaion2) {
-    final ClientQuarantineConversation conversation = (ClientQuarantineConversation) converstaion2;
+      final SocketChannel channel, final QuarantineConversation quarantineConversation) {
+    final ClientQuarantineConversation conversation =
+        (ClientQuarantineConversation) quarantineConversation;
```
RoiEXLab: I feel like this might be a good use case for the var keyword

game-core/src/main/java/games/strategy/net/LocalNoOpMessenger.java
```diff
@@ -64,6 +65,12 @@ public ILoginValidator getLoginValidator() {
     return null;
   }
 
+  @Override
+  public void setForgotPasswordConversation(
+      final ForgotPasswordConversation forgotPasswordConversation) {
+    throw new UnsupportedOperationException("Password reset not supported by local servers.");
```
RoiEXLab: Isn't there some sort of interface for the lobby or something?
Everything but the lobby shouldn't have support for that

game-core/src/main/java/games/strategy/net/nio/ClientQuarantineConversation.java
```diff
@@ -130,6 +132,12 @@ public Action message(final Serializable serializable) {
           }
           localName = strings[0];
           serverName = strings[1];
+          if (strings.length > 2) {
+            passwordChangeRequired =
+                strings[2] != null
+                    && strings[2].equals(ServerQuarantineConversation.CHANGE_PASSWORD);
+          }
```
RoiEXLab: Not blaming you for this, but I have no idea in the slightest what's happening here

game-core/src/main/java/games/strategy/net/nio/ServerQuarantineConversation.java
```diff
@@ -77,20 +87,30 @@ public Action message(final Serializable serializable) {
...
-                            remoteName,
-                            remoteMac,
-                            channel.socket().getRemoteSocketAddress()))
-                    .orElseGet(() -> PlayerNameValidation.serverSideValidate(remoteName));
-            send(error);
-            if (error != null) {
+            if (response.containsKey(LobbyLoginResponseKeys.FORGOT_PASSWORD)) {
+              send(
+                  forgotPasswordConversation.handle(channel.socket().getInetAddress(), remoteName));
               step = Step.ACK_ERROR;
```
RoiEXLab: The more I see this code the more I feel like this is unneccessarily complex

game-core/src/main/java/games/strategy/net/nio/ServerQuarantineConversation.java
```diff
@@ -101,7 +121,7 @@ public Action message(final Serializable serializable) {
                     remoteName, channel.socket().getInetAddress(), serverMessenger.getNodes());
           }
           // send the node its assigned name and our name
-          send(new String[] {remoteName, serverMessenger.getLocalNode().getName()});
+          send(new String[] {remoteName, serverMessenger.getLocalNode().getName(), error});
```
RoiEXLab: No idea what's happening here either

lobby/src/main/java/org/triplea/lobby/server/login/forgot/password/create/PasswordEmailSender.java
```diff
@@ -0,0 +1,57 @@
...
+import org.triplea.lobby.server.EnvironmentVariable;
+
+/** Sends a temporary password to a target user. */
+class PasswordEmailSender implements BiConsumer<String, String> {
+  private static final String FROM = "no-reply@triplea-game.org";
+
+  @Override
+  public void accept(final String email, final String generatedPassword) {
+    if (EnvironmentVariable.LOCAL_DEV.getBoolean()) {
+      // do not send emails from local dev.
```
RoiEXLab: I had good experiences with https://ethereal.email/ which is a no-op SMTP server, you can use to verify everything works.
You might want to be allow configuring such a service here

lobby/src/main/java/org/triplea/lobby/server/login/forgot/password/create/PasswordEmailSender.java
```diff
@@ -0,0 +1,57 @@
...
+import javax.mail.MessagingException;
+import javax.mail.Session;
+import javax.mail.Transport;
+import javax.mail.internet.InternetAddress;
+import javax.mail.internet.MimeMessage;
+import org.triplea.lobby.server.EnvironmentVariable;
+
+/** Sends a temporary password to a target user. */
+class PasswordEmailSender implements BiConsumer<String, String> {
+  private static final String FROM = "no-reply@triplea-game.org";
```
RoiEXLab: Is this actually a valid email address?
I mean SMTP allows any address, but that might cause the mails to end up in the spam filter.

lobby/src/main/java/org/triplea/lobby/server/login/forgot/password/create/PasswordEmailSender.java
```diff
@@ -0,0 +1,57 @@
...
+
+    final Properties props = new Properties();
+    final Session session = Session.getDefaultInstance(props, null);
+
+    try {
+      final Message message = new MimeMessage(session);
+      message.setFrom(new InternetAddress(FROM, "no-reply"));
+      message.addRecipient(Message.RecipientType.TO, new InternetAddress(email, email));
+      message.setSubject("Temporary Password");
+      message.setText(createMailBody(generatedPassword));
```
RoiEXLab: Perhaps out of scope for this PR but plain text is really ugly.
This should really be a multipart email.
Have a look at some simple templates I created for the NodeJS replacement of MARTI: https://github.com/triplea-game/dice-server-js/tree/master/public

lobby/src/main/java/org/triplea/lobby/server/login/forgot/password/create/PasswordEmailSender.java
```diff
@@ -0,0 +1,57 @@
...
+      Transport.send(message);
+    } catch (final MessagingException e) {
+      throw new EmailSendFailure(e);
+    } catch (final UnsupportedEncodingException e) {
```
RoiEXLab: Didn't miss this Exception at all. It's a good thing most classes work using StandardCharsets these days, but I still wonder where exactly this Exception is coming from

lobby/src/main/java/org/triplea/lobby/server/login/forgot/password/create/TempPasswordPersistence.java
```diff
@@ -0,0 +1,40 @@
...
+        pass -> new HashedPassword(RsaAuthenticator.hashPasswordWithSalt(pass)),
+        hashedPass -> BCrypt.hashpw(hashedPass.value, BCrypt.gensalt()));
+  }
+
+  boolean storeTempPassword(final String username, final String generatedPassword) {
+    tempPasswordDao.invalidateTempPasswords(username);
+
+    final HashedPassword hashedPass = passwordHasher.apply(generatedPassword);
+    final String tempPass = hashedPasswordBcrypter.apply(hashedPass);
+    return tempPasswordDao.insertTempPassword(username, tempPass);
```
RoiEXLab: To be fair I don't think it's necessary to hash a random token that doesn't originate from user input and expires after a short amount of time, but if you want to treat it like a password to keep all the other logic the same that's fine too I guess.

lobby/src/main/java/org/triplea/lobby/server/login/forgot/password/verify/TempPasswordVerification.java
```diff
@@ -0,0 +1,36 @@
...
+  public boolean checkTempPassword(final String username, final String hashedPassword) {
+    final String tempPassword = tempPasswordDao.fetchTempPassword(username).orElse(null);
+    if (tempPassword == null) {
+      return false;
+    }
+
+    if (!BCrypt.checkpw(hashedPassword, tempPassword)) {
+      return false;
+    }
+    tempPasswordDao.invalidateTempPasswords(username);
```
RoiEXLab: Maybe that's just me because I always think of password resets in terms of tokens.
But shouldn't the password be invalidated after using it in either case?

lobby/src/test/java/org/triplea/lobby/server/login/LobbyLoginValidatorIntegrationTest.java
```diff
@@ -134,13 +147,24 @@ void testLogin() {
...
-            .apply(challenge -> new HashMap<>(RsaAuthenticator.newResponse(challenge, "wrong"))),
+            .apply(
+                challenge -> {
+                  final Map<String, String> map = new HashMap<>();
+                  map.put(
+                      LobbyLoginResponseKeys.RSA_ENCRYPTED_PASSWORD,
+                      RsaAuthenticator.encrpytPassword(
+                          challenge.get(LobbyLoginChallengeKeys.RSA_PUBLIC_KEY), "wrong"));
+                  return map;
+                }),
```
RoiEXLab: Collections.singletonMap ?

lobby/src/test/java/org/triplea/lobby/server/login/forgot/password/create/PasswordGeneratorTest.java
```diff
@@ -0,0 +1,33 @@
...
+   */
+  @Test
+  void verifyPasswordChanges() {
+    final var passwordGenerator = new PasswordGenerator();
+
+    final Set<String> strings = new HashSet<>();
+    for (int i = 0; i < 1000; i++) {
+      final String generated = passwordGenerator.generatePassword();
+      assertThat(strings.contains(generated), is(false));
+      strings.add(generated);
```
RoiEXLab: Imagine this test failing once in the next 100 years or something ^^

lobby/src/main/java/org/triplea/lobby/server/login/LobbyLoginValidator.java
```diff
@@ -181,6 +190,11 @@ private static String getBanDurationBreakdown(final Timestamp stamp) {
           response,
           pass -> {
             if (hashedPassword.isBcrypted()) {
+              // TODO: update tests to verify tempPasswordVerification branch.
```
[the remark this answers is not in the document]
DanVanAtta: yes, it's still valid; I do not plan on doing this in this PR. With luck some break-up of the login validator will happen and it'll become more trivial to test.

lobby/src/main/java/org/triplea/lobby/server/login/LobbyLoginValidator.java
```diff
@@ -181,6 +190,11 @@ private static String getBanDurationBreakdown(final Timestamp stamp) {
...
+              if (tempPasswordVerification.checkTempPassword(username, pass)) {
+                return ServerQuarantineConversation.CHANGE_PASSWORD;
+              }
+
```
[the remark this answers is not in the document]
DanVanAtta: I'm not sure if I follow. If we return a null string, that will look like a success response back to the client. In essence we're returning a magic string here that the client would act on to trigger the change password prompt.

game-core/src/main/java/games/strategy/net/ClientMessengerFactory.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import org.triplea.lobby.common.LobbyConstants;
+import org.triplea.lobby.common.login.LobbyLoginChallengeKeys;
+import org.triplea.lobby.common.login.LobbyLoginResponseKeys;
+import org.triplea.lobby.common.login.RsaAuthenticator;
+
+/** Default implementation of {@link IClientMessenger}. */
```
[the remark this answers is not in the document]
DanVanAtta: Indeed, copy/paste error. I've fixed it in: 25609a4c1

game-core/src/main/java/games/strategy/engine/lobby/client/login/ChangePasswordPanel.java
DanVanAtta: We only really need to make the dialog not visible. The dialog creation code is notable, after we set the dialog to visible it is blocking until no longer visible. The next line then disposes the dialog. 

I updated this method to be a more simple 'not visible': ed4157d2c
Checking out the dispose code there is quite a lot there, probably better then to not 'double dispose'.

Overall it is a pattern for swing JDialog to work like this, essentially to have a sequence like:
```
dialog.setVisible(true);
dialog.dispose();
```

game-core/src/main/java/games/strategy/engine/lobby/client/login/ForgotPasswordPanel.java
DanVanAtta: It very likely would have. Compare the the two new panel classes to: (CreateUpdateAccountPanel)[https://github.com/triplea-game/triplea/blob/master/game-core/src/main/java/games/strategy/engine/lobby/client/login/CreateUpdateAccountPanel.java] and you'll see a lot of similarity. So in essence we're looking at glorified copy/pastes. It's not consistent if these panels deviate too much in their implementation.

game-core/src/main/java/games/strategy/engine/lobby/client/login/ForgotPasswordPanel.java
DanVanAtta: Turns out this is okay, the dispose calll happens after the panel is no longer visible. `setVisible(true)` on a dialog blocks until it is no longer visible, if you hunt down the creation code near the bottom of the class, the dispose is there.

game-core/src/main/java/games/strategy/engine/lobby/client/login/LobbyLogin.java
DanVanAtta: In part, the account creation panel uses enums, so I want to try and stay consistent with that implementation. I'm not sure if a boolean return value is actually that much clear or not. Sure it's a few lines less code, but less explicit. I think perhaps something of a wash either way, in which case the consistency with the copy/pasted implementation wins the argument for me.

game-core/src/main/java/games/strategy/engine/lobby/client/login/LoginPanel.java
DanVanAtta: Probably even LOGGED_IN maybe.

game-core/src/main/java/games/strategy/net/ClientMessenger.java
DanVanAtta: Sure: 3fb3d8e8a
The real change here is to fix a spelling error in the parameter. That probably should have been done in a different PR. It's not good practice to throw in random refactorings when adding a feature, particularly when deeper and more meaningful fixes/refactorings are more called for to really fix some of the problems we have.

While fixing spelling, the method name is misspelled, fixed in: c5c340d752c9a76e9d1fa372183378655bdf18f9

game-core/src/main/java/games/strategy/net/LocalNoOpMessenger.java
DanVanAtta: The server code is shared between lobby and 'client' servers, people hosting locally. There is a client login validator which checks if a user is 'minibanned', ie: in-memory banned, and the lobby has a more full-blown validation that checks passwords against DB. The 'set' method here is almost something of a callback so that in lobby sub-project we can create the server quarantine conversation and then inject lobby specific code.

The forgot password module is tucked behind the login validation. While thinking about that, it's probably better to add a NPE check for the forgot password conversation: 9e963529b

Considering it for a moment longer, a NPE check and no-op is a silent bug, probably best to be an assertion even: e0e837905

This code in: ServerModel.java is pretty important to see, it's the client server that instantiates the login module. Notice the circular dependency between server messenger and the validator, that prevents us from injecting the validator as a constructor arg.
```
      this.serverMessenger =
          new ServerMessenger(props.getName(), props.getPort(), objectStreamFactory);
      final ClientLoginValidator clientLoginValidator = new ClientLoginValidator(serverMessenger);
      clientLoginValidator.setGamePassword(props.getPassword());
      serverMessenger.setLoginValidator(clientLoginValidator);
```

game-core/src/main/java/games/strategy/net/nio/ClientQuarantineConversation.java
DanVanAtta: After successful login the server sends back an array with the client name:

ServerQuarantineConversation.java:128
```
          // send the node its assigned name and our name
          send(new String[] {remoteName, serverMessenger.getLocalNode().getName(), error});
          // send the node its and our address as we see it
          send(
              new InetSocketAddress[] {
                (InetSocketAddress) channel.socket().getRemoteSocketAddress(),
                serverMessenger.getLocalNode().getSocketAddress()
              });
```

New to this update is the third element in the array "error". Typically that will be null, in the change password case it'll be set to the magic string "CHANGE_PASSWORD".

So here we have the client code parsing out the third array element to see if it contains the magic string. Appending to the array avoids having to do deeper changes to the quarantine conversation flow.

game-core/src/main/java/games/strategy/net/nio/ServerQuarantineConversation.java
DanVanAtta: You're telling me :smile: 
This code is, I think, some of the oldest and basically is not touched because it's brittle, complex, and very important.

game-core/src/main/java/games/strategy/net/nio/ServerQuarantineConversation.java
DanVanAtta: Server side counterpart of the part that was confusing, see: https://github.com/triplea-game/triplea/pull/4991#discussion_r314853517

lobby/src/main/java/org/triplea/lobby/server/login/forgot/password/create/PasswordEmailSender.java
DanVanAtta: That's a pretty neat service. If it were simply a matter of swapping out the configured host, that would be appealing. On the other hand if we have to add any additional code to work with it, then we're no longer really testing the original mail sending code and have introduced a new liability and point of complexity that is not actually verifying the prod code. It's also a benefit to avoid network calls in local dev.

The flag is kinda hacky, but it is straight forward and effective. If we can overcome the problems I stated above, and configure this cleanly (and we'd still need a flag to turn off the network communication), I wouldn't necessarily have objection. IMO it's better to focus on delivering the feature for now and revisiting later as we have time and desire.

lobby/src/main/java/org/triplea/lobby/server/login/forgot/password/create/PasswordEmailSender.java
DanVanAtta: AFAIK it is valid; Sending a test mail, the mail did not land in spam folder. Anything specific to say it's not valid?

lobby/src/test/java/org/triplea/lobby/server/login/forgot/password/create/PasswordGeneratorTest.java
DanVanAtta: I suspect probably even more rare than 100 years. If it were to fail, I'd guess a dev probably would ignore it.. Ideally not test failure is ignored, they all should have some work put in to make them deterministic or fixed otherwise. 

There is non-determinism in this test, but it's almost a good thing as we increase test coverage each time we run the test. I added a note recommending how to correct the code if we see a failure: e182b430a

lobby/src/test/java/org/triplea/lobby/server/login/LobbyLoginValidatorIntegrationTest.java
DanVanAtta: The test mutates the collection that is returned, line 59:
```
    return responseGetter -> {
      final Map<String, String> response = responseGetter.apply(challenge);
      response.putIfAbsent(LobbyLoginResponseKeys.EMAIL, EMAIL);
      response.putIfAbsent(
          LobbyLoginResponseKeys.LOBBY_VERSION, LobbyConstants.LOBBY_VERSION.toString());
      return loginValidator.verifyConnection(challenge, response, name, mac, address);
    };
```
This test could use some work to be more explicit, probably the map should be a constant instead of something derived by a function. Less DRY and more copy/paste would be good for this test IMO.

lobby/src/main/java/org/triplea/lobby/server/login/forgot/password/verify/TempPasswordVerification.java
DanVanAtta: Slightly confused what the 'either' case refers to. The password here is invalidated if it is used successfully. If we invalidated it on the negative case, then I could try to login as someone else and their temp password would suddenly be invalid. Also if there is a copy/paste mistake and some extra whitespace is captured, a user could invalidate their temp password mistakenly.

I think then this code is correct, let me know  if still see it otherwise.

lobby/src/main/java/org/triplea/lobby/server/login/forgot/password/create/TempPasswordPersistence.java
DanVanAtta: There actually was no expiration. I added in one here: f23c3e5ae694b1e5f8ffe457ae67ad831603a5a8

We would want to treat the temp token with the same security. If someone can read DB values, and we had a weaker hashing strategy, it'd be an attack vector. Instead of cracking the password hashes, someone would just need to request a password reset and then they could crack the easier hash instead.

lobby/src/main/java/org/triplea/lobby/server/login/forgot/password/create/PasswordEmailSender.java
DanVanAtta: Lines 31 + 32:

```
      message.setFrom(new InternetAddress(FROM, "no-reply"));
      message.addRecipient(Message.RecipientType.TO, new InternetAddress(email, email));
```

DanVanAtta: Specifically the `InternetAddress` constructor

lobby/src/main/java/org/triplea/lobby/server/login/forgot/password/create/PasswordEmailSender.java
DanVanAtta: The email should not be frequent, IMO a short plain text looked okay enough to me :smile: 

If we want to improve it, sure :man_shrugging: 
The testing would need to be repeated though; after banging my head against encryption for around 9 hours, and banging my head against the client/server quarantine conversation for about 5, I'm very ready to move on :smiley:

game-core/src/main/java/games/strategy/engine/lobby/client/login/LobbyLogin.java
RoiEXLab: What about checking if the return value is == to OK ?
This way you get to keep the enum and I get the simplicity of a simple if statement ^^

game-core/src/main/java/games/strategy/net/nio/ServerQuarantineConversation.java
RoiEXLab: > brittle, complex, and very important

Legacy Code in a nutshell 😄

lobby/src/main/java/org/triplea/lobby/server/login/forgot/password/create/PasswordEmailSender.java
RoiEXLab: Let me clarify:
I wasn't aware that triplea-game.org is a valid mail address.
AFAIK when the source email server and the from email address don't match, common email servers just put them directly into the spam folder, but I'm not sure how accurate this information is.

Basically I'm wondering what exactly the java mail API does. Is it just talking to some SMTP server somewhere, or is it actually an SMTP server implementation itself that just sends mails directly to other POP3 or IMAP servers?

lobby/src/main/java/org/triplea/lobby/server/login/forgot/password/verify/TempPasswordVerification.java
RoiEXLab: From your perspective the code is indeed correct.
AFAIK common password reset mechanisms send you a link with a token as GET parameter, where it makes perfectly sense to invalidate the actual token when a false token is issued. You possibly can't mistype a hyperlink.
For the password on the other hand it does make sense, as you correctly explained.

game-core/src/main/java/games/strategy/engine/lobby/client/login/LobbyLogin.java
DanVanAtta: At that point the enum does not make sense as you mentioned. The switch statement does not bother me.

The following does not read well:
```
if(forgotPasswordPanel.show(parentWindow)) {
:
:
}
```

We'd have to refactor the `show` method to be called something else. That would then break the symmetry with the "peer" panel classes. The ROI here I do not think is favorable here. Given that I don't mind the switch statement at all, for me the ROI is negative.

lobby/build.gradle
```diff
@@ -12,6 +12,8 @@ dependencies {
     implementation project(':java-extras')
     implementation project(':game-core')
     implementation project(':lobby-db-dao')
+
+    implementation 'javax.mail:mail:1.5.0-b01'
```
RoiEXLab: I'm not sure about this, but I believe there's a newer, maintained version of this:
https://search.maven.org/artifact/jakarta.mail/jakarta.mail-api/1.6.3/jar
Looks like this library was taken over by the eclipse foundation. Telling by the GitHub repository it looks very similar: https://github.com/eclipse-ee4j/mail

game-core also seems to depend on a similar version of the API: `com.sun.mail:javax.mail:1.6.2` unifying all the versions would be desireable.

lobby/src/main/java/org/triplea/lobby/server/login/forgot/password/create/PasswordEmailSender.java
DanVanAtta: > AFAIK when the source email server and the from email address don't match, common email servers just put them directly into the spam folder, but I'm not sure how accurate this information is

That really depends on the email provider. I wonder how accurate that rule works as it's possible to have email relays. Mail from `no-replay@amazon.com` for example I would confidently bet does not come from a `amazon.com` server. That mail would come from some random ec2 instance that would probably have `amazon.com` as a domain suffix, but it would not be `amazon.com` proper.


---------

> Is it just talking to some SMTP server somewhere

Basically. It can be configured to do more.

game-core/src/main/java/games/strategy/engine/lobby/client/login/LobbyLogin.java
RoiEXLab: I was actually thiking of this:
```java
if (forgotPasswordPanel.show(parentWindow) == ReturnValue.OK) {

}
```

lobby/src/main/java/org/triplea/lobby/server/login/forgot/password/verify/TempPasswordVerification.java
DanVanAtta: Indeed! That hyperlink would need to go to a public website with a password reset form. That seemed like a lot more infrastructure and moving parts then issuing a temporary password.

Issuing a temporary password is also not that uncommon, so I don't think we are necessarily doing anything too novel here either.

lobby/build.gradle
DanVanAtta: That is `mail-api`; I recall reading something that if you want to actually send mail, you need the implementation dependency. Given the testing done, I'm not sure if there is any concrete benefit to going for a different version.

game-core/src/main/java/games/strategy/engine/lobby/client/login/LobbyLogin.java
DanVanAtta: That is not too bad, I like the suggestion. I scanned the class, and we have this:
```
  private @Nullable LobbyClient loginToServer() {
    final LoginPanel loginPanel = new LoginPanel(LobbyLoginPreferences.load());
    final LoginPanel.ReturnValue returnValue = loginPanel.show(parentWindow);
    switch (returnValue) {
      case LOGON:
        return login(loginPanel);
      case CANCEL:
        return null;
      case CREATE_ACCOUNT:
        return createAccount();
      case FORGOT_PASSWORD:
        forgotPassword();
        return null;
      default:
        throw new AssertionError("unknown login panel return value: " + returnValue);
    }
  }
```

The other 3 switch statements could be simplified, but for now they are consistent with the first switch block. That makes me think it's better to not mix the patterns.

lobby/build.gradle
RoiEXLab: From what I can tell the newer version doesn't throw UnsupportedEncodingException

lobby/src/test/java/org/triplea/lobby/server/login/LobbyLoginValidatorIntegrationTest.java
RoiEXLab: Well, you could wrap it in a HashMap to get a one liner, but whatever not that important anyways.
This is one of the cases where C# has superior language features like initializers that would allow such things to be done per implementation

lobby/build.gradle
DanVanAtta: Removing exactly 2 lines for the catch in a small class strikes me as a very marginal benefit.

Is there a non 'mail-api' version? Is this worth the research and effort we are putting into it?

DanVanAtta: (And worth the re-testing)

RoiEXLab: Nevermind doesn't seem to actually be the case :(

DanVanAtta: > com.sun.mail:javax.mail:1.6.2 unifying all the versions would be desireable.

I would agree to reducing dependencies and keeping them the same. I'll re-test with the com.sun dependency. I'm a bit surprised though we can / should use a `com.sun` given the license we use.

DanVanAtta: Actually...  Are we *sure* that `com.sun.mail` is a cool dependency to have??

I could see us getting away with that as a server side dependency, but if we distribute that with the game-core, it could become a bigger deal.

DanVanAtta: Looks like the com.sun.mail library is GPLv2, should be okay then. I still have my test program available, can verify in 5 minutes...

RoiEXLab: Well I think its desireable to use a library that is well-maintained, even if we don't get any large benefits from it.
Not necessarily as part of this PR, but as a follow-up PR I think it would make sense to bump the versions to the latest jakarta version. IMO it's not that unlikely that this would work right out of the box because it doesn't look like the API changed at all.

game-core/src/main/java/games/strategy/engine/lobby/client/login/LoginPanel.java
RoiEXLab: Any intentions to change it as part of this PR or out of scope in your opinion?

lobby/src/main/java/org/triplea/lobby/server/login/LobbyLoginValidator.java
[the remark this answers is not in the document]
RoiEXLab: Ah, I shoud've clarified I was reffering to the password column in the "user table", so we force a migration to the new system this way, but wouldn't actually have to use some sort of migration logic.
This would allow us to drop the column entirely and just add a non-null constraint for new values and just fill the other passwords with randomly generated ones so we don't have any null values there or something

lobby/build.gradle
DanVanAtta: Updated package, re-testing looked good: 3227d4781a83f005a612621ddaccd3f922b2eec9

game-core/src/main/java/games/strategy/engine/lobby/client/login/LoginPanel.java
DanVanAtta: Out of scope, no need to change it in this branch

lobby/src/main/java/org/triplea/lobby/server/login/LobbyLoginValidator.java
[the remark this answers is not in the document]
DanVanAtta: I think we're on the same page with forgot password being a nice feature to help with the migration.

For any user where we use a dummy password, it would be good to trigger the forgot password feature, generate a temp password and send an email. We should perhaps plan that migration though.

One point to be sure about, there is no verification of whose account you request for password reset. It would be bad to disable a password when a temp password is requested, it could be used to annoy someone and invalidate their password. That is a vector for abuse of the system, it's one reason for the audit table to store inet address so we can ban those users if we get reports about abuse.

lobby/build.gradle
RoiEXLab: Nice 👍 

If you have the time, consider checking if https://github.com/eclipse-ee4j/mail/releases compiles, just to be sure this is actually the same library with a new owner.
If it compiles might be worth updating at some point in the future

PR #4994, opened 2019-08-16
Refactor: games.stratety.triplea reduce scope and use lombok

DanVanAtta:
- Replace getters and setters with lombok
- Reduce method and variable scopes


game-core/src/main/java/games/strategy/triplea/ui/AbstractUiContext.java
```diff
@@ -28,32 +30,30 @@
 @Log
 public abstract class AbstractUiContext implements UiContext {
 
-  protected static final String UNIT_SCALE_PREF = "UnitScale";
```
DanVanAtta: Decreased scope lead to re-ordering, nothing removed or added

PR #4995, opened 2019-08-16
Couple Automated Code Fixes (IDEA Analysis)

DanVanAtta:
## Overview
Automated fixes found by a couple of static analysis tools

## Functional Changes
None

## Manual Testing Performed
Hosted a game, connected to it locally, played through a round with a couple of AIs

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/TripleAPlayer.java
+++ b/game-core/src/main/java/games/strategy/triplea/TripleAPlayer.java
@@ -711,10 +711,10 @@ public abstract class TripleAPlayer extends AbstractHumanPlayer implements ITrip
   }
 
   @Override
-  public boolean confirmMoveHariKari() {
+  public void confirmMoveHariKari() {
     final String question =
         "All units in destination territory will automatically die, do you still want to move?";
-    return ui.getOk(question);
+    ui.getOk(question);
   }
 
   @Override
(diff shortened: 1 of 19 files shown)


game-core/src/main/java/games/strategy/triplea/TripleAPlayer.java
```diff
@@ -711,10 +711,10 @@ public boolean confirmMoveKamikaze() {
   }
 
   @Override
-  public boolean confirmMoveHariKari() {
+  public void confirmMoveHariKari() {
     final String question =
         "All units in destination territory will automatically die, do you still want to move?";
-    return ui.getOk(question);
+    ui.getOk(question);
```
RoiEXLab: This seems odd.
This smells like the original plan of the code designer was to query this method to get the confirmation, but it looks like the UI handles the result automatically

game-core/src/main/java/games/strategy/triplea/ai/AbstractAi.java
```diff
@@ -103,9 +103,7 @@ public boolean confirmMoveKamikaze() {
   }
 
   @Override
-  public boolean confirmMoveHariKari() {
-    return false;
-  }
+  public void confirmMoveHariKari() {}
```
RoiEXLab: Here the issue becomes very clear.
From a readers perspective this looked like the Ai was denying HariKari in general.
However if you think about it confirmation of some sort only makes sense for human players, the Ai is not going to be like "ok, right I didn't notice I would lose those units".
So move this method to a different interface maybe?

PR #4999, opened 2019-08-16
Fix default run configurations for Intellij Idea.

OrdoFlammae:
## Overview
This pull request fixes a slight error in the run configurations for Intellij.

## Manual Testing Performed
- ..

### DIFF
--- a/.idea/runConfigurations/HeadedGameRunner.xml
+++ b/.idea/runConfigurations/HeadedGameRunner.xml
@@ -1,7 +1,7 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="HeadedGameRunner" type="Application" factoryName="Application" singleton="false" nameIsGenerated="true">
     <option name="MAIN_CLASS_NAME" value="org.triplea.game.client.HeadedGameRunner" />
-    <module name="triplea.game-headed" />
+    <module name="triplea.game-headed.main" />
     <option name="WORKING_DIRECTORY" value="$PROJECT_DIR$/game-headed" />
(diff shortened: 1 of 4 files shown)


DanVanAtta:
Great to see the first contribution @OrdoFlammae , welcome

I think we might have been ping-ponging between having this run config vs not

DanVanAtta:
@OrdoFlammae did you import the checked in IDEA settings (in folder: `.ide-intellij/`)? If not, does it change the configs for you?

RoiEXLab:
> I think we might have been ping-ponging between having this run config vs not

@DanVanAtta That's exactly what happened, this PR broke my configuration.

OrdoFlammae:
@DanVanAtta, I did import the IDEA settings. Just in case, I just re-imported them, and nothing has really changed. @RoiEXLab, how does this break your configuration. Sorry, I had no idea that would happen (probably should've guessed, since this hadn't happened yet).

RoiEXLab:
Well my IDEA project expects the "one module per gradle project" layout. Whereas your configuration seemingly uses the "one module per source set" approach.
If your have the project open, you can open a window using the bar on the right where it says "Gradle", then select the wrench 🔧 and somewhere there should be a checkbox that allows you to configure this setting.

OrdoFlammae:
@RoiEXLab, that setting appears to be non-existent in recent versions of IDEA (I'm using 2019.2).

RoiEXLab:
Yeah, I just double checked too:
Screenshot

Interesting

PR #5000, opened 2019-08-17
Remove Mac Java 8 Handler code

DanVanAtta:
## Overview
Min version of java is 11, remove handler code for java8.

## Functional Changes
- none expected

## Manual Testing Performed
- none


DanVanAtta:
@RoiEXLab expanded the scope here and removed the reflective access. Please take a look when you get the chance

DanVanAtta:
@RoiEXLab good consideration. Going to merge now to lock in this iteration. It probably is the case that we can just inline the code. Perhaps it's useful to become stateful and unit tested; leaving that consideration for the next developer.

PR #5013, opened 2019-08-18
Fixes #4990 on client side.

### LINKED ISSUE #4990: Create account allows blank email
### Repro
- launch latest Triplea lobby + client
- connect to lobby
- create new account
- fill in a name and password, leave email blank
- click ok

Notice: account is created without a email
Expected: account should not be created

### Discussion

We have a non-null constraint on DB and we have no empty emails in production. I suspect we should prefer allowing null or non-empty but not default to empty emails. This query is more intuitive IMO:
```
select * from lobby_user where email is not null;
```

vs:
```
select * from lobby_user where email != '';
```

### Expected fix
Require a non-empty email when creating account

OrdoFlammae:
Empty emails are not allowed to be input when creating an account with client.

Interesting note: when the tests were written for email validation, an empty email was expected to pass.


DanVanAtta:
It looks like the empty email validation in some cases is okay and expected, hence the interesting side-note.

Checking usages of the email validation, in DiceServerEditor.java:195
```
    final boolean ccValid =
        setLabelValid(PlayerEmailValidation.isValid(ccAddress.getText()), ccLabel);
```
The CC field I think is optional and is hence okay to blank.

I suspect the validation update will need to go directly in: `CreateUpdateAccountPanel.java`

PR #5020, opened 2019-08-19
Fixes #4990.

### LINKED ISSUE #4990: Create account allows blank email

OrdoFlammae:
## Overview
This pull request fixes #4990.

## Functional Changes
Empty email addresses are no longer accepted when creating an account.

## Manual Testing Performed
Made sure the email text-box doesn't accept empty emails.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/login/CreateUpdateAccountPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/login/CreateUpdateAccountPanel.java
@@ -277,6 +277,9 @@ public final class CreateUpdateAccountPanel extends JPanel {
           "Invalid name",
           JOptionPane.ERROR_MESSAGE);
       return;
+    } else if (usernameField.getText().length() == 0) {
+      JOptionPane.showMessageDialog(
+          this, "You must enter an email", "No Email", JOptionPane.ERROR_MESSAGE);
     } else if (passwordField.getPassword().length == 0) {
       JOptionPane.showMessageDialog(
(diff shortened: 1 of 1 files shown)


DanVanAtta:
@OrdoFlammae good to see the fix :+1: 

Few notes:
- Server side should likely also protect against the empty email case. Software tends to have a funny of habit of where if anything can happen, it does. Specifically server should check for empty email and the DB should have a constraint to disallow the invalid data. The server check is to defend against custom clients and possible bugs. DB constraint is another layer of data validation,  see 'database as fortress design principle':
  - [97 Things Every Software Architect Should Know](https://books.google.com/books?id=HDknEjQJkbUC&pg=PA46&lpg=PA46&dq=database+as+fortress&source=bl&ots=SKH90P7GmR&sig=ACfU3U1eShodycFbJCS677Q63DajPVy2mg&hl=en&sa=X&ved=2ahUKEwj8je7ru5TkAhWN9Z4KHUsVBrEQ6AEwCnoECBIQAQ#v=onepage&q=database%20as%20fortress&f=false)

PR #5023, opened 2019-08-19
Don't set GameData to null

### LINKED ISSUE #5022: NullPointerException when trying to load installed maps
### Engine version
1.10.15336

### Map name
any installed map (latest versions)

### Can you describe how to trigger the error? (eg: what sequence of actions will recreate it?)
Start Triple A, hit "Select Map", select any map.
A window pops up:
npe

 I have got a "NullPointerException" in the error.log:

### Do you have the exact error text? Please copy/paste if so

[error.log](https://github.com/triplea-game/triplea/files/3517278/error.log)

### Instead of this error, what should have happened?
TripleA should allow me to choose the game-type, select the players and start the game.

### Any additional information that may help

Interestingly after some tries I can load the map as expected and continue.
So this error seems to be maybe related to a fresh installation?

RoiEXLab:
## Overview
Should hopefully fix #5022 

## Functional Changes
Bugfix.

## Manual Testing Performed
Verified Games are loaded normally.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/ui/DefaultGameChooserEntry.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/ui/DefaultGameChooserEntry.java
@@ -11,13 +11,14 @@ import java.net.URI;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.logging.Level;
+import javax.annotation.Nonnull;
 import lombok.extern.java.Log;
 import org.triplea.java.UrlStreams;
 
 @Log
 final class DefaultGameChooserEntry implements GameChooserEntry {
   private final URI url;
-  private GameData gameData;
+  @Nonnull private GameData gameData;
   private boolean gameDataFullyLoaded = false;
   private final String gameNameAndMapNameProperty;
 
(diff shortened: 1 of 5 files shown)


game-core/src/main/java/games/strategy/engine/framework/ui/DefaultGameChooserEntry.java
```diff
@@ -43,8 +44,6 @@ public void fullyParseGameData() throws GameParseException {
     // TODO: We should be setting this in the the constructor. At this point, you have to call
     // methods in the
     // correct order for things to work, and that is bads.
-    gameData = null;
-
```
RoiEXLab: This is the core change

ron-murhammer: @RoiEXLab Uh, any idea why this was setting it to null? Concerned there could be side effects of removing this.

RoiEXLab: Well in case parsing fails, this null assignment replaces the shallow parsed GameData with null.
But I'd argue that this is a wrong behaviour because the way the class works is that it parses a shallow version in the constructor that throws an exception if that fails, so an object without corresponding GameData can't be constructed.
But I double checked all usages (and removed null checks) if that helps

ron-murhammer: Alright. I'm good moving forward with this change then.

PR #5029, opened 2019-08-21
Revert "Fix default run configurations for Intellij Idea."

DanVanAtta:
Reverts triplea-game/triplea#4999

Ping-pong back to previous configuration. To have the latest work I need to rename project modules to have a ".main" suffix and then re-importing from gradle can often break this. 

@OrdoFlammae , @RoiEXLab, if either of you can find the "trick" to get IDEA to the right configuration, we desperately could use documentation for it.


OrdoFlammae:
I'm confused. Why do you need to rename project modules to have a .main suffix? They already have it.

RoiEXLab:
@DanVanAtta have a look at the discussion in #4999 there you can see how I resolved the problem.
It looks like the default configuration of intellij changed, thus when importing triplea with a clean installation it requires the "new"/current run configuration

DanVanAtta:
Cool, sounds like that legacy option needs to be checked unless getting a clean install of IDEA. Considering this is purely an upgrade concern, everyone impacted is just us, going forward default config should work, meaning we do not need to document this.

Issue #5024, opened 2019-08-20
CheckStyle and spotlessApply conflict


OrdoFlammae:
Apparently CheckStyle and spotlessApply have different ideas about how to do things. CheckStyle wants the ending curly brace of every method to be on its own line, while, when spotlessApply runs, it moves the ending curly brace of empty methods to the same line as the declaration. Which side do we want to go with? So far, we've been using the spotlessApply way of things, so should we just remove the one check from the CheckStyle?

DanVanAtta:
@OrdoFlammae do you have any example Java files or code examples where you've seen this behavior?

OrdoFlammae:
For example, in AaInMoveUtil.java, the current code is
```
AaInMoveUtil() {}
```

CheckStyle complains if it isn't
```
AaInMoveUtil() {
}
```

This happens around 356 times currently in 253 files.

DanVanAtta:
The gradle build checks for violations and fails if there are any. I suspect maybe you're not using the checked-in checkstyle configuration. Could you check that you've configured the checkstyle plugin to use it?

It should look something like this:
![Screenshot from 2019-08-21 12-15-41](https://user-images.githubusercontent.com/12397753/63461481-91b9c280-c40d-11e9-81d5-4d1d043ad464.png)

We've a checked-in settings config as well under `/.ide-intellij`, importing that might auto-configure this for you. It'd be interesting to know if that is or is not working as well.

OrdoFlammae:
It looks like the problem is that my Checkstyle version is later than yours (8.23).

DanVanAtta:
I'm able to repro the problem after bumping up the checkstyle version. It does seem the checkstyle rules are in conflict.
I submitted a PR to remove the conflicting checkstyle rules: https://github.com/triplea-game/triplea/pull/5031

RoiEXLab:
Fixed in #5031

PR #5031, opened 2019-08-21
Remove checkstyle rules that conflict with google format

DanVanAtta:
## Overview

Google format will conditionally place right curly on same line when there is empty content for: while loops, for loops, class def, method def

This update removes the checkstyle rule that checks for right curly to be on its own line which conflicts with the above cases. The google-format will correct this and spotless gradle plugin will enforce this.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Code Cleanup or refactor
[ ] Code Removal
[X] Configuration Change
[ ] Bug fix

 

## Testing

[X] Manually tested
[ ] No testing done

### Manual Testing
In IDEA with checkstyle version 8.23, verified there is no checkstyle violation with the following:
```
  static {
  }

  AaInMoveUtil() {}

  public void initialize(final IDelegateBridge bridge) {
    this.bridge = bridge;
    this.player = bridge.getPlayerId();
    for (int i = 0; i < 2; i++) {}
    while (true) {}
  }

  private static class Test {}
```

## Additional Review Notes
Addresses: https://github.com/triplea-game/triplea/issues/5024

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -144,7 +144,7 @@ subprojects {
     }
 
     checkstyle {
-        toolVersion = '8.12'
+        toolVersion = '8.23'
         configFile = rootProject.file('config/checkstyle/checkstyle.xml')
         configProperties = [samedir: configFile.parent]
     }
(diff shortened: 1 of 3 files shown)


DanVanAtta:
Also pushed an update to checkstyle version, after that update the gradle run of checkstyle reproduces the problem we see when IDEA is using latest checkstyle version

Issue #5035, opened 2019-08-22
'Shift' Click to Unload Transports no longer works


DanVanAtta:
### Engine version
1.10.14690+

### Can you describe how to trigger the error? (eg: what sequence of actions will recreate it?)
- Load a transport with a land unit
- Shift Click SZ containing transport
- Try to unload on an adjacent land territory
- Notice: error Message

### Instead of this error, what should have happened?
Land units on sea transport should unload

### Any additional information that may help
I bisected the error to this commit: 1860c849a61a9f003abe70b3645487690dcc2d63 (tag: 1.10.14690) Fix incorrect transports unloading

DanVanAtta:
https://github.com/triplea-game/triplea/pull/5037 fixes this, but possibly breaks the bug-fix that originally introduced the probelm

ron-murhammer:
@DanVanAtta I think this will have side effects and probably isn't the best way to fix this. Closing in favor of #5049.

ron-murhammer:
Closing based on #5049.

PR #5037, opened 2019-08-22
Fix unload all transport units

DanVanAtta:
## Overview

Fixes shift click movement offloading units from transports. Fixed by removing a 'negation' line that adds the sea units to the current selected units.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Code Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[X] Bug fix
 

#5035

This line adds seazone units to selection which are then attempted to be moved to land: 
```
units.addAll(CollectionUtils.getMatches(selectedUnits, getUnloadableMatch().negate()));
```

Why this did not cause problems before is curious to me..

Removed line adding sea zone units.

## Testing

[X] Manually tested
[ ] No testing done

### Manual Testing
With revised rules, load units to a transport, with air units on a carrier. Shift-click the SZ and move the units to neighboring land.
 

## Additional Review Notes

- I don't understand why this problem was introduced exactly in: https://github.com/triplea-game/triplea/commit/1860c849a61a9f003abe70b3645487690dcc2d63, some extra scrutiny would be appreciated. Notably for any use-cases or rule sets which maybe this regresses on.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
@@ -526,7 +526,13 @@ public class MovePanel extends AbstractMovePanel {
         private Predicate<Unit> getUnloadableMatch() {
           // are we unloading everything? if we are then we dont need to select the transports
           return PredicateBuilder.of(Matches.unitIsOwnedBy(getCurrentPlayer()))
-              .and(Matches.unitIsLand())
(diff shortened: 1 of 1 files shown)


ron-murhammer:
@DanVanAtta I think this will have side effects and probably isn't the best way to fix this. Closing in favor of #5049.

PR #5039, opened 2019-08-22
Unit scroller improvements

DanVanAtta:
## Overview

(1) Notify when all units moved via popup. If this is annoying to users, it can be turned off with a 'do not show again' button or turned off/on via game settings.
(2) If all units in current territory are moved, 'c' will focus on next territory instead of focusing on the current (and empty) territory
(3) Sleep only movable units in a territory. For example landing bombers with an AA gun, it's common to only want the AA gun to be slept.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Code Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[X] Bug fix
 
### User facing changes
See overview

Hit the 'm' key for previous unit, should see an exception.

Terrritory list is returned as an immutable list, reversing that in-place throws UnSupportedOperation.

Copy the immutable territory list to a mutable list and reverse that list instead.

## Testing

[X] Manually tested
[ ] No testing done

### Manual Testing
- Use 'n' to select next territory, move all units out of that territory. Press 'c' and notice that the next territory is selected 
- Press 'm' to select previous territory, notice no errors.
- Press 's' to sleep all units, notice pop-up, click 'turn-off' and repeat again in next move phase and notice pop-up is not shown
- For example with a territory with AA gun, move a unit to that territory and ensure it has no movement left. During non-combat, when AA gun is up, click sleep, then during next move phase verify the unit that had exhausted movement does not come up in the scroller

## Screens Shots

Screenshot from 2019-08-21 23-49-41
Screenshot from 2019-08-21 23-50-40
Screenshot from 2019-08-21 23-50-56

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/unit/scroller/UnitScroller.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/unit/scroller/UnitScroller.java
@@ -7,9 +7,11 @@ import games.strategy.engine.data.Territory;
 import games.strategy.engine.data.Unit;
 import games.strategy.triplea.attachments.TerritoryAttachment;
 import games.strategy.triplea.delegate.Matches;
+import games.strategy.triplea.settings.ClientSetting;
 import games.strategy.triplea.ui.MapPanel;
 import java.awt.BorderLayout;
 import java.awt.Component;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
@@ -21,10 +23,12 @@ import javax.swing.BoxLayout;
 import javax.swing.JButton;
 import javax.swing.JComponent;
 import javax.swing.JLabel;
+import javax.swing.JOptionPane;
 import javax.swing.JPanel;
 import javax.swing.SwingUtilities;
 import org.triplea.java.PredicateBuilder;
 import org.triplea.java.collections.CollectionUtils;
+import org.triplea.swing.DialogBuilder;
 import org.triplea.swing.JPanelBuilder;
 import org.triplea.swing.SwingComponents;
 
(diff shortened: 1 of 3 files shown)


game-core/src/main/java/games/strategy/triplea/ui/unit/scroller/UnitScroller.java
```diff
@@ -241,39 +247,81 @@ public void centerOnCurrentMovableUnit() {
...
+        JOptionPane.showOptionDialog(
+            mapPanel,
+            "All units have moved",
+            "All units moved",
+            JOptionPane.YES_NO_OPTION,
+            JOptionPane.PLAIN_MESSAGE,
+            null,
+            new String[] {"OK", "Do not show again"},
+            "OK");
+    if (result == 1) {
```
RoiEXLab: JOptionPane should have constants for this.
I'm pretty sure intellij has a quick fix for this, but maybe it doesn't work because the assignment and the actual comparison aren't the same. But don't know

DanVanAtta: Swing is returning the index position of the array passed in as the second to the last argument. The array is arbitrary, so there couldn't be constants for arbitrary index positions. It is in contrast to a YES_NO dialog or the YES_NO_CANCEL dialog which IIRC does return magic constants as well. Those constants I think are also '0', '1', and '2', but the relationship there is only coincidental, they just happen to be the same value.

RoiEXLab: Fair point, didn't know that. Perhaps this indicates it should be documented what the 1 stands for?

DanVanAtta: It is a shortcoming of the swing API. I'm not sure how it can be commented on without more or less saying  "see the javadocs of JOptionPane"

Having a better API is probably a better answer. TBH I think this falls under the 'know your library' category. Swing has a horrible API like this.

We have other examples, should we document them all? Do you have any kind of comment in mind that would not just be copy/pasted next to every usage and be redundant to the javadoc on the API?

Examples:

ViewMenu.java:150
```
              final int result =
                  JOptionPane.showOptionDialog(
                      frame,
                      panel,
                      "Choose Map Scale",
                      JOptionPane.OK_CANCEL_OPTION,
                      JOptionPane.PLAIN_MESSAGE,
                      null,
                      new String[] {"OK", "Cancel"},
                      0);
              if (result != 0) {
                return;
              }
```

ViewMenu:382
```
              final int result =
                  JOptionPane.showOptionDialog(
                      frame,
                      ui,
                      "Edit Map Font and Color",
                      JOptionPane.YES_NO_CANCEL_OPTION,
                      JOptionPane.PLAIN_MESSAGE,
                      null,
                      options,
                      2);
              if (result == 1) {
```

RoiEXLab: Maybe a single comment "Check if second button was pressed" seems reasonable?

If you disagree feel free to merge.

DanVanAtta: @RoiEXLab thought of something, an intermediate variable to describe the condition seems to work, updated in: 0a98c9296

PR #5043, opened 2019-08-22
Properly catch unwrapped exception

### LINKED ISSUE #5027: Game Over Exception
### Engine version
1.10.15353

### Map name
Experienced with Napoleonic Empires

### Can you describe how to trigger the error? (eg: what sequence of actions will recreate it?)
While testing the solution to https://github.com/triplea-game/triplea/issues/5022 I noticed this:
Load the map. Start a local game. Exit the game.
gameover

### Do you have the exact error text? Please copy/paste if so

```
Aug. 21, 2019 9:01:31 VORM. org.triplea.game.client.HeadedGameRunner lambda$main$0
SCHWERWIEGEND: Game Over
games.strategy.engine.GameOverException: Game Over
	at games.strategy.engine.player.DefaultPlayerBridge.getRemoteDelegate(DefaultPlayerBridge.java:58)
	at games.strategy.triplea.TripleAPlayer.canUnitsFight(TripleAPlayer.java:385)
	at games.strategy.triplea.TripleAPlayer.move(TripleAPlayer.java:339)
	at games.strategy.triplea.TripleAPlayer.start(TripleAPlayer.java:142)
	at games.strategy.engine.framework.ServerGame.waitForPlayerToFinishStep(ServerGame.java:580)
	at games.strategy.engine.framework.ServerGame.runStep(ServerGame.java:450)
	at games.strategy.engine.framework.ServerGame.startGame(ServerGame.java:307)
	at games.strategy.engine.framework.startup.launcher.LocalLauncher.launchInternal(LocalLauncher.java:56)
	at games.strategy.engine.framework.startup.launcher.LocalLauncher.launchInternal(LocalLauncher.java:29)
	at games.strategy.engine.framework.startup.launcher.AbstractLauncher.lambda$launch$0(AbstractLauncher.java:16)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

RoiEXLab:
## Overview
This fixes a bug introduced by #5018 

## Functional Changes
[ ] New map or map update
[ ] New Feature
[ ] Code Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[x] Bug fix

Fixes #5027

I missed a case where the exception was directly thrown an wasn't wrapped into a RuntimeException.

By catching GameOverException explicitly as well.

## Testing
[x] Manually tested
[ ] No testing done

### Manual Testing
I verified the error no longer appears after closing a local game, and verified no unintentional bugs were introduced to the networked variants.
Note that I'm not 100% sure if the recent Exception catching changes were done too well: I have the suspicion that since we're now properly catching all the exceptions, users don't get a savegame when an opponent disconnects because previously this caused an exception to be propagated to all connected users, and now the Exception is caught and everything safely terminates.
I'm not 100% sure how exactly this mechanic is supposed to behave, so I'd kindly ask you to double check this.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/ServerGame.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/ServerGame.java
@@ -307,6 +307,10 @@ public class ServerGame extends AbstractGame {
           runStep(false);
         }
       }
+    } catch (final GameOverException e) {
+      if (!isGameOver) {
+        log.log(Level.SEVERE, "GameOverException raised, but game is not over", e);
+      }
     } catch (final RuntimeException e) {
       if (e.getCause() instanceof GameOverException) {
         if (!isGameOver) {
--- a/game-core/src/main/java/games/strategy/engine/framework/ServerGame.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/ServerGame.java
@@ -311,14 +311,6 @@ public class ServerGame extends AbstractGame {
       if (!isGameOver) {
         log.log(Level.SEVERE, "GameOverException raised, but game is not over", e);
       }
-    } catch (final RuntimeException e) {
-      if (e.getCause() instanceof GameOverException) {
-        if (!isGameOver) {
-          log.log(Level.SEVERE, "GameOverException raised, but game is not over", e);
-        }
-      } else {
-        throw e;
-      }
     }
   }
 


game-core/src/main/java/games/strategy/engine/framework/ServerGame.java
```diff
@@ -307,6 +307,10 @@ public void startGame() {
           runStep(false);
         }
       }
+    } catch (final GameOverException e) {
```
DanVanAtta: It looks like the `if` condition below could be modified to account for this, ie:
`if (e instanceof GameOverException || e.getCause() instanceof GameOverException) {`

Which makes it seem like the bug is that instead of `GameOverException` being the wrapped exception, it *is* the exception now. Considering we throw any other exception type, the entire catch block for `RuntimeException` and its nested if statement now looks unnecessary.

DanVanAtta:
I verified the autosave feature is still good, launched two games via terminal and connected to a locally hosted game. I force killed the client and alternatively killed the host. Beyond seeing chat panel bug remaining after game close, everything looked fine.  :+1:

game-core/src/main/java/games/strategy/engine/framework/ServerGame.java
RoiEXLab: Yeah I wasn't sure what to prefer at this point.
The whole exception thing is a complete mess.
I initially tried making GameOverException checked, but ultimately failed because at some points it's treated like an unchecked exception to terminate a game and at some other points in the code it gets treated as something unexpected that needs to be handled in a special way.
Combine that with the fact that at some point the exception is transmitted over the network and everything is completely unusable.

If you have the time you might want to have a look yourself, maybe you see something I missed

DanVanAtta: I think you can keep it simple here, the RuntimeException block is now buggy because `e.getCause()` is no longer GameOverException.

Catching the right exception type is all that is needed now as any other exception is rethrown. This is an example of 'control-flow-by-exception-handling', just not clean.. Perhaps better would be to have a game stop listener, have that get executed and set a 'game-terminated' flag which would then exit this loop. That's a decent sized refactor rather than a bug fix. Pretty much I think you can remove the `catch(RuntimeException)` block and this should be good to go.

DanVanAtta: > I initially tried making GameOverException checked, but ultimately failed because at some points it's treated like an unchecked exception to terminate a game and at some other points in the code it gets treated as something unexpected that needs to be handled in a special way.

:grinning:  Oh, TripleA! Not only do we have control-flow-by-exception-handling, but we have a violation of SRP too : )

It's a good call out how messy that is.
I suspect somewhere in there is the 'ghosting' bug, the system needs perhaps a good rework to be simpler and more pro-active to verify connectivity.

RoiEXLab: In the PR that changed the exception types a bit there was a comment somewhere that raised the question if it would make sense to have GameOverException extend MessengerException.
After having to deal with this, I realised why just a couple hours ago: In local games this exception is just the "terminate game" exception.
In networked games this exception is used to somehow communicate disconnects that lead to the end of the game, that's why everything is so messed up.

RoiEXLab: > I think you can keep it simple here, the RuntimeException block is now buggy because e.getCause() is no longer GameOverException.

Are you sure about that? I'm pretty sure I put that check there for a reason, but at the same time it could've been me being overly cautious

DanVanAtta: I ran it through a debugger, the if statement fails because `e` is an instance of `GameOverException`

RoiEXLab: Ok, but did you verify the case when you're hosting a networked game?
For local games, this was clearly my mistake, but I could imagine that for networked games the exception is wrapped at some point and rethrown when a player, or the host itself disconnects

RoiEXLab: @DanVanAtta I'll get some sleep now.
Feel free to push the desired changes directly to this branch (you should be able to do so via the web client as well), if you want this merged ASAP.

Otherwise I'll commit the changes tomorrow based on your answer.

DanVanAtta: @RoiEXLab this is your bug fix, please walk through a debugger in that case to verify.

RoiEXLab: Good observations @DanVanAtta I wasn't able to hit this catch block either. Removed it.

DanVanAtta: Thanks for verifying!

PR #5049, opened 2019-08-22
Ensure non-unloading units can move along route

ron-murhammer:
## Overview
Fixes shift click movement offloading units from transports. Fixed by removing a 'negation' line that adds the sea units to the current selected units (#5035).

This bug was introduced when fixing #4835. Before that fix, the engine was honoring the specific transports that the user selected to unload and was just randomly picking which units to unload. That fix corrected that issue but also then tried to move all other non-unloadable selected units to the territory even if they couldn't legally move there (ex. sea units onto land).

This now only adds in non-unloadable units which can actually legally move along the route + the units from the transports that the user selected.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Code Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[x] Bug fix

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
@@ -262,7 +262,8 @@ public class MovePanel extends AbstractMovePanel {
             // Load Bombers with paratroops
             if ((!nonCombat || isParatroopersCanMoveDuringNonCombat(getData()))
                 && TechAttachment.isAirTransportable(getCurrentPlayer())
-                && selectedUnits.stream()
+                && selectedUnits
+                    .stream()
                     .anyMatch(Matches.unitIsAirTransport().and(Matches.unitHasNotMoved()))) {
               final PlayerId player = getCurrentPlayer();
               // TODO Transporting allied units
@@ -555,7 +556,8 @@ public class MovePanel extends AbstractMovePanel {
             units.addAll(
                 getUnitsToUnload(
                     route, CollectionUtils.getMatches(selectedUnits, getUnloadableMatch())));
-            units.addAll(CollectionUtils.getMatches(selectedUnits, getUnloadableMatch().negate()));
+            units.addAll(
+                CollectionUtils.getMatches(unitsThatCanMoveOnRoute, getUnloadableMatch().negate()));
             if (units.isEmpty()) {
               cancelMove();
               return;
@@ -984,7 +986,8 @@ public class MovePanel extends AbstractMovePanel {
   private static Predicate<Unit> areOwnedUnitsOfType(
       final Collection<Unit> units, final PlayerId owner) {
     return mainUnit ->
-        units.stream()
+        units
+            .stream()
             .filter(unit -> unit.getOwner().equals(owner))
             .anyMatch(Matches.unitIsOfType(mainUnit.getType()));
   }
--- a/game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
@@ -262,8 +262,7 @@ public class MovePanel extends AbstractMovePanel {
             // Load Bombers with paratroops
             if ((!nonCombat || isParatroopersCanMoveDuringNonCombat(getData()))
                 && TechAttachment.isAirTransportable(getCurrentPlayer())
-                && selectedUnits
-                    .stream()
+                && selectedUnits.stream()
                     .anyMatch(Matches.unitIsAirTransport().and(Matches.unitHasNotMoved()))) {
               final PlayerId player = getCurrentPlayer();
               // TODO Transporting allied units
@@ -986,8 +985,7 @@ public class MovePanel extends AbstractMovePanel {
   private static Predicate<Unit> areOwnedUnitsOfType(
       final Collection<Unit> units, final PlayerId owner) {
     return mainUnit ->
-        units
-            .stream()
+        units.stream()
             .filter(unit -> unit.getOwner().equals(owner))
             .anyMatch(Matches.unitIsOfType(mainUnit.getType()));
   }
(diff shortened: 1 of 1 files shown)


ron-murhammer:
@DanVanAtta There appears to be formatting inconsistencies in the new profile. As when I run the format it changes 2 lines which cause the build to fail formatting checks.

DanVanAtta:
The update is an improvement as there is no error message and land units unload, but the shift click does not include fighters from carriers.

Attached is a save with that scenario setup with a german carrier and transport: [carrier_and_transport_unload.tsvg.zip](https://github.com/triplea-game/triplea/files/3532142/carrier_and_transport_unload.tsvg.zip)

ron-murhammer:
@DanVanAtta That appears to actually be a separate bug as if you test in the current stable release (and probably all releases before) that same issue exists.

To fix that, most likely the issue is in `updateUnitsThatCanMoveOnRoute`.

game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
```diff
@@ -555,7 +556,8 @@ private void selectEndPoint(final Territory territory) {
             units.addAll(
                 getUnitsToUnload(
                     route, CollectionUtils.getMatches(selectedUnits, getUnloadableMatch())));
-            units.addAll(CollectionUtils.getMatches(selectedUnits, getUnloadableMatch().negate()));
+            units.addAll(
+                CollectionUtils.getMatches(unitsThatCanMoveOnRoute, getUnloadableMatch().negate()));
```
DanVanAtta: The 'negate' is confusing. A comment to clear it up or adding a new matcher so the condition is not negated would help. It's hard to understand why we are taking the units that cannot move on the route and adding that to the `units` collection.

ron-murhammer: I think you are misunderstanding. We are taking the non-unloadable units (essentially air/sea units) and adding them in.

DanVanAtta: I know I am misunderstanding :)
That is the problem, this is a very confusing thing to read! It seems like the code is doing the opposite with the 'negate'. With the selected units we seems like we were grabbing all units in the territory as a tautology. It's not clear why this addition is needed and what's it's really doing. I don't think future code maintainers would have better luck.

OrdoFlammae: A comment goes a long way.....

DanVanAtta:
Tested 7xxx, we get the same behavior. It requires multiple rounds of shift clicking to get the planes to move. #5037 had an update to grab them, can we get parity with that fix or a fast follow?

ron-murhammer:
@DanVanAtta I'm fairly certain #5037 didn't fix including planes either are you sure you tested that?

As the change adds planes into the list of units that get passed to the player selection for unloading transports but after they choose which transports to unload, it doesn't add the planes in. It also causes a number of side effects like having to select which transports to unload even when you are unloading all units because you are now passing air units to the unload method.

But yes, I'm planning to take a look at seeing if fixing having planes move during the unload is doable.

DanVanAtta:
@ron-murhammer I took a look at the formatting problem and it's a version issue. Please see: https://github.com/triplea-game/triplea/issues/5051, there is an unofficial plugin that could work, if so that is a less than ideal path forward.

Otherwise, if the code could be improved to make it more clear, in-code, what/why the negate is needed and what it is doing, this would be an ideal fix for now.

DanVanAtta:
> @DanVanAtta I'm fairly certain #5037 didn't fix including planes either are you sure you tested that?

If all transports are homogonous, same units, then shift-clicking will unload all. Otherwise there is a prompt of which transports to unload and then the planes will not move. I did not notice any other negative consequences from #5037 after playing through a full game of revised.

ron-murhammer:
@DanVanAtta I'd love to see a screenshot or video of it working then as when I tested it, it didn't include planes and made me choose transports even though I'm unloading all of them. All I did was load up revised, edit 2 russian transports next to caucasus, load 4 units, move fighter to that sea zone, and unload all.

DanVanAtta:
With #5037

### Case 1
1 inf, 1 transport, 1 carrier with a fig:
Screenshot from 2019-08-22 17-05-33
Screenshot from 2019-08-22 17-05-56
Screenshot from 2019-08-22 17-06-06


### Case 2
1 carrier, 1 fig, 2 transports, 4 inf
Screenshot from 2019-08-22 17-07-09
Screenshot from 2019-08-22 17-07-16
Screenshot from 2019-08-22 17-07-21

### Case 3 (non-homgonous)
2 transports, 3 inf, 1 carrier, 1 fig

Screenshot from 2019-08-22 17-08-09
Screenshot from 2019-08-22 17-08-13
Screenshot from 2019-08-22 17-08-17
Screenshot from 2019-08-22 17-08-22

-----------

I do not disagree that #5037 is pretty hacky given the above, and it does not really matter that the original behavior is also hacky. There are a number of problems to fix:
- shift clicking should move all available units, planes included
- the semi-transparent units shown when moving all should include planes (#5037 fails there)
- when moving off a transport, the movement indicator says "1/0", it should say "1/1" or "1/1+" if there are figs with more than one movement. eg:
Screenshot from 2019-08-22 17-10-40
Screenshot from 2019-08-22 17-11-37

ron-murhammer:
@DanVanAtta Yep, that's consistent with what I see and I was referring to case 3 which technically introduces a new bug where you now have to select which transports even when unloading all.

I think we should move forward with this PR as it does fix the blocker regression issue. I think creating a new non-blocker issue to further improve the other points we discussed (including planes, fixing hover over image, movement counter).

DanVanAtta:
@ron-murhammer is there a valuable comment that can be added to explain the 'negate' line? It's taking a really lot of effort to understand that, even with your explanation I'm still having trouble grokking it.

ron-murhammer:
@DanVanAtta Added comments. Not sure if that helps but open to suggestions if those aren't useful.

DanVanAtta:
@ron-murhammer it helps  :smile:

PR #5056, opened 2019-08-23
Allow air units to move when unloading

ron-murhammer:
## Overview
- Improve unloading moves so that if air units are selected they are moved as well
- Follow up to discussion in #5049

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Code Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix

image

image


DanVanAtta:
Noticing the warning icon in the first screenshot - perhaps that is a bug or is it there to indicate the SZ units are not moving?

Thanks for taking care of the update @ron-murhammer  - it's an excellent usability enhancment

PR #5061, opened 2019-08-23
Update LICENSE

DanVanAtta:
Remove all markdown, actually revert to a previous version here.


DanVanAtta:
Hopefully third time will be the charm here..

Issue #4959, opened 2019-08-11
"N" Key Does not work after moving units (Multiplayer only)


DanVanAtta:
### Problem

After moving a unit while playing a multiplayer game, the "n" key will no longer work and does not select the next available unit.

Workaround: click the map, 'n' key then works again.

### Repro

- Host a game
- Start the game
- Press 'n' key to cycle to a next unit
- Move any unit
- Press 'n' key again

Notice: nothing happens
Expected: 'n' key should cycle to next unit

Of note, this is not a problem for local/single player games.

DanVanAtta:
Adding release blocker to this; The additional keys 'm', 'space bar' and 'c' also do not work. With the unit-scroller feature this becomes a prominent bug and makes that feature essentially broken. 

I'm working on a fix for this..

PR #5062, opened 2019-08-23
Replace key listeners with key bindings

DanVanAtta:
## Overview
Convert some key listeners to key bindings. Key bindings do not need exact component focus and more reliably fire. Fixes problem where component focus is lost after moving units in multiplayer game making hotkeys no longer work. Because focus is not required, the 'setfocusable(false)' flag on buttons is no longer required.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Code Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[X] Bug fix

https://github.com/triplea-game/triplea/issues/4959

Swing focus can be lost relatively easily

Use key bindings instead of key listeners

> Key listeners are also difficult if the key binding is to be active when the component doesn't have focus. Some of the advantages of key bindings are they're somewhat self documenting, take the containment hierarchy into account, encourage reusable chunks of code (Action objects), and allow actions to be easily removed, customized, or shared. Also, they make it easy to change the key to which an action is bound. Another advantage of Actions is that they have an enabled state which provides an easy way to disable the action without having to track which component it is attached to.
(https://docs.oracle.com/javase/tutorial/uiswing/misc/keybinding.html)

## Testing

[X] Manually tested
[ ] No testing done

### Manual Testing
- Launched a network game, moved a unit, used hotkeys  'f', 'n', 'm'

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
@@ -169,7 +169,7 @@ import org.triplea.util.Tuple;
 
 /** Main frame for the triple a game. */
 @Log
-public final class TripleAFrame extends JFrame implements KeyBindings {
+public final class TripleAFrame extends JFrame implements KeyBindingSupplier {
   private static final long serialVersionUID = 7640069668264418976L;
   private final LocalPlayers localPlayers;
   private final GameData data;
@@ -751,7 +751,7 @@ public final class TripleAFrame extends JFrame implements KeyBindings {
     uiContext.addShutdownWindow(this);
   }
 
-  private void addKeyBindings(final KeyBindings... keyBindings) {
+  private void addKeyBindings(final KeyBindingSupplier... keyBindings) {
     Arrays.stream(keyBindings)
         .map(Supplier::get)
         .map(Map::entrySet)
--- a/game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
@@ -94,7 +94,7 @@ import java.awt.Toolkit;
 import java.awt.event.ActionEvent;
 import java.awt.event.FocusAdapter;
 import java.awt.event.FocusEvent;
-import java.awt.event.KeyAdapter;
+import java.awt.event.InputEvent;
 import java.awt.event.KeyEvent;
 import java.awt.event.KeyListener;
 import java.awt.event.WindowAdapter;
@@ -114,6 +114,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -143,6 +144,7 @@ import javax.swing.JTabbedPane;
 import javax.swing.JTextArea;
 import javax.swing.JToggleButton;
 import javax.swing.JToolTip;
+import javax.swing.KeyStroke;
 import javax.swing.ListCellRenderer;
 import javax.swing.ListSelectionModel;
 import javax.swing.Popup;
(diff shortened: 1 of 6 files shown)


game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
```diff
@@ -669,9 +671,8 @@ public void focusGained(final FocusEvent e) {
     final MovePanel movePanel = new MovePanel(data, mapPanel, this);
     actionButtons = new ActionButtons(data, mapPanel, movePanel, this);
 
-    SwingUtilities.invokeLater(() -> mapPanel.addKeyListener(getFullScreenListener()));
+    addKeyBindings(movePanel, this);
     SwingUtilities.invokeLater(() -> mapPanel.addKeyListener(getArrowKeyListener()));
-    SwingUtilities.invokeLater(() -> mapPanel.addKeyListener(movePanel.getCustomKeyListeners()));
     SwingUtilities.invokeLater(() -> mapPanel.addKeyListener(getFlagToggleKeyListener(this)));
```
RoiEXLab: I believe we could simply a lot of code if similar places were migrated as well.
The whole "input map"/key listener thing is super inconsistent over the codebase

game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
```diff
@@ -167,7 +169,7 @@
 
 /** Main frame for the triple a game. */
 @Log
-public final class TripleAFrame extends JFrame {
+public final class TripleAFrame extends JFrame implements KeyBindings {
```
RoiEXLab: I don't quite agree with the term "KeyBindings" here.
In my opinion the interfaces name should make clear that it only supplies a collection of bindings, but doesn't actually manage them.
Maybe that's just me being silly, but just wanted to point out that this confused me for a moment

DanVanAtta: The thinking was an implementor "HasKeyBeindings". Any suggestions for alternatives?

game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
DanVanAtta: The other methods are pretty complex, we don't gain quite that much as the logic code in the listeners is a good bit greater than the infrastructure code of the key listener. I decided to leave it alone for now as that migration was a large effort. I had started down that path and it looked like a pretty expensive rabbit hole.

game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
RoiEXLab: `KeyControllable`?

game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
RoiEXLab: I must admit that `getFlagToggleKeyListener` is a mess.
This was one of my first contributions to the TripleA project, and it basically keeps a Thread alive as long as the key is held down to make sure that we get some sort of "key up" event even if the window loses focus.

Probably not the best idea to realize that but it worked I guess 🤷‍♂

game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
DanVanAtta: Not of a fan of that one,  `KeyBindingListener` maybe. TBH I'm pretty happy with `KeyBindings`

RoiEXLab: `KeyBindingListener` it is! 😄

DanVanAtta: Are we sure that is a better name?
A listener responds to an event, in this case the event could be interpreted as the key binding itself. IE: when a key is bound, fire this listener, rather than responding to a key press event. In that case we really want a 'KeyPressListener'; and we also have the issue that we have an API returning many listeners.

`KeyPressHandler` is maybe an okay name. Or maybe `KeyBindingSupplier`, but I feel like we're getting into the kingdom of nouns problem.

RoiEXLab: `KeyBindingSupplier` is pretty descriptive, because well it implements supplier.
I'd go with that one

DanVanAtta: There is a small issue of singular vs plural as this interface is returning many key bindings and not just one. It reads well enough though, I'm okay going with that name.

PR #5070, opened 2019-08-23
Simplify IDisplay, IRemotePlayer, IGame

RoiEXLab:
## Overview
This PR simplifies the object graph of IDisplay, IRemotePlayer and IGame, mainly by merging interfaces that are only used together anyways.
Since the we're network incompatible with 1.10 anyways, this seems like a good opportunity.

## Functional Changes
- [ ] New map or map update
- [ ] New Feature
- [ ] Feature update or enhancement
- [x] Code Removal
- [x] Code Cleanup or refactor
- [ ] Configuration Change
- [ ] Bug fix

## Testing

- [ ] Covered by existing automated tests
- [ ] Covered by newly added automated tests
- [x] Manually tested
- [ ] No testing done

### Manual Testing
 I verified I could play a normal game over the network as well as a local game with AI.


game-core/src/main/java/games/strategy/engine/player/IRemotePlayer.java
```diff
@@ -1,15 +1,259 @@
 package games.strategy.engine.player;
```
DanVanAtta: I see IGamePlayer and IRemotePlayer are merged.

There is an attempt to make an abstraction in the code base to hide when elements are actually across the network such that the network layer becomes transparent.

Are all players going to be `IRemotePlayer` instances? Have you considered perhaps renaming `IRemotePlayer` to `IGamePlayer`, would that make sens at all?

game-core/src/main/java/games/strategy/triplea/TripleA.java
```diff
@@ -32,41 +31,29 @@
...
-      // set sound channel to null to handle the case of shutdown being called multiple times.
-      // If/when shutdown is called exactly once, then the null assignment should be unnecessary.
-      soundChannel = null;
-    }
-
-    if (display != null) {
-      if (game != null) {
-        game.removeDisplay(display);
-      }
-      display.shutDown();
```
DanVanAtta: Is this shutdown a no-op? Why can it be removed? Is it located somewhere else now?

game-core/src/main/java/games/strategy/triplea/TripleA.java
```diff
@@ -98,7 +83,7 @@ public void startGame(
 
   @Override
   public Class<? extends IRemote> getRemotePlayerType() {
-    return ITripleAPlayer.class;
```
DanVanAtta: :+1:   Good to see ITripleAPlayer be consolidated and the hierarchy collapsed. It was a useful abstraction when there were multiple game types, now that things are focused it is past time for it to be simplified.

game-core/src/main/java/games/strategy/triplea/delegate/AbstractPlaceDelegate.java
```diff
@@ -540,7 +541,7 @@ private String moveAirOntoNewCarriers(
     }
     final List<Unit> fighters = producer.getUnitCollection().getMatches(ownedFighters);
     final Collection<Unit> movedFighters =
-        getRemotePlayer().getNumberOfFightersToMoveToNewCarrier(fighters, producer);
+        bridge.getRemotePlayer().getNumberOfFightersToMoveToNewCarrier(fighters, producer);
```
DanVanAtta: Are you sure this is safe? Just checking as I've ran across examples where the local data was not the same as the data on the bridge. It manifested as unexpected null data as the bridge data was only partially maintained.

RoiEXLab: 100% sure.
I noticed I haven't actually created seperate commits for this, but I basically inlined the getRemotePlayer() method call using IntelliJ refactoring

game-core/src/main/java/games/strategy/triplea/TripleA.java
RoiEXLab: Yeah, I made shutdown part of AbstractGame, so the display is shut down whenever the display is set to null.

game-core/src/main/java/games/strategy/engine/framework/AbstractGame.java
```diff
@@ -7,53 +7,60 @@
...
   protected final Vault vault;
-  protected IGameModifiedChannel gameModifiedChannel;
-  protected final PlayerManager playerManager;
   protected boolean firstRun = true;
 
-  protected AbstractGame(
+  IGameModifiedChannel gameModifiedChannel;
+
+  final Map<PlayerId, IRemotePlayer> gamePlayers = new HashMap<>();
+  final PlayerManager playerManager;
```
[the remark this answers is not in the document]
RoiEXLab: True, haven't even noticed that

game-core/src/main/java/games/strategy/engine/player/IRemotePlayer.java
RoiEXLab: I'm open to renaming stuff.
I'd have actually renamed the interface to just `IPlayer`/`Player` because there is only a single player left.
Basically all classes that implemented `IGamePlayer` implemented `ITripleAPlayer` as well, both of those interfaces extended `IRemotePlayer`, which is why I could merge those 3 interfaces no problem

DanVanAtta: Perhaps `GamePlayer` instead of `IRemotPlayer`?

DanVanAtta:
Remaining updates are pretty trivial @RoiEXLab (renaming), feel free to tackle it and merge or merge now.

game-core/src/main/java/games/strategy/engine/player/IRemotePlayer.java
RoiEXLab: Would you be unhappy with just `Player`? 
I don't think that name is taken yet

DanVanAtta: I'm not used to that as a name yet, still used to `GamePlayer`. It feels more specific, as opposed to say a player in lobby or a player in a network game staging screen. It might be my habituation though.

Ultimately no, with the above caveat - `Player` seems fine

RoiEXLab: IMO just `Player` is consistent with `PlayerId` or `TripleAPlayer`, so that's what I'll be going with.

game-core/src/main/java/games/strategy/triplea/TripleA.java
RoiEXLab: Side note: `TripleADisplay#shutDown` actually calls `TripleAFrame#shutDown`, so without this call the game window wouldn't close after the game stops.

Issue #5019, opened 2019-08-19
Testing process is non-streamlined and unintuitive


OrdoFlammae:
This is a minor issue, but I think there should be a better way to test the project before submitting pull requests. Right now, the process involves starting a docker process and a test http-server before running a shell-script that runs a bunch of checks. I see a few problems with this process. First, I can't do it inside my IDE, which means that every time I want to check something, I have to navigate outside of the IDE to do it, which breaks up the development process. Secondly, following from not having it in the IDE, I have to have multiple windows running around just to test the code. If we had the build process inside the IDE, it would be easier to run the multiple processes. Lastly, why are we using shell scripts to do this when we can just make another Gradle configuration to do it? That is what Gradle is for, right?

Ultimately, what I would like would be one Gradle task that runs through all the other necessary Gradle tasks, and does the whole process. Or maybe two, if one is too impractical, one to start up the server, and one to run all the tests (in which case, only one really needs to be implemented, the testing task will be simple).

I will be working on this in my spare time, but I would like some input from the development team who have been working on this project for a lot longer than me (ha ha). Do you think this is a good idea? Would you have any suggestions for implementing this?

RoiEXLab:
For unit tests alone you shouldn't have to leave your IDE at all. You can create a test config in IntelliJ IDEA and use that to run all the unit tests, not that you'll have to exclude JUnit 5 tags called "integration" for this to exclude integration tests that require a postgresql service running in the background with a database called ta_users IIRC, everything else should work there no problem.
 I must admit that I mostly rely on the travis build to run all the tests, but during development I find it more convinient to just run invididual classes directly from the IDE anyways.

You have a point though

DanVanAtta:
Testing of TripleA I would suggest is a critical issue. 


There are some interesting papers including TripleA (though somewhat dated and they look at TripleA from when it was hosted in sourceforge). One paper essentially finds that our tests have a number of problems (91% of tests contain test smells and it finds there is a correlation between test smells and maintenance effort):
- http://www.cs.loyola.edu/~binkley/papers/icsm-12-smells.pdf

Another defines metrics to find poor testing means the codebase is inefficient to work with (unmaintainable and brittle): https://testroots.org/assets/papers/athanasiouTSE.pdf


> (Page 17) 6.3.2  Hypothesis 2 : The relation between test code qualityand throughput

[...]

PR #5076, opened 2019-08-24
Add buttons for highlight units and flags, change flag toggle mode

DanVanAtta:
## Overview
Flags:
(1) Update flag display toggle to rotate through modes. Previously pressing L would toggle flags on for a short period. Now 'L' key will do the same as clickign the new flags button.
(2) Add new flags button. Clicking it once will turn on small flags, again for large flags, again to turn off.
(3) With a button added, remove flag selection sub-menu from the menu bar

Highlight Units:
- Add a button that does the same thing as pressing 'F' to highlight movable units.
- Highlight units button disappears when it is not 'your' turn

Update About Menu to have an Credits Section:
    - Adds a new section for credit citations in the about menu
    - New image icons do not have to accredition when they are under 100x100, it feels like good practice to give these images credit.

## Functional Changes

[ ] New map or map update
[X] New Feature
[ ] Code Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix

## Testing

[X] Manually tested
[ ] No testing done

## Screens Shots
Screenshot from 2019-08-23 20-05-41
Screenshot from 2019-08-23 20-06-07
Screenshot from 2019-08-23 22-16-02

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/FlagDrawMode.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/FlagDrawMode.java
@@ -22,7 +22,7 @@ class FlagDrawMode {
             UnitsDrawer.UnitFlagDrawMode.class,
             prefs.get(
                 UnitsDrawer.PreferenceKeys.DRAW_MODE.name(),
-                UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG.toString()));
+                UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG.name()));
     UnitsDrawer.setUnitFlagDrawMode(setting, prefs);
     UnitsDrawer.enabledFlags =
         prefs.getBoolean(
@@ -34,7 +34,7 @@ class FlagDrawMode {
       prefs.putBoolean(UnitsDrawer.PreferenceKeys.DRAWING_ENABLED.name(), true);
       prefs.put(
           UnitsDrawer.PreferenceKeys.DRAW_MODE.name(),
-          UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG.toString());
+          UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG.name());
       UnitsDrawer.setUnitFlagDrawMode(UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG, prefs);
       // check if small flags, if so then draw large flags
     } else if (UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG
@@ -47,7 +47,7 @@ class FlagDrawMode {
       prefs.putBoolean(UnitsDrawer.PreferenceKeys.DRAWING_ENABLED.name(), true);
       prefs.put(
           UnitsDrawer.PreferenceKeys.DRAW_MODE.name(),
-          UnitsDrawer.UnitFlagDrawMode.LARGE_FLAG.toString());
+          UnitsDrawer.UnitFlagDrawMode.LARGE_FLAG.name());
       UnitsDrawer.setUnitFlagDrawMode(UnitsDrawer.UnitFlagDrawMode.LARGE_FLAG, prefs);
       // otherwise we had large flags, turn drawing flags off
     } else {
@@ -55,7 +55,7 @@ class FlagDrawMode {
       prefs.putBoolean(UnitsDrawer.PreferenceKeys.DRAWING_ENABLED.name(), false);
       prefs.put(
           UnitsDrawer.PreferenceKeys.DRAW_MODE.name(),
-          UnitsDrawer.UnitFlagDrawMode.LARGE_FLAG.toString());
+          UnitsDrawer.UnitFlagDrawMode.LARGE_FLAG.name());
       UnitsDrawer.setUnitFlagDrawMode(UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG, prefs);
     }
     mapPanel.resetMap();
--- a/game-core/src/main/java/games/strategy/triplea/ui/FlagDrawMode.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/FlagDrawMode.java
@@ -1,0 +1,63 @@
+package games.strategy.triplea.ui;
+
+import games.strategy.triplea.ui.screen.UnitsDrawer;
+import java.util.prefs.Preferences;
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+
+/**
+ * Class to change how unit flags are drawn on the map. Unit flags are flags drawn underneath or
+ * below units to make it easier to identify which nation they belong to. By default flag drawing is
+ * turned off, toggling once turns small flags on, toggle again to turn on large flags and again to
+ * turn off.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+class FlagDrawMode {
+
+  static void toggleNextDrawMode(final MapPanel mapPanel) {
+
+    final Preferences prefs = Preferences.userNodeForPackage(FlagDrawMode.class);
+    final UnitsDrawer.UnitFlagDrawMode setting =
+        Enum.valueOf(
+            UnitsDrawer.UnitFlagDrawMode.class,
+            prefs.get(
+                UnitsDrawer.PreferenceKeys.DRAW_MODE.name(),
+                UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG.toString()));
+    UnitsDrawer.setUnitFlagDrawMode(setting, prefs);
+    UnitsDrawer.enabledFlags =
+        prefs.getBoolean(
+            UnitsDrawer.PreferenceKeys.DRAWING_ENABLED.name(), UnitsDrawer.enabledFlags);
+
+    // draw mode is off, toggle small flag
+    if (!prefs.getBoolean(UnitsDrawer.PreferenceKeys.DRAWING_ENABLED.name(), false)) {
+      UnitsDrawer.enabledFlags = true;
+      prefs.putBoolean(UnitsDrawer.PreferenceKeys.DRAWING_ENABLED.name(), true);
+      prefs.put(
+          UnitsDrawer.PreferenceKeys.DRAW_MODE.name(),
+          UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG.toString());
+      UnitsDrawer.setUnitFlagDrawMode(UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG, prefs);
+      // check if small flags, if so then draw large flags
+    } else if (UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG
+        .toString()
+        .equals(
+            prefs.get(
+                UnitsDrawer.PreferenceKeys.DRAW_MODE.name(),
+                UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG.toString()))) {
+      UnitsDrawer.enabledFlags = true;
+      prefs.putBoolean(UnitsDrawer.PreferenceKeys.DRAWING_ENABLED.name(), true);
+      prefs.put(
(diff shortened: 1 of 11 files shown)


game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
```diff
@@ -772,14 +767,19 @@ private void addKeyBindings(final KeyBindingSupplier... keyBindings) {
...
         () ->
             gameCenterPanel.setDividerLocation(
                 (gameCenterPanel.getDividerLocation()
                         <= gameCenterPanel.getMaximumDividerLocation())
                     ? 1.0
                     : gameCenterPanel.getLastDividerLocation()));
+    bindings.put(
+        KeyBindingSupplier.fromKeyEventCode(KeyEvent.VK_L),
+        () -> FlagDrawMode.toggleNextDrawMode(mapPanel));
+    return bindings;
```
RoiEXLab: Since java 9 there exists Map.of to create maps with a small amount of keys

game-core/src/main/java/games/strategy/triplea/ui/FlagDrawMode.java
```diff
@@ -0,0 +1,63 @@
...
+
+  static void toggleNextDrawMode(final MapPanel mapPanel) {
+
+    final Preferences prefs = Preferences.userNodeForPackage(FlagDrawMode.class);
+    final UnitsDrawer.UnitFlagDrawMode setting =
+        Enum.valueOf(
+            UnitsDrawer.UnitFlagDrawMode.class,
+            prefs.get(
+                UnitsDrawer.PreferenceKeys.DRAW_MODE.name(),
+                UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG.toString()));
```
RoiEXLab: I know this is copied code, but I'm pretty sure `UnitsDrawer.UnitFlagDrawMode.valueOf("something")` would work as well.

game-core/src/main/java/games/strategy/triplea/ui/FlagDrawMode.java
```diff
@@ -0,0 +1,63 @@
...
+          UnitsDrawer.PreferenceKeys.DRAW_MODE.name(),
+          UnitsDrawer.UnitFlagDrawMode.LARGE_FLAG.toString());
+      UnitsDrawer.setUnitFlagDrawMode(UnitsDrawer.UnitFlagDrawMode.LARGE_FLAG, prefs);
+      // otherwise we had large flags, turn drawing flags off
+    } else {
+      UnitsDrawer.enabledFlags = false;
+      prefs.putBoolean(UnitsDrawer.PreferenceKeys.DRAWING_ENABLED.name(), false);
+      prefs.put(
+          UnitsDrawer.PreferenceKeys.DRAW_MODE.name(),
+          UnitsDrawer.UnitFlagDrawMode.LARGE_FLAG.toString());
```
RoiEXLab: Using `Enum#name()` is safer, as name can't be overriden and is guaranteed to be the correct name that can be parsed again.

game-core/src/main/java/games/strategy/triplea/ui/FlagDrawMode.java
```diff
@@ -0,0 +1,63 @@
...
+        prefs.getBoolean(
+            UnitsDrawer.PreferenceKeys.DRAWING_ENABLED.name(), UnitsDrawer.enabledFlags);
+
+    // draw mode is off, toggle small flag
+    if (!prefs.getBoolean(UnitsDrawer.PreferenceKeys.DRAWING_ENABLED.name(), false)) {
+      UnitsDrawer.enabledFlags = true;
+      prefs.putBoolean(UnitsDrawer.PreferenceKeys.DRAWING_ENABLED.name(), true);
+      prefs.put(
+          UnitsDrawer.PreferenceKeys.DRAW_MODE.name(),
+          UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG.toString());
```
RoiEXLab: Same here with name VS toString

game-core/src/main/java/games/strategy/triplea/ui/FlagDrawMode.java
```diff
@@ -0,0 +1,63 @@
...
+        .toString()
+        .equals(
+            prefs.get(
+                UnitsDrawer.PreferenceKeys.DRAW_MODE.name(),
+                UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG.toString()))) {
+      UnitsDrawer.enabledFlags = true;
+      prefs.putBoolean(UnitsDrawer.PreferenceKeys.DRAWING_ENABLED.name(), true);
+      prefs.put(
+          UnitsDrawer.PreferenceKeys.DRAW_MODE.name(),
+          UnitsDrawer.UnitFlagDrawMode.LARGE_FLAG.toString());
```
RoiEXLab: Name VS toString

game-core/src/main/java/games/strategy/triplea/ui/FlagDrawMode.java
```diff
@@ -0,0 +1,63 @@
...
+    if (!prefs.getBoolean(UnitsDrawer.PreferenceKeys.DRAWING_ENABLED.name(), false)) {
+      UnitsDrawer.enabledFlags = true;
+      prefs.putBoolean(UnitsDrawer.PreferenceKeys.DRAWING_ENABLED.name(), true);
+      prefs.put(
+          UnitsDrawer.PreferenceKeys.DRAW_MODE.name(),
+          UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG.toString());
+      UnitsDrawer.setUnitFlagDrawMode(UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG, prefs);
+      // check if small flags, if so then draw large flags
+    } else if (UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG
+        .toString()
```
RoiEXLab: Same here

DanVanAtta:
> As is I don't think this is a beneficial change.

Not at all? It's okay for a magic hotkey "L" to be known by nobody and never used?

>  It is dividing flag settings from the rest of the settings 

The settings are already scattered and are a hodge.Highlight is similar and does not even have a setting or UI control.

>  and not really considering maps that have many resources which already fill up most of the bottom bar.

Not enough space on the bottom bar is not compelling. We added space there for when connected to dice servers, was it not a problem then? We also do have space, but we need to refine our error messages and space given to the territory indicator. It's an existing problem, 80px is not making it that much worse.

Grouping controls is till valuable, and consistency of information vs functional displays I find to be more compelling.

@ron-murhammer please going forward offer suggestions of what you would accept or do think could be improved. The "I don't like this, you're not solving anything" kind of response is not encouraging. I also think we are drastically way too used to MAJOR problems, "the normalcy of the absurd" and we over-index on insignificant problems and underindex on the ones we are used to.

---------

Per the forum thread, I'm relocating the buttons to be grouped in the unit scroller. "Center", "highlight" and "flags" are somewhat similar and this avoids adding functional controls to the bottom bar.

DanVanAtta:
@ron-murhammer , a good read:  https://blog.submain.com/toxic-code-review-culture/  , notably the section: "Saying “This Is Wrong”"

game-core/src/main/java/games/strategy/triplea/ui/FlagDrawMode.java
DanVanAtta: Only need to point out items once. I'm looking for a good reading that went into detail and am turning up a blank. I recall the reasoning was to avoid the 'avalance' effect of many comments. Leaving a review that has many, many comments works against the reviews goals and is subtly saying "you did a shit job, you don't know how to code". I wish I could find that resource for the reading, it was interesting.

This section had some interesting and pertinent points IMO as well:
> https://blog.asana.com/2016/12/7-ways-to-uplevel-your-code-review-skills/
> 7. Always give approval, unless you can prove that there is a bug

The UX change here is significant though, it's okay to let a review be open for a bit to fully consider it and to simply state that as well.

DanVanAtta:
@ron-murhammer , @RoiEXLab I recommend trying out UX changes before jumping into reviews; I think that could have lead to more valuable feedback. I think we've hit a good example of a number of review anti-patterns and are suffering from being myopic. I'm also pretty aware of how much we have to do and our rate of progress. Such a problem is an existential problem to this project, see this comment: https://github.com/triplea-game/triplea/issues/5019#issuecomment-523631088 (@ron-murhammer please note the referenced scholarly articles that look at TripleA, perhaps you may find them interesting, at the very least it's pretty cool that Master's and PHD's are pulling data from TripleA)

game-core/src/main/java/games/strategy/triplea/ui/FlagDrawMode.java
RoiEXLab: True, but in this case this is my own code I wrote almost 3 years in the past 😅 
I agree that not all of those comments are necessary, but I felt like there were some hidden toString statements in unexpected places

RoiEXLab:
@DanVanAtta I don't have any preference with the UX, so I only really looked at the individual code snippets. Leaving it up to the others to sort out the details

DanVanAtta:
Latest, trying it out, the center units and highlight next to each other work much better than I expected. Flag is not ideal, but it is related to finding units and there are other problems with it already  that are somewhat mitigated.

Screenshot from 2019-08-24 13-02-25

The scroller does get a bit busy IMO, but I suspect that is habituation. I think we may have found a decent compromise.

game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
DanVanAtta: Let's leave that for a future refactor please, perhaps we can go across the board and apply that update to more places than just here.

game-core/src/main/java/games/strategy/triplea/ui/FlagDrawMode.java
DanVanAtta: Understood, the thing to be careful about is the perception of "Oh, I know this code, it's mine and it is not great. Could you please do these extra things beyond the scope of your goal to make improvements I should have made some time ago? You know, while you're here.."

To be clear, I think I understand your motivation and I appreciate improving the code. Just want to convey that it runs a risk of the above perception. 

At the same time it's difficult to know how exactly we clean up the TripleA codebase. I think the answer is to move faster and pick our battles. We perhaps should let smaller things slide and do a better job of adding tests and larger rewrites where most of these points could be mooted. The danger I see is moving slowly on small cleanups that would be removed with a more significant rewrite. At the same time, it's hard to know if we'll ever get those tests and the more significant rewrite. It is a balance I suppose that is not easy to strike.

DanVanAtta:
@ron-murhammer in your hands now. Updated

game-core/src/main/java/games/strategy/triplea/ui/menubar/ViewMenu.java
```diff
@@ -435,72 +431,6 @@ private void addLockMap() {
...
-      final ButtonGroup group,
-      final ActionListener action,
-      final boolean selected) {
-    final JRadioButtonMenuItem buttonItem = new JRadioButtonMenuItem(text);
-    buttonItem.addActionListener(action);
-    buttonItem.setSelected(selected);
-    group.add(buttonItem);
-    return buttonItem;
-  }
-
```
ron-murhammer: I would like to keep the unit flags as a view menu option here as they fit well with the other unit UI settings such as unit size, show units, etc.

DanVanAtta: Okay, updated: 7ff1c8e80
What's your long term vision @ron-murhammer for the options and controls of this feature?

ron-murhammer: Well for unit flags in particular, I'd like them to be per map settings like unit size. In terms of location, they are more of a UI setting and would like to eventually refactor the top settings toolbar into more of a modal window like most games have.

Issue #5060, opened 2019-08-23
F Key highlights wrong units after ending turn - Unit Scroller Keys active during all turn phases


DanVanAtta:
### Can you describe how to trigger the error? (eg: what sequence of actions will recreate it?)
- During purchase, use the 'n', 'm' or 'c' keys to cycle through units
  - Notice: unit scroller work even though it is not a move phase
- End a turn, during purchase phase, press 'f' key
  - The units of the previous turn light up. For example, ending russian turn, during German purchase, the russian units light up

### Instead of this error, what should have happened?
- Unit scroller keys should be no-op when the unit scroller buttons are not visible
- Highlight units should only work during move phase

![Screenshot from 2019-08-22 23-18-28](https://user-images.githubusercontent.com/12397753/63570794-2bc75b00-c533-11e9-9d64-87024e418fc5.png)

DanVanAtta:
'F' key problem is solved with #5077. Unit scroller keys still need to be fixed.

DanVanAtta:
Not sure if the unit scroller being active all the time is actually a problem. Closing for now.

PR #5077, opened 2019-08-24
Fix highlight units, correct turn

DanVanAtta:
## Overview
Action panel current player does not increment correctly on first phase of turn (at least). This fix grabs the actual current player which allows the highlight current units to highlight the right units. Previously on a purchase phase the units of the previous turn would be higlighted.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Code Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[X] Bug fix

https://github.com/triplea-game/triplea/issues/5060

Previous 'currentPlayer' value does not update correctly. (Why is unknown)

Use gameData getSequence current player which does update to be the latest player

## Testing

[X] Manually tested
[ ] No testing done


RoiEXLab:
Would be nice to figure out the root cause

Issue #3783, opened 2018-08-11
Forbid Usernames With No Letters


Cernelius:
Currently in lobby there is a user called "---".

That is not against any rules I know of, but I think it would be better if names like that or also numbers-only are not allowed.

Also maybe up to 12 characters only.

Cernelius:
I meant that being merely engine enforced, in the moment you try to log in; not the mods banning or adding it to the rules or anything like that.

Cernelius:
Now there is one called "----", instead (not sure if the same user).

DanVanAtta:
@Cernelius reviving this old topic, would you mind sharing a recommendation for validation rules? We have a lot of flexibility for what to reject/accept, I'm curious if you have any thoughts for a starting point.

prastle:
We could give you a long list but the simple answer is an l can't look like an I
 in the forum or on logins nor any other letter. L i

DanVanAtta:
Well, we need an exact requirement list to code up. The easiest thing is to allow anything. Any other rules we need to explicitly add.

I'm thinking:
- min length 3
- max length 15
- alpha-numeric only, no symbols

It's not worth going to town on the rules, they should be simple here or it'll be confusing and/or painful to code.

RoiEXLab:
@DanVanAtta A thing to consider is potentially asian characters we want to allow.

Cernelius:
These are my suggestions, but under them a few users will have to remake their names:

1. Must be 5 to 20 characters long.
2. Nothing is case sensitive (Cernel=cernel=CERNEL).
3. Only latin letters plus the W, arabic numbers, the underscore and the short dash are acceptable (so no spaces either).
4. Must start with a letter (not with numbers or dash or underscore) (which implies they will have at least a letter in them).

Cernelius:
Right, this is off topic, but both in the forum and the lobby please be sure to adopt fonts that are highly distinctive (typical example, the lower case "L" being easy to tell apart from the upper case "I" (the first one should be curved to the right on top and the second one should have horizontal bars both above and below the vertical body)).

panther2:
When the forums of axisandallies .org switched to NodeBB recently, we experienced PBF-problems with usernames that contained spaces or certain special characters.
For details please see: 
https://www.axisandallies.org/forums/topic/32495/new-triplea-with-new-forums-user-does-not-exist-error
and
https://www.axisandallies.org/forums/post/1237061

[...]

PR #5079, opened 2019-08-24
Disallow login with name that is already logged in

DanVanAtta:
## Overview
Add a case insensitive check that a player name does not match any that are logged in. For example, if player "TEST" is logged, the name "test" would be rejected. The check for registered user name is case insensitive, it should then not be possible to use this to block registered users from logging in.

Addresses requested naming requirement item (2) "Nothing is case sensitive (Cernel=cernel=CERNEL)." from: https://github.com/triplea-game/triplea/issues/3783#issuecomment-465382253

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Code Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix

## Testing

[X] Manually tested
[ ] No testing done

### Manual Testing
Note: Had to hack out the duplicate login name check and hack out the filter for same-machine login to verify that the duplicate name check worked.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/lobby/PlayerNameValidation.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/PlayerNameValidation.java
@@ -1,6 +1,7 @@
 package games.strategy.engine.lobby;
 
 import com.google.common.annotations.VisibleForTesting;
+import java.util.Collection;
 import lombok.NoArgsConstructor;
 import org.triplea.lobby.common.LobbyConstants;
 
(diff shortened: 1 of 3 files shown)


DanVanAtta:
@RoiEXLab no, see: https://github.com/triplea-game/triplea/pull/5079/files#diff-8fe620f6ffa239c46f6b25104ce93ce3R120, which filters out same network when considering if already logged in with the same name. Of note, that logic is already after we 'dupe' names if they are coming from the same network.

Issue #4875, opened 2019-06-07
Version Numbering Problem and Maybe the Final Discussion!!!


DanVanAtta:
Now that we are two version numbers plus build number there is a problem:

Example: v1.10 and v1.11 are not expected to be compatible.

We are currently in a situation where we *want* users to do a compatible upgrade from version 1.9.x to a later version of 1.9. If we only have 1.10 and build number, we will *not* be able to indicate this with version number.

I think there are three options before us:

### (1) use build number when communicating to users, eg: 1.10+23525
Pretending the problem is with 1.10, and that we wanted to let users know they should upgrade, we'd tell them to "upgrade to the latest release: 1.10.23525"

### (2) Add in a build number, eg: 1.10.0+23525
Again, pretending we have this problem with 1.10.0, we'd ask users to "upgrade to version 1.10.1"

(note this very handy diagram that explains what each number represents: https://user-images.githubusercontent.com/12397753/51069660-64d83e00-15e8-11e9-8961-4c9c3d393308.png)

### (3) Use the first number for incompatible and second number is compatible, eg: 2.0+23525
Again pretending we had the problem with 1.10, we'd tell users they'd need to go to 1.11 and 1.11 would be compatible with 1.10. Version 2.0 would not be compatible with 1.x

## Second topic: use a plus side for splitting build number

I don't think this one is controversial, on the last topic of version numbers it was recommended to use a "+" to designate the build number, eg: `1.10+23525`

### My thoughts

- AFAIK it's not a best practice to use build number extensively nor for external communications. This makes me shy away from option 1
- I like option 3 as it makes good use of the first version number, I'm a bit skeptical we'll go to a version 2 or a version 3 ever, and I don't think it'll be really easy to create a definition of when the major version should be adjusted.
- I like option 2 as it is essentially semver

What are people's votes? Please indicate if you have strong feelings as well or if you're just voting. Let me know if the problem statement could use more clarification. Sorry to drag this topic up again, I hope we can hit resolution on it this time.

RoiEXLab:
Difficult question: I'd actually prefer 2 or 3 with a slight tendency to 3

ssoloff:
Vote: (3) (which seems more like semver to me)

DanVanAtta:
I always thought semver had 4 numbers rather than 3. :man_shrugging: 
Appreciate the vote!

[...]

PR #4048, opened 2018-09-15
Replace GameEnginePropertyReader with ProductConfiguration

ssoloff:
## Overview

Follow up to #4029.

It was noted that "game engine properties" doesn't really make sense any longer for the following reasons:

1. It only contains a single version property.
1. That version is applicable to more than just the game engine; it applies to every artifact produced from this repo (e.g. game client, lobby server, HTTP server, etc.).

Therefore, this PR generalizes the "game engine version" into a "product version."  Overall, not much changes functionally from #4029.  It's mostly an exercise in renaming things.  Although, there were some build system changes described under Functional Changes below.

## Functional Changes

* Changed how the product version is generated during the Gradle build:
    * _game-core/src/main/resources/META-INF/triplea/product.properties_ is now the Single Source of Truth for the first four components (major, minor, point, micro) of the product version (previously, one had to change both _game_engine.properties_ and _.travis.yml_ when any of these components were updated).
    * _.travis.yml_ is now the Single Source of Truth for the fifth component (build number a.k.a. qualifier) of the product version.
    * Gradle is now the Single Source of Truth for the complete five-component product version (i.e. it merges the value from _product.properties_ and the build number specified in the project property `productVersionQualifier`).  Previously, this was spread around the Gradle buildscripts and _.travis.yml_.

## Refactoring Changes

* Renamed `GameEnginePropertyReader` to `ProductConfiguration` and moved it to the `org.triplea.common.config.product` package.  `org.triplea.common` was chosen because this applies to more than just the game client.  Even though the lobby server has a legitimate dependency on `game-core` at this time, I suspect we'll see more code that fits into a non-game package that applies to multiple projects.  At some point in the future, we'll probably even extract a `common` subproject in the Gradle build.
* Moved the existing types in `g.s.engine.config` to `o.t.common.config` for the same reason as the previous point.
* Applied `version = getProductVersion()` to _all_ Gradle subprojects instead of duplicating it in each subproject buildscript.

## Manual Testing Performed

* Built a full release in my fork and verified the following:
    * All artifacts have the correct version in their file names.
    * All artifacts have the correct version in their fat jar names.
    * Running the headed client from the following artifacts displayed the correct product/engine version on the main menu:
        * Portable build
        * Unix installer
        * Windows x86_64 installer
* Verified running the headed client from Eclipse did not produce any unintended side effects with a full version string of `1.9.0.0.@qualifier@` (see below).

## Additional Review Notes

* Some notes on how the version qualifier is replaced:
[...]

PR #5080, opened 2019-08-24
Simplify Version and set to 2.0.0

DanVanAtta:
## Overview

Set version to 2.0.0; Fits with new version criteria where first value is engine compatibility, second is compatible release number, third is build number. See: https://github.com/triplea-game/triplea/issues/4875 for more details and discussion.

(1) The 'lesserThan' API usage can be replaced by using 'GreaterThan' and swapping argument positions

(2) Remove '@buildId@' replacement in Gradle and use '0' for the dev build. This can be done by having travis overwrite the properties file with the build number, at which point gradle can read the property value directly without doing a token replacement. Meanwhile we can set the local version number to be '2.0.0' and then local builds do not need to do a token replacement to 'dev'. Since the last number is the build number is always increasing, a '0' for the build number clearly indicates a 'dev' build and we do not need to use a special 'dev' string to represent it.

(3) Simplify parsing of Version, use 'splitter' instead of a regex

(4) Use lombok to simplify Version getters and Setters, equals and hashcode

(5) Remove EngineVersion wrapper and move those methods to Version itself

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Code Removal
[X] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix

## Testing

[X] Manually tested
[ ] No testing done

## Additional Review Notes

For follow-up and future updates:
- ensure compatibility logic for engine is updated to look only at 'major' number
- consider map compatibility


game-core/src/main/resources/META-INF/triplea/product.properties
```diff
@@ -1 +1 @@
-version = 1.10.@buildId@
+version = 2.0.0
```
RoiEXLab: I'm not quite sure if I like this change.
I feel like at some point someone is going to increment the "patch" number in semver terms and wonder why everything breaks

RoiEXLab:
@ssoloff I'd appreciate if you could have a quick look on the diff because you introduced the buildid thingy IIRC

ssoloff:
> @ssoloff I'd appreciate if you could have a quick look on the diff because you introduced the buildid thingy IIRC

I guess the biggest thing I see is that the build is no longer entirely reproducible from Gradle alone--the build ID replacement piece is now part of the Travis buildscript.  (Actually, that may have been partially broken in a previous change when [this line](https://github.com/triplea-game/triplea/pull/5080/files#diff-354f30a63fb0907d4ad57269548329e3L24) was added in #4892.  The idiomatic way to get the full product/release version would be to let Gradle tell you what it is; see the [similar code](https://github.com/triplea-game/triplea/blob/master/.travis/push_tag#L9) in the _.travis/push_tag_ script.  You could also just specify a wildcard in the jar filename to avoid needing to know the release version altogether.)  The nice thing about the previous solution was that one could reproduce the identical release artifacts from a local build simply by running `./gradlew -PbuildId=whatever release`.  I used that feature in the past to provide private builds to users for testing (so as not to provide several builds with the same `dev` ID with the side effect that the user may not  know which one they are testing).

Also, by modifying _product.properties_ in place, the build is no longer idempotent.  While, in this case, it only affects Travis, a non-idempotent build is generally a Bad Thing.  The previous solution sent the modified _product.properties_ to the _build_ folder, the canonical location for such things.

I'm pretty sure these issues were discussed in #4048 and some of the PRs that preceded it.

game-core/src/main/resources/META-INF/triplea/product.properties
DanVanAtta: That would indicate a misunderstanding of what the numbers represent and how they are used.

We could leave off the last '.0' and append it with the build number. I like though having in dev the build number to look like it would in production and have the same number of digits. That is purely for consistency between dev and prod, to reduce differences  which means our testing of dev carries over more to production. Having behavior that is production only is risky and can't be tested by definition.

DanVanAtta:
Thanks for the comments, it's a good discussion.
In large part this first update is a simplification; As can be observed by the net removal of lots of code, we built up a lot of over-complexity around version number.

DanVanAtta:
> The nice thing about the previous solution was that one could reproduce the identical release artifacts from a local build simply by running ./gradlew -PbuildId=whatever release

Interesting use-case; I could see us adding in a flag to allow this. On the other hand, I'm not sure if it's truly good to provide a version that looks like an official artifact. If wanting to provide a user with a given version, then if it's built, it is available on releases. If we need to modify the artifact, then it's a different version and should not have the same build number.

> Also, by modifying product.properties in place, the build is no longer idempotent. While, in this case, it only affects Travis, a non-idempotent build is generally a Bad Thing. 

I don't quite understand how that is the case. If we re-trigger travis, I think it uses the same build number, everything is done from clean and the same update operation is done. If testing locally, the regex for '.0$' would fail to match after the first iteration and you would wind up with the same build number (but at the cost of a modified file). The local testing use-case for this is not terribly significant, given how often we modify the build, a few modified files while testing this out if we need to make changes I think is a pretty minor cost compared to say for example needing to get the license key, re-downloading artifacts that do land in the working directory and also show up as new files that need to be deleted.

DanVanAtta:
Once these simplifications change land, we need to still update game logic to behave correctly for compatibility flags. Notably version compat should only look at the leading digit in the version string.

ssoloff:
I still think it's a step backwards to take build logic that is already managed by Gradle and move it to _.travis.yml_.  As has been discussed before, _.travis.yml_ is simply the glue between Travis and the TripleA build system, and it should ideally contain only that which is absolutely necessary to have Travis run the Gradle build and deploy the artifacts.

DanVanAtta:
@ssoloff I really appreciate the consideration and the time and effort into this discussion.

I will make two assertions:
- I think it's perfectly fine to allow for the build number of "0" to represent the "dev" build number. I actually prefer that as then the version number better matches the format that we would see it in production.
- Build artifacts should be immutable and build artifacts with the same build number should match. This can *always* be hacked of course, but in general if there is updated code, the build number should not match. Therefore I think if we custom build an artifact for someone, it should not have a build number matching any existing one, with special understanding that "0" for build number is a custom build.


My thinking is that:
- (1) we already have elements from the Travis build process injected into Gradle.
- (2) Without running travis, we get a bogus build id value "dev" that then needs to be handled with special case logic in gradle and code.
- (3) If we do not run gradle and build from source (IDE build), we have a similar problem where we have a '@buildId@' value in version, but this results in a 3rd variant for what the version number would look like when running the app.

Hence, if you simulate the build process locally, you either have:
(A) set the travis build number (eg: 5555), in which case you get "2.0.5555"
(B) Run from gradle -> "2.0.dev"
(C) Run from source -> "2.0.@buildId@"

Following this update, the build number becomes unified for cases (B) and (C) for local simulation, running through either gradle or building from source will result in the same "2.0.0" build number.


To expound on (1) above, travis is involved with the gradle build already. Notably it grabs the travis build number from env variable and then makes that variable available to gradle. That is done in 'setup_gradle' (and so this is the part where I think where we already had travis interacting with gradle):


```
GRADLE_USER_HOME=~/.gradle
GRADLE_USER_PROPERTIES="${GRADLE_USER_HOME}/gradle.properties"

mkdir -p "$GRADLE_USER_HOME"
echo "buildId=$TRAVIS_BUILD_NUMBER" >> "$GRADLE_USER_PROPERTIES"
```

Hence, the previous process I would summarize as the following 3 steps:

(1) Travis passes build number to gradle properties file, making the build number available to gradle
(2) Gradle updates product.properties replacing '@buildId@' with the build number
(3) Gradle reads version number from product.properties

In this update this is reduced to 2 steps:

(1) Travis updates product.properties
(2) Gradle reads version number from product.properties


So in both cases I see travis as interacting with gradle. The goal of this update is then to short-circuit the logic, simplify it so that gradle does not have to any computations and travis is still the same interacting with gradle through one file. My hope is that this will create easier to understand build logic, with less variable replacement and fewer moving pieces. Given this update does drop a number of lines from the build file, I think there is evidence of that outcome in the diff. Hence, I hope this would not be considered a step backwards but instead a simplification of how travis is interacting with gradle.

DanVanAtta:
@RoiEXLab I added another update simplifying things a bit further: https://github.com/triplea-game/triplea/pull/5080/commits/fe167738fe2cdc64aca42ff3af276660cc78c9ee, that removes some dead code from the build process

ssoloff:
> To expound on (1) above, travis is involved with the gradle build already. Notably it grabs the travis build number from env variable and then makes that variable available to gradle. That is done in 'setup_gradle' (and so this is the part where I think where we already had travis interacting with gradle)

Right, that's the "glue" that ties the Travis `TRAVIS_BUILD_NUMBER` environment variable to the Gradle `buildId` project property.  It's equivalent to passing `-PbuildId=$TRAVIS_BUILD_NUMBER` to every invocation of Gradle within _.travis.yml_.  It's a one-way data dependency from Travis to Gradle.

> In this update this is reduced to 2 steps:
>
> (1) Travis updates product.properties
> (2) Gradle reads version number from product.properties

Not quite.  You also have:

(3) Travis reads version number from _product.properties_ in order to run the HTTP server.

So that you now have the (rather obscure) path to _product.properties_ in three locations spread across two subsystems.  Whereas, prior to #4892, this path was managed in a single location within Gradle alone.

I think some of the confusion over who should be responsible for what comes from an artificial dependency Travis has on the release version in this set of lines:

https://github.com/triplea-game/triplea/blob/7623a208c0fc6c908d2cb5c89fe3f8c82ba4d66a/.travis.yml#L24-L26

As mentioned above, it shouldn't be necessary to know the release version to launch the HTTP server.  In addition to the proposals mentioned in my original comment, it should also be possible to spin up this server via Gradle directly using the `:http-server:run` task.  That would obviate the need for querying the release version completely at this point, as well as hiding the detail of needing to run the `shadowJar` task beforehand.

Issue #5010, opened 2019-08-17
Game Freeze During Battle - Retreat Option Prompted too Early


DanVanAtta:
## Problem
- Started a battle and saw a retreat dialog appear before the first battle dice were rolled
- Note, no buttons are rendered or clickable, game seems effectively frozen

Expected:
- no retreat dialog, attack dice to be rolled

Stack trace from thread dumps:
```
"Thread-74" #167 prio=6 os_prio=0 cpu=60567.05ms elapsed=4875.98s tid=0x00007f930c38c800 nid=0x259c in Object.wait()  [0x00007f93265b4000]
   java.lang.Thread.State: WAITING (on object monitor)
	at java.lang.Object.wait(java.base@11.0.2/Native Method)
	- waiting on <no object reference available>
	at java.lang.Object.wait(java.base@11.0.2/Object.java:328)
	at java.awt.EventQueue.invokeAndWait(java.desktop@11.0.2/EventQueue.java:1361)
	- waiting to re-lock in wait() <0x0000000726a151a0> (a java.awt.EventQueue$1AWTInvocationLock)
	at java.awt.EventQueue.invokeAndWait(java.desktop@11.0.2/EventQueue.java:1342)
	at javax.swing.SwingUtilities.invokeAndWait(java.desktop@11.0.2/SwingUtilities.java:1480)
	at org.triplea.swing.SwingAction.invokeAndWaitResult(SwingAction.java:130)
	at org.triplea.swing.SwingAction.invokeAndWait(SwingAction.java:102)
	at games.strategy.triplea.ui.BattlePanel.lambda$showBattle$11(BattlePanel.java:210)
	at games.strategy.triplea.ui.BattlePanel$$Lambda$2058/0x00000008009dfc40.run(Unknown Source)
	at org.triplea.java.Interruptibles.lambda$await$0(Interruptibles.java:41)
	at org.triplea.java.Interruptibles$$Lambda$50/0x0000000800106c40.get(Unknown Source)
	at org.triplea.java.Interruptibles.awaitResult(Interruptibles.java:78)
	at org.triplea.java.Interruptibles.await(Interruptibles.java:39)
	at games.strategy.triplea.ui.BattlePanel.showBattle(BattlePanel.java:208)
	at games.strategy.triplea.ui.display.TripleADisplay.showBattle(TripleADisplay.java:45)
	at jdk.internal.reflect.GeneratedMethodAccessor76.invoke(Unknown Source)
[...]

PR #5086, opened 2019-08-25
Use invokeLater instead of invokeAndWait

DanVanAtta:
## Overview

Bit of a blind fix to try and avoid deadlock in BattlePanel with 'invokeAndWait'.

We are seeing deadlock happen with a call to 'setVisible' from within EDT while another thread is calling invokeAndWait. With luck by calling 'invokeLater' instead, the 'setVisible' call will be queued to be called after the other thread executes invokeAndWait.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Code Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[X] Bug fix

https://github.com/triplea-game/triplea/issues/5010

Deadlock in EDT calling setVisible. How this happens exactly is not clear.

Blind fix, try using 'invokeLater' instead of invokeAndWait

## Testing

[X] Manually tested
[ ] No testing done

### Manual Testing
Verified could not repro any problems fighting combats with save game attached to https://github.com/triplea-game/triplea/issues/5010

## Additional Review Notes
There is only really a couple lines changed at the beginning of the updated method, lots of indentation changes.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/BattlePanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/BattlePanel.java
@@ -205,68 +205,63 @@ public final class BattlePanel extends ActionPanel {
       final boolean isAmphibious,
       final BattleType battleType,
       final Collection<Unit> amphibiousLandAttackers) {
-    Interruptibles.await(
-        () ->
-            SwingAction.invokeAndWait(
-                () -> {
-                  if (battleDisplay != null) {
-                    cleanUpBattleWindow();
-                    currentBattleDisplayed = null;
-                  }
-                  if (!getMap().getUiContext().getShowMapOnly()) {
-                    battleDisplay =
-                        new BattleDisplay(
-                            getData(),
-                            location,
-                            attacker,
-                            defender,
-                            attackingUnits,
-                            defendingUnits,
-                            killedUnits,
-                            attackingWaitingToDie,
-                            defendingWaitingToDie,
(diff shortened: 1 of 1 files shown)


DanVanAtta:
I'll emphasize this is a blind-fix, a guess, I'm not at all sure if invokeLater will be a magic bullet; though it seems it could work.

@ron-murhammer I hadn't done too much background research into latest changes. Without a repro case it's really difficult to know when such a race condition was introduced. There have not been too many recent changes to BattlePanel.java for some time. This refactor did touch it:

```
commit 72fc1857159eb9c3d40ff26f4c86e6ea73ef0ced (tag: 1.10.15294)
Author: Dan Van Atta <dhvatta@gmail.com>
Date:   Fri Aug 16 14:50:18 2019 -0700

   Refactor: run refresh code on EDT (#4997)
```
https://github.com/triplea-game/triplea/pull/4997/files

Looking at that diff again, it does look clean. I suspect perhaps the race condition might not actually be from BattlePanel and itself. Each time hitting this bug there is another window that is trying to be displayed. Perhaps when BattlePanel is rendering, the `setVisible` call is blocking for a resource held by the other window.

***Important***: I did discover that the game is not completely frozen! A person can simply close the second popup window with the 'X' button in the upper right corner. It's obvious that can be done as the UI is for the most part locked up and does look completely frozen. If we continue to see this problem and fail to solve it, we could tell users of the next release to be aware of the situation and that is how they would fix it.

PR #5089, opened 2019-08-25
Name de-dupe by MAC instead of by IP

DanVanAtta:
## Overview
Previous feature to enforce one player name per IP is switched to be one player name per MAC. After this update, logging in with the same device multiple times would always force teh same name. The switch from IP to MAC is to avoid a situation where a 'LAN party' (minus host) are all stuck using the same name.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix

## Testing

[ ] Manually tested
[ ] No testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
+++ b/game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
@@ -1,13 +1,12 @@
 package games.strategy.net;
 
 import com.google.common.base.Preconditions;
+import com.google.common.collect.Multimap;
 import games.strategy.engine.lobby.PlayerNameValidation;
-import java.net.InetAddress;
 import java.util.Collection;
 import java.util.Comparator;
-import java.util.HashSet;
+import java.util.Map;
 import java.util.Optional;
-import java.util.stream.Collectors;
 import lombok.AccessLevel;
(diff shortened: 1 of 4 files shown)


game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
```diff
@@ -22,37 +21,42 @@
...
     Preconditions.checkArgument(PlayerNameValidation.serverSideValidate(desiredName) == null);
+    Preconditions.checkNotNull(loggedInMacsToNames);
 
-    String currentName = findLoggedInName(ipAddress, loggedInNodes).orElse(desiredName);
-    if (currentName.length() > 50) {
-      currentName = currentName.substring(0, 50);
+    String currentName = desiredName;
+
+    if (!isBotName(desiredName)) {
+      if (currentName.length() > 50) {
```
DanVanAtta: The name truncation now happens before `findExistingLoggedInNameFromSameMac`; Technically a bug fix, by doing the truncation after `findLoggedInName`, the name de-dupe in `findLoggedInName` would not have matched the truncated name (which previously occurred after `findLoggedInName`)

PR #5093, opened 2019-08-26
Add commentary to DB helper scripts

DanVanAtta:
## Overview
Documentation update to describe the scripts. 'launch_db' is updated
to be a bit more useful as it's rare to need to run 'build_docker'.
Instead 'launch_db' now captures the common commands run after any
system restart.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix
[X] Documentation Change

## Testing

[ ] Manually tested
[x] No testing done

### DIFF
--- a/lobby-db/launch_db
+++ b/lobby-db/launch_db
@@ -1,15 +1,14 @@
 #!/bin/bash
 
-# This script is mostly in part documentation and convenience. Here
-# we tie together the full flow of building a docker image, launching
-# it, and then executing flyway to have a DB schema installed.
-
+# Convenience script to get a local DB up and running on docker
+# with some sample data.
 
 set -eu
 
 path=$(dirname $0)
 
-$path/build_docker
 $path/run_docker
+sleep 5
 $path/run_flyway
+$path/load_sample_data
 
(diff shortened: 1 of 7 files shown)


lobby-db/launch_db
```diff
@@ -1,15 +1,14 @@
...
+# Convenience script to get a local DB up and running on docker
+# with some sample data.
 
 set -eu
 
 path=$(dirname $0)
 
-$path/build_docker
 $path/run_docker
+sleep 5
```
RoiEXLab: Ah the typical "db isn't ready yet" problem with docker ^^

RoiEXLab: Netcat should be installed on every linux system, so this could actually easily be done.
Or you might want to try out something like this: https://github.com/eficode/wait-for which is basically a script that can be used to wait for a service to start

RoiEXLab: Maybe I'll do when I have more time. Currently I can't afford doing things that will take an unknown amount of time, next month this should get a little bit better

PR #5094, opened 2019-08-26
Add email to forgot password and move forgot password logic to http-server

DanVanAtta:
## Overview

(1) For forgot password, request user enter their email in addition to their username
- This is so we can more truly avoid fake password requests. A users email is not known publicly while username is. By requesting both username and email, we are requesting a non-public piece of information that only the user should know.
- We need to also ask for username in additional to email as emails are not unique in TripleA but usernames are.

(2) Move forgot password logic to http-server
- This should simplify the overall interaction and removes logic from the complicated ServerQuarantineConversation.
- The forgot password logic was in the lobby code for a few reasons:
  - was not clear how or if emails could be sent from http server
  - was questionable how and if password hashing could be re-used.
- Email was previously not part of the lobby socket communication as adding that additional piece of information was not compatible with the previous flow. It was not readily apparent if it could have been cleanly added, so no email was at the time a limitation of re-using LobbyServerQuarantineConversation. With the logic moved to HTTP Server, it's very straight forward to add email to the request JSON being sent to the server.

(3) Minor: In non-prod we continue to *not* send a temp password email, but we will now log it.  Previously the email notificationit was a complete no-op and there was no way to know the temp password (in non-prod).

(4) Stop looking at XFORWARDED_HEADER when looking for IP address in http requests. It can be forged, if someone is using a Proxy we'll read that IP address and its okay.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[X] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix

## Testing

[X] Manually tested
[ ] No testing done

### Manual Testing
- launched lobby + http server  + lobby db with sample data
- clicked 'play online' > forgot password
- user = "test"; email = "email@email.com"
- verified saw success message, grabbed temp password from http-server log
- logged in with temp password, verified lobby showed change password message
- changed password, logged out
- logged back in with newly changed password

### Before
Screenshot from 2019-08-25 23-50-56

### After
Screenshot from 2019-08-25 22-30-58


RoiEXLab:
Before reviewing this:
Can we check how many duplicate emails we have in the DB?
If there are reasonably few, we could maybe tell them that we're going to change their email to something like `0+name@server.com`, the `0+` prefix should work most of the time without any problems and if we ever run into Problems we should be able to handle those manually

java-extras/src/main/java/org/triplea/java/Sha512Hasher.java
```diff
@@ -0,0 +1,48 @@
+package org.triplea.java;
```
RoiEXLab: I'd have gone with `org.triplea.crypto`, because this obviously is java code and I don't think it adds much value here.

lobby-db-dao/src/main/java/org/triplea/lobby/server/db/dao/TempPasswordDao.java
```diff
@@ -32,23 +33,27 @@
...
 
+  @SqlQuery("select id from lobby_user where username = :username and email = :email")
+  Optional<Integer> lookupUserIdByUsernameAndEmail(
+      @Bind("username") String username, @Bind("email") String email);
+
   @SqlUpdate(
       "update temp_password_request"
           + " set date_invalidated = now()"
-          + " where lobby_user_id = (select id from lobby_user where username = :username)"
+          + " where lobby_user_id = :userId"
```
RoiEXLab: 👍

RoiEXLab:
@DanVanAtta I wasn't actually considering doing anything within this pr regarding that topic.

Also 607 is a lot more than I hoped for. 
If we had like 50 that would've been a whole different story

java-extras/src/main/java/org/triplea/java/Sha512Hasher.java
RoiEXLab: `org.triplea.hash` seems fine.
To my knowledge the `org.triplea.java` package only exist for this class, that's why I was suggesting to adjust the name to be more fitting for the classes - or rather class - inside

http-server/src/test/java/org/triplea/server/error/reporting/ErrorReportControllerIntegrationTest.java
```diff
@@ -9,7 +9,9 @@
 import org.triplea.http.client.error.report.ErrorUploadRequest;
 import org.triplea.http.client.error.report.ErrorUploadResponse;
 import org.triplea.server.http.AbstractDropwizardTest;
+import org.triplea.test.common.Integration;
 
+@Integration
```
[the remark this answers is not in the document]
RoiEXLab: > I wonder a bit if we need the `@Integration` test flag anymore at all.

In general or just for all `AbstractDropwizardTest` classes?

The only reason we/I introduced it is so intergration tests wouldn't count towards coverage.

java-extras/src/main/java/org/triplea/java/Sha512Hasher.java
RoiEXLab: Ahhh ok that makes sense then.
I must have missed this folder

RoiEXLab:
> Agree 607 is a bit surprising. I suspect the large number is due to the missing 'forgot password' feature, so people recreated accounts.

Good Theory. We have a "last-login-time" store somewhere, don't we? We could analyse if this is actually the case by checking if only one of the accounts has a recent time

PR #5114, opened 2019-08-28
Update travis to delete existing tag if it exists

DanVanAtta:
## Overview
Potentially a build fix, it looks like introducing 'set -eu' is failing the build. Previously the build would blindly try to delete tags and we ignored any failures there. With 'set -eu', this potentially causes a failure.

The update here adds a 'git ls-remote --tags' to look for tags and then conditionally executes the delete tag if the target tag is found.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[X] Configuration Change
[X] Bug fix

## Testing

[ ] Manually tested
[ ] No testing done

### Manual Testing
Verified script still passes shellcheck 
Verified grep -q command executes the second part of the command on tag found case and not when tag is not found.

### DIFF
--- a/.travis/do_release
+++ b/.travis/do_release
@@ -102,14 +102,21 @@ function triggerGithubReleaseWithTagPush() {
   git config --global user.email "tripleabuilderbot@gmail.com"
   git config --global user.name "tripleabuilderbot"
 
-  echo "Attempting to delete tag $TAG_VALUE, this is so we will recreate the release if this build is being re-run. Ignore any errors from this"
-  git push -q "https://$GITHUB_PERSONAL_ACCESS_TOKEN_FOR_TRAVIS@github.com/triplea-game/triplea" --delete "refs/tags/$TAG_VALUE"
-  echo "Done deleting existing tag if it existed."
+  local -r gitRepo="https://$GITHUB_PERSONAL_ACCESS_TOKEN_FOR_TRAVIS@github.com/triplea-game/triplea"
 
+  # Sometimes assets will fail to upload but we'll still have a tag created
+  # Check if the current tag exists, if so, delete it. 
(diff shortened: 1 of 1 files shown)


.travis/do_release
```diff
@@ -102,14 +102,21 @@ function triggerGithubReleaseWithTagPush() {
...
   git config --global user.name "tripleabuilderbot"
 
-  echo "Attempting to delete tag $TAG_VALUE, this is so we will recreate the release if this build is being re-run. Ignore any errors from this"
-  git push -q "https://$GITHUB_PERSONAL_ACCESS_TOKEN_FOR_TRAVIS@github.com/triplea-game/triplea" --delete "refs/tags/$TAG_VALUE"
-  echo "Done deleting existing tag if it existed."
+  local -r gitRepo="https://$GITHUB_PERSONAL_ACCESS_TOKEN_FOR_TRAVIS@github.com/triplea-game/triplea"
 
+  # Sometimes a tag will be created from a previous build and assets will fail to upload.
+  # Check if the current tag exists, if so, delete it.
+  git ls-remote --tags origin | grep -q "2.0.15585$" && \
```
RoiEXLab: I don't think you meant to grep for `2.0.15585` here

DanVanAtta:
> However I'd like to point out that I was convinced there was a mechanism that already overwrote any existing tags, not sure why this change is required, because it shouldn't

Existing mechanism in our travis build? AFAIK that mechanism was a  "attempt it and let it fail". The conditional is perhaps a bit cleaner if only for the messaging.

DanVanAtta:
I was also thinking the tag push fail was causing us to have an exit code 1, it turns out it was not; but again, it's probably better to be explicit about this than not so we can avoid the explanatory echo statement documenting a potential no-op

PR #5116, opened 2019-08-28
Add debug statements to travis 'do_release'

DanVanAtta:
## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix

## Testing

[ ] Manually tested
[ ] No testing done

### DIFF
--- a/.travis/do_release
+++ b/.travis/do_release
@@ -129,7 +129,9 @@ function pushMaps() {
   ## do git stuff, check if there is a diff, if so, commit and push it
   (
     cd website
+    git status
     if ! git diff-index --quiet HEAD --; then
+      echo "Executing diff-index if"
       git add --all _maps/
       git commit -m "Bot: update map files after game engine build ${TRAVIS_BUILD_NUMBER:-}"
       git push -fq origin master


RoiEXLab:
IIRC @ssoloff once used some sort of debug mode to access the install4j key in travis, might be useful in the future to avoid having debug PRs like this

DanVanAtta:
It's crazy this fixed the build.
@RoiEXLab the install4j step is already completed, the piece that failed is a clone of website was checked and then YAML splitter was run to update maps. The if statement is supposed to detect if any files had changed, and if so the if block is executed. Based on the failure, the if statement was true and the if block was executed, but the commit inside of it had nothing to commit and then failed.

We have been seeing that error message previously on green builds, we did not though have the "set -e" flag to halt the build if we encounter errors.

RoiEXLab:
> It's crazy this fixed the build.

ikr 😅

DanVanAtta:
yah, sadly fixes like that means we actually don't know what fixed the build. There is a dependency on the website repo here, but AFAIK nothing changed there.. 

I'm thinking to spend some time iterating on this, I think the first step is to see how stable the build is, rebuilding it, and then reverting the fix and adding it back in again to see if it is really making a difference.

Issue #2502, opened 2017-10-14
PBEM: Warnings displayed in console when sending email


ssoloff:
#### Environment

TripleA 1.9.0.0.7131

#### Reproduction Steps

* Start a PBEM game.
* Select a provider (I used Gmail) and enter your credentials.
* Click **Test Email**.

#### Actual behavior

The console is displayed with the following warnings: 

```
Oct 14, 2017 4:43:38 PM javax.mail.Session loadResource
WARNING: expected resource not found: /META-INF/javamail.default.providers
Oct 14, 2017 4:43:38 PM javax.mail.Session loadResource
WARNING: expected resource not found: /META-INF/javamail.default.address.map
```

#### Expected behavior

No warnings should appear in the console.

#### Notes

The email is sent without any issue.  This is just an annoying UX issue because the console is displayed unexpectedly.

I only noticed this recently.  I suspect the root cause is the upgrade of the `com.sun.mail:mailapi` and `com.sun.mail:smtp` dependencies in #2486.

ssoloff:
Fixed in 1.9.0.0.7286 via #2541.

PR #5118, opened 2019-08-29
Use jakarta dependency

RoiEXLab:
## Overview
This PR changes the JavaMail dependency to use the maintained "jakarta" artifact, which is the version now maintained by the eclipse foundation.
This change should change absolutely nothing as stated on their website: https://eclipse-ee4j.github.io/mail/#Latest_News
There is a 1.6.4 version that contains some bugfixes, but I wasn't feeling confident enough to use that directly

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[x] Configuration Change
[ ] Bug fix

## Testing

[ ] Manually tested
[x] No testing done

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,4 +40,4 @@ this Program grant you additional permission to convey the resulting work.
 
 Library | Group ID | Artifact ID | SPDX License ID
 :-- | :-- | :-- | :--
-JavaMail | com.sun.mail | javax.mail | GPL-2.0-only
+Jakarta Mail | com.sun.mail | jakarta.mail | GPL-2.0-only
--- a/build.gradle
+++ b/build.gradle
@@ -71,6 +71,7 @@ subprojects {
         mockitoVersion = '2.24.5'
         postgresqlVersion = '42.2.5'
         sonatypeGoodiesPrefsVersion = '2.2.5'
+        jakartaMailVersion = '1.6.4'
     }
 
     repositories {
--- a/game-core/build.gradle
+++ b/game-core/build.gradle
@@ -12,7 +12,7 @@ dependencies {
     implementation project(':swing-lib')
     implementation 'org.snakeyaml:snakeyaml-engine:1.0'
     implementation 'com.googlecode.soundlibs:jlayer:1.0.1.4'
-    implementation 'com.sun.mail:javax.mail:1.6.2'
+    implementation 'com.sun.mail:jakarta.mail:1.6.3'
     implementation 'commons-cli:commons-cli:1.4'
     implementation 'commons-codec:commons-codec:1.11'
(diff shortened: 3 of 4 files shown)


DanVanAtta:
@RoiEXLab I suppose this should be tested to make sure we don't run into any unpleasant surprises in case our existing configuration or any updates for any reason do not work. It's a bit painful too that one of the remaining undocumented setups on the server is email. When I inherited system admin of the servers, none of the deployment or configuration was documented, AFAIK email configuration is the last holdout needing documentation.

DanVanAtta:
For testing, I created a sample app that sent myself an email, the email sender is a file that can be copy/pasted and then a main method added to send the mail. I put this in a new project, and ran a shadow jar to package it. From there I `scp`'d the file to the server. eg:
`scp ../path/to/shadow/jar admin@lobby.triplea-game.org:~`

Then it was a matter of executing the jar with `java -jar ...shadowJar`. It does sound like a PITA, it is, but FWIW I was surprised it did only take about 10 minutes.

RoiEXLab:
@DanVanAtta well it compiles and the website linked above states that there are no changes at all, except for the new maven location.
I double checked this is the correct dependency (thus should have all the runtime requirements), so there really should be no surprises.

I mean if you have a really bad feeling about this I can try to test this tomorrow, but I don't think there's a real problem here

DanVanAtta:
Let's test it, better sure than not. Should only need to be done once per dependency update. Ideally it's automated or we incorporate that stubbed email receiver you mentioned, we're not there yet though and once verified we'll then know anything that uses the email sending modules will be okay.

RoiEXLab:
Note to myself: when I'm going to test this anyways, I can
1. Update to 1.6.4
2. Use only the SMTP runtime dependency, we don't need imap and pop3

RoiEXLab:
Ok, ran into the same issue: https://github.com/triplea-game/triplea/issues/2502
Thanks @ssoloff for saving me from the headache 😅 

@DanVanAtta I currently only verified the dependency for game-core, it still works. If that makes you feel confident enough, feel free to merge, otherwise I'll try to check it on the lobby server this evening or something.

DanVanAtta:
@RoiEXLab you were able to send an email without any issue using the new dependency? It does not matter at all which project it was in, so long as our server had no issue sending an email.

RoiEXLab:
Yes, but I'm not sure if both projects actually call the same code

DanVanAtta:
Should be good enough, maybe worth checking @RoiEXLab

RoiEXLab:
You convinced me, they both have all the dependencies and they both can send emails

DanVanAtta:
Thank you for doing that verification!

PR #5122, opened 2019-08-29
Fix pre-release to release bad enum value problem for flag settings

DanVanAtta:
## Overview
Null out flag value when flag values are turned off.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[X] Bug fix

Reported in: https://forums.triplea-game.org/topic/1531/menu-absent-new-bug-in-old-version/5

Newest release sets an enum value that is not present in previous release. When the previous release reads that value, it errors out with 'enum value not found'. To fix this, update latest to null out the enum flag value (even though it is not necessary). This way when flags are turned off and a null value is present, the latest release will read a null value and would not error out.

## Testing

[ ] Manually tested
[X] No testing done


DanVanAtta:
Merging to get this pre-release built and distributed

PR #5129, opened 2019-08-30
Introduce UserJdbiDao Refactor

DanVanAtta:
## Overview

(1) Rename UserLookupDao -> UserJdbiDao
(2) Move #updateLastLogin to UserJdbiDao
(3) Update API of #login(String, HashedPassword) -> #login(String, String)
(4) Cleanup LobbyLoginValidator, remove unused TempPasswordDao property

Eventually the updates will support more refactors in LobbyLoginValidator which in turn supports changes in login functionality to add a 'user needs to change password functionality'

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[X] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix

## Testing

[ ] Manually tested
[ ] No testing done


lobby-db-dao/src/test/java/org/triplea/lobby/server/db/dao/UserJdbiDaoTest.java
```diff
@@ -0,0 +1,43 @@
...
+  @Test
+  void getPassword() {
+    assertThat(userDao.getPassword(USERNAME), isPresentAndIs(PASSWORD));
+    assertThat(userDao.getPassword("DNE"), isEmpty());
+  }
+
+  @Test
+  void updateLastLogin() {
+    assertThat(userDao.updateLastLoginTime(USERNAME), is(1));
+    assertThat(userDao.updateLastLoginTime("DNE"), is(0));
```
RoiEXLab: Wouldn't it make sense to check  if this actually updates the time here?
I know this isn't 100% trivial because we can't mock the DB now() function, but at least checking if the new date is more recent would be a win IMO

DanVanAtta: Indeed this would be normally verified with an expected dataset. You're right about the 'now' function. One problem is there are no queries against that column, reading it would require a test-only DB command. It's possible, but we'd add complication in the test and production coupling to the test-code.

The other alternative would be to parameterize the date function that is passed in so we do not use the DB 'now()'. It's really ideal to use 'now()' as then the DB can take care of the date, no time zone or formatting issue, and the java method has fewer parameters.

I don't know if either of those two choices are really worth it for what is otherwise a pretty simple DB update. To some extent knowing we can execute the query against a live-DB is one of the biggest wins we get and would catch most problems we could have. For example if the DB column name had a typo, or the table, some of the more likely problems with a simple query where there is a refactor of table or columns names, we would catch those.

Of some note, the return value of the method was updated to be an 'int' and not a 'void' method call in order to enable this level testing so we are not *just* invoking the DB query.

RoiEXLab:
Ah, sorry forgot we're still mid-debug

DanVanAtta:
Truly bad practice to debug like this in prod. One really should instead invest the time to reproduce and iterate quickly locally. I was a bit silly in trying to debug it as we only needed to see the output from git diff-index to see what it was telling us. Regardless, there is a PR building now to remove all the debug stuff and would open up normal merges again.

PR #5146, opened 2019-08-31
Faster Builds with Parallel travis VMs 

DanVanAtta:
## Overview
Break up build tasks across a couple more Travis VMs by using build matrix. Should help parallelize build tasks and allow for a faster overall runtime of the build.

Release task is broken out and pre-emptively downloads images while we expect other tasks to still be running. This way if we get to the deploy step (merge builds), then the images will already be downloaded and we can proceed with artifact creation.

Five parallel jobs is chosen specifically as that seems to be the max number of VMs that Travis will spin up for a project fork. Beyond that and we have jobs waiting for previous jobs to terminate to free up VMs. Hence, five seems to be the maximum parallelization available from travis.

Tasks load balanced in an attempt that each of the 5 VMs will take the same amount of time.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[x] Configuration Change
[ ] Bug fix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,17 +9,17 @@ matrix:
       - TRIPLEA_RELEASE=true
   - jdk: openjdk11
     env: 
-      - TASKS="test jacocoTestReport integTest"
-      - INSTALL=true
+      - TASKS="test jacocoTestReport"
       - PUSH_REPORTS=true
       - TRIPLEA_RELEASE=false
   - jdk: openjdk11
     env: 
-      - TASKS="validateYamls spotlessCheck checkstyleMain checkstyleTest"
+      - TASKS="validateYamls spotlessCheck checkstyleMain checkstyleTest pmdMain"
       - TRIPLEA_RELEASE=false
   - jdk: openjdk11
     env: 
-      - TASKS=pmdMain
+      - TASKS=integTest
+      - INSTALL=true
       - TRIPLEA_RELEASE=false
   - jdk: openjdk12
     env: 
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,27 @@ language: java
 matrix:
   include:
   - jdk: openjdk11
-    env: TRIPLEA_RELEASE=true
+    env:
+      - TASKS="shadowJar downloadAssets"
+      - TRIPLEA_RELEASE=true
+  - jdk: openjdk11
+    env: 
+      - TASKS="test jacocoTestReport integTest"
+      - INSTALL=true
+      - PUSH_REPORTS=true
+      - TRIPLEA_RELEASE=false
+  - jdk: openjdk11
+    env: 
+      - TASKS="validateYamls spotlessCheck checkstyleMain checkstyleTest"
+      - TRIPLEA_RELEASE=false
+  - jdk: openjdk11
+    env: 
+      - TASKS=pmdMain
+      - TRIPLEA_RELEASE=false
   - jdk: openjdk12
-    env: TRIPLEA_RELEASE=false
+    env: 
+      - TASKS=test
+      - TRIPLEA_RELEASE=false
 addons:
   postgresql: "10"
   apt:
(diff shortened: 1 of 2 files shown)


DanVanAtta:
Last build was kinda slow, regardless slowest build was done in 7m30s. This is compared to the 15m it used take for branch builds. Given integTest is on its own, there is no improving that build time further unless we can speed up or break up the integTests.

Screenshot from 2019-08-30 20-22-41

RoiEXLab:
Hmm I have mixed feelings on this one.
1. I think this is a nice improvement to build times
2. On the other hand, this introduces a lot of complexity into the build process and executes a lot of tasks (like compiling) more than once.

Also a minor concern: will this affect throughput if suddenly the a single pr needs more than double the vms? Will other PRs have to wait longer?

DanVanAtta:
I remember when the build was under 3 minutes : )

Build times are actually pretty important for efficiency. I'm not sure how many times you've had the experience to have a build fail multiples times, 5, 10 minutes into the build process. Part of the problem I think is that build times creep up so gradually that it always feels acceptable and just as how it is.  For example, hey, the build time is only 20 seconds slower, merge, and repeat. 

Going from 15 minutes to 7 is a pretty big jump. I think it'd be ideal if we could further improve that to be under 5 minutes and get back into the 3-4 minute range. The full release build taking 25 minutes now is getting up there, it was painful building patches the other day and took most of a few hours just to make small tweaks. We can ill afford such inefficiency.

Regardless, to address your points:
- To some extent we were compiling redundantly, grant it was with JDK11 and JDK12, arguably different, but there was overlap. It's the same situation if a compile fails there multiple times vs for multiple task VMs. I'm not sure what the real problem is, if all the builds fail 2 minutes in, then you check one and see there is a compile error. Otherwise we can potentially gain a lot more information to see for example that maybe both a test and checkstyle fails. Before you only got the one piece of info and then later got the other (5 minutes vs 5~10 minutes and 10 minutes, so 5 minutes vs waiting 15~20 minutes).


- For complexity, the configuration is a bit more complex, I think the task flag makes it pretty clear and you get to see them listed in the build result window so it's clear what is executing what. The VMs are also independent, it's pretty straigtht forward IMO as it's just multiple builds that are not interacting with each other. We had some complexity already to run gradle as part of the install task, where that does not belong simply to get faster feedback. I don't think it can always be quite expected for a build script to be beautiful, it does have the requirement to give feedback quickly. Regardless, the running tasks in install 'mess' is cleaned up and we have some flags to skip steps that are not going to be needed so we can get faster feedback. There is more complexity in the build process, but it's there regardless of parallel runs or not.

Is there anything in specific that seems potentially confusing?



> Also a minor concern: will this affect throughput if suddenly the a single pr needs more than double the vms? Will other PRs have to wait longer?

Maybe. We do see the JDK12 build often starting slower and some builds wait for that to complete, it looks like its due to VM allocation. Hard to say if it'll change in the future. Assuming the 5 VM limit that I observed holds, let's look at build times:

1 build -> 7 minutes
2 builds -> 14 minutes, in sequence
3 builds -> 21 minutes

For the previous process:
1 build -> 15 minutes
2 builds -> 15 minutes
3 builds -> 30 minutes

Looks like we get better throughput overall. Since we're spreading more tasks to more VMs, we're applying more horsepower to the overall job.

It's also worth a mention as well that speeding up builds is a best practice, as can be implied from the important of fast turnarounds. Travis itself recommends speeding up builds in this manner: https://docs.travis-ci.com/user/speeding-up-the-build/

RoiEXLab:
Thanks for the clarification.
The travis confog becoming more complex is a tradeoff, but I think you convinced me that this tradeoff is worth it.

Side-Note: When github actions finally launches, wr might want to give the ci feature a shot. Maybe it's faster

DanVanAtta:
> Side-Note: When github actions finally launches, wr might want to give the ci feature a shot. Maybe it's faster

It's an exciting feature to see in the works. I'm pretty sure there is some aspect of it that will be useful to us!

RoiEXLab:
Well the first thing I'll try to do with it is setting everything up so that the website repo is updated with a current version string whenever we release a stable, instead of querying the github api each time with js

RoiEXLab:
Errm something went wrong 
https://github.com/triplea-game/triplea/releases/tag/untagged-581c729baae7e824a823

DanVanAtta:
We already update the list of maps, it's feasible. Though, checking/unchecking the prerelease box is a pretty nice, single location for releasing. Given we roll back sometimes, it would be difficult to commit the value statically.

DanVanAtta:
I'm looking into it @RoiEXLab

DanVanAtta:
Speaking of build problems, I'll trade you for this one @RoiEXLab : https://travis-ci.org/triplea-game/triplea-game.github.io

RoiEXLab:
Dang it, I knew I specified that file somewhere, I just couldn't find it.
Thanks for notifying me

PR #5148, opened 2019-08-31
JMenuItem Checkbox Persistence

DanVanAtta:
## Overview
- Remove setting "show tabbed production"
- Refactor multiple JMenuItem checkboxes to use swing build API.
- Set the following checkboxes up to remember their setting:
  - Show comment log
  - Show chat times
- Simplify boolean client setting, default to false

To support the persistence we add a new interface to ClientSetting boolean values that can be injected into a JMenuItemCheckBoxMenuItemBuilder

## Functional Changes

[ ] New map or map update
[ ] New Feature
[x] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix

## Testing

[x] Manually tested
[ ] No testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/settings/ClientSetting.java
+++ b/game-core/src/main/java/games/strategy/triplea/settings/ClientSetting.java
@@ -41,6 +41,7 @@ import lombok.extern.java.Log;
  *
  * @param <T> The type of the setting value.
  */
+@SuppressWarnings("StaticInitializerReferencesSubClass")
 @Log
 public abstract class ClientSetting<T> implements GameSetting<T> {
   public static final ClientSetting<Integer> aiPauseDuration =
@@ -52,9 +53,9 @@ public abstract class ClientSetting<T> implements GameSetting<T> {
   public static final ClientSetting<Integer> battleCalcSimulationCountLowLuck =
       new IntegerClientSetting("BATTLE_CALC_SIMULATION_COUNT_LOW_LUCK", 500);
   public static final ClientSetting<Boolean> confirmDefensiveRolls =
-      new BooleanClientSetting("CONFIRM_DEFENSIVE_ROLLS", false);
+      new BooleanClientSetting("CONFIRM_DEFENSIVE_ROLLS");
   public static final ClientSetting<Boolean> confirmEnemyCasualties =
-      new BooleanClientSetting("CONFIRM_ENEMY_CASUALTIES", false);
+      new BooleanClientSetting("CONFIRM_ENEMY_CASUALTIES");
   public static final ClientSetting<String> defaultGameName =
       new StringClientSetting("DEFAULT_GAME_NAME_PREF", "Big World : 1942");
   public static final ClientSetting<String> defaultGameUri =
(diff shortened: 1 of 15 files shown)


game-core/src/main/java/games/strategy/triplea/settings/ClientSetting.java
```diff
@@ -41,6 +41,7 @@
  *
  * @param <T> The type of the setting value.
  */
+@SuppressWarnings("StaticInitializerReferencesSubClass")
```
RoiEXLab: I saw this warning too, is it safe to ignore?

DanVanAtta:
@RoiEXLab FWIW there are a good number of client settings mutated outside of engine preferences, eg:
```
public static final ClientSetting<Boolean> notifyAllUnitsMoved =
public static final ClientSetting<String> defaultGameUri =
public static final ClientSetting<String> moderatorApiKey =
public static final ClientSetting<Boolean> firstTimeThisVersion =
public static final ClientSetting<String> lastCheckForEngineUpdate =
public static final ClientSetting<String> lastCheckForMapUpdates =
public static final ClientSetting<String> emailServerHost =
```

I can speak to the original intent behind ClientSetting as I introduced the class originally. Really it was just a facade, a helper API in front of the system preferences code. It was to hide the persistence details and simplify the interaction when getting properties. In essence, such a code clean up that we see here was the original intent. Removing direct interaction from the properties as well gives us options to change how we persist those settings.

A next iteration saw all the various settings we had everywhere and desired to make a single settings window to aggregate everything. It was also an issue that the setings window from main menu had its own unique settings that could not be altered from within the game, and the game settings was its own window. Unifying those settings windows to a single window was a goal.

It was though not a goal to make the settings window the one place to toggle settings. For example we could place a UI control to update a setting and still also have the settings window setting to modify it. This is just to make it easier to find. There is a UX principle of 'locality' that we should try to incorporate more, meaning that settings that control features should be in/near those features. For example I've been wanting to put a "gear icon" somewhere in the battle window to open a drop-down list of the battle relevant settings that can be changed. This way when in a battle, you can see/update those settings directly from that window.

Longer term, every persisted setting being in a single class does not scale super well and violates the open-closed principle. I think we have a good bit of runway (time) before it becomes a more significant problem.

RoiEXLab:
@DanVanAtta I don't really have any objections to use the same code, I just think it might make it easier to understand if this long list of ClientSettings (that also cause a warning) would be grouped into logically connected chunks

game-core/src/main/java/games/strategy/triplea/settings/ClientSetting.java
DanVanAtta: The class is plastered with these warnings : (
We've had this configuration and warnings for a long time. If it's a problem, it'll be a problem whether we ignore the warnings or not. There is a practicality for ignoring the warnings so we don't mask other warnings.

The text from the analysis describes this problem as a potential threading problem:

> Reports classes that refer to their own subclasses in their static initializers or in static fields. Such references can cause JVM-level deadlocks in multithreaded environment, when one thread tries to load superclass and another thread tries to load subclass at the same time.

DanVanAtta:
> Also not necessarily part of this PR, but in general because this list only gets longer and longer

Indeed. If put into groups, I wonder if we might have the same problem as java package grouping, any grouping you decide is going to be 'wrong' and arbitrary. (Side-note, the rule of thumb I've come to see is that package grouping should be based on visibility/scoping and not much else).

I think alpha-sorting the settings would help the management, some of the settings are not terribly well named. I'm not opposed to a nicer grouping, though it's difficult to see what would work really well. Notably there are a number of settings where only one or two of them really fit together. I wonder if we can convey that grouping more with common name prefixes so they stay together when sorted.

I've had a thought in the back of mind whether client settings all really need to be static variables and if we should make them local variables to be instantiated and used by the classes that need them. It's a radical change though, and many of these settings do have "action-at-a-distance", are set in one place and have an effect somewhere very different. At least having a unified way to do this helps avoids things like static variables in the sound checkbox class. The consistency of static state access through ClientSetting is not a horrible way to manage that problem (essentially lump the problem areas together and put a box around it, then there is a known pattern and its less of an issue).

PR #5152, opened 2019-08-31
Simplify flag view menu options and refactor code

DanVanAtta:
## Overview
- Unify the flag toggling code between the hot key action listener and view menu
- Simplify the view menu options and instead of a 'on/off' switch with a radio button, instead replace that with three options.
- Convert the radio buttons to simple menu items. This is actually a hack so that the radio button does not need to be kept in sync with updates fired from hotkey. It would be more ideal to have radio buttons, but then we would need to have a state change listener to update the radio selection when the hotkey for flags is pressed. To avoid this, we can convert to simple menu items. A user should be able to able to see pretty clear which selection is active, losing the radio button is not a terrible loss.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[X] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix

## Testing

[X] Manually tested
[ ] No testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/FlagDrawMode.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/FlagDrawMode.java
@@ -1,7 +1,7 @@
 package games.strategy.triplea.ui;
 
+import games.strategy.triplea.settings.ClientSetting;
 import games.strategy.triplea.ui.screen.UnitsDrawer;
-import java.util.prefs.Preferences;
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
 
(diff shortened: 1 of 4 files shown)


Cernelius:
You can remove the medium flag option, I think, leaving only on/off.

game-core/src/main/java/games/strategy/triplea/ui/FlagDrawMode.java
```diff
@@ -12,51 +12,28 @@
...
     } else {
-      UnitsDrawer.enabledFlags = false;
-      prefs.putBoolean(UnitsDrawer.PreferenceKeys.DRAWING_ENABLED.name(), false);
-      prefs.put(
-          UnitsDrawer.PreferenceKeys.DRAW_MODE.name(),
-          UnitsDrawer.UnitFlagDrawMode.LARGE_FLAG.name());
-      UnitsDrawer.setUnitFlagDrawMode(UnitsDrawer.UnitFlagDrawMode.SMALL_FLAG, prefs);
+      toggleDrawMode(UnitsDrawer.UnitFlagDrawMode.NONE, mapPanel);
     }
-    mapPanel.resetMap();
```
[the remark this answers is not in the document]
DanVanAtta: Good consideration, I was wondering about maybe doing something that would cycle through the draw modes. I'm of two minds here:
(1) when the code is external to the enum, it might be better to be explicit rather than use a more complex algorithm. One should always be careful of "clever" and consider total time to understand. Explicit and simple is worth a lot.

(2)  We do have a feature envy code smell where we're interacting with the enum in a way that could be self-encapsulated. If the logic for next draw mode were encapsulated, I don't think having such code like this in an enum is that too unexpected or difficult. It also has the nice benefit to localize any changes to the enum, so we can add/remove enum values and we don't have to update external classes that use the enum.

Update is made in: 2e5cea29a; this block now looks like this:
```
  /** Toggles to the next flag draw mode. */
  static void toggleNextDrawMode(final MapPanel mapPanel) {
    final var currentDrawMode = ClientSetting.unitFlagDrawMode.getValueOrThrow();
    toggleDrawMode(currentDrawMode.nextDrawMode(), mapPanel);
  }
```

DanVanAtta:
@Cernelius I do not necessarily disagree, I think you could be right. Removing features is not to be taken lightly though. I don't think we have the evidence that the small flag draw mode is completely unused. In part, it's difficult evidence to acquire. At some point I suspect TripleA will start to write a 'feature usage log' that can be uploaded. It'd be a very useful thing to see if some features are completely unused and if others are used more often then we expect.

game-core/src/main/java/games/strategy/triplea/ui/screen/UnitsDrawer.java
```diff
@@ -58,8 +55,14 @@
...
   /** Identifies the location where a nation flag is drawn relative to a unit. */
   public enum UnitFlagDrawMode {
-    LARGE_FLAG,
-    SMALL_FLAG
+    NONE,
+    SMALL_FLAG,
+    LARGE_FLAG;
+
+    public UnitFlagDrawMode nextDrawMode() {
+      final var values = UnitsDrawer.UnitFlagDrawMode.values();
```
RoiEXLab: Is it really necessary to put the full enum name here?
If yes I'd actually be surprised because I'd consider values() a static method of every enum

DanVanAtta: No, copy/paste artifact. Any static method call can be fully qualified, AFAIK as an unwritten convent, we conventionally do not do this. Updated: adf776ea8

PR #5157, opened 2019-08-31

DanVanAtta:
## Overview

(2) Add 'on tag is missing' to skip tag builds. This shoudl avoid dupliate builds after merges to master as the tag that is pushed to trigger deployments triggers its own (redundant) build.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[X] Configuration Change
[ ] Bug fix

## Testing

[X] Manually tested
[ ] No testing done

Could see the configuration working for travis build against my fork.

## Screens Shots

Typical build will look like this:
Screenshot from 2019-08-31 14-10-45

On an early iteration I did not have the flag to turn off the deployment step, post-merge the build should look like this (and ideally the deploy will succeed!):
Screenshot from 2019-08-31 14-10-25

## Additional Review Notes
Travis install script did not really make a lot of sense as a name as there is no longer an install step. The script was renamed and broken out into two purpose specific scripts.


DanVanAtta:
Hmm, the deploy step is showing up even though on forked builds it did not. It's not necessarily a problem but would slow things down. The deploy VM fires up and if everything is disabled then it would slow things down while booting. I'm not sure why a fork build would see a difference vs a PR build.

DanVanAtta:
Okay, turns out the 'branch' name is 'master'. I'm not sure how the previous configuration then was successfully excluding PR builds from the deploy step. The deploy stage is now gone though for this PR build.

PR #5158, opened 2019-08-31

DanVanAtta:

(2) Add 'on tag is missing' to skip tag builds. This should avoid dupliate builds after merges to master as the tag that is pushed to trigger deployments triggers its own (redundant) build.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix

## Testing

[ ] Manually tested
[ ] No testing done


DanVanAtta:
Having trouble with travis auto-building the PR now. It seems not like it when PR builds are canceled.

PR #5159, opened 2019-08-31

DanVanAtta:
## Overview

(2) Add 'on tag is missing' to skip tag builds. This should avoid dupliate builds after merges to master as the tag that is pushed to trigger deployments triggers its own (redundant) build.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[x] Configuration Change
[ ] Bug fix

## Testing

[x] Manually tested
[ ] No testing done


.travis.yml
```diff
@@ -1,52 +1,58 @@
...
+      addons: {postgresql: "10"}
+      script:
+        - ./.travis/setup_database &
+        - ./.travis/launch_http_server
+        - ./gradlew integTest
+    - stage: verify
+      env: DESCRIPTION="JDK12 unitTests"
+      jdk: openjdk12
+      script:
+        - ./gradlew --parallel test
```
RoiEXLab: I mean obviously we don't need to run checkstyle here again, but I think integration tests, might be an equally good indicator that something breaks in a future version

DanVanAtta: IIRC error prone gave us the most trouble on upgrade, so we're covered there running that.

integTests I'm really on the fence. If we can make it a bit faster than the JDK11 integ test by cutting a step, that would be excellent. Integ tests being the slowest build is what we wait for, if the JDK12 VM keeps taking a bit longer to spin up, it could push out the total build time, hence I'm on the fence about it.

DanVanAtta: I was thinking to go ahead and see how this works and we can update it. Early results are bad.

1) integ tests are not reliable in parallel right now: https://github.com/triplea-game/triplea/issues/5163

If we run both tests and integTests, then the tests will be run in non-parallel mode and would be pretty slow. We could execute the test and 'integTest' as two different commands, that would make things a bit slower again.

2) Results for runtime are not good. integTest is a slow task, for JDK11 builds the test and integTest were broken up to help speed up the overall build time. Having those two tasks together adds to our slowest task.

In terms of performance results, this build gave it a try: https://travis-ci.org/DanVanAtta/triplea/builds/579469507; The JDK12 test + integTest task took 9m43s. The JDK11 integTest task took 8m5s. That represents a 20% increase in test speed, we'd expect perhaps even slower if we turned off the parallel flag.

PR #5160, opened 2019-08-31
Fix lock map not working bug

DanVanAtta:
## Overview
'getDefaultValue()' for a boolean condition will typically return false. 'getDefaultValue().orElse(false)' is always false. This update fixes this and uses the intended 'getValue()' option.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[X] Bug fix

[X ] Manually tested
[ ] No testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/settings/BooleanClientSetting.java
+++ b/game-core/src/main/java/games/strategy/triplea/settings/BooleanClientSetting.java
@@ -30,6 +30,6 @@ public final class BooleanClientSetting extends ClientSetting<Boolean>
 
   @Override
   public boolean getSetting() {
-    return getDefaultValue().orElse(false);
+    return getValue().or(this::getDefaultValue).orElse(false);
   }
 }


game-core/src/main/java/games/strategy/triplea/settings/BooleanClientSetting.java
```diff
@@ -30,6 +30,6 @@ public void saveSetting(final boolean value) {
 
   @Override
   public boolean getSetting() {
-    return getDefaultValue().orElse(false);
+    return getValue().or(this::getDefaultValue).orElse(false);
```
ssoloff: Just noting that `ClientSetting#getValue()` already delegates to `getDefaultValue()` if required, so the `.or()` call here should be unnecessary.  Also, since this implementation always requires a default value to be specified (i.e. `defaultValue` is never `null`), the `.orElse(false)` will never be called.

In light of these two considerations, the implementation of this method should degenerate to just `getValueOrThrow()`.

PR #5166, opened 2019-09-01
Add tests for comparator in DiceRoll.java

RoiEXLab:
## Overview
This is one of the methods that causes flaky coverage, the comparator isn't always invoked in a deterministic way, so I wrote tests that would cover all if cases all of the time.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix

## Testing

[ ] Manually tested
[ ] No testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
@@ -37,6 +37,8 @@ import org.triplea.java.collections.IntegerMap;
 import org.triplea.util.Triple;
 import org.triplea.util.Tuple;
 
+import com.google.common.annotations.VisibleForTesting;
+
 /**
  * Used to store information about a dice roll.
  *
@@ -424,7 +426,8 @@ public class DiceRoll implements Externalizable {
     sortAaHighToLow(units, data, defending, new HashMap<>());
   }
 
-  private static void sortAaHighToLow(
+  @VisibleForTesting
+  static void sortAaHighToLow(
       final List<Unit> units,
       final GameData data,
       final boolean defending,
--- a/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
@@ -1,5 +1,6 @@
 package games.strategy.triplea.delegate;
 
+import com.google.common.annotations.VisibleForTesting;
 import games.strategy.engine.data.GameData;
 import games.strategy.engine.data.PlayerId;
 import games.strategy.engine.data.Territory;
@@ -36,8 +37,6 @@ import org.triplea.java.collections.CollectionUtils;
 import org.triplea.java.collections.IntegerMap;
 import org.triplea.util.Triple;
 import org.triplea.util.Tuple;
-
-import com.google.common.annotations.VisibleForTesting;
 
 /**
  * Used to store information about a dice roll.
--- a/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
@@ -431,32 +431,16 @@ public class DiceRoll implements Externalizable {
       final GameData data,
       final boolean defending,
       final Map<Unit, Tuple<Integer, Integer>> unitPowerAndRollsMap) {
-    final Comparator<Unit> comparator =
-        (u1, u2) -> {
-          final Tuple<Integer, Integer> tuple1 =
-              getMaxAaAttackAndDiceSides(
-                  Collections.singleton(u1), data, defending, unitPowerAndRollsMap);
-          final Tuple<Integer, Integer> tuple2 =
-              getMaxAaAttackAndDiceSides(
-                  Collections.singleton(u2), data, defending, unitPowerAndRollsMap);
-          if (tuple1.getFirst() == 0) {
-            if (tuple2.getFirst() == 0) {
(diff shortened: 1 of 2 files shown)


game-core/src/test/java/games/strategy/triplea/delegate/DiceRollTest.java
```diff
@@ -741,4 +750,61 @@ void testSbrRolls() {
...
+    @Mock private Unit unit4;
+    @Mock private Unit unit5;
+
+    private final Map<Unit, Tuple<Integer, Integer>> unitPowerAndRollsMap = new HashMap<>();
+    private final List<Unit> units = new ArrayList<>();
+
+    @BeforeEach
+    void setUp() {
+      units.addAll(List.of(unit1, unit2, unit3, unit4, unit5));
+      for (final var unit : units) {
```
DanVanAtta: I think it would help maintainers skip a few steps if this for statement were extracted to a method or a comment is added. Can something be done to make it a bit more direct/obvious what is happening?

game-core/src/test/java/games/strategy/triplea/delegate/DiceRollTest.java
```diff
@@ -741,4 +750,61 @@ void testSbrRolls() {
...
+        when(unit.getType()).thenReturn(unitTypeMock);
+      }
+    }
+
+    @Test
+    void testAttacking() {
+      int index = 4;
+      for (final var unit : units) {
+        when(((UnitAttachment) unit.getType().getAttachment("")).getOffensiveAttackAa(any()))
+            .thenReturn(index / 2);
```
DanVanAtta: What does 'index / 2' represent? Would that question be answered by extracting the value to a variable?

game-core/src/test/java/games/strategy/triplea/delegate/DiceRollTest.java
```diff
@@ -741,4 +750,61 @@ void testSbrRolls() {
...
+      DiceRoll.sortAaHighToLow(units, gameData, false, unitPowerAndRollsMap);
+      assertThat(units.get(0), is(unit1));
+      assertThat(units.get(1), is(unit2));
+      assertThat(units.get(2), is(unit3));
+      assertThat(units.get(3), is(unit4));
+      assertThat(units.get(4), is(unit5));
+    }
+
+    @Test
+    void testDefending() {
```
DanVanAtta: A javadoc comment on how this method is setup to test I think would go a long way. Looking at the loop and comparing to the last test case, it kinda seems like we would get a reverse sort order. Why do we expect the order that is being asserted?

RoiEXLab: I'm pretty sure the sort algorithm used by java is stable, so because multiple values share the same value, they'll stay in their initial order

game-core/src/test/java/games/strategy/triplea/delegate/DiceRollTest.java
RoiEXLab: Not really, in order to test the `getFirst() == 0` cases, I had to introduce duplicate keys somehow to get this edge case covered.

A comment would help though IMO

game-core/src/test/java/games/strategy/triplea/delegate/DiceRollTest.java
RoiEXLab: Initially I had all the setup in this method, but because only parts of it were used in each test, this was against the strict stubbing policy of mockito, that's why I moved the stuff in such a weird way, I think I can fix that

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
```diff
@@ -424,37 +425,19 @@ private static void sortAaHighToLow(
     sortAaHighToLow(units, data, defending, new HashMap<>());
   }
 
-  private static void sortAaHighToLow(
+  @VisibleForTesting
```
DanVanAtta: Instead of making this visible for testing, any thoughts to extract this method to a package-protected utility class, something like "DiceRollSorter"?

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
```diff
@@ -424,37 +425,19 @@ private static void sortAaHighToLow(
...
-          }
-          return 0;
-        };
-    units.sort(comparator);
+    units.sort(
+        Comparator.comparing(
+            unit ->
+                getMaxAaAttackAndDiceSides(
+                    Collections.singleton(unit), data, defending, unitPowerAndRollsMap),
+            Comparator.<Tuple<Integer, Integer>, Boolean>comparing(tuple -> tuple.getFirst() == 0)
```
DanVanAtta: A comment to describe the compare algo could be nice to have as well. Maybe extracting to a method would work too so the method name can be used to describe the operation that is occurring.

RoiEXLab: I mean if I understood the algorithm that would be a feasible thing to do, but I don't unfortunately.

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
RoiEXLab: Not feasible for this PR.
The method makes use of `getMaxAaAttackAndDiceSides`, which could be used in a static way, but at this point we'd create a utility class that uses the class it is called from

PR #5167, opened 2019-09-01
Restore Travis Deploy Step & Skip Install Step

DanVanAtta:
## Overview

(1) 'env' variables are not available to stage 'if' conditions. To make sense of that, an if statement for a stage is evaluated by the travis coordinator process whether to spin up a VM to launch the travis script. An env variable, is a variable available on the VM. The travis deploy step was not running because 'env(TRAVIS_PULL_REQUEST)' was evaluating a non-existent variable, presumably to an empty string. This did not match the value 'false', and so the 'if' condition was never true. When the VM were to fire up though, then the value TRAVIS_PULL_REQUEST would have been populated and made available to the VM, but not before.

(2) I noticed the 'install' step was executing and was running a default 'gradle assemble' command. Overall this did not slow things down too much, but it did create a 'jar' file that is not necessary. To turn off the install step, we pass a magic value 'skip' to it.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[x] Configuration Change
[x] Bug fix

## Testing

[x] Manually tested
[ ] No testing done

Verified on fork that the expected deploy step showed up, required a slightly hacked script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ jobs:
     - stage: verify
       env: DESCRIPTION="static verification"
       jdk: openjdk11
+      install: skip
       script:
         ## shellcheck all shell scripts
         - find .travis/ -type f | xargs grep -lE "^#\!/bin/bash$" | xargs shellcheck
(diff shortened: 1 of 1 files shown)


DanVanAtta:
Going to go ahead and merge so we can get the build fix in.

PR #5169, opened 2019-09-01
Add tests for UnitTypeComparator.java

RoiEXLab:
## Overview
Part two of #5166 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature update or enhancement
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix

## Testing

[ ] Covered by existing automated tests
[x] Covered by newly added automated tests
[ ] Manually tested
[ ] No testing done

PR #5170, opened 2019-09-01
Add characterisation tests for DummyPlayer#selectCasualties

RoiEXLab:
## Overview
Final part 3 of #5169 and #5166 
This should hopefully fix flaky coverage on PRs, future will tell.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix

## Testing

[ ] Manually tested
[ ] No testing done

### DIFF
--- a/game-core/src/test/java/games/strategy/triplea/odds/calculator/DummyPlayerTest.java
+++ b/game-core/src/test/java/games/strategy/triplea/odds/calculator/DummyPlayerTest.java
@@ -17,6 +17,7 @@ import games.strategy.triplea.delegate.data.CasualtyList;
 import java.util.Collections;
 import java.util.List;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 
 class DummyPlayerTest {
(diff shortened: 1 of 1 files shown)


game-core/src/test/java/games/strategy/triplea/odds/calculator/DummyPlayerTest.java
```diff
@@ -0,0 +1,189 @@
...
+            null,
+            null,
+            null,
+            false,
+            null,
+            defaultCasualties,
+            null,
+            null,
+            false);
+    assertThat(details.getDamaged(), is(damaged));
```
DanVanAtta: `damaged` is an expected list? If so, renaming the variable to `expectedDamage` could help make that more obvious (obvious is good, it means faster comprehension).

Assertion comments I think would go a long way in helping to keep these tests maintainable to describe why we expect the assertions to hold. Will that reasoning remain obvious to you after a few months/years?

game-core/src/test/java/games/strategy/triplea/odds/calculator/DummyPlayerTest.java
```diff
@@ -0,0 +1,189 @@
...
```
DanVanAtta: I'm seeing a consistent three assertions, and only one looks to be changing. I wonder if the tests could be more focused. For one aspect, given the first and third assertions are always the same, it could be that the method under test happens to produce the right results for this case but not in all cases. I'm pretty convinced the 'getKilled' is relatively well exercised.

I wonder then if we can make the tests 'more' focused on specific behaviors by adding some nesting. The 3 existing tests would be a single nested class for the `details.getKilled` check. We'd then add two additional nesting to test specifically the `getAutoCalculated` and `getDamaged`. It's also a question whether the full setup is needed to test those two methods.
What are your thoughts @RoiEXLab ?

game-core/src/test/java/games/strategy/triplea/odds/calculator/DummyPlayerTest.java
RoiEXLab: Well this list should contain damaged units?
I'd assume that the headed player implementation uses this list in order to indicate which 2-hit battleships or something already took a hit, or should take a hit as part of the result of the method.
Just guesses though

game-core/src/test/java/games/strategy/triplea/odds/calculator/DummyPlayerTest.java
RoiEXLab: I tried to reduce the boilerplate a little bit. Tell me what you think

DanVanAtta: Looking good. Boiler plate in test is not necessarily bad; though there is a trade-off maintenance cost of updating many things when things change. So a bit of a balance I suppose.

To be direct, what's the most painful in tests is when they fail and it's not clear what they were testing or should have passed to begin with. I'm a big fan of assertion messages describing conditions, things like `assertThat("this should pass because..", a, is(b));`

I also really like javadocs on test methods that describe the strategy being employed to test.

Overall it's really easy know why tests should pass when writing them, and why each case is necessary, but then over time that context is lost.

Hence, anything that can be done to help carry forward the information about why the test cases are expected to pass, what they are testing, could be of big value. On the flip side, having many similar tests to update and them failing, and not being sure why they even should have passed in the first place or what they were testing, can become a very large maintenance cost.

PR #5178, opened 2019-09-03
Add new test to be run by Travis: Smoke testing

DanVanAtta:
## Overview
Add a new sub-project for smoke testing. This is end-to-end verification where we avoid
validating logic and we send requests into the system and validate that we get no error
messages.

A travis job is added to run the test. The job is powered by a shell script that coordinates
the steps for the testing:
- start lobby
- start bot
- verify the two have started and are listening on respective ports
- use smoke-test Java client to connect to lobby
- use smoke-test Java client to connect to bot
- verify logs of bot and lobby do not contain any SEVERE logging.

If the bot fails to connect to lobby, it will continue running but will log a SEVERE error.
Thus, we check logs to detect that condition.

If either lobby or bot fail to startup completely, we'll get a fail fast error using 'nc' (netcat)
which checks if the lobby or bot ports are open.

On exit, bot and lobby are stopped and logs are printed to help with debug. We will always print logs so that on successful builds we have a point of comparison, particularly to help compare against failure builds.

Note:
- The smoke test is not integrated with gradle check command as it depends on the current system state and closes current lobby and bot. If there is any manual testing then this could quickly get in the way.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[x] Configuration Change
[ ] Bug fix

## Testing

[x] Manually tested
[ ] No testing done

### Manual Testing

- Ran locally, verified various failure permutations are spotted by the test. For example, no lobby running, lobby failing to run due to DB down, no bot running
- An example of the job output can be found at: https://travis-ci.org/DanVanAtta/triplea/jobs/580343505  (or from this branch PR build)


DanVanAtta:
I actually gave that try, it's a different sort of 'headless-client'. It's really not too bad to take available players in a bot game, and it's kinda cool to join the bot game and then run the automation and see a 3rd player come in and take all the player slots.

Where it got tricky:
- UI code is heavily mixed in with model code. It was doable to inject an interface for the UI interactions, but doing that well will take time
- We're missing ability to select AI players on bots

I hacked in some code to set all the players to be AI players. The game took a very long time to get going and did not look to ever really get past 'launching.'. Eventually the AI players did start to play! But, they made no moves and just kept ending their current turn phase.

I think where we would want to go is to have the AI players actually move and let it play for a few minutes before terminating the game and then checking for errors.

PR #5184, opened 2019-09-09
Revert "Fix map position"

ron-murhammer:
Reverts triplea-game/triplea#5180

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/MapPanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/MapPanel.java
@@ -637,6 +637,8 @@ public class MapPanel extends ImageScrollerLargeView {
     super.paint(g2d);
     g2d.scale(scale, scale);
     g2d.clip(new Rectangle2D.Double(0, 0, getImageWidth(), getImageHeight()));
+    int x = model.getX();
+    int y = model.getY();
     final List<Tile> images = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


ron-murhammer:
@RoiEXLab This fix causes a really bad side effect where you can't vertically scroll maps anymore. Reverting.

RoiEXLab:
Hmm I was pretty sure I checked that.
Basically we need to set x and y = 0 on small maps, maybe you are able to come up with better criteria?

RoiEXLab:
@ron-murhammer I just double checked using Hexglobe. I can vertical-scroll just fine.
Not that scrolling is only disabled when the map is too small for the screen, so chances are your screen is simply too "high" so the map can't wrap around correctly (every part is is only displayed once, but I don't quite see how reverting this change fixes this).

So I'd highly appreciate if you could go into more detail here please, and if this takes longer to resolve, revert the revert because this bug makes small maps literally unplayable (Release Blocker-level).

ron-murhammer:
@RoiEXLab I tested revised and game of thrones. I tried all the various ways of vertically scrolling the map down and none of them did anything. I was essentially stuck on the top edge of the map. I'll retest later today to double check if you can't reproduce it.

RoiEXLab:
@ron-murhammer Thanks for clarifying I'm able to reproduce it. I'll try to come up with another condition

RoiEXLab:
Created #5186

PR #5185, opened 2019-09-09
Hide lobby client messengers & Refactor LobbyGameTableModel

DanVanAtta:
## Overview
Main goal is to 'hide' messengers interaction behind `LobbyClient.java`. In other words LobbyClient should better and more fully encapsulate `Messengers.java`

LobbyGameTableModel was refactored because it was very entangled with the messenger. That 'model' class is a swing table model, so it's actually more UI component then it is actually model. Most commits included are the refactor of that class. The big update there was to extract the backing data model into a data model class. Once that was done the test of the game table model was updated to test the new data model class, greatly simplifying that test. With that update the game table model is no longer even aware of the lobby client or messengers, helping to achieve the goal of hiding the 'messengers'.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix

## Testing

[x] Manually tested
[ ] No testing done

Connected to a local lobby, hosted a game and exited game. Verified that the hosted game showed up in the game list and was removed on exit.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/ui/GameListModel.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/ui/GameListModel.java
@@ -11,21 +11,36 @@ class GameListModel {
   private final List<Tuple<GUID, GameDescription>> gameList = new ArrayList<>();
 
   /**
-   * <p>Adds or updates a game identified by GUID.
+   * Checks if a game identified by GUID is in the data model already or not.
    *
-   * @param gameId GUID identifying a game to insert or if existing, to update.
+   * @return True if the table data model contains the given GUID, false otherwise.
+   */
+  boolean containsGame(final GUID gameId) {
+    return findGame(gameId) != null;
+  }
+
+  /**
+   * Adds a game identified by GUID.
+   *
+   * @param gameId GUID identifying the new game to insert.
+   * @param description Description for the new game.
+   */
+  void add(final GUID gameId, final GameDescription description) {
+    gameList.add(Tuple.of(gameId, description));
+  }
+
+  /**
+   * Updates a game identified by GUID.
+   *
+   * @param gameId GUID identifying the game to update.
    * @param description The new description for the game.
    * @return Returns the row index of the updated game.
    */
-  Optional<Integer> updateOrAdd(final GUID gameId, final GameDescription description) {
+  int update(final GUID gameId, final GameDescription description) {
     final Tuple<GUID, GameDescription> toReplace = findGame(gameId);
-    if (toReplace == null) {
-      gameList.add(Tuple.of(gameId, description));
-      return Optional.empty();
-    }
     final int replaceIndex = gameList.indexOf(toReplace);
     gameList.set(replaceIndex, Tuple.of(gameId, description));
-    return Optional.of(replaceIndex);
+    return replaceIndex;
   }
 
   private Tuple<GUID, GameDescription> findGame(final GUID gameId) {
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/ui/GameListModel.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/ui/GameListModel.java
@@ -1,5 +1,6 @@
 package games.strategy.engine.lobby.client.ui;
 
+import com.google.common.base.Preconditions;
 import games.strategy.net.GUID;
 import java.util.ArrayList;
 import java.util.List;
@@ -37,6 +38,7 @@ class GameListModel {
    * @return Returns the row index of the updated game.
    */
   int update(final GUID gameId, final GameDescription description) {
+    Preconditions.checkNotNull(gameId);
     final Tuple<GUID, GameDescription> toReplace = findGame(gameId);
     final int replaceIndex = gameList.indexOf(toReplace);
     gameList.set(replaceIndex, Tuple.of(gameId, description));
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/ui/GameListModel.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/ui/GameListModel.java
@@ -1,0 +1,59 @@
+package games.strategy.engine.lobby.client.ui;
+
+import games.strategy.net.GUID;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import org.triplea.lobby.common.GameDescription;
+import org.triplea.util.Tuple;
+
+class GameListModel {
+  private final List<Tuple<GUID, GameDescription>> gameList = new ArrayList<>();
+
+  /**
+   * <p>Adds or updates a game identified by GUID.
+   *
+   * @param gameId GUID identifying a game to insert or if existing, to update.
+   * @param description The new description for the game.
+   * @return Returns the row index of the updated game.
+   */
+  Optional<Integer> updateOrAdd(final GUID gameId, final GameDescription description) {
+    final Tuple<GUID, GameDescription> toReplace = findGame(gameId);
+    if (toReplace == null) {
+      gameList.add(Tuple.of(gameId, description));
+      return Optional.empty();
+    }
+    final int replaceIndex = gameList.indexOf(toReplace);
+    gameList.set(replaceIndex, Tuple.of(gameId, description));
+    return Optional.of(replaceIndex);
+  }
+
+  private Tuple<GUID, GameDescription> findGame(final GUID gameId) {
+    return gameList.stream()
+        .filter(game -> game.getFirst().equals(gameId))
+        .findFirst()
+        .orElse(null);
+  }
+
+  Optional<Integer> removeGame(final GUID gameId) {
+    final Tuple<GUID, GameDescription> gameToRemove = findGame(gameId);
+    if (gameToRemove != null) {
+      final int index = gameList.indexOf(gameToRemove);
+      gameList.remove(gameToRemove);
+      return Optional.of(index);
+    }
+    return Optional.empty();
+  }
+
+  int size() {
+    return gameList.size();
+  }
+
+  GameDescription getGameDescriptionByRow(final int i) {
+    return gameList.get(i).getSecond();
+  }
+
+  GUID getGameGuidByRow(final int rowIndex) {
+    return gameList.get(rowIndex).getFirst();
+  }
+}
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyFrame.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyFrame.java
@@ -52,7 +52,7 @@ public class LobbyFrame extends JFrame {
     chatPlayers.setChat(chat);
     chatPlayers.setPreferredSize(new Dimension(200, 600));
     chatPlayers.addActionFactory(this::newAdminActions);
-    final LobbyGamePanel gamePanel = new LobbyGamePanel(this.client.getMessengers());
+    final LobbyGamePanel gamePanel = new LobbyGamePanel(client, new LobbyGameTableModel(client));
     final JSplitPane leftSplit = new JSplitPane();
     leftSplit.setOrientation(JSplitPane.VERTICAL_SPLIT);
     leftSplit.setTopComponent(gamePanel);
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyFrame.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyFrame.java
@@ -53,7 +53,8 @@ public class LobbyFrame extends JFrame {
     chatPlayers.setPreferredSize(new Dimension(200, 600));
     chatPlayers.addActionFactory(this::newAdminActions);
 
-    final LobbyGameTableModel tableModel = new LobbyGameTableModel(client);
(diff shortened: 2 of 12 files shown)


game-core/src/main/java/games/strategy/engine/lobby/client/ui/GameListModel.java
```diff
@@ -0,0 +1,76 @@
...
+
+import com.google.common.base.Preconditions;
+import games.strategy.net.GUID;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import org.triplea.lobby.common.GameDescription;
+import org.triplea.util.Tuple;
+
+class GameListModel {
```
DanVanAtta: The code in this class is extracted from LobbyGameTableModel.

One notable update is it has a 'containsGame' API now, the table model has some small logic to do a conditional check if a game exists and then it calls 'add' or 'update' as appropriate.

game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyFrame.java
```diff
@@ -52,7 +52,12 @@ public LobbyFrame(final LobbyClient client, final LobbyServerProperties lobbySer
     chatPlayers.setChat(chat);
     chatPlayers.setPreferredSize(new Dimension(200, 600));
     chatPlayers.addActionFactory(this::newAdminActions);
-    final LobbyGamePanel gamePanel = new LobbyGamePanel(this.client.getMessengers());
+
+    final LobbyGameTableModel tableModel =
+        new LobbyGameTableModel(client.listGames(), client.isAdmin());
+    client.addGameChangeListener(tableModel);
```
DanVanAtta: Before LobbyGameTableModel had an inner object that was an instance of the listener class, the model class tooke as parameter messengers and then registered the inner listener with the messengers. This violates IOC principle, and secondly the inner listener class is not needed when the game table model can implement that interface directly.

Once the listener is not a private variable, we can set up the listener here, outside of the game table model constructor.

game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyGamePanel.java
```diff
@@ -29,30 +28,23 @@
...
   private JButton joinGame;
-  private JButton bootGame;
   private LobbyGameTableModel gameTableModel;
-  private final Messengers messengers;
+  private final LobbyClient lobbyClient;
   private JTable gameTable;
 
-  LobbyGamePanel(final Messengers messengers) {
-    this.messengers = messengers;
-    createComponents();
```
DanVanAtta: The private methods here were just combined into the constructor.

game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyGamePanel.java
```diff
@@ -29,30 +28,23 @@
...
-    setupListeners();
-    setWidgetActivation();
-  }
+  LobbyGamePanel(final LobbyClient lobbyClient, final LobbyGameTableModel lobbyGameTableModel) {
+    this.lobbyClient = lobbyClient;
+    this.gameTableModel = lobbyGameTableModel;
 
-  private void createComponents() {
-    hostGame = new JButton("Host Game");
+    final JButton hostGame = new JButton("Host Game");
```
DanVanAtta: Removing private methods allows some variables to become method local.

game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyGamePanel.java
```diff
@@ -99,66 +91,54 @@ private void createComponents() {
...
-  private void setupListeners() {
     hostGame.addActionListener(e -> hostGame());
     joinGame.addActionListener(e -> joinGame());
     bootGame.addActionListener(e -> bootGame());
-    gameTable.getSelectionModel().addListSelectionListener(e -> setWidgetActivation());
+    gameTable
+        .getSelectionModel()
+        .addListSelectionListener(
+            e -> {
+              final boolean selected = gameTable.getSelectedRow() >= 0;
```
DanVanAtta: This code is moved from the 'setWidgetActivation' method

game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyGamePanel.java
```diff
@@ -99,66 +91,54 @@ private void createComponents() {
...
-              gameTable.clearSelection();
-            }
-            mouseOnGamesList(e);
-          }
-
-          @Override
-          public void mouseReleased(final MouseEvent e) {
-            mouseOnGamesList(e);
-          }
+        new MouseListenerBuilder()
```
DanVanAtta: Using the new MouseListenerBuilder here to clean up the mouse listener code. The implementation bodies were moved to private helper methods.

game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyGameTableModel.java
```diff
@@ -1,25 +1,13 @@
 package games.strategy.engine.lobby.client.ui;
```
DanVanAtta: A more heavy refactor was done to this class, drops dependencies on Messengers altogether and simplified the test, which in turn also no longer depended on messengers.

game-core/src/main/java/org/triplea/lobby/common/GameDescription.java
```diff
@@ -54,4 +58,13 @@ public boolean isBot() {
     return hostName.startsWith(HeadlessGameServer.BOT_GAME_HOST_NAME_PREFIX)
         && HeadlessGameServer.BOT_GAME_HOST_COMMENT.equals(comment);
   }
+
+  public String getFormattedBotStartTime() {
```
DanVanAtta: This is moved from lobby game table model class, the 'null' check is added.

game-core/src/test/java/games/strategy/engine/lobby/client/ui/GameListModelTest.java
```diff
@@ -0,0 +1,63 @@
+package games.strategy.engine.lobby.client.ui;
```
DanVanAtta: This test is actually a simplified version of LobbyGameTableModelTest and is renamed.

game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyGamePanel.java
```diff
@@ -228,16 +223,13 @@ private void joinGame() {
     // we sort the table, so get the correct index
     final GameDescription description =
         gameTableModel.get(gameTable.convertRowIndexToModel(selectedIndex));
-    GameRunner.joinGame(description, messengers);
+    GameRunner.joinGame(description, lobbyClient.getPlayerName());
```
RoiEXLab: Off-Topic:
I have been thinking about what needs to be done in order to move more classes up to `game-headed`, and one of the first things that need to be done is getting rid of all the static helper methods in `GameRunner` because this class is the root of all the headed code.

Besides that, spawning a new JVM to launch a game is still a bad idea IMO, but unfortunately the whole game logic relies on it currently. (Too many `System.exit(0)`s everywhere)

game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyGameTableModel.java
```diff
@@ -154,19 +138,11 @@ public Object getValueAt(final int rowIndex, final int columnIndex) {
       case Comments:
         return description.getComment();
       case Started:
-        return formatBotStartTime(description.getStartDateTime());
+        return description.getFormattedBotStartTime();
       case GUID:
         return gameList.get(rowIndex).getFirst();
       default:
         throw new IllegalStateException("Unknown column:" + column);
```
RoiEXLab: Just wanted to point out that we usually throw `AssertionError` on impossible conditions

game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyGameTableModel.java
```diff
@@ -154,19 +138,11 @@ public Object getValueAt(final int rowIndex, final int columnIndex) {
       case Comments:
         return description.getComment();
       case Started:
-        return formatBotStartTime(description.getStartDateTime());
```
RoiEXLab: Interesting choice.
I believe I was the last person to touch this method and for me it seemed perfectly reasonable that the "UI" part gets to decide how the Instant is formatted.

If you insist this is the right thing to do, then the same should apply to `description.isPassworded()` as well as `description.getRound()` IMO, because in those cases the description only supplies the raw data, which is then formatted by this class.

game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyGameTableModel.java
```diff
@@ -26,27 +24,25 @@
...
   }
 
-  private final List<Tuple<GUID, GameDescription>> gameList = new ArrayList<>();
-
+  private final GameListModel gameListModel;
   private final boolean isAdmin;
 
   LobbyGameTableModel(final Map<GUID, GameDescription> gameList, final boolean isAdmin) {
     this.isAdmin = isAdmin;
+    gameListModel = new GameListModel();
```
RoiEXLab: This field can be initialized in-line

game-core/src/test/java/org/triplea/lobby/common/GameDescriptionTest.java
```diff
@@ -47,4 +49,22 @@ void isBotShouldReturnFalseWhen() {
...
+
+      assertDoesNotThrow(gameDescription::getFormattedBotStartTime);
+    }
+
+    @Test
+    void shouldNotThrowExceptionWhenStartedDateIsNotNull() {
+      final GameDescription gameDescription =
+          GameDescription.builder().startDateTime(Instant.now()).build();
+
+      assertDoesNotThrow(gameDescription::getFormattedBotStartTime);
```
RoiEXLab: Thinking about this, this test is kinda silly, but I guess the reasoning here is that the time differes depending on the region, so we can't really reliably test different locales.

game-core/src/test/java/games/strategy/engine/lobby/client/ui/LobbyGameTableModelTest.java
```diff
@@ -50,13 +50,6 @@ void addGame() {
     assertThat(gameListModel.size(), is(2));
   }
 
-  @Test
-  void updateGameWithNullGuidIsIgnored() {
-    gameListModel.update(null, GameDescription.builder().build());
-    assertThat(
-        "expect row count to remain 1, null guid is bogus data", gameListModel.size(), is(1));
```
RoiEXLab: Actually testing for an exception might be not that bad

game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyGamePanel.java
DanVanAtta: We also *need* to get rid of the JVM launch to be listed on Steam. I've been wondering if we should go ahead and do the game listing even if we don't get the UI refresh complete and work on that post-Steam launch. Then again, the JVM launch is a hard blocker for it.

TripleAFrame is a similar one to GameRunner, between there is some ground to cover but would go a long ways.

The code that has been moved into game-headed and lobby has, I've found, been really helpful.

I've debated a bit personally whether we want a 3rd package, one for the UI layer. Essentially we'd have "game-headed-swing" > "game-logic" > "game-core". The top layer would inject interface implementations that bind the UI to swing. I suspect it's maybe premature to go that far, but it's something I've been considering in the back of my mind.

game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyGameTableModel.java
DanVanAtta: Good call-out.  It did look like a feature envy code smell having the method in 'tableModel'. Passing a formatter object to 'description' could be a way to god too, I suspect that is over-engineered though.

Game table model though has too many responsibilities as is. It's basically a UI class, IMO should be focused mostly on layout and satisfying the swing table contract.

game-core/src/test/java/org/triplea/lobby/common/GameDescriptionTest.java
DanVanAtta: Funny enough, did get a NPE when formatting a null date and that did need to be fixed. Sometimes verifying you get no error is a reasonable minimal level of testing, better than nothing. I'm pretty sure we probably could set the locale for the test and make it more deterministic; I suspect there the test could be made stronger/better.

game-core/src/test/java/games/strategy/engine/lobby/client/ui/LobbyGameTableModelTest.java
DanVanAtta: It's just testing then the precondition statement; I think that is at least questionable. It's not as questionable as testing a getter or setter, but I would lean towards considering that as testing more framework and the language rather than our code. 

If a method were more complex, that level of testing/scrutiny could be good safety, and I do have memories where certain illegal arg conditions were not detected properly. I don't think this is clearly a good case for it though

PR #5186, opened 2019-09-09
"Fix Map Position" Bugfix

RoiEXLab:
## Overview
Follow-Up to #5184
See #5180 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[x] Bug fix

## Testing

[x] Manually tested
[ ] No testing done


DanVanAtta:
If we keep having trouble here, IMO it would be really worthwhile to extract the logic pieces so that they could be tested. That's generally the best way I know to actually be sure a section of code is correct and stays correct.

PR #5192, opened 2019-09-10
Colorize unit images

ron-murhammer:
## Overview
- Implement feature request: https://forums.triplea-game.org/topic/1548/dyeing-colorizing-unit-images
- Add ability for engine to colorize images. This will take a color code (only apply its hue and saturation) or take a HSB value (apply all 3 values, increasing brightness is often not ideal so kind of an advanced use at your own risk). It also ignores any pixels that are close to black (brightness < 10%) as especially if increasing brightness these get faded and make images look less sharp.
- Add ability to flip colorized unit images
- Add ability to place unit images in unit folder directly which are then shared across all nations (it will first check for an image in the nation folder and if not found then check the unit folder and then check the images included with TripleA). This will allow flexibility if not all units can be shared or aren't monocolored.
- Add new properties to map.properties:
  - *A color or HSB value* - apply to unit images for each nation (this will be applied to ALL units of that nation unless exempt). This will allow map makers to potentially use a single set of unit images and specify colors for each nation. Also allow users to override the default unit image color just like they can for territory colors.
  - *Flip unit* - flip colorized unit images for each nation (this will be applied to ALL units of that nation unless exempt)
  - *A list of units exempt from colorization* - this could include things like factories that are the same color for all nations or units that were already colorized by hand for each nation like AA radar or damaged battleships

## Functional Changes

[ ] New map or map update
[X] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix

## Testing

[X] Manually tested
[ ] No testing done

### Testing
- Tested being able to load unit image that are placed directly in the units folder and that the priority of unit images is: nations folder in map, units folder in map, nations folder in engine assets, units folder in engine assets
- Tested the colorization and flipping of units images with the following properties and result

**Properties**
```
units.transform.color.Germans=FF0000
units.transform.flip.Germans=true
units.transform.color.Russians=00BBBB
units.transform.brightness.Russians=25
units.transform.color.British=00FF00
units.transform.ignore=factory,fighter
```

**Result**
311a4a2b-7b46-404b-afad-18804f151b28-image

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
+++ b/game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
@@ -228,9 +228,7 @@ public class ResourceLoader implements Closeable {
    */
   public @Nullable URL getResource(final String inputPath) {
     final String path = resourceLocationTracker.getMapPrefix() + inputPath;
-    return getMatchingResources(path).stream()
-        .findFirst()
-        .orElse(getMatchingResources(inputPath).stream().findFirst().orElse(null));
+    return findResource(path).or(() -> findResource(inputPath)).orElse(null);
   }
 
   /**
@@ -244,16 +242,15 @@ public class ResourceLoader implements Closeable {
   public @Nullable URL getResource(final String inputPath, final String inputPath2) {
     final String path = resourceLocationTracker.getMapPrefix() + inputPath;
     final String path2 = resourceLocationTracker.getMapPrefix() + inputPath2;
-    return getMatchingResources(path).stream()
-        .findFirst()
-        .orElse(
-            getMatchingResources(path2).stream()
-                .findFirst()
-                .orElse(
-                    getMatchingResources(inputPath).stream()
-                        .findFirst()
-                        .orElse(
-                            getMatchingResources(inputPath2).stream().findFirst().orElse(null))));
+    return findResource(path)
+        .or(() -> findResource(path2))
+        .or(() -> findResource(inputPath))
+        .or(() -> findResource(inputPath2))
+        .orElse(null);
+  }
+
+  private Optional<URL> findResource(final String searchPath) {
+    return getMatchingResources(searchPath).stream().findFirst();
   }
 
   private List<URL> getMatchingResources(final String path) {
--- a/game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
+++ b/game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
@@ -233,6 +233,28 @@ public class ResourceLoader implements Closeable {
         .orElse(getMatchingResources(inputPath).stream().findFirst().orElse(null));
   }
 
+  /**
+   * Returns the URL of the resource at the specified path or {@code null} if the resource does not
+   * exist.
+   *
+   * @param inputPath (The name of a resource is a '/'-separated path name that identifies the
+   *     resource. Do not use '\' or File.separator)
+   */
+  public @Nullable URL getResource(final String inputPath, final String inputPath2) {
+    final String path = resourceLocationTracker.getMapPrefix() + inputPath;
+    final String path2 = resourceLocationTracker.getMapPrefix() + inputPath2;
+    return getMatchingResources(path).stream()
+        .findFirst()
+        .orElse(
+            getMatchingResources(path2).stream()
+                .findFirst()
+                .orElse(
+                    getMatchingResources(inputPath).stream()
+                        .findFirst()
+                        .orElse(
+                            getMatchingResources(inputPath2).stream().findFirst().orElse(null))));
+  }
+
   private List<URL> getMatchingResources(final String path) {
     try {
       return Collections.list(loader.getResources(path));
--- a/game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
+++ b/game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
@@ -235,10 +235,11 @@ public class ResourceLoader implements Closeable {
 
   /**
    * Returns the URL of the resource at the specified path or {@code null} if the resource does not
-   * exist.
+   * exist. Tries the given 2 paths in order first in the map resources then engine resources.
    *
    * @param inputPath (The name of a resource is a '/'-separated path name that identifies the
    *     resource. Do not use '\' or File.separator)
+   * @param inputPath2 Same as inputPath but this takes second priority when loading
    */
   public @Nullable URL getResource(final String inputPath, final String inputPath2) {
     final String path = resourceLocationTracker.getMapPrefix() + inputPath;
--- a/game-core/src/main/java/games/strategy/triplea/image/ImageColorizer.java
+++ b/game-core/src/main/java/games/strategy/triplea/image/ImageColorizer.java
@@ -1,0 +1,91 @@
+package games.strategy.triplea.image;
+
+import java.awt.Color;
+import java.awt.image.BufferedImage;
+
+/**
+ * Class used to colorize images. Based on finding the luminance from the original image then
+ * applying hue, saturation, and brightness to colorize it.
+ */
+public class ImageColorizer {
+
+  private static final int MAX_COLOR = 256;
+  private static final float LUMINANCE_RED = 0.2126f;
+  private static final float LUMINANCE_GREEN = 0.7152f;
+  private static final float LUMINANCE_BLUE = 0.0722f;
+
+  /**
+   * Apply color to the given image. This takes the hue and saturation of the given color and
+   * applies it to the image. It ignores the brightness of the color which can cause images to
+   * become poor quality.
+   */
+  public static void applyColor(final Color color, final BufferedImage image) {
+    final float[] hsb = Color.RGBtoHSB(color.getRed(), color.getGreen(), color.getBlue(), null);
+    applyHsb((int) (hsb[0] * 360), (int) (hsb[1] * 100), 0, image);
+  }
+
+  /**
+   * Apply HSB values to the given image.
+   *
+   * @param hue 0 to 360
+   * @param saturation 0 to 100
+   * @param brightness -100 to 100, which adjusts the given image's brightness
+   * @param image the image to transform
+   */
+  public static void applyHsb(
+      final int hue, final int saturation, final int brightness, final BufferedImage image) {
+
+    // Create color lookup table for luminance values
+    final int[] redLookup = new int[MAX_COLOR];
+    final int[] greenLookup = new int[MAX_COLOR];
+    final int[] blueLookup = new int[MAX_COLOR];
+    final float hueFloat = hue / 360f;
+    final float saturationFloat = saturation / 100f;
+    for (int i = 0; i < MAX_COLOR; i++) {
+      final float brightnessFloat = i / 255f;
+      final Color color = Color.getHSBColor(hueFloat, saturationFloat, brightnessFloat);
+      redLookup[i] = (color.getRed());
+      greenLookup[i] = (color.getGreen());
+      blueLookup[i] = (color.getBlue());
+    }
+
+    // Loop through image to colorize each pixel
+    for (int x = 0; x < image.getWidth(); x++) {
+      for (int y = 0; y < image.getHeight(); y++) {
+        final Color color = new Color(image.getRGB(x, y), true);
+
+        // Don't worry about colorization if pixel is transparent
+        if (color.getAlpha() == 0) {
+          continue;
+        }
+
+        // Avoid colorization if pixel is close to black so image stays sharp
+        final float[] hsb = Color.RGBtoHSB(color.getRed(), color.getGreen(), color.getBlue(), null);
+        if (hsb[2] < 0.1) {
+          continue;
+        }
+
+        // Find luminance and use lookup table to set resulting color
+        final int lum = findLuminance(color, brightness);
+        final Color finalColor =
+            new Color(redLookup[lum], greenLookup[lum], blueLookup[lum], color.getAlpha());
+        image.setRGB(x, y, finalColor.getRGB());
+      }
+    }
+  }
+
+  private static int findLuminance(final Color color, final int brightness) {
+    int lum =
+        (int)
+            (color.getRed() * LUMINANCE_RED
+                + color.getGreen() * LUMINANCE_GREEN
+                + color.getBlue() * LUMINANCE_BLUE);
+    if (brightness > 0) {
+      lum = (int) (lum * (100f - brightness) / 100f);
+      lum += 255f - (100f - brightness) * 255f / 100f;
+    } else if (brightness < 0) {
+      lum = (int) ((lum * (brightness + 100f)) / 100f);
+    }
+    return lum;
+  }
+}
--- a/game-core/src/main/java/games/strategy/triplea/image/ImageColorizer.java
+++ b/game-core/src/main/java/games/strategy/triplea/image/ImageColorizer.java
@@ -14,6 +14,8 @@ public class ImageColorizer {
   private static final float LUMINANCE_GREEN = 0.7152f;
   private static final float LUMINANCE_BLUE = 0.0722f;
 
+  private ImageColorizer() {}
+
   /**
    * Apply color to the given image. This takes the hue and saturation of the given color and
    * applies it to the image. It ignores the brightness of the color which can cause images to
@@ -44,9 +46,9 @@ public class ImageColorizer {
     for (int i = 0; i < MAX_COLOR; i++) {
       final float brightnessFloat = i / 255f;
       final Color color = Color.getHSBColor(hueFloat, saturationFloat, brightnessFloat);
-      redLookup[i] = (color.getRed());
-      greenLookup[i] = (color.getGreen());
-      blueLookup[i] = (color.getBlue());
+      redLookup[i] = color.getRed();
+      greenLookup[i] = color.getGreen();
+      blueLookup[i] = color.getBlue();
     }
 
     // Loop through image to colorize each pixel
@@ -82,7 +84,7 @@ public class ImageColorizer {
                 + color.getBlue() * LUMINANCE_BLUE);
     if (brightness > 0) {
       lum = (int) (lum * (100f - brightness) / 100f);
-      lum += 255f - (100f - brightness) * 255f / 100f;
+      lum = (int) (lum + (255f - (100f - brightness) * 255f / 100f));
     } else if (brightness < 0) {
       lum = (int) ((lum * (brightness + 100f)) / 100f);
     }
--- a/game-core/src/main/java/games/strategy/triplea/image/ImageTransformer.java
+++ b/game-core/src/main/java/games/strategy/triplea/image/ImageTransformer.java
@@ -1,29 +1,32 @@
 package games.strategy.triplea.image;
 
 import java.awt.Color;
+import java.awt.geom.AffineTransform;
+import java.awt.image.AffineTransformOp;
 import java.awt.image.BufferedImage;
 
 /**
- * Class used to colorize images. Based on finding the luminance from the original image then
- * applying hue, saturation, and brightness to colorize it.
+ * Class used to transform images including colorizing and flipping. Colorizing is based on finding
+ * the luminance from the original image then applying hue, saturation, and brightness to colorize
+ * it. More information about luminance can be found here:
+ * https://en.wikipedia.org/wiki/Relative_luminance.
  */
-public class ImageColorizer {
+public class ImageTransformer {
 
   private static final int MAX_COLOR = 256;
   private static final float LUMINANCE_RED = 0.2126f;
   private static final float LUMINANCE_GREEN = 0.7152f;
   private static final float LUMINANCE_BLUE = 0.0722f;
 
-  private ImageColorizer() {}
+  private ImageTransformer() {}
 
   /**
-   * Apply color to the given image. This takes the hue and saturation of the given color and
-   * applies it to the image. It ignores the brightness of the color which can cause images to
-   * become poor quality.
+   * Apply color and brightness to the given image. This uses the hue and saturation of the given
+   * color. It ignores the brightness of the color and uses the given brightness value.
    */
-  public static void applyColor(final Color color, final BufferedImage image) {
+  public static void colorize(final Color color, final int brightness, final BufferedImage image) {
     final float[] hsb = Color.RGBtoHSB(color.getRed(), color.getGreen(), color.getBlue(), null);
-    applyHsb((int) (hsb[0] * 360), (int) (hsb[1] * 100), 0, image);
+    colorize((int) (hsb[0] * 360), (int) (hsb[1] * 100), brightness, image);
   }
 
   /**
@@ -34,7 +37,7 @@ public class ImageColorizer {
    * @param brightness -100 to 100, which adjusts the given image's brightness
    * @param image the image to transform
    */
-  public static void applyHsb(
+  private static void colorize(
       final int hue, final int saturation, final int brightness, final BufferedImage image) {
 
     // Create color lookup table for luminance values
@@ -44,7 +47,7 @@ public class ImageColorizer {
     final float hueFloat = hue / 360f;
     final float saturationFloat = saturation / 100f;
     for (int i = 0; i < MAX_COLOR; i++) {
-      final float brightnessFloat = i / 255f;
+      final float brightnessFloat = (float) i / (MAX_COLOR - 1);
       final Color color = Color.getHSBColor(hueFloat, saturationFloat, brightnessFloat);
       redLookup[i] = color.getRed();
       greenLookup[i] = color.getGreen();
@@ -89,5 +92,12 @@ public class ImageColorizer {
       lum = (int) ((lum * (brightness + 100f)) / 100f);
     }
     return lum;
+  }
+
+  public static BufferedImage flipHorizontally(final BufferedImage image) {
+    final AffineTransform tx = AffineTransform.getScaleInstance(-1, 1);
+    tx.translate(-image.getWidth(null), 0);
+    final AffineTransformOp op = new AffineTransformOp(tx, AffineTransformOp.TYPE_NEAREST_NEIGHBOR);
+    return op.filter(image, null);
   }
 }
--- a/game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
+++ b/game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
@@ -7,6 +7,7 @@ import games.strategy.triplea.ResourceLoader;
 import games.strategy.triplea.attachments.UnitAttachment;
 import games.strategy.triplea.delegate.Matches;
 import games.strategy.triplea.delegate.TechTracker;
+import games.strategy.triplea.ui.mapdata.MapData;
 import games.strategy.triplea.util.UnitCategory;
 import games.strategy.ui.Util;
 import java.awt.AlphaComposite;
@@ -15,6 +16,8 @@ import java.awt.Dimension;
 import java.awt.Graphics2D;
 import java.awt.Image;
 import java.awt.Toolkit;
+import java.awt.geom.AffineTransform;
+import java.awt.image.AffineTransformOp;
 import java.awt.image.BufferedImage;
 import java.net.URL;
 import java.util.HashMap;
@@ -46,6 +49,7 @@ public class UnitImageFactory {
   // Scaling factor for unit images
   private double scaleFactor;
   private ResourceLoader resourceLoader;
+  private MapData mapData;
 
   public UnitImageFactory() {}
 
@@ -55,13 +59,15 @@ public class UnitImageFactory {
       final int initialUnitWidth,
       final int initialUnitHeight,
       final int initialUnitCounterOffsetWidth,
-      final int initialUnitCounterOffsetHeight) {
+      final int initialUnitCounterOffsetHeight,
+      final MapData mapData) {
     unitIconWidth = initialUnitWidth;
     unitIconHeight = initialUnitHeight;
     unitCounterOffsetWidth = initialUnitCounterOffsetWidth;
     unitCounterOffsetHeight = initialUnitCounterOffsetHeight;
     this.scaleFactor = scaleFactor;
     resourceLoader = loader;
+    this.mapData = mapData;
     clearImageCache();
   }
 
@@ -115,7 +121,7 @@ public class UnitImageFactory {
     if (images.containsKey(fullName)) {
       return Optional.of(images.get(fullName));
     }
-    final Optional<Image> image = getBaseImage(baseName, player);
+    final Optional<Image> image = getBaseImage(baseName, player, type);
     if (image.isEmpty()) {
       return Optional.empty();
     }
@@ -142,18 +148,53 @@ public class UnitImageFactory {
       final String baseImageName, final PlayerId id, final ResourceLoader resourceLoader) {
     // URL uses '/' not '\'
     final String fileName = FILE_NAME_BASE + id.getName() + "/" + baseImageName + ".png";
-    final URL url = resourceLoader.getResource(fileName);
+    final String fileName2 = FILE_NAME_BASE + baseImageName + ".png";
+    final URL url = resourceLoader.getResource(fileName, fileName2);
     return Optional.ofNullable(url);
   }
 
-  private Optional<Image> getBaseImage(final String baseImageName, final PlayerId id) {
+  private Optional<Image> getBaseImage(
+      final String baseImageName, final PlayerId id, final UnitType type) {
     final Optional<URL> imageLocation = getBaseImageUrl(baseImageName, id);
     Image image = null;
     if (imageLocation.isPresent()) {
       image = Toolkit.getDefaultToolkit().getImage(getBaseImageUrl(baseImageName, id).get());
       Util.ensureImageLoaded(image);
+      if (!mapData.ignoreColorizingUnit(type.getName())) {
+        if (mapData.getUnitColor(id.getName()).isPresent()) {
+          final Color color = mapData.getUnitColor(id.getName()).get();
+          image = convertToBufferedImage(image);
+          ImageColorizer.applyColor(color, (BufferedImage) image);
+          if (mapData.shouldFlipUnit(id.getName())) {
+            image = flipImage((BufferedImage) image);
+          }
+        } else if (mapData.getUnitHsb(id.getName()).isPresent()) {
+          final int[] hsb = mapData.getUnitHsb(id.getName()).get();
+          image = convertToBufferedImage(image);
+          ImageColorizer.applyHsb(hsb[0], hsb[1], hsb[2], (BufferedImage) image);
+          if (mapData.shouldFlipUnit(id.getName())) {
+            image = flipImage((BufferedImage) image);
+          }
+        }
+      }
     }
     return Optional.ofNullable(image);
+  }
+
+  private static BufferedImage convertToBufferedImage(final Image image) {
+    final BufferedImage newImage =
+        new BufferedImage(image.getWidth(null), image.getHeight(null), BufferedImage.TYPE_INT_ARGB);
+    final Graphics2D g = newImage.createGraphics();
+    g.drawImage(image, 0, 0, null);
+    g.dispose();
+    return newImage;
+  }
+
+  private static BufferedImage flipImage(final BufferedImage image) {
+    final AffineTransform tx = AffineTransform.getScaleInstance(-1, 1);
+    tx.translate(-image.getWidth(null), 0);
+    final AffineTransformOp op = new AffineTransformOp(tx, AffineTransformOp.TYPE_NEAREST_NEIGHBOR);
+    return op.filter(image, null);
   }
 
   /**
@@ -188,7 +229,7 @@ public class UnitImageFactory {
     if (icons.containsKey(fullName)) {
       return Optional.of(icons.get(fullName));
     }
-    final Optional<Image> image = getBaseImage(baseName, player);
+    final Optional<Image> image = getBaseImage(baseName, player, type);
     if (image.isEmpty()) {
       return Optional.empty();
     }
@@ -271,7 +312,7 @@ public class UnitImageFactory {
   public Dimension getImageDimensions(final UnitType type, final PlayerId player) {
     final String baseName = getBaseImageName(type, player, false, false);
     final Image baseImage =
-        getBaseImage(baseName, player)
+        getBaseImage(baseName, player, type)
             .orElseThrow(
                 () ->
                     new RuntimeException(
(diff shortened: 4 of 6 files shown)


game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
```diff
@@ -233,6 +233,29 @@ public boolean hasPath(final String path) {
...
+    return getMatchingResources(path).stream()
+        .findFirst()
+        .orElse(
+            getMatchingResources(path2).stream()
+                .findFirst()
+                .orElse(
+                    getMatchingResources(inputPath).stream()
+                        .findFirst()
+                        .orElse(
+                            getMatchingResources(inputPath2).stream().findFirst().orElse(null))));
```
ron-murhammer: This isn't pretty especially with the new google formatting. Open to ideas on refactoring this.

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
```diff
@@ -142,20 +148,55 @@ public Image getImage(final UnitCategory unit) {
       final String baseImageName, final PlayerId id, final ResourceLoader resourceLoader) {
     // URL uses '/' not '\'
     final String fileName = FILE_NAME_BASE + id.getName() + "/" + baseImageName + ".png";
-    final URL url = resourceLoader.getResource(fileName);
+    final String fileName2 = FILE_NAME_BASE + baseImageName + ".png";
+    final URL url = resourceLoader.getResource(fileName, fileName2);
```
ron-murhammer: This is where it now also considers the unit folder directly.

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
```diff
@@ -142,20 +148,55 @@ public Image getImage(final UnitCategory unit) {
...
+          }
+        } else if (mapData.getUnitHsb(id.getName()).isPresent()) {
+          final int[] hsb = mapData.getUnitHsb(id.getName()).get();
+          image = convertToBufferedImage(image);
+          ImageColorizer.applyHsb(hsb[0], hsb[1], hsb[2], (BufferedImage) image);
+          if (mapData.shouldFlipUnit(id.getName())) {
+            image = flipImage((BufferedImage) image);
+          }
+        }
+      }
```
ron-murhammer: This is the core of the new functionality where it checks the new map properties and applies necessary colorization and flipping to the unit image.

game-core/src/main/java/games/strategy/triplea/image/ImageColorizer.java
```diff
@@ -0,0 +1,93 @@
...
+      final int hue, final int saturation, final int brightness, final BufferedImage image) {
+
+    // Create color lookup table for luminance values
+    final int[] redLookup = new int[MAX_COLOR];
+    final int[] greenLookup = new int[MAX_COLOR];
+    final int[] blueLookup = new int[MAX_COLOR];
+    final float hueFloat = hue / 360f;
+    final float saturationFloat = saturation / 100f;
+    for (int i = 0; i < MAX_COLOR; i++) {
+      final float brightnessFloat = i / 255f;
```
DanVanAtta: Is the 255f value always equal to MAX_COLOR-1? If so, should we derive the value instead of hardcoding it?

game-core/src/main/java/games/strategy/triplea/image/ImageColorizer.java
```diff
@@ -0,0 +1,93 @@
...
+import java.awt.image.BufferedImage;
+
+/**
+ * Class used to colorize images. Based on finding the luminance from the original image then
+ * applying hue, saturation, and brightness to colorize it.
+ */
+public class ImageColorizer {
+
+  private static final int MAX_COLOR = 256;
+  private static final float LUMINANCE_RED = 0.2126f;
```
DanVanAtta: Are there any resources or references to give background information on these luminosity constants?

game-core/src/main/java/games/strategy/triplea/image/ImageColorizer.java
```diff
@@ -0,0 +1,93 @@
...
+   * @param hue 0 to 360
+   * @param saturation 0 to 100
+   * @param brightness -100 to 100, which adjusts the given image's brightness
+   * @param image the image to transform
+   */
+  public static void applyHsb(
+      final int hue, final int saturation, final int brightness, final BufferedImage image) {
+
+    // Create color lookup table for luminance values
+    final int[] redLookup = new int[MAX_COLOR];
```
DanVanAtta: Do the lookup tables save that much computation? Looking at java.awt.Color, it looks like we might be saving a few floating point operations. I wonder if there is any real performance gain here. I would suspect if there were a caching gain to be had, it would be by caching input color to final color.

game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
DanVanAtta: I'd approach a refactor by first removing the potentially unused second parameter. The usage of this method would then look like:

```
    return Optional.ofNullable(resourceLoader.getResource(FILE_NAME_BASE + id.getName() + "/" + baseImageName + ".png"))
        .or(resourceLoader.getResource(FILE_NAME_BASE + baseImageName + ".png")
```


For the method itself, the repeated part is:
```
return getMatchingResources(path).stream()
        .findFirst()
```

To fix that, I'd add a helper method, eg:

```
private Optional<URL> findResource(final String searchPath) {
   return getMatchingResources(searchPath)
        .stream()
        .findFirst()
}
```

Then using that helper method, this method becomes:

```
     return findResource(resourceLocationTracker.getMapPrefix() + inputPath)
                    .or(() ->  findResource(inputPath))
                    .orElse(null);

```

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
```diff
@@ -337,6 +346,68 @@ public int getDefaultUnitsStackSize() {
     return Math.max(0, Integer.parseInt(stack));
   }
 
+  /** Returns the unit color associated with the player named {@code playerName}. */
+  public Optional<Color> getUnitColor(final String playerName) {
+    // already loaded, just return
+    if (unitColors.containsKey(playerName)) {
```
DanVanAtta: Properties seems to store data in memory in its own map, is there a benefit to the `unitColors` cache?

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
```diff
@@ -337,6 +346,68 @@ public int getDefaultUnitsStackSize() {
...
+      return Optional.ofNullable(unitColors.get(playerName));
+    }
+    // look in map.properties
+    final Color color;
+    try {
+      color = getColorProperty(PROPERTY_UNITS_COLOR_PREFIX + playerName);
+    } catch (final Exception e) {
```
DanVanAtta: Nit, catching raw Exception can be considered a code smell. If we have a bug somewhere and throw a NPE, the error message might be incorrect. If we're going to throw an exception anyways, would we be just as well off if `getColorProperty` threw that exception (unchecked)? In this case this method would not need a try/catch block ata ll.

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
```diff
@@ -142,20 +148,55 @@ public Image getImage(final UnitCategory unit) {
...
+      if (!mapData.ignoreColorizingUnit(type.getName())) {
+        if (mapData.getUnitColor(id.getName()).isPresent()) {
+          final Color color = mapData.getUnitColor(id.getName()).get();
+          image = convertToBufferedImage(image);
+          ImageColorizer.applyColor(color, (BufferedImage) image);
+          if (mapData.shouldFlipUnit(id.getName())) {
+            image = flipImage((BufferedImage) image);
+          }
+        } else if (mapData.getUnitHsb(id.getName()).isPresent()) {
+          final int[] hsb = mapData.getUnitHsb(id.getName()).get();
```
DanVanAtta: Nit: The if/else block contains some redundant parts.

```
image = convertToBufferedImage(image);
 
if (mapData.getUnitColor(id.getName()).isPresent()) {
  ImageColorizer.applyColor(color, (BufferedImage) image);
 } else if (mapData.getUnitHsb(id.getName()).isPresent()) {
    final int[] hsb = mapData.getUnitHsb(id.getName()).get();
    ImageColorizer.applyHsb(hsb[0], hsb[1], hsb[2], (BufferedImage) image);
 }

if (mapData.shouldFlipUnit(id.getName())) {
  image = flipImage((BufferedImage) image);
 }
```

Further though, the branching logic here mirrors a bit what is done in the colorizer class.

Notably:
```
    final float[] hsb = Color.RGBtoHSB(color.getRed(), color.getGreen(), color.getBlue(), null);
    applyHsb((int) (hsb[0] * 360), (int) (hsb[1] * 100), 0, image);
```

I think then if the map data method returned a normalized HSB directly, then there would be no need for branching logic here and everything would be just HSB.

It is a bit nasty for map data to do that computation, I've also been considering if it really should know as much about value parsing as it does as well. To that extent, I'd recommend to create a strategy object that takes a mapData object, eg:

```
class UnitColorParser {
  private final MapData mapData;
   public Optional<int[]> getUnitHsb(PlayerId player) {
            // read map Data property
            // do the string splitting
            // if it is a color value, convert to HSB
            // return hsb value
   }
}
```
The test case for such a class would be pretty straight forward and help ensure we got all the error cases correct. We probably should reject extreme and out of bound values.

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
```diff
@@ -142,20 +148,55 @@ public Image getImage(final UnitCategory unit) {
...
 
-  private Optional<Image> getBaseImage(final String baseImageName, final PlayerId id) {
+  private Optional<Image> getBaseImage(
+      final String baseImageName, final PlayerId id, final UnitType type) {
     final Optional<URL> imageLocation = getBaseImageUrl(baseImageName, id);
     Image image = null;
     if (imageLocation.isPresent()) {
       image = Toolkit.getDefaultToolkit().getImage(getBaseImageUrl(baseImageName, id).get());
       Util.ensureImageLoaded(image);
+      if (!mapData.ignoreColorizingUnit(type.getName())) {
```
DanVanAtta: IMO `UnitImageFactory` is mainly about finding and loading images from disk.

The fact it has unit highlight is a bit of a code smell where it's no longer really even a factory but is also a  `UnitImageRenderer`.

It would be nice to have a full-blown adherence to the 'open-closed' principle be applied to this class, but it would be a pretty heavy lift. In the meantime, I also note we have a `mapData` dependency injected into this class, a very heavy dependency. Notably, [interface segregation principle](https://en.wikipedia.org/wiki/Interface_segregation_principle) comes to mind.

Just as the logic for unit colorization is externalized, I think the image rendering logic could also be externalized to, or moved to the colorizer class. In which case, the code here would be something like:

```
private final UnitColorTransform unitColorTransformer;
private final ImageOrientationTransform orientationTransform

image = unitColorTransform.apply(image);
image = orientationTransform.apply(image);
```

With the above, if the map data flag is not set, then we'd have a no-op and just return the parameter value. The above decreases the cyclomatic complexity and keeps UnitImageFactory from knowing everything about how images are drawn.

What do you think?

DanVanAtta:
Forgot to mention, I wanted to suggest storing the colorized 'fallback' images in a "default" folder rather than in the units folder. I think that might keep the images more organized, cohesive to the existing folder/file structure. Any thoughts there @ron-murhammer ?

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
```diff
@@ -142,20 +148,55 @@ public Image getImage(final UnitCategory unit) {
...
+  private static BufferedImage convertToBufferedImage(final Image image) {
+    final BufferedImage newImage =
+        new BufferedImage(image.getWidth(null), image.getHeight(null), BufferedImage.TYPE_INT_ARGB);
+    final Graphics2D g = newImage.createGraphics();
+    g.drawImage(image, 0, 0, null);
+    g.dispose();
+    return newImage;
+  }
+
+  private static BufferedImage flipImage(final BufferedImage image) {
```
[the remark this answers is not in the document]
ron-murhammer: Renamed and moved to ImageTransformation class.

game-core/src/main/java/games/strategy/triplea/image/ImageColorizer.java
ron-murhammer: Updated.

game-core/src/main/java/games/strategy/triplea/image/ImageColorizer.java
ron-murhammer: Added more comments to class referencing luminosity.

game-core/src/main/java/games/strategy/triplea/image/ImageColorizer.java
ron-murhammer: Some but didn't test performance improvement. For large images this could make a noticeable difference (think 1000x1000 pixels) as anything in double for loop is costly. For unit images they tend to be around 50x50 so not a huge deal but I think its better to just leave this since its a generic image transformation class.

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
ron-murhammer: Honestly, I'm not sure. This is how its done for `playerColors` so I followed the same pattern. Gonna consider examining that further out of scope for this PR.

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
ron-murhammer: Yeah, the exception handling in a lot of this class is questionable. I had followed a similar pattern as playerColors but decided to go ahead and refactor this a bit to clean it up.

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
ron-murhammer: Refactored this and updated to handle brightness instead of HSB property. Its definitely a step in the right direction I think.

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
ron-murhammer: Probably a valid point and I decided to at least move the image flipping outside of this. This could probably be taken further but would probably be better to look at the class as a whole since several other methods like highlight already do even more image manipulation then this new functionality.

ron-murhammer:
@DanVanAtta Made some updates based on your feedback. I decided to go with swapping the HSB property with brightness as I tend to agree and forum seemed to slightly favor that as well.

As for a separate folder for the shared unit images, I consider that but didn't think it was worth it. Also most of the other types of images are directly in the corresponding folder so I think its more consistent then.

game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
ron-murhammer: Refactored to use a helper method. Though your suggestion is slightly incorrect as it would change the order of the checks. We want to check the folders in this order: map nations, map units, engine nations, engine units.

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
DanVanAtta: That modularity does not strike me as correct. Why not have the image manipulation also be in colorizer, why make unit image factory do that? It seems we're building on problems. I really see problems with introducing mapData as a dependency. That would torpedo testing, violates interface segregation principle, and if we build on it will take large effect to undo. I don't think modifying existing code to this extent to add in case statements is the way to go. Easier to avoid this now. I do think it's simple to extract this new code to a new class, instead of injecting map data, inject the new class.

DanVanAtta:
> I consider that but didn't think it was worth it

Why not? The code to search there would be straight forward and arguably more cohesive as instead of shortening the search path we're altering the search folder. I fear a map with 30 unit images, 10 nations, is going to be a mess when looking at the 'units' folder.

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
ron-murhammer: I don't want to push TripleA specific implemention into `ImageTransformer` as its meant to be a generic image manipulation class that could even one day be extracted into a separate library if it became useful outside of this project. Really the only image logic left here is checking all the map data properties to determine if and what to transform. The transformation itself is in `ImageTransformer`.

Where that logic belongs you could debate but given how the classes are structured currently probably fits best where it is unless a new class is introduced. Particularly this chunk of code:
```
        if (mapData.getUnitColor(id.getName()).isPresent()) {
          final Color color = mapData.getUnitColor(id.getName()).get();
          final int brightness = mapData.getUnitBrightness(id.getName());
          ImageTransformer.colorize(color, brightness, (BufferedImage) image);
        }
        if (mapData.shouldFlipUnit(id.getName())) {
          image = ImageTransformer.flipHorizontally((BufferedImage) image);
        }
      }
```

ron-murhammer:
@DanVanAtta The main reason is consistency. If you look at all the other image folders, the shared images are directly in the folders. You can see all the folder in POS as an example: https://github.com/triplea-maps/the_pact_of_steel/tree/master/map.

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
```diff
@@ -55,13 +57,15 @@ public void setResourceLoader(
       final int initialUnitWidth,
       final int initialUnitHeight,
       final int initialUnitCounterOffsetWidth,
-      final int initialUnitCounterOffsetHeight) {
+      final int initialUnitCounterOffsetHeight,
+      final MapData mapData) {
```
DanVanAtta: I'm still pretty concerned with the new MapData  dependency being wired into the class. It strikes me as similar to the swing 'parent' component dependency that is wired throughout the codebase and we're having a lot of trouble now because of it. 

I don't think there is any better time than now to avoid the new dependency. It also would seem a lot cleaner to have a focused strategy object take its place and avoid adding more rendering code to unit image factory. The unit highlight is in the same league, but at least it does not introduce a very heavy, new dependency.

DanVanAtta:
> @DanVanAtta The main reason is consistency. If you look at all the other image folders, the shared images are directly in the folders. You can see all the folder in POS as an example: https://github.com/triplea-maps/the_pact_of_steel/tree/master/map.

Thanks for the reference @ron-murhammer , though I'm not quite spotting any great examples. Could you point me to a specific folder? I see a number of examples where we have files and then the folder contained is more "extras" that should probably not be checked in to that location.

The units folder is consistent with no unit images in it, just more folders. I think it's quite clean looking as-is: https://github.com/triplea-maps/the_pact_of_steel/tree/master/map/units, adding a 'default' folder I believe would maintain that.

game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
```diff
@@ -228,9 +228,29 @@ public boolean hasPath(final String path) {
...
+
+  /**
+   * Returns the URL of the resource at the specified path or {@code null} if the resource does not
+   * exist. Tries the given 2 paths in order first in the map resources then engine resources.
+   *
+   * @param inputPath (The name of a resource is a '/'-separated path name that identifies the
+   *     resource. Do not use '\' or File.separator)
+   * @param inputPath2 Same as inputPath but this takes second priority when loading
+   */
+  public @Nullable URL getResource(final String inputPath, final String inputPath2) {
```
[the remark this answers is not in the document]
ron-murhammer: I think the only way to take this further is have the single arg version call the new double arg or combine them together to take vararg or a list. That would increase the complexity of the new method since you need to handle a null or empty second parameter then. My general feel is there isn't much left to be gained here and the code is pretty clean after the helper method addition.

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
```diff
@@ -48,15 +48,19 @@
   public static final String PROPERTY_UNITS_SCALE = "units.scale";
   public static final String PROPERTY_UNITS_WIDTH = "units.width";
   public static final String PROPERTY_UNITS_HEIGHT = "units.height";
+  public static final String PROPERTY_UNITS_COUNTER_OFFSET_WIDTH = "units.counter.offset.width";
+  public static final String PROPERTY_UNITS_COUNTER_OFFSET_HEIGHT = "units.counter.offset.height";
+  public static final String PROPERTY_UNITS_STACK_SIZE = "units.stack.size";
+  public static final String PROPERTY_UNITS_COLOR_PREFIX = "units.color.";
+  public static final String PROPERTY_UNITS_COLOR_BRIGHTNESS_PREFIX = "units.color.brightness.";
+  public static final String PROPERTY_UNITS_COLOR_FLIP_PREFIX = "units.color.flip.";
```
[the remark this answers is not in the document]
ron-murhammer: Both color and flip uses the same ignore list as I had a hard time imagining you wanting 2 separate ignore lists. Generally, I don't users will be flipping images unless you are already coloring them. I consider it more of an extension of coloring images than a separate feature.

ron-murhammer:
@DanVanAtta I think you're missing my point. So look at the resources folder specifically: https://github.com/triplea-maps/the_pact_of_steel/tree/master/map/resources. The shared resource images are placed directly in that folder. Now image I eventually want nation specific icons for resources so I then allow nation folders inside there. If I go and create some default folder for unit images this doesn't align with all the other image folders where images are directly in the root folder.

game-core/src/main/java/games/strategy/triplea/image/ImageTransformer.java
```diff
@@ -0,0 +1,103 @@
...
+   */
+  public static void colorize(final Color color, final int brightness, final BufferedImage image) {
+    final float[] hsb = Color.RGBtoHSB(color.getRed(), color.getGreen(), color.getBlue(), null);
+    colorize((int) (hsb[0] * 360), (int) (hsb[1] * 100), brightness, image);
+  }
+
+  /**
+   * Apply HSB values to the given image.
+   *
+   * @param hue 0 to 360
```
DanVanAtta: Does the game disallow the out of range values anywhere? Beyond the javadoc, would a precondition check here help ensure we do not continue with invalid input?

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
ron-murhammer: While I don't particularly like injecting `MapData` here, I'm not really seeing a quick win to avoid it in a way consistent with how the classes are structured at the moment. The only thing I can imagine doing without a major refactor is instead create a new class (something like `UnitImageRender`) and inject `MapData` into that then inject that class into `UnitImageFactory` instead. 

My general feel around that is it ends up being a half baked refactoring where a lot of the other methods in `UnitImageFactory` need to be included in that refactoring to essentially split what `UnitImageFactory` does now into 2 separate functions. And even doing that you still end up with having to inject `MapData` into that new class unless you also go and do a major refactoring to `MapData` to split it into smaller pieces.

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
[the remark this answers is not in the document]
DanVanAtta: Ah, didn't realize they were using the same ignore list. I personally did not draw the same connection that the two were related. "color.flip" I would interpret naively as inverting the color, not changing the orientation.

Is it possible to use 'flip' without using colors? It seems the two concepts should maybe not be so coupled. It might be okay maybe to duplicate some ignore logic to have two independent lists in case we do want to start coloring factories but not flip them. I wonder if that level of flexibility would actually be a good thing.

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
DanVanAtta: > The only thing I can imagine doing without a major refactor is instead create a new class (something like UnitImageRender) and inject MapData into that then inject that class into UnitImageFactory instead.

Yes! :smile: 


>  half baked refactoring where a lot of the other methods in UnitImageFactory need to be included in that refactoring to 

You have a good point, but the other elements in UnitImageFactory would not necessarily go into that same strategy object, but different ones. Thinking about SRP, it woudl be just as inappropriate to have one class know how to do every kind of image rendering, just as it's not "appropriate" for UnitImageFactory to know how to do all image rendering.

I think the wins with the half baked refactoring are:
- we stop the bleeding
- do not make an existing problem worse
- we would not be doing any kind of throw-away work, the extracted code would not need to be touched, the existing code would then build on that pattern and continue isolating and splitting functionality in a similar pattern

> And even doing that you still end up with having to inject MapData into that new class unless you also go and do a major refactoring to MapData to split it into smaller pieces.

True, though I think having just the one dependency on MapData will be easier to deal with compared to having many parameters and mapData being one of them.
 
>  If there is a quick win here I'm all for it but everything I see ends up a major effort or it ends up half baked.

I think the half baked version is the quick win. I view it much less as half baked compared to not making an existing problem worse.

DanVanAtta:
> @DanVanAtta I think you're missing my point. So look at the resources folder specifically: https://github.com/triplea-maps/the_pact_of_steel/tree/master/map/resources. The shared resource images are placed directly in that folder. Now image I eventually want nation specific icons for resources so I then allow nation folders inside there. If I go and create some default folder for unit images this doesn't align with all the other image folders where images are directly in the root folder. Folder structures are also hierarchical for a reason, I don't really see much of a reason not to take advantage of that and start creating "magic named" folders like default/shared/something.

Personally, I think the folders already have a large degree of magic naming, "units", and the "player names", etc.. I'm not sure I quite have the same view of hierarchy. Personally I would expect the highest level to take precedence, rather than the lowest level taking precedence, but that is my subjective view on it (ie: first found wins when descending the tree vs last found wins).

But, indeed I was missing your point. Thank you for pointing out the behavior of resources folder, it was not obvious looking at it. I suppose that is pretty compelling even if I don't necessarily like the original decision. It does make some sense when there are no player level overrides. Though I think it would have been clearer to begin with if items were in a 'default' folder, in that case looking at the resources folder, it would be more obvious that you could add overrides there.

TL;DR - introducing 'default' folder I do agree now would be inconsistent with how the other folders work.

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
DanVanAtta: > or create a new class for these 20 or so lines

I don't have any problem with a 20 line class file. It encapsulates the image rendering code and buffers UnitImageFactory from knowing about MapData. I suspect such a class might be readily testable, which is a good sign for design.

IMO it's similar to how the raw colorizer code is extracted so we don't have UnitImageFactory dealing with the raw colorizer directly nor map data. I think that creates a pretty nice design tree where we have map data and colorizer coming together in one module (class), and then unit image factory uses that class.

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
DanVanAtta: I strongly suspect the caching is a cache on top of a cache already. I think the issue is the original "read from map.properties" comment was misleading/not correct, leading to the extra caching.

Not ideal to keep as is, would be a bit simpler without the cache -  but okay.. We can leave it for another day. A TODO comment indicating this uncertainty in caching would be good to see, but not strictly necessary.

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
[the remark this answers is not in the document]
ron-murhammer: Open to naming it something besides flip. I don't think there are many actual use cases for flipping units where you aren't colorizing them so separate lists I think are unnecessary and starting to over engineer it.

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
ron-murhammer: I guess we just see different value in the half baked version vs consistency. I feel like the minimal improvements are outweighed by the inconsistency it creates across the rest of the class. This can then cause confusion for the next developer that reads this class and wonders "why is the colorizing check logic in a separate class but the highlighting logic is in this class?" If classes are frequently getting updated and we believe the rest of the class would be refactored in short order then I see the benefit of the short term inconsistency to create the pattern to follow. But when the reality is it probably won't be touched again for a long time, we then live with the inconsistency for a long time which actually makes the code base harder to work with.

game-core/src/main/java/games/strategy/triplea/image/ImageTransformer.java
ron-murhammer: Everything but brightness is derived from Color (which is already validated) so those values have to be valid. I added a validation when reading the brightness property to ensure its within the allowed range. I think those checks are sufficient for now. If this class would become a library or be used in other areas then adding validation of parameters would probably be a good idea.

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
ron-murhammer: Resolving this comment so we can just have the conversation in one comment thread that I responded to.

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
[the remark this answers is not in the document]
DanVanAtta: Personally I'm like for the prefix to be `units.image.*`

Perhaps one of these may strike you, some suggestions I could think of:
```
units.flip.horizontal=true
units.rotate.horizontal=true
units.image.rotate=180
units.image.rotate.horizontal=true
units.image.rotate=horizontal
```

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
DanVanAtta: Hmm, reading this class, my reaction is, "oh damn.. the unit highlight code is here". Having other, unrelated rendering logic extracted would leave the same reaction and avoid two WTFs of "oh wtf, all sorts of rendering code is here"

So I don't see it as half baked at all, it's strategically inserting new code without modifying a bunch of existing code. In general adding various if statements and a large chunk of new logic to existing code is a design smell (open-closed principle violation, probably SRP violation as well).

I fear that if we don't come back to this soon, then it'll be even harder to refactor later and it'll be yet more work to do. Worse yet, if we add more rendering logic, we could apply the same logic and have a third thing.. And then there is more dependency on mapData, maybe we bring in a GameData object... 

I really just don't see the inconsistency, having all rendering code in one place is not consistent IMO, it's just a complete lack of modularity and organization of code.

game-core/src/main/java/games/strategy/triplea/image/ImageTransformer.java
DanVanAtta: Given it's a public API, the defensive programming part of me, and the 'fail early' principle makes me feel like the parameters should be validated, given they are not constrained by type. There is a calculation that occurs to get us to a HSB value, any bugs there could send incorrect data here. Having the checks would nip that in the bud. Any such incorrect values perhaps would not go on to be detected and we'd get bad rendering without necessarily knowing why. I'd rather see the bug revealed in testing as an exception then later post-deployment.

game-core/src/main/java/games/strategy/triplea/image/ImageColorizer.java
DanVanAtta: Interesting performance results.

Original: 1.8ms
No-lookup table: 1.7ms
Caching the before to after color: 2.1ms


The code with the cached lookup table is below (the no-lookup table is the same but without the hashmap):

```
  private static final Map<Color, Color> lookupTable = new HashMap<>();
  private static void colorize(
      final int hue, final int saturation, final int brightness, final BufferedImage image) {

    // Create color lookup table for luminance values
    final float hueFloat = hue / 360f;
    final float saturationFloat = saturation / 100f;

    // Loop through image to colorize each pixel
    for (int x = 0; x < image.getWidth(); x++) {
      for (int y = 0; y < image.getHeight(); y++) {
        final Color color = new Color(image.getRGB(x, y), true);

        // Don't worry about colorization if pixel is transparent
        if (color.getAlpha() == 0) {
          continue;
        }

        if(lookupTable.containsKey(color)) {
          image.setRGB(x, y, lookupTable.get(color).getRGB());
          continue;
        }

        // Avoid colorization if pixel is close to black so image stays sharp
        final float[] hsb = Color.RGBtoHSB(color.getRed(), color.getGreen(), color.getBlue(), null);
        if (hsb[2] < 0.1) {
          continue;
        }

        // Find luminance and use lookup table to set resulting color
        final int lum = findLuminance(color, brightness);
        final Color finalColor = computeColor(hueFloat, saturationFloat, lum);
        lookupTable.put(color, finalColor);
//            new Color(redLookup[lum], greenLookup[lum], blueLookup[lum], color.getAlpha());
        image.setRGB(x, y, finalColor.getRGB());
      }
    }
  }


  private static Color computeColor(final float hueFloat, float saturationFloat, int lum) {
    int i = lum;
//    for (int i = 0; i < MAX_COLOR; i++) {
      final float brightnessFloat = (float) i / (MAX_COLOR - 1);
      return Color.getHSBColor(hueFloat, saturationFloat, brightnessFloat);
  }
```

It's very surprising that the array lookup is *slower*

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
DanVanAtta: Another point of view, `ImageColorizer` I kinda liked; with ImageTransformer it actually seems more inconsistent now than before. At a bit of a deeper level really. UnitImageFactory has several operations of 'image->image', and ImageTransformer has similar. So we have scaling and highlight in unit image factory, and we have coloring and flipping in ImageTransformer. 

I suppose to really be consistent it would all be one class.. Personally I have *zero* problem with small, focused classes; having small and easy to understand, well tested modules IMO is a good way to go. To that extent I kinda liked "ImageColorizer" as a class and would have pulled out the 'flip' code to something like "ImageRotator"

With the above perspective, could be suggested that have any external class at all is not consistent. I'm spotting in this review that I suspect it could be simpler to extract even more from UnitImageFactory, and the extracted and new code perhaps might not be just a single new class.

game-core/src/main/java/games/strategy/triplea/image/ImageTransformer.java
DanVanAtta: My mistake, it is a private API. I'm not used to seeing fully formal javadocs on private methods.

Looking at that now, I wonder a bit why the two methods are not combined now. Going a bit further, it is a bit odd that for example we multiply 'hue' by 360 to then just divide it by 360.

game-core/src/main/java/games/strategy/triplea/image/ImageColorizer.java
ron-murhammer: Well its not much different either way with those results. I would be interested for larger images if you see the same results.

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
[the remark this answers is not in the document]
ron-murhammer: So rotate is actually a different image operation: https://www.fotojet.com/features/photo-editor/rotate-image.html. I'll think about remaining these properties a bit. I think the main issue is actually just the prefix and the reason is I added flip at the very end. I think it should probably be something like `units.transform` or `units.image`. Essentially the same change I made where I originally have the class `ImageColorizer` and then changed it to `ImageTransformer`.

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
ron-murhammer: I actually think that the other image transformation functions in `UnitImageFactory` potentially belong in `ImageTransformer` like scaling/highlighting as these are generic image transformations and IMO fit well together. I don't really think you need a separate class for each of these given that they are closely related and fairly small amount of code. Now if eventually `ImageTransformer` became too large then you could break it into smaller pieces.

game-core/src/main/java/games/strategy/triplea/image/ImageTransformer.java
ron-murhammer: The problem is really that the java color apis deal with decimal representation of HSB values which isn't very common. You could argue to combine them now though I think there is value in keeping them separate as the actual algorithm is meant to work on HSB values and for TripleA it just makes more sense to use color code and brightness. Once I did that the original HSB method is no longer called outside the class so is private but I think still worth keeping as is.

DanVanAtta: Ok. :+1: 
 IMO this is an opportunity to make the 2 line method, where one line is another method call, take on more responsibility. At that point 'colorize' would go to a package protected module so that it's usage is then encapsulated.

ron-murhammer:
@DanVanAtta Updated the method name. 

If its really only a 5% gain even on that larger size image then I would tend to agree that its probably not worth the complexity. Though looking quickly at your code, I don't believe you are handling the alpha of the image as you commented this line and don't see to reference the alpha anywhere else:
```
new Color(redLookup[lum], greenLookup[lum], blueLookup[lum], color.getAlpha());
```

And at least the most straightforward way to handle adding the alpha is then to create a second Color object from the values of the one you create with HSB. I think that is actually where the main performance gain is as with the lookup table you just created the first color object 256 times rather than for every pixel in the image.

DanVanAtta:
@ron-murhammer interesting read. It's a bit of a shame we can't discuss with higher bandwidth; I don't have quite the same view of consistency and think that we break consistency a bit more than maintain it in some ways with this update. The args is a great example, the previous was consistent about pulling properties off of MapData, then we pass MapData itself (inconsistent), and as mentioned the rendering code is not completely contained in UnitImageFactory. I more was advocating for more design be given to injecting this new code.

Sometimes I do not appreciate the consistency argument very much as it seems cover for injecting yet more functionality somewhere it does not belong. Generally if a chunk of logic is made more complicated, the whole thing needs to be extracted so the cyclomatic complexity of the original code can be maintained or improved. I find if anything, that is perhaps the biggest source of code decay. It's incremental, an additional parameter here, another if statement, a loop inside, another if statement, then suddenly the code is a bear to refactor, it is left for the next person; and then they say "well, it was already a mess and I just added an if statement".

Re: performance

I could have mis-measured it, I attempted to extract the array logic, but perhaps missed it. There is `PerfTimer` available to do quick measurements. EG:
```
try(PerfTimer timer = PerfTimer.start("timer-name")) {
  // code to measure
}
```
When the above exists, an info log statement is emitted.

PR #5210, opened 2019-09-13
Add a stalebot config, close issues after 120 days, PRs after 3 weeks.

DanVanAtta:
Stalebot config documentation is at: https://github.com/probot/stale
A default template is available here: https://probot.github.io/apps/stale/
Submitted config is mostly default, with addition to a 3 week PR threshold.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[x] Configuration Change
[ ] Bug fix
[ ] Other:

## Testing

[ ] Manually tested
[x] No testing done

## Additional Review Notes
Happy for this to be a 'working', iterative PR; feedback on the thresholds and text are very welcome

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -8,12 +8,12 @@ exemptLabels:
   - "Impact: Bad Game Rules"
   - "Impact: Game Crashing"
 # Label to use when marking an issue as stale
-staleLabel: 'ice box - close and revisit later'
+staleLabel: 'Stale'
 
 markComment: >
   The TripleA team regrets this issue was not solved sooner. To keep focus
   on latest issues and the most pressing tasks, 
-  This issue has been automatically marked as stale because it has not 
+  this issue has been automatically marked as stale because it has not 
   had recent activity.  If the issue may be closed, please do so. 
   If there are remaining items, we encourage for those items to be 
   resubmitted as new, independent, concise tasks. The strategy
@@ -26,5 +26,5 @@ pulls:
    markComment: >
      This pull request has been automatically marked as stale because it has not had
      recent activity. We are eager to see this work completed, please update and
-     re-open as soon as possible, thank you for your contributions!
+     re-open as soon as possible.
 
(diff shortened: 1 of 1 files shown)


ron-murhammer:
Looks pretty good. I'd probably lean a bit more conservatively to start and maybe to 180 days for issues and 30 days for PRs but I'm fine moving forward with 120/21 if you feel that's better. I'd say let's give @RoiEXLab a day or so to comment then merge.

DanVanAtta:
Interesting consideration for 180 vs 120

That is a threshold difference of March 17 or May 17. Looking for some data, these issues would be impacted:

Screenshot from 2019-09-13 15-43-30

Looking at those issues, they seem to be pretty stale and I don't think will get any further attention even if another 30-60 days go by to put them over the 180 threshold.  The stalebot adds a warning and I think the clock resets if there is a comment; getting a warning sooner is probably better to avoid an issue from getting even older before we pay attention to it again.

The CLA issue is a good example where we should be mindful of staleness thresholds (seems like we should resolve that one and give it some attention to close the matter). That issue is almost argument for us marking issues as stale at say 90 days, but increasing the close window to something like 30 or 60 days. Currently it's stale at 120, and close after 3. Maybe stale at 90 and close at +30 or +60 could make more sense.

@RoiEXLab eager to hear your thoughts and @ron-murhammer if you have any further thoughts given the above considerations.

RoiEXLab:
I'm fine with the current config 👍

Interestingly enough the bot/lobby restart issue should be resolved. Last time the lobby restarted no problem.
Bots restarted as well.
However I'm not sure if the bot reconnect threshold does currently make sense though, so it might still be possible, that it takes way too long for bots to reconnect to the lobby. I recall someone submitting a PR which fiddled around with this code, but I don't remember if the time was reduced.

ron-murhammer:
Looks like we are good on this. @DanVanAtta I'll let you take the final look over and merge it.

DanVanAtta:
Thanks for going over the config. I updated the stale label to be 'Stale' instead of ice-box. If an issue is re-cut into new issues and closed, we might wind up still having the ice-box label to revisit. I think it would also be useful to know what we manually chose to revisit vs the topics that just dropped off as stale.

Second, made a couple minor text tweaks.

PR #5209, opened 2019-09-13
Add Snyk badge to README

ron-murhammer:
## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature update or enhancement
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix
[ ] Other:

## Testing

[ ] Covered by existing automated tests
[ ] Covered by newly added automated tests
[ ] Manually tested
[ ] No testing done

PR #5212, opened 2019-09-14
Remove duplicate slash

RoiEXLab:
Just noticed this in #5209


RoiEXLab:
Looks like there's still some flaky coverage

PR #5217, opened 2019-09-16
Update to User-Roles and add Auth Layer to HttpServer

DanVanAtta:
(1) Update to use user-roles
 - Instead of having flags at DB level for 'isAdmin', we instead of a string
   field to indicate the user role. Available user roles are 'ADMIN', 'MODERATOR',
   'PLAYER' and 'ANONYMOUS'. The latter two are essentially the same for now,
   in the future there may be some differences, for example allowing a PLAYER to
   update their account but not an anonymous user.

(2) Add Http Server Authentication and Authorization
 - Use framework components to validate users are authenticated and authorized.
   We use 'OAuth' as that maps to api tokens, BasicCredentials would be username
   and password, and given we represent an API-key with a user-id on the database,
   we do not need the username value, just the API key. Note this is in contrast
   with a JWT token which would not need a database lookup at all.
 - Config allows for caching of user credentials so we will not always
   need to do a DB lookup to validate an API key
 - DB is updated to have a new 'api_key' table to store the API keys used for
   authentication.

(3) Update HttpServer 'Controller' integration tests to use utility methods
 - There was some redundancy in the tests, 'DropWizardTest' is now a top-level
   abstract test with two subclasses, BasicEndpointTest and ProtectedEndpointTest.
   The latter will help automatically verify authentication is required to access
   an endpoint.

(4) Fix up Moderators Tab in Toolbox
 - Some code to generate API keys was left in, that was fixed up.

 

## Functional Changes

[ ] New map or map update
[x] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix
[ ] Other:

## Testing

[x] Manually tested
[ ] No testing done

Launched a lobby and http-server, connected with the 'test/test' command and demo'ed the features of the moderator toolbox.


http-server/src/test/java/org/triplea/server/http/DropwizardTest.java
```diff
@@ -0,0 +1,51 @@
...
+    @Override
+    public void beforeAll(final ExtensionContext context) {
+      if (!started) {
+        started = true;
+        try {
+          support.before();
+        } catch (final RuntimeException e) {
+          // ignore, server is already started
+        }
+        context.getRoot().getStore(GLOBAL).put("dropwizard-startup", this);
```
RoiEXLab: Is this store ever accessed?

http-server/src/test/java/org/triplea/server/http/DropwizardTest.java
```diff
@@ -0,0 +1,51 @@
...
+      implements BeforeAllCallback, ExtensionContext.Store.CloseableResource {
+    private static boolean started = false;
+
+    private static final DropwizardTestSupport<AppConfig> support =
+        new DropwizardTestSupport<>(ServerApplication.class, "configuration-prerelease.yml");
+
+    @Override
+    public void beforeAll(final ExtensionContext context) {
+      if (!started) {
+        started = true;
```
RoiEXLab: In case we ever want to run our tests in parallel: Maybe using `AtomicBoolean` might be a little bit "safer" here.

http-server/src/test/java/org/triplea/server/http/DropwizardTest.java
```diff
@@ -0,0 +1,51 @@
...
+        } catch (final RuntimeException e) {
+          // ignore, server is already started
+        }
+        context.getRoot().getStore(GLOBAL).put("dropwizard-startup", this);
+      }
+    }
+
+    @Override
+    public void close() {
+      support.after();
```
RoiEXLab: TBH I have no Idea how the JUnit5 extension model works, but how do you tell that close is only called after all tests were executed and why isn't there a similar thing for instantiation?
The [documentation](https://junit.org/junit5/docs/current/user-guide/#extensions-keeping-state) suggests that the constructor might be the corresponding initializer

lobby-db-dao/src/main/java/org/triplea/lobby/server/db/data/UserRole.java
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Class to hold constants representing the possible set of user role values. <br>
+ * Note: this is not an enum to allow these values to be referenced from annotations.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class UserRole {
+  public static final String ADMIN = "ADMIN";
+  public static final String MODERATOR = "MODERATOR";
+  public static final String PLAYER = "PLAYER";
+  public static final String ANONYMOUS = "ANONYMOUS";
```
RoiEXLab: What about using an enum here?
I believe somewhere else in the code we should have a similar enum (ADMIN, NOT_ADMIN IIRC)

lobby/src/main/java/org/triplea/lobby/server/db/UserController.java
```diff
@@ -140,15 +140,16 @@ public String getUserEmailByName(final String username) {
...
     try (Connection con = connection.get();
         PreparedStatement ps = con.prepareStatement(sql)) {
       ps.setString(1, username);
       try (ResultSet rs = ps.executeQuery()) {
         if (!rs.next()) {
           return false;
         }
-        return rs.getBoolean("admin");
+        final String role = rs.getString("role");
+        return "ADMIN".equals(role) || "MODERATOR".equals(role);
```
RoiEXLab: I'm not sure about the compile-time dependencies, but it would be nice if those strings were stored in single global constants instead of hardcoding them here.

http-server/src/main/java/org/triplea/server/access/AuthenticatedUser.java
```diff
@@ -0,0 +1,32 @@
...
+ * that throws an {@code UnsupportedOperationException} if accessed.
+ */
+@Builder
+public class AuthenticatedUser implements Principal {
+  @Getter @Nullable private final Integer userId;
+  @Getter @Nonnull private final String userRole;
+
+  @Override
+  public String getName() {
+    throw new UnsupportedOperationException("Name lookup is not done on authentication.");
```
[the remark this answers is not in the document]
DanVanAtta: > I'm curious to know if we should implement it this way at some point.

Is a link missing in the comment? :smile:  
The reasoning behind unsupported was because 'getName' would reasonably return the player name. We do not use that field, looking it up from DB just so we can fill in an implementation that we do not use seemed wrong.

http-server/src/main/java/org/triplea/server/access/AuthenticatedUser.java
```diff
@@ -0,0 +1,32 @@
...
+import org.triplea.lobby.server.db.data.UserRole;
+
+/**
+ * AuthenticatedUser is anyone that has successfully logged in to the lobby, whether via anonymous
+ * account or registered account. Because we must extend Principle, we have a no-op 'getName' method
+ * that throws an {@code UnsupportedOperationException} if accessed.
+ */
+@Builder
+public class AuthenticatedUser implements Principal {
+  @Getter @Nullable private final Integer userId;
```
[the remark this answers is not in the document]
DanVanAtta: `0` would then be a magic number. To avoid that, using an explicit null.

http-server/src/test/java/org/triplea/server/http/DropwizardTest.java
DanVanAtta: Not sure, it's moved code; I did not validate whether it was originally needed or not.

lobby-db-dao/src/main/java/org/triplea/lobby/server/db/data/UserRole.java
DanVanAtta: An enum makes sense, but it takes a lot of hackery to then use strings in annotations. My first approach was an enum, but then you wind up creating a static inner class to hold the string values, and you map the enum constants to strings, and then you wind up needing to override @ToString and then there is confusion whether 'name()' or 'toString' should be used.

Seemed easier in the end to just use string constants. There is a comment in javadoc to indicate an enum was considered: ` Note: this is not an enum to allow these values to be referenced from annotations.`

lobby/src/main/java/org/triplea/lobby/server/db/UserController.java
DanVanAtta: It was a pretty late change to push 'UserRole' to the DB level, originally these constants were not available to 'lobby'. In the end I did push 'UserRole' from 'http-server' to 'lobby-db', so the constants can be used here now for sure. 

Updated in: d2d1ce369

http-server/src/test/java/org/triplea/server/http/DropwizardTest.java
DanVanAtta: Modifying moved code more properly belongs in a follow-up update. As a PR author, it greatly increased test scope to modify moved code. That can be bad either because it adds more overall work, or that testing is skipped, which is super-dangerous and a source of bugs.

Regardless, it is updated to use AtomicBoolean now:  63ff37578

http-server/src/test/java/org/triplea/server/http/DropwizardTest.java
DanVanAtta: The `@Close` is not called without this. I added a comment: ca32d865d88a505efc2fe145254ab074e13606bc

http-server/src/test/java/org/triplea/server/http/DropwizardTest.java
DanVanAtta: Seems to be the magic of the extension. When this code was originally introduced it was tested and we've been using it for some time now too; so there it's battle-tested and we can have some confidence it works. More practically, removing the context registration call and this is no longer called, so it's Junit5 extension magic happening here.

DanVanAtta:
@RoiEXLab thanks for the initial look, updated and/or responded to all comments

http-server/src/main/java/org/triplea/server/access/AuthenticatedUser.java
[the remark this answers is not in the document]
RoiEXLab: Well at some point we need the name of a player in some controller class when more and more code migrates to the HTTP scheme. So I was wondering if this was just a case of "too difficult" for almost no benefit, or rather "doesn't work well with our codebase", because the name is already looked up using another fancy way.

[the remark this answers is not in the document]
DanVanAtta: If you have a user-id, then you have the name in effect. 
In general the codebase assuming usernames are globally unique, and immutable, is not great. So i think if you need the name, it's probably to display to the user, but the client already knows the local name. If you do need name, then it should be looked up by user_id. We could do that here on authorization, but YAGNI and it'd be bad to encourage for user name to be used and not user_id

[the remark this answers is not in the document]
RoiEXLab: Hmm I was actually thinking about doing this lookup as part of this method in some sense, so whenever this object is created, the code looks up the display name and stores it along with the user-id. This way the name would be cached in memory and reduces traffic for the database.
Alternatively the object could of course also hold the jdbi information, but I'm not sure if that's a good thing to do. This approach would keep the name up-to-date, simply hiding the complext DB logic somewhere.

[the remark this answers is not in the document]
DanVanAtta: > This way the name would be cached in memory and reduces traffic for the database.

On the backend that probably should not be needed. Using 'user_id' is preferred. For example, if we update another table row to mark which user performed an action, it should be by user-id, that goes for many cases, like a potential 'chat-history'. Essentially 'user-name' is in effect a display-level concern, a concern for the UI, the backend should not really care about user-names (since user names are actually a many-to-many relationship relative to users).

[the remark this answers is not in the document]
DanVanAtta: To be explicit, TripleA using 'username' as PK is a very early mistake in data modeling. The data modeling was in some ways kinda primitive.

PR #5218, opened 2019-09-16
Update PR Template - Remove Bug Section & Cleanup

DanVanAtta:
- Remove the dedicated bug section. Context such as root cause, how the bug is fixed belong in the commit comment. Any notes we write in a PR should be useful for the reviewer, we want to minimize this so that there is more context in the commit history and not in the PR review history.

- Update formatting a bit so that bug link goes next to the 'Bug fix' checkbox, keeps things more cohesive

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[x] Configuration Change
[ ] Bug fix
[ ] Other:

## Testing

[ ] Manually tested
[x] No testing done

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,26 +7,15 @@
 
 
 ## Functional Changes
-<!-- Put an X next to an item, if other, please specify -->
+<!-- Put an X next any that apply -->
 [ ] New map or map update
 [ ] New Feature
 [ ] Feature update or enhancement
 [ ] Feature Removal
 [ ] Code Cleanup or refactor
 [ ] Configuration Change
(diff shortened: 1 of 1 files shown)


DanVanAtta:
We should probably think about this one. "Problem" was introduced as a label so we don't have arguments over if something is a bug or not. Bad game rules can be considered a bug, we don't want to have issues that have many redundant labels. 
It's also possible to just comment on an issue, eg: "Is this still a problem?" to restart the timer. On one hand that could be a good way to prune the lower priority issues that are dropping away, but on the other it might not be the best process.

For consideration, do we want to rename the "Impact.*" rules to be "Impact: Incorrect Behavior", or add a new label "Impact: Incorrect UI Behavior"

PR #5219, opened 2019-09-16
Add "Impact: Incorrect UI Behavior" label to stalebot exclusion label

DanVanAtta:
- This update will have stalebot exclude issues labeled with 'Bug'

Looking at problems that have been marked as stale, it does not quite seem appropriate
to stop tracking those problems altogether. For example, a set of Mac OS only problems
I was tempted to salvage by putting into a wiki list, but then why not just keep
tracking them as bugs?

We will incur the problem where some bugs will actually become stale, and we'll have
to check that they are still up-to-date. It seems better to still allow a long tail
of a bug queue than to try and track those problems in another manner.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[x] Configuration Change
[ ] Bug fix
[ ] Other:

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -7,6 +7,7 @@ exemptLabels:
   - "Release Blocker"
   - "Impact: Bad Game Rules"
   - "Impact: Game Crashing"
+  - "Impact: Incorrect UI Behavior"
 # Label to use when marking an issue as stale
 staleLabel: 'Stale'
 


DanVanAtta:
We should probably think about this one. "Problem" was introduced as a label so we don't have arguments over if something is a bug or not. Bad game rules can be considered a bug, we don't want to have issues that have many redundant labels.
It's also possible to just comment on an issue, eg: "Is this still a problem?" to restart the timer. On one hand that could be a good way to prune the lower priority issues that are dropping away, but on the other it might not be the best process.

For consideration, do we want to rename the "Impact.*" rules to be "Impact: Incorrect Behavior", or add a new label "Impact: Incorrect UI Behavior"

DanVanAtta:
I think I convinced myself that we should use a specific "Impact-Type" label to represent this.

I updated the brand-new "bug" label to be "Impact: Incorrect UI Behavior" (https://github.com/triplea-game/triplea/labels/Impact%3A%20Incorrect%20UI%20Behavior)

PR #5234, opened 2019-09-17
Add territory effect movement cost

ron-murhammer:
- Implement feature request: https://forums.triplea-game.org/topic/1183/terrain-effects-for-movement
- Adds new XML TerritoryEffectAttachment option "movementCostModifier" which allows map makers to set movement costs
- Adds new XML property "Enter Territories With Higher Movement Costs Then Remaining Moves"
- Update all routing logic to handle movement costs and refactor code base to handle them properly
- Update AI to handle movement costs
- Various bug fixes related to code I was already touching

## Functional Changes

[ ] New map or map update
[X] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[X] Manually tested
[ ] No testing done

I'm mostly using TWW and updating its territoryEffectAttachments for manual testing. Also testing lots of regular maps to ensure this didn't cause any unintended side effects or major performance issues.

**XML**
```
        <attachment name="territoryEffectAttachment" attachTo="Mountain" javaClass="TerritoryEffectAttachment" type="territoryEffect">
            <option name="movementCostModifier" value="germanStrategicBomber" count="5"/>
        </attachment>
```

```
        <property name="Enter Territories With Higher Movement Costs Then Remaining Moves" value="true" editable="false">
            <boolean/>
        </property>
```

**Shows route finding avoiding high cost territory**
image

**Showing that a strat bomber that has 6 movement can't enter Austria as it would cost: 1 first move + 6 (1 base + 5 mountain) = 7.**
image


game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
```diff
@@ -67,6 +68,16 @@ protected static int getInt(final String value) {
     }
   }
 
+  /** Throws an error if format is invalid. */
+  protected static BigDecimal getBigDecimal(final String value) {
```
DanVanAtta: This tool seems to be in an odd place, and it's only used by one child class. What do you think about making this a private static utility function or moving it to a utils class? `java-extras` would be a good hope, perhaps something like `BigDecimalFactory`

ron-murhammer:
So this is a pretty massive PR that touches a lot of core functionality of the engine around movement and routes. I've tried to do it all in one go for the most part except for a few TODOs and ignoring scrambling (which never really fully worked for distance > 1 anyways). I also tried to take this opportunity to refactor a lot of the route methods to improve the API and avoid multiple ways of doing the same thing.

I still have a few additional things that I'd like to include in this but figured I put it up since its functional and probably about 80-90% done. Here are remaining items:
- Add XML property "Units Can Enter Territories With Higher Movement Costs Then Remaining Moves"
- More manual testing to ensure no issues across a variety of maps
- Consider more unit testing
- Try to properly implement `getNeighborsByMovementCost`

game-core/src/main/java/games/strategy/engine/data/GameMap.java
```diff
@@ -271,80 +272,110 @@ public Territory getTerritory(final String s) {
...
   /**
-   * Returns the shortest route between two territories or null if no route exists.
+   * Returns all neighbors within a certain distance of the starting territory that match the
+   * condition. Does NOT include the original/starting territory in the returned Set.
    *
-   * @param t1 start territory of the route
-   * @param t2 end territory of the route
+   * <p>TODO: update to properly consider movement cost not just distance
+   *
+   * @param territory referring territory
```
DanVanAtta: nit: not sure what is meant by 'referring' territory, perhaps "reference territory"?

game-core/src/main/java/games/strategy/engine/data/GameMap.java
```diff
@@ -271,80 +272,110 @@ public Territory getTerritory(final String s) {
...
+   * @param movementLeft maximal distance of the neighboring territories
    */
-  public Route getRoute(final Territory t1, final Territory t2) {
-    return getRoute(t1, t2, Matches.territoryIsLandOrWater());
+  public Set<Territory> getNeighborsByMovementCost(
+      final Territory territory, final Unit unit, final BigDecimal movementLeft) {
```
DanVanAtta: Why BigDecimal for movement left and not double?

game-core/src/main/java/games/strategy/engine/data/GameMap.java
```diff
@@ -271,80 +272,110 @@ public Territory getTerritory(final String s) {
...
+    Preconditions.checkArgument(
```
DanVanAtta: Good to see the precondition check, given that `Route` can have null territories, I wonder if it would be more robust include another precondition check for non-null territory and non-null unit.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
```diff
@@ -271,80 +272,110 @@ public Territory getTerritory(final String s) {
...
+        movementLeft.compareTo(BigDecimal.ZERO) >= 0,
+        "MovementLeft must be non-negative: " + movementLeft);
+    if (movementLeft.compareTo(BigDecimal.ZERO) == 0) {
+      return Collections.emptySet();
+    }
+    final Set<Territory> start = getNeighbors(territory);
```
DanVanAtta: Is the set of territories truly the `start`? Should the variable be named `neighbors`?

game-core/src/main/java/games/strategy/engine/data/GameMap.java
```diff
@@ -271,80 +272,110 @@ public Territory getTerritory(final String s) {
...
+    if (movementLeft.compareTo(BigDecimal.ONE) <= 0) {
```
DanVanAtta: Is this an impossible condition? It was asserted movement left is non-negative at the start of the method, and the previous if statement checks for zero movement left, which should leave no case for zero or less. Is there something I'm overlooking?

The logic seems complex enough that it would be a good candidate for test. That would help ensure no bugs and give us confidence for any future updates that we are not breaking the method and save a lot to understand every case. `GameMap` perhaps is also not the best home for every calculation relating to game map, to test this one would need to control the `getNeighbors` function, and it would be straight forward. To do that, a utility class with a passed in function to calculate neighbors could be mocked or stubbed, making testing pretty straight forward.

game-core/src/main/java/games/strategy/engine/data/RouteFinder.java
```diff
@@ -66,19 +65,65 @@
     return Optional.empty();
   }
 
+  Optional<Route> findRouteByCost(final Territory start, final Territory end) {
```
ron-murhammer: This is the core and most important method of the PR. I'd appreciate focused review here and anything that can help with performance.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
```diff
@@ -271,80 +272,110 @@ public Territory getTerritory(final String s) {
...
-   * @param cond condition that covered territories of the route must match
+   * <p>TODO: update to properly consider movement cost not just distance
    */
-  @Nullable
-  public Route getRoute(final Territory t1, final Territory t2, final Predicate<Territory> cond) {
-    checkNotNull(t1);
-    checkNotNull(t2);
-
-    return new RouteFinder(this, cond).findRoute(t1, t2).orElse(null);
+  public Set<Territory> getNeighborsByMovementCost(
```
DanVanAtta: This method looks nearly identical to the one above. Can the above method call this method?
Maybe something like:
```
getNeighborsByMovementCost(territory, unit, movementleft, () -> true);
```

game-core/src/main/java/games/strategy/engine/data/GameMap.java
```diff
@@ -271,80 +272,110 @@ public Territory getTerritory(final String s) {
...
+   * movement costs per territory, it takes the max movement cost for each territory of all the
+   * units when comparing movement costs across different routes.
    *
-   * @param start start territory of the route
-   * @param end end territory of the route
-   * @param matches Map of territory matches for covered territories
-   * @return a composite route between two territories
+   * @param cond condition that covered territories of the route must match
+   * @param units checked against canals and for movement costs
+   * @param player player to check canal ownership
```
DanVanAtta: If this method is specific to canal, should that be baked into the method name? eg, something like: `getRouteForUnitsOverCanal`

game-core/src/main/java/games/strategy/triplea/attachments/TerritoryEffectAttachment.java
```diff
@@ -98,6 +101,36 @@ public int getCombatEffect(final UnitType type, final boolean defending) {
     return defending ? combatDefenseEffect.getInt(type) : combatOffenseEffect.getInt(type);
   }
 
+  private void setMovementCostModifier(final String value) throws GameParseException {
```
DanVanAtta: It would be good to see this tested, at least as an integration test that the parsing is good. Perhaps add an example attachment to an XML or see about decomposing the logic so that this can be an isolated unit. Another alternative is to mark this as `@VisibleForTesting` and have the method return value so that it can be tested directly.

game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
ron-murhammer: Actually many of the methods of this class are helpers for reading XML strings and converting them to the appropriate types.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
ron-murhammer: Yeah, those param java docs aren't really useful. Removed them.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
ron-murhammer: I want to avoid precision issues. Movement is kind of like a currency so fits better as a BigDecimal type.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
ron-murhammer: Added a check for unit. Territory is actually already checked in getNeighbors.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
ron-murhammer: Renamed.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
ron-murhammer: This is comparing to 1. So movementLeft could be say 0.5.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
ron-murhammer: Combined.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
ron-murhammer: It used to but at this point its more than that. Its essentially the route method if you know the units you want to move as it checks canals and movement costs.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
DanVanAtta: Are there cases where you saw precision problems? Will future engineers think this is over-engineered?

game-core/src/main/java/games/strategy/engine/data/GameMap.java
DanVanAtta: Public APIs tend to be treated a bit special as they can be invoked from new locations and one cannot generally assume that you know all the caller paths into the method. Private methods are a bit of an exception because if you raise the scope of a method, then it is a buyer-beware situation.

It's one benefit even to reducing scope, you can be less stringent about validation and asserting state. If we don't get a NPE in very short order, 'fail-early' principle says to fail as soon as an error condition could be detected, which is either the caller of this method or as soon as the method starts executing.

I guess in other ways, it doesn't hurt, does it?

game-core/src/main/java/games/strategy/engine/data/GameMap.java
DanVanAtta: Ah, I see. It's just a bit difficult to grok: `movementLeft.compareTo(BigDecimal.ONE) <= 0`
Takes more effort to understand than would be really desirable. If not resorting to using doubles, then extracting the condition to a boolean variable so it is named could reduce the overhead to understand the condition.

game-core/src/main/java/games/strategy/engine/data/RouteFinder.java
DanVanAtta: As a first observation, there could be more abstraction or decomposition. Just noting we've maps and number types and not too much that is higher order.

For example, I suspect `final Map<Territory, BigDecimal> routeCosts = new HashMap<>();` could be replaced with a `CostCalculator` and then hide a number of details and be more easily tested on its own.

You could have methods like "addCost", or "compareTerritory", or "isBetterRoute", seems like a number of the logic conditionals in this method could be folded into it.

re: performance, @ron-murhammer Do you know the method call frequency or current runtime?

game-core/src/main/java/games/strategy/engine/data/GameMap.java
DanVanAtta: Could be worth creating a proper abstraction, just like java `Money` class, eg: `MovementPoints`

The below reads well:
```
public Set<Territory> getNeighborsByMovementCost(
      final Territory territory, final Unit unit, final MovementPoints movementLeft)
```

At that rate the internal data representation would be encapsulated and we could change it more easily and/or more easily verify the needed precision. The encapsulation really would help the verification as we'd know all access to/from the data and so we could verify the precision requirements and refactor if needed/desired.

There could be potential for some natural data operations and help avoid the 'compareTo' method of BigDecimal

game-core/src/main/java/games/strategy/triplea/delegate/TerritoryEffectHelper.java
```diff
@@ -81,4 +84,36 @@ private static boolean unitTypeKeepsBlitz(final UnitType type, final Territory l
...
+  /**
+   * Finds movement cost for each unit by adding 1 plus any territory effects and then returns the
+   * max movement across all units. If no territory effects then just returns the base cost of 1.
+   */
+  public static BigDecimal getMaxMovementCost(final Territory t, final Collection<Unit> units) {
+    if (getEffects(t).isEmpty()) {
+      return BigDecimal.ONE;
+    }
+    BigDecimal max = new BigDecimal(Integer.MIN_VALUE);
+    for (final Unit unit : units) {
```
DanVanAtta: It would be interesting if over a large unit stack if there are performance savings to skip computing movement cost of the same unit type with the same movement left. I think we'd first need to know the runtime of this method with 100 or 200 units first, then it might be worth optimizing if it's a bit slower. Before any peformance optimization, or refactoring, the method should get a test so it would be safe to change.

game-core/src/test/java/games/strategy/triplea/delegate/WW2V3Year41Test.java
```diff
@@ -1368,7 +1366,7 @@ void testAirCanLandWithAlliedFighters() {
     addTo(sz40, carrier(gameData).create(1, germans));
     addTo(sz40, fighter(gameData).create(1, italians(gameData)));
     addTo(madagascar, fighter(gameData).create(2, germans));
-    final Route route = gameData.getMap().getRoute(madagascar, sz40);
+    final Route route = gameData.getMap().getRoute(madagascar, sz40, Matches.always());
```
DanVanAtta: What was the motivation for having an 'always match' parameter? Seems like this is a case for method overloading and having a simplified API that only takes the two parameters, let the class handle under-the-covers that the third parameter is defaulted to a tautology.

RoiEXLab:
@DanVanAtta Sorry, I'm currently ill, so I won't be able to review non-trivial changes the next couple of days 😷

DanVanAtta:
I was there recently, rest up @RoiEXLab , get better soon!

game-core/src/main/java/games/strategy/engine/data/RouteFinder.java
ron-murhammer: This is part of route finding so every time you highlight units and move the cursor around. @RoiEXLab did some performance testing when he rewrote the route finding algorithm. I haven't run any numbers yet just manual testing on maps so far.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
ron-murhammer: So the double precision problems are fairly common. A very simple example is if you say would want to have road territories which have a 0.2 movement cost (so 1 move unit can move 5 territories). So a simple test that shows the problem would be like this:
```
  void testDouble() {
    final double d1 = 1;
    final double d2 = 0.2;
    final double result = d1 - d2 - d2 - d2 - d2 - d2;
    assertEquals(result, 0);
  }
```

You'll see that it fails as doubles can't handle the precision properly. So we need to do 1 of 3 options of which I think BigDecimal is probably the best:
1. Use BigDecimal
2. Use double/float but round every time you do any comparison of values (painful and very open to developers forgetting to and having edge condition bugs)
3. Create some sort of BigDecimal-like class like Money/MovementPoints which from what we need functionality wise would end up essentially being BigDecimal

So I decided to just go with BigDecimal though option 3 (some wrapper around most likely BigDecimal) could be used. Just didn't see enough value in introducing that complexity.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
ron-murhammer: I mean I think you can go too far with Preconditions and add them to every method for every argument. Its just a lot of boiler plate code that really if you are already checking it in another place is just doing nothing. There is definitely value in many cases of failing fast and ensuring things like NPEs don't happen deep inside methods. But there are also logical boundaries to how things are used. There are a lot of similar methods in this call due to offering different parameter combinations that then use a private method to check the territory isn't null.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
ron-murhammer: Given how often comparisons are used, its probably overkill as we are probably talking about 100 condition extraction for a very simple comparison.

game-core/src/main/java/games/strategy/triplea/delegate/TerritoryEffectHelper.java
ron-murhammer: Yeah, I actually had thought the same thing. I decided at least for now, that it wasn't probably a major performance difference so would wait to consider it if performance seems slow.

game-core/src/main/java/games/strategy/engine/data/RouteFinder.java
RoiEXLab: Hmm the code doesn't really look like a common Dijkstra algorithm, which may indicate the performance isn't optimal.
Might be worth going for A* even if the performance is bad

game-core/src/test/java/games/strategy/triplea/delegate/WW2V3Year41Test.java
ron-murhammer: Ah, I actually figured you'd enjoy this change. The reason is that the method was mostly only used in test code and its generally a bad method to use. You almost always should have some condition for the route you want to find based on the units that you expect to move. Most of these tests really should have a condition and are technically passing either because it isn't checking the route closely or the only route between them happens to meet the requirements. In this example, it should really have a condition that air units can actually move into the territories along the route.

game-core/src/main/java/games/strategy/engine/data/RouteFinder.java
ron-murhammer: @RoiEXLab Yeah, its more of an adaption to the existing route finding. My guess is its not optimal but performance seems ok. I figure we can also always come back and refactor it since its very contained. Open to help here :)

game-core/src/main/java/games/strategy/engine/data/GameMap.java
DanVanAtta: Perhaps better safe than sorry - ok. Though I'm concerned at lack of abstraction, have you put any thought to wrapping the internal data type? The comparisons of BigDecimal seem a bit 'raw' or 'low-level'. For example my misunderstanding of the other if/else chain is a bad sign. I've always thought the idea of a code sitting down with a fine whiskey by a fireplace to study 10 lines of code for an evening is an anti-pattern, not developers being smart (in fact, the 'smart' developer I think is just a myth, there's lots of code to go through, if it does not take little amounts of time to understand, it does not scale).

game-core/src/main/java/games/strategy/engine/data/GameMap.java
DanVanAtta: Hmm, I still think that bakes in assumptions about how the public method is used. The limit is validity of all arguments passed in. NPE checks are sometimes not needed as you'll get a NullPointer in the first few lines anyways. If a value is unexpectedly null and then is propagated deeper into the system, it's a failure to detect an error condition. I don't think adding one extra line of check is going to really hurt or make this harder to understand.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
DanVanAtta: Does this PR really contain around 100 condition checks? From what I've seen it's about a  dozen. 

To be direct, this is hard to understand:
```
if (movementLeft.compareTo(BigDecimal.ONE) <= 0) {
```
It's not impossible, nor particularly difficult, but it's not KISS

game-core/src/main/java/games/strategy/triplea/delegate/TerritoryEffectHelper.java
DanVanAtta: I agree to not do premature optimization, at the same time getting some quick perf measurements should not take too much effort. PerfTimer is pretty fast to use, have you experimented with it yet?

DanVanAtta:
@ron-murhammer I took another review pass, not as a criticism, but observation, my impression is the existing code and new code are all relatively imperative and low level. I think some key abstractions could help raise it up and make things easier to understand, essentially a few refactoring passes to add more structure and abstractions. Perhaps such an effort could boost understandability in a significant way. Otherwise, some of these changes are pretty complex to review, will still need some more review iterations.

game-core/src/main/java/games/strategy/triplea/attachments/TerritoryEffectAttachment.java
```diff
@@ -98,6 +101,36 @@ public int getCombatEffect(final UnitType type, final boolean defending) {
...
+    final String[] s = splitOnColon(value);
+    if (s.length < 2) {
+      throw new GameParseException(
+          "movementCostModifier must have a count and at least one unitType" + thisErrorMsg());
```
DanVanAtta: An example of a valid syntax would perhaps make this exception message far more useful and easier to resolve if it were to happen. Otherwise I notice that "must have a count" might not be clear.

game-core/src/main/java/games/strategy/triplea/attachments/TerritoryEffectAttachment.java
```diff
@@ -98,6 +101,36 @@ public int getCombatEffect(final UnitType type, final boolean defending) {
...
+    }
+    final Iterator<String> iter = Arrays.asList(s).iterator();
+    final BigDecimal effect = getBigDecimal(iter.next());
+    while (iter.hasNext()) {
+      final String unitTypeToProduce = iter.next();
+      final UnitType ut = getData().getUnitTypeList().getUnitType(unitTypeToProduce);
+      if (ut == null) {
+        throw new GameParseException("No unit called: " + unitTypeToProduce + thisErrorMsg());
```
DanVanAtta: Printing out the unit type list could make this exception message more helpful. For example if there is a small typo and it's not easier to tell what it is, having the list of valid values could make it more clear where/what went wrong.

game-core/src/main/java/games/strategy/engine/data/RouteFinder.java
ron-murhammer: Agree. I was waiting to get rid of the old implementation until after I was finished and tested performance. Performance comparison was pretty negligible for route finding so I've combined them now.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
ron-murhammer: Updated it a little bit. It reads fine to me so if you have further suggestions let me know as sometimes the author is just too close to it.

game-core/src/main/java/games/strategy/triplea/attachments/TerritoryEffectAttachment.java
ron-murhammer: This is tested indirectly by some of the unit tests that were added. I kind of go back and forth on unit testing all these XML parsing methods as there are a lot of them and most are fairly simple. For now, I'm gonna consider it out of scope given the rest of the methods in this class aren't directly unit tested either.

game-core/src/main/java/games/strategy/triplea/delegate/TerritoryEffectHelper.java
ron-murhammer: So I ended up trying to use PerfTimer for some route finding tests but couldn't figure out how it works. Where does the output go? I was assuming the in game console window but didn't seem to go there. The javadocs for it seem out of date and it references a class I couldn't even find.

game-core/src/main/java/games/strategy/engine/data/Route.java
```diff
@@ -139,43 +141,42 @@ public Territory getStart() {
...
+      movementCost = movementCost.add(TerritoryEffectHelper.getMovementCost(t, unit));
+    }
+    return movementCost;
+  }
+
+  public BigDecimal getMovementCostIgnoreEnd(final Unit unit) {
+    BigDecimal movementCost = BigDecimal.ZERO;
+    final List<Territory> territories =
+        steps.size() > 0 ? steps.subList(0, steps.size() - 1) : steps;
+    for (final Territory t : territories) {
```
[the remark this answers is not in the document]
ron-murhammer: Updated.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
```diff
@@ -271,80 +251,95 @@ public Territory getTerritory(final String s) {
...
-    return getRoute(start, end, Matches.territoryIs(end).or(match));
+  /**
+   * Returns the shortest route between two territories so that covered territories except the start
+   * and end match the condition or null if no route exists. Doesn't pass in any units so doesn't
+   * check canals or movement costs.
+   *
+   * @param cond condition that covered territories of the route must match
+   */
+  @Nullable
+  public Route getRoute(
```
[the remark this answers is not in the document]
ron-murhammer: As its structured now, RouteFinder is only used by GameMap. All other classes call into the various getRouteXXX methods in GameMap. Whether that is ideal could be discussed but gonna say that is out of scope for this.

game-core/src/main/java/games/strategy/engine/data/Route.java
```diff
@@ -21,45 +23,53 @@
...
-    for (final Territory t : route) {
+    for (final Territory t : territories) {
       add(t);
     }
   }
 
+  /** Set the start of this route. */
+  private void setStart(final Territory newStartTerritory) {
+    checkNotNull(newStartTerritory);
+    start = newStartTerritory;
```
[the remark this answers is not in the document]
ron-murhammer: Agree. There used to be more constructors so it was more difficult. Updated.

game-core/src/main/java/games/strategy/triplea/attachments/TerritoryEffectAttachment.java
ron-murhammer: Count is a pretty common term in the game XML. This message aligns with the rest of the messages in this class so I think its better to leave it.

game-core/src/main/java/games/strategy/triplea/attachments/TerritoryEffectAttachment.java
ron-murhammer: There are some maps with way too many unit types to display them all. For example, I think TWW has probably about 100.

game-core/src/main/java/games/strategy/triplea/delegate/TerritoryEffectHelper.java
DanVanAtta: It does an info logging, which is routed to console. It might be that 'debug' logging needs to be turned on in console.

ron-murhammer: I tried switching it to debug as well.

game-core/src/test/java/games/strategy/engine/data/RouteFinderTest.java
```diff
@@ -97,7 +102,78 @@ void testNoRouteOnInvalidGraph() {
...
-        routeFinder.findRoute(territories.get(0), territories.get(territories.size() - 1));
+        routeFinder.findRouteByDistance(
+            territories.get(0), territories.get(territories.size() - 1));
+    assertFalse(optRoute.isPresent());
+  }
+
+  @Test
+  void testFindRouteByCost() {
+    final RouteFinder routeFinder = new RouteFinder(map, t -> true, new ArrayList<>(), player);
+    final Optional<Route> optRoute =
```
DanVanAtta: Bit of nit, OptionalMatchers test-utility could be used to clean this up a good bit, eg:

```
final RouteFinder routeFinder = new RouteFinder(map, t -> true, new ArrayList<>(), player);

 final Optional<Route> result =
        routeFinder.findRouteByCost(territories.get(0), territories.get(territories.size() - 1));

final Route expectedResult = new Route(Stream.of(0, 3, 6, 8).map(territories::get).collect(Collectors.toList()));
assertThat(result, isPresentAndIs(expectedResult));
```

The 'assertThat' is also a good place to add a commentary string of why we have the expected results, why are territories 0, 3, 6, and 8 the expected territories? It could be reverse engineered, but that level of effort does not scale well (consider a scenario where this is just one of 50 failing tests, knowing why a test is supposed to pass is a best-practice. 

Of some interest, there were even a few master's papers written about testing that looked specifically at TripleA (we did not score very well):

of note, this paragraph: http://www.cs.loyola.edu/~binkley/papers/icsm-12-smells.pdf
> (Page 1) The  second  study  is  a controlled  experiment  involving  twenty  master  students.  Itis  aimed  at  analyzing  whether  the  presence  of  test  smellsaffects  the  comprehension  of  source  code  during  software maintenance.  In  the  study  we  asked  subjects  to  perform different program comprehension tasks and we measured thesubjects’  performance  using  both  correctness  and  the  timespent to perform a task.

> Collected  data  from  the  first  study  shows  that  there  is  a high  diffusion  of  the  test  smells  in  both  open  source  and industrial  software  systems.  In  addition,  the  second  studyprovides  evidence  that  test  smells  have  a  strong  negativeimpact on program comprehension and maintenance.

The other interesting paper that included TripleA if you have not come across it yet: https://testroots.org/assets/papers/athanasiouTSE.pdf

game-core/src/main/java/games/strategy/performance/PerfTimer.java
```diff
@@ -7,7 +7,7 @@
  * Provides a high level API to the game engine for performance measurements. This class handles the
  * library details and sends output to 'PerformanceConsole.java' <br>
```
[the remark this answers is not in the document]
ron-murhammer: Yeah, decided to fix the 1 character java doc typo and initially didn't realize there were other issues. But yeah a separate PR to update and explain a bit more on its usage would be good.

game-core/src/test/java/games/strategy/triplea/delegate/TerritoryEffectHelperTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+  @Test
+  void testGetMaxMovementCostZero() throws Exception {
+    final GameData twwGameData = TestMapGameData.TWW.getGameData();
+    final PlayerId germans = GameDataTestUtil.germany(twwGameData);
+    final Territory sicily = territory("Sicily", twwGameData);
+    final BigDecimal result =
+        TerritoryEffectHelper.getMaxMovementCost(
+            sicily, GameDataTestUtil.unitType("germanInfantry", twwGameData).create(1, germans));
+    assertTrue(result.compareTo(BigDecimal.ZERO) == 0);
```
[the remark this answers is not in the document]
ron-murhammer: Agree. Updated.

game-core/src/test/java/games/strategy/engine/data/RouteFinderTest.java
ron-murhammer: Agree that they could be better. I'm gonna skip updating this one since there are already a bunch of existing unit tests in this class and it follows that same pattern. A separate PR to improve the unit tests in this class would be good.

DanVanAtta: Okay, but let's be careful of a "The bureaucracy is expanding to meet the needs of the expanding bureaucracy"

PR #5237, opened 2019-09-18
Add PostConditions utility

DanVanAtta:
To avoid oddity of using "PreConditions" for "PostConditions", introduce utility class "PostConditions"
See javadoc of added class for additional details, useful as 'always-on' assertions.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[x] Other:  Add code Utility 

## Testing

[ ] Manually tested
[ ] No testing done

## Additional Review Notes

Not yet used, to be used in upcoming PRs

### DIFF
--- a/java-extras/src/main/java/org/triplea/java/PostConditions.java
+++ b/java-extras/src/main/java/org/triplea/java/PostConditions.java
@@ -7,7 +7,7 @@ import lombok.NoArgsConstructor;
  * Class to complement guava "PreConditions" and assert "post" state, for example verifying the
  * state of a return value when a method exits. Java assertions normally would serve this function,
  * but they are not enabled by default. This class serves as an 'always-on' alternative to assertion
- * and provides room for additional functionality such as error messages when asserted states fail.
+ * and provides room for nicer APIs and additional functionality when asserted states fail.
  * <br>
  * Example Usage:
  *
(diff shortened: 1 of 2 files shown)


DanVanAtta:
With luck the javadoc describes why assert is not used:
> This class serves as an 'always-on' alternative to assertion
 * and provides room for additional functionality such as error messages when asserted states fail.

I'm personally not concerned about performance for relatively simple assertions. I think there is some long history to that concern and the performance concern is perhaps legacy - computers are just blindingly fast these days.

As an example of some places to use this, DB updates that we expect to always succeed. The DB udpate returns the number of rows updated, so we could return a boolean and check that the row count is '1'. The caller often cannot really do anything with that, it throwing an exception based on that condition is a bit late and violates 'fail-early' principle (we could have thrown the exception sooner). So basically the postcondition is to allow that. Otherwise the usage is hopefully going to be for documentation and to allow for more rigorous constraints and additional programmatic verification of correctness.

RoiEXLab:
The `assert` keyword does support error messages. I always get the syntax wrong though.
IIRC its `assert true : "Message"` or something

DanVanAtta:
I did not know that they supported messages, updated javadoc: df512eeca
I think that the syntax is not obvious/easy is perhaps more reason. Though at end of day, the assertions not being on in production is actually a bit the problem. Really we want them for local testing and prerelease, but in general 'always-on' is the goal; on during test-only is not worth much, adding additional config to turn them on seems like something we can miss or just extra complication, particularly compared to the additional benefits.

RoiEXLab:
👍

PR #5238, opened 2019-09-18
Refactor RoleAuthorizer

DanVanAtta:
- Refactor RoleAuthorizer logic, flip the way we check whether to grant access, define rules in terms of the users given role rather than defining the rules relative to the requestsed role.

- Avoid using 'isAdmin' and 'isModerator' methods in AuthenticatedUser, push knowledge of roles more to RoleAuthorizer and allow for easier, more comprehensive tests to be written against RoleAuthorizer.

- Add utility method 'getUserIdOrThrow' to assert for a non-null user id

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[ ] Manually tested
[ ] No testing done


http-server/src/main/java/org/triplea/server/access/AuthenticatedUser.java
```diff
@@ -22,11 +24,12 @@ public String getName() {
...
-
-  public boolean isModerator() {
-    return userRole.equals(UserRole.MODERATOR);
+  public int getUserIdOrThrow() {
+    Preconditions.checkState(
+        !userRole.equals(UserRole.ANONYMOUS),
+        "Anonymous user is expected to have a null user id, "
+            + "it is the only case expected where user id would be null");
+    return Optional.ofNullable(userId)
+        .orElseThrow(() -> new AssertionError("Expected to have a non-null user id"));
```
ron-murhammer: Not really understanding this method. It throws an error if the user role is anonymous or if the user has a null user id? Also maybe I'm just not used to seeing it written this way but seems a bit strange to create an optional of an Integer to then just immediately check if its null? Couldn't you just directly check if `userId == null`?

Also given this change, do any of the code paths that now call this method properly handle the expections or if this happens does it kill everything?

http-server/src/test/java/org/triplea/server/access/RoleAuthorizerTest.java
```diff
@@ -18,45 +17,60 @@
...
+    when(authenticatedUser.getUserRole())
+        .thenReturn(UserRole.ADMIN)
+        .thenReturn(UserRole.MODERATOR)
+        .thenReturn(UserRole.PLAYER)
+        .thenReturn(UserRole.ANONYMOUS);
+
+    assertThat(roleAuthorizer.authorize(authenticatedUser, UserRole.ANONYMOUS), is(true));
+    assertThat(roleAuthorizer.authorize(authenticatedUser, UserRole.ANONYMOUS), is(true));
+    assertThat(roleAuthorizer.authorize(authenticatedUser, UserRole.ANONYMOUS), is(true));
+    assertThat(roleAuthorizer.authorize(authenticatedUser, UserRole.ANONYMOUS), is(true));
```
ron-murhammer: These unit tests seem backwards to me? Shouldn't an admin return true for all and anonymous only return true for anonymous role?

http-server/src/test/java/org/triplea/server/moderator/toolbox/moderators/ModeratorsControllerTest.java
```diff
@@ -52,7 +53,7 @@ void getModerators() {
 
   @Test
   void isSuperModPositiveCase() {
-    when(authenticatedUser.isAdmin()).thenReturn(true);
+    when(authenticatedUser.getUserRole()).thenReturn(UserRole.ADMIN);
```
ron-murhammer: Are "admin" and "super mod" the same thing?

http-server/src/main/java/org/triplea/server/access/AuthenticatedUser.java
DanVanAtta: > if the user role is anonymous or if the user has a null user id? 

Yes. We expect user id to be null only with anonymous accounts. If the user id were null, it's indirect to say "the user id was null", instead we can say "this user anonymous, they do have a user id"

To avoid predicting that we will always have the 4 roles, and in case we do not update this code, we check both situations. It can also be thought of as this way, we have a validity check, "user id != null", and the precondition check is documentation.

> Also maybe I'm just not used to seeing it written this way but seems a bit strange to create an optional of an Integer to then just immediately check if its null? Couldn't you just directly check if userId == null?

```
Optional.ofNullable(userId).orElseThrow(..)
```
Is the same as:
```
if(userId == null) {
  throw ...
} else {
 return userId;
}
```
I suppose the difference is imperative vs functional, one says what to do, the other says how to do it.

> Also given this change, do any of the code paths that now call this method properly handle the expections or if this happens does it kill everything?

Kills everything by design. The code paths where we could get a null user id would NPE out anyways. WE could put NPE guards everywhere, but in each case we want the code to terminate anyways. Having no user-id when we expect one is effectively a security breach.

DanVanAtta: Let me know if that properly answers your questions @ron-murhammer

http-server/src/test/java/org/triplea/server/access/RoleAuthorizerTest.java
DanVanAtta: This unit test is basically saying:

Can an admin authenticate as anonymous? yes
Can a moderator authenticate as anonymous ? yes
etc..

It's not quite the API I would have chosen for this, but there is a constraint to work within the framework. Creating a delegate object to handle authorization I think would add too much indirection.

http-server/src/test/java/org/triplea/server/moderator/toolbox/moderators/ModeratorsControllerTest.java
DanVanAtta: Yes; alignment of terms would be nice, but it's not in scope of this refactor. With the DB now updated to have roles, we can use ADMIN as a term.  Before we had an 'isAdmin' column in DB that meant moderator. Now that column is 'role', so we can use 'admin' without confusing 'admin' with 'moderator'. 'SuperMod' is only in the toolbox, which is still beta and we have not completely baked in the term 'super mod' just yet.

http-server/src/main/java/org/triplea/server/access/AuthenticatedUser.java
ron-murhammer: Ok. I think my confusion is maybe the precondition message: "Anonymous user is expected to have a null user id, it is the only case expected where user id would be null". That reads kind of odd to me and not sure what its trying to tell me. If I saw that in say the logs after some sort of issue, I don't think I would understand what happened. Is there a better way to phrase that?

The other thing that strikes me as a little inconsistent is one of the 2 checks is done in a precondition while the other is then done as part of the return statement. Would it be more clear to check both as preconditions? Essentially more explicitly saying this method requires the user to not be anonymous and not have a null user ID?

http-server/src/test/java/org/triplea/server/access/RoleAuthorizerTest.java
ron-murhammer: Ah, I see now. I read the test methods in the opposite way where I thought it was saying let's check admin user against the 4 roles rather than let's check each user type against admin role. I think maybe updating the unit test names to "authenticateAsAdmin" instead of "authenticateAdmin" would make that more clear.

http-server/src/test/java/org/triplea/server/moderator/toolbox/moderators/ModeratorsControllerTest.java
ron-murhammer: Ok. I'm fine leaving it out of scope for now. It would be good long term for this overall project to get to pretty concise and consistent terminology so its a bit easier to understand.

DanVanAtta: Generally agree; I think part of the problem is we don't have a domain term yet for 'admin', the role simply did not really exist. We could call it 'council member', but that does not capture the role well.

There are some glaring naming problems that do not match the domain:
- a player is called a node instead of as a player
- a 'game' refers to an XML when that is more a map. We have a 'select' map option, not a 'select game' option. Which 'maps' do you like, refers to specific XML variants, not to a set of XMLs

With that said, w.r.t ADMIN, does that seem reasonable? I think 'moderator' and 'player' are pretty clear. Anonymous is roughly well understood as opposed to registered or a 'player'. 'Super-mod' is something I made up when doing toolbox 2.0, largely to not collide with the 'isAdmin' DB column.

http-server/src/test/java/org/triplea/server/access/RoleAuthorizerTest.java
DanVanAtta: Good suggestion, I'll take a look at the test method names and see if they can be improved.

prastle:
I would sug (not that I know anything about code) 4 levels.  Anon, Player = registered, Moderator, Admin or Dev? or maybe 5 levels

http-server/src/main/java/org/triplea/server/access/AuthenticatedUser.java
DanVanAtta: The precondition message should be improved, that is pretty clear.

Ideally we would never see these messages in logs. The idea is to enforce strong validity constraints and never let these error conditions get past testing, the software simply fails/blows up before we ever get to production. On the other side of the coin, it's 99% likely a developer bug/error if we ever have these bad states, so we should never really see these situations.

The precondition being documentation I think makes sense, it's better than a comment and more direct and adheres to 'fail-early' just slightly better. Checking both conditions as a precondition is a bit odd. The method checks for null and throws or returns a non-null ID, that is what the method does. Using a precondition to check that is a bit odd.

I'm avoiding the assumption a bit that the only case with a null user id is when the user is authenticated as anonymous.  We could have a bug somewhere that maybe leads to that situation. It's a bit the arrow of causality, an anonymous user implies a null user id, but a null user id might not always necessarily imply an anonymous user. Ideally it is bidirectional and basically assumes we have no bugs and are correct about enforcing this invariant. Hence the wiggle-room and why the two checks are not the same.

DanVanAtta:
REGISTERED by itself seems odd, and it's not mutually exclusive as moderators and admins are registered too. 

Admin does not mean a dev, it's someone that is able to make players moderators and/or make moderators admins. Toolbox 2.0 allows for that : )

REGISTERED_PLAYER is an option, but seems a bit wordy. 

This could work: ANONYMOUS_PLAYER, PLAYER, MODERATOR, ADMIN , but the ANONYMOUS_PLAYER again seems a bit wordy.

I"m curious if any of those really jump out for anyone.

prastle:
By registered I meant to refer to the fact that a reg player is not Anon.  Thus perhaps Anon Player Mod Admin and Dev

prastle:
As an after thought (and off topic sorry) but maybe anon cannot receive any of your future ideas of Stats and games etc or the ability to slap?

DanVanAtta:
It is currently: Anon Player Mod Admin

'Dev' currently is not granted anything special, they are one of the above roles. We need to be sure it is used somewhere to avoid unused code and unnecessary complexity.

> anon cannot receive any of your future ideas of Stats and games etc or the ability to slap?

Let's try to stay on topic, but this can be easy to answer.. Anon is not tracked in DB, so no, they would not acquire stats (stats would be a reason to become non-anon :grin: )

Removing slap is fundamental, handicapping a role just because we don't really like it, seems wrong. Slap is useful; while it can be abused, slapping is common, particularly when meeting people to start a game.

prastle:
To Ccarify I was thinking Dev as Super mod.  Admin suffices.  Also removing stats might be enough to encourage less anon all gd ty :)

DanVanAtta:
There are actually different kinds of dev, casual contributors, major contributors, and maintainers. The first submits PRs every now and then, the second submits many PRs and is active in github issue conversations, and the latter fixes other peoples bugs when it comes time to release :grin: 

Maintainers have access to DB, anyone with DB access probably may as well be an admin. But not all admins will be dev's. (in part, toolbox 2.0 allows for promotion/demotion of moderators so that it's not up to only the dev's to manage it; We want to minimize the things that only devs can do).

DanVanAtta:
@prastle I'm entertaining the naming discussion as we are on the topic, and naming is actually quite difficult to change once set, and is very important for understanding. Coding is about creating abstractions, hiding details, being able to talk in terms of modules and not bits.

For example, we take the list of strings and mutate it to get a list of ids which we then filter by the ids that are greater than zero, compared to: we take the player list and then filter for the registered users. The first is low-level and has no abstraction, the second one is readily understood and is more abstract.

So it's no bother to try and get this right.

prastle:
I guess my thought is simple.   Admin have the powers I used to have (except massive changes)  new release will clarify that and hopefully add new explanation through wiki or forum).  Only dev = ROOT

prastle:
So it is completely up to u if that makes sense or not. I was just offering my 2 cents

DanVanAtta:
Gotchya @prastle ; the role discussion is something that has consequences in code, it's not necessarily a team structure concept. We are modeling the code after the team structure. I think you're thinking more about project structure. The lobby server code has no use to know about a dev, there are no dev only endpoints. It could make sense if we created a lobby tool to put it in 'maintenance mode' maybe, that would be something triggered by dev

DanVanAtta:
Hmm, at the risk of expounding a mistake, we should try to keep PR conversation related to the exact code that is changed and specific to it. This latter conversation deserved it's own discussion thread and perhaps should be its own issue. I think we exhausted the convo, the current set of roles are about as good as we can get it seems.

prastle:
Cool np

DanVanAtta:
@ron-murhammer updated, let me know what you think.

7d7965f  - updates precondition message
30bda4610  - updates the test method names (what a difference a word makes, excellent suggestion @ron-murhammer )

Issue #3462, opened 2018-06-15
Document Deprecated Items


Cernelius:
I'd like to make a Wiki in here with the full list of all deprecated items currently supported (stuff that works but that you should never use, in new maps), likely split between game elements and skin elements. I need a developer to list me here all of them, also referring to not obvious things (for example, are jpg and jpeg for the smallMap deprecated or not?).

Example: "isInfantry".

Cernelius:
A list of things that come to my mind, in no particular order:

- Is jpg and jpeg deprecated for the smallMap or anywhere else?
- Should we add a mention about TwoIfBySea in pos2?
- Are the long attachment names deprecated too?

@ron-murhammer Are you sure you want to track (non skin) deprecates in pos2 at all? Functionally, I think, there, you could tell only the things that people are supposed to use, to keep it cleaner, and record deprecates here in the Wiki. For example, I would just change all to short attachment names, instead of having for each the long one documented as deprecate, as well. Also, I don't believe things like TowIfBySea have been tracked in there, like ever.
Also, pos2 might not have all deprecates, as some might impact both the game and the skin, and pos2 is, of course, xml only.

Cernelius:
A search with non case sensitive "deprecate" in pact_of_steel_2.xml returned only "isInfantry".

Cernelius:
Generally speaking, this ticket is meant to catch everything that has been added up after the upgrades to 1.9.0.0.3266 documented in here:
https://github.com/triplea-game/triplea/wiki/Upgrade-Maps-Information
Plus anything already deprecated that has not been documented or updated from 1.8.0.9 (example: TwoIfBySea).

ron-murhammer:
I made some POS2 XML updates. Here are the things I know are deprecated:
- isInfantry
- isParatroop
- isMechanized
- costPU
- artillerySupportable
- artillery
- unitSupportCount

Cernelius:
Thanks @ron-murhammer.

I believe you are or should be incorrect on a couple items. As you can read in the Wiki I linked, "isParatroop" and "isMechanized" used to be deprecated (actually, in the 1.8.0.9 they were already supported doing nothing; so already beyond deprecation), but are now supposed to be entirely removed, or at least this is what I believe @DanVanAtta did, and I documented.
https://github.com/triplea-game/triplea/wiki/Upgrade-Maps-Information

[...]

PR #3367, opened 2018-04-14
Short class names for delegates

RoiEXLab:
This Sounds like a straight forward change, but I'd like to hear your opinion about it, because it consists of 3 different parts that I want your opinion on.
1. Allow short hand names for delegates in `games.strategy.triplea.delegate`, because of the name conflicts for delegates in `games.strategy.twoIfbySea.delegate`, those names need to be spelled out regardless of this change (because usage of those delegates [is rare](https://github.com/search?p=1&q=org%3Atriplea-maps+twoIfBySea&type=Code&utf8=%E2%9C%93))
2. Noticing that we have a special TestDelegate type that is only being invoked by test code (or at least this is what I hope to be happening with this name). I didn't add a short version for that, and made the behaviour of attachments consistent with this, because there's also a special TestAttachment class. However this strongly hints at either a flaw with testing (test attachments should be only available for test cases and not for production code), or bad naming (why is it a **Test**Attachment then?)
3. As a consequence of 1) and based on the [feedback of this thread](https://forums.triplea-game.org/topic/595/short-attachment-names) add a warning for any maps that use twoIfBySea delegates and inform the potential map maker that those shouldn't be used. ("Will be removed in a future version of TripleA")

/cc @DanVanAtta @ron-murhammer @ssoloff

PR #5242, opened 2019-09-18
[Not for Merge] Drop TwoIfBySeaEndTurnDelegate

DanVanAtta:
Per deprecation comments, was waiting for incompatible release to drop this code.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[x] Manually tested
[ ] No testing done

- Launched a local game.
- Looked for TwoIfBySea in map repo, not found.


Cernelius:
The current prerelease is not map incompatible with the current stable.

Cernelius:
This is what @RoiEXLab said:
https://forums.triplea-game.org/topic/595/short-attachment-names/17
By "will be removed in a future version" I just wanted to strongly indicate the twoIfBySea delegates shouldn't be used to enable us to either rename it in an incompatible release, or to incorporate them into the existing delegates and to remove them from the twoIfBySea package afterwards.

game-core/src/main/java/games/strategy/engine/data/gameparser/XmlGameElementMapper.java
```diff
@@ -101,23 +101,10 @@ public XmlGameElementMapper(
         .put("TechActivationDelegate", TechActivationDelegate::new)
         .put("TechnologyDelegate", TechnologyDelegate::new)
         .put("UserActionDelegate", UserActionDelegate::new)
-        .putAll(newTwoIfBySeaDelegateFactories())
         .putAll(auxiliaryDelegateFactoriesByTypeName)
         .build();
   }
 
-  @SuppressWarnings(
-      "deprecation") // required for map compatibility; remove upon next map-incompatible release
```
DanVanAtta: @RoiEXLab would you mind clarifying this comment? What exactly can be removed? The delegate 'TwoIfBySeaEndTurnDelegate" is used in `class_variations/../four_if_by_sea.xml`

DanVanAtta:
@Cernelius can this PR be resurrected? https://github.com/triplea-maps/classic_variations/pull/3/files

If we made that update, it looks like the delegate code being removed here would become unused. The comment quoted: https://github.com/triplea-game/triplea/pull/5242#issuecomment-532683765 from @RoiEXLab is pretty confusing.

@RoiEXLab, can you clarify the state of the twoIfBySea end turn delegate, is it slated to be removed? Is there a deprecation path?

RoiEXLab:
@DanVanAtta here's the story I recall.
I believe I asked the map makers some time ago if they knew what made the twoIfBySea delegates special. IIRC no one really seemed to know, there were only a couple maps that actually used them.
So ideally those few maps were migrated to use the other delegates, so the twoIfBySea delegates can be removed from the code

DanVanAtta:
Okay, sounds like the update: https://github.com/triplea-maps/classic_variations/pull/3/files @Cernelius proposed would do that.

@Cernelius do you know if that closed PR is still valid, can we re-open it?

Cernelius:
Unless I'm missing anything, what that PR would still does is merely fully substituting the deprecated and redundant "twoIfBySea" delegate with the regular one, that I understand it is already recalling anyways. So that is no more no less valid that doing the same thing again, on a new PR. You can merge that PR if you would be doing the same thing yourself.

The problem with this, is that, apparently, the games using the "twoIfBySea" appear to be supposed to work differently than they do. I suppose this delegate was, therefore, doing something else, in the past. Currently, all those maps are substantially bugged, as they recall something that is not really anymore supported, but I'm not even sure of this. The only problem with substituting the "twoIfBySea" delegate with the regular one is hiding this matter, so that you will not know anymore what are those maps, if you'll even figure out what that delegate is supposed to do. I'm not sure converting those games like that is really better than deleting them (and deleting them may be a good choice, if they are based on something that is not anymore supported).

With a huge warning here I'm really guessing, I think that the "twoIfBySea" was a delegate for allowing cumulative placement abilities per territory. That is, if having 1 factory in a territory allows to place X, having N factories allows to place NX.

Finally, I did that only for that game, but I understand there are a few others currently using the "twoIfBySea" delegate. I don't think I care at all about (or would ever play) any of them, personally.

Cernelius:
Basically, if anyone want to bother updating all those maps to use the regular delegate, instead of the "twoIfBySea" one, it would be better to track them all in a Wiki, as likely bugged (partially unsupported) games, with a mention that they used to work with that delegate ("twoIfBySea"). Or maybe noting it somewhere in their repositories (where?)?

I don't think it makes sense to merge that PR of mine without converting all other games having "twoIfBySea" too. That PR was intended to be the first one of a number of them, covering all cases. I don't even know why I did it at all, since I don't care about any of those games; I suppose I just wanted to help @RoiEXLab.

Cernelius:
References:

This is the original main PR touching this matter:
https://github.com/triplea-game/triplea/pull/3367

Here I asked about this matter again, and nobody gave me an actual answer on the "twoIfBySea" item:
https://github.com/triplea-game/triplea/issues/3462

Here is a bug report that I opened more than 4 years ago, about two maps, that is apparently related to the "twoIfBySea" delegate (I don't like or play that map; I was just spectating it):
https://sourceforge.net/p/triplea/bugs/1102/

As I said, I'm really dubious whether it would be a better move to just delete all those games (as well as deleting the maps having only those games), instead of updating them. And I think they should be tracked as probably bugged if updated, but I don't know how.

DanVanAtta:
Appreciate the background info on this @Cernelius ; @RoiEXLab I think you're the only one that has the background context to solve the TODO comment without investigating for a few hours the full extent of the background, I'll leave it to you to solve the TODO here and drop the delegate however would be appropriate.

RoiEXLab:
@DanVanAtta Sorry to disappoint, but I really don't have a clue on how to tackle this task.
So I'm really relying on map makers as well.

PR #5245, opened 2019-09-18
Replace Error-Report rate limiting with library

DanVanAtta:
Instead of 'rolling-our-own' rate limiting, there is a library that can do this with annotations.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

I'm a bit on the fence for testing, not sure how/if we should really test this. If we trigger the limit, it could interfere with other tests as well. Perhaps manual verification is enough to know it is working, then we can trust the annotations from there.


http-server/src/main/java/org/triplea/server/http/AppConfig.java
```diff
@@ -23,29 +23,6 @@
...
-   * new API keys and will need cache values to expire. Of note, cache values are by IP address, so
-   * a single IP address expiring from cache can decrement the max failed attempt by more than one.
-   */
-  public static final long FAILED_API_KEY_CACHE_EXPIRATION = 10;
-
-  /**
-   * How many failed attempts we will allow for any given IP address before that IP address is
-   * locked out from further API key validation attempts.
-   */
-  public static final int MAX_API_KEY_FAILS_BY_IP = 4;
```
DanVanAtta: API key expiration and max fails were already unused, being cleaned up here. Max reports per day is moved.

http-server/src/main/java/org/triplea/server/error/reporting/CreateIssueStrategy.java
```diff
@@ -34,17 +32,12 @@
...
   @Override
   public ErrorUploadResponse apply(final ErrorReportRequest errorReportRequest) {
-    if (allowErrorReport.test(errorReportRequest.getClientIp())) {
-      final ErrorUploadResponse errorUploadResponse = sendRequest(errorReportRequest);
+    final ErrorUploadResponse errorUploadResponse = sendRequest(errorReportRequest);
 
-      errorReportingDao.insertHistoryRecord(errorReportRequest.getClientIp());
-      errorReportingDao.purgeOld(Instant.now().minus(2, ChronoUnit.DAYS));
+    errorReportingDao.insertHistoryRecord(errorReportRequest.getClientIp());
+    errorReportingDao.purgeOld(Instant.now().minus(365, ChronoUnit.DAYS));
```
RoiEXLab: Why this change from 2 days to a year?

DanVanAtta: The year limit is a bit arbitrary, was debating between 3 months, 6 months, 1 year, and 3 years.

Before the table was used for rate limiting, now it's only good for stats/audit. Because we no longer read the table, but append to it only, the length of the table is less significant. In other words, a large table would cause the read query to be slow, now it does not matter if it's large. It's a bit arbitrary how long to keep stats for. A year seems long enough for interesting data and data older than that is perhaps not needed and less interesting.

PR #5247, opened 2019-09-19
AI without capital shouldn't do politics

asvitkine:
AI without capital shouldn't do politics. Otherwise, it's confusing when eliminated players declare war and other random actions when they clearly are in no position to do so.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[ X] Bug fix:  
[ ] Other:   

## Testing

[ ] Manually tested
[ X] No testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/ProAi.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/ProAi.java
@@ -254,7 +254,7 @@ public class ProAi extends AbstractAi {
         } else if (stepName.endsWith("Politics")) {
           ProData.initializeSimulation(this, dataCopy, player);
           // Can only do politics if this player still owns its capital.
-          if (ProData.myCapital != null && ProData.myCapital.getOwner().equals(player)) {
+          if (ProData.myCapital == null || ProData.myCapital.getOwner().equals(player)) {
             final PoliticsDelegate politicsDelegate = DelegateFinder.politicsDelegate(dataCopy);
             politicsDelegate.setDelegateBridgeAndPlayer(bridge);
             final List<PoliticalActionAttachment> actions = politicsAi.politicalActions();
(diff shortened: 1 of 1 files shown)


Cernelius:
I'm personally more bothered by the fact that completely eliminated players cannot receive political actions (so I'm stuck eventually liberating their capitals, in case I was allied with them), but the defaults here I believe are supposed to be what happens in WWII Global...

asvitkine:
Is WWII Global an official board game or just a community map?

Can you link to the rules for it?

In terms of this patch, this is not actually changing the rule set. It's just changing the behavior of the AI. So human players can continue to do what they can do right now.

In theory, this could instead be made more sophisticated instead - for example players who've lost their capital to have logic to not declare wars since that's generally detrimental. But I thought this would be the simplest fix.

(There's a separate argument of whether this should be codified in the rules instead - i.e. you lose the ability to do politics after capital is lost - but that seems like a bigger can of worms, hence why I've opted for this version currently.)

Happy to listen to what you think!

ron-murhammer:
@asvitkine I'm not sure this change would work across all maps as I believe there are some with no capitals. I think my question is what maps have you played where the AI was declaring war after they had lost their capital?

asvitkine:
Ah right, it should be "capital == null || controls capital". I can change it to that.

For the map, its napoleonic wars ffa.

asvitkine:
Apologies for the delay. I updated the logic per the above.

I think the better long-term solution is to make the Politics AI actually do reasonable things - e.g. declare war on enemies that it can attack reasonably and sue for peace when it's in trouble. And make consistent decision when replying to politics actions vs. initiating them.

Right now, it's mostly random.

Still, I think this is a reasonable improvement in the meantime, before we can get some reasonable Politics AI logic in place.

PR #5252, opened 2019-09-20
Game Listing Server Backend

DanVanAtta:
Adds HTTP endpoints for 'game listing' functionality: {post, remove, update, boot, keep-alive, fetch}

There are two unimplemented listeners for notifying existing, connected,
clients for game updates, those are marked with TODOs and will land in future updates.

'GameReaper', a utility to enforce keep-alive and prune 'dead' games implements
the keep-alive feature. See the javadoc of that class for further details.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[ ] Manually tested
[ ] No testing done

### DIFF
--- a/http-clients/src/main/java/org/triplea/http/client/lobby/HttpLobbyClient.java
+++ b/http-clients/src/main/java/org/triplea/http/client/lobby/HttpLobbyClient.java
@@ -2,15 +2,21 @@ package org.triplea.http.client.lobby;
 
 import java.net.URI;
 import lombok.Getter;
+import org.triplea.http.client.lobby.game.listing.GameListingClient;
 import org.triplea.http.client.lobby.moderator.toolbox.HttpModeratorToolboxClient;
 
 /** Holder class for the various http clients that access lobby resources. */
 @Getter
 public class HttpLobbyClient {
   private final HttpModeratorToolboxClient httpModeratorToolboxClient;
-  // TODO: Project#12 Add additional http clients
+  private final GameListingClient gameListingClient;
 
-  public HttpLobbyClient(final URI lobbyUri, final String apiKey) {
+  private HttpLobbyClient(final URI lobbyUri, final String apiKey) {
     httpModeratorToolboxClient = new HttpModeratorToolboxClient(lobbyUri, apiKey);
+    gameListingClient = GameListingClient.newClient(lobbyUri, apiKey);
+  }
+
+  public static HttpLobbyClient newClient(final URI lobbyUri, final String apiKey) {
+    return new HttpLobbyClient(lobbyUri, apiKey);
   }
 }
--- a/http-clients/src/main/java/org/triplea/http/client/lobby/game/listing/GameListingClient.java
+++ b/http-clients/src/main/java/org/triplea/http/client/lobby/game/listing/GameListingClient.java
@@ -1,0 +1,57 @@
+package org.triplea.http.client.lobby.game.listing;
+
+import java.net.URI;
+import java.util.List;
+import lombok.AllArgsConstructor;
+import org.triplea.http.client.AuthenticationHeaders;
+import org.triplea.http.client.HttpClient;
+
+/**
+ * Http client for interacting with lobby game listing. Can be used to post, remove, boot, fetch and
+ * update games.
+ */
+@AllArgsConstructor
+public class GameListingClient {
+
+  public static final String BOOT_GAME_PATH = "/lobby/games/boot-game";
+  public static final String FETCH_GAMES_PATH = "/lobby/games/fetch-games";
+  public static final String KEEP_ALIVE_PATH = "/lobby/games/keep-alive";
+  public static final String POST_GAME_PATH = "/lobby/games/post-game";
+  public static final String UPDATE_GAME_PATH = "/lobby/games/update-game";
+  public static final String REMOVE_GAME_PATH = "/lobby/games/remove-game";
+
+  private final AuthenticationHeaders authenticationHeaders;
+  private final GameListingFeignClient gameListingFeignClient;
+
+  public static GameListingClient newClient(final URI serverUri, final String apiKey) {
+    return new GameListingClient(
+        new AuthenticationHeaders(apiKey),
+        new HttpClient<>(GameListingFeignClient.class, serverUri).get());
+  }
+
+  public List<LobbyGameListing> fetchGameListing() {
+    return gameListingFeignClient.fetchGameListing(authenticationHeaders.createHeaders());
+  }
+
+  public String postGame(final LobbyGame lobbyGame) {
+    return gameListingFeignClient.postGame(authenticationHeaders.createHeaders(), lobbyGame);
+  }
+
+  public void updateGame(final String gameId, final LobbyGame lobbyGame) {
+    gameListingFeignClient.updateGame(
+        authenticationHeaders.createHeaders(),
+        UpdateGameRequest.builder().gameId(gameId).gameData(lobbyGame).build());
+  }
+
+  public boolean sendKeepAlive(final String gameId) {
+    return gameListingFeignClient.sendKeepAlive(authenticationHeaders.createHeaders(), gameId);
+  }
+
+  public void removeGame(final String gameId) {
+    gameListingFeignClient.removeGame(authenticationHeaders.createHeaders(), gameId);
+  }
+
+  public void bootGame(final String gameId) {
+    gameListingFeignClient.bootGame(authenticationHeaders.createHeaders(), gameId);
+  }
+}
--- a/http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListing.java
+++ b/http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListing.java
@@ -1,0 +1,204 @@
+package org.triplea.server.lobby.game.listing;
+
+import com.google.common.annotations.VisibleForTesting;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+import lombok.Builder;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+import org.triplea.http.client.lobby.game.listing.LobbyGame;
+import org.triplea.http.client.lobby.game.listing.LobbyGameListing;
+import org.triplea.lobby.server.db.dao.ModeratorAuditHistoryDao;
+
+/**
+ * Class that stores the set of games in the lobby. Games are identified by a combination of two
+ * values, the api-key of the user that created the game and a UUID assigned when the game is
+ * posted. Keep in mind that the 'gameId' is a publicly known value. Therefore we use API key, for
+ * example, to ensure that other users can't invoke endpoints directly to remove the games of other
+ * players.<br>
+ *
+ * <h2>Keep-Alive</h2>
+ *
+ * The game listing has a concept of 'keep-alive' where games, after posting, need to send
+ * 'keep-alive' messages periodically or they will be de-listed. If a game has been de-listed, and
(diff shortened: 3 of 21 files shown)


http-clients/src/main/java/org/triplea/http/client/lobby/HttpLobbyClient.java
```diff
@@ -2,15 +2,21 @@
...
 
-  public HttpLobbyClient(final URI lobbyUri, final String apiKey) {
+  private HttpLobbyClient(final URI lobbyUri, final String apiKey) {
     httpModeratorToolboxClient = new HttpModeratorToolboxClient(lobbyUri, apiKey);
+    gameListingClient = GameListingClient.newClient(lobbyUri, apiKey);
+  }
+
+  public static HttpLobbyClient newClient(final URI lobbyUri, final String apiKey) {
+    return new HttpLobbyClient(lobbyUri, apiKey);
   }
```
ron-murhammer: What's the reasoning for having the `newClient` method vs using the constructor?

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListing.java
```diff
@@ -0,0 +1,204 @@
...
+  @Builder
+  @EqualsAndHashCode
+  @Getter
+  static class GameId {
+    @NonNull private final String apiKey;
+    @NonNull private final String id;
+  }
+
+  @VisibleForTesting
+  class GameNotFound extends RuntimeException {
```
ron-murhammer: For these custom exceptions. would it be better to suffix them with 'Error' or 'Exception'?

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameReaper.java
```diff
@@ -0,0 +1,43 @@
...
+
+  GameReaper(final int duration, final TimeUnit timeUnit) {
+    this(CacheBuilder.newBuilder().expireAfterWrite(duration, timeUnit).build());
+  }
+
+  /**
+   * Returns set of game ids to be reaped. These are games that have not received a keep alive
+   * within the cut-off time.
+   */
+  Collection<GameListing.GameId> findDeadGames(final Collection<GameListing.GameId> gameIds) {
```
ron-murhammer: Any reason for needing to have `GameListing.GameId` instead of importing it?

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListingController.java
```diff
@@ -0,0 +1,115 @@
...
+import org.triplea.server.access.AuthenticatedUser;
+
+/** Controller with endpoints for posting, getting and removing games. */
+@Builder
+@Produces(MediaType.APPLICATION_JSON)
+@Path("/")
+@AllArgsConstructor(
+    access = AccessLevel.PACKAGE,
+    onConstructor_ = {@VisibleForTesting})
+@RolesAllowed(UserRole.ANONYMOUS)
```
ron-murhammer: Can you clarify what `@RolesAllowed(UserRole.ANONYMOUS)` is indicating here?

http-clients/src/main/java/org/triplea/http/client/lobby/HttpLobbyClient.java
DanVanAtta: It's more consistent with the other the http clients, they nearly all have a factory method rather than public constructor

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListing.java
DanVanAtta: Exception maybe, but IMO it's kingdom of nouns situation, it's already clear they are exceptions. An example usage:
```
           .orElseThrow(() -> new GameNotFound(gameId));
```
I don't think that needs an 'Exception' suffix to be more clear.

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListingController.java
DanVanAtta: It means you need to be authenticated to access the endpoint. Anonymous is lowest authentication level, so anyone that is authenticated can access. It's analogous to an "unquarantined" socket connection. In other words, users that are not banned.

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameReaper.java
DanVanAtta: It's style mostly I suppose. I've a slight preference for namespacing inner classes rather than importing them. I don't think we have a strong convention implicit or explicit around this. Going forward, if to be called out all the time, we should add adopt a code convention around this (open an issue, get consensus, update the [wiki](https://github.com/triplea-game/triplea/wiki/Code-Conventions))

My 2 cents on that, I'm opposed as I had to by hand add the import statement, there are essentially no imports where I need to do that. I've the IDE setup to import automatically when there is only one choice and I've added filter lists so that nearly all imports are done automatically. To fix imports, I merely need to open a file. I don't see much value in a convention to explicitly import inner classes, IMO it's not going to save that much time and IMO reduces some clarity by being less explicit. I generally prefer inner classes to be namespaced, and if not then made to be proper stand-alone classes. I could see reasons to do that here for `GameId`, my initial inclination was to see how it develops as it's mostly meant to be internal to GameListing and hide the fact that ApiKey is attached to a game.

Regardless, in this instance, I've only weak preferences..  Updated: 3a92622a6

http-clients/src/main/java/org/triplea/http/client/lobby/HttpLobbyClient.java
ron-murhammer: Ok. I don't really see much value but not a big deal.

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListing.java
ron-murhammer: I guess I prefer having the suffix as its probably more consistent with the rest of the exceptions and makes it easier to read. I think some of these like "GameNotFound" could be say an entry in a game list or a response object to a user that clicked on a game that didn't exist. If you prefer to leave as is I'm ok with that. I'll leave it up to you.

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameReaper.java
ron-murhammer: Ok, I think there are a few other places as well. Can you do a quick search on "GameListing.GameId". I tend to prefer the import as it just makes the code easier to read IMO and more concise.

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListingController.java
ron-murhammer: Ok. It reads a little odd but I see what you mean. I think its just sounds a bit odd to have an "anonymous" role and that all the other user types would be able to authenticate to be "anonymous" role. Where as user vs mod vs admin is a more natural progression. But that might just be the reality of how its structured.

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListing.java
DanVanAtta: Principle of least surprise does apply; then again, it's a language level problem that all exceptions are suffixed Exception and the language probably should never have done that. Fighting the language does violate principle of least surprise, I think this where people start to not like Java as a language (particularly bad in J2EE and previous versions, lots of harmful ceremony built into the language).

I'm inclined to see how this plays out, the exception is pretty local and always is surrounded in context for now.

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameReaper.java
DanVanAtta: There was one more place in the same class, now fixed: 
7099bc8 

The test code uses the non-import version, it's consistent there for now and not worth fixing IMO. I somewhat regret now updating it at all, this is where I start to push back. The legibility difference is absolutely minor, meanwhile this is an import that needs to be done by hand. I don't think there should be many imports of inner classes as that kinda violates encapsulation, inner classes really should be package level and utility level things, not used as public imports (otherwise, make them proper classes). So for me, the inner class is a callout that we're dealing with a very localized entity. `GameReaper` and `GameListing` are very closely related, hence the "reaching in" of `GameReaper` into `GameListing`. Having `GameId` be stand-alone I don't think would be good as it clutters up the package, and makes it seem like it would be okay to use in many places. For the most part, the API-key should not be used as part of the identity of lobby games, and that I want hidden as much as possible.

The reason game reaper uses GameId at all is convenience of API, so we can avoid having to strip api-key for the parameter call to add it back when processing the result. There's certainly a code smell here, but I'm trying to balance that against having to do a lot of iteration work on the caller side to remove/attach API-key. In another respect, the lack of import and namespacing I think calls out the code-smell, IMO the real fix is generally to eliminate that vs using an import.

http-clients/src/main/java/org/triplea/http/client/lobby/game/listing/GameListingClient.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+@AllArgsConstructor
+public class GameListingClient {
+
+  public static final String BOOT_GAME_PATH = "/lobby/games/boot-game";
+  public static final String FETCH_GAMES_PATH = "/lobby/games/fetch-games";
+  public static final String KEEP_ALIVE_PATH = "/lobby/games/keep-alive";
+  public static final String POST_GAME_PATH = "/lobby/games/post-game";
+  public static final String UPDATE_GAME_PATH = "/lobby/games/update-game";
+  public static final String REMOVE_GAME_PATH = "/lobby/games/remove-game";
```
RoiEXLab: Just a small side-note, if you were using HTTP methods besides `GET` and `POST` like `PUT`, `PATCH` and `DELETE`, this would allow to use the same uri string for multiple actions, which at least IMO makes it simpler to work with.
Example: `PUT /lobby/game`, `PATCH /lobby/game` and `DELETE /lobby/game` vs. `POST /lobby/games/post-game`, `POST /lobby/games/update-game` and `POST /lobby/games/remove-game`.

Of course those methods have different implications like the caching mechanisms that come with `GET` and make promises to have or not have side-effects, or to be indempotent, so that would need checking, but it might make it easier to make assumptions about the API in the future.
Obviously the overall benefit in this case is rather small, and REST really only comes to shine on larger use-cases, but I still think it would be nice to use this here.

DanVanAtta: PUT usually implies a file upload, right?

REST AFAIK also implies self-documentation. Similar to the github api for example, a request to the root URL tells you which other URLs are available. I think what we'd done here is simply JSON over HTTP. Using status codes well I think is a win, proper use the HTTP verbs, I'd agree the benefit is for us kinda marginal.

On the other hand, I do want to see an example of HTTP PUT with file upload. There is a lot of utility in that. We could replace the forum poster with a feign client and open up a lot of options like 'cloud save-games' and upload of map files and images.

RoiEXLab: AFAIK `PUT` only implies the creation of a resource. What defines such a resource and from what kind of data it is created isn't defined further.
So creating a game entry could be seen as creating a resource.

DanVanAtta: We could do something like:

```
POST /games
PUT /games/<GAME_ID>
GET /games/<GAME_ID>
PATCH /games/<GAME_ID>
```

The difference between PUT and PATCH seems ambiguous. Passing the 'game_id' through parameter or URL string is not very much different. I'm not sure there is that much value for having more faithful adherence to REST. We're not necessarily doing REST here.

PR #5255, opened 2019-09-21
Add simple check for non-null and non-empty lobby host value.

DanVanAtta:
## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/login/LobbyServerProperties.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/login/LobbyServerProperties.java
@@ -11,6 +11,7 @@ import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.ToString;
 import org.apache.http.client.utils.URIBuilder;
+import org.triplea.java.ArgChecker;
 
 /**
  * Server properties.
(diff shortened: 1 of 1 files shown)


prastle:
In layman's terms will this fix the crashed ghosted bots?

DanVanAtta:
No, ghosting is a much deeper problem rooted somewhere in some gnarly/complex Java socket code that was written 17 years ago :grin: 

This check is validity constraint so that the code blows up sooner rather than later; we get a more sensible error message sooner in the system rather than later with something else failing.

prastle:
Sigh KK

DanVanAtta:
@prastle this PR should though, at least, once the backend infrastructure is used: https://github.com/triplea-game/triplea/pull/5252

Of particular note, the '[GameReaper](https://github.com/triplea-game/triplea/pull/5252/files#diff-60b7dd1a97e376266cb0d5c32a9b5ef9)' that collects dead games and removes them.

It'll be a pretty big lift though to go from the using the classic-java lobby of today to use the http server with game reaper, but a big step is taken in that direction.

The fix for ghosting is going to be two fold:
(1) better handling of lost connections
(2) games will be required to send a 'keep-alive' message, if one is not received then game-reaper should kick in and remove them.

With luck too this should allow for an auto-reconnect to lobby.
Game hosts will send keep-alives, if the lobby dies, game hosts will see no response. When lobby is back, the keep alive message will be acknowledged with a 'false' instead of true. At that point the game hosts will re-post their games to the resurrected lobby.

PR #5257, opened 2019-09-21
Convert JPanelBuilder to a type-safe builder

DanVanAtta:
Calling a layout method on JPanelBuilder now returns a specific layout builder.
This allows for layout specific methods to be available only after invoking
that layout. For example, before a flowlayout could be used with a 'addNorth'
method, specific to border layout.

In addition, as a new feature, We add a new method to GridBagLayout to have an
'add(Component, GridbagConstraint)' method.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[x] Manually tested
[ ] No testing done

Checked various UI windows to look for rendering problems

## Additional Review Notes

SimpleGridBagLayout is not yet used. Most of the test code added revolves around it.

### DIFF
--- a/swing-lib/src/main/java/org/triplea/swing/jpanel/JPanelBuilder.java
+++ b/swing-lib/src/main/java/org/triplea/swing/jpanel/JPanelBuilder.java
@@ -1,103 +1,47 @@
 package org.triplea.swing.jpanel;
 
-import com.google.common.base.Preconditions;
-import java.awt.BorderLayout;
 import java.awt.Component;
 import java.awt.Dimension;
-import java.awt.FlowLayout;
-import java.awt.GridBagLayout;
-import java.awt.GridLayout;
-import java.awt.LayoutManager;
-import java.util.ArrayList;
-import java.util.Collection;
-import javax.annotation.Nonnull;
 import javax.swing.BorderFactory;
-import javax.swing.Box;
-import javax.swing.BoxLayout;
 import javax.swing.JComponent;
-import javax.swing.JLabel;
 import javax.swing.JPanel;
 import javax.swing.border.Border;
 import javax.swing.border.EtchedBorder;
-import lombok.AllArgsConstructor;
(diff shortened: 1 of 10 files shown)


swing-lib/src/main/java/org/triplea/swing/jpanel/JPanelBuilder.java
```diff
@@ -197,125 +78,40 @@ public JPanelBuilder horizontalAlignmentCenter() {
...
-   * create the layout until after we create the component. Thus we use a flag to create it, rather
-   * than creating and storing the layout manager directly as we do for the other layouts such as
-   * gridLayout.
-   */
-  private enum BoxLayoutType {
-    NONE,
-    HORIZONTAL,
-    VERTICAL
+  public BoxLayoutBuilder boxLayoutHorizontal() {
+    return new BoxLayoutBuilder(this, BoxLayoutBuilder.BoxLayoutOrientation.HORIZONTAL);
```
RoiEXLab: Not exactly how I envisioned it but pretty similar.
I'd have expected the "specific" builders to extend the original builder, so that all common actions could stay there.
Not sure which approach is better, because my approach has the downside that the layout could be defined multiple times and this could only be checked at run-time, not at compile-time

DanVanAtta: That's an interesting way to extend the common options to all builders.

I'd rather there be a stronger 'type-safe' way to do the builders, the classic builder pattern forces you to specify each required option, in order. I think to that extent, having an ordering requirement to specify the common options is okay. On the other hand, avoiding inheritance is significant, I could see additional issues coming up with inheritance (for example, method overriding could start making things very complicated)

Issue #5253, opened 2019-09-20
Save files don't properly load when double clicked on macOS


NickClark:
**Describe the bug**
When opening a save file on macOS, it opens TripleA, but doesn't load the file.

**To Reproduce**
Double Click any save file. Notice that when TripleA starts, no save file is selected.

**Expected behavior**
The opened file should be selected. Even better if it could jump back into the game.

**OS (please put an X in the appropriate box):**
[] Windows
[x] Mac
[] Linux

RoiEXLab:
Hey @NickClark , thanks for bringing this to our attention.
Which version of TripleA are you using?

NickClark:
I've tried both 1.9.0.0.13066 and a recent 2.0 nightly

RoiEXLab:
Ok, I dug a little bit into the issue.

Looks like this never actually worked because macos opens files a bit different than all other operating systems :(
I'll try a blind fix and would appreciate if @NickClark could verify it actually works once we have a build artifact (aka a new pre-release)

RoiEXLab:
Opened #5262

NickClark:
I'll take a look. Where's your CI's build artifacts hosted? Or is this not ready to test yet?

RoiEXLab:
Ok, the latest pre-release should have the changes, let us know if it works.

NickClark:
Tested v 2.0.16018 Still doesn't work with a newly created game

RoiEXLab:
Wait, what do you mean with "newly created"?

NickClark:
Meaning I didn't use an 1.9 save, I made sure to make a new 2.0 save file.

RoiEXLab:
Ok, I'll have to try something else then.
There wasn't any error message either, right?

NickClark:
No... is there a way to turn on extra logging?

RoiEXLab:
Well there is, but unhandled exceptions always trigger an error window, so you won't get anything out of that.

RoiEXLab:
Ok, can you try https://github.com/triplea-game/triplea/releases/tag/2.0.16048 ?
I don't expect the feature to work, but you should get an info message when trying.

NickClark:
No info message shows... hmm. It opens the app... and it is the right version. Side note (just so you have my environment details), the window that opens reports itself as Java version 11.0.4 and Engine Version as 2.0.16048. Loaded Savegame reported as "-"

RoiEXLab:
[...]

PR #5262, opened 2019-09-22
Open Savegames on MacOS correctly

RoiEXLab:
More or less a blind fix, I hope this works, but am unable to test it.

Note that there are 2 commits. The first commit is what is supposed to be the fix, the second commit does some minor refactoring.
I noticed that there are 2 places where a quit listener is set in macos, if they are both triggered, they would replace each other, not sure what to do about that...

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[x] Bug fix: #5253
[ ] Other:   

## Testing

[ ] Manually tested
[x] No testing done

### DIFF
--- a/game-headed/src/main/java/org/triplea/game/client/HeadedGameRunner.java
+++ b/game-headed/src/main/java/org/triplea/game/client/HeadedGameRunner.java
@@ -67,15 +67,11 @@ public final class HeadedGameRunner {
             SwingUtilities.invokeLater(
                 () -> DownloadMapsWindow.showDownloadMapsWindowAndDownload(mapName));
           });
-      MacOsIntegration.setOpenFilesHandler(
-          list ->
-              list.stream()
-                  .findAny()
-                  .ifPresent(
-                      file -> {
-                        System.setProperty(CliProperties.TRIPLEA_GAME, file.getAbsolutePath());
-                        GameRunner.showMainFrame();
-                      }));
+      MacOsIntegration.setOpenFileHandler(
+          file -> {
+            System.setProperty(CliProperties.TRIPLEA_GAME, file.getAbsolutePath());
+            GameRunner.showMainFrame();
+          });
     }
 
     if (HttpProxy.isUsingSystemProxy()) {
--- a/game-headed/src/main/java/org/triplea/game/client/HeadedGameRunner.java
+++ b/game-headed/src/main/java/org/triplea/game/client/HeadedGameRunner.java
@@ -58,7 +58,7 @@ public final class HeadedGameRunner {
     ArgParser.handleCommandLineArgs(args);
 
     if (SystemProperties.isMac()) {
-      MacOsIntegration.addOpenUriHandler(
+      MacOsIntegration.setOpenUriHandler(
           uri -> {
             final String mapName =
                 URLDecoder.decode(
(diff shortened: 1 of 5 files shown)


game-headed/src/main/java/org/triplea/game/client/HeadedGameRunner.java
```diff
@@ -66,6 +67,15 @@ public static void main(final String[] args) {
             SwingUtilities.invokeLater(
                 () -> DownloadMapsWindow.showDownloadMapsWindowAndDownload(mapName));
           });
+      MacOsIntegration.setOpenFilesHandler(
+          list ->
```
DanVanAtta: Why is this a list?
I'm noting that as picking the first element we find could cause ordering dependencies or other incorrect behavior. Is this to cover if a person selects many files and opens them all?

DanVanAtta: Would it make for a cleaner API if  `MacOsIntegration.setOpenFilesHandler(` were to take a `File`

DanVanAtta: 2nd question, is this only going to work if TripleA is already open? It's a bit odd to register a file handler for opening files? Isn't the idea that opening save game files would open TripleA and TripleA would not already need to be open.

RoiEXLab: > Why is this a list?

No idea, it's probably for the case of many files opened.

> Would it make for a cleaner API if  `MacOsIntegration.setOpenFilesHandler(` were to take a `File`

Simplifying this might be an option, but currently this class is just a generic wrapper for the desktop api, all of the other methods don't really try to hide details either.

Aand no,this will work even if triplea is not opened.
See, on Linux and Windows file handling causes the os to launch a new process for every single file. Macos on the other hand doesn't really like having multiple processes, instead it prefers a single one with multiple windows.
So when triplea isn't opened already it will launch an instance and instantly send a file open event, similar to uri handling, and if another file is opened, the same process receives another event, thinking about it, this might have unforseen consequences when this happens during a game, besides the fact that there might be a race condition here anyways (the event might be fired before the main menu is instantiated

DanVanAtta: Wait, we had our own custom logic in MacOS integration, it denigrated to just a wrapper, and now because of that we can't use it to hide details and it just needs to be a pass-thru wrapper?

- IMO the wrapper should try to hide details or it's better to add deprecation tags and not use it.
- Some commentary would be good to avoid confusion on how the file handles works, I don't know if it'll be clear to future maintainers how this works.

DanVanAtta: > all of the other methods don't really try to hide details either.

They do not have similar details to hide and we've already discussed it probably makes sense to inline them.

The API added does not make sense and is not well documented. Should we really just pick a file out of many, or perhaps open no files and make it up to the user to pick the right one? A List API where many elements are ignored and only one is active, or allowing for a no-op if an empty list is not terribly well defined, the behavior is very non-deterministic.

RoiEXLab: @DanVanAtta I simplified the API, let me know what you think.

DanVanAtta: Much simpler, thank you!

PR #5265, opened 2019-09-22
Refactor: Simplify GameListing

DanVanAtta:
- Combine GameReaper with GameListing and store list of games in cache directly.
- Avoid paranoid checking of wrong API key with existing game ids, instead just no-op or create a new game, treate game ID as a true combination of UUID and API-Key
- Add value object 'api-key' for strong typing, double checking that String parameters were in the right order was a bit time-consuming and seemed error-prone
- With the gamelisting stored in cache, values are expunged automatically without needing a background thread to do so.
- Remove not yet used game event listeners, keep things simple for now until the listeners are actually used.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[ ] Manually tested
[ ] No testing done


http-server/src/main/java/org/triplea/server/access/ApiKey.java
```diff
@@ -0,0 +1,9 @@
+package org.triplea.server.access;
+
+import lombok.Value;
+
+/** Simple value object for strong typing. */
+@Value(staticConstructor = "of")
+public class ApiKey {
+  private final String value;
+}
```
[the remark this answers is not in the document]
DanVanAtta: It's identical to the other ApiKey, it looks it'll probably be moved to http-clients to be shared.

DanVanAtta:
> However I have a small concern about the api: Is it the case that the client periodically checks for game listing updates?

It looks likely yes. To know about ghosted games dropping off, we'd need a loop to periodically check the internal state and then emit updates, or alternatively we have the clients regularly check for updates. I was leaning towards websockets thinking it was a bit much to send full game data every time, on the other hand I think that is premature optimization talking and the KB or two of data is really a pittance when we are talking about update tolerances over a few seconds and not sub-second. The client polling for updates has another benefit as well in case any update messages are missed, it's a stronger synchronization back to the server, even if not as immediate.


> If not then this list is going to not be synchronised at all over different clients

How do you mean?

RoiEXLab:
> How do you mean?

I just wanted to point out that I wasn't able to find a mechanism that broadcasted a removal whenever a game entry changed, so I just wanted to make sure there's a client mechanism to make up for that

DanVanAtta:
There are two ways for a removal to happen, one is for an expiry or second an explicit remove. There was a 'removeListener' previously that went to a no-op that was intended to be wired to a client. It's premature at this point to make that decision of whether that will be done via websocket or polling.

The explicit remove was wired to that listener, and the expiry was not wired at all and had a TODO. To wire the expiry, we would have needed a poller thread server side. Having the cache expire entries is pretty clean, checking server-side in a thread for expired entries is not clean. As is, it looks like the simplest way to achieve client side updates will be through periodic checks of current state.

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListing.java
```diff
@@ -37,116 +38,50 @@
...
-        });
+  boolean updateGame(final ApiKey apiKey, final String gameId, final LobbyGame lobbyGame) {
+    final GameId listedGameId = new GameId(apiKey, gameId);
+    return Optional.ofNullable(games.getIfPresent(listedGameId))
+        .map(
+            game -> {
+              games.put(listedGameId, lobbyGame);
+              return true;
+            })
+        .orElse(false);
```
RoiEXLab: Haven't actually tried this, but what about this:
```java
return games.asMap().replace(listedGameId, lobbyGame) != null
```

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListing.java
```diff
@@ -163,42 +98,35 @@ private void removeGame(final GameId gameId) {
...
-    removeGame(gameToRemove);
-    log.info("Moderator {} booted game: {}, hosted by: {}", moderatorId, gameId, hostName);
-    auditHistoryDao.addAuditRecord(
-        ModeratorAuditHistoryDao.AuditArgs.builder()
-            .moderatorUserId(moderatorId)
-            .actionName(ModeratorAuditHistoryDao.AuditAction.BOOT_GAME)
-            .actionTarget(hostName)
-            .build());
+    games.asMap().entrySet().stream()
+        .filter(entry -> entry.getKey().id.equals(gameId))
```
RoiEXLab: This really makes me feel like there should be 2 different maps
1. gameId string -> API key
2. gameId string -> LobbyGame

Instead of
gameId object -> Lobby game

DanVanAtta: The fact that 'post' game generates a game id prevents us from having duplicates, but in theory it would be okay for there to be a duplicate gameId so long as the API keys were different. On the client side that would not work, hence it's not allowed. More fundamentally though the 'game id' is a true compound key, searching by gameId and wrong API key returns no results. 

The double map I see as having a few problems:
- multiple caches that need to be kept in sync
- parallel data structure problem (it's similar to two arrays where there is correlation between the pair-wise elements)
- logic for not found when gameId is there but API key is not would be a bit more complicated. Recall that gameId is publicly known, someone could try to update a game by passing a gameId, but their API key would not match. The lookup logic would be a bit more complicated rather than a simple 'not found' case.

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListing.java
DanVanAtta: Good suggestion, I'd break it up into two lines though, we're doing many different operations on the single line, bit confusing.

DanVanAtta: Updated (3fb6320f1) to this:
```
    final var listedGameId = new GameId(apiKey, gameId);
    final var existingValue = games.asMap().replace(listedGameId, lobbyGame);
    return existingValue != null;
```

PR #5267, opened 2019-09-24
Wire Https Port to Server Games

DanVanAtta:
- Use LobbyServerProperties object to pass lobby properties, this results in simplifications
where we no longer need to get lobby properties from LobbyClient

- Pass HTTPs port to bot via args.

Net result is HTTPS port should become available to hosted games.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[ ] Manually tested
[ ] No testing done

### DIFF
--- a/.travis/run_smoke_tests
+++ b/.travis/run_smoke_tests
@@ -74,6 +74,7 @@ function startBot() {
     -Dtriplea.name=Bot01Testing \
     -Dtriplea.lobby.host=localhost \
     -Dtriplea.lobby.port=3304 \
+    -Dtriplea.lobby.https.port=8080 \
     -Dtriplea.lobby.game.supportPassword= \
     -Dtriplea.map.folder=~/triplea/downloadedMaps/ \
     -jar "$BOT_JAR_PATH" > "$BOT_OUTPUT" 2>&1 &
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyGamePanel.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyGamePanel.java
@@ -3,6 +3,7 @@ package games.strategy.engine.lobby.client.ui;
 import games.strategy.engine.framework.GameRunner;
 import games.strategy.engine.framework.startup.ui.ServerOptions;
 import games.strategy.engine.lobby.client.LobbyClient;
+import games.strategy.engine.lobby.client.login.LobbyServerProperties;
 import games.strategy.net.INode;
 import games.strategy.net.Node;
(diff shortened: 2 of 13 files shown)


game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyGamePanel.java
```diff
@@ -155,7 +161,10 @@ private void mouseOnGamesList(final MouseEvent e) {
 
     final JPopupMenu menu = new JPopupMenu();
 
-    getUserGamesListContextActions().forEach(menu::add);
+    Arrays.asList(
+            SwingAction.of("Join Game", this::joinGame),
+            SwingAction.of("Host Game", () -> hostGame(lobbyServerProperties)))
+        .forEach(menu::add);
```
RoiEXLab: I don't think having a list makes this more readable at this point.

.travis/run_smoke_tests
```diff
@@ -74,6 +74,7 @@ function startBot() {
     -Dtriplea.name=Bot01Testing \
     -Dtriplea.lobby.host=localhost \
     -Dtriplea.lobby.port=3304 \
+    -Dtriplea.lobby.https.port=8080 \
```
RoiEXLab: Typically HTTPS is 443, whereas 8080 is typically used as alternative port for HTTP (80) which might be confusing.

Anyways what I'm trying to say is that IMO this suggests that the transport (i.e. TLS) is an essential part for the communication to work, so if `https` is just supposed to represent the HTTP based lobby system, I'd actually vote to drop the s.

DanVanAtta: I don't think I'm really thrilled that we call it 'https' or 'http' at all, we're bubbling up technology details.

We do call it 'https' in other places, notably client settings, I think we shot ourselves in the foot already. Of course, until we get some local certs installed and start using https locally, the port must be 8080. The intent is that we will get it switched to be https, having https not be available locally would be a bad move.

game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyGamePanel.java
DanVanAtta: Personally I think that says something about the helper method: ` getUserGamesListContextActions()`.
Having the values be one per line IMO is a readability win.

To get parity with before, we could extract the list to an intermediate variable, but I think that would not be much cleaner in the end and highlights the suspect extracted method (which was excessively indirect)

PR #5280, opened 2019-09-26
Refactor: Convert 'String' to value object 'PlayerName'

DanVanAtta:
First iteration to mass convert unbounded strings to domain object 'PlayerName'.
Refactor begins with updating the API of of 'Node' to have a 'getPlayerName'
method with objective to eventually to convert the internal data type of
'Node.name' to be 'PlayerName' and be strongly typed whenever accessed.

Another goal behind this conversion is to avoid passing 'Node' data where
typically only one property of the Node, name, is used.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[ ] Manually tested
[x] No testing done

## Additional Review Notes

Notable benefits:
- domain objects! Good step, if we add validation/constraints then we'll be approaching  'domain driven security' (eg: https://www.youtube.com/watch?v=9mGsLcruhwQ "Domain Driven Security")
- type-safe, can't accidentally use incompatible types as parameters or variables, get a warning when comparing incompatible types

The big driver here is to start isolating node data from player name data. We currently pass a list of nodes to chat, we do not need IP addresses, but because it is attached to node, we pass it anyways and thereby leak player IP addresses.

This update is not complete, to avoid a completely overwhelming review it will be done in iterations. We also need to strike up a domain terminology conversation, notably because there is room for error and further updates start to run up against other Strings that represent other domain entities. It's notable for example that 'player' has multiple meanings (in the code, and even within the same sub-domain).

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/chat/ChatPlayerPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/ChatPlayerPanel.java
@@ -1,5 +1,6 @@
 package games.strategy.engine.chat;
 
+import games.strategy.engine.lobby.PlayerName;
 import games.strategy.net.INode;
 import java.awt.BorderLayout;
 import java.awt.Dimension;
@@ -15,6 +16,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import javax.imageio.ImageIO;
 import javax.swing.Action;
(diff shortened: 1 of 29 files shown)


game-core/src/main/java/games/strategy/engine/chat/ChatPlayerPanel.java
```diff
@@ -255,9 +253,6 @@ private String getDisplayString(final INode node) {
       }
       extra = extra + " (" + sb + ")";
     }
-    if (extra.length() == 0) {
-      return node.getName();
-    }
     return node.getName() + extra;
```
RoiEXLab: Those length == 0 checks are so weird

game-core/src/main/java/games/strategy/engine/chat/ChatPlayerPanel.java
```diff
@@ -255,9 +253,6 @@ private String getDisplayString(final INode node) {
       }
       extra = extra + " (" + sb + ")";
```
RoiEXLab: Side-Note: This looks like the existing stringbuilder should be used for that

game-core/src/main/java/games/strategy/engine/chat/ChatPlayerPanel.java
DanVanAtta: Yeah, this whole thing can be simplified quite a bit. The `getNotesForNode` method too, in the end should become more or less a simple concatenation

DanVanAtta:
@RoiEXLab we can/should add validation in the 'of' factory method. Having fancy string wrappers is still valuable assuming we achieve useful domain modeling. HashedPassword I think was a failure because of the logic it contained and there were multiple types of hashed passwords. It was also used in cases when the password was potentially not hashed. "Password" might simply have been better. 

In the meantime I thought it was really interesting to see that 'player' takes on multiple meanings and in some string->string maps it was actually confusing/not clear what the values represented. I made a mistake in this update actually replacing a few 'playerNames' with the type that were not actually player names. 

The distinction of 'node name' vs 'player name' is worth considering, domain modelling would suggest to not invent names like that and to stick with the domain terminology.

PR #5281, opened 2019-09-26
Refactor: Simplify LobbyWatcher

DanVanAtta:
- Split 'reverse' connection check to its own class
- Add additional constructors to LobbyWatcher to simplify constructor logic
- Use Headless static calls to determine if local player is headless or human
- Do not instantiate lobby watcher unless we are actually connecting to lobby,
  creates expectation for non-null system props to know how to connect to lobby.
- Change return type of creating lobby watcher from nullable to optional.
  Empty optional is returned when the lobby is not available.
- Change API of 'testGame' to be simpler, we just need a boolean back to know
  if lobby could make the 'reverse' connection. Previously returned the actual
  address, which is not needed and not necessary to display as an error message.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[x] Manually tested
[ ] No testing done

Launched a local lobby:
- added a headless bot, verified player count, joined bot host, verified player count.
- hosted a game on local lobby, verified player count

### DIFF
--- a/game-core/src/main/java/org/triplea/game/server/HeadlessServerSetup.java
+++ b/game-core/src/main/java/org/triplea/game/server/HeadlessServerSetup.java
@@ -1,11 +1,15 @@
 package org.triplea.game.server;
 
+import static games.strategy.engine.framework.CliProperties.LOBBY_HOST;
+import static games.strategy.engine.framework.CliProperties.LOBBY_PORT;
+
 import games.strategy.engine.framework.startup.launcher.ILauncher;
 import games.strategy.engine.framework.startup.mc.GameSelectorModel;
 import games.strategy.engine.framework.startup.mc.IRemoteModelListener;
 import games.strategy.engine.framework.startup.mc.ServerModel;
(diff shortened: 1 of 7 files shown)


game-core/src/main/java/org/triplea/game/server/HeadlessServerSetup.java
```diff
@@ -34,23 +37,30 @@
...
+            .orElseThrow(CouldNotConnectToLobby::new);
+
+    lobbyWatcher.setInGameLobbyWatcher(watcher);
     lobbyWatcher.setGameSelectorModel(gameSelectorModel);
+    LocalServerAvailabilityCheck.builder()
+        .localNode(model.getMessenger().getLocalNode())
+        .controller(watcher.getRemoteMessenger().getLobbyGameController())
+        .errorHandler(log::severe)
+        .build()
+        .run();
```
RoiEXLab: Good thing this check is separate now 👍

PR #5282, opened 2019-09-26
Prune lobby welcome message

DanVanAtta:
- Remvoe upgrade text, if users have not upgraded by now, the continued lobby message is not going to make a difference.

- Remove help mention, there is a 'help' menu that users should be able to find without too much trouble

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[x] Other:   Lobby welcome text changed

### DIFF
--- a/lobby_server.yaml
+++ b/lobby_server.yaml
@@ -3,19 +3,14 @@
   port: 3304
   https_port: 5432
   message: |
-    The latest TripleA has important feature and security updates, please be sure you have updated in the last couple months:
-    https://triplea-game.org/download/
-
+    Welcome to TripleA, we are pleased you are here.
+    
     Join the TripleA ranked ladder by posting here:
     https://forums.triplea-game.org/topic/1068/invitation-requests-to-the-new-ladder-and-tournament-site
 
-    For questions or help, please visit the forums: https://forums.triplea-game.org/
-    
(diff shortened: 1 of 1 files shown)


Cernelius:
I'd change "Welcome to TripleA, we are pleased you are here" to only "Welcome to the TripleA lobby".

DanVanAtta:
It is a bit saccharine, but I don't think it hurts @Cernelius and avoids "welcome to..." from being a bit overly dry IMO. We could think to maybe remove the message. Long term I think we probably need to create a 'terms and rules of lobby' pop-up for first-time logins where a person has to agree once. Then we could remove most of the welcome message, I think it'll be most effective if it changes every few weeks and also is not more than a sentence or two.

DanVanAtta:
Merging for now, we can continue following-up and discussing changes.

Cernelius:
The other item that I was pointing out is that should welcome the user to the lobby, rather than to TripleA in general.

At most I would add something like "enjoy your stay".

DanVanAtta:
I missed that, good suggestion. I'm personally tempted to axe the welcome message line, not sure if it is really needed.

PR #5283, opened 2019-09-26
Add information dialog on mac os File Open event

RoiEXLab:
This PR is supposed to act as sort of debug information for #5253

## Functional Changes
[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[x] Other:  Debug information

## Testing
[ ] Manually tested
[x] No testing done

### DIFF
--- a/game-headed/src/main/java/org/triplea/game/client/HeadedGameRunner.java
+++ b/game-headed/src/main/java/org/triplea/game/client/HeadedGameRunner.java
@@ -74,7 +74,8 @@ public final class HeadedGameRunner {
                 () ->
                     JOptionPane.showMessageDialog(
                         null,
-                        "Unfortunately opening save-games via the OS is currently not supported on macOS.",
+                        "Unfortunately opening save-games via the OS"
+                            + " is currently not supported on macOS.",
                         "Unsupported feature",
                         JOptionPane.INFORMATION_MESSAGE));
             System.setProperty(CliProperties.TRIPLEA_GAME, file.getAbsolutePath());
--- a/game-headed/src/main/java/org/triplea/game/client/HeadedGameRunner.java
+++ b/game-headed/src/main/java/org/triplea/game/client/HeadedGameRunner.java
@@ -20,6 +20,7 @@ import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
 import java.util.logging.Level;
 import javafx.application.Application;
+import javax.swing.JOptionPane;
 import javax.swing.SwingUtilities;
 import lombok.extern.java.Log;
 import org.triplea.game.client.ui.javafx.TripleA;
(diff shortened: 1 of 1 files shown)


DanVanAtta:
@RoiEXLab , it looks like we're abandoning this feature.

I was wondering about it and was thinking to bring up the topic of whether we really wanted to support opening save game files directly, perhaps simply opening the game is enough. There are a number of apps where that is all that happens.

I wonder if we need to call this out. Ideally the save game file would be selected, in-game, but we should look to options. The Mac specific code causes headaches for the Debian distro, removing as much of it as possible would be a win on that front.

RoiEXLab:
I thought having a "sorry this is not supported" Message would be better for now  instead of it just doing nothing.
Obviously I plan to remove it once I figure out why the code is not working.

I think we should "keep" this feature just for consistency. It works on Windows and Linux distros, why shouldn't it work on Mac as well?

Regarding the Debian issues. I think in this case there shouldn't be any severe problems with this. I haven't looked at the Linux specific code of the openJDK, but at least for Windows the implementation for the desktop api are no-op methods.
Since we're requiring java 11+ we are also exclusively using public api, so that really shouldn't be an issue anymore.

DanVanAtta:
Worth double checking, I could be wrong, it may be that the Debian folks have been patching TripleA to remove all Mac specific code. There's a problem where there are downstream changes not present in the upstream (this repository).

> I thought having a "sorry this is not supported" Message would be better for now instead of it just doing nothing.

Assuming the game opens and then does nothing more, I wonder if the message might be a bigger surprise than the save game not being loaded. 

> It works on Windows and Linux distros, why shouldn't it work on Mac as well?

Fair question, my response:
- users on Mac are not going to be comparing with the other distro's, ignorance is bliss
- removing OS specific code is a win for us, kinda is not fun to lose the better UX, but we certainly profit for easier testing and maintenance

DanVanAtta:
@RoiEXLab I really don't know whether it's better to "silently fail" here, or if it's really a user expectation and that they think the game is bugged compared to not being as good as it could be.

RoiEXLab:
Good points raised, however I want to point out the main purpose of this PR: Provide debug information. So I'd be fine if this PR was merged and reverted in the next commit, I just need a build artifact to check if this code is ever run, or if it's doing something completely different which would make our entire discussion redundant

DanVanAtta:
>  I want to point out the main purpose of this PR: Provide debug information.

I don't think that was communicated, probably ideally the code would be marked with a TODO stating the current is temporary.

If we choose to drop this feature for mac only, notably because we won't ever be sure if still works (and so would be very brittle code for us, we'd have no confidence in changes), that would also moot this conversation as well.

RoiEXLab:
> I don't think that was communicated

Well it's part of the description, but doesn't really matter now.

DanVanAtta:
I suppose. But keep in mind though that the important commentary should land in commit comments. Any notes in PRs are only really valuable for the lifespan of the review. It's a tragedy when to understand code you have to go back to when it was merged and find a corresponding PR to read the comments in the PR. In other words, commentary/documentation needs to be captured in code and commit comments, the PR comments are of low value. 

I wasn't too sure what was meant exactly by 'debug' and the message in the code change seemed pretty authoritative and there was not marking it was temporary, hence the confusion.

RoiEXLab:
Yeah I figured I could've made it more obvious

PR #5285, opened 2019-09-27
Fix: Headless bot select player option

DanVanAtta:
To show a 'select' player button, we need to pass a null value to the
checker method. Updating to value objects, we instead had a null
value wrapped in a non-null method, which the checker method then
did not deal with properly.

This update fixes the situation by restoring previous behavior where
null player names are passed as null. It is tempting to allow the
value object 'PlayerName' to simply hold a null (so we could check
if the contained value is null rather than checking teh value object),
but for now leaning towards future-behavior where we would
enforce non-null values to be held by PlayerName.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[x] Bug fix:  
[ ] Other:   

## Testing

[x] Manually tested
[ ] No testing done

## Screens Shots
Current master has no select player buttons:
Screenshot from 2019-09-27 11-09-58

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/ClientSetupPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/ClientSetupPanel.java
@@ -68,7 +68,10 @@ public class ClientSetupPanel extends SetupPanel {
               name, playerNamesAndAlliancesInTurnOrder.get(name), enabledPlayers.get(name));
       playerRows.add(playerRow);
       SwingUtilities.invokeLater(
-          () -> playerRow.update(PlayerName.of(players.get(name)), disableable.contains(name)));
+          () ->
+              playerRow.update(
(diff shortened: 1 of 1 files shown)


DanVanAtta:
@RoiEXLab before we release, we should get validation in place. non-null, no new lines, and less than the Db lenght are minimum. The only complication is how to combine with existing player validation. I'm thinking `PlayerNameValidator` should be made into a subclass of `PlayerName` and renamed to `Validator`. Then the 'of' method could assert no validation errors and we can still re-use the validation without needing to instantiate a `PlayerName` object.

The domain terminology is also starting to really clash, TripleA started with 'Player' to represent factions or nations, then when multiplayer was introduced, an "actual" player was called a "node" (which is not good domain modeling, nobody talks about nodes)

PR #5289, opened 2019-09-27
Refactor: Split Chat from Communication Mechanism

DanVanAtta:
Main thrust of this update is to split the communication mechanism
from chat. This is not a complete update but moves a long way, future
updates should be able to drop in new 'ChatTransmitter' implementations
to allow for chat to be backed by a different transport technology. In other 
words, this update attempts to 'hide' messengers interaction from Chat

Notable updates:
- IP address of chatters is (more) hidden

This update also tries to hide/remove 'node' from Chat usage. Of
significant, the front-end should have fewer if any node references.
This was a security issue as node carries IP address with it, before
this update the IP addresses of all chat participants is broadcasted
to all other chat clients.

- StatusController squashed into ChatController

Mainly a simplification change, reduces number of remotes.

- ChatParticpant data object introduced, represents a player name
  plus tag (moderator or not moderator).

Change summary:
- Remove INode from Chat, use ChatParticipant instead
- Move 'messengers' interaction out of Chat to be behind 'ChatTransmitter'
- Squash 'StatusController' into 'ChatController'
- Update status data representation, instead of parallel maps
  of "player -> tag" and "player -> status", create a data
  object "ChatParticipant" that is a combination of name and
  tag, and change data structure to a single map "chatparticipant -> status"
- Update headless chat 'setChat' method to be an unsupported operation.
  Headless chat is instantiated just once and it appears the cancel
  path is never invoked to unset the chat. Of note, the 'history' copy
  that was done in the 'setChat' method can be removed as well since
  we only 'set' with a new chat object, once, the history is always
  empty.
- Use empty status to represent no status instead of null.
- Remove chat version numbering, was not the best guard for ordering
  and is strictly not necessary
- Remove chat latch and synchronization, latch is unnecessary and
  synchronization is not needed or would have wrapped what are thread-safe
  or atomic operations.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[x] Manually tested
[ ] No testing done

- hosted local network game, set status and chatted
- connected to a local lobby, verified boot/ban, chat and status setting
- connected to a local bot with 2 headed game-clients, chatted, set status, started a game, chatted-in game, set status,  and disconnected with one account


game-core/src/main/java/games/strategy/engine/chat/Chat.java
```diff
@@ -155,140 +51,141 @@ public void ping() {}
...
-    final LinkedHashSet<String> current = getTagText(tag);
-    notesMap.put(node, current);
+    chatHistory.add(new ChatMessage(message, from));
+    for (final IChatListener listener : listeners) {
+      listener.addMessage(message, from);
+    }
+    // limit the number of messages in our history.
+    while (chatHistory.size() > 1000) {
+      chatHistory.remove(0);
+    }
```
RoiEXLab: I know that this is kept code, but when `chatHistory` is a `CopyOnWriteArrayList`, `remove(0)` will be a super expensive call.
I'd recommend using guavas [`EvictingQueue`](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/EvictingQueue.java), it should automatically remove the elements, but is a cyclic buffer.
If thread-safety is a requirement, using `Collections#synchronizedCollection` should do the trick.

game-core/src/main/java/games/strategy/engine/chat/HeadlessChat.java
```diff
@@ -44,25 +42,11 @@ public Chat getChat() {
...
-      synchronized (this.chat.getMutex()) {
-        allText = new StringBuilder();
-        for (final ChatMessage message : this.chat.getChatHistory()) {
-          addChatMessage(message.getMessage(), message.getFrom().getValue());
-        }
-      }
-    } else {
-      updatePlayerList(Collections.emptyList());
-    }
+    throw new UnsupportedOperationException("Headless bots do not support resetting of chat");
```
RoiEXLab: Would you mind explaining the reasoning behind this change?

game-core/src/main/java/games/strategy/engine/chat/Chat.java
```diff
@@ -155,140 +51,141 @@ public void ping() {}
...
-        return;
-      }
-      final List<INode> playerNames = new ArrayList<>(nodes);
-      Collections.sort(playerNames);
-      for (final IChatListener listener : listeners) {
-        listener.updatePlayerList(playerNames);
-      }
+    final List<ChatParticipant> playerNames =
+        chatters.keySet().stream()
+            .sorted(Comparator.comparing(c -> c.getPlayerName().getValue()))
```
RoiEXLab: Is it important the list is sorted?

game-core/src/main/java/games/strategy/engine/chat/Chat.java
```diff
@@ -155,140 +51,141 @@ public void ping() {}
...
-    final IChatController controller = messengers.getRemoteChatController(chatName);
-    messengers.addChatChannelSubscriber(chatChannelSubscriber, chatChannelName);
-    final Tuple<Map<INode, Tag>, Long> init = controller.joinChat();
-    final Map<INode, Tag> chatters = init.getFirst();
-    nodes = new ArrayList<>(chatters.keySet());
-    chatInitVersion = init.getSecond();
-    latch.countDown();
-    assignNodeTags(chatters);
+    sentMessagesHistory = new SentMessagesHistory();
+    chatters = Optional.ofNullable(chatTransmitter.connect()).orElse(new HashMap<>());
```
RoiEXLab:
```suggestion
    chatters = Optional.ofNullable(chatTransmitter.connect()).orElseGet(HashMap::new);
```

game-core/src/main/java/games/strategy/engine/chat/Chat.java
```diff
@@ -1,150 +1,46 @@
...
-              nodes.remove(node);
-              notesMap.remove(node);
-              updateConnections();
-            }
-            for (final IChatListener listener : listeners) {
-              listener.addStatusMessage(node.getPlayerName() + " has left");
-            }
-          }
-        }
+  private final List<IChatListener> listeners = new CopyOnWriteArrayList<>();
```
RoiEXLab: For listeners this data structure is ideal.

game-core/src/main/java/games/strategy/engine/chat/Chat.java
```diff
@@ -155,140 +51,141 @@ public void ping() {}
...
-     * and will see any messages broadcasted by the server.
-     *
-     * 2. Call the init message on the server remote. Any add or join messages sent from the server
-     * will wait until we receive the init return value.
-     *
-     * 3. When we receive the init message response, initialize our state and release the latch
-     * so all pending messages will get processed. Messages may be ignored if the server
-     * version is incorrect. This all seems a lot more involved than it needs to be.
-     */
-    final IChatController controller = messengers.getRemoteChatController(chatName);
```
RoiEXLab: So you did actually modify the class with this weird constraints.
You have my full respect ^^

game-core/src/main/java/games/strategy/engine/chat/HeadlessChat.java
DanVanAtta: This method is never called in a headless context, the exception is here so we can be notified in case I missed something.

It was a progressive step to see that this method did nothing:
- the history copy was always a copy of an empty list
- update player list is a no-op.

Seems we had a copy/paste of this method and maintained it "blindly"

game-core/src/main/java/games/strategy/engine/chat/Chat.java
DanVanAtta: Thanks, I think one big thing I learned is "getRemote" returns a server endpoint and calls it. "getBroadcaster" invokes all client endpoints via the server.

The code behind this was tricky, I spent a few hours last night debugging a remote not found exception, it was kinda painful. On plus side, it all should now be a bit more hidden behind the transmitter.

game-core/src/main/java/games/strategy/engine/chat/Chat.java
DanVanAtta: Line 196, it was sorted previously, it might not be. It seems clients could be expecting players to be in sorted order.

game-core/src/main/java/games/strategy/engine/chat/Chat.java
RoiEXLab: Ok, I just double checked this actually works: It does!

DanVanAtta: Good suggestion: 7e84adcf9

game-core/src/main/java/games/strategy/engine/chat/ChatPlayerPanel.java
```diff
@@ -253,7 +255,7 @@ private String getDisplayString(final INode node) {
       }
```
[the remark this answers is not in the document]
DanVanAtta: I don't know, it's hard to decipher what is happening. Using magic constants like this is kinda bad practice. For now tried to stay focused on the API split, it was a bear to get that done, spent most of a day just staring at this code wondering how to un-knot it.

game-core/src/main/java/games/strategy/engine/chat/ChatPlayerPanel.java
```diff
@@ -253,7 +255,7 @@ private String getDisplayString(final INode node) {
...
       extra = extra + " (" + sb + ")";
     }
-    return node.getName() + extra;
+    return chatParticipant.getPlayerName() + extra;
```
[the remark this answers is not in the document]
DanVanAtta: I believe you're right, though we should try to break it up just slightly, it's a bit too heavy as one line. One logic operation per line is the rule of thumb that I think applies here.

game-core/src/main/java/games/strategy/engine/chat/JavaSocketChatTransmitter.java
```diff
@@ -0,0 +1,104 @@
...
+      }
+
+      @Override
+      public void ping() {}
+
+      @Override
+      public void statusChanged(final PlayerName playerName, final String status) {
+        chatClient.statusUpdated(playerName, status);
+      }
+    };
```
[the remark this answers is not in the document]
DanVanAtta: It does and does not. As an inner class, it stays bundled, so we have one logical unit that is interacting with messengers.

lobby/src/main/java/org/triplea/lobby/server/ModeratorController.java
```diff
@@ -83,29 +89,36 @@ private void assertUserIsAdmin() {
...
-                            () ->
-                                new IllegalStateException(
-                                    "Failed to find user: " + modNode.getName())))
-                .actionName(ModeratorAuditHistoryDao.AuditAction.BOOT_USER_FROM_LOBBY)
-                .actionTarget(node.getName())
-                .build());
+
+    serverMessenger.getNodes().stream()
+        .filter(n -> n.getPlayerName().equals(playerName))
+        .findAny()
```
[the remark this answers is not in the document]
DanVanAtta: Repetition count and size is not too bad, but I don't mind factoring out the common lines, sure: 89a70eb35

game-core/src/main/java/games/strategy/engine/chat/JavaSocketChatTransmitter.java
```diff
@@ -0,0 +1,104 @@
...
+        chatClient.participantAdded(chatParticipant);
+      }
+
+      @Override
+      public void speakerRemoved(final PlayerName playerName) {
+        chatClient.participantRemoved(playerName);
+      }
+
+      @Override
+      public void ping() {}
```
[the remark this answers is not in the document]
DanVanAtta: I was focused on getting the class split up and not dependent on Messengers, it was a pretty difficult update and did not go smoothly the first 2 times I tried it. It would be nice to see the method removed for the simplification, but without contrary evidence we should assume it's needed.

game-core/src/main/java/games/strategy/engine/chat/JavaSocketChatTransmitter.java
```diff
@@ -0,0 +1,104 @@
...
+      }
+
+      @Override
+      public void slapOccurred(final PlayerName slappedPlayer) {
+        final PlayerName slapper = MessageContext.getSender().getPlayerName();
+        if (slappedPlayer.equals(playerName)) {
+          chatClient.slappedBy(slapper);
+        } else {
+          chatClient.eventMessage(slappedPlayer + " was slapped by " + slapper);
+        }
```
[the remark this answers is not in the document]
DanVanAtta: Which two interfaces specifically? If referring to chat client and chat channel, i think the answer might be no. Chat channel has dependencies to the Java NIO socket mechanism. ChatClient on the other hand we can potentially have multiple, a second example that I hope to have come along would be HttpChatClient

game-core/src/main/java/games/strategy/engine/chat/HeadlessChat.java
DanVanAtta: Would be great to move out of the interface, beyond the mutability aspect. It's risky to make too many changes, this is a pretty brittle are of the code, I've a bad taste in my mouth after spending 2 hours debugging a remote not found error before this was ready to submit. 

Only question I think is who follows up, let me know if that is a refactor you'd like to work on.

DanVanAtta:
@RoiEXLab added a few updates, please resolve anything that is not still outstanding. Really appreciate the review!

game-core/src/main/java/games/strategy/engine/chat/JavaSocketChatTransmitter.java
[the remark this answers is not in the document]
RoiEXLab: Thanks for clarifying 👍

game-core/src/main/java/games/strategy/engine/chat/HeadlessChat.java
RoiEXLab: Sure, I can create a PR for that

game-core/src/main/java/games/strategy/engine/chat/ChatPlayerPanel.java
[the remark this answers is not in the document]
RoiEXLab: I don't know either, but if you ask me I'd drop this kind of character stripping

RoiEXLab:
Ok, you addressed my most important points and most of the remaining stuff is either potentially risky or out of scope.
I'll create follow-up PRs for a couple of things and would kindly ask you to do the same for the other comments that aren't straight forward like checking if ping is actually needed etc.

RoiEXLab:
Ok @DanVanAtta I created PRs for all of the easy-to-fix PRs, let me know your thoughts on the remaining two points.
I'd appreciate if you would try to have a look at them.

game-core/src/main/java/games/strategy/engine/chat/JavaSocketChatTransmitter.java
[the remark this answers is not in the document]
DanVanAtta: Did some research on this, `ping()` was introduced in: 0c9f4f675679cc2618bc6a93d1bd2f7dcecd69c0, with comment:

> Creating a simple Ping method inside of the Chat Controller, that will ping all computers once every minute, hopefully improving everyone's connections. (Sean Bridges & veqryn)


We do set the TCP keep-alive option:
>     socket.setKeepAlive(true);

This SO seems to mention some of the gotchyas: https://stackoverflow.com/questions/1480236/does-a-tcp-socket-connection-have-a-keep-alive

> Gotchas
2 Hour Default

> Since the process doesn't start until a connection has been idle for two hours by default, stale TCP connections can linger for a very long time before being pruned. This can be especially harmful for expensive connections such as database connections.
Keep-Alive is Optional

> According to RFC 1122 4.2.3.6, responding to and/or relaying TCP Keep-Alive packets is optional

--------------

If lobby chat is converted to http, the keep-alive mechanism there will certainly send packets frequently enough to avoid any of the above limits. In-game chats should likely get traffic, at least if a game is started. I would be concerned a about 'staging' screens where players queue up and do a long joining period, eg: someone hosts and then over the course of hours players slowly join up without any conversation happening. 

On the other hand, TCP keep-alive is supposed to keep the connection open without us manually sending messages. I suspect we could probably drop this feature, but I would be afraid we'd find out the hard way where/how it was needed.

@RoiEXLab any thoughts based on that research?

game-core/src/main/java/games/strategy/engine/chat/Chat.java
```diff
@@ -155,140 +51,141 @@ public void ping() {}
...
     }
-    final LinkedHashSet<String> tagText = new LinkedHashSet<>();
-    if (tag == Tag.MODERATOR) {
-      tagText.add(TAG_MODERATOR);
+    chatters.put(chatParticipant, "");
+    updateConnections();
+    for (final IChatListener listener : listeners) {
+      listener.addStatusMessage(chatParticipant.getPlayerName() + " has joined");
+      if (chatSoundProfile == ChatSoundProfile.GAME_CHATROOM) {
+        ClipPlayer.play(SoundPath.CLIP_CHAT_JOIN_GAME);
```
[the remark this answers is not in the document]
DanVanAtta: HeadlessChat probably does not need any concept of a sound channel at all, any sounds it plays are a no-op anyways. I think I might be better off trying to stay focused on the server migration efforts and next release. Hopefully this update makes future updates easier and isolates the particularly brittle parts so we can update things with more confidence.

PR #5292, opened 2019-09-28
Flaky coverage

RoiEXLab:
Another iteration that tries to get rid of flaky coverage

**I'd recommend reviewing with whitespace changes off**

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[x] No manual testing done
[ ] Manually tested

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/odds/calculator/DummyPlayer.java
+++ b/game-core/src/main/java/games/strategy/triplea/odds/calculator/DummyPlayer.java
@@ -1,5 +1,6 @@
 package games.strategy.triplea.odds.calculator;
 
+import com.google.common.annotations.VisibleForTesting;
 import games.strategy.engine.data.GameData;
 import games.strategy.engine.data.PlayerId;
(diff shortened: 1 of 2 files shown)


RoiEXLab:
Of course Decoder wants attention now as well

DanVanAtta:
Looks fine :+1: 
 Kinda curious where/what the inconsistent coverage was, I did not spot it.

RoiEXLab:
Well I'm really relying on codecov here to see what kind of coverage changes all of the time.
I assume that the flaky coverage is caused by non-deterministic behaviour of tests that test battles: Sometimes dice roll differently, so specific code paths may never be triggered

DanVanAtta:
I think I realized late, after you mention it, one way to fix that is to test the code-path that is sometimes tested explicitly, in which case it is always tested. Not a bad strategy  :+1:

PR #5293, opened 2019-09-29
Split Update account into update password and update email

DanVanAtta:
(1) Split up the create or update account panel into two, one for
   updating password, another for updating email. To do this
   we update the lobby menu to have the two selection options.
   There is already an update password panel, so just needed
   to create a new update email panel.

(2) Re-wire backend of update email to go to http server instead
  of lobby. Allows for related lobby controller methods to be dropped.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[ ] No manual testing done
[x] Manually tested

- Logged in, changed email, verified email was updated in DB
- Logged in, changed password, logged out and then back in with updated password.
  - verified 'remember' password remembers the updated password


game-core/src/main/java/games/strategy/engine/framework/startup/mc/SetupPanelModel.java
```diff
@@ -119,39 +120,44 @@ public void login() {
...
-                        () ->
-                            DialogBuilder.builder()
-                                .parent(lobbyFrame)
-                                .title("Password Not Updated")
-                                .errorMessage(
-                                    "Password not updated, you will need to request a new "
-                                        + "temporary password to log in again.")
-                                .showDialog());
+                try {
+                  passwordChanged =
```
RoiEXLab: Your thoughts on putting a `do {} while(!passwordChanged);` loop here?
Without this loop people have to manually re-click a button to get the same dialog. Just a thought, popups can also by super annoying if users want to completely cancel the whole login procedure anyways

RoiEXLab:
btw. I appreciate that more and more logic is moved to the HTTP client 👍 
Soon enough we'll be able to drop our custom RSA logic.

game-core/src/main/java/games/strategy/engine/framework/startup/mc/SetupPanelModel.java
DanVanAtta: Few thoughts:
- This pop-up is shown on login after using a temp password, so there is no button to be clicked. That does make me think the message text should be updated to tell the user to use the "account > change password" feature before they log out.
- If there are any persistent errors on reset password, the do-while loop would effectively become infinite.
- The recurring pop-up can be a bit annoying, instead it would be a better UX to disallow for the window to be closed or cancelled until the 'do' condition is fulfilled. In this case we could add a prompt like "are you really sure you want to close this window?"

DanVanAtta:
Thanks for the review @RoiEXLab . The custom RSA logic is a part of the iceberg, eventaully all NIO socket communication will be between client and game server, or client and headless, it should be much easier to update and reason about at that point. I think the next biggest focus we should have it to try and convert headless to just be a network Relay. We almost have the code for it with the 'chatBroadcaster', in essense when you host, if you do not have a public IP address then you instead connect to the relay and the IP address of the relay is advertised as your hosting address. The relay then just sends messages back and forth. If we achieve that, then anyone can host, we reduce the number of needed bots as we are then just IO bound and not memory bound (so we save about 50% of total operational costs), *and* *ALL* of the headless code can be deleted!

DanVanAtta:
I'll work on updating the error message text for the "your temporary password is no longer valid" and update and merge.

RoiEXLab:
Well I wouldn't say that all headless code can be deleted, because it shares a lot of classes with the headed server code, but yes removing all the code that is specifically for headless should be a goal.
There are a couple of things to consider though:
1. Tournaments might become trickier to organise, they require a trusted server to prevent faking dice rolls. Ideally UI would just be a layer on top of the logic that could simply be turned off, so we could still occasionally have competitive instances or come up with a different way to achieve the same.
2. We should consider security: If the real host can be anyone we need to double check that no malicious data can be sent in a way that compromises the users system.
Specifically I'm thinking about serialised objects that may be altered in a way that violates constraints checked in the constructor or worse stuff that may lead to DoS or remote code execution, so we need to be extra careful to not just blindly accept everything.

DanVanAtta:
>     Tournaments might become trickier to organise, they require a trusted server to prevent faking dice rolls. Ideally UI would just be a layer on top of the logic that could simply be turned off, so we could still occasionally have competitive instances or come up with a different way to achieve the same.

Perhaps, I think we can work it out. For p2p, using hmac's could be used to verify the system has not changed. The way 'trust' is currently established for dice roller is for example there is an email sent with every dice roll. That can be matched up with the dice rolls in the game history to verify.

To another extent, anything that requires trust, for example dice rolls, we could potentially create endpoints for those specific functionalities. I do not think we need a heavy 'bot' server to actually have verified games.

>  If the real host can be anyone we need to double check that no malicious data can be sent in a way that compromises the users system.

Isn't that a problem with today's hosting software? Not everyone joins bot games to launch servers, if we have such vulnerabilities, they would be existent today and/or with a relay.

RoiEXLab:
Yes, this is probably already a problem today, but no reason to just ignore it, just something to keep in mind

PR #5294, opened 2019-09-29
ChatModel simplification

RoiEXLab:
Follow-up to #5289

This PR moves methods out of the `ChatModel` interface that clearly don't belong there.
I must admit that the "workaround" for the `ChatModel#getViewComponent` method is not really clean, but `HeadlessChat` really shouldn't depend on swing either.

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[ ] No manual testing done
[x] Manually tested

I verified I could open the usual screens where a chat is displayed.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/chat/ChatPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/ChatPanel.java
@@ -60,7 +60,6 @@ public class ChatPanel extends JPanel implements ChatModel {
     return chatMessagePanel.getAllText();
   }
 
-  @Override
   public void setChat(final Chat chat) {
     chatMessagePanel.setChat(chat);
     chatPlayerPanel.setChat(chat);
--- a/game-core/src/main/java/games/strategy/engine/chat/ChatPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/ChatPanel.java
@@ -3,9 +3,7 @@ package games.strategy.engine.chat;
 import games.strategy.engine.chat.Chat.ChatSoundProfile;
 import games.strategy.net.Messengers;
 import java.awt.BorderLayout;
-import java.awt.Component;
 import java.awt.Dimension;
-import java.util.Optional;
 import javax.swing.DefaultListCellRenderer;
 import javax.swing.JPanel;
 import javax.swing.JSplitPane;
@@ -79,10 +77,5 @@ public class ChatPanel extends JPanel implements ChatModel {
 
   public ChatMessagePanel getChatMessagePanel() {
     return chatMessagePanel;
-  }
-
-  @Override
-  public Optional<Component> getViewComponent() {
-    return Optional.of(this);
   }
 }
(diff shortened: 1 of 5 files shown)


DanVanAtta:
I think it would be good if we could find a way to avoid `instanceof` checks. 

Looking to that, I think there are three things we could look into:

(1) Do we even need the `((ChatPanel) chatModel).setChat(null)` method call at all? If it can't be outright removed, perhaps the chat would lose scope if we changed that to `chatModel = null`


(2) If we need the 'setChat' method, then perhaps we can create a second interface `HeadedChatModel` that extends `ChatModel` and then has our `setChat` method, assuming we can't just remove that method. Then, at least the `instanceof` checks could be verifying an instance of an interface type rather than concrete implementation.

(3) Parallel variables. I think this is probably my utmost preference for cleaner API's and avoiding instanceof checks. The idea is instead of doing an instance of check for example, just pass the concrete type as a second variable. EG: change this constructor API:
```
  MainPanel(
      final GameSelectorPanel gameSelectorPanel,
      final Consumer<MainPanel> launchAction,
      @Nullable final ChatModel chatModel,
      final Runnable cancelAction) {
```

To:
```
  MainPanel(
      final GameSelectorPanel gameSelectorPanel,
      final Consumer<MainPanel> launchAction,
      @Nullable final ChatModel chatModel,
      @Nullable final ChatPanel chatPanel,
      final Runnable cancelAction) {
```

ChatModel should really not be aware of the UI component, the relationship is more ideally 'the other way around'. The control flow with that kind of structure would be more explicit and not rely on an instanceof check (at the cost of passing around optionally null values, but IMO that is well compensated for by removing the instanceof).  

@RoiEXLab what do you think of those ideas?

RoiEXLab:
Yeah was thinking about option 1 as well.
I'll have a look tomorrow to see what is feasible.

RoiEXLab:
@DanVanAtta I wasn't able to find a good solution for the component change.
It could be very well the case that I'm missing something obvious, but the abstraction is a couple layers deep.

Basically the problem comes down to the first problem I solved for the HeadlessChat class. Because ServerModel can't guarantee the type is going to be ChatPanel (even though it always should be for headed servers, which is the only case we need to consider here) there's no way to pass the more specific type through to main panel without casting at some point.
Also not all "SetupPanels" have a ChatModel obviously (like the local game launcher) so thats another thing to consider.

I'd really appreciate if you could have a look, maybe you're able to see something I missed, or have an entirely different idea that enhances the interface structure as a whole

DanVanAtta:
Hmm, I'm not sure if the abstraction is necessarily a couple layers deep, but the details are leaked a few layers deep, almost a lack of abstraction. IMO keep iterating and removing dependencies. No awesome good suggetions for you at the moment I regret.

PR #5295, opened 2019-09-29
Simplify chat truncation

RoiEXLab:
Noticed this during #5294

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[x] No manual testing done
[ ] Manually tested

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/chat/HeadlessChat.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/HeadlessChat.java
@@ -73,12 +73,6 @@ public class HeadlessChat implements IChatListener, ChatModel {
         .start();
   }
 
-  private void trimLengthIfNecessary() {
-    if (allText.length() > MAX_LENGTH) {
-      allText.delete(0, MAX_LENGTH / 2);
-    }
-  }
-
   private void addChatMessage(final String originalMessage, final String from) {
     trimLengthIfNecessary();
 
@@ -92,6 +86,12 @@ public class HeadlessChat implements IChatListener, ChatModel {
   public void addStatusMessage(final String message) {
     trimLengthIfNecessary();
     allText.append("--- ").append(message).append(" ---\n");
+  }
+
+  private void trimLengthIfNecessary() {
+    if (allText.length() > MAX_LENGTH) {
+      allText.delete(0, MAX_LENGTH / 2);
+    }
   }
 
   @Override
--- a/game-core/src/main/java/games/strategy/engine/chat/HeadlessChat.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/HeadlessChat.java
@@ -17,7 +17,7 @@ public class HeadlessChat implements IChatListener, ChatModel {
   // roughly 1000 chat messages
   private static final int MAX_LENGTH = 1000 * 200;
   private Chat chat;
-  private StringBuilder allText = new StringBuilder();
+  private final StringBuilder allText = new StringBuilder();
   private final ChatFloodControl floodControl = new ChatFloodControl();
 
   public HeadlessChat(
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/engine/chat/HeadlessChat.java
```diff
@@ -73,11 +73,14 @@ public void addMessageWithSound(final String message, final PlayerName from, fin
         .start();
   }
 
-  private void addChatMessage(final String originalMessage, final String from) {
-    final String currentAllText = allText.toString();
-    if (currentAllText.length() > MAX_LENGTH) {
-      allText = new StringBuilder(currentAllText.substring(MAX_LENGTH / 2));
+  private void trimLengthIfNecessary() {
```
DanVanAtta: nit: method ordering is 'wrong' on this one. It's interesting to see a lot of classes where updates occurred by extracting a method, and instead of placing it below, it was easier to place it above. Reading the code gets kinda interesting as you keep scrolling up to find the next method in the method call.

game-core/src/main/java/games/strategy/engine/chat/HeadlessChat.java
```diff
@@ -73,11 +73,14 @@ public void addMessageWithSound(final String message, final PlayerName from, fin
...
+    if (allText.length() > MAX_LENGTH) {
+      allText.delete(0, MAX_LENGTH / 2);
```
DanVanAtta: Do you understand where the divide by 2 comes in? This one kinda confused me, if the length is over max, why are we dividing by 2 and not just truncating?

game-core/src/main/java/games/strategy/engine/chat/HeadlessChat.java
```diff
@@ -17,7 +17,7 @@
   // roughly 1000 chat messages
   private static final int MAX_LENGTH = 1000 * 200;
   private Chat chat;
-  private StringBuilder allText = new StringBuilder();
+  private final StringBuilder allText = new StringBuilder();
   private final ChatFloodControl floodControl = new ChatFloodControl();
 
   public HeadlessChat(
```
DanVanAtta: We may want to investigate where/how messages get into HeadlessChat and if they should be truncated at the chat controller level. I think the problem is that we use a 'broadcaster' for chat instead of a remote. It might be better for chat to be sent by getting a remote, that would only call a method on chat controller. Controller would then use a broadcaster to relay the message. In that way, instead of clients blasting out messages through the controller to all other clients, we could have some centralized code. What is otherwise happening is a message that is too long is being broadcast and then truncated server side.

I'm thining about this also in the context of a headlessChat, we likely should not need any truncation code here at all, either becasue the controller has done this work for us, or, because it's a bot, it's just not needed.

game-core/src/main/java/games/strategy/engine/chat/HeadlessChat.java
RoiEXLab: I believe this "divide by 2" is to make sure that the string builder doesn't have to call delete again when the next message comes in, similar to how usually arraylist implementations double their allocated space whenever the underlying array gets full to maintain O(n) on average.

This shouldn't be a big deal now that the deletion operation isn't that expensive anymore, but I wouldn't bet on it

game-core/src/main/java/games/strategy/engine/chat/HeadlessChat.java
RoiEXLab: I believe this is supposed to be a buffer for chat messages. Once the storage gets full, the code deletes the oldest half of all messages and therefore regains half of the space again.
There's probably a better way to achieve that using `EvictingQueue` or something, but this is probably outside of the scope of this PR.

DanVanAtta: Okay. Kinda convoluted IMO, the 'allText' is only here as a buffer for when a moderator requests the chat log of a bot. I think the the coding here lost vision of its intent/purpose

PR #5296, opened 2019-09-29
Simplify player status string building

RoiEXLab:
Follow-up to #5289

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[x] No manual testing done
[ ] Manually tested

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/chat/ChatPlayerPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/ChatPlayerPanel.java
@@ -1,5 +1,6 @@
 package games.strategy.engine.chat;
 
+import com.google.common.base.Ascii;
 import games.strategy.engine.lobby.PlayerName;
 import java.awt.BorderLayout;
 import java.awt.Dimension;
@@ -239,16 +240,10 @@ public class ChatPlayerPanel extends JPanel implements IChatListener {
     }
 
     final String extra = chatParticipant.isModerator() ? " " + TAG_MODERATOR : "";
-    final String status = chat.getStatus(chatParticipant.getPlayerName());
+    final String status = Ascii.truncate(chat.getStatus(chatParticipant.getPlayerName()), 25, "");
 
     final String suffix =
-        status.isEmpty()
-            ? ""
-            : " ("
-                + status
-                    .substring(0, Math.min(status.length(), 25))
-                    .replaceAll("[\\u0300-\\u036F]", "")
-                + ")";
+        status.isEmpty() ? "" : " (" + status.replaceAll("[\\u0300-\\u036F]", "") + ")";
     return chatParticipant.getPlayerName() + extra + suffix;
   }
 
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/engine/chat/ChatPlayerPanel.java
```diff
@@ -238,24 +239,11 @@ private String getDisplayString(final ChatParticipant chatParticipant) {
...
-    final StringBuilder sb = new StringBuilder();
-    if (status.length() > 0) {
-      if (status.length() > 25) {
-        status = status.substring(0, 25);
-      }
-      for (int i = 0; i < status.length(); i++) {
-        final char c = status.charAt(i);
-        if (c >= '\u0300' && c <= '\u036F') { // skip combining characters
-          continue;
-        }
```
RoiEXLab: Note that this PR drops this replacement.
However if there are any objections to this, I can just revert the latest commit to get a version that keeps this replacement.

DanVanAtta: We should understand it before dropping it; hopefully there was a good reason it was added to begin with. It's really annoying to see code like this, at least there was some commentary noted to help, but the exact logic is elusive.

game-core/src/main/java/games/strategy/engine/chat/ChatPlayerPanel.java
```diff
@@ -238,24 +239,11 @@ private String getDisplayString(final ChatParticipant chatParticipant) {
       return "";
     }
 
-    String extra = chatParticipant.isModerator() ? " " + TAG_MODERATOR : "";
-    String status = chat.getStatus(chatParticipant.getPlayerName());
+    final String extra = chatParticipant.isModerator() ? " " + TAG_MODERATOR : "";
+    final String status = Ascii.truncate(chat.getStatus(chatParticipant.getPlayerName()), 25, "");
```
DanVanAtta: Have you had any thoughts whether chat truncation be done at the client, just the server, or both server and client?

RoiEXLab: I believe @ssoloff dealt with sanitation before, maybe he knows something about it

game-core/src/main/java/games/strategy/engine/chat/ChatPlayerPanel.java
RoiEXLab: Well I checked the unicode table of this particular range and it's like the comment states:
All of those characters are "combination chars" that give a ^ to a character. According to wikipedia those are called [diacritic signs](https://en.wikipedia.org/wiki/Diacritic) as a whole.

So removing those should really not have a noticeable effect.

game-core/src/main/java/games/strategy/engine/chat/ChatPlayerPanel.java
DanVanAtta: Sorry, got a bit rhetorical. The truncation probably should be done in chat controller, unfortunately chat messages are sent via broad cast, so they are blindly forwarded.

PR #5297, opened 2019-09-29
Use ternary expression

RoiEXLab:
Follow-up to #5289

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[x] No manual testing done
[ ] Manually tested

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/chat/PlayerChatRenderer.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/PlayerChatRenderer.java
@@ -8,6 +8,7 @@ import games.strategy.engine.lobby.PlayerName;
 import games.strategy.net.INode;
 import games.strategy.triplea.ui.UiContext;
 import java.awt.Component;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -29,7 +30,6 @@ import javax.swing.SwingConstants;
  */
 public class PlayerChatRenderer extends DefaultListCellRenderer {
   private static final long serialVersionUID = -8195565028281374498L;
-  private final int maxIconCounter;
   private final Map<String, List<Icon>> iconMap = new HashMap<>();
   private final Map<String, Set<String>> playerMap = new HashMap<>();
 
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/engine/chat/PlayerChatRenderer.java
```diff
@@ -99,16 +99,12 @@ private void setIconMap() {
                 .collect(Collectors.toList());
         maxIconCounter = Math.max(maxIconCounter, icons.size());
         playerMap.put(playerNode.getPlayerName().getValue(), players);
-        if (uiContext == null) {
-          iconMap.put(playerNode.getPlayerName().getValue(), null);
-        } else {
-          iconMap.put(playerNode.getPlayerName().getValue(), icons);
-        }
+        iconMap.put(playerNode.getPlayerName().getValue(), uiContext == null ? null : icons);
```
DanVanAtta: One odd thing, if `uiContext` and icons have a relationship here that is not direct or necessarily clear. EG: what happens to `icons` when uiContext is null? It seems like there is a potential for computations and then it's all ignored later when some *other* variable is null.

Would it be yet cleaner if `icons` were assigned to `null` when `uiContext` is `null`?

RoiEXLab: Good point, will simplify something else even 👍

RoiEXLab:
Ok @DanVanAtta I altered the code a bit to avoid filling the map with null values.
Instead the entries are now just omitted if the icon factory is null.

While writing this I realized that the factory can never be null in the existing code, so I will just turn this into a precondition

RoiEXLab:
Ok, I restructured the code a little bit more.
In theory the for loop could be converted into a stream expression, but given that this would mean 3 different streams, we'd end up with more code so I decided not to do it.

game-core/src/main/java/games/strategy/engine/chat/PlayerChatRenderer.java
```diff
@@ -74,41 +89,16 @@ private String getNodeLabelWithPlayers(final PlayerName playerName) {
...
-          iconMap.put(playerNode.getPlayerName().getValue(), null);
-        } else {
-          iconMap.put(playerNode.getPlayerName().getValue(), icons);
-        }
-      }
-    }
   }
 
-  public int getMaxIconCounter() {
-    return maxIconCounter;
```
RoiEXLab: I double checked this is only called once per game instance, so this change is ok performance wise

DanVanAtta:
Certainly simpler, good work @RoiEXLab  :+1:

PR #5302, opened 2019-09-30
Refactor: HttpController base class and ErrorReportController consistency

DanVanAtta:
The goal of this refactor is stronger consistency across http controller and clients. Very notably, I keep forgetting to add the `@Path("")` annotation to new controllers, the common base-class I hope will help avoid that error.

Change summary:

1) Add a base class for HttpControllers called HttpController. Base classes
tend to be anti-patterns, except it seems to work well when the base class is
little more than the common annotations. It's easy to forget to add the @Path("")
annotation, by offering a common class to extend, with luck that detail will be
easier to include and also the class will provide a location for documentation
on how to build an endpoint.

2) ErrorUpload* was inconsistently named from its controller and was renamed ErrorReport*
It was also missing a test, the rename and addition of test case brings it in-line to
the standards of the other controller+client classes.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[x] No manual testing done
[ ] Manually tested


http-server/src/main/java/org/triplea/server/error/reporting/ErrorReportController.java
```diff
@@ -9,38 +9,37 @@
...
+  public ErrorReportResponse uploadErrorReport(
+      @Context final HttpServletRequest request, final ErrorReportRequest errorReport) {
 
     if (errorReport.getBody() == null || errorReport.getTitle() == null) {
       throw new IllegalArgumentException("Missing error report body and/or title");
     }
 
     return errorReportIngestion.apply(
-        ErrorReportRequest.builder()
+        org.triplea.server.error.reporting.ErrorReportRequest.builder()
```
RoiEXLab: Why are there 2 `ErrorReportRequest` classes?

http-server/src/main/java/org/triplea/server/error/reporting/ErrorReportRequest.java
```diff
@@ -4,13 +4,12 @@
...
 import lombok.Getter;
-import org.triplea.http.client.error.report.ErrorUploadRequest;
 
 /** Data object combining error report JSON and IP address from the incoming http request. */
 @Builder
 @Getter
 @EqualsAndHashCode
 public class ErrorReportRequest {
-  @Nonnull private final ErrorUploadRequest errorReport;
+  @Nonnull private final org.triplea.http.client.error.report.ErrorReportRequest errorReport;
```
RoiEXLab: Ok, this is confusing

http-server/src/main/java/org/triplea/server/error/reporting/ErrorReportController.java
DanVanAtta: Name collision, not always the end of the world, but the handling is so close to one another that I agree it's confusing. I resolved this by splitting up the second ErrorReportRequest, it was a value object holding IP and the original error report. It's now split up and the second value object deleted.

RoiEXLab:
👍
Note that the build is failing

PR #5304, opened 2019-09-30
Refactor: Move connectivity check to http

DanVanAtta:
Migrate connectivity test to http server from java socket server.
The connectivity check is the verification that kicks in after hosting
a game where the server tries to establish a 'reverse' connection
to the host to make sure the host is accessible from the internet.

This update moves that check to HTTP server.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[ ] No manual testing done
[x] Manually tested

- launched a local lobby with bot
- joined local lobby, hosted a game

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/LocalServerAvailabilityCheck.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/LocalServerAvailabilityCheck.java
@@ -1,9 +1,8 @@
 package games.strategy.engine.framework.startup.ui;
 
-import games.strategy.net.INode;
 import java.util.function.Consumer;
 import lombok.Builder;
-import org.triplea.lobby.common.ILobbyGameController;
+import org.triplea.http.client.lobby.game.ConnectivityCheckClient;
 import org.triplea.util.ExitStatus;
 
 /**
@@ -12,15 +11,20 @@ import org.triplea.util.ExitStatus;
  */
 @Builder
 public class LocalServerAvailabilityCheck {
-  private final ILobbyGameController controller;
-  private final INode localNode;
+  private final ConnectivityCheckClient connectivityCheckClient;
+  private final int localPort;
   private final Consumer<String> errorHandler;
 
+  /**
+   * Starts a thread that requests the server to do a 'reverse' connection back to the local client.
+   * This is to ensure that our local IP address is public facing and that other computers on the
+   * internet can establish connections to this (the local) host.
+   */
   public void run() {
     // if we lose our connection, then shutdown
     new Thread(
             () -> {
-              if (!controller.testGame(localNode)) {
+              if (!connectivityCheckClient.checkConnectivity(localPort)) {
                 // if the server cannot connect to us, then quit
                 errorHandler.accept(
                     "Your computer is not reachable from the internet.\n"
@@ -29,7 +33,7 @@ public class LocalServerAvailabilityCheck {
                         + "(The firewall exception must be updated every time a new version of "
                         + "TripleA comes out.)\n"
                         + "And that your Router is configured to send TCP traffic on port "
-                        + localNode.getPort()
+                        + localPort
                         + " to your local ip address.\n"
                         + "See 'How To Host...' in the help menu, at the top of the lobby "
                         + "screen.");
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/ServerSetupPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/ServerSetupPanel.java
@@ -1,6 +1,9 @@
 package games.strategy.engine.framework.startup.ui;
 
 import static games.strategy.engine.framework.CliProperties.LOBBY_HOST;
+import static games.strategy.engine.framework.CliProperties.LOBBY_HTTPS_PORT;
+import static games.strategy.engine.framework.CliProperties.LOBBY_PORT;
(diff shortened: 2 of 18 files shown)


game-core/src/main/java/games/strategy/engine/framework/startup/ui/ServerSetupPanel.java
```diff
@@ -100,11 +105,24 @@ private void createLobbyWatcher() {
...
+                  .localPort(model.getMessenger().getLocalNode().getPort())
                   .errorHandler(errorHandler)
-                  .localNode(model.getMessenger().getLocalNode())
                   .build()
                   .run();
+
+              System.clearProperty(LOBBY_HOST);
+              System.clearProperty(LOBBY_PORT);
+              System.clearProperty(LOBBY_HTTPS_PORT);
+              System.clearProperty(TRIPLEA_NAME);
```
RoiEXLab: Could you explain why this change was introduced?

game-core/src/test/java/games/strategy/engine/framework/startup/ui/InGameLobbyWatcherTest.java
```diff
@@ -46,15 +46,6 @@ void shouldReturnPrimaryValueWhenPrimaryValueSet() {
...
     }
 
-    @Test
-    void shouldClearPrimaryValueWhenPrimaryValueSet() {
-      givenPrimaryValueSet();
-
-      getLobbySystemProperty(KEY);
-
-      assertThat(System.getProperty(KEY), is(nullValue()));
-    }
```
RoiEXLab: What about reversing this test now that the clear statement is removed?

http-server/src/main/java/org/triplea/server/lobby/game/ConnectivityController.java
```diff
@@ -0,0 +1,42 @@
...
+  @RateLimited(
+      keys = {KeyPart.IP},
+      rates = {@Rate(limit = 10, duration = 1, timeUnit = TimeUnit.MINUTES)})
+  @RolesAllowed(UserRole.ANONYMOUS)
+  public boolean checkConnectivity(@Context final HttpServletRequest request, final Integer port) {
+    Preconditions.checkArgument(port > 0, "Port must be a positive number, was: " + port);
+    Preconditions.checkArgument(
+        port < Math.pow(2, 16), "Port must be less than max value (2^16), was: " + port);
+
+    return connectivityCheck.test(new InetSocketAddress(request.getRemoteAddr(), port));
```
RoiEXLab: I wonder what was your motivation was to move this check to the server?

game-core/src/main/java/games/strategy/engine/framework/startup/ui/LocalServerAvailabilityCheck.java
```diff
@@ -12,15 +11,20 @@
...
+  /**
+   * Starts a thread that requests the server to do a 'reverse' connection back to the local client.
+   * This is to ensure that our local IP address is public facing and that other computers on the
+   * internet can establish connections to this (the local) host.
+   */
   public void run() {
     // if we lose our connection, then shutdown
     new Thread(
             () -> {
-              if (!controller.testGame(localNode)) {
```
RoiEXLab: Hmm I'm not seeing that this method is removed, are there any usages?

DanVanAtta: Casualty of working on this from multiple machines, change was made but did not make it into this branch. Now removed: 949f6e9b8bd32f243cd5c89c839186c5f335cb12

game-core/src/main/java/games/strategy/engine/framework/startup/ui/ServerSetupPanel.java
DanVanAtta: Turns out INode.getAddress().getHostAddress()... etc.. returns the interface in addition to IP address. Hence this warning https://github.com/triplea-game/triplea/pull/5304/files#diff-5c823a8b563c30a1fc7db84682121912R26

That can't be used to create a valid URI, and we need to create one for the HttpLobbyClient. We do have the host address available, but lobby watcher has the side effect of clearing it (so that way on the next game launch the magic system property does not cause the game to launch in a hosting mode). But, we do need lobby watcher as it creates a connection which gives us an api key. So.. the clear is removed from lobby watcher and moved to after we create the http client so that we can get the host address from props and also get the API key.

game-core/src/test/java/games/strategy/engine/framework/startup/ui/InGameLobbyWatcherTest.java
DanVanAtta: It's okay to delete tests sometimes :smile: 
On another level, it's pretty special that the lobby watcher cleared a system property, I don't think we need to be explicit about it being left alone. Ideally we would move this check to code that is testing where we do now have the property clear, though that is a bit prohibitive/not worth it.

http-server/src/main/java/org/triplea/server/lobby/game/ConnectivityController.java
DanVanAtta: Removes a dependency on messengers, eventually with the goal of having no java NIO lobby left.

PR #5306, opened 2019-09-30
Refactor: Remove deprecated Observable from ImageScrollModel

DanVanAtta:
Convert 'Observable' to a listener.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[x] Code Cleanup or refactor
[ ] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[x] No manual testing done
[ ] Manually tested

### DIFF
--- a/game-core/src/main/java/games/strategy/ui/ImageScrollModel.java
+++ b/game-core/src/main/java/games/strategy/ui/ImageScrollModel.java
@@ -1,6 +1,7 @@
 package games.strategy.ui;
 
-import java.util.Observable;
+import java.util.ArrayList;
+import java.util.Collection;
 
 /**
  * Model for an ImageScroller. Generally one large view and one small view will be connected to the
@@ -8,7 +9,7 @@ import java.util.Observable;
  *
  * <p>notifies its observers when changes occur.
  */
-public class ImageScrollModel extends Observable {
+public class ImageScrollModel {
   @SuppressWarnings("checkstyle:MemberName")
   private int x;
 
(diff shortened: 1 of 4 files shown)


game-core/src/main/java/games/strategy/ui/ImageScrollModel.java
```diff
@@ -1,14 +1,15 @@
...
 /**
  * Model for an ImageScroller. Generally one large view and one small view will be connected to the
  * same model.
  *
  * <p>notifies its observers when changes occur.
  */
-public class ImageScrollModel extends Observable {
+public class ImageScrollModel {
   @SuppressWarnings("checkstyle:MemberName")
   private int x;
```
RoiEXLab: I wonder if this suppression is still required?
This class doesn't look like it's actually implementing serializeable, maybe subclasses are doing it?
Anyways we should be able to remove it anyways?

DanVanAtta: Still needed, I would caution though about comments in PRs that are on existing code or areas that are outside of the intended update, there are SO many things that could be updated. 
 
Screenshot from 2019-09-30 11-36-42

On another note, it's very notable that the listener relies on the listeners to have references to this class and/or are just relying on side effects. It's pretty dirty.. Ideally a listener is given input of what has changed as opposed to the tangle web of dependencies that we have :(

meanwhile, the observer deprecation warnings are clogging up the compiler output, makes it difficult to find actual errors or new problems, I feel like this is an easy enough type of fix to chip away at.

PR #5307, opened 2019-09-30
Update PULL_REQUEST_TEMPLATE.md

DanVanAtta:
Minor, remove spaces in checkbox area. The space is needed for an actual checkbox list, but we do not have that; otherwise the space means there is a small amount of manual formatting when submitting PRs.

 

## Functional Changes

[ ] New map or map update
[ ] New Feature
[ ] Feature Removal
[ ] Code Cleanup or refactor
[x] Configuration Change
[ ] Bug fix:  
[ ] Other:   

## Testing

[x] No manual testing done
[ ] Manually tested

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -8,21 +8,21 @@
 
 ## Functional Changes
 <!-- Put an X next any that apply -->
-[ ] New map or map update
-[ ] New Feature
-[ ] Feature update or enhancement
-[ ] Feature Removal
-[ ] Code Cleanup or refactor
-[ ] Configuration Change
-[ ] Bug fix:  <!-- Link to bug issue or forum post here -->
-[ ] Other:   <!-- Please specify -->
+[] New map or map update
+[] New Feature
+[] Feature update or enhancement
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,21 +8,21 @@
...
-[ ] Bug fix:  <!-- Link to bug issue or forum post here -->
-[ ] Other:   <!-- Please specify -->
+[] New map or map update
+[] New Feature
+[] Feature update or enhancement
+[] Feature Removal
+[] Code Cleanup or refactor
+[] Configuration Change
+[] Bug fix:  <!-- Link to bug issue or forum post here -->
+[] Other:   <!-- Please specify -->
```
RoiEXLab: Well using monospace fonts this makes a difference, because with this change the boxes no longer have the same width once you place an x in one of them, but ok

DanVanAtta: yeah, small hesitation there, but it's okay, the 'wide' checkbox should draw more attention to what is checked. Meanwhile we avoid having to do that annoying delete of the space :grin:

PR #5309, opened 2019-10-01
Add count for support bonus type

ron-murhammer:
Addresses: https://forums.triplea-game.org/topic/1560/stack-unit-support

A count can now be specified for a supportAttachment bonusType which specifies how many units can stack support onto the given support target (previously it was always just 1). The count can be any numerical value with 1 being the default but anything less than 1 will represent infinite stacking (0, -1, etc). Infinite stacking for strength support is not advised because you still have the dice sides limit it but infinite stacking for roll modifiers can make sense.

**New bonusType optional count parameter**
```
<option name="bonusType" value="Cover" count="3"/>
```

**Infantry receives support from up to 2 artillery**
```
        <attachment name="supportAttachmentArtillerygerman" attachTo="germanArtillery" javaClass="UnitSupportAttachment" type="unitType">
            <option name="unitType" value="germanInfantry:germanMarine:germanCombatEngineer:germanAlpineInfantry:germanParatrooper"/>
            <option name="side" value="offence"/>
            <option name="bonus" value="1"/>
            <option name="number" value="1"/>
            <option name="bonusType" value="ArtilleryBonus" count="2"/>
            <option name="impArtTech" value="false"/>
            <option name="players" value="Germany"/>
        </attachment>
```

**Infantry receives support from up to 3 static defenses (entrenchment, fortification)**
```
        <attachment name="supportAttachmentEntrenchmentgerman" attachTo="germanEntrenchment" javaClass="UnitSupportAttachment" type="unitType">
            <option name="unitType" value="germanInfantry:germanMarine:germanCombatEngineer:germanAlpineInfantry:germanParatrooper:italianInfantry:italianMarine:italianCombatEngineer:italianAlpineInfantry:italianParatrooper:japaneseInfantry:japaneseMarine:japaneseCombatEngineer:japaneseAlpineInfantry:japaneseParatrooper"/>
            <option name="side" value="defence"/>
            <option name="bonus" value="1"/>
            <option name="number" value="2"/>
            <option name="bonusType" value="FortBonus" count="3"/>
            <option name="impArtTech" value="false"/>
            <option name="players" value="Germany"/>
        </attachment>
        <attachment name="supportAttachmentFortificationgerman" attachTo="germanFortification" javaClass="UnitSupportAttachment" type="unitType">
            <option name="unitType" value="germanInfantry:germanMarine:germanCombatEngineer:germanAlpineInfantry:germanParatrooper:italianInfantry:italianMarine:italianCombatEngineer:italianAlpineInfantry:italianParatrooper:japaneseInfantry:japaneseMarine:japaneseCombatEngineer:japaneseAlpineInfantry:japaneseParatrooper"/>
            <option name="side" value="defence"/>
            <option name="bonus" value="2"/>
            <option name="bonusType" value="FortBonus" count="3"/>
            <option name="impArtTech" value="false"/>
            <option name="players" value="Germany"/>
        </attachment>
```

## Functional Changes

[] New map or map update
[x] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[x] Manually tested

Ran a few maps like revised on all AI to ensure no regressions. Edited TWW XML to test the new parameter and opened the BC to see the calculated power.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
@@ -37,6 +37,10 @@ public class UnitSupportAttachment extends DefaultAttachment {
     @Nonnull String name;
     @Nonnull Integer count;
 
+    public int getCount() {
+      return count < 0 ? Integer.MAX_VALUE : count;
+    }
+
     boolean isOldArtilleryRule() {
       return name.equals(Constants.OLD_ART_RULE_NAME);
     }
--- a/game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
@@ -18,7 +18,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
-import org.triplea.util.Tuple;
+import javax.annotation.Nonnull;
+import lombok.Value;
 
 /**
  * An attachment for instances of {@link UnitType} that defines properties for unit types that
@@ -26,6 +27,17 @@ import org.triplea.util.Tuple;
  */
 public class UnitSupportAttachment extends DefaultAttachment {
   private static final long serialVersionUID = -3015679930172496082L;
+
+  /** Type to represent name and count */
+  @Value
+  public static class BonusType {
+    @Nonnull String name;
+    @Nonnull Integer count;
+
+    boolean isOldArtilleryRule() {
+      return name.equals(Constants.OLD_ART_RULE_NAME);
+    }
+  }
 
   private Set<UnitType> unitType = null;
   private boolean offence = false;
@@ -38,7 +50,7 @@ public class UnitSupportAttachment extends DefaultAttachment {
   private int number = 0;
   private boolean allied = false;
   private boolean enemy = false;
-  private Tuple<Integer, String> bonusType = null;
+  private BonusType bonusType = null;
   private List<PlayerId> players = new ArrayList<>();
   private boolean impArtTech = false;
   // strings
@@ -218,16 +230,17 @@ public class UnitSupportAttachment extends DefaultAttachment {
   private void setBonusType(final String type) throws GameParseException {
     final String[] s = splitOnColon(type);
     if (s.length > 2) {
-      throw new GameParseException("bonusType can only have value and count" + thisErrorMsg());
+      throw new GameParseException(
+          "bonusType can only have value and count: " + type + thisErrorMsg());
     }
     if (s.length == 1) {
-      bonusType = Tuple.of(1, s[0]);
+      bonusType = new BonusType(s[0], 1);
     } else {
-      bonusType = Tuple.of(getInt(s[0]), s[1]);
+      bonusType = new BonusType(s[1], getInt(s[0]));
     }
   }
 
-  private void setBonusType(final Tuple<Integer, String> type) {
+  private void setBonusType(final BonusType type) {
     bonusType = type;
   }
 
@@ -314,7 +327,7 @@ public class UnitSupportAttachment extends DefaultAttachment {
     return offence;
   }
 
-  public Tuple<Integer, String> getBonusType() {
+  public BonusType getBonusType() {
     return bonusType;
   }
 
@@ -350,7 +363,7 @@ public class UnitSupportAttachment extends DefaultAttachment {
   private static Set<UnitType> getTargets(final GameData data) {
     Set<UnitType> types = null;
     for (final UnitSupportAttachment rule : get(data)) {
-      if (rule.getBonusType().getSecond().equals(Constants.OLD_ART_RULE_NAME)) {
+      if (rule.getBonusType().isOldArtilleryRule()) {
         types = rule.getUnitType();
         if (rule.getName().startsWith(Constants.SUPPORT_RULE_NAME_OLD_TEMP_FIRST)) {
           // remove it because it is a "first", which is just a temporary one made to hold target
@@ -376,8 +389,7 @@ public class UnitSupportAttachment extends DefaultAttachment {
 
   static void setOldSupportCount(final UnitType type, final GameData data, final String count) {
     for (final UnitSupportAttachment rule : get(data)) {
-      if (rule.getBonusType().getSecond().equals(Constants.OLD_ART_RULE_NAME)
-          && rule.getAttachedTo() == type) {
+      if (rule.getBonusType().isOldArtilleryRule() && rule.getAttachedTo() == type) {
         rule.setNumber(count);
       }
     }
@@ -386,7 +398,7 @@ public class UnitSupportAttachment extends DefaultAttachment {
   static void addTarget(final UnitType type, final GameData data) throws GameParseException {
     boolean first = true;
     for (final UnitSupportAttachment rule : get(data)) {
-      if (rule.getBonusType().getSecond().equals(Constants.OLD_ART_RULE_NAME)) {
+      if (rule.getBonusType().isOldArtilleryRule()) {
         rule.addUnitTypes(Collections.singleton(type));
         first = false;
       }
--- a/game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
@@ -9,6 +9,7 @@ import games.strategy.engine.data.MutableProperty;
 import games.strategy.engine.data.PlayerId;
 import games.strategy.engine.data.UnitType;
 import games.strategy.triplea.Constants;
+import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -30,7 +31,9 @@ public class UnitSupportAttachment extends DefaultAttachment {
 
   /** Type to represent name and count */
   @Value
-  public static class BonusType {
+  public static class BonusType implements Serializable {
+    private static final long serialVersionUID = -7445551357956238314L;
+
     @Nonnull String name;
     @Nonnull Integer count;
 
--- a/game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
@@ -18,6 +18,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
+import org.triplea.util.Tuple;
 
 /**
  * An attachment for instances of {@link UnitType} that defines properties for unit types that
@@ -37,7 +38,7 @@ public class UnitSupportAttachment extends DefaultAttachment {
   private int number = 0;
   private boolean allied = false;
   private boolean enemy = false;
-  private String bonusType = null;
+  private Tuple<Integer, String> bonusType = null;
   private List<PlayerId> players = new ArrayList<>();
   private boolean impArtTech = false;
   // strings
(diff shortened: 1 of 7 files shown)


game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
```diff
@@ -37,7 +38,7 @@
   private int number = 0;
   private boolean allied = false;
   private boolean enemy = false;
-  private String bonusType = null;
+  private Tuple<Integer, String> bonusType = null;
```
DanVanAtta: I think we're trying to phase out `Tuple` (@RoiEXLab (?))

This can be replaced pretty readily by a value object, eg:

```
@Value
private static class {
  @Nonnull Integer ...
  @Nonnull String ...
}
```

We get benefit that the values become labelled. It's not immediately clear what the 'string' or even the 'int' represent. Following up with naming/abstractions, should this variable still be called 'bonusType?'

game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
```diff
@@ -214,7 +215,19 @@ private void resetNumber() {
     number = 0;
   }
 
-  private void setBonusType(final String type) {
+  private void setBonusType(final String type) throws GameParseException {
+    final String[] s = splitOnColon(type);
+    if (s.length > 2) {
+      throw new GameParseException("bonusType can only have value and count" + thisErrorMsg());
```
DanVanAtta: It might not be clear which value was the offending value, does it make sense to print out the 'pre-split' value?

game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
```diff
@@ -301,7 +314,7 @@ public boolean getOffence() {
     return offence;
   }
 
-  public String getBonusType() {
+  public Tuple<Integer, String> getBonusType() {
```
DanVanAtta: I'd encourage using lombok here, in that case there would be no need to update the getter.

game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
```diff
@@ -337,7 +350,7 @@ static void addRule(final UnitType type, final GameData data, final boolean firs
   private static Set<UnitType> getTargets(final GameData data) {
     Set<UnitType> types = null;
     for (final UnitSupportAttachment rule : get(data)) {
-      if (rule.getBonusType().equals(Constants.OLD_ART_RULE_NAME)) {
+      if (rule.getBonusType().getSecond().equals(Constants.OLD_ART_RULE_NAME)) {
```
DanVanAtta: To note potential further benefit of a value object, a value object for bonus type could encapsulate this logic and remove redundancy, eg:
```if(rule.getBonusType().isOldArtilleryRule())```

File: game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java [hunk 5]
```diff
@@ -1069,27 +1069,33 @@ private static int getSupport(
...
-          if (supportersLeft != null) {
-            final Set<Unit> supporters = supportersLeft.keySet();
-            if (!supporters.isEmpty()) {
-              final Unit u = supporters.iterator().next();
-              supportUnitsLeft.get(rule).add(u, -1);
-              if (supportUnitsLeft.get(rule).getInt(u) <= 0) {
-                supportUnitsLeft.get(rule).removeKey(u);
-              }
-              unitSupportMap.computeIfAbsent(u, i -> new IntegerMap<>()).add(unit, rule.getBonus());
+          final int numSupportToApply =
```
DanVanAtta: Is this code covered by any existing tests? How safe it is it to update?

game-core/src/test/resources/Total_World_War_Dec1941.xml
```diff
@@ -16755,7 +16755,7 @@
       <option name="dice" value="strength"/>
       <option name="bonus" value="1"/>
       <option name="number" value="1"/>
-      <option name="bonusType" value="ArtilleryBonus"/>
+      <option name="bonusType" value="ArtilleryBonus" count="2"/>
```
DanVanAtta: :+1:

game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
```diff
@@ -214,7 +215,19 @@ private void resetNumber() {
     number = 0;
   }
 
-  private void setBonusType(final String type) {
+  private void setBonusType(final String type) throws GameParseException {
+    final String[] s = splitOnColon(type);
```
DanVanAtta: Just occurred to me, with a value object, we can have a constructor encapsulate this logic entirely, eg:

```
bonusType = new BonusType(type);
```
That opens up the parsing for easy testing and encapsulates the construction logic.

game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
ron-murhammer: Ok. I'm pretty neutral on tuple vs a small nested class. Though there is a pretty good amount of tuple use so probably should consider adding a note about that in coding standards and it would probably be better to have a pure refactoring PR to address at least the majority of it to push for consistency.

game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
ron-murhammer: Added.

game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
ron-murhammer: Updated.

game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
ron-murhammer: Updated.

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
ron-murhammer: Indirectly, yes. There are dice roll tests that check support being applied as well as some battle tests. I also now added a direct unit test for finding total power for support.

game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
DanVanAtta: One of the cases where commentary in PR did not make it into the code. Agree it should be documented, probably ideally as deprecated.

It's been an ongoing process to remove, note for example "Triplet" is removed. There is far too much code to remove in one go for it to be done in one PR. "Pair" objects are common, there are many in the code base, it's consistency within a class, package, project, and we never had that consistency in package or project.

For some time we have not been adding new dependencies.

DanVanAtta:
Understanding a bit better, I really wonder about '0' being infinite. I'm concerned it is a bit of a magic number. I wonder if there could be a case where an initial support count is '0', but say a technology can be made that would increase the value, for example, infantry with a 'combined arms tech' start to give a bonus to tanks.

I wonder if we can't just make this explicit, eg:
```
<option name="bonusType" value="Cover" unlimited="true"/>
```

@ron-murhammer what do you think? In part I was looking at how the '0' is translated to a max integer, which seems a bit odd that we go from 0 to a max, but then we do a min against that max. I wonder as well if that logic can't be simplified so that the 'max' is not more or less just ignored, but particularly avoiding the complexity where a zero is translated to another value that is then effectively a no-op.

ron-murhammer:
@DanVanAtta 0 and -1 or anything less than those are used as infinite for a number of other properties. If you think there is a case for setting to 0 then I'm open to having that essentially just mean it doesn't apply and only <0 is infinite.

DanVanAtta:
You seem opposed to an explicit "unlimited" or "infinite" string, any reason to not have that be an explicit property? As noted, tech advances perhaps could seek to adjust the support limit and plausibly could introduce a use-case for a '0'.  More fundamentally, I would like to avoid the magic number altogether.

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
```diff
@@ -1069,27 +1069,33 @@ private static int getSupport(
       final Predicate<UnitSupportAttachment> ruleFilter) {
     int givenSupport = 0;
     for (final List<UnitSupportAttachment> bonusType : supportsAvailable) {
+      final int bonusTypeCount = bonusType.get(0).getBonusType().getCount();
+      int maxPerBonusType = bonusTypeCount >= 1 ? bonusTypeCount : Integer.MAX_VALUE;
```
DanVanAtta: I'm looking for ways this could be simpler, if you spot anything @ron-murhammer , please go for it.

Looking at these two lines, it's a bit odd as we converting a 'zero' value to a max integer. There seems to be quite a lot packed in these two lines.

I wonder if we can help this by encapsulating the 'infinite' knowledge into bonusType.

Perhaps we could have 'getBonusType().getCount()' contain the logic directly and have that method convert the '0' to be a 'max_value'?

In which case we would have:

```
      int maxPerBonusType = bonusType.get(0).getBonusType().getCount();;
```

That also dovetails nicely with having an explicit flag for 'infinite', the bonus type can take care of that detail without 'DiceRoll' knowing about the details of unit bonus allotment (arguably inappropriate coupling).

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
```diff
@@ -1069,27 +1069,33 @@ private static int getSupport(
...
-            final Set<Unit> supporters = supportersLeft.keySet();
-            if (!supporters.isEmpty()) {
-              final Unit u = supporters.iterator().next();
-              supportUnitsLeft.get(rule).add(u, -1);
-              if (supportUnitsLeft.get(rule).getInt(u) <= 0) {
-                supportUnitsLeft.get(rule).removeKey(u);
-              }
-              unitSupportMap.computeIfAbsent(u, i -> new IntegerMap<>()).add(unit, rule.getBonus());
+          final int numSupportToApply =
+              Math.min(
```
DanVanAtta: Building on previous comment, perhaps we can make our value object 'bonusType' more useful and encapsulate more of this logic. This is logic that 'DiceRoll' really should not know about, dice rolls should not know about rules and support bonuses.

Cernelius:
>You seem opposed to an explicit "unlimited" or "infinite" string, any reason to not have that be an explicit property? As noted, tech advances perhaps could seek to adjust the support limit and plausibly could introduce a use-case for a '0'. More fundamentally, I would like to avoid the magic number altogether.

I really despise this "-1 is infinite duh" convention, and opened a feature request about it here:
https://forums.triplea-game.org/topic/912/a-general-way-to-define-infinite

Tho, in all fairness, either you rework all or nothing. So I would go with the -1 thing here, for now, rather than being inconsistent with anything else.

I would personally prefer that a "+" means positive infinite and a "-" means negative infinite.

In the case of this property, this would be my suggestion of infinite:
`<option name="bonusType" value="ArtilleryBonus" count="+"/>`

If not supporting two infinite, then I would have infinite being an empty count, that would be:
`<option name="bonusType" value="ArtilleryBonus" count=""/>`

I don't like having any letters in numeral fields. For the usualy infinity symbol, I don't like using symbols that are not part of common keyboards (meaning I would only use symbols available for both the QWERTY and DVORAK keyboards).

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
ron-murhammer: Agree. Updated.

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
ron-murhammer: I think this is actually better here than in bonusType. Neither is ideal and DiceRoll itself should probably be broken up into a few classes.

ron-murhammer:
@DanVanAtta Refactored a bit more. Without a major refactoring of DiceRoll, I don't think there is much more than can be done here.

DanVanAtta:
Thanks for reviewing that @ron-murhammer ; I recognize the value of adding features, and it's unfair that we have to pay the technical debt now and before. I think the bar is that we just do not increase complexity, keep same or decrease and we'll be in better shape.

I appreciate the update :+1: 
Agree a larger overhaul is needed, but it feels like the complexity added is no longer *that* much more.

DanVanAtta:
@Cernelius agree, good point on how the other attributes are working

PR #5313, opened 2019-10-03
Chat Refactors: IChatListener breakup, chat history fix, constructor injection

DanVanAtta:
(1) Break up IChatListener interface into two, interface segragation principle. Part of
interface was used for message box, the other part was used for the player listing.

(2) Fix bug where 'slap messages' are recorded into chat history 'n' times instead
of just once.

(3) Update HeadlessChat constructor, dependency injection principle. Instead of the
constructor new'ing up a 'Chat' object, we inject it.

(4) Change sound profile of HeadlessChat to 'no sound'. The sounds played on a headless
server are no-ops, we may as well use the 'no sound' profile.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[] Manually tested

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/chat/Chat.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/Chat.java
@@ -29,7 +29,8 @@ public class Chat implements ChatClient {
 
   private final ChatTransmitter chatTransmitter;
 
-  private final List<IChatListener> listeners = new CopyOnWriteArrayList<>();
+  private final List<ChatMessageListener> chatMessageListeners = new CopyOnWriteArrayList<>();
+  private final List<ChatPlayerListener> chatPlayerListeners = new CopyOnWriteArrayList<>();
 
   @Getter(AccessLevel.PACKAGE)
   private final SentMessagesHistory sentMessagesHistory;
@@ -71,9 +72,7 @@ public class Chat implements ChatClient {
             .sorted(Comparator.comparing(c -> c.getPlayerName().getValue()))
             .collect(Collectors.toList());
 
-    for (final IChatListener listener : listeners) {
-      listener.updatePlayerList(playerNames);
-    }
+    chatPlayerListeners.forEach(listener -> listener.updatePlayerList(playerNames));
   }
 
   @Override
@@ -83,9 +82,7 @@ public class Chat implements ChatClient {
       return;
     }
     chatHistory.add(new ChatMessage(message, from));
-    for (final IChatListener listener : listeners) {
-      listener.addMessage(message, from);
-    }
+    chatMessageListeners.forEach(listener -> listener.addMessage(message, from));
   }
 
   @Override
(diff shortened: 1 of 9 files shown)


game-core/src/main/java/games/strategy/engine/chat/Chat.java
```diff
@@ -112,26 +110,22 @@ public void participantRemoved(final PlayerName playerName) {
...
   }
 
   @Override
   public void slappedBy(final PlayerName from) {
     final String message = "You were slapped by " + from;
-    for (final IChatListener listener : listeners) {
-      chatHistory.add(new ChatMessage(message, from));
-      listener.addMessageWithSound(message, from, SoundPath.CLIP_CHAT_SLAP);
-    }
+    chatHistory.add(new ChatMessage(message, from));
```
DanVanAtta: Bugfix is here, notice that 'chatHistory.add(...)' is in the for loop, yet does not use any of the loop variables. Debugging, on a single slap this is called 'n' times. This line was moved outside of the for-loop so that it is called just once.

Perhaps also worth noting that the chat history line is outside of listener for-loops in other locations as well: https://github.com/triplea-game/triplea/pull/5313/files#diff-8ffdcb54556ca7b66cd59050e599e646R84

game-core/src/main/java/games/strategy/engine/chat/ChatMessageListener.java
```diff
@@ -1,12 +1,9 @@
...
 
 import games.strategy.engine.lobby.PlayerName;
-import java.util.Collection;
-
-/** An interface to allow for testing. */
-public interface IChatListener {
-  void updatePlayerList(Collection<ChatParticipant> players);
 
+/** Callback interface for a component that is interested in chat messages. */
+public interface ChatMessageListener {
```
ron-murhammer: Don't most of our interfaces today start with "I"? Are we moving away from that standard? Just concerned about inconsistency.

game-core/src/main/java/games/strategy/engine/chat/Chat.java
```diff
@@ -95,12 +92,13 @@ public void participantAdded(final ChatParticipant chatParticipant) {
...
-      if (chatSoundProfile == ChatSoundProfile.GAME_CHATROOM) {
-        ClipPlayer.play(SoundPath.CLIP_CHAT_JOIN_GAME);
-      }
-    }
+    chatMessageListeners.forEach(
+        listener -> {
+          listener.addStatusMessage(chatParticipant.getPlayerName() + " has joined");
+          if (chatSoundProfile == ChatSoundProfile.GAME_CHATROOM) {
+            ClipPlayer.play(SoundPath.CLIP_CHAT_JOIN_GAME);
+          }
```
RoiEXLab: I had a look at the code where chatSoundProfile is being set, and it looks like it depends on the usage type this chat instance is supposed to be used for.
Any ideas how this information might relate to the kind of listeners registered?
If it turns out that specific listeners are only registered when the sound profile is for the chatroom, then this ClipPlayer invokation could easily be moved into the appropriate listeners.
Otherwise we might want to consider passing the chatSoundProfile so the listener can decide what to do.

game-core/src/main/java/games/strategy/engine/chat/ChatMessageListener.java
RoiEXLab: AFAIK we don't really use the I prefix for newer interfaces, but existing interfaces stay that way

game-core/src/main/java/games/strategy/engine/chat/Chat.java
RoiEXLab: There even is an `addMessageWithSound` method, this is so weird.

RoiEXLab: IMO it should really be the listeners decision to play sound or not, obviously the class should pass the "kind" of message to the listeners so they can decide based on that.

DanVanAtta: Sounds about right @RoiEXLab . If we push the play sound functionality to listeners and in more appropriate places, I would guess eventually the `ChatSoundProfile` may no longer even be needed .

PR #5315, opened 2019-10-05
Improve politics UI

asvitkine:
This change makes a number of improvements to the politics UI:
   - Political action buttons are now placed in the table, making it clearer what the current stance is towards the target country and other countries without having to correlate between two different panels. The bottom actions panel is removed.
   - The layout of the political grid is improved to make each column be of equal width to make for a more pleasing visual layout.
   - The "No Actions" button is renamed to "No Further Actions" as that is more accurate.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[X] Manually tested

Played Napoleonic Wars FFA and Feudal Japan and viewed the Politics panel, both during the Politics turn, and as a standalone status via the menu.

### Before
before

### After
after


DanVanAtta:
@Heppisorus curious if you've any thoughts on this update, either for now or things that we can follow up with.

game-core/src/main/java/games/strategy/triplea/ui/PoliticalStateOverview.java
```diff
@@ -43,36 +46,45 @@ public PoliticalStateOverview(
     drawPoliticsUi();
   }
 
+  private void addCell(final Component cell, final Insets insets, final int x, final int y) {
```
DanVanAtta: Code nitpick: Please check the method ordering convention:
https://github.com/triplea-game/triplea/wiki/Code-Conventions#depth-first-method-ordering

We want to avoid new methods being added *above* where they are used. Specifically here "addCell" should be below the method that first uses it.

game-core/src/main/java/games/strategy/triplea/ui/PoliticalStateOverview.java
```diff
@@ -43,36 +46,45 @@ public PoliticalStateOverview(
...
+            GridBagConstraints.CENTER,
+            GridBagConstraints.BOTH,
+            insets,
+            0,
+            0));
+  }
+
   /** does the actual adding of elements to this panel. */
   private void drawPoliticsUi() {
     this.setLayout(new GridBagLayout());
```
DanVanAtta: If we have fixed width columns, would a standard 'GridLayout' do the job?

game-core/src/main/java/games/strategy/triplea/ui/PoliticalStateOverview.java
```diff
@@ -97,50 +109,38 @@ private void drawPoliticsUi() {
...
+    if (background != null) {
+      p.setBackground(background);
+    }
+    return p;
   }
 
   /** Gets a label showing the colored relationshipName between these two players. */
   private JPanel getRelationshipLabel(final PlayerId player1, final PlayerId player2) {
+    if (player1.equals(player2)) {
+      return wrapInJPanel(new JLabel(PoliticalStateOverview.LABEL_SELF), null);
```
DanVanAtta: code nit: try to avoid passing null values. https://github.com/triplea-game/triplea/wiki/Code-Conventions#null-handling

I think the helper method is perhaps not as needed as it could be.

We've a 'builder' helper library to fix some of the swing API problems, using that this is relatively straight forward:
```
new JPanelBuilder().add(new JLabel(PoliticalStateOverview.LABEL_SELF)).build();
```

We do not have a background property on `JPanelBuilder` it would be good to add, without doing that work, the other variant of the method call is pretty straight forward as well (line 168):

```
JPanel panel = new JPanelBuilder().addnew JLabel(relType.getName()).build();
panel.setBackgroundColor(relColor);
return panel;
```

Part of the motivation is to avoid different variable states, where 'null' background color means that parameter is unused, raises questions like "when is this null?" when is it not? What happens if it is. We also try to annotate nullable variables/values with "@Nullable" so that IDE can help detect NPE's for us, but first it's better to avoid null.

FrostionAAA:
I think it should be considered to have columns be the width of the widest player name instead of a fixed width, if this is the current suggestion. In maps with many players it would be nice to have it as narrow as possible, to avoid too much scrolling.

The “after” picture seems to have wider than necessary columns.

game-core/src/main/java/games/strategy/triplea/ui/PoliticalStateOverview.java
```diff
@@ -157,17 +157,16 @@ private JPanel getRelationshipLabel(final PlayerId player1, final PlayerId playe
...
+    return getRelationshipComponent(player1, player2, relType, getRelationshipTypeColor(relType));
   }
 
-  private JComponent getRelationshipComponent(
-      final PlayerId player1, final PlayerId player2, final RelationshipType relType) {
+  protected JPanel getRelationshipComponent(
+      final PlayerId player1,
+      final PlayerId player2,
+      final RelationshipType relType,
+      final Color relColor) {
```
DanVanAtta: Is this parameter nulllable? The color is computed from the relationship type, so we probably can avoid the null parameter by moving 'getRelationshipTypeColor' to be within this method. Considering as well the parameter is only used when 'editable' is false, that might be a good move. (Going by principle that ideally all method parameters are always used; which seems to be a problem in this method already, but perhaps better to not exacerbate it)

Going that route, one could have:
```
JPanel panel = new JPanelBuilder().add(new JLabel(relType.getName()).build();
Optional.ofNullable(getRelationshipTypeColor(relType)).ifPressent(panel::setBackgroundColor);
return panel;
```

game-core/src/main/java/games/strategy/triplea/ui/PoliticalStateOverview.java
```diff
@@ -157,17 +157,16 @@ private JPanel getRelationshipLabel(final PlayerId player1, final PlayerId playe
         data.releaseReadLock();
       }
     }
-    final JComponent relationshipLabel = getRelationshipComponent(player1, player2, relType);
-    final JPanel relationshipLabelPanel = new JPanel();
-    relationshipLabelPanel.add(relationshipLabel);
-    relationshipLabelPanel.setBackground(getRelationshipTypeColor(relType));
-    return relationshipLabelPanel;
+    return getRelationshipComponent(player1, player2, relType, getRelationshipTypeColor(relType));
```
DanVanAtta: Side-comment, this helper method is a bit ugly, in existing code. I'm seeing a "mixing of abstractions" code smell. Notably, above, we have very low level code, where we're getting locks on game data, computing a relationship type at very low level, then on this last line the abstraction goes to very high level "getRelationshipComponent".

To fix, we'd probably need to re-split the helper methods, so we coudl have stuff like:
```
RelationshipType relType = computeRelationship();
return createRelationshipComponent(label);
```
Sigh, no action needed on this comment, just noting that the existing code structure could use help.

ron-murhammer:
It would be good to see what Global 40 2nd looks like before and after as that is probably the most popular map with politics.

game-core/src/main/java/games/strategy/triplea/ui/PoliticsPanel.java
```diff
@@ -57,61 +55,81 @@
...
+                          1.0,
+                          GridBagConstraints.WEST,
+                          GridBagConstraints.HORIZONTAL,
+                          insets,
+                          0,
+                          0);
+                    }
+
+                    @Override
+                    protected JPanel getRelationshipComponent(
```
DanVanAtta: Overriding concrete methods, general usage of inheritance can create a number of problems. For example, in the `PoliticalStateOverview` class, it's not going to be clear we have dependencies on any method that is called by this override method. It's a 'spooky action at distance', where changes in `PoliciticalStateOverview` impact this class.

A good rule of thumb is to simply avoid overriding concrete types. I think this could potentially be fixed by extracting the common components into an abstract class. `getRelationshipComponent` would then be an abstract method, and there would be two child classes, `PoliticalStateOverview` and `PoliticsPanel`. This way the common code is in an abstract class and because it is abstract, it's more clear that is shared code. Hopefully that seems like a good design idea, creates a more tree like structure where children add to functionality of the parent rather than changing the behavior of the parent. 

TL:DR, I'd like to see us avoid an override of a concrete method. It can create a lot of complexity/confusion and breaks modularity. The common code could also be extracted to another class and we could use composition, so that the common code is then invoked instead of being inherited. eg:
```
doRendering() {
  RelationshipType relType = relationshipCalculator.calculateRelationship();
  return getRelationshipComponent(relType);
}
```
Given the existing structure, the abstract base class might be the easier route.  @asvitkine any thoughts on this one?

game-core/src/main/java/games/strategy/triplea/ui/PoliticsPanel.java
```diff
@@ -57,61 +55,81 @@
...
+                      panel.add(status, makeGridBagConstraints(insets, row++));
+                      if (!player1.equals(getCurrentPlayer())) {
+                        return panel;
+                      }
 
-              final JSplitPane splitPane =
-                  new JSplitPane(JSplitPane.VERTICAL_SPLIT, overviewScroll, choiceScroll);
-              splitPane.setOneTouchExpandable(true);
-              splitPane.setDividerSize(8);
+                      for (final PoliticalActionAttachment paa : validPoliticalActions) {
```
DanVanAtta: nit: abbreviations should probably be avoided: `paa` and favor spelling it out instead. I know it's existing, but given the relatively heavy modifications in this section, I think it probably makes sense to go ahead and fix it.

The cost to write this variable out fully is one-time, meanwhile it helps anyone reading the code (it's one less thing for them to keep track of in their mind when trying to scan/or study this section of code).

game-core/src/main/java/games/strategy/triplea/ui/PoliticsPanel.java
```diff
@@ -57,61 +55,81 @@
...
-                                  ? 20
-                                  : 0))),
-                      (overviewScroll.getPreferredSize().height > availHeightOverview
-                          ? availHeightOverview
-                          : (overviewScroll.getPreferredSize().height
-                              + (validPoliticalActions.isEmpty() ? 26 : 0)
-                              + (overviewScroll.getPreferredSize().width > availWidth ? 20 : 0)))));
+                  new PoliticalStateOverview(getData(), getMap().getUiContext(), false) {
+
+                    private GridBagConstraints makeGridBagConstraints(
```
DanVanAtta: For consideration, there is a 'GridBagConstraintBuilder' that might be of help to you here.

Cernelius:
I'm dubious on the first item. I might have a slight preference for how it works now, actually.

By the way, can you add an option for hiding players from all "Politics Panel" tabs (at the discretion of the mapmaker, similarly to the "isHidden" option)? For example, for cases like the GameSetup and MinorClans of Feudal Japan or Dutch, Mongolians, Neutral_Axis and Neutral_Allies of WWII Global. I've also a map that could really use this, though not released yet.

Cernelius:
The "No Further Actions" would be better as just "Close", as that is just closing the windows, maybe to look at the map, not really deciding anything the like (that is done with the "Done" button in action).

Cernelius:
>I think it should be considered to have columns be the width of the widest player name instead of a fixed width

I much agree and immediately thought the same.

DanVanAtta:
"Done" is very consistent with other such screens, like purchase.
Going forward in this conversation, let's please keep discussion to the existing changes before we discuss too much the forward looking updates, like re-ordering of columns.

> I think it should be considered to have columns be the width of the widest player name instead of a fixed width

I think that is the case here, the widest column is calculated and then applied to all of them.

asvitkine:
Changing the button to "Done" sounds good.

For the width of the columns, the code is already calculating the width dynamically. I believe it's the buttons in the example screenshot that are making things the columns wider than the name of the countries. I can see if we could have less padding around them, for example. Or make the text smaller. However, this would go against the other suggestion of making the button text larger.

For the code changes, I can look at making them once we have agreement on the UI.

DanVanAtta:
@asvitkine be aware you'll need to drive a good chunk of that agreement conversation. Getting consensus on UI changes I've found to be one of the more difficult things about this project as there are so many opinions and so many "by-the-way" type of suggestions. I suspect there will be majority support for the underlying support, you'll need to create that forum post to get that kind of feedback.

asvitkine:
Thanks for asking how it would look like in Global 40. The change as-is does not work well for that map, since there's multiple declare war actions, each targeting a separate alliance of players.

Here's how it looks with my change (poorly due to the buttons not conveying sufficient information):
global_40_2nd_after

For reference, here's how it looks now:
global_40_2nd_before

A question was asked earlier about what happens if the window is too wide? In that case, scroll bars are shown, here's an example (here I just resized the window, but I confirmed that the same thing happens if the window would otherwise be too big):
scrollbars

Anyway, it's clear this doesn't work for the map. I could see a few options:
 1. We have a version of the existing UI and either use it for maps where there is ambiguity, like this, or provide a way to toggle between the two UIs. In terms of using for specific maps, we could either have the engine detect this (e.g. if it sees multiple actions with same name and different targets, it switches to old UI), or have the mapmaker specify. If the latter, then I worry there could be debate about which maps should use this UI vs. the old one.
  2. We could try to provide more info in the buttons. For example, listing other country icons.
  3. We could list all actions that affect multiple countries in the old UI as before.

I like option 3 the most. This way, any simple actions could be inline, but any actions affecting multiple countries would be down below as before.

Another variant on 3 is to always build the bottom panel, but hide (i.e. start with it closed), if those actions are all simple actions that are shown inline (which will look most similar to the new proposed UI, but lets people expand the list). Show it only when there are more complex actions (that declare war on multiple countries).

DanVanAtta:
> (2) We could try to provide more info in the buttons. For example, listing other country icons.

I like this as an option, keeps the UI more cohesive.

> (3) We could list all actions that affect multiple countries in the old UI as before.

If we go this route, the UI becomes a bit inconsistent. Given the behavior is different across maps, maybe it won't be very noticeable, but offhand seems like a negative.

> (1) provide a way to toggle between the two UIs

Seems inconsistent again, not sure if we would want to support the code with this level of complexity. IMO either the new UI is preferable or it's not.

ron-murhammer:
@asvitkine @DanVanAtta Yeah, I had a feeling looking at the initial design that it wouldn't work quite so well for all maps. I think the best thing to do would be create a thread in the forum with some examples and options to get feedback from map makers and others in the community (maybe they'll have an even better idea that we didn't consider). I also think we probably need to get 3-5 maps that have combined hopefully a good representation of politics would be ideal for examples and testing (global 40 and Nap are 2 good ones to start).

In terms of the 3 options you listed, here are my initial thoughts on them:
1. The flexibility of being able to either have map makers choose or have the engine select a default by some specified rules would be nice and also keeps the risk of this change low. The drawbacks would be that it might be difficult to come to rules to default to one or the other, if a map setting then my gut is most maps won't update to it so low adoption/usage, and its kind of creating 2 different politics interfaces which can be more difficult for users to adapt to from map to map (ex. I play global 40 which say uses the old then I try Nap FFA which uses the new and get confused where the buttons are).
2. I don't think that will work well given the number of nations in some cases even just in the Global 40 example though I'd be open to seeing how that looks. Also I'm not sure if you are already but should probably make the button tooltip have the text that displays to the right of the buttons in the current one.
3. This is probably my least favorite option as now you have "user actions" in potentially multiple places (in the table and below the table). That generally makes the user experience confusing and not cohesive.

Cernelius:
>"Done" is very consistent with other such screens, like purchase.

Pointing out that I'm not a fan of having "Done" on the window then "Done" again in the action tab (but this is off topic) in phases like purchase, there is a difference here. When you are purchasing, you define your purchase by clicking on "Done". With user/political actions, you not only define but even irreversibly confirm your action upon clicking on the button; those are the "Done and Confirmed" part of it. So what you do clicking that thing in the middle is absolutely nothing more than to "Close" the window. "Done" would be consistent if the politics etc. would work in that you select all your actions, then look at them, then click on "Done", and send them all at the same time, but this is not how it works.

Cernelius:
At the moment, I've nothing else to say than what I already did. I'm personally fairly good with how all this stuff work at the moment; tested it a lot, and never found any real flaw. The only thing I can really think of is being able to hide all those pointless players (Dutch, etc.), so to keep the information more concise.

Only some warnings:

Don't forget that with political actions you can change relationships in general, not necessarily only between the turn player and another single player (though this is how it usually works). And, in general, don't forget that I don't believe you can find any maps that use everything that you can do in politics. To be sure, you should rather create a map yourself, using each and every possibility offered by politics (paying for making actions, etc.).

What you can do in politics may be expanded anytime in the future, to be more like user actions. I actually believe this would be good; so a more unbounded format may be advisable, even if the current one would be fully tailored on the current possibilities (only changing relationships, possibly at a resources cost).

I suggest keeping the political and user actions format fairly similar, as politics is really just a reduced version of user actions, with the politics screen on top.

asvitkine:
Since we don't yet have an agreement of the direction of the UI, what I'm going to do is to split off the hopefully non-controversial parts of this change (renaming the button and maxing columns same width) into a separate PR that can land first. Once that's in, we can continue to discuss the remaining changes and find the next steps.

Thanks for all the feedback.

asvitkine:
I've sent out https://github.com/triplea-game/triplea/pull/5320 for the changes to the column widths and button name.

Cernelius:
To more or less achieve your aim, I'm thinking you can go the way of intermixing the politic panel with the actions, by putting all actions (kept in the current format) immediately below the row of a player they are influencing, excluding the maker. This is sort of what you were doing, except that it would keep the current format and it would be under the whole row for the player, instead of just under the entry for the player in the row of the turn player.

DanVanAtta:
@asvitkine given the better widths have been merged, this can be closed? You're following up on forums for the button changes?

asvitkine:
Let's close this PR for now. Since the other part of this change landed already, this no longer cleanly applies. I think if I want to pursue better UI here, I'll end up recoding it, maybe trying out Cernelius' suggestion and discussing on the forums.

PR #5318, opened 2019-10-06
Move sound invocation to ChatMessageListener implementation

RoiEXLab:
After seeing this multiple times now, I decided to fix this, turned out easier than expected.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[x] No manual testing done
[] Manually tested

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/chat/Chat.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/Chat.java
@@ -4,8 +4,6 @@ import com.google.common.collect.EvictingQueue;
 import games.strategy.engine.lobby.PlayerName;
 import games.strategy.engine.message.MessageContext;
 import games.strategy.net.Messengers;
-import games.strategy.sound.ClipPlayer;
-import games.strategy.sound.SoundPath;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -42,23 +40,13 @@ public class Chat implements ChatClient {
       Collections.synchronizedCollection(EvictingQueue.create(1000));
 
   private final ChatIgnoreList ignoreList = new ChatIgnoreList();
-  private final ChatSoundProfile chatSoundProfile;
   @Getter private final PlayerName localPlayerName;
   private final Collection<BiConsumer<PlayerName, String>> statusUpdateListeners =
       new ArrayList<>();
 
-  /** A profile defines the sounds to use for various chat events. */
-  public enum ChatSoundProfile {
-    LOBBY_CHATROOM,
-    GAME_CHATROOM,
-    NO_SOUND
-  }
-
-  public Chat(
-      final Messengers messengers, final String chatName, final ChatSoundProfile chatSoundProfile) {
+  public Chat(final Messengers messengers, final String chatName) {
     this.localPlayerName = messengers.getLocalNode().getPlayerName();
     chatTransmitter = new JavaSocketChatTransmitter(this, chatName, messengers);
-    this.chatSoundProfile = chatSoundProfile;
     sentMessagesHistory = new SentMessagesHistory();
     chatters = Optional.ofNullable(chatTransmitter.connect()).orElseGet(HashMap::new);
     updateConnections();
@@ -91,12 +79,7 @@ public class Chat implements ChatClient {
     chatters.put(chatParticipant, "");
     updateConnections();
     chatMessageListeners.forEach(
-        listener -> {
-          listener.addStatusMessage(chatParticipant.getPlayerName() + " has joined");
-          if (chatSoundProfile == ChatSoundProfile.GAME_CHATROOM) {
-            ClipPlayer.play(SoundPath.CLIP_CHAT_JOIN_GAME);
-          }
-        });
+        listener -> listener.addStatusMessage(chatParticipant.getPlayerName() + " has joined"));
   }
 
   @Override
--- a/game-core/src/main/java/games/strategy/engine/chat/Chat.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/Chat.java
@@ -117,8 +117,7 @@ public class Chat implements ChatClient {
   public void slappedBy(final PlayerName from) {
     final String message = "You were slapped by " + from;
     chatHistory.add(new ChatMessage(message, from));
-    chatMessageListeners.forEach(
-        listener -> listener.addMessageWithSound(message, from, SoundPath.CLIP_CHAT_SLAP));
+    chatMessageListeners.forEach(listener -> listener.addSlapMessage(message, from));
   }
 
   @Override
--- a/game-core/src/main/java/games/strategy/engine/chat/ChatMessageListener.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/ChatMessageListener.java
@@ -6,7 +6,7 @@ import games.strategy.engine.lobby.PlayerName;
 public interface ChatMessageListener {
   void addMessage(String message, PlayerName from);
 
-  void addMessageWithSound(String message, PlayerName from, String sound);
+  void addSlapMessage(String message, PlayerName from);
 
   void addStatusMessage(String message);
 }
--- a/game-core/src/main/java/games/strategy/engine/chat/ChatMessagePanel.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/ChatMessagePanel.java
@@ -222,9 +222,13 @@ public class ChatMessagePanel extends JPanel implements ChatMessageListener {
     addMessageWithSound(message, from, SoundPath.CLIP_CHAT_MESSAGE);
   }
 
-  /** thread safe. */
   @Override
-  public void addMessageWithSound(final String message, final PlayerName from, final String sound) {
+  public void addSlapMessage(final String message, final PlayerName from) {
+    addMessageWithSound(message, from, SoundPath.CLIP_CHAT_SLAP);
+  }
+
+  private void addMessageWithSound(
+      final String message, final PlayerName from, final String sound) {
     SwingAction.invokeNowOrLater(
         () -> {
           if (from == null || chat == null) {
--- a/game-core/src/main/java/games/strategy/engine/chat/ChatMessagePanel.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/ChatMessagePanel.java
@@ -68,8 +68,16 @@ public class ChatMessagePanel extends JPanel implements ChatMessageListener {
               chat.updateStatus(status.trim());
             }
           });
+  private final ChatSoundProfile chatSoundProfile;
 
-  public ChatMessagePanel(final Chat chat) {
+  /** A profile defines the sounds to use for various chat events. */
+  public enum ChatSoundProfile {
+    LOBBY,
+    GAME
+  }
+
+  public ChatMessagePanel(final Chat chat, final ChatSoundProfile chatSoundProfile) {
+    this.chatSoundProfile = chatSoundProfile;
     init();
     setChat(chat);
   }
@@ -82,7 +90,7 @@ public class ChatMessagePanel extends JPanel implements ChatMessageListener {
     setSize(300, 200);
   }
 
-  public String getAllText() {
+  String getAllText() {
     return text.getText();
   }
 
@@ -305,6 +313,9 @@ public class ChatMessagePanel extends JPanel implements ChatMessageListener {
                 e);
           }
         });
+    if (chatSoundProfile == ChatSoundProfile.GAME) {
+      ClipPlayer.play(SoundPath.CLIP_CHAT_JOIN_GAME);
+    }
   }
 
   /** Show only the first n lines. */
--- a/game-core/src/main/java/games/strategy/engine/chat/ChatPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/ChatPanel.java
@@ -1,6 +1,6 @@
 package games.strategy.engine.chat;
 
-import games.strategy.engine.chat.Chat.ChatSoundProfile;
+import games.strategy.engine.chat.ChatMessagePanel.ChatSoundProfile;
 import games.strategy.net.Messengers;
 import java.awt.BorderLayout;
 import java.awt.Dimension;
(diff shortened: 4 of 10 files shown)


DanVanAtta:
@RoiEXLab would you mind doing some manual testing that sounds still play? Given TripleA, you never know... I've recently had some big surprises when firing up the game for code that I thought safe to change.

game-core/src/main/java/games/strategy/engine/chat/ChatMessageListener.java
```diff
@@ -6,7 +6,7 @@
 public interface ChatMessageListener {
   void addMessage(String message, PlayerName from);
 
-  void addMessageWithSound(String message, PlayerName from, String sound);
+  void addSlapMessage(String message, PlayerName from);
```
DanVanAtta: Any thought to rename this to `addSlapEvent`? 
There is also a false consistency of the "add" prefix, `addStatusMessage` is badly named, since it's setting a status message. But for slap event, 'add' is not the worst verb, but I'm struck as it's not really a message but more a chat event.

game-core/src/main/java/games/strategy/engine/chat/ChatMessageListener.java
```diff
@@ -6,7 +6,7 @@
...
 
   void addStatusMessage(String message);
```
RoiEXLab: This is a thing I'm not quite sure about:
All those kind of messages are basically the same, so I was thinking maybe it makes sense to merge all those methods and add a `MessageType` enum.
This would make sense because 2 out of 3 implementations have the same implementation for `addMessage` and `addSlapMessage`, so they call the other method anyways.
However if we did this for `addSlapMessage` and `addMessage`, we'd have to do this for `addStatusMessage` as well if we want to stay consistent. But then we'd have to pass null for `PlayerName from` because this information is not present in this case.
That's why I decided to do it this way for now. Let me know if you think the other approach is better in your opinion

game-core/src/main/java/games/strategy/engine/chat/ChatMessagePanel.java
```diff
@@ -68,8 +68,16 @@
...
+  private final ChatSoundProfile chatSoundProfile;
 
-  public ChatMessagePanel(final Chat chat) {
+  /** A profile defines the sounds to use for various chat events. */
+  public enum ChatSoundProfile {
+    LOBBY,
+    GAME
+  }
+
+  public ChatMessagePanel(final Chat chat, final ChatSoundProfile chatSoundProfile) {
```
DanVanAtta: Any thoughts/consideration to using a factory pattern to hide the `ChatSoundProfile` enum?
eg:
```
public static ChatMessagePanel newLobbyChatPanel() { .. }
public static ChatMessagePanel newGameChatPanel() { .. }
```

RoiEXLab:
@DanVanAtta Yeah sure, I just wanted to get the PR out quickly.
I'll report back once I tested it.

game-core/src/main/java/games/strategy/engine/chat/ChatMessagePanel.java
```diff
@@ -222,9 +230,13 @@ public void addMessage(final String message, final PlayerName from) {
...
   }
 
-  /** thread safe. */
   @Override
-  public void addMessageWithSound(final String message, final PlayerName from, final String sound) {
+  public void addSlapMessage(final String message, final PlayerName from) {
+    addMessageWithSound(message, from, SoundPath.CLIP_CHAT_SLAP);
+  }
+
+  private void addMessageWithSound(
```
DanVanAtta: I think we're still left with a SRP violation.

What do you think of making the private method to be focused on playing a sound?

eg:

```
  public void addSlapMessage(final String message, final PlayerName from) {
    addChatMessage(message, from);
    playSound(from, SoundPath.CLIP_CHAT_SLAP);
  }
```

game-core/src/main/java/games/strategy/engine/chat/ChatMessageListener.java
DanVanAtta: I think this is okay for now. In part the question is do we have the control flow be at the caller or callee, ie: this is a chat message, this is a player joined message. I think keeping strong types and enumerating the different event types on this interface is good for now. This way the listener can be a true listener and it's caller will identify and then route messages to it.

RoiEXLab:
Ok, looks like Sound was actually played more often than it was supposed to.
I fixed this by adding a boolean parameter, not that elegant, but I guess good enough for now. Let me know if you have any suggestions how this can be improved.
Haven't tried lobby chat so far (I'm still trying to setup docker on windows, but I almost got it), but since this one isn't supposed to play sound anyways it's probably not that important.

game-core/src/main/java/games/strategy/engine/chat/ChatMessagePanel.java
RoiEXLab: Good point, I'm not sure if this will have unintended consequences though:
If this listener is called on the EDT, this operation is blocking, otherwise it's executed asynchronously. So there are a handful of options.
I'd assume that the nio listeners are not running on the EDT for simplicity, otherwise this problem would be trivial to solve. Let me know if you have any insights about which Thread invokes this listener.

The options:
1. Run `SwingAction.invokeNowOrLater` directly in `addMessage` and `addSlapMessage`, and turn `addMessageWithSound` into 2 blocking methods.
2. Turn `addMessageWithSound` into 2 methods that each invoke `SwingAction.invokeNowOrLater` separately and execute them one by one. This way the sound might not be playing synchronized with the actual chat event.
3. Like 2., but it's guaranteed to be asynchronous even when executed on the EDT.

**However there's a problem with all of those approaches:**
The flood control check is done asynchronously as well, and obviously the sound should only be played if floodControl allows it. I believe the advertised "thread-safety" is mainly coming from the fact that everything is executed on the EDT, so I wouldn't feel confident moving those checks to the current thread :(
Thoughts?

game-core/src/main/java/games/strategy/engine/chat/ChatMessagePanel.java
RoiEXLab: This wouldn't be easy to achieve: The `ChatSoundProfile` is passed a couple of layers down:
https://github.com/triplea-game/triplea/pull/5318/files#diff-9aeb0682c4d70742aca95a41ebca9d8bL27-R30

So a factory pattern would actually make things more complicated.

game-core/src/main/java/games/strategy/engine/chat/ChatMessageListener.java
RoiEXLab: I wanted to stay consistent with `addMessage` and `addStatusMessage`, because in the end all of those "events" somehow add a message on a display or in a log or something.
I'd appreciate if you would open a PR with name proposals for all 3 methods that suit better in your opinion, but I think this discussion is out-of-scope for this PR.

DanVanAtta: Sure, it's a false consistency though, it's being consistent with something that is incorrect. It would lead to something like  "addPlayerJoinedMessage", which is not a message. It's also a question what exactly are we adding to? It's an implementation detail that there is a chat history, and that could potentially be removed.

So, `addSlapEvent`, are you opposed to that? IMO it's probably better like "playerWasSlapped", or "slapEventOccurred", or "handleSlap", but fundamentally I'd prefer a distinction between messages and chat events (messages could be considered a subset, but not everything is a message, like a player joining, which should arguably be part of this interface)

DanVanAtta: I got a bit carried away, "addStatusMessage" is incorrect, it's not adding, it's setting, and a status is not at all a message.

game-core/src/main/java/games/strategy/engine/chat/ChatMessagePanel.java
DanVanAtta: Okay. The 'passed' a couple layers down is a problem in of itself, too many layers sit on top and certainly was just 'wired' through. Perhaps this is a place where DI would help.

#Nothing further needed.

game-core/src/main/java/games/strategy/engine/chat/ChatMessagePanel.java
DanVanAtta: It might not be thread safe, if flood control is checked on 'n' parallel threads, are all the needed values marked as volatile to guarantee each thread will see the updated values from other threads?

I think the split could be done by moving the SwingAction async logic to the top level, so we then just invoke two methods from within that thread rather than having the lower-level methods do so.

The flood control check IMO would be better also moved to a top layer, fits better with "imperative shell, functional core" design pattern (in other words, control flow logic at a higher level calling functional components). From another perspective, the 'addChatMessage' method is incorrectly named as it does not always do that, it checks flood control and then has conditions based on that.

Another benefit to moving control-flow logic to an imperative shell is it's easier to understand larger sections of the code. Currently if we inlined all the code, it would be all the same for understanding. Today, you have to go through all the lines of every method invoked to know what will happen, the method migth be a no-op, it might not be, to know you have to look at them. One trick to maintainable code is to divide and conquer what someone needs to know about. For example "playSoundMethod" that maybe plays a sound, you have to read it all. But if you see there is a condition on the method call, and you know the condition is false, then that's an entire private method you do not have to read as you know it's not executed (rather than executed but resulting in a no-op).

game-core/src/main/java/games/strategy/engine/chat/Chat.java
```diff
@@ -79,7 +79,8 @@ public void participantAdded(final ChatParticipant chatParticipant) {
     chatters.put(chatParticipant, "");
     updateConnections();
     chatMessageListeners.forEach(
-        listener -> listener.addStatusMessage(chatParticipant.getPlayerName() + " has joined"));
+        listener ->
+            listener.addStatusMessage(chatParticipant.getPlayerName() + " has joined", true));
```
DanVanAtta: Ah yeah, this reminds me, `addStatusMessage`, is NOT EVEN A STATUS!! There is a player 'status' in chat, it's the text that goes in parentheses next to a players name. I forget often that this is actually a `addChatEvent`, and is not at all a player status...

I think we might just need to fix up the API, the boolean parameter is pretty message. EG:
```
listener.joinEvent(PlayerName)
listener.slapEvent(PlayerName)
```

game-core/src/main/java/games/strategy/engine/chat/ChatMessageListener.java
DanVanAtta: Sorry, I was incorrect, 'addStatusMessage' is just confusing, as we have an important concept of a player status, and this 'addStatusMessage' has nothing to do with that. The 'addStatusMessage' is really just a generic chat event. Sigh..

game-core/src/main/java/games/strategy/engine/chat/Chat.java
RoiEXLab: Yeah ikr, I confused it too when I was testing stuff.

RoiEXLab:
@DanVanAtta So I think we can both agree that the interface has quite some problems that make it less than ideal.
However this PR really was supposed to just move the ClipPlayer inside the listener interface which worked without a doubt, so I'd really appreciate if you would create a follow-up PR to address all the suggestions you already mentioned.
If there are any suggestions you're sure they are within the scope of this PR, like a simple rename of a method that is **_not_** completely re-done in a follow-up PR anyways, I'll try to address them of course.
Otherwise I'd appreciate a merge.

DanVanAtta:
The boolean parameter is pretty ugly, it goes to the interface names and indicates we need a new interface method for the player joined. @RoiEXLab considering you've added the boolean parameter to get around how the sound profile is dealt with, it feels like it is in scope. If you want to follow up, you can feel free to merge and do another iteration to fix the boolean parameter.

RoiEXLab:
Fair point @DanVanAtta 
I wasn't able to find a good way to "rework" the whole thing to just use "events" instead of getting a message string, simply because there are 2 implementations of the interface and I don't want to duplicate the code for every kind of event, so I went with a simpler solution to just create more message methods.

Naming is not really good, but that should solve the ugly boolean problem.
Let me know if that's ok for you now.

game-core/src/main/java/games/strategy/engine/chat/ChatClient.java
```diff
@@ -34,7 +34,7 @@
   /**
    * An event message is a message from the server that is a non-chat message, eg: "x slapped y".
```
DanVanAtta: javadoc is out of date

game-core/src/main/java/games/strategy/engine/chat/ChatMessageListener.java
```diff
@@ -8,5 +8,9 @@
 
   void addSlapMessage(String message, PlayerName from);
 
-  void addStatusMessage(String message, boolean joined);
+  void addSlapMessage(String message);
+
+  void addPlayerJoinedMessage(String message);
+
+  void addPlayerLeftMessage(String message);
```
DanVanAtta: side-note, eventually we should follow up to avoid the kingdom of nouns problem, the "add.*Message" is noisy.

eg:
```
messageReceived
playerJoined
playerLeft
slap
```

Conveys same info as the 'add.*Message'. I don't really care if you change it now, but I'm nervous about the 'consistency' argument when we add another method and force it to fall into the kingdom of nouns naming anti-pattern.

game-core/src/main/java/games/strategy/engine/chat/ChatMessageListener.java
```diff
@@ -4,13 +4,13 @@
 
 /** Callback interface for a component that is interested in chat messages. */
 public interface ChatMessageListener {
-  void addMessage(String message, PlayerName from);
+  void messageReceived(String message, PlayerName from);
 
-  void addSlapMessage(String message, PlayerName from);
+  void slap(String message, PlayerName from);
```
DanVanAtta: slap and slap could be a bit confusing. One is sending a slap, the other is receiving a slap. I'd suggest "slap" and "slapped", the latter to indicate a slap has been received.

If you agree, please fix and then merge!  :shipit:

PR #5320, opened 2019-10-09
 Improve Politics UI: Equal-width columns and "Done" button.

asvitkine:
Improve Politics UI: Equal-width columns and "Done" button.

This is split off from https://github.com/triplea-game/triplea/pull/5315 and limits changes to the following:
  - Makes columns on the politics panel equal width for a better visual appearance
  - Renames "No Actions" button to "Done" for consistency with other UIs and to reduce confusion since previously some actions might have been performed. Also maps Shift-D and Escape to it, for consistency with other dialogs.
  - Also applies some feedback from the other PR's review and cleans up the code.

This PR does not include any changes to the political action buttons unlike the previous PR, since there was no consensus on that part of the change. It

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[X] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[X] Manually tested

Manually tested by loading up the politics UI on Napoleonic Wars FFA and Global 40 2nd.

### Before

before

### After

after_simple


game-core/src/main/java/games/strategy/triplea/ui/PoliticalStateOverview.java
```diff
@@ -34,6 +37,7 @@
   private final GameData data;
   private final boolean editable;
   private final Set<Triple<PlayerId, PlayerId, RelationshipType>> editChanges = new HashSet<>();
+  private int maxColumnWidth;
```
DanVanAtta: As mentioned in the other PR, is grid layout not an option to get fixed width columns?

game-core/src/main/java/games/strategy/triplea/ui/PoliticsPanel.java
```diff
@@ -126,8 +126,7 @@
                       0));
               final JButton noActionButton =
                   new JButton(
-                      SwingAction.of(
-                          "No Actions", event -> politicalChoiceDialog.setVisible(false)));
+                      SwingAction.of("Done", event -> politicalChoiceDialog.setVisible(false)));
```
DanVanAtta: Is there a 'close' listener attached anywhere? The other buttons now uniformly have "shift+d" mapped to them, this would be missing that.

game-core/src/main/java/games/strategy/triplea/ui/PoliticalStateOverview.java
asvitkine: No, for two reasons:
 1. There are horizontal separators bars that are added to separate the rows. These are added as a row.
 2. The width of the first column does not need to be the same as the others. It has to be as wide as the widest country name - but that may not be the max width, for example if a political status string is longer.

game-core/src/main/java/games/strategy/triplea/ui/PoliticsPanel.java
asvitkine: Added.

game-core/src/main/java/games/strategy/triplea/ui/PoliticalStateOverview.java
DanVanAtta: (1) Does adding `new JSeparator()` behave differently for grid vs gridbag layout?
(2) Looking at the values in the first column, it's the same label as in the first horizontal row, 'flag+country'. Am I missing something or wouldn't the widest country name of the first column be equal to the widest country name listed in the first row?

game-core/src/main/java/games/strategy/triplea/ui/PoliticsPanel.java
DanVanAtta: Please double check these via manual testing if you have not already. We have an unresovled bug where key bindings are always active, I suspect typing in chat "D" would close the politics window, but that is an existing problem we need to solve.

game-core/src/main/java/games/strategy/triplea/ui/PoliticalStateOverview.java
DanVanAtta: To clarify, for (1), adding multiple separators do not join together? Or does it create visual gaps?

DanVanAtta: Ah, sorry, did not read this well closely enough:

>  but that may not be the max width, for example if a political status string is longer.

Does the layout still look good with really long political status? Do we have examples of that? If one political status is very long, does that make all the columns very wide (unnecessarily so?) It seems a small price that *one* column is unnecessarily wide for simplicity of not having grid bag layout. That line of thought makes me wonder if we really always want all columns to be equal width then. Perhaps we could solve it by having longer political status messages wrap to preserve column widths. Ignoring the first column, a long political status making all columns very wide might be adverse.

asvitkine: The behavior will be that all the country columns will be made that wide.

I'm not sure we need to worry about this though. Such long political statuses might not actually exist on any maps. If they do, then we'll hear about it in a bug report and can adjust this logic. With the current logic, it would be simple enough to adjust to have some max width for the struts added.

For the separators, currently we add 1 per line. With GridLayout, we'd have to add N instead. So this will increase complexity. And then we'd have to make sure the code ensures there's no gaps and it restricts us in the future (i.e. we can't have the separators be different height than the cells with GridLayout). So I don't think we should go that route.

game-core/src/main/java/games/strategy/triplea/ui/PoliticsPanel.java
asvitkine: This code is applying to the Politics dialog, which is modal. While it is active, it's not possible to type D in chat, because the main window can't receive focus.

I can confirm that typing D in chat activates the shortcut in the main window when this dialog is not open, but this is not related to my change.

game-core/src/main/java/games/strategy/triplea/ui/PoliticalStateOverview.java
DanVanAtta: > I'm not sure we need to worry about this though. Such long political statuses might not actually exist on any maps. 

That is argument for grid layout, though I would agree we can add a max width later and controlling the height of the separators is compelling.

game-core/src/main/java/games/strategy/triplea/ui/PoliticsPanel.java
DanVanAtta: Thanks for double checking that :smiley:

PR #5321, opened 2019-10-09
Remove duplicate dependency in http-clients build.gradle file

DanVanAtta:
(1) Remove duplicated dependency on openfeign
(2) Fix formatting of testImplementation dependency

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[x] No manual testing done
[] Manually tested

### DIFF
--- a/http-server/build.gradle
+++ b/http-server/build.gradle
@@ -30,6 +30,7 @@ dependencies {
     implementation "io.dropwizard:dropwizard-core:$dropwizardVersion"
     implementation "io.dropwizard:dropwizard-jdbi3:$dropwizardVersion"
     implementation "io.dropwizard:dropwizard-auth:$dropwizardVersion"
+    implementation 'io.github.openfeign:feign-gson:10.2.0'
     implementation 'org.jdbi:jdbi3-core:3.8.2'
     implementation 'org.jdbi:jdbi3-sqlobject:3.8.2'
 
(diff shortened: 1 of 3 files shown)


http-server/build.gradle
```diff
@@ -30,6 +30,7 @@ dependencies {
     implementation "io.dropwizard:dropwizard-core:$dropwizardVersion"
     implementation "io.dropwizard:dropwizard-jdbi3:$dropwizardVersion"
     implementation "io.dropwizard:dropwizard-auth:$dropwizardVersion"
+    implementation 'io.github.openfeign:feign-gson:10.2.0'
```
RoiEXLab: What about extracting the version number to a global variable just like the dependencies above are doing it?

RoiEXLab:
Ah nevermind, it was obvious but just outside of the diff.
Figured it out on my own

http-server/build.gradle
DanVanAtta: Since the value is repeated, it makes sense. I wonder if we are adding too much ceremony to our dependencies. Do you recall @RoiEXLab why we switched from `compile` to `implementation`?

RoiEXLab: IIRC this was because gradle updated their build model and made `compile` deprecated.
I believe this has some further implications regarding transitive dependencies, but maybe that was something else I recall from the android build plugin (where there is also `api` as an option)

DanVanAtta: The transitive dependency behavior does seem to be a major difference, it was a bit of a surprise. I'm wondering which is better, explicit dependencies or reduced repetition by inheriting transitive dependencies.

PR #5322, opened 2019-10-09
Isolate Chat Messengers Dependencies and Chat Refactor Improvements

DanVanAtta:
Messengers isolating changes:
(1) Inject ChatTransmitter dependency into Chat
(2) Update ChatClient#messageReceived API to take in playerName,
    this way we can pass in the player name from MessengersChatTransmitter
    and not rely on messengers specific code
(3) Break circular dependency in ChatTransmitter constructor on ChatClient,
    use new setter: ChatTransmitter#setChatClient instead

Chat cleanups/refactors:
(4) Move 'status' to ChatParticipant, avoids map of
    'chatParticipant -> status' and converts to set of 'chatParticipant'
(5) Rename 'JavaSocketChatTransmitter' -> 'MessengersChatTransmitter'

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[x] Manually tested

Did a quick smoke-test, connected to a local lobby and confirmed chat and setting player status worked.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/chat/Chat.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/Chat.java
@@ -2,15 +2,12 @@ package games.strategy.engine.chat;
 
 import com.google.common.collect.EvictingQueue;
 import games.strategy.engine.lobby.PlayerName;
-import games.strategy.engine.message.MessageContext;
-import games.strategy.net.Messengers;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
-import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.BiConsumer;
@@ -33,7 +30,7 @@ public class Chat implements ChatClient {
   @Getter(AccessLevel.PACKAGE)
   private final SentMessagesHistory sentMessagesHistory;
 
-  private final Map<ChatParticipant, String> chatters;
+  private final Collection<ChatParticipant> chatters;
 
   @Getter
   private final Collection<ChatMessage> chatHistory =
@@ -44,17 +41,18 @@ public class Chat implements ChatClient {
   private final Collection<BiConsumer<PlayerName, String>> statusUpdateListeners =
       new ArrayList<>();
 
-  public Chat(final Messengers messengers, final String chatName) {
-    this.localPlayerName = messengers.getLocalNode().getPlayerName();
-    chatTransmitter = new JavaSocketChatTransmitter(this, chatName, messengers);
+  public Chat(final ChatTransmitter chatTransmitter) {
+    this.localPlayerName = chatTransmitter.getLocalPlayerName();
+    this.chatTransmitter = chatTransmitter;
+    chatTransmitter.setChatClient(this);
     sentMessagesHistory = new SentMessagesHistory();
-    chatters = Optional.ofNullable(chatTransmitter.connect()).orElseGet(HashMap::new);
+    chatters = Optional.ofNullable(chatTransmitter.connect()).orElseGet(HashSet::new);
     updateConnections();
   }
 
   private void updateConnections() {
     final List<ChatParticipant> playerNames =
-        chatters.keySet().stream()
+        chatters.stream()
             .sorted(Comparator.comparing(c -> c.getPlayerName().getValue()))
             .collect(Collectors.toList());
 
@@ -62,8 +60,7 @@ public class Chat implements ChatClient {
   }
 
   @Override
-  public void messageReceived(final String message) {
-    final PlayerName from = MessageContext.getSender().getPlayerName();
+  public void messageReceived(final PlayerName from, final String message) {
     if (isIgnored(from)) {
       return;
     }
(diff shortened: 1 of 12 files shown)


game-core/src/main/java/games/strategy/engine/chat/Chat.java
```diff
@@ -44,26 +41,26 @@
...
+    chatters = Optional.ofNullable(chatTransmitter.connect()).orElseGet(HashSet::new);
     updateConnections();
   }
 
   private void updateConnections() {
     final List<ChatParticipant> playerNames =
-        chatters.keySet().stream()
+        chatters.stream()
             .sorted(Comparator.comparing(c -> c.getPlayerName().getValue()))
             .collect(Collectors.toList());
```
RoiEXLab: I wonder if it would make sense to turn `chatters` into a `TreeSet` or something and use this custom comparator.
This has multiple benefits:
1. We don't need to sort it every time a connection is updated
2. The lookup below in linear time `O(n)`, could potentially be done in `O(log n)` using [`TreeSet#ceiling(E)`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/TreeSet.html#ceiling(E)) or something. I haven't found a guarantee for this time-complexity but after googling this, the implementation seems to suggest it.

Of course this only works if there aren't multiple `ChatParticipant`s with the same player name.

game-core/src/main/java/games/strategy/engine/chat/Chat.java
```diff
@@ -125,10 +125,10 @@ void updateStatus(final String status) {
   }
 
   String getStatus(final PlayerName playerName) {
-    return chatters.entrySet().stream()
-        .filter(n -> n.getKey().getPlayerName().equals(playerName))
+    return chatters.stream()
+        .filter(n -> n.getPlayerName().equals(playerName))
         .findAny()
```
RoiEXLab: Even if you're not convinced by my suggestion above, I'd appreciate if this lookup would be extracted to a private helper method

game-core/src/main/java/games/strategy/engine/chat/ChatTransmitter.java
```diff
@@ -1,18 +1,17 @@
...
 public interface ChatTransmitter {
-  /**
-   * Connects to chat, adds current player and returns the set of players that are in chat mapped to
-   * their status.
-   */
-  Map<ChatParticipant, String> connect();
+  void setChatClient(ChatClient chatClient);
+
+  /** Connects to chat, adds current player and returns the set of players that are in chat. */
+  Collection<ChatParticipant> connect();
```
RoiEXLab: Hmm are you sure it's good to go with a such a broad collection type?
I feel like specifying it's a set, makes it clear that there are no duplicate values.

game-core/src/main/java/games/strategy/engine/chat/MessengersChatTransmitter.java
```diff
@@ -4,38 +4,39 @@
...
+  public void setChatClient(final ChatClient chatClient) {
     chatChannelSubscriber = chatChannelSubscriber(chatClient);
   }
 
   private IChatChannel chatChannelSubscriber(final ChatClient chatClient) {
     return new IChatChannel() {
       @Override
       public void chatOccurred(final String message) {
-        chatClient.messageReceived(message);
+        chatClient.messageReceived(MessageContext.getSender().getPlayerName(), message);
```
RoiEXLab: I still don't like the `MessageContext` class :/

DanVanAtta: Totally, please take note where this line is being removed [from in Chat.java](https://github.com/triplea-game/triplea/pull/5322/files#diff-8ffdcb54556ca7b66cd59050e599e646L66)

That blew up when switching out the chat transmitter type.

game-core/src/main/java/games/strategy/engine/chat/Chat.java
```diff
@@ -44,26 +41,26 @@
   private final Collection<BiConsumer<PlayerName, String>> statusUpdateListeners =
       new ArrayList<>();
 
-  public Chat(final Messengers messengers, final String chatName) {
-    this.localPlayerName = messengers.getLocalNode().getPlayerName();
-    chatTransmitter = new JavaSocketChatTransmitter(this, chatName, messengers);
+  public Chat(final ChatTransmitter chatTransmitter) {
+    this.localPlayerName = chatTransmitter.getLocalPlayerName();
+    this.chatTransmitter = chatTransmitter;
+    chatTransmitter.setChatClient(this);
```
RoiEXLab: Ah, another thing I think is weird.
I really feel like there shouldn't be a need to have this kind of circular dependency, but ok 🤔

game-core/src/main/java/games/strategy/engine/chat/MessengersChatTransmitter.java
RoiEXLab: Looking forward to the day where this class can be safely removed ^^

game-core/src/main/java/games/strategy/engine/chat/Chat.java
DanVanAtta: I said in the commit comment the circular dependency is broken, to one extent it is, but really it is not. More fundamentally, Chat as a constructor arg to ChatTransmitter, and now ChatTransmitter as an arg to Chat, that was an unworkable dependency, hence the setter.

PR #5323, opened 2019-10-09
build.gradle consistency updates: dependencies

DanVanAtta:
1) Use variables for all version numbers
2) Alpha-sort all dependencies and version variables
3) Use double quotes for all dependencies (notably including project dependencies)

---

Updates dropwizard version from 1.3.14 to 1.3.15

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[x] No manual testing done
[] Manually tested

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -65,14 +65,41 @@ subprojects {
     version = getProductVersion()
 
     ext {
-        dropwizardVersion = '1.3.14'
+        awaitilityVersion = '3.1.6'
+        commonsCliVersion = '1.4'
+        commonsCodecVersion = '1.11'
+        commonsIoVersion = '2.6'
+        commonsMathVersion = '3.6.1'
+        databaseRiderVersion = '1.5.2'
+        dropWizardTestingVersion = '1.3.12'
+        dropwizardVersion = '1.3.15'
+        equalsVerifierVersion = '3.1.4'
+        errorProneVersion = '2.3.3'
+        guavaVersion = "27.0.1-jre"
+        guavaVersion = '27.0.1-jre'
+        hamcrestJsonVersion = '0.2'
+        hamcrestOptionalVersion = '2.0.0'
         hamcrestVersion = '2.0.0.0'
-        jbcryptVersion = '0.4'
-        junitJupiterVersion = '5.4.0'
-        mockitoVersion = '2.24.5'
-        postgresqlVersion = '42.2.5'
-        sonatypeGoodiesPrefsVersion = '2.2.5'
         jakartaMailVersion = '1.6.4'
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -65,36 +65,60 @@ subprojects {
...
     ext {
-        dropwizardVersion = '1.3.14'
+        awaitilityVersion = '3.1.6'
+        commonsCliVersion = '1.4'
+        commonsCodecVersion = '1.11'
+        commonsIoVersion = '2.6'
+        commonsMathVersion = '3.6.1'
+        databaseRiderVersion = '1.5.2'
+        dropWizardTestingVersion = '1.3.12'
+        dropwizardVersion = '1.3.15'
```
DanVanAtta: Note, dropwizard version updated here from 1.3.14 to 1.3.15

For the most part this is just a refactoring update, there are one or two other places where version numbers were brought inline.

DanVanAtta:
> Hmm, I'm not convinced extracting the version number for all dependencies.

Good consideration, my thoughts:
- having some versions be variables, others not seemed inconsistent
- we've a lot of ceremony to add dependencies, at least having one way to do things, even if elaborate - simplifies. Part of the idea is to reduce the number of decisions and refactoring. For example, if a dep was added to a new sub-project, suddenly you need to create a variable.
- found some issues, some version numbers were not the same, and we had inconsistency with adding the variables. Perhaps notably it was due to the refactoring issue when adding a new dep to a new sub-project.

DanVanAtta:
One more point:

> Use double quotes for all dependencies (notably including project dependencies)

It looks like gradle prefers as a style to use single quotes. If some deps had variables, others not, it was kinda ugly to see a mixture of single quotes and double quotes. It was also annoying to refactor a dep to use double quotes or go back to single quotes depending on if it had a variable. So for example, we add a brand new sub-project that had the same dep as another, then suddenly we need to refactor the other sub-project to have double quotes and a variable, even if it is not being touched. To re-iterate, I also really don't like how we have to 'think' whether to use a variable or not, it's not an important question to have to consider.

DanVanAtta:
Need this merged to avoid resolving conflicts. I'm pretty strong on this:
1. We *failed* to consistently use variables for the common dependencies
2. It is busy work to worry about whether to use variables or not, we should be concerned/worrying about other things and not counting how many sub-projects use a given dependency or worrying if something is shared now vs not, and converting single/double quotes.

In other words, if we don't use variables everywhere, how can we ensure future success where we failed before?

PR #5324, opened 2019-10-09
Remove empty project

RoiEXLab:
Noticed this, and it seems unused...


RoiEXLab:
Oh and btw @DanVanAtta it would be nice if the HTTP server had it's own intelliJ launch configuration

DanVanAtta:
@RoiEXLab naming is not ideal, though launch config is checked in:  https://github.com/triplea-game/triplea/blob/master/.idea/runConfigurations/ServerApplication.xml

DanVanAtta:
@RoiEXLab naming is not ideal, though launch config is checked in:  https://github.com/triplea-game/triplea/blob/master/.idea/runConfigurations/ServerApplication.xml

PR #5325, opened 2019-10-09
Add domain-data subproject

DanVanAtta:
1 commit on top of: https://github.com/triplea-game/triplea/pull/5323

### Overview

    
New sub-project for domain-entities that are heavily shared between multiple, potentially low-level sub-projects. For example, at the DB layer, it makes sense to have API's to accept 'PlayerName', it has become a bit untenable to keep the DB layer to pure primitive java entities. It is also getting very odd that common domain objects are converging to 'http-clients' (where they do not belong).
    
It is a risk we'll add too many or inappropriate domain objects to the new sub-project, we should be judicious about having only true domain and heavily shared objects land in 'domain-data'. Specifically, with multiple sub-domains, a term can very well be defined in multiple sub-domains and have different meanings. If we place such objects in 'domain-data', the multiple means will conflict, this is something to avoid.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[x] No manual testing done
[] Manually tested


RoiEXLab:
Oh and someone needs to remove/fix the synk project configurations.
The old project causes the build to fail and I believe new projects won't get picked up either.

DanVanAtta:
Chat is getting closer to ready, it's 3~5 very different sets of changes. There is some need to get things moving. This update is pretty spread out in the number of files affected and was easy to cut individually. I'm not sure if the others will be as easy to carve off and we they may wind up landing as a slow-to-review large PR.

W.r.t to this update, I've been really on the fence about having a domain data subproject. Twice before I created it and decided better against it. I wonder how many objects we'll wind up having in it, and how we can handle conflicts of sub-domains using terminology different. Just notably "Player" in maps means something a bit different compared to a "game player". Game-core and Maps seems to be the biggest sub-domains we have, I think AI is another though it is married pretty well to game-core.

The deciding factor here was the need/desire to use domain objects at the DB layer and http-client shared code layer. Notably for DB I wanted to create a wrapper around API keys to hide that they are hashed before being stored. One of the original idea's for DB layer was to only use primitive types as dependencies, but for such a wrapper that started to seem less ideal.

I think we'll find out how things work when we have new sub-domains / sub-projects carved off from the 'bottom' of the project. Notably things like "map parsing". Within game play there are a number of sub-domains like the core engine, the loading process, the lobby client.

I think at the end of the day, if we find that domain-data is too broad or not going to be useful enough, we can perhaps re-merge domain objects easily enough if we do not change their packages.

PR #5336, opened 2019-10-11
Consolidate TestData to utility classes

DanVanAtta:
- Reduces redundancy
- Extracts data setup from unit tests
- Allows for future data changes to be done in one place

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[x] No manual testing done
[] Manually tested

### DIFF
--- a/game-core/src/test/java/games/strategy/engine/lobby/client/ui/GamePollerTaskTest.java
+++ b/game-core/src/test/java/games/strategy/engine/lobby/client/ui/GamePollerTaskTest.java
@@ -9,7 +9,6 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 import feign.FeignException;
-import java.time.Instant;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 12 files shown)


DanVanAtta:
> Hmm, not sure if I like that there are 3 classes with the same name and almost the same content, this might cause some confusion in the future.

How would you envision that confusion manifesting? I'm thinking that given they all serve the same purpose, the scoping of the different objects available will prevent any errors/issues.

RoiEXLab:
I always think of the most common scenario how I navigate through code:
Let's say I create a PR with some refactoring changes and forget fixing the tests. The build fails and in the logs I see the class name with the issue.
I go back to the code, change the data and for some reason the test still fails.
Until I realise that I was simply looking at the wrong class at least 5 minutes will have passed.

I don't think this is an absolute blocker, I'm just saying that this might potentially be a think too look out for

DanVanAtta:
Compilation errors should be highlighted in-file, the path is fully qualified in logs, and searching for a class and seeing multiple results I hope is all enough where it'll never take 5 minutes: 
Screenshot from 2019-10-11 18-23-04

DanVanAtta:
Agree there could be confusion perhaps, but I think it's a bigger problem if the classes served different purposes and weren't standardized in form/purpose. I don't think artificially naming them otherwise at this juncture quite makes sense either.

PR #5334, opened 2019-10-11
Add Chat 'Connected' API and use 'set' for chatters

DanVanAtta:
- Connected API is to provide a callback for chat transmitter
  to update with the current list of players. Previously the
  current list of players was a synchronous-only operation done
  at the time of connection. With websockets that will become
  an async operation necessitating a callback.

- Use 'set' for chatters on Client. Chatters joining chat receive
  a connected message and will receive a join message for their
  own connection. The 'set' prevents the duplicate add between
  the initial player list and the join message.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature update or enhancement
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[x] Other:  Add forward looking API  

## Testing

[x] No manual testing done
[] Manually tested

PR #5337, opened 2019-10-11
Wire IP address to api-key-generator

DanVanAtta:
Future looking update to send IP data to Api key generation. Future
updates will store IP address with API-keys. This way when we ban
an IP, we can reference the corresponding API keys and mark them
as invalid.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[x] No manual testing done
[] Manually tested

### DIFF
--- a/http-server/src/main/java/org/triplea/server/access/ApiKeyGenerator.java
+++ b/http-server/src/main/java/org/triplea/server/access/ApiKeyGenerator.java
@@ -1,8 +1,10 @@
 package org.triplea.server.access;
 
 import com.google.common.hash.Hashing;
+import java.net.InetAddress;
 import java.nio.charset.StandardCharsets;
 import java.util.UUID;
+import java.util.function.Function;
 import java.util.function.Supplier;
 import javax.annotation.Nonnull;
 import lombok.Builder;
@@ -12,7 +14,7 @@ import org.triplea.lobby.server.db.dao.UserJdbiDao;
 
 // TODO: Project#12 This is copy/pasted from lobby sub-project
 @Builder
-public class ApiKeyGenerator implements Supplier<ApiKey> {
+public class ApiKeyGenerator implements Function<InetAddress, ApiKey> {
 
   @Nonnull private final ApiKeyDao apiKeyDao;
   @Nonnull private final UserJdbiDao userDao;
@@ -39,7 +41,7 @@ public class ApiKeyGenerator implements Supplier<ApiKey> {
   }
 
   @Override
-  public ApiKey get() {
+  public ApiKey apply(final InetAddress inetAddress) {
     final ApiKey key = keyMaker.get();
     apiKeyDao.storeKey(null, key.getValue());
     apiKeyDao.deleteOldKeys();
--- a/lobby/src/main/java/org/triplea/lobby/server/api/key/ApiKeyGeneratorFactory.java
+++ b/lobby/src/main/java/org/triplea/lobby/server/api/key/ApiKeyGeneratorFactory.java
@@ -21,6 +21,7 @@ public final class ApiKeyGeneratorFactory {
             .apiKeyDao(jdbi.onDemand(ApiKeyDao.class))
             .userDao(jdbi.onDemand(UserJdbiDao.class))
             .build();
+    // TODO: Project#12 replace ApiKeyGeneartor with ApiKeyDaoWrapper
     return (name, ip) -> keyGenerator.apply(name);
   }
 }
--- a/lobby/src/main/java/org/triplea/lobby/server/api/key/ApiKeyGeneratorFactory.java
+++ b/lobby/src/main/java/org/triplea/lobby/server/api/key/ApiKeyGeneratorFactory.java
@@ -1,6 +1,7 @@
 package org.triplea.lobby.server.api.key;
 
-import java.util.function.Function;
+import java.net.InetAddress;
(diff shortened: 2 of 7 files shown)


http-server/src/main/java/org/triplea/server/access/ApiKeyGenerator.java
```diff
@@ -39,7 +41,7 @@ private static String randomHash() {
   }
 
   @Override
-  public ApiKey get() {
+  public ApiKey apply(final InetAddress inetAddress) {
```
RoiEXLab: Why is the parameter added here?
Doesn't seem to be used anywhere.

lobby/src/main/java/org/triplea/lobby/server/api/key/ApiKeyGeneratorFactory.java
```diff
@@ -13,11 +14,13 @@
...
-        .userDao(jdbi.onDemand(UserJdbiDao.class))
-        .build();
+  public static BiFunction<PlayerName, InetAddress, ApiKey> newApiKeyGenerator(final Jdbi jdbi) {
+    final ApiKeyGenerator keyGenerator =
+        ApiKeyGenerator.builder()
+            .keyMaker(ApiKeyGenerator.createKeyMaker())
+            .apiKeyDao(jdbi.onDemand(ApiKeyDao.class))
+            .userDao(jdbi.onDemand(UserJdbiDao.class))
+            .build();
+    return (name, ip) -> keyGenerator.apply(name);
```
RoiEXLab: Why is this parameter being added if it's ignored anyways?

http-server/src/main/java/org/triplea/server/access/ApiKeyGenerator.java
DanVanAtta: Nope, it's not. This class will be deleted entirely in short order. The wiring is 'future looking'. It's basically the same as a feature flag, any code behind a feature flag is just as dead as this parameter.

DanVanAtta:
@RoiEXLab if you want to see the future version, check out: https://github.com/triplea-game/triplea/pull/5333/files

That is still a 2k line PR, the next step, all the DB updates, is perhaps 1000 lines of updates. At least avoiding the wiring changes to be included there can let us focus on how the DB queries are updates and not all the plumbing code.

lobby/src/main/java/org/triplea/lobby/server/api/key/ApiKeyGeneratorFactory.java
DanVanAtta: From the overview:

> Future looking update to send IP data to Api key generation. Future
updates will store IP address with API-keys.

DanVanAtta:
It's also better to discuss now if it does not make sense to store IP address in the api-key table compared to realizing all those updates to discuss it then. Horse has already left the barn, but it would at least save the break-up work; Spending multiple days doing nothing but breaking up a large update is not necessarily enjoyable, on the other hand submitting a PR over 500 is unreviewable. It's a joke IMO to think that kind of PR gets the right level of scrutiny. I think it's more important going forward here that we keep in mind whether these are the right changes, if a few lines can be improved here or there really does not matter compared to the overall effort.

DanVanAtta:
@RoiEXLab I feel the line you drew was perhaps arbitrary. Did you notice that in this update: https://github.com/triplea-game/triplea/pull/5334, the new `connected` method is not called anywhere? I'm not sure if that is rightly worse than an unused parameter, but it's arguably the same ballpark. Hence I'm just very surprised you drew a line now on this and not before.

RoiEXLab:
Haven't noticed that bit in the other PR, likely because of the large size.
Anyways, feel free to merge.

DanVanAtta:
Please note the number, #5334 is merged. @RoiEXLab happy to discuss PR strategies if you'd like. I'd agree that this specific update is reaching pretty far to have *something* that is separate and mergable. It's not necessarily easy to land very large updates and keep things reviewable. I think such an update where we have dead/unused code is probably okay so long as:
- we have the code marked where there is more work. In this case I'm missing the TODOs that would note that (I did not add those in to avoid merge conflicts, it's laziness on my part so no real excuse)
- the submitter is sure they are going to have the capacity to follow up and complete the TODOs in short order. Having 'dangling' work like this living for a long time would be bad.

I think we can all recall some of the initial clean-ups of TripleA that we did where there were relatively massive and incomplete features that were merged.

http-server/src/main/java/org/triplea/server/access/ApiKeyGenerator.java
DanVanAtta: The class is marked with:
```
// TODO: Project#12 This is copy/pasted from lobby sub-project
```
I'm skipping adding a new TODO here as the whole class will be deleted in future iterations.

The other location that was of question, I added a TODO to mark that we will use the unused parameter in future iterations.

lobby/src/main/java/org/triplea/lobby/server/api/key/ApiKeyGeneratorFactory.java
DanVanAtta: Added a TODO here to note that this will be updated.

PR #5340, opened 2019-10-13
Add tooltips to history country nodes.

asvitkine:
This allows mouse-overing the nodes to see the country name, which is useful for maps with lots of countries where the player does not always remember the icons.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[X] Manually tested

Tested by opening a saved game with history and mouse-overing the nodes.

### Before

tooltip2_before

### After

tooltip2_after


DanVanAtta:
@asvitkine, I'm curious, why the "merge pull request" commit?

ie:
```
Merge pull request #4 from triplea-game/master
asvitkine committed 8 days ago
```

FWIW, in case it's of any use, my process to remove such commits:
-  I set my local env by cloning the main repo, I keep the remote called 'origin', then I add a remote for my fork and call that 'fork'. 
-  Then when cutting new branches, I use a script called [gup](https://github.com/DanVanAtta/git_tools/blob/master/gup) to rebase to the latest, which usually drops off any merge commits.

---------

Just kinda curious, since we do squash-rebase merges, the additional commit drops off anyways, so it does not matter too much.

game-core/src/main/java/games/strategy/triplea/ui/history/HistoryPanel.java
```diff
@@ -447,12 +450,13 @@ public Component getTreeCellRendererComponent(
       if (value instanceof Step) {
         final PlayerId player = ((Step) value).getPlayerId();
         if (player != null) {
+          final String text = value.toString() + " (" + player.getName() + ")";
           if (uiContext != null) {
             super.getTreeCellRendererComponent(tree, value, sel, expanded, leaf, row, haveFocus);
             icon.setImage(uiContext.getFlagImageFactory().getSmallFlag(player));
             setIcon(icon);
+            setToolTipText(text);
           } else {
```
DanVanAtta: Side note, it seems odd for UI text to be null, I wonder if that is a dead code branch :thinking:

asvitkine:
> Would it make sense to instead to have a tooltip on just the country flag?

That was actually my plan originally - but looks like the country flag is just an icon and not a JComponent itself, so it didn't seem like that could easily be done without redoing the structure.

DanVanAtta:
I was afraid/concerned it was an icon. The UI might have alignment issues if we have a flag icon and then a label as two different components. Looking at the diff, it looked like that perhaps could have been an option.

I think probably most ideal is to split the two components (flag and text), and have the hover text for the phase text be a 'summary' to avoid needing to open the history node. I think we can iterate on this if it's feasible, in the meantime let's merge so we capture the benefit of this update.

PR #5341, opened 2019-10-13
Fix slow save loading on "Start Local Game".

asvitkine:
This fixes a bug with very slow save game loading when on the "Start Local Game" screen.
The cause was that MainPanel.accept() would add itself as an observer to the panel each time it's called, causing an exponential number of instances of itself being added to the observer list. The fix is to ensure there are no duplicate observers.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Bug fix: 
[] Other:   

## Testing

[] No manual testing done
[X] Manually tested

Opened "Start Local Game" and used "Open Saved Game" button. Before this change, it would take a long time to load. With this change, it's instant.

Also confirmed with debug logging in MainPanel.accept(). The above operation would previously result in accept() being called thirty thousand times. Now, it's called 17 times.


game-core/src/main/java/games/strategy/engine/framework/startup/ui/SetupPanel.java
```diff
@@ -32,7 +32,9 @@
   private final List<Consumer<SetupPanel>> listeners = new CopyOnWriteArrayList<>();
 
   public void addObserver(final Consumer<SetupPanel> observer) {
-    listeners.add(observer);
+    if (!listeners.contains(observer)) {
```
DanVanAtta: - I wonder if the observer list should actually be a set?
- Is there another fix out there so we don't try to add duplicate Observers?

asvitkine: > I wonder if the observer list should actually be a set?

I considered changing it from CopyOnWriteArrayList to CopyOnWriteArraySet. But I think set iteration order is not defined and I didn't want to change any behaviour beyond this fix.

> Is there another fix out there so we don't try to add duplicate Observers?

Well, the calling code could ask this class if it's already an observer and not add itself, but that seems extra complexity for no reason.

DanVanAtta: My guess/bet would be that the call order does not matter. Considering the caller spams 'addObserver', any ordering dependency is perhaps more accidental then intentional.

> Well, the calling code could ask this class if it's already an observer and not add itself, but that would just be extra complexity without a benefit.

Sounds reasonable.

PR #5342, opened 2019-10-13
Make "Flag Display Mode" menu items mutually exclusive with radio buttons.

asvitkine:
Before this change, the items did not have decorators indicating which is selected. This change adds this functionality.
Note: In 1.9, this submenu was different and there were checkboxes next to the items, but the changes to this menu in 2.0 made the options mutually exclusive already, so this PR expands on that by making them use a proper radio UI.

## Functional Changes
[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[X] Manually tested

Testing:
1.  Open the "Flags Display Mode" submenu in the View menu on the game.
2. Verify that an item is selected and matches the display mode.
3. Toggle the state and verify that the menu items are updated with an indicator about which is active.
4. Toggle the state from the "Unit Scroller" UI and check that the selected menu item gets updated accordingly.
 
### Before
before_radio

### After

radio

Note: On Mac, radio menu items are still displayed with checkboxes. On other platforms, the indicators may be different.


DanVanAtta:
@asvitkine I'm afraid to admit you're running into a hack of mine.

The hack is that the 'flag display mode' button in the unit scroller area is not hooked up to the menu. That means if you change the flag display mode by clicking the button or using the hotkey, the menu becomes out of date.

For example:
- use menu to set small flags
- cycle to a combat move phase, click the flag button to see large flags
- now go back to the menu, notice 'small flags' is still selected

By not indicating a current state, we were getting around this problem, hence the hack.

Side-note, it would be ideal if the button icon also reflected the current state. IE: small flags would be a small flag icon, large flags would be the current 'large' flag icon, and no flags would perhaps do a circle with bar over the flag.

I think the choices we have are as follows:
- remove the menu option (my original preference but there is not consensus around this)
- remove the button in unit scroller and the key binding (1.9 had a hot-key that was unbound to any UI controls, the button was added to solve that problem)
- remove the selection indicator from the menu (essentially the current 2.0 behavior)
- add a listener to the key-binding/button to update the menu option (probably ideal)

asvitkine:
Thanks for the extra context. I'll take a look at what it would take to make (d) happen.

asvitkine:
Please take another look.

I implemented updating of the menu based on the underlying state. We only need it when the menu is getting opened, which works well.

DanVanAtta:
Good work on this @asvitkine :+1:

PR #5343, opened 2019-10-13
Alternative SetupPanel listener fix

RoiEXLab:
This is an alternative fix to #5341 that avoids checking if the listener does actually exist already.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[x] Manually tested

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/SetupPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/SetupPanel.java
@@ -9,7 +9,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.Consumer;
 import javax.annotation.Nullable;
 import javax.swing.Action;
@@ -29,18 +28,16 @@ public abstract class SetupPanel extends JPanel implements SetupModel {
   private static final long serialVersionUID = 4001323470187210773L;
   private static final String SET_ALL_DEFAULT_LABEL = "Default";
 
-  private final List<Consumer<SetupPanel>> listeners = new CopyOnWriteArrayList<>();
+  @Nullable private Consumer<SetupPanel> listener;
 
-  public void addPanelChangedListener(final Consumer<SetupPanel> listener) {
-    listeners.add(listener);
+  public void setPanelChangedListener(final Consumer<SetupPanel> listener) {
+    this.listener = listener;
   }
 
-  public void removePanelChangedListener(final Consumer<SetupPanel> listener) {
-    listeners.remove(listener);
-  }
-
-  void fireListeners() {
-    listeners.forEach(consumer -> consumer.accept(this));
+  void fireListener() {
+    if (listener != null) {
+      listener.accept(this);
+    }
   }
 
   @Nullable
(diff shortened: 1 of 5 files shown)


DanVanAtta:
@asvitkine has better context on this update than I do. @asvitkine, I'm curious for your thoughts here.

asvitkine:
I like the name change from Observer to listener, since Observer is confusing with java.util.Observer.

I'm not sure how I feel about having the logic on the calling side, rather than in the API. Yes, technically the bug is on the calling side and this is fixing it. But the downside are:
  - It still leaves the possibility of other things registering themselves multiple times, which doesn't seem like it would ever be desired.
   - The calling code is more complex.
   - The list gets modified, which is more expensive than just checking membership and not adding.

RoiEXLab:
There are some benefits to this approach as well:
Namely Listeners are now cleaned up when they are no longer needed, reducing the risk of a memory leak. Currently they exist for the lifetime of all the setuppanels and keep a reference to the mainpanel.
Also it's more consistent with the way the majority of listeners work in the triplea codebase: Almost all of them use a list as backing datastructure, and a lot of then provide a way to remove the listener when it's no longer needed.

asvitkine:
That's a good point. Perhaps the best of both worlds would be to have a helper class for lists of observers that could internally do a check that an observer is not added twice when add() is called?

Then we could use this class for all observer lists and it could find such bugs in other places.

DanVanAtta:
Good points! I think before a helper class we should convert to a `Set`. Removing a non-existent `set` member is a no-op and we can avoid the helper class. I'm 95% sure we do not need ordering of listeners.

DanVanAtta:
@RoiEXLab I'm not positive about the 'removeListener'. There are lots of examples of 'removeListener' in the code, but I think those are largely Java legacy. Listeners in the very old days often were implemented with static variables, hence the need to remove listeners. AFAIK modern listener patterns do not explicitly remove listeners and instead rely on objects to fall out of scope. 

In this case, if the 'panel' where we are removing a listener does not fall out of scope, we have a pretty large memory leak. Removing the listener reference to it reduces the footprint of the leak, but we still have a large leak. Any thoughts @RoiEXLab ?

RoiEXLab:
@DanVanAtta In general I agree that remove should be unecessary in most cases.
In this case however because the main panel does allow setting the currently active panel, I think removing the listeners makes sense.

A thought from my side: It could be the case that only one listener is ever really registered, so it might be sufficient to just turn this list into a single field

DanVanAtta:
@RoiEXLab Do you know if the remove listener is strictly necessary? 

I agree it might be easier to have a single element member. That conversion could moot the above question perhaps.    Leaning on YAGNI, if we need multiple we can update it.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/SetupPanel.java
```diff
@@ -31,13 +31,15 @@
 
   private final List<Consumer<SetupPanel>> listeners = new CopyOnWriteArrayList<>();
 
-  public void addObserver(final Consumer<SetupPanel> observer) {
-    if (!listeners.contains(observer)) {
-      listeners.add(observer);
-    }
+  public void addPanelChangedListener(final Consumer<SetupPanel> observer) {
```
DanVanAtta: nit: 'observer' parameter name looks out of place now that we have listeners

RoiEXLab:
@DanVanAtta single field seems to work just fine, let me know if you're fine with my approach

PR #5346, opened 2019-10-13
launch_db: Busy wait loop and cleanup output

DanVanAtta:
(1) Current db launch script simply waits 5 seconds for DB to startup.
    This update changes that to add a wait loop that checks for DB
    startup every 0.2s and continues if it has started.
(2) Cleanup launch_db output, remove 'set -x' and add echo output to
    log the current action being taken.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[x] Other:  dev utility script enhancement 

## Testing

[] No manual testing done
[x] Manually tested

- Also ran shellcheck and fixed up shellcheck issues in 'launch_db'

### Before
Screenshot from 2019-10-13 10-52-38

### After
Screenshot from 2019-10-13 10-52-05

## Additional Review Notes

The stop of DB docker container is there to kill any dead containers. We may want to enhance this script in the future to detect if docker is running, if so then execute 'drop_db' to clear database and then re-run flyway and sample data. That should be faster. On the other hand, ensuring that we are going to successfully relaunch DB is of value (drop-db will fail if there are active connections, stopping container should likely kill those connections).

### DIFF
--- a/lobby-db/launch_db
+++ b/lobby-db/launch_db
@@ -34,7 +34,7 @@ function waitForDatabaseToStart() {
     tryAttempt=$((tryAttempt+1))
     # timeout after 10s
     if [ "$tryAttempt" -gt 50 ]; then
-      echo "Aborting DB startup timed out"
+      echo "Aborting DB startup (timed out)"
       exit 1
     fi
   done
--- a/lobby-db/launch_db
+++ b/lobby-db/launch_db
@@ -3,8 +3,7 @@
 # Convenience script to get a local DB up and running on docker
 # with some sample data.
 
-set -u
-set -e
+set -uu
 
 path=$(dirname "$0")
 "$path/run_docker"
--- a/lobby-db/launch_db
+++ b/lobby-db/launch_db
@@ -3,12 +3,40 @@
 # Convenience script to get a local DB up and running on docker
 # with some sample data.
 
-set -eu
+set -u
+set -e
 
-path=$(dirname $0)
+path=$(dirname "$0")
+"$path/run_docker"
 
-$path/run_docker
-sleep 5
-$path/run_flyway
-$path/load_sample_data
+set +e
+echo -n "Waiting for Database start"
+status=1
+tryAttempt=0
+while [ "$status" != 0 ]; do
+  echo 'select 1' \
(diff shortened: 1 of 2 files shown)


lobby-db/launch_db
```diff
@@ -3,12 +3,40 @@
 # Convenience script to get a local DB up and running on docker
 # with some sample data.
 
-set -eu
+set -u
+set -e
```
RoiEXLab: Why this change?
Shouldn't it be the exact same?

lobby-db/launch_db
```diff
@@ -3,12 +3,40 @@
...
-$path/load_sample_data
+set +e
+echo -n "Waiting for Database start"
+status=1
+tryAttempt=0
+while [ "$status" != 0 ]; do
+  echo 'select 1' \
+     | psql -h localhost -U lobby_user lobby_db 2> /dev/null \
+     | grep -q '1 row'
+  status=$?
```
RoiEXLab: In bash you can inline this check into the while condition.
This way there's no need to declare a status variable at all.

See https://stackoverflow.com/a/5274386

DanVanAtta: Indeed, it's preferred even. Something was going wrong with it though... I gave up after 20 minutes.

lobby-db/launch_db
DanVanAtta: Left over from debugging, should be converted back to how it was.

DanVanAtta: Fixed: cb976f3d7

lobby-db/launch_db
DanVanAtta: Just needed to spend a moment longer, managed to get it correctly formatted/syntaxed : )

It actually really helped to have had a working version first and then to iterate to get it right, my first attempt was to have the 'while' loop directly invoke the corresponding command via a variable.. 

Updated 223d6eb

DanVanAtta:
Did another iteration to cleanup the the script. Now extracted the 'wait for db' piece to a function. Please let me know what you think @RoiEXLab

PR #5351, opened 2019-10-16
Fix ClassCastException on JOptionPanes if Escape is pressed.

asvitkine:
Fix ClassCastException on JOptionPanes if Escape is pressed.

Repro instructions for the bug:
  1. Play a solo map of Global 40 2nd with all humans.
  2. On Germany combat, send a Bomber to London to Bomb.
  3. On the dialog to intercept, hit Escape.

Without this fix, a ClassCastException happens.
Screen Shot 2019-10-15 at 11 15 55 PM

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[X] Manually tested

See above.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
@@ -1539,7 +1539,7 @@ public final class TripleAFrame extends JFrame implements KeyBindingSupplier {
                 if (!dialog.isVisible()) {
                   return;
                 }
-                // Note: No cast to String, as we get JOptionPane.DEFAULT_OPTION on Escape.
+                // Note: getValue() can return either an int (user pressed escape), or a String.
                 final Object option = optionPane.getValue();
                 if (option.equals(optionNone)) {
                   unitPanels.clear();
@@ -1716,6 +1716,7 @@ public final class TripleAFrame extends JFrame implements KeyBindingSupplier {
                 if (!dialog.isVisible()) {
                   return;
                 }
+                // Note: getValue() can return either an int (user pressed escape), or a JButton.
                 final Object option = optionPane.getValue();
                 if (option == noneButton) {
                   choosers.clear();
@@ -1811,7 +1812,7 @@ public final class TripleAFrame extends JFrame implements KeyBindingSupplier {
                 if (!dialog.isVisible()) {
                   return;
                 }
-                // Note: No cast to String, as we get JOptionPane.DEFAULT_OPTION on Escape.
+                // Note: getValue() can return either an int (user pressed escape), or a String.
                 final Object option = optionPane.getValue();
                 if (option.equals(optionNone)) {
                   selection.clear();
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
```diff
@@ -1539,7 +1539,8 @@ public Component getListCellRendererComponent(
                 if (!dialog.isVisible()) {
                   return;
                 }
-                final String option = ((String) optionPane.getValue());
+                // Note: No cast to String, as we get JOptionPane.DEFAULT_OPTION on Escape.
```
DanVanAtta: 2 questions:
1) Will this comment make sense to someone reading the code by itself? IE: it makes sense seeing this comment in this diff, but without the original cast to string, would the comment make sense? Perhaps the note should be a warning that we can have either a String value or a `JOptionPane.DEFAULT_OPTION` be returned.

2) Comparing Object to String is perhaps slightly questionable. Would `String.valueOf(optionaPanel.getValue())` do the trick? In that case we'd be comparing String to String.

asvitkine: I think it does, as someone may have the same thought as you in your point (2) - where they may think it's odd that we're comparing Object to String and try to make the code make more sense by casting to a String. So I think the warning is appropriate. We could phrase it differently, if you have a suggestion though.

Re: 2 - I don't see a problem. In Java, it's OK to .compare() any object to any other (as long as you don't call .compare() on a null). So I don't see why we'd need the extra complexity and unnecessary work of String.valueOf().

DanVanAtta: It's not a problem per-say to compare object to string, but it can be brittle when comparing differing types and it can be a source of errors. Because of this, it may be examined more closely then it needs to be and/or flagged in static analysis.

No matter, perhaps it could be argued either way.

To consider, we have a 3rd usage of `optionPane.getValue()` in TripleAFrame:1718

```
                final Object option = optionPane.getValue();
                if (option == noneButton) {
```

Perhaps no comment at all is appropriate and this falls under the category of "know your API?"

If there would be a comment, we'd probably want it on the 3rd location, as a suggestion, something like:
```
// Note: getOption() can return either an int (user pressed escape), or a String
// Note: getOption() can return either an int (user pressed escape), or a JButton
```

I'm not sure if we should sign up to having a warning comment on every usage of JOptionPane. IMO a wrapper API of some sort might be a good way to go so that we fix the underlying API. Something like:

```
final Optional<String> option = SwingComponents.readOption(optionPane, String.class);
```

Summing up, the three options I can suggest (maybe there is something even better, but these are what I can suggest):
1. Add a comment on all 3 usages
2. No comment on any of the usages (and consistently assign to `Object` as you have)
3. Use a wrapper API that is safe

A 4th option that would involve a large refactor would be to use a custom builder component and replace the usage of the option dialog completely with a builder that solves the API problems. It's probably better to go with one of the above 3 options first though before we tackle that level of change.

asvitkine: Commenting on all 3 usages makes sense. I definitely prefer it over no-comment as this API is quite subtle. Done - went with your wording suggestion.

I agree that the 4th option would be best, but as you say, that's a larger change.

DanVanAtta:
Good to see this fixed @asvitkine , thank you  :+1:

Issue #5352, opened 2019-10-16
Parallel Integration Test Mayhem


DanVanAtta:
DB data between sub-projects is causing problems when building sub-projects in parallel. AFAIK devs should be using `./gradlew --parallel ...`  (or the `./verify` script)

An example of the problem, http-server, lobby, and DB-dao sub-projects all interact with DB data. We have a race-condition over which ones will clear and insert data into DB yielding non-deterministic test output. Adding more DB unit tests increases pressure on this race condition and it's hit a tipping point in some in-flight work I have.

How to solve? Some options:
1) We could no longer support parallel sub-project builds. The cost is slower build times for everyone on local development.
2) Move all tests that interact with a live database to a specific sub-project. Removing the DB dependency from all other sub-projects would allow sub-projects to be built in parallel. This sub-project could be called something like 'system-test' and be the only one that spins up live resources and interacts with DB. Removing the "real" DB dependency dependency from all other sub-projects would allow us to build in parallel sub-projects in parallel and not worry about tests trampling one-another.
3) Remove dependencies on live DB data. I'm not sure if this is feasible. It's a catch-22 for the lobby controller's, those are due to be removed in about 4 or 5 PRs from now, those PRs are running into this problem. The http server system test would be difficult to not have hit a live DB and we lose test coverage if we are not integrating with a real DB.

(2) seems to be the only viable option. Before making this update I want to get the opinions of the other contributors. Please let me know if the problem is not clear.

RoiEXLab:
From the options you've given I'd lean towards option 2 as well, however I wonder about this scenario:
Is it possible to isolate the test cases from each other? If we created a temporary db for each test case that would eliminate the problems, right?

DanVanAtta:
Having 'n' databases could be a challenge. We'd have to solve the performance problems, how to clone a baseline database and/or run DB migrations quickly, second configuration would be a challenge for dynamic ports, instance to be spun up at the right time. If we go with an in-memory route, something like H2, it still would be costly to spin that up per test and we would still have an integration test problem as eventually we would want to communicate to a real database and do so from multiple sub-projects.

Additive only data could be a good way to go so we avoid the deletion of data, which is the root cause behind these problems. Though, I don't know if that plays well with `@ExpectedData` tags. Second, we might have some brittleness with identifiers. For example, 'user_role' table IDs values would become 'magic' identifiers spread around, we could not add to that table as it has UK constraints.

[...]

PR #5354, opened 2019-10-16
User-role table and add role 'HOST'

DanVanAtta:
With user-role being tracked in two tables, to avoid duplication of user role info,
a user_role table is being created.

HOST role is added to indicate a role that has no username or role id. This is the
role that is specific for 'lobby watcher' (a specific connection used to connect
to lobby to send game status updates).

To support further updates on serve side, AuthenticatedUser is being updated to
have username information.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[x] No manual testing done
[] Manually tested


DanVanAtta:
This will probably fail due to: https://github.com/triplea-game/triplea/issues/5352

If we move all DB-dependent tests to a dedicated sub-project, the prod code updates will not change and we can get a head start on reviewing those pieces.

DanVanAtta:
I was expecting a failure as the local CLI build fails. Parallel sub-project builds look to be busted. Migrating to a single integration-testing sub-project seems to have fixed the issue. Given the overlap in code updates (merge conflicting), I'll hold off on submitting that update until after this has been merged.

RoiEXLab:
@DanVanAtta I'll try to review this PR this evening

http-server/src/main/java/org/triplea/server/access/RoleAuthorizer.java
```diff
@@ -9,27 +9,38 @@
...
-        return requestedRole.equals(UserRole.ANONYMOUS);
+        return anonymousPermissions(requestedRole);
+      case UserRole.HOST:
+        return requestedRole.equals(UserRole.HOST);
       default:
         throw new AssertionError("Unrecognized user role: " + user.getUserRole());
     }
   }
+
+  private static boolean adminPermissions(final String requestedRole) {
```
RoiEXLab: `adminPermissions` is a confusing name in my opinion.
Without looking at the method above I'd actually think this method returns true if the passed role has admin permissions, instead of if the admin is authorized to perform actions for the specified role.
So I'd suggest renaming the methods to `isAdminAuthorizedFor` or something. Can't really think of something better right now, haven't convinced myself yet 😅

lobby-db-dao/src/main/java/org/triplea/lobby/server/db/dao/api/key/ApiKeyDaoWrapper.java
```diff
@@ -0,0 +1,96 @@
...
+    final ApiKey key = keyMaker.get();
+    final int hostRoleId = userRoleDao.lookupHostRoleId();
+    insertKey(null, null, key, ip, hostRoleId);
+    return key;
+  }
+
+  /** Creates (and stores in DB) a new API key for registered or anonymous users. */
+  public ApiKey newKey(final PlayerName playerName, final InetAddress ip) {
+    Preconditions.checkArgument(playerName != null);
+    Preconditions.checkArgument(ip != null);
```
RoiEXLab: Same here in this class with the direct null check

lobby-db-dao/src/test/java/org/triplea/lobby/server/db/data/ApiKeyUserDataTest.java
```diff
@@ -0,0 +1,98 @@
...
+        Arrays.asList("0", UserRole.ANONYMOUS), //
+        Arrays.asList(String.valueOf(USER_ID), UserRole.PLAYER), //
+        Arrays.asList(String.valueOf(USER_ID), UserRole.MODERATOR), //
+        Arrays.asList(String.valueOf(USER_ID), UserRole.ADMIN));
+  }
+
+  @ParameterizedTest
+  @MethodSource("mustHaveUserName")
+  void assertInvalidStatesRolesThatMustHaveName(final List<String> data) throws Exception {
+    givenResults(Integer.parseInt(data.get(0)), data.get(1), null);
```
RoiEXLab: You can pass multiple arguments using parameterized tests.
I believe the type you'd use instead of a `List` is called `Arguments`, so `mustHaveUsername` would return a `List<Arguments>`
See https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-sources-MethodSource for examples

lobby-db/sample_data.sql
```diff
@@ -6,11 +6,20 @@ delete
...
 from access_log;
+delete
+from user_role;
+
+insert into user_role(id, name)
+values (1, 'ADMIN'), -- user can add/remove admins and add/remove moderators, has boot/ban privileges
+       (2, 'MODERATOR'), -- user has boot/ban privileges
+       (3, 'PLAYER'), -- standard registered user
+       (4, 'ANONYMOUS'), -- users that are not registered, they do not have an entry in lobby_user table
+       (5, 'HOST'); -- AKA LobbyWatcher, special connection for hosts to send game updates to lobby
```
RoiEXLab: Hmmm given the check I commented on earlier, would it work implementing something like "permission levels", so we'd just have to compare an int for check for permission, or is the permission model too diverse?

lobby/src/main/java/org/triplea/lobby/server/db/UserController.java
```diff
@@ -52,11 +52,27 @@ public boolean doesUserExist(final String username) {
...
       final String name, final String email, final HashedPassword hashedPassword) {
+
+    try (Connection con = connection.get();
+        PreparedStatement ps =
+            con.prepareStatement("insert into user_role(id, name) values(10, 'ANONYMOUS')")) {
+
+      ps.execute();
+      con.commit();
+    } catch (final SQLException e) {
+      // ignore
```
RoiEXLab: I think we should at least log it 🤷‍♂

lobby/src/test/java/org/triplea/lobby/server/db/BadWordControllerIntegrationTest.java
```diff
@@ -5,13 +5,15 @@
...
 import com.github.database.rider.junit5.DBUnitExtension;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 import org.triplea.lobby.server.config.TestLobbyConfigurations;
 import org.triplea.test.common.Integration;
 
+@Disabled
```
RoiEXLab: Is it worth adding a project comment here?

RoiEXLab: Same with the other tests

lobby-db-dao/src/test/java/org/triplea/lobby/server/db/data/ApiKeyUserDataTest.java
DanVanAtta: Trying `Arguments` here, access is by object array instead of `List<String>`, lead to more casting:
```
  private static List<Arguments> mustHaveUserName() {
    return Arrays.asList(
        Arguments.of("0", UserRole.ANONYMOUS), //
        Arguments.of(String.valueOf(USER_ID), UserRole.PLAYER), //
        Arguments.of(String.valueOf(USER_ID), UserRole.MODERATOR), //
        Arguments.of(String.valueOf(USER_ID), UserRole.ADMIN));
  }

  @ParameterizedTest
  @MethodSource("mustHaveUserName")
  void assertInvalidStatesRolesThatMustHaveName(final Arguments data) throws Exception {
    givenResults(Integer.parseInt(String.valueOf(data.get()[0])), String.valueOf(data.get()[1]), null);

    assertPostconditionFailure();
  }
```

It looks like from the reference that Arguments is good when we need a stream operation before using the args, ie:
```
public class MyArgumentsProvider implements ArgumentsProvider {

    @Override
    public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
        return Stream.of("apple", "banana").map(Arguments::of);
    }
}
```

The cast to get an integer from `Object` is a bit nasty, as we cast to a string first then to an integer, and then we have another cast to get the second string. `List<String>` is not horrible here, we could do `List<Object>` but then we'll have the same casting issues. Seems a wash, `List<String>` has the least casting.

lobby/src/test/java/org/triplea/lobby/server/db/BadWordControllerIntegrationTest.java
DanVanAtta: It should not be disabled actually. Our `verify` script broke some time ago, this update really made that bad and I could no longer ever get clean runs as there was some conflict between one test or another. This was disabled as a way to try and get around that, eventually I submitted this expecting to see the tests conflict with one another and was a bit surprised they did not.
Removed the disabled: 2f062fb53

lobby/src/main/java/org/triplea/lobby/server/db/UserController.java
DanVanAtta: oh, this is a bit of a shameful hack. The `UserController` code is all dead code in super short order.
The reason the exception is ignored is because we are expecting a UK error indicating the value is already in DB. The only time we would expect that is while running tests, hence, the super-hack. I don't remember why I went for such an ugly approach, i think it's because the test-code was calling these methods for setup while testing the same logic. 

I added some TODOs and comments about this: 221069234

lobby-db/sample_data.sql
DanVanAtta: That was getting tempting. It's important to note that `host` is a singleton permission, only `host` is authorized for the `host` role. That breaks the integer based check with a 'ranking'. Second, for authorization we should lean to having white-lists as much as possible, be explicit so we can avoid accidental new roles coming in.

It might make more sense to have a many to one relationship for users to user roles. I've been debating that a bit. In that case an admin would have 4 roles associated and we could do direct matching. EG:

```
private boolean isAuthorized(String targetRole, List<String> userRoles) {
  return userRoles.contains(targetRole); 
}
```

The thing that is stopping me from doing that is for now things are still pretty simple, but I'm concerned a bit about the data insertion logic. Going that route, we'd probably want to define a 'role-group', which would map a 'group' to many roles, and we could use that as a join table to translate 'user' to a list of roles through their role group.

It's a decision of whether we want the role mappings to be defined in DB or in code. I'm kinda liking going with a 'role-group', having it all be defined in DB, but at this point the value to switch seems kinda low, probably better to avoid the code-churn and revisit if we have more work with roles.

lobby-db-dao/src/main/java/org/triplea/lobby/server/db/dao/api/key/ApiKeyDaoWrapper.java
DanVanAtta: Updated: d40f45d61

http-server/src/main/java/org/triplea/server/access/RoleAuthorizer.java
DanVanAtta: It does read better with: `adminAuthorizedFor`. The 'is' prefix IMO gets to be a bit wordy.

359135935

http-server/src/main/java/org/triplea/server/access/AuthenticatedUser.java
```diff
@@ -16,20 +16,33 @@
...
+
+    return name;
   }
 
   public int getUserIdOrThrow() {
     Preconditions.checkState(
-        !userRole.equals(UserRole.ANONYMOUS), "Anonymous users always have a null user id");
-    return Optional.ofNullable(userId)
-        .orElseThrow(() -> new AssertionError("Error, expected to have a non-null user id"));
+        userId != null, "This method is called when we expect user id to not be null");
```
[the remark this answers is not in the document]
DanVanAtta: Was a bit curious about this, looking at the javadocs:

> checkArgument(boolean expression) Ensures the truth of an expression involving one or more parameters to the calling method.

> checkNotNull(T reference) Ensures that an object reference passed as a parameter to the calling method is not null.

> checkState(boolean expression)
Ensures the truth of an expression involving the state of the calling instance, but not involving any parameters to the calling method.

It looks like in this case `checkState` is appropriate. I used that also because it really is the data state that is invalid. `null` values are actually fine, but in combination with other values, it indicates a bad data state in DB.

In short, it looks like when checking args, `checkArgument` or `checkNotNull` are the ones to use.

http-server/src/main/java/org/triplea/server/access/AuthenticatedUser.java
```diff
@@ -16,20 +16,33 @@
...
+  @Getter @Nonnull private final ApiKey apiKey;
+  @Nullable private final String name;
 
+  @Nullable
   @Override
   public String getName() {
-    throw new UnsupportedOperationException("Name lookup is not done on authentication.");
+    Preconditions.checkState(
+        (name == null && userRole.equals(UserRole.HOST))
+            || (name != null && !userRole.equals(UserRole.HOST)),
```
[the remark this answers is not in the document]
DanVanAtta: :+1:  
Updated: 4bbe92055

lobby-db-dao/src/test/java/org/triplea/lobby/server/db/data/ApiKeyUserDataTest.java
RoiEXLab: Ah, you misunderstood my point:
When using Arguments the content gets expanded, so you can do this:
```java
  @ParameterizedTest
  @MethodSource("mustHaveUserName")
  void assertInvalidStatesRolesThatMustHaveName(final int id, final String role) throws Exception {
    givenResults(id, role, null);

    assertPostconditionFailure();
  }
```

RoiEXLab: Of course this makes the assumption that you the types match the types you passed to the arguments object

DanVanAtta: Ah, that was good to learn about, I did not get that from the documentation. It is much cleaner with `Arguments`, no casting at all, updated: 2fcae9823

RoiEXLab: AFAIK it works with a simple 2 dimensional array or other container structures as well

PR #5355, opened 2019-10-17
Upgrade Gradle to 5.6.2 and suppress output warnings

DanVanAtta:
(1) Update gradle version 5.6.2, 5.6 brings with it pmd incremental analysis flag
(2) Turn on PMD incremental analysis flag
(3) Suppress deprecated usage warnings, we are ignoring them for now and meanwhile
    compiler output is spammed and errors become difficult to find.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[x] Manually tested

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -122,7 +122,6 @@ subprojects {
     tasks.withType(JavaCompile).configureEach {
         options.compilerArgs += [
             '-Xlint:all,-processing',
-            '-Xlint:all,-deprecation',
             '-Xmaxwarns', '1000'
         ]
         options.encoding = 'UTF-8'
--- a/build.gradle
+++ b/build.gradle
@@ -122,6 +122,7 @@ subprojects {
     tasks.withType(JavaCompile).configureEach {
         options.compilerArgs += [
             '-Xlint:all,-processing',
+            '-Xlint:all,-deprecation',
             '-Xmaxwarns', '1000'
         ]
         options.encoding = 'UTF-8'
@@ -205,6 +206,7 @@ subprojects {
         consoleOutput = true
         ruleSetFiles = files(rootProject.file('config/pmd/pmd.xml'))
         ruleSets = []
+        incrementalAnalysis = true
     }
 
     spotless {
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.4.1-all.zip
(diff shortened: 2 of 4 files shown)


build.gradle
```diff
@@ -205,6 +206,7 @@ subprojects {
         consoleOutput = true
         ruleSetFiles = files(rootProject.file('config/pmd/pmd.xml'))
         ruleSets = []
+        incrementalAnalysis = true
```
RoiEXLab: I believe this might actually slow things down for travis.
@ssoloff once commented on the warning, but I don't recall exact details

DanVanAtta: IIRC his comment was that there was not a good way to turn off the warning, that incremental was not supported on gradle.

> I believe this might actually slow things down for travis.

Why do you suspect that? 

Seems like it is simply caching results: https://pmd.github.io/latest/pmd_userdocs_incremental_analysis.html
I would be a bit surprised if that were noticeably slow. 

Failing turning on incremental, it would be good to find a way to suppress the output warnings. It was a significant problem finding errors in a lot of 'noisy' log messages.

DanVanAtta: This was the comment/message about the warning from @ssoloff  that I was recalling (see the additional comments section): https://github.com/triplea-game/triplea/pull/4637

ssoloff: My comment was that the warning wasn't suppressable at the time PMD was introduced into TripleA (see #4637).  It looks like the `incrementalAnalysis` property was just recently introduced in Gradle 5.6, so I have no idea how it might or might not affect build performance.  The only thing I'd recommend is to verify that, with this option enabled, a known PMD violation still triggers a build failure.

RoiEXLab: Ah ok, misconception on my side, I had remembered the conversation a little bit different.

build.gradle
```diff
@@ -122,6 +122,7 @@ subprojects {
     tasks.withType(JavaCompile).configureEach {
         options.compilerArgs += [
             '-Xlint:all,-processing',
+            '-Xlint:all,-deprecation',
```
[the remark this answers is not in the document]
DanVanAtta: I share the concern. It is impractical when needing to actually find errors. The balance between 'nice-to-have hopefully we'll get to it and meanwhile we ignore this output' vs 'this is getting in the way and takes up real time when finding errors which *must* be fixed', the latter needs to win out.

Since the concern is noise in output, we could tune down the number of warnings, perhaps instead of 1000, do 10.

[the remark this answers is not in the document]
DanVanAtta: I removed the deprecation suppression for now so we can move forward. I'll submit a follow-up PR, I'm thinking reducing the warnings might be the better way to go.

Otherwise, I tested out PMD violations, adding/removing them, they are still detected (and I'll note, *much* faster than before)

PR #5356, opened 2019-10-17
Update Drop-DB script to always work, kill connections

DanVanAtta:
Update the drop_db script to kill active connections. This allows it
to always run, 'drop_db' fails if there are active connections.

This is particularly useful if background daemon threads are lingering
and have active DB connections.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[x] Manually tested

### DIFF
--- a/lobby-db/drop_db
+++ b/lobby-db/drop_db
@@ -6,6 +6,8 @@
 
 DB=lobby_db
 PSQL="psql -h localhost -U lobby_user -w -d postgres"
+
+echo "select pg_terminate_backend(pid) from pg_stat_activity where datname='$DB';" | $PSQL
 echo "drop database $DB" | $PSQL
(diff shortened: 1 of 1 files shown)


lobby-db/drop_db
```diff
@@ -6,6 +6,8 @@
 
 DB=lobby_db
 PSQL="psql -h localhost -U lobby_user -w -d postgres"
+
+echo "select pg_terminate_backend(pid) from pg_stat_activity where datname='$DB';" | $PSQL
```
RoiEXLab: Interesting, sql really makes things weird if no data needs to actually be fetched ^^

PR #5359, opened 2019-10-18
Reduce gradle build compiler warning count

DanVanAtta:
Reduce number of compiler warnings displayed on gradle builds
to reduce noise in build output. Important when we need
to find error messages amongst the output.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[x] Manually tested

## Additional Review Notes
Main goal here is so we can have cleaner output and not necessarily ignore compiler warnings altogether. Happy to discuss appropriate thresholds, 15 is chosen so we get a good sample of warnings but the output is still not overwhelming.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -122,7 +122,7 @@ subprojects {
     tasks.withType(JavaCompile).configureEach {
         options.compilerArgs += [
             '-Xlint:all,-processing',
-            '-Xmaxwarns', '1000'
+            '-Xmaxwarns', '15'
         ]
         options.encoding = 'UTF-8'
         options.errorprone {


RoiEXLab:
The reason I originally introduced this is to avoid warnings being swallowed by the compiler, meaning if there are 15 known warnings and a new one is introduced, you won't notice it at all.
I wonder if there's an application that just analyses new compiler warnings 🤔

DanVanAtta:
> if there are 15 known warnings and a new one is introduced, you won't notice it at all.

As-is, you wouldn't notice either. We need to keep in mind there are different dimensions to warnings as well, notably IDE which can be throttled up and down. It would be easy to introduce lots of warnings that are detected on one IDE but not another, another way to introduce issues without noticing.

IMO the warnings are a problem to chip away at or for someone who wants to contribute but is not yet ready to embark on a larger project. I suspect most new contributors will want to jump in with a project though, so it's pretty hypothetical. At the end of the day we have to be sure to weigh value, whether it's better to fix warnings or to push forward to 2.0, javaFx, etc.. Hence, chipping away at the problem IMO is the right way to go, so throttling down the warning output so it's there but not noisy I think is maybe a good compromise.

asvitkine:
Should we just disable some warning categories?

For example, most warnings seem to be from deprecated Observer class use.
We could disable it via:
 -           '-Xlint:all,-processing',
+           '-Xlint:all,-processing,-deprecation',

We could still file an issue in the issue tracker to handle those. But in the meantime, it would be nice not to spam developers about it.

Alternatively, maybe there's some way to check in a "expected warnings" file and make the build output just the diff? Does gradle offer any way to postprocess the warning output with custom code?

DanVanAtta:
I don't think there is a right answer here other than to chip away at the problem and get it fixed.

My initial thought was also to suppress the deprecation warnings. There are two downsides to doing that:
1. It'll be easy to forget about the warnings
2. To solve the warnings, it'll require a temporary code change that will need to be undone. Not a big deal, but still a small hassle and annoyance.

We really do want output that is just a `sort | uniq -c` of the warning types, we only really care about the types and counts until we actually go to fix the problems, in which case we want the full details. We could massage the output to be more readable and send the full data to a report file. I suspect though it would wind up being misplaced effort and add complexity we probably don't really need.

PR #5365, opened 2019-10-22
Don't use fixed width tabs on mac look and feel.

asvitkine:
It especially looks bad with the Mac native look and feel, but even on other look and feels, the tabs weren't fixed width anyway since some of them were wider.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[X] Manually tested

### Before
Screen Shot 2019-10-21 at 8 32 11 PM

### After
Screen Shot 2019-10-21 at 8 32 36 PM

### DIFF
--- a/swing-lib/src/main/java/org/triplea/swing/SwingComponents.java
+++ b/swing-lib/src/main/java/org/triplea/swing/SwingComponents.java
@@ -140,7 +140,7 @@ public final class SwingComponents {
   }
 
   public static JTabbedPane newJTabbedPane(final int width, final int height) {
-    final JTabbedPane tabbedPane = new JTabbedPaneWithFixedWidthTabs();
+    final JTabbedPane tabbedPane = new JTabbedPane();
     tabbedPane.setPreferredSize(new Dimension(width, height));
     return tabbedPane;
   }
--- a/swing-lib/src/main/java/org/triplea/swing/SwingComponents.java
+++ b/swing-lib/src/main/java/org/triplea/swing/SwingComponents.java
@@ -139,11 +139,13 @@ public final class SwingComponents {
             });
   }
 
-  public static JTabbedPane newJTabbedPane() {
-    return newJTabbedPane(900, 600);
+  public static JTabbedPane newJTabbedPane(final int width, final int height) {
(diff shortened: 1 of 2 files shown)


DanVanAtta:
We really should try to avoid OS specific code as much as possible. Mainly for testing considerations. For example, it would be easy to break a different OS and any OS specific code arguably needs testing on each OS, multiplying the manual test burden.

This is how it looks for linux and I assume windows too:
Screenshot from 2019-10-21 18-42-42

Does it make sense to not use fixed width tabs altogether?

DanVanAtta:
Side-note: We might be able to reduce overflow by combining & reducing the number of tabs. That seems like a follow-up change that could be done in parallel.

asvitkine:
> We really should try to avoid OS specific code as much as possible. Mainly for testing considerations. For example, it would be easy to break a different OS and any OS specific code arguably needs testing on each OS, multiplying the manual test burden.
> 
> This is how it looks for linux and I assume windows too:
> Screenshot from 2019-10-21 18-42-42
> 
> Does it make sense to not use fixed width tabs altogether?

Sure, done!

swing-lib/src/main/java/org/triplea/swing/SwingComponents.java
```diff
@@ -139,11 +139,13 @@ public void actionPerformed(final ActionEvent e) {
             });
   }
 
-  public static JTabbedPane newJTabbedPane() {
-    return newJTabbedPane(900, 600);
+  public static JTabbedPane newJTabbedPane(final int width, final int height) {
+    final JTabbedPane tabbedPane = new JTabbedPaneWithFixedWidthTabs();
```
DanVanAtta: Was this line supposed to be `newJTabbedPane()`?
I'm a bit confused as I thought we were removing fixed width tabs, and they seem to still be around. Wondering what I'm missing.

asvitkine: Whoops, yes. Fixed.

DanVanAtta:
:+1:

PR #5368, opened 2019-10-26
Validate Account Creation on Key-Type

DanVanAtta:
Moves validation of 'create account' from when a user clicks submit
to while they enter data, does not enable 'submit' button until fields are valid.

Update details:
(1) Extract key type validation functionality to a utility class 'KeyTypeValidator'.
(2) Add 'KeyTypeValidator' to CreateAccountPanel
(3) Add a validation label to Create Account Panel to display validation message.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[x] Manually tested

### After
Screenshot from 2019-10-25 22-40-41

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/login/CreateAccountPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/login/CreateAccountPanel.java
@@ -275,7 +275,6 @@ public final class CreateAccountPanel extends JPanel {
     }
   }
 
-  // TODO: Project#12 show validation message as a label (?)
   private Optional<String> validationMessage() {
     if (!Arrays.equals(passwordField.getPassword(), passwordConfirmField.getPassword())) {
       return Optional.of("Passwords must match");
@@ -313,7 +312,7 @@ public final class CreateAccountPanel extends JPanel {
   }
 
   public String getPassword() {
-    // TODO: Md5-Deprecation SHA512 hash password on client side here
+    // TODO: Project#12 SHA512 hash password on client side here
     return new String(passwordField.getPassword());
   }
 
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/login/CreateAccountPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/login/CreateAccountPanel.java
@@ -227,12 +227,7 @@ public final class CreateAccountPanel extends JPanel {
     buttons.add(cancelButton);
 
     cancelButton.addActionListener(e -> close());
-    okButton.addActionListener(
-        e -> {
-          if (okButton.isEnabled()) {
-            okPressed();
-          }
-        });
+    okButton.addActionListener(e -> okPressed());
 
     SwingComponents.addEnterKeyListener(
         this,
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/login/CreateAccountPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/login/CreateAccountPanel.java
@@ -5,12 +5,14 @@ import games.strategy.engine.lobby.PlayerNameValidation;
 import games.strategy.triplea.settings.ClientSetting;
 import games.strategy.ui.Util;
 import java.awt.BorderLayout;
+import java.awt.Color;
 import java.awt.FlowLayout;
 import java.awt.GridBagConstraints;
 import java.awt.GridBagLayout;
 import java.awt.Insets;
 import java.awt.Window;
 import java.util.Arrays;
+import java.util.Optional;
 import javax.annotation.Nullable;
 import javax.swing.BorderFactory;
 import javax.swing.ImageIcon;
@@ -23,6 +25,7 @@ import javax.swing.JPanel;
 import javax.swing.JPasswordField;
 import javax.swing.JTextField;
 import org.triplea.swing.JCheckBoxBuilder;
+import org.triplea.swing.KeyTypeValidator;
 import org.triplea.swing.SwingComponents;
 
 /** The panel used to create a new lobby account or update an existing lobby account. */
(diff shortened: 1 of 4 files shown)


game-core/src/main/java/games/strategy/engine/lobby/client/login/CreateAccountPanel.java
```diff
@@ -275,14 +313,15 @@ public ReturnValue show(final Window parent) {
   }
 
   public String getPassword() {
+    // TODO: Md5-Deprecation SHA512 hash password on client side here
```
RoiEXLab: Maybe I'm missing something, but what exactly does this comment imply?

game-core/src/main/java/games/strategy/engine/lobby/client/login/CreateAccountPanel.java
```diff
@@ -195,19 +198,75 @@
...
+          if (okButton.isEnabled()) {
+            okPressed();
+          }
+        });
+
+    SwingComponents.addEnterKeyListener(
+        this,
+        () -> {
+          if (okButton.isEnabled()) {
+            okPressed();
```
RoiEXLab: Would it make sense to move this if check inside the `okPressed` method?
I know that you don't like methods silently being no-ops if some certain condition is true, but I think if the method is properly renamed this would be worth it to reduce code-duplication?

game-core/src/main/java/games/strategy/engine/lobby/moderator/toolbox/tabs/moderators/ModeratorsTab.java
```diff
@@ -91,24 +88,13 @@ private Component buildAddModeratorPanel(final JTable table) {
     // This is a very fancy listener that will validate the entered requested new moderator
     // name exists (as the user types) before we enable the submit button.
     // To avoid sending too many requests to server we have a back-off period to batch requests.
```
RoiEXLab: This comment no longer belongs here, the code was moved to `KeyTypeValidator`

swing-lib/src/main/java/org/triplea/swing/KeyTypeValidator.java
```diff
@@ -0,0 +1,44 @@
...
+      final Predicate<String> dataValidation,
+      final Consumer<Boolean> action) {
+    DocumentListenerBuilder.attachDocumentListener(
+        textComponent,
+        () ->
+            new Thread(
+                    () -> {
+                      if (!validationIsInFlight) {
+                        validationIsInFlight = true;
+                        Interruptibles.sleep(200);
```
RoiEXLab: Does this really provide a benefit?
I mean for server-side validation having such a timeout definitely makes sense because the requets take their time, but for local validation this is completely overkill IMO

swing-lib/src/main/java/org/triplea/swing/KeyTypeValidator.java
```diff
@@ -0,0 +1,44 @@
...
+      final JTextComponent textComponent,
+      final Predicate<String> dataValidation,
+      final Consumer<Boolean> action) {
+    DocumentListenerBuilder.attachDocumentListener(
+        textComponent,
+        () ->
+            new Thread(
+                    () -> {
+                      if (!validationIsInFlight) {
+                        validationIsInFlight = true;
```
RoiEXLab: Also this class is not really thread safe, using AtomicBoolean would fix that.

game-core/src/main/java/games/strategy/engine/lobby/client/login/CreateAccountPanel.java
DanVanAtta: Fixed in: https://github.com/triplea-game/triplea/pull/5368/commits/956d225394a68cc1ddd944422b3a4ce1213e275a

swing-lib/src/main/java/org/triplea/swing/KeyTypeValidator.java
```diff
@@ -0,0 +1,51 @@
...
+                        // returns us stale data, that new request will have a chance to
+                        // operate on the most recent.
+                        validationIsInFlight.set(false);
+
+                        final String textData = textComponent.getText().trim();
+                        final boolean valid = dataValidation.test(textData);
+
+                        SwingUtilities.invokeLater(() -> action.accept(valid));
+                      }
+                    })
```
RoiEXLab: I'm still not quite convinced having a 200ms timeout is necessary, but I guess if you spent time polishing this class you'd disagree.
I'm curious to know how exactly this affects user experience

game-core/src/main/java/games/strategy/engine/lobby/client/login/CreateAccountPanel.java
RoiEXLab: Definitely an improvement, even though I still think it's weird checking if the button is enabled before "triggering a button pressed event".
I really feel like there should be a built-in way to link certain textfields to a button so this check would be performed automatically

DanVanAtta: Maybe the original problem is that the 'okButtonPressed' method has actually nothing to do with the button. Arguably it's badly named as its named after the implementation that triggers the action and not after the action itself. Hence, the 'enter' key press is mapped to the same action, bypasses the button altogether. The button enabled is our state tracking if the fields are valid or not...

> I really feel like there should be a built-in way to link certain textfields to a button so this check would be performed automatically

Not that I'm aware of. KeyTypedListener is about as close as it gets. In the swing rewrite of the launch screen there were a few components like this that were created FWIW..

swing-lib/src/main/java/org/triplea/swing/KeyTypeValidator.java
DanVanAtta: In essence we're queuing and then de-duping validation events. 200ms is not actually a timeout, it's a delay.

I did not spend too much time on this code, it was largely extracted, generalized and then re-used. The 200ms was fine-tuned on the first iteration, this iteration it was tested and responded well enough.

DanVanAtta:
> I'll try to have a look on your bigger PR tomorrow.

Thanks! NB, quite a bit of code has been queuing, I've a lot more for follow-up. Hopefully it won't take too long to get it all polished and reviewed, the 2.0 release is highly anticipated.

prastle:
@RoiEXLab @DanVanAtta   Stated in a different format... All Map makers have been trying to use the new code... TripleA is basically at a standstill until a new testing lobby and /or stable is released.

DanVanAtta:
@prastle indeed, also speaks to the need for more compatibility; lots of the game-to-game and map-to-game compatibility constraints are possible to fix.

PR #5371, opened 2019-10-28
Http-server login components

DanVanAtta:
Adds backend components for http-server to do login. Four types of
login are effectively supported:
  - anonymous: no password login
  - registered: standard login with bcrypt hashed password
  - legacy: standard login but with a md5 hashed password. To aid
    migration, the front-end wills end the plain text password of
    the user over https. This allows us to re-hash legacy passwords
    and store them as brypted.
  - temp password: one-time login using a temporary password, after
    login with temp_password, the temp passwords of the user are
    invalidated.

Notes:
- http-server login not currently used by front-end, that will land
  in upcoming updates
- some tests and functionality are excluded or disabled, they are
  waiting for future updates which will update DB to have system-id
  instead of Mac, and/or waiting for http-server chat components to land.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[x] Other: Feature migration  

## Testing

[x] No manual testing done
[] Manually tested

## Additional Review Notes
- As always, naming suggestions are welcome
- There is still quite a lot of changes on-top of these, to avoid merge conflicts I may favor doing follow-up changes, please indicate what seems to be a deal breaker vs something that can be followed-up with.
- as noted in overview, a number of tests and functionalities will land later, they are missing so that other changes are not looped in and keeps this iteration more bite-sized

### DIFF
--- a/http-server/src/main/java/org/triplea/server/user/account/login/authorizer/legacy/Md5Crypt.java
+++ b/http-server/src/main/java/org/triplea/server/user/account/login/authorizer/legacy/Md5Crypt.java
@@ -1,0 +1,98 @@
+package org.triplea.server.user.account.login.authorizer.legacy;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.apache.commons.codec.digest.Md5Crypt.md5Crypt;
+
+import java.nio.charset.StandardCharsets;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * A collection of methods for using the FreeBSD MD5-crypt hash algorithm.
+ *
+ * @see <a
+ *     href="https://www.usenix.org/legacyurl/md5-crypt">https://www.usenix.org/legacyurl/md5-crypt</a>
+ * @see <a href="https://www.systutorials.com/docs/linux/man/n-md5crypt/">
+ *     https://www.systutorials.com/docs/linux/man/n-md5crypt/ </a>
+ */
+final class Md5Crypt {
+  private static final String MAGIC = "$1$";
+  private static final Pattern HASHED_VALUE_PATTERN =
+      Pattern.compile(
+          "^" + MAGIC.replace("$", "\\$") + "([\\.\\/a-zA-Z0-9]{1,8})\\$([\\.\\/a-zA-Z0-9]{22})$");
+  private static final byte[] EMPTY_KEY_BYTES = new byte[0];
+
+  private Md5Crypt() {}
+
+  /**
+   * Hashes the specified password using the specified salt.
+   *
+   * @param password The password to be hashed.
+   * @param salt The salt. May begin with {@code $1$} and end with {@code $} followed by any number
+   *     of characters. No more than eight characters will be used. If empty, a new random salt will
+   *     be used.
+   * @return The hashed password.
+   * @deprecated MD5-crypt is not secure for hashing sensitive information, such as passwords. Use a
+   *     different hashing algorithm, such as bcrypt.
+   */
+  @Deprecated
+  public static String hashPassword(final String password, final String salt) {
+    return hash(password, salt);
+  }
+
+  /**
+   * Hashes the specified value using the specified salt.
+   *
+   * <p><b>WARNING:</b> This method should not be used to hash sensitive information.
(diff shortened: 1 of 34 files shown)


http-server/src/main/java/org/triplea/server/user/account/login/authorizer/legacy/Md5Crypt.java
```diff
@@ -0,0 +1,98 @@
...
+ * @see <a
+ *     href="https://www.usenix.org/legacyurl/md5-crypt">https://www.usenix.org/legacyurl/md5-crypt</a>
+ * @see <a href="https://www.systutorials.com/docs/linux/man/n-md5crypt/">
+ *     https://www.systutorials.com/docs/linux/man/n-md5crypt/ </a>
+ */
+final class Md5Crypt {
+  private static final String MAGIC = "$1$";
+  private static final Pattern HASHED_VALUE_PATTERN =
+      Pattern.compile(
+          "^" + MAGIC.replace("$", "\\$") + "([\\.\\/a-zA-Z0-9]{1,8})\\$([\\.\\/a-zA-Z0-9]{22})$");
```
RoiEXLab: Might be worth considering using `Pattern.quote(MAGIC)`

http-server/src/main/java/org/triplea/server/user/account/login/authorizer/legacy/Md5Crypt.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  private static String normalizeSalt(final String salt) {
+    if (salt.isEmpty()) {
+      return newSalt();
+    }
+
+    return String.format(
+        "%1.8s",
+        salt.replaceFirst("^" + MAGIC.replace("$", "\\$"), "")
```
RoiEXLab: Same here with regex escaping

http-server/src/main/java/org/triplea/server/user/account/login/authorizer/temp/password/TempPasswordLogin.java
```diff
@@ -0,0 +1,35 @@
...
+    return tempPasswordDao
+        .fetchTempPassword(loginRequest.getName())
+        .map(
+            tempPassword -> {
+              if (passwordChecker.test(loginRequest.getPassword(), tempPassword)) {
+                tempPasswordDao.invalidateTempPasswords(loginRequest.getName());
+                return true;
+              } else {
+                return false;
+              }
```
RoiEXLab: The else case is unecessary at this point

DanVanAtta: True, we should perhaps revisit this as a convention.

1) the block structure is easier to grok when there is an explicit else, the return statements line up.
2) Arguably it's easier to understand, when you see if/else block you know that the method terminates in the conditionals and not possibly in some and not others.
3) Perhaps most compelling to me, I had a bug where in one case I missed the 'return' in the 'if' block, the code fell through to the final return and returned the wrong value, eg:

```
if( condition ) {
  // this return statement was missing
}
return false;
```

On the other hand, written as an `if/else` block, the above would have given a compilation error.

So while the `} else {` is unnecessary, removing it is maybe not the best, less I'm not sure is more in this case.

http-server/src/main/java/org/triplea/server/user/account/login/authorizer/legacy/Md5Crypt.java
DanVanAtta: I should have called out better that Md5Crypte is legacy/copied code. Ideally we would delete it soon. Given it's not tested, I'm reluctant to modify it.

DanVanAtta: (Though, there will be an integration test landing, so that might be some confidence to modify it, but again because it's copied, and also temporary/legacy, there is some value in leaving it as-is too).

RoiEXLab: Yeah I figured, I mean the only reason that string is escaped is because `$` has a special meaning in regex

PR #5376, opened 2019-10-29
Allow 'Toast.java' parent to either be Window or JFrame

DanVanAtta:


DanVanAtta:
> Have you considered using JOptionPane.getFrameForComponent and passing a single parent Component field?

I don't think so, the use-case is the new location where this is needed only has a `Window` reference. Perhaps upstream we could do that, though I think we might still wind up being stuck with a `Window` instead of frame.

---------


I'll revisit this to fix up the compile error and @ mention you @RoiEXLab  when this is ready.

DanVanAtta:
I'm no longer seeing where the window component is needed. Closing this in the meantime.

PR #5379, opened 2019-10-29
Rename method 'isAdmin' to 'isModerator'

DanVanAtta:
De-conflicts the name from the 'admin' role compared to 'moderator'


game-core/src/main/java/games/strategy/engine/lobby/client/LobbyClient.java
```diff
@@ -28,7 +28,7 @@ public LobbyClient(
     this.anonymousLogin = anonymousLogin;
     final var moderatorController =
         (IModeratorController) messengers.getRemote(IModeratorController.REMOTE_NAME);
-    admin = moderatorController.isAdmin();
+    moderator = moderatorController.isAdmin();
```
RoiEXLab: Hmm, why is the method still called isAdmin?
Shouldn't it be renamed as well?

DanVanAtta: Perhaps, though `ModeratorController` will be cleaned-up/deleted, not too much point to update it.

RoiEXLab: Hmm leaving this decision up to you.
Feel free to merge if you don't think it's worth Renaming.

DanVanAtta: It's extra effort and will create a small merge conflict that would need to be resolved nonetheless. The original branch has this class just completely deleted and it was not even a concern. It feels very much like overhead at this point.

PR #5383, opened 2019-10-30
Add GenericWebsocketClient

DanVanAtta:
Adds a foundational class to make working with websockets a bit simpler.
The generic websocket client allows for basic websocket connectivity.
The generic client is specialized for dealing with JSON messages
and handles conversion of incoming messages from JSON string to
a java object type.


http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * Non-blocking send of a message to the server. Implementation note: data is sent as a JSON
+   * string, this method handles conversion of the parameter object to JSON.
+   *
+   * @param message The data object to send to the server.
+   */
+  public void send(final OutgoingT message) {
+    final String json = gson.toJson(message);
+    CompletableFutureUtils.logExceptionWhenComplete(
+        CompletableFuture.runAsync(() -> client.sendMessage(json), threadPool),
```
RoiEXLab: Ok I dug into the code of the library and I'm pretty sure the send operation is asynchronous already. (Like I expected it to work from my experience with javascript)
So using a worker pool is probably not speeding things up by a noticeable amount.
Of course it might make sense to parallelize json serialization if the objects are large, but sending doesn't seem to be blocking at all.

DanVanAtta: This `sendMessage` is calling:

```
    waitForConnection();
    if (client.isOpen()) {
      client.send(message);
    } else {
      log.severe("Failed to establish connection to server.");
    }
  }
```

Note the `waitForConnection`

DanVanAtta:
> I'm curious to know your thought process that lead to this code. The abstraction is good, but I think you're underestimating the abstraction the library already provides for you.

Please also let me know if the javadoc of the class was not quite sufficient to capture this.

RoiEXLab:
I'll try to create a explanation in a couple of hours that tries to explain my reasoning in detail.
A lot of my assumptions are based off the "vanilla javascript api" where I'm not quite sure how much is protocol specification, and how much is just api design, so I'll try to double check with the source code of the library

RoiEXLab:
Ok here we go, I had a look at the code and I think I get now the problems you were trying to avoid.
So I want to reconsider some of my auggestions.

So If I understand you correctly you want to have a completely asynchronous way of sending messages with no blocking at any point whatsoever?

So most of the api is asynchronous already, the only problem is that messages need to be queued up until the connection is open, right?

Here's my alternative suggestion to address those issues:
Use a singleThreadExecutor that acts as a message queue.
In the constructor instead of calling `connect()`, submit
```java
() -> client.connectBlocking()
```
In `sendMessage` you add a simple if check:
```
if (client.isOpen()/* && executor queue is empty or shutdown already in
order to ensure the message order is correct */) {
  executor.shutDown();
  client.send(message);
} else {
  executor.submit(() -> client.send(message))
}
```
This way the "waitForConnection" loop is avoided, and once a connection is established the spawned thread will shut down eventually reducing the performance footprint, while keeping everything asynchronous.

Obviously this is just some example code to outline my idea, it definitely needs polishing.

Let me know what you think, or if I missed an important detail.
This approach should be completely asynchronous without any concurrency issues

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
DanVanAtta: @RoiEXLab I extracted the socket connection establishing code to a new class 'WebSocketConnector', removed the wait loop and implemented the single threadpool blocking mechanism in the new connector class. I still have the 'threadpool' here though. The threadpool here will not guarantee ordering if there is any buffering, but I think that is okay. Errorprone really insists that the future object of 'sendMessage' be checked, which is a weak but actual reason for the threadpool to exist so we can have some code to do that logging. The conversion to JSON should be a pretty fast operation, all data is in memory, so even with a large payload we'd expect nano or micro-second timescales.

Let me know your thoughts on the updates.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnector.java
```diff
@@ -0,0 +1,77 @@
...
+
+  void initiateConnection() {
+    executorService = Executors.newSingleThreadExecutor();
+    connectionThread = executorService.submit((Callable<Boolean>) webSocketClient::connectBlocking);
+    executorService.shutdownNow();
+  }
+
+  void waitUntilConnectionIsOpen() {
+    try {
+      executorService.awaitTermination(connectTimeoutMillis, TimeUnit.MILLISECONDS);
```
RoiEXLab: Ok, actually didn't think of it this way.
Now you're really using the blocking method + executor as a fancy way to create some sort of latch to block other threads.
Definitely something I could live with, but I mean a `CountDownLatch` +a custom on open implementation of the connection is opened asynchronously would do the trick as well with less complexity.

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
```diff
@@ -0,0 +1,95 @@
...
+ * Furthermore, this class provides a listener API to be notified of incoming message events, a send
+ * API, and it automatically converts incoming and outgoing messages to JSON string.
+ *
+ * @param <IncomingT> Message type we expect to receive from the server.
+ * @param <OutgoingT> Message type we send to the server.
+ */
+@Log
+public class GenericWebSocketClient<IncomingT, OutgoingT> implements WebSocketConnectionListener {
+  private static final Gson gson = new Gson();
+  private final ExecutorService threadPool = Executors.newFixedThreadPool(4);
```
RoiEXLab: Ok, I'm still not quite getting this, please correct me if I'm missing something.
Why 4 Threads?
The send operation is asynchronous to my knowledge, so using 4 seperate threads shouldn't really speed things up by a noticeable amount, right?

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnector.java
```diff
@@ -0,0 +1,77 @@
...
+  @VisibleForTesting
+  WebSocketConnector(final WebSocketClient webSocketClient, final int connectTimeoutMillis) {
+    this.webSocketClient = webSocketClient;
+    this.connectTimeoutMillis = connectTimeoutMillis;
+  }
+
+  void initiateConnection() {
+    executorService = Executors.newSingleThreadExecutor();
+    connectionThread = executorService.submit((Callable<Boolean>) webSocketClient::connectBlocking);
+    executorService.shutdownNow();
```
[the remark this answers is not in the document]
DanVanAtta: Indeed, that was supposed to be shutdown, fixed: 5a5430959

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnector.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  void waitUntilConnectionIsOpen() {
+    try {
+      executorService.awaitTermination(connectTimeoutMillis, TimeUnit.MILLISECONDS);
+      if (!connectionThread.get() || !webSocketClient.isOpen()) {
+        throw new CouldNotConnect();
+      }
+    } catch (final InterruptedException | ExecutionException e) {
+      throw new CouldNotConnect(e);
```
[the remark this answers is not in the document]
DanVanAtta: Indeed. Also throwing is probably not the right behavior. We likely want to no-op, an exception would show a prompt to a user rather than letting the application abort or proceed to terminate. Updated in: b845467cd, sets interrupt and does a no-op if current thread is interrupted.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnector.java
DanVanAtta: How do you incorporate the timeout in that case? 

Of concern, let's say we have a timeout and we queue messages to send in the same threadpool. If the time to flush all messages takes longer than the timeout, then we'll have an incorrect error not because the connection could not be opened, but because instead we did not send messages in time.

To some extent the original code, though a bit manual, is starting to look not too bad given the cleverness and interplay to avoid having a manual, blocking, timeout. I'm starting to ponder if we hit diminishing returns.

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
DanVanAtta: It was not really to speed things up, the 4 count is perhaps from earlier decisions and maybe longer applied.

Going into the implementation, we land here to this bottleneck:
```
	private void write( List<ByteBuffer> bufs ) {
		synchronized(synchronizeWriteObject) {
			for( ByteBuffer b : bufs ) {
				write( b );
			}
		}
	}
```
The 4 threads then will certainly not be any faster than 1. Should we then have a threadpool at all? I think yes to avoid blocking on the above synchronize block. It probably makes sense to include the gson operation in the thread as well if we are going to go async sending to our async send queue.

DanVanAtta: Updated to be a threadpool of one and put the gson operation on the new thread: 5e0ca037618220ae3da88ea5f8dc273a52df2055

Added a TODO to revisit the perf gain: c245621542473b56c704d35ca3a6b862a4ae5c17

There are a number of integration tests I'm holding back where it would be easier to verify that, I do not want to integrate those branches with this one and then again with master, and then potentially again if there are yet more changes; there are enough merge conflicts to resolve without re-resolving the same ones multiple times and making them worse with transitory changes if this branches changes further.

DanVanAtta:
@RoiEXLab updated, I think the big disconnect is incorporating a timeout. I prefer avoiding countdown latches whenever possible.

DanVanAtta:
> I'm curious to know your thought process that lead to this code. The abstraction is good, but I think you're underestimating the abstraction the library already provides for you.

To comment, this code started about 3 months ago and was put to the side. 5-7 weeks ago it was resurrected and cleaned up and started out as a spike to connect to a websocket server and quickly turned into an integration test where the implementation was driven by the sequence:
```
connect
send
wait for response
close
```
From that, without blocking on the connect, something was needed to ensure that the connect was completed before sending. The extra complexity is to try and hide that delay between connect and send from any clients of this abstraction so that those clients could 'think' that they can send right away after connecting.

DanVanAtta:
> I believe the changes go into the right direction, not quite my original idea, but definitely better than the while loop

I'm not sure I'm entirely convinced the while loop was that inferior. It had some advantages:
- isolated a chunk of logic to one, well-scoped place
- was relatively easy to understand/read
- no involvement of additional threading logic (ie: thread interrupts, single-threaded thread pools, thread executors and how those operate with timeouts)

Regardless, I do want to see this move forward, this component is blocking a number of the next pieces that could land.

DanVanAtta:
Thanks! Sounds good @RoiEXLab  :+1:

PR #5386, opened 2019-10-30
Add ClientMessageEnvelope & Factory

DanVanAtta:
Add objects that will be sent from client to server over websocket.
The message envelope is designed to be encoded into a JSON string
and contains three fields. The envelope contains a message type and
payload. Based on message type the server will know what kind of object
exists in the payload. Last, the message envelope contains the clients
API-key to be used by the server for client identification and filtering.


http-clients/src/main/java/org/triplea/http/client/lobby/chat/events/client/ClientMessageEnvelope.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public ApiKey getApiKey() {
+    return ApiKey.of(apiKey);
+  }
+
+  public ClientMessageType getMessageType() {
+    try {
+      return ClientMessageType.valueOf(messageType);
+    } catch (final IllegalArgumentException e) {
+      return ClientMessageType.UNKNOWN;
```
RoiEXLab: 👍 For this error handling

PR #5387, opened 2019-10-30
Update CompletableFutureUtils to accept a Consumer

DanVanAtta:
Make CompletableFutureUtils more generic by accepting a Consumer instead
of doing a logging action itself. This does two things:
- removes dependency on using java.util logger
- allows for client code to specify the desired logging level, potentially
  warn or info as desired.

### DIFF
--- a/java-extras/src/main/java/org/triplea/java/concurrency/CompletableFutureUtils.java
+++ b/java-extras/src/main/java/org/triplea/java/concurrency/CompletableFutureUtils.java
@@ -3,10 +3,8 @@ package org.triplea.java.concurrency;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
 import lombok.experimental.UtilityClass;
-import lombok.extern.java.Log;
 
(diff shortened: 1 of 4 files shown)


java-extras/src/main/java/org/triplea/java/concurrency/CompletableFutureUtils.java
```diff
@@ -1,38 +1,26 @@
...
-import com.google.common.annotations.VisibleForTesting;
 import java.util.concurrent.CompletableFuture;
-import java.util.logging.Level;
-import java.util.logging.Logger;
+import java.util.function.Consumer;
+import lombok.experimental.UtilityClass;
 import lombok.extern.java.Log;
 
 /** A collection of useful methods for working with instances of {@link CompletableFuture}. */
 @Log
```
RoiEXLab: Is this annotation still required?

DanVanAtta: Good catch, removed in 0c311bd

PR #5391, opened 2019-10-31
Http-Server Chat Components

DanVanAtta:
Adds backend http-server components for chat. Note, the endpoint powering
this functionality will land in a future change, for now these are just
the implementation classes that will eventually be wired to the server.

Design notes:
- "chatters" is a main data structure to keep track of the current chat participants.
- "ChatSocketController" is a thin controller class that opens a websocket, listens
  for messages and connections. This class is not yet registered with the server
  configuration, and is not yet active. Websockets are registered via static class
  reference, injecting objects into it is difficult and is done by injecting such
  objects into the user session. Thus, this class tries to be thin, it provides
  the websocket endpoints, grabs a 'MessagingService' from user properties, then
  delegates all further actions to the 'MessagingService'
- "ChatEventProcessor" is an important command and control class to identify
  the message that has been received and initiate the appropriate processing.

### DIFF
--- a/http-server/src/test/java/org/triplea/server/lobby/chat/InetExtractorTest.java
+++ b/http-server/src/test/java/org/triplea/server/lobby/chat/InetExtractorTest.java
@@ -1,0 +1,21 @@
+package org.triplea.server.lobby.chat;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+import java.net.InetAddress;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+class InetExtractorTest {
+
+  private static final String IP_ADDRESS = "127.0.0.1";
+
+  @Test
+  void verify() throws Exception {
+    final Map<String, Object> propertiesMap =
+        Map.of(InetExtractor.IP_ADDRESS_KEY, "/" + IP_ADDRESS + ":42840");
+
+    assertThat(InetExtractor.extract(propertiesMap), is(InetAddress.getByName(IP_ADDRESS)));
+  }
+}
--- a/http-server/src/test/java/org/triplea/server/lobby/chat/MessageBroadcasterTest.java
+++ b/http-server/src/test/java/org/triplea/server/lobby/chat/MessageBroadcasterTest.java
@@ -4,7 +4,7 @@ import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
-import com.google.common.collect.Sets;
+import java.util.Set;
 import java.util.function.BiConsumer;
 import javax.websocket.Session;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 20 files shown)


http-server/src/test/java/org/triplea/server/lobby/chat/InetExtractorTest.java
```diff
@@ -0,0 +1,21 @@
...
+import org.junit.jupiter.api.Test;
+
+class InetExtractorTest {
+
+  private static final String IP_ADDRESS = "127.0.0.1";
+
+  @Test
+  void verify() throws Exception {
+    final Map<String, Object> propertiesMap =
+        Map.of(InetExtractor.IP_ADDRESS_KEY, "/" + IP_ADDRESS + ":42840");
```
RoiEXLab: I'm curious to know where this address format originates from

http-server/src/test/java/org/triplea/server/lobby/chat/MessageBroadcasterTest.java
```diff
@@ -0,0 +1,42 @@
...
+  @Mock private Session session1;
+  @Mock private Session session2;
+  @Mock private ServerMessageEnvelope serverEventEnvelope;
+
+  @Mock private BiConsumer<Session, ServerMessageEnvelope> singleMessageSender;
+  @InjectMocks private MessageBroadcaster messageBroadcaster;
+
+  @Test
+  void accept() {
+    when(session0.getOpenSessions()).thenReturn(Sets.newHashSet(session0, session1, session2));
```
RoiEXLab: Actually java has now those kind of factory methods as well, see https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Set.html#of(E,E,E)

http-server/src/test/java/org/triplea/server/lobby/chat/MessagingServiceTest.java
```diff
@@ -0,0 +1,186 @@
...
+      when(session.getUserProperties()).thenReturn(userPropertiesMap);
+
+      messagingService.handleMessage(session, JSON_MESSAGE);
+
+      verify(messageBroadcaster, never()).accept(any(), any());
+    }
+
+    /**
+     * In this case we have a valid API key, though the ClientEnvelope we de-serialized from JSON
+     * string is invalid. This is represented by teh eventProcessing returning zero server messages.
```
RoiEXLab: "teh" typo

DanVanAtta: Thanks! Fixed: 86d60b5d1

http-server/src/test/java/org/triplea/server/lobby/chat/MessageBroadcasterTest.java
DanVanAtta: Thanks, I was having trouble finding that method, updated: 9d7259218

http-server/src/test/java/org/triplea/server/lobby/chat/InetExtractorTest.java
DanVanAtta: In INetExtractor:
```
  public static InetAddress extract(final Map<String, Object> userSession) {
    // expected format '/127.0.0.1:42840'
    final String ipString = String.valueOf(userSession.get(IP_ADDRESS_KEY)).substring(1);
```
The `userSession` comes from the `Session` object that is part of the websocket methods, ie: `session.getUserProperties`

In some initial research on how to get this value, the IP address being placed on the userSession is specific to the container, for us, 'jetty'. The format is post-priori, essentially I verified that on a running server we had this property, printed it, and hence the expected format.

PR #5392, opened 2019-11-01
Rename hashed-mac to system-id

DanVanAtta:
Data length is changing to fit an expected UUID (36 characters long)

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[x] No manual testing done
[] Manually tested

## Additional Review Notes
I've been a bit on the fence whether to enforce the system-id to be an exact 36 expected characters, or if that is over-constrained. Curious if there are thoughts on that. If there is consensus to go with that, I'd be happy to mark some TODOs and update with a follow-up (first I'd like to get ModeratorController dropped, after which time we won't have to update code that is going to be deleted anyways).


prastle:
Please explain Moderator drop a bit... @DanVanAtta

RoiEXLab:
@prastle
@DanVanAtta is currently migrating all existing features to the new server system, so by dropping he's referring to the old code that becomes obsolete once everything is migrated

PR #5401, opened 2019-11-02
Wire LobbyChatTransmitter to LobbyChatClient

DanVanAtta:
Adds client side infrastructure for lobby websocket chat.
LobbyChatTransmitter is a new class that can be swapped out for a
'MessengersChatTransmitters' as they both implement the same chat API.

The ChatTransmitter is wired to a set of listener APIs and send
APIs that are backed by an http component that eventually sends
and receives requests from the GenericWebsocket connection.

As part of this wiring is an event handler that handles server
messages and routes them back to the appropriate listener.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[x] No manual testing done
[] Manually tested

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/chat/LobbyChatTransmitter.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/LobbyChatTransmitter.java
@@ -1,0 +1,71 @@
+package games.strategy.engine.chat;
+
+import games.strategy.engine.lobby.client.LobbyClient;
+import java.util.Collection;
+import org.triplea.domain.data.PlayerName;
+import org.triplea.http.client.lobby.chat.ChatParticipant;
+import org.triplea.http.client.lobby.chat.LobbyChatClient;
+
+/**
+ * Chat transmitter designed to work with lobby, sends and receives messages over Websocket. This
+ * class provides a send API and does wiring to connect callbacks from a {@code LobbyChatClient} to
+ * a {@code ChatClient}.
+ */
+// TODO: Project#12 test-me
+public class LobbyChatTransmitter implements ChatTransmitter {
+  private final LobbyChatClient lobbyChatClient;
+  private final PlayerName localPlayerName;
+
+  public LobbyChatTransmitter(final LobbyClient lobbyClient) {
+    this.localPlayerName = lobbyClient.getPlayerName();
+    this.lobbyChatClient = lobbyClient.getHttpLobbyClient().getLobbyChatClient();
+  }
+
+  @Override
+  public void setChatClient(final ChatClient chatClient) {
+    lobbyChatClient.addPlayerStatusListener(chatClient::statusUpdated);
+    lobbyChatClient.addPlayerLeftListener(chatClient::participantRemoved);
+    lobbyChatClient.addPlayerJoinedListener(chatClient::participantAdded);
+    lobbyChatClient.addChatMessageListener(chatClient::messageReceived);
+    lobbyChatClient.addConnectedListener(chatClient::connected);
+
+    lobbyChatClient.addPlayerSlappedListener(
+        slapEvent -> {
+          if (slapEvent.getSlapped().equals(localPlayerName)) {
+            chatClient.slappedBy(slapEvent.getSlapper());
+          } else {
+            chatClient.playerSlapped(slapEvent.getSlapper() + " slapped " + slapEvent.getSlapped());
+          }
+        });
+  }
+
+  @Override
+  public Collection<ChatParticipant> connect() {
+    return lobbyChatClient.connect();
+  }
+
+  @Override
+  public void disconnect() {
+    lobbyChatClient.close();
+  }
+
+  @Override
+  public void sendMessage(final String message) {
+    lobbyChatClient.sendChatMessage(message);
+  }
+
+  @Override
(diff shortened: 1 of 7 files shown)


game-core/src/main/java/games/strategy/engine/chat/LobbyChatTransmitter.java
```diff
@@ -0,0 +1,71 @@
...
+    lobbyChatClient.addConnectedListener(chatClient::connected);
+
+    lobbyChatClient.addPlayerSlappedListener(
+        slapEvent -> {
+          if (slapEvent.getSlapped().equals(localPlayerName)) {
+            chatClient.slappedBy(slapEvent.getSlapper());
+          } else {
+            chatClient.playerSlapped(slapEvent.getSlapper() + " slapped " + slapEvent.getSlapped());
+          }
+        });
```
RoiEXLab: Could it ever be the case that this method is called multiple times so that a `ChatClient` instance must stay in memory for the lifetime of the `lobbyChatClient` even when there are no other references?
I think lobbyChatClient should provide a way to clear out old listeners if a new one is set, or alternatively this method should prevent calling itself multiple times

prastle:
I am not a code Guru like u2 But yes I had to close three rooms tonight.  Excellent pt @RoiEXLab. Left over peeps behind.

http-clients/src/main/java/org/triplea/http/client/lobby/chat/LobbyChatClient.java
```diff
@@ -1,10 +1,98 @@
...
+  public void sendChatMessage(final String message) {
+    webSocketClient.send(outboundMessageFactory.sendMessage(message));
+  }
+
+  public void close() {
+    webSocketClient.close();
+  }
+
+  public Collection<ChatParticipant> connect() {
+    webSocketClient.send(outboundMessageFactory.connectToChat());
```
RoiEXLab: Ok, this confused me for a minute:
`connectToChat()` sounded to me like this method connects to the chat and then this method sends some sort of acknowledgement packet to the server which I found weird.
Looking the code up it makes complete sense now, but you might want to consider renaming the method in the future.

RoiEXLab:
@prastle It's more a garbage collection concern, rather than a ghost room concern if I understand you correctly.

prastle:
You understood me correctly.  I hope we can deal with garbage bots somehow :) Or "Ghost Bots"

game-core/src/main/java/games/strategy/engine/chat/LobbyChatTransmitter.java
DanVanAtta: We've essentially a glorified callback, perhaps that perspective will make this make more sense. To another extent, with multiple chat clients, all objects should be going out of scope at the same time. If multiple listeners are interested in chat events, when the chat goes away, so should all the listeners. 

It's not ideal this is even a `set` method, that is largely a constraint of working with the existing chat code. It does not also make sense to have singleton instances as un-registering other listeners would be a surprising side effect as well. The dependency is a bit on its head overall, as we would want the listeners to be constructor args to `lobbyChatClient`, but that creates a circular dependency.

DanVanAtta:
@prastle ghost bots should already be taken care. Of concern is ghost chatters. It seems like there could be an underlying 'ping/pong' keep-alive mechanism that would disconnect any ghost chatters. We should verify that behavior though when it comes time to test.

Otherwise we're concerned/discussing possible memory leaks.

http-clients/src/main/java/org/triplea/http/client/lobby/chat/LobbyChatClient.java
DanVanAtta: I'm not sure of any renaming that would help.
`connect` for `lobbyChatClient` we are stuck with because of how the previous chat worked. The (API) users of the chat transmitter interface are more aware than they should be of async connect vs sync connect, but ultimately we are trying to hide that detail.

For the message factory, a 'connect' message is an explicit connection to chat, logical. Until that is done, chat messages are not routed back. There is also a difference between connecting to the server, and actually connecting to the chat. At some point as well blocking needs to be added to rejected the banned users.

Let me know if you can come up with any ideas for better naming, I think at trouble here is just the multiple different API perspectives

RoiEXLab: @DanVanAtta Ah, I wasn't clear enough:
What I was trying to say is that I didn't realise `connectToChat()` just builds a message to send, and doesn't immediately send it as the verb 'connect' implies.

I mean obviously the name of the variable gives it away that the class is just a factory, but confused me for a while

game-core/src/main/java/games/strategy/engine/chat/LobbyChatTransmitter.java
RoiEXLab: I'd agree that it's problematic that this looks like a setter from the outside.
If this method would rather be an add* method I probably wouldn't even have bothered to mention this, because in this case nobody expects this method to replace any potential previous entry.

PR #5409, opened 2019-11-03
Simplify WebSocket client

RoiEXLab:
This is my design approach to the whole websocket client thing.
Let me know what you think @DanVanAtta 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[x] No manual testing done
[] Manually tested


http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -89,16 +86,14 @@ void close() {
   void connect() {
     Preconditions.checkState(!client.isOpen());
     Preconditions.checkState(!closed);
-    webSocketConnector.initiateConnection();
+    Interruptibles.await(
+        () -> client.connectBlocking(DEFAULT_CONNECT_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS));
```
RoiEXLab: I hardcoded the timeout here, if you think making this configurable for testing is better, let me know

RoiEXLab: Also `connectBlocking()` would return false if anything goes wrong according to the documentation, I wonder if we should use that to allow for better error handling

DanVanAtta: It might make sense to let the integration test land before we update this too much. With that test we'll be able to be sure we have feature parity.

Correct me if I'm mistaken here, I think we may run into an issue with 'send'. The `GenericWebSocketClient` sees the connect as blocking, but it's free to call 'send' whenever it wants. If we then get into the send code, the connection might not yet be open.

RoiEXLab: Let me clear things up for you:
`WebSocketConnection#connect` is blocking so in a single threaded environment if that method is called on initialisation, it blocks until the connection is either established or failed, so any subsequent calls to send occur after a successful or unsuccessful connection, but the state is unambiguous.
So here's the thing with `GenericWebsocketClient`: In the constructor it adds the blocking connect call to the single threaded executor, in an asynchronous way, so the constructor itself isn't blocking, which is the goal.
So whenever there is a call to send before the connection has been established, this send action will be added to the executor queue as well, but because the executor is single-threaded it will not be executed before the initial blocking connection action has been finished.
This obviously doesn't allow us to use more than 1 Thread in the executor, but if we ever get to the point where 1 thread is too low for the amount of incoming messages we have a completely different problem anyways.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -89,16 +86,14 @@ void close() {
...
   }
 
   void sendMessage(final String message) {
     Preconditions.checkState(!closed);
-    webSocketConnector.waitUntilConnectionIsOpen();
     Postconditions.assertState(Thread.currentThread().isInterrupted() || client.isOpen());
```
DanVanAtta: A place to update, no longer a postcondition and interrupted check no longer needed.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
DanVanAtta: Okay, that makes sense to queue the requests and have connection go first. There are some updates needed:
- javadoc of the class containing that threadpool should document that sequencing
- GenericWebsocketClient should guard against 'send' being called before 'connect'. It seems like if we get a call in that ordering, that send would fail or block, when before it could have been called at the same time or out of order
- The javadoc of this class should be updated as well, it's out of date
- The TODO statement on the 'send' method in generic websocket is out of date, we now require the single threaded threadpool. Perhaps commenting on the threadpool that it's single threaded to be effectively a blocking queue. 
- I suspect the test cases should be updated to check for the new behavior. 

That is hopefully relatively exhaustive of a list. Again I'd like to see this class and set of updates be run against the in-flight integ test that verifies a chat flow.

DanVanAtta:
@RoiEXLab there are some last updates, see above for a list, should be fast to knock-out, and then verify that the updated code still passes the tests in: https://github.com/triplea-game/triplea/pull/5417

DanVanAtta:
@RoiEXLab the test I was mentioning was updated. I agree this is a simpler approach, if the documentation/TODOs are all updated, and the tests still pass, this should be good to go.

RoiEXLab:
Yeah, I'll try to get this PR done this evening

RoiEXLab:
@DanVanAtta Updated, let me know what you think

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
```diff
@@ -49,7 +51,9 @@ public GenericWebSocketClient(
     this.messageListener = messageListener;
     client = webSocketClient;
     client.addListener(this);
-    client.connect();
+    CompletableFutureUtils.logExceptionWhenComplete(
+        CompletableFuture.runAsync(client::connect, threadPool),
```
DanVanAtta: If there is an exception on connect, will that bubble up and notify the user? The underlying code had a parameterized timeout to essentially test that case. Is this failure mode still tested so we can be sure we have the expected behavior?

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
```diff
@@ -68,7 +71,10 @@ public void send(final OutgoingT message) {
 
   /** Non-blocking close of the websocket connection. */
   public void close() {
-    client.close();
+    CompletableFutureUtils.logExceptionWhenComplete(
+        CompletableFuture.runAsync(client::close, threadPool),
+        e -> log.log(Level.WARNING, "Failed to close client", e));
```
DanVanAtta: Do you remember if warnings are shown (as a pop-up) to users by default?

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
```diff
@@ -61,59 +55,36 @@ void onError() {
...
 
     @Test
-    void connect() {
+    void connect() throws Exception {
       webSocketConnection.connect();
 
-      verify(webSocketConnector).initiateConnection();
+      verify(webSocketClient)
+          .connectBlocking(
+              WebSocketConnection.DEFAULT_CONNECT_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
```
DanVanAtta: Very nitpicky, I'm not sure if we gain much by verifying the parameters of this test, we could make this more flexible by just asserting that any arg is passed to connect blocking.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -74,31 +69,30 @@ void addListener(final WebSocketConnectionListener listener) {
...
+   */
   void sendMessage(final String message) {
     Preconditions.checkState(!closed);
-    webSocketConnector.waitUntilConnectionIsOpen();
-    Postconditions.assertState(Thread.currentThread().isInterrupted() || client.isOpen());
-    // if we aborted waiting for the connection, current thread will be interrupted, do a no-op.
-    if (!Thread.currentThread().isInterrupted()) {
-      client.send(message);
-    }
+    client.send(message);
```
DanVanAtta: Any thoughts to checking `client.isOpen()`? I vaguely recall cases where that was an important check during system testing. It would be good to see any error behavior here be covered by unit tests so that we can verify the handling for these edge-cases that are hard to reproduce manually.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -74,31 +69,30 @@ void addListener(final WebSocketConnectionListener listener) {
...
+  /**
+   * Initiates a websocket connection. Must be called before {@link #sendMessage(String)}. This
+   * method blocks until the connection has either successfully been established, or the connection
+   * failed.
+   */
   void connect() {
     Preconditions.checkState(!client.isOpen());
     Preconditions.checkState(!closed);
-    webSocketConnector.initiateConnection();
+    Interruptibles.await(
```
DanVanAtta: The return value of the interruptible is not used. If we go on to call `send`, we may get an odd exception. Do you have any thoughts about this error-case scenario and how to handle it? Perhap same considerations for when we fail to establish a connection.

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
RoiEXLab: `ErrorMessageHandler` defines `Level.WARNING` as minimum level

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
RoiEXLab: Hmm, good point, I don't think that'll actually be the case. I'll add that

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
DanVanAtta: Thanks for checking, we may want to consider the error handling; errors tend to have users overreact and given we're dealing with a thick client, logging is of very limited value.

Perhaps if the connection winds up being closed, we can `info` or `debug` log this so there is no pop-up, effectively handling the error. If the connection is still open, we could consider a retry, though I'm not sure if that is really worth it.

RoiEXLab: I'll leave it up to you to decide this, the log statements are all in one place now, so it's easy to change them all.

RoiEXLab:
@DanVanAtta Updated, please have another look

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -77,12 +78,18 @@ void close() {
...
   void connect() {
     Preconditions.checkState(!client.isOpen());
     Preconditions.checkState(!closed);
-    Interruptibles.await(
-        () -> client.connectBlocking(DEFAULT_CONNECT_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS));
+    final Result<Boolean> result =
+        Interruptibles.awaitResult(
+            () -> client.connectBlocking(DEFAULT_CONNECT_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS));
+    if (!result.completed || !result.result.orElse(false)) {
+      throw new RuntimeException("Failed to connect to a websocket at " + client.getURI());
```
DanVanAtta: Instead of throwing RuntimeException, it might be good to resurrect the `CouldNotConnect` exception from `WebSocketConnector`:
```
  /** Exception indicating connection to server failed. */
  @VisibleForTesting
  static final class CouldNotConnect extends RuntimeException {
    private static final long serialVersionUID = -5403199291005160495L;

    private static final String ERROR_MESSAGE = "Error, could not connect to server";

    CouldNotConnect() {
      super(ERROR_MESSAGE);
    }

    CouldNotConnect(final Exception e) {
      super(ERROR_MESSAGE, e);
    }
  }
```


I note this as the audience for this error message is the average game player, not developers. A game player will not necessarily know about a websocket, what it's used for, or will they really even care. They'll want to know what the error means to them and the game.  

I don't know if this is that easy to fix. We probably should ideally surface this as a boolean return value and allow the client code decide how to message the failure. The only problem is that in turn, `GenericWebsocketClient` should do the same thing, but we have an async connect which makes it tricky.

There can be a number of solutions, if of any help, I can suggest two courses of action:
1) Surface a boolean return value to `GenericWebSocketClient`, and pass an error message factory to `GenericWebsocketClient` to display on case of failure to connect.
2) Perhaps simpler, just fix up the error message. I think it's debatable if we need to report the URI to the client (could be noise, the URI is completely transparent to users, so it might not make sense to report it). could be something like:
```
throw new  CouldNotConnect()
```
or:
```
throw new CouldNotConnect(client.getURI());
:
:
    CouldNotConnect(URI uri) {
      super(ERROR_MESSAGE + " at " + uri);
    }
```

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
```diff
@@ -70,11 +71,32 @@ void setup() {
...
+      when(webSocketClient.connectBlocking(anyLong(), any())).thenReturn(true);
       webSocketConnection.connect();
 
       verify(webSocketClient).connectBlocking(anyLong(), any());
     }
 
+    @Test
+    void connectFails() {
+      webSocketConnection.setClient(
+          new WebSocketClient(URI.create("wss://server.invalid")) {
```
DanVanAtta: Does this fail quickly, or is there a timeout? I wonder in regards to test runtime.

On a minor note, it took some reverse engineering to figure out why we expect a RuntimeException. Perhaps that could be fixed by adding a `@DisplayName` annotation, eg:
```
@DisplayName("Verify connect fails with exception when connecting to a non-existent endpoint");
```

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
```diff
@@ -81,10 +82,19 @@ void close() {
...
       verify(webSocketClient).send(MESSAGE);
     }
+
+    @Test
+    void sendMessageFailsIfConnectionNotOpened() {
+      when(webSocketClient.isOpen()).thenReturn(false);
+
+      assertThrows(IllegalStateException.class, () -> webSocketConnection.sendMessage(MESSAGE));
+
+      verify(webSocketClient, times(0)).send(MESSAGE);
```
DanVanAtta: nit: `Mockito.never()` AFAIK is preferred over `times(0)`  (in the same way as `isEmpty()` is preferred over `length == 0`)

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
DanVanAtta: I think the consideration is more or less as follows:
- are we going to be spamming users with error notifications? For example, will this be a N+1 error notification, will there be some other error beforehand and then a user sees this error message popup.
- Do we want this error reported? A popup will have a 'report to triplea' attached to it, is this something that we will want users to report to us?
- Do we want to expect the user to restart TripleA after seeing this? For better or worse, most users assume on any error that they should restart TripleA. I've observed that behavior often enough I consider it a rule of thumb, it's a bit funny sometimes when errors are relatively benign, but people go through the trouble of rehosting a game and restarting everything.

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
RoiEXLab: Good call, I knew there was something to achieve this, but I couldn't remember that's what it was called

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
RoiEXLab: I just tested this again, it takes slightly longer than normal unit test. This test case takes about 5 seconds to execute. I'll try to inject a mock socket factory to speed things up, the tradeoff is that we might end up being further away from reality

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -77,12 +78,18 @@ void close() {
...
    * failed.
+   *
+   * @throws RuntimeException If the connection fails
    */
   void connect() {
     Preconditions.checkState(!client.isOpen());
     Preconditions.checkState(!closed);
-    Interruptibles.await(
-        () -> client.connectBlocking(DEFAULT_CONNECT_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS));
+    final Result<Boolean> result =
```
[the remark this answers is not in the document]
RoiEXLab: That works I guess 🤷‍♂

RoiEXLab:
@DanVanAtta Adressed your comments, this should be good to go now, please have another look

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
DanVanAtta: Thanks for checking, have you considered the trade-off of toggling the timeout to be something very short, like 5ms?

RoiEXLab: Considered that, but by looking at the code it seems that the timeout is already 0?
Not sure if this is the value for "default timeout" or something but I haven't actually tried using 1

DanVanAtta:
I do wonder if the test case would be a good bit simpler to have a short timeout:
- avoids a number of mock calls, this is very good as the test becomes less coupled to implementation
- avoids the need for the socket factory injection, which adds more coupling
- the test documentation/commentary can likely be simpler

Again, feel free to fix and merge or merge and follow-up with the test fixes. Curious on your thoughts to that regard about test coupling/simplicitiy.

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
DanVanAtta: I'm a bit confused, I was expecting this chunk of code to control the connectoin timeout:
```
    final Result<Boolean> connectionAttempt =
        Interruptibles.awaitResult(
            () -> client.connectBlocking(DEFAULT_CONNECT_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS));
```

My thinking is we could go back to connect timeout to be parameterized so that test could control it and then inject a very short timeout value. Thoughts? I'm also curious where you are spotting the timeout of '0'.

RoiEXLab: Ah, of course. I'll blame it on the fact that I'm pretty tired right now ^^
The 0 timeout actually represents the internal timeout of WebSocketClient. I had to dig into the source code to be able to make the test work, so that's where I found it.

Anyways I feel like this is going back and forth, which is why I'd suggest to merge this PR now, so you can try to find a solution that 1. works and 2. is better in your opinion.
I also think that making the external timeout/our custom timeout super short only tests a different scenario where the connection actually takes too long to establish (maybe caused by high packet loss) vs a non-responding server.

I also tried using a single WebSocketClient mock for all the test cases to keep things simpler, however it seems like this actually achieved the opposite for this one test case.

In any case I'll recommend you fiddle around with the test code a little. I'd say it's in a fairly decent shape right now, but I agree that it can be improved, I'm just not sure how exactly it should be improved.

DanVanAtta: Sounds good :+1:

PR #5411, opened 2019-11-03
Prerelease deployment updates

DanVanAtta:
A series of updates to get us on track to be able to run a deployment from travis as a post-build step. As noted in the README, we now have a path open to us with travis build stages to run deployments after artifacts are deployed. Previously we would have staged ansible scripts from a control server, downloaded artifacts out-of-band and run the deployment, because we have that 'post-step' now, we can do this from travis directly.

Included here are a series of changes that gets along on this path:
- a number of fixes so that the deployment finishes/works
- adding infrastructure and notes about ansible-vault so we can use encrypted values and not need as many environment variables.
- update README with some rough notes and updated to better reflect this latest plan

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[x] Manually tested

## Additional Review Notes
Still todo:
- we need to deploy certs properly 
- fix up our code to use https.
- avoid breaking local development, and use certs + https locally
- update travis configuration to execute the deployments


infrastructure/ansible/roles/lobby_database/tasks/main.yml
```diff
@@ -20,6 +20,7 @@
       - postgresql
       - postgresql-contrib
       - libpq-dev
+      - python-psycopg2
```
RoiEXLab: Is this python 2 dependency really needed? :(
python 2 is supposed to be end of life at the end of this year

DanVanAtta: Yes, I was tempted to remove the python 3 dependency, though that limits the number of environments we can deploy to. Vagrant environment uses the python 3 version. Ubuntu 18.04 LTS needed the python 2 dep.

DanVanAtta:
Snyk project is removed.

PR #5413, opened 2019-11-03
Small optimization to Route constructor.

asvitkine:
No need to create a temporary list and array.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[X] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[X] No manual testing done
[] Manually tested

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/Route.java
+++ b/game-core/src/main/java/games/strategy/engine/data/Route.java
@@ -40,18 +40,19 @@ public class Route implements Serializable, Iterable<Territory> {
   private final Territory start;
   private final List<Territory> steps = new ArrayList<>();
 
-  public Route(final List<Territory> territories) {
-    this.start = checkNotNull(territories.get(0));
-    for (int i = 1; i < territories.size(); i++) {
-      add(territories.get(i));
-    }
-  }
-
-  public Route(final Territory start, final Territory... territories) {
+  public Route(final Territory start, final List<Territory> territories) {
     this.start = checkNotNull(start);
     for (final Territory t : territories) {
       add(t);
     }
+  }
+
+  public Route(final List<Territory> territories) {
+    this(territories.get(0), territories.subList(1, territories.size()));
+  }
+
+  public Route(final Territory start, final Territory... territories) {
+    this(start, List.of(territories));
   }
 
   /** Add the given territory to the end of the route. */
--- a/game-core/src/main/java/games/strategy/engine/data/Route.java
+++ b/game-core/src/main/java/games/strategy/engine/data/Route.java
@@ -41,9 +41,10 @@ public class Route implements Serializable, Iterable<Territory> {
   private final List<Territory> steps = new ArrayList<>();
 
   public Route(final List<Territory> territories) {
-    this(
-        territories.get(0),
-        territories.subList(1, territories.size()).toArray(new Territory[territories.size() - 1]));
+    this.start = checkNotNull(territories.get(0));
+    for (int i = 1; i < territories.size(); i++) {
+      add(territories.get(i));
+    }
   }
 
   public Route(final Territory start, final Territory... territories) {


game-core/src/main/java/games/strategy/engine/data/Route.java
```diff
@@ -41,9 +41,10 @@
   private final List<Territory> steps = new ArrayList<>();
 
   public Route(final List<Territory> territories) {
-    this(
```
DanVanAtta: One downside I see is that we have two constructors now with differing implementations. Perhaps debatable, but it can be easier to reason about a class when there is one constructor that has logic and all the other constructors invoke that constructor. If there is an update for example to how this class is initialized, we might forget to update both constructors. Or, when reading this class, there are two constructors that need to be taken into account when reasoning about the state of the class.

Is there a way we can avoid the extra list but still have effectively one constructor?

DanVanAtta: Maybe the other constructor can be updated to invoke this one cleanly?

RoiEXLab: I just want to point out that `List#subList` doesn't actually copy all elements in a list, but rather returns a read-only view of the elements in this area, so the only optimization that's happening here is that a single object less is allocated which is negligible IMO

asvitkine: The toArray() call does though. It creates an array of the proper size and toArray() copies the entries to that array.

RoiEXLab: Hmm I wonder if the other constructor actually makes sense this way.
I could imagine that in most cases such a split constructor makes things more difficult to deal with. In that case the whole discussion would become redundant

RoiEXLab: Also small observation, I'd say it's more elegant to do a foreach loop over the sublist tbh, but in the end it doesn't really make a difference

ron-murhammer: Its funny as I think this is close to what I originally had then @DanVanAtta and I discussed and he strongly preferred having the constructors invoke each other so I changed it. I really don't have a strong preference here.

DanVanAtta: Indeed, we seem to have gone circles here, I'll create an issue for a convention on this so we can capture it. I've seen a number of cases where multiple constructor logics is a straight-up anti-pattern. When reviewing any class code, with multiple constructors, the reasoning of state is just that much more difficult as you must know the code path that lead you to that point to know how the class was initialized. This is a pretty simple case, but it's still the same point. If there is no good way to unify the two constructors, I'd recommend going with a factory.

DanVanAtta: Sorry to pile on a bit, responding to:

> I don't think we can make the other constructor call this one without creating some extra temporaries. 

In java, tightly scoped objects, objects that are created and then quickly fall out of scope and are GC'd, are incredibly cheap. Generally the optimization should be for tightly scoped objects and developer maintenance.

asvitkine: I agree that small objects are cheap. But the current code create an array and fills it from the sublist for no reason.

And the existing code is more complex than my suggested replacement. (territories.subList(1, territories.size()).toArray(new Territory[territories.size() - 1]) is quite a statement)

I agree that all things being equal, having a single constructor impl is better. But in this case, the class is very simple and having two is more readable than the current single constructor impl (where we have the sublist toArray() monstrosity above) and more efficient.

I don't think we should apply the "only a single constructor impl" as a hard rule, but as a guidance when doing so doesn't introduce other issues (in this case, complexity and performance).

As a case in point, I checked the Google Java style guide (who've written millions of Java code I'm sure) - and it doesn't mention such a rule.

asvitkine: The other option is to go with 3 ctors:

  public Route(final Territory start, final List<Territory> territories) {
    this.start = checkNotNull(start);
    for (final Territory t : territories) {
      add(t);
    }
  }

  public Route(final Territory start, final Territory... territories) {
    this(start, List.of(territories));
  }

  public Route(final List<Territory> territories) {
    this(territories.get(0), territories.subList(1, territories.size()));
  }

We still have some temporaries and now there's an extra constructor, but it allows both to reuse the code.

DanVanAtta: I don't necessarily disagree. I am concerned about a slippery slope though (what happens when we have a 3rd constructor, and similar arguments), there is some value in sticking to a hard and fast rule (to an extent). Google Java Style guide does not cover a lot, for example we tend to favor the points in effective java, those are outside of the scope of formatting rules.

We should back up and consider what we are gaining. The code was already one form, we are potentially ping-ponging back, which is not a good sign. The performance/space complexity gain is very minimal, we are just fixing up one slightly hard to read line and adding other costs involved.

My recommendation is look for the bigger win here. Perhaps a factory can be used to improve this, perhaps we can fix the internal representation of the class to not the explicit tracking of the `start` territory and just maintain a simple list.

Looking to the usages of these constructors, the subList is a pretty minimal concern, there seems to be lots of room for improvement. Again, let's look to the clear wins and avoid the code churn because one line of code seems offensive.

DanVanAtta: Just read your comment @asvitkine , the 3 ctor option looks good

asvitkine: OK, let's go with that. I still find it a bit unfortunate, but I do think it's an improvement over what's currently there in terms of readability and performance.

PR #5415, opened 2019-11-04
Optimize TripleAUnit.getTransporting().

asvitkine:
This is called extensively by ProAI and this change improves the runtime significantly.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[X] Manually tested

### Before
Profile showing 22 seconds being spent in isTransporting() called by AI.
Screen Shot 2019-11-03 at 8 08 45 PM

### After
Profile showing this time disappears with this change.
Screen Shot 2019-11-03 at 8 09 28 PM

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/TripleAUnit.java
+++ b/game-core/src/main/java/games/strategy/triplea/TripleAUnit.java
@@ -110,8 +110,7 @@ public class TripleAUnit extends Unit {
    * method if possible.
    */
   public List<Unit> getTransporting() {
-    final boolean isTransport = UnitAttachment.get(getType()).getTransportCapacity() > 0;
-    if (isTransport) {
+    if (Matches.unitCanTransport().test(this) || Matches.unitIsCarrier().test(this)) {
       // we don't store the units we are transporting
       // rather we look at the transported by property of units
       for (final Territory t : getData().getMap()) {
--- a/game-core/src/main/java/games/strategy/triplea/TripleAUnit.java
+++ b/game-core/src/main/java/games/strategy/triplea/TripleAUnit.java
@@ -110,12 +110,15 @@ public class TripleAUnit extends Unit {
    * method if possible.
    */
   public List<Unit> getTransporting() {
-    // we don't store the units we are transporting
-    // rather we look at the transported by property of units
-    for (final Territory t : getData().getMap()) {
-      // find the territory this transport is in
-      if (t.getUnitCollection().contains(this)) {
-        return getTransporting(t.getUnitCollection());
+    final boolean isTransport = UnitAttachment.get(getType()).getTransportCapacity() > 0;
+    if (isTransport) {
+      // we don't store the units we are transporting
+      // rather we look at the transported by property of units
+      for (final Territory t : getData().getMap()) {
+        // find the territory this transport is in
+        if (t.getUnitCollection().contains(this)) {
+          return getTransporting(t.getUnitCollection());
+        }
       }
     }
     return Collections.emptyList();


asvitkine:
Test failures seem unrelated. Something with WebSocket?

RoiEXLab:
Yup, looks like a flaky test, my currently active PR will touch this part of the code anyways, so no reason to worry for now.
I restarted the build.
I'd like to hear @ron-murhammer's opinion on this

DanVanAtta:
It is worth double checking a number of things:
- how is mech infantry handled? Does it increment transport capacity or is another property used
- On attachment, there is a `getCarrierCapacity()`, are planes marked as 'transported by?'

I'm not terribly comfortable seeing refactoring/optimizations to untested code. "Better to optimize correct code than it is to correct optimize code". @asvitkine would you be opposed to building in some testing around this logic. This would give more confidence to the current update and in the future save time by ensuring this knowledge is kept and tested anytime this code is updated.

Any unit test could suffer from just testing our own assumptions which could prove to be incorrect. Before doing that I would recommend first adding some in-place testing. Basically have both versions of the method, compare the respective return values and ensure they are the same. Then if you play test with: transports, mech infantry, carriers, cargo planes, paratroops; we could be confident we've correctness. Then if we land a unit test, we'll never need to repeat that exercise.

asvitkine:
If we want to do significantly more work on this than something simple like this, I actually suggest it would be better to start tracking transported units explicitly in a collection.

This way, we don't need to do any work to find the units that are being transported. As otherwise, even with this change, it's still expensive to compute this for actual transports. For example, the algorithm to load units into transports calls into this for actual transports - and this happens multiple times as the move is validated and so forth.

If this sounds reasonable, I can look at doing this. Of course, we can also find ways to add unit tests for that as well.

ron-murhammer:
@asvitkine That is definitely the longer term solution but given that we already have some bad bugs in the pre-release and I want to release soon, I'd prefer to avoid a massive change to transported unit tracking right now.

asvitkine:
That's fair.

However, I do feel like getting this fixed for the next release should be something we should do. The AI spending 20s or more in this function for units that are not transports is just sad.

I can definitely check carrier capacity too. But I'm not sure how ground transports are handled.  What map has those? I assume the units would be defined in the XML somewhere and I can just . look how they're set up?

DanVanAtta:
mechanized infantry is an example of ground transport. Would recommend setting up a map, enabling that tech, saving and reloading with AI. For full confidence, really recommend adding some in-game testing to check the two different return values are always the same.  With that setup, verifying mechanized infrantry, and a few standard/sample games, our confidence would be pretty high that this is not risky.

That code would be pretty easy to straight-forward to setup, ie:

```
 public List<Unit> getTransporting() {
     var control = getTransportingControl();
     var experiment = getTransportingExperiment();
     if(!control.equals(experiment) {
          throw new RuntimeException("not equal");
     }
     return experiment;
  }
 



  private List<Unit> getTransportingControl() {
    // we don't store the units we are transporting
    // rather we look at the transported by property of units
    for (final Territory t : getData().getMap()) {
      // find the territory this transport is in
      if (t.getUnitCollection().contains(this)) {
        return getTransporting(t.getUnitCollection());
      }
    }
        return Collections.emptyList();
   }


private List<Unit> getTransportingExperiment() {
    final boolean isTransport = UnitAttachment.get(getType()).getTransportCapacity() > 0;
    if (isTransport) {
      // we don't store the units we are transporting
      // rather we look at the transported by property of units
      for (final Territory t : getData().getMap()) {
        // find the territory this transport is in
        if (t.getUnitCollection().contains(this)) {
          return getTransporting(t.getUnitCollection());
        }
      }
    }
    return Collections.emptyList();
}
```

DanVanAtta:
Do agree on release priority,  we've 10 open pre-release issues, 4 assigned to me, I'm wrapping up http-lobby server migration which will resolve those 4, the other 6 need volunteers & work.

asvitkine:
Looking more closely, if I'm understanding the code right, the way to set the "transportedBy" property is via:

    change.add(ChangeFactory.unitPropertyChange(unit, transport, TripleAUnit.TRANSPORTED_BY));

This is done in exactly one place (it's cleared in a few others), which is TransportTracker.loadTransportChange()

This is called from the following places:
  1. EditDelegate.addUnits() - this only looks at loading land units on sea transports.
  2. InitializationDelegate.initTransportedLandUnits() - this also matches only unitIsLand() onto unitIsTransport() - where unitIsTransport() checks getTransportCapacity() and that it's a sea unit.
  3. MovePerformer.markTransportsMovement() - two calls here, this seems to also handle paratroopers.
  4. TransportTracker.reloadTransports() - this reloads units onto transports that were unloaded. Looks like this is called during scrambling and when transports retreat from combat.
  5. And some test code in WW2V3Year41Test

So in conclusion, I think checking getTransportCapacity() and getCarrierCapacity() should be reasonable.

Probably,  I can use Matches.unitIsCarrier() and Matches.unitCanTransport() functions for these, since they exist and abstract away the details.

asvitkine:
So I changed the if to:
```
    if (Matches.unitCanTransport().test(this) || Matches.unitIsCarrier().test(this)) {
```

Now that I'm relying on standard APIs for testing this, I'm hoping this should address concerns.

asvitkine:
> mechanized infantry is an example of ground transport. Would recommend setting up a map, enabling that tech, saving and reloading with AI. For full confidence, really recommend adding some in-game testing to check the two different return values are always the same. With that setup, verifying mechanized infrantry, and a few standard/sample games, our confidence would be pretty high that this is not risky.
> 
> That code would be pretty easy to straight-forward to setup, ie:
> 
> ```
>  public List<Unit> getTransporting() {
>      var control = getTransportingControl();
>      var experiment = getTransportingExperiment();
>      if(!control.equals(experiment) {
>           throw new RuntimeException("not equal");
>      }
>      return experiment;
>   }
>  
> 
> 
> 
>   private List<Unit> getTransportingControl() {
>     // we don't store the units we are transporting
>     // rather we look at the transported by property of units
>     for (final Territory t : getData().getMap()) {
>       // find the territory this transport is in
>       if (t.getUnitCollection().contains(this)) {
>         return getTransporting(t.getUnitCollection());
>       }
>     }
>         return Collections.emptyList();
>    }
> 
> 
> private List<Unit> getTransportingExperiment() {
>     final boolean isTransport = UnitAttachment.get(getType()).getTransportCapacity() > 0;
>     if (isTransport) {
>       // we don't store the units we are transporting
>       // rather we look at the transported by property of units
>       for (final Territory t : getData().getMap()) {
>         // find the territory this transport is in
>         if (t.getUnitCollection().contains(this)) {
>           return getTransporting(t.getUnitCollection());
>         }
>       }
>     }
>     return Collections.emptyList();
> }
> ```

Thanks for the suggestion, let me try that.

asvitkine:
I did the test @DanVanAtta suggested and no problems were detected. I also verified manually that mech infantry was otherwise working. Looks like it's just used to help other units move, but doesn't have any "state" of units being transported by it.

ron-murhammer:
@asvitkine Yeah, land transports don't truly "transport" and essentially check the units that they enable to move along with them during the move since they are done in one move vs sea transports doing load-move-unload so need to track the state.

DanVanAtta:
:+1:

PR #5416, opened 2019-11-04
Batch AI amphibious moves.

asvitkine:
Prior to this change, the AI would load and move transports one unit at a time.
So moving 20 units via transports would take 40 moves:
  - 20 x (1 load, 1 load, 1 move, 1 unload).

With this change, it's just 3 moves (when they all have the same destination):
  1. Load 20 units
  2. Move 10 transports
  3. Unload 20 units.

This makes a big difference when playing with AI since this makes
the AI moves be much faster when amphibious moves are involved.

Additionally, the code is changed to make the player pass the mapping from units to transports to the backend, so that the AI could convey this directly without having the backend recalculate it and potentially get it wrong.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[X] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing
[] No manual testing done
[X] Manually tested

Loaded a save game where US has a lot of stuff to move. Observed the moves it made.
Loaded a save game where UK has a combat move involving moving transports, picking up units and landing them. Observed the moves it made.
Also added some unit tests.

### Before

Screen Shot 2019-11-03 at 10 35 17 PM

### After

Screen Shot 2019-11-03 at 10 38 37 PM

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
@@ -1,0 +1,157 @@
+package games.strategy.triplea.ai.pro.util;
+
+import games.strategy.engine.data.Route;
+import games.strategy.engine.data.Unit;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+
+/**
+ * MoveBatcher implements an algorithm that batches AI moves together, reducing the number of
+ * distinct moves the AI does (and thus the visual delay) by combining moves together.
+ */
+public class MoveBatcher {
+  private static class Move {
+    private final ArrayList<Unit> units;
+    private final Route route;
+    private final HashSet<Unit> transportsToLoad;
+
+    Move(final ArrayList<Unit> units, final Route route, final HashSet<Unit> transportsToLoad) {
+      this.units = units;
+      this.route = route;
+      this.transportsToLoad = transportsToLoad;
+    }
+
+    Move(final ArrayList<Unit> units, final Route route) {
+      this(units, route, null);
+    }
+
+    Move(final Unit unit, final Route route, final Unit transportToLoad) {
+      this(mutableSingletonList(unit), route, new HashSet<Unit>(List.of(transportToLoad)));
+    }
+
+    private static ArrayList<Unit> mutableSingletonList(final Unit unit) {
+      return new ArrayList<>(Collections.singletonList(unit));
+    }
+
+    boolean isTransportLoad() {
+      return this.transportsToLoad != null;
+    }
+
+    boolean canMergeWith(final Move other) {
+      return other.isTransportLoad() == isTransportLoad() && route.equals(other.route);
+    }
+
+    boolean mergeWith(final Move other) {
+      if (canMergeWith(other)) {
+        // Merge units and transports.
+        units.addAll(other.units);
+        if (isTransportLoad()) {
+          transportsToLoad.addAll(other.transportsToLoad);
+        }
+        return true;
+      }
+      return false;
+    }
+
+    void addTo(
+        final List<Collection<Unit>> moveUnits,
+        final List<Route> moveRoutes,
+        final List<Collection<Unit>> transportsToLoad) {
+      moveUnits.add(units);
+      moveRoutes.add(route);
+      transportsToLoad.add(this.transportsToLoad);
+    }
+  }
+
+  private final ArrayList<ArrayList<Move>> moveSequences = new ArrayList<ArrayList<Move>>();
+
+  /**
+   * Starts a new sequence. This must be called before calling any add*() methods. A sequence
+   * indicates a logical dependency relationship between the moves in that sequence.
+   */
+  public void newSequence() {
+    moveSequences.add(new ArrayList<Move>());
+  }
+
+  /**
+   * Adds a transport load move. Ownership of the params is transferred to this object.
+   *
+   * @param unit The unit to move.
+   * @param route The route to move on.
+   * @param transport The transport to load.
+   */
+  public void addTransportLoad(final Unit unit, final Route route, final Unit transport) {
+    addMove(new Move(unit, route, transport));
+  }
+
+  /**
+   * Adds a move for the specified units. Ownership of the params is transferred to this object.
+   *
+   * @param units The units to move. Note that type is ArrayList explicitly to prevent an
+   *     unmodifiable List.of() being passed.
+   * @param route The route to move on.
+   */
+  public void addMove(final ArrayList<Unit> units, final Route route) {
+    addMove(new Move(units, route));
+  }
+
+  private void addMove(final Move newMove) {
+    final ArrayList<Move> sequence = moveSequences.get(moveSequences.size() - 1);
+    if (!sequence.isEmpty() && sequence.get(sequence.size() - 1).mergeWith(newMove)) {
+      return;
+    }
+    sequence.add(newMove);
+  }
+
+  /**
+   * Emits the accumulated moves to the passed output params.
+   *
+   * @param moveUnits The units to move.
+   * @param moveRoutes The routes to move along
+   * @param transportsToLoad The transports to be loaded.
+   */
+  public void batchAndEmit(
+      final List<Collection<Unit>> moveUnits,
+      final List<Route> moveRoutes,
+      final List<Collection<Unit>> transportsToLoad) {
+    int i = 0;
+    for (final var sequence : moveSequences) {
+      if (!sequence.isEmpty()) {
+        mergeSequences(sequence, moveSequences.subList(i + 1, moveSequences.size()));
+        for (final Move move : sequence) {
+          move.addTo(moveUnits, moveRoutes, transportsToLoad);
+        }
+      }
+      i++;
+    }
+  }
+
+  /**
+   * Attempts to merge a sequence of moves with other sequences. Any sequence merged into this
+   * sequence will have its moves cleared.
+   *
+   * @param sequence The sequence to merge other sequences into.
+   * @param sequences The sequences to try to merge into the sequence.
+   */
+  private static void mergeSequences(
+      final ArrayList<Move> sequence, final List<ArrayList<Move>> sequences) {
+    for (final var otherSequence : sequences) {
+      boolean merge = (otherSequence.size() == sequence.size());
+      for (int j = 0; merge && j < sequence.size(); j++) {
+        merge = sequence.get(j).canMergeWith(otherSequence.get(j));
+      }
+      if (!merge) {
+        continue;
+      }
+      for (int j = 0; j < sequence.size(); j++) {
+        if (!sequence.get(j).mergeWith(otherSequence.get(j))) {
+          throw new RuntimeException("Unexpected");
+        }
+      }
+      otherSequence.clear();
+    }
+  }
+}
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
@@ -4,10 +4,10 @@ import games.strategy.engine.data.Route;
 import games.strategy.engine.data.Unit;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import javax.annotation.Nullable;
 
 /**
  * MoveBatcher implements an algorithm that batches AI moves together, reducing the number of
@@ -17,12 +17,12 @@ public class MoveBatcher {
   private static class Move {
     private final ArrayList<Unit> units;
     private final Route route;
-    private final HashMap<Unit, Unit> unitsToTransports;
+    private final @Nullable HashMap<Unit, Unit> unitsToTransports;
 
     private Move(
         final ArrayList<Unit> units,
         final Route route,
-        final HashMap<Unit, Unit> unitsToTransports) {
+        final @Nullable HashMap<Unit, Unit> unitsToTransports) {
       this.units = units;
       this.route = route;
       this.unitsToTransports = unitsToTransports;
@@ -33,14 +33,10 @@ public class MoveBatcher {
     }
 
     Move(final Unit unit, final Route route, final Unit transportToLoad) {
-      this(mutableSingletonList(unit), route, new HashMap<>(Map.of(unit, transportToLoad)));
+      this(new ArrayList<>(List.of(unit)), route, new HashMap<>(Map.of(unit, transportToLoad)));
     }
 
-    private static ArrayList<Unit> mutableSingletonList(final Unit unit) {
-      return new ArrayList<>(Collections.singletonList(unit));
-    }
-
-    boolean isTransportLoad() {
+    private boolean isTransportLoad() {
       return this.unitsToTransports != null;
     }
 
@@ -81,7 +77,7 @@ public class MoveBatcher {
   }
 
   /**
-   * Adds a transport load move. Ownership of the params is transferred to this object.
+   * Adds a transport load move.
    *
    * @param unit The unit to move.
    * @param route The route to move on.
@@ -92,7 +88,7 @@ public class MoveBatcher {
   }
 
   /**
-   * Adds a move for the specified units. Ownership of the params is transferred to this object.
+   * Adds a move for the specified units. The units list must not be changed after adding the move.
    *
    * @param units The units to move. Note that type is ArrayList explicitly to prevent an
    *     unmodifiable List.of() being passed.
@@ -112,6 +108,9 @@ public class MoveBatcher {
 
   /**
    * Emits the accumulated moves to the passed output params.
+   *
+   * <p>TODO: #5416 Refactor the IMoveDelegate and ProMoveUtils code to use a List of Move objects,
+   * instead of these parallel lists.
    *
    * @param moveUnits The units to move.
    * @param moveRoutes The routes to move along
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
@@ -33,10 +33,7 @@ public class MoveBatcher {
     }
 
     Move(final Unit unit, final Route route, final Unit transportToLoad) {
-      this(
-          mutableSingletonList(unit),
-          route,
-          new HashMap<>(Map.of(unit, transportToLoad)));
+      this(mutableSingletonList(unit), route, new HashMap<>(Map.of(unit, transportToLoad)));
     }
 
     private static ArrayList<Unit> mutableSingletonList(final Unit unit) {
@@ -155,7 +152,8 @@ public class MoveBatcher {
       }
       for (int j = 0; j < sequence.size(); j++) {
         if (!sequence.get(j).mergeWith(otherSequence.get(j))) {
-          throw new RuntimeException("Unexpected");
+          throw new IlegalStateException(
+              "Could not merge move despite checking canMergeWith() earlier.");
         }
       }
       otherSequence.clear();
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
@@ -36,7 +36,7 @@ public class MoveBatcher {
       this(
           mutableSingletonList(unit),
           route,
-          new HashMap<Unit, Unit>(Map.of(unit, transportToLoad)));
+          new HashMap<>(Map.of(unit, transportToLoad)));
     }
 
     private static ArrayList<Unit> mutableSingletonList(final Unit unit) {
@@ -73,14 +73,14 @@ public class MoveBatcher {
     }
   }
 
-  private final ArrayList<ArrayList<Move>> moveSequences = new ArrayList<ArrayList<Move>>();
+  private final ArrayList<ArrayList<Move>> moveSequences = new ArrayList<>();
 
   /**
    * Starts a new sequence. This must be called before calling any add*() methods. A sequence
    * indicates a logical dependency relationship between the moves in that sequence.
    */
   public void newSequence() {
-    moveSequences.add(new ArrayList<Move>());
+    moveSequences.add(new ArrayList<>());
   }
 
   /**
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
@@ -152,7 +152,7 @@ public class MoveBatcher {
       }
       for (int j = 0; j < sequence.size(); j++) {
         if (!sequence.get(j).mergeWith(otherSequence.get(j))) {
-          throw new IlegalStateException(
+          throw new IllegalStateException(
               "Could not merge move despite checking canMergeWith() earlier.");
         }
       }
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
@@ -144,14 +144,14 @@ public class MoveBatcher {
       final ArrayList<Move> sequence, final List<ArrayList<Move>> sequences) {
     for (final var otherSequence : sequences) {
       boolean merge = (otherSequence.size() == sequence.size());
-      for (int j = 0; merge && j < sequence.size(); j++) {
-        merge = sequence.get(j).canMergeWith(otherSequence.get(j));
+      for (int i = 0; merge && i < sequence.size(); i++) {
+        merge = sequence.get(i).canMergeWith(otherSequence.get(i));
       }
       if (!merge) {
         continue;
       }
-      for (int j = 0; j < sequence.size(); j++) {
-        if (!sequence.get(j).mergeWith(otherSequence.get(j))) {
+      for (int i = 0; i < sequence.size(); i++) {
+        if (!sequence.get(i).mergeWith(otherSequence.get(i))) {
           throw new IllegalStateException(
               "Could not merge move despite checking canMergeWith() earlier.");
         }
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
@@ -5,8 +5,9 @@ import games.strategy.engine.data.Unit;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashSet;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 /**
  * MoveBatcher implements an algorithm that batches AI moves together, reducing the number of
@@ -16,12 +17,15 @@ public class MoveBatcher {
   private static class Move {
     private final ArrayList<Unit> units;
     private final Route route;
-    private final HashSet<Unit> transportsToLoad;
+    private final HashMap<Unit, Unit> unitsToTransports;
 
-    Move(final ArrayList<Unit> units, final Route route, final HashSet<Unit> transportsToLoad) {
+    private Move(
+        final ArrayList<Unit> units,
+        final Route route,
+        final HashMap<Unit, Unit> unitsToTransports) {
       this.units = units;
       this.route = route;
-      this.transportsToLoad = transportsToLoad;
+      this.unitsToTransports = unitsToTransports;
     }
 
     Move(final ArrayList<Unit> units, final Route route) {
@@ -29,7 +33,10 @@ public class MoveBatcher {
     }
 
     Move(final Unit unit, final Route route, final Unit transportToLoad) {
-      this(mutableSingletonList(unit), route, new HashSet<Unit>(List.of(transportToLoad)));
+      this(
+          mutableSingletonList(unit),
+          route,
+          new HashMap<Unit, Unit>(Map.of(unit, transportToLoad)));
     }
 
     private static ArrayList<Unit> mutableSingletonList(final Unit unit) {
@@ -37,7 +44,7 @@ public class MoveBatcher {
     }
 
     boolean isTransportLoad() {
-      return this.transportsToLoad != null;
+      return this.unitsToTransports != null;
     }
 
     boolean canMergeWith(final Move other) {
@@ -49,7 +56,7 @@ public class MoveBatcher {
         // Merge units and transports.
         units.addAll(other.units);
         if (isTransportLoad()) {
-          transportsToLoad.addAll(other.transportsToLoad);
+          unitsToTransports.putAll(other.unitsToTransports);
         }
         return true;
       }
@@ -59,10 +66,10 @@ public class MoveBatcher {
     void addTo(
         final List<Collection<Unit>> moveUnits,
         final List<Route> moveRoutes,
-        final List<Collection<Unit>> transportsToLoad) {
+        final List<Map<Unit, Unit>> unitsToTransports) {
       moveUnits.add(units);
       moveRoutes.add(route);
-      transportsToLoad.add(this.transportsToLoad);
+      unitsToTransports.add(this.unitsToTransports);
     }
   }
 
@@ -111,18 +118,18 @@ public class MoveBatcher {
    *
    * @param moveUnits The units to move.
    * @param moveRoutes The routes to move along
-   * @param transportsToLoad The transports to be loaded.
+   * @param unitsToTransports The map from units to transports to be loaded.
    */
   public void batchAndEmit(
       final List<Collection<Unit>> moveUnits,
       final List<Route> moveRoutes,
-      final List<Collection<Unit>> transportsToLoad) {
+      final List<Map<Unit, Unit>> unitsToTransports) {
     int i = 0;
     for (final var sequence : moveSequences) {
       if (!sequence.isEmpty()) {
         mergeSequences(sequence, moveSequences.subList(i + 1, moveSequences.size()));
         for (final Move move : sequence) {
-          move.addTo(moveUnits, moveRoutes, transportsToLoad);
+          move.addTo(moveUnits, moveRoutes, unitsToTransports);
         }
       }
       i++;
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
@@ -245,7 +245,6 @@ public final class ProMoveUtils {
         final var loadedUnits = new ArrayList<Unit>();
         final var remainingUnitsToLoad = new ArrayList<Unit>();
 
-
         if (TransportTracker.isTransporting(transport)) {
           loadedUnits.addAll(amphibAttackMap.get(transport));
         } else {
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
@@ -17,6 +17,7 @@ import games.strategy.triplea.delegate.remote.IMoveDelegate;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -27,6 +28,57 @@ import org.triplea.util.Tuple;
 /** Pro AI move utilities. */
 public final class ProMoveUtils {
   private ProMoveUtils() {}
+
+  private static class Move {
+    private final ArrayList<Unit> units;
+    private final Route route;
+    private final ArrayList<Unit> transportsToLoad;
+
+    Move(final ArrayList<Unit> units, final Route route, final ArrayList<Unit> transportsToLoad) {
+      this.units = units;
+      this.route = route;
+      this.transportsToLoad = transportsToLoad;
+    }
+
+    Move(final ArrayList<Unit> units, final Route route) {
+      this(units, route, null);
+    }
+
+    Move(final Unit unit, final Route route, final Unit transportToLoad) {
+      this(mutableSingletonList(unit), route, mutableSingletonList(transportToLoad));
+    }
+
+    private static ArrayList<Unit> mutableSingletonList(final Unit unit) {
+      return new ArrayList<>(Collections.singletonList(unit));
+    }
+
+    boolean isTransportLoad() {
+      return this.transportsToLoad != null;
+    }
+
+    boolean mergeWith(final Move other) {
+      if (other != null
+          && other.isTransportLoad() == isTransportLoad()
+          && route.equals(other.route)) {
+        // Merge units and transports.
+        units.addAll(other.units);
+        if (isTransportLoad()) {
+          transportsToLoad.addAll(other.transportsToLoad);
+        }
+        return true;
+      }
+      return false;
+    }
+
+    void addTo(
+        final List<Collection<Unit>> moveUnits,
+        final List<Route> moveRoutes,
+        final List<Collection<Unit>> transportsToLoad) {
+      moveUnits.add(units);
+      moveRoutes.add(route);
+      transportsToLoad.add(this.transportsToLoad);
+    }
+  }
 
   /**
    * Calculates normal movement routes (e.g. land, air, sea attack routes, not including amphibious,
@@ -177,6 +229,7 @@ public final class ProMoveUtils {
     final GameData data = ProData.getData();
     final GameMap map = data.getMap();
 
+    final HashMap<Territory, ArrayList<Move>> movesMap = new HashMap<>();
     // Loop through all territories to attack
     for (final Territory t : attackMap.keySet()) {
 
@@ -185,10 +238,15 @@ public final class ProMoveUtils {
       for (final Unit transport : amphibAttackMap.keySet()) {
         int movesLeft = TripleAUnit.get(transport).getMovementLeft().intValue();
         Territory transportTerritory = ProData.unitTerritoryMap.get(transport);
+        ArrayList<Move> moves = movesMap.get(transportTerritory);
+        if (moves == null) {
+          moves = new ArrayList<>();
+          movesMap.put(transportTerritory, moves);
+        }
 
         // Check if units are already loaded or not
-        final List<Unit> loadedUnits = new ArrayList<>();
-        final List<Unit> remainingUnitsToLoad = new ArrayList<>();
+        final ArrayList<Unit> loadedUnits = new ArrayList<>();
+        final ArrayList<Unit> remainingUnitsToLoad = new ArrayList<>();
         if (TransportTracker.isTransporting(transport)) {
           loadedUnits.addAll(amphibAttackMap.get(transport));
         } else {
@@ -204,12 +262,9 @@ public final class ProMoveUtils {
             for (final Unit amphibUnit : remainingUnitsToLoad) {
               if (map.getDistance(transportTerritory, ProData.unitTerritoryMap.get(amphibUnit))
                   == 1) {
-                moveUnits.add(Collections.singletonList(amphibUnit));
-                transportsToLoad.add(Collections.singletonList(transport));
                 final Route route =
                     new Route(ProData.unitTerritoryMap.get(amphibUnit), transportTerritory);
-                moveRoutes.add(route);
-                unitsToRemove.add(amphibUnit);
+                moves.add(new Move(amphibUnit, route, transport));
                 loadedUnits.add(amphibUnit);
               }
             }
@@ -283,13 +338,11 @@ public final class ProMoveUtils {
               }
             }
             if (territoryToMoveTo != null) {
-              final List<Unit> unitsToMove = new ArrayList<>();
+              final ArrayList<Unit> unitsToMove = new ArrayList<>();
               unitsToMove.add(transport);
               unitsToMove.addAll(loadedUnits);
-              moveUnits.add(unitsToMove);
-              transportsToLoad.add(null);
               final Route route = new Route(transportTerritory, territoryToMoveTo);
-              moveRoutes.add(route);
+              moves.add(new Move(unitsToMove, route));
               transportTerritory = territoryToMoveTo;
             }
           }
@@ -311,11 +364,46 @@ public final class ProMoveUtils {
 
         // Unload transport
         if (!loadedUnits.isEmpty() && !t.isWater()) {
-          moveUnits.add(loadedUnits);
-          transportsToLoad.add(null);
           final Route route = new Route(transportTerritory, t);
-          moveRoutes.add(route);
+          moves.add(new Move(loadedUnits, route));
         }
+      }
+    }
+
+    // Re-order and batch the moves. That is, move all the units together, then
+    // move all the transports together, then unload them all.
+    for (final ArrayList<Move> moves : movesMap.values()) {
+      // First, add all the transport loads.
+      int i = 0;
+      for (final Move move : moves) {
+        if (move != null && move.isTransportLoad()) {
+          // Find all others with the same route to merge with.
+          mergeMoves(move, moves, i + 1);
+          move.addTo(moveUnits, moveRoutes, transportsToLoad);
+          moves.set(i, null);
+        }
+        i++;
+      }
+
+      // Then, add all the transport and unload moves, merging moves together.
+      // Since we process the moves in order, no special logic is needed to make
+      // sure transports move before unloading.
+      i = 0;
+      for (final Move move : moves) {
+        if (move != null) {
+          mergeMoves(move, moves, i + 1);
+          move.addTo(moveUnits, moveRoutes, transportsToLoad);
+        }
+        i++;
+      }
+    }
+  }
+
+  private static void mergeMoves(
+      final Move move, final ArrayList<Move> moves, final int startIndex) {
+    for (int i = startIndex; i < moves.size(); i++) {
+      if (move.mergeWith(moves.get(i))) {
+        moves.set(i, null);
       }
     }
   }
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
@@ -162,7 +162,7 @@ public final class ProMoveUtils {
    *
    * @param moveUnits Receives the unit groups to move.
    * @param moveRoutes Receives the routes for each unit group in {@code moveUnits}.
-   * @param transportsToLoad Receives the transport groups for each unit group in {@code moveUnits}.
+   * @param unitsToTransports Receives the maps from units to transports to load.
    * @param attackMap Specifies the territories to be attacked. Will be updated to reflect any
    *     transports unloading in a specific territory.
    */
@@ -170,7 +170,7 @@ public final class ProMoveUtils {
       final PlayerId player,
       final List<Collection<Unit>> moveUnits,
       final List<Route> moveRoutes,
-      final List<Collection<Unit>> transportsToLoad,
+      final List<Map<Unit, Unit>> unitsToTransports,
       final Map<Territory, ProTerritory> attackMap,
       final boolean isCombatMove) {
 
@@ -315,7 +315,7 @@ public final class ProMoveUtils {
       }
     }
 
-    moves.batchAndEmit(moveUnits, moveRoutes, transportsToLoad);
+    moves.batchAndEmit(moveUnits, moveRoutes, unitsToTransports);
   }
 
   /**
@@ -434,7 +434,7 @@ public final class ProMoveUtils {
   public static void doMove(
       final List<Collection<Unit>> moveUnits,
       final List<Route> moveRoutes,
-      final List<Collection<Unit>> transportsToLoad,
+      final List<Map<Unit, Unit>> transportsToLoad,
       final IMoveDelegate moveDel) {
 
     final GameData data = ProData.getData();
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
@@ -30,17 +30,17 @@ public final class ProMoveUtils {
   private ProMoveUtils() {}
 
   private static class Move {
-    private final ArrayList<Unit> units;
+    private final List<Unit> units;
     private final Route route;
-    private final ArrayList<Unit> transportsToLoad;
+    private final List<Unit> transportsToLoad;
 
-    Move(final ArrayList<Unit> units, final Route route, final ArrayList<Unit> transportsToLoad) {
+    Move(final List<Unit> units, final Route route, final List<Unit> transportsToLoad) {
       this.units = units;
       this.route = route;
       this.transportsToLoad = transportsToLoad;
     }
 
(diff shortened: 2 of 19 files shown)


asvitkine:
Test failures seem unrelated. Something with WebSocket?

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
```diff
@@ -178,6 +230,7 @@ public static void calculateAmphibRoutes(
 
     final GameData data = ProData.getData();
 
+    final HashMap<Territory, ArrayList<Move>> movesMap = new HashMap<>();
```
RoiEXLab: I'd appreciate if you could try to use interface types wherever possible, I.e.
`Map` instead of `HashMap` or `List` instead of `ArrayList`.
This way we could easily swap out the implementation at a later time if it no longer servers our needs

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
```diff
@@ -317,11 +370,46 @@ public static void calculateAmphibRoutes(
...
+        i++;
+      }
+    }
+  }
+
+  private static void mergeMoves(
+      final Move move, final ArrayList<Move> moves, final int startIndex) {
+    for (int i = startIndex; i < moves.size(); i++) {
+      if (move.mergeWith(moves.get(i))) {
+        moves.set(i, null);
```
RoiEXLab: Could you explain why you're explicitly nulling out the entries instead of clearing/removing them?

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
RoiEXLab: This applies to all other parts of your PR as well btw

asvitkine: Calling methods on interfaces is slower than on concrete tables. The latter is done through a vtable (so just get the pointer at the given offset), while the former requires scanning through an itable, see:
https://lukasatkinson.de/2018/interface-dispatch/#java-itables

So in local code (i.e. not in parameters and return values), I do prefer using concrete types for that reason. For actual APIs, keeping things as interfaces makes sense - since we don't want to make restrictions on callers. But when it's all local (within a function or in a private helper class), why make the code slower for very minimal benefit?

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
asvitkine: Removing would make this loop O(n^2) instead of O(n).

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
```diff
@@ -186,10 +239,15 @@ public static void calculateAmphibRoutes(
       for (final Unit transport : amphibAttackMap.keySet()) {
         int movesLeft = TripleAUnit.get(transport).getMovementLeft().intValue();
         Territory transportTerritory = ProData.unitTerritoryMap.get(transport);
+        ArrayList<Move> moves = movesMap.get(transportTerritory);
+        if (moves == null) {
+          moves = new ArrayList<>();
+          movesMap.put(transportTerritory, moves);
+        }
```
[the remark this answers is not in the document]
asvitkine: Thanks, I was looking for something like that but couldn't find it. Will do.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
RoiEXLab: Fair enough, I missed the if case that only "removes" certain entries.

You might want to document that somewhere so future developers will be aware of that

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
RoiEXLab: Interesting article you linked there 👍

However the article also notes that in practice the JIT Compiler optimizes away some of that slower lookup, so it likely won't be noticeable at all.

If you really insist on having concrete types for local variables, I'd kindly ask you to use the java 11 `var` keyword to keep the code shorter, but I personally dislike having concrete implementations even for helper methods. We ran into some issues in the past that make it sheer impossible to change the implementation type, because it is being passed so many layers deep and I really want to avoid that.

asvitkine: True that JIT should help. I like the var suggestion for locals. Will go with that.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
```diff
@@ -317,11 +370,46 @@ public static void calculateAmphibRoutes(
...
+          mergeMoves(move, moves, i + 1);
+          move.addTo(moveUnits, moveRoutes, transportsToLoad);
+        }
+        i++;
+      }
+    }
+  }
+
+  private static void mergeMoves(
+      final Move move, final ArrayList<Move> moves, final int startIndex) {
```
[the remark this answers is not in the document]
asvitkine: In this case, yes - because we need to be able to index into it. A LinkedList would not work.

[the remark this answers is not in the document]
RoiEXLab: Fair point, I'd recommend adding a Javadoc that explicitly states exactly that.

ron-murhammer:
@asvitkine If I remember correctly there is actually a reason that units are loaded 1 at a time and its because the multiple unit loading logic isn't perfect which can potentially cause the load to not load all the units into the proper transports. This is particularly important for maps that have different types of transports with varying capacities and units with different transport costs.

Without extensive testing, I'd prefer not to make this change.

asvitkine:
Another option is to have them as separate moves still, but reorder them
(using this logic) and change when we call pause() below to not call it
between these loads.

But maybe we can just fix the loading logic? Do you have a suggestion of
how to test?

ron-murhammer:
@asvitkine Yeah, I think the key is either:
1. Ensure each unit is loaded into the proper transport (I didn't review your code changes but at least as it sits today, you can't rely on the engine to properly do this 100%)
2. Update the engine loading logic to ensure it always does it 100% optimized. This is probably not feasible as this is essentially a bin packing problem so the potential combinations explodes.

As long as we ensure 1 then if we want to change the pauses or how it writes to history, I'm open to that. From a test perspective, it would be great to add some unit tests on maps with various transport capacities. For manual testing, you should be able to pick something like WaW and even manually try to load lots of units with varying transport costs into lots of varying capacity transports that you know can fit and you'll see the engine not always do it properly. I believe the engine uses a simple load largest units first algorithm which 95%+ does fairly well and is very fast.

https://en.wikipedia.org/wiki/Bin_packing_problem

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
DanVanAtta: I'd back going with the more generic type:
- it's a pseudo-java convention, not just in this project, but has been going for some time. The 'least-surprise' principle means to try and follow expectations of other programmers
- it encourages to not depend concrete types unnecessarily, good point about locals, though even then there is something to be said about having just the type or API you need.
- we should be careful of micro-optimizations. The runtime difference of tracing a few a pointers vs one is pretty certainly order nano-second. We have cases where game-data fetches are retrieved over network for clients, which is 6 to 9 orders of magnitude more significant. To that point, I really recommend the code-horror article [the infinite space between words](https://blog.codinghorror.com/the-infinite-space-between-words/), it illustrates that point well. To that extent, the JVM is quite good about optimizations that at a local level, you don't need to worry about it so much.
- last, for the most part the rest of the codebase uses as generic of types as possible, it's worthwhile to keep it consistent

> I like the var suggestion for locals 

Let's be careful to use this well. So far, AFAIK, var is best used when you have a local where the name and type match, eg: `Map< > map`, or something like: `Image image`. Names that are identical to the type can be a code smell. Regardless, using `var` can be dangerous if the type information is not clear.

asvitkine: I agree that var is not good when type info is not clear - e.g. if you're assigning the return value of some method. But if we're just creating a local collection - i.e. = new HashMap<...>(); then I think that's not a concern. Do you agree?

i.e. the suggestion is for a line like:
final HashMap<Territory, ArrayList<Move>> movesMap = new HashMap<>();

To become:

final var movesMap = new HashMap<Territory, ArrayList<Move>>();

So we're not losing any type info there.

DanVanAtta: Sounds good, for reference,  [Effective Java Item 64](https://kea.nu/files/textbooks/new/Effective%20Java%20%282017%2C%20Addison-Wesley%29.pdf)  "Item 64:  Refer to objects by their interfaces"

asvitkine:
> @asvitkine Yeah, I think the key is either:
> 
> 1. Ensure each unit is loaded into the proper transport (I didn't review your code changes but at least as it sits today, you can't rely on the engine to properly do this 100%)
> 2. Update the engine loading logic to ensure it always does it 100% optimized. This is probably not feasible as this is essentially a bin packing problem so the potential combinations explodes.
> 
> As long as we ensure 1 then if we want to change the pauses or how it writes to history, I'm open to that. From a test perspective, it would be great to add some unit tests on maps with various transport capacities. For manual testing, you should be able to pick something like WaW and even manually try to load lots of units with varying transport costs into lots of varying capacity transports that you know can fit and you'll see the engine not always do it properly. I believe the engine uses a simple load largest units first algorithm which 95%+ does fairly well and is very fast.
> 
> https://en.wikipedia.org/wiki/Bin_packing_problem

Thanks, I'll take a look. If the current algorithm for loading units onto transports is not optimal, then we should be able to just fix it. For example, here's a straight forward bin packing algorithm that guarantees optimal results:
https://security-consulting.icu/blog/2014/08/bin-packing-algorithm-java/#bin-packing-first-fit-algorithm

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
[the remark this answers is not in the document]
asvitkine: Done.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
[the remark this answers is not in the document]
asvitkine: Done.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
asvitkine: Done.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
asvitkine: OK, I'll go with var for local collections that are created with new and with interface types otherwise. I think this should be a good compromise. Done.

asvitkine:
> Thanks, I'll take a look. If the current algorithm for loading units onto transports is not optimal, then we should be able to just fix it. For example, here's a straight forward bin packing algorithm that guarantees optimal results:
> https://security-consulting.icu/blog/2014/08/bin-packing-algorithm-java/#bin-packing-first-fit-algorithm

Hmm, actually that article seems misleading. It says "The Bin Packing Problem can also be solved by an algorithm, called first fit decreasing." after presenting a brute force solution. But according to Wikipedia, "first fit decreasing" is actually an approximate algorithm and not exact. So that doesn't seem ideal.

Looking at how TripleA implements this, it looks like it already uses a similar technique (sorts by size decreasing and attempts to load), in TransportUtils.mapTransportsToLoad().

So we can try making that implementation find the optimal solution instead - or perhaps fall back to a slower algorithm that finds an optimal solution if the first one didn't work.

Another option is to pass the mapping of units to transports as part of the move to the backend, so it doesn't have to recompute it.

ron-murhammer:
@asvitkine Correct. Unless you do a true brute force and try every possible combination which explodes very quickly as bin packing is an NP Hard problem, you will just be approximating the optimal solution.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
ron-murhammer: Uh. I think adopting broad use of `var` is probably something that is a broader discussion. I don't particularly like it and generally I haven't seen it used much in Java projects. It would be better to do a refactoring just focusing on `var` rather than introducing it here.

asvitkine:
> @asvitkine Correct. Unless you do a true brute force and try every possible combination which explodes very quickly as bin packing is an NP Hard problem, you will just be approximating the optimal solution.

Wikipedia mentions some papers for fairly fast optimal solutions described in some research papers, but I couldn't find implementations we could use.

Also, we may be able to do something simpler than brute force since our problem is a bit constrained (e.g. we have only specific unit costs and specific transport types - and a lot of combinations are equivalent to others). Of course, if we want to go that route, then we definitely should have some tests.

However, I'm tempted to just go with the approach of plumbing the mapping through. I think it can be as simple as replacing the transportsToLoad collection to be a map of unit to transport. But will have to check if all call sites can provide that or not.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
DanVanAtta: Agree broader discussion, couple quick mentions:
- We've already 58 uses

```
$ find . -name "*.java" | xargs cat | grep -c "var " 
58
```
- not a lot of projects are java 10+
- usage needs to be strategic when type information is obvious, eg: `Image image = new Image()` to `var image = new Image()`

Effective Java does not yet have a recommendation on var, I'm pretty sure it would be "use var judiciously" with careful notes about typing, similar to [(P203) - Item 45](https://kea.nu/files/textbooks/new/Effective%20Java%20%282017%2C%20Addison-Wesley%29.pdf) "Use streams judiciously"

> Overusing streams makes programs hard to read and maintain

>  In  the  absence  of  explicit  types,  careful  naming  of  lambdaparameters is essential to the readability of stream pipelines.

asvitkine:
I also discovered that this current approach doesn't handle the sequence of "move transport, load unit, move transport, unload". Looking into making that work as well as adding tests.

(Also looking at adding tests for our transport packing algorithm.)

asvitkine:
Refactored this quite a bit - with a new helper class and unit tests. The new algorithm uses the concept of "move sequences" to as a simple dependency relationship between moves and ensures to not reorder moves within a sequence and merge equivalent sequences.

I added unit tests for the new class including test coverage for the case the previous code was not handling (move transport, load, move transport, unload).

This still doesn't address the concern that the game logic may not handle the loading of transports properly - I'm still planning to make changes to ensure that doesn't happen by passing a map of units to transports to the backend.

Although separately I think the current algorithm should not have any problems handling the standard A&A set up of transport of 5 capacity and units of 3 and 2 weight. I don't think there are any edge cases there given how it works, but let me know if I'm missing something. But there are cases with other transport capacities and unit sizes I think it won't handle - like loading a transport of size 5 and another of size 4 with units of sizes 4,3,2 - where the optimal would be to have 3 and 2 go to 5 and 4 go to 4, which I don't think the current code guarantees. But this should be moot when I make the desired unit to transport mapping be passed to the backend, though we can separately still improve the algorithm.

asvitkine:
I think this should be ready for another look. Thanks!

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
```diff
@@ -0,0 +1,164 @@
...
+
+    Move(final ArrayList<Unit> units, final Route route) {
+      this(units, route, null);
+    }
+
+    Move(final Unit unit, final Route route, final Unit transportToLoad) {
+      this(
+          mutableSingletonList(unit),
+          route,
+          new HashMap<Unit, Unit>(Map.of(unit, transportToLoad)));
```
RoiEXLab:
```suggestion
          new HashMap<>(Map.of(unit, transportToLoad)));
```
Pretty sure that works

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
```diff
@@ -0,0 +1,164 @@
...
+      boolean merge = (otherSequence.size() == sequence.size());
+      for (int j = 0; merge && j < sequence.size(); j++) {
+        merge = sequence.get(j).canMergeWith(otherSequence.get(j));
+      }
+      if (!merge) {
+        continue;
+      }
+      for (int j = 0; j < sequence.size(); j++) {
+        if (!sequence.get(j).mergeWith(otherSequence.get(j))) {
+          throw new RuntimeException("Unexpected");
```
RoiEXLab: Ok, this is a bad message.
I'm someone that says that there's no such thing as an expected exception in general, because exception really are an extraordinary event per definition.
So 2 options here:
1. This is supposed to be an impossible condition, in this case throwing an `IlegalStateException` would be more appropriate here. It should contain a better message why this state is illegal though.
2. This exception might happen from time to time, in this case a simple message explaining what happened would be sufficient already.

game-core/src/test/java/games/strategy/triplea/ai/pro/util/MoveBatcherTest.java
```diff
@@ -0,0 +1,130 @@
...
+  private final Route sz12ToAlgeria = new Route(sz12, algeria);
+  private final Route algeriaToSz18 = new Route(sz12, algeria);
+  private final Unit inf1 = infantry(gameData).create(americans(gameData));
+  private final Unit inf2 = infantry(gameData).create(americans(gameData));
+  private final Unit tank1 = armour(gameData).create(americans(gameData));
+  private final Unit tank2 = armour(gameData).create(americans(gameData));
+  private final Unit transport1 = transport(gameData).create(americans(gameData));
+  private final Unit transport2 = transport(gameData).create(americans(gameData));
+
+  public MoveBatcherTest() throws Exception {}
```
RoiEXLab: This constructor is redundant

game-core/src/test/java/games/strategy/triplea/ai/pro/util/MoveBatcherTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+  private static ArrayList<Unit> unitList(final Unit... units) {
+    return new ArrayList<Unit>(List.of(units));
```
RoiEXLab:
```suggestion
    return new ArrayList<>(List.of(units));
```

I assume the list must be mutable?

game-core/src/test/java/games/strategy/triplea/delegate/MoveValidatorTest.java
```diff
@@ -117,45 +117,45 @@ void testValidateMoveForRequiresUnitsToMove() throws Exception {
...
+            toMove, r, germans, Map.of(), Map.of(), false, null, twwGameData);
     assertFalse(results.isMoveValid());
 
     // Add allied owned germanRail to destination so move succeeds
     final PlayerId japan = GameDataTestUtil.japan(twwGameData);
     addTo(easternGermany, GameDataTestUtil.germanRail(twwGameData).create(1, japan));
     results =
         MoveValidator.validateMove(
-            toMove, r, germans, Collections.emptyList(), new HashMap<>(), false, null, twwGameData);
+            toMove, r, germans, Map.of(), Map.of(), false, null, twwGameData);
```
RoiEXLab: I really hope the map is never modified

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
```diff
@@ -0,0 +1,164 @@
...
+        final List<Collection<Unit>> moveUnits,
+        final List<Route> moveRoutes,
+        final List<Map<Unit, Unit>> unitsToTransports) {
+      moveUnits.add(units);
+      moveRoutes.add(route);
+      unitsToTransports.add(this.unitsToTransports);
+    }
+  }
+
+  private final ArrayList<ArrayList<Move>> moveSequences = new ArrayList<ArrayList<Move>>();
```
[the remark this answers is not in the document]
DanVanAtta: Agree, for example the constructor ` Move(final ArrayList<Unit> units, final Route route) {`,
if we change the implementation type, then we'll have to refactor the constructor, perhaps unnecessarily. It also implies that we really want this exact type when perhaps an `Iterable` or `Collection` are the API.
There are other reasons to go with the interface type:
- it's a general expectation in Java to code to interfaces, "least surprise" principle
- rest of the code base is pretty consistent to use the most generic interface type possible

ron-murhammer:
@asvitkine What testing have you done so far after the last round of changes to pass the transport map? Given the large scope of this change that involves touching UI, delegate, and AI code I think at a minimum we need:
1. Run a few all AI games on a few different maps and let run to completion to ensure no errors and they are properly loading/unloading transports
2. Do some manual testing of loading/unloading transports on a few different maps to ensure player loading/unloading from the UI works properly

asvitkine:
> @asvitkine What testing have you done so far after the last round of changes to pass the transport map? Given the large scope of this change that involves touching UI, delegate, and AI code I think at a minimum we need:
> 
> 1. Run a few all AI games on a few different maps and let run to completion to ensure no errors and they are properly loading/unloading transports
> 2. Do some manual testing of loading/unloading transports on a few different maps to ensure player loading/unloading from the UI works properly

I've done the following tests so far:
  1. Testing on Great War with US and UK AIs loading and unloading troops (verified same exact moves happen as without this change, but batched).
  2. Played on Revised - testing loading and unloading units (and checking various things like trying to load transports that had been used in combat move afterwards) and moving planes through sea zones which triggers some AC logic.
  3. Letting AIs play through on revised.

Happy to do more tests if you have suggestions.

game-core/src/test/java/games/strategy/triplea/ai/pro/util/MoveBatcherTest.java
asvitkine: It's needed because some of the initializing of the variables above can throw Exception, so compiler requires that the constructor is declared to throw it.

game-core/src/test/java/games/strategy/triplea/delegate/MoveValidatorTest.java
asvitkine: It is not, I checked.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
asvitkine: It's the first case, since earlier we call canMergeWith() on all the moves on line 151. Changed as you suggested.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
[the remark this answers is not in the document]
asvitkine: So before I wrote a unit test, I actually had the API accept List instead of ArrayList. Then, in the unit test, I passed in List.of() which compiled just fine and gave runtime errors since the implementation relies on being able to add to it.

And also to iterate by index (see mergeSequences()) - so LinkedList would also not be a good choice.

Given that using ArrayList concrete types avoids both those issues (i.e. runtime errors if someone tries to pass in List.of() that's not modifiable and performance issues if someone tries to pass in LinkedList), I would we prefer we keep ArrayList as the type here.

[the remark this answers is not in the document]
DanVanAtta: Interesting reasoning, though I don't know if I hold quite the same view. Let's say for example that someone else has a Set, or a different kind of list, to use the API presented they have to cast their object to an ArrayList, a very specific data type. That perhaps is not necessary as you don't actually need the API of an ArrayList, just the API of a List or a Collection, so you're forcing them to change their data type to suit your implementation needs.

The reveal that passing in an immutable list is interesting, I don't know if constraining the parameter is the right fix. We do have a convention to favor immutability and mark everything as final. If we use mutable data structures and then nothing is marked as final because it's all mutable, then the convention is without purpose. I think it's also then a bit striking as well that the input parameter is modified. 'Out' or 'in/out' parameters I'll suggest are best avoided, they're essentially side effects and force people using or reading usages of this code to know about the implementation. In part, we want anyone using this API, or reading code that uses it, to be unburdened with knowing the API. From a human scaling perspective, it's very limiting, to understand any code that uses this code, you really need to know "this" code, and so on, which means to modify anything at a top-level part of the abstraction, you have to be familiar with everything below it. It's a high-bar for entry, means we spend a lot of time examining underlying code before being able to make a modification.

I don't mean to be piling on, just conveying that these kind of decisions can add up to real maintenance cost in the form of having to "know" about a lot of code to be able to change just a few lines.

Another really easy way to fix the immutability issue is to avoid using an 'in' parameter, instead make of copy of the parameter, modify that, and then return the modified result.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
```diff
@@ -0,0 +1,164 @@
...
+        return true;
+      }
+      return false;
+    }
+
+    void addTo(
+        final List<Collection<Unit>> moveUnits,
+        final List<Route> moveRoutes,
+        final List<Map<Unit, Unit>> unitsToTransports) {
+      moveUnits.add(units);
```
DanVanAtta: This is jumping out to me, sorry I don't quite have the time to look at this in detail. Is there a way though where we can invert the data dependency relationship? To explain, we are making exclusive use of 'out' parameters, I think it's potentially a violation of encapsulation. Any caller to this code must know the state of the current object to then know the state of the 'out' parameters. That is potentially not going to be good because whenever reading the caller code, you need to mentally keep track, or a tally, of the state of this object to understand that code.

I'll try to take a deeper look tomorrow, just wanted to get this conversation started and to highlight this design area for further review.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
```diff
@@ -0,0 +1,162 @@
...
+    boolean isTransportLoad() {
+      return this.unitsToTransports != null;
+    }
+
+    boolean canMergeWith(final Move other) {
+      return other.isTransportLoad() == isTransportLoad() && route.equals(other.route);
+    }
+
+    boolean mergeWith(final Move other) {
+      if (canMergeWith(other)) {
```
DanVanAtta: Is there a clean way to split the "canMerge" concern from the "merge" concern?

It seems like it could make sense for someone to execute this by itself, eg:
```mergetWith(other)``` and ignore the return value. In which case we might have a silent no-op. On the other hand, the usage in an if statement has side effects beyond returning a 'true/false'

File: game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java [hunk 13]
```diff
@@ -0,0 +1,162 @@
...
+
+    Move(final Unit unit, final Route route, final Unit transportToLoad) {
+      this(mutableSingletonList(unit), route, new HashMap<>(Map.of(unit, transportToLoad)));
+    }
+
+    private static ArrayList<Unit> mutableSingletonList(final Unit unit) {
+      return new ArrayList<>(Collections.singletonList(unit));
+    }
+
+    boolean isTransportLoad() {
```
DanVanAtta: nit: if possible, please mark methods/variables with the most restrictive scope, eg: private in this case. It helps readers, I did wonder where/how this was being used outside of this class to then realize it's internal only.

Beyond the nit, can we represent an unloaded transport with an empty collection rather than null?

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
asvitkine: This is just for interfacing with the existing way moves are handled, see ProMoveUtils.java and IMoveDelegate.java.

I agree that this is messy and I think we should just refactor things so there's a single Move object that's used - instead of these weird parallel lists. But this change is already large enough and it's out of scope for it.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
```diff
@@ -0,0 +1,164 @@
...
+   * @param unit The unit to move.
+   * @param route The route to move on.
+   * @param transport The transport to load.
+   */
+  public void addTransportLoad(final Unit unit, final Route route, final Unit transport) {
+    addMove(new Move(unit, route, transport));
+  }
+
+  /**
+   * Adds a move for the specified units. Ownership of the params is transferred to this object.
```
DanVanAtta: Would you clarify what is meant by: " Ownership of the params is transferred to this object." ?

game-core/src/test/java/games/strategy/triplea/ai/pro/util/MoveBatcherTest.java
```diff
@@ -0,0 +1,130 @@
...
+    final List<Collection<Unit>> moveUnits = new ArrayList<>();
+    final List<Route> moveRoutes = new ArrayList<>();
+    final List<Map<Unit, Unit>> unitsToTransports = new ArrayList<>();
+    moves.batchAndEmit(moveUnits, moveRoutes, unitsToTransports);
+
+    assertEquals(3, moveUnits.size());
+    assertEquals(3, moveRoutes.size());
+    assertEquals(3, unitsToTransports.size());
+
+    assertEquals(unitList(inf1, tank1, tank2, inf2), moveUnits.get(0));
```
DanVanAtta: If this assertion were to fail, would it be clear to someone why it was expected to pass in the first place? Could that be answered with minimal effort? If a change were made, or a series of changes made, and all assertions in this test were to fail, how much time would it take for someone to determine why the assertions were valid in the first place and hence the necessary modification?

To reduce this cost, as a first pass, I recommend adding a comment to each assertion, for example `assertEquals("these units were unloaded together, should be grouped together", unitLIst(...), moveUnits.get(0))`

game-core/src/test/java/games/strategy/triplea/ai/pro/util/MoveBatcherTest.java
```diff
@@ -0,0 +1,130 @@
...
+    assertEquals(null, unitsToTransports.get(0));
+
+    assertEquals(unitList(tank1, inf1, inf2), moveUnits.get(1));
+    assertEquals(brazilToSz18, moveRoutes.get(1));
+    final var expected = Map.of(tank1, transport1, inf1, transport2, inf2, transport2);
+    assertTrue(Maps.difference(expected, unitsToTransports.get(1)).areEqual());
+
+    assertEquals(unitList(transport1, tank1, transport2, inf1, inf2), moveUnits.get(2));
+    assertEquals(sz18ToSz12, moveRoutes.get(2));
+    assertEquals(null, unitsToTransports.get(2));
```
DanVanAtta: Out of curiosity, why are null elements stored? Wouldn't it be just as well to not even store the element? EG: is an index out of bounds perhaps expected here?

game-core/src/test/java/games/strategy/triplea/delegate/MoveValidatorTest.java
```diff
@@ -117,45 +117,45 @@ void testValidateMoveForRequiresUnitsToMove() throws Exception {
...
     assertTrue(results.isMoveValid());
 
     // Remove germanRail from destination so move fails
     GameDataTestUtil.removeFrom(easternGermany, germanRail);
     results =
         MoveValidator.validateMove(
-            toMove, r, germans, Collections.emptyList(), new HashMap<>(), false, null, twwGameData);
+            toMove, r, germans, Map.of(), Map.of(), false, null, twwGameData);
```
DanVanAtta: We should probably confirm this with a convention, though the `Map.of()` seems a bit odd when there is `Collections.emptyMap()`. I feel like the var-args variant should be used when there is at least one parameter. For example, it would also be a bit odd and seemingly indirect to use `Map.of(new Unit[0[)` (which is equivalent to `Map.of()`), when `Collections.emptyMap()` is the same thing and more direct (and there is no unused parameter that is passed)

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
DanVanAtta: Okay, that sounds fair enough. Would you mind commenting the code to document that? "TODO: #5416 follow-up Refactor to avoid the parallel lists, the side effects are to support existing code" (super-paraphrased).

I don't mind us doing what's needed to get things done, it's not all going to be pretty, but we can perhaps mitigate it a bit and have pity on the next reader of the code with some call-out comments of where we know things are not ideal and would ideally be fixed up later.

DanVanAtta:
I'll step back and let @RoiEXLab and @ron-murhammer finish the review, too many reviews is not always helpful. Certainly happy to give suggestions, do feel free to ask.

Otherwise, this seems like a nice update in a number of ways, faster, better AI - also it seems like an update to a relatively difficult part of the code as well. I jumped in for concern that we nonetheless try to keep the maintenance costs as low as we can, and I was curious about the design decisions taken.

game-core/src/test/java/games/strategy/triplea/ai/pro/util/MoveBatcherTest.java
asvitkine: It's the same messy parallel list business.

unitsToTransports.get(2) corresponds to moveRoutes.get(2) to moveUnits.get(2).

game-core/src/main/java/games/strategy/triplea/delegate/data/MoveDescription.java
```diff
@@ -50,11 +46,11 @@ public String toString() {
...
 
-  public Collection<Unit> getTransportsThatCanBeLoaded() {
-    if (transportsThatCanBeLoaded == null) {
-      return Collections.emptyList();
+  public Map<Unit, Unit> getUnitsToTransports() {
+    if (unitsToTransports == null) {
+      return Map.of();
     }
-    return transportsThatCanBeLoaded;
+    return unitsToTransports;
```
RoiEXLab: What about preventing `unitsToTransports` to be null in the first place?
(This check could be done in the constructor already)

game-core/src/main/java/games/strategy/triplea/delegate/data/MoveDescription.java
```diff
@@ -13,17 +12,17 @@
...
-  private final Collection<Unit> transportsThatCanBeLoaded;
+  private final Map<Unit, Unit> unitsToTransports;
   private final Map<Unit, Collection<Unit>> dependentUnits;
 
   public MoveDescription(
       final Collection<Unit> units,
       final Route route,
-      final Collection<Unit> transportsThatCanBeLoaded,
+      final Map<Unit, Unit> unitsToTransports,
       final Map<Unit, Collection<Unit>> dependentUnits) {
```
RoiEXLab: Not your fault, but we try to annotate nullable fields and parameters with `@Nullable` in new code/refactored code, so it would be nice if you could do the same when you explicity pass null somewhere

game-core/src/main/java/games/strategy/triplea/delegate/MoveValidator.java
```diff
@@ -1171,8 +1168,8 @@ private static MoveValidationResult validateTransport(
     }
     // If there are non-sea transports return
     final boolean seaOrNoTransportsPresent =
```
RoiEXLab: The variable name and the comment doesn't match the behaviour🤔 
It should probably be called `seaTransportPresent`.
In any case the expression can be simplified:
- `!seaOrNoTransportsPresent`
- == `!(unitsToTransports.isEmpty() || unitsToTransports.values().stream().anyMatch(Matches.unitIsSea().and(Matches.unitCanTransport())))`
- == `!unitsToTransports.isEmpty() && unitsToTransports.values().stream().noneMatch(Matches.unitIsSea().and(Matches.unitCanTransport()))`

game-core/src/test/java/games/strategy/triplea/delegate/MoveValidatorTest.java
asvitkine: I agree that it would be good to make a decision on this and then apply it consistently going forward.

Here's why I think Map.of() is better than Collections.emptyMap() and similar for List.of() vs Collections.emptyList():

- It's shorter to write. This makes a difference for code readability especially when used in long parameter lists - often allowing the code to be formatted much more concisely.
- Map.of() is already useful for cases where the size is non-0. If we'll already be using it for those cases, it's better to remember just one API rather than multiple - i.e. no need to ever use Collections.emptyList().

Let me know if this makes sense - if so, I suggest documenting this somewhere - and maybe as a follow up, we can do a bulk replace of existing Collections.emptyList() and emptyMap() uses to make everything consistent.

game-core/src/test/java/games/strategy/triplea/ai/pro/util/MoveBatcherTest.java
asvitkine: Moving gameData to this constructor would also require moving everything else - since they reference gameData.

This would double the number of lines required, so I prefer not to do that. Let me know if you disagree.

RoiEXLab: Ah I see, thanks for the clarification.
No, keeping everything as-is seems cleaner to me.
Not sure if it's worth adding a comment explaining the java mechanism, doesn't matter too much

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
asvitkine: Also, it's possible this all becomes moot after the refactoring to not use parallel lists.

Because after that refactoring, we may be passing Move objects around instead of these list params, which I either will allow us to be both efficient and nice as an API (i.e. whereas ArrayList doesn't have a swap() method that can be done efficiently, if we have a wrapper object like Move(), we may be able to do this - or if we make Move objects immutable, then this is also moot.)

game-core/src/test/java/games/strategy/triplea/delegate/MoveValidatorTest.java
DanVanAtta: @asvitkine this is a matter of what one is used to, which is okay, we can set expectations.

I'm curious, in your opinion, doesn't `Map.of()` seem a bit indirect though? It's another way of saying "emptyMap", hence the indirection. In other words, instead of saying "empty map", we're saying "map with zero elements", it's the same, but just slightly more round-about IMO. Is the analogy to "size() == 0" vs "empty()" a reasonable analogy, or do you view that differently?

Either way, we don't have a solid convention one way or another. Consistency argument would suggest to use 'empty*", but I don't think it's worthwhile to hold things up. We can extract this line of discussion to: https://github.com/triplea-game/triplea/issues/5489

DanVanAtta:
I'd agree on the generic vs concrete types, I'd probably prefer even to see `Collection` as the interface type unless we truly need the full `List` API. Arguing with effective java is difficult as well: Effective Java Item 52: http://thefinestartist.com/effective-java/52

For now, we can move past that concern and put the convention conversation to: https://github.com/triplea-game/triplea/issues/5489

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
```diff
@@ -185,10 +187,12 @@ public static void calculateAmphibRoutes(
       for (final Unit transport : amphibAttackMap.keySet()) {
         int movesLeft = TripleAUnit.get(transport).getMovementLeft().intValue();
         Territory transportTerritory = ProData.unitTerritoryMap.get(transport);
+        moves.newSequence();
 
         // Check if units are already loaded or not
-        final List<Unit> loadedUnits = new ArrayList<>();
-        final List<Unit> remainingUnitsToLoad = new ArrayList<>();
+        final var loadedUnits = new ArrayList<Unit>();
```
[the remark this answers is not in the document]
ron-murhammer: I believe its needed. You need to specify `Unit` either on the left or right side. Otherwise it would initialize to an ArrayList of objects.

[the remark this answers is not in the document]
DanVanAtta: Sure enough, I was thinking you'd get a 'type cannot be inferred'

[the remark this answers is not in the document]
ron-murhammer: I'm fairly sure it will let you do it though I would think IDEs/checkstyle/etc might give you a warning. Its one of the reasons I don't really like `var` in a lot cases as its kind of dangerous if you don't know what you are doing.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
```diff
@@ -0,0 +1,161 @@
...
+
+    Move(final ArrayList<Unit> units, final Route route) {
+      this(units, route, null);
+    }
+
+    Move(final Unit unit, final Route route, final Unit transportToLoad) {
+      this(new ArrayList<>(List.of(unit)), route, new HashMap<>(Map.of(unit, transportToLoad)));
+    }
+
+    private boolean isTransportLoad() {
```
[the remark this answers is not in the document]
asvitkine: It means - "is the type of this move a transport load"? i.e. as part of this move, units are being loaded on a transport.

So none of your above names apply.

We can name it isTransportLoadMove() - but this a method of a class called Move, so the Move at the end is redundant - it's like if List.isEmpty() was called List.isEmptyList() - sure it's correct, but kind of redundant.

Perhaps adding Javadoc would be sufficient for you here?

Other options are - isMoveTypeTransportLoad() - or isLoadingUnits()... Not sure I like them better than current name if you're OK with just Javadoc.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
```diff
@@ -0,0 +1,161 @@
...
+        final ArrayList<Unit> units,
+        final Route route,
+        final @Nullable HashMap<Unit, Unit> unitsToTransports) {
+      this.units = units;
+      this.route = route;
+      this.unitsToTransports = unitsToTransports;
+    }
+
+    Move(final ArrayList<Unit> units, final Route route) {
+      this(units, route, null);
```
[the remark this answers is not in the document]
asvitkine: To be clear, null entries is how the three parallel lists in ProMoveUtils are handling transports already. So changing it here would require a bigger refactoring. But the right bigger refactoring is to get rid of the three parallel lists in the first place. We agreed that that should be done in a separate follow-up PR.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
```diff
@@ -0,0 +1,161 @@
...
+  public void newSequence() {
+    moveSequences.add(new ArrayList<>());
+  }
+
+  /**
+   * Adds a transport load move.
+   *
+   * @param unit The unit to move.
+   * @param route The route to move on.
+   * @param transport The transport to load.
```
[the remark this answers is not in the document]
asvitkine: The transport is not moving. It's stationary at the destination of the route and being loaded by the unit. I can clarify in the method doc.

I don't think we want to do any abstraction until we do the refactoring to get rid of the parallel lists.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
```diff
@@ -0,0 +1,161 @@
...
+   */
+  public void addTransportLoad(final Unit unit, final Route route, final Unit transport) {
+    addMove(new Move(unit, route, transport));
+  }
+
+  /**
+   * Adds a move for the specified units. The units list must not be changed after adding the move.
```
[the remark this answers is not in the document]
asvitkine: Did you see my lengthy reply to your earlier comment about this? It's in the thread where you mentioned the previous version of the comment about "taking ownership".

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
```diff
@@ -0,0 +1,161 @@
...
+  private void addMove(final Move newMove) {
+    final ArrayList<Move> sequence = moveSequences.get(moveSequences.size() - 1);
+    if (!sequence.isEmpty() && sequence.get(sequence.size() - 1).mergeWith(newMove)) {
+      return;
+    }
+    sequence.add(newMove);
+  }
+
+  /**
+   * Emits the accumulated moves to the passed output params.
```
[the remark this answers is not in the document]
asvitkine: I think it should be:

Collection<Move> moves = moveBatcher.batchAndEmit();

But this related to the previously discussed refactoring about making IMoveDelegate and ProMoveUtils operate on Move objects instead of three parallel lists.

Doing so would be an even bigger change than this PR. Can we do it in the next PR like agreed? I feel like all your comments are around this - which is an existing bad pattern in the current code that I'm not changing, just making my class work with it.

Like, let's not put effort into polishing this existing ugly mess if the plan is to refactor it. Because that's just wasted effort.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
asvitkine: But this means if someone calls mergeWith() it can result in illegal moves being produced. Are you sure this is what you want? It seems bad for the API to allow this to happen.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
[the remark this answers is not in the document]
DanVanAtta: > I don't think we want to do any abstraction until we do the refactoring to get rid of the parallel lists.

I'm not sure I agree with this, seems to be circular reasoning:
- the abstraction of the current code is too low to allow for abstractions to be created until the code is refactored to have abstractions.

Generally any update, new code, should be itself clean. I think I'm left confused, the unit is moving, we have the route the unit moves along, and we have a transport that is not moving. Is the transport at the terminus of the route? Why is the transport involved in this scenario? 

(as far any CR comment, please assume anyone else reading the code would have the same confusion/issues, we should be looking for a code update and/or javadoc to help clear it up, which we can discuss before commit, but ultimately the best outcome is a code update of course)

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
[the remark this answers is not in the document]
DanVanAtta: Sounds reasonable. I can't read your mind and what you intend for a future change, it's almost just as well to assume there will be no further updates, and even then it's not certain what those updates are necessarily going to be. Adding various TODO statements on what will be changed could be helpful to communicate this and to help guide future changes.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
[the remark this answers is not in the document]
DanVanAtta: I'm curious how have you seen it being dangerous? Type cannot be inferred is a compiler error, it's for when the non-generic type (IE: List or Map) cannot be inferred. For generics, defaulting up to an object is likely going to also run into an error later unless you are actually pulling objects from the collections. On another note, this usage I think is okay, the scope of the variable is too large anyways, where the variable is used, you can't read the declaration anyways, so it does not matter either way.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
DanVanAtta: Nice APIs are worth quite a bit, impossible to be used incorrectly, means no bugs and fast development as you don't have to study any API you use, you can just use it. For maintenance, it also means modularity and isolation, to reason about the state of this object, you have to know about the callers, and from the caller perspective, you have to know about the implementation of this object and its callers.

With respect to runtime cost, anything order nano-second is too fast to worry about, premature optimization. I ran an experiment once to check the differing runtimes over different JDKs, from 1.2 to 8. In Java 8, new object creation can be faster than calling a no-op method, things have gotten really fast and well optimized.  For appreciation of how fast a ns is, I very much like this blog post: https://blog.codinghorror.com/the-infinite-space-between-words/

This is Java after-all, IMO avoiding in/out or out parameters is good, copying in parameters and returning the value is a standard'ish approach, nobody is going to fault you for the list copy.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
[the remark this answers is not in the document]
DanVanAtta: Maybe `isTransporting`?
A private method probably should be self-explanatory through it's name, I feel like javadoc'ing a one-liner private method means we have not come up with a good name/abstraction.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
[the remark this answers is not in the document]
DanVanAtta: Okay, I like how you annotated the nullables later so at least we have that safe-guard. I'm sure there is a lot of refactoring to be done. I suppose it's of help if the locations for update are marked so that it is communicated for review and other developers if they happen to read the code, and so we don't forget.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
```diff
@@ -0,0 +1,162 @@
...
+   * @param sequences The sequences to try to merge into the sequence.
+   */
+  private static void mergeSequences(
+      final ArrayList<Move> sequence, final List<ArrayList<Move>> sequences) {
+    for (final var otherSequence : sequences) {
+      boolean merge = (otherSequence.size() == sequence.size());
+      for (int i = 0; merge && i < sequence.size(); i++) {
+        merge = sequence.get(i).canMergeWith(otherSequence.get(i));
+      }
+      if (!merge) {
```
DanVanAtta: Question looking at this, `merge` has the value of the last sequence element, it ignores if any intermediary values were merged. Is that really the intent, to check if the last merge was success?

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
DanVanAtta: That sounds like it would be even more brittle.

Making 'mergeWith' functional would be a nice way to go, maybe out of scope/difficult given how much we rely on mutability.

We could make `mergeWith` do a precondition check that `canMerge` is true, from a usage point of view, maybe it's a wash. Reluctantly, we may need for the mutability issues to be fixed before we can fix this one.

You would agree we have a SRP violation? It does seem we have two things wrapped in one.

As a functional variant that does not blow up with precondition, an `Optional<Move>` being returned could be a nice API to indicate we were able to create a valid merge.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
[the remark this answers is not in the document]
DanVanAtta: I missed it, sorry. My reaction is roughly the same though, to avoid mutability of parameters.

DanVanAtta:
@asvitkine the last question I have is perhaps the only thing outstanding, I can understand the existing code is not ideal. 

Do you plan to follow up with some refactoring passes to try and fix some of the issues we've been going back and forth on? If that follow-up is going to be as soon as this is merged, then I'm inclined to get that done. If you won't have time to follow-up soon, or not sure when, then it's probably a good move to pick a "GREP" token for your TODOs so they all have the same prefix, perhaps: `TODO: @asvitkine `, and mark the places that would be refactored so you can come back to it later and in the meantime other dev's can know when reading that there is work in progress.

ron-murhammer:
@DanVanAtta I'm fine putting some sort of token in the TODOs but I'd prefer it to be like a keyword or issue ID not any names.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
asvitkine: The condition on line 147 will break out of the loop as soon as merge is false.

asvitkine:
> @asvitkine the last question I have is perhaps the only thing outstanding, I can understand the existing code is not ideal.
> 
> Do you plan to follow up with some refactoring passes to try and fix some of the issues we've been going back and forth on? If that follow-up is going to be as soon as this is merged, then I'm inclined to get that done. If you won't have time to follow-up soon, or not sure when, then it's probably a good move to pick a "GREP" token for your TODOs so they all have the same prefix, perhaps: `TODO: @asvitkine `, and mark the places that would be refactored so you can come back to it later and in the meantime other dev's can know when reading that there is work in progress.

Ah, sorry if that wasn't clear. Yes, I was planning to work on that refactoring right after this is submitted. Hence why I was frustrated that we kept discussing the code that I was planning to refactor right after to address the issues - I just didn't want to grow this PR even more with those changes here.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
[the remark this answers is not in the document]
asvitkine: Just to be clear, the existing code works on the three parallel lists. So new code cannot avoid working on those three parallel lists if it needs to work with the existing code. So either we refactor the whole thing in this PR - which I'm proposing not to do, so as to not make this PR even bigger - or we leave the TODO to do it in the next PR.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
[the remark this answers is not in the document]
asvitkine: isTransporting() sounds like it's a "move of units together with transport" not a move to load units onto a transport. I assume the confusion is with the word "load" since it can mean different things. Do we want different terminology? IsEmbarkation()? But then we're introducing new terminology which might be more confusing.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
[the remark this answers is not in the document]
asvitkine: Ok, let's just copy the lists for now. I lost the energy to care about this aspect of performance. If we must live with less efficient code to get this in, then so be it. It's not as bad as other things happening in TripleA that I see when profiling it, so probably better to save my energy on fixing those instead.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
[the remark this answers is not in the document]
DanVanAtta: `unitsAreLoadingOntoTransport()`?
Failing that, yeah, a quick javadoc just in case.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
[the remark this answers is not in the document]
DanVanAtta: Sorry to have worn you down, API usability/brittleness is in part how we got to a bad place. I do rate that as a pretty big concern. I'm happy seeing this as part of a follow-up

DanVanAtta:
@ron-murhammer sounds fine, PR number probably would be a good one for this kind of followup.

PR #5417, opened 2019-11-04
Add acceptance/integration tests for websocket and websocket chat

DanVanAtta:
1. Add a test to verify websocket server can be launched and we can
   use an out-of-the-box client to connect to it. Verify basic
   actions like send text and closing the connection.
2. Add an acceptance test for chat, connect to the chat websocket
   and simulate two chatters connecting and exercise all chat
   functionality (eg: set status, slap, send messages).

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[x] Other: tests  

## Testing

[x] No manual testing done
[] Manually tested

## Additional Review Notes
- would like review to mostly focus on how easy it is to understand the tests

### DIFF
--- a/integration-testing/build.gradle
+++ b/integration-testing/build.gradle
@@ -7,6 +7,7 @@ configurations {
 }
 
 dependencies {
+    // https://mvnrepository.com/artifact/org.java-websocket/Java-WebSocket
     runtimeOnly "org.postgresql:postgresql:$postgresqlVersion"
     testImplementation "com.github.database-rider:rider-junit5:$databaseRiderVersion"
     testImplementation "com.sun.mail:jakarta.mail:$jakartaMailVersion"
@@ -25,6 +26,7 @@ dependencies {
     testImplementation "org.jdbi:jdbi3-core:$jdbiVersion"
     testImplementation "org.jdbi:jdbi3-sqlobject:$jdbiVersion"
     testImplementation "org.mindrot:jbcrypt:$jbcryptVersion"
+    testImplementation "org.java-websocket:Java-WebSocket:$javaWebsocketVersion"
     testImplementation project(":test-common")
     testImplementation project(':domain-data')
     testImplementation project(':game-core')
--- a/integration-testing/build.gradle
+++ b/integration-testing/build.gradle
@@ -21,6 +21,7 @@ dependencies {
     testImplementation "io.github.openfeign:feign-gson:$openFeignVersion"
     testImplementation "javax.activation:activation:$javaxActivationVersion"
     testImplementation "javax.xml.bind:jaxb-api:$jaxbApiVersion"
+    testImplementation "org.awaitility:awaitility:$awaitilityVersion"
     testImplementation "org.jdbi:jdbi3-core:$jdbiVersion"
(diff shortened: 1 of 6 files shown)


RoiEXLab:
There's a build failure

integration-testing/src/test/java/org/triplea/http/server/LobbyWebsocketClientTest.java
```diff
@@ -0,0 +1,53 @@
...
+        };
+
+    assertThat(client.isOpen(), is(false));
+    client.connect();
+
+    await().pollDelay(10, TimeUnit.MILLISECONDS).atMost(1, TimeUnit.SECONDS).until(client::isOpen);
+    client.send("sending! Just to make sure there are no exception here.");
+
+    // small wait to process any responses
+    Thread.sleep(10);
```
RoiEXLab: Is this really needed?

integration-testing/src/test/java/org/triplea/http/server/LobbyWebsocketClientTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+    client.close();
+    await()
+        .pollDelay(10, TimeUnit.MILLISECONDS)
+        .atMost(100, TimeUnit.MILLISECONDS)
+        .until(() -> !client.isOpen());
```
RoiEXLab: I mean I get that we're just testing a library here, but in case any those tests ever fail, what would be do about it?
Probably nothing, because library change is too much and it worked in the past, so there must be a workaround

integration-testing/src/test/java/org/triplea/http/server/LobbyWebsocketClientTest.java
DanVanAtta: if there are any errors on send, because send is async, unless we pause, we might not even have yet sent the message to server.

integration-testing/src/test/java/org/triplea/http/server/LobbyWebsocketClientTest.java
DanVanAtta: This was one of the very first tests written before any websocket code landed, probably good to keep that context in mind. If for any reason the websocket connection does not open, is not configured right, or any of the flow with the wiring for message service is broken, we're verifying that level of connectivity. This is in another sense an integration test that does not rely on the `GenericWebsocketClient`, so if we see any issues we know there is something wrong on the server side. Hence, it's not really a library test, we are firing up a server, opening ports, and using the most basic client to connect to it. Perhaps the name `ClientTest` is misleading, this is a test using a `WebsocketClient` connecting to the lobby.

integration-testing/build.gradle
```diff
@@ -7,6 +7,7 @@ configurations {
 }
 
 dependencies {
+    // https://mvnrepository.com/artifact/org.java-websocket/Java-WebSocket
```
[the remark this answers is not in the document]
DanVanAtta: Left over from copy/pasting the import in from mvnrepository, removed: 5262423aa

PR #5421, opened 2019-11-04
Deprecate PlayerNameValidation methods & reduce dependency on magic l…

DanVanAtta:
With game listings moved to Http-Server, the updated system design
will not rely on naming conventions to identify automated hosts.

With this in mind, validation between server side and client side
becomes unified and can be the same logic.

The validation logic between PlayerName and PlayerNameValidation
then also becomes identical, we should stop using PlayerNameValidation
in favor of PlayerName. PlayerName can be accessed from multiple
projects, hence that is the favored copy.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[x] No manual testing done
[] Manually tested

### DIFF
--- a/game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
+++ b/game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
@@ -5,7 +5,6 @@ import games.strategy.engine.lobby.PlayerNameValidation;
 import java.util.Collection;
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
-import org.triplea.lobby.common.LobbyConstants;
 
 /** Utility class that will assign a name to a newly logging in player. */
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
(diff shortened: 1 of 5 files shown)


game-core/src/main/java/games/strategy/net/PlayerNameAssigner.java
```diff
@@ -22,16 +21,14 @@
...
   public static String assignName(
       final String desiredName, final String mac, final Collection<String> loggedInNames) {
-    Preconditions.checkArgument(PlayerNameValidation.serverSideValidate(desiredName) == null);
+    Preconditions.checkArgument(PlayerNameValidation.validate(desiredName) == null);
     Preconditions.checkNotNull(mac);
     Preconditions.checkNotNull(loggedInNames);
 
     String currentName = desiredName;
 
-    if (!isBotName(desiredName)) {
```
RoiEXLab: Does this mean we can name our bots now whatever we like? 😱

Issue #5233, opened 2019-09-17
NPE in Unit Scroller


ron-murhammer:
**Describe the bug**
- Running latest pre-release with an all AI game on revised, I got the following NPE which appears to have to do with a potential null player:
```
SEVERE: null
java.lang.NullPointerException
	at games.strategy.triplea.attachments.TechAttachment.get(TechAttachment.java:63)
	at games.strategy.triplea.delegate.TechTracker.hasRocket(TechTracker.java:117)
	at games.strategy.triplea.image.UnitImageFactory.getBaseImageName(UnitImageFactory.java:239)
	at games.strategy.triplea.image.UnitImageFactory.getImageDimensions(UnitImageFactory.java:304)
	at games.strategy.triplea.ui.unit.scroller.AvatarPanelFactory.createUnitStackImage(AvatarPanelFactory.java:79)
	at games.strategy.triplea.ui.unit.scroller.AvatarPanelFactory.buildPanel(AvatarPanelFactory.java:51)
	at games.strategy.triplea.ui.unit.scroller.UnitScroller.lambda$6(UnitScroller.java:167)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:313)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:770)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:721)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:715)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:85)
	at java.desktop/java.awt.EventQueue.dispatchEvent(EventQueue.java:740)
	at java.desktop/java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:203)
	at java.desktop/java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:124)
	at java.desktop/java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:113)
	at java.desktop/java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:109)
	at java.desktop/java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:101)
	at java.desktop/java.awt.EventDispatchThread.run(EventDispatchThread.java:90)
[...]

PR #5432, opened 2019-11-05
Add null check as player can be null during some phases

ron-murhammer:
Addresses #5233

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[x] Manually tested

Tested scenarios outlined in bug report and now don't see any errors.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/unit/scroller/UnitScroller.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/unit/scroller/UnitScroller.java
@@ -69,7 +69,7 @@ public class UnitScroller {
       "Press 'L' key or click this button to toggle flags";
 
   private Collection<Unit> skippedUnits = new HashSet<>();
-  private Collection<Unit> sleepingUnits = new HashSet<>();
+  private final Collection<Unit> sleepingUnits = new HashSet<>();
   private Territory lastFocusedTerritory;
 
   private final GameData gameData;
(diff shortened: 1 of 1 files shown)


ron-murhammer:
@DanVanAtta Please take a look at the fix here. There are a bunch of places you could add a null check but this seemed like the easiest given the way its structured.

game-core/src/main/java/games/strategy/triplea/ui/unit/scroller/UnitScroller.java
```diff
@@ -163,8 +163,10 @@ private void drawUnitAvatarPane(final Territory t) {
     SwingUtilities.invokeLater(
         () -> {
           selectUnitImagePanel.removeAll();
-          selectUnitImagePanel.add(
-              avatarPanelFactory.buildPanel(matchedUnits, currentPlayerSupplier.get()));
+          if (currentPlayerSupplier.get() != null) {
+            selectUnitImagePanel.add(
+                avatarPanelFactory.buildPanel(matchedUnits, currentPlayerSupplier.get()));
```
DanVanAtta: The logic might be more unified if we had a single list with an aggregated data type, and rendering was a no-op only if the given list was empty, rather than a no-op if either player is null, or list is empty. This seems fine though.

Issue #5286, opened 2019-09-27
KeyBindings Fire When Chatting


DanVanAtta:
**Describe the bug**
Keybindings are firing even when typing in chat

**To Reproduce**
- Host a local network game
- Type the word "fun", notice the 'f' and 'n' keystrokes activate key bindings

**Expected behavior**
Key bindings should not fire when typing in chat

PR #5455, opened 2019-11-09
Fix KeyBindings firing while chatting

DanVanAtta:
While a user is typing in a text component, we do not want keybindings
to fire. For example, pressing space key skips current unit, typing
a few words would skip a few units.

To fix this we can set a keyboard focus listener to fire when keyboard
focus changes. We then check if focus is on a text component, if so
we disable key binding actions. When focus is off of the text component,
we re-enable the key binding actions.

Addresses: https://github.com/triplea-game/triplea/issues/5286

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Bug fix:  https://github.com/triplea-game/triplea/issues/5286 
[] Other:   

## Testing

[] No manual testing done
[x] Manually tested


RoiEXLab:
Some merge conflicts showed up

PR #5460, opened 2019-11-12
Use URI for server instead of host+port combination

DanVanAtta:
This update changes the triplet "lobby.host", "lobby.port"
and "lobby.https.port" to be a single property "lobby.uri".
This drops configuration to connect to the old lobby and
allows us to define whether we will use HTTPS or HTTP
by configuration rather than hardcoding the protocol. This
is particularly useful as local development will likely continue
to be direct connection to (http) dropwizard while deployed contexts
will be to nginx fronting a dropwizard server over https.

** Configuration Updates and Consolidation of live versions file

'latest_version.properties' and 'lobby_server.yaml' are being combined
into one file. There are a few semantic changes to this configuration
file on top of the combined format:
- engine will treat version numbers as a "min-supported" instead of
  looking for exact version matches.
- error_message concept is replaced with 'inactive' flag. We had an
  error_message field before to allow us to notify users when they
  are connecting to an out-of-date lobby and that they should upgrade.
  The 'inactive' flag will instead trigger client side out-of-date
  messaging.

** Lobby Override URI Test-Setting Update

The engine setting for override URI is refreshed with this change.
The setting used to be a series of text field boxes and is replaced
with a radio button to select between a 'remote lobby' or 'local lobby'.
If the 'local lobby' radio button is selected, then a localhost URI
is auto-filled and used. Otherwise the 'remote lobby' has a free-form
text field for custom override values.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[x] Manually tested

- Verified could host from a local lobby.
- The live_servers.yaml file is not live and yields an error in the background, we can do live testing of the updated configuration mechanism post-merge.

## Additional Review Notes
Please check for:
 - clarity of documentation
 - new mechanism of downloading server configs (EG: the new config file makes sense)
 - test case completeness

Also of note:
- `ContentReader` is refactored to use a `ContentDownloader` to allow for direct access to an `InputStream` to read remote http resources. The direct access is particularly useful so that we can swap out the source of the `InputStream` in test contexts.
- We've a number of new tests for the new `live_servers.yml` file, an integration test to read a sample file, and there is a validation test to verify the 'current' `live_servers.yml`.


game-core/src/main/java/org/triplea/live/servers/ServerYamlParser.java
```diff
@@ -0,0 +1,58 @@
...
+      throw new IllegalArgumentException("Invalid yaml format");
+    }
+
+    if (yamlProps == null) {
+      throw new IllegalArgumentException("Invalid yaml file");
+    }
+
+    final String latest =
+        (String)
+            Optional.ofNullable(yamlProps.get("latest"))
```
RoiEXLab: I don't want to get into the formatting argument here, but I think it's cleaner to do the string casting inside `Optional#ofNullable`

game-core/src/test/java/games/strategy/engine/framework/map/download/ContentDownloaderIntegrationTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+@Integration
+class ContentDownloaderIntegrationTest extends AbstractClientSettingTestCase {
+  @Test
+  void downloadFile() throws Exception {
+    try (CloseableDownloader contentDownloader =
+        new ContentDownloader(URI.create(UrlConstants.DOWNLOAD_WEBSITE))) {
+
+      final List<String> content =
+          IOUtils.readLines(contentDownloader.getStream(), StandardCharsets.UTF_8);
```
RoiEXLab: Warning, opinion:
Not really a fan of commons-io, starting with java 8 I believe the standard provides sufficient API, in this example you could use:
```java
try (var bufferedReader = new BufferedReader(new InputStreamReader(contentDownloader.getStream(), StandardCharsets.UTF_8))) {
  final List<String> content = bufferedReader.lines().collect(Collectors.toList())
}
```
If hamcrest has matchers for streams, then of course there's no need for the collection step

Also when operating with files directly, the Files class has some nicer APIs: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Files.html

game-core/src/test/java/org/triplea/live/servers/LiveServersFetcherIntegrationTest.java
```diff
@@ -0,0 +1,55 @@
...
+      public InputStream getStream() {
+        return asInputStream(TestDataFileReader.readContents(source));
+      }
+
+      @Override
+      public void close() {
+        // no-op
+      }
+    };
+  }
```
RoiEXLab: I feel like it would be easier to mock this class 🤔

game-core/src/test/java/org/triplea/live/servers/ServerYamlValidityTest.java
```diff
@@ -0,0 +1,98 @@
...
+    assertThat(liveServers.getLatestEngineVersion().isGreaterThan(new Version("0.0")), is(true));
+  }
+
+  @Test
+  void uriValuesDoNotEndInSlash() {
+    liveServers.getServers().stream()
+        .map(ServerProperties::getUri)
+        .filter(Objects::nonNull)
+        .map(URI::toString)
+        .forEach(uriString -> assertThat(uriString, not(StringEndsWith.endsWith("/"))));
```
RoiEXLab: Hmm isn't there an `assertAll` method?
In case there are no servers found, or filtered out completely this test wouldn't actually test anything, so at least testing the stream is not empty would be nice.
The same applies for all the other similar tests in this class

game-core/src/test/java/org/triplea/live/servers/ServerYamlValidityTest.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  @Test
+  void noVersionDuplicates() {
+    final long uniqCount =
+        liveServers.getServers().stream()
+            .map(ServerProperties::getMinEngineVersion)
+            .distinct()
+            .count();
+    assertThat((long) liveServers.getServers().size(), is(uniqCount));
```
RoiEXLab: Ok, didn't expect this cast to long here, but it makes somewhat sense

test-common/build.gradle
```diff
@@ -1,6 +1,7 @@
 description = 'Test utility library, generic test utilities useful for TripleA projects'
 
 dependencies {
+    implementation "commons-io:commons-io:$commonsIoVersion"
```
RoiEXLab: I already mentioned it, I don't think it's worth using this dependency, test-common could just implement helper methods in 2 or 3 lines and we have the same functionality without any additional dependencies

test-common/src/main/java/org/triplea/test/common/StringToInputStream.java
```diff
@@ -0,0 +1,16 @@
...
+import java.nio.charset.StandardCharsets;
+import lombok.experimental.UtilityClass;
+import org.apache.commons.io.IOUtils;
+
+@UtilityClass
+public class StringToInputStream {
+
+  /** Converts a {@code String} to an equivalent {@code InputStream}. */
+  public static InputStream asInputStream(final String inputString) {
+    return IOUtils.toInputStream(Strings.nullToEmpty(inputString), StandardCharsets.UTF_8);
```
RoiEXLab:
```suggestion
    return new ByteArrayInputStream(Strings.nullToEmpty(inputString).getBytes(StandardCharsets.UTF_8));
```

test-common/src/main/java/org/triplea/test/common/TestDataFileReader.java
```diff
@@ -0,0 +1,64 @@
...
+        // project. Check up a directory in case context is sub-project.
+        .or(() -> readFromProjectRoot("../" + filePath))
+        .or(() -> readFromResources(filePath))
+        .orElseThrow(() -> new TestDataFileNotFound(filePath));
+  }
+
+  private Optional<String> readFromProjectRoot(final String filePath) {
+    try {
+      final File f = new File(filePath);
+      final List<String> lines = IOUtils.readLines(new FileInputStream(f), Charsets.UTF_8);
```
RoiEXLab:
```suggestion
      final List<String> lines = Files.readAllLines(Paths.get(filePath), StandardCharsets.UTF_8);
```

test-common/src/main/java/org/triplea/test/common/TestDataFileReader.java
```diff
@@ -0,0 +1,64 @@
...
+    } catch (final IOException e) {
+      return Optional.empty();
+    }
+  }
+
+  @SuppressWarnings("ConstantConditions")
+  private Optional<String> readFromResources(final String filePath) {
+    final ClassLoader classLoader = TestDataFileReader.class.getClassLoader();
+    try (InputStream inputStream = classLoader.getResourceAsStream(filePath)) {
+      return Optional.of(IOUtils.toString(inputStream, StandardCharsets.UTF_8));
```
RoiEXLab:
```suggestion
      return Optional.of(new String(inputStream.readAllBytes(), StandardCharsets.UTF_8));
```

I think that's enough examples for now, I hope I got my point across

game-core/src/main/java/games/strategy/engine/framework/map/download/ContentReader.java
```diff
@@ -85,24 +64,8 @@ void downloadToFile(final String uri, final File file) throws IOException {
     checkNotNull(file);
 
     try (FileOutputStream os = new FileOutputStream(file)) {
-      downloadInternal(
+      downloadAndApplyAction(
           uri, is -> os.getChannel().transferFrom(Channels.newChannel(is), 0L, Long.MAX_VALUE));
```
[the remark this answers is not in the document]
RoiEXLab: Also just noticed this method exists, which simplifies this even further:
```suggestion
          uri, is -> Files.copy(is, file.toPath()));
```

game-core/src/main/java/games/strategy/engine/framework/map/download/ContentDownloader.java
```diff
@@ -0,0 +1,79 @@
...
+public final class ContentDownloader implements CloseableDownloader {
+  private final CloseableHttpClient httpClient;
+
+  @Getter(onMethod_ = @Override)
+  private final InputStream stream;
+
+  private final CloseableHttpResponse response;
+
+  public ContentDownloader(final URI uri) throws IOException {
+    this(HttpClients.custom().disableCookieManagement().build(), uri, HttpProxy::addProxy);
```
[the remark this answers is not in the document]
DanVanAtta: I haven't used Feign for generic HTTP, always for JSON response, but I suppose nothing would stop it.

The `HttpClient` linked would be an upgrade to the code that was modified in this class, having 3 different objects to close and the `getEntity()` to return an `InputStream` all seem a bit unwieldy.

There is a question whether to migrate or not. I would be concerned about 'lava-flow' and code churn. I'm not sure if the ROI is really that great. This HTTP client would benefit, and we have the 'forum-poster' that is now inconsistent. The 'forum-poster' client I've thought would be better off as a feign client. I think for consistency, because of the API it operates with is fundamentally JSON based, I'd vote for now we use Feign for the JSON API's consistently. Otherwise, someone would have to evaluate the pros/cons of migrating away from Feign and hopefully be prepared to make an across-the-board update. I think ROI comes into play though, a Reactive interface is nice, Feign interfaces are nice, IMO we are perhaps in a good-enough place and maybe better to focus on other things and avoid the churn. With all that said, refactoring `ContentDownloader` to use the reactive API or feign would be good IMO.

game-core/src/main/java/games/strategy/engine/framework/map/download/ContentReader.java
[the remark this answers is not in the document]
DanVanAtta: This seems safe, but I'm a bit nervous about committing updates without testing them, also nervous about increasing the test scope of the update. There is something to be said about reducing how many things that can go wrong.

test-common/src/main/java/org/triplea/test/common/TestDataFileReader.java
DanVanAtta: updated: 685cda865

When merging, (as always) let's please be sure to remove/prune the intermediary commit messages.

"Removing 'io.commons' from test-common" would be particularly misleading.

test-common/build.gradle
DanVanAtta: No objection, thanks for the suggestions
Updated: 685cda865c5fd9a4efb1187843dcccf29ab6a59b

game-core/src/main/java/games/strategy/triplea/settings/SelectionComponentFactory.java
```diff
@@ -417,10 +421,97 @@ public void reset() {
...
+    final String localhost = "http://localhost:8080";
+
+    final JTextField uriField = new JTextFieldBuilder().build();
+
+    final JRadioButton remoteServer = new JRadioButton("Remote Server");
+    remoteServer.setSelected(
+        ClientSetting.lobbyUseLocalhostOverride
+            .getValue()
+            .map(useLocalhost -> !useLocalhost)
+            .orElse(true));
```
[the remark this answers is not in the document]
DanVanAtta: Agree it's a bit confusing, too many negations.
Extracting to a variable makes sense. I also kinda cringe whenever we use a `ClientSetting<Boolean>` with the optional getter API, the optional is overkill/complication as the boolean setting value can never be non-null. I changed the setting type to be a `BooleanClientSetting` so we can avoid the Optional API and remove the 'orElse' logic: 724d47ebf

Side-note, the 'orElse' logic for a boolean client setting is technically dead code, ie:
```
value = booleanClientSettings.getValue().orElse(false);
```
Is the effectively the same as:
```
value =  (booleanClientSetting.getValue() == null) ? booleanClientSetting.getValue() : false;
```
Hence, the 'else' block is effectively dead code, so it's a bit overly complicated.

Updated: 724d47ebf

game-core/src/main/java/games/strategy/engine/lobby/client/login/LobbyLogin.java
```diff
@@ -49,7 +52,15 @@ public LobbyLogin(final Window parent, final ServerProperties serverProperties)
    */
   public void promptLogin() {
     if (serverProperties.isInactive()) {
-      showError("Could not connect to server", serverProperties.getServerErrorMessage().get());
+      SwingComponents.showDialogWithLinks(
+          DialogWithLinksParams.builder()
+              .title("Lobby Not Available")
+              .dialogText(
+                  String.format(
+                      "Your version of TripleA is too old, please download the latest: <a>%s</a>",
```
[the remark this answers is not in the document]
DanVanAtta: You're right, the existing is wrong. 
Fixed to have an href: b1659effd

game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadConfiguration.java
```diff
@@ -1,25 +1,21 @@
...
 
   static {
-    final Supplier<CloseableHttpClient> httpClientSupplier =
-        () -> HttpClients.custom().disableCookieManagement().build();
-    contentReader = new ContentReader(httpClientSupplier);
-    downloadLengthReader = new DownloadLengthReader(httpClientSupplier);
+    contentReader = new ContentReader(ContentDownloader::new);
+    downloadLengthReader =
+        new DownloadLengthReader(() -> HttpClients.custom().disableCookieManagement().build());
   }
```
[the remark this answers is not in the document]
DanVanAtta: Reasonably simplification given where this has landed. I was pretty focused on other objective sin the iterations, the stream closures gave a number of troubles. 

Updated: 2cbcfabea

game-core/src/main/java/games/strategy/engine/framework/map/download/ContentReader.java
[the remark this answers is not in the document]
DanVanAtta: Updated: 46e80d077b40799943dbb809d82aadd3b9d23f95

But then did some quick testing, and hit:
```
Testing:
SEVERE: Failed to download: https://github.com/triplea-maps/diplomacy-map_skin1/archive/master.zip
java.nio.file.FileAlreadyExistsException: /tmp/triplea1756697996669965286tmp
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:94)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:116)
	at java.base/sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:215)
	at java.base/java.nio.file.spi.FileSystemProvider.newOutputStream(FileSystemProvider.java:478)
	at java.base/java.nio.file.Files.newOutputStream(Files.java:219)
	at java.base/java.nio.file.Files.copy(Files.java:3066)
	at games.strategy.engine.framework.map.download.ContentReader.lambda$downloadToFile$0(ContentReader.java:65)
	at games.strategy.engine.framework.map.download.ContentReader.downloadAndApplyAction(ContentReader.java:50)
	at games.strategy.engine.framework.map.download.ContentReader.downloadToFile(ContentReader.java:65)
	at games.strategy.engine.framework.map.download.DownloadFile.lambda$newDownloadThread$2(DownloadFile.java:64)
	at java.base/java.lang.Thread.run(Thread.java:834)
```
And reverted (46e80d): 
c1dd7b9bcf7f524ffa1981d4558bb995cd26c90d

game-core/src/main/java/games/strategy/triplea/settings/SelectionComponentFactory.java
```diff
@@ -417,10 +421,97 @@ public void reset() {
...
+      public void reset() {
+        remoteServer.setSelected(!ClientSetting.lobbyUseLocalhostOverride.getValueOrThrow());
+        localhostServer.setSelected(ClientSetting.lobbyUseLocalhostOverride.getValueOrThrow());
+        uriField.setText(ClientSetting.lobbyUriOverride.getValue().map(URI::toString).orElse(""));
+        uriField.setEnabled(!ClientSetting.lobbyUseLocalhostOverride.getValue().orElse(false));
+      }
+    };
   }
 
+  // TODO: UNUSED-CODE remove this method
```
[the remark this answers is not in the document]
DanVanAtta: Yes, something I spotted while working on these updates. Noted the TODO so we'd have a better chance of coming back and fixing it.

game-core/src/main/java/org/triplea/live/servers/CurrentVersionSelector.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Override
+  public ServerProperties apply(final LiveServers liveServers) {
+    Preconditions.checkNotNull(liveServers);
+    Preconditions.checkArgument(!liveServers.getServers().isEmpty());
+
+    return liveServers.getServers().stream()
+        .filter(this::lessThanOrEqualToCurrentVersion)
+        .max(maxVersion())
+        .orElseThrow(() -> new IllegalArgumentException(noVersionFoundErrorMessage(liveServers)));
```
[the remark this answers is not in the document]
DanVanAtta: Thanks, some gnarly test-code for it, was a bit fun to have a pure algorithm problem for a change.

game-core/src/main/java/org/triplea/live/servers/FetchingCache.java
```diff
@@ -0,0 +1,60 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.util.Collections;
+import java.util.Optional;
+import java.util.function.Function;
+import lombok.Builder;
+import org.triplea.java.function.ThrowingSupplier;
+
+// TODO: test-client setting override
```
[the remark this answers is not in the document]
DanVanAtta: No, it's not! You should be nervous to see such TODOs and our convention served exactly its purpose here. This TODO was meant to be done before PR, it was helpful for me to come back and make the update, and that was done, but I did not remove the TODO message.

Updated (removed): 69f12e92e

game-core/src/main/java/org/triplea/live/servers/ServerYamlParser.java
DanVanAtta: As a general comment, whenever I find myself prefacing a review statement with something like:
> I don't want to get into the formatting argument here,

I would wonder to myself if the issue is important enough to be worth commenting on. The goal of CR is not to find and perfect every line detail, but to be sure there is reasonable clarity, no errors, design and technical debt are reasonable. One thing I've personally struggled with in reviews is to start at a high level first and then go low-level last.

With that said, I've no real preference one way or another, inside does have some benefits, not a bad clean-up; updated: c03e66b01

game-core/src/main/java/games/strategy/engine/lobby/client/login/LobbyLogin.java
```diff
@@ -57,8 +57,9 @@ public void promptLogin() {
               .title("Lobby Not Available")
               .dialogText(
                   String.format(
-                      "Your version of TripleA is too old, please download the latest: <a>%s</a>",
-                      UrlConstants.DOWNLOAD_WEBSITE))
+                      "Your version of TripleA is too old, please download the latest:"
+                          + "<br><a href=%s>%s</a>",
```
RoiEXLab: This might work, because HTML parser are not known for their strictness, but usually the URL is put in quotes

game-core/src/main/java/games/strategy/engine/framework/map/download/ContentReader.java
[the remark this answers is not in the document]
RoiEXLab: Ah, in this case you need to pass https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/StandardCopyOption.html#REPLACE_EXISTING as third param, didn't knew that either

game-core/src/main/java/org/triplea/live/servers/ServerYamlParser.java
DanVanAtta: To continue debating with myself, and unsolicited advice, @RoiEXLab happy to get the code cleaner, let's be sure to not lose sight of the forest and be sure the high level is right.

game-core/src/main/java/games/strategy/engine/framework/map/download/ContentReader.java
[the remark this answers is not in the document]
DanVanAtta: It's a bit curious we're using the same temp file to begin with, it might have issues with concurrent file downloads. Let's put a pin in this one and update post-merge (for the sake of testing and focus)

game-core/src/test/java/org/triplea/live/servers/ServerYamlValidityTest.java
DanVanAtta: Updated to add a check that we have at least one active server configured: ebc19b26d

game-core/src/test/java/org/triplea/live/servers/ServerYamlValidityTest.java
DanVanAtta: Indeed, same here.

Would be nice to have a `countInt` method, also a bit a weakness of hamcrest that leaving the cast off yields no compilation error, yet the assertion fails due to mismatched type. Might be argument for us to create a numeric matcher that will do a value comparison of long to int.

test-common/src/main/java/org/triplea/test/common/TestDataFileReader.java
DanVanAtta: Thank you for providing the sample code. I did not take wholesale as imports were needed, regardless the snippet made the update easy :+1:

game-core/src/main/java/games/strategy/engine/lobby/client/login/LobbyLogin.java
DanVanAtta: Probably better to not leave it to *might*, agree with the call-out, updated: 1e9afeb45

game-core/src/test/java/org/triplea/live/servers/LiveServersFetcherIntegrationTest.java
DanVanAtta: Added a missing `@Annotation` tag: c1fb6b28b

I was a bit on the fence with using a mock. Integration tests classically should not use them. Though, it did seem like it would be simpler, so I gave it a shot. One interesting thing, there was a failure when running a second test in the suite, it was because the configuration is cached. It turns out the test was wrong, that both sets of tests were verifying the same fetched configuration.

I added some cache clearing code and updated to use a mock: fabea08b7

DanVanAtta:
Thanks for checking this over @RoiEXLab ; I think I've responded to everything/updated all, let me know if I missed anything or if you see anything else on another review pass.

It's very possible as well this PR drops the last dependency on the java lobby code (in smoke test). Post-merge, we'll be able to delete all of that code and then start to simplify the socket network code as it'll always be specific to game-clients connecting to game-hosts :rocket:

game-core/src/main/java/games/strategy/engine/framework/map/download/ContentDownloader.java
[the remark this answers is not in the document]
RoiEXLab: Just tried migrating away from the websocket lib were using, and whilst definitely feasible it's currently not worth my time: The standard lib has a completely different API structure, which would require us to change a lot of the class structure

PR #5464, opened 2019-11-12
Infrastructure: Update http run_server script

DanVanAtta:
1. Script can be no-arg and rely on default args assuming we will deploy configuration.yml
   file to proper directory.
2. Also corrects script arg "server". It is currently configured as a flag rather than
   as an arg. By using default args, we solve this.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[] Manually tested

### DIFF
--- a/infrastructure/ansible/roles/http_server/templates/run_server.j2
+++ b/infrastructure/ansible/roles/http_server/templates/run_server.j2
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 cd $(dirname $0)
-java -jar -jar {{ http_server_jar }} -server {{ http_server_config_file }}
+java -jar -jar {{ http_server_jar }}


DanVanAtta:
It's pretty straight-forward, just a conditional in `main` method to check for args: https://github.com/triplea-game/triplea/blob/master/http-server/src/main/java/org/triplea/server/http/ServerApplication.java#L69

RoiEXLab:
Ah, thanks for the clarification, I was wondering if this would be some sort of weird vagrant mechanism 😅

RoiEXLab:
Oh and btw. I haven't had the time so far to review the large pr.
Because it's the first one in the list I didn't merge all the approved PRs so far, so if they're unrelated feel free to merge

DanVanAtta:
Cool, thanks @RoiEXLab 
Vagrant is not doing a lot other than simulating a linode instance. For the most part the latest round of development is live against prerelease. I likely should be doing that against vagrant so that there is a dogfooding benefit.

PR #5465, opened 2019-11-12
Infrastructure role, variable renaming & cleanups

DanVanAtta:
1. Rename DB role from "lobby_database" to "postgres"
2. Move application specification schema install from
   role: "postgres"->"flyway"
3. Rename hosts "lobby" to "dropwizardHosts"
4. Rename hosts "lobby_database" to postgres
5. Rename vagrant host from 'lobby' to 'vagrantHost'
6. Fix up variables names, remove lobby variable references
7. Remove 'ansibleMasterHost' reference, we'll deploy from Travis
   and not a dedicated ansible control server.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[] Manually tested


infrastructure/ansible/roles/flyway/tasks/main.yml
```diff
@@ -1,3 +1,41 @@
+- name: Create user on DB for application use
```
DanVanAtta: These are moved tasks from the old lobby_database role.

infrastructure/ansible/roles/postgres/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
+- name: Ensure latest bash and ssl
```
DanVanAtta: These tasks were in 'lobby_database', they are still specific to database. The other parts of 'lobby-database' tasks have been moved to 'flyway' role. The two sets of tasks are split based on what is specific to application vs generic to setting up the postgres database. So for example, if we were to deploy multiple schemas, or deploy multiple postgreses, we could just add more 'flyway'-like roles or re-use this postgres role.

infrastructure/ansible/roles/flyway/tasks/main.yml
```diff
@@ -1,3 +1,41 @@
...
+    login_password: "{{ flyway_db_password }}"
+    encrypted: yes
+    priv: "ALL"
+    role_attr_flags: NOSUPERUSER,NOCREATEDB
+    state: "present"
+
+# This is a hack because the previous tasks were supposed to set
+# the DB level password for our user, but it typically does not.
+# To achieve this, we'll run the native SQL commands to set the
+# user DB password.
```
RoiEXLab: So this is a vagrant bug?

DanVanAtta: To be sure, this is not new, we're copying code from the old postgres role to the flyway role.

The problem is in the `postgresql_user` task. Ansible commands map to python scripts, the problem is in the [postgresql_user python script](
https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/database/postgresql/postgresql_user.py).
Essentially the problem is the 'postgresql_user' task does not actually set the users password.

Issue #2921, opened 2018-01-28
Travis broken due to exceeding GitHub LFS quota


ssoloff:
I'm not sure how, but we've either exceeded the 1 GiB storage quota or 1 GiB bandwidth quota imposed by GitHub LFS.  Travis builds are failing with this error:

```
Error downloading object: src/main/resources/org/triplea/client/ui/javafx/images/background.png (fc82f76): Smudge error: Error downloading src/main/resources/org/triplea/client/ui/javafx/images/background.png (fc82f760ee88fbd0ee86ebbe58c7bf263601ab882c12d6cc1806d554fafe354d): batch response: This repository is over its data quota. Purchase more data packs to restore access.
```

I'm going to guess we're exceeding the bandwidth quota due to the frequency of our Travis builds.

Relevant link: https://help.github.com/articles/about-storage-and-bandwidth-usage/

RoiEXLab:
@ron-murhammer Could you check?
https://help.github.com/articles/viewing-storage-and-bandwidth-usage-for-an-organization/

prastle:
Noob ? but can we just buy more space?  Just a ? @RoiEXLab

ssoloff:
Just was able to look at this further now...  As the docs say, it affects forks, too.  The Travis build on my fork just failed for the same reason.

I'm going to create a test branch with LFS disabled and see if that is at least a possible solution.

prastle:
Off topic but I need the Lobby yaml update for the new TOC when ya guys get it working please.
https://forums.triplea-game.org/topic/545/revised-tournament-of-champions-season-12-toc-12

ssoloff:
Merge now requires a green Travis build, no?  I think you need admin rights to bypass.

RoiEXLab:
Oh right. 
Forgot about that @ssoloff

ssoloff:
@RoiEXLab It figures something like this would happen right after we enabled all the new constraints. :smile:

ssoloff:
I followed [this procedure](https://stackoverflow.com/a/41961459/3900879), and everything seems to be working on that branch.  I cloned the test branch into a separate folder and confirmed it retrieved the images from Git and not LFS.

@DanVanAtta @RoiEXLab @ron-murhammer Do we want to disable LFS now so the build isn't blocked?

RoiEXLab:
@ssoloff No objections on not using Git LFS all together.

ssoloff:
[...]

PR #5481, opened 2019-11-15
Fix builds, Hack Travis to use Java 11.0.5

DanVanAtta:
1. Custom install Java 11.0.5 on the JDK11 builds.
      The version of Java11 that is installed by travis seems baked
      in, oracle Java 11 does not help, nor does using apt to
      try and upgrade the java package to a more recent version.
      The custom install uses a Java 11.0.5 hosted in triplea-game/assets
      and installs it in a similar way that travis would install java.
    
    2. For runtime, use '--no-daemon' on gradle scripts, shaves about a minute.
      The DL of JDK11 slows things down (177MB DL), this second part of the
      update is to try and help compensate for the slowdown.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] No manual testing done
[x] Manually tested

- Verified travis build of branch for forked repo passed and checked the output.

### DIFF
--- a/.travis/install_java_11.0.5
+++ b/.travis/install_java_11.0.5
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+(
+  cd ~ || exit 1
+  wget "https://github.com/triplea-game/assets/raw/master/jdk-11.0.5_linux-x64_bin.tar.gz"
+  tar -xzf jdk-11.0.5_linux-x64_bin.tar.gz
+)
+
--- a/.travis/install_java_11.0.5
+++ b/.travis/install_java_11.0.5
@@ -2,7 +2,7 @@
 
 (
   cd ~ || exit 1
-  wget "https://github.com/triplea-game/assets/raw/master/jdk-11.0.5_linux-x64_bin.tar.gz"
+  wget "https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.5%2B10/OpenJDK11U-jdk_x64_linux_hotspot_11.0.5_10.tar.gz"
   tar -xzf jdk-11.0.5_linux-x64_bin.tar.gz
 )
 
--- a/.travis/install_java_11.0.5
+++ b/.travis/install_java_11.0.5
@@ -3,6 +3,6 @@
 (
   cd ~ || exit 1
   wget "https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.5%2B10/OpenJDK11U-jdk_x64_linux_hotspot_11.0.5_10.tar.gz"
-  tar -xzf jdk-11.0.5_linux-x64_bin.tar.gz
+  tar -xzf OpenJDK11U-jdk_x64_linux_hotspot_11.0.5_10.tar.gz
 )
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,11 @@ if: tag is blank
 #   https://sormuras.github.io/blog/2017-12-08-install-jdk-on-travis.html
 # Java12 does not have this issue, we do not download the Java11 script
 # for that JVM or else we would be overriding Java12 with Java11.  
+#
+# Additional note: The export commands seem to not apply when extracted
+# to a script and seemingly only work when located in the main travis config
+# here.
+#
 
 jobs:
   include:
@@ -25,12 +30,9 @@ jobs:
       env: DESCRIPTION="static verification"
       jdk: openjdk11
       before_install:
-        - cd ~
-        - wget "https://github.com/triplea-game/assets/raw/master/jdk-11.0.5_linux-x64_bin.tar.gz"
-        - tar -xzf jdk-11.0.5_linux-x64_bin.tar.gz
-        - export JAVA_HOME="$(pwd)/jdk-11.0.5"
+        - ./.travis/install_java_11.0.5
+        - export JAVA_HOME=~/jdk-11.0.5
         - export PATH=${JAVA_HOME}/bin:$PATH
-        - cd -
         - java --version
       install: skip
       script:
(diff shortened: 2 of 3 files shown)


DanVanAtta:
I was wondering why this build failed - we're hitting a rate-limit. 60 downloads per hour. The JDK file is stored in git-lfs, github does not allow files over 100MB, hence it's in git-lfs, which is delegated to the service that is rate-limited. With 4 downloads per build, we can only do 12 builds an hour. 

Seems we need to find a different location to host the java 11.0.5 file.

I went to the effort of storing the 11.0.5 file as any place to download it seems to be behind an oracle login wall. There might be a different location with a latest jdk that has a simple link where we can download the file repeatedly (but maybe we'll hit a rate limit..).

I'm thinking we may need to resort to hosting assets on one our servers.. It's getting late here, I'm gonna sleep on this, any ideas are welcome here for file hosting options.

DanVanAtta:
We might be able to pass an API key with the download request to increase the rate limit, not sure though.. just an idea.

.travis/install_java_11.0.5
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+(
+  cd ~ || exit 1
+  wget "https://github.com/triplea-game/assets/raw/master/jdk-11.0.5_linux-x64_bin.tar.gz"
```
RoiEXLab: An option would be to use the AdoptOpenJDK package, but it seems the archive variant is hosted on github as well:
https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.5%2B10/OpenJDK11U-jdk_x64_linux_hotspot_11.0.5_10.tar.gz

Might be worth trying though, maybe they have higher limits for their archives.
My other ideas:
1. Setup a simple Webserver/reuse nginx on the forum server to host the file
2. Travis has some sort of caching system, so if we download the jdk as usual and just let travis cache it, we might not hit the rate limit at all.

ssoloff:
I'd also refer you to #2921.  In addition to rate limiting quotas, there's an overall 1 GiB storage and bandwidth quota per month (unless you pay for more).  That's why we had to drop Git LFS when we first tried it about two years ago.  After a week's worth of builds, we ran out of quota (and that was only storing the JFX client graphics in LFS).

(BTW, I don't think you can simply download an LFS-hosted file over HTTP.  I think you need an actual LFS-enabled Git client.  If you look at the logs, the `wget` command in the build is only downloading a 134-byte file, which is the Git LFS file descriptor that points to the actual blob.  You could try to hack your own sans-Git solution using [this](https://kamranicus.com/posts/2017-06-14-downloading-git-lfs-files-from-tfs-vsts) as a guide.)

A quick solution may be to just create a manual release in the `assets` repo and upload the file there (I don't know if GitHub releases have a per-file size limit less than the ~172 MiB required.)

A longer term solution would be to create a TripleA organization account on JCenter and use that to host binary artifacts that don't need to be versioned.  I think we talked about this a while back, but it never went anywhere.  IMO, a proper artifact repository would solve a lot of the bloat we have in the `assets` repo.

.travis/install_java_11.0.5
RoiEXLab:
```suggestion
  wget "https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.5%2B10/OpenJDK11U-jdk_x64_linux_hotspot_11.0.5_10.tar.gz"
```
I'll just go ahead and try if this works

RoiEXLab:
Ok, this seems to work.
Is everyone fine with this solution? @ssoloff @DanVanAtta

ssoloff:
> Is everyone fine with this solution?

:+1:

RoiEXLab:
I'll just go ahead and merge this so the build works again, feel free to follow up with more PRs

RoiEXLab:
I restarted the builds on all the failing PRs, looks good

panther2:
@RoiEXLab Shouldn't 2.0.16548 include Java 11.0.5 now? 
Instead it is 
tra
Or what am I missing?

RoiEXLab:
Good point, but this only affects the version were building against on travis, not the version that's actually being bundled with triplea.
But we should definitely consider using the same version for both.
It looks like github has it's own ci solution now that from my understanding fully relies on either docker images or javascript environments, so we could complete customize our build environment which allows us to keep our version up-to-date and synchronised at all times

DanVanAtta:
### RE: git LFS file descriptor

The git LFS file descriptor was interesting, for a good hour or two the original file was downloaded, then eventually the file was replaced by the file descriptor at which point the DL broke. It seemed like if you're not rate limited, then the redirect on download goes to the actual file, but if rate limited then you get the file descriptor. I'm a bit curious if that is actually the case..

### RE: setting up an artifact repository

I certainly can see merits to this. We no longer need to DL most of the bundleded JDKs, so there is certainly a lot less there. I wonder a bit how long we'll need to use our own JDK for builds, Travis may very well update their build image before JCenter's hard cutover to HTTPs. My understanding is they plan to do that in a few weeks and what we experienced was an intentional 24 hour outage.

Bottom line, if it's just images and audio assets eventually, is there value in a repository compared to hosting files in a github repo?

### RE: Travis Caching + OpenJDK

I'm a bit worried about rate limiting on the OpenJDK download and the liability of us getting blocked or that not being available. Using OpenJDK is plenty okay, I was not able to find that link a few nights ago when hacking at this. Regardless, with the large file downloads, setting up Travis to cache these files sounds like a win.

### RE: Github CI

It would be nice if the YAML configuration were more 1:1, the documentation looks to still be pretty fresh. There is advantage to having a slightly more reliable build system. It's worth recalling that while Travis has given us a few issues, it's worked pretty well for many builds.

The better integration into github also seems to be a benefit. I'd wonder about:
- build speeds, would it be comparable
- docker images, would we have to maintain these ourselves?
- configuration complexity and script migration, can we re-use most of our build scripts or is there a lot of re-work that we would have to do?
- environment variables/secrets. Soon we'll want/need to store a password for ansible vault, we'll need a way to inject that value into a script (likely as an env variable). 

I suspect nothing will be a full blocker, but there are some questions to answer. Last, we should probably be careful about investing into something that replaces a mostly adequate solution; replacing the CI only adds so much value.

DanVanAtta:
Re: Java Version

We may want to consider going with the 6 month release version instead of LTS. Another way to solve the custom DL problem is simply to use JDK12 or JDK13 that does not have the out-of-date JDK11 problem.

RoiEXLab:
@DanVanAtta Regarding GitHub CI, I don't think we should be rushing for this, however I think I'll try to play around with it soon™ just to see how well it's usable.

DanVanAtta:
Sounds good @RoiEXLab , I suspect a prototype to demonstrate and compare capabilities is appropriate. There might be room for using both (maybe?), at the very least gathering more information would be great. Agree no rush on doing so.

PR #5474, opened 2019-11-14
Update asset bundle - pick up new unit-scroller image

DanVanAtta:
updated asset bundle replaces the 'unit_sleep' 'moon' icon
for a 'guard' icon. This is motivated to avoid the false
antonym of the 'sun' icon (highlight) and 'moon' icon
(unit_sleep).

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature update or enhancement
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[x] Other:  image update

PR #5488, opened 2019-11-16
Update asset download for missing unit scroller images

ron-murhammer:
Missing unit scroller images causing errors

### DIFF
--- a/game-headed/build.gradle
+++ b/game-headed/build.gradle
@@ -48,7 +48,7 @@ clean.doFirst {
 task downloadAssets {
     doLast {
         download {
-            src "https://github.com/triplea-game/assets/releases/download/18/game_headed_assets.zip"
+            src "https://github.com/triplea-game/assets/releases/download/23/game_headed_assets.zip"
             dest "$projectDir/.assets/assets.zip"
             overwrite false
         }


ron-murhammer:
@DanVanAtta I think this unfortunately shows another down side of splitting those assets out and any update to them then requires devs to remember to update the gradle build to pull a new version. Merging this.

DanVanAtta:
Which image was missing?
I don't think it necessarily mattered that the logic was split if the image reference was just wrong. Having the assets split does have advantages, notably that we can update image names and choose when we pick them up. The build also automatically updates images to be losslessly compressed, that would be painful on the 'main build'. And to another extent, we simply do not have much space left in the main repo, it's slow to clone and we are not that far away from 1 gig, it's not viable to keep assets there.

> them then requires devs to remember to update the gradle build to pull a new version. 

IE, this is actually a feature, I wonder where the issue was as I was not aware that the names had changed. There was also a PR for this update, I wonder what happened to it.

DanVanAtta:
It is also nice to be able to work on images without needing to put those as part of the PR that uses them. Blocking code+image work on review is more painful compared to iterating on images and allow those to be updated without dependency on code. For example the 'wake-all' image is committed but not used. IF it were part of the code base, it'd be a dead asset and should never have been merged (and if merged, it re-opens the slippery slope where we had accumulated many unused assets, it was a task to identify how many were unused, there were quite a number. It's also painful to have those as one would be nervous update updating them in case it breaks something non-obvious. Support for dead-code, or dead-images is quite demoralizing and inefficient).

DanVanAtta:
The update was in-flight: https://github.com/triplea-game/triplea/pull/5474

Looks like we needed to merge in-order, or include the update with the dependency change. Probably the latter, I'm not sure if it was realized there was actually a dependency change. So a few things came along to contribute here, because we had a problem, we should not ignore there are many benefits. For example, update no20 in assets was ignored and reverted, it was *really* good to not have to do that experimentation in the main repo, wasting space and time, and was *really* good to have a place where we can avoid CR delay.

PR #5490, opened 2019-11-16
Let's see if this works

RoiEXLab:
I'm just trying out this standard template that just runs Unit Test on JDK 12


DanVanAtta:
@RoiEXLab do you have Travis enabled to build your fork of TripleA? I don't object to some live testing. It's a nice tool to have available for travis to build your fork (I tend to use that for a lot of CI build development)

RoiEXLab:
@DanVanAtta @ron-murhammer 6m build time seems pretty good.
Might be partly due to the gradle update (they promised faster incremental builds in the 6.0 update), but it seems to be at least on par with travis.


@DanVanAtta I have travis enabled on my fork, I'm not sure if it's still properly setup though 😅 
Anyways I don't intend to merge this at all, just wanted to check if it's really that easy to get started and it is.
This gradle template allows us to setup a JDK and a gradle command within the config file you see is added here. So I really just had to change 2 lines to get this running. I was initially trying to use JDK 13 , but as always there are some compatibility issues with errorprone (and lombok probably)

DanVanAtta:
All good, happy to see the experimentation either way.

RoiEXLab:
Ok, we gained some experience in #5507

PR #5491, opened 2019-11-16
Fix unit tests

RoiEXLab:
This fixes some unit tests that fail on my german windows system. (Some failures were due to localization of java, some due to different system architecture, see the commits)

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Bug fix:  
[] Other:   

## Testing

[] Manually testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/KeyBindingSupplier.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/KeyBindingSupplier.java
@@ -1,8 +1,6 @@
 package games.strategy.triplea.ui;
 
-import com.google.common.base.Preconditions;
 import java.awt.event.InputEvent;
-import java.awt.event.KeyEvent;
 import java.util.Map;
 import java.util.function.Supplier;
 import javax.swing.KeyStroke;
@@ -49,7 +47,8 @@ public interface KeyBindingSupplier extends Supplier<Map<KeyStroke, Runnable>> {
   static KeyStroke fromKeyEventCode(final int code, final ModifierKey modifierKey) {
     final KeyStroke result = KeyStroke.getKeyStroke(code, modifierKey.modifierMask);
     if (result.toString().contains("UNKNOWN")) {
-      throw new IllegalArgumentException("Be sure to use a constant from 'KeyEvent', unknown key constant: " + code);
+      throw new IllegalArgumentException(
+          "Be sure to use a constant from 'KeyEvent', unknown key constant: " + code);
     }
     return result;
   }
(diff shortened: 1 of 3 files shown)


game-core/src/main/java/games/strategy/triplea/ui/KeyBindingSupplier.java
```diff
@@ -47,9 +47,10 @@ static KeyStroke fromKeyEventCode(final int code) {
   }
 
   static KeyStroke fromKeyEventCode(final int code, final ModifierKey modifierKey) {
-    Preconditions.checkArgument(
-        !KeyEvent.getKeyText(code).toUpperCase().contains("UNKNOWN"),
-        "Be sure to use a constant from 'KeyEvent', unknown key constant: " + code);
-    return KeyStroke.getKeyStroke(code, modifierKey.modifierMask);
+    final KeyStroke result = KeyStroke.getKeyStroke(code, modifierKey.modifierMask);
+    if (result.toString().contains("UNKNOWN")) {
```
DanVanAtta: Is `toUpperCase` not needed?

game-core/src/main/java/org/triplea/debug/TextUtils.java
```diff
@@ -7,7 +7,7 @@ private TextUtils() {}
 
   static String textToHtml(final String text) {
     return "<html>"
-        + HtmlEscapers.htmlEscaper().escape(text).replaceAll(System.lineSeparator(), "<br/>")
+        + HtmlEscapers.htmlEscaper().escape(text).replaceAll("\r?\n", "<br/>")
```
DanVanAtta: Are you sure we need `\r`? AFAIK it's not actually needed in most places where we have it. I think we have it in so many places due to confusion vs actually needing it.
Isn't `lineSeperator()` also preferred? I'm curious how this breaks.

game-core/src/test/java/games/strategy/triplea/settings/PathClientSettingTest.java
```diff
@@ -15,12 +16,15 @@
   final class EncodeValueTest {
     @Test
     void shouldReturnEncodedValue() {
+      final String separator = File.separator;
```
DanVanAtta: :+1: 
We probably have more of these elsewhere.

game-core/src/main/java/org/triplea/debug/TextUtils.java
RoiEXLab: I don't know if it's actually needed, but in case it is it will be replaced as well.
The reason why this test didn't work is the following:
The test case passed some string along the lines of `Test\nString`, where on linux the `\n` got replaced with `<br>`, but on linux the test was trying to replace `\r\n` in the string which obviously isn't part of it, so the string stayed the same.

That's why the `\r` is optional, it would probably still work if it got ommited, but \r is really uneccesary

game-core/src/main/java/games/strategy/triplea/ui/KeyBindingSupplier.java
RoiEXLab: Nope, the `UNKNOWN` in `toString()` is all caps already

DanVanAtta: Okay, I think I put that in there in case it changes in the future or is system dependent.

PR #5492, opened 2019-11-16
AI fix v3 subs submerging

ron-murhammer:
Address: https://forums.triplea-game.org/topic/105/ai-development-discussion-and-feedback/181

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProOddsCalculator.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProOddsCalculator.java
@@ -136,8 +136,6 @@ public class ProOddsCalculator {
       final List<Unit> attackingUnits,
       final List<Unit> defendingUnits,
       final boolean checkSubmerge) {
-    final GameData data = ProData.getData();
-
     final boolean hasNoDefenders =
         defendingUnits.stream().noneMatch(Matches.unitIsNotInfrastructure());
     final boolean isLandAndCanOnlyBeAttackedByAir =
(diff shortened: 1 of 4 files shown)


game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProOddsCalculator.java
```diff
@@ -176,6 +186,7 @@ public ProBattleResult callBattleCalculator(
       final List<Unit> attackingUnits,
       final List<Unit> defendingUnits,
       final Set<Unit> bombardingUnits,
+      final boolean checkSubmerge,
```
DanVanAtta: Any objection to using an enum instead of boolean parameter? The "true, true" invocation is difficult to understand, requires a redirect to look at this code, slows down reading comprehension. Alternatively a static factory method could be employed to encapsulate the boolean parameters to be class-internal.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProOddsCalculator.java
```diff
@@ -217,23 +228,40 @@ public ProBattleResult callBattleCalculator(
...
-        && !defendingTransportedUnits.isEmpty()) { // Add TUV swing for transported units
+    if (t.isWater() && !defendingTransportedUnits.isEmpty()) {
       final double transportedUnitValue =
           TuvUtils.getTuv(defendingTransportedUnits, ProData.unitValueMap);
       tuvSwing += transportedUnitValue * winPercentage / 100;
     }
 
+    // Remove TUV and add to remaining units for defenders that can submerge before battle
+    if (tuvSwing > 0
+        && checkSubmerge
```
DanVanAtta: are all 4 of these boolean conditions to represent "canSubmergeBeforeBattle"? If so, extracting to a variable or method could make the comment redundant and the code a bit easier to grok

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProOddsCalculator.java
```diff
@@ -217,23 +228,40 @@ public ProBattleResult callBattleCalculator(
...
+        && Properties.getSubRetreatBeforeBattle(data)
+        && defendingUnits.stream().anyMatch(Matches.unitCanEvade())
+        && attackingUnits.stream().noneMatch(Matches.unitIsDestroyer())) {
```
DanVanAtta: side-note, it really seems like there should a rules class to do this logic. If there are other units that can prevent subs from submerging, seems this would break. 

Does v2 or v1 rules allow destroyers to block sub-submerge? It might be moot if there are no destroyers on those maps, but could be worth considering that the 'isDestroyer' condition is truly universal.

ron-murhammer: The isDestroyer is actually checking the unit property not whether the unit itself is a destroyer. I can't remember if v1 handles it the same as the rest but at the moment all the AI logic assumes that isDestroyer property prevents submerging subs.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProOddsCalculator.java
ron-murhammer: Agree. Updated.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProOddsCalculator.java
ron-murhammer: Moved the boolean parameters out of public methods so its just internal implementation in this class.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProOddsCalculator.java
DanVanAtta: I think v1 can get away with it because there are no destroyer units.

PR #5497, opened 2019-11-17
PR Auto-Correct

DanVanAtta:
Assuming maintainer edits is enabled, this update will have the builder-bot
run spotlessApply and push any updates back to the PR branch. Also, builder-bot
will look for any unnecessary lombok '@Log' tags and remove them if there
is no logging in the respective class file.

This opens up a framework for additional auto-corrections as we can think
of them.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[x] Manually testing done

- dry-run: https://github.com/triplea-game/triplea/pull/5498
- tested on a forked repo with travis

## Additional Review Notes

For discussion and to be decided:
- edits from maintainers is required for this to work, perhaps it gives us a point of pause. Then again, 'spotlessCheck' fails if there are not updates, and if there are auto-corrections and we can't update the branch to fix them, the branch will fail all the same
- I added the auto-correction to be as a first step of static analysis while all the tests run in parallel. It may make sense to add the auto-correction as a first step before we then run all tests and checks in parallel. This would slow down build time. To consider, it could make sense to have the auto-correct to merge to the latest master, that way when we test/build we'll be testing from the head of master. If we do that, then it probably makes even more sense to do the auto-correction as first step to everything else.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,8 @@ jobs:
       script:
         ## shellcheck all shell scripts
         - find .travis/ -type f | xargs grep -lE "^#\!/bin/bash$" | xargs shellcheck
-        - ./gradlew --no-daemon --parallel validateYamls spotlessCheck checkstyleMain checkstyleTest pmdMain
+        - ./.travis/auto-correct
+        - ./gradlew --parallel validateYamls spotlessCheck checkstyleMain checkstyleTest pmdMain
     - stage: verify
       env: DESCRIPTION="JDK11 integTest"
       jdk: openjdk11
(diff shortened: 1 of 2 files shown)


DanVanAtta:
Travis does not seem to like draft PR builds very much (and keeps trying to pull from the wrong PR):
```
The command "eval git fetch origin +refs/pull/5496/merge: " failed 3 times.
```

PR #5498, opened 2019-11-17
Auto correct test

DanVanAtta:
Assuming maintainer edits is enabled, this update will have the builder-bot
run spotlessApply and push any updates back to the PR branch. Also, builder-bot
will look for any unnecessary lombok '@Log' tags and remove them if there
is no logging in the respective class file.

This opens up a framework for additional auto-corrections as we can think
of them.

 

## Functional Changes

[] New map or map update
[x] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Bug fix:  
[] Other:   

## Testing
[x] Manually testing done

- dry-run: https://github.com/triplea-game/triplea/pull/5498
- tested on a forked repo with travis

## Additional Review Notes

For discussion and to be decided:
- edits from maintainers is required for this to work
- I added the auto-correction to be as a first step of static analysis while all the tests run in parallel. It may make sense to add the auto-correction as a first step, then once completed, run everything else. This would make builds a bit slower. It makes a lot more sense if we do something like automatically rebase to the latest master.

### TODO:

- [ ]  abort the auto-correct if building from a fork.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,8 @@ jobs:
       script:
         ## shellcheck all shell scripts
         - find .travis/ -type f | xargs grep -lE "^#\!/bin/bash$" | xargs shellcheck
-        - ./gradlew --no-daemon --parallel validateYamls spotlessCheck checkstyleMain checkstyleTest pmdMain
+        - ./.travis/auto-correct
+        - ./gradlew --parallel validateYamls spotlessCheck checkstyleMain checkstyleTest pmdMain
     - stage: verify
       env: DESCRIPTION="JDK11 integTest"
       jdk: openjdk11
--- a/.travis/auto-correct
+++ b/.travis/auto-correct
@@ -2,6 +2,7 @@
 
 CURRENT_BRANCH=${TRAVIS_PULL_REQUEST_BRANCH:-$TRAVIS_BRANCH}
 
+git remote -v
 
 function main() {
   runSpotless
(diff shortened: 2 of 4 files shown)


RoiEXLab:
@DanVanAtta I think the problem is that on PullRequest the secrent environment variables are not available for security reasons.
I don't think there's a good way to work around that using just travis.
This might be the opportunity to try out github actions. Because of the integration with github, there's an exception for the github auth token there, so the authentication would still work:
https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets#using-encrypted-secrets-in-a-workflow
From what I was able to google, there are several "actions" out there already that can do the individual steps.

Something along the lines of this might already do the trick:
```
name: CI

on: [push]

jobs:
  build:

    runs-on: ubuntu-latest

    steps:
    - uses: actions/checkout@v1
    - name: Set up JDK 11
      uses: actions/setup-java@v1
      with:
        java-version: 11
    - name: Run the bash scrip
      run: ./gradlew spotlessApply
    - name: Commit changes # This is the step name that will be displayed in your runs
      uses: EndBug/add-and-commit@v2.1.0 # You can change this to use a specific version
      with: # See more info about inputs below
        author_name: Your Name
        author_email: mail@example.com
        message: "Your commit message"
        path: "."
        pattern: "*.java"
        force: false
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # Leave this line unchanged

```

This combines the standard images with this one I found in the marketplace: https://github.com/marketplace/actions/add-commit

DanVanAtta:
Seems like some more support/evidence for going with github actions. Otherwise could spot @RoiEXLab , auto-correction might be unworkable at a PR level.

Issue #5499, opened 2019-11-17
ProMoveUtils and IMoveDelegate should be refactored to work on Move objects


asvitkine:
ProMoveUtils and IMoveDelegate should be refactored to work on Move objects.

This issue is to capture the refactoring proposed as a follow up to:
https://github.com/triplea-game/triplea/pull/5416

Currently, three parallel lists are used by the Pro AI to keep track of moves. This is messy, instead it should be a list of Move objects. I plan to address this.

ron-murhammer:
@asvitkine Looks like there are some bugs in the current functionality. I think it would be better to do some more testing before further refactoring.

asvitkine:
Thanks for the heads up. Agree about fixing any issues with the current functionality before more refactoring. (Which is why I wanted to do the refactoring in separate PR. :)

DanVanAtta:
To super nit on terminology, refactoring is a process where code that is already tested is simplified. Changing code without writing tests for it first is re-writing. Refactoring is 'safe' so long as the tests are good, TripleA is devious because most code can hardly be tested without first being re-written.

I mostly call this out so that our response is not to "be more safe, don't simplify, be surgical", but instead to increase test coverage which pays dividends in safety of changes and the potential simplicity that can ensue. IE: the response should not be "don't make changes", it should be more "write better/more tests".

PR #5503, opened 2019-11-17
Refactor move code to use MoveDescription objects.

asvitkine:
Refactor move code to use MoveDescription objects.

This simplifies the code and avoids the pattern of parallel output list params that was being used in the AI code.

This PR still leaves some TODOs behind that I plan to address in follow-up PRs.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[X] Code Cleanup or refactor
[] Configuration Change
[X] Bug fix:  https://github.com/triplea-game/triplea/issues/5499
[] Other:   

## Testing

[] Manually testing done


asvitkine:
This is not yet ready to be merged, but sending it for early feedback to make sure there are no concerns with this direction. This is the refactoring I alluded to when working on the AI batching change.

Note: There are still TODOs here indicating additional follow up work to be done to improve things further. For instance, for now I'm not yet consolidating MoveDescription with MoveBatcher.Move, although I plan to later.

game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
```diff
@@ -23,7 +21,13 @@ public MoveDescription(
       final Map<Unit, Collection<Unit>> dependentUnits) {
     super(units);
     this.route = route;
```
DanVanAtta: Nit: can simplify the NPE check with guava, eg:
```
this.route = Precondtions.checkNotNull(route);
```

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
```diff
@@ -1,35 +1,35 @@
...
 import java.util.List;
 import java.util.Map;
-import javax.annotation.Nullable;
 
 /**
  * MoveBatcher implements an algorithm that batches AI moves together, reducing the number of
  * distinct moves the AI does (and thus the visual delay) by combining moves together.
  */
 public class MoveBatcher {
+  // TODO: Merge/replace with MoveDescription.
```
DanVanAtta: nit: should include a 'grep' or a 'grouping' token with the TODO. Something so we can track it back to the original task, we've a problem of decades old TODO comments that do not make a lot of sense and decided to avoid that would be by marking/grouping TODO comments so we can find them. An issue, or a PR number would probably be good choices.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
```diff
@@ -415,77 +403,70 @@ public static void calculateBombingRoutes(
...
-          final Route r2 = moveRoutes.get(j);
-          if (r.equals(r2)) {
-            moveUnits.get(j).addAll(moveUnits.get(i));
-            moveUnits.remove(i);
-            moveRoutes.remove(i);
-            i--;
-            break;
-          }
+    // TODO: Use MoveBatcher here - or ideally at the time the moves are being generated.
+    for (int i = 0; i < moves.size(); i++) {
```
DanVanAtta: Is this block checked by automated test? How safe is the update here?

game-core/src/main/java/games/strategy/triplea/ai/weak/WeakAi.java
```diff
@@ -2,6 +2,7 @@
 
 import com.google.common.collect.Streams;
```
DanVanAtta: Side-note: `WeakAI` we want to delete. The only reason we do not delete is for battle simulation. We new up a `WeakAI` to provide logic for the battle simulation. If we can extract that individual logic and supply it to battle simulation / battle calc, then we could remove this AI class entirely.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
asvitkine: Done.

game-core/src/main/java/games/strategy/triplea/ai/weak/WeakAi.java
asvitkine: Ack.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
asvitkine: I don't think it is.

We should be careful to not diverge from the previous logic. I adjusted it to more closely match the previous logic, please double check. Thanks!

DanVanAtta:
@asvitkine a merge conflict showed up, resolving that with luck will unblock the next travis build,

DanVanAtta:
Hmm, I must have made a mistake somewhere when looking at merge conflicts. I think maybe I did not fetch your remote before simulating the merge.

DanVanAtta:
Ah yeah, I recall some of those were existing :grin: 
@asvitkine I cut a branch and resolved conflicts: https://github.com/triplea-game/triplea/compare/master...DanVanAtta:refactor_move_code_merged?expand=1

It's looking weird submitting a PR into your branch.
I think you should be able to cherry-pick the merge commit and resolve the merge conflicts (then run spotlessApply to clean up any cruft):
```
git cherry-pick  6e0add5 
```

game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
```diff
@@ -0,0 +1,80 @@
...
+    super(Collections.unmodifiableCollection(units));
+    this.route = Preconditions.checkNotNull(route);
+    this.unitsToTransports = Collections.unmodifiableMap(unitsToTransports);
+    this.dependentUnits =
+        Collections.unmodifiableMap(
+            dependentUnits.entrySet().stream()
+                .collect(
+                    Collectors.toMap(
+                        Map.Entry::getKey,
+                        e -> Collections.unmodifiableSet(new HashSet<>(e.getValue())))));
```
RoiEXLab: Just learned that https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Set.html#copyOf(java.util.Collection)

Is a thing, which performs an immutable copy of a set.

asvitkine: Thanks, done!

asvitkine:
> Ah yeah, I recall some of those were existing 😁
> @asvitkine I cut a branch and resolved conflicts: https://github.com/triplea-game/triplea/compare/master...DanVanAtta:refactor_move_code_merged?expand=1
> 
> Something does not look right submitting a PR into your branch.
> I think you should be able to cherry-pick the merge commit and resolve the merge conflicts with that (then run spotlessApply to clean up any cruft):
> 
> ```
> git cherry-pick  6e0add5 
> ```

The cherry-pick didn't find it, but I synced my upstream/master and merged from it and resolved conflicts. That seemed to have worked, although I'm getting some compile errors locally for files I'm not changing. Not sure.

/Projects/triplea_fork/triplea/game-core/src/test/java/org/triplea/game/startup/SetupModelTest.java:9: error: cannot find symbol
import games.strategy.engine.random.IRemoteDiceServer;

game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import lombok.Getter;
+
+/** Describes an action that moves one or more units along a specific route. */
+public class MoveDescription extends AbstractMoveDescription {
+  private static final long serialVersionUID = 2199608152808948043L;
+  @Getter private final Route route;
+  @Getter private final Map<Unit, Unit> unitsToTransports;
+  @Getter private final Map<Unit, Collection<Unit>> dependentUnits;
```
DanVanAtta: Have you considered using Guava `Multimap`? https://guava.dev/releases/23.0/api/docs/com/google/common/collect/Multimap.html

It is essentially the same datastructure as `Map<X Collection<Y>>`.

game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+import lombok.Getter;
+
+/** Describes an action that moves one or more units along a specific route. */
+public class MoveDescription extends AbstractMoveDescription {
+  private static final long serialVersionUID = 2199608152808948043L;
+  @Getter private final Route route;
```
DanVanAtta: minor, if `@Getter` is on all variables, you can put it on the class and have it in one spot.

game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+import lombok.Getter;
+
+/** Describes an action that moves one or more units along a specific route. */
+public class MoveDescription extends AbstractMoveDescription {
```
DanVanAtta: There is not much in `AbstractMoveDescription`, it might be easier to just have a variable for 
```
  private final Collection<Unit> units;
```
The benefits is we avoid the coupling from inheritance, one of the stronger/strongest forms of coupling out there.

game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
```diff
@@ -0,0 +1,77 @@
...
+
+  @SuppressWarnings("UndefinedEquals")
+  private static boolean collectionsAreEqual(final Collection<Unit> a, final Collection<Unit> b) {
+    // https://stackoverflow.com/questions/1565214/is-there-a-way-to-check-if-two-collections-contain-the-same-elements-independen
+    return HashMultiset.create(a).equals(HashMultiset.create(b));
+  }
+
+  @Override
+  public String toString() {
+    return "Move message route:" + route + " units:" + getUnits();
```
DanVanAtta: I'm a bit confused by "Move message", what does message mean in this context?
Have you considered using lombok `@ToString` to avoid a custom toString implementation?

Same thing for `hashCode` and `equals`, you can use `@EqualsAndHashcode` to avoid writing those on your own.

asvitkine: I'm not sure. This class is being moved from games.strategy.triplea.delegate.data package and I didn't change the implementation of this method.

Let me know if you prefer I change it to use @ToString assuming nothing relies on this message.

For equals & hashCode, I'm not sure the defaults would work since the collections may have different underlying implementations. I just care about the contents, not what their underlying types are.

game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
asvitkine: Looks like it's used in AbstractUndoableMove.getDescriptionObject(). Not sure its easy to decouple given it's not the only subclass.

game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
asvitkine: I'd rather not make this change bigger than it is. This type is already in use throughout the code.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
```diff
@@ -411,40 +401,34 @@ public static void calculateBombingRoutes(
...
+          final Route r2 = moves.get(j).getRoute();
           if (r.equals(r2)) {
-            moveUnits.get(j).addAll(moveUnits.get(i));
-            moveUnits.remove(i);
-            moveRoutes.remove(i);
+            final var mergedUnits = new ArrayList<Unit>();
+            mergedUnits.addAll(moves.get(j).getUnits());
+            mergedUnits.addAll(moves.get(i).getUnits());
+            moves.set(i, new MoveDescription(mergedUnits, r));
+            moves.remove(i);
```
DanVanAtta: I don't think I understand the intent of these two lines. We set element 'i', then we remove element 'i'. What's the effect of the 'set'?

game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
DanVanAtta: Sounds okay, didn't quite realize how much of this class was just a move vs update.

game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
DanVanAtta: Is it possible to decouple *just* this one subclass? 

NB: It turns out this update is breaking change to save game serialization. This is a good time to make such changes. After the 2.0 release is launched, and we will need to maintain save-game compatibility.

game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
DanVanAtta: I'd hope nothing relies on this message, it seems here for debug.
W.r.t equals, we can represent the internal collection as `Set` to indicate we care that ordering does not matter.

DanVanAtta: Regardless, since we're doing a file move mostly, this kind of update would be best left for follow-up.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMoveUtils.java
DanVanAtta: @asvitkine please double check this one, I think it might supposed to be `remove(j)`

Running through a debugger, I noticed 'move' `i` gets the aggregation, then we it gets removed. It looks like it would make sense if it were `j` being removed.

Example below, with 'move @ index 2' being merged with 'move @ index 3'

Screenshot from 2019-11-21 20-58-51
Screenshot from 2019-11-21 21-00-00
Screenshot from 2019-11-21 21-01-47

asvitkine: Thanks for noticing, this was indeed a bug.

It was supposed to be set(j), not i. It's meant to match the previous logic, which was:

```
            moveUnits.get(j).addAll(moveUnits.get(i));
            moveUnits.remove(i);
            moveRoutes.remove(i);
```

i.e. making entry at index j be the merged move and removing i.

Fixed.

DanVanAtta:
@asvitkine Thanks for the update. I don't think there should be anything left. If you would, for some quick smoke testing would be healthy to verify that an AI vs AI game plays cleanly. I'll hopefully do a final pass tomorrow (about 14 hours from now)

asvitkine:
> @asvitkine Thanks for the update. I don't think there should be anything left. If you would, for some quick smoke testing would be healthy to verify that an AI vs AI game plays cleanly. I'll hopefully do a final pass tomorrow (about 14 hours from now)

Done. Played a game of Revised with a combination of Hard/Fast/Easy AIs (but mostly hard). Went to Round 17 with Axis winning.

game-core/src/test/java/games/strategy/triplea/ai/pro/util/MoveBatcherTest.java
```diff
@@ -67,27 +71,28 @@ public void testMoveUnitsWithMultipleTransports() {
...
     // Check move to load all the land units onto transports.
-    assertEquals(
+    var expected =
         new MoveDescription(
             List.of(inf1, tank1, tank2, inf2),
             brazilToSz18,
-            Map.of(inf1, transport1, tank1, transport1, inf2, transport2, tank2, transport2)),
-        moves.get(0));
+            Map.of(inf1, transport1, tank1, transport1, inf2, transport2, tank2, transport2));
+    assertThat(moves.get(0), is(expected));
```
DanVanAtta: side-note: This is not a bad place where a string comment would help spell out why we expect this assertion to hold.

game-core/src/test/java/games/strategy/triplea/ai/pro/util/MoveBatcherTest.java
```diff
@@ -113,25 +118,26 @@ public void testTransportsPickingUpUnitsOnTheWay() {
     //   Load tank and two infantry onto the transports.
```
DanVanAtta: side-note, the 'favor using helper functions' could help reduce the length of this test and perhaps make it easier to understand, eg:

```
givenTransportsMovedIntoPosition();
givenTankAndInfantryLoadedOntoTransport();
givenTransportedUnitsAreMoved();
givenUnitsUnloaded();

assertThat(moves, hasSize(4));
```

PR #5505, opened 2019-11-17
AI unit purchase upgrade logic improvements

ron-murhammer:
Address some of the feedback here: https://forums.triplea-game.org/topic/1518/ai-interesting-moves-and-points-of-improvement-on-revised/11

Specifically: Round 18: Russia is fighting for its life. It decides to buy Cavalry instead of defensive units.

The AI now considers the strategic value of the factory territory and if its high (close to enemies) then prefer strong units rather than high movement units.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
@@ -2338,11 +2338,16 @@ class ProPurchaseAi {
     }
   }
 
+  /**
+   * Determine efficiency value for upgrading to the given purchase option. If the strategic value
+   * of the territory is low then favor high movement units as its far from the enemy otherwise
+   * favor high defense.
+   */
   private static double findUpgradeUnitEfficiency(
       final ProPurchaseOption ppo, final double strategicValue) {
-    return strategicValue >= 1
-        ? ppo.getAttackEfficiency() * ppo.getDefenseEfficiency() * ppo.getCost() / ppo.getQuantity()
-        : ppo.getAttackEfficiency() * ppo.getMovement() * ppo.getCost() / ppo.getQuantity();
+    final double multiplier =
+        (strategicValue >= 1) ? ppo.getDefenseEfficiency() : ppo.getMovement();
+    return ppo.getAttackEfficiency() * multiplier * ppo.getCost() / ppo.getQuantity();
   }
 
   private IntegerMap<ProductionRule> populateProductionRuleMap(
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
@@ -2269,37 +2269,34 @@ class ProPurchaseAi {
           break;
         }
 
-        // Determine best long range attack option (prefer air units)
-        ProPurchaseOption bestAttackOption = null;
-        double maxAttackEfficiency =
-            minPurchaseOption.getAttackEfficiency()
-                * minPurchaseOption.getMovement()
-                * minPurchaseOption.getCost()
-                / minPurchaseOption.getQuantity();
+        // Determine best upgrade option (prefer air units)
+        // TODO: ensure map has carriers or air unit has range to reach enemy land mass
+        ProPurchaseOption bestUpgradeOption = null;
+        double maxEfficiency =
+            findUpgradeUnitEfficiency(minPurchaseOption, placeTerritory.getStrategicValue());
         for (final ProPurchaseOption ppo : purchaseOptionsForTerritory) {
           if (ppo.getCost() > minPurchaseOption.getCost()
               && (ppo.isAir()
                   || placeTerritory.getStrategicValue() >= 0.25
                   || ppo.getTransportCost() <= minPurchaseOption.getTransportCost())) {
-            double attackEfficiency =
-                ppo.getAttackEfficiency() * ppo.getMovement() * ppo.getCost() / ppo.getQuantity();
+            double efficiency = findUpgradeUnitEfficiency(ppo, placeTerritory.getStrategicValue());
             if (ppo.isAir()) {
-              attackEfficiency *= 10;
+              efficiency *= 10;
             }
             if (ppo.getCarrierCost() > 0) {
               final int unusedLocalCarrierCapacity =
                   ProTransportUtils.getUnusedLocalCarrierCapacity(
                       player, t, placeTerritory.getPlaceUnits());
               final int neededFighters = unusedLocalCarrierCapacity / ppo.getCarrierCost();
-              attackEfficiency *= (1 + neededFighters);
+              efficiency *= (1 + neededFighters);
             }
-            if (attackEfficiency > maxAttackEfficiency) {
-              bestAttackOption = ppo;
-              maxAttackEfficiency = attackEfficiency;
+            if (efficiency > maxEfficiency) {
+              bestUpgradeOption = ppo;
+              maxEfficiency = efficiency;
             }
           }
         }
-        if (bestAttackOption == null) {
+        if (bestUpgradeOption == null) {
           airAndLandPurchaseOptions.remove(minPurchaseOption);
           continue;
         }
@@ -2327,16 +2324,25 @@ class ProPurchaseAi {
         placeTerritory.getPlaceUnits().removeAll(unitsToRemove);
         ProLogger.trace(t + ", removedUnits=" + unitsToRemove);
         for (int i = 0; i < unitsToRemove.size(); i++) {
-          if (resourceTracker.hasEnough(bestAttackOption)) {
-            resourceTracker.purchase(bestAttackOption);
+          if (resourceTracker.hasEnough(bestUpgradeOption)) {
+            resourceTracker.purchase(bestUpgradeOption);
             final List<Unit> newUnit =
-                bestAttackOption.getUnitType().create(bestAttackOption.getQuantity(), player, true);
+                bestUpgradeOption
+                    .getUnitType()
+                    .create(bestUpgradeOption.getQuantity(), player, true);
             placeTerritory.getPlaceUnits().addAll(newUnit);
             ProLogger.trace(t + ", addedUnit=" + newUnit);
           }
         }
       }
     }
+  }
+
+  private static double findUpgradeUnitEfficiency(
+      final ProPurchaseOption ppo, final double strategicValue) {
+    return strategicValue >= 1
+        ? ppo.getAttackEfficiency() * ppo.getDefenseEfficiency() * ppo.getCost() / ppo.getQuantity()
+        : ppo.getAttackEfficiency() * ppo.getMovement() * ppo.getCost() / ppo.getQuantity();
   }
 
   private IntegerMap<ProductionRule> populateProductionRuleMap(


asvitkine:
I'm not familiar with this logic - but if the goal is to upgrade units, wouldn't it be better to always choose strictly better units in general?

ron-murhammer:
@asvitkine This logic is used towards the end of the AIs purchase logic to try to use any remaining PUs (essentially replace buying cheaper units with more expensive ones). Previously it was trying to buy units that were more expensive, higher movement, higher attack, and prefer air. Now it does some basic check one if its close to an enemy front and if so then instead prefer more expensive, higher attack, higher defense, and prefer air (essentially care less about movement and more about defense).

asvitkine:
Thanks, I agree that this change is positive.

What I'm wondering about is - if the AI previously made a decision that it wanted say an infantry unit, it seems wrong to upgrade that to a unit that's not strictly better (like a cavalry unit), since whatever factors it used to decide on the infantry might be invalidated by that "upgrade". (i.e. why not always pick a unit that's >= for all three attributes of attack/defense/movement).

I guess one challenge is that Air units are never strictly better than ground, due to how they work (i.e. they can't remain in newly captured territory), but maybe we we can still have the logic allow for that (by not considering that aspect).

ron-murhammer:
@asvitkine Yeah the challenge is it really depends on the map and unit set. Some maps just don't have a lot of options for "strictly better". It also depends on why they built a certain unit (defense, attack, fodder, etc). And yes often air units will be a bit weaker in attack or defense. If for instance its looking to upgrade say a tank (3/3/2) that its placing a 5-6 territories from the front and I have the option of bomber (4/1/6) that would probably be a good choice. Now if I'm about to lose the factory that its being placed in then that would be a bad choice as I want more defense and care less about movement.

What I've learned, is the AI logic needs to be pretty flexible to handle all the various maps and unit sets so its better to avoid hard limits where possible.

asvitkine:
That makes sense. I guess ideally we'd have a way to encode the desired upgrade on the original purchase decision - so that when this logic happens, it can take that into account. But that would be a much larger change.

The other option is to first try to find strictly better, but still fall back to logic similar to this if it's not possible.

As an aside, would it make sense to try to add better test coverage as we're making these changes? e.g. since you're now adding logic to detect a specific condition (i.e. territory to defend), perhaps a test can set that up and check that it's working as intended?

ron-murhammer:
@asvitkine The difficult part about testing is creating the input data to test which in most cases is a fairly full fledged map situation. So generally the only complete data is starting map XMLs. At one point, I started collecting save games that I wanted to turn into an AI test suite but never got around to it. Some of the small utility methods can be tested but that is often checking a very small part of the overall AI. I'm open to ideas around testing AI code but its a bit of a unique animal.

DanVanAtta:
@ron-murhammer I'm certainly an advocate for unit testing, the benefits are correctness and design. Unit testing is good for finding bugs. I'm not sure if you've read "Working with Legacy Code", it's also extremely useful for making modifications. Let's say someone wanted to update that method, having existing unit tests could ensure that existing use-cases are not broken. 

In terms of AI testing, I think decision modularity is the key. It's not a bad regression/acceptance test to create certain scenarios and verify the AI decision making, but at scale that will not cover everything. What I mean is the permutations of possible decisions is too much to cover through that kind of testing, it's the integration scam pitfall: https://www.youtube.com/watch?v=VDfX44fZoMc

What might be better is to define decision modules where we can better control the input decisions and then verify the outputs. I would envision that to complement unit testing so we know we have correctness at the unit level, then using those units/strategy objects we can have decision modules to operate on those and create rule sets that could be verified more easily.

ron-murhammer:
@DanVanAtta I don't disagree with you but I think its a bit harder than it seems on the surface. The AI is generally less about making incremental decisions and more about making incremental scores/calculations. But I do agree that there are probably around 3 levels of AI testing:
- Unit testing (method level)
- Calc/decision level (given this scenario what does the AI decide or score around factor X)
- Overall AI phase test (given this scenario what does the AI purchase, move, attack, place, etc)

IMO, the one that would actually be the most meaningful to start with is the last one. It won't cover all permutations but even covering a small number of scenarios and walking through most of the AI code would provide a large benefit.

So in this example, the utility method is calculating a unit upgrade efficiency score for each potential purchase option to determine which one it thinks would be best. In theory the 3 levels of tests would be:
- Test that a unit upgrade efficiency is properly calculated for various inputs
- Test that given a scenario with this map, these units to purchase, and this remaining PUs what units are upgraded
- Test the overall AI purchase phase given a scenario

game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
```diff
@@ -2269,37 +2269,34 @@ private void upgradeUnitsWithRemainingPUs(
...
 
-        // Determine best long range attack option (prefer air units)
-        ProPurchaseOption bestAttackOption = null;
-        double maxAttackEfficiency =
-            minPurchaseOption.getAttackEfficiency()
-                * minPurchaseOption.getMovement()
-                * minPurchaseOption.getCost()
-                / minPurchaseOption.getQuantity();
+        // Determine best upgrade option (prefer air units)
+        // TODO: ensure map has carriers or air unit has range to reach enemy land mass
```
DanVanAtta: nit, missing the TODO "tag"

game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
```diff
@@ -2339,6 +2338,13 @@ private void upgradeUnitsWithRemainingPUs(
     }
   }
 
+  private static double findUpgradeUnitEfficiency(
```
DanVanAtta: Any objection to javadoc'ing this method? I'm thinking capturing the context of what we are doing, the motivation/philosophy behind this calculation would be useful context to maintain going forward? 

For example, it's a bit curious why/when we can swap defense efficiency for movement

game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
```diff
@@ -2339,6 +2338,13 @@ private void upgradeUnitsWithRemainingPUs(
...
+      final ProPurchaseOption ppo, final double strategicValue) {
+    return strategicValue >= 1
+        ? ppo.getAttackEfficiency() * ppo.getDefenseEfficiency() * ppo.getCost() / ppo.getQuantity()
```
DanVanAtta: nit, there is some redundancy in the two conditional branches. Seems can be consolidated with something like:
```
double multiplier = (strategicValue >= 1) ? ppo.getDefenseEfficiency() : ppo.getMovement();
return ppo.getAttackEfficiency() * multiplier * ppo.getCost() / ppo.getQuantity();
```
What do you think?

DanVanAtta:
@ron-murhammer left some review comments. 

I lost my response to your last comment, I think we agree to a good extent.

W.r.t:
```
    Unit testing (method level)
    Calc/decision level (given this scenario what does the AI decide or score around factor X)
    Overall AI phase test (given this scenario what does the AI purchase, move, attack, place, etc)
```
Adding tests in reverse order given above is a reasonable approach for working with a legacy code base. The high level tests, while not thorough, give enough safety that the a person can start drilling down and retro-fitting code to have tests. On the other hand though, just retro-fitting code with unit tests is not super valuable. It's also a good idea for us to try and structure updates in multiple phases if we can so we can get those tests harnesses, then update the logic and test in tandem in addition to ensuring that new code meets the quality bar we'd want (and I'd call out that updating existing code that is not tested is not refactoring, it's a re-write, it's new code).

game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
ron-murhammer: Not really relevant here as this comment isn't directly related to this PR. Just something I noticed.

game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
ron-murhammer: Added.

game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
ron-murhammer: Updated.

game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
DanVanAtta: In this case, a user name is maybe a good tag. I don't think anyone else can complete this TODO. 
More options:
- create an AI improvements project, can use the project number.
- Generic "AI-IMPROVEMENTS"  tag

game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
```diff
@@ -2338,11 +2338,16 @@ private void upgradeUnitsWithRemainingPUs(
     }
   }
 
+  /**
+   * Determine efficiency value for upgrading to the given purchase option. If the strategic value
```
DanVanAtta: :+1:

game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
ron-murhammer: Given all the AI code is packaged together, I don't a TODO tag really adds anything as you can just search for all the TODOs in the AI code.

DanVanAtta: Maybe this item should go to your personal backlog? Are you planning on fixing this imminently? Maybe "AI" is a good enough tag.

How can a person easily refine a search to just AI code? Should we be forced to go through such an exercise if looking for TODO statements that are 'orphaned'? At some point soon, all TODOs without a tag would/should probably get dropped. We've many like this where only the person that wrote them knows what needs to be done, sometimes even the original author doesn't even have that context anymore.

ron-murhammer: Go to games.strategy.triplea.ai package. I'd like to think I could speak to all the TODOs in the AI code since I probably wrote all of them :)  Honestly for now, I'd say unless there is a PR/issue to tie a TODO to, I don't think adding some filler tag is worth it. At some point, if we do that clean up, I'm going to have to review all the existing AI TODOs anyways.

DanVanAtta: > Go to games.strategy.triplea.ai package

If we do a search for TODOs at a top level, we'd still have to filter out the package. 

I'm also concerned we're eliding a convention. It seems "except for AI" code to be a big exception. IMO it would be pretty simple to add an AI tag, my concern would be very satisfied, and we'd easily know the new vs old TODOs within that code. Any that are still current would presumably get such a tag. I'm also concerned we could easily wind up with "yup, that's current, yup, that too. Will I ever get to any of them? Probably not. Can anyone else pick up this TODO and fix it? no". Seems like we would not be actually fixing the problem of dangling TODOs that we had decided we wanted to stop making worse.

DanVanAtta: We can revisit the convention in this case. I've also found there are gaps to it. 
The convention as written was not meant to be an advisory, if so, it would have been a trivial suggestion, which is not something worth even writing down as a convention.

https://github.com/triplea-game/triplea/wiki/Code-Conventions#todo-comments-attach-a-tracking-token-5249


Here is an issue to revisit the convention: https://github.com/triplea-game/triplea/issues/5549

If we go forward, we could be saying that it's okay to push forward on conventions where we do not agree and then we revisit them. IMO it would be worth just placating a silly convention just for the sake that we can say that we follow conventions when they are written down.

With that said, our focus probably should be on issue #5549 to see how we can revisit the original problem we were trying to solve.

DanVanAtta:
FYI, on merge here, I updated the commit comment body to include the description from the CR:
"The AI now considers the strategic value of the factory territory and if its high (close to enemies) then prefer strong units rather than high movement units."

Issue #5500, opened 2019-11-17
Latest Version Check Log Message When Launching From IDE


ron-murhammer:
I see this in the logs when launching pre-release from IDE:
```
Nov 16, 2019 10:56:32 PM org.triplea.live.servers.LiveServersFetcher latestVersion
INFO: (No network connection?) Failed to get server properties
java.io.IOException: Unexpected status code (404)
	at games.strategy.engine.framework.map.download.ContentDownloader.<init>(ContentDownloader.java:64)
	at games.strategy.engine.framework.map.download.ContentDownloader.<init>(ContentDownloader.java:43)
	at org.triplea.live.servers.LiveServersFetcher.lambda$0(LiveServersFetcher.java:29)
	at org.triplea.live.servers.FetchingCache.fetchLiveServers(FetchingCache.java:53)
	at org.triplea.live.servers.FetchingCache.get(FetchingCache.java:31)
	at org.triplea.live.servers.FetchingCache.get(FetchingCache.java:1)
	at org.triplea.live.servers.LiveServersFetcher.latestVersion(LiveServersFetcher.java:44)
	at games.strategy.engine.auto.update.EngineVersionCheck.checkForLatestEngineVersionOut(EngineVersionCheck.java:28)
	at games.strategy.engine.auto.update.UpdateChecks.checkForUpdates(UpdateChecks.java:26)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

DanVanAtta:
The live file is referenced by a typo 'yaml' vs 'yml' suffix.
We could have avoided this by pushing the configuration file to live first, but that would have been a bit odd/risky as the file format was not fully vetted and it would have been unused.. 

#5506 has a fix for this problem and updates the build to verify we won't have any bad links. Interestingly a number of links had a redirect and we had a second one that was also broken.

PR #5506, opened 2019-11-18
Check links

DanVanAtta:
### Add shell script to verify links in UrlConstants.java are valid
    
    If we have any invalid links in UrlConstants.java, after this
    update we'll get a build failure in the static analysis step
    of the travis build.

###  Fix links in UrlConstants
    
    - Fix two broken links:
      - yaml vs yml typo
      - second is a removed page from 'github/../docs', updated page is in wiki
    - Fix redirect:
      - http->https
      - https://www.triplea-game.org -> https://triplea-game.org

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[x] Bug fix: #5500 
[] Other:   

## Testing

[x] Manually testing done

- Verified we would see a build failure with invalid links: https://travis-ci.org/DanVanAtta/triplea/builds/613252556

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,7 @@ jobs:
       script:
         ## shellcheck all shell scripts
         - find .travis/ -type f | xargs grep -lE "^#\!/bin/bash$" | xargs shellcheck
+        - .travis/check-links || travis_terminate 1;
         - ./gradlew --no-daemon --parallel validateYamls spotlessCheck checkstyleMain checkstyleTest pmdMain
     - stage: verify
       env: DESCRIPTION="JDK11 integTest"
--- a/.travis/check-links
+++ b/.travis/check-links
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+set -eu
+
+URI_FILE=$(find . -type f -path "*/src/main/java/*" -name "UrlConstants.java")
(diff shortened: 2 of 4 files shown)


DanVanAtta:
Agree there is a degree of brittleness. We need to weigh that against the brittleness of the links themselves. It was as surprise that half of our links were redirected and we had another one that was broken.

If we ever get a link that needs to be multi-line, perhaps we can use a properties file. I suppose ultimately it's not a problem until it is, historically we have not changed/added links too frequently. Ideally we would not bake in too many links at all.

> If we ever rename the class, this will fail. This is avoidable and can be fixed rather fast.

It should be pretty easy to identify this and rename the file that we are targetting. The path is not hardcoded, so we can move it around freely without worry.

> So I'm not really sure if using grep to do this check is the most reliable idea.

Technically sed :grin: , but same point. We could remove concatenations with sed as well if push really comes to shove, best to keep it simple though.

> If you're crazy enough you could compile this class manually and write other java code than goes through all fields via reflection and checks the values/prints them for curl to do the same.

Indeed we could do this in java code, I'm a bit concerned about the ROI. If we see any issues down the road, this might be the thing to do really, and put that in a test.

Issue #5487, opened 2019-11-16
Auto SpotlessApply (?)


DanVanAtta:
Builds failing because spotlessApply was not run are pretty common and a bit annoying to see that kind of failure. We could set up travis to auto-run it and commit and changes. If the 'merge conflict' is bad it's not too hard to force push over the auto-committed change, and usually it's rare for a branch to need modification once it's been pushed and is up for PR.

Thoughts?

RoiEXLab:
Good idea.
Since spotless was introduced I really got into the habit of not caring about formatting at all, so I usually run the spotless task from intelliJ just before commiting (If I would use IntelliJ for commits as well, I could even configure that as a pre-commit hook, but I really prefer my other GUI solution right now), but there are times where I just forget this, so a clear win IMO

DanVanAtta:
@RoiEXLab as you noted, we can't push back to a branch from a PR with any kind of secret token. That seems to torpedo this idea. We could have a bot do a follow-up PR to correct any issues, but that would get very messy.

PR #5507, opened 2019-11-18
Start creating an auto-formatter

RoiEXLab:
Another shot at #5487

### DIFF
--- a/.github/workflows/auto-format.yml
+++ b/.github/workflows/auto-format.yml
@@ -14,7 +14,7 @@ jobs:
       with:
         java-version: 11
     - name: Run Spotless
-      run: ./gradlew spotlessApply
+      run: ./gradlew --parallel spotlessApply
     - name: Commit changes
       uses: EndBug/add-and-commit@v2.1.0
       with:
@@ -25,4 +25,4 @@ jobs:
         pattern: "*.java"
         force: false
       env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # Leave this line unchanged
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/auto-format.yml
+++ b/.github/workflows/auto-format.yml
@@ -1,0 +1,28 @@
+name: Auto-Formatting
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK 11
+      uses: actions/setup-java@v1
+      with:
+        java-version: 11
+    - name: Run Spotless
+      run: ./gradlew spotlessApply
+    - name: Commit changes
+      uses: EndBug/add-and-commit@v2.1.0
+      with:
+        author_name: tripleabuilderbot
+        author_email: tripleabuilderbot@gmail.com
+        message: "Auto-Formatting"
+        path: "."
+        pattern: "*.java"
+        force: false
+      env:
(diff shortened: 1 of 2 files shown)


RoiEXLab:
@DanVanAtta Ok, this worked better than expected. I really just ended up using the yaml I posted in https://github.com/triplea-game/triplea/pull/5498#issuecomment-554765530
Feel free to add some steps that also check the lombok annotations etc. using the bash scripts you created for #5498 in a follow-up PR.
Have a look at the gradlew step in the yaml, it should be the whole bash shell you can use.

.github/workflows/auto-format.yml
```diff
@@ -0,0 +1,28 @@
...
+      uses: EndBug/add-and-commit@v2.1.0
+      with:
+        author_name: tripleabuilderbot
+        author_email: tripleabuilderbot@gmail.com
+        message: "Auto-Formatting"
+        path: "."
+        pattern: "*.java"
+        force: false
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # Leave this line unchanged
```
DanVanAtta: Where are the secrets defined?
Perhaps it would make sense to do a wiki write-up to document how the github actions/workflow is configured.

.github/workflows/auto-format.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK 11
+      uses: actions/setup-java@v1
+      with:
+        java-version: 11
+    - name: Run Spotless
+      run: ./gradlew spotlessApply
```
DanVanAtta: We may want `--parallel` here in case it speeds up this action

RoiEXLab:
>     * do you know if github actions are kicked off in parallel to a travis build, or after?

The actions are completely independent of travis, so parallel, just like LGTM or the synk check
>     * do you know if we can sequence workflow files, or does github actions pick the ordering for us?

All "actions" are completely independent of each other, so every yaml file gets their own docker environment, if you want some actions performed one after another, you'll have to create dedicated steps in a single file, not sure how much configuration is allowed there.
I read something about build matrix functionality, but I don't know how exactly it works.

>     * are actions just a matter of dropping in a new file into `workflows`?

Yes.

>     * is there a way to extract common steps so we are not defining the JDK across each file?

I don't know actually. I haven't found a way so far.
However in the repository settings you can configure env variables, that you could use in every yaml file, so that way you could synchronize the main version, but you'll have to define it each time.

.github/workflows/auto-format.yml
RoiEXLab: In the settings of this repo.
`secrets.GITHUB_TOKEN` however is always present, and is an exception to the other secret tokens: It's present even in PRs

.github/workflows/auto-format.yml
RoiEXLab: Good idea. AFAIK the container is allowed to use 2 cores.

DanVanAtta:
Found this:

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobs

> A workflow run is made up of one or more jobs. Jobs run in parallel by default. To run jobs sequentially, you can define dependencies on other jobs using the jobs.<job_id>.needs keyword.

.github/workflows/auto-format.yml
RoiEXLab: The documentation can be found here:
https://help.github.com/en/actions/automating-your-workflow-with-github-actions

I can only tell you what I read there...

DanVanAtta:
Opened up an issue so we can continue the discussion of setting up github actions CI

Issue #5344, opened 2019-10-13
Update unit scroller images and make them look good


DanVanAtta:
Anyone wanting to pick this up is welcome, otherwise I'll pick this back up perhaps in a week.

Task 1: Update unit scroller images with images containing a drop shadow: https://forums.triplea-game.org/post/27360

Task 2: if the images do not look good on mac, with a light colored theme, then we need to solve with likely one of these two approaches:
A) add some logic to invert colors
B) add some logic to use an alternative icon which will have inverted colors

DanVanAtta:
- Buttons are now narrow enough to not be cut-off on the right hand side on initial map render
- Updated the button images

![Screenshot from 2019-11-17 16-23-32](https://user-images.githubusercontent.com/12397753/69016779-a4815c00-0956-11ea-9c2b-e495d11fb03d.png)

DanVanAtta:
Napoleonic empires has a mini-map that is even more narrow than revised.

I'm thinking adding another 20px or so to the side-bar is probably a better fix at this point.

The unit move history also easily gets cut-off and adds scrollbars:
![Screenshot from 2019-11-17 18-10-06](https://user-images.githubusercontent.com/12397753/69019684-84599900-0966-11ea-9694-9696fd03224e.png)

DanVanAtta:
Napoleonic empires has a mini-map that is even more narrow than revised.

I'm thinking adding another 20px or so to the side-bar is probably a better fix at this point.

The unit move history also easily gets cut-off and adds scrollbars:
![Screenshot from 2019-11-17 18-10-06](https://user-images.githubusercontent.com/12397753/69019684-84599900-0966-11ea-9694-9696fd03224e.png)

RoiEXLab:
Can this be closed now that #5509 is merged?

DanVanAtta:
yes, the minimum requirements of drop-shadows and not being cut-off seem to be met.

PR #5509, opened 2019-11-18
Set min-size of action panel to 240px

DanVanAtta:
Allows enough space for unit-scroller and unit-move-history
to typically not be cut-off on the right hand side.

Additionally, if there is not enough room for unit scroller,
re-sizing will yield a min-size, so you will not be able to get
the action panel back to it's original (cut-off) width.

This update sets the min width of the action panel to match the
effective min width.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[x] Manually testing done

- checked action panel sizing on napoleanic wars (has perhaps the smallest known mini-map, which
typically controls the initial width of the action panel)

### After
Screenshot from 2019-11-17 19-15-52

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/ActionPanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/ActionPanel.java
@@ -3,6 +3,7 @@ package games.strategy.triplea.ui;
 import games.strategy.engine.data.GameData;
 import games.strategy.engine.data.PlayerId;
 import games.strategy.triplea.Properties;
+import java.awt.Dimension;
 import java.util.concurrent.CountDownLatch;
 import javax.swing.BoxLayout;
 import javax.swing.JPanel;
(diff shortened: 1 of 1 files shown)


DanVanAtta:
@ron-murhammer I'm not sure we will get that resolved very quickly. We may be hitting the definition of bike-shedding. Please check out the screenshot in: https://github.com/triplea-game/triplea/issues/5344, the unit history images are cut-off. It does not look like the game was really ever intended to have minimaps less than 250px. I'd recommend we move forward with this fix and let people play-test it and give their feedback. 

From your own suggest: https://forums.triplea-game.org/post/27292

> redrum Admin 2 months ago

> @LaFayette Yeah, I believe the right panel width is solely based on the minimap width for the map. Probably should just consider setting a hard minimum which the unit scroller requires. Though to @Cernel point, seeing if there are some ways to make its required width a bit smaller would be good.

Please note work was done to make the unit-scroller more narrow: https://github.com/triplea-game/triplea/pull/5473, we could go to 32x32x icons, but they do not look good. Considering that the unit history gets cut-off when the action panel is less than 250px, the hard minimum seems reasonable and fixes multiple problems. If we do not go with this hard minimum, then we'll leave the unit history as a problem.

I was tempted to also put the unit scroller in a scrollable area, but it seems like the usability would be very poor.

Again, I'd recommend we merge, I'll update the forum post with the latest, we can get play-tested feedback, and then it's a one-line change if we want to go back or alter it.

PR #5510, opened 2019-11-18
Unit Scroller Updates: "station", rename images + wake-all button

DanVanAtta:
1. Rename "sleep" concept to "station"
2. Update button names, slightly simpler names and use "station.png"
3. Add a wake-all button to activate stationed and skipped units.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[x] Manually testing done

- launched a napoleonic wars map:
 -  skipped all units, then waked them
 - did a mix of station+skip units, then waked them
 - skipped+stationed unit in one move phase, proceeded to next move phase, verified wake-all

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/unit/scroller/UnitScroller.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/unit/scroller/UnitScroller.java
@@ -67,7 +67,6 @@ public class UnitScroller {
   private static final String WAKE_ALL_TOOLTIP =
       "Press 'w' or click this button to activate all skipped or stationed units";
 
-
   private static final String HIGHLIGHT_MOVABLE_TOOLTIP =
       "Press 'F' key or click this button to highlight movable units";
   private static final String FLAG_TOGGLE_BUTTON_TOOLTIP =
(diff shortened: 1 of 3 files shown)


ron-murhammer:
@DanVanAtta I personally prefer "sleep" over "station" as I think its more commonly used in other games. Also "wake" is clearly the opposite of "sleep" but not really clearly the opposite of "station". Where did the forum thread end up on this?

DanVanAtta:
I'm afraid to say the forum thread wound up with analysis paralysis and deadlocked. Once forward progress is made, I'm guessing there will be some more momentum and feedback. Most people do not want to get involved and too much time has been spent going back and forth.

I was not inclined to change from 'sleep'. The updated icons, which have already been merged, uses a 'sun' instead of a lantern for highlight. That creates a false contrast with the 'moon' for sleep, so that was changed to a shield. 'sleep' + shield do not really jive, hence the term 'station'.

I'd recommend we take that up in the forum thread and see how it pans out. The original sleep discussion dragged out for weeks and was ultimately distracting. Let's please put a pin in the terminology decision and make a final decision when 2.0 is released.

PR #5511, opened 2019-11-18
Add a workflow file to run unit-tests

DanVanAtta:
## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] Manually testing done

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -1,0 +1,15 @@
+name: Auto-Formatting
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK 11
+      uses: actions/setup-java@v1
+      with:
+        java-version: 11
+    - name: Run unit tests
+      run: ./gradlew --parallel testJava
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -1,7 +1,16 @@
 name: Auto-Formatting
 
+on:
+  pull_request:
+    types: opened
+    jobs:
+      rebase:
+        - uses: actions/checkout@master
+        - name: Automatic Rebase
+          uses: cirrus-actions/rebase@master
+          env:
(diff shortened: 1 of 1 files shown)


.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,29 @@
...
+    types: opened
+    jobs:
+      rebase:
+        - uses: actions/checkout@master
+        - name: Automatic Rebase
+          uses: cirrus-actions/rebase@master
+          env:
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+on: [push]
+jobs:
```
RoiEXLab: @DanVanAtta You might already have figured this out on your own, but each workflow (yaml file) can only have global triggers for all jobs.
Besides that you're defining the value for `on` 2 times, which is invalid yaml spec-wise.
However you are able to define all the triggers at once, and add an `if` key to the individual jobs to prevent them from running if the false trigger is the cause.

See https://help.github.com/en/actions/automating-your-workflow-with-github-actions/contexts-and-expression-syntax-for-github-actions#github-context and https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idif

DanVanAtta:
One thing I'm still wondering, how to get the actions output to appear in the PR. 
Is that a function of the `on` condition? We may want to revisit the 'push' for the spotlessApply in that case, personally I only want that to run when a PR is open. I very often push branches just to make a backup and get them off my machine. These intermediary branches do not need to be updated.

RoiEXLab:
@DanVanAtta Well it's displayed in the checks tab for PR related triggers:
Screenshot

However I noticed that the auto-format build doesn't trigger when the latest commit was created by the build step itself, presumably to avoid infinite commit loops.

> Is that a function of the `on` condition?

I guess so.

> We may want to revisit the 'push' for the spotlessApply in that case, personally I only want that to run when a PR is open.

Good point, I was thinking about excluding master, but this restriction makes more sense after all.

PR #5528, opened 2019-11-18
Change trigger for auto-formatting

RoiEXLab:
Per #5511
TEST


RoiEXLab:
Ok, the trigger is not what I was searching for

PR #5529, opened 2019-11-18
Check different trigger

RoiEXLab:
Second attempt at #5511


RoiEXLab:
Ok, this should work now.
Apparently "synchronize" is the PR type that is triggered whenever a new commit is pushed to a PR.

RoiEXLab:
Ok, now the branch is on a merged state, so the commit is no longer done on the original branch.
I don't have the time to look into this right now, but maybe @DanVanAtta has an idea on how to fix this problem.

Issue #5524, opened 2019-11-18
Convention? Favor Collections.emptyList() over List.of()


DanVanAtta:
Favor:
```
Collection<String> empty = Collections.emptyList();
```
Instead of:

```
Collection<String> empty = List.of();
```

--------------

### Decision

Favor `List.of()`

ron-murhammer:
No strong opinion either way from me.

DanVanAtta:
Thanks!
Me and @asvitkine have weighed in, any thoughts on this @RoiEXLab & @ssoloff ?

asvitkine:
One other point:

I didn't address your comparison to .isEmpty() vs. == 0 earlier, let me do so now. In my mind, I don't think it's the same.

Some of the reasons to prefer .isEmpty() aren't relevant here:
 1. When not using .isEmpty(), you have to then debate whether it should be ".size() == 0" or ".size() < 1" (and ".size() > 0" or ".size() != 0" for the inverse). Without a convention, all of those end up being used in a codebase. So isEmpty() eliminates that decision/inconsistency with the alternative. This doesn't apply for List.of() since there's only one way to do List.of() for an empty list.
 2. The ".size() == 0" check is more complex - it includes a method call, an operator and a value to compare against. Whereas isEmpty() is just a method call - so simpler at the callsite. Again, List.of() doesn't have any issue here - and can be argue to be simpler at the callsite since it's shorter than the alternatives.

So given those 2 points, I don't consider drawing the analogy List.of() to ".size() == 0" as fair - the latter has very clear reasons why you'd want to prefer isEmpty(), that don't apply to List.of().

DanVanAtta:
Good points @asvitkine . I'm curious if you have any comment to:

- equivalence between `List.of()` and `List.of(new String[0])`. This is bothersome as seems we have an unused arg. Variatic args, my impression, are typically for 1 or more args, zero args case seems a bit unnatural.

- `List.of()` is same as `Arrays.asList()`, previously `emptyList()` was still preferred, why would the `List.of()` API change that? Why have a `emptyList()` api at all if there was `Arrays.asList()`

DanVanAtta:
FWIW, some counts on how many we have of each:
```
dan@dan-desk:~/work/triplea$ find . -type f -name "*java" | xargs grep -E "List.of\\(\\)|Set.of\\(\\)|Map.of\\(\\)"  | wc -l
39
[...]

PR #5534, opened 2019-11-19
Use Map.of/List.of/Set.of over Collections and Arrays.asList

DanVanAtta:
As discussed in: https://github.com/triplea-game/triplea/issues/5524

Favor using List.of instead Collections.emptyList, Collections.singletonList,
and Arrays.asList.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[x] Manually testing done

- Ran through 2 AI vs AI games 

## Additional Review Notes

- Not entirely safe as `Arrays.asList` is mutable.  The manual testing hopefully found everywhere we would have a problem. I've some concern we may see issues in the file menus, with luck we'll find those, if any, with further prerelease testing
- Not 100% of usagse of emtpyList/emptySet were converted, some tests relied on being able to add `null`, in those cases those tests were left alone.

### DIFF
--- a/game-core/src/test/java/games/strategy/triplea/delegate/WW2V3Year41Test.java
+++ b/game-core/src/test/java/games/strategy/triplea/delegate/WW2V3Year41Test.java
@@ -82,9 +82,7 @@ import games.strategy.triplea.delegate.data.TechResults;
 import games.strategy.triplea.delegate.remote.IAbstractPlaceDelegate;
 import games.strategy.triplea.xml.TestMapGameData;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -364,7 +362,7 @@ class WW2V3Year41Test {
     addTo(british(gameData), transport(gameData).create(1, british(gameData)), gameData);
     final String error =
         del.placeUnits(
-            Collections.emptyList(),
+            List.of(),
             territory("United Kingdom", gameData),
             IAbstractPlaceDelegate.BidMode.NOT_BID);
     assertNull(error);
@@ -1009,7 +1007,7 @@ class WW2V3Year41Test {
             AbstractMoveDelegate.getBattleTracker(gameData).getPendingBattle(attacked);
     final List<String> steps = battle.determineStepStrings(true);
     assertEquals(
-        Arrays.asList(
+        List.of(
                 attacker + SUBS_SUBMERGE,
                 defender + SUBS_SUBMERGE,
                 attacker + SUBS_FIRE,
@@ -1051,7 +1049,7 @@ class WW2V3Year41Test {
             AbstractMoveDelegate.getBattleTracker(gameData).getPendingBattle(attacked);
     final List<String> steps = battle.determineStepStrings(true);
     assertEquals(
-        Arrays.asList(
+        List.of(
                 defender + SUBS_SUBMERGE,
                 defender + SUBS_FIRE,
                 attacker + SELECT_SUB_CASUALTIES,
@@ -1099,7 +1097,7 @@ class WW2V3Year41Test {
             AbstractMoveDelegate.getBattleTracker(gameData).getPendingBattle(attacked);
     final List<String> steps = battle.determineStepStrings(true);
     assertEquals(
-        Arrays.asList(
+        List.of(
                 attacker + SUBS_SUBMERGE,
                 attacker + SUBS_FIRE,
                 defender + SELECT_SUB_CASUALTIES,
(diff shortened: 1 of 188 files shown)


asvitkine:
Is it possible to wait on this until my move refactor pr is landed? Presumably this change should be easy to rebase since it's just a search and replace and spotless apply.

Also one case to be careful of is that Collections.singletonList(null) is allowed but not List.of(null).

DanVanAtta:
@asvitkine we try to review/merge in order to add predictability to conflicts. With that said, we have not really tested that policy with WIP PRs. If having to choose, I would generally pick the PR that touches more files. While you may see conflicts in a dozen lines, this PR is very likely to see conflicts in hundreds of lines. If we try a package move while this update is in-flight, the conflicts will be pretty killer.

This update was a bit more than just search and replace, missing imports had to be fixed by hand. We had plenty examples where the `List` import was not present, eg: `Collection<> someCollection = Collections.emptyList()`

> Also one case to be careful of is that Collections.singletonList(null) is allowed but not List.of(null).

Indeed, we had one example of that. Using null values in lists is pretty archaic, happily there was only the one example to fix. The update from `Arrays.asList` to `List.of` was a big concern as we're going from mutable to immutable. Note though, I had the AI play against itself for about an hour before cutting this PR.

> Is it possible to wait on this until my move refactor pr is landed?

I checked for merge conflicts, it does not conflict.

asvitkine:
> @asvitkine we try to review/merge in order to add predictability to conflicts. With that said, we have not really tested that policy with WIP PRs. If having to choose, I would generally pick the PR that touches more files. While you may see conflicts in a dozen lines, this PR is very likely to see conflicts in hundreds of lines. If we try a package move while this update is in-flight, the conflicts will be pretty killer.
> 
> This update was a bit more than just search and replace, missing imports had to be fixed by hand. We had plenty examples where the `List` import was not present, eg: `Collection<> someCollection = Collections.emptyList()`
> 
> > Also one case to be careful of is that Collections.singletonList(null) is allowed but not List.of(null).
> 
> Indeed, we had one example of that. Using null values in lists is pretty archaic, happily there was only the one example to fix. The update from `Arrays.asList` to `List.of` was a big concern as we're going from mutable to immutable. Note though, I had the AI play against itself for about an hour before cutting this PR.
> 
> > Is it possible to wait on this until my move refactor pr is landed?
> 
> I checked for merge conflicts, it does not conflict. 👍

Okay, sounds good. Thanks for checking!

DanVanAtta:
@asvitkine thanks for taking the look! I'm not sure how to handle your feedback. There are 500+ locations changed, I don't think it's reasonable for us to fix up every one and them and have full confidence nothing is wrong. TripleA is super brittle, the chance of this being reverted if we do that would be pretty high, and then it would be for nought. Even if we did make all updates we could, the scope could increase by quite a bit (starts us down a rabbit hole), and/or we might wind up spending a lot more time testing.

We could add TODO statements to places where we maybe a collection copy is not needed, but I would be concerned we'd add them in locations where we're not sure and are speculating. I think overall we should try to follow-up with further clean-ups. The existing scope was really to remove the empty lists, I already spent 2 and half hours on this update, the ROI of spending another two hours is kinda poor. 

Do you have any thoughts or suggestions of good ways to make sure we don't lose the review feedback and successfully follow-up?

asvitkine:
> @asvitkine thanks for taking the look! I'm not sure how to handle your feedback. There are 500+ locations changed, I don't think it's reasonable for us to fix up every one and them and have full confidence nothing is wrong. TripleA is super brittle, the chance of this being reverted if we do that would be pretty high, and then it would be for nought. Even if we did make all updates we could, the scope could increase by quite a bit (starts us down a rabbit hole), and/or we might wind up spending a lot more time testing.
> 
> We could add TODO statements to places where we maybe a collection copy is not needed, but I would be concerned we'd add them in locations where we're not sure and are speculating. I think overall we should try to follow-up with further clean-ups. The existing scope was really to remove the empty lists, I already spent 2 and half hours on this update, the ROI of spending another two hours is more than I'm willing to invest.
> 
> Do you have any thoughts or suggestions of good ways to make sure we don't lose the review feedback and successfully follow-up?

That's fair. I'm fine if we want to do it in a follow up. I can send a PR if you want after you land this.

game-core/src/test/java/games/strategy/triplea/delegate/WW2V3Year41Test.java
```diff
@@ -1166,7 +1164,7 @@ void testAttackDestroyerAndSubsAgainstSubAndDestroyer() {
         invocation -> {
           final Collection<Unit> selectFrom = invocation.getArgument(0);
           return new CasualtyDetails(
-              Collections.singletonList(selectFrom.iterator().next()), new ArrayList<>(), false);
+              List.of(selectFrom.iterator().next()), new ArrayList<>(), false);
```
asvitkine: List.of() can be used for the next param as well. I checked CasualtyDetails and it makes a copy.

DanVanAtta:
> That's fair. I'm fine if we want to do it in a follow up. I can send a PR if you want after you land this.

That would be really great @asvitkine! I can certainly help review it

PR #5539, opened 2019-11-19
Invert deployment ordering of nginx and dropwizard

DanVanAtta:
Nginx fronts dropwizard (http-server) transparently, we need dropwizard to be
running for the nginx forwarding to work. Nginx depends on dropwizard,
not the other way round. This update inverts deployment ordering so that
dropwizard is deployed before nginx.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[] Manually testing done

### DIFF
--- a/infrastructure/ansible/site.yml
+++ b/infrastructure/ansible/site.yml
@@ -31,8 +31,8 @@
   strategy: free
   roles:
     - java
+    - http_server
     - nginx
-    - dropwizard
 
 - hosts: botHosts
   gather_facts: no


infrastructure/ansible/site.yml
```diff
@@ -31,8 +31,8 @@
   strategy: free
   roles:
     - java
+    - http_server
     - nginx
-    - dropwizard
```
RoiEXLab: I'm not sure if this is "re-ordering"

DanVanAtta: :grin:  The perils of moving "independent" updates. Here though the commit commentary was incomplete. This is both a fix and a re-order. Running this playbook yields an error right now. The 'dropwizard' role DNE. The role name comes from the sub-folder name in the 'roles' folder: https://github.com/triplea-game/triplea/tree/master/infrastructure/ansible/roles, so this must be 'http_server'

I've updated the commit comment to reflect the correction.

RoiEXLab: I figured something like that, thanks for clarifying ^^

DanVanAtta: :+1:

PR #5542, opened 2019-11-20
Websocket fixes for HTTPS & Disconnect Handling

DanVanAtta:
1. Fix disconnect window appearing when client closes the websocket connection.
This is done by splitting the concept of a connection lost listener and connection
closed listener. The closed listener is executed every time we close connection,
where client or server initiated. The connection lost listener is executed only
when the connection is closed without the client calling the close method. This
is done at the implementation level by removing connection lost listeners when
'close' is called.

2. Fix missing lobby shutdown on chat connection closed. Previously, if server
disconnected, the lobby would close but the game-headed application would continue
to run. With a shutdown call, the game-client lobby window now closes completely
and its background process, and the main game-headed initial screen is shown.

3. Fix protocol for secure websocket. Our current http client is able to convert
a http connection to websocket (ws) connection without issue. Converting an
'https' connection to 'wss' is not automatic and we must use the correct protocol
to begin with. We use 'http' for local development, 'https' for deployed server.
In this update, if we see a 'https' protocol, when establishing a websocket connection
we'll then update the protocol to be a 'wss' (secure websocket).

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Bug fix:  
[] Other:   

## Testing

[x] Manually testing done

- tested these updates in tandem with working on prerelease server and HTTPS.


http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
```diff
@@ -53,7 +59,14 @@ public GenericWebSocketClient(
...
         CompletableFuture.runAsync(client::connect, threadPool),
-        e -> log.log(Level.WARNING, "Failed to open connection with server", e));
+        e -> log.log(Level.SEVERE, "Failed to open connection with server", e));
+  }
+
+  @VisibleForTesting
+  static URI swapHttpsToWssProtocol(final URI uri) {
+    return uri.getScheme().equals("https")
+        ? URI.create(uri.toString().replace("https", "wss"))
+        : uri;
```
RoiEXLab: Why aren't http links just converted to `ws://`?

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
```diff
@@ -90,6 +109,8 @@ public void messageReceived(final String message) {
 
   @Override
   public void connectionClosed(final String reason) {
+    connectionClosedListeners.forEach(
+        connectionLostListener -> connectionLostListener.accept(reason));
     connectionLostListeners.forEach(
         connectionLostListener -> connectionLostListener.accept(reason));
```
RoiEXLab: Ok, what exactly is the difference between those 2 listeners?

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
DanVanAtta: I'm not entirely sure. The http upgrade request is very happy over `http`, but is not for `https`. Since it's a secure connection, `wss` is needed, `ws` also does not work either (maybe we could write a forwarding rule for that, seems simpler for a number of reasons to go directly with the correct protocol here)

DanVanAtta: The reason I went with an overwrite of the URI is to keep the 'override' configuration simpler, and also because this turned out to be a concern for only https.

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
DanVanAtta: Please let me know if these javadocs don't quite answer that question:
```
  /** These are called if connection is disconnected (by server). */
  private final Collection<Consumer<String>> connectionLostListeners = new ArrayList<>();
  /** These are called whenever connection is closed, whether by us or server. */
  private final Collection<Consumer<String>> connectionClosedListeners = new ArrayList<>();
```

RoiEXLab: Ah, sorry I should've been more clear. I get that this was the intended difference, but right now the only usage I see is in line 112, where the exact same thing is passed the connection lost listeners, the only difference being the order in which both are triggered.

So it currently looks like there is now difference from a code perspective.

DanVanAtta: The fix here is maybe not quite as clean as it could be (ie: slightly hacky), notice in "close" we clear the connection lost listeners.

This means if we have a 'connectionClosed' by the underlying socket, we'll call both listeners. If we call 'close' ourselves, then the 'connectionLost' listener set will be empty and we'll still call the 'connectionClosedListeners'.

RoiEXLab: Ahh, thanks for clarifying.
Maybe worth adding in a comment to explain this to future versions of ourselves?

DanVanAtta:
Fixed up conflicts and added some commentary. @RoiEXLab should be ready for for another pass assuming I didn't miss anything.

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
DanVanAtta: :grin:   7bbe6f9 
Done

PR #5547, opened 2019-11-20
Infrastructure UFW port configuration & remove Nginx cert generation

DanVanAtta:
1. Move UFW port management to individual roles

UFW ports were previously opened by using a group_var override that
would take effect in the 'security' role and open the right ports.
When a server has overlapping roles, eg: 'bot', 'lobby', the group_var
is not additive and becomes non-determistic whehter we will get 'bot'
or 'lobby' ports.

This update removes this mechanism and instead distributes the UFW
port management to each role that needs to open ports.

2. Remove NGINX task to generate certs

Certbot will be used to create certs, the cert generation task is
not needed.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[x] Configuration Change
[x] Bug fix:  
- Bot ports were not being opened.
[] Other:   

## Testing

[x] Manually testing done
- Verified as part of deploying to prerelease.triplea-game.org


RoiEXLab:
Some merge conflicts showed up

PR #5548, opened 2019-11-20
Update NGINX config for ciphers & websocket port forwarding

DanVanAtta:
- Add config to forward http upgrade request, enables websocket
- Fix proxy target port to refer to dropwizard 8080
- Update ciphers to secure set
  - From: https://cipherli.st/
  - Verified strong ciphers with:
    https://www.ssllabs.com/ssltest/analyze.html?d=prerelease.triplea-game.org
- Update cert locations to refer to locations generated by certbot

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Bug fix:  
[] Other:   

## Testing

[x] Manually testing done
- deployed to prerelease.triplea-game.org and verified connectiity
- used `wscat` to verify configurations for websocket connection

### DIFF
--- a/infrastructure/ansible/roles/nginx/templates/etc_nginx_sites_enabled_default.j2
+++ b/infrastructure/ansible/roles/nginx/templates/etc_nginx_sites_enabled_default.j2
@@ -6,16 +6,14 @@ server {
 server {
     listen 443;
     server_name {{ inventory_hostname }};
-
-    ssl_certificate           {{ nginx_ssl_certificate }};
-    ssl_certificate_key       {{ nginx_ssl_certificate_key }};
+    ssl_certificate /etc/letsencrypt/live/{{ inventory_hostname }}/fullchain.pem; # managed by Certbot
+    ssl_certificate_key /etc/letsencrypt/live/{{ inventory_hostname }}/privkey.pem; # managed by Certbot
 
     ssl on;
     ssl_session_cache  builtin:1000  shared:SSL:10m;
-    ssl_protocols  TLSv1.1 TLSv1.2;
-    ssl_ciphers HIGH:!aNULL:!eNULL:!EXPORT:!CAMELLIA:!DES:!MD5:!PSK:!RC4;
+    ssl_protocols  TLSv1.2 TLSv1.3;
+    ssl_ciphers EECDH+AESGCM:EDH+AESGCM;
     ssl_prefer_server_ciphers on;
-
     access_log            /var/log/nginx/java_server.access.log;
 
     location / {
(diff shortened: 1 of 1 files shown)


infrastructure/ansible/roles/nginx/templates/etc_nginx_sites_enabled_default.j2
```diff
@@ -6,16 +6,14 @@ server {
...
-    ssl_certificate_key       {{ nginx_ssl_certificate_key }};
+    ssl_certificate /etc/letsencrypt/live/{{ inventory_hostname }}/fullchain.pem; # managed by Certbot
+    ssl_certificate_key /etc/letsencrypt/live/{{ inventory_hostname }}/privkey.pem; # managed by Certbot
 
     ssl on;
     ssl_session_cache  builtin:1000  shared:SSL:10m;
-    ssl_protocols  TLSv1.1 TLSv1.2;
-    ssl_ciphers HIGH:!aNULL:!eNULL:!EXPORT:!CAMELLIA:!DES:!MD5:!PSK:!RC4;
+    ssl_protocols  TLSv1.2 TLSv1.3;
+    ssl_ciphers EECDH+AESGCM:EDH+AESGCM;
```
RoiEXLab: Excellent change, but it would be nice to see the other changes as well suggested on https://cipherli.st/ since we have the benefit of being able to control both the client and the server 👍 


I'm not sure if our current server supports TLS 1.3, depends on the Ubuntu version I guess.
In any case I'm pretty sure it doesn't work on the forum server.

infrastructure/ansible/roles/nginx/templates/etc_nginx_sites_enabled_default.j2
```diff
@@ -26,9 +24,16 @@ server {
       proxy_set_header        X-Forwarded-Proto $scheme;
 
       # Fix the “It appears that your reverse proxy set up is broken" error.
-      proxy_pass          http://localhost:4567;
+      proxy_pass          http://localhost:8080;
       proxy_read_timeout  90;
 
-      proxy_redirect      http://localhost:443 https://localhost:{{ http_server_port }};
+      proxy_redirect      https://localhost:443 https://localhost:{{ http_server_port }};
```
RoiEXLab: One thing to consider: This reverse proxy continues to be a proxy, so I believe lobby ip records etc. will point to the nginx proxy instead of the "actual source", so reintroducing the X-Forwarded-For logic might be necessary here, to let nginx tell the server which was the original IP.

DanVanAtta: Cool, let's keep as advisory and see what happens when it's deployed, we'll know pretty quickly if we need the x-forwarded port if the access-log IPs are incorrect.

infrastructure/ansible/roles/nginx/templates/etc_nginx_sites_enabled_default.j2
DanVanAtta: Should be easy to do, any objection if I add those in as a follow-up?

RoiEXLab: No objection, adding in the dhparams is requires an extra key generation step anyways, or making them static and putting it in the repository shouldn't be a problem either (see https://security.stackexchange.com/questions/94390/whats-the-purpose-of-dh-parameters )

PR #5550, opened 2019-11-20
Rename 'OddsCalculator' -> 'BattleCalculator'

DanVanAtta:
1. Rename OddsCalculator related classes to BattleCalculator
Players and the code both refer to the BattleCalculator as the BattleCalculator.

2. Fix collision with BattleCalculator class, rename to 'CasualtySelector'
- BattleCalculator was a utility class with various casualty selection methods
on it (as per javadoc). Renaming to 'CasualtySelector' better aligns with
that purpose.
- Moved one method out of 'CasualtySelector' to its point of usage, after this
method move most methods in the 'CasualtySelector' are now relativey cohesive.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Bug fix:  
[x] Other: Naming consistency  

## Testing

[x] Manually testing done
- Quick smoke testing to get the screenshots.

### Before
Screenshot from 2019-11-20 00-19-35

### After
Screenshot from 2019-11-20 00-38-45

## Additional Review Notes

- The code/documentation is quite consistent about calling the battle calculator the battle calculator. It seems the original 'battleCalculator' class took the domain name that we would expect for the actual 'BattleCalculator', or alternatively 'OddsCalculator' never used its proper domain name. Regardless, it's been a point of confusion for a very long time (and it always takes me some solid time to find the 'real' BattleCalculator whenever I search for it)

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/CasualtySelector.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/CasualtySelector.java
@@ -43,10 +43,10 @@ import org.triplea.util.Tuple;
  * all over the place.
  */
 @Log
-public class BattleCalculator {
+public class CasualtySelector {
   private static final Map<String, List<UnitType>> oolCache = new ConcurrentHashMap<>();
 
-  private BattleCalculator() {}
+  private CasualtySelector() {}
 
   public static void clearOolCache() {
     oolCache.clear();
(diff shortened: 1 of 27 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/CasualtySelector.java
```diff
@@ -1190,44 +1190,4 @@ private static boolean isChooseAa(final GameData data) {
...
     return Properties.getPartialAmphibiousRetreat(data);
   }
-
-  /**
-   * This returns the exact Power that a unit has according to what DiceRoll.rollDiceLowLuck() would
-   * give it. As such, it needs to exactly match DiceRoll, otherwise this method will become
-   * useless. It does NOT take into account SUPPORT. It DOES take into account ROLLS. It needs to be
-   * updated to take into account isMarine.
-   */
-  public static int getUnitPowerForSorting(
```
DanVanAtta: This method is moved, everything else is a straight rename.

DanVanAtta:
My 2 cents, the model class should be `BattleSimulator` so that the `BattleCalculator` uses a `BattleSimulator` to simulate battles. The feature itself, is called the 'battle calculator'. 

NB, the model split right now is not perfect, we have `IBattleCalculator` and `ConcurrentBattleCalculator` - and then it gets into the issue as well that we fire up a Dummy AI to re-use the casualty selection logic for the benefit of the simulator. 

I suppose one thing at a time.. This last iteration of taking 5 minutes to find the right "BattleCalculator" had me finally ready to change this.

RoiEXLab:
That'll work as well IMO.
Regarding the implementation, I recall several failed attempts to simply/optimize the concurent calculator, because the way it actually works is really hacky.
It basically creates a bunch of game data clones, one for each thread, simulates the battles using those instances and then reverts all of those changes before the next thread can re-use the GameData instance.
I really feel like there should be a way to dynamically create a wrapper GameData object that acts as some sort of write layer, which allows read operations on the underlying object, but all write operations are only applied to the wrapper. So that at the end of each simulation the wrapper can just be discarded without any further cleanup.

RoiEXLab:
(Just ignore this rant, at this point I'm too scared to even look at the code again, way too much time wasted there)

DanVanAtta:
I hear ya, it's interesting when you need to do things multiple times to find the right path.

My 2 cents FWIW:
- I've always felt a full game data object should not be needed, we only need the set of units and their current stats
- It seems there should be a battle rules module that uses that data, then in a functional style would compute results which can then be incorporated back to run the next round of combat
- The battle simulator IMO would use this battle rules module and take the "part" of game data that it needs to know the set of units and their stats and then be able to simulate multiple iterations of combat using the battle rules.

I think a big key is segmenting the GameData object. We do run into other issues that makes it particularly difficult where so much is coupled and mutable. Considering the impact for AI to improve the simulations, I wonder/think one day we'll get there.

PR #5555, opened 2019-11-21
Update NGINX SSL settings

DanVanAtta:
Add params from https://cipherli.st/

TODO:
Add dhparam private key

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manually testing done

## Additional Review Notes

- Draft for early preview and commentary


infrastructure/ansible/roles/nginx/templates/etc_nginx_sites_enabled_default.j2
```diff
@@ -10,12 +10,26 @@ server {
...
 
     ssl on;
-    ssl_session_cache  builtin:1000  shared:SSL:10m;
     ssl_protocols  TLSv1.2 TLSv1.3;
     ssl_ciphers EECDH+AESGCM:EDH+AESGCM;
     ssl_prefer_server_ciphers on;
     access_log            /var/log/nginx/java_server.access.log;
 
+    ssl_prefer_server_ciphers on; 
+    ssl_dhparam /etc/nginx/dhparam.pem; # openssl dhparam -out /etc/nginx/dhparam.pem 4096
```
RoiEXLab: Just wanted to point out that this file doesn't need to be a secret according to this so question:
https://security.stackexchange.com/questions/94390/whats-the-purpose-of-dh-parameters

It's just some precomputed values.

RoiEXLab: (Because sou called it "dhparam private key")

DanVanAtta:
@RoiEXLab I suspect it might be better for you to take over this branch/PR. Would you like to, or is it close enough as is? A DH param key file being public is good news. If you can check it into `ansible/roles/http-server/files`, I can add the tasks.yml command, though still pretty straight forward:
```
- name: deploy key file
   src: 'dh-params.pem'
   dest: /dest/dh-params.pem
   mode: 600
```

RoiEXLab:
@DanVanAtta Added the file, now you just need to make ansible move it into the right directory for nginx to pick it up 👍

DanVanAtta:
@RoiEXLab is this a good training example for you to ramp up on how to add this for ansible?
- I'm a single point of failure for the ansible configuration, it would be good for someone else to be hands-on with it.
- We should get the same configuration going for the forums server. It's manually deployed right now, which is a risk/liability. For example, I was unaware of the SSL settings that were applied, we'll lose that info over time until we can get it into configuration.
- The mail configurations should land in ansible so we are not back in the same situation of collectively not knowing how it is deployed

https://docs.ansible.com/ansible/latest/modules/file_module.html

RoiEXLab:
Ok, @DanVanAtta I think I managed to get it done.
I'm unsure about the details (ended up using the copy module instead of file because I wasn't sure it would allow the symlink to the ansible file), but after all I think I understand the basics now.
I was thinking about using the openssl command directly, but discarded this idea because I really don't want to try something too crazy at the beginning.
However I just saw this (first google result when trying to add an if check for a file): https://raymii.org/s/tutorials/Ansible_-_Only_if_a_file_exists_or_does_not_exist.html

So well, let's go with that I suppose. Even though it's not a security risk to publish this file, it certainly has a reason why this file is supposed to be generated

infrastructure/ansible/roles/nginx/tasks/main.yml
```diff
@@ -4,6 +4,18 @@
     name: nginx
     state: present
 
+- name: generate dhparam.pem
+  become: true
+  shell:
+    cmd: openssl dhparam -out /etc/nginx/dhparam.pem 4096
+    creates: /etc/nginx/dhparam.pem
```
DanVanAtta: Good usage of `creates`, it keeps the script idempotent and the 'change' and 'new' correct. It's ideal when an ansible script is run for it to actually only report changed values if something has changed.

I noticed that `/etc/nginx/dhparam.pem` is repeated a number of times in this task file. Sometimes that will be enough to justify extracting to a variable so we can be sure that everything is kept in sync. 
EG:
1) add to defaults.yml:
```
nginx_dh_param_file: /etc/nginx/dhparam.pem
```

Replace usages in tasks/main.yml to look like this:
```
dest: "{{ nginx_dh_param_file }}"
```

Note the double quotes above, those are needed otherwise the YAML parser see `dest: {` and thinks  that the value being defined is  map.

But to bring it home, because the value is also used in the nginx config, the variable becomes arguably more important:
1. ensures the file stays in sync
2. helps document that the file generated here is the same one used in configuration. We can trace the value of the variables to find usages and the correlation can then be discovered that it's a generated file, so we know where it comes from.

DanVanAtta:
Looks good @RoiEXLab, thanks for taking the stab. Did you test the configuration on vagrant or prerelease by any chance?

I'll comment/note that the ansible docs are pretty good reference all-in-all:
https://docs.ansible.com/ansible/latest/modules/shell_module.html#shell-module; but great find/usage of the 'creates' arg. 

I just had the one callout to extract the output file to a variable, reduces the duplication within the task file, but most importantly creates a correlation to the deployed configuration file that should be kept in sync. Otherwise looks good to me.

infrastructure/ansible/roles/nginx/tasks/main.yml
RoiEXLab: In this case, the nginx config template should also contain it.

RoiEXLab:
> Did you test the configuration on vagrant or prerelease by any chance?

Unfortunately not, haven't had the time to read through your documentation on how to do that yet.
If you want to merge this ASAP, feel free to do the remaining change today, or I'll try to do it during the next 2 days when I have the time

DanVanAtta:
There's no rush here @RoiEXLab , hence it's maybe a particularly good opportunity to help spread ansible deployment knowledge. Arguably even, we should fix the existing prerelease deployment problems (https://github.com/triplea-game/triplea/issues/5571) before we merge this so we don't add additional changes to a broken build.

RoiEXLab:
Can you give me a hint on how to verify if this actually works?
A link to the documentation maybe?

DanVanAtta:
This describes how to set up vagrant:
https://github.com/triplea-game/triplea/blob/master/infrastructure/README-VAGRANT-DEV.md

It's going to be ideal as you won't need to mess with prerelease. The execute script to deploy to vagrant is: `./run_ansible_vagrant`

TBH I have not run the vagrant deploy in some time, let me know if you run into any issues and I'll investigate and update documentation/scripts.

Or, you can run against prerelease (less ideal, please go ahead with this if vagrant gives you issues):
1. Create a file "vault_password" in `infrastructure/` (it is in .gitignore, do not check it in)
2. Add the vault password to that file, see the secrets doc
3. Install ansible, check that `ansible-playbook` is available as a command
4. Run: `./run_deployment -d -v -i ansible/inventory/prerelease.yml`
  - The '-d' is for diff output, you'll see the contents of how each file is updated
  - '-v' is for verbose, gives a bit more debug information

RoiEXLab:
@DanVanAtta Tried doing the local install on WSL, but ultimately failed because at some point something failed to connect to localhost there (I assume it was the vagrant instance that wasn't starting for some reason, had some issues with line endings as well, but that was just git on windows being weird some times, ended up converting everything).

So I could try setting up a virtual machine, and try again there, or I could try it on prerelease as suggested, but tbh I don't really know what I'm supposed to look out for to verify everything works.

Also I have a couple of questions regarding prerelease:
Is the repo already checked out there, if yes, where?
Am I supposed to delete everything afterwards, or just leave everything as-is for the next test?

DanVanAtta:
> I assume it was the vagrant instance that wasn't starting for some reason

I won't be too much of help here as the issue with vagrant and getting any kind of OS set up. Perhaps virtual box was missing? Vagrant is just a unified API to virtual hosting, using virtual box.

Having instructions on how to  get set up on windows would be a nice level of completeness.

It is worth considering if we would just want people to set up a linode for temporary usage, though the round-trip there is slower and it's also limiting to those that have linode access. So probably ideal that we can keep that as a local developer workflow.


> but tbh I don't really know what I'm supposed to look out for to verify everything works.

In essense:
- 1. that the ansible execution runs to completion without any hard errors
- 2. The end result is a running software stack. Probably best to get it set up on a 'master' branch to know what the baseline is supposed to look like. 


> Also I have a couple of questions regarding prerelease:
Is the repo already checked out there, if yes, where?


No, it's not. Ansible works by firing commands over SSH. Essentially it's glorified shell scripts. The ansible modules are just python wrappers where you can state what state you want.

For example:

```
file:
   type: directory
   name: some name
```
I don't think that is a real example, but to demonstrate, under the covers that translates to:
```
if [ ! -d 'some name' ]; then
   mkdir 'some name';
fi
```
Then it would check permissions, etc... and the above is executed over the network as SSH commands.


> Am I supposed to delete everything afterwards, or just leave everything as-is for the next test?

A key feature/desing of ansible is its idempotent and declarative. It 'asserts' state and if the state is not correct it updates the system state to be correct. So it should not matter too much. We can have 'hanging' configuration though if we deploy to a certain filename, and then we change the name of that file. Ansible in that case would have no way to know about the old name and would not know to delete it unless explicitly done.

In essence, to thoroughly test:
- run ansible, verify it all works
- run ansible again, verify/minimized reported changes
- delete OS, recreate it
- run ansible again, verify it all works
- run ansible again, verify again changes are minimized.

I think that last flow highlights why a VM is useful.

RoiEXLab:
Ok, second try, this time using a XUbuntu VM.
Failed at the exact same step:
```
robin@robin-VirtualBox:~/Schreibtisch/triplea/infrastructure$ ./run_ansible_vagrant
+ source infrastructure-machine/secrets
./run_ansible_vagrant: Zeile 4: infrastructure-machine/secrets: Datei oder Verzeichnis nicht gefunden
+++ pwd
++ basename /home/robin/Schreibtisch/triplea/infrastructure
+ CUR_FOLDER=infrastructure
+ '[' infrastructure '!=' infrastructure ']'
++ sed 's/.*=\s*//' ../game-core/src/main/resources/META-INF/triplea/product.properties
+ export VERSION=2.0.0
+ VERSION=2.0.0
+ echo 'Parsed version number: 2.0.0'
Parsed version number: 2.0.0
+ buildLobbyJar
+ local jarFile=triplea-lobby-2.0.0.jar
+ local targetFolder=./ansible/roles/lobby/files
+ '[' '!' -e ./ansible/roles/lobby/files/triplea-lobby-2.0.0.jar ']'
+ cd ../
+ '[' '!' -e lobby/build/libs/triplea-lobby-2.0.0.jar ']'
+ ./gradlew shadowjar

BUILD SUCCESSFUL in 1s
23 actionable tasks: 23 up-to-date
+ cd infrastructure
+ mkdir -p ./ansible/roles/lobby/files
+ cp lobby/build/libs/triplea-lobby-2.0.0.jar ./ansible/roles/lobby/files
cp: Aufruf von stat für 'lobby/build/libs/triplea-lobby-2.0.0.jar' nicht möglich: Datei oder Verzeichnis nicht gefunden
+ buildBotJar
+ local jarFile=triplea-game-headless-2.0.0.jar
+ local targetFolder=./ansible/roles/bot/files/
+ '[' '!' -e ./ansible/roles/bot/files//triplea-game-headless-2.0.0.jar ']'
+ buildMigrations
+ local targetFolder=./ansible/roles/flyway/files/
+ local zipFile=migrations.zip
+ '[' '!' -e ./ansible/roles/flyway/files//migrations.zip ']'
+ cd ../lobby-db
+ '[' '!' -e build/distributions/migrations.zip ']'
+ cd ../infrastructure
+ mkdir -p
mkdir: fehlender Operand
„mkdir --help“ liefert weitere Informationen.
+ cp ../lobby-db/build/distributions/migrations.zip ./ansible/roles/flyway/files/
cp: reguläre Datei './ansible/roles/flyway/files/' kann nicht angelegt werden: Ist kein Verzeichnis
+ ansible-playbook -D -v ansible/site.yml -i ansible/inventory/vagrant
Using /home/robin/Schreibtisch/triplea/infrastructure/ansible.cfg as config file
 [WARNING]: Found both group and host with same name: vagrant


PLAY [postgresHosts,dropwizardHosts,botHosts] *************************************************************************************************************************************************************

TASK [apt_update : add mirror for faster download] ********************************************************************************************************************************************************
fatal: [vagrant]: UNREACHABLE! => {"changed": false, "msg": "SSH Error: data could not be sent to remote host \"127.0.0.1\". Make sure this host can be reached over ssh", "unreachable": true}

NO MORE HOSTS LEFT ****************************************************************************************************************************************************************************************
	to retry, use: --limit @/home/robin/Schreibtisch/triplea/infrastructure/ansible/site.retry

PLAY RECAP ************************************************************************************************************************************************************************************************
vagrant                    : ok=0    changed=0    unreachable=1    failed=0
```

I follow the rules described in https://github.com/triplea-game/triplea/blob/master/infrastructure/README-VAGRANT-DEV.md I even installed `openssh-server` because this wasn't installed.
Any ideas? (System is in german, because I'm used to it ^^)

DanVanAtta:
Thanks for bearing with it @RoiEXLab . There were a couple things wrong, should be fixed with: https://github.com/triplea-game/triplea/pull/5682

One note, the 'apt update' step takes a *really* long time, so it may appear frozen. I commented out the step in `ansible/site.yml` to verify that it can proceed past it.


The port mappings in `infrastructure/VagrantFile` are out of date. https://github.com/triplea-game/triplea/pull/5683 updates them, after that you should be able to access spitfire running on the vagrant from your host machine with "localhost:9090"

DanVanAtta:
>  #5683 updates them, after that you should be able to access spitfire running on the vagrant from your host machine with "localhost:9090"

To clarify and be sure to be clear, once the ansible install finishes, *then* the lobby should be running and available. `cd triplea/infrastructure; vagrant ssh` to gain access to the vagrant host to help debug if it is not.

DanVanAtta:
Got the following error @RoiEXLab .

We need to fix the setup script that places `migrations.zip` into the ansible role folder. 
```
TASK [flyway : extract migrations] ****************************************************************************************************************************************************
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Could not find or access 'migrations.zip'\nSearched in:\n\t/home/dan/work/triplea/infrastructure/ansible/roles/flyway/files/migrations.zip\n\t/home/dan/work/triplea/infrastructure/ansible/roles/flyway/migrations.zip\n\t/home/dan/work/triplea/infrastructure/ansible/roles/flyway/tasks/files/migrations.zip\n\t/home/dan/work/triplea/infrastructure/ansible/roles/flyway/tasks/migrations.zip\n\t/home/dan/work/triplea/infrastructure/ansible/files/migrations.zip\n\t/home/dan/work/triplea/infrastructure/ansible/migrations.zip on the Ansible Controller.\nIf you are using a module and expect the file to exist on the remote, see the remote_src option"}

```

I've not 100% sure of the problem just yet. The fact we have two 'run' scripts for ansible is a bit of a problem as they need to be kept in sync.

This function from run-ansible-vagrant appears fine, but the migrations.zip file is not present in the flyway/files folder.
```
function buildMigrations() {
  local targetFolder=./ansible/roles/flyway/files/
  local zipFile=migrations.zip
  if [ ! -e $targetFolder/$zipFile ]; then
    (
      cd ../lobby-db
      if [ ! -e build/distributions/$zipFile ]; then
        ../gradlew release
      fi
      cd ../infrastructure
      mkdir -p $targetFoldee
      cp ../lobby-db/build/distributions/$zipFile $targetFolder
    )
  fi
}

buildMigrations

```

Will need to debug this further..

infrastructure/ansible/roles/nginx/templates/etc_nginx_sites_enabled_default.j2
```diff
@@ -10,12 +10,26 @@ server {
...
     ssl on;
-    ssl_session_cache  builtin:1000  shared:SSL:10m;
     ssl_protocols  TLSv1.2 TLSv1.3;
     ssl_ciphers EECDH+AESGCM:EDH+AESGCM;
     ssl_prefer_server_ciphers on;
     access_log            /var/log/nginx/java_server.access.log;
 
+    ssl_prefer_server_ciphers on; 
+    ssl_dhparam /etc/nginx/dhparam.pem;
+    ssl_ecdh_curve secp384r1; # Requires nginx >= 1.1.0
```
DanVanAtta: @RoiEXLab we should know the version of nginx that we are deploying, the commentary should not really be here, or the infrastructure should take care to inject the right values for the right versions.

DanVanAtta: To this extent, we should change the apt install to install a specific version:
```
  apt:
    name: nginx
    state: present
```
It might be better to add a comment on where the documentation is for these parameters, and/or create a wiki documentation page or add a README to the nginx role to describe why and what these parameters are doing.

infrastructure/ansible/roles/nginx/templates/etc_nginx_sites_enabled_default.j2
```diff
@@ -10,12 +10,26 @@ server {
...
+    ssl_session_cache shared:SSL:10m;
+    ssl_session_tickets off; # Requires nginx >= 1.5.9
+    ssl_stapling on; # Requires nginx >= 1.3.7
+    ssl_stapling_verify on; # Requires nginx => 1.3.7
+    resolver $DNS-IP-1 $DNS-IP-2 valid=300s;
+    resolver_timeout 5s; 
+    add_header Strict-Transport-Security "max-age=63072000; includeSubDomains; preload";
+    add_header X-Frame-Options DENY;
+    add_header X-Content-Type-Options nosniff;
+    add_header X-XSS-Protection "1; mode=block";
```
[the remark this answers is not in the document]
DanVanAtta: This is good context. If we ever have a web browser user agent hit nginx, we would almost just as well reject the request. That would be a bit extreme as it might impair our efforts to do development and debug, but I think it speaks to our expectations/requirements, we really only expect TripleA app to be hitting nginx for the purpose of getting to data over http.

infrastructure/ansible/roles/nginx/templates/etc_nginx_sites_enabled_default.j2
```diff
@@ -10,12 +10,26 @@ server {
...
 
+    ssl_prefer_server_ciphers on; 
+    ssl_dhparam /etc/nginx/dhparam.pem;
+    ssl_ecdh_curve secp384r1; # Requires nginx >= 1.1.0
+    ssl_session_timeout  10m;
+    ssl_session_cache shared:SSL:10m;
+    ssl_session_tickets off; # Requires nginx >= 1.5.9
+    ssl_stapling on; # Requires nginx >= 1.3.7
+    ssl_stapling_verify on; # Requires nginx => 1.3.7
+    resolver $DNS-IP-1 $DNS-IP-2 valid=300s;
```
DanVanAtta: Why valid for 300s?

infrastructure/ansible/roles/nginx/templates/etc_nginx_sites_enabled_default.j2
DanVanAtta: That specific update is arguably out of scope for now, but it's something we shoudl do before we spin up new prod.

DanVanAtta:
https://github.com/triplea-game/triplea/pull/5712 fixes the deployment to vagrant. @RoiEXLab you should be able to confirm a baseline on master after that is merged.

Apologies for not fixing/jumping on vagrant sooner, as soon as you can get a box running, I forgot how relatively pain-free that process was and it is much more efficient than debugging on a live server.

RoiEXLab:
Hmm I'm still getting this error when running
```bash
vagrant up
./run_ansible_vagrant
```
on master:
```
TASK [http_server : deploy jar file] ******************************************************************************************************************************************************************************************************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: 	/home/robin/Schreibtisch/triplea/infrastructure/ansible/triplea-http-server-2.0.0.jar
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Could not find or access 'triplea-http-server-2.0.0.jar'\nSearched in:\n\t/home/robin/Schreibtisch/triplea/infrastructure/ansible/roles/http_server/files/triplea-http-server-2.0.0.jar\n\t/home/robin/Schreibtisch/triplea/infrastructure/ansible/roles/http_server/triplea-http-server-2.0.0.jar\n\t/home/robin/Schreibtisch/triplea/infrastructure/ansible/roles/http_server/tasks/files/triplea-http-server-2.0.0.jar\n\t/home/robin/Schreibtisch/triplea/infrastructure/ansible/roles/http_server/tasks/triplea-http-server-2.0.0.jar\n\t/home/robin/Schreibtisch/triplea/infrastructure/ansible/files/triplea-http-server-2.0.0.jar\n\t/home/robin/Schreibtisch/triplea/infrastructure/ansible/triplea-http-server-2.0.0.jar"}
	to retry, use: --limit @/home/robin/Schreibtisch/triplea/infrastructure/ansible/site.retry
```

Am I missing something?

DanVanAtta:
@RoiEXLab the run script before it executes `ansible-playbook` should be running gradle to generate the deployment artifacts. Check the gradle command that does a shadow jar of http-server, and check the folder where it is supposed to be moved: `/home/robin/Schreibtisch/triplea/infrastructure/ansible/roles/http_server/files/triplea-http-server-2.0.0.jar`

RoiEXLab:
Ok, looks like the `./run_ansible_vagrant` script never actually moves any http-server jar anywhere, probably the issue.
I moved it manually there and this step worked no problem

infrastructure/ansible/roles/nginx/templates/etc_nginx_sites_enabled_default.j2
RoiEXLab: I agree that the comments are not required here, but I don't think it's necessary to lock the nginx version, once we can be sure that all the options are available using the current version in the repository

RoiEXLab:
@DanVanAtta I was finally able to test this config, and fixed the issues with the config.
However I noticed that somehow the portforwarding seems broken. I wasn't able to connect to http://localhost:9090 for some reason, it timed out.
Not sure if the server is actually running?

infrastructure/ansible/roles/nginx/tasks/main.yml
DanVanAtta: @RoiEXLab exactly, hence the suggestion for a variable so that the value is unified. Otherwise it can be changed in one or the other and become out of sync. The usage of variables to keep config in sync is part of the reason to use something like ansible, it makes deployments very difficult to break when a value is changed (as all dependent values then are automatically updated). Hardcoded values imply the value is used in just that one place.

infrastructure/ansible/roles/nginx/templates/etc_nginx_sites_enabled_default.j2
DanVanAtta: We can get an out of sync state without locking version. We may for example test one version now and production then picks up a different one. When upgrading, we likely want to be explicit about it and will want to set a version number at some point to initiate the upgrade. Otherwise, without a version number, we're essentially saying "any" version will do, and that is not the case. For utilities like 'curl', it's appropriate to get any version, for a "main" system component where it matters quite a bit, I suspect we are better off with a specific version.

DanVanAtta:
In case it helps, FYI one can connect to the vagrant server with `vagrant ssh`

> Ok, looks like the ./run_ansible_vagrant script never actually moves any http-server jar anywhere, probably the issue.
I moved it manually there and this step worked no problem

Good discovery. It does look like `run_ansible_vagrant` was before http-server even. We've some work to do.

> I wasn't able to connect to http://localhost:9090 for some reason, it timed out.

This is due to firewall rule.  Eventually, without these updates, I get:
Screenshot from 2019-12-14 19-17-14

I'm not sure exactly what to make of that error and why we didn't get it on prerelease.

RoiEXLab:
@DanVanAtta This sort of error is to be expected because you're using the self signed certificate for nginx.
However if you're able to export this certificate to the machine you're running triplea on and install it you should be fine.

And which firewall rule are you referring to?

And I already remembered the `vagrant ssh` command, it helped me debug the nginx config issues, because an invalid config only fails the nginx restart step without any further details.

RoiEXLab:
@DanVanAtta I was able to successfully verify this works using the self-signed certificate, I added some documentation on how to make it work on ubuntu (windows is even simpler, you double click it and start the import assistant).
However when creating an account I got some weird error about password encoding using password `12345`, which then got me into an infinite loop of popups until I got a rate-limiting exception. (429 status).
In your case you'll have to delete the current vagrant certificate, turns out the self signed certificate needs a "common name" (CN), so HTTPS is satisfied.
I set it to use `localhost`, so you can connect using `https://localhost:9000`, but `https://127.0.0.1:9000`, `https://[::1]:9000` or any alternative IP for the machine will probably not work.

RoiEXLab:
Error
Reproduced the error with `house` as password, looks like it's not being hashed properly?
https://github.com/triplea-game/triplea/blob/4a22b54ee153168091a4d537db1f5725cc011a8b/http-server/src/main/java/org/triplea/server/user/account/create/PasswordValidation.java#L16

Note that whenever I click "OK" or try to exit the dialog, a new dialog pops up with the same message until I hit rate limiting

RoiEXLab:
Ok, I think this should be all of your comments addressed.
Feel free to correct any grammar/spelling mistakes in the README-DEV, but otherwise I think this is ready to merge after almost a month 😅

DanVanAtta:
Thanks for engaging on this @RoiEXLab . It seems you're more comfortable with the ansible aspect. Hopefully our bus factor has now increased from 1 to 2 on the infrastructure code :raised_hands:

PR #5556, opened 2019-11-21
Add automatic prerelease deployment to travis

DanVanAtta:
- Sets up Travis to do a post-release task to deploy prerelease.
This is done in short by running ansible for the prerelease hosts.

- Update README with some ansible quick overview notes and typical
commands for running deployments.

- Update 'run_deployment' script to be more generic and allow
passing of arbitrary args to ansible. This way we can choose to
use the '-d' flag on local but not in travis. '-d' shows file
diffs of what will change on remote servers, very useful for
development and debugging, but would otherwise be dangerous
as any passwords that are written to files on remote servers
would show up in that diff.

 

## Functional Changes

[] New map or map update
[x] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manually testing done
- tested on forked travis build: https://travis-ci.org/DanVanAtta/triplea/jobs/614915797 (I added a hack to set the stage to be 'verify' so that the job would run concurrently allowing for faster development iterations. The 'deploy' stage being defined last should be the last thing run)

## Additional Review Notes
- README diff is not very clean. The running ansible section was moved up and combined with new examples of how to execute 'run_deployment. The first section, overview is brand new. Those are the first two sections, the rest of the README is mostly unchanged (but the diff shows additions/removal because of the re-ordering), so pay attention to just the overview and the running ansible sections.
- I've updated travis to have the new required "ANSIBLE_VAULT_PASSWORD" env variable. That unlocks ansible-vault which decrypts the deployment SSH key. Deployment in ansible is done via SSH commands, the SSH key allows for server-to-server communication. Eventually we'll also embed encrypted secrets in the ansible configuration, ansible-vault will use the same key to decrypt those values.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -122,7 +122,7 @@ jobs:
       install:
         - sudo apt update
         - sudo apt -y install sshpass
-        - sudo pip install ansible
+        - pip install ansible
       language: python
       python: "3.6"
       script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,7 +84,7 @@ jobs:
       script:
         - ./.travis/setup_database
         - ./.travis/run_smoke_tests
-    - stage: deploy
+    - stage: upload artifacts
       if: (branch = master) and (repo = 'triplea-game/triplea') and (type != 'pull_request')
       install: skip
       before_install:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -117,6 +117,22 @@ jobs:
           tags: false
           repo: triplea-game/triplea
           branch: master
+    - stage: deploy prerelease
+      if: (branch = master) and (repo = 'triplea-game/triplea') and (type != 'pull_request')
+      install:
+        - sudo apt update
+        - sudo apt -y install sshpass
+        - sudo pip install ansible
```
RoiEXLab: Depending on the way travis is setup here, you might not even need to use sudo for this command.

Regardless there has to be a way to install python packages on travis without sudo

.travis.yml
```diff
@@ -117,6 +117,22 @@ jobs:
           tags: false
           repo: triplea-game/triplea
           branch: master
+    - stage: deploy prerelease
+      if: (branch = master) and (repo = 'triplea-game/triplea') and (type != 'pull_request')
+      install:
+        - sudo apt update
+        - sudo apt -y install sshpass
```
RoiEXLab: I don't think travis is allowed to run sudo commands, I think you need to declare the packages as usual (see other stages)

.travis.yml
DanVanAtta: Was able to remove the `sudo` on pip: 8eaa2fae2

.travis.yml
DanVanAtta: AFAIK travis can run sudo commands. By default even it'll run all commands as sudo. We have it disabled though: https://github.com/triplea-game/triplea/pull/5556/files/3d25ae1aa1091c4e4c02179f72e33428bcae8f2c#diff-354f30a63fb0907d4ad57269548329e3R2

Sudo is allowed for all Travis OS except windows: https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system

I checked if we could get away without the sudo on either apt update or install, we need it on both:
-  https://travis-ci.org/DanVanAtta/triplea/jobs/615871270
- https://travis-ci.org/DanVanAtta/triplea/jobs/615865889

I don't know why sudo is not needed on the other apt install commands, perhaps they are unnecessary and those packages are installed? I'm not sure.

DanVanAtta:
Reason for the apt was due to the documentation I was following, the `apt update` was also needed. Looks like using 'addons' is possible and avoids the `apt update` requirement, results in a faster builder.

PR #5559, opened 2019-11-22
AI Improve Defend Territory Check

ron-murhammer:
Address round 7 example from https://forums.triplea-game.org/topic/1518/ai-interesting-moves-and-points-of-improvement-on-revised/16

The AI sometimes will be too cautious when determining which territories to defend so update the TUV check to check the max of min TUV swing and 0. Essentially defend the territory as long as it ends up with a negative TUV for the attacker (<0) or its less than the min result (have a bunch of units that attacked into a territory so want to reinforce it for a better result).

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/ProNonCombatMoveAi.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/ProNonCombatMoveAi.java
@@ -1170,7 +1170,8 @@ class ProNonCombatMoveAi {
         }
 
         // Check if its worth defending
-        if ((result.getTuvSwing() - holdValue) > patd.getMinBattleResult().getTuvSwing()
+        if ((result.getTuvSwing() - holdValue)
+                > Math.max(0, patd.getMinBattleResult().getTuvSwing())
             || (!hasHigherStrategicValue
                 && (result.getTuvSwing() + extraUnitValue / 2)
                     >= patd.getMinBattleResult().getTuvSwing())) {


game-core/src/main/java/games/strategy/triplea/ai/pro/ProNonCombatMoveAi.java
```diff
@@ -1170,7 +1170,8 @@ private void moveUnitsToDefendTerritories(
         }
 
         // Check if its worth defending
```
DanVanAtta: Might be worth updating this comment to define "worth defending"

DanVanAtta: (for potential follow-up)

ron-murhammer: Hmm. Not sure how to do it effectively. I think the comment would end up being several lines long and essentially explaining the if conditions. This would be just the comment for the first half of the condition: "Essentially defend the territory as long as it ends up with a negative TUV for the attacker (<0) or its less than the min result (have a bunch of units that attacked into a territory so want to reinforce it for a better result)."

Thoughts?

DanVanAtta: We certainly have danger of an out of date comment. Extracting the conditionals to a method and using intermediate values could make it self-explanatory.

As an example to outline the idea:

```
if(worthDefending( ..)) {
 areSuccessful = true;
}

:
:
private static boolean worthDefending(...) {
  boolean netTuvSwing = ...
  boolean strategicValue = ... 
  return netTuvSwing || strategicValue
}
```

ron-murhammer: Yeah, that's probably a better option than a very long comment. Though even doing that I think it'll be difficult to fully capture all the logic/reasoning behind it.

PR #5564, opened 2019-11-22
Implement User-Ban Filter

DanVanAtta:
Adds a filter to the dropwizard http-server to check if a user is
banned by IP or system-id. If so, their request is rejected.

TODO: future iterations should consider caching a white list
and/or black list so that we are not sending a request to DB
on every http request.

After this update, http-server will reject requests from banned users.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

## Additional Review Notes

- Existing http clients/requests had to be updated to send a system-id header, so the updates are beyond adding just the filter.
- 'BanDurationFormatter' is moved code so that we can re-use that logic
- The main update is adding: BannedPlayerFilter, updating UserBanDao to provide a lookup method
- DB is updated to remove a constraint for ban expiries, enables test code to insert expired bans, and second we add an index for faster ban lookup
- UserBanDao has the table column names extracted to a new class so that we can share those column names for multiple record lookup objects.

### DIFF
--- a/http-server/src/main/java/org/triplea/server/access/BannedPlayerFilter.java
+++ b/http-server/src/main/java/org/triplea/server/access/BannedPlayerFilter.java
@@ -1,0 +1,64 @@
+package org.triplea.server.access;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
+import java.time.Clock;
+import java.time.Duration;
+import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+import javax.ws.rs.container.PreMatching;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status;
+import javax.ws.rs.ext.Provider;
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.RequiredArgsConstructor;
+import org.jdbi.v3.core.Jdbi;
+import org.triplea.http.client.SystemIdHeader;
+import org.triplea.http.client.lobby.moderator.BanDurationFormatter;
+import org.triplea.lobby.server.db.dao.user.ban.BanLookupRecord;
+import org.triplea.lobby.server.db.dao.user.ban.UserBanDao;
+
+@Provider
+@PreMatching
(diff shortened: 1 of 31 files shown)


http-server/src/main/java/org/triplea/server/access/BannedPlayerFilter.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public static BannedPlayerFilter newBannedPlayerFilter(final Jdbi jdbi) {
+    return new BannedPlayerFilter(jdbi.onDemand(UserBanDao.class), Clock.systemUTC());
+  }
+
+  @Override
+  public void filter(final ContainerRequestContext requestContext) {
+    if (Strings.emptyToNull(request.getHeader(SystemIdHeader.SYSTEM_ID_HEADER)) == null) {
+      // missing system id header, abort the request
+      requestContext.abortWith(Response.status(Status.FORBIDDEN).entity("Invalid request").build());
```
RoiEXLab: If you ask me, the status code should be `UNAUTHORIZED` rather than `FORBIDDEN`

lobby-db-dao/src/main/java/org/triplea/lobby/server/db/dao/BadWordsDao.java
```diff
@@ -17,6 +17,9 @@
   @SqlUpdate("delete from bad_word where word = :word")
   int removeBadWord(@Bind("word") String badWordToRemove);
 
+  /*  @SqlQuery("select exists (select * from game_hosting_api_key where key = :apiKey)")
+  boolean keyExists(@Bind("apiKey") String apiKey);*/
+  // TODO: change the count to an exists
```
RoiEXLab: Obligatory TODO warning, also there's a comment containing another query

DanVanAtta: Ah, i was mostly kicking this PR out the door last night before heading to bed, meant to put this file to another branch. Thinking I'll remove it from this PR.

http-server/src/main/java/org/triplea/server/access/BannedPlayerFilter.java
DanVanAtta: https://github.com/triplea-game/triplea/pull/5569

PR #5565, opened 2019-11-23
Fix pixelated territory name rendering.

asvitkine:
Without this change, territory names look very pixelated. See screenshots.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix:  pixelated territory names
[] Other:   

## Testing

[X] Manual testing done

Loaded Great War and looked at territory name rendering.

### Before
Screen Shot 2019-11-23 at 10 37 23 AM

### After
Screen Shot 2019-11-23 at 10 38 07 AM

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/screen/Tile.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/screen/Tile.java
@@ -62,6 +62,8 @@ public class Tile {
             RenderingHints.VALUE_ALPHA_INTERPOLATION_QUALITY);
         g.setRenderingHint(
             RenderingHints.KEY_INTERPOLATION, RenderingHints.VALUE_INTERPOLATION_BICUBIC);
+        g.setRenderingHint(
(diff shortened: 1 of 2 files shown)


asvitkine:
> I'm not sure, but IIRC other IDrawables usually reset the graphic rendering hints after drawing in order not to influence other IDrawables.
> Of course if you want to set this as a global hint, you should be able to set it in `MapPanel` where other hints are set as well.

Thanks for the pointer. Doing it in MapPanel does not seem to work. I didn't trace why.

This code is already setting other properties that it doesn't reset - like font and color. Seems strange to reset the hint but not other things it's setting.

asvitkine:
I found that this is called from Tile.getImage() which sets other rendering hints. Moved it there now.

RoiEXLab:
Ah, of course. I forgot that tiles are being drawn and cached in a different step, good work 👍

PR #5568, opened 2019-11-23
Add controller tests for login and create account

DanVanAtta:
## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[x] Other: automated tests   

## Testing

[] Manual testing done

### DIFF
--- a/integration-testing/src/test/java/org/triplea/server/http/BasicEndpointTest.java
+++ b/integration-testing/src/test/java/org/triplea/server/http/BasicEndpointTest.java
@@ -25,7 +25,7 @@ public abstract class BasicEndpointTest<T> extends DropwizardTest {
   /** Use this to verify an endpoint that returns data. */
   protected <X> X verifyEndpointReturningObject(final Function<T, X> methodRunner) {
     final X result = methodRunner.apply(clientBuilder.apply(localhost));
-    assertThat(methodRunner.apply(clientBuilder.apply(localhost)), notNullValue());
+    assertThat(result, notNullValue());
     return result;
   }
 
--- a/integration-testing/src/test/java/org/triplea/server/http/BasicEndpointTest.java
+++ b/integration-testing/src/test/java/org/triplea/server/http/BasicEndpointTest.java
@@ -23,8 +23,10 @@ public abstract class BasicEndpointTest<T> extends DropwizardTest {
   }
 
   /** Use this to verify an endpoint that returns data. */
-  protected void verifyEndpointReturningObject(final Function<T, ?> methodRunner) {
+  protected <X> X verifyEndpointReturningObject(final Function<T, X> methodRunner) {
+    final X result = methodRunner.apply(clientBuilder.apply(localhost));
     assertThat(methodRunner.apply(clientBuilder.apply(localhost)), notNullValue());
+    return result;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


integration-testing/src/test/java/org/triplea/server/http/BasicEndpointTest.java
```diff
@@ -23,8 +23,10 @@ protected void verifyEndpointReturningVoid(final Consumer<T> methodRunner) {
   }
 
   /** Use this to verify an endpoint that returns data. */
-  protected void verifyEndpointReturningObject(final Function<T, ?> methodRunner) {
+  protected <X> X verifyEndpointReturningObject(final Function<T, X> methodRunner) {
+    final X result = methodRunner.apply(clientBuilder.apply(localhost));
     assertThat(methodRunner.apply(clientBuilder.apply(localhost)), notNullValue());
```
RoiEXLab:
```suggestion
    assertThat(result, notNullValue());
```
The space difference at the beginning of the diff seems to be just a visual bug, no space is actually added

integration-testing/src/test/java/org/triplea/server/user/account/login/CreateAccountControllerIntegrationTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+  @Test
+  void duplicateAccountCreateFails() {
+    verifyEndpointReturningObject(client -> client.createAccount(USERNAME_1, EMAIL_1, PASSWORD_1));
+
+    final CreateAccountResponse result =
+        verifyEndpointReturningObject(
+            client -> client.createAccount(USERNAME_1, EMAIL_1, PASSWORD_1));
+
+    assertThat(result, not(is(CreateAccountResponse.SUCCESS_RESPONSE)));
```
RoiEXLab:
```suggestion
    assertThat(result, is(not(CreateAccountResponse.SUCCESS_RESPONSE)));
```

Should work as well, it's probably nicer to read.

integration-testing/src/test/java/org/triplea/server/user/account/login/LoginControllerIntegrationTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    assertThat(response.getFailReason(), nullValue());
+    assertThat(response.getApiKey(), notNullValue());
+
+    // second login should behind the scenes do a password upgrade, transparent to the client.
+    response =
+        verifyEndpointReturningObject(
+            client -> client.login(LEGACY_PASSWORD_USER, LEGACY_PASSWORD));
+    assertThat(response.getFailReason(), nullValue());
+    assertThat(response.getApiKey(), notNullValue());
```
RoiEXLab: I personally like to wrap hamcrest matchers in `is()` because it makes the assertion very nice to read (assert that x is null value), but really a style preference

DanVanAtta: Interesting point. Probably better for us to discuss if we want to make that a standard/convention. I'll consider it further. Would you be willing/able to do the bulk correction/update if so?

PR #5570, opened 2019-11-24
Update BadWordsDao existence check from 'count' to 'exists'

DanVanAtta:
## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/integration-testing/src/test/java/org/triplea/lobby/server/db/dao/BadWordsDaoTest.java
+++ b/integration-testing/src/test/java/org/triplea/lobby/server/db/dao/BadWordsDaoTest.java
@@ -53,7 +53,7 @@ class BadWordsDaoTest extends DaoTest {
   }
 
   @ParameterizedTest
-  @ValueSource(strings = {"zz", "", "some word not containing any bad words"})
+  @ValueSource(strings = {"zz", "", "some string not containing any bad words"})
   void notBadWordContains(final String notInBadWords) {
     assertThat(badWordsDao.containsBadWord(notInBadWords), is(false));
   }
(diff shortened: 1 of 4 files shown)


integration-testing/src/test/java/org/triplea/lobby/server/db/dao/BadWordsDaoTest.java
```diff
@@ -48,17 +49,12 @@ void removeBadWord() {
...
-
-  @SuppressWarnings("unused")
-  private static List<String> notBadWordContains() {
-    return List.of("zz", "", "some word not containing any bad words");
+    assertThat(badWordsDao.containsBadWord(badWord), is(true));
   }
 
   @ParameterizedTest
-  @MethodSource
+  @ValueSource(strings = {"zz", "", "some word not containing any bad words"})
```
RoiEXLab:
```suggestion
  @ValueSource(strings = {"zz", "", "some string not containing any bad words"})
```
"word not containing word" sounds somewhat wrong to me 🤔

lobby-db-dao/src/main/java/org/triplea/lobby/server/db/dao/BadWordsDao.java
```diff
@@ -18,8 +18,6 @@
   int removeBadWord(@Bind("word") String badWordToRemove);
 
   @SqlQuery(
-      "select count(bw.word) "
-          + "from bad_word bw "
-          + "where lower(:word) like '%' || lower(bw.word) || '%'")
-  int containsBadWord(@Bind("word") String word);
+      "select exists (select * from bad_word where lower(:word) like '%' || lower(word) || '%')")
```
RoiEXLab: The power of SQL

DanVanAtta: FWIW, it's actually a significant improvement to use `exists` over `count`:
https://blog.jooq.org/2016/09/14/avoid-using-count-in-sql-when-you-could-use-exists/

Runtime performance is:

> 40x on PostgreSQL

RoiEXLab: Not surprising to see a performance improvement since postgres now no longer has to go through all entries to get the exact number, but instead can abort once it finds a single entry.
Factor 40 is a lot though

PR #5572, opened 2019-11-24
Simplify ContentReader

RoiEXLab:
Follow-up to #5460
## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/map/download/ContentReader.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/map/download/ContentReader.java
@@ -4,11 +4,11 @@ package games.strategy.engine.framework.map.download;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
-import java.nio.channels.Channels;
+import java.nio.file.Files;
+import java.nio.file.StandardCopyOption;
 import java.util.Optional;
 import java.util.logging.Level;
 import lombok.AllArgsConstructor;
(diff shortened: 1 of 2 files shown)


game-core/src/main/java/games/strategy/engine/framework/map/download/ContentReader.java
```diff
@@ -63,9 +63,7 @@ void downloadToFile(final String uri, final File file) throws IOException {
     checkNotNull(uri);
     checkNotNull(file);
 
-    try (FileOutputStream os = new FileOutputStream(file)) {
-      downloadAndApplyAction(
-          uri, is -> os.getChannel().transferFrom(Channels.newChannel(is), 0L, Long.MAX_VALUE));
-    }
+    downloadAndApplyAction(
+        uri, is -> Files.copy(is, file.toPath(), StandardCopyOption.REPLACE_EXISTING));
```
DanVanAtta: @RoiEXLab are sure downloading multiple maps in parallel will have no issues?

RoiEXLab: Well if it worked before without any issues, it will continue to work without any issues.
If you're concerned about the lambda, the only reason a lambda is being used here is to make it hard to misuse the passed input stream: The stream is closed by `downloadAndApplyAction`.
So the only real change here is that we
1. No longer use java.nio.channels
2. let java.nio.Files handle the outputstream logic for us, instead of creating a stream on our own.

RoiEXLab: But in case you're not convinced I just downloaded 4 maps concurrently without any issues (when downloading, got some non 200 status codes from github before actually starting the download)

RoiEXLab:
@DanVanAtta I just realized we could create a GitHub action that labels any PR based on the ticked checkbox ^^

game-core/src/main/java/games/strategy/engine/framework/map/download/ContentReader.java
DanVanAtta: Thanks for checking, wanted to be sure we were using unique temp files per download.

RoiEXLab: I'm pretty sure that's still the case, but as mentioned already it shouldn't really matter regarding this change.

The only difference I found by the way when having a look at the jdk source code, is that `FileOutputStream` seems to overwrite or append existing files if they are present, whereas the nio variant straight up deletes any existing file before creating a new one when the REPLACE_EXISTING flag is present.

PR #5573, opened 2019-11-24
Simplify scrambling code. No functional changes.

asvitkine:
I was working on another PR that was refactoring this code to a helper class that can be used in another context (I want to use it from the UI code to warn a player when attacking a territory where the defender can scramble) and noticed that the current code is more complex than it needs to be. I decided to do this simplification first rather than coupling it with my other change.

In particular, it was using a data structure that had an extra inner collection for no reason (it always ever had one entry). This change simplifies that and also fixes some comments that were badly formatted.

Behaviour should not change.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[X] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[X] Manual testing done

Tested out scrambling on WWII Global 2nd Edition, with german subs attacking 3 territories around UK where planes from Scotland and London (which has UK and French planes) can scramble.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/BattleDelegate.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/BattleDelegate.java
@@ -815,13 +815,13 @@ public class BattleDelegate extends BaseTripleADelegate implements IBattleDelega
       }
       boolean scrambledHere = false;
 
-      // Remove any units that were already scrambled.
+      // Remove any units that were already scrambled to other territories.
       int unitsLeft = 0;
       final var scramblers = scramblersByTerritoryPlayer.get(terrPlayer);
-      for (final Territory from : scramblers.entrySet()) {
+      for (final Territory from : scramblers.keySet()) {
         final Collection<Unit> unitsToScramble = scramblers.get(from).getSecond();
         unitsToScramble.retainAll(from.getUnitCollection());
-        unitsLeft += unitsToScramble.getSecond().size();
+        unitsLeft += unitsToScramble.size();
       }
 
       if (unitsLeft > 0) {
--- a/game-core/src/main/java/games/strategy/triplea/delegate/BattleDelegate.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/BattleDelegate.java
@@ -826,109 +826,111 @@ public class BattleDelegate extends BaseTripleADelegate implements IBattleDelega
       if (unitsLeft == 0) {
         continue;
       }
-      final Map<Territory, Collection<Unit>> toScramble =
-          getRemotePlayer(defender).scrambleUnitsQuery(to, scramblers);
-      if (toScramble == null) {
-        continue;
-      }
-      // verify max allowed
-      if (!scramblers.keySet().containsAll(toScramble.keySet())) {
-        throw new IllegalStateException("Trying to scramble from illegal territory");
-      }
-      for (final Territory t : scramblers.keySet()) {
-        if (toScramble.get(t) == null) {
+      {
+        final Map<Territory, Collection<Unit>> toScramble =
+            getRemotePlayer(defender).scrambleUnitsQuery(to, scramblers);
+        if (toScramble == null) {
           continue;
         }
-        if (toScramble.get(t).size() > getMaxScrambleCount(scramblers.get(t).getFirst())) {
-          throw new IllegalStateException(
-              "Trying to scramble "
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/BattleDelegate.java
```diff
@@ -819,28 +814,26 @@ private void doScrambling() {
         continue;
       }
       boolean scrambledHere = false;
-      for (final var scramblers : scramblersByTerritoryPlayer.get(terrPlayer)) {
```
asvitkine: This loop is no longer needed now that we removed the extra Collection that was previously scramblersByTerritoryPlayer.get(terrPlayer).

We still need the inner loop - it was previous a while() loop on an Iterator, but I switched it to a simpler for loop below on line 821.

game-core/src/main/java/games/strategy/triplea/delegate/BattleDelegate.java
```diff
@@ -819,28 +814,26 @@ private void doScrambling() {
...
-        // verify that we didn't already scramble any of these units
-        final Iterator<Territory> territoryIter = scramblers.keySet().iterator();
-        while (territoryIter.hasNext()) {
-          final Territory t = territoryIter.next();
-          scramblers.get(t).getSecond().retainAll(t.getUnitCollection());
-          if (scramblers.get(t).getSecond().isEmpty()) {
-            territoryIter.remove();
```
asvitkine: This logic was removing an entry from the scramblers map when it no longer had units.
This was fine, but really it's unnecessary complex. I replaced it with logic to just count
the remaining units instead sand only proceed if there are any left, which is equivalent.

asvitkine: Actually, looks like this is causing a behaviour changes - although very subtle.

This can result in empty map entries being passed to the player via scrambleUnitsQuery(), if there is at least one entry that's not empty. This actually causes the UI for scrambling to show a box with no units rather than not show a box. Since I didn't intend a behavior change here, let me restore the previous logic.

asvitkine: Restored the logic with a slightly better construct.

game-core/src/main/java/games/strategy/triplea/delegate/BattleDelegate.java
asvitkine: The inner loop is now a removeIf() instead.

game-core/src/main/java/games/strategy/triplea/delegate/BattleDelegate.java
```diff
@@ -819,28 +814,28 @@ private void doScrambling() {
...
-          }
-        }
-        if (scramblers.isEmpty()) {
-          continue;
-        }
+
+      final var scramblers = scramblersByTerritoryPlayer.get(terrPlayer);
```
ron-murhammer: Thoughts on specifying the type instead of using var here? Its a bit hard to understand given that scramblers is a fairly complex type.

asvitkine: I can if you prefer. It definitely provides value, but at the same time the type is very long:

Map<Territory, Tuple<Collection<Unit>, Collection<Unit>>>

asvitkine: I ended up doing it. Done.

asvitkine: Ah, I missed that this was already merged. I'll send a new PR to do that.

RoiEXLab: It's a complete type-mess here, that's why I didn't say anything regarding that topic, I don't think the type actually helps with the general confusion.
Having some dedicated types instead of tuple would definitely help here

asvitkine: Agreed.

ron-murhammer: Yeah, this is definitely a case where a custom type would be useful but I think you get at least some value with specifying the type. It makes it so you don't have to go read a bunch of extra code to figure out what the type is.

PR #5576, opened 2019-11-25
Cleanup code and make battle drawable deterministic.

asvitkine:
Cleans up some code in BattleDrawable and removes some nondeterminism.

The nondeterminism was due to the fact that Set does not have a deterministic iteration order. With the previous logic, for non-water territories it was not deterministic whether the stripe color of the attacker would be used or not - if there are two other warring players there. Now, if there is an attacker (i.e. someone at war with territory owner), their stripe color will be used.

(Though there is still nondeterminism if there is more than one attacker player.)

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[X] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
[] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/screen/drawable/BattleDrawable.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/screen/drawable/BattleDrawable.java
@@ -31,32 +31,29 @@ public class BattleDrawable extends TerritoryDrawable {
       final GameData data,
       final Graphics2D graphics,
       final MapData mapData) {
+    final Territory territory = data.getMap().getTerritory(territoryName);
     final Set<PlayerId> players = new HashSet<>();
-    for (final Unit u : data.getMap().getTerritory(territoryName).getUnitCollection()) {
+    for (final Unit u : territory.getUnitCollection()) {
       if (!TripleAUnit.get(u).getSubmerged()) {
         players.add(u.getOwner());
       }
     }
-    final Territory territory = data.getMap().getTerritory(territoryName);
     PlayerId attacker = null;
     boolean draw = false;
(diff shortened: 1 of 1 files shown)


ron-murhammer:
@asvitkine Do you have an example of when this would happen (screenshot or save game)?

asvitkine:
> Ok, your changes seem fine to me, but I don't quite get where the code turns more deterministic TBH.
> I'm probably missing something though

So previous, for the !territory.isWater() case, we had the following logic:

```
    for (final PlayerId p : players) {
        if (data.getRelationshipTracker().isAtWar(p, territory.getOwner())) {
          attacker = p;
          draw = true;
          break;
        }

        for (final PlayerId p2 : players) {
          if (data.getRelationshipTracker().isAtWar(p, p2)) {
            draw = true;
            break;
          }
        }
    }
```

So it can exit the loop if either the first condition or the second condition matches. More specifically, if a player P is at war with the territory owner or is at war with some other player who's there.

But the two exit conditions are not equivalent, in the first case, "attacker" is set, which changes the highlight color.

The nondeterminism comes with the fact that the first loop's order is undefined.

So let's say there's two players in the list, P1 who's at war with the territory owner and P2 who's at war with let's say P3 who's also there.

If we check P1 first, then we'll mark him as an attacker and use his color. But if we check P2 first, then we don't mark P1 as an attacker.

But which happens is nondeterministic since we're iterating a list.

Does that help explain things?

Now, I don't actually know if this can happen in practice - it may only happen if some map has a trigger that spawn some units for a player somewhere during a different player's combat step or some such. Anyway, the new code structure in this PR fixes that source of nondeterminism.

PR #5579, opened 2019-11-25
Fix potential NPE, return empty collection instead of null

DanVanAtta:
Usages of `getWhatTerritoriesThisIsUsedInConvoysFor` will NPE if
we return null, the callers do not use `null` checks but instead
`isEmpty` checks that would otherwise NPE.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix: fix potential NPE 
[] Other:   

## Testing

[] Manual testing done

## Additional Review Notes

Example location that could NPE: https://github.com/DanVanAtta/triplea/blob/ca0f1361a4efe3847aed2e982ac3587694e3a891/game-core/src/main/java/games/strategy/triplea/attachments/TerritoryAttachment.java#L700

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/attachments/TerritoryAttachment.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/TerritoryAttachment.java
@@ -614,11 +614,11 @@ public class TerritoryAttachment extends DefaultAttachment {
    * Returns the collection of territories that make up the convoy route containing the specified
    * territory or {@code null} if the specified territory is not part of a convoy route.
    */
-  public static Set<Territory> getWhatTerritoriesThisIsUsedInConvoysFor(
+  public static Collection<Territory> getWhatTerritoriesThisIsUsedInConvoysFor(
       final Territory t, final GameData data) {
     final TerritoryAttachment ta = TerritoryAttachment.get(t);
     if (ta == null || !ta.getConvoyRoute()) {
-      return null;
+      return new HashSet<>();
     }
     final Set<Territory> territories = new HashSet<>();
     for (final Territory current : data.getMap().getTerritories()) {
@@ -696,7 +696,7 @@ public class TerritoryAttachment extends DefaultAttachment {
       if (!convoyAttached.isEmpty()) {
         sb.append("Needs: ").append(MyFormatter.defaultNamedToTextList(convoyAttached)).append(br);
       }
-      final Set<Territory> requiredBy = getWhatTerritoriesThisIsUsedInConvoysFor(t, getData());
+      final Collection<Territory> requiredBy = getWhatTerritoriesThisIsUsedInConvoysFor(t, getData());
       if (!requiredBy.isEmpty()) {
         sb.append("Required By: ")
             .append(MyFormatter.defaultNamedToTextList(requiredBy))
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/triplea/attachments/TerritoryAttachment.java
```diff
@@ -614,11 +614,11 @@ private void resetBlockadeZone() {
    * Returns the collection of territories that make up the convoy route containing the specified
    * territory or {@code null} if the specified territory is not part of a convoy route.
    */
-  public static Set<Territory> getWhatTerritoriesThisIsUsedInConvoysFor(
+  public static Collection<Territory> getWhatTerritoriesThisIsUsedInConvoysFor(
```
RoiEXLab: Hmm, defining Set as return type usually implies that there are no duplicates, so I'd actually like to keep the interface defined as Set

DanVanAtta: The javadoc states collection:
> Returns the collection of territories that make up the convoy route containing the specified
   * territory 

We probably should update L623 as well:
```
 final Set<Territory> territories = new HashSet<>();
```
The fact no duplicates are returned is an attribute of the implementation. We've a problem where we over-specify 'Set' or 'List' in many places where we rely on nor need either API. Looking back at BattleTracker, it looks like a good bit of this over-specification came from the original 2003 code, are we continuing bad habits?

You can also have a List containing all unique elements as well, the fact that you get set does not mean that no other collection type will not also include duplicates. IE: `Set` type is not the only collection type or situation where you may see all unique values.

DanVanAtta: I updated the javadoc as it also stated that the return value was nullable: 5831782402ad94ffe26756f66d4a88e9d8520605

DanVanAtta: Checking for any 3rd party references on best practices here, there's not too much. The examples I have found do encourage returning `Collection` type:


- https://www.codejava.net/java-core/collections/18-java-collections-and-generics-best-practices#PreferInterface
- https://www.javaguides.net/2018/06/java-collection-framework-best-practices.html

Examples of in-line code assign to `List`, but those examples are for local scoped variables. Given we are setting a contract, the more generic of interface type seems to be the rule of thumb.

ron-murhammer: I personally prefer to return Set since the caller then knows you have unique elements.

DanVanAtta: IDK, arguably if a caller can generalize to a `Collection`, then they should. IE: don't depend on a `Set` return type if you all you need is an iterator. In that case if the callee changes, the caller can be unchanged. 

I can see some merits to both sides. My concerns for this codebase:
- continues a pattern of using concrete types and/or Set/List when all we need is Collection. There is a mix, most of the v1 code uses the older types, looks like when Chris got involved much was made to depend on Collection
- It's not often that a caller actually cares about uniqueness, it pushes details that we maybe should not be concerned about. #bikeshedding  It's very similar I think to the underscore naming pattern, it pushes a lot of unnecessary details higher in importance than is actually necessary, by the time you're done with the ceremonies you can then worry about the actual business code
- We probably should be using `Collection` as internal type. It's the implementation that defines the uniqueness and ordering behavior, not the type. IE:
```
Collection<String> strings = new HashSet<>();
```
If we then have 'getters', it's going to be odd that some methods return `Collection` and others return `Set`.

PR #5582, opened 2019-11-25
Fix potential NPEs

DanVanAtta:
## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

[] Manual testing done


DanVanAtta:
In these cases I don't think we would have seen proper failure, would have been random NPE. From what I can tell it's bugs, we check for null in other locations, then it looks like we forget to do so later. These issues were highlighted by IDE, it detects the null check and then in the same code block that there is a null dangerous call.

In one case for example there is a call to a helper method for input, on error it shows a message and then returns null. In that case we'd fast-follow with a NPE in the caller method.

PR #5585, opened 2019-11-25
Replace constant-valued variables with the constant value

DanVanAtta:
## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/Route.java
+++ b/game-core/src/main/java/games/strategy/engine/data/Route.java
@@ -448,13 +448,13 @@ public class Route implements Serializable, Iterable<Territory> {
   private Tuple<ResourceCollection, Boolean> getFuelCostsAndIfChargedFlatFuelCost(
       final Unit unit, final GameData data, final boolean ignoreFlat) {
     final ResourceCollection resources = new ResourceCollection(data);
-    boolean chargedFlatFuelCost = false;
     if (Matches.unitIsBeingTransported().test(unit)) {
-      return Tuple.of(resources, chargedFlatFuelCost);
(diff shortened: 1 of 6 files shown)


game-core/src/main/java/games/strategy/engine/data/Route.java
```diff
@@ -448,13 +448,13 @@ public static ResourceCollection getMovementFuelCostCharge(
   private Tuple<ResourceCollection, Boolean> getFuelCostsAndIfChargedFlatFuelCost(
       final Unit unit, final GameData data, final boolean ignoreFlat) {
     final ResourceCollection resources = new ResourceCollection(data);
-    boolean chargedFlatFuelCost = false;
     if (Matches.unitIsBeingTransported().test(unit)) {
-      return Tuple.of(resources, chargedFlatFuelCost);
+      return Tuple.of(resources, false);
```
RoiEXLab: The more tuples I see, the worse they seem to get. No idea what kind of value this method actually returns

ron-murhammer: Actually returns what the method name describes: fuel costs (ResourceCollection) and if charged flat fuel cost (boolean).

DanVanAtta: No disagreement, but it's one more thing to remember. If you can't also see the method name at the same time, it's yet another layer to understand. Whenever you have to reference other code to understand the current line, it's a bit of a maintenance code-smell.

PR #5586, opened 2019-11-26
Allow build or download of deployment artifacts

DanVanAtta:
- Updates prerelease configuration to build artifacts instead of download
- Sets up for auto-prod deployments, allows for prod artifacts to be downloaded

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done
Did some quick smoke testing, verified works on prerelease:
`./run_deployment 2.0.16821 -v --diff -i ansible/inventory/prerelease`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,7 +96,7 @@ jobs:
         - ./.travis/setup_gpg
         ## Update product version to include build number.
         ## EG: replace "2.0.0" to be "2.0.15555";
-        - sed -i "s/.0$/.$TRAVIS_BUILD_NUMBER/" game-core/src/main/resources/META-INF/triplea/product.properties
+        - sed -i "s/.0 *$/.$TRAVIS_BUILD_NUMBER/" game-core/src/main/resources/META-INF/triplea/product.properties
         - ./.travis/do_release
       deploy:
         jdk: openjdk11
@@ -129,7 +129,7 @@ jobs:
       python: "3.6"
       script:
         - sed -i "s/.0$/.$TRAVIS_BUILD_NUMBER/" game-core/src/main/resources/META-INF/triplea/product.properties
-        - version=$(cat game-core/src/main/resources/META-INF/triplea/product.properties | sed 's/.* //')
+        - version=$(cat game-core/src/main/resources/META-INF/triplea/product.properties | tr -d ' ' | sed 's/.*=//')
         - eval "$(ssh-agent -s)"
         - cd infrastructure
         - ./build_artifacts
--- a/.travis.yml
+++ b/.travis.yml
@@ -128,13 +128,30 @@ jobs:
       language: python
       python: "3.6"
       script:
+        - sed -i "s/.0$/.$TRAVIS_BUILD_NUMBER/" game-core/src/main/resources/META-INF/triplea/product.properties
+        - version=$(cat game-core/src/main/resources/META-INF/triplea/product.properties | sed 's/.* //')
         - eval "$(ssh-agent -s)"
-        - version=$(sed "s/.0$/.$TRAVIS_BUILD_NUMBER/" game-core/src/main/resources/META-INF/triplea/product.properties)
         - cd infrastructure
-        - export ANSIBLE_HOST_KEY_CHECKING=False
+        - ./build_artifacts
         - echo "$ANSIBLE_VAULT_PASSWORD" > vault_password
         - ./run_deployment "$version" -i ansible/inventory/prerelease
-
+    - stage: deploy production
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -128,13 +128,30 @@ jobs:
       language: python
       python: "3.6"
       script:
+        - sed -i "s/.0$/.$TRAVIS_BUILD_NUMBER/" game-core/src/main/resources/META-INF/triplea/product.properties
+        - version=$(cat game-core/src/main/resources/META-INF/triplea/product.properties | sed 's/.* //')
```
RoiEXLab: I'm confused about the sed command at the end, I suppose it's supposed to remove whitespaces from the string if present?
I believe that in case this string ever end up ending with whitespace like `1.2.3    `,  the whole string will be replaced with an empty string because `.` matches any char

.travis.yml
```diff
@@ -128,13 +128,30 @@ jobs:
...
+      addons:
+        apt:
+          packages:
+            - sshpass
+      install:
+        - pip install ansible
+      language: python
+      python: "3.6"
+      script:
+        - version="2.0.16817"
```
RoiEXLab: Is this supposed to be hardcoded?
It doesn't seem convenient having to update this file whenever we want to push a new version, so I'm guessing you just put that there to verify this step on your fork

.travis.yml
```diff
@@ -128,13 +128,30 @@ jobs:
...
+        - ./.travis/download_artifacts "$version" 
+        - eval "$(ssh-agent -s)"
+        - cd infrastructure
+        - echo "$ANSIBLE_VAULT_PASSWORD" > vault_password
+        # - ./run_deployment "$version" -i ansible/inventory/production
```
RoiEXLab: I guess this should be uncommented, otherwise the previous cd command seems a little bit uneccessary.

.travis.yml
DanVanAtta: No, it is hardcoded. We won't push a new version too often.
 
Travis env variable was another option, but we lose the record of when updates are made.

It's perhaps important to note this is just the version of the downloaded artifacts. We will run the rest of the deployment configuration on every update. This will allow us to deploy/remove keys and do security or configuration updates without changing the running binaries.

Triggering builds on branches or tags was also another option but also not great either as we can't update prod with the latest non-source code updates.

.travis.yml
DanVanAtta: inventory/production is not configured yet. The rest of the changes here should fix the build.
The heavy lift is certainly in `run_deployment`, so this task is not doing too much just yet other than exercising the download logic.

.travis.yml
DanVanAtta: The regex removes everything up until the last space. The format of the version number has been stable for a relatively long time, ie: `version = 2.0`

Indeed, trailing space would break the regex. I'd be more concerned about the disappearance of the leading space, but again the version number has been quite stable. Given we are unlikely to change version number except for a few times every year, is it worth a more complicated regex?

RoiEXLab: Well we don't really expect the version string itself to contain spaces, so doing a global sed (sed with global flag) to replace spaces with an empty string will do, right? 
Nothing too complicated

DanVanAtta: If there is a trailing space, the previous `sed` higher up in the file is going to fail too:

```
        - ./.travis/setup_gpg
        ## Update product version to include build number.
        ## EG: replace "2.0.0" to be "2.0.15555";
        - sed -i "s/.0$/.$TRAVIS_BUILD_NUMBER/" game-core/src/main/resources/META-INF/triplea/product.properties
```

DanVanAtta: Updated to guard against unexpected white space: 12f07ab83

.travis.yml
DanVanAtta: @ron-murhammer heads up that updating this version value is how we would change the version of the running software in prod.

@RoiEXLab , I'd recommend we extract this conversation to an issue so we can be sure to be on the same page, and/or consider if there are other ways to inject and maintain the version we wish to deploy.

DanVanAtta: For follow-up to this topic: https://github.com/triplea-game/triplea/issues/5602
I'd recommend we avoid blocking this PR as we are still landing new features and can have room to iterate. It's also the case that we are not yet updating production in any way, so a new PR could cleanly show the changes, rather than waiting, losing context, and then having to re-review everything to regain the context (and avoids further merge conflicts)

PR #5587, opened 2019-11-26
Remove RSA keys, leave ed25519 keys

DanVanAtta:
ed25519 is very secure and was the key type requested
in key creation instructions:

https://github.com/triplea-game/infrastructure/blob/master/docs/server_access.md#generate-ssh-key

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done


ron-murhammer:
@DanVanAtta Not sure its worth removing those keys and making @RoiEXLab and I generate keys again. Thoughts @RoiEXLab ?

DanVanAtta:
@ron-murhammer it's a one-time operation. Given the keys are posted publicly, using a very strong encryption is seemingly worth-while.  Your keys are the only ones that are not ed25519.

RoiEXLab:
I'm a bit mixed on this, I mean I get that elliptic curves seem to be superior to typical prime factorization, but at the same time RSA isn't viewed as insecure to my knowledge, and since it's only being used to pass a challenge response test, it doesn't really affect the security of the overall connection.
So if there aren't any drawbacks in convenience at some point I really don't see a need to remove the keys, but at the same time I will prefer newer algorithms in the future.

DanVanAtta:
On my side, and for future keys, it's difficult to know the strength of a RSA key added. If it's RSA, we have to be sure it has enough bits. On the other hand, ED25519 would be consistent, the keys are nice and short, and there is no question on if it was generated with enough secure bits. On a last consideration, instructions for how to generate it are provided, it takes less than 5 minutes to generate such a key.

DanVanAtta:
@RoiEXLab 
>  RSA isn't viewed as insecure to my knowledge

That's true so long as enough bits are used, it can require non-default settings.


https://en.wikipedia.org/wiki/RSA_(cryptosystem)#Attacks_against_plain_RSA

> There are a number of attacks against plain RSA as described below.

>   When encrypting with low encryption exponents (e.g., e = 3) and small values of the m, (i.e., m < n1/e) the result of me is strictly less than the modulus n. In this case, ciphertexts can be easily decrypted by taking the eth root of the ciphertext over the integers.

>  Some experts believe that 1024-bit keys may become breakable in the near future or may already be breakable by a sufficiently well-funded attacker, though this is disputable

> Keys of 512 bits have been shown to be practically breakable in 1999 when RSA-155 was factored by using several hundred computers, and these are now factored in a few weeks using common hardware. Exploits using 512-bit code-signing certificates that may have been factored were reported in 2011.[30] A theoretical hardware device named TWIRL, described by Shamir and Tromer in 2003, called into question the security of 1024 bit keys. It is currently recommended that n be at least 2048 bits lon


Hence, there are questions whether non-default settings were used. It's nicer to ellide all these questions and have consistent formats and not take the exceptions nor risk

DanVanAtta:
Added a key from RoiEXLab. 

To sum up:
- I don't want have to police people how any key was generated now or going forward
- Ed25519 was asked for in the instructions
- It should be fast and easy to rotate keys, generating a key is not a big deal: `ssh-keygen -t ed25519`

DanVanAtta:
@ron-murhammer if there is an emergency and you absolutely need to log in to the prerelease server, we can add your key and deploy it pretty quickly. Otherwise I object to the precedent that any key can be added, and I do not want to ever need to ask people how they generated a RSA key. Last, rotating keys should be pretty easy, if we can't generate and replace a key as if it were not a big deal, that is also a problem.

This is currently blocked, I'd like to get this moving forward to reduce PR queue size and to pop it from my mental stack of things I'm worrying about/have in flight.

PR #5598, opened 2019-11-26
Convert Optional(Long) data member to @Nullable Long

DanVanAtta:
For class member variables, nullable is preferred over optional.
We also remove the volatile keyword as that does not have
the desired effect on mutable objects (the reference to the object
will be volatile but the state of the object can still get out of date)

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix: remove volatile keyword that was not doing what we expected 
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/map/download/MapDownloadProgressListener.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/map/download/MapDownloadProgressListener.java
@@ -1,10 +1,10 @@
 package games.strategy.engine.framework.map.download;
 
 import java.util.Optional;
+import javax.annotation.Nullable;
 import javax.swing.JProgressBar;
 import javax.swing.SwingUtilities;
 import org.apache.commons.io.FileUtils;
-import org.triplea.java.OptionalUtils;
 
 /**
  * A listener of map download progress events that updates the associated controls in the UI.
@@ -17,8 +17,8 @@ final class MapDownloadProgressListener {
   private static final int MAX_PROGRESS_VALUE = 100;
 
   private final DownloadFileDescription download;
-  private volatile Optional<Long> downloadLength = Optional.empty();
   private final JProgressBar progressBar;
+  @Nullable private Long downloadLength;
 
   MapDownloadProgressListener(
       final DownloadFileDescription download, final JProgressBar progressBar) {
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/engine/framework/map/download/MapDownloadProgressListener.java
```diff
@@ -73,12 +74,13 @@ private void updateProgressBarWithPercentComplete(
 
   void downloadUpdated(final long currentLength) {
     final String toolTipText = String.format("Installing to: %s", download.getInstallLocation());
-    OptionalUtils.ifPresentOrElse(
```
RoiEXLab: I just realised we no longer need this helper method at all, this is a method of Optional that was introduced in a java version 9-11

DanVanAtta: Indeed, it's nice to see the optional API expanded as much as it has.

game-core/src/main/java/games/strategy/engine/framework/map/download/MapDownloadProgressListener.java
```diff
@@ -17,8 +17,8 @@
   private static final int MAX_PROGRESS_VALUE = 100;
 
   private final DownloadFileDescription download;
-  private volatile Optional<Long> downloadLength = Optional.empty();
   private final JProgressBar progressBar;
+  @Nullable private Long downloadLength;
```
ssoloff: Note that, with the above change, instances of this type are no longer thread-safe, as stated in the class Javadocs.

`downloadLength` is written by one thread and read by another.  While reading/writing an object reference is atomic in the sense that it guarantees no partial read/write from/to memory, changes to the reference made by one thread are not guaranteed to be visible to other threads without some kind of synchronization (so-called "safe publication").  Please see sections 3.4 and 3.5 of _Java Concurrency in Practice_ (specifically, section 3.4.2: _Example: Using Volatile to Publish Immutable Objects_, applies here).

DanVanAtta: I agree, we have a mutable reference to an immutable object - added volatile back in: https://github.com/triplea-game/triplea/pull/5617

PR #5601, opened 2019-11-26
Remove 'ifPresentOrElse' utility

DanVanAtta:
The feature is now built in directly to the JDK, we can use
    'Optional.ifPresentOrElse' instead of our own helper method.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done


DanVanAtta:
Agree we could probably make more progress on removal. This one was particularly clean and relevant. I'll leave it up for grabs if anyone wants to follow-up.

PR #5607, opened 2019-11-27
Remove volatile keyword on mutable objects

DanVanAtta:
Being flagged as a bug by sonar analysis

> Marking an array volatile means that the array itself will always be read fresh and never thread cached, but the items in the array will not be. Similarly, marking a mutable object field volatile means the object reference is volatile but the object itself is not, and other threads may not see updates to the object state.

> This can be salvaged with arrays by using the relevant AtomicArray class, such as AtomicIntegerArray, instead. For mutable objects, the volatile should be removed, and some other method should be used to ensure thread-safety, such as synchronization, or ThreadLocal storage.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadFile.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadFile.java
@@ -25,7 +25,7 @@ final class DownloadFile {
 
   private final DownloadFileDescription download;
   private final DownloadListener downloadListener;
-  private volatile DownloadState state = DownloadState.NOT_STARTED;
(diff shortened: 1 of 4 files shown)


DanVanAtta:
Good points. Though, it seems that `volatile` for mutable objects is not really doing anything, just misleading. I'm generally skeptical on most locking and thread-safety code in TripleA, much is just sprinkled in. Regardless, I'd agree a better fix would be worth taking up.

PR #5609, opened 2019-11-27
Fix arithmetic expressions done as integer

DanVanAtta:
Fixes a few bugs flagged by sonar.

> When arithmetic is performed on integers, the result will always be an integer. You can assign that result to a long, double, or float with automatic type conversion, but having started as an int or long, the result will likely not be what you expect.

> For instance, if the result of int division is assigned to a floating-point variable, precision will have been lost before the assignment. Likewise, if the result of multiplication is assigned to a long, it may have already overflowed before the assignment.

> In either case, the result will not be what was expected. Instead, at least one operand should be cast or promoted to the final type before the operation takes place.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/ui/GameChooser.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/ui/GameChooser.java
@@ -66,7 +66,7 @@ public class GameChooser extends JDialog {
     leftPanel.setLayout(new GridBagLayout());
     final JLabel gamesLabel = new JLabel("Games");
     gamesLabel.setFont(
-        gamesLabel.getFont().deriveFont(Font.BOLD, gamesLabel.getFont().getSize() + 2));
+        gamesLabel.getFont().deriveFont(Font.BOLD, gamesLabel.getFont().getSize() + 2.0F));
     leftPanel.add(
         gamesLabel,
         new GridBagConstraints(
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPurchaseOption.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPurchaseOption.java
@@ -78,9 +78,9 @@ public class ProPurchaseOption {
     if (isInfra) {
       hitPoints = 0;
     }
-    attack = unitAttachment.getAttack(player) * quantity;
+    attack = (double) unitAttachment.getAttack(player) * quantity;
     amphibAttack = attack + 0.5 * unitAttachment.getIsMarine() * quantity;
-    defense = unitAttachment.getDefense(player) * quantity;
+    defense = (double) unitAttachment.getDefense(player) * quantity;
     transportCost = unitAttachment.getTransportCost() * quantity;
     carrierCost = unitAttachment.getCarrierCost() * quantity;
     isAir = unitAttachment.getIsAir();
@@ -255,7 +255,8 @@ public class ProPurchaseOption {
     final double distance = Math.max(0, enemyDistance - 1.5);
     final int moveValue = isLandTransport ? (movement + 1) : movement;
     // 1, 2, 2.5, 2.75, etc
-    final double moveFactor = 1 + 2 * (Math.pow(2, moveValue - 1) - 1) / Math.pow(2, moveValue - 1);
+    final double moveFactor =
+        1.0 + 2.0 * (Math.pow(2, moveValue - 1.0) - 1.0) / Math.pow(2, moveValue - 1.0);
     return Math.pow(moveFactor, distance / 5);
   }
 
--- a/game-core/src/main/java/games/strategy/triplea/ui/RouteDescription.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/RouteDescription.java
@@ -53,7 +53,7 @@ class RouteDescription {
     diffX *= diffX;
     int diffY = end.y - other.end.y;
     diffY *= diffY;
-    final int endDiff = (int) Math.sqrt(diffX + diffY);
+    final int endDiff = (int) Math.sqrt((double) diffX + diffY);
     return endDiff < 6;
   }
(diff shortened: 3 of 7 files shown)


game-core/src/main/java/games/strategy/engine/framework/ui/GameChooser.java
```diff
@@ -66,7 +66,7 @@ private void layoutComponents() {
     leftPanel.setLayout(new GridBagLayout());
     final JLabel gamesLabel = new JLabel("Games");
     gamesLabel.setFont(
-        gamesLabel.getFont().deriveFont(Font.BOLD, gamesLabel.getFont().getSize() + 2));
+        gamesLabel.getFont().deriveFont(Font.BOLD, gamesLabel.getFont().getSize() + 2.0F));
```
RoiEXLab:
```suggestion
        gamesLabel.getFont().deriveFont(Font.BOLD, gamesLabel.getFont().getSize() + 2f));
```
I don't think we use `F` for floats, just `f`?
In any case the trailing `.0` is not required

game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPurchaseOption.java
```diff
@@ -255,7 +255,8 @@ private double calculateLandDistanceFactor(final int enemyDistance) {
     final double distance = Math.max(0, enemyDistance - 1.5);
     final int moveValue = isLandTransport ? (movement + 1) : movement;
     // 1, 2, 2.5, 2.75, etc
-    final double moveFactor = 1 + 2 * (Math.pow(2, moveValue - 1) - 1) / Math.pow(2, moveValue - 1);
+    final double moveFactor =
+        1.0 + 2.0 * (Math.pow(2, moveValue - 1.0) - 1.0) / Math.pow(2, moveValue - 1.0);
```
RoiEXLab: Haven't checked, but maybe when using `1d` and `2d` instead of `1.0` and `2.0` it might fit into a single line?
But not really important, similar thing like the style preference earlier.
It would be interesting to see if we consistently use one form over the other, or it's just a 50:50 chance each time

game-core/src/main/java/games/strategy/triplea/ui/RouteDescription.java
```diff
@@ -53,7 +53,7 @@ public boolean equals(final Object o) {
     diffX *= diffX;
     int diffY = end.y - other.end.y;
     diffY *= diffY;
-    final int endDiff = (int) Math.sqrt(diffX + diffY);
+    final int endDiff = (int) Math.sqrt((double) diffX + diffY);
```
RoiEXLab: Ok, this is really picky here, I mean I get why sonarqube complains, but I don't think this will ever become an issue, nobody expects diffX + diffY to be out of range for an int, but in range of a float.
That's a good example why I'm not 100% agreeing with sonarqube everytime

game-core/src/main/java/games/strategy/engine/framework/ui/GameChooser.java
DanVanAtta: We have more `f` than uppercase `F`, but it is mixed.
The `.0` I like as `1f` and `2f` look like hex numbers.

game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPurchaseOption.java
DanVanAtta: It appears we have no examples of `1d` or `2d`, we have examples of `1.0d`
`1.0` AFAIK is a double, the `1.0d` should be unnecessary.

TBH I think there are other issues beyond trying to have this fit on one line. The numerator is seemingly magic, the addition and subtraction operations do not make sense off-the-bat, this takes a good bit of reverse engineering to understand. Focusing on formatting IMO is going to be the wrong focus for right now.

game-core/src/main/java/games/strategy/triplea/ui/RouteDescription.java
DanVanAtta: > That's a good example why I'm not 100% agreeing with sonarqube everytime

There is tuning to be done to downgrade or upgrade various alarms to account for false positives. On the other topic, arguing with static analysis IMO only goes so far. While yes, we can say we know more about the code than analysis and frankly are "smarter", on the other hand there is something to be said about code that is simple enough and valid enough that it passes any/all static analysis. 

While TripleA is not safety critical, the following was an interesting read and the practice I think follows for most any software project to minimize defects:

> The Power of Ten – Rules for Developing Safety Critical Code
> NASA/JPL Laboratory for Reliable Software 

> 10.Rule: All code must be compiled, from the first day of development, with allcompiler warnings enabled at the compiler’s most pedantic setting. All code must compile with these setting without any warnings. All code must be checked daily with at least one, but preferably more than one, state-of-the-art static source code analyzer and should pass the analyses with zero warnings.  

http://pixelscommander.com/wp-content/uploads/2014/12/P10.pdf

DanVanAtta:
Going to merge, down to just one item `2.0f` vs `2.0F`

I don't think it's worth the 5 minutes to fix.

If we want to stay consistent, @RoiEXLab I recommend we open a convention about this, or otherwise not care. 

@RoiEXLab I wonder as well if we should favor opening convention issues rather than PR commenting on the nitpicky stuff. I don't mind hashing it out, but it seems like it may take some time to work through a very long laundry list of conventions. I also think we should keep in mind the value/focus of a PR, there were some real problems with arithmetic operations done as integer, we should probably be focused on whether that was done well and any stylistic items extract to an issue conversation which can then be enforced or mass updated later.

PR #5618, opened 2019-11-27
Travis update: Run sonar only on master

DanVanAtta:
- Sonar uploading reports on PR builds is a problem as fixes will
    be reported and then any subsequent run for merged builds might
    not include those fixes, showing a false regression.
    
    - Combine jacoco test coverage step with 'test and integTest' step
    
    - Clean up description labels further, the JDK language is displayed
    already.
 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

## Additional Review Notes

One of the more salient reasons for this update, sonar is uploading data on each PR. This gives us an inconsistent history as when a PR builds, potentially reducing counts, we'll see less bugs in the sonarcloud dashboard. Then, if any other PR is built, or if master is built, it'll upload results that won't include the bug fixes. This results in slightly confusing valleys in the graphs of bugs over time:

Screenshot from 2019-11-27 13-49-16

For some consideration, we'll be adding another job for master branch builds, it's tempting to see if we can consolidate some of the jobs as Travis only runs so many "VMs" in parallel. Perhaps with some additional context/history we'll have better insight to how to balance the build jobs. It is  something for us to keep in mind as we move forward

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ jobs:
         - .travis/check-links || travis_terminate 1;
         - ./gradlew --parallel validateYamls spotlessCheck checkstyleMain checkstyleTest pmdMain pmdTest
     - stage: verify
-      env: "JDK11 Tests"
+      env: "Unit & Integration Tests"
       jdk: openjdk11
       cache:
         directories:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -62,9 +62,10 @@ jobs:
       install: skip
       script:
         - ./.travis/setup_database &
-        - ./gradlew --parallel test integTest
+        - ./gradlew --parallel test integTest jacocoTestReport
+        - bash <(curl -s https://codecov.io/bash)  # upload coverage report - https://github.com/codecov/example-gradle
     - stage: verify
-      env: "JDK12 Unit Tests"
+      env: "Unit Tests"
```
DanVanAtta: The unit and integ tests take 11 minutes, the rest of the stages in verify take about 7. 
If we remove the JDK12 tests, I think we would probably benefit to stay with 4 concurrent stages and move the integ tests to their own stage to help reduce that 11 minute runtime.

PR #5621, opened 2019-11-28
Add Client components for RemoteAction polling

DanVanAtta:
- Adds classes to be used by client (in this case game-server-hosts)
to query for players that have been banned or disconnected.
- Any banned/disconnected player will then be disconnected from the
game server host.
- Also included is a method to check if a player is banned, this
will be used to check new players joining a server to see if they
are allowed to join.

Next steps:
- this is part of a multi-part update and is not yet used.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

## Additional Review Notes
https://github.com/triplea-game/triplea/pull/5616 provides updates so we can instantiate the 'poller' created in this PR in a sane place and have the needed data

Follow up updates will mostly focus on further backend updates:
- code to connect this PR with #5616
- implementation of remote actions client
- implementation of http server moderator actions controller

### DIFF
--- a/game-core/src/test/java/org/triplea/game/server/ban/check/PlayerDisconnectActionTest.java
+++ b/game-core/src/test/java/org/triplea/game/server/ban/check/PlayerDisconnectActionTest.java
@@ -22,7 +22,7 @@ import org.triplea.domain.data.PlayerName;
 @ExtendWith(MockitoExtension.class)
 class PlayerDisconnectActionTest {
   private static final INode NODE_1 =
-      new Node("node-1", InetSocketAddress.createUnresolved("1.1.1.1", 5000));
+      new Node("node-1", InetSocketAddress.createUnresolved("99.99.99.100", 5000));
 
   private static final String NODE_2_NAME = "node-2";
   private static final INode NODE_2 =
--- a/game-core/src/test/java/org/triplea/game/server/ban/check/PlayerDisconnectActionTest.java
+++ b/game-core/src/test/java/org/triplea/game/server/ban/check/PlayerDisconnectActionTest.java
@@ -1,0 +1,108 @@
+package org.triplea.game.server.ban.check;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.atLeastOnce;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import games.strategy.net.INode;
+import games.strategy.net.IServerMessenger;
+import games.strategy.net.Node;
+import java.net.InetSocketAddress;
+import java.util.Set;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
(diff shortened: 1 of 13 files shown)


game-core/src/test/java/org/triplea/game/server/ban/check/PlayerDisconnectActionTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.triplea.domain.data.PlayerName;
+
+@ExtendWith(MockitoExtension.class)
+class PlayerDisconnectActionTest {
+  private static final INode NODE_1 =
+      new Node("node-1", InetSocketAddress.createUnresolved("1.1.1.1", 5000));
```
RoiEXLab: Side-Note:
1.1.1.1 is an actual IP that's being used. See https://1.1.1.1

http-clients/src/main/java/org/triplea/http/client/lobby/game/remote/actions/RemoteActions.java
```diff
@@ -0,0 +1,29 @@
...
+@EqualsAndHashCode
+public class RemoteActions {
+  private final Collection<BannedPlayer> bannedPlayers;
+  private final Collection<String> serversToShutdown;
+
+  public boolean serverIsRequestedToShutdown(final InetAddress myPublicIpAddress) {
+    return serversToShutdown != null
+        && serversToShutdown.stream()
+            .map(IpAddressParser::fromString)
+            .anyMatch(myPublicIpAddress::equals);
```
RoiEXLab: Hmm it looks like most of the implementations of `serversToShutdown` are sets.
Would it work converting InetAddress to a string an performing a `serversToShutdown.contains(myPublicIpAddress.toString())`?
Alternatively converting serversToShutdown to a collection of InetAddresses upon creation?

DanVanAtta: I'm afraid neither may work. For Strings, IPV6 can be represented in multiple formats, so string representations could be brittle. As `RemoteActions` can be sent as  JSON payload, I'd rather avoid using nested objects where we can.

game-core/src/test/java/org/triplea/game/server/ban/check/PlayerDisconnectActionTest.java
DanVanAtta: Thanks, probably better to keep the value to be a true dummy value: 0cb3424bd

http-clients/src/main/java/org/triplea/http/client/lobby/game/remote/actions/RemoteActions.java
RoiEXLab: Bummer, but ok

PR #5622, opened 2019-11-28
Replace HashedMac with SystemId

DanVanAtta:
Overall removes HashedMac and replaces with SystemId.
A key benefit is we no longer call external applications,
a requirement for listing in Steam. The previous MacFinder
code, as a fallback, called things like 'ipconfig' to find mac.

Change Overview:
- Replace MacFinder with SystemIdLoader
- Extract SystemId loading from 'SystemIdHeaders'
  to new class 'SystemIdLoader'

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

## Additional Review Notes
- 'SystemIdLoader' is mostly all moved code from 'SystemIdHeaders', same for the test code for it.


domain-data/src/main/java/org/triplea/domain/data/SystemIdLoader.java
```diff
@@ -0,0 +1,60 @@
...
+import lombok.extern.java.Log;
+
+/** Loads a SystemId from persistence. */
+@UtilityClass
+@Log
+public class SystemIdLoader {
+  private static final String SYSTEM_KEY = "system-id-key";
+
+  @Setter(value = AccessLevel.PACKAGE, onMethod_ = @VisibleForTesting)
+  private static PreferencesPersistence preferencesPersistence =
```
DanVanAtta: The code in this class is largely just moved from [SystemIdHeader](https://github.com/triplea-game/triplea/pull/5622/files#diff-81cfc0624ee3ee51ecd9cea802dd76c1), and instead of creating a Map of system id key to system id, this class just focuses on returning (and/or generating) a system id key.

game-core/src/main/java/games/strategy/engine/framework/startup/login/ClientLoginValidator.java
```diff
@@ -92,8 +91,6 @@ public String verifyConnection(
 
     if (hashedMac == null) {
       return ErrorMessages.UNABLE_TO_OBTAIN_MAC;
-    } else if (!MacFinder.isValidHashedMacAddress(hashedMac)) {
```
DanVanAtta: System id does not have the same hashing applied to it, as it's random. Since the system id is client-side generated, any validity requirements based on hashing can be elided. Hence, we're making it more difficult for system-id/mac to be spoofed, but ultimately we're not really adding security.

I'm curious if we think we should try to make it a bit harder for someone. The scenario is basically someone changes their system id by custom compiling code, do we think that a hash guarantee is going to actually stop them if they get an error for invalid system id, or will the attacker be sophisticated enough already to hash their injected value so it passes?

game-core/src/main/java/games/strategy/net/ClientMessenger.java
```diff
@@ -83,9 +84,7 @@ public ClientMessenger(
           }
         } catch (final ConnectException e) {
           throw new RuntimeException(
-              String.format(
-                  "Could not connect host: %s, port: %s, name: %s, mac: %s", host, port, name, mac),
-              e);
+              String.format("Could not connect host: %s, port: %s", host, port), e);
```
DanVanAtta: name and mac are removed from this message, it's not yet used and not part of trying to connect to a remote host. Instead of updating the 'mac' name here, it's just removed.

game-core/src/test/java/games/strategy/engine/vault/VaultTest.java
```diff
@@ -71,6 +72,8 @@ void testLocal() throws Exception {
    * Passes when run individually. Fails when run as part of a suite that consists of multiple
    * server/vault tests.
    */
+  @Disabled
```
DanVanAtta: Noticed these tests that were very old-school disabled, instead of having had '@Ignore' put on them, they had the '@Test' removed. The '@Disabled' is added so we see the disabled test and have a better chance to address the non-determinism or just drop the test code altogether.

game-core/src/main/java/games/strategy/engine/framework/startup/login/ClientLoginValidator.java
RoiEXLab: I think we can all agree here that the system can be tricked with ease if someone actually tried, so I wouldn't bother too much with the details.
The only people getting to this point, won't get tricked by a simple additionall step

PR #5626, opened 2019-11-29
Reduce game runner dependencies

RoiEXLab:
This PR tries to remove some of the dependencies to GameRunner in order to be able to make everything more modular in the future.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done

~Haven't tested the changes yet, thus the draft status~


RoiEXLab:
If you have alternative suggestions to solve this Problem, I'd like to hear them

game-core/src/main/java/games/strategy/engine/data/properties/FileProperty.java
```diff
@@ -121,31 +124,39 @@ private static File getFileUsingDialog(final String... acceptableSuffixes) {
...
     }
-    final Optional<File> selectedFile =
-        GameRunner.showFileChooser(
-            new FileFilter() {
-              @Override
-              public boolean accept(final File file) {
-                if (file == null) {
-                  return false;
-                } else if (file.isDirectory()) {
+    Optional<File> selectedFile;
```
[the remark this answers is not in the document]
RoiEXLab: Good catch, I'll try to fix this

game-core/src/main/java/games/strategy/engine/data/GameParser.java
```diff
@@ -938,9 +937,6 @@ private void parseEditableProperty(
       case "boolean":
         editableProperty = new BooleanProperty(name, null, Boolean.parseBoolean(defaultValue));
         break;
-      case "file":
-        editableProperty = new FileProperty(name, null, defaultValue);
-        break;
```
RoiEXLab: This is the only sort-of breaking change in this PR (and the resulting change in game.dtd).
None of the existing maps in `triplea-maps` seem to use this feature at all.
I highly doubt this really works, because the file is internally represented as `java.io.File` which can't represent files within a zip file at all, so all this property could do is reference files somewhere else on a users PC which likely isn't what we intend to do with such a property. I'm not 100% sure on this one, but AFAIK this property would need to be used by the engine explicitly (can't be referenced by the map itself somewhere), so all that really breaks is a feature that was no-op before, right?

I'd like to hear your opinion on this @ron-murhammer

RoiEXLab: Also I have a batch of other "sort-of" breaking changes that I will submit after this PR to avoid merge conflicts

ron-murhammer: @RoiEXLab This is fine to remove. I'm fairly sure that the FileProperty was only used by one of the old puzzle games that we removed support for already. It was used to load a background image or something like that.

RoiEXLab: Ah thanks for the clarification.
That makes sense 👍

DanVanAtta:
Good work @RoiEXLab

PR #5632, opened 2019-11-30
Configure code-climate third party plugins

DanVanAtta:
Adds additional plugins to code climate
for additional linting.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

## Additional Review Notes
Adding an initial config that sounds like it should be right, some iteration may be needed.

The full list of analysis plugins is available here: https://docs.codeclimate.com/docs/list-of-engines

The plugins selected are the ones from that list that seem like they could be of use for us.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -2,7 +2,7 @@ plugins:
   # https://docs.codeclimate.com/docs/git-legal
   git-legal:
     enabled: true
-      allow_strong_copyleft: true
+    allow_strong_copyleft: true
   # https://docs.codeclimate.com/docs/fixme
   fixme:
     enabled: true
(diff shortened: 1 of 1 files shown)


.codeclimate.yml
```diff
@@ -0,0 +1,17 @@
+plugins:
+  # https://docs.codeclimate.com/docs/git-legal
+  git-legal:
+    enabled: true
+      allow_strong_copyleft: true
```
RoiEXLab:
```suggestion
    allow_strong_copyleft: true
```
This should be what causes the build to fail

DanVanAtta:
Looks like these plugins find a large number of items. One of which are the TODOs, so we'll finally have a list of them that we can start working off.

The licensing plugin is worth a double check, I think the 'strong copyleft' setting is something we want as we are GPL

PR #5635, opened 2019-12-01
Remove CollectionProperty

RoiEXLab:
This is a safe change, this class was only used to create an editor for a `Collection` field in `tools.map.making.MapProperties`.
Because no field exists, this is defacto dead code.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature update or enhancement
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done

PR #5636, opened 2019-12-01
Remove ComboProperty

RoiEXLab:
This is a potentially breaking change, similar to the one in #5626, potentially related to #5635.
I couldn't find a map (via github search) that actually uses this property so I decided to remove it. If this is required for some reason

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done


DanVanAtta:
Do existing maps use the DTD that is on master?
If so, DTD changes could break those maps, even on old versions of TripleA. Please let me know if I'm missing something or wrong about that. I recall vaguely that DTD updates were particularly painful because of that (and for some of those reasons, I'd be in favor of us removing a TripleA DTD altogether so we would not have those issues, but unfortunately DTD injects code which we'd have to add to each map)

RoiEXLab:
@DanVanAtta existing maps use the dtd that is shipped with the engine that loads the map, that's why I consider this a "potentially breaking change".
However I didn't find any maps that use this property, so it won't affect any existing official maps if I didn't miss something.

DanVanAtta:
I agree this is not used. Kinda curious what manual testing looked like @RoiEXLab , but does not matter too much.

```
~/work/maps$ find . -name "*xml" -type f  | while read file; do  grep -F "<list" "$file"; done | wc -l
0
```

RoiEXLab:
> Kinda curious what manual testing looked like @RoiEXLab , but does not matter too much.

Almost expected this answer 😅 
Really I only verified the map making tools still work no problem (even though this change shouldn't have affected them at all), and "manually verified" that this is not used in any map i.e. I didn't write a script for this, I used github search and checked the 15 or so occurences

PR #5640, opened 2019-12-02
Travis check scripts: Unmatched tests

DanVanAtta:
- Adds two checker scripts to travis build and the `verify` script to look for tests and source files that do not have matching names or do not have matching packages.
- Fixed any issues found

### find-unmatched-test-packages

- loops over all source files
- computes an expected test file name
- looks for that test file name
- if found, checks the package of the test file name against the package of the source file
- if the two packages do not match, we report the source file and test file as non-matching

### find-orphaned-tests
- loops over all test files
  - filters out tests in `integration-testing`, any tests marked with  `@Integration`, `@ExtendedUnitTest` and filters out abstract tests
- for each test file, computes an expected source file path
- checks if the expected source file exists
- if the expected source file does not exist, reports the test file and the expected source file
 
## Fix summary
- A number of unit tests were not renamed to match the source file
- Md5Crypt and its test were flagged, Md5Crypt in game-core is unused and was removed.
- A number of tests were in the wrong package and were moved to match

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[x] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done

Ran scripts locally and fixed any issues. Verified scripts had an exit code of `1` if any issues were found and an exit code of `0` when all issues were fixed.

### find-unmatched-test-packages
This was the initial output:
Screenshot from 2019-12-01 13-23-51

I added a "how to fix" message, with that message the output looks like this:
Screenshot from 2019-12-01 16-22-14

### find orphaned tests

Screenshot from 2019-12-01 15-36-42

## Additional Review Notes
- Find-orphaned-tests adds constraints on how we can name tests by creating a dependency between test name and source file. So far that can be hacked around by using `@ExtendedUnitTest`. I suspect the meaning of that annotation might evolve.


.travis/find-unmatched-test-packages
```diff
@@ -0,0 +1,62 @@
...
+
+       ## chop off the filename and the '/src/test/java/' prefix
+       ## EG: ./project/src/test/java/org/triplea/package/File.java -> org/triplea/package
+       testPackage=$(echo "$testFileMatch" | \
+            sed 's|\(.*\)/.*\.java$|\1|' | sed 's|.*/java/||')
+
+       ## chop off the filename and the '/src/main/java/' prefix
+       ## EG: ./project/src/main/java/org/triplea/package/File.java -> org/triplea/package
+       srcPackage=$(echo "$file" | \
+            sed 's|\(.*\)/.*\.java$|\1|' | sed 's|.*/java/||')
```
RoiEXLab: This regex feels a little bit weird:
```regex
(.*)/.*\.java$
```

Because in theory for a string `/a/b/c.java`
Group 1 could be either one of
- `/a/b`
 - `/a`
 - ` `

Maybe there's a convention to regex engines, but I'd really feel a little bit safer if the
regex would look a little bit more like this:
```regex
(.*)/[^/]*\.java$
```
or
```regex
(.*)/.*?\.java$
```
Not sure what sed supports.
Same thing a couple of lines above

DanVanAtta: I assume you're referring to the first regex @RoiEXLab .
`sed` does longest possible match. For `/a/b/c.java`, group 1 will be: `a/b`
I dont' quite follow the examples given, I feel like I should probably get more up to speed with javascript and/or perl regexes, I'm very used to GNU regex. Could you explain how those variants would be more robust?

RoiEXLab: Yes,
```regex
[^/]
```

Would match all characters except the forward slash, so it is guaranteed to be just the filename without any of the directory stuff.
```regex
.*?
```
The `?` after the `*` is the lazy operator which explicitly makes this part try to match as few characters as possible, which will achieve the same result a little bit more explicit and not possibly rely on some undefined regex engine behaviour (maybe a different linux distro uses some other implementation of sed?)

DanVanAtta: The first suggestion which guards against non-greedy matchers works.
```
dan@dan-desk:~/work/triplea$ echo "./test-common/src/main/java/org/triplea/test/common/security/TestSecurityUtils.java" | sed 's|\(.*\)/[^/]*\.java$|\1|'
./test-common/src/main/java/org/triplea/test/common/security
```

Unless I manged something, the second one does not:
```
dan@dan-desk:~/work/triplea$ echo "./test-common/src/main/java/org/triplea/test/common/security/TestSecurityUtils.java" | sed 's|\(.*\)/.*?\.java$|\1|'
./test-common/src/main/java/org/triplea/test/common/security/TestSecurityUtils.java
```

`sed` is greedy matching, that should not vary by distro. 

http://sed.sourceforge.net/sedfaq5.html

>  The RE '.*' is designed to be "greedy" (i.e., matching as many characters as possible)

5.4. My RE isn't matching/deleting what I want it to. (Or, "Greedy vs. stingy pattern matching")

PR #5642, opened 2019-12-02
Add custom style checks to build

DanVanAtta:
Custom script added to Travis to enforce team style conventions.

Items checked for:
- Use Java9 collection methods like "List.of" instead of "Collections.emtpyList"
- Avoids legacy collections like "Vector" and "hashTable"
- Collections are assigned to their interface type, not concrete, eg: "List<String> list = "
- SQL tables are singular
- SQL has no camel-casing

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[x] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done
- Ran script to verify
- injected items that it should catch and verified
- verified exit code is 1 when items are caught

## Screens Shots
Screenshot from 2019-12-02 01-04-52
Screenshot from 2019-12-02 01-05-00

## Additional Review Notes
Draft for now to avoid merge conflicts and to ensure consensus before issues are fixed.


DanVanAtta:
Crud, didn't click 'Draft' button, marking the title as 'DRAFT'..

DanVanAtta:
Latest output:
Screenshot from 2019-12-02 17-44-11

I think the set of checks is about stabilitized.

@RoiEXLab , @ron-murhammer , please have a look at the check list: https://github.com/triplea-game/triplea/pull/5642/files#diff-526da0609b1ce2f46bdea160d082d57fR21, let me know if they look good.

DanVanAtta:
Found another check to add, `Comparator` specifies that the parameter should be `@nonnull` or else it throw a NPE. 
Screenshot from 2019-12-02 22-19-12

DanVanAtta:
@ron-murhammer @RoiEXLab , I'd like the configured rules to be confirmed before fixing the java code violations. We should probably be sure to give quick reviews to draft PRs, they are only opened for the purpose of getting feedback, if we wait for it to be finalized, it defeats the purpose of opening the draft in the first place.

RoiEXLab:
Ah, sorry I missed this.
I'm fine with the rules, however I want to point out that there are comparators that can handle null properly.
See https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Comparator.html#nullsFirst(java.util.Comparator)

asvitkine:
> Found another check to add, `Comparator` specifies that the parameter should be `@nonnull` or else it throw a NPE.
> Screenshot from 2019-12-02 22-19-12

I'm confused, I thought we preferred types to be marked `@Nullable` and types not marked that should be assumed to be non-null (in new code).

DanVanAtta:
I might have been looking at an old javadoc, JDK11 states this on the compareTo method:
```
NullPointerException - if an argument is null and this comparator does not permit null arguments
```

Noting **this comparator does not permit null arguments**


I'll remove the `@Nonnull` check as it looks like nullable is now okay. I do get an IDE warning that the parameter is still marked as `@Nonnull` on the parent method.

> I'm confused, I thought we preferred types to be marked @Nullable and types not marked that should be assumed to be non-null (in new code).

@asvitkine generally yes. We very much want `@Nullable` where it applies. `@Nonnull` usages is either for helping lombok and code generation, or if we just want to be really explicit about a value not being null.

In this case there is an IDE warning on `comparator` stating that the override is changing the parent method `@Nonnull` parameter to be implicitly nullable. I think that could be a Liskov Substitution principle violation. Regardless, the JDK 11 javadoc, and existence of comparators that handle null make this seem to be an over-constraint.

DanVanAtta:
I disabled the failing checks, will folllow-up with iterations that fix and enable the disabled checks one by one.

Issue #5646, opened 2019-12-03
"Must complete Battle in 110 Sea Zone first" shown in all-AI game


asvitkine:
**Describe the bug**
This is on tip of tree, playing all Hard AI game of Global 2nd edition.
 
**Engine Version or SHA**
c3709c28a46f060b96cfd65e3c80f6337ac5c357

**To Reproduce**
1. Run a Global 2nd edition game with all players set to Hard AI.

asvitkine:
To repro, you may need to get this fix in first or else it might hit an issue earlier: https://github.com/triplea-game/triplea/pull/5647

Adding instrumentation, the battle being attempted is in Normandy Bordeaux. But actually looking at the AI moves in question, it's not obvious why the sea battle must be completed first. In my example game, the units attacking Normandy are unrelated to the units in the sea battle - they consist of units just moving from land territories and not involving the sea battle.

Perhaps the engine is incorrectly associating the sea battle as a dependency?

DanVanAtta:
Is this the same problem as: https://github.com/triplea-game/triplea/issues/5216 @asvitkine ?

> When progressing to combat phase, a pop-up appears warning player should fight one battle first, but combat window is also opened and neither the combat window nor the warning dialog can be closed.

asvitkine:
I'm not sure it's the same. In this case, I didn't observe any freeze, just this popup but the AI game kept going.

asvitkine:
For reference, it doesn't happen with 16749 - which is an old build I had around.

I'm going to try bisecting.

asvitkine:
Looks like it's not deterministic because some games, Germany simply doesn't attack Normandy on turn 1. So not obvious when it started. I'll try to just debug it instead of bisecting.

asvitkine:
Interesting, if I load up the game after germany does its combat move when the bug happens, and play the turn out as a human, I don't get this dependent battle warning.

asvitkine:
Ok, so the cause seems to be this logic related to scrambling:
```
      // now make sure any amphibious battles that are dependent on this 'new' sea battle have their
      // dependencies set.
      if (to.isWater()) {
        for (final Territory t : data.getMap().getNeighbors(to, Matches.territoryIsLand())) {
          final IBattle battleAmphib = battleTracker.getPendingBattle(t, false, BattleType.NORMAL);
          if (battleAmphib != null) {
            if (!battleTracker.getDependentOn(battle).contains(battleAmphib)) {
[...]

PR #5653, opened 2019-12-04
Fix dependent territory scramble and unmodifiable collection error.

asvitkine:
This PR fixes https://github.com/triplea-game/triplea/issues/5646, where an incorrect battle dependency was being added due to scrambling, which also caused dependent error popups with AI players.

The old logic was as follows:

```When scrambling to a sea territory, mark all adjacent land battles as dependent on it.```

That was not correct, as it did not take into account if adjacent land battles actually were amphibious battles from the sea territory to which units were being scrambled. This was a long standing bug that started to be triggered in turn 1 of Global 2nd edition due to new AI decisions.

In addition to this fix, also fixes a different recent bug with an unmodifiable collection being modified.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix:  https://github.com/triplea-game/triplea/issues/5646
[] Other:   

## Testing

[X] Manual testing done

Did the following manual tests:

On Global 2nd edition:

1. As Germany on 1st turn, send a ship to Sea Zone 110 and a land unit to Normandy.
2. As UK, scramble planes to defend.
3. In combat move, make sure that Normandy battle is not dependent on Sea Battle by doing it first.

Second test also on Global 2nd edition:

1. As Germany on 1st turn, use edit mode to remove enemy units from Sea Zone 110 and add a german transport to adjacent sea zone.
2. As combat move, load a unit on the transport, move to SZ110 and attack Normandy with it.
3. As UK, scramble planes to defend.
4. In this case, the Normandy battle should depend on SZ110 battle unlike in the previous case.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/BattleDelegate.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/BattleDelegate.java
@@ -875,7 +875,7 @@ public class BattleDelegate extends BaseTripleADelegate implements IBattleDelega
             }
             // after that is applied, we have to make a map of all dependencies
             final Map<Unit, Collection<Unit>> dependenciesForMfb =
-                TransportTracker.transportingWithAllPossibleUnits(attackingUnits);
+                new HashMap<>(TransportTracker.transportingWithAllPossibleUnits(attackingUnits));
             for (final Unit transport :
                 CollectionUtils.getMatches(attackingUnits, Matches.unitIsTransport())) {
               // however, the map we add to the newly created battle, cannot hold any units that are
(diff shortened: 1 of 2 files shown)


asvitkine:
@DanVanAtta thoughts on the codeclimate warnings? I can change the code to use a continue statement on null battle to reduce the nesting, but this would be changing the existing control flow - since my current change is not touching that.

ron-murhammer:
@panther2 Can you confirm this is a bug in Global 40 in the current stable release?

@asvitkine I'm fine with the code climate warnings as you aren't really touching that code and this is small bug fix.

DanVanAtta:
Re: code-climate, new code we should probably keep clean. Touching code with an existing warning seems okay to leave be.

panther2:
@ron-murhammer Indeed it is a bug as in the first above scenario the battles should occur independently. In the second scenario of course the sea battle has to be resolved at first (although in that case the attacking transport would not move there as it cannot win the sea battle).

DanVanAtta:
Thank you for the research @panther2 

Side-note, if we have lots of code-climate issues, and if we are constantly touching code that is getting flagged, and ignore because it is existing, I wonder if we can re-arrange that process somehow so we can fix those issues without burdening PRs too much. Technical debt is not fun stuff.

DanVanAtta:
Good work finding and fixing this problem @asvitkine, I wonder how much longer this would have been a bug had you not found it.

PR #5661, opened 2019-12-05
Enable and test http-endpoint Roles

DanVanAtta:
1. We need a jersey config to activate allowed roles checking
2. Updated ProtectedEndpoint test to verify disallowed roles

The previous behavior was that if you could authenticate then you
could access any endpoint (@AllowedRoles was not active).

The prevous protectedEndpoint test verified that an admin
could access an endpoint and that an invalid API key (could
not authenticate) could not access an endpoint. This test
setup masked the problem that roles were not being enforced,
we only verified authentication.

This update updates the protected endpoint test to accept
a new enum 'AllowedUserRole' that specifies which role
can access an endpoint. To try and keep the test permutations
constrained, we'll verify that a user with that exact role
can access that endpoint, and second we'll test that a sample
of other (valid) roles cannot access that endpoint.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

[] Manual testing done


RoiEXLab:
Something is wrong with the travis build.
From the log it looks like everything went just fine, but the last command executed exited with a non 0 exit code.

DanVanAtta:
Codecov probably should not have run, note:

```
> Task :integration-testing:integTest

org.triplea.server.lobby.game.hosting.GameHostingControllerTest > sendGameHostingRequest() FAILED

    org.triplea.http.client.HttpInteractionException: Status 500 reading GameHostingFeignClient#sendGameHostingRequest(Map)

> Task :integration-testing:integTest

105 tests completed, 1 failed
The command "./gradlew --parallel test integTest jacocoTestReport" exited with 1.
```

The above is a failure I've only seen on travis, passes locally. The 500 on server is the reason for the recent urgency to get server output back online. It's going to be nearly impossible to debug unless we can get server output.

DanVanAtta:
With output from server, was able to find the error on server:
```
ERROR [2019-12-06 02:09:27,272] org.triplea.lobby.server.db.JdbiDatabase: Exception executing SQL: insert into game_hosting_api_key(key, ip) values(:key, :ip::inet)

! org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "game_hosting_api_key_pkey"

!   Detail: Key (id)=(1) already exists.
```

This is because we create a key with this ID already:
```
game_hosting_api_key:
  - id: 1
    key: 06dbbb9b6ac87d97f9acca120ae4784d0eaf6865ea99788a389a384da8ab0709e77af2bfe4f4e82c6e6d375ae256aa95c2fa99ce97ce65981cfd1340257a441a
    # key = sha512(HOST)
    ip: "127.0.0.1"
    date_created: 2010-01-01 23:59:20.0
```
The above record is new for this update.

The reason I do *not* see this locally is because the ID sequence generator I have locally is not reset. Hence when the sequence starts out fresh, generates ID (1) for an insert, we get a collision and a UK violation.

The fix is simple enough, just had to change the ID value in the above to '80000'

Issue #5664, opened 2019-12-05
Merging Integration-Testing  & lobby-db into http-server


DanVanAtta:
Motivation:
- we may recall that when http-server is started from integration tests that Travis does not show output from it. Generally this is okay, but is a huge problem if we get a travis only 500.
- https://github.com/triplea-game/triplea/pull/5661 now is generating a travis only 500.

History:
- we had this fixed before by launching an http-server from jar
- logback pathing issues broken when we included integration-testing, all classpaths got more or less mangled together and we wind up with a classdef not found error launching the http-server
- we moved forward with the integration-testing subproject and removed the independent jar file launch and lost server output
- before integration-testing subproject, we had a nondeterministic build as DB updates when testing http-server ran they could conflict with lobby-db

Proposal:
- merge up sub-projects and re-enable jar launch of http-server

Benefits:
- fewer sub-projects. Now that we no longer have lobby using DB, having all tests depending on DB in one sub-project keeps the same solution we wanted by having all such tests in the 'integration-testing' subproject.
- better cohesion of DB code with where it is used.
- we regain output of http-server in travis

Challenges
- in-flight PRs need this debug output, yet the merge conflicts may potentially be bad. Our new post-merge review process likely will come in handy here.

RoiEXLab:
Actually I wondered why there were still 2 separate projects.
So I don't see a reason to not go for it 👍

PR #5665, opened 2019-12-05
Merge 'integration-testing' subproject into 'http-server'

DanVanAtta:
Begins execution on: https://github.com/triplea-game/triplea/issues/5664

We merge the integ-testing into http-server so that we can re-enable launching http server in travis from jar file. That will provide for us dropwizard output in the travis output logs. That is essential for debugging a travis only server 500. Recalling, we originally had a problem with integration-testing when we did this as the classpath somehow got mangled and http-server failed during execution with a ClassNotFound error for a logback class. 

Furthermore, 'lobby' sub-project is now dropped, so we can simplify the sub-project structure a bit, we now have this option as before we did not.

One cost to this update is we can not and no longer should run http-server unit tests in parallel as we will then have db unit tests running with that. 

Some key benefits:
- we had a duplicated drop wizard config in http-server and integration-testing that was ideal to merge up, now those are merged
- the DB updates/testing/code is all very specific to http-server, having that in the same project is better cohesion.
- having the controller tests in the same project is also more cohesive. We can now more cleanly check test coverage by comparing the source files with test files and can better see/expect a 1:1 mapping.
- we have some simplifications with dependency management, the build.gradle files being merged means we only introduce a few new dependencies into http-server for the benefit of test, meanwhile we drop the build.gradle file from integration-testing
- integration-testing had many dependencies, notably game-core. Removing that is a sig

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[x] Other:  project restructure 

## Testing

[x] Manual testing done
- confirmed ServerApplication launcher still works
 

## Additional Review Notes
In integration-testing build.gradle we had a build exclusion for an SLF4J library. That seems to no longer be needed at all, it was not migrated into http-server build.gradole.


.travis.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         - .travis/check-links || travis_terminate 1;
         - .travis/find-unused-dbunit-datasets || travis_terminate 1
         - .travis/find-unmatched-test-packages || travis_terminate 1;
-        - .travis/find-orphaned-tests || travis_terminate 1;
+        # - .travis/find-orphaned-tests || travis_terminate 1;
```
RoiEXLab: Are you planning on uncommenting this line some time in the future or did you mean to remove this entirely?

DanVanAtta: Uncomment, it's a really useful check to have, it's helped a couple times already and on the initial create found a number of tests that were in the wrong package or were not renamed after their source file.

The tests for classes in `lobby-db` items now though are being flagged as in the wrong package, we got around that previously by excluding any test in 'integration-testing'. I'm thinking we'll merge 'lobby-db' into 'http-server' in next next steps.

PR #5667, opened 2019-12-06
Make MoveDescription do basic validation on the route parameter.

asvitkine:
Make MoveDescription do basic validation on the route parameter. This addresses a TODO from the MoveDescription refactoring, removing this validation from the AI code and moving it to the MoveDescription ctor.

Making this change revealed some cases where WeakAI could be generating such invalid moves, which this PR is also fixing.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[X] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[X] Manual testing done

Made AIs play a game on Revised and verified no errors raised.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
+++ b/game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
@@ -26,7 +26,7 @@ public class MoveDescription extends AbstractMoveDescription {
       final Map<Unit, Collection<Unit>> dependentUnits) {
     super(Collections.unmodifiableCollection(units));
     this.route = Preconditions.checkNotNull(route);
-    if (route.getEnd() == null || route.getStart() == null || route.hasNoSteps()) {
+    if (route.hasNoSteps()) {
       throw new IllegalStateException("Invalid route: " + route);
     }
     this.unitsToTransports = Collections.unmodifiableMap(unitsToTransports);
--- a/game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
+++ b/game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
@@ -26,6 +26,9 @@ public class MoveDescription extends AbstractMoveDescription {
       final Map<Unit, Collection<Unit>> dependentUnits) {
     super(Collections.unmodifiableCollection(units));
     this.route = Preconditions.checkNotNull(route);
+    if (route.getEnd() == null || route.getStart() == null || route.hasNoSteps()) {
+      throw new IllegalStateException("Invalid route: " + route);
+    }
     this.unitsToTransports = Collections.unmodifiableMap(unitsToTransports);
     this.dependentUnits =
         Collections.unmodifiableMap(
--- a/game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
+++ b/game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
@@ -26,9 +26,7 @@ public class MoveDescription extends AbstractMoveDescription {
       final Map<Unit, Collection<Unit>> dependentUnits) {
     super(Collections.unmodifiableCollection(units));
     this.route = Preconditions.checkNotNull(route);
-    if (route.hasNoSteps()) {
-      throw new IllegalStateException("Invalid route: " + route);
-    }
+    Preconditions.checkArgument(route.hasSteps());
     this.unitsToTransports = Collections.unmodifiableMap(unitsToTransports);
     this.dependentUnits =
         Collections.unmodifiableMap(
--- a/game-core/src/main/java/games/strategy/triplea/ai/weak/WeakAi.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/weak/WeakAi.java
@@ -223,13 +223,6 @@ public class WeakAi extends AbstractAi {
 
   private static void doMove(final List<MoveDescription> moves, final IMoveDelegate moveDel) {
     for (final MoveDescription move : moves) {
-      // TODO: #5499 Validate this when MoveDescription is constructed.
-      if (move.getRoute().getEnd() == null
-          || move.getRoute().getStart() == null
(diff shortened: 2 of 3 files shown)


asvitkine:
Looks like I hit this case on Revised. Investigating.

Marked as WIP. DO NOT MERGE.

DanVanAtta:
Direction is looking good

asvitkine:
> Looks like I hit this case on Revised. Investigating.
> 
> Marked as WIP. DO NOT MERGE.

OK, should be good now. Fixed issues with WeakAI where moves with invalid routes were being generated.

asvitkine:
> Direction is looking good

Should be ready for merge now.

game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
```diff
@@ -26,6 +26,9 @@ public MoveDescription(
       final Map<Unit, Collection<Unit>> dependentUnits) {
     super(Collections.unmodifiableCollection(units));
     this.route = Preconditions.checkNotNull(route);
+    if (route.getEnd() == null || route.getStart() == null || route.hasNoSteps()) {
```
DanVanAtta: A single territory route is not valid for a move description? Can you clarify if that is fully intended.

``` route.getStart() == null```  is guaranteed by route?

Side-note, kinda nice to have the single constructor, but it looks like the start is asserted as not null:
```
  public Route(final Territory start, final List<Territory> territories) {
    this.start = checkNotNull(start);
    for (final Territory t : territories) {
      add(t);
    }
  }

  public Route(final List<Territory> territories) {
    this(territories.get(0), territories.subList(1, territories.size()));
  }

  public Route(final Territory start, final Territory... territories) {
    this(start, List.of(territories));
  }
```

It also seems like `getEnd()` can not be null either (start is guaranteed to be non-null, if steps is non-empty then we guarantee that we do not add any null elements to it):

```
  public Territory getEnd() {
    if (steps.isEmpty()) {
      return start;
    }
    return steps.get(steps.size() - 1);
  }
```
```
  /** Add the given territory to the end of the route. */
  private void add(final Territory territory) {
    checkNotNull(territory);
    if (territory.equals(start) || steps.contains(territory)) {
      throw new IllegalArgumentException(
          String.format(
              "Loops not allowed in steps, route: %s, new territory: %s", this, territory));
    }
    steps.add(territory);
  }
```

File: game-core/src/main/java/games/strategy/triplea/ai/weak/WeakAi.java [hunk 2]
```diff
@@ -303,7 +298,9 @@ private static void doMove(final List<MoveDescription> moves, final IMoveDelegat
           // move toward the start of the amphib route
           if (firstSeaZoneOnAmphib != null) {
             final Route r = getMaxSeaRoute(data, t, firstSeaZoneOnAmphib, player);
```
DanVanAtta: It may be worth marking `getMaxSeaRoute` as nullable, it would allow for IDE to help us detect such NPE cases.

game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
asvitkine: Thanks for checking, agreed. Removing these.

Keeping route.hasNoSteps() check, since I think that's still valid. My understanding is each move must have steps (i.e. a unit is moving from somewhere to somewhere). Let me know if you think that's not the case.

game-core/src/main/java/games/strategy/triplea/ai/weak/WeakAi.java
asvitkine: Done.

game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
DanVanAtta: > My understanding is each move must have steps (i.e. a unit is moving from somewhere to somewhere). Let me know if you think that's not the case.

Sounds reasonable, but TBH I'm not sure. Would you mind double checking on this?

game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
```diff
@@ -26,6 +26,9 @@ public MoveDescription(
       final Map<Unit, Collection<Unit>> dependentUnits) {
     super(Collections.unmodifiableCollection(units));
     this.route = Preconditions.checkNotNull(route);
+    if (route.hasNoSteps()) {
+      throw new IllegalStateException("Invalid route: " + route);
+    }
```
RoiEXLab: Guavas `Preconditions#checkState` makes this a one-liner

DanVanAtta:
Just waiting for double-check/confirmation that a single territory `MoveDescription` is invalid: https://github.com/triplea-game/triplea/pull/5667/files/950b64d6b04087b4312209d8735228c4c43ff354#diff-2cf07b5adc74d8fe27e4d85bed6ce698

game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
asvitkine: Done

game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
asvitkine: I wasn't able a way to create such a move from the UI. For example, in MovePanel.updateUnitsThatCanMoveOnRoute(), it clears the move if the route has no steps.

In terms of whether a move is "valid", well MoveValidator does consider it as valid per:

```
    final MoveValidationResult result = new MoveValidationResult();
    if (route.hasNoSteps()) {
      return result;
    }
```

But I think nothing would happen in this case when the move is performed. And I also played a full AI game with no such routes being generated. So I think this is safe.

PR #5669, opened 2019-12-06
Launch (spitfire) http-server from Jar on travis

DanVanAtta:
The out-of-band server startup sends output to travis console.
The server does typically start up if needed from integration tests,
but when that happens Travis does not print any output from the server.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Problem fix:  
[x] Other:  Travis build configuration change 

## Testing

[] Manual testing done

### Before
Screenshot from 2019-12-05 17-32-00

### After
Screenshot from 2019-12-05 17-30-45
Screenshot from 2019-12-05 17-30-57

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,8 @@ jobs:
         - java --version
       install: skip
       script:
-        - ./.travis/setup_database &
+        - .travis/setup_database
+        - .travis/launch-http-server
         - ./gradlew --parallel test integTest jacocoTestReport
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -64,7 +64,8 @@ jobs:
         - java --version
       install: skip
       script:
-        - ./.travis/setup_database &
```
DanVanAtta: The removal of the background task `&` is intentional, we really want the database pieces to complete before http-server starts. The server does abort-launch if DB is not available.

PR #5673, opened 2019-12-06
Extract common websocket listener code to base classes

DanVanAtta:
- Adding more message type enums leads to redundant code to 'ChatServerMessageType'. This
code was abstracted to a common interface 'WebsocketMessageType'.
- Updated RemoteActionsMessageType to use the common interface 'WebsocketMessageType'.
- The websocket listener itself became also very boilerplate, common code
was extracted to a common base class 'WebsocketListener'

Benefits:
- if adding a new listener websocket, you can extend WebsocketListener and just need
to define one abstract method where we plugin a call to the message type enum 'valueOf'
- message type enum becomes very boilerplate and is easily tested. Once the listener
and message type are defined and tested, then focus can be spent on the actual listeners
and no more code is needed for the actual websocket communication (though care should be
taken to make sure we close our websocket connection).

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done
- connected to a local lobby and chatted

## Additional Review Notes
- We're essentially unifying all of the websocket listener code here
- Websocket listener code is becoming abstract, we're taking common code from message type enum and putting into a "base interface" and doing the same for the websocket listener class (common code moved to an abstract base class)
- This sets up a standard way for websocket listeners, paves the way for a future update to add a websocket listener for game listing.


http-clients/src/main/java/org/triplea/http/client/web/socket/messages/WebsocketMessageType.java
```diff
@@ -0,0 +1,30 @@
...
+  default <X> void sendPayloadToListener(
+      final ServerMessageEnvelope serverMessageEnvelope, final T listener) {
+    Preconditions.checkArgument(
+        serverMessageEnvelope.getMessageType().equals(toString()),
+        String.format(
+            "Unexpected message type: %s, wanted message type: %s",
+            serverMessageEnvelope.getMessageType(), toString()));
+    final X payload = (X) serverMessageEnvelope.getPayload(getClassType());
+    ((Consumer<X>) getListenerMethod().apply(listener)).accept(payload);
+  }
```
RoiEXLab: Had to think about our discussion about this, and now that this is a generic interface I really feel like the `<X>` generic type should be part of the interface declaration.
Previously this method was "safe", because the only usages were in the same class, where everything obviously had the same type.
Actually the code could have been changed to guarantee compile-time safety by changing the constructor:
```java
enum Test {
 ABC(String.class, o -> (s -> {}));

 <X> Test(Class<X> classType, Function<T, Consumer<X>> consumer) {
    // Set fields here
 }
}
```

Obviously the underlying code still would've been weird to a certain degree, because in this case were obviously limited by the compiler, but there would be no way of messing up the types at all.

Now that this is an interface there aren't any guarantees that the types returned by `getClassType()` does in any way match the function returned by `getListenerMethod()`, which defeats the purpose of generics and type safety IMO.

However I'm not only here to complain, I think I have a solution that
1. Solves the weird casting issue and
2. Still allows us to use enums

The only limitation is that the generic type "X" can't be checked anywhere except in the implementing class.

Here's the plan (I only tested the important things, so I might have missed a minor detail).

This interface get's a new generic parameter X, and both non-default methods, use that type.
Then you create a new wrapper interface, defining this exact method, implementations are supposed to delegate this call to a `TypeWrapper<T, ?>` (or alternatively a `TypeWrapper<T, ?>` getter), that is implemented by all current `WebsocketMessageType` enums to keep the logic roughly the same.
Then you create a single `TypeWrapper<T, X>` that implements this method here and the changed interface, that is currently being used by all enums.
All enums that currently implement this interface are changed to implement the newly created interface, and only accept a `TypeWrapper` object in their constructor (or create this object based on constructor parameters in a type safe manner), which is then being used to implement the logic here.

I assume this was probably really confusing, so have a look at my test implementation that does this without defining a new interface for enums to implement:
```java
import java.util.function.Consumer;
import java.util.function.Function;

class ServerMessageEnvelope {
 <X> X getPayload(Class<X> clazz) {
   return null;
 }
}

interface WebsocketMessageType<T, X> {
  Class<X> getClassType();

  Function<T, Consumer<X>> getListenerMethod();

  default void sendPayloadToListener(
      final ServerMessageEnvelope serverMessageEnvelope, final T listener) {
    final X payload = serverMessageEnvelope.getPayload(getClassType());
    getListenerMethod().apply(listener).accept(payload);
  }
}

class TypeWrapper<T, X> implements WebsocketMessageType<T, X> {

  private final Class<X> clazz;
  private final Function<T, Consumer<X>> function;

  TypeWrapper(Class<X> clazz, Function<T, Consumer<X>> function) {
    this.clazz = clazz;
    this.function = function;
  }



  @Override
  public Class<X> getClassType() {
    return clazz;
  }

  @Override
  public Function<T, Consumer<X>> getListenerMethod() {
    return function;
  }
}

enum Test {
 ABC(new TypeWrapper<>(Integer.class, a -> (s -> {}))),
 DEF(Double.class, a -> (s -> {}));

 private TypeWrapper<String, ?> wrapper;

 Test(TypeWrapper<String, ?> wrapper) {
   this.wrapper = wrapper;
  }

  <X> Test(Class<X> clazz, Function<String, Consumer<X>> function) {
   this.wrapper = new TypeWrapper<>(clazz, function);
  }

  void sendPayloadToListener(
      final ServerMessageEnvelope serverMessageEnvelope, final String listener) {
    wrapper.sendPayloadToListener(serverMessageEnvelope, listener);
  }
}
```

I'll review the rest of this PR later this day, took my quite some time to check my code idea.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebsocketListener.java
```diff
@@ -0,0 +1,100 @@
...
+        listeners);
+  }
+
+  public void setListeners(final ListenersTypeT listeners) {
+    this.listeners = listeners;
+  }
+
+  public void close() {
+    webSocketClient.close();
+  }
```
RoiEXLab: Any thoughts about letting this class implement `Closeable`?
This way static analysis might catch if people forget to close this, and we don't have to rely on people to read the javadoc.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebsocketListener.java
```diff
@@ -0,0 +1,100 @@
...
+   * @param websocketPath Path on the server to the websocket to which we will listen.
+   * @param listeners Listener object that contains listeners for each message type we would expect
+   *     to receive from server.
+   */
+  protected WebsocketListener(
+      final URI hostUri, final String websocketPath, final ListenersTypeT listeners) {
+    this(
+        new GenericWebSocketClient(
+            URI.create(hostUri + websocketPath),
+            "Failed to connect to " + URI.create(hostUri + websocketPath)),
```
RoiEXLab: Hmm is it really necessary to create a URI object for the error message?
If `URI.create("http://example.com").toString()` really returns something different than `"http://example.com" then I would consider putting this error message inside the `GenericWebSocketClient` constructor, or accept a lambda that maps a URI object to a message so that the object doesn't have to be created 2 times.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebsocketListener.java
```diff
@@ -0,0 +1,100 @@
...
+    this.listeners = listeners;
+  }
+
+  public void close() {
+    webSocketClient.close();
+  }
+
+  @Override
+  public void accept(final ServerMessageEnvelope serverMessageEnvelope) {
+    Preconditions.checkState(listeners != null);
```
RoiEXLab: Question:
Listeners can be set in the constructor or by using the setter.
In which case is the setter required? Because if they can be set in the constructor and aren't modified ever, it would be easy to just add some pre-condition in the constructor to avoid this scenario.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebsocketListener.java
```diff
@@ -0,0 +1,100 @@
...
+   */
+  protected abstract MessageTypeT readMessageType(ServerMessageEnvelope serverMessageEnvelope);
+
+  private Optional<MessageTypeT> readMessageTypeValue(
+      final ServerMessageEnvelope serverMessageEnvelope) {
+    try {
+      return Optional.of(readMessageType(serverMessageEnvelope));
+    } catch (final IllegalArgumentException e) {
+      // expect this to happen when we try to use an enum 'valueOf' method for a type
+      // that does not match the enum.
```
RoiEXLab: Hmm might be worth logging this as info message?
Might be useful for future debugging 🤷‍♂

http-clients/src/main/java/org/triplea/http/client/web/socket/messages/WebsocketMessageType.java
RoiEXLab: Also found this https://openjdk.java.net/jeps/301

Exactly our use-case ^^

http-clients/src/main/java/org/triplea/http/client/web/socket/WebsocketListener.java
DanVanAtta: It would be nice, but the interactions are not confined to a single method.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebsocketListener.java
DanVanAtta: If you do not specify the listener in the constructor, then you need to call the setter. This is to support the chat case. I suspect with a moderator lift we could get the chat listeners to be created at the same time as the listeners. I'm not sure how easy that would be as, IIRC, we create the chat websocket essentially on login, we then pass chat to the lobby frame which is the thing that owns and creates the listeners.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebsocketListener.java
DanVanAtta: Tempting, though it gets back to the slight surprises that it only seems there is ever just one websocket connection and all server messages go to all client websockets.

http-clients/src/main/java/org/triplea/http/client/web/socket/messages/WebsocketMessageType.java
DanVanAtta: > Now that this is an interface there aren't any guarantees that the types returned by getClassType() does in any way match the function returned by getListenerMethod(), which defeats the purpose of generics and type safety IMO.

The extraction I believe is equivalent code. It should not have changed any type guarantees. Without a generic typed enum, I suspect we're going to be stuck. We are casting from a JSON String to a type, because the deserialization depends on the serialization, there will be a fundamental 'not type safe', it's unavoidable at some level.

Jeps/301 does look like a match, regrettably it's still just a candidate change though: https://bugs.openjdk.java.net/browse/JDK-8170351


I commend your example @RoiEXLab and trying to fix the type safety issues, though we can only go so far in any one refactoring iteration, I feel like we're getting a bit further from what is 'prudent'. The updates you're describing seem to be more logical/easier once the code has had this first transformation. I think we should take this level of refactor updates to an issue ticket for out-of-band follow-up. The update here is lifting code and not really changing it. We have 2 listeners now, a 3rd is imminent and being worked on, the purpose of this update was to unify the first 2 and pave the way for the third.


### Constructor change

I gave this update a try in the chat message type enum:
```
enum Test {
 ABC(String.class, o -> (s -> {}));

 <X> Test(Class<X> classType, Function<T, Consumer<X>> consumer) {
    // Set fields here
 }
}
```

I was not able to get it to compile. Either the 'T' or the 'Consumer' type conflicted.


### TypeWrapper
The TypeWrapper could have possibility. Though I'm a bit confused on:

```
 private TypeWrapper<String, ?> wrapper;
```

Why is that as String?


```
 ABC(new TypeWrapper<>(Integer.class, a -> (s -> {}))),
 DEF(Double.class, a -> (s -> {}));
```

Why does Double not need a type wrapper but integer does?


### Follow up

It seems we're blocked, the above comments are awkward as I can't comment on a diff. I think it's evidence we're looking at further refactorings. Would you mind explaining excactly what you think should be changed that is within the intent of this update @RoiEXLab? Please also clarify if you think that extracting common code to a base-class actually does change type-safety, I'm curious how that would work as the classes shoudl be equivalent.

Last, I'm concerned about the test/brittleness effort. I spent about 60 minutes last night debugging chat after this refactor went too far and I had to roll back a number of updates to get it working again. As mentioned, because we deserialize from JSON, the code is not going to be fully type-safe. So I don't think we can fully solve it, it's fundamental to the JSON deserialization.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebsocketListener.java
DanVanAtta: I'm starting to forget why the error message is parameterized. I'd like to personally see it be updated so that 'new' of the websocket does not create a connection and that be done explicitly later. It's been a challenge in test to work with that. It's a balance though as we would then need to be sure to call a 'connect' function. 

At the moment I don't know if it's clear that any 'genericWebSocketClient' will always want the same connection error message. Because we have multiple websockets now, knowing the path to the websocket seems handy. You did not that `hostUri` is of type URI? So we're creating the fully qualified path, because we're calling another constructor we are pretty forced to do the create twice. It's a little ugly, IMO ultimately the technical debt for it is extremely localized and not impactful. It seemed like then a decent option compared to going to other lengths.

http-clients/src/main/java/org/triplea/http/client/web/socket/messages/WebsocketMessageType.java
RoiEXLab: I updated my original comment to give some hints, but it's really hard explaining it.
That's why I created a PR onto your branch to make it more clear: https://github.com/DanVanAtta/triplea/pull/7

http-clients/src/main/java/org/triplea/http/client/web/socket/WebsocketListener.java
RoiEXLab: I don't quite follow: All of the initialization is done in the constructor right, so closing can be done all the time, right?
Why doesn't this match the `Closeable` interface then?

http-clients/src/main/java/org/triplea/http/client/web/socket/WebsocketListener.java
RoiEXLab: Well I don't think accepting a `Function<URI, String>` isn't a lot more complicated than accepting a plain string, but on the other hand you could create a private static helper method that creates a `GenericWebSocketClient` for you and re-uses the URI object.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebsocketListener.java
DanVanAtta: >  All of the initialization is done in the constructor right, so closing can be done all the time, right?

I'm starting to be a bit confused now  :confused: 
Perhaps to clarify, the chat socket is closed when a user closes the lobby frame. It cannot be closed all the time, we need the connection to stay open as long as the user has the lobby frame open so they can see chat messages as they come in. The websocket is opened when the lobby frame is created, the close is done in a listener method. I don't see a way to use a 'try-with-resources' block as the open and close are done in very different methods at different times.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebsocketListener.java
DanVanAtta: > Well I don't think accepting a Function<URI, String> isn't a lot more complicated 

That could be done, but conceptually it's somewhat difficult. We pass a URI to a super class, where we also pass a function, where we assume the input to that function will be the same URI that we passed to the super class. I don't think client classes should be quite that aware of the super-class implementation.

I don't know if DRY'ing up the URI creation is the right thing to do, it could be the wrong fix. I'm concerned about code churn in this area. I'd prefer to leave a code duplication code smell in favor of seeing how the usages develop as we add more examples.

http-clients/src/main/java/org/triplea/http/client/web/socket/messages/WebsocketMessageType.java
DanVanAtta: Thanks @RoiEXLab , though perhaps you can appreciate that the update you are suggesting is significant.

The trade-off made here was roughly as follows:
- fixes duplication
- unifies the pattern for listeners
- removes testing burden and simplifies
- allows the next listener to drop in with nicer tests, no throw-away work for unnecessary tests, and consistent to the above pattern
- type non-safety added (IMO it's annoying, but not egregious)

DanVanAtta:
@RoiEXLab I think the bar of 'requested changes' for better type safety is a bit high. I'd like this iteration to go first and then the suggestion you had to land as a second iteration.

RoiEXLab:
@DanVanAtta I wasn't sure if you wrote https://github.com/triplea-game/triplea/pull/5673#discussion_r355093242 before having a look at https://github.com/DanVanAtta/triplea/pull/7
So I wasn't sure if you wanted to merge the PR in this branch.
Is your modification proposal final, or do you have other plans with the PR-PR-branch?

PR #5684, opened 2019-12-06
Remove guarded name constants

DanVanAtta:
'admin' was a special user name for chat connections in lobby. With the lobby re-write,
this should no longer be used. We can then drop all of the cosntants from LobbyConstants.

PlayerName and PlayerNameValidation converge once that check is removed as well, this
update replaces references from PlayerNameValidation to PlayerName.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done
- hosted a game locally
- connected to the game as 'admin', verified chat works.


DanVanAtta:
yeah, agree, the former logic relied on magic chatter names and always created a chat connection for an 'admin' player and the lobby watchers.

PR #5692, opened 2019-12-07
Fix bad lag when trying to move lots of units from a territory.

asvitkine:
On some maps, like Great War, it's often the case that a very large amount of units is accumulated on some borders, such as between Italy and Austria - e.g. 200+ units.

When trying to move these units, before this change, the game UI lags a lot. This change improves that.

When a complete move is invalid, the existing algorithm tries every tail sublist of the sorted unit list. This change improves that by having the sublist logic skip equivalent units (i.e. skip over all infantry that have same movement).

A few other improvements are made - such as the existing code being inconsistent about whether it took a read lock while performing move validation - this duplicated code is moved into a helper function and made to take the lock consistently.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[X] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix:  
[] Other:   

## Testing

[X] Manual testing done

I didn't have a compatible save game ready, so used edit mode to add a bunch of units to a German territory on the western front in Great War. Tested moving these units in the UI and
observed it is much less laggy (though still noticeably slow, but not freezing for many
seconds anymore). Save game attached:

[ww1_test.tsvg.zip](https://github.com/triplea-game/triplea/files/3935367/ww1_test.tsvg.zip)

With the above save game, I did some measurements of how long it takes to validate a move to the following territories:

Before the change:
 Time to: Bastogne = 4189 ms
 Time to: Picardy = 2800 ms
 Time to: Compeigne = 2748 ms
 Time to: Aisne River = 2737 ms

After the change:
 Time to: Bastogne = 183 ms
 Time to: Picardy = 51 ms
 Time to: Compeigne = 56 ms
 Time to: Aisne River = 58 ms

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/UnitComparator.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/UnitComparator.java
@@ -3,6 +3,7 @@ package games.strategy.triplea.delegate;
 import games.strategy.engine.data.PlayerId;
 import games.strategy.engine.data.Route;
 import games.strategy.engine.data.Unit;
+import games.strategy.engine.data.UnitType;
 import games.strategy.triplea.TripleAUnit;
 import games.strategy.triplea.attachments.UnitAttachment;
 import games.strategy.triplea.util.TransportUtils;
@@ -105,6 +106,20 @@ public final class UnitComparator {
       // Sort by increasing movement normally, but by decreasing movement during loading
       if (left1.compareTo(left2) != 0) {
         return (route != null && route.isLoad()) ? left2.compareTo(left1) : left1.compareTo(left2);
+      }
+
+      // Sort units by type first.
+      final UnitType t1 = u1.getType();
+      final UnitType t2 = u2.getType();
+      if (!t1.equals(t2)) {
+        // Land transportable units should have higher priority than non-land transportable ones,
+        // when all else is equal.
+        final int isLandTransportable1 = UnitAttachment.get(t1).getIsLandTransportable() ? 1 : 0;
+        final int isLandTransportable2 = UnitAttachment.get(t2).getIsLandTransportable() ? 1 : 0;
+        if (isLandTransportable1 != isLandTransportable2) {
+          return isLandTransportable2 - isLandTransportable1;
+        }
+        return Integer.compare(t1.hashCode(), t2.hashCode());
       }
 
       return Integer.compare(u1.hashCode(), u2.hashCode());
--- a/game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
+++ b/game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
@@ -7,6 +7,8 @@ import games.strategy.engine.data.PlayerId;
 import games.strategy.engine.data.Route;
 import games.strategy.engine.data.Unit;
 import games.strategy.engine.data.UnitCollection;
+import games.strategy.triplea.TripleAUnit;
+import games.strategy.triplea.attachments.TechAttachment;
 import games.strategy.triplea.delegate.AbstractMoveDelegate;
 import games.strategy.triplea.delegate.AbstractMoveDelegate.MoveType;
 import games.strategy.triplea.delegate.Matches;
@@ -15,12 +17,14 @@ import games.strategy.triplea.delegate.UndoableMove;
 import games.strategy.triplea.delegate.UnitComparator;
 import games.strategy.triplea.delegate.data.MoveValidationResult;
 import games.strategy.triplea.delegate.data.MustMoveWithDetails;
+import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.Predicate;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import org.triplea.java.collections.CollectionUtils;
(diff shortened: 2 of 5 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/UnitComparator.java
```diff
@@ -107,6 +107,11 @@ private UnitComparator() {}
         return (route != null && route.isLoad()) ? left2.compareTo(left1) : left1.compareTo(left2);
       }
 
+      // Sort units by type first.
+      if (!u1.getType().equals(u2.getType())) {
+        return Integer.compare(u1.getType().hashCode(), u2.getType().hashCode());
+      }
+
```
RoiEXLab: Would you mind explaining why we end up comparing hash codes?

I mean I see that this is being used a couple of lines below as well, but I don't understand it there either.

game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
```diff
@@ -1147,41 +1149,33 @@ private void updateUnitsThatCanMoveOnRoute(final Collection<Unit> units, final R
...
-    try {
-      final MoveDescription move =
-          new MoveDescription(bestWithDependents, route, unitsToTransports, dependentUnits);
-      allResults =
-          AbstractMoveDelegate.validateMove(
-              moveType, move, getCurrentPlayer(), nonCombat, getUndoableMoves(), getData());
-    } finally {
-      getData().releaseReadLock();
-    }
+    best = Lists.reverse(best);
```
RoiEXLab: Not really a fan of re-assigning this variable, because it makes things confusing.
I'd appreciate if you could assign this wrapper to a new independent `bestReversed` variable. (If you want you can even use `Collection#removeIf` with an inverted `Predicate` instead if the repeated `CollectionUtils#getMatches` which avoids having to recreate a list every time and make the variable final)

game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
```diff
@@ -1147,41 +1149,33 @@ private void updateUnitsThatCanMoveOnRoute(final Collection<Unit> units, final R
...
-        lastResults =
-            AbstractMoveDelegate.validateMove(
-                moveType, move, getCurrentPlayer(), nonCombat, getUndoableMoves(), getData());
+        // Try to find a valid sublist, skipping equivalent units.
+        final Unit firstSkipUnit = best.get(0);
+        int startIndex = 1;
+        while (startIndex < best.size()
+            && unitsAreEquivalent(firstSkipUnit, best.get(startIndex))) {
+          startIndex++;
+        }
```
RoiEXLab: I have a hard time understanding why your new code is supposed to be equivalent.

Why can "equivalent" units be skipped, and what makes them "equivalent" in this scenario?

game-core/src/main/java/games/strategy/triplea/delegate/UnitComparator.java
asvitkine: The idea is that if you have 100 infantry and 100 tanks, then if the 100 infantry are all equivalent (same unit type, same movement left), we can skip all 100 instead of skipping one at a time since if one can't move, the others can't either.

RoiEXLab: Ah ok, makes sense. I'd still like @ron-murhammer to have a second look at this 👍

game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
```diff
@@ -1216,6 +1210,30 @@ private void updateUnitsThatCanMoveOnRoute(final Collection<Unit> units, final R
     unitsThatCanMoveOnRoute = new ArrayList<>(bestWithDependents);
   }
 
+  // Whether the two units are equivalent for the purposes of movement.
+  private boolean unitsAreEquivalent(final Unit u1, final Unit u2) {
+    final BigDecimal left1 = TripleAUnit.get(u1).getMovementLeft();
+    final BigDecimal left2 = TripleAUnit.get(u2).getMovementLeft();
+    return u1.isEquivalent(u2) && left1.equals(left2);
```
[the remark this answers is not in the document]
asvitkine: Thanks for suggesting testing land transports.

Indeed, it doesn't handle that case well. But actually even without this change, it seems it's buggy - that's because the sort order of units is buggy - so if you select 3 infantry, 1 artillery and 2 tanks (with mech infantry tech), it's not guaranteed the game figures out that sending those units 2 territories away allows the 2 infantry and 2 tanks to go - because of the arbitrary sort order of infantry and artillery! This is with the current code without this change.

But with this change, it doesn't work at all - thanks for noticing!

Let me see how I can support that, I suspect we just need logic to determine how many units could be assisted and make sure they're sorted last.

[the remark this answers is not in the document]
asvitkine: I have an initial version that seems to solve it, though I'd like to do more testing, including another unit test.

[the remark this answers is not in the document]
ron-murhammer: Ok. Once you think this is ready let me know. If you can add some unit tests and a few screenshots to this PR that would be very helpful. I'd like to test this a bit myself manually with a few different maps as well.

asvitkine:
Still working on it.

While writing a unit test, I discovered that the behavior of getMovableUnitsComparator() is very strange. The sort order is different depending on if some units don't have enough moves to complete the route or not. This seems very unintuitive - because ideally it would sort them identical - by movement (and other considerations like dependent units), regardless of the route.

I'm trying to figure out if this behavior is really required - because it is very confusing.

RoiEXLab:
@asvitkine Just wanted to say that I appreciate your effort that not only tries to refactor code, but also questions the legitimacy of existing code 👍

asvitkine:
FYI: My plan with this PR is to rework it on top of my refactoring PR #5708, along with tests for the new behavior changes.

asvitkine:
I've rebased this now on my other refactor PR and added a unit test for mech infantry.

asvitkine:
Now rebased on master. @RoiEXLab can you take another look and merge if it looks good?

PR #5694, opened 2019-12-08
Small optimization: use Integer.parseInt() directly.

asvitkine:
Small optimization: use Integer.parseInt() directly when checking for exception, instead of getInt() which rethrows a different exception. This was showing up in profiles when profiling unit movement UI - as this is called often from territoryIsPassableAndNotRestricted() -> getListedTerritories().

I was seeing the IllegalStateException ctor that the previous code called through taking up ~700ms on a given profile - which went away with this change.

Screen Shot 2019-12-08 at 4 10 24 PM

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[X] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/attachments/AbstractRulesAttachment.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/AbstractRulesAttachment.java
@@ -376,13 +376,13 @@ public abstract class AbstractRulesAttachment extends AbstractConditionsAttachme
         // See if the first entry contains the number of territories needed to meet the criteria
         try {
           // check if this is an integer, and if so set territory count
-          getInt(name);
+          Integer.parseInt(name);
           if (mustSetTerritoryCount) {
             haveSetCount = true;
             setTerritoryCount(name);
           }
           continue;
-        } catch (final Exception e) {
+        } catch (final NumberFormatException e) {
           // territory name is not an integer; fall through
         }
       }


game-core/src/main/java/games/strategy/triplea/attachments/AbstractRulesAttachment.java
```diff
@@ -376,13 +376,13 @@ protected void validateNames(final String[] terrList) {
...
           // check if this is an integer, and if so set territory count
-          getInt(name);
+          Integer.parseInt(name);
           if (mustSetTerritoryCount) {
             haveSetCount = true;
             setTerritoryCount(name);
           }
           continue;
-        } catch (final Exception e) {
+        } catch (final NumberFormatException e) {
```
RoiEXLab: I just had a look at `setTerritoryCount`.
Looks like the value can only ever be `"each"` or a number encoded as string, otherwise it throws an exception.

So I'd actually recommend the following:
replace `setTerritoryCount(name)` with `territoryCount = Integer.parseInt(name)`, but with intermediate variable so the exception is also thrown when `mustSetTerritoryCount` is false and then drop `Integer.parseInt(name)` in line 379.

If you want you can also overload `setTerritoryCount` to accept an int and update all the callers that convert an int to a string before invoking this method

RoiEXLab:
Actually, nevermind I'll create a follow-up PR

PR #5698, opened 2019-12-10
Fix None label displayed in casualties panel after casualties added.

asvitkine:
See before/after screenshots. This bug exists in 1.9 too.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix:  
[] Other:   

## Testing

[X] Manual testing done

Do a battle and observe what the casualties panels show after first round.

### Before

Screen Shot 2019-12-10 at 1 49 32 AM

### After

Screen Shot 2019-12-10 at 1 55 18 AM

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
@@ -204,10 +204,14 @@ public class BattleDisplay extends JPanel {
     final JPanel casualtyPanel;
     if (playerId.equals(defender)) {
       casualtyPanel = casualtiesInstantPanelDefender;
-      casualtyPanel.remove(labelNoneDefender);
+      if (!killedUnits.isEmpty()) {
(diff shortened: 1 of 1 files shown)


asvitkine:
> I'm not 100% sure how this panel works, but isn't there a case where no units are killed, so this panel needs to be re-added?

I don't think so. I don't see any code that would be doing this.

RoiEXLab:
> I don't think so. I don't see any code that would be doing this.

Yeah, I don't do either, that's what is bothering me. I would've expected that it should exist, so that in a round without any hits "None" is being displayed again.

ron-murhammer:
Hmm. I think the problem is that the casualties panel is kind of an in between of total casualties and casualties by round. At the moment, I believe this is a good fix to at least remove the "none" once there are casualties. The way it works I think is it only adds something per combat round if there was a casualty that round so you don't get "none" for every round that didn't have any casualties only get it before there are any casualties.

PR #5699, opened 2019-12-10
Enhance type safety

RoiEXLab:
Basically https://github.com/DanVanAtta/triplea/pull/7 rebased with suggestions.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/messages/MessageTypeListenerBinding.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/messages/MessageTypeListenerBinding.java
@@ -10,7 +10,7 @@ import lombok.Getter;
 @AllArgsConstructor
 @Getter
 @Immutable
-public class WebsocketMessageWrapper<T, X> implements WebsocketMessageType<T> {
+public class MessageTypeListenerBinding<T, X> implements WebsocketMessageType<T> {
 
   private final Class<X> classType;
   private final Function<T, Consumer<X>> listenerMethod;
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/messages/MessageTypeListenerBinding.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/messages/MessageTypeListenerBinding.java
@@ -7,6 +7,11 @@ import lombok.Value;
 
 @Value(staticConstructor = "of")
 @Immutable
+// The function passed as second param isn't
+// guaranteed to be immutable, but we are
+// going to trust the future-devs to use
+// stateless lambdas.
+@SuppressWarnings("Immutable")
 public class MessageTypeListenerBinding<T, X> {
 
   private final Class<X> classType;
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/messages/MessageTypeListenerBinding.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/messages/MessageTypeListenerBinding.java
@@ -1,6 +1,5 @@
 package org.triplea.http.client.web.socket.messages;
 
-import com.google.common.base.Preconditions;
 import com.google.errorprone.annotations.Immutable;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -12,15 +11,9 @@ public class MessageTypeListenerBinding<T, X> {
 
   private final Class<X> classType;
   private final Function<T, Consumer<X>> listenerMethod;
-  private final String name;
 
   public void sendPayloadToListener(
       final ServerMessageEnvelope serverMessageEnvelope, final T listener) {
-    Preconditions.checkArgument(
-        serverMessageEnvelope.getMessageType().equals(name),
-        String.format(
-            "Unexpected message type: %s, wanted message type: %s",
-            serverMessageEnvelope.getMessageType(), name));
     final X payload = serverMessageEnvelope.getPayload(getClassType());
     getListenerMethod().apply(listener).accept(payload);
   }
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/messages/MessageTypeListenerBinding.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/messages/MessageTypeListenerBinding.java
@@ -4,19 +4,16 @@ import com.google.common.base.Preconditions;
 import com.google.errorprone.annotations.Immutable;
 import java.util.function.Consumer;
 import java.util.function.Function;
-import lombok.AllArgsConstructor;
-import lombok.Getter;
+import lombok.Value;
 
-@AllArgsConstructor
-@Getter
+@Value(staticConstructor = "of")
 @Immutable
-public class MessageTypeListenerBinding<T, X> implements WebsocketMessageType<T> {
+public class MessageTypeListenerBinding<T, X> {
 
   private final Class<X> classType;
   private final Function<T, Consumer<X>> listenerMethod;
   private final String name;
 
-  @Override
   public void sendPayloadToListener(
       final ServerMessageEnvelope serverMessageEnvelope, final T listener) {
     Preconditions.checkArgument(
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/messages/WebsocketMessageType.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/messages/WebsocketMessageType.java
@@ -1,5 +1,7 @@
 package org.triplea.http.client.web.socket.messages;
 
+import com.google.common.base.Preconditions;
+
 /**
  * Interface representing a websocket message type that knows about the correct type
  * to extract from a payload object from a {@code ServerMessageEnvelope}.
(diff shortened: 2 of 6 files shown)


DanVanAtta:
@RoiEXLab thanks for working on this. I've been experimenting with your PR/PR and was looking for ways where we could remove the message 'factory' classes. For example, the below is very boilerplate:

```
  public ServerMessageEnvelope newBannedPlayer(final InetAddress bannedIp) {
    return ServerMessageEnvelope.packageMessage(
        ServerRemoteActionMessageType.PLAYER_BANNED.toString(), bannedIp);
  }
```

I was looking for a way where we could restate it as:
```
  public ServerMessageEnvelope newBannedPlayer(final InetAddress bannedIp) {
        return ServerRemoteActionMessageType.PLAYER_BANNED
                                     .packageMessage(bannedIp);
  }
```

@RoiEXLab do you see a way we could achieve that? I'm not looking for that to be done in this PR, but would not want us to move away from that direction. I was hoping the message binder object could support that, but was even considering having a second "helper" object, similar to the message binder to deliver that.

Going even further, it would be nice if we could even unify the server and client message envelopes. The client envelopes, instead of having a payload, apikey, and type string, they would instead have a type string and contain a "clientMessage" that contains the api key and payload string, ie:
```
class ClientMessage {
   ApiKey apiKey;
   String messagePayload;
}
```

I'm ultimately wondering if you're able to think of any kind of proof of concept/demo code where we could remove the factory messages. If so, it seems like this would a first step, removing factories would be the next, then if feasible unifying the client and server message envelopes.

RoiEXLab:
@DanVanAtta I think I can come up with something, I'll try to create a PR tomorrow.
Would be nice if we could get this merged in the meantime

DanVanAtta:
I'm only hesitant from my experiments where the generics conflicted and it
did not seem workable.

RoiEXLab:
> I'm only hesitant from my experiments where the generics conflicted and it
did not seem workable.

Would you mind going into details?

DanVanAtta:
Can't be too concrete as it was all experimental. In short, the following was the problem:
```
- private final Function<ChatMessageListeners, Consumer<?>> listenerMethod;
+  private final MessageTypeListenerBinding<ChatMessageListeners, ?> messageTypeListenerBinding;
```
In specific that the wildcard type matched the consumer and not the consumed type. I was not able to re-arrange it so that there was type safety on the send method. I think our focus should be there, as it's more important to make sure that the send and listener methods line up type-wise. IMO it is more important as the error will be more subtle, the two messages are packaged and opened in different locations in the code, it's more likely they could get out of sink, and in particularly if client and server are updated independent of one another (which perhaps is another problem)

RoiEXLab:
@DanVanAtta
I think I get now what you're trying to achieve.
You want to be able to do something like this, right?
```java
EnumType.STRING_MESSAGE.send("abc");
EnumType.INT_MESSAGE.send(1);
```
So that the send method only accepts the correct type right? Unfortunately this won't be possible using enums, we'd have to make them proper objects to achieve this. Currently all enum values within an enum have to be the same when it comes to method signature.
You could however create a dedicated enum for each class type to solve that problem, but it makes other things more complicated, but when using interfaces instead of concrete enum types this could be feasible.

From a compiler standpoint, it's not necessary to have this generic type-safety though, we could make the message binding accept a plain object, and check the type as a precondition using the classType, which is then passed to the `ServerMessageEnvelope` where it is turned into a string.
This envelope can then be passed to the send method as usual.

RoiEXLab:
Regardless I'm not sure how this is directly related this PR.
Is there anything I could improve code-wise?
Because I think it makes much more sense to merge this PR now and then make the code fit afterwards if it needs to match a new requirement

DanVanAtta:
Indeed for the send. I realize about the enum issue, so was thinking we
could get the bind object from the enum to use for packaging.

I don't know if I'd really call this a new requirement. I want the send and
unwrap to be type safe, type safety between unwrap and listener is
significant but less error prone since that happens in the same class and
is readily tested.

Hence, I'd like to know what a type safe serialization and deserialization
would look like.

DanVanAtta:
In other words, we might be focusing on the wrong requirement here. Second,
we need to make sure server/client differences will be obvious and easily
handled.

DanVanAtta:
The wildcard for consumer type perhaps makes more sense as a problem, if we
get the bind object from enum to do send, then the type for the consumed
object seems to not line up

RoiEXLab:
Perhaps I should ask a different question.
What do I have to do to get this PR approved?
I agree that this code doesn't solve all the problems, but I think it's a small improvement over the current code, so definitely worth merging.

DanVanAtta:
I don't know if this the right direction if we have to undo it to get type
safety on message packaging. I'd be looking for a path forward to that more
important objective and if this complements that

RoiEXLab:
@DanVanAtta If you ask me the only "real" way to get type safety on serialization is if we completely drop the current enum scheme in favour of either multiple enums grouped by serialization type, where we need to partially implement an `Enum#valueOf` method that delegates to the proper enum type,

or by using static final instances of objects with different generic parameters, where we'd have to put all of the objects into a map and manually implement the valueOf functionality.

So in those 2 cases this change is redundant, but at the same time wouldn't hurt either because the class could still be re-used as-is (by delegating everything to it), but it wouldn't be necessary to have that extra layer of abstraction.

In case we decide to keep the enum structure as-is I don't think we can have compile-time type safety (runtime-checks however are easy to implement using the current structure) for serialization.
In this scenario this PR is IMO definitely a step in the right direction, because it allows us to keep common logic regarding serialization in a single place, and doesn't rely on default methods not being overriden by an enum.

Even if you're not convinced that this code will match future requirements, it's literally not even a 100 line change, so undoing it can be done in a matter of minutes, nothing too crazy.

That's why I'd vote to merge this PR, it's such a small change that implements compile-time safety for 2 enums, nothing to have a week long dicussion about

DanVanAtta:
Biggest goals are reduction of boilerplate and getting type safety in
packaging. Given we will need to test the enum to listener bindings
anyways, I'm not super worried about type safety here as we will detect it.

If we can wire an object to a listener, it seems we should be able to for a
second that packages. The wildcard in function type was an impediment when
giving that a try.

Agree not worth lots of convos, I'm concerned by the churn and I don't know
when I'll be able to review this. Focus please on a prototype for message
send type safety.

It sounds like the map might be the right way to go.

http-clients/src/main/java/org/triplea/http/client/web/socket/messages/MessageTypeListenerBinding.java
```diff
@@ -0,0 +1,25 @@
+package org.triplea.http.client.web.socket.messages;
+
+import com.google.errorprone.annotations.Immutable;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import lombok.Value;
+
+@Value(staticConstructor = "of")
+@Immutable
+// The function passed as second param isn't
```
DanVanAtta: IMO, perhaps not worth the comment and the warning suppression to mark immutable. Could be just as well to leave it as effectively immutable.

http-clients/src/main/java/org/triplea/http/client/web/socket/messages/MessageTypeListenerBinding.java
```diff
@@ -0,0 +1,25 @@
...
+// guaranteed to be immutable, but we are
+// going to trust the future-devs to use
+// stateless lambdas.
+@SuppressWarnings("Immutable")
+public class MessageTypeListenerBinding<T, X> {
```
DanVanAtta: Nit: I'm not a fan that template types need to end in 'T', in the case of more than one though, it's a lot more descriptive to spell out the types and what they are meant to represent. EG:
```
<ListenersT, PayloadT>
```
Is a lot more descriptive compared to:
```
<T, X>
```

http-clients/src/main/java/org/triplea/http/client/web/socket/messages/WebsocketMessageType.java
```diff
@@ -1,30 +1,29 @@
...
-import java.util.function.Consumer;
-import java.util.function.Function;
 
 /**
- * Interface representing a websocket message type that knows how to extract a payload object from a
- * {@code ServerMessageEnvelope} and also knows how to send that payload to a parameter listener.
+ * Interface representing a websocket message type that knows about the correct type to extract from
+ * a payload object from a {@code ServerMessageEnvelope}.
+ *
+ * <p>Delegates {@link #sendPayloadToListener(ServerMessageEnvelope, Object)} to {@link
```
DanVanAtta: nit, implementation detail in this comment. More important to javadoc the purpose and intent of an interface. I think the commentary here is describing  a bit too much about how the class does what it does as opposed to what the class is supposed to do. Would probably also be good to have a javadoc describe expected contexts and usages where we would use the interface, what contracts we would be guaranteeing when implementing it.

DanVanAtta:
>  Focus please on a prototype for message
send type safety.

Belay that, focus should be on the release blocker 2.0 bug(s) and QA testing 2.0. The pattern we have is good enough for moving game listing to a websocket. Please mention me somewhere if you want to start tackling that.

DanVanAtta:
Side-note, i re-wrote the commit comment some, we probably should establish a standard, the commentary was useful for review, but IMO commentary for review is the least valuable and least permanent there is. We should be focusing the value retention for those that are investigating bugs, exploring code, and looking for history to determine what changed where and why.

PR #5701, opened 2019-12-11
Improve restart of nginx to use systemctl module

DanVanAtta:
Ansible best practice is to avoid using shell commands and
to prefer the built in modules. This update uses the built-in
systemctl module to execute the nginx restart

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

## Additional Review Notes
Note, after this update, we still will fail with nginx restart. There is a chicken-and-egg problem where the installed config requires 'letsencrypt' to have been run. We'll likely need to install a dummy or hardcoded SSL cert on the letsencrypt path so that the path is valid. On new servers, the ordering is tenous, we may need to run letsencrypt after ansible fails to get things to be correct, then run ansible. Ideally we'd bake in the 'letsencrypt' module and not need to manage this ourselves by hand.

### DIFF
--- a/infrastructure/ansible/roles/nginx/tasks/main.yml
+++ b/infrastructure/ansible/roles/nginx/tasks/main.yml
@@ -20,5 +20,7 @@
 
 - name:  relaunch nginx
   become: true
-  shell: systemctl restart nginx
+  systemd:
+    name: nginx
+    state: restarted
 


DanVanAtta:
True, though #5555 needs to be verified. If it's not merge-ready, it does not have a "lock" on the code it touches, particularly so to fixes that are beyond the single-focus of the update. This line needs to be fixed more as we do not want to restart nginx on every run of ansible, disconnecting all users when nothing changed is something to be fixed. On prerelease, we change the installer file each time, so we can get away with it, when we run ansible against prod on every commit, this must be fixed. (So there are more changes that are going to happen here, on these lines, this update is a hopefully reasonable first iteration to that series).

PR #5702, opened 2019-12-11
Fix flyway CLI install task

DanVanAtta:
- Do not redownload flyway zip file, if zip file already exists
we can skip DL (otherwise fails), and similarly if zip file is
already extracted we can skip the extraction

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/infrastructure/ansible/roles/flyway/tasks/main.yml
+++ b/infrastructure/ansible/roles/flyway/tasks/main.yml
@@ -36,13 +36,25 @@
   shell: echo "alter role {{ flyway_db_user }} with password '{{ flyway_db_password }}';"  | sudo -u postgres psql
   become_user: postgres
 
+- name: check for flyway download 
+  stat:
+    path: /tmp/flyway-cli.tar.gz
+  register: flyway_download
+
 - name: Download flyway CLI
+  when: flyway_download.stat.exists == False
   get_url:
     url: "{{ flyway_download_location }}"
     dest: /tmp/flyway-cli.tar.gz
     mode: 777
 
+- name: check for extracted flyway
+  stat:
+    path: /tmp/flyway-{{ version }}
+  register: flyway_extraction
+
 - name: extract flyway
+  when: flyway_extraction.stat.exists == False
   unarchive:
     copy: no
     src: /tmp/flyway-cli.tar.gz
--- a/infrastructure/ansible/roles/flyway/tasks/main.yml
+++ b/infrastructure/ansible/roles/flyway/tasks/main.yml
@@ -1,3 +1,10 @@
+- name: create service user to run flyway
+  become: true
+  user:
+    name: "{{ flyway_user }}"
+    create_home: yes
+    system: yes
+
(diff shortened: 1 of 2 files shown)


infrastructure/ansible/roles/flyway/tasks/main.yml
```diff
@@ -36,13 +36,25 @@
...
     dest: /tmp/flyway-cli.tar.gz
     mode: 777
 
+- name: check for extracted flyway
+  stat:
+    path: /tmp/flyway-{{ version }}
+  register: flyway_extraction
+
 - name: extract flyway
+  when: flyway_extraction.stat.exists == False
```
RoiEXLab: According to the docs of the unarchive module, you could use the `creates` property instead of using a stat.

infrastructure/ansible/roles/flyway/tasks/main.yml
```diff
@@ -36,13 +36,25 @@
   shell: echo "alter role {{ flyway_db_user }} with password '{{ flyway_db_password }}';"  | sudo -u postgres psql
   become_user: postgres
 
+- name: check for flyway download 
+  stat:
+    path: /tmp/flyway-cli.tar.gz
+  register: flyway_download
+
 - name: Download flyway CLI
+  when: flyway_download.stat.exists == False
```
RoiEXLab: Also the default behaviour of the `get_url` module is not to overwrite any existing files according to the docs:
See parameter `force`:
https://docs.ansible.com/ansible/latest/modules/get_url_module.html

infrastructure/ansible/roles/flyway/tasks/main.yml
DanVanAtta: "creates" is a nice new feature, happy to use it here.

infrastructure/ansible/roles/flyway/tasks/main.yml
DanVanAtta: This update is required, or you get this error:
 "msg": "Destination /tmp/flyway-cli.tar.gz is not writable",

`force` is not ideal as we get a DL and overwrite every time. It turns out if we set a checksum on the get_url task, then it knows enough when to skip download task. I think it's designed this way to have you clean up downloaded files, and/or overwrite them, or set a checksum for them. In any of these cases, if the download fails or is corrupted, then the playbook can be repeated and it'll repair any issues. If you skip download with a corrupt file, then the re-run would not fix the issues.

DanVanAtta:
Iterated one more time and fixed what I think are the most salient and remaining problems in the flyway tasks. Previously if we run the role twice it would have error'ed out on the download file. After fixing that, we still had 'changed' reports. I fixed those up, after running the task a second time we get a desirable report of no changes:

Screenshot from 2019-12-11 21-26-03

PR #5705, opened 2019-12-12
Update run ansible vagrant script to allow tag args

DanVanAtta:
Pass any script parameters to ansible command. This allows us to
invoke ansible with tags, eg: `run_ansible_vagrant -t flyway`

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/infrastructure/run_ansible_vagrant
+++ b/infrastructure/run_ansible_vagrant
@@ -69,5 +69,6 @@ function buildMigrations() {
 buildMigrations
 
 ansible-playbook -D -v \
-     ansible/site.yml \
+    "$@" \
+    ansible/site.yml \
     -i ansible/inventory/vagrant 


infrastructure/run_ansible_vagrant
```diff
@@ -69,5 +69,6 @@ function buildMigrations() {
 buildMigrations
 
 ansible-playbook -D -v \
-     ansible/site.yml \
+    "$@" \
```
RoiEXLab: Does this work when it's inside quotes?

DanVanAtta: I would expect for it to be break, if it were to break, on any token containing a space. The `-t <tag>` param worked, I expect it should work always. I have not tested with many args though. The other element, this was modeled after  what we have in run_deployment, I don't remember exactly if multiple args were tested there as well (I think so, but my recollection is fuzzy).

```
ansible-playbook \
  --vault-password-file "$VAULT_PASSWORD_FILE" \
 "$@" \
 ansible/site.yml
```

PR #5707, opened 2019-12-13
Fix incorrect units loaded when loading a transport.

### LINKED ISSUE #5706: Loading transports with multiple units regressed since 1.9
**Describe the bug**
While testing my movement refactoring changes, I noticed that loading transports is not working great when multiple units are selected and seems to be worse than earlier builds. In particular, moving a selection of 3 infantry and 1 tank onto a transport results in 1 tank being loaded only, whereas previously, 1 infantry and 1 tank would be loaded.

**Engine Version or SHA**
1667ee6365fbdea45abbd34180e2d4f08ccb19eb

**To Reproduce**
1. On Global 40 2nd edition...
2. As germany, select 3 infantry and 1 tank in poland
3. Try to load them on the transport

**Expected behavior**
1 infantry and 1 tank should be loaded.

**OS (please put an X in the appropriate box):**
[] Windows
[X] Mac
[] Linux

**Additional context**
This is not related to my recent MovePanel changes - the bug exists without them. But it could be due to my refactoring of unit to transport mapping. I will try to bisect.

asvitkine:
This regressed in #5416, because that commit caused unitsToTransports to only be computed once, instead of updated each time the unit list was filtered in the function.

This fixes #5706.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix: #5706
[] Other:   

## Testing

[X] Manual testing done

On Global 40 2nd edition...
As germany, select 3 infantry and 1 tank in poland
Try to load them on the transport

1 infantry and 1 tank should be loaded.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
@@ -7,6 +7,7 @@ import games.strategy.engine.data.PlayerId;
 import games.strategy.engine.data.Route;
 import games.strategy.engine.data.Territory;
 import games.strategy.engine.data.Unit;
+import games.strategy.engine.data.UnitCollection;
 import games.strategy.engine.data.UnitType;
 import games.strategy.engine.framework.LocalPlayers;
 import games.strategy.triplea.Properties;
@@ -1122,19 +1123,9 @@ public class MovePanel extends AbstractMovePanel implements KeyBindingSupplier {
       return;
     }
     getMap().hideMouseCursor();
-    // TODO kev check for already loaded airTransports
-    Map<Unit, Unit> unitsToTransports = Map.of();
-    if (MoveValidator.isLoad(units, dependentUnits, route, getData(), getCurrentPlayer())) {
-      final Collection<Unit> transportsToLoad =
-          route
-              .getEnd()
-              .getUnitCollection()
-              .getMatches(
-                  Matches.unitIsTransport().and(Matches.alliedUnit(getCurrentPlayer(), getData())));
-      unitsToTransports = TransportUtils.mapTransports(route, units, transportsToLoad);
-    }
+    final Collection<Unit> transportsToLoad = getPossibleTransportsToLoad(units, route);
     List<Unit> best = new ArrayList<>(units);
-    // if the player selects a land unit and other units when the
+    // if the player selects a land unit and other units then
     // only consider the non land units
     if (route.getStart().isWater() && route.getEnd().isWater() && !route.isLoad()) {
       best = CollectionUtils.getMatches(best, Matches.unitIsLand().negate());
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
```diff
@@ -1191,6 +1182,17 @@ private void updateUnitsThatCanMoveOnRoute(final Collection<Unit> units, final R
     unitsThatCanMoveOnRoute = new ArrayList<>(lastResult.getUnitsWithDependents());
   }
 
+  private Collection<Unit> getPossibleTransportsToLoad(
+      final Collection<Unit> units, final Route route) {
+    // TODO kev check for already loaded airTransports
```
DanVanAtta: I'm with code-climate for fixing or removing this TODO, does the todo comment make any sense to anyone? I'm not sure what we should be checking and if it's not already taken care of somehow. Or is this method just incomplete and/or wrong from the get-go by not considering all transport types. Did you gain any insights @asvitkine  into this when working on this code?

game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
```diff
@@ -1191,6 +1182,17 @@ private void updateUnitsThatCanMoveOnRoute(final Collection<Unit> units, final R
...
+    if (MoveValidator.isLoad(units, dependentUnits, route, getData(), getCurrentPlayer())) {
```
DanVanAtta: nit, could go for a ternary expression and make this a bit more concise. It adds more to each line but the single return statement at front should make it conceptually easier to understand (the reader knows right away we are returning). 
eg:
```
return MoveValidator.isLoad(units, dependentUnits, route, getData(), getCurrentPlayer())
    ? route.getEnd().getUnitCollection().getMatches(
          Matches.unitIsTransport().and(Matches.alliedUnit(getCurrentPlayer(), getData()))
    : List.of();
```

game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
asvitkine: I've not had a chance to test this yet, but note that a corresponding TODO exists in MoveValidator.isLoad().

I think it's better to keep the TODO until we have a chance to validate this. This can be done by expanding test coverage for MovableUnitsFilter which I'm adding in the next PR - though my priority there is to add test cases for the other things first - like transports loading and mech infantry.

game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
asvitkine: I find your example harder to read. Note that I did refactor this to a helper to allow the assignment in the caller to be done to a final variable. I find overall the helper + two returns is more readable than the other alternatives like your suggestion. But I'm curious what others think.

DanVanAtta: IMO depends on expectations and what a person is used to. The multiple return points has higher cognitive complexity, the two are otherwise equivalent. It's not easy to read either way.

PR #5708, opened 2019-12-13
Refactor movable units filtering to a helper class.

asvitkine:
This decouples the UI code in MovePanel from the logic of filtering which units could move on a given route.

No functionality changes. This refactoring will allow us to more easily unit test the functionality as we make changes to it, which I plan to do more of in a follow-up PR. This PR contains just a basic unit test.

Note: This PR is branched from my other two PRs that are pending review. I'll rebase it on head once they land.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[X] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[X] Manual testing done

Basic testing of movement on a couple of maps.


game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
```diff
@@ -0,0 +1,205 @@
...
+import javax.annotation.Nullable;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import org.triplea.java.collections.CollectionUtils;
+
+/**
+ * Utility class for filtering a unit list to the subset of those units that can move on a route.
+ */
+public final class MovableUnitsFilter {
+  /** The result of the filter operation. */
```
DanVanAtta: nit, what do you think about renaming this class to `FilterOperationResult`? That way we can omit the clarifying comment.

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
```diff
@@ -0,0 +1,205 @@
...
+  @Getter
+  public static class Result {
+    // The filtered units and dependents that can move on the route.
+    private final List<Unit> unitsWithDependents;
+    // An error message, if no units could move on the route.
+    @Nullable private String errorMessage;
```
DanVanAtta: Would this potentially be more clear named "noUnitsCanMoveMessage"?
Similar for the warning, "someUnitsCanMoveMessage"

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
```diff
@@ -0,0 +1,205 @@
...
+    // A warning message, if some, but not all units could move on the route.
+    @Nullable private String warningMessage;
+
+    public Result(
+        final MoveValidationResult allUnitsResult,
+        final MoveValidationResultWithDependents lastResult) {
+      this.unitsWithDependents = lastResult.getUnitsWithDependents();
+      if (!allUnitsResult.isMoveValid()) {
+        final @Nullable final String message = getErrorOrWarningMessage(allUnitsResult);
```
DanVanAtta: Interesting that: `final @Nullable final` compiles :grin: 

AFAIK nullable is not used on local variables as the scope is pretty well managed. Recommend to use Optional to call this out.

```
getErrorOrWarningMessage(allUnitsResult)
   .ifPresent(msg -> 
              if( !lastResult.getResult().isMoveValid() ) {
                     this.errorMessage = msg;
              } else {
                    this.warningMessage = msg;
              }
        );
```
Implicit to the above, the return type of `getErrorOrWarningMessage` is `Optional<String>`. Given the above code, I wonder what you would think about splitting the error and warning conditions? I think that may help with a potential SRP violation code-smell. The setters would then be:


```
this.errorMessage = getErrorMessage(allUnitsResult).orElse(null);
this.warningMessage = getWarningMessage(allUnitsResult).orElse(null);
```

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  /**
+   * Filters the units and their dependents to a subset of units that can move on the route.
+   *
+   * @param units The units to filter.
+   * @param dependentUnits The dependent units map.
+   * @return The result.
+   */
+  public Result filterUnitsThatCanMove(
```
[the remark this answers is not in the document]
DanVanAtta: IIRC the agreement for code cov was to comment on warnings on why they were not appropriate to be fixed. Given we've extracted code to a new class, I would think it plausible to avoid code climate warnings. Is there strong reason to allow this warning? Ditto for the other code-climate warnings.

game-core/src/test/java/games/strategy/triplea/util/MovableUnitsFilterTest.java
```diff
@@ -0,0 +1,115 @@
...
+public class MovableUnitsFilterTest {
+  final GameData gameData = TestMapGameData.REVISED.getGameData();
+  final PlayerId germans = germans(gameData);
+  final Territory germany = territory("Germany", gameData);
+  final Territory easternEurope = territory("Eastern Europe", gameData);
+  final Territory kareliaSsr = territory("Karelia S.S.R.", gameData);
+  final UnitType infantryType = infantry(gameData);
+  final UnitType armourType = armour(gameData);
+
+  public MovableUnitsFilterTest() throws Exception {}
```
DanVanAtta: nit, empty constructor not needed in junit5 tests

game-core/src/test/java/games/strategy/triplea/util/MovableUnitsFilterTest.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Predicate;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.triplea.java.collections.CollectionUtils;
+
+public class MovableUnitsFilterTest {
+  final GameData gameData = TestMapGameData.REVISED.getGameData();
```
DanVanAtta: nit, these variables can be declared static: https://github.com/triplea-game/triplea/wiki/Java-Code-Conventions#mark-values-and-methods-as-static-where-possible-5489

game-core/src/test/java/games/strategy/triplea/util/MovableUnitsFilterTest.java
```diff
@@ -0,0 +1,115 @@
...
+            gameData, germans, route, false, AbstractMoveDelegate.MoveType.DEFAULT, List.of());
+    return filter.filterUnitsThatCanMove(units, Map.of());
+  }
+
+  private Collection<Unit> germanyUnits(final Predicate<Unit> predicate) {
+    return CollectionUtils.getMatches(germany.getUnits(), predicate);
+  }
+
+  @Test
+  @DisplayName("moving infantry and tanks two territories should filter to just tanks")
```
DanVanAtta: :+1: The display name helps make it clear what the test is doing, excellent to see

game-core/src/test/java/games/strategy/triplea/util/MovableUnitsFilterTest.java
```diff
@@ -0,0 +1,115 @@
...
+  void onlyTanksCanMoveTwo() {
+    final Route route = new Route(germany, easternEurope, kareliaSsr);
+    final Collection<Unit> units = germanyUnits(Matches.unitIsOfTypes(infantryType, armourType));
+    assertThat(units, hasSize(5));
```
DanVanAtta: nit, it seems like this assertion belongs to a different test. Perhaps "assumeThat" may be more appropriate: https://howtodoinjava.com/junit5/junit-5-assumptions-examples/

Caution, would be good to confirm that the assumeThat behaves as expected, IIRC I experimented with it before and it did not abort the test when the assumption was false. If it works, it's more fitting than an assertion.

game-core/src/test/java/games/strategy/triplea/util/MovableUnitsFilterTest.java
```diff
@@ -0,0 +1,115 @@
...
+    final Collection<Unit> justTanks = germanyUnits(Matches.unitIsOfType(armourType));
+    assertThat(justTanks, hasSize(2));
+
+    final var result = filterUnits(route, units);
+    assertThat(result.getErrorMessage(), is(nullValue()));
+    assertThat(result.getWarningMessage(), is("Not all units have enough movement"));
```
DanVanAtta: nit, brittle test code smell. If we change this string in production code, this test will fail unnecessarily. Recommend extracting the string to a constant, eg:
```
@VisibleForTesting
static String NOT_ALL_UNITS_CAN_MOVE = "Not all units have enough movement";
```

I wonder a bit too if maybe the string that is displayed should be a downstream concern. For example, if we were to internationalize the strings (not on our radar, but for the sake of argument), then it will be odd to have the dependency on the string value. Perhaps a cleaner API would return an enum result, then the consumer could map that result to a string. That would make this test read more like:

```
 assertThat(result.getValue(), is(SOME_UNITS_CAN_MOVE));
```

The other two enum values I would presume would be:
```
ALL_UNITS_CAN_MOVE
NO_UNITS_CAN_MOVE
```
This also avoids a good bit of nullable args. Nullable return types  are essentially a SRP violation code-smell, they represent values that have different meanings depending on the value. IE: if null, this variable means one thing (a flag), if not-null, it's a display string for the user. Having enum values would then imply non-null return type, and code that is closer to the front-end could decide about which display string to use.

game-core/src/test/java/games/strategy/triplea/util/MovableUnitsFilterTest.java
asvitkine: It's needed when it throws exception - which is needed because TestMapGameData.REVISED.getGameData() can throw.

Perhaps we should make getGameData() throw a RuntimeException so it's not needed?

@DanVanAtta WDYT?

game-core/src/test/java/games/strategy/triplea/util/MovableUnitsFilterTest.java
asvitkine: How much of the code do you want me to refactor in one PR? Changing this would require changing MoveValidator in a significant way. It has lots of other error messages it can give, such as "Not enough transports" and a bunch of other cases that should not be possible given how this called is called (e.g. "Can only move friendly units").

I'm trying to keep each PR to a reasonable size to make them easier to review and minimize risk, so I don't think this change should be done here.

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
```diff
@@ -0,0 +1,202 @@
...
+          AbstractMoveDelegate.validateMove(moveType, move, player, nonCombat, undoableMoves, data);
+    } finally {
+      data.releaseReadLock();
+    }
+
+    return new MoveValidationResultWithDependents(
+        result, result.isMoveValid() ? unitsWithDependents : List.of());
+  }
+
+  private List<Unit> addMustMoveWith(
```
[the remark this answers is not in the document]
asvitkine: @DanVanAtta I'm moving this method. I actually simplified it slightly by removing an extra condition - i.e. can just check whether .get() returned null instead of checking membership, but still codeclimate complains it's too complex.

Given the type of code in TripleA, this seems an unreasonable complaint. I think this method is of reasonable complexity for TripleA and I'd be happy if most of the code of TripleA is of this complexity. (Most is much more complex.)

Perhaps we should change the codeclimate threshold? The alternative is to make a ton of small helper methods which introduces its own complexity (i.e. makes classes longer and hard to understand how things fit together).

@DanVanAtta Do you have a better suggestion on what to do here? I mean, certainly we could change helpers...

For example, some ideas I see:
  - Make bestWithDependents a Set, which can avoid the contains() call, but then we have to convert to ArrayList on return.
  - We could make MustMoveWithDetails have some method to do some of this functionality. For example, instead of calling of getMustMoveWith().get(u), we could call a method that returns the units it must move with - never null - which can avoid the null check.

But even with all that, it's not obvious this will reduce the complexity enough. Is there a way to test codeclimate locally - or do I need to upload new patch sets and wait for codeclimate to comment on the PR?

game-core/src/test/java/games/strategy/triplea/util/MovableUnitsFilterTest.java
asvitkine: I think this throws, which I'm not sure it can be static? See my other comment - maybe we make it throw RuntimeException to avoid this?

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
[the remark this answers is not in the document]
asvitkine: @DanVanAtta suggestion on what can be simplified here?

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
asvitkine: But the way the code is set up right now is it's the same behavior?

Do you mean just having getErrorMessage() and getWarningMessage() both call getErrorOrWarningMessage() based on lastResult?

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
asvitkine: Isn't it better to have a comment on each class anyway? I'm OK with renaming, but I'm not sure that's a good reason not to have a comment?

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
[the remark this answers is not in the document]
asvitkine: Also, the logic in this function doesn't make sense to me, particularly this part:

```
      final Collection<Unit> mustMoveWith = mustMoveWithDetails.getMustMoveWith().get(u);
      if (mustMoveWith != null) {
        for (final Unit m : mustMoveWith) {
          if (!bestWithDependents.contains(m)) {
            bestWithDependents.addAll(mustMoveWith);
          }
        }
      }
```

It seems ```mustMovewith``` gets added multiple types for each unit in ```mustMoveWith``` - which seems incorrect. It seems the correct thing is to have the inner statement be ```bestWithDependents.add(m);``` - but this PR is not meant to change behavior here... so....

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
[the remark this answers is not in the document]
asvitkine: This specific comment by the way was on a previous version of the code before I moved some functionality to helpers. Now the function is ~30 lines of code.

One thing that feels like it would make sense to move is the logic inside the  ```!allUnitsResult.isMoveValid()``` block, but actually doing this exercise reveals that a lot of params have to be passed: lastUnitsResult, dependentUnits, route, best, etc. Which ends up being pretty complex.

I could make some of those be instance variables of the class. Right now, it's a bit arbitrary which parameters are passed to the ctor and which are passed to this function - so we could move everything to the ctor, which could simplify some of the internal passing of parameters between functions at the cost of keeping more state in the class. It's not obvious that's a win over just having the logic here.

By the way, the function is already significantly simplified over the original. Perhaps would should agree that "strictly better" changes are OK even if they trigger codeclimate warnings? @DanVanAtta WDYT?

game-core/src/test/java/games/strategy/triplea/util/MovableUnitsFilterTest.java
asvitkine: I sent a separate PR to change getGameData() that would allow this to be removed.

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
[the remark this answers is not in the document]
asvitkine: I now made a helper getInitialUnitList() which should make codeclimate happier. Personally, I don't find this actually helps readability much - because now it's less obvious what's going on - and the flow was already sequential. But it reduces lines of code here and complexity, so I guess it makes the metrics happy...

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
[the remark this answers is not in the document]
asvitkine: For now, I've simplified this function so that the null check is avoided and also reduced the number of params needed for MoveValidator.getMustMoveWith(), which should hopefully satisfy codeclimate...

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
```diff
@@ -0,0 +1,203 @@
...
```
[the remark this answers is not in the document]
asvitkine: Looks like my change to reduce the number of parameters passed to MoveValidator did not sufficiently reduce the complexity of this method.

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
asvitkine: Maybe. Another option is to just have message and an enum of (all units can move, some units can move, no units can move) and a message.

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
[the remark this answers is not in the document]
DanVanAtta: @asvitkine you're right that we should not just make the metrics happy and seek to fundamentally improve the code. I think solving the code climate issues largely requires good faith efforts to spend more time resolving the issues. I recently had a method too large issue and eventually found a good way to shorten it up. My first reaction to the warning was negative though and thought the method seemed 'fine'.

IMO it's fine to respond to some of the warnings with "existing warning, this is moved code", and keep it short. I think we can all agree it's better if we can get these issues fixed, otherwise I wonder when if ever we'd fix them.

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
DanVanAtta: Good comments are good, particularly to explain the purpose and utility of a class. On the other hand this comment I think could be baked into the code and the code would be more self documenting. In this example, a comment describing exactly what a filter operation is perhaps could be useful. It could also be after a rename that the class is self-explanatory enough that the commentary is not needed.

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
DanVanAtta: Looking at other parts of code, it looks like having an enum be the result would lead to a better separation of concerns.

For example, MovePanel already 'owns' this status string:
```
      setStatusWarningMessage("Not all units can move there");
```

It's a bit odd to have some status strings owned by a filter class, and others by the panel. We can avoid the nullity SRP issues as well by having a single result flag, and just let the MovePanel own the status/error/warning strings.

WDYT @asvitkine ?

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
DanVanAtta: There are multiple ways to achieve the objective here. IMO converting to an enum flag solves a number of concerns, including this one. Otherwise, your hunch to use multiple helper methods sounds reasonable.

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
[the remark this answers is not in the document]
DanVanAtta: Explicitly, if our goal to fix method too long warnings is to shorten 2 or 3 lines, then I'd agree we're only making the metrics happy. I suspect a method that is 'that' long should be broken up a good deal. If we just extract one or two lines to a helper method, the level of abstraction could become 'mixed'. EG: a long method with only low level code, if we extract just a bit to a method call, the method call will be a 'higher level' of abstraction, in which case we lots of low level code and then a method call. Hence, extracting more/most of the method into helper function is probably a good way to go.

game-core/src/test/java/games/strategy/triplea/util/MovableUnitsFilterTest.java
DanVanAtta: I was wrong about the empty constructor not being needed here, I think Roi had the same comment on another PR.

game-core/src/test/java/games/strategy/triplea/util/MovableUnitsFilterTest.java
DanVanAtta: I agree with that direction. It's not great we have to use full-blown XML maps IMO to do testing, for now though it's certainly an improvement and helps enable cleaner tests.

DanVanAtta:
@asvitkine how do you feel about the enum conversation? Do you disagree with the possible benefits, is it something you'd prefer for a follow-up, or are you okay incorporating it here?

asvitkine:
> @asvitkine how do you feel about the enum conversation? Do you disagree with the possible benefits, is it something you'd prefer for a follow-up, or are you okay incorporating it here?

Ok, let me try the enum approach. I'llm also try to address the comment about the unit test to do the assume thing.

game-core/src/test/java/games/strategy/triplea/util/MovableUnitsFilterTest.java
asvitkine: It does not appear that assumeThat exists in the version of JUnit we have. assumingThat() exists, but it takes an Executable, which requires making the test flow kind of ugly. So I've left as is.

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
asvitkine: Renamed the class. Kept the comment though.

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  /**
+   * Filters the units and their dependents to a subset of units that can move on the route.
+   *
+   * @param units The units to filter.
+   * @param dependentUnits The dependent units map.
+   * @return The result.
+   */
+  public FilterOperationResult filterUnitsThatCanMove(
```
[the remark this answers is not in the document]
asvitkine: Proposing to land as-is. We can think about ways to simplify things in future PRs - or perhaps consider changing the threshold.

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
asvitkine: Went with the enum, but still keeping the message as coming from MoveValidator.

asvitkine:
@DanVanAtta - can this be merged now? I have several other changes blocked by this.

game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
```diff
@@ -1123,119 +1107,37 @@ private void updateUnitsThatCanMoveOnRoute(final Collection<Unit> units, final R
...
-      }
-      if (!lastResult.getResult().isMoveValid()) {
-        setStatusErrorMessage(message);
+
+    final MovableUnitsFilter unitsFilter =
+        new MovableUnitsFilter(getUnitOwner(units), route, nonCombat, moveType, getUndoableMoves());
+    final var result = unitsFilter.filterUnitsThatCanMove(units, dependentUnits);
+    switch (result.getStatus()) {
+      case NO_UNITS_CAN_MOVE:
+        setStatusErrorMessage(result.getWarningOrErrorMessage().orElseThrow());
```
DanVanAtta: Just to convey, I was thinking the 'getWarningOrErrorMessage' String would be defined & set here. In that case the 'result' object would be a pure result without any strings in it. NBD.

DanVanAtta:
Thanks for the updates. LGTM :+1:

PR #5709, opened 2019-12-13
Simplify, route end and start are never null, remove null checks.

DanVanAtta:
## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

## Additional Review Notes
Just removing conditionals that are always true or always false. Perhaps easiest to review with white space changes turned off.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/RouteScripted.java
+++ b/game-core/src/main/java/games/strategy/engine/data/RouteScripted.java
@@ -37,17 +37,9 @@ public class RouteScripted extends Route {
   }
 
   @Override
-  public Territory getEnd() {
-    if (super.getEnd() == null) {
-      return super.getStart();
-    }
-    return super.getEnd();
-  }
-
-  @Override
   public Territory getTerritoryAtStep(final int i) {
     try {
-      if (super.getEnd() == null || super.getTerritoryAtStep(i) == null) {
+      if (super.getTerritoryAtStep(i) == null) {
         return super.getStart();
       }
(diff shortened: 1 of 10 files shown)


game-core/src/main/java/games/strategy/engine/data/RouteScripted.java
```diff
@@ -36,18 +36,10 @@ public int numberOfSteps() {
...
-      return super.getStart();
-    }
-    return super.getEnd();
-  }
-
   @Override
   public Territory getTerritoryAtStep(final int i) {
     try {
-      if (super.getEnd() == null || super.getTerritoryAtStep(i) == null) {
+      if (super.getTerritoryAtStep(i) == null) {
```
RoiEXLab: getTerritoryAtStep has to be non-null as well, right?
So the only case this code does anything is when i is outside of the array bounds and an exception is thrown.

RoiEXLab: Just had a look at the other part of the method, this is really ugly :(

DanVanAtta: It was also interesting how much copy/paste there was in this clean-up, many of the null checks were clearly just copied from other parts of the code.

> getTerritoryAtStep has to be non-null as well, right?

Likely should be non-null as we disallow adding null territories. It might be easy to find and clean up as well.

PR #5712, opened 2019-12-13
Add ansible parameter and variable for nginx certs

DanVanAtta:
Make nginx certs a variable. If cert file does not exist,
a cert is generated.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

## Additional Review Notes
After this update, the full ansible deploy to a vagrant VM completes successfully.

### DIFF
--- a/infrastructure/ansible/roles/nginx/tasks/main.yml
+++ b/infrastructure/ansible/roles/nginx/tasks/main.yml
@@ -6,7 +6,7 @@
 
 - name: create SSL keys if needed
   become: true
-  command: openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout {{ nginx_ssl_certificate_key }} -out {{ nginx_ssl_certificate }} -batch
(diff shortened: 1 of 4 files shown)


infrastructure/ansible/roles/nginx/tasks/main.yml
```diff
@@ -4,6 +4,12 @@
     name: nginx
     state: present
 
+- name: create SSL keys if needed
+  become: true
+  command: openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout {{ nginx_ssl_certificate_key }} -out {{ nginx_ssl_certificate }} -batch
```
RoiEXLab: Not sure about the letsencrypt config, but at least on prod I prefer to generate 4096 bit keys, might make sense to keep that consistent

RoiEXLab: I udated the command, I realized that cerbot isn't part of ansible so far, so nevermind.
Just something to keep in mind.

PR #5716, opened 2019-12-15
 Change TestMapGameData.getGameData() to not throw Exception.

asvitkine:
It's not expected that the enumerated test game maps have errors loading - so simplify the code so that tests don't have to specify that they throw Exception.

Updates tests accordingly, which allows a lot of code to be simplified.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[X] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done


DanVanAtta:
Good update. IMO as a next step we likely want to reduce the number of game XML maps used for test and see if we can decouple code from needing a full XML to be tested. Those are medium/longer term efforts though.

PR #5717, opened 2019-12-15
Automate localhost install of cert from vagrant machine

DanVanAtta:
Automates installation of NGINX SSL certificate to localhost (ansible control machine)
when deploying to a vagrant VM. Assumes localhost is linux-flavored. If running on windows OS,
will likely need to do a linux VM to simulate the control machine, and then can use vagrant
from within the VM to simulate a target machine.

Change summary:
1. Use vagrant group_vars to add override for vagrant env to place certs in '/vagrant/.vagrant/'
The '/vagrant/' folder is a special mount on the virtual machine that maps to the local machine:
  'triplea/infrastructure/'. Hence '/vagrant/.vagrant/' maps to 'triplea/infrastructure/.vagrant'

2. Use run_ansible_vagrant script to install generated cert on localhost:
 A) symlink the nginx cert from 'infrastructure/.vagrant/(cert-file)' to '/usr/local/share/ca-certificates/'
 B) when done with ansible deployment, run 'update-ca-certificates -f' to refresh certificates
    and pick up any new or updated certs in '/usr/local/share/ca-certificates/'

3. Update cert installation step to run if either cert file does not exist.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[x] Other:  vagrant deployment automation 

## Testing

[x] Manual testing done

### Clean install
```
./gradlew clean
cd infrastructure
vagrant halt
vagrant destroy
vagrant up
./run_ansible_vagrant
```
- Then connected to lobby on "https://localhost:8000"

#### A
- removed the certifcate file in 'infrastructure/.vagrant/' 
- re-ran automation (`./run_ansible_vagrant`)
- noticed the headed game had to be restarted to pick up the updated certs
- reconnected to lobby (https://localhost:8000)

#### B
- removed the certificate on the VM in /etc/nginx
- re-ran automation (`./run_ansible_vagrant`)
- restarted headed game
- reconnected to lobby (https://localhost:8000)

### DIFF
--- a/infrastructure/README-VAGRANT-DEV.md
+++ b/infrastructure/README-VAGRANT-DEV.md
@@ -45,26 +45,10 @@ sudo chown $USER:$USER -R ~/triplea/infrastructure/.vagrant/
 ### HTTPS Config
 
 Ansible will setup a self-signed certificate to be used by nginx.
-So in order to actually access the server
-without any exception thrown by the TLS layer,
-you'll have to import this certificate into your OS' root certificates.
-To extract the certificate on to your system,
-you'll have to install the vagrant-scp plugin like this:
-```bash
-vagrant plugin install vagrant-scp
-```
-Then you can copy it to a convenient place on your system in order to import it.
-On Ubuntu you'd run:
-```bash
-sudo vagrant scp :/etc/nginx/cert.crt /usr/local/share/ca-certificates/triplea_vagrant.crt
-sudo chmod 644 /usr/local/share/ca-certificates/triplea_vagrant.crt
-sudo update-ca-certificates
-```
-to import the certificate into your local keystore.
-Now you can use the self-signed certificate like any other CA-validated certificate.
-You'll have to do this every time the self-signed certificate expires
-or is replaced manually.
-Currently it's valid for 356 days.
+`run_ansible_vagrant` will symlink this certificate from the vagrant virtual
+machine to your local '/usr/local/share/ca-certificates' where it will
+be picked up and added as a trusted certificate.
+
 
 ## Check Results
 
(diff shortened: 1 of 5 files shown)


infrastructure/README-VAGRANT-DEV.md
```diff
@@ -45,26 +45,10 @@ sudo chown $USER:$USER -R ~/triplea/infrastructure/.vagrant/
...
-```
-to import the certificate into your local keystore.
-Now you can use the self-signed certificate like any other CA-validated certificate.
-You'll have to do this every time the self-signed certificate expires
-or is replaced manually.
-Currently it's valid for 356 days.
+`run_ansible_vagrant` will symlink this certificate from the vagrant virtual
+machine to your local '/usr/local/share/ca-certificates' where it will
+be picked up and added as a trusted certificate.
+
```
RoiEXLab: Might be worth pointing out how to force a recreation of the certificate in case it runs out.

DanVanAtta: We should probably automate that in some manner. That way we can be confident the renewal code will work. If we are only renewing for vagrant, it's not worth it and we should figure out how to automate letsencrypt renewal.

RoiEXLab: Well letsencrypt renewal is the easiest thing ever, run certbot once and it will setup a cronjob for you (using those exact certificate settings you just provided) and if you provide an email it will even send a notification if cert renewal fails for some reason.
The notification saved me so many times. The forum server was one of the "early adopters" of letsencrypt so back in the day there wasn't an automated setup for nginx, which means I had to setup nginx myself accordingly, which wasn't done properly at first.

TL;DR Install certbot, run it, it gets renewed automatically.

RoiEXLab: Besides that, letsencrypt certs are valid for 90 days 'only' for security, so certificate that are valid for a larger period should be avoided in production.

Issue #5720, opened 2019-12-16
Paratroopers broken


asvitkine:
**Describe the bug**
Loading and moving paratroopers currently runs into two exceptions:

java.lang.IllegalArgumentException
        at com.google.common.base.Preconditions.checkArgument(Preconditions.java:128)
        at games.strategy.engine.data.MoveDescription.<init>(MoveDescription.java:29)
        at games.strategy.triplea.ui.MovePanel$1.selectUnitsToMove(MovePanel.java:308)
        at games.strategy.triplea.ui.MovePanel$1.unitsSelected(MovePanel.java:129)
        at games.strategy.triplea.ui.MapPanel.notifyUnitSelected(MapPanel.java:499)
        at games.strategy.triplea.ui.MapPanel$3.mouseReleased(MapPanel.java:227)

And when fixing that:

Caused by: java.lang.UnsupportedOperationException
        at java.base/java.util.Collections$UnmodifiableCollection.addAll(Collections.java:1070)
        at games.strategy.triplea.delegate.MovePerformer.markTransportsMovement(MovePerformer.java:410)
        at games.strategy.triplea.delegate.MovePerformer$3.execute(MovePerformer.java:169)
        at games.strategy.triplea.delegate.ExecutionStack.execute(ExecutionStack.java:34)
        at games.strategy.triplea.delegate.MovePerformer.populateStack(MovePerformer.java:322)
        at games.strategy.triplea.delegate.MovePerformer.moveUnits(MovePerformer.java:76)
        at games.strategy.triplea.delegate.MoveDelegate.performMove(MoveDelegate.java:690)

Both are caused by my commits - will work on fixing.

**Engine Version or SHA**
224059ffde15bf204451650477c5e40b2a541d70

**To Reproduce**
1. On WW2v3 1942, give russia paratroopers tech
2. Select bomber in Russia
3. Load up an infantry on the dialog that pops up
4. Move it to Finland

**Expected behavior**
It should move without error. Instead, it hits the above errors.

**Screenshots**

**OS (please put an X in the appropriate box):**
[] Windows
[X] Mac
[] Linux

**Zipped Save Game**
[paratroopers_test.tsvg.zip](https://github.com/triplea-game/triplea/files/3966030/paratroopers_test.tsvg.zip)

**Additional context**
The above works on 1.9.
However, even on 1.9, I noticed another bug - which is that trying to go through a sea zone, like Black Sea doesn't work - gives a "Not enough transports error".

PR #5721, opened 2019-12-16
Fix paratroopers.

asvitkine:
Paratroopers were broken by some recent commits.
This fixes the issues.

There were two problems:
  1. MoveDescription was not allowed to be constructed with a Route with no steps, but turns out that it was not needed at all and that codepath could just be removed.
  2. We were trying to modify some unmodifiable collections, which is fixed by making copies.

Additional, cleans up a couple things in related code - using isEmpty() instead of size() checks and removing an unnecessary null check.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix:  https://github.com/triplea-game/triplea/issues/5720
[] Other:   

## Testing

[X] Manual testing done

Loading paratrooper as described in the issue.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
@@ -299,7 +299,7 @@ public class MovePanel extends AbstractMovePanel implements KeyBindingSupplier {
                       candidateAirTransportsMatch);
               // candidateAirTransports.removeAll(selectedUnits);
               candidateAirTransports.removeAll(dependentUnits.keySet());
-              if (unitsToLoad.size() > 0 && candidateAirTransports.size() > 0) {
+              if (!unitsToLoad.isEmpty() && !candidateAirTransports.isEmpty()) {
                 final Collection<Unit> airTransportsToLoad =
                     getAirTransportsToLoad(candidateAirTransports);
                 selectedUnits.addAll(airTransportsToLoad);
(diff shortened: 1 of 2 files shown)


game-core/src/main/java/games/strategy/triplea/ui/MovePanel.java
```diff
@@ -299,20 +299,14 @@ private void selectUnitsToMove(
...
                 if (!airTransportsToLoad.isEmpty()) {
                   final Collection<Unit> loadedAirTransports =
                       getLoadedAirTransports(route, unitsToLoad, airTransportsToLoad, player);
                   selectedUnits.addAll(loadedAirTransports);
-                  final Map<Unit, Unit> unitsToTransports =
-                      TransportUtils.mapTransportsToLoad(loadedAirTransports, airTransportsToLoad);
-                  final MoveDescription message =
-                      new MoveDescription(
-                          loadedAirTransports, route, unitsToTransports, dependentUnits);
-                  setMoveMessage(message);
```
ron-murhammer: Can you clarify why this can be removed?

asvitkine: setMoveMessage() is for the move that will be returned via waitForMove(), but this interaction does not actually result in a move that is sent to the backend, instead this just initiates the move and the move that will be returned is when the final destination is selected. I've verified that moving paratroopers still works with this removed.

ron-murhammer: Ok. Yeah, I looked at this code more closely and can't figure out why the move message would need set since this is just finding the selected units to move. The move message should always be set when you select an end point for the units and actually have a route.

PR #5723, opened 2019-12-18
Improve algorithm for loading units onto transports in move UI.

asvitkine:
Previously, a move of 3 infantry and 2 tanks to one transport would not result in the algorithm choosing 1 tank and 1 infantry to load - instead, it loaded just one tank.

This is because it sorted the list as {inf, inf, inf, tank, tank} and tried all tail sublists of that list to see what fits.

This change fixes that by just using the mapTransports() algorithm to choose the units that can be loaded and adds unit tests.

Also some more refactoring to make codeclimate happy.

Note: This is an improvement over 1.9 which has the same behavior as current master.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix:  
[] Other:   

## Testing

[X] Manual testing done

On Revised, as Germany try to move 3 infantry and 2 tanks onto a single transport that's in sz5. One tank and infantry should be loaded with this change. Before, only a single tank was.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
+++ b/game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
@@ -120,35 +120,54 @@ public final class MovableUnitsFilter {
     final MoveValidationResult allUnitsResult = lastResult.getResult();
 
     if (!allUnitsResult.isMoveValid()) {
-      // if the player is invading only consider units that can invade
-      if (!nonCombat
-          && route.isUnload()
-          && Matches.isTerritoryEnemy(player, data).test(route.getEnd())) {
-        best = CollectionUtils.getMatches(best, Matches.unitCanInvade());
-        lastResult = validateMoveWithDependents(best, dependentUnits, transportsToLoad);
-      }
-      final boolean hasLandTransports =
-          TechAttachment.isMechanizedInfantry(player)
-              && best.stream().anyMatch(Matches.unitIsLandTransport());
-      final Predicate<Unit> isLandTransportable = Matches.unitIsLandTransportable();
-
-      while (!best.isEmpty() && !lastResult.getResult().isMoveValid()) {
-        final Unit firstSkipUnit = best.get(0);
-        int startIndex = 1;
-        // Check if we can skip more than one unit if they are equivalent (e.g. all units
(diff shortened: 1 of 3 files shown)


game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
```diff
@@ -87,81 +86,105 @@ public FilterOperationResult(
...
+  private MoveValidationResultWithDependents chooseSubsetOfUnitsThatCanMove(
+      final List<Unit> units,
+      final Collection<Unit> transportsToLoad,
+      final MoveValidationResultWithDependents initialResult) {
+    if (!transportsToLoad.isEmpty()) {
+      final List<Unit> allUnits = addMustMoveWith(units);
+      final Collection<Unit> loadedUnits =
+          TransportUtils.mapTransports(route, allUnits, transportsToLoad).keySet();
+      return validateMoveWithDependents(loadedUnits, transportsToLoad);
+    }
```
RoiEXLab: This if case does make sense to me, but when is transportsToLoad empty?
And what does the rest of the method do in that case?
I really have no clue about the logic

asvitkine: The rest of the logic is actually unchanged - refactored from the previous code.

game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
```diff
@@ -87,81 +86,105 @@ public FilterOperationResult(
...
+    MoveValidationResultWithDependents lastResult = initialResult;
+    List<Unit> best = units;
+    // if the player is invading only consider units that can invade
+    if (isInvading()) {
+      best = CollectionUtils.getMatches(best, Matches.unitCanInvade());
+      lastResult = validateMoveWithDependents(best, List.of());
+    }
+
+    final boolean hasLandTransports = hasLandTransports(units);
+    best.sort(getUnitComparator(best).reversed());
```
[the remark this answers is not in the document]
asvitkine: This was actually in getInitialUnitList() previously, but the sorting was only needed before the sublist logic (below this line), so I moved it further down.

PR #5727, opened 2019-12-20
Simplify addMustMoveWith().

asvitkine:
The previous logic was more complex than it needed to be.
This version is equivalent and is quite a bit simpler.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[X] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
+++ b/game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
@@ -21,6 +21,7 @@ import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/triplea/util/MovableUnitsFilter.java
```diff
@@ -223,15 +224,10 @@ private MoveValidationResultWithDependents validateMoveWithDependents(
...
-      final Collection<Unit> mustMoveWith = mustMoveWithDetails.getMustMoveWithForUnit(u);
-      for (final Unit m : mustMoveWith) {
-        if (!bestWithDependents.contains(m)) {
-          bestWithDependents.addAll(mustMoveWith);
-        }
-      }
+      bestWithDependents.addAll(mustMoveWithDetails.getMustMoveWithForUnit(u));
     }
-    return bestWithDependents;
+    return new ArrayList<>(bestWithDependents);
```
RoiEXLab: I briefly checked the callers of this method, and I'm almost sure you could return a Set here directly.
Of course this would require some refacoring in other places to accept `Collection` instead of `List`, but that should really be it. Avoids the need to create a list-copy of the set

asvitkine: It might be possible, but quickly checking it would make this PR much bigger - since the type ends up making it to unitsWithDependents on FilterOperationResult and thus part of the API.

Let me look at doing that in a separate PR after this one instead.

RoiEXLab: Yeah of course, I didn't want you to do this as part of this PR.
So that's fine

RoiEXLab:
I was thinking of changing the implementation type to `LinkedHashMap` to preserve the insertion order, but then I realized that this diff actually has a subtle change in behaviour.
Previously the method could potentially return a list with duplicate elements given the right conditions.
I saw a couple of checks for this down the call hierarchy, so these would become redundant if they are only ever supplied by this piece of code. Looking forward to your follow-up PR.

PR #5728, opened 2019-12-21
Do not re-create swing components for every change

RoiEXLab:
Instead of re-creating the components every time a new territory is hovered over, we can simply modify the contents instead.
Avoiding some overhead in redrawing all the components.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done


RoiEXLab:
I did some more or less accurate tests before, but for the sake of accuracy I just ran the tests again.
This change roughly saves 1-2ms on average in this method on my machine.
However I realized that the equality check I implemented is redundant, so that's a bonus

RoiEXLab:
What's perhaps noteworthy:
Previously if the new territory was null, it took almost 2ms most of the time to remove all the components, now this takes < 1ms because the elements are only being hidden, not removed.

PR #5729, opened 2019-12-21
Tile locking update

RoiEXLab:
I realized that the engine did lock the UI Thread every time isDirty was called, which wastes a lot of time.
Instead the if check was changed to somewhat hide the inner logic and only lock when necessary.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/MapPanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/MapPanel.java
@@ -677,15 +677,15 @@ public class MapPanel extends ImageScrollerLargeView {
     final Rectangle2D.Double mainBounds =
         new Rectangle2D.Double(x, y, getScaledWidth(), getScaledHeight());
     drawTiles(g2d, images, data, mainBounds, undrawnTiles);
-    if (routeDescription != null && mouseShadowImage != null && routeDescription.getEnd() != null) {
-      final AffineTransform t = new AffineTransform();
-      t.translate(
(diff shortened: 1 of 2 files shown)


game-core/src/main/java/games/strategy/triplea/ui/MapPanel.java
```diff
@@ -677,15 +677,15 @@ public void paint(final Graphics g) {
...
-    }
     if (routeDescription != null) {
+      if (mouseShadowImage != null && routeDescription.getEnd() != null) {
+        final AffineTransform t = new AffineTransform();
+        t.translate(
+            normalizeX(routeDescription.getEnd().getX() - getXOffset()),
+            normalizeY(routeDescription.getEnd().getY() - getYOffset()));
+        t.translate(mouseShadowImage.getWidth() / -2.0, mouseShadowImage.getHeight() / -2.0);
+        g2d.drawImage(mouseShadowImage, t, this);
+      }
```
ron-murhammer: Not completely following this change. Can you add some context?

RoiEXLab: This is just a semantic change.
Previously both if cases checked if RouteDescription is non null, now there's a single if check and a nested if statement.

It's just a random change I had lying around locally.

PR #5734, opened 2019-12-22
Adjust `@Nullable` annotations

RoiEXLab:
I noticed that the `@Nullable` annotations are applied a bit inconsistent for the `MapSelectionListener` interface.

This PR fixes this inconsistency and removes redundant null checks, I verified that the only values ever supplied are non-null there.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
@@ -229,9 +229,6 @@ class EditPanel extends ActionPanel {
       new DefaultMapSelectionListener() {
         @Override
         public void territorySelected(final Territory territory, final MouseDetails md) {
-          if (territory == null) {
-            return;
-          }
           if (currentAction == changeTerritoryOwnerAction) {
             final TerritoryAttachment ta = TerritoryAttachment.get(territory);
             if (ta == null) {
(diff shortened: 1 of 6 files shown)


game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
```diff
@@ -229,9 +229,6 @@ private void selectUnitsToRemove(
       new DefaultMapSelectionListener() {
         @Override
         public void territorySelected(final Territory territory, final MouseDetails md) {
-          if (territory == null) {
```
DanVanAtta: Are you sure this is non-nullable, or were we missing the @Nullable annotation?

Side-note on a larger question beyond the scope of this update: I wonder if we would have been better off marking things both as '@Nullable' and/or as "@Nonnull" until we got everything marked. Once everything is marked, then we would be safe to know that a missing '@Nullable' truly means '@Nonnull'

DanVanAtta: bump @RoiEXLab

RoiEXLab: Thanks for the reminder.
I did verify this is non-nullable.
I thought about adding a non-null annotation as well, but wasn't really sure about that

PR #5735, opened 2019-12-22
Use doubles over BigDecimal

RoiEXLab:
Just stumbled over this a week ago, it's super weird having to deal with BigDecimals here, not really necessary at this point, rounding errors are ok, so we save some overhead.

@ron-murhammer Do you know why BigDecimals are being used everywhere?
I assume that the type is being used for currencies, and somehow managed to spread all over the codebase?

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/RouteFinder.java
+++ b/game-core/src/main/java/games/strategy/engine/data/RouteFinder.java
@@ -3,7 +3,6 @@ package games.strategy.engine.data;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import games.strategy.triplea.delegate.TerritoryEffectHelper;
-import java.math.BigDecimal;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


ron-murhammer:
@RoiEXLab I believe we need to use BigDecimal here to avoid precision issues with doubles when having decimal movement costs.

RoiEXLab:
@ron-murhammer Are you sure?
I do realize that this might introduce rounding issues, but I believe that because this is 'just' the route finding algorithm it would also work somewhat decently if we rounded the doubles to ints.

Can you think of a specific example where rounding is an issue?

ron-murhammer:
@RoiEXLab Fairly sure. Its related to this PR which add potential decimal movement costs: https://github.com/triplea-game/triplea/pull/5234

Essentially movement needs to be treated like a resource/currency as you are going to do comparisons of a unit's remaining movement and the territory movement cost so the values need to have full precision.

DanVanAtta:
My 2 cents, 'MovementPoints' perhaps is a high enough of a concept it could be abstracted to its own type. There are a number of operations that would likely fit with it (perhaps methods accepting routes, units, and others to do our calculations).

Once extracted to a type, the ultimate answer IMO here is to leave it to a unit test to verify whether we need the precision or not. The abstracted type would make that easy to test and also allow for us to swap out any underlying data representation and keep the scope of change to one class.

The move calculations already being landed, I think we may be past a point where that is palatable. Ultimately IMO we need tests to tell us whether we have this constraint or not, and to enforce it for future maintainers to know if they have violated the requirement or not when we update this code next.

RoiEXLab:
I'm not really convinced that doubles are not precise enough, but meh

PR #5741, opened 2019-12-23
Add JavaFX map selection screen

RoiEXLab:
This PR adds a map selection screen for the JavaFX client.
You get to this screen by clicking on "Local Game" in the game-selection screen.
Currently it's only displaying a placeholder image, but maybe in the future we can update the xml to include an image to be shown as thumbnail.

Next step is to add a player-selection screen and then we could in theory start a local game using the new client.
So maybe in Summer 2020, we get to a point where all "new" UI development can be done in JavaFX, and we can all partially replace existing components.

## Functional Changes

[] New map or map update
[x] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done

## Screens Shots

Screenshot1
Screenshot2


game-headed/src/test/java/org/triplea/game/client/ui/javafx/screens/MapSelectionTest.java
```diff
@@ -0,0 +1,111 @@
...
+    final Node node = mock(Pane.class);
+    mapSelection.setPreviewContainer(node);
+
+    mapSelection.closeDetails();
+
+    verify(node).setVisible(false);
+  }
+
+  @Test
+  void testDetailsButtonShowsDetails() throws Exception {
```
DanVanAtta: side-note and nit: IMO the `test` prefix can be removed from the test method names without losing anything. `detailsButtonShowsDetails()` is just as descriptive IMO.

Of more importance though, this is a somewhat lengthy test. It might benefit from extracting the setup code into helper methods. To explain through a rough example:
```
givenPreviewWithWebEngine(webEngine);
givenMapSelectionWithContainerAndWindow(previewContainer, previewWindow);
givenGameDataWithMapName("Test name");

:
:

mapSelection.showDetails();

    verify(previewContainer).setVisible(true);
    verify(webEngine).loadContent("Test Notes\n Test Name \n");
```

Though, IMO it would be even better if we could have decompose the logic a bit more. Notably so that 'showDetails' returns a value (or uses another object that returns a value for us). This way we can test the results of a return value and not need to use `verify`, which is essentially testing for side-effects.  I think perhaps it indicates the underlying logic would benefit from strategy objects to perform the logic rather than owning the logic itself. 

```

File: game-headed/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker [hunk 2]
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
DanVanAtta: This is new, could you provide a reference for what this file is doing or quickly explain it? Does this take the place of `@ExtendWith` at all?

game-headed/src/main/java/org/triplea/game/client/parser/DefaultGameDetector.java
```diff
@@ -0,0 +1,28 @@
+package org.triplea.game.client.parser;
+
+import games.strategy.engine.framework.ui.GameChooserEntry;
+import games.strategy.engine.framework.ui.GameChooserModel;
+import java.util.Set;
+import java.util.function.Consumer;
+import javafx.concurrent.Service;
+import javafx.concurrent.Task;
+
+public class DefaultGameDetector extends Service<Set<GameChooserEntry>> implements GameDetector {
```
DanVanAtta: Would it be possible to use composition instead of inheritance for the `Service`?

RoiEXLab: Well this is just the way the service class works for JavaFX, not sure how to do that differently

game-headed/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
RoiEXLab: Ah, this is just a file that enables mockito to be able to mock final classes, which is required to mock the webview class

RoiEXLab:
The reason for the separate window is simply screen size.
If you can come up with a system that doesn't take a full 1920x1080 to display correctly I can do another iteration on this.
However I think it should be a priority to get working components first and Polish them afterwards.

game-headed/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
DanVanAtta: Interesting, thanks for the explanation

DanVanAtta:
Thanks for responses, @RoiEXLab just the couple comments about potentially improving the maintainability of the tests, please let me know what you think.

RoiEXLab:
@DanVanAtta Updated

game-headed/src/main/java/org/triplea/game/client/ui/javafx/screens/MapSelection.java
```diff
@@ -0,0 +1,149 @@
...
+      final GameDetector gameDetector, final BiFunction<String, String, String> linkLocalizer) {
+    this.gameDetector = gameDetector;
+    this.linkLocalizer = linkLocalizer;
+  }
+
+  private Node createMapListing(final GameChooserEntry gameChooserEntry) {
+    final var button = new Button(gameChooserEntry.getGameName());
+    // Placeholder image
+    final var imageView =
+        new ImageView(new Image("https://triplea-game.org/images/missing_map.png", true));
```
DanVanAtta: - What happens when the map download screen is opened without a network connection?
- Should 'missing_map.png' be hosted on triplea-game.org? Would it be better to bundle the missing_map.png with the game so that a network fetch is not needed? Any concern if 'missing_map.png' were moved and we did not realize game engines needed it? Is there also risk that we repeat this pattern for other images, how do we keep track of which images are hosted on website and used by game engines vs those that are used only by website?

game-headed/src/main/java/org/triplea/game/client/ui/javafx/screens/MapSelection.java
```diff
@@ -0,0 +1,149 @@
...
+    button.setGraphic(imageView);
+    button.setWrapText(true);
+    button.setPrefSize(200, 200);
+    button.setContentDisplay(ContentDisplay.TOP);
+    button.setOnAction(
+        event -> {
+          selectedGame = gameChooserEntry;
+          selectButton.setDisable(false);
+          detailsButton.setDisable(false);
+          mapContainer.getChildren().forEach(node -> node.getStyleClass().remove("selected"));
```
DanVanAtta: If the structure of mapContainer is changed, will this break? What kind of children nodes do we expect here? Would it be safer to be more direct about the unsetting of 'selected' by assigning the nodes of mapContainer to a collection directly on construction, then we could be more explicit about which nodes are operating on.

game-headed/src/main/java/org/triplea/game/client/ui/javafx/screens/MapSelection.java
```diff
@@ -0,0 +1,149 @@
...
+          button.getStyleClass().add("selected");
```
DanVanAtta: Would it make sense to create a 'css constants' class to hold constants defined in CSS? Would that be helpful for new maintainers to be able to determine where the 'selected' string is coming from?

game-headed/src/main/java/org/triplea/game/client/ui/javafx/screens/MapSelection.java
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  @Override
+  public void connect(final ScreenController<FxmlManager> screenController) {
+    this.screenController = screenController;
+  }
+
+  @Override
+  public void onShow() {
+    if (loaded) {
```
DanVanAtta: When is this class instantiated? Can 'onShow' be moved to constructor, any problems with that?

game-headed/src/main/java/org/triplea/game/client/ui/javafx/util/FxmlManager.java
```diff
@@ -19,7 +19,9 @@
 
   GAME_SELECTION_CONTROLS("/org/triplea/game/client/ui/javafx/fxml/GameSelectionControls.fxml"),
 
-  MAIN_MENU_CONTROLS("/org/triplea/game/client/ui/javafx/fxml/MainMenuControls.fxml");
+  MAIN_MENU_CONTROLS("/org/triplea/game/client/ui/javafx/fxml/MainMenuControls.fxml"),
+
+  MAP_SELECTION("/org/triplea/game/client/ui/javafx/fxml/MapSelection.fxml");
```
DanVanAtta: nit, to avoid unnecessary diffs in future, ending enum with comma and putting semi on new line will keep changes a bit easier to grok. EG:
```
MAP_SELECTION("/org/triplea/game/client/ui/javafx/fxml/MapSelection.fxml"),
;
```

game-headed/src/test/java/org/triplea/game/client/ui/javafx/screens/MainMenuPaneTest.java
```diff
@@ -33,21 +33,24 @@ void testGetNode() {
   void testInitialize() throws Exception {
     final NavigationPane mock = mock(NavigationPane.class);
     final StackPane mock2 = mock(StackPane.class);
-    FieldSetter.setField(null, Application.class.getDeclaredField("userAgentStylesheet"), "");
+    final Field userAgentStyleSheetField =
+        Application.class.getDeclaredField("userAgentStylesheet");
+    userAgentStyleSheetField.setAccessible(true);
+    userAgentStyleSheetField.set(null, "");
```
DanVanAtta: What's the 'set' of null needed for? Could this be extracted to a method to make it self-documenting, or alternatively worth explaining in a quick comment.

game-headed/src/test/java/org/triplea/game/client/ui/javafx/screens/MainMenuPaneTest.java
```diff
@@ -64,6 +67,9 @@ void testInitialize() throws Exception {
 
   @AfterAll
   static void cleanup() throws Exception {
-    FieldSetter.setField(null, Application.class.getDeclaredField("userAgentStylesheet"), null);
+    final Field userAgentStyleSheetField =
+        Application.class.getDeclaredField("userAgentStylesheet");
+    userAgentStyleSheetField.setAccessible(true);
+    userAgentStyleSheetField.set(null, null);
```
DanVanAtta: ditto on hard to grok nulling, what's the purpose to set null null?

game-headed/src/test/java/org/triplea/game/client/ui/javafx/screens/MapSelectionTest.java
```diff
@@ -0,0 +1,121 @@
...
+  void mapSelectionObjectReturnsCorrectNode() {
+    final Node node = mock(Node.class);
+    mapSelection.setRoot(node);
+
+    assertThat(mapSelection.getNode(), is(equalTo(node)));
+  }
+
+  @SuppressWarnings("unchecked")
+  @Test
+  void backKeyWorks() {
```
DanVanAtta: nit, is this testing the 'back key'? Or are we testing the back function?

game-headed/src/test/java/org/triplea/game/client/ui/javafx/screens/MapSelectionTest.java
```diff
@@ -0,0 +1,121 @@
...
+    final Node previewContainer = mock(Node.class);
+    mapSelection.setPreviewContainer(previewContainer);
+
+    final WebView previewWindow = mock(WebView.class);
+    final var webEngine = mockWebEngine();
+    when(previewWindow.getEngine()).thenReturn(webEngine);
+    mapSelection.setPreviewWindow(previewWindow);
+
+    final GameChooserEntry gameChooserEntry = mockGameChooserEntryWithNotes();
+    mapSelection.setSelectedGame(gameChooserEntry);
```
DanVanAtta: nit, the setup code is a bit hard to grok. It might be worthwhile to further extract/hide it behind a 'given' method, eg:

```
givenPreviewWindowWithSelectedGameWithNotesAndName(mapNotes,mapName);

mapSelection.showDetails();

verify(previewContainer).setVisible(true);
verify(webEngine).loadContent(mapNotes.trim() + mapName);
```

game-headed/src/test/java/org/triplea/game/client/ui/javafx/screens/MapSelectionTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+    mapSelection.showDetails();
+
+    verify(previewContainer).setVisible(true);
+    verify(webEngine).loadContent(mapNotes.trim() + mapName);
+  }
+
+  private WebEngine mockWebEngine() throws Exception {
+    final Class<?> clazz = Class.forName("com.sun.glass.ui.Screen");
+    final Field screensField = clazz.getDeclaredField("screens");
```
DanVanAtta: Using reflection to set a component of the UI framework seems to potentially be dangerous or maybe even evil. Can you explain why alternatives were not feasible? I wonder a bit if the javafx code is mixed with the UI flow code a bit too much and if the two can be more decoupled. If we can move logic a bit further away from JavaFX, we may be able to avoid tests that need JavaFX component knowledge.

game-headed/src/main/java/org/triplea/game/client/ui/javafx/screens/MapSelection.java
```diff
@@ -0,0 +1,149 @@
...
+import lombok.AccessLevel;
+import lombok.Setter;
+import org.triplea.game.client.parser.DefaultGameDetector;
+import org.triplea.game.client.parser.GameDetector;
+import org.triplea.game.client.ui.javafx.screen.ControlledScreen;
+import org.triplea.game.client.ui.javafx.screen.ScreenController;
+import org.triplea.game.client.ui.javafx.util.FxmlManager;
+import org.triplea.util.LocalizeHtml;
+
+public class MapSelection implements ControlledScreen<ScreenController<FxmlManager>> {
```
DanVanAtta: High-level / important: I strongly suspect this class should have almost no logic in it. Similar to how http controllers delegate most everything, a similar pattern here could help us avoid having to include FX components in test. More specifically, if all logic were delegated, then we'd just have to ensure the delegated logic is invoked and we could test that logic independently and not be coupled to UI layer. WDYT?

RoiEXLab: Well in an ideal world this class should define what all the UI components inside this class are actually doing, so all of the UI logic regarding this exact screen should be here

game-headed/src/test/java/org/triplea/game/client/ui/javafx/screens/MapSelectionTest.java
RoiEXLab: It's a little bit hacky.
The problem in this case is the static initializer block of WebEngine, which is obviously executed once we try to mock this class, so this reflective call avoids some logic that tries to determine the primary screen etc. to be used by instances of the class.

I mean we could try using a framework like https://github.com/TestFX/TestFX + https://github.com/TestFX/Monocle which avoids this problem by "correctly" setting up a JavaFX environment for the tests, but for now this hack seems just good enough for me

game-headed/src/test/java/org/triplea/game/client/ui/javafx/screens/MainMenuPaneTest.java
RoiEXLab: Again, a JavaFX hack, basically equivalent to `Application.userAgentStylesheet = null` to fix some issues with static initializers.

The only reason this test code is touched, because previously I used the internal (non-api) mockito methods to simplify this, but this caused some errorprone warnings to show up.

game-headed/src/main/java/org/triplea/game/client/ui/javafx/screens/MapSelection.java
RoiEXLab: We could do that, but the styles are really managed in the base.css class which is completely unrelated to java, so the actual file could still be something completely different

game-headed/src/main/java/org/triplea/game/client/ui/javafx/screens/MapSelection.java
RoiEXLab: Well no, but maybe.
In the FXML file `mapContainer` is a `TilePane`, with a intermediate loading icon.
Once the screen is shown and we have calculated all the available maps asynchronously, we clear this container and add all the games as buttons with icons to this container.
So this particular container is completely managed by the java code in this class once the games are loaded, so nothing can really break from outside influence at this point, but this isn't necessarily the case for all containers that are managed dynamically.

game-headed/src/main/java/org/triplea/game/client/ui/javafx/screens/MapSelection.java
RoiEXLab: 1. Good question. I would assume nothing (except for a stacktrace in the console) and a missing image, because I tried putting an svg there which isn't supported by JavaFX which did exactly this (without the stacktrace though). I'd have to test this though.
2. It's probably better to bundle it with the game. Initially I wasn't sure if I wanted to implement "real" thumbnails as part of this PR, which is why I used some random placeholder image for this.

game-headed/src/test/java/org/triplea/game/client/ui/javafx/screens/MapSelectionTest.java
DanVanAtta: Monocle could be something that could help, but the point I'm trying to convey is that I don't think we should have any/much logic at the javafx layer. IE: all interactions to JavaFX are probably best done as part of injected strategy objects. This way we have 'pure' UI code that wires into strategy objects that do everything else.

In other words, the code we are testing seems to be too close to the UI code and we do not have them well enough decoupled.

DanVanAtta: Said in yet another way, we might possibly be creating some of the same mistakes as we had with the swing code, where the UI layer and the view handling are not decoupled. Monocle would probably be useful that the UI is rendered and invokes the backend layer we create.

game-headed/src/main/java/org/triplea/game/client/ui/javafx/screens/MapSelection.java
RoiEXLab: I think I might need to explain the pattern to you (or it might be better if you tried to build a simple screen on your own/polish an existing one, similar to my ansible experience).
Basically the *Controller classes are instantiated by JavaFX once for every FXML file, so controllers are essentially singletons, somewhat similar to dependency injection in spring.

`onShow()` was introduced because when I started working on this I had a severe freeze on startup, which I thought would come from the heavy game loading.
`onShow()` is only called when a screen is switched to/made visible on the stack, so this way the freeze would only occur once when people would actually try to play a game, which felt way better.

In the end it turned out that the freeze was because all of the thumbnails were loaded synchronously on the JavaFX thread, so once I changed this addition is not really needed anymore.
In fact if we moved the "loading" mechanism somewhere else, we could have the maps load in the background, so that the user might not even notice that anything took 3 seconds to load because everything is already loaded when they actually start a game.
So thanks for reminding me.

This class is initialized on startup, along with all the other FXML controllers, however at this point none of the dependency injection has occured.
What you want to do is add a method like this
```java
@FXML
void initialize() {
  // initialize everything here
}
```
which is called by the FXML loader once the Nodes have been injected.

game-headed/src/test/java/org/triplea/game/client/ui/javafx/screens/MapSelectionTest.java
RoiEXLab: Somwhere inbetween.
The FXML defines that this method is called whenever the button is pressed, but we can't really check using conventional unit tests that the FXML actually binds the button to this function.
Using JavaFX we could go for end-to-end tests where we simulate a user clicking through all the menus to see if they're bound correctly, but for a simple unit test, this is probably as close as we can get.

RoiEXLab:
@DanVanAtta I don't claim to have the best pattern possible, and I'm still experimenting with a couple of things.
Most of the current structure turned out like this, because this way IntelliJ provides excellent support for it.
If you rename a `fxml:id="abc"` in the FXML file, it renames the corresponding field in the controller as well and vice-versa.
If you change a method's name in your controller class, it renames the value in the corresponding FXML with it and vice-versa.
Which makes it really pleasant to do refactorings.

We could in theory create instances of our controllers and pass them to JavaFX when we load an FXML file to havbe more freedom in that regard, but then we'd lose all of the pleasant stuff I just mentioned (and more).

I don't have a working screen system that allows passing information between screens yet.
We could use the "global" javafx storage (see [`Stage#getProperties()`](https://openjfx.io/javadoc/11/javafx.graphics/javafx/stage/Window.html#getProperties())), which is "global" because we only ever use a single stage for everything. (You can use substages to embed 3D rendering etc. in another stage, but that might only become relevant once I re-do the rendering logic to use hardware acceleration and nicer effects. This might break this idea, but not sure, haven't tried.)
No idea if that's better than coming up with a system similar to androids [Intent](https://developer.android.com/reference/android/content/Intent) class, which basically allows a screen to define a payload which is passed to another screen/app.

In any case I'd agree that it would definitely be possible to tie the logic less to the UI, but I definitely haven't seen an example of it yet.
So if you can have a look at the code and come up with a better system than the current one, I'd happily apply it everywhere.

DanVanAtta:
@RoiEXLab please mark a TODO for this: https://github.com/triplea-game/triplea/pull/5741#discussion_r361219596 and add some quick comments on the 'null, null' testing issues. With that feel free to merge and we can iterate on the pattern/testing to see if we can get it better.

RoiEXLab:
@DanVanAtta Updated

game-headed/src/main/java/org/triplea/game/client/ui/javafx/screens/MapSelection.java
RoiEXLab: I just checked using an invalid domain, just like I expected nothing happens.
The button stays blank, and not even an exception is logged

game-headed/src/main/java/org/triplea/game/client/ui/javafx/screens/MapSelection.java
```diff
@@ -0,0 +1,144 @@
...
+  @VisibleForTesting
+  MapSelection(
+      final GameDetector gameDetector, final BiFunction<String, String, String> linkLocalizer) {
+    this.gameDetector = gameDetector;
+    this.linkLocalizer = linkLocalizer;
+  }
+
+  private Node createMapListing(final GameChooserEntry gameChooserEntry) {
+    final var button = new Button(gameChooserEntry.getGameName());
+    // TODO Placeholder image, ship with engine
```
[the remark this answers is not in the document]
RoiEXLab: @DanVanAtta This is the TODO

PR #5743, opened 2019-12-23
Change ProData to not be static.

asvitkine:
Change ProData to not be static.

This class caches various state related to the Pro AI. It currently has a bunch of static non-final members - i.e. globals, which is not good design. For example, it would prevent allowing the engine to run two independent games in the same program. This is also responsible for 20 different sonarcloud issues here:
https://sonarcloud.io/project/issues?id=triplea-game-sonar&open=AW6pgYL4gYcRdnXYRPNB&resolved=false&types=VULNERABILITY

This PR changes the class to be an instance that's owned by ProAi and passed to methods that need it.

Note: I'm sure additional refactoring can be done to improve the code further after this change, but I recommend making such changes in a follow up. In particular, I don't think it makes sense to try to fix all the codeclimate issues in this PR - many of them are existing - and the ones that are not (e.g. methods going from 4 params to 5) are still improved by this PR by virtue of not relying on global variables after the change.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[X] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[X] Manual testing done

Tested all-AI game on `Revised` to ensure no crashes.


ron-murhammer:
@asvitkine I agree with the direction and generally wanted to refactor this for a while now. I also agree please keep the scope of the PR pretty limited otherwise it'll be very difficult to review. I'm fine ignoring the codeclimate comments and leaving those for follow on PRs.

asvitkine:
Should be ready for review. I tested an all-AI game on Revised to make sure nothing seems broken.

asvitkine:
@ron-murhammer are you good with this being merged?

asvitkine:
@ron-murhammer @RoiEXLab @DanVanAtta 

So is there interest in proceeding with this? If not, I can abandon. It seems given no one merged this, a bunch of conflicts cropped up. I'm not super enthusiastic to resolve them all if this will just sit without any attention as before and accumulate more conflicts. So either we should decide we want this, and I can resolve the conflicts and we can get it merged promptly before accumulating new ones, or I can just abandon this.

RoiEXLab:
I think both @DanVanAtta and I waited for @ron-murhammer's approval on this.
But I think we can merge if you want to fix all of the merge conflicts and address any issues later. Thoughts @DanVanAtta?

ron-murhammer:
@asvitkine The code here looks good. My one concern is making sure that the AI is using the correct game data in all situations as it makes a copy of the game data to simulate a turn for purchases/placements. It looks like you ran into that looking at the commit history. As long as you feel confident that there isn't any places that it changed which game data (real vs copy) that its using then I'm good with this.

asvitkine:
@ron-murhammer 

Ok, I rebased it on master.

asvitkine:
I played another all-AI game on Revised to make sure it works.

The unit test failure seems unrelated?

RoiEXLab:
Yeah, no worries this is a current issue on master

ron-murhammer:
@asvitkine Ok. Merging this. I'll try to do some AI testing to make sure nothing was impacted.

PR #5746, opened 2019-12-24
Group battle related classes from 'delegate' package together

DanVanAtta:
Creates a new sub-package in 'delegate', 'delegate.battle' and groups
battle related classes together.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

## Screens Shots
The classes below in the 'delegate.battle' package used to all be in the 'delegate' package.

Screenshot from 2019-12-23 20-11-10


DanVanAtta:
@RoiEXLab updated the delegate finding regex: 1ff5f9d2c32ef0dccc4f7c599911925ea32a9b2a, let me know if you have any comments.

RoiEXLab:
@DanVanAtta didn't even think of the XmlMapper class.
I the regex is unnecessary, we replace the strings only for backwards compatibility with xmls that use the fully qualified name, so unless you're planning to update those to use the new package name (which you shouldn't for many reasons) this change does only allow map makers to use longer strings instead of the old form

DanVanAtta:
The updated package name made me think the mismatched string was coming from the java class package name. The XML packages were updated: https://github.com/triplea-game/triplea/pull/5746/files#diff-f1bafe5ddd636563c640cb1a80eec2a7

It seems we should update the test files to not use the extended package name format and perhaps keep one with the previous fully qualified name so we can ensure backward compatibility.

RoiEXLab:
> It seems we should update the test files to not use the extended package name format and perhaps keep one with the previous fully qualified name so we can ensure backward compatibility.

Exactly. I think we did it the other way round, changing a single string, but keeping all the others, but basically yeah.

game-core/src/test/java/games/strategy/engine/data/gameparser/XmlGameElementMapperTest.java
```diff
@@ -33,7 +33,7 @@ void shouldReturnDelegateWhenNamePresent() {
     @Test
     void shouldReturnDelegateWhenFullyQualifiedNamePresent() {
       final Optional<IDelegate> result =
-          xmlGameElementMapper.newDelegate("games.strategy.triplea.delegate.BattleDelegate");
+          xmlGameElementMapper.newDelegate("games.strategy.triplea.delegate.battle.BattleDelegate");
```
RoiEXLab: This change is still incorrect

DanVanAtta: Missed that,  :+1: 
Fixed in: 58dafc8

PR #5749, opened 2019-12-24
Ensure JOptionDialog is rendered in swing thread

DanVanAtta:
Fixes potential error if error dialog is requested outside of the EDT thread.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

[] Manual testing done


DanVanAtta:
The precondition would check for EDT, right? The component render fails if not on EDT, in effect we kinda have such a precondition. At issue though is that we don't know we have a non-EDT path until we hit it :(

Isolating the swing code more completely and ensuring we "do it right" in the isolated code perhaps is the best solution.

PR #5750, opened 2019-12-24
Fix websocket send race-condition, make send wait and block.

DanVanAtta:
Summary:
- remove single threaded threadpool for websocket connect and send
- make connect be non-blocking
- have send wait for connection to be open (now blocking)

Problem description:
Observed preconditions failures in 'send' for connection closed.
The previous flow relied on single-threaded operations to ensure
we opened a connection and then sent data over an open connection.

This was ordered by queuing operations onto a single-threaded thread
pool. The suspected problem is that (A) the connect thread spawns
a new thread (B) and waits for it to be finished before we spawn
a new thread (C) to do the send. If (C) is spawned before (B), we
hit a race condition and attempt to send befor the websocket connection
is open.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
@@ -9,10 +9,10 @@ import java.util.concurrent.TimeUnit;
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.Setter;
+import org.awaitility.Awaitility;
+import org.awaitility.core.ConditionTimeoutException;
 import org.java_websocket.client.WebSocketClient;
 import org.java_websocket.handshake.ServerHandshake;
-import org.triplea.java.Interruptibles;
-import org.triplea.java.Interruptibles.Result;
 import org.triplea.java.timer.ScheduledTimer;
 import org.triplea.java.timer.Timers;
 
@@ -28,10 +28,15 @@ import org.triplea.java.timer.Timers;
  * </ul>
  */
 class WebSocketConnection {
-  private static final int DEFAULT_CONNECT_TIMEOUT_MILLIS = 5000;
   private final Collection<WebSocketConnectionListener> listeners = new HashSet<>();
+  private final URI serverUri;
 
   private boolean closed = false;
+
+  @Setter(
+      value = AccessLevel.PACKAGE,
+      onMethod_ = {@VisibleForTesting})
+  private int connectTimeoutMillis = 5000;
 
   @Getter(
       value = AccessLevel.PACKAGE,
@@ -44,6 +49,7 @@ class WebSocketConnection {
   private final ScheduledTimer pingSender;
 
   WebSocketConnection(final URI serverUri) {
+    this.serverUri = serverUri;
     client =
         new WebSocketClient(serverUri) {
           @Override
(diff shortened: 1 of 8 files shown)


http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -115,7 +121,15 @@ void connect() {
...
-    Preconditions.checkState(client.isOpen());
+    if (!client.isOpen()) {
+      try {
+        Awaitility.await()
+            .atMost(connectTimeoutMillis, TimeUnit.MILLISECONDS)
+            .until(() -> client.isOpen());
+      } catch (final ConditionTimeoutException ignored) {
+        throw new CouldNotConnect(serverUri);
+      }
+    }
```
RoiEXLab: IIRC the reason why we introduced the complex logic in the first place was to avoid any kind of blocking behaviour.
Does UI call this method directly? If so I'd actually recommend making `awaitUntilOpen` a separate method, so that the UI can spawn a Thread that waits until the connection is actually open and then enables the UI components to send stuff, whereas this method just gets a precondition that checks if the method is actually open.
This way we could avoid a UI freeze. If this method is invoked on any other Thread besides the EDT, then this is perfectly fine.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -115,7 +121,15 @@ void connect() {
    */
   void sendMessage(final String message) {
     Preconditions.checkState(!closed);
-    Preconditions.checkState(client.isOpen());
+    if (!client.isOpen()) {
+      try {
+        Awaitility.await()
+            .atMost(connectTimeoutMillis, TimeUnit.MILLISECONDS)
+            .until(() -> client.isOpen());
```
RoiEXLab:
```suggestion
            .until(client::isOpen);
```

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
DanVanAtta: If UI spawns a connection and then itself issues a 'send', we'll need to block on the 'send'. I don't think we can guarantee that we'll never have that condition. We could try to do something like have 'connect' return a lock token and then require the token to send. IMO just blocking on send is simpler.

DanVanAtta: Some quick research, we do both the 'connect' and 'send' from the EDT. Below is where we call connect (it is through the constructor of the lobby chat transmitter):

```
  public LobbyFrame(final LobbyClient lobbyClient, final ServerProperties serverProperties) {
    super("TripleA Lobby");
    setDefaultCloseOperation(JFrame.DISPOSE_ON_CLOSE);
    setIconImage(JFrameBuilder.getGameIcon());
    this.lobbyClient = lobbyClient;
    setJMenuBar(new LobbyMenu(this));
    final ChatTransmitter chatTransmitter = new LobbyChatTransmitter(lobbyClient);
    final Chat chat = new Chat(chatTransmitter);
```
 
Then the 'connect' message below is the first time we send data over websocket:
```
 public Chat(final ChatTransmitter chatTransmitter) {
    this.localPlayerName = chatTransmitter.getLocalPlayerName();
    this.chatTransmitter = chatTransmitter;
    chatTransmitter.setChatClient(this);
    sentMessagesHistory = new SentMessagesHistory();
    Optional.ofNullable(chatTransmitter.connect()).ifPresent(chatters::addAll);
    updateConnections();
  }
```

DanVanAtta:
Will follow up on non-blocking additions.

Issue #5681, opened 2019-12-06
Unable to Join Lobby (2.0.17022)


Cernelius:
Using TripleA 2.0.16996 I can join lobby.

Using TripleA 2.0.17061 I cannot join lobby.

I see the lobby, with the name I choose in it, but I cannot interact with it, and I see a window telling me "Disconnected fr...".

I'm on Windows 10.

p.s.: Lately I'm frequently experiencing cases in which a new release disables me from loading the savegames of a previous one, that much lowers my propension to update my prerelease, especially if I have any games going with other persons. Is that normal? It would be of much help if something like the second digit would be increased any time savegame compatibility is consciously broken.

DanVanAtta:
Were you getting a 403 error @Cernelius ? 

We recently fixed a problem where authorization was not being applied. That lead to cases where the auth rules were not correctly defined (now that they are being applied, this is noticeable).

> lately I'm frequently experiencing cases in which a new release disables me from loading the savegames of a previous one, that much lowers my propension to update my prerelease, especially if I have any games going with other persons. Is that normal?

yes and no. It speaks to the absolute brittleness of TripleA and how hard it is to work with. I can guarantee that 90% of those changes that caused that would have looked innocent and were not at all intended to have that effect. We would only know by firing up a save game manually. Given we've had about 50 merges in the last week, that is 50 times we would have done it, in prep of a PR there is about 3-5 rounds of testing, so we're talking a repetition of that test 150 or 250 times in just the last week.

We're in a non-compatible mode right now, which means there is no guarantee for one version to be compatible with the next. Using a prerelease is a "at your risk" activity, it's discouraged to do so for the sake of just picking up new features. Beyond other intents, it is intended for testing and feature preview.


With that said, it looks like we have a different/new issue on connecting to lobby in the latest @ b5fa8391f9dbb6f47d5fc6aca610e3dea7dcfd9d

```
SEVERE: null
java.lang.NullPointerException
	at games.strategy.engine.chat.Chat.lambda$getStatus$15(Chat.java:140)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:176)
	at java.base/java.util.HashMap$KeySpliterator.tryAdvance(HashMap.java:1624)
	at java.base/java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:127)
	at java.base/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:502)
[...]

PR #5751, opened 2019-12-24
Add config to restart http_server (spitfire) if changed

DanVanAtta:
Addresses: #5681 (Unable to Join Lobby)

We were running a stale version and had changed the server-client
data format causing errors on connection. Restarting the server
to pick up the latest version fixed the problem. The previous
deployment configuration did not restart the service, with this
update if we change the http-server jar file or server config
then we'll restart the service.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done

Tested with:
```
cd infrastructure
vagrant up
./run_ansible_vagrant -t http_server
```

Then altered http_server code, ran a gradle clean, and re-ran the above verifying the 'changed' code above was triggered.

### DIFF
--- a/infrastructure/ansible/roles/http_server/tasks/main.yml
+++ b/infrastructure/ansible/roles/http_server/tasks/main.yml
@@ -7,6 +7,7 @@
 
 - name: deploy jar file
   become: true
+  register: deploy_jar_file
   copy:
     src: "{{ http_server_jar }}"
     dest: "{{ http_server_home_folder }}/{{ http_server_jar }}"
@@ -24,6 +25,7 @@
 
 - name: deploy server config file
   become: true
+  register: deploy_config_file
   template:
     src: configuration.yml.j2
     dest: "{{ http_server_home_folder}}/configuration.yml"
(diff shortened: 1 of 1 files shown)


infrastructure/ansible/roles/http_server/tasks/main.yml
```diff
@@ -44,8 +46,17 @@
...
   become: true
   service:
     name: http_server
     state: started
     enabled: yes
+
+- name: restart service if new jar file deployed
+  when: (deploy_jar_file.changed) or (deploy_config_file.changed)
+  become: true
+  service:
```
RoiEXLab: You might want to consider using `systemd` explicitly, see https://stackoverflow.com/a/44099145

DanVanAtta: Agree it's a good modernizing move.

PR #5757, opened 2019-12-26
Non-blocking websocket send

DanVanAtta:
First, do a blocking connect on a new thread and add a timeout to it. This way
if we fail to connect we will call an error handler to notify the connection
failed. Before, if we never connected, there was never a timeout nor notification.

Second, add a send queue to websocket sender. On send, check if connection is open,
if not open, add the message to send to the send queue. On connection open,
we will flush the send queue. We have a sync block to make sure that
before we check if the connection is open, if we are flushing messages, that
we cannot queue messages. This is to avoid a race condition where after
we check if the connection is open, if it opens then and we then
flush the send queue and clear it, we want to avoid adding any new messages
to the send queue.

Change summary:
- Blocking connect on a new thread with a timeout invokes error handler
- Wire error handler to websocket sender
- Make send non-blocking by using a send queue
- Minor: add lombok getter/setter for test and code cleanup

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done
- connected to a local lobby and verified chat
- disabled local lobby websockets, connected and verified error behavior (pop-up appears, chat does not work)

## Additional Review Notes
- please pay extra attention to the two synchronization blocks added on the send queue in WebsocketConnection:

### DIFF
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
@@ -8,7 +8,6 @@ import java.util.HashSet;
 import java.util.Queue;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.logging.Level;
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
@@ -8,12 +8,13 @@ import java.util.HashSet;
 import java.util.Queue;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
+import java.util.logging.Level;
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.Setter;
+import lombok.extern.java.Log;
 import org.java_websocket.client.WebSocketClient;
 import org.java_websocket.handshake.ServerHandshake;
 import org.triplea.java.timer.ScheduledTimer;
@@ -30,6 +31,7 @@ import org.triplea.java.timer.Timers;
  *   <li>Issuing periodic keep-alive messages (ping) to keep the websocket connection open
  * </ul>
  */
+@Log
 class WebSocketConnection {
   @VisibleForTesting static final int DEFAULT_CONNECT_TIMEOUT_MILLIS = 5000;
 
@@ -115,11 +117,6 @@ class WebSocketConnection {
    * @throws IllegalStateException Thrown if connection is already open (eg: connect called twice).
    * @throws IllegalStateException Thrown if connection has been closed (ie: 'close()' was called)
    */
-  // Suppression: Ignoring future return values ignores any exceptions thrown by the code
-  // that completes the future. In this case we ignore the returned future from 'whenComplete'
-  // to avoid blocking. We would not expect the ping sender start or error handler to throw
-  // exceptions.
-  @SuppressWarnings("FutureReturnValueIgnored")
   void connect(final Consumer<String> errorHandler) {
     Preconditions.checkState(!client.isOpen());
     Preconditions.checkState(!closed);
@@ -132,22 +129,29 @@ class WebSocketConnection {
               } else {
                 errorHandler.accept("Failed to connect to: " + serverUri);
               }
+            })
+        .exceptionally(
+            throwable -> {
+              log.log(Level.SEVERE, "Unexpected exception starting socket ping sender", throwable);
+              return false;
             });
   }
 
   private CompletableFuture<Boolean> connectAsync() {
     final CompletableFuture<Boolean> completableFuture = new CompletableFuture<>();
     // execute the connection attempt
-    new Thread(() -> {
-      boolean connected;
-      try {
-        connected =
-            client.connectBlocking(DEFAULT_CONNECT_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
-      } catch (final InterruptedException ignored) {
-        connected = false;
-      }
-      completableFuture.complete(connected);
-    }).start();
+    new Thread(
+            () -> {
+              boolean connected;
+              try {
+                connected =
+                    client.connectBlocking(DEFAULT_CONNECT_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
+              } catch (final InterruptedException ignored) {
+                connected = false;
+              }
+              completableFuture.complete(connected);
+            })
+        .start();
     return completableFuture;
   }
 
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
@@ -135,25 +135,19 @@ class WebSocketConnection {
             });
   }
 
-  // Suppression: Ignoring future return values ignores any exceptions thrown by the code
-  // that completes the future. Connect blocking does not throw any exceptions, there are
-  // no exceptions to ignore, we can ignore the future returned by 'submit'.
-  @SuppressWarnings("FutureReturnValueIgnored")
   private CompletableFuture<Boolean> connectAsync() {
     final CompletableFuture<Boolean> completableFuture = new CompletableFuture<>();
     // execute the connection attempt
-    Executors.newSingleThreadExecutor()
-        .submit(
-            () -> {
-              boolean connected;
-              try {
-                connected =
-                    client.connectBlocking(DEFAULT_CONNECT_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
-              } catch (final InterruptedException ignored) {
-                connected = false;
-              }
-              return completableFuture.complete(connected);
-            });
+    new Thread(() -> {
+      boolean connected;
+      try {
+        connected =
+            client.connectBlocking(DEFAULT_CONNECT_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
+      } catch (final InterruptedException ignored) {
+        connected = false;
+      }
+      completableFuture.complete(connected);
+    }).start();
     return completableFuture;
   }
 
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
@@ -5,12 +5,15 @@ import com.google.common.base.Preconditions;
 import java.net.URI;
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.Queue;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.Setter;
-import org.awaitility.Awaitility;
-import org.awaitility.core.ConditionTimeoutException;
 import org.java_websocket.client.WebSocketClient;
 import org.java_websocket.handshake.ServerHandshake;
 import org.triplea.java.timer.ScheduledTimer;
@@ -28,15 +31,18 @@ import org.triplea.java.timer.Timers;
  * </ul>
  */
 class WebSocketConnection {
+  @VisibleForTesting static final int DEFAULT_CONNECT_TIMEOUT_MILLIS = 5000;
+
   private final Collection<WebSocketConnectionListener> listeners = new HashSet<>();
+  /**
+   * If sending messages before a connection is opened, they will be queued. When the connection is
+   * opened, the queue is flushed and messages will be sent in order.
+   */
+  private final Queue<String> queuedMessages = new ConcurrentLinkedQueue<>();
+
   private final URI serverUri;
 
   private boolean closed = false;
-
-  @Setter(
-      value = AccessLevel.PACKAGE,
-      onMethod_ = {@VisibleForTesting})
-  private int connectTimeoutMillis = 5000;
 
   @Getter(
       value = AccessLevel.PACKAGE,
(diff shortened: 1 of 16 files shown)


http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -95,53 +109,71 @@ void close() {
...
+
+  // Suppression: Ignoring future return values ignores any exceptions thrown by the code
+  // that completes the future. Connect blocking does not throw any exceptions, there are
+  // no exceptions to ignore, we can ignore the future returned by 'submit'.
+  @SuppressWarnings("FutureReturnValueIgnored")
+  private CompletableFuture<Boolean> connectAsync() {
+    final CompletableFuture<Boolean> completableFuture = new CompletableFuture<>();
+    // execute the connection attempt
+    Executors.newSingleThreadExecutor()
+        .submit(
```
RoiEXLab: In this case you might want to use the `Thread` constructor directly, the indirection over the thread executor will likely only cause overhead and a FutureReturnValueIgnored warning

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -95,53 +109,71 @@ void close() {
...
+
+    connectAsync()
+        .whenComplete(
+            (connected, throwable) -> {
+              if (connected && throwable == null) {
+                pingSender.start();
+              } else {
+                errorHandler.accept("Failed to connect to: " + serverUri);
+              }
+            });
```
RoiEXLab: Well technically we could get an error here, maybe the errorHandler throws some exception, or the ping sender has some issue.
In both cases it would just silently fail.

> We would not expect the ping sender start or error handler to throw exceptions.

I don't think exceptions are expected at any time really unless you're going for a pythonic EAFP style:
https://docs.python.org/3/glossary.html#term-eafp 🤷‍♂ 

This is an perfectly fine example to use `CompletableFutureUtils`

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -95,53 +109,71 @@ void close() {
...
+  // that completes the future. Connect blocking does not throw any exceptions, there are
+  // no exceptions to ignore, we can ignore the future returned by 'submit'.
+  @SuppressWarnings("FutureReturnValueIgnored")
+  private CompletableFuture<Boolean> connectAsync() {
+    final CompletableFuture<Boolean> completableFuture = new CompletableFuture<>();
+    // execute the connection attempt
+    Executors.newSingleThreadExecutor()
+        .submit(
+            () -> {
+              boolean connected;
```
RoiEXLab: I'm really curious why this variable can't be final.
I get that in most cases try + catch can't guarantee that a variable is only "set once", but in this case it should be safe?
Is this just a compiler oddity?

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -95,53 +109,71 @@ void close() {
...
    *
-   * @throws IllegalStateException If the connection hasn't been opened yet. {@link #connect()}
-   *     needs to be called first.
+   * @throws IllegalStateException If the connection has been closed.
    */
   void sendMessage(final String message) {
     Preconditions.checkState(!closed);
-    if (!client.isOpen()) {
-      try {
-        Awaitility.await()
```
RoiEXLab: Is the dependency now unused?

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -95,53 +109,71 @@ void close() {
...
 
-    private static final String ERROR_MESSAGE = "Error, could not connect to server at %s";
-
-    CouldNotConnect(final URI uri) {
-      super(String.format(ERROR_MESSAGE, uri));
+    // Synchronized to make sure that the connection does not open right after we check it.
+    // If the connection is not yet open, the synchronized block should guarantee that we add
+    // to the queued message queue before we start flushing it.
+    synchronized (queuedMessages) {
+      if (!client.isOpen()) {
```
RoiEXLab: I wonder if the value of `client.isOpen()` changes before or after `WebSocketClient#onOpen` is run.
Regardless there is a (really) unlikely race-condition here because the isOpen state isn't synchronized as well.

I'd recommend introducing a local field open that tracks this state independent of implementation

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
```diff
@@ -54,60 +67,142 @@ void onError() {
...
-    void connect() {
-      webSocketConnection.connect();
+    @DisplayName("Verify connect failing invokes error handler and pinger is not running")
+    void connectionFailure() throws Exception {
+      givenWebSocketConnects(false);
+
+      webSocketConnection.connect(errorHandler);
+
+      // minimal amount of sleep to allow new thread to spawn up and connection attempt to fail.
+      Thread.sleep(50L);
```
RoiEXLab: You could make the connect method return a `Future`, which would allow you to await it in this test, which avoids this ugly sleep.
Same for the other method a couple of lines below

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -95,53 +109,71 @@ void close() {
...
+        .submit(
+            () -> {
+              boolean connected;
+              try {
+                connected =
+                    client.connectBlocking(DEFAULT_CONNECT_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
+              } catch (final InterruptedException ignored) {
+                connected = false;
+              }
+              return completableFuture.complete(connected);
```
RoiEXLab: Any chance `WebSocketClient#connectBlocking` throws any other exception apart from `InterruptedException`?
I'd feel safer if we would add a `catch(Exception) {}` block that rejects the future, to prevent some user of this API to wait indefinitely

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
DanVanAtta: Indeed, compiler oddity, the assignment in the catch block flags a compiler error when made final.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
DanVanAtta: Indeed, good catch, removed in: 18b0bacfd

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
DanVanAtta: `CompletableFutureUtils` would potentially just add another `whenComplete` clause. The error handling would be the same, so we'd except a failure there just the same and would not really be adding much.

For refernece:
```
    future.whenComplete(
        (result, ex) -> {
          if (ex != null) {
            exceptionHandler.accept(ex);
          }
        });
```

It's also interesting that we have a suppress warning clause, perhaps the `whenComplete` can be transformed into an `exceptionally` clause in `CompletableFutureUtils` (in which case, maybe the helper class would not be needed at all).

e48b6d909 updates this and removes this second warning suppression.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
DanVanAtta: Good suggestion, updated: 14c24387f

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
DanVanAtta: `isOpen` is set before `onOpen` is called. For reference, the 'isOpen' value comes from `readyState` when `readyState == ReadyState.OPEN`

***WebsocketImpl.java***
```
	private void open( Handshakedata d ) {
		log.trace( "open using draft: {}",draft );
		readyState = ReadyState.OPEN;
		try {
			wsl.onWebsocketOpen( this, d );
		} catch ( RuntimeException e ) {
			wsl.onWebsocketError( this, e );
		}
	}
```

***WebsocketClient.java***
```
	@Override
	public final void onWebsocketOpen( WebSocket conn, Handshakedata handshake ) {
		startConnectionLostTimer();
		onOpen( (ServerHandshake) handshake );
		connectLatch.countDown();
	}
```

> Regardless there is a (really) unlikely race-condition here because the isOpen state isn't synchronized as well.

The purpose of the synchronization blocks is to avoid any race-condition. I'm curious where you are spotting such a situation.

If we get to the point of code where we start flushing messages, then `client.isOpen` will be true, because we have a lock at that point, we should be blocked from getting into this code block and when we do, `client.isOpen` will evaluate to true and we will not add messages to the queue.

I think I'm happy trusting the implementation on this to not change the behavior `isOpen` being toggled before `onOpen` is called. That seems like a drastic change for them, also strikes me a bit as a bit too much untrusting of the library.

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
DanVanAtta: A unit test that is 100ms faster (best case) is very valuable, with thousands of tests, 100ms is actually a really slow test. Though, we don't quite have a 100% unit test...

Anyways, this is a good suggestion and also allows us to verify the connect return result as well for a stronger test. Updated: 3100d056d

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
DanVanAtta: `WebSocketConnection` is package protected, it's used by `GenericWebsocketConnection`. Thus the API usage is controlled. I would not want/expect for the visibility of `WebsocketConnection` to be raised, it really is meant to be used through `GenericWebsocketConnection`.

Here is the code behind `connect`, please note that it already catches exceptions very aggressively (and hence we do not need to do it, would be redundant; I originally had such a catch block and after reviewing the implementation code noted that it was not necessary):
```
	public void run() {
		InputStream istream;
		try {
			boolean isNewSocket = false;
			if (socketFactory != null) {
				socket = socketFactory.createSocket();
			} else if( socket == null ) {
				socket = new Socket( proxy );
				isNewSocket = true;

			} else if( socket.isClosed() ) {
				throw new IOException();
			}

			socket.setTcpNoDelay( isTcpNoDelay() );
			socket.setReuseAddress( isReuseAddr() );

			if( !socket.isBound() ) {
				socket.connect( new InetSocketAddress( uri.getHost(), getPort() ), connectTimeout );
			}

			// if the socket is set by others we don't apply any TLS wrapper
			if (isNewSocket && "wss".equals( uri.getScheme())) {

				SSLContext sslContext = SSLContext.getInstance("TLSv1.2");
				sslContext.init(null, null, null);
				SSLSocketFactory factory = sslContext.getSocketFactory();
				socket = factory.createSocket(socket, uri.getHost(), getPort(), true);
			}

			istream = socket.getInputStream();
			ostream = socket.getOutputStream();

			sendHandshake();
		} catch ( /*IOException | SecurityException | UnresolvedAddressException | InvalidHandshakeException | ClosedByInterruptException | SocketTimeoutException */Exception e ) {
			onWebsocketError( engine, e );
			engine.closeConnection( CloseFrame.NEVER_CONNECTED, e.getMessage() );
			return;
		}

		writeThread = new Thread( new WebsocketWriteThread(this) );
		writeThread.start();

		byte[] rawbuffer = new byte[ WebSocketImpl.RCVBUF ];
		int readBytes;

		try {
			while ( !isClosing() && !isClosed() && ( readBytes = istream.read( rawbuffer ) ) != -1 ) {
				engine.decode( ByteBuffer.wrap( rawbuffer, 0, readBytes ) );
			}
			engine.eot();
		} catch ( IOException e ) {
			handleIOException(e);
		} catch ( RuntimeException e ) {
			// this catch case covers internal errors only and indicates a bug in this websocket implementation
			onError( e );
			engine.closeConnection( CloseFrame.ABNORMAL_CLOSE, e.getMessage() );
		}
		connectReadThread = null;
	}
```

DanVanAtta:
@RoiEXLab incorporated suggestions/responded, please have another look.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
RoiEXLab: The race condition:
Given 2 Threads A and B.
A is the UI (or UI triggered) Thread and B the Thread that actually handles the TCP logic.
So Thread B is performing a websocket handshake and executes until the point where the state is set to `ReadyState.OPEN`. Then the scheduler decides to pause Thread B and resumes with Thread A which is about to send/queue a message. Thread A reaches the synchronized block, checks if `client.isOpen()` which returns true at this point and the message is being sent directly (likely without an exception if I'm not missing something). Then Thread A is done and is stopped.
The scheduler then starts executing Thread B again, enters the synchronized block of the `onOpen` method, sends all existing messages and clears the queue afterwards.
Note that this probably sends all messages correctly, however possibly out of order which could in theory cause some weird issues.
That's why I suggested keeping track of the state ourselves, so we can set the state while synchronized to `queuedMessages`.
However I realized that this might not even be necessary.
If we checked `client.isOpen() && queuedMessages.isEmpty()` instead, we would avoid sending messages out of order as long as the `isOpen` state is being set before the `onOpen` method is run.
Note that in this case it is important to use `isEmpty()` over `size() == 0`, [the javadocs state](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ConcurrentLinkedQueue.html#size()):
> Beware that, unlike in most collections, this method is NOT a constant-time operation. Because of the asynchronous nature of these queues, determining the current number of elements requires an O(n) traversal. Additionally, if elements are added or removed during execution of this method, the returned result may be inaccurate. Thus, this method is typically not very useful in concurrent applications.

Whereas no such comment is found on the `isEmpty()` method.
Out of curiosity I checked the implementation and it doesn't depend on size at all, so it's constant time after all.
What you should consider, even though it might not make a difference:
You already synchronize the access to the collection, so you could easily a simple `ArrayDeque` instead of a more complex concurrent implementation to avoid the overhead.

DanVanAtta:
`isEmpty()` would need to merely check the head pointer of the list, so it makes sense it is constant time.

@RoiEXLab thanks for pointing out the race condition. Happily all messages would get sent, but out of order.

I'll follow-up with the remaining suggestions.

PR #5761, opened 2019-12-27
Remove save game folder setting

DanVanAtta:
- The setting option to specify save game folder is not terribly useful
- Check for parent folders 'f.getParentFolder()' can NPE if the save
  game folder setting is set to a non-existent folder. Fix this by using
  a simpler "f.mkdirs" call that both creates folders and avoids a
  potential NPE.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/AutoSaveFileUtils.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/AutoSaveFileUtils.java
@@ -4,8 +4,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static games.strategy.engine.framework.GameDataFileUtils.addExtension;
 import static org.triplea.java.StringUtils.capitalize;
 
-import com.google.common.annotations.VisibleForTesting;
-import games.strategy.triplea.settings.ClientSetting;
+import games.strategy.engine.framework.ui.SaveGameFileChooser;
 import java.io.File;
 import java.nio.file.Paths;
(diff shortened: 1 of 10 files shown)


DanVanAtta:
> This feature removes the ability to freely define a savegame folder

Yes, somehow I was able to update it without noticing and then saw a NPE on select game when you define a folder that DNE. The ability to custom define a savegame folder seems to have been over-eager on my point some time ago to make such settings configurable. I'm not aware of any use-case for this or any user requests for this feature.

> but makes it dependend on the configured triplea folder instead

In essence the default value becomes the only value. It would be great though to have the save game folder constant not live with the save game file chooser component. Fixing that starts going down a bigger rabbit hole though, hence the scope was limited to where it is for at least now..

panther2:
@DanVanAtta 
>Yes, somehow I was able to update it without noticing and then saw a NPE on select game when you define a folder that DNE. The ability to custom define a savegame folder seems to have been over-eager on my point some time ago to make such settings configurable. I'm not aware of any use-case for this or any user requests for this feature.

Ok, then let me rise awareness. I have always custom defined a savegame folder, as I hate to be determined by the operating system when saving my personal files. I have different partitions for OS and programs on the one hand and my user-files, including my save games on the other hand.

PR #5762, opened 2019-12-27
WebSocketConnection concurrency updates

DanVanAtta:
- Use non-concurrent linked list implementation, since it is accessed under a lock, access
  is effectively single-threaded and we can use a non-concurrent linked list implementatoin.
- Use boolean tracking variable for connection is open check. State of variable is changed
  only under synchronization. This is to avoid potential race conditions between the point
  when the websocket state is marked as open and when 'onOpen' is called. Under such a
  a condition, messages could be sent out of order.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
@@ -3,6 +3,7 @@ package org.triplea.http.client.web.socket;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import java.net.URI;
+import java.util.ArrayDeque;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.LinkedList;
@@ -37,13 +38,13 @@ class WebSocketConnection {
    * If sending messages before a connection is opened, they will be queued. When the connection is
    * opened, the queue is flushed and messages will be sent in order.
    */
-  private final Queue<String> queuedMessages = new LinkedList<>();
+  private final Queue<String> queuedMessages = new ArrayDeque<>();
 
   /**
    * State variable to track open connection, this value is set and checked under the same
    * synchronization lock.
    */
-  private volatile boolean connectionIsOpen = false;
+  private boolean connectionIsOpen = false;
 
   private final URI serverUri;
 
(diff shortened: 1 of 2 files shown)


http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -37,7 +37,13 @@
    * If sending messages before a connection is opened, they will be queued. When the connection is
    * opened, the queue is flushed and messages will be sent in order.
    */
-  private final Queue<String> queuedMessages = new ConcurrentLinkedQueue<>();
+  private final Queue<String> queuedMessages = new LinkedList<>();
```
DanVanAtta: Went with linked list, constant time append and not having to pre-allocate space are positives.

RoiEXLab: Error-prone will likely fail.
Even though it's faster in theory, modern CPUs are just better at allocating blocks instead of random memory

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -37,7 +37,13 @@
...
+
+  /**
+   * State variable to track open connection, this value is set and checked under the same
+   * synchronization lock.
+   */
+  private volatile boolean connectionIsOpen = false;
```
RoiEXLab: Not even necessary to make this volatile, also only accessed with mutex.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
DanVanAtta: Sure enough:

> /home/travis/build/triplea-game/triplea/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java:40: error: [JdkObsolete] It is very rare for LinkedList to out-perform ArrayList or ArrayDeque. Avoid it unless you're willing to invest a lot of time into benchmarking. Caveat: LinkedList supports null elements, but ArrayDeque does not.

PR #5780, opened 2019-12-28
Fix single element arrays

RoiEXLab:
Stumbled across this when checking the code for #5010

I'd recommend having a look at the commits one by one.
Should make things clearer

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/AaInMoveUtil.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/AaInMoveUtil.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


DanVanAtta:
Looks good, though curious why AtomicReference and not just have a reference?

RoiEXLab:
@DanVanAtta Well you can't set outer local variables from lambdas, because those might not exist anymore when the lambda is run.
In case you're reffering to the [Reference class](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ref/Reference.html),  that's just an abstract class for memory optimizations.

DanVanAtta:
Ah, makes sense, thanks.

PR #5782, opened 2019-12-28
Enable test now that authentication is enforced

DanVanAtta:
- Also, fix a precondition check bug in protected endpoint test.


DanVanAtta:
Important to note, nested test classes do not work properly with dbunit.

PR #5784, opened 2019-12-28
Replaces usages of 'collections.' with java9 API

DanVanAtta:
Replaces usages of 'collections.' with java9 API,
EG: Collections.emptyMap() -> Map.of()


DanVanAtta:
PointFileReaderWriterTest presented a roadblock to fixing all 'legacy' collection usages. The test seems pretty brittle, possibly relying on map ordering, beyond that it's not clear why the test was supposed to pass and asserts the expectations that it does. I removed the updates to that test class, that is the last class before we can set 'use java9 collection API' to enforcing.

http-server/src/main/java/org/triplea/server/http/web/socket/MessageBroadcaster.java
```diff
@@ -20,7 +21,8 @@
   public void accept(
       final Collection<Session> sessions, final ServerMessageEnvelope serverEventEnvelope) {
     log.info("Broadcasting: {}", serverEventEnvelope);
-    sessions
+
+    ImmutableSet.copyOf(sessions)
```
RoiEXLab: Why is this copy required?

game-core/src/main/java/games/strategy/triplea/delegate/MoveValidator.java
```diff
@@ -286,10 +285,10 @@ private static String validateCanal(
     // Check each unit 1 by 1 to see if they can move through necessary canals on route
     String result = null;
     final Set<Unit> unitsThatFailCanal = new HashSet<>();
+    final Set<Unit> setWithNull = new HashSet<>();
+    setWithNull.add(null);
```
RoiEXLab: Not to sound mean or something, but this seems like a prime example to my why the "old" methods still have their use.
So unless you find a way to avoid null usage in every place like this (why do we even need a singleton, could an empty set work as well?) I don't think this is a step forward

RoiEXLab: Also hashset is mutable, whereas the singleton is not.

game-core/src/test/java/games/strategy/engine/data/properties/MapPropertyTest.java
```diff
@@ -23,27 +23,32 @@
...
     }
 
     @Test
     void shouldReturnTrueWhenValueIsMapAndKeysAreCompatibleAndValuesAreCompatible() {
-      assertThat(mapProperty.validate(ImmutableMap.of(KEY_1, 11, KEY_2, 22)), is(true));
-      assertThat(mapProperty.validate(Collections.singletonMap(null, 33)), is(true));
+      assertThat(mapProperty.validate(Map.of(KEY_1, 11, KEY_2, 22)), is(true));
+      final Map<String, Integer> map = new HashMap<>();
+      map.put(null, 33);
+      assertThat(mapProperty.validate(map), is(true));
```
RoiEXLab: Alternative approach:
The script does allow the old methods if null is hardcoded

http-server/src/main/java/org/triplea/server/http/web/socket/MessageBroadcaster.java
DanVanAtta: Hit a concurrent modification error.

game-core/src/main/java/games/strategy/triplea/delegate/MoveValidator.java
DanVanAtta: The mutability of collections depends on the contents. Singleton is unmodifiable, but it's not necessarily immutable. Favoring immutability is much more about removing setters.

DanVanAtta: To this example, Unit is mutable, the singleton collection wrapper around it does not change that.

game-core/src/test/java/games/strategy/engine/data/properties/MapPropertyTest.java
DanVanAtta: Maybe, the null hardcoded would lead to false positives. Please note the script is not yet enabled. The test for PointFileReaderWriter has major problems, I'll need your help likely to understand why the test is expected to pass to begin with and seems to depend on non-deterministic ordering.

`null` in a collection is a smell, I suspect the actual code smell here is a 'data clump' code smell. Fixing that IMO is the right fix, not hacking the script.

DanVanAtta:
@RoiEXLab , I hope we can agree that the syntax change is not important:
- adding a null to a list is something we do not want. If doing so, then not being able to use syntactic sugar to make it easy seems fine.
- The fix for not adding null is a larger scope, and something that should perhaps be fixed

Mutability:
- Modifiability of collections is not making things immutable. I don't think we're super awesome about mutability, lots of legacy code. We should be enforcing that all variables be declared final and all data types not have setters to really get there. For better or worse we're at a best-effort approach, which is a bit problematic.

Convention:
- The conventions are meant to be much more hard and fast rules than not. If we say, "we have the convention, but at the first inconvenience, we can abandon it", then there's no point to the convention. We're going to argue just the same in PR about doing what we want anyways.

Perfect code:
- Sometimes you have to take a step back before you can move forward. 

Contracts and scope:
- The modifiable collection is being injected as part of test, the code under test does not require an unmodifiable collection, nor should it.
- The lack of sugary syntax is very limited, we're talking about just a couple extra lines in test code. Is there any real importance here? I think in conclusion respecting the convention, even if it has the effect that dirty code becomes even dirtier, is not necessarily a bad thing. To some extent, it's part of the point, if you want to put nulls in collections, we don't make it easy.

game-core/src/test/java/games/strategy/engine/data/properties/MapPropertyTest.java
DanVanAtta: It'd be easier to instead to define an annotation that skips custom style checks. If we do so, then I'm concerned the conventions become toothless, in which case what was the point of defining them?

DanVanAtta: IE: conventions == "do this unless you really can't", vs "Please do this if it's convenient and something you already agree with". Conventions are meant to make people do things that they don't necessarily agree with for consistency of the project. Not everyone agrees with 2 space indents or line length. As Ron mentioned, better to be consistently bad vs inconsistently good.

DanVanAtta:
@RoiEXLab bump, please respond.

RoiEXLab:
I agree that null inside a collection is most likely a smell, however I really do appreciate the syntactic sugar that the singleton/immutable map/set etc. methods provide even for null values and I'd say that avoiding this sugar is clearly a step back.
But in the end we might want to avoid null values, so I agree having the "ugly syntax" might be a good motivation to fix those places.
One thing I wonder: Would it make sense to create a class with methods that reproduce the old behaviour, just so that we can mark it deprecated and the IDE highlights its usage? It would also allow us to find all usages over find usages functionality

DanVanAtta:
@RoiEXLab you're still requesting changes here, what are those? I would agree that not using the nicest sugar syntax is a step back, though:
- it violates convention
- it is done in order to allow code we do not want to see and is a larger problem than syntax

PR #5786, opened 2019-12-30
Websocket ban

DanVanAtta:
## Short summary:
- Client-side: Show reason code in connection failed error messages
- Client-side: Unify connection lost and connection closed listener
  by logging at a low level.
- Client-side: fix connection closed message not appearing
- Server-side: kill active websocket connections of banned players 
  when they are banned.

### Close active websocket connections of banned players

Request filtering will block banned players from creating new
connections, as of yet there was no mechanism to disconnect
websocket sessions of banned players. This update adds a new
class called 'SessionSet' that will listen to banned player
events, and will disconnect any sessions for that player.

Each server side websocket has an event queue or a handler
object to process websocket requets. To implement this
feature, we create a 'SessionSet' for each one, inject
the 'SessionSet' into each handler, and we register
each session set to listen to banned events.

###  Improve http client error handling

Detailed summary:
- Improve error message display when unable to connect to lobby.
  Show the http reason code, an http entity body is not consistent
  on error responses, we can instead place connection error information
  in the status code reason phrase

- Improve detetion of connection closed vs connection lost. The
  'remote' boolean can be used to detect at a low level if
  the server closed the connection or if the client. If the server
  closed the connection, then we can log (which triggers a popup)
  an error. With the decision to do logging at a low level, connection
  closed listener and connection lost listener become unified and
  no longer need a 'string' error message and become simple 'Runnables'

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### Before
Screenshot from 2019-12-29 19-19-09

### After

Screenshot from 2019-12-29 19-17-43

When connected and banned, this error message pops-up, before the lobby window would just close:
Screenshot from 2019-12-29 16-28-14

## Additional Review Notes
- commits are broken up by server side and client side updates. They are mostly needed together for the testing to work out. A third commit is mostly temporary until a race condition with listeners can be fixed.


DanVanAtta:
Hmm, well the hack fix of course is not quite enough, will need to resolve the smoke-test error.  In the meantime review can get a head start as the bulk of changes are stable. It's odd we are seeing a race condition now with listeners being added to chat client and had not before.

http-server/src/main/java/org/triplea/server/http/web/socket/SessionSet.java
```diff
@@ -0,0 +1,80 @@
...
+  @Getter(value = AccessLevel.PACKAGE, onMethod_ = @VisibleForTesting)
+  private final Set<Session> sessions = new HashSet<>();
+
+  public void put(final Session session) {
+    sessions.add(session);
+  }
+
+  public Collection<Session> values() {
+    removeClosedSessions();
+    return ImmutableSet.copyOf(sessions);
```
RoiEXLab: You could use `Set#copyOf` here

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListingEventQueue.java
```diff
@@ -25,36 +21,19 @@
   private final BiConsumer<Collection<Session>, ServerMessageEnvelope> broadcaster;
 
   @Getter(value = AccessLevel.PACKAGE, onMethod_ = @VisibleForTesting)
-  private final Map<String, Session> sessions = new HashMap<>();
+  private final SessionSet sessionSet;
```
RoiEXLab: As mentioned in #5802 we need a way to have a concurrent SessionSet

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListingEventQueue.java
```diff
@@ -25,36 +21,19 @@
...
 
   void addListener(final Session session) {
     Preconditions.checkNotNull(session);
     Preconditions.checkNotNull(session.getId());
```
RoiEXLab: Now that the ID is no longer required, we don't need it to be non-null

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListingWebsocket.java
```diff
@@ -25,11 +23,6 @@ private GameListingEventQueue getEventQueue(final Session session) {
         (GameListingEventQueue) session.getUserProperties().get(GAME_LISTING_EVENT_QUEUE));
   }
 
-  @OnClose
-  public void close(final Session session, final CloseReason closeReason) {
-    getEventQueue(session).removeListener(session.getId());
-  }
-
```
RoiEXLab: Just so I get this correctly: Starting with this update we lazy-remove any closed sessions?

http-server/src/main/java/org/triplea/server/remote/actions/RemoteActionsWebSocket.java
```diff
@@ -29,11 +27,6 @@ private RemoteActionsEventQueue getEventQueue(final Session session) {
         (RemoteActionsEventQueue) session.getUserProperties().get(ACTIONS_QUEUE_KEY));
   }
 
-  @OnClose
-  public void close(final Session session, final CloseReason closeReason) {
-    getEventQueue(session).removeSession(session);
-  }
-
   /**
```
RoiEXLab: Also mentioned in #5802
This class and GameListingWebsocket are almost identical

http-server/src/test/java/org/triplea/server/access/BannedPlayerFilterTest.java
```diff
@@ -104,8 +104,9 @@ private void verifyForbiddenRequest() {
 
       final Response response = responseCaptor.getValue();
       assertThat(response.getStatus(), is(Status.UNAUTHORIZED.getStatusCode()));
-      assertThat((String) response.getEntity(), StringContains.containsString("5 minutes"));
-      assertThat((String) response.getEntity(), StringContains.containsString(BAN_ID));
+      assertThat(
+          response.getStatusInfo().getReasonPhrase(), StringContains.containsString("5 minutes"));
+      assertThat(response.getStatusInfo().getReasonPhrase(), StringContains.containsString(BAN_ID));
```
RoiEXLab: Usually we statically import those methods, don't we?

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListingWebsocket.java
RoiEXLab: Wouldn't this possibly result in a ghost entry for an extended amount of time until another change request is made?

http-server/src/main/java/org/triplea/server/http/ServerApplication.java
```diff
@@ -141,8 +144,25 @@ public void run(final AppConfig configuration, final Environment environment) {
...
-    final var gameListingEventQueue = GameListingEventQueueFactory.newGameListingEventQueue();
+
+    final SessionSet remoteActionSessions = new SessionSet();
+    final SessionSet gameListingSessions = new SessionSet();
+    final SessionSet chatSessions = new SessionSet();
+    final BannedPlayerEventHandler bannedPlayerEventHandler =
+        BannedPlayerEventHandler.builder()
+            .sessionSets(
+                Set.of(
+                    remoteActionSessions, //
```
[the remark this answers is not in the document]
DanVanAtta: yes, it's there to make the formatter to keep the formatting to one-per-line and not same line.

http-server/src/test/java/org/triplea/server/access/BannedPlayerFilterTest.java
DanVanAtta: Updated: 34d959a6a

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListingEventQueue.java
DanVanAtta: Removed: 52b859842

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListingEventQueue.java
DanVanAtta: Use concurrent set: c664a7f5e

Additional thread safety updates: aa8f957bc (message broadcaster receives other collections from more locations than here)

http-server/src/main/java/org/triplea/server/http/web/socket/SessionSet.java
DanVanAtta: Updated: 6e198db64

http-server/src/main/java/org/triplea/server/http/web/socket/SessionSet.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /** Just in case we fail to remove a closed session from listeners, remove any closed sessions. */
+  private void removeClosedSessions() {
+    // find all sessions that are not open
+    final Set<Session> closedSessions =
+        sessions.stream().filter(Predicate.not(Session::isOpen)).collect(Collectors.toSet());
+    // remove each of the closed sessions
+    closedSessions.forEach(sessions::remove);
```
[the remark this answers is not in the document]
DanVanAtta: Updated: 639874b08

http-server/src/main/java/org/triplea/server/remote/actions/RemoteActionsWebSocket.java
DanVanAtta: Yes, I was looking for a way to share the annotations and pass an implementation, but the annotation scanner does not pick up annotated methods from a parent class. We could consider merging all websockets together into just one based on connection source. If that seems like a good idea, we can  continue that conversation in an issue and follow-up.

RoiEXLab: Yeah, I have no problem with that

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListingWebsocket.java
RoiEXLab: Bump

http-server/src/main/java/org/triplea/server/http/web/socket/SessionSet.java
```diff
@@ -0,0 +1,76 @@
...
+    // remove any closed sessions
+    sessions.removeIf(Predicate.not(Session::isOpen));
+  }
+
+  public Collection<Session> getSessionsByIp(final InetAddress serverIp) {
+    removeClosedSessions();
+    return sessions.stream()
+        .filter(s -> InetExtractor.extract(s.getUserProperties()).equals(serverIp))
+        .filter(Session::isOpen)
+        .collect(Collectors.toSet());
```
RoiEXLab: Any thoughts about returning `Stream<Session>` here?
The one non-test usage performs a for-each on this anyways, so that wouldn't be a problem.
As a nice bonus, `closeSessionsByIp` could re-use the logic

http-server/src/main/java/org/triplea/server/http/web/socket/SessionSet.java
```diff
@@ -0,0 +1,76 @@
...
+  public void closeSessionsByIp(final InetAddress ip) {
+    sessions.stream()
+        .filter(Session::isOpen)
+        .filter(s -> InetExtractor.extract(s.getUserProperties()).equals(ip))
+        .forEach(
+            s -> {
+              try {
+                s.close();
+              } catch (final IOException e) {
+                log.warn("Could not close session (session is still open? {}", s.isOpen(), e);
```
RoiEXLab:
```suggestion
                log.warn("Could not close session (session is still open? {})", s.isOpen(), e);
```

http-server/src/main/java/org/triplea/server/http/web/socket/SessionSet.java
```diff
@@ -0,0 +1,76 @@
...
+ * <ol>
+ *   This class has two main purposes:
+ *   <li>Unify common session tracking code
+ *   <li>Provide a mechanism to close already established connections that have been banned. We will
+ *       rely on request filtering to prevent banned sessions from creating a new websocket
+ *       connection, this part is needed to remove existing connections of a banned user.
+ * </ol>
+ */
+@Slf4j
+public class SessionSet {
```
RoiEXLab: You could add a `@ThreadSafe` annotation to this class as some sort of documentation

http-server/src/test/java/org/triplea/server/lobby/chat/event/processing/ChatEventProcessorTest.java
```diff
@@ -46,10 +47,15 @@
...
 
-  private final ChatEventProcessor chatEventProcessor = new ChatEventProcessor(new Chatters());
+  private final ChatEventProcessor chatEventProcessor =
+      new ChatEventProcessor(new Chatters(), new SessionSet());
 
   @Mock private Chatters chatters;
 
+  @SuppressWarnings("unused")
+  @Mock
+  private SessionSet sessionSet;
```
RoiEXLab: Why is this unused?

http-server/src/test/java/org/triplea/server/lobby/game/listing/GameListingEventQueueTest.java
```diff
@@ -140,7 +124,7 @@ void gameUpdated() {
                 GameListingMessageType.GAME_UPDATED.name(), lobbyGameListing)));
     assertThat(
         "Verify closed session is removed",
-        gameListingEventQueue.getSessions().values(),
+        gameListingEventQueue.getSessionSet().values(),
         hasSize(1));
   }
 }
```
RoiEXLab: Slightly out-of-scope, but could you re-do the unchecked warning suppression from https://github.com/triplea-game/triplea/pull/5802/files#diff-748e01ab941a190e4392897c626bf28aR72-R116 ?

http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListingWebsocket.java
DanVanAtta: Turns out the 'close' is called whether client or server closes. If we only do a lazy close, then it gets weird in test with just one connection (nothing closes), it also looked like chatter is less reliable to drop. Went with instead to use 'onClose' to close connections and no lazy close at all. IMO important to do the close in one place (DRY)

DanVanAtta:
PR is back again to original failure with listeners race condition, will try to get it figured out and respond to latest CR feedback.

http-server/src/test/java/org/triplea/server/lobby/game/listing/GameListingEventQueueTest.java
RoiEXLab: Bump

http-clients/src/main/java/org/triplea/http/client/lobby/chat/LobbyChatClient.java
```diff
@@ -41,37 +41,37 @@ public static LobbyChatClient newClient(
...
 
   public void setChatMessageListeners(final ChatMessageListeners chatMessageListeners) {
-    setListeners(chatMessageListeners);
+    new WebsocketListenerBinding<>(webSocketClient, chatMessageListeners) {
+      @Override
+      protected ChatServerMessageType readMessageType(
+          final ServerMessageEnvelope serverMessageEnvelope) {
+        return ChatServerMessageType.valueOf(serverMessageEnvelope.getMessageType());
+      }
+    };
```
RoiEXLab: If you don't know that the constructor has side-effects and registers itself as part of the constructor, this line looks very weird.
I think you should be able to move the registration thingy to a seperate method to make this more explicit

Cernelius:
"Banned 1 hours"? Is that communicating the duration of the ban itself? I think it would be helpful the most, for the banned user, exactly to know the remaining time the ban is going to last, but having both the info about the original duration of the ban and how much of it is currently left would be the best.

Also probably off topic, but I hope it will be very clear for the user whether the ban is general or just name-related (something like "please, change your user name and rejoin").

Personal preference, I would not add the "for violating lobby rules" part.

http-clients/src/main/java/org/triplea/http/client/lobby/chat/LobbyChatClient.java
DanVanAtta: The constructor having side-effects becomes obvious unless someone believes this is dead-code. Having such a construct is very odd though, I would agree.  

I think this would get fixed soon: https://github.com/triplea-game/triplea/issues/5845; fixing it now would more likely result in code-churn and would be quite piece-meal.

DanVanAtta:
@Cernelius 
> "Banned 1 hours"? Is that communicating the duration of the ban itself?


Yes


>  I think it would be helpful the most, for the banned user, exactly to know the remaining time the ban is going to last, but having both the info about the original duration of the ban and how much of it is currently left would be the best.

While maybe helpful, IMO it's being too nice to a banned user. They are lucky they are getting let back into the game at all. Letting them know the original ban duration is not strictly needed, it's arguably nice to know, but really it's not that important to know (it's maybe more important for moderators, but we'll have that information in the banned-users table).


> Also probably off topic, but I hope it will be very clear for the user whether the ban is general or just name-related (something like "please, change your user name and rejoin").

This is something that should probably be checked as part of the next release QA (which is in theory imminient and/or happening now). 


> Personal preference, I would not add the "for violating lobby rules" part.

It is to distinguish between a disconnect and a ban. A disconnect can be done for any reason, a ban is ultimately for violating lobby rules. We could add more fine-grained reasons, for now it's easier to make the blanket statement which should always be true. I also very much like the emphasis that there are actually written lobby rules out there, my impression it's only moderators that have actually read these. A code of conduct is worthless if it's a secret code, the actions from moderators appear arbitrary otherwise even if following a code.

DanVanAtta:
@Cernelius good points/questions, though if there is more follow-up and/or discussion, as most of the commentary is about already updated code, that conversation would hijack this thread, best to open those questions in a new issue.

PR #2924, opened 2018-01-30
PointFileReaderWriter: Require caller to close streams

ssoloff:
Companion to #2889 but for the `writeXXX()` methods.

`PointFileReaderWriter` uses a resource-management idiom that is different from the rest of our code: it takes ownership of the `OutputStream` passed to its methods and closes the stream before they return.  This can lead to confusion because the calling code appears to leak the stream it creates.  For consistency, this PR modifies the code such that the code that creates the stream is responsible for closing it.

#### Functional changes

None.

#### Refactoring changes

* `PointFileReaderWriter` methods no longer close their `OutputStream` parameter.
* Handling a `null` mapping in `writeOneToMany()` was never realized by any of the current callers, so it was replaced with a `checkNotNull` precondition.
* The signature of `writeOneToMany()` was changed to use a `mapping` parameter of type `Map<String, List<Point>>` instead of `Map<String, Collection<Point>>` for symmetry with `readOneToMany()` (as well as consistency with all other read/write methods in this class).  This allowed the removal of a few unnecessary copies in callers that were created simply to workaround the invariance of the generic collections.
* Replaced several concrete types with interface types where possible (e.g. `HashMap` -> `Map`).

#### Testing

Several characterization tests were added to detect regressions in `PointFileReaderWriter`.

#### Notes

Please view with whitespace changes ignored (`?w=1`).

PR #5801, opened 2019-12-31
Convert PointFileReaderWriterTest to use Java9 Collection APIs

DanVanAtta:
Convert PointFileReaderWriterTest to use Java9 Collection APIs
    
EG: "Collections.emptyMap" -> "Map.of"
    
J9 API usage conforms to TripleA decided team coding convention.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/game-core/src/test/java/org/triplea/util/PointFileReaderWriterTest.java
+++ b/game-core/src/test/java/org/triplea/util/PointFileReaderWriterTest.java
@@ -4,7 +4,6 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.aMapWithSize;
 import static org.hamcrest.Matchers.hasKey;
 import static org.hamcrest.Matchers.is;
-import static org.hamcrest.core.StringContains.containsString;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
@@ -490,6 +489,7 @@ final class PointFileReaderWriterTest {
     @Test
     void shouldWriteOnePointPerName() throws Exception {
       final Map<String, Point> pointsByName =
+          // ImmutableMap#of guarantees iteration ordering
           ImmutableMap.of(
               "United Kingdom", new Point(1011, 1021),
               "Germany", new Point(2011, 2021),
@@ -498,9 +498,13 @@ final class PointFileReaderWriterTest {
       final String content =
           writeToString(os -> PointFileReaderWriter.writeOneToOne(os, pointsByName));
 
-      assertThat(content, containsString("United Kingdom  (1011,1021)"));
-      assertThat(content, containsString("Germany  (2011,2021)"));
-      assertThat(content, containsString("Eastern United States  (3011,3021)"));
+      assertThat(
+          content,
+          is(
+              ""
+                  + "United Kingdom  (1011,1021) \r\n"
+                  + "Germany  (2011,2021) \r\n"
+                  + "Eastern United States  (3011,3021) "));
     }
   }
 
@@ -518,7 +522,8 @@ final class PointFileReaderWriterTest {
     @Test
     void shouldWriteMultiplePointsPerName() throws Exception {
       final Map<String, List<Point>> pointListsByName =
-          Map.of(
+          // ImmutableMap#of guarantees iteration ordering
+          ImmutableMap.of(
               "Belarus",
                   List.of(new Point(1011, 1021), new Point(1012, 1022), new Point(1013, 1023)),
               "54 Sea Zone", List.of(new Point(2011, 2021), new Point(2012, 2022)),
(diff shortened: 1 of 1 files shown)


game-core/src/test/java/org/triplea/util/PointFileReaderWriterTest.java
```diff
@@ -505,13 +498,9 @@ void shouldWriteOnePointPerName() throws Exception {
...
-      assertThat(
-          content,
-          is(
-              ""
-                  + "United Kingdom  (1011,1021) \r\n"
-                  + "Germany  (2011,2021) \r\n"
-                  + "Eastern United States  (3011,3021) "));
+      assertThat(content, containsString("United Kingdom  (1011,1021)"));
+      assertThat(content, containsString("Germany  (2011,2021)"));
+      assertThat(content, containsString("Eastern United States  (3011,3021)"));
```
RoiEXLab: Line endings are somewhat important to check.
If everything is in the same line that's a syntax error

game-core/src/test/java/org/triplea/util/PointFileReaderWriterTest.java
```diff
@@ -600,39 +585,41 @@ void shouldWriteMultiplePlacementsPerName() throws Exception {
...
-                  + "Belarus  <  (1011,1021)  (1012,1022)  (1013,1023)  > \r\n"
-                  + "54 Sea Zone  <  (2011,2021)  (2012,2022)  (2013,2023)  > "
-                  + " <  (2111,2121)  (2112,2122)  (2113,2123)  > \r\n"
-                  + "Philippines  <  (3011,3021)  (3012,3022)  (3013,3023)  > "
+          containsString(
+              "54 Sea Zone  <  (2011,2021)  (2012,2022)  (2013,2023)  > "
+                  + " <  (2111,2121)  (2112,2122)  (2113,2123)  > "));
+      assertThat(
+          content,
+          containsString(
```
RoiEXLab: Same here with line endings

game-core/src/test/java/org/triplea/util/PointFileReaderWriterTest.java
DanVanAtta: The ordering of the three lines is non-deterministic. It was actually a major flaw in the test previously that a non-deterministic ordering was being relied upon. We could assert that we see 2 counts of the line terminator, though it may not mean much as it could appear anywhere in the string.

RoiEXLab: Hmm we could change the code to generate files with unix-style line endings, then we could check the line endings for every line

ssoloff: > It was actually a major flaw in the test previously that a non-deterministic ordering was being relied upon.

I believe this test did not previously rely on a non-deterministic ordering.  Guava's [`ImmutableMap#of()`](https://guava.dev/releases/26.0-jre/api/docs/com/google/common/collect/ImmutableMap.html#of-K-V-K-V-K-V-K-V-K-V-) factory methods guarantee that the returned map entries are iterated in the same order as the arguments.  Whereas, the Java `Map#of()` factory methods do not make the same guarantee.

DanVanAtta: I did not know that about `ImmutableMap#of`. I'm still concerned:
- the API takes a `Map`, typically the ordering is not expected to be guaranteed
- the test code assigned to a `Map`, we're relying on the contract of a specific implementation which as was the case here, is error prone as the implementation was changed and caused a breakage
- the test assumes a specific implementation, notably that the elements would be processed in-order, that does not seem to be required for correctness
- relying on specific nuances of Guava seems to increase barrier to entry

Are there any ideas of how to avoid the guava immutable map here? I don't think the test coverage for the line endings is actually that profound, we're testing the `collect` method on streams:

```
        mapping.entrySet().stream()
            .map(entry -> entry.getKey() + " " + pointToString(entry.getValue()))
            .collect(Collectors.joining("\r\n")),
```

ssoloff: > Are there any ideas of how to avoid the guava immutable map here?

I guess you could change the implementation to be deterministic instead of the test.  That is, sort the stream by key and simply document that fact in the Javadocs.  I'm not aware of any downstream consumer that would be affected by that change.

IIRC, the reason the above proposal wasn't implemented at the time the original tests were written is because the entire set of `PointFileReaderWriter` tests were added as a single group of characterization tests prior to doing some refactoring (see #2889 and #2924).  So, by definition, the implementation wasn't supposed to be changed before the tests were complete.  The idea to go back and make the implementation deterministic instead of the test was probably just lost in the follow-up noise.

game-core/src/test/java/org/triplea/util/PointFileReaderWriterTest.java
DanVanAtta: To be pedantic, there is a 3rd example as well. I don't think there is a need to point out every example once a general point has been made and it's clear that it's a general point that applies to every example that could be found.

game-core/src/test/java/org/triplea/util/PointFileReaderWriterTest.java
DanVanAtta: Certainly a number of issues present here, for the sake of moving forward, went back to `ImmutableMap#of` with: df0e2ec58

PR #5802, opened 2019-12-31
Session concurrency fix?

RoiEXLab:
Alternative approach to #5793 
I'd recommend reviewing the commits one by one, because I'm not 100% sure the third commit is safe.

## Functional Changes
[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  #5793
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListingEventQueue.java
+++ b/http-server/src/main/java/org/triplea/server/lobby/game/listing/GameListingEventQueue.java
@@ -3,6 +3,7 @@ package org.triplea.server.lobby.game.listing;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.function.BiConsumer;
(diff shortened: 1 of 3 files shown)


RoiEXLab:
CodeClimate has a point, the 2 classes are almost identical

DanVanAtta:
Really appreciate the proof of concept here. This would conflict pretty badly with: https://github.com/triplea-game/triplea/pull/5786

Also please check out `SessionSet` from that same PR, it's almost the same kind of change as here, but it incorporates more redundant code.

DanVanAtta:
@RoiEXLab I'm not sure this is complete altternative to: https://github.com/triplea-game/triplea/pull/5793 (the change here is good, but I'd probably still want #5793 nonetheless).

The API in #5793 is `Collection`, the copy is done so that the method itself enforces the contract it needs for correctness. This reduces coupling as the method no longer requires to be passed a thread-safe collection, it establishes (at least this element of) thread-safety on its own.

RoiEXLab:
> I'd probably still want #5793 nonetheless

When copying the collection (using ImmutableSet.of or similar functions) you end up iterating over the whole collection. If we really have concurrency problems, this iteration process could end up failing as well. The only difference being a smaller time window.
So the only difference this would make is in case the iteration itself would change the underlying collection like this:
```java
final var list = new ArrayList<>(List.of("a", "b", "c"));
for (final String s : list) {
  if (s.equals("a")) {
    // throws ConcurrentModificationException
    list.remove(0);
  }
}
```
whereas
```java
final var list = new ArrayList<>(List.of("a", "b", "c"));
for (final String s : List.copyOf(list)) {
  if (s.equals("a")) {
    // works fine
    list.remove(0);
  }
}
```

So unless you think a 'MessageSender' implementation will ever have to mess around with the actual implementation, #5793 is probably just unecessary overhead

RoiEXLab:
Part of #5786

PR #5808, opened 2020-01-01
Fix freeze with retreat dialog

asvitkine:
Fix freeze with retreat dialog. This was caused by the code allowing multiple retreat dialogs to be shown - when clicking the action button multiple times, and the retreat dialog from a previous battle blocking new UI thread code (for the next battle) from running.

Also addresses a similar issue with the select casualties dialog. Previously, you could click the button multiple times to get multiple dialogs shown. When a dialog is still shown when the battle ends, it would also result in a freeze. With this PR, only one dialog can be shown at a time (via making the button disabled while it's shown).

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[X] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix:  https://github.com/triplea-game/triplea/issues/5010
[] Other:   

## Testing

[X] Manual testing done

Tested that it's not possible to bring up multiple retreat dialogs by clicking the action button multiple times. Also tested that without this change, the problem was reproducible when having multiple retreat dialogs open and ending battle with not all of them having been closed (it would freeze).

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
@@ -392,11 +392,7 @@ public class BattleDisplay extends JPanel {
             options,
             cancel);
     // dialog dismissed
-    if (choice == JOptionPane.CLOSED_OPTION) {
-      return false;
-    }
-    // wait
-    if (choice == JOptionPane.CANCEL_OPTION) {
+    if (choice == JOptionPane.CLOSED_OPTION || choice == JOptionPane.CANCEL_OPTION) {
       return false;
     }
     // remain
@@ -457,19 +453,14 @@ public class BattleDisplay extends JPanel {
             options,
             no);
     // dialog dismissed
-    if (choice == JOptionPane.CLOSED_OPTION) {
-      return false;
-    }
-    // wait
-    if (choice == JOptionPane.CANCEL_OPTION) {
+    if (choice == JOptionPane.CLOSED_OPTION || choice == JOptionPane.CANCEL_OPTION) {
       return false;
     }
     // remain
     if (choice == JOptionPane.NO_OPTION) {
       return true;
     }
-    // if you have eliminated the impossible, whatever remains, no matter how improbable,
-    // must be the truth
+
     // retreat
     if (possible.size() == 1) {
       retreatTo.set(possible.iterator().next());
--- a/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
@@ -392,15 +392,15 @@ public class BattleDisplay extends JPanel {
             options,
             cancel);
     // dialog dismissed
-    if (choice == -1) {
+    if (choice == JOptionPane.CLOSED_OPTION) {
       return false;
     }
     // wait
-    if (choice == 2) {
+    if (choice == JOptionPane.CANCEL_OPTION) {
       return false;
     }
     // remain
-    if (choice == 1) {
+    if (choice == JOptionPane.NO_OPTION) {
       return true;
     }
     // submerge
@@ -457,7 +457,7 @@ public class BattleDisplay extends JPanel {
             options,
             no);
     // dialog dismissed
-    if (choice == -1) {
+    if (choice == JOptionPane.CLOSED_OPTION) {
       return false;
     }
     // wait
--- a/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
@@ -572,6 +572,7 @@ public class BattleDisplay extends JPanel {
 
                 @Override
                 public void actionPerformed(final ActionEvent e) {
+                  actionButton.setEnabled(false);
                   final String messageText = message + " " + btnText + ".";
                   if (chooser == null || chooserScrollPane == null) {
                     chooser =
@@ -622,6 +623,7 @@ public class BattleDisplay extends JPanel {
                           options,
                           focus);
                   if (option != 0) {
+                    actionButton.setEnabled(true);
                     return;
                   }
                   final List<Unit> killed = chooser.getSelected(false);
@@ -636,10 +638,10 @@ public class BattleDisplay extends JPanel {
                     final CasualtyDetails response = new CasualtyDetails(killed, damaged, false);
                     casualtyDetails.set(response);
                     dicePanel.removeAll();
-                    actionButton.setEnabled(false);
                     actionButton.setAction(nullAction);
                     continueLatch.countDown();
                   }
+                  actionButton.setEnabled(true);
                 }
               });
         });
--- a/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
@@ -357,44 +357,55 @@ public class BattleDisplay extends JPanel {
         SwingAction.of(
             "Submerge Subs?",
             e -> {
-              final String ok = "Submerge";
-              final String cancel = "Remain";
-              final String wait = "Ask Me Later";
-              final String[] options = {ok, cancel, wait};
-              final int choice =
-                  JOptionPane.showOptionDialog(
-                      BattleDisplay.this,
-                      message,
-                      "Submerge Subs?",
-                      JOptionPane.OK_CANCEL_OPTION,
-                      JOptionPane.PLAIN_MESSAGE,
-                      null,
-                      options,
-                      cancel);
-              // dialog dismissed
-              if (choice == -1) {
-                return;
-              }
-              // wait
-              if (choice == 2) {
-                return;
-              }
-              // remain
-              if (choice == 1) {
+              actionButton.setEnabled(false);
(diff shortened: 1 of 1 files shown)


asvitkine:
Looks like the same problem happens with select casualties - i.e. the button can be clicked multiple times and if one of the dialogs is still up when the battle ends, it freezes the game.

I can either look at fixing that in separate PR or as part of this one.

asvitkine:
> Looks like the same problem happens with select casualties - i.e. the button can be clicked multiple times and if one of the dialogs is still up when the battle ends, it freezes the game.
> 
> I can either look at fixing that in separate PR or as part of this one.

Now addressed as part of this PR.

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
```diff
@@ -407,69 +418,83 @@ private Territory getRetreatInternal(final String message, final Collection<Terr
...
+        });
     mapPanel.getUiContext().addShutdownLatch(latch);
     Interruptibles.await(latch);
     mapPanel.getUiContext().removeShutdownLatch(latch);
-    SwingUtilities.invokeLater(() -> actionButton.setAction(nullAction));
+
     return retreatTo.get();
   }
 
+  private boolean showRetreatDialog(
```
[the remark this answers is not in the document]
asvitkine: This is refactored from an existing method that was more complex. Ignoring.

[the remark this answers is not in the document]
asvitkine: This is refactored from an existing method that was more complex. Ignoring.

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
```diff
@@ -407,69 +418,83 @@ private Territory getRetreatInternal(final String message, final Collection<Terr
...
+            BattleDisplay.this,
+            comp,
+            message,
+            JOptionPane.OK_CANCEL_OPTION,
+            JOptionPane.PLAIN_MESSAGE,
+            null);
+    if (option == JOptionPane.OK_OPTION) {
+      if (comp.getSelection() != null) {
+        retreatTo.set(comp.getSelection());
+        return true;
```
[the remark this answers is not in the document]
asvitkine: This is refactored from an existing method that was more complex. Ignoring.

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
```diff
@@ -407,69 +418,83 @@ private Territory getRetreatInternal(final String message, final Collection<Terr
...
+      }
+    }
+
+    return false;
```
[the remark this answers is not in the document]
asvitkine: This is refactored from an existing method that was more complex. Ignoring.

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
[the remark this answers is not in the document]
asvitkine: This is refactored from an existing method that was more complex. Ignoring.

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
```diff
@@ -407,69 +418,83 @@ private Territory getRetreatInternal(final String message, final Collection<Terr
...
+    // wait
+    if (choice == JOptionPane.CANCEL_OPTION) {
+      return false;
+    }
+    // remain
+    if (choice == JOptionPane.NO_OPTION) {
+      return true;
+    }
+    // if you have eliminated the impossible, whatever remains, no matter how improbable,
+    // must be the truth
```
RoiEXLab: This is a great sherlock holmes quote, but this should be remove because I don't get what it's trying to imply? 😅

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
```diff
@@ -611,10 +638,10 @@ public void actionPerformed(final ActionEvent e) {
                     final CasualtyDetails response = new CasualtyDetails(killed, damaged, false);
                     casualtyDetails.set(response);
                     dicePanel.removeAll();
-                    actionButton.setEnabled(false);
```
RoiEXLab: Do you know why the button was previously disabled here?
Or if you don't, why did you have to change it?

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
```diff
@@ -407,69 +418,83 @@ private Territory getRetreatInternal(final String message, final Collection<Terr
...
+            options,
+            no);
+    // dialog dismissed
+    if (choice == JOptionPane.CLOSED_OPTION) {
+      return false;
+    }
+    // wait
+    if (choice == JOptionPane.CANCEL_OPTION) {
+      return false;
+    }
```
RoiEXLab: If you want you could merge the 2 if cases together

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
```diff
@@ -357,46 +357,57 @@ private Territory getSubmerge(final String message) {
...
+            JOptionPane.PLAIN_MESSAGE,
+            null,
+            options,
+            cancel);
+    // dialog dismissed
+    if (choice == JOptionPane.CLOSED_OPTION) {
+      return false;
+    }
+    // wait
+    if (choice == JOptionPane.CANCEL_OPTION) {
```
RoiEXLab: Same here with merged if cases.

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
```diff
@@ -357,46 +357,57 @@ private Territory getSubmerge(final String message) {
...
+        });
     mapPanel.getUiContext().addShutdownLatch(latch);
     Interruptibles.await(latch);
     mapPanel.getUiContext().removeShutdownLatch(latch);
-    SwingUtilities.invokeLater(() -> actionButton.setAction(nullAction));
     return retreatTo.get();
   }
 
+  private boolean showSubmergeDialog(
+      final String message, final AtomicReference<Territory> retreatTo) {
```
RoiEXLab: Not mandatory, but I think it would make a nicer API if you returned some sort of `RetreatResult` object here:
```java
@AllArgsConstructor
@Getter
class RetreatResult {
  private final boolean confirmed;
  @Nullable
  private final Territory targetTerritory;
}
```
Instead of setting an atomic reference. I'm not quite happy with this approach though, even though it would preserve existing logic.

Instead I'd rather go with this approach:
```java
  @Nullable
  Territory getRetreat(
      final String message, final Collection<Territory> possible, final boolean submerge) {
    final Territory territory = (!submerge || possible.size() > 1)
        ? getRetreatInternal(message, possible)
        : getSubmerge(message);
    return territory.equals(getBattleLocation()) ? null : territory;
}
```

And then both `showSubmergeDialog` and `showRetreatDialog`, return an empty optional at places where false is currently returned and an optional with the retreat territory (the current battle territory instead of null for the remain case) else.
This is a far more "logical" approach and the "current territory" could even get passed out to calling classes to avoid null in a larger refactoring in another PR.

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
asvitkine: It's not strictly necessary for this PR, but this makes things consistent with the other logic.

This code was both disabling the button and setting it to nullAction. In other code, the button was set to nullAction but not disabled. We could do either both or just setting to nullAction, but this case was inconsistent - so the change is to make it consistent.

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
asvitkine: I agree that it would be better, but perhaps we can try to minimize such large changes in this PR and do the refactoring without logic changes in a follow-up?

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
asvitkine: Done.

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
asvitkine: Done.

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
asvitkine: Done.

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
RoiEXLab: No worries. I'll submit a follow-up once this is merged, mostly wrote this for myself so I can understand why it's not that easy. ^^

DanVanAtta:
This is good to see. It's not clear though if #5010 is recreated by clicking action button multiple times, that could be yet another problem fixed here. The problem in #5010 is triggered when selecting a new battle, AFAIK without any retreat dialogs open (though perhaps there is a retreat dialog open in the background, with luck that is the problem)

Issue #4525, opened 2019-01-06
Selecting "No" in the "Leave Game" or "Exit Program" dialog box causes "Quit" menu item to not pop up "Exit Program" dialog box (nothing appears, game does not exit).


tvleavitt:
### Engine version

1.9.0.0.13066

### My Operating System

OS X Mojave 10.14.2

### Map name

World War II v6 1941

### Can you describe how to trigger the error? (eg: what sequence of actions will recreate it?)

Select "Quit" after starting a game, then click "No" when prompted in the "Exit Program" dialog. Or, select "Leave Game" and do the same thing when prompted. Subsequent selection of the "Quit" menu item yields no dialog box, and the program fails to exit. This appears to persist even after selecting "Leave Game" and closing a map. Only restarting the entire program fixes it.

### Do you have the exact error text? Please copy/paste if so

No error text.

### Instead of this error, what should have happened?

An "Exit Program" dialog box should pop up, allowing the program to be exited.

### Any additional information that may help

This also happens on a later 1.10 test build.

No save game attached, as this is easily reproduced and not specific to a particular map.

panther2:
Is that OS-related? I cannot reproduce the issue on Windows.

ssoloff:
@panther2 Yes, this is specific to macOS.

I hesitated earlier today to create a new **macOS** label for issues specific to Macs.  However, your two recent comments made my decision for me. :smile:  I'm not sure if the other devs will agree, but macOS-specific problems are kind of special for us at the moment because it's very hard for us to address them due to lack of resources.  Therefore, I felt calling them out loudly with a label is probably not a bad thing.

PR #5810, opened 2020-01-01
Fix quit dialog not appearing on Mac when dismissed.

asvitkine:
Fix quit dialog not appearing on Mac when dismissed.
We were not calling quitResponse.cancelQuit() when quit was canceled, which was needed.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix:  https://github.com/triplea-game/triplea/issues/4525
[] Other:   

## Testing

[X] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyFrame.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyFrame.java
@@ -131,7 +131,7 @@ public class LobbyFrame extends JFrame {
             .toSwingAction());
   }
 
-  public void shutdown() {
+  public boolean shutdown() {
     setVisible(false);
     dispose();
     new Thread(
@@ -140,5 +140,6 @@ public class LobbyFrame extends JFrame {
               GameRunner.exitGameIfFinished();
             })
         .start();
+    return true;
   }
 }
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyFrame.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyFrame.java
@@ -11,6 +11,7 @@ import games.strategy.engine.framework.GameRunner;
 import games.strategy.engine.lobby.client.LobbyClient;
 import games.strategy.engine.lobby.client.ui.action.BanPlayerModeratorAction;
 import games.strategy.engine.lobby.client.ui.action.DisconnectPlayerModeratorAction;
+import games.strategy.triplea.ui.QuitHandler;
 import games.strategy.triplea.ui.menubar.LobbyMenu;
 import java.awt.BorderLayout;
 import java.awt.Dimension;
@@ -28,7 +29,7 @@ import org.triplea.swing.JFrameBuilder;
 import org.triplea.swing.SwingComponents;
 
 /** The top-level frame window for the lobby client UI. */
-public class LobbyFrame extends JFrame {
+public class LobbyFrame extends JFrame implements QuitHandler {
   private static final long serialVersionUID = -388371674076362572L;
 
   @Getter private final LobbyClient lobbyClient;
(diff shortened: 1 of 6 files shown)


game-core/src/main/java/games/strategy/triplea/ui/menubar/FileMenu.java
```diff
@@ -112,7 +112,7 @@ private void addExitMenu() {
     // Mac OS X automatically creates a Quit menu item under the TripleA menu,
     // so all we need to do is register that menu item with triplea's shutdown mechanism
     if (isMac) {
-      MacOsIntegration.setQuitHandler(frame::shutdown);
+      MacOsIntegration.setQuitHandler(() -> frame.shutdown());
```
RoiEXLab: Now that shutdown returns a boolean, shouldn't this change be unecessary?

game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyFrame.java
```diff
@@ -140,5 +140,6 @@ public void shutdown() {
               GameRunner.exitGameIfFinished();
             })
         .start();
+    return true;
```
RoiEXLab: Hmm Methods returning the same objects regardless of execution path seem weird to me.
I'd actually move this true statement up one level and keep this method returning void, or alternatively have both frames implement an `QuitHandler` interface for consistency or something that will then be used.

game-core/src/main/java/games/strategy/triplea/ui/MacOsIntegration.java
```diff
@@ -40,8 +41,16 @@ public static void setOpenFileHandler(final Consumer<File> handler) {
...
-    Desktop.getDesktop().setQuitHandler((quitEvent, quitResponse) -> handler.run());
+    Desktop.getDesktop()
+        .setQuitHandler(
+            (quitEvent, quitResponse) -> {
+              if (handler.getAsBoolean()) {
+                quitResponse.performQuit();
+              } else {
+                quitResponse.cancelQuit();
+              }
+            });
```
RoiEXLab: This reminds me of #5283
Any chance if we're using the files handler the wrong way?
#5253

game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyFrame.java
asvitkine: Implementing QuitHandler sounds interesting, I can take a look at doing it.

asvitkine: I gave it a shot (didn't update the PR) and I'm not convinced it ends up cleaner.

At least, for TripleAFrame, if we implement the interface, then the non-Mac code that needs to attach the action to a JMenuItem would need to call it, which ends up looking like:

```
  public void shutdown() {
	handleQuitRequestWith(null, new QuitResponse() {
		@Override
		void cancelQuit() {
		}
		@Override
		void performQuit() {
		}
	});
  }
```

Which is messy. Or we make handleQuitRequestWith() call shutdown(), but then shutdown() needs to return a boolean as now.

So it doesn't end up cleaning that my current code. I could make LobbyFrame's shutdown() not return a boolean and handle it in the caller, but then it would be inconsistent.

I guess we could keep the version of setQuitHandler() that takes a Runnable (keep both signatures), and then LobbyFrame.shutdown() could be reverted to return void, while TripleAFrame would keep the code as-is.

What do you prefer @RoiEXLab ?

RoiEXLab: Oh, that's not what I meant.
I was thinking about creating an interface like this
```java
interface QuitHandler {
  boolean shutdown();
}
```
which is then being implemented by both frames and instead of `BooleanSupplier` you pass the object directly

asvitkine: Ah, I see. Done.

PR #5811, opened 2020-01-01
Use abstract object for retreat

RoiEXLab:
Follow-up to #5808
Couldn't use optional, because subs already use the current territory with a different meaning, and also couldn't move this object a layer outside to prevent special null meaning magic because I didn't want to break network compatibility for games.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
@@ -461,6 +461,11 @@ public class BattleDisplay extends JPanel {
       return RetreatResult.retreatTo(possible.iterator().next());
     }
 
+    return showRetreatOptions(message, possible);
+  }
+
+  private RetreatResult showRetreatOptions(
+      final String message, final Collection<Territory> possible) {
     final RetreatComponent comp = new RetreatComponent(possible);
     final int option =
         JOptionPane.showConfirmDialog(
--- a/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
@@ -44,9 +44,12 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.Timer;
 import java.util.TimerTask;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
+import java.util.logging.Level;
 import javax.annotation.Nullable;
 import javax.swing.AbstractAction;
 import javax.swing.Action;
@@ -72,6 +75,7 @@ import javax.swing.table.TableCellRenderer;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
+import lombok.extern.java.Log;
 import org.triplea.java.Interruptibles;
 import org.triplea.java.collections.CollectionUtils;
 import org.triplea.swing.SwingAction;
@@ -79,6 +83,7 @@ import org.triplea.swing.SwingComponents;
 import org.triplea.util.Tuple;
 
 /** Displays a running battle. */
+@Log
 public class BattleDisplay extends JPanel {
   private static final long serialVersionUID = -7939993104972562765L;
   private static final String DICE_KEY = "D";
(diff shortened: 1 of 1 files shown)


DanVanAtta:
@asvitkine I think you might be the best reviewer for this update. Please let me know if you won't have an opportunity to take a look at this update in the next couple days.

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
```diff
@@ -375,8 +367,33 @@ private Territory getSubmerge(final String message) {
     return retreatTo.get();
   }
 
-  private boolean showSubmergeDialog(
-      final String message, final AtomicReference<Territory> retreatTo) {
+  private Action getPlayerAction(
```
asvitkine: This still feels a bit clunky - since it's not just some random player action, but something that returns a RetreatResult.

How about just not having this method and inlining it into getRetreat() instead?

e.g.

```
final String title;
final Suppler<RetreatResult> showDialog;
if (!submerge || possible.size() > 1) {
  title = "Retreat?";
  showDialog = () -> showSubmergeDialog(message), retreatTo, latch);
} else {
  title = "Submerge Subs?";
  showDialog = () -> showSubmergeDialog(message);
}
```

And then the inlined SwingAction.of() call.

Otherwise it's hard to follow what's going on especially since there is a lot of stuff happening in this class already - so adding all these new methods just introduces more complexity.

Another option is to move the logic into separate classes for the logic, but not sure of the best set up there...

... If we step back a bit, even with this refactoring, I feel that this classes suffers a lot from mixing game logic with Swing UI code. Ultimately we'd probably want to recode things such that low level concepts - like use of JOptionPane - are abstracted away, but the business logic remains. But that's probably too big of a change for what you were after here.

Another though - I wonder if the Java Future interface would be better suited here than the combination of CountDownLatch and AtomicReference. Doesn't Future provide exactly what we need?

RoiEXLab: Using future is an excellent idea, I'll try that.

RoiEXLab:
@asvitkine Updated

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
```diff
@@ -340,43 +349,66 @@ void notifyRetreat(final Collection<Unit> retreating) {
...
+    final String title;
+    final Supplier<RetreatResult> supplier;
+    if (!submerge || possible.size() > 1) {
+      title = "Retreat?";
+      supplier = () -> showRetreatDialog(message, possible);
+    } else {
+      title = "Submerge Subs?";
+      supplier = () -> showSubmergeDialog(message);
+    }
+    final Action action = getPlayerAction(title, supplier, future);
```
asvitkine: I think you missed the other part of my comment where I suggested not having this getPlayerAction() method and just inlining it here with the changes above, since it's only called once now.

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
```diff
@@ -340,43 +349,66 @@ void notifyRetreat(final Collection<Unit> retreating) {
...
-            "Submerge Subs?",
-            e -> {
-              actionButton.setEnabled(false);
-              if (showSubmergeDialog(message, retreatTo)) {
-                actionButton.setAction(nullAction);
-                latch.countDown();
-              }
-              actionButton.setEnabled(true);
-            });
+    final CompletableFuture<Territory> future = new CompletableFuture<>();
```
asvitkine: Nit: Move future to be right above the "final Action action =" line.

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
```diff
@@ -340,43 +349,66 @@ void notifyRetreat(final Collection<Unit> retreating) {
...
+      return future.get();
+    } catch (final InterruptedException e) {
+      Thread.currentThread().interrupt();
+    } catch (final ExecutionException e) {
+      log.log(Level.INFO, "UiContext shut down before supplying result", e);
+    } finally {
+      mapPanel.getUiContext().removeActive(rejectionCallback);
+    }
+    return null;
+  }
```
asvitkine: Seems like this should live somewhere else and be generic to the return type. Since it's using UiContext, perhaps we can add the method on UiContext instead?

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
RoiEXLab: True, but if the method gets much longer code climate will complain.
That's why the method exists.

game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
RoiEXLab: Valid point, but I saw this kind of pattern in a lot of places, so that would be a larger cleanup.
I plan renaming `removeActive` and `addActive` to `addShutdownHook` and `removeShutdownHook` and replace the `Active` interface with runnable in a follow-up PR.
I think the current status is definitely an improvement for now, but I agree that it can be improved further.
I'd move this method as part of this incoming PR then.
Any further refactoring is not on my priority list though, so it would be up to you in case you want to apply this pattern elsewhere.

PR #5822, opened 2020-01-04
Change bcrypt dependency

RoiEXLab:
Follow-up to #5804
It looks like our jbcrypt library is no longer actively maintained.
That's why I decided to migrate to a what seems to be modern variant of the library https://github.com/patrickfav/bcrypt with good documentation and a modern approach to this task.
I'm about 99% sure that I chose the correct "default settings", we should consider increasing the amount of rounds (currently 10), or moving to a new hashing algorithm like argon2 when we're already at it anyways. See https://medium.com/@mpreziuso/password-hashing-pbkdf2-scrypt-bcrypt-and-argon2-e25aaf41598e

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -75,6 +75,7 @@ subprojects {
 
     ext {
         awaitilityVersion = '4.0.2'
+        bcryptVersion = '0.9.0'
         commonsCliVersion = '1.4'
         commonsCodecVersion = '1.14'
         commonsIoVersion = '2.6'
@@ -95,7 +96,6 @@ subprojects {
         jaxbApiVersion = '2.3.1'
         jaxbCoreVersion = '2.3.0.1'
         jaxbImplVersion = '2.3.2'
-        bcryptVersion = '0.9.0'
         jdbiVersion = '3.12.0'
         jlayerVersion = '1.0.1.4'
         junitJupiterVersion = '5.5.2'
--- a/build.gradle
+++ b/build.gradle
@@ -95,7 +95,7 @@ subprojects {
         jaxbApiVersion = '2.3.1'
         jaxbCoreVersion = '2.3.0.1'
         jaxbImplVersion = '2.3.2'
-        jbcryptVersion = '0.4'
+        bcryptVersion = '0.9.0'
         jdbiVersion = '3.12.0'
         jlayerVersion = '1.0.1.4'
         junitJupiterVersion = '5.5.2'
--- a/game-core/build.gradle
+++ b/game-core/build.gradle
@@ -1,5 +1,3 @@
-import org.apache.tools.ant.filters.ReplaceTokens
-
 description = 'TripleA core library containing code shared between headed and headless versions'
 
 ext {
--- a/game-core/build.gradle
+++ b/game-core/build.gradle
@@ -14,7 +14,6 @@ dependencies {
     implementation "org.apache.commons:commons-math3:$commonsMathVersion"
     implementation "org.apache.httpcomponents:httpclient:$apacheHttpComponentsVersion"
     implementation "org.apache.httpcomponents:httpmime:$apacheHttpComponentsVersion"
-    implementation "at.favre.lib:bcrypt:$bcryptVersion"
     implementation "org.snakeyaml:snakeyaml-engine:$snakeYamlVersion"
     implementation project(":domain-data")
     implementation project(":http-clients")
--- a/game-core/build.gradle
+++ b/game-core/build.gradle
@@ -14,7 +14,7 @@ dependencies {
     implementation "org.apache.commons:commons-math3:$commonsMathVersion"
     implementation "org.apache.httpcomponents:httpclient:$apacheHttpComponentsVersion"
     implementation "org.apache.httpcomponents:httpmime:$apacheHttpComponentsVersion"
-    implementation "org.mindrot:jbcrypt:$jbcryptVersion"
+    implementation "at.favre.lib:bcrypt:$bcryptVersion"
     implementation "org.snakeyaml:snakeyaml-engine:$snakeYamlVersion"
     implementation project(":domain-data")
     implementation project(":http-clients")
--- a/http-server/src/main/java/org/triplea/server/user/account/PasswordBCrypter.java
+++ b/http-server/src/main/java/org/triplea/server/user/account/PasswordBCrypter.java
@@ -1,6 +1,7 @@
 package org.triplea.server.user.account;
 
 import at.favre.lib.crypto.bcrypt.BCrypt;
+import at.favre.lib.crypto.bcrypt.LongPasswordStrategies;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.io.BaseEncoding;
@@ -18,7 +19,7 @@ public class PasswordBCrypter implements Function<String, String> {
   public String apply(final String password) {
     // TODO: Md5-Deprecation remove hashing here, move to client-side
     final String hashed = hashPasswordWithSalt(password);
-    return BCrypt.withDefaults().hashToString(10, hashed.toCharArray());
+    return BCrypt.with(LongPasswordStrategies.none()).hashToString(10, hashed.toCharArray());
   }
 
   @VisibleForTesting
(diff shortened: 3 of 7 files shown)


build.gradle
```diff
@@ -95,7 +95,7 @@ subprojects {
         jaxbApiVersion = '2.3.1'
         jaxbCoreVersion = '2.3.0.1'
         jaxbImplVersion = '2.3.2'
-        jbcryptVersion = '0.4'
+        bcryptVersion = '0.9.0'
```
ssoloff: Looks like someone went to the trouble to sort this list lexicographically; might want to move this entry up.

game-core/build.gradle
```diff
@@ -16,7 +14,7 @@ dependencies {
     implementation "org.apache.commons:commons-math3:$commonsMathVersion"
     implementation "org.apache.httpcomponents:httpclient:$apacheHttpComponentsVersion"
     implementation "org.apache.httpcomponents:httpmime:$apacheHttpComponentsVersion"
-    implementation "org.mindrot:jbcrypt:$jbcryptVersion"
+    implementation "at.favre.lib:bcrypt:$bcryptVersion"
```
ssoloff: I don't see any other changes in the `game-core` project.  Is this dependency still needed?

DanVanAtta:
Test coverage should confirm 'legacy' cases. Notably the [http login integration](https://github.com/triplea-game/triplea/blob/master/http-server/src/test/java/org/triplea/server/user/account/login/LoginControllerIntegrationTest.java) test:

1. Logs a user in with a bcrypted password.
2. Uses forgot password functionality to login (IIRC that is a temp bcrypted password)
3. Checks legacy md5 password case and that the password is upgraded to bcrypt. Now with a secure connection we can send plaint text the password to server over https and this allows us to do a re-hash. When md5 is gone we'll be able to SHA512 the passworsds on client side before sending the password.

ssoloff:
> Test coverage should confirm 'legacy' cases

Just to clarify ... By "legacy-bcrypted password," I meant a case whereby a password has been hashed using `org.mindrot.jbcrypt.BCrypt#hashpw()` and is now verified using `at.favre.lib.crypto.bcrypt.BCrypt#verify()`, which would happen if 1.9 lobby DB records with bcrypted passwords are migrated forward to the 2.0 lobby DB.  It's unlikely there will be a problem, but it doesn't hurt to check it just in case one of the `at.favre.lib.crypto.bcrypt.BCrypt` configuration parameters wasn't chosen correctly to match the hard-coded `org.mindrot.jbcrypt.BCrypt` configuration.

game-core/build.gradle
RoiEXLab: Apparently not.
Good catch

DanVanAtta:
Indeed, correct me if I'm wrong, but I believe we have that scenario @ssoloff :+1: 
Bcrypt hash:
https://github.com/triplea-game/triplea/blob/master/http-server/src/test/resources/datasets/integration.yml#L65

Login test:
https://github.com/triplea-game/triplea/blob/master/http-server/src/test/java/org/triplea/server/user/account/login/LoginControllerIntegrationTest.java#L42

Verify:
https://github.com/triplea-game/triplea/pull/5822/files#diff-93582380eda20526e7d2cf6c506f9241R13

http-server/src/main/java/org/triplea/server/user/account/login/authorizer/BCryptHashVerifier.java
```diff
@@ -1,14 +1,17 @@
...
-import org.mindrot.jbcrypt.BCrypt;
 import org.triplea.server.user.account.PasswordBCrypter;
 
 public class BCryptHashVerifier implements BiPredicate<String, String> {
   @Override
   public boolean test(final String password, final String databaseBcrypted) {
     // TODO: Md5-Deprecation Move SHA512 hashing to client side
     final String hashedPassword = PasswordBCrypter.hashPasswordWithSalt(password);
-    return BCrypt.checkpw(hashedPassword, databaseBcrypted);
+    return BCrypt.verifyer(null, LongPasswordStrategies.none())
```
DanVanAtta: The level of abstraction is perhaps inconsistent here, would you agree @RoiEXLab ?

Notably, we have a high level call `PasswordBCrypter.hashPasswordWithSalt`, which abstracts away parameters and configuration, then we have a more low level call to verify.

Second, the params to `verifyer` seem to be chosen for specific reasons, but it's not immediately clear (missing context) what those reasons were.

I think you could resolve both concerns by moving the `verifyer` call to a similar static method. That method could then be documented to note why the parameters are chosen as they are, and we'd have a consistent abstraction in this method. For example:
```
final String hashedPassword = PasswordBCrypter.hashPasswordWithSalt(password);
return BcryptedPasswordVerifyer.verify(hashedPassword);
```

WDYT @RoiEXLab ?

(A third benefit, we would not need to repeat the verifyer parameters in [test](https://github.com/triplea-game/triplea/pull/5822/files#diff-d9b0f1911d3493fb0dffedbe799f1d81R31), DRY)

http-server/src/main/java/org/triplea/server/user/account/PasswordBCrypter.java
```diff
@@ -18,7 +19,7 @@
   public String apply(final String password) {
     // TODO: Md5-Deprecation remove hashing here, move to client-side
     final String hashed = hashPasswordWithSalt(password);
-    return BCrypt.hashpw(hashed, BCrypt.gensalt());
+    return BCrypt.with(LongPasswordStrategies.none()).hashToString(10, hashed.toCharArray());
```
DanVanAtta: Really similar to my other comment, I think we may benefit to create a utility method around this hasing method:
- provides a good place to document decisions/parameters:
  - why '10'?
  - why 'LongPasswordStrategies.none()'?
- Avoids repetition between source code and test (DRY)

http-server/src/main/java/org/triplea/server/user/account/PasswordBCrypter.java
```diff
@@ -18,7 +19,7 @@
   public String apply(final String password) {
     // TODO: Md5-Deprecation remove hashing here, move to client-side
     final String hashed = hashPasswordWithSalt(password);
-    return BCrypt.hashpw(hashed, BCrypt.gensalt());
```
DanVanAtta: Are we still using a bcrypt salt anywhere?
If not, for this PR or a fast-follow, we can clean up the configuration that was intended to inject a salt:
- https://github.com/triplea-game/triplea/blob/master/http-server/src/main/java/org/triplea/server/http/AppConfig.java#L57
- https://github.com/triplea-game/triplea/blob/master/http-server/configuration.yml#L12
- https://github.com/triplea-game/triplea/blob/master/infrastructure/ansible/roles/http_server/templates/configuration.yml.j2#L8
- https://github.com/triplea-game/triplea/blob/master/infrastructure/ansible/roles/http_server/defaults/main.yml#L20

ssoloff:
@DanVanAtta Yup, that should cover it.  Thanks for pointing it out.  :+1:

http-server/src/main/java/org/triplea/server/user/account/PasswordBCrypter.java
RoiEXLab: I'm not sure what you mean by that.
Obviously a salt is still generated for the individual passwords, you just don't have to manually supply them any longer.

RoiEXLab:
@DanVanAtta Updated, let me know what you think.

http-server/src/main/java/org/triplea/server/user/account/PasswordBCrypter.java
DanVanAtta: Good news - That means the `AppConfig` scaffolding added for it, and never wired fully to be used, can be removed as it'll never be used.

DanVanAtta:
Sorry @RoiEXLab , it'll be some time before i can really review this again in further detail. Recommend @ssoloff taking a look at the updates, or doing a final self-review pass and merging.

PR #5842, opened 2020-01-12
AI add buying zero move units

ron-murhammer:
Address forum requests: https://forums.triplea-game.org/topic/1707/simple-trigger-help/20

Add AI support to purchase 0 move and/or construction units which are generally used as static defenses. These are fairly important in some maps.

## Testing

[x] Manual testing done

Ran several all AI games on NWO, NML, and revised to ensure there are no issues and the AI appropriately considers buying bunkers in NWO and trenches in NML.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
@@ -2569,8 +2569,6 @@ class ProPurchaseAi {
                 ppt.getTerritory(),
                 purchaseTerritory.getTerritory(),
                 isBid)) {
-
-          // Place max number of units
           final List<Unit> constructions =
               CollectionUtils.getMatches(unitsToPlace, Matches.unitIsConstruction());
           unitsToPlace.removeAll(constructions);
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
@@ -23,6 +23,7 @@ import games.strategy.triplea.ai.pro.util.ProBattleUtils;
 import games.strategy.triplea.ai.pro.util.ProMatches;
 import games.strategy.triplea.ai.pro.util.ProOddsCalculator;
 import games.strategy.triplea.ai.pro.util.ProPurchaseUtils;
+import games.strategy.triplea.ai.pro.util.ProPurchaseValidationUtils;
 import games.strategy.triplea.ai.pro.util.ProTerritoryValueUtils;
 import games.strategy.triplea.ai.pro.util.ProTransportUtils;
 import games.strategy.triplea.ai.pro.util.ProUtils;
@@ -730,7 +731,7 @@ class ProPurchaseAi {
             new ArrayList<>(defensePurchaseOptions);
         allDefensePurchaseOptions.addAll(zeroMoveDefensePurchaseOptions);
         final List<ProPurchaseOption> purchaseOptionsForTerritory =
-            ProPurchaseUtils.findPurchaseOptionsForTerritory(
+            ProPurchaseValidationUtils.findPurchaseOptionsForTerritory(
                 proData,
                 player,
                 allDefensePurchaseOptions,
@@ -743,7 +744,7 @@ class ProPurchaseAi {
         while (true) {
 
           // Select purchase option
-          ProPurchaseUtils.removeInvalidPurchaseOptions(
+          ProPurchaseValidationUtils.removeInvalidPurchaseOptions(
               player,
               startOfTurnData,
               purchaseOptionsForTerritory,
@@ -959,9 +960,9 @@ class ProPurchaseAi {
 
       // Remove options that cost too much PUs or production
       final List<ProPurchaseOption> purchaseOptionsForTerritory =
-          ProPurchaseUtils.findPurchaseOptionsForTerritory(
+          ProPurchaseValidationUtils.findPurchaseOptionsForTerritory(
               proData, player, specialPurchaseOptions, t, isBid);
-      ProPurchaseUtils.removeInvalidPurchaseOptions(
+      ProPurchaseValidationUtils.removeInvalidPurchaseOptions(
           player,
           startOfTurnData,
           purchaseOptionsForTerritory,
@@ -1046,13 +1047,13 @@ class ProPurchaseAi {
 
       // Determine most cost efficient units that can be produced in this territory
       final List<ProPurchaseOption> landFodderOptions =
-          ProPurchaseUtils.findPurchaseOptionsForTerritory(
+          ProPurchaseValidationUtils.findPurchaseOptionsForTerritory(
               proData, player, purchaseOptions.getLandFodderOptions(), t, isBid);
       final List<ProPurchaseOption> landAttackOptions =
-          ProPurchaseUtils.findPurchaseOptionsForTerritory(
+          ProPurchaseValidationUtils.findPurchaseOptionsForTerritory(
               proData, player, purchaseOptions.getLandAttackOptions(), t, isBid);
       final List<ProPurchaseOption> landDefenseOptions =
-          ProPurchaseUtils.findPurchaseOptionsForTerritory(
+          ProPurchaseValidationUtils.findPurchaseOptionsForTerritory(
               proData, player, purchaseOptions.getLandDefenseOptions(), t, isBid);
 
       // Determine enemy distance and locally owned units
@@ -1079,7 +1080,7 @@ class ProPurchaseAi {
       for (final Iterator<Unit> it = unplacedUnits.iterator(); it.hasNext(); ) {
         final Unit u = it.next();
         if (remainingUnitProduction > 0
-            && ProPurchaseUtils.canUnitsBePlaced(proData, List.of(u), player, t, isBid)) {
+            && ProPurchaseValidationUtils.canUnitsBePlaced(proData, List.of(u), player, t, isBid)) {
           remainingUnitProduction--;
           unitsToPlace.add(u);
           it.remove();
@@ -1094,7 +1095,7 @@ class ProPurchaseAi {
       while (true) {
 
         // Remove options that cost too much PUs or production
-        ProPurchaseUtils.removeInvalidPurchaseOptions(
+        ProPurchaseValidationUtils.removeInvalidPurchaseOptions(
             player,
             startOfTurnData,
             landFodderOptions,
@@ -1102,7 +1103,7 @@ class ProPurchaseAi {
             remainingUnitProduction,
             unitsToPlace,
             purchaseTerritories);
-        ProPurchaseUtils.removeInvalidPurchaseOptions(
+        ProPurchaseValidationUtils.removeInvalidPurchaseOptions(
             player,
             startOfTurnData,
             landAttackOptions,
@@ -1110,7 +1111,7 @@ class ProPurchaseAi {
             remainingUnitProduction,
             unitsToPlace,
             purchaseTerritories);
-        ProPurchaseUtils.removeInvalidPurchaseOptions(
+        ProPurchaseValidationUtils.removeInvalidPurchaseOptions(
             player,
             startOfTurnData,
             landDefenseOptions,
@@ -1336,10 +1337,10 @@ class ProPurchaseAi {
 
       // Determine units that can be produced in this territory
       final List<ProPurchaseOption> purchaseOptionsForTerritory =
-          ProPurchaseUtils.findPurchaseOptionsForTerritory(
+          ProPurchaseValidationUtils.findPurchaseOptionsForTerritory(
               proData, player, purchaseOptions.getFactoryOptions(), maxTerritory, isBid);
       resourceTracker.removeTempPurchase(maxPlacedOption);
-      ProPurchaseUtils.removeInvalidPurchaseOptions(
+      ProPurchaseValidationUtils.removeInvalidPurchaseOptions(
           player,
           startOfTurnData,
           purchaseOptionsForTerritory,
@@ -1560,7 +1561,7 @@ class ProPurchaseAi {
 
           // Determine sea and transport units that can be produced in this territory
           final List<ProPurchaseOption> seaPurchaseOptionsForTerritory =
-              ProPurchaseUtils.findPurchaseOptionsForTerritory(
+              ProPurchaseValidationUtils.findPurchaseOptionsForTerritory(
                   proData,
                   player,
                   purchaseOptions.getSeaDefenseOptions(),
@@ -1580,7 +1581,7 @@ class ProPurchaseAi {
             }
 
             // Select purchase option
-            ProPurchaseUtils.removeInvalidPurchaseOptions(
+            ProPurchaseValidationUtils.removeInvalidPurchaseOptions(
                 player,
                 startOfTurnData,
                 seaPurchaseOptionsForTerritory,
@@ -1788,7 +1789,7 @@ class ProPurchaseAi {
 
         // Determine sea and transport units that can be produced in this territory
         final List<ProPurchaseOption> seaPurchaseOptionsForTerritory =
-            ProPurchaseUtils.findPurchaseOptionsForTerritory(
+            ProPurchaseValidationUtils.findPurchaseOptionsForTerritory(
                 proData,
                 player,
                 purchaseOptions.getSeaDefenseOptions(),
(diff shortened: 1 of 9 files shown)


ron-murhammer:
Looks like there are a few codeclimate issues around the length/complexity of some of the helper methods for validating unit production. These are already refactored from the original method. I'm not sure whether its worth further breaking these up or if its easier to keep the logic together. Open to feedback here.

DanVanAtta:
Extract class, replace parameter with method and/or introduce parameter object refactorings all probably could help with the codeclimate issues. To keep logic together, it does not always have to be broken up across methods, but perhaps introducing new classes.

Cernelius:
I suggest Domination 1914 NML (trenches) as test map, as it is also interesting for having immobile units that provide hitpoints only (thus highly complementary to land units, differently from the NWO bunkers, that are substantially self-sufficient). Maybe also, then, lowering their cost to 2 PUs only, see if the AI will be down to buy significantly more of them.

The only issue is that trenches, there, are made worth less by the fact that there is gas, having an anti-trenches behaviour, and the AI cannot see this. So, not seeing this, the AI should probably buy more trenches than a good player would.

Moving forward, the other map I suggest is Rome Total War, where you have fairly efficient immobile infrastructures that provide, instead, defensive powers (and are capturable). Again a case of high complementarity with mobile units (and specifically with archers, through support), but substantially opposite to D1914NML.

ron-murhammer:
Alright. I took a pass on refactoring. I think there are some improvements though now that I've split a class in half, CodeClimate thinks its all new code (doesn't consider moved methods) so is now I went up to 17 issues! I think my patience to battle with CodeClimate further is probably at about its end for this PR.

Cernelius:
I think this is a major step for the AI, as I believe purchasable immobile combat units can really add a lot to games, on multiple levels (starting from realism: Maginot Line and Atlantic Wall anyone?).

Just to be sure, does this mean that the AI will also consider buying immobile infrastructures (like the forts or Rome Total War)?

1. For example, if I have a regular AA Gun, except only that has movement 0, will the AI considering purchasing it?

2. For example, if I have a regular AA Gun, except that has movement 0 and target some other units than just air (possibly all units in the game), will the AI considering purchasing it?

3. For example, if I have an infrastructure unit that has movent 0 but positive defence value, will the AI considering purchasing it?

4. For example, if I have an infrastructure unit that has movent 0 but defensive support abilities (usually either bonus to allied or malus to enemy units or both), will the AI considering purchasing it?

The fort of Rome Total War is an example of both point 3 and point 4 above.

Just to know what I would be going to expect, once passing to the testing phase (I've actually a game under development in which I have "castle" units, that are immobile defensive infrastructures, though they are also built in two steps, because you just don't dump concrete off the sky, so I guess nevermind until that is supported too).

Cernelius:
The other major item would be distinguishing the case of truly immobile units from the case of movement 0 units that are actually supposed to be sea/land/air transported, instead, though, currently, it is not simply possible to sea transport movement 0 unit, as loading counts as moving for the loading unit itself.

game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
```diff
@@ -2551,6 +2571,10 @@ private void addUnitsToPlaceTerritory(
                 isBid)) {
 
           // Place max number of units
```
DanVanAtta: Is this comment now misplaced? It's hard to see how it relates to the next couple lines, seems like it relates to lines below it.

The comment block also seems to be blocking off a logical step, would be a good candidate for an extract method refactor.

game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
```diff
@@ -2551,6 +2571,10 @@ private void addUnitsToPlaceTerritory(
...
+          final List<Unit> constructions =
+              CollectionUtils.getMatches(unitsToPlace, Matches.unitIsConstruction());
+          unitsToPlace.removeAll(constructions);
```
DanVanAtta: Why are constructions being removed from unitsToPlace and added to `ppt`. 

Second, would `ppt` provide better encapsulation if it provided a method to add constructions rather than exposing a mutable datastructure?

EG:
```
 ppt.getPlaceUnits().addAll(constructions);
```
to:
```
 ppt.addConstructions(constructions); // this method could assert that all units match 'unitIsconstruction'
```

game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPurchaseOption.java
```diff
@@ -71,6 +72,10 @@
     final Resource pus = data.getResourceList().getResource(Constants.PUS);
     cost = productionRule.getCosts().getInt(pus);
     costs = productionRule.getCosts();
+    isConstruction = unitAttachment.getIsConstruction();
```
DanVanAtta: These 4 data members seem very related. Have you considered creating a type around it?

game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPurchaseOption.java
```diff
@@ -252,6 +257,9 @@ public double getAmphibEfficiency(
   }
 
   private double calculateLandDistanceFactor(final int enemyDistance) {
+    if (movement <= 0) {
+      return 0.1;
```
DanVanAtta: What's the rationale for 0.1?

game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPurchaseTerritory.java
```diff
@@ -64,7 +65,8 @@ public ProPurchaseTerritory(
   public int getRemainingUnitProduction() {
     int remainingUnitProduction = unitProduction;
     for (final ProPlaceTerritory ppt : canPlaceTerritories) {
-      remainingUnitProduction -= ppt.getPlaceUnits().size();
+      remainingUnitProduction -=
+          CollectionUtils.countMatches(ppt.getPlaceUnits(), Matches.unitIsConstruction().negate());
```
DanVanAtta: It's a bit difficult to follow this line. An intermediate variable maybe would help, or adding a `Matches` for 'notConstruction' to avoid the negate. Have you considered unit testing this method to help document and verify it?

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProPurchaseValidationUtils.java
```diff
@@ -0,0 +1,279 @@
...
+import games.strategy.triplea.delegate.AbstractPlaceDelegate;
+import games.strategy.triplea.delegate.Matches;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Predicate;
+import org.triplea.java.collections.CollectionUtils;
+
+/** Pro AI purchase utilities. */
```
DanVanAtta: Is this comment here because it has to be? I wonder as `ValidationUtils` seems to be more specific than the comment implies. Second, have you considered using `@UtilityClass` to avoid the private constructor?

game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
ron-murhammer: Yeah, removed this comment.

game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
ron-murhammer: Essentially the method removes units from unitsToPlace and adds them into the ProPurchaseTerritory structure. The reason for this is that the AI decides how many units to build in a territory but then needs to map them to which factory is actually placing the (mostly only matters for sea territories).

Eventually, I do agree that trying to push this and a few other methods into ProPurchaseTerritory would be best just didn't want to tackle that in this PR. If you look at ProPurchaseTerritory its mostly just a data structure currently and was created to begin grouping some of the data fields together to avoid passing around lots of lists but hasn't really gotten much functionality added into it yet.

game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPurchaseOption.java
ron-murhammer: Not really. I'd have to think more about this class in general to see if/how it would be best to start breaking it up a bit. Compared to most AI classes though this is probably one of the better ones.

game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPurchaseOption.java
ron-murhammer: Want it to be essentially an order of magnitude worse than 1 move units. Don't want to return 0 as this is used to multiply with other values to get a kind of purchase value score for units.

game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPurchaseTerritory.java
ron-murhammer: Updated.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProPurchaseValidationUtils.java
ron-murhammer: Updated comment. Added that annotation to this class and ProPurchaseUtils.

ron-murhammer:
@DanVanAtta Made some updates based on your review.

@Cernelius Responses:
For example, if I have a regular AA Gun, except only that has movement 0, will the AI considering purchasing it?
- Depends which type of AA (battle vs SBR) and what other stats it has. Generally, SBR AA purchases are treated completely separately and won't buy 0 move units but this could be added. Battle AA would be consider with other units just probably end up low priority since it puts no value on the AA.

For example, if I have a regular AA Gun, except that has movement 0 and target some other units than just air (possibly all units in the game), will the AI considering purchasing it?
- Depends on type of AA (battle vs SBR). If you have an AA Gun with battle AA that has 0 movement and isConstruction then the AI should consider building it though it will be low priority since it doesn't value the AA (so HP/attack/defense would make it more likely or if its the only option).

For example, if I have an infrastructure unit that has movent 0 but positive defence value, will the AI considering purchasing it?
- It should consider it. Updating NML or NWO by replacing bunker or trench with your unit so its the only construction option then setting up a scenario where it needs the extra defense at say a factory would be a good way to test.

For example, if I have an infrastructure unit that has movent 0 but defensive support abilities (usually either bonus to allied or malus to enemy units or both), will the AI considering purchasing it?
- It should consider it.

game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPurchaseOption.java
DanVanAtta: Would you mind adding a comment in code? That is actually a bit more involved than I had expected.

Cernelius:
>Add AI support to purchase 0 move and construction

Now I wonder if this is actually about movement 0 units or units that are both movement 0 and construction? I mean, will movement 0 units be considered only if they are also constructions (like in NWO and D1914NML)?

Does that mean that this supports both 0 movement units and construction units, or that it supports units that are both?

Like, is this a thing also for the Mortella_Tower of Napoleonic Empires and the fort of 270BC (both normal placement, albeit with mobile factories in the game)?

game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPurchaseOption.java
ron-murhammer: Added.

ron-murhammer:
@Cernelius It will consider 0 move and/or construction units (they are just often used together). This change was primarily focused on 0 move units as there was a hard check preventing the AI from buying them. But there needed to be some enhancements to how the AI deals with construction units as well and it needed to handle the various construction limit parameters. 

It should consider buying Mortella_Tower of Napoleonic Empires and the fort of 270BC but generally values 0 move units low and only to defend high strategic territories like factories.

Cernelius:
>Want it to be essentially an order of magnitude worse than 1 move units.

I don't think it is necessary being that extreme. Just take as example the Mortella_Tower and the Fortress of Napoleonic Empires. The mortella tower is defence 2 and cost 3 and the fortress is defence 4 and cost 5. Chasseurs are att/def 1/3, cost 5 and movement 1. So, a movement 1 unit that is att/def 0/2 may be balanced costing 3.5, that is only 0.5 PUs more than the cost of the Mortella Tower.

You can just treat immobile units has having attack 0 in all cases, then value them at **something from 0.5 to 0.8 the same movement 1 unit**, not as much as 0.1 only!

Cernelius:
The Mortella_Tower, despite being only slightly better than what you can get at movement 1, is fairly consistently bought in Napoleonic Empires, despite being clearly a bad unit. The Fortress is even worse than the Mortella Tower, and basically just a last-minute unit for desperate situations.

ron-murhammer:
@Cernelius It might be better to move this conversation to the forum as its gonna kind of get lost here once this PR is merged.

I realize the an order of magnitude is a bit extreme but the point is I don't want the AI to overbuy 0 move units as that could cause very bad behavior. I essentially only want it to buy them when it absolutely needs them for defense. I may eventually adjust this some but would rather start overly conservative. 

I actually rarely buy mortella towers and fortresses in NE and only in situations where I really need to squeeze every PU out or need to defend a factory territory from constant naval invasion threat. IMO those units are pretty minor parts of NE and are more often traps for inexperienced players that waste PUs building too many of them.

Cernelius:
Sure, but what I was saying is that Mortella_Tower, let's say it is attack/defence 0/2, is a unit that, if it would move 1, would have a balanced cost of 3 or 3.5 PUs, no more. Instead, at movement 0, costs 3 PUs. This means that unit is priced at something like 90% than its balanced cost if it would be able to move 1 and attack (at attack power 0).

So, if you just set immobile at 0.5 the value of movement 1, provided that immobile is seen as always having attack value 0, then you are already valuing it far less than what most TripleA games do. So that's already very conservative, and also 0.8 I'm pretty sure would make the AI buy Mortella_Tower only very rarely.

The way you are going, I'm pretty sure Mortella_Tower will be never bought, since it is not even a unit you want to drop at the last minute, to maximize defence (that is Fortress).

ron-murhammer:
@Cernelius 0.5 would probably be enough but the point is I really only want the AI to buy 0 move units when it has no other options for now. So it will still buy Mortella_Tower but only to defend a factory when it has 3 PUs left over, it will buy bunkers and trenches because its the only construction defense option. Something like 0.8 would probably cause issues as the AI could then potentially build them kind of randomly in places that don't really even need them as there is a good amount of randomness added into purchasing.

I'm open to adjusting this in the future and there are additional improvements that can be made around this.

Cernelius:
You are also at risk that mapmakers will start going nuts with all kind of hacks to make the AI buy those immobile at the 0.1 valuation, like triggering in and out some super stats, during purchase.

ron-murhammer:
@Cernelius Well, anything is possible. I tested on NML and NWO/WaW which are by far the most popular maps that have 0 move units that are frequently bought. I think my point is that 0 move units are a somewhat unique type and are more situational (shouldn't just be compared to all other 1+ move units when purchasing in any factory). There are certain situations when they should be purchased and modeling that in the AI is the goal.

Cernelius:
I don't care much this discussion getting lost, as all I'm saying, as far as units balance, are things that if someone cannot agree, there's probably no point talking with anyways.

Ok, well, yes anyways probably Napoleonice Empires is the best example of, instead, movement 0 units that have nothing special, so that can be taken as a good example of balancing such units, and we can agree Napoleonic Empires does it in a very cautious way, making movement 0 units barely worth buying at all, and very marginal.

The other example is the Fortress, that is defence 4 and cost 5.

We can consider a unit that is att/def 1/4 and movement 1. All considered, the balanced cost for such a unit would be 6.5 PUs, rounding it to 0.5 PUs.

So, the balanced cost for an att/def 0/4 and movement 1 unit (you can still use it as fodder, in attack) may be 6 PUs.

Hence, the Fortress at 5 PUs is priced at something like 83% of what it may be priced if it would also be a movement 1 and attack 0 unit.

So, I think that for your aim of assuring an extreme level of marginalism on normal purchase choices, rather than fixing it at 10%, one could go with the rule of thumb of fixing it at 50% of what appers to be the usual map balancing, that is 50% of 80%, thus 40%.

So I would suggest you at least upping that value from the current 0.1 to a still very conservative 0.4, if you really want to be extremely cautious.

Then, the ability of using units to place in excess of placement limits, like the bunker of NWO or the trenches of D1914NML, thus having a premium price for the ability of placing more than my production limits, is probably something that should be considered beside the fact of being movement 0 at all, as nothing mandates the two being related in any ways (one could make a map with mercenaries that are constructions and cost more than normal units at same value, to represent raising such troops once you cannot get any more of your men).

PR #5849, opened 2020-01-19
Remove throws on flag rendering and default to 'NONE'

DanVanAtta:
- Replace case statement for flag rendering with if/else
- Default to flag rendering of 'NONE', if flag rendering is
 otherwise not 'small' or 'large', we will default to NONE

This avoids potential error condition if we change the flag rendering
name or somehow have no flag rendering stored in ClientSettings.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

## Additional Review Notes

Updates here are mostly whitespace changes, just a few lines are changed.
Use this link to review without white space changes: https://github.com/triplea-game/triplea/pull/5849/files?utf8=%E2%9C%93&diff=unified&w=1

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/screen/UnitsDrawer.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/screen/UnitsDrawer.java
@@ -137,37 +137,39 @@ public class UnitsDrawer extends AbstractDrawable {
 
     final UnitFlagDrawMode drawMode =
         ClientSetting.unitFlagDrawMode.getValue().orElse(UnitFlagDrawMode.NONE);
-    if (drawMode == UnitFlagDrawMode.LARGE_FLAG) {
-      // If unit is not in the "excluded list" it will get drawn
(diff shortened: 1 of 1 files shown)


DanVanAtta:
I do like the switch. With two cases handled it becomes more questionable. Secondly the following was a bit easier to loop in with the if/else block so we have one level of control flow rather than two:
```
  img.ifPresent(image -> drawUnit(graphics, image, bounds, data));
```

DanVanAtta:
yeah, i'm on the fence, I like the switch, agree the default would work well. The two are pretty equivalent, switch also has a potential for compiler warning if we add cases (though if we do, we'd certainly update this block). If doing this a second time, using a switch would perhaps be the way to go. The first rev of this update simply converted the switch statement, then I looped in the `img.ifPresent` and simplest way to do that was to extend the if/else blocks.

PR #5851, opened 2020-01-19
Remove toggle flags button and 'L' hotkey for it

DanVanAtta:
Removes toggle flags button in unit scroller the 'L' hotkey to toggle flags.

 

## Functional Changes

[] New map or map update
[] New Feature
[x] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done

Verified hotkeys and rendering of unit scroller.

## Screens Shots

Screenshot from 2020-01-18 22-51-39

## Additional Review Notes

Next steps for unit scroller:
- change unit highlight to be an on/off toggle and bound to a view menu item, remove unit highlight button
- update unit scroller image to highlight on hover and click action will take the place of the 'center' button
- consider updating unit scroller to be 'always' shown similar to the units-to-place panel (with a similar expand/collapse button)


DanVanAtta:
Ignoring code climate warning, duplication is slightly intentional. The alternative to inlining the builder method was to expand the helper method to have 5 args. The many-args and indirection is harder to understand than the builder being directly present.

game-core/src/main/java/games/strategy/triplea/ui/AbstractMovePanel.java
```diff
@@ -238,10 +238,7 @@ protected final void display(final GamePlayer gamePlayer, final String actionLab
         () -> {
           removeAll();
           add(movedUnitsPanel(gamePlayer, actionLabel));
-          add(
-              getUnitScrollerPanel(
-                  frame.getLocalPlayers(),
-                  () -> FlagDrawMode.toggleNextDrawMode(frame.getMapPanel())));
```
RoiEXLab: Is this method still used?

Cernelius:
I'd rather use Ctrl and numbers to add same or bigger movement units from zones to the battle calculator, but, remaining on topic, I would say Ctrl+0 makes more sense for no flags than Ctrl+1.

ron-murhammer:
@DanVanAtta Any thoughts on hotkey alternatives and maybe a single toggle hotkey to cycles through rather than individual hotkeys for each flag mode? Not sure I really like using ctrl+numbers in this case.

DanVanAtta:
@ron-murhammer 

> Any thoughts on hotkey alternatives

Would be in favor, but found no three side by side that are available that would also make sense. Perhaps you've any suggestions?

> maybe a single toggle hotkey to cycles through rather than individual hotkeys for each flag mode? 

That gets us back to the icon to represent exactly that hotkey. We could go back to a dual menu option, one to toggle on/off (with a hotkey), and a radio for small/large (no hotkey). 

I wonder more fundamentally if we actually need a hotkey at all.

> Not sure I really like using ctrl+numbers in this case.

I don't either really (as mentioned).

game-core/src/main/java/games/strategy/triplea/ui/AbstractMovePanel.java
DanVanAtta: AFAIK we did not remove any references to a `display(` method in this PR, please note if I'm wrong about that. Given that, the method should still be used if it was used before at all.

DanVanAtta:
Tempted to move forward and have us debate hotkeys and update when we resolved that debate. Seems to be a lot of heartache over this feature and churn. I'm not sure exactly how we can be more decisive, but we need to be better about looking at the UX failures we have here and come up with a single plan to solve them. We seem to keep bouncing between requirements on this feature.

For the sake of getting to other dependent changes, will just remove all hotkeys on flags for now, that solves the UX problems if not introduces a new one where power users toggling this feature frequently would want a hotkey.

PR #5852, opened 2020-01-19
Encrypt passwords/secrets for prelease & production

DanVanAtta:
- Encrypt passwords using ansible vault
- Drop from configuration.yml, bcrypted_salt, not used and will not be needed.
- Add utility script 'create_secret' to make it a bit easier to create secrets
- Set password for the 'postgres' db user

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done
FWIW ran ansible deployment against a local vagrant

## Additional Review Notes

The changed passwords are likely to bork prerelease. I'm planning to create a new server to replace prerelease where we'll deploy from fresh. At that same time we can make any last 2.0 migration file changes.


DanVanAtta:
Ignoring now the rest of the codeclimate warnings. The README file needs updates to be shorter/more concise, should focus on that first before fixing style issues. (IE: fixing style issues now may be more churn than benefit)

infrastructure/ansible/roles/http_server/defaults/main.yml
```diff
@@ -14,7 +14,6 @@ http_server_db_port: "5432"
 
 http_server_db_name: "{{ lobby_db_name }}"
 http_server_db_user: "{{ lobby_db_user }}"
-http_server_db_password: "{{ lobby_db_password }}"
 
-## TODO: encrypt all of the below variables
-http_server_bcrypt_salt: $2a$10$IhIXWg4HkQRWrZqjj9kV0u
+create_issues_github_api_token: test
```
RoiEXLab: Test?

infrastructure/create_secret
```diff
@@ -0,0 +1,42 @@
...
+  echo "This script expects a second file to contain the ansible vault password"
+  echo "named 'vault_password'"
+  exit 1
+}
+
+variable_name=$1
+
+set -eu
+
+if [ ! -f "secret" ]; then
```
RoiEXLab: Hmm, not sure if I like the pattern of relying on some specific files to exist.
In my opinion we should be able to pass a filename and/or fall back to a default value.

infrastructure/README.md
```diff
@@ -1,26 +1,56 @@
...
+on production, and also defined in "group_vars/vagrant.yml" to
+have another value when deploying to a local vagrant server.
+
+The inventory files should all have the same structure and
+define hostgroups that line up with the "group_var" file names.
+
+
+### Notes on variables
+
+* do not use `vars` folder
```
RoiEXLab: Maybe a stupid question, but what exactly is the vars folder, and why shouldn't it be used?

DanVanAtta:
> Thoughts about sticking to the codeclimate styleguide for markdown?

I commented on the violations, to recap:
- some are existing, so a bit much to fix them all up
- the README needs more fundamental work, rewritten & made concise. No point to fix violations if the text changes completely on the next rev.

infrastructure/README.md
DanVanAtta: Role vars are number 15 out of the 22 levels of precedence and places where variables can be defined: https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#ansible-variable-precedence

Why not use vars?
It's a similar problem to using an overly strong CSS selector, when trying to override it, it can be really difficult and artificial to find a selector with a stronger strength.

In our case we should not need vars, just use defaults; essentially favoring the weakest precedence to give the most flexibility for overrides. 

'vars' are probably best when re-using an existing role via some sort of import, even then I'm not sure if that's really a cleaner why to go compared to using a group_var maybe, can depend on the context. Regardless, the point is that we're limiting where you can define variables to 3 or 4 locations to simplify that list of 22 down to just a few.

infrastructure/create_secret
DanVanAtta: The files are temporary and are added to .gitignore to avoid accidental checkin. The file is there so you don't type the command on prompt (avoid accidental typo) nor have to add the secret via command line arg (insecure).

Using 'create_secret' is one time, the vault password file and secret file should be removed afterwards. See the README update for the command that is being replaced, this is really a wrapper around a one-liner.

Let me know if that makes sense.

infrastructure/ansible/roles/http_server/defaults/main.yml
DanVanAtta: Note the value in the configuration.yml used for local development and automated tests:
- https://github.com/triplea-game/triplea/blob/master/http-server/configuration.yml#L6

Note that the value is specific, used as a magic flag to not send API requests if the value is equal to test:
- https://github.com/triplea-game/triplea/blob/master/http-clients/src/main/java/org/triplea/http/client/github/issues/GithubIssueClient.java
- https://github.com/triplea-game/triplea/blob/master/http-server/src/main/java/org/triplea/server/error/reporting/CreateIssueStrategy.java#L41

With this PR, note that the vagrant/prerelease/production configuration.yml is parameterized for this value:
- https://github.com/triplea-game/triplea/pull/5852/files#diff-448e030bac09cb5b6eaf84f89d4b5d54R2

The value we have here is the default, so we will use that unless there is an override present. That is intentional so that 'vagrant' will work without any overrides.

Then finally note the 'linode' group_var override that will override this for linode environments:
- https://github.com/triplea-game/triplea/pull/5852/files#diff-8a634d36d006538798cff653e9e28b53R18

Finally, we update inventory to define prerelease and production as children of 'linode' so that the override will take effect for those environments:
- https://github.com/triplea-game/triplea/pull/5852/files#diff-b3216fe9376fa99bf180b01fe4bd1a55R16
- https://github.com/triplea-game/triplea/pull/5852/files#diff-6f7a9a9749d18379ef4b4f7cf267c018R19

RoiEXLab: Thanks for clarifying 👍

infrastructure/create_secret
RoiEXLab: Are all files temporary?
If they are, why not create files in `/tmp` and reference them in other parts of the script?
This way there's no chance it can be checked in accidentally.

Either way, the current system "enforces" some arbitrary filenames, that might conflict with pre-existing files for some reason (when used). The typo argument is valid, but at least I'm personally using tab-autcompletion for files to prevent exactly that.
I just think it's more natural to be explicit about the files your specifying this as input.
To me this feels like if grep would only accept input files called "input" in the current directory, and you'd have to copy files or use symlinks to run it on any other file.

Different thought: What about using STDIN for the secret? Avoids the problem with the secret being stored in bash history, in if you're really paranoid you can also use `read -s SECRET` to read from the terminal in a hidden way and store them into `$SECRET`.

DanVanAtta: > Are all files temporary?

Mostly temporary, we should not be keeping secrets, either the vault password or unencrypted secret files on our systems for very long. Those values represent a security concern/vulnerability and should be deleted in a reasonable time frame.

> If they are, why not create files in /tmp and reference them in other parts of the script?

Being in `/tmp` might help them be cleaned up, but probably better for it to be explicit as we should know what we're doing when dealing with secrets. 

The vault-password file is used by ansible-playbook commands proper, having it be in a known location is preferable. 

Second, '/tmp' tends to have more lenient read permissions, isn't it a bit more locked down to have it in a folder owned by only the current user?

Re: typo -- I'm concerned about typo's using STDIN. 

> I just think it's more natural to be explicit about the files your specifying this as input.

The issue I see with this is then parameter ordering becomes extremely significant. If we go the args flags, I seriously raise the prospect we are over-engineering this.

> To me this feels like if grep would only accept input files called "input" in the current directory, and you'd have to copy files or use symlinks to run it on any other file.

The idea is you encrypt one secret at a time. Update 'secret' with the secret, pass the parameter name as a variable to this script, copy the output into a '.yml' file, repeate until you encrypted all the variables you want, then delete the 'secrets' file. Once done testing, delete also the 'vault-password' file.

RoiEXLab: Ok, I get why tmp files are maybe not the right way to go. Tmp has special permissions so that only the owner of a file can actually delete it, I guess if adjusting the read permission, we probably could keep the file private, blocked for other users. But this is just my assumption at this point.

I personally think that the read command is very intriguing; there's no need for a magic file to exist, and if you're still concerned about potential typos, you could still pipe a magic file into the script so it stays the same.

Let me know what you think.
If you disagree, feel free to merge but I'm convinced this is definitely an option worth considering

PR #5854, opened 2020-01-19
Enable lower_snake_case validation of SQL and squash migrations

DanVanAtta:
1. Squash SQL migrations files together
2. Enable validation to disallow camel casing in SQL migrations
3. Fix camelCased names in SQL migration file

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/lobby-db/src/main/resources/db/migration/V2.00.00__lobby.sql
+++ b/lobby-db/src/main/resources/db/migration/V2.00.00__lobby.sql
@@ -1,4 +1,4 @@
-comment on database lobby_db is 'The Database of the TripleA Lobby';
+comment on database lobby_db is 'Database used by lobby';
 
 create table bad_word
 (
@@ -32,7 +32,7 @@ values (1, 'ADMIN'),     -- user can add/remove admins and add/remove moderators
        (2, 'MODERATOR'), -- user has boot/ban privileges
        (3, 'PLAYER'),    -- standard registered user
        (4, 'ANONYMOUS'), -- users that are not registered, they do not have an entry in lobby_user table
-       (5, 'HOST'); -- AKA LobbyWatcher, special connection for hosts to send game updates to lobby
+       (5, 'HOST'); -- AKA 'Lobby Watcher', special connection for hosts to send game updates to lobby
 
 create table lobby_user
 (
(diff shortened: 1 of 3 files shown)


lobby-db/src/main/resources/db/migration/V2.00.00__lobby.sql
```diff
@@ -49,7 +49,7 @@ alter table lobby_user
     owner to lobby_user;
 alter table lobby_user
     add constraint lobby_user_pass_check check (password IS NOT NULL OR bcrypt_password IS NOT NULL);
-comment on table lobby_user is 'Stores information about TripleA Lobby users.';
+comment on table lobby_user is 'Stores information about Triplea Lobby users.';
```
RoiEXLab: Not sure about that, but don't we usually use `TripleA` as default way to write it?

DanVanAtta: Yes, but it's also camel-cased too..

RoiEXLab: Not quite sure on how to reply to your statement.
I mean shouldn't we agree on some way or another?

DanVanAtta: I'm not sure i follow what kind of suggestion you would be proposing @RoiEXLab , could you walk me through what you might want to see?

The rule of looking for camel case in SQL files is pretty naive, just any lower case letter next to upper case. I don't think the ROI for special casing 'TripleA' to that is significant enough to worry about, and I would be concerned that we could see examples where 'TripleA' were used in a table or column name (in general, or let alone as a camel cased variant)

RoiEXLab: Ahh, that's the missing piece I was missing: There's a script that checks the casing in this file, which isn't smart enough to differenciate between regular syntax and strings, right?
In this case feel free to merge, I can't think of a good way to enhance the script to ignore strings. I mean for this particular case it might be sufficient to run the file through sed first and replace all strings with the empty string, but this could potentially cause problems with escape sequences

PR #5871, opened 2020-01-23
Enhance suicide functionality and deprecate old suicide options

ron-murhammer:
Addresses feature request: https://forums.triplea-game.org/topic/1579/suicide-munitions-unit-with-support

- Add new isSuicideOnAttack and isSuicideOnDefense properties to deprecate isSuicide and global property "Defending Suicide and Munition Units Do Not Fire"
- Suicide units will now fire either during first strike phase (subs) or normal phase rather than a separate hacked phase before first strike
- Suicide units can now receive and give support
- Fix bug where first strike units weren't actually receiving support even though the battle window showed it: https://forums.triplea-game.org/topic/1181/total-world-war-december-1941-3-0-0-6/456
- Implement remaining target limiting properties canNotTarget and canNotBeTargetBy: https://forums.triplea-game.org/topic/1603/cannottarget-not-working
- Lots of refactoring to allow all of these to be implemented and to better centralize a lot of the functionality

## Functional Changes

[] New map or map update
[x] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done

Tested suicide units on NML and Iron War. Tested revised and global to make sure regular battles and subs/destroyers still function properly. Ran a few all AI games.


game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
```diff
@@ -1752,19 +1753,45 @@ private void resetCanDieFromReachingMaxDamage() {
...
 
   private void setIsSuicide(final String s) {
-    isSuicide = getBool(s);
+    setIsSuicide(getBool(s));
   }
 
   private void setIsSuicide(final Boolean s) {
-    isSuicide = s;
+    setIsSuicideOnAttack(s);
+    setIsSuicideOnDefense(s && !Properties.getDefendingSuicideAndMunitionUnitsDoNotFire(getData()));
```
DanVanAtta: Why is this the condition for suicide on defense?
`!Properties.getDefendingSuicideAndMunitionUnitsDoNotFire(getData()));`

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
```diff
@@ -1752,19 +1753,45 @@ private void resetCanDieFromReachingMaxDamage() {
...
+    setIsFirstStrike(s);
```
DanVanAtta: Should first strike truly be a side effect of is suicide? Does this limit our options in case we want a unit that is just 'isSuicide'? Say someone has a non-attacking unit, or an attacking unit that is suicide, but not first strike, how could that be handled?

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
```diff
@@ -1752,19 +1753,45 @@ private void resetCanDieFromReachingMaxDamage() {
...
+  }
+
+  private void setIsSuicideOnAttack(final String s) {
```
DanVanAtta: Should the 's' convention be carried forward? Would it be more clear if the parameter were simply spelled out?

game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
```diff
@@ -493,8 +493,12 @@ private Matches() {}
...
   }
 
-  public static Predicate<Unit> unitIsSuicide() {
-    return unit -> UnitAttachment.get(unit.getType()).getIsSuicide();
+  public static Predicate<Unit> unitIsSuicideOnAttack() {
+    return unit -> UnitAttachment.get(unit.getType()).getIsSuicideOnAttack();
+  }
+
+  public static Predicate<Unit> unitIsSuicideOnDefense() {
+    return unit -> UnitAttachment.get(unit.getType()).getIsSuicideOnDefense();
```
DanVanAtta: Side-note, do you recall the rationale to not attach this method as a property of `Unit`? I suspect it would be pretty clearly better encapsulation.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
```diff
@@ -1752,19 +1753,45 @@ private void resetCanDieFromReachingMaxDamage() {
   }
 
   private void setIsSuicide(final String s) {
-    isSuicide = getBool(s);
+    setIsSuicide(getBool(s));
   }
 
   private void setIsSuicide(final Boolean s) {
```
DanVanAtta: Would it maybe be more clear and direct to have a 'isSucicideOnAttack' property? It seems to get at that property, one has to be pretty clever to set a combination of values.

I wonder as well if simply adding those two properties would make it easier for this update to be backward compatible. I'm very concerned that the property is not directly from the XML but is instead derived, it seems like it is quite complex compared to other properties that are usually 1:1.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1380,17 +1379,18 @@ private void fireNavalBombardment(final IDelegateBridge bridge) {
   }
 
   private void fireSuicideUnitsAttack() {
+    // TODO: Remove isSuicideOnAttack check and update to use canNotTarget/canNotBeTargetedBy
```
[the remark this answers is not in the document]
DanVanAtta: Kinda agree with the TODO climate warning unless we have a reference to who or what owns this task.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1404,35 +1404,33 @@ private void fireSuicideUnitsAttack() {
     if (suicideAttackers.isEmpty() || attackedDefenders.isEmpty()) {
       return;
     }
-    final boolean canReturnFire = (!isSuicideAndMunitionCasualtiesRestricted());
-    final List<Unit> allEnemyUnitsAliveOrWaitingToDie = new ArrayList<>();
-    allEnemyUnitsAliveOrWaitingToDie.addAll(defendingUnits);
+    final boolean canReturnFire = (!Properties.getSuicideAndMunitionCasualtiesRestricted(gameData));
```
DanVanAtta: Should `canReturnFire` be a method of `Properties`? Would that be more direct and re-usable? Any concern that we might duplicate or confuse this check somewhere else?

game-core/src/test/java/games/strategy/triplea/delegate/DiceRollTest.java
```diff
@@ -63,19 +63,23 @@ void testSimple() {
...
     // infantry attacks
     final DiceRoll roll3 =
-        DiceRoll.rollDice(infantry, false, russians, bridge, battle, "", territoryEffects, null);
+        DiceRoll.rollDice(
+            infantry, false, russians, bridge, battle, "", territoryEffects, null, infantry);
     assertThat(roll3.getHits(), is(1));
     // infantry attack
     final DiceRoll roll4 =
-        DiceRoll.rollDice(infantry, false, russians, bridge, battle, "", territoryEffects, null);
+        DiceRoll.rollDice(
```
DanVanAtta: Would a method overload be cleaner to avoid passing null parameter when the last parameter is unused?

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
DanVanAtta: Second, if 'isFirstStrike' is set, would this conflict and/or introduce a race condition for which property is actually set?
EG: isSuicide = true & isFirstStrike = false (or vice versa)

If this also gets reset, will that turn off firstStrike for units that have it? Are we guaranteed to have it be turned back on?

I'm feeling pretty convinced this property should be more independent of others, and allow 'isSuicide' to be more legacy and/or operate mostly as it used to.

DanVanAtta:
@ron-murhammer some concerns here:
1. not sure if the existing changes to properties are the best direction, that perhaps we should keep properties decoupled and favor keeping existing properties working as they were. 
2. scale of review is getting to be really large, potentially well beyond anything that can be reasonably and well reviewed.  Could lead back to a frustrating number of round trips for both you and reviewer, particularly with a mixture of larger scale review items and many minor items.

What's the end-point of updates to this branch? Is this changeset already too large and perhaps should be a draft and broken up into smaller pieces? What is the test plan for these updates, will unit and integration tests land to verify the updates or will we be working with manual testing?

DanVanAtta:
Another question, what is the final spec that you are coding @ron-murhammer , are you implementing the initial suggestion as proposed by @FrostionAAA in the mentioned forum thread, or a variant of that?

game-core/src/test/java/games/strategy/triplea/delegate/DiceRollTest.java
ron-murhammer: Possibly. I may add some overloads to this but wanted to ensure I updated all the necessary places that needed updated first and not miss any.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
ron-murhammer: Not really following you here. So essentially what this does is add 2 new unit options `isSuicideOnAttack` and `isSuicideOnDefense` as often you don't want units to suicide in both cases (think I don't want missiles to suicide on defense, etc). This then deprecates `isSuicide` which for backwards compatibility now just sets these new properties. Eventually after this release and if all maps using `isSuicide` are updated then we could remove it all together.

game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
ron-murhammer: All the unit properties are part of UnitAttachment. This is really just the overall structure of the XML and code base.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
ron-murhammer: I'd rather keep consistency for now. Easier to just update them all at once. Though honestly most of these are so simple that the parameter name isn't a huge deal.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
ron-murhammer: In theory no but this is for backwards compatibility because that is how suicide units work currently. Moving forward maps shouldn't use isSuicide and set either or both of the 2 new options.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
ron-murhammer: Because this is a hack property that was added to make it so all suicide units in a map wouldn't suicide on defense. This is essentially deprecated now and just used for backwards compatibility.

ron-murhammer:
@DanVanAtta So I think I responded to all your code comments. The code here should be mostly completed now though I still need to test a bunch of things.

1. not sure if the existing changes to properties are the best direction, that perhaps we should keep properties decoupled and favor keeping existing properties working as they were.
- The goal is to deprecate isSuicide and the global property "Defending Suicide and Munition Units Do Not Fire". These just set the new properties for backwards compatibility.

2. scale of review is getting to be really large, potentially well beyond anything that can be reasonably and well reviewed. Could lead back to a frustrating number of round trips for both you and reviewer, particularly with a mixture of larger scale review items and many minor items.
- Yeah, this is a big one but it required some significant changes to the core battle phase code in order to consolidate how units are grouped together. I realize this could have a long review cycle and hope that for the most part there are a number of code improvements already and most other things can be pushed to follow ups.

What's the end-point of updates to this branch? Is this changeset already too large and perhaps should be a draft and broken up into smaller pieces? What is the test plan for these updates, will unit and integration tests land to verify the updates or will we be working with manual testing?
- I think this should be pretty much done though needs more testing. This would be very difficult to break into pieces and really until I started working through it, wasn't clear exactly all the areas that needed updated.

Another question, what is the final spec that you are coding @ron-murhammer , are you implementing the initial suggestion as proposed by @FrostionAAA in the mentioned forum thread, or a variant of that?
- I updated the initial PR post to capture everything being done here. It is primarily based on that thread and discussion there though also a related bug fix and implementing the final isSub properties canNotTarget and canNotBeTargetedBy.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
DanVanAtta: I'm not entirely sure it's consistent, I think there is argument for it being copy/pasta. Otherwise, we're left with just a single letter variable name representing the type, 'String'. Or does the 's' represent 'suicide', or 'isSuicide'? Having to decode what a variable single character represents is a small burden, but a burden nonetheless, that makes it harder to grok/read code.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
DanVanAtta: I think this is where map versions would be incredibly useful. We could decouple this based on specified map version. 

Have you considered making isSuicde have attribute values? eg: 'suicideOn=attack', 'suicideOn=all'

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
DanVanAtta: Would you mind adding a comment to the code to clarify point? I'm pretty sure others would have the same question casually reading this and they would not have the benefit of this PR review conversation for reference.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
DanVanAtta: Essentially I was suggesting to leave the 'isSuicide' property unchanged and instead just add the new properties. It's more clear to me now that you're removing the old 'isSuicide' property.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
```diff
@@ -689,10 +690,18 @@ private void setCanNotTarget(final Set<UnitType> value) {
 
   public Set<UnitType> getCanNotTarget() {
     if (canNotTarget == null) {
-      canNotTarget =
-          new HashSet<>(
-              CollectionUtils.getMatches(
-                  getData().getUnitTypeList().getAllUnitTypes(), Matches.unitTypeIsAir()));
+      if (isSuicideOnAttack && isFirstStrike) { // Support isSuicide shortcut
```
DanVanAtta: nit, not sure a casual reader will understand the comment to 'support isSuicide "shortcut"', they may wonder how it is a shortcut, and what 'isSuicide' is necessarily even referring to.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
```diff
@@ -689,10 +690,18 @@ private void setCanNotTarget(final Set<UnitType> value) {
...
+        canNotTarget =
+            new HashSet<>(
+                CollectionUtils.getMatches(
+                    getData().getUnitTypeList().getAllUnitTypes(),
+                    Matches.unitTypeIsSuicideOnAttack().or(Matches.unitTypeIsSuicideOnDefense())));
```
DanVanAtta: nit, it's just the matches type that varies depending on the if condition, seems though the if/else could be simplified.

Beyond the code low-level cleanup, why is the 'isSub' and 'isSuicide' properties opposites of one another? IE: why is if 'isSuicide' is false, then 'isSub' is true?

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
ron-murhammer: Added. Also marked these as deprecated.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
DanVanAtta: > "Easier to just update them all at once."

I have trouble agreeing as there are many more instances 'to fix' now. It's not clear they would all get the same fix.

DanVanAtta: Actually, backing up a bit, the private setter is really odd. Perhaps would it be easier if we just inlined this? Then when invoked via method reference, it would imply that we have more going on then a simple setter. Someone reading the code would be no worse off seeing any of the inconsistent cases, either they'd see the simple setter inline, or they would see a method reference that is either more complex or is a simple setter (which we could later come back to clean up).

ron-murhammer: I believe the double setters (1 string, 1 actual type) has to do with the XML parsing for attachments. The attachment classes in general are very rough and primarily there for to hold the values from the map XML parsing.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
ron-murhammer: Ok. Do you have an alternative suggestion?

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
ron-murhammer: Possibly though I was concerned with it being harder to read if too much of it was inlined.

They aren't true opposites but because this is for supporting the deprecated properties, current maps never have isSub and isSuicide on the same unit as they didn't work together.

game-core/src/test/java/games/strategy/triplea/delegate/DiceRollTest.java
ron-murhammer: Updated.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
DanVanAtta: I don't think that would qualify as too much inlining, it helps a lot as it makes clear what is different between the two cases. You can assign to an intermediary variable if the inline syntax feels awkward.

> They aren't true opposites...

Point taken, a casual reader would be very confused why we can assume with one condition being false that we fall into this other (seemingly very unrelated) case.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
DanVanAtta: Some options:
- remove the comment
- expound on the comment and provide links/references for context
- bake the 'isSuicide' into matches. 
- Make 'CanTarget' a property of "UnitType", add explanations there

The last is probably my preferred approach followed by adding a specific matcher. I like the last approach as it has the best encapsulation and is the most easily tested (this section of code would be great to have tests for, it would make review easier and would lend more confidence this is working as intended).

Is there a legit timeline of when we can retrofit all maps to not have 'isSuicide'? IMHO the bulk map update experiences we have had caused far more pain than they are worth, and we do not even have ownership of all maps, and there was a lot of pain there. Removing map features I don't think is feasible and is part of why I think it's important we add versioning to map syntax/formats. Regardless of your viewpoint there, having a 'Matches.isSuicide' is independent of it is an actual unit property and the matches is, like here, a composite property.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
DanVanAtta: > I believe the double setters (1 string, 1 actual type) has to do with the XML parsing for attachments. The attachment classes in general are very rough and primarily there for to hold the values from the map XML parsing.

I believe I understood that; My point is the indirection here does not seem useful, inlining these methods would reduce the indirection.

Re: variable name, most of the examples have been touched/added here and the new setter methods could have different variable names altogether. I don't really trust that we ever will come back and fix them.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import lombok.Getter;
+import lombok.Setter;
+import org.triplea.java.collections.CollectionUtils;
+
+/** Group of firing units and their targets. */
+@Getter
+@Setter
```
DanVanAtta: It does not look like this setter is used. Can it be removed? Is this class actually immutable? If not, it looks like it would not take too much to make it so, is that feasible? Is there a reason for mutability?

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import lombok.Getter;
+import lombok.Setter;
+import org.triplea.java.collections.CollectionUtils;
+
+/** Group of firing units and their targets. */
+@Getter
```
DanVanAtta: It looks like the getter here is used only internally. Unless I missed a usage, do you have any thoughts on removing the getter or restricting the visibility of the get method? 

Looking at the class, and the class properties, it raises a code smell if we return the internal data, it makes it seem as if the class were incompletely providing operations on the data members. IE: if this class provides operations on two internal sets, then what's the point to return the internal data unless this class is incomplete in doing so.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
```diff
@@ -0,0 +1,94 @@
...
+@Setter
+public class TargetGroup {
+
+  private final Set<UnitType> firingUnitTypes;
```
DanVanAtta: It looks like this is always a singleton set, do I see that correctly? Why store this as a set if it always has size == 1?

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
```diff
@@ -0,0 +1,94 @@
...
+
+  public Collection<Unit> getFiringUnits(final Collection<Unit> units) {
+    return CollectionUtils.getMatches(units, Matches.unitIsOfTypes(firingUnitTypes));
+  }
+
+  public Collection<Unit> getTargetUnits(final Collection<Unit> units) {
+    return CollectionUtils.getMatches(units, Matches.unitIsOfTypes(targetUnitTypes));
+  }
+
+  public String getMessage(final GamePlayer player) {
```
DanVanAtta: Nit, this method can be marked as static. It also seems to not really belong to this class as well (knowing it's static helps highlight that). Any thoughts to simply inline this to its single usage and avoid the indirection?

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
```diff
@@ -0,0 +1,94 @@
...
+    return player.getName() + " fire, ";
+  }
+
+  /**
+   * Find all target groupings for firing units and enemy units based on canNotTarget and
+   * canNotBeTargetedBy attributes. Also takes into account if any destroyers are present which
+   * cancel canNotBeTargetedBy. Sort all the target groups so most restrictive appear first.
+   */
+  public static List<TargetGroup> newTargetGroups(
```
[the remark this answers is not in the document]
DanVanAtta: A test of this method would help review and add confidence this is completely working as expected. Such a test would give design feedback as well and potentially help drive further break down in responsibility to make it easier to test and understand.

For example, extracting 'findTargets' to its own module would be a first simplification that would help decompose the problem, ie: a targetFinder that is perhaps injected into this class.

game-core/src/test/java/games/strategy/triplea/delegate/DiceRollTest.java
```diff
@@ -63,19 +63,19 @@ void testSimple() {
         .thenAnswer(withValues(1)); // infantry attack does not hit at 1 (0 based)
     // infantry defends
     final DiceRoll roll =
-        DiceRoll.rollDice(infantry, true, russians, bridge, battle, "", territoryEffects, null);
+        DiceRoll.rollDice(infantry, true, russians, bridge, battle, territoryEffects);
```
DanVanAtta: :+1:   Good to see unused/placeholder args falling away.

game-core/src/test/java/games/strategy/triplea/delegate/WW2V3Year41Test.java
```diff
@@ -1143,12 +1129,12 @@ void testAttackDestroyerAndSubsAgainstSubAndDestroyer() {
     final List<String> steps = battle.determineStepStrings(true);
     assertEquals(
         List.of(
-                attacker + SUBS_FIRE,
-                defender + SELECT_SUB_CASUALTIES,
+                attacker + FIRST_STRIKE_UNITS_FIRE,
```
DanVanAtta: :+1:   I like quite a bit we are making the ordering more generic and list 'first strike units' rather than assuming any/all first strike units are subs.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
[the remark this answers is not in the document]
DanVanAtta: Why is this TODO not being done now?

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
DanVanAtta: Having a method `Properties.canReturnFire(gameData)` seems like it would much simpler to understand here and would be much better encapsulation. Currently `MustFightBattle` is bloated and knows about far too much, it should not necessarily be in change of interpretting game rules like this.

DanVanAtta: Noticing now as well we have the logic repeated: https://github.com/triplea-game/triplea/pull/5871/files/7e244f4d9c1c60f021bacdaf21c2606c914d4dbd#diff-009c223dfe7c716d07dc90f27899c933R1446, seems like an actual DRY violation where the idea of 'canReturnFire' is computed in multiple places, moving this logic would help with that problem.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1805,25 +1805,16 @@ private void defenderRetreatSubs(final IDelegateBridge bridge) {
...
-          .deadUnitNotification(battleId, attacker, deadUnits, dependentUnits);
-      bridge
-          .getDisplayChannelBroadcaster()
-          .deadUnitNotification(battleId, defender, deadUnits, dependentUnits);
-      remove(deadUnits, bridge, battleSite, null);
-    }
+    final List<Unit> deadUnits = new ArrayList<>();
+    deadUnits.addAll(CollectionUtils.getMatches(attackingUnits, Matches.unitIsSuicideOnAttack()));
+    deadUnits.addAll(CollectionUtils.getMatches(defendingUnits, Matches.unitIsSuicideOnDefense()));
+    bridge
```
DanVanAtta: This deadUnit notification only appears in one branch of the unmodified code, for a case of `isDefendingSuicideAndMunitionUnitsDoNotFire`, why should this be shown now? Is this a no-op perhaps, is there something subtle that is not obvious?

game-core/src/test/java/games/strategy/triplea/delegate/PacificTest.java
```diff
@@ -108,20 +108,44 @@ void testNonJapanAttack() {
...
-            infantryUs, true, americans, bridge, mock(IBattle.class), "", territoryEffects, null);
+            infantryUs,
+            true,
+            americans,
+            bridge,
+            mock(IBattle.class),
+            "",
+            territoryEffects,
+            null,
+            infantryUs);
```
DanVanAtta: It's striking we have the same parameter passed in twice to a method. IMHO it is a pretty significant API code smell. Perhaps an extract parameter object refactor would fix that?

IMO the existing API already has significant technical debt, I don't think that is fair reason to add more potential debt for someone else to attempt to fix. It's also very notable that a number of other parameters are already useless, seems like this API is already stretched.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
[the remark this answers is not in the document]
ron-murhammer: This was a temporary TODO that has already been removed...

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
ron-murhammer: I think you're missing the big picture here. This entire class and really all the attachment classes follow this pattern of methods. Now whether that is an ideal pattern or not we can discuss but that is really a much larger issue than this PR and I would prefer to have a separate issue for it.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
ron-murhammer: Uh, I think this was all removed. Are you looking at the latest diff and commits?

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
ron-murhammer: So the issue with that direction is that we really want to isolate and reduce where isSuicide is used as otherwise its confusing as whether the property is still being used throughout the delegates (which it isn't as its now just setting other properties). If I saw a matches isSuicide then I'm going to think its reading that property unless I go and specifically read that method. The idea here is to keep all the isSuicide functionality in the unit attachment so that it isn't leaked to the other classes.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
ron-murhammer: Updated.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
ron-murhammer: Updated to remove comment.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
ron-murhammer: Ah yes. Good catch. Originally I was using the setters but had reworked to avoid that.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
ron-murhammer: The getters are used in MustFightBattle to pull the information out.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
ron-murhammer: Nope. This holds all the unit types that have the same list of targets. They are added 1 by 1 as targets are checked.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
ron-murhammer: This is a bit of a forward looking method as the message ideally would be more detailed to include which unit types are firing or being targeted.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
ron-murhammer: I think you are looking at outdated code here.

game-core/src/test/java/games/strategy/triplea/delegate/PacificTest.java
ron-murhammer: I think you are looking at outdated code here as well.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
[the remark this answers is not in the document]
DanVanAtta: Cool, thanks for taking care of, did not properly notice the 'outdated' tag.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
DanVanAtta: I don't think I am necessarily:
- At 4300+ lines, this class is a monster. At such scale, it's arguable there are any patterns, it's a task just to grok it to see if anything is consistent or what it even contains.

This is a counter-example to the 'pattern':

```
        .put(
            "transportCost",
            MutableProperty.ofMapper(
                DefaultAttachment::getInt,
                this::setTransportCost,
                this::getTransportCost,
                () -> -1))
```


To help avoid unnecessary indirection I'm suggesting to follow the example of the counter-example and inline the setter/getters if they are simple.

Second, the class violates our preferred method ordering with arbitrarily located getter/setters that ought to be below where they are used.

Third, if not explicitly listed, we have a preference to spell out variables and not use single letter variables. I'm not positive if the 's' makes sense or is the right prefix, and seeing `s && ...` is not as easy to understand as it could be to understand.

I can understand pushback on the first and second points, though I don't think either of us should make that the hill we choose to die on, the third point seems far less reasonable to ignore as it's more about the care of this update as it is adding new code that should really not be adding at least that element of technical debt.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
DanVanAtta: Not sure, seems to indicate I need to start over on the entire review  :disappointed:

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
DanVanAtta: A TODO tagged with the forum discussion would have been ideal to have communicated that. Do you have an expectation when that would land (near-future, far-future?)

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
DanVanAtta: I see that `getFiringUnits` and `getTargetUnits` are used in `MustFightBattle`, though the only usages of the generated getters are in `TargetGroup`. Can you point me to where they are in `MustFightBattle`, am I looking at the wrong diff or something?

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  public String getMessage(final GamePlayer player) {
+    return player.getName() + " fire, ";
+  }
+
+  /**
+   * Find all target groupings for firing units and enemy units based on canNotTarget and
+   * canNotBeTargetedBy attributes. Also takes into account if any destroyers are present which
+   * cancel canNotBeTargetedBy. Sort all the target groups so most restrictive appear first.
```
DanVanAtta: Could you clarify or define 'restrictive' in this context? It's not clear to me from this context what that would exactly mean.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
```diff
@@ -0,0 +1,92 @@
...
+        if (firingGroup.getTargetUnitTypes().equals(targets)) {
+          firingGroup.getFiringUnitTypes().add(unitType);
+          isAdded = true;
+          break;
+        }
+      }
+      if (!isAdded) {
+        firingGroups.add(new TargetGroup(unitType, targets));
+      }
+    }
```
DanVanAtta: Have you considered re-writing this with a stream+map+collect to remove the intermediary variables and the 'added' local tracking variable? Any thoughts to adding an explanatory comment to help indicate what operation we are doing here?

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
```diff
@@ -0,0 +1,92 @@
...
+    return sortFiringGroups(firingGroups);
+  }
+
+  private static Set<UnitType> findTargets(
+      final UnitType unitType, final Set<UnitType> unitTypes, final Set<UnitType> enemyUnitTypes) {
+    final Set<UnitType> targets = new HashSet<>(enemyUnitTypes);
+    targets.removeAll(UnitAttachment.get(unitType).getCanNotTarget());
+    if (!unitTypes.stream().anyMatch(Matches.unitTypeIsDestroyer())) {
```
DanVanAtta: nit, `noneMatch` should be able to replace the `!...anyMatch()` and help avoid the negation

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
```diff
@@ -1751,20 +1756,54 @@ private void resetCanDieFromReachingMaxDamage() {
...
 
+  @Deprecated
   private void setIsSuicide(final Boolean s) {
-    isSuicide = s;
+    setIsSuicideOnAttack(s);
+    // Global property controlled whether isSuicide units would suicide on defense
+    setIsSuicideOnDefense(s && !Properties.getDefendingSuicideAndMunitionUnitsDoNotFire(getData()));
+    setIsFirstStrike(s);
+    if (s) {
+      canNotTarget = null;
```
DanVanAtta: The set to null might be potentially dangerous / NPE if `setCanNotTarget` is invoked after `setIsSuicide`, noting on line 683:

```
  private void setCanNotTarget(final String value) throws GameParseException {
    final String[] s = splitOnColon(value);
    for (final String u : s) {
      final UnitType ut = getData().getUnitTypeList().getUnitType(u);
      if (ut == null) {
        throw new GameParseException("canNotTarget: no such unit type: " + u + thisErrorMsg());
      }
683:      canNotTarget.add(ut);
    }
  }
```

Would that be a potential problem?

DanVanAtta: Second, the set to null seems to rely that any access to `canNotTarget` will go through a `get` method which has the side effect of also setting `canNotTarget`. Could this be fixed and/or be made more explicit by doing an assignment here of that value? (this seems kinda hacky, but it perhaps it makes things explicit at least):

```
if(s) {
  canNotTarget = getCanNotTarget();
}
```

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
ron-murhammer: I'd actually argue the larger the class the more consistency matters as its harder to understand if you see something being done multiple different ways. 

That being said, I misunderstood what you were suggesting about in lining. I'm open to using one of the helper mappers to reduce which setter/getter/reseter are needed. Updated this.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
ron-murhammer: Ah I see. Yeah, I reduced the visibility of the getters.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
ron-murhammer: Pushed it in line out of this class for now.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
ron-murhammer: Updated.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
ron-murhammer: Open to it. If you want to suggest how you'd do it with streams then I'm glad to take a look. I didn't see an easy way to do it that was easy to understand.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
ron-murhammer: Good catch. Updated.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
ron-murhammer: setCanNotTarget and setIsSuicide should never be used in the same map. setIsSuicide is just for existing map compatibility.

That's correct. The reason for this is you can't set the canNotTarget until all the XML unit attachments are parsed (as you need to read some of the unit properties) so has to happen essentially after all unit parsing is done. This was the easiest way I could think to do it and the only way to access it is through the getter (null is essentially being used as a place holder to flag that canNotTarget needs to still be calculated). This is a bit hacky but without adding some sort of placeholder state or extra parsing steps, I couldn't think of a better way to do it.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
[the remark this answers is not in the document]
ron-murhammer: Refactored this a bit to break it up and rename some variables. I would eventually like to add unit tests to this since it is a pretty isolated functionality that should be easy to test. Not sure I'll have the time to do that in the next few days though so might be a better follow up.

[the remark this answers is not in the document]
ron-murhammer: Add unit tests.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
DanVanAtta: I understand that `setCanNotTarget` and `setIsSuicide` are now being considered incompatible settings. Will the failure mode when that is done be a NPE? I suspect eventually, on some map, we'll see the two settings be combined and the NPE would be an unhelpful error message.

If we could get it so that the system fails with an exact error message (EG: unsupported option combination) and tells the user what/where, that would be quite valuable and we would handling this situation well enough.  

Could we possibly achieve that by adding a null-guard in the 'setCanNotTarget' method and then adding a post-processing check if the two settings have been used together? Second question, do we really need to set `canNotTarget` to null? If we can avoid the null, then it may simplify the need for null-checks and perhaps we could have the two options exist with each other, but perhaps the behavior would be non-deterministic.

ron-murhammer: Yeah, you'll get a NPE pointing to that method which will make it pretty obvious what the issue is. I don't expect it to ever happen and it would only be a map maker which would see it and probably just post on the forum. We can add checks to try to check for the null and return a better message just wondering if we are over engineering something that I don't think will ever happen.

You could avoid setting the null by setting some other say boolean field that then is checked instead of checking for the null in the getter though you are then adding additional temporary fields into a class that already has a lot of members variables. Not sure if that is really better or worse but was the other option I considered.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
DanVanAtta:
```
  private static void addToTargetGroups(
      final UnitType unitType, final Set<UnitType> targets, final List<TargetGroup> targetGroups) {
    targetGroups.stream()
        .filter(targetGroup -> targetGroup.getTargetUnitTypes().equals(targets))
        .findAny()
        .ifPresentOrElse(
            targetGroup -> targetGroup.getFiringUnitTypes().add(unitType),
            () -> targetGroups.add(new TargetGroup(unitType, targets)));
  }
```

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
DanVanAtta: Personally I would skip the generated getters altogether for more directness. It also makes it clear at the call site what we are doing and a reader would not have to verify where the getter is and if it could be overloaded.

Good enough though marking the getter private to reduce its scope. :+1:

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
DanVanAtta: Actually.. maybe not, taking another look..

DanVanAtta: Okay, didn't take long, yeah, I think the above should be equivalent.

DanVanAtta:
Summarizing:

Two open items:
* (minor/syntax): https://github.com/triplea-game/triplea/pull/5871#discussion_r373802347
* (important/error handling): https://github.com/triplea-game/triplea/pull/5871#discussion_r373801817

Beyond those two, I think we are pretty close to good. I'd like to take another pass to check out the logic in further detail.

DanVanAtta:
Ah, I'm forgetting, I have not checked out the added unit test (yay!) yet. If that looks good and the logic detailed double-check all looks good, then would be truly down to just the two items noted above.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
DanVanAtta: It's always surprising the data combinations/states that eventually wind up happening. Realistically we could maybe see it if a map were being retrofitted, or an existing map copied and then updated to be something else and the map maker missed an occurrence of the extra option.

Here's perhaps an idea:
- check for a null value of `canNotTarget` in the setter, rather than NPE, throw an illegal state at the beginning of the setter. That would provide a way to give a direct message to indicate that an incompatible setting option was used.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
ron-murhammer: Yep that works. Updated.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
ron-murhammer: Added null check and custom exception message.

game-core/src/test/java/games/strategy/triplea/delegate/battle/TargetGroupTest.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  @Test
+  void testNewTargetGroupsForOneUnitAndOneEnemyUnit() {
+    final GamePlayer germany = GameDataTestUtil.germany(twwGameData);
+    final List<Unit> units = GameDataTestUtil.germanInfantry(twwGameData).create(1, germany);
+    final GamePlayer britain = GameDataTestUtil.britain(twwGameData);
+    final List<Unit> enemyUnits = GameDataTestUtil.britishInfantry(twwGameData).create(1, britain);
+    final List<TargetGroup> result = TargetGroup.newTargetGroups(units, enemyUnits);
+    assertEquals(1, result.size());
```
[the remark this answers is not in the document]
DanVanAtta: FWIW, if it makes the static imports any easier, I keep a file around with the imports below, then use clean-up imports to remove any that I do not use when done writing a test:


```
import static org.hamcrest.collection.IsCollectionWithSize.hasSize;
import static org.junit.jupiter.api.Assertions.assertThrows;
import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
import static com.github.npathai.hamcrestopt.OptionalMatchers.isEmpty;
import static com.github.npathai.hamcrestopt.OptionalMatchers.isPresent;
import static com.github.npathai.hamcrestopt.OptionalMatchers.isPresentAndIs;
import static org.hamcrest.MatcherAssert.assertThat;
import static org.hamcrest.Matchers.is;
import static org.hamcrest.Matchers.not;
import static org.hamcrest.Matchers.nullValue;
import static org.mockito.ArgumentMatchers.any;
import static org.mockito.ArgumentMatchers.anyString;
import static org.mockito.ArgumentMatchers.eq;
import static org.mockito.Mockito.never;
import static org.mockito.Mockito.verify;
import static org.mockito.Mockito.when;
import org.hamcrest.text.IsEmptyString;
```

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
```diff
@@ -0,0 +1,90 @@
...
+                target -> !UnitAttachment.get(target).getCanNotBeTargetedBy().contains(unitType))
+            .collect(Collectors.toSet());
+  }
+
+  private static void addToTargetGroups(
+      final UnitType unitType, final Set<UnitType> targets, final List<TargetGroup> targetGroups) {
+    if (targets.isEmpty()) {
+      return;
+    }
+    targetGroups.stream()
```
DanVanAtta: This looks strange, could you check this @ron-murhammer ? It looks like `targetGroups` is always empty.  Was the iteration supposed to be on `targets`?

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
```diff
@@ -0,0 +1,90 @@
...
+    targets.removeAll(UnitAttachment.get(unitType).getCanNotTarget());
+    return unitTypes.stream().anyMatch(Matches.unitTypeIsDestroyer())
+        ? targets
+        : targets.stream()
+            .filter(
+                target -> !UnitAttachment.get(target).getCanNotBeTargetedBy().contains(unitType))
+            .collect(Collectors.toSet());
+  }
+
+  private static void addToTargetGroups(
```
DanVanAtta: nit, `addToTargetGroups` could be a bit more descriptive. Notably, I'm curious/not sure what exactly we are adding to target groups. I'm not honestly 100% sure how to describe what we're adding to target groups here, sorry to not have good suggestions for you on-hand.

Second, bit of a side note, if you're able to see a way, would be nice as well to be able to return a result rather than modifying the input param `targetGroups`.  In which case up on line 55 we could have something like:
```
targetGroups.addEligibleTargets(unitType, targets);
```

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  /**
+   * Find all target groupings for firing units and enemy units based on canNotTarget and
+   * canNotBeTargetedBy attributes. Also takes into account if any destroyers are present which
+   * cancel canNotBeTargetedBy. Sort all the target groups so those with the least number of targets
+   * appear first.
+   */
+  public static List<TargetGroup> newTargetGroups(
+      final Collection<Unit> units, final Collection<Unit> enemyUnits) {
```
DanVanAtta: I see AI is tearing through this method quite frequently while simulating battles. Would it make sense to early return an empty list here if either of the input parameters are empty? I'm wondering if that might help performance a bit, particularly if unit stacks get to be very large, and secondly it might help a code maintainer grok this method to quickly know this method is effectively a no-op if either parameter is empty.

ron-murhammer: Uh we could though I don't expect it to have much performance impact (probably <1%). If you did some tests and saw differently then I'd definitely look to add that in.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
ron-murhammer: I can't think of any better name. Essentially its adding the unit type and targets to either an exist target group or adding a new target group.

A method call like that wouldn't be possible unless you made TargetGroups a separate class instead of a list.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
ron-murhammer: Its correct. It will be empty on the first call to this method but then future calls will be looking at existing TargetGroup in the list to see whether to add to that group or add a new TargetGroup.

game-core/src/test/java/games/strategy/triplea/delegate/battle/TargetGroupTest.java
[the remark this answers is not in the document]
ron-murhammer: Done.

game-core/src/test/java/games/strategy/triplea/delegate/battle/TargetGroupTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+  final GameData twwGameData = TestMapGameData.TWW.getGameData();
+
+  @Test
+  void testNewTargetGroupsWithEmptyUnitsAndEnemyUnits() {
+    final List<TargetGroup> result = TargetGroup.newTargetGroups(List.of(), List.of());
+    assertEquals(0, result.size());
+  }
+
+  @Test
```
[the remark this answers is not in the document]
ron-murhammer: Done.

game-core/src/test/java/games/strategy/triplea/delegate/battle/TargetGroupTest.java
```diff
@@ -0,0 +1,95 @@
...
+import org.junit.jupiter.api.Test;
+
+class TargetGroupTest extends AbstractDelegateTestCase {
+
+  final GameData twwGameData = TestMapGameData.TWW.getGameData();
+
+  @Test
+  void testNewTargetGroupsWithEmptyUnitsAndEnemyUnits() {
+    final List<TargetGroup> result = TargetGroup.newTargetGroups(List.of(), List.of());
+    assertEquals(0, result.size());
```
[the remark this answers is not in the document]
ron-murhammer: Done.

ron-murhammer:
@DanVanAtta Comments should be addressed. I will say that because most of the unit tests don't follow those conventions, it makes it much more painful as there aren't many examples and often I'll start from existing unit tests which then use different conventions. Starting to rethink whether those conventions are worth it if we aren't going to mass update existing unit tests.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
DanVanAtta: Ah, I see it now. It's a bit counter-intuitive for a loop or a stream to have an effect when the iterated item is empty.

It could be worthwhile IMO then, for explicitness/clarity, to special case this situation. IE:

```
if(targetGroups.isEmpty()) {
  targetGroiups.add(new TargetGroup(unitType, targets);
  return;
}

```

Though, I think it's now more clear to me how we could avoid the mutable 'in/out' parameter, IMHO it would be even better if we returned a value (and also the name of this method becomes a bit easier to refine), eg:

```

 private static Optional<TargetGroup>  findTargetInTargetGroups( final UnitType unitType, final Set<UnitType> targets, final List<TargetGroup> targetGroups) {
     return targetGroups.stream()
              .filter(targetGroup -> targetGroup.getTargetUnitTypes().equals(targets))
              .findAny()
}
```

Then the `for` loop where this is used becomes a bit more clear and the overall control-flow IMO I think is more evident:

```
    for (final UnitType unitType : unitTypes) {
      final Set<UnitType> targets = findTargets(unitType, unitTypes, enemyUnitTypes);
     targetGroups.add(
           findTargetInTargetGroups(unitType, targets, targetGroups).orElseGet(() -> new TargetGroup(unitType, targets)));
    }
```

WDYT?

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
DanVanAtta: Yeah, I don't think there would be a huge performance win, maybe if the stack sizes were very large. It's more an early termination to make it easier for a reader to see when the method is a no-op, similar to the if-guard you have in another method (line 74)
```
    if (targets.isEmpty()) {
      return;
    }
```

DanVanAtta:
> I will say that because most of the unit tests don't follow those conventions

Well, certainly almost none follow the `@DisplayName` nor the assertion comment convention. Those two conventions are to fix a pretty large problem. The 'problem' is too consistent to not make worse? So we'll have more tests that are going to be very hard to understand when we come back and do not have context? Please recall as well the two studies on unit testing that included TripleA where TripleA was rated very low for maintainability of tests, a big factor against it was the ease to understand tests. Also recall that test smells were inversely correlated to ease of development.


FWIW, there are actually more test files now that use hamcrest matchers than not:

```
dan@dan-desk:~/work/triplea$ find . -name "*.java" -type f | xargs grep -l "import static org.junit.jupiter.api.Assertions"  | wc -l
133

dan@dan-desk:~/work/triplea$ find . -name "*.java" -type f | xargs grep -l "import static org.hamcrest.MatcherAssert" | wc -l  
266
```

DanVanAtta:
side-note, the conventions doc is there so that everyone can/should read it and basically take what's written there as the practice to follow for any new or modified code (despite any other ways you see of how something is done already). It's saying, 'going forward, this is how you should decide to do this - and if you see examples otherwise, know that it is not the preferred practice.' 

Hopefully we can agree there is value in having such agreements and would agree that depite whether we really like them or not, it is the job of author and reviewer to spot places where we can follow agreed-on project conventions.

game-core/src/test/java/games/strategy/triplea/delegate/battle/TargetGroupTest.java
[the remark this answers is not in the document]
DanVanAtta: Cool, thanks, this comment:
```
   "Verify 2 target groups if sub and fighters attack sub and fighters as subs can only "
          + "target subs")
```
- was actually really key to understanding the rest of them, IMHO it helped quite a bit.  :+1:

DanVanAtta:
Test looks good.

One new comment to perhaps clarify the TargetGroup logic. Thank you for the clarification on how that worked, with that new understanding I should be able to trace through it more competently. I think Wednesday I'll get the chance. I don't think there would be any more feedback beyond the one comment, I'm expecting the trace-through to be good. 

If you address the comment @ron-murhammer , i'm happy doing the trace-through as a post-merge review item. Good work on this update on the whole, the feature is a good direction, and I do like you added the TargetGroup concept and also how you broke up the logic methods that support it.

DanVanAtta:
Couple points/updates here that have been nagging me once I've slept on this:
- Avoiding side-effects in the 'addTargets' method I think is actually subtly more important than I initially realized:
  - The empty stream iteration to add to the stream is an odd pattern that is not very typical
  - We've a convention to avoid mutating parameters: https://github.com/triplea-game/triplea/wiki/Java-Code-Conventions#do-not-mutate-parameters-avoid-out-and-avoid-inout-parameters-5489
  - Changing the method to return a value really improves the logic flow, otherwise the logic of the main for loop is split between methods, reducing the value of splitting the methods.

 - It would be great to go a bit deeper in the canNotTarget test and assert the exact unit properties that make it so a sub cannot be targetted. In this way if we change the XML to upgrade it or what not, the precondition will have been documented and we'll know if the properties just need to be updated or if there is a problem in the targetting logic. Using an assertion comment is probably going to be ideal to set up this precondition state.

I can't help but wonder if the canNotTarget updates would have been easier for everyone as well on its own branch. I think it highlights a bit that we are not very good about jumping on PRs to review them. It encourages larger PRs that take longer to review, compared to knowing if something is bite-sized and submitted then it will be merged in short order (and hence no need to add more functionality to it). Regardless, that seems to be water under the bridge. 

The two updates mentioned above IMO would be quite valuable for understandability of the target group code, and the update to test would go a long way to capturing context of why the test is working the way it does.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
ron-murhammer: I'm fairly certain that isn't equivalent as it would add the same TargetGroup to the list multiple times. Now you could probably change to an ordered set instead of a list and I think that would then work.

So I generally agree with trying to avoid mutable parameters but this alternative seems harder to understand as you really shouldn't be adding to targetGroups if you find identical targets and instead add to an existing TargetGroup.

So probably if you want mutable parameters then you could do something like:
```
Optional<TargetGroup> targetGroup = findTargetInTargetGroups(unitType, targets, targetGroups);
if (targetGroup.isPresent()) {
// add unit type to existing target group
} else {
// create and add new target group to list
}
```

Cernelius:
Probably you have already covered this matter, but be sure to force a stalemate if all remaining units cannot target each other and both attackers and defenders cannot retreat (but never any stalemate if the attacker can still retreat anything (theorically, you should oblige it to retreat)).

However, maybe you also want to expand to the case both cannot hit each other normally, but there are still some targeted (AA) strikes that can do so, instead.

Also I don't like the "cannot target" and "cannot be targeted" definition, as it seems to imply that it is the attacker that picks casualties. I would rather say "cannot hit" and "cannot be hit", as it is the privilege of the defender to, then, assign such hits, amongst the eligible units.

game-core/src/main/java/games/strategy/triplea/delegate/battle/TargetGroup.java
DanVanAtta: Perhaps being more verbose is okay. I do think we need a change here as this was not easy to grok. Let me please clarify what I think are the problems so that we can be on the same page:

1. (biggest): The pattern of streaming on an empty list to then append to that list is a bit novel and surprising. I had to double check that it even worked. This perhaps could be solved by special casing the empty list scenario and appending, otherwise do the find.

2. The break up of:
```
for loop
   do something
   do more stuff
done
```
Is not necessarily helpful. The 'do more stuff' piece does not take all of the logic or even a complete step out of the for loop. To understand the for loop, you really have to jump down two methods to see what side-effects occur. Having the 'do more stuff' function be functional could fix this as you would have different levels of abstraction, rather than part of the for loop simply moved elsewhere (indirection as opposed to abstraction). This could also be solved by simply inlining, but I think we want an abstraction, ie: a method call to calculate a result that is then used in the for loop.

3. The mutable in parameter, this I think is probably more a symptom of the second problem.

With that stated, I think we'll at least be on the same page. Also for consideration, I've been wondering if we want a multimap instead of a list. The multimap would take care for us of the adding a new element if none is there, otherwise we append to an existing list. That perhaps could simplify/remove some of the helper methods.

ron-murhammer: Updated. Not sure where else to take it. I think we generally agree on the drawbacks but they aren't necessarily easy to solve without adding in other drawbacks as the logic here is fairly complex (we aren't just making a CRUD app). We are also getting to the point of essentially debating the design of a few methods which while important, we've already taken a few passes on and this PR is pretty large and been open for a while. I think at some point we have to push forward and consider follow on PRs if there are some ideas on how to further improve it.

DanVanAtta: Generally agree, some quick comments:
- tactical design should be reasonable and not add debt unless there is a really strong trade-off to be made. Quick follow-ups for cleanup are not really a good thing.
- large PRs will tend to be open for a long time as they cannot be reviewed quickly. The size of a PR will also lead to more areas of feedback and it's less efficient to come back and re-review. Consider it to another extent, if someone wanted to game the review system, then to get less review, just open larger PRs. I think the rule of thumb is about 30 minutes per 200 lines of code (https://smartbear.com/learn/code-review/best-practices-for-peer-code-review/). Hence, at 600 lines, you're requiring a dedicated 90 minutes. Given our time limitations, that means at least multiple days for a first pass only. 

I think perhaps the better way to go would be more iterations, smaller changes, and decoupling updates. If for example the TargetGroup were all on its own, it probably could have been more efficiently reviewed and would not have blocked anything else. It is water under the bridge at this point though, and I also appreciate how much effort it takes to break up larger changes as I certainly have, multiple times, spent literally days breaking up 600-1000 lines PRs

DanVanAtta:
Travis cache issues killed the last build. 
I merged latest master to this branch, should pass build then.

PR #5874, opened 2020-01-24
Refactor and hover listener consolidate

DanVanAtta:
Series of refactors around MapPanel, recommend to review commit-by-commit.

A big update here is applying DRY principle to 'hover units'. Both TripleAFrame and MapPanel kept track of the units that are current mouse hovered. TripleAFrame did this to for 'i' key info purposes and MapPanel for unit highlight. The update removes the state tracking of current hovered units from TripleAFrame and instead TripleAFrame accesses this data from MapPanel.

Otherwise, MapPanel is moved to its own class and there are a series of changes to better encapsulate MapPanel data.

Ultimately this update tries to get a start on addressing some of the technical debt in MapPanel and TripleAFrame, working with unit highlight lead quickly to many debt problems..

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done


game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanelSmallView.java
```diff
@@ -1,14 +0,0 @@
-package games.strategy.triplea.ui.panels.map;
-
-import games.strategy.triplea.ui.mapdata.MapData;
-import games.strategy.ui.ImageScrollModel;
-import games.strategy.ui.ImageScrollerSmallView;
-import java.awt.Image;
-
-public class MapPanelSmallView extends ImageScrollerSmallView {
```
RoiEXLab: Ok, this change came unexpected 😮

game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
```diff
@@ -481,10 +480,7 @@ public void windowGainedFocus(final WindowEvent e) {
     tooltipManager = new MapUnitTooltipManager(mapPanel);
     mapPanel.addMapSelectionListener(mapSelectionListener);
     mapPanel.addMouseOverUnitListener(
-        (units, territory) -> {
-          unitsBeingMousedOver = units;
-          tooltipManager.updateTooltip(getUnitInfo());
-        });
+        (units, territory) -> tooltipManager.updateTooltip(getUnitInfo()));
```
RoiEXLab: Why not just pass units to `getUnitInfo()`?
This way there's no need to expose a getter, or am I missing something?

game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
```diff
@@ -341,6 +342,18 @@ public GameData getData() {
     return gameData;
   }
 
+  public boolean getEditMode() {
+    final boolean isEditMode;
+    // use GameData from mapPanel since it will follow current history node
```
RoiEXLab: I guess this comment does no longer make 100% sense (if it ever did).

One thing that I think is weird with this change in general:
Even if the method itself seems to delegate 50% of it's method to the MapPanel class, the only usages are part of TripleAFrame which seems counter-intuitive to me 🤔

game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
```diff
@@ -341,6 +342,18 @@ public GameData getData() {
...
+    gameData.acquireReadLock();
+    try {
+      isEditMode = BaseEditDelegate.getEditMode(gameData);
```
RoiEXLab: Just a small thing I noticed, there's no need to have an extra variable for this, this line can return the result directly.

game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
```diff
@@ -140,7 +140,7 @@ public void gameDataChanged(final Change change) {
             tileManager.resetTiles(gameData, uiContext.getMapData());
             SwingUtilities.invokeLater(
                 () -> {
-                  recreateTiles(getData(), uiContext);
+                  recreateTiles(gameData, uiContext);
```
RoiEXLab: This is actually a thing I wonder which "style" to prefer, maybe something we should mention in our code guidelines

game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
DanVanAtta: I did give that a shot, the second usage of `getUnitInfo` IIRC does not work well with it. If it did work, then perhaps the changes then became just extremely involved and would not have been reviewable. Perhaps after this PR it can be done more cleanly, or the original problem in the second usage would become clear.

game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
DanVanAtta: I'm confused. This method delegates basically everything to MapPanel. The `getEditMode` method being in `TripleAFrame` seems to be classic feature envy code smell. The fact `TripleAFrame` is the only consumer, does not mean it should also know/manipulate the internal data state of `MapPanel` by getting internal data and then managing and interpretting that data state.

THe comment still kinda makes sense perhaps. MapPanel, when in history mode, gets a clone of the game data. It's subtle and hard to really grok the data state of MapPanel (not well encapsulated at all, but also MapPanel is so bloated, it's hard to keep track of the data state anyways)

game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
DanVanAtta: Agree, but for a follow-up. Didn't want to drastically update things too much to keep the review managable and make it clear this is moved code. The amount of cleanup that should be done in MapPanel and TripleAFrame goes really, really deep.

game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
DanVanAtta: IIRC, there is a sonar lint violation about it. Using a method to access local data is an anti-pattern IMO. The method can be overriden, which then can cause unexpected breaks in methods that were previously working. Even if there are no breakages, it becomes very non-intuitive looking at code how it would work as the methods can behave differently depending on the instance type. This removes the ability to reason about a class by only looking at its code, you have to understand the usage history and system state to know exactly how the class will behave. Hence, anti-pattern, it makes it hard to reason how this class will work and breaks encapsulation.

game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
RoiEXLab: Good point about the encapsulation.
If the 50% was what confused you: Just an arbitrary number, didn't want to write 100% because that's not really accurant either

game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
RoiEXLab: Interesting, weird nevertheless.

PR #5877, opened 2020-01-24
Use non-primitive type to fix meaningless builder @Nonnull

DanVanAtta:


DanVanAtta:
@RoiEXLab perhaps, it's a language limitation though as primitives will always get a non-null value.

RoiEXLab:
Well it kind of is, but there's really no reason that lombok can't just add a boolean for every primitive field that is set to true whenever the field is actually set to a "real" value.
So yeah, `@NonNull` is probably not intended to mark mandatory fields (explicitly setting a field to null wouldn't work either), but IMO there should be a way.

DanVanAtta:
> `@NonNull is probably not intended to mark mandatory fields`

Wait, isn't @NonNull there basically to mark a mandatory field?

The annotation works across other contexts, like AllArgsConstructor IIRC. Knowing to use non-primitive types is a downside of lombok, a piece of trivia to know in order to use the library well. FWIW, at least there is a warning about it now. I've certainly ran across prod bugs due to a primitive not being set.

RoiEXLab:
> Wait, isn't @nonnull there basically to mark a mandatory field?

Not really, it just adds a null check for the field, and the builder just declares all the possible fields using their default value NULL, which is the null pointer for objects, and `0` for primitives.
So when actually building if any object-field does still have the initial null-pointer the null-check fails and we get an exception.
Making the field "required" is just a nice side-effect.

And well `@AllArgsConstructor` (to my knowledge) always creates a constructor that contains all fields of the class. If the fields have constraints, the contrains like non-nullity are checked, but that's it.
`@RequiredArgsConstructor` on the other hand defines "required fields" as fields that are
> final [...] and fields with constraints such as `@NonNull`.

See https://projectlombok.org/api/lombok/RequiredArgsConstructor.html

RoiEXLab:
On the bright side, JDK 14 brings [Records](https://openjdk.java.net/jeps/359), which could potentially make lombok redundant in the future, or at least fewer features of it

PR #5880, opened 2020-01-25
MapData - find islands refactor

DanVanAtta:
- Extract, test and refactor island finding algorithm
- Simplify usage of island data in AutoPlacementFinder, better encapsulate island data in MapData.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

## Additional Review Notes
Lot of small refactoring steps, sequence of refactoring can be reviewed commit-by-commit.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/mapdata/IslandTerritoryFinder.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/mapdata/IslandTerritoryFinder.java
@@ -34,7 +34,8 @@ class IslandTerritoryFinder {
     // map sea territories to: sea territory name -> islands contained by that sea territory
     return seaTerritories.stream()
         .collect(
-            Collectors.toMap(sea -> sea, findIslandsForSeaTerritory(landTerritories, polygons)));
+            Collectors.toMap(
+                Function.identity(), findIslandsForSeaTerritory(landTerritories, polygons)));
   }
 
   private static Set<String> filterSeaTerritories(final Set<String> territoryNames) {
--- a/game-core/src/main/java/games/strategy/triplea/ui/mapdata/IslandTerritoryFinder.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/mapdata/IslandTerritoryFinder.java
@@ -3,7 +3,6 @@ package games.strategy.triplea.ui.mapdata;
 import games.strategy.ui.Util;
 import java.awt.Polygon;
 import java.awt.Rectangle;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -29,18 +28,16 @@ class IslandTerritoryFinder {
     final Set<String> seaTerritories = filterSeaTerritories(polygons.keySet());
     final Set<String> landTerritories = filterNotSeaTerritories(polygons.keySet());
 
+    // Function where given a territory name, give us a polygon for that territory
     final Function<String, Polygon> polygonLookup =
         territoryName -> polygons.get(territoryName).iterator().next();
 
-    final Map<String, Set<String>> contains = new HashMap<>();
-
-    for (final String seaTerritory : seaTerritories) {
-      final Set<String> islands =
-          findIslandsForSeaTerritory(seaTerritory, landTerritories, polygonLookup);
-
-      contains.put(seaTerritory, islands);
-    }
-    return contains;
+    // map sea territories to: sea territory name -> islands contained by that sea territory
+    return seaTerritories.stream()
+        .collect(
+            Collectors.toMap(
+                sea -> sea,
+                sea -> findIslandsForSeaTerritory(sea, landTerritories, polygonLookup)));
   }
 
   private static Set<String> filterSeaTerritories(final Set<String> territoryNames) {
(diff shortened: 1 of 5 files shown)


game-core/src/main/java/games/strategy/triplea/ui/mapdata/IslandTerritoryFinder.java
```diff
@@ -0,0 +1,74 @@
...
+   *     contained by each corresponding sea territory.
+   */
+  static Map<String, Set<String>> findIslands(final Map<String, List<Polygon>> polygons) {
+    final Set<String> seaTerritories = filterSeaTerritories(polygons.keySet());
+    final Set<String> landTerritories = filterNotSeaTerritories(polygons.keySet());
+
+    // map sea territories to: sea territory name -> islands contained by that sea territory
+    return seaTerritories.stream()
+        .collect(
+            Collectors.toMap(sea -> sea, findIslandsForSeaTerritory(landTerritories, polygons)));
```
RoiEXLab: Would `Function.identity()` work here?

game-core/src/main/java/games/strategy/triplea/ui/mapdata/IslandTerritoryFinder.java
```diff
@@ -0,0 +1,74 @@
...
+            .collect(Collectors.toSet());
+  }
+
+  /** Checks that a given land territory would be contained by a given sea territory. */
+  private Predicate<String> landIsContainedBySeaTerritory(
+      final String seaTerritory, final Map<String, List<Polygon>> polygons) {
+
+    // Function where given a territory name, give us a polygon for that territory
+    final Function<String, Polygon> polygonLookup =
+        territoryName -> polygons.get(territoryName).iterator().next();
```
RoiEXLab: 2 things:
1. This is a map of lists, so instead of using `list.iterator().next()`, you could as well use `list.get(0)` directly
2. Does it make sense to check the size of the list before? If the lists was empty for some reason, this would result in an exception. If there are more than 1 polygon (which is mentioned somewhere in the javadocs) this would be silently ignore all the other polygons

game-core/src/main/java/games/strategy/triplea/ui/mapdata/IslandTerritoryFinder.java
```diff
@@ -0,0 +1,74 @@
...
+   * Finds all island territories given a set of territory names and their corresponding polygons.
+   * An island is defined as a land territory that is contained within a sea territory.
+   *
+   * <p>Warning: territories may consist of multiple polygons, only the first polygon is checked.
+   *
+   * @param polygons Mapping of territory names to the polygons representing that territories.
+   * @return A mapping of sea territory names to the name of land territories that are islands
+   *     contained by each corresponding sea territory.
+   */
+  static Map<String, Set<String>> findIslands(final Map<String, List<Polygon>> polygons) {
```
RoiEXLab: Another quick thought:
What about making this class non-static and make the polygon map an actual field?
This way a lot of lambdas could be turned into private methods and there would be no need to pass this map to almost every method.

game-core/src/main/java/games/strategy/triplea/ui/mapdata/IslandTerritoryFinder.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+class IslandTerritoryFinder {
+
+  /**
+   * Finds all island territories given a set of territory names and their corresponding polygons.
+   * An island is defined as a land territory that is contained within a sea territory.
```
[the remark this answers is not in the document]
DanVanAtta: Updated javadoc to be a bit more clear on this: 99fc238388954393f21e5bf4566f299171efadb9

game-core/src/main/java/games/strategy/triplea/ui/mapdata/IslandTerritoryFinder.java
DanVanAtta: It seems to, yes: 30db1cf9b

game-core/src/main/java/games/strategy/triplea/ui/mapdata/IslandTerritoryFinder.java
DanVanAtta: Original implementation here; I think the polygon reader ensures that we get at least one polygon. I suspect there is further improvement to be had here.

game-core/src/main/java/games/strategy/triplea/ui/mapdata/IslandTerritoryFinder.java
DanVanAtta: Not a bad thought, though I think that might be going the wrong direction. It's probably better to try and collapse methods where it makes sense rather than storing state.

DanVanAtta:
@RoiEXLab agree there should be a O(n) solution here. Overall the finder is pretty fast, was 22ms, about 39ms after this update (on World at War). This I think is a place where a lot of engineering instincts, training, and particularly interviews and interview prep teach us the wrong the thing. This is simply not the place for faster performance, the objective we want is to reduce maintenance cost and human time spent on this code - essentially tuck it away, let it work, and never invest time in it and maybe one day throw it away for something else. I realized this while working on this, it was a bit of a journey to see that there were not any really easy performance improvements.

PR #5882, opened 2020-01-26
Make MapData control lifecycle of resource loader and remove close method

DanVanAtta:
Updates MapData to always instantiate a resource loader itself and to close
it when construction is done. This ensures that the resource loader is closed
and removes the burden of other classes from having to remember to call
close on map data.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

## Additional Review Notes

Use this link to hide whitespace changes: https://github.com/triplea-game/triplea/pull/5882/files?utf8=%E2%9C%93&diff=unified&w=1

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
+++ b/game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
@@ -7,8 +7,11 @@ import com.google.common.base.Splitter;
 import games.strategy.engine.ClientFileSystemHelper;
 import games.strategy.engine.framework.map.download.DownloadMapsWindow;
 import games.strategy.engine.framework.startup.launcher.MapNotFoundException;
+import java.awt.Image;
+import java.io.ByteArrayInputStream;
 import java.io.Closeable;
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.MalformedURLException;
@@ -22,9 +25,11 @@ import java.util.Optional;
 import java.util.logging.Level;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
+import javax.imageio.ImageIO;
 import lombok.Getter;
 import lombok.extern.java.Log;
 import org.triplea.java.UrlStreams;
+import org.triplea.java.function.ThrowingSupplier;
 import org.triplea.swing.SwingComponents;
 
 /**
@@ -275,5 +280,32 @@ public class ResourceLoader implements Closeable {
 
     return UrlStreams.openStream(url)
         .orElseThrow(() -> new IllegalStateException("Failed to open an input stream to: " + path));
+  }
+
+  public ThrowingSupplier<InputStream, IOException> optionalResource(final String path) {
+    return () ->
+        Optional.ofNullable(getResourceAsStream(path))
+            .orElseGet(() -> new ByteArrayInputStream(new byte[0]));
+  }
+
+  public ThrowingSupplier<InputStream, IOException> requiredResource(final String path) {
+    return () ->
+        Optional.ofNullable(getResourceAsStream(path))
+            .orElseThrow(() -> new FileNotFoundException(path));
+  }
+
+  public Optional<Image> loadImage(final String imageName) {
+    final URL url = getResource(imageName);
+    if (url == null) {
+      // this is actually pretty common that we try to read images that are not there. Let the
+      // caller
+      // decide if this is an error or not.
+      return Optional.empty();
+    }
+    try {
+      return Optional.of(ImageIO.read(url));
+    } catch (final IOException e) {
+      throw new IllegalStateException(e);
+    }
   }
 }
--- a/game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
+++ b/game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
@@ -305,7 +305,8 @@ public class ResourceLoader implements Closeable {
     try {
       return Optional.of(ImageIO.read(url));
     } catch (final IOException e) {
-      throw new IllegalStateException(e);
+      log.log(Level.SEVERE, "Image loading failed: " + imageName, e);
+      return Optional.empty();
     }
   }
 }
--- a/game-core/src/main/java/games/strategy/triplea/ui/HeadedUiContext.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/HeadedUiContext.java
@@ -56,7 +56,7 @@ public class HeadedUiContext extends AbstractUiContext {
 
   @Override
   protected void internalSetMapDir(final String dir, final GameData data) {
-    if(resourceLoader != null) {
+    if (resourceLoader != null) {
       resourceLoader.close();
     }
     resourceLoader = ResourceLoader.getMapResourceLoader(dir);
--- a/game-core/src/main/java/games/strategy/triplea/ui/HeadedUiContext.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/HeadedUiContext.java
@@ -56,6 +56,9 @@ public class HeadedUiContext extends AbstractUiContext {
 
   @Override
   protected void internalSetMapDir(final String dir, final GameData data) {
+    if(resourceLoader != null) {
+      resourceLoader.close();
+    }
     resourceLoader = ResourceLoader.getMapResourceLoader(dir);
     mapData = new MapData(dir);
     // DiceImageFactory needs loader and game data
--- a/game-core/src/main/java/games/strategy/triplea/ui/HeadedUiContext.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/HeadedUiContext.java
@@ -173,6 +173,7 @@ public class HeadedUiContext extends AbstractUiContext {
   @Override
   public void shutDown() {
     super.shutDown();
+    resourceLoader.close();
   }
 
   @Override
--- a/game-core/src/main/java/games/strategy/triplea/ui/HeadedUiContext.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/HeadedUiContext.java
@@ -57,10 +57,7 @@ public class HeadedUiContext extends AbstractUiContext {
   @Override
   protected void internalSetMapDir(final String dir, final GameData data) {
     resourceLoader = ResourceLoader.getMapResourceLoader(dir);
-    if (mapData != null) {
-      mapData.close();
-    }
-    mapData = new MapData(resourceLoader);
+    mapData = new MapData(dir);
     // DiceImageFactory needs loader and game data
     diceImageFactory = new DiceImageFactory(resourceLoader, data.getDiceSides());
     final double unitScale =
@@ -176,7 +173,6 @@ public class HeadedUiContext extends AbstractUiContext {
   @Override
   public void shutDown() {
     super.shutDown();
-    mapData.close();
   }
 
   @Override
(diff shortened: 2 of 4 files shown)


game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
```diff
@@ -276,4 +281,31 @@ public boolean hasPath(final String path) {
     return UrlStreams.openStream(url)
         .orElseThrow(() -> new IllegalStateException("Failed to open an input stream to: " + path));
   }
+
```
DanVanAtta: These three methods are moved, no updates to them.

game-core/src/main/java/games/strategy/triplea/ui/HeadedUiContext.java
```diff
@@ -57,10 +57,7 @@
   @Override
   protected void internalSetMapDir(final String dir, final GameData data) {
     resourceLoader = ResourceLoader.getMapResourceLoader(dir);
-    if (mapData != null) {
-      mapData.close();
```
DanVanAtta: Removing this kind of call to `close()` is the main impetus for this update.

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
```diff
@@ -127,14 +122,16 @@
...
   private final Map<String, Image> effectImages = new HashMap<>();
-  private final ResourceLoader resourceLoader;
 
-  public MapData(final String mapNameDir) {
-    this(ResourceLoader.getMapResourceLoader(mapNameDir));
-  }
+  @Nullable private final Image vcImage;
+  @Nullable private final Image blockadeImage;
+  @Nullable private final Image errorImage;
+  @Nullable private final Image warningImage;
```
DanVanAtta: These 4 images were loaded when called, leveraging the fact that `MapData` had a reference to a `resourceLoader`. With the `resourceLoader` reference available now only on construction, we now load these images in the constructor.

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
```diff
@@ -224,52 +216,42 @@ public boolean scrollWrapY() {
...
     for (final String name : centers.keySet()) {
-      final Optional<Image> territoryNameImage = loadTerritoryNameImage(name);
+      final Optional<Image> territoryNameImage = loadTerritoryNameImage(resourceLoader, name);
 
       territoryNameImage.ifPresent(image -> territoryNameImages.put(name, image));
     }
     return territoryNameImages;
   }
 
-  private Optional<Image> loadTerritoryNameImage(final String imageName) {
```
DanVanAtta: The updates to this method are a simplification:
1. We chain the `if( img.isEmpty() )`  check to the optional with an 'or' check
2. Remove exception handling, none of the wrapped code (please double check) throws.

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
```diff
@@ -765,34 +747,20 @@ private Rectangle getBoundingRectWithTranslate(
...
   public Optional<Image> getErrorImage() {
-    return loadImage("misc/error.gif");
+    return Optional.of(errorImage);
   }
 
   public Optional<Image> getWarningImage() {
-    return loadImage("misc/warning.gif");
-  }
-
-  private Optional<Image> loadImage(final String imageName) {
```
DanVanAtta: `loadImage` is simply moved.

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
```diff
@@ -811,15 +779,18 @@ private Rectangle getBoundingRectWithTranslate(
...
 
   public Optional<Image> getTerritoryEffectImage(final String effectName) {
-    // TODO: what does this cache buy us? should we still keep it?
-    if (effectImages.get(effectName) != null) {
-      return Optional.of(effectImages.get(effectName));
-    }
-    Optional<Image> effectImage = loadImage("territoryEffects/" + effectName + "_large.png");
-    if (effectImage.isEmpty()) {
-      effectImage = loadImage("territoryEffects/" + effectName + ".png");
+    try (ResourceLoader loader = ResourceLoader.getMapResourceLoader(mapNameDir)) {
```
DanVanAtta: This method looks up territory effects at runtime. It's not clear if we could load them all in construction very easily in case a map would add territory effects as some sort of trigger. Hence we need to instantiate a resource loader here to support dynamic lookup of territory effects. Not ideal, but a trade-off..

game-core/src/main/java/tools/image/AutoPlacementFinder.java
```diff
@@ -260,56 +260,44 @@ private void calculate() {
...
+      textOptionPane.appendNewLine(name + ": " + points.size());
+    }
+    textOptionPane.appendNewLine("\r\nAll Finished!");
+    textOptionPane.countDown();
+    final String fileName =
+        new FileSave("Where To Save place.txt ?", "place.txt", mapFolderLocation).getPathString();
+    if (fileName == null) {
+      textOptionPane.appendNewLine("You chose not to save, Shutting down");
       textOptionPane.dispose();
-    } catch (final Exception e) {
```
DanVanAtta: This catch block would never have been triggered from `new MapData` to begin with, and is removed.

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
RoiEXLab: Not sure about the exception part.
`loadImage` explicitly rethrows an exception when there's an IOException thrown.
Because we have a URL-null check before, this shoudl happen rarely, but in case the image file has an unsupported format or something, this method would throw an exception.

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
```diff
@@ -811,15 +779,18 @@ private Rectangle getBoundingRectWithTranslate(
...
+      // TODO: what does this cache buy us? should we still keep it?
+      if (effectImages.get(effectName) != null) {
+        return Optional.of(effectImages.get(effectName));
+      }
+      Optional<Image> effectImage =
+          loader.loadImage("territoryEffects/" + effectName + "_large.png");
+      if (effectImage.isEmpty()) {
+        effectImage = loader.loadImage("territoryEffects/" + effectName + ".png");
```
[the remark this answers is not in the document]
RoiEXLab: Thinking about this again, you could even use it twice, by prepending `Optional.ofNullable(effectImages.get(effectImages))`, but in this case we might end up re-inserting the element into effect images

game-core/src/main/java/tools/image/AutoPlacementFinder.java
RoiEXLab: Not sure about that either.
This catches any potential RuntimeExceptions as well and loadImage does still explicitly throw one, so the constructor could definitely throw an exception that could land here.

game-core/src/main/java/games/strategy/triplea/ui/HeadedUiContext.java
```diff
@@ -57,10 +57,7 @@
   @Override
   protected void internalSetMapDir(final String dir, final GameData data) {
     resourceLoader = ResourceLoader.getMapResourceLoader(dir);
```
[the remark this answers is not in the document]
DanVanAtta: Looks like it used to be in `shutdown` and indirectly via closing MapData. That was very indirect, I missed that nuance (also a complete violation of encapsulation..) 

Good call-out/catch, added an explicit close to the `shutdown` method: 87e8c7ed8

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
```diff
@@ -127,14 +122,16 @@
...
+  @Nullable private final Image blockadeImage;
+  @Nullable private final Image errorImage;
+  @Nullable private final Image warningImage;
+  private final String mapNameDir;
 
-  public MapData(final ResourceLoader loader) {
-    resourceLoader = loader;
+  public MapData(final String mapNameDir) {
+    this.mapNameDir = mapNameDir;
+    final ResourceLoader loader = ResourceLoader.getMapResourceLoader(mapNameDir);
```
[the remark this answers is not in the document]
DanVanAtta: The nesting of try statements gets a bit nasty, not a bad idea nonetheless, added: cc36734b6

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
DanVanAtta: I believe you're right. I added the previous logging to 'loadImage': 8340c9776

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
```diff
@@ -765,34 +747,20 @@ private Rectangle getBoundingRectWithTranslate(
...
   }
 
   public Optional<Image> getBlockadeImage() {
-    return loadImage("misc/blockade.png");
+    return Optional.ofNullable(blockadeImage);
   }
 
   public Optional<Image> getErrorImage() {
-    return loadImage("misc/error.gif");
+    return Optional.of(errorImage);
```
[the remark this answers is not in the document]
DanVanAtta: Yes. Initially I was debating whether to honor that previous contract as these are images shipped with the game engine. Regardless, updated: 07502296e

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
```diff
@@ -765,34 +747,20 @@ private Rectangle getBoundingRectWithTranslate(
...
-      // caller
-      // decide if this is an error or not.
-      return Optional.empty();
-    }
-    try {
-      return Optional.of(ImageIO.read(url));
-    } catch (final IOException e) {
-      throw new IllegalStateException(e);
-    }
+    return Optional.of(warningImage);
```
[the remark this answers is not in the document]
DanVanAtta: 07502296e

game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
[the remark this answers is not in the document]
DanVanAtta: Good points. I wonder if I updated this further in the branch where this commit comes from. For now trying to limit the scope as this is mostly existing code.

game-core/src/main/java/tools/image/AutoPlacementFinder.java
DanVanAtta: I'm less sure now as well. Though, looking at the error handling, we would have shown a user 4 pop-ups, and none of which were very specific. 

Considering this is in the map-making `tools`, I think letting it throw is probably okay.  @RoiEXLab WDYT?

game-core/src/main/java/games/strategy/triplea/ui/HeadedUiContext.java
RoiEXLab: Hmm I think we still have the initial Problem, but on a higher level.
Previously the code did call `mapData.close()` whenever `internalSetMapDir` is executed in order to close the underlying resourceloader that might still exist.
Now MapData is fine, but in case this method is run twice (or more), we have n - 1 unclosed instances of ResourceLoader lying around, so it looks like this code needs to come back, I'm afraid :(

DanVanAtta: I agree to the issue you raised. The code that seemingly needs to come back it seems is a check for the previous resource loader and to close that. The resource loader is passed around a lot and references are kept to it. If we can avoid those references from being stored, we could safely close `resourceLoader` at the end of the `internalSetMapDir` method, which would long-term be more ideal IMHO.

PR #5894, opened 2020-02-01
Infra auto cert

DanVanAtta:
- Automate certbot letsencrypt and cert renewal
- Fix nginx restarting on every deployment run, only restart if config changes

## Additional Review Notes
- Commit-by-commit review may help break down the changesets and make review somewhat easier

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done

Preliminary:
```
~/work/triplea/infrastructure$ echo "***vault-password-omitted***" > vault_password
```

Full deployment command:
```
~/work/triplea/infrastructure$ ./run_deployment 2.0.17779 -i ansible/inventory/prerelease
```

Nginx Deployment command:
```
~/work/triplea/infrastructure$ ./run_deployment 2.0.17779 -i ansible/inventory/prerelease -t nginx --diff
```

CertBot Deployment command:
```
~/work/triplea/infrastructure$ ./run_deployment 2.0.17779 -i ansible/inventory/prerelease -t certbot --diff
```

- Deployed Nginx
  - Verified:
    - /etc/nginx/sites-enabled/default contained generated cert files

- Deployed Certbot
  - Verified:
    - /etc/nginx/sites-enabled/default contained letsencrypt cert files

- Deployed Nginx
  - Verified:
    - /etc/nginx/sites-enabled/default contained letsencrypt cert files
    - ansible reported no changes

Deleted:
  - /etc/letsencrypt
  - /etc/nginx/cert.*

- Deployed Nginx
  - Verified:
    - /etc/nginx/sites-enabled/default contained generated cert files
    - ansible reported no changes

- Deployed Certbot
  - Verified:
    - /etc/nginx/sites-enabled/default contained letsencrypt cert files

- Connected to prerelease lobby with headed-game lobby using URL (in test tab): https://prerelease.triplea-game.org
  - Verified:
    - chat works
  - Noticed:
    - Seeing a presumably false-positive error message for a connection problem connecting to websocket URL

- Connected to prerelease lobby with a second headed-game lobby using URL (in test tab): https://prerelease.triplea-game.org
  - Verified:
    - chat works and visible to all connected headed-games

- Deleted the Prerelease linode
- Created a new prerelease linode nanode
  - ubuntu 18.04 LTS
  - used the well known root password
  - selected the well known ssh key from linode account (we each should add this in the linode account settings)

- Updated namecheap A & AAAA DNS records for prerelease to reflect new IPv4 and IPv6 addresses
- Waited for DNS changes to take effect

- Ran full deployment
  - Verified:
    - /etc/nginx/sites-enabled/default contained letsencrypt cert files

- Connected to prerelease lobby with headed-game lobby using URL (in test tab): https://prerelease.triplea-game.org
  - Verified:
    - chat works

- Deployed Nginx
  - Verified:
    - established prerelease lobby connection from headed-game instance was not interrupted
    - /etc/nginx/sites-enabled/default contained letsencrypt cert files
   
- Deployed CertBot
  - Verified:
    - established prerelease lobby connection from headed-game instance was not interrupted

Deleted:
  - /etc/letsencrypt
  - /etc/nginx/cert.*

- Deployed Nginx
  - Verified:
    - /etc/nginx/sites-enabled/default contained generated cert files

- Deployed Certbot
  - Verified:
    - /etc/nginx/sites-enabled/default contained letsencrypt cert files

 - Connected to prerelease lobby with headed-game lobby using URL (in test tab): https://prerelease.triplea-game.org
  - Verified:
    - chat works

### DIFF
--- a/infrastructure/README.md
+++ b/infrastructure/README.md
@@ -80,19 +80,22 @@ examples
 
 # To see 'diff' output, which shows how each file on server is being updated
 #  (caution!! this can expose secret values to stdout)
-./run_deployment 2.0.1000 -d -i ansible/inventory/prerelease
+./run_deployment 2.0.1000 --diff -i ansible/inventory/prerelease
 
 # To see SSL debug output:
+./run_deployment 2.0.1000 -vvvv -i ansible/inventory/prerelease
+
+# To see debug output from each command:
 ./run_deployment 2.0.1000 -vvv -i ansible/inventory/prerelease
 
 # To deploy to just bots, use tags, '-t' (see playbook for tag names)
-./run_deployment 2.0.1000 -vvv -t bots -i ansible/inventory/prerelease
+./run_deployment 2.0.1000 -t bots -i ansible/inventory/prerelease
 ```
 
 Production deployment is only a matter of specifying the production inventory file.
 
 ```
-./run_deployment 2.0.1000 -vvv -i ansible/inventory/production
+./run_deployment 2.0.1000 -i ansible/inventory/production
 ```
 
 ## PreRelease and Production Deployments
@@ -165,47 +168,13 @@ ansible-vault encrypt --vault-password-file=vault_password ansible_ssh_key.ed255
 
 # Https Certificate Installation
 
-Currently done manually.
+The 'certbot' role will:
+  - run lets-encrypt and create a publicly signed SSL cert.
+  - sets up a weekly renewal cronjob that will renew the SSL cert if it
+    is within 30 days of expiry
 
-## certbot from letsencrypt
-
-```bash
-sudo apt-get update
-sudo apt-get install software-properties-common
-sudo add-apt-repository universe
-sudo add-apt-repository ppa:certbot/certbot
-sudo apt-get update
-sudo apt-get install certbot python-certbot-nginx 
-
-sudo certbot --nginx -m tripleabuilderbot@gmail.com --agree-tos
-```
-
-Create CAA DNS records
+For each domain running SSH, a CAA DNS record needs to be created (one time):
 
 ![Screenshot from 2019-11-19 13-06-13](https://user-images.githubusercontent.com/12397753/69196411-48980e00-0ae3-11ea-9130-61e1fd5368b3.png)
 
-Everything that goes well, should look like:
-```
-Congratulations! You have successfully enabled
-https://prerelease.triplea-game.org
-
-You should test your configuration at:
-https://www.ssllabs.com/ssltest/analyze.html?d=prerelease.triplea-game.org
-- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-
-IMPORTANT NOTES:
- - Congratulations! Your certificate and chain have been saved at:
-   /etc/letsencrypt/live/prerelease.triplea-game.org/fullchain.pem
-   Your key file has been saved at:
-   /etc/letsencrypt/live/prerelease.triplea-game.org/privkey.pem
-   Your cert will expire on 2020-02-17. To obtain a new or tweaked
-   version of this certificate in the future, simply run certbot again
-   with the "certonly" option. To non-interactively renew *all* of
-   your certificates, run "certbot renew"
- - Your account credentials have been saved in your Certbot
-   configuration directory at /etc/letsencrypt. You should make a
-   secure backup of this folder now. This configuration directory will
-   also contain certificates and private keys obtained by Certbot so
-   making regular backups of this folder is ideal.
-``````
 
--- a/infrastructure/README.md
+++ b/infrastructure/README.md
@@ -81,11 +81,11 @@ examples
 
 ### Useful flags
 
-  * `-v`: Verbose output
-  * `-vvv`: Debug output
-  * `-vvvv`: SSL Debug output
-  * `--diff`: Shows differences in updated templates and files
-  * `-t`: Tags
+* `-v`: Verbose output
+* `-vvv`: Debug output
+* `-vvvv`: SSL Debug output
+* `--diff`: Shows differences in updated templates and files
+* `-t`: Tags
 
 ### Examples with Tags
 
--- a/infrastructure/README.md
+++ b/infrastructure/README.md
@@ -67,34 +67,45 @@ defined in the prerelease inventory file.
 Assuming a vault password file named 'vault_password' is created in the same
 folder and contains the ansible vault password, run: 
 
-``bash
+```bash
 ./run_deployment [version_to_install] [ansible_args]
 ```
 
 examples
+
+
+### Typical Deployment Command
 ```bash
 ./run_deployment 2.0.1000 -i ansible/inventory/prerelease
+```
 
-# To see verbose output
-./run_deployment 2.0.1000 -v -i ansible/inventory/prerelease
+### Useful flags
+  * -v: Verbose output
+  * -vvv: Debug output
+  * -vvvv: SSL Debug output
+  * --diff: Shows differences in updated templates and files
+  * -t : Tags
 
-# To see 'diff' output, which shows how each file on server is being updated
-#  (caution!! this can expose secret values to stdout)
-./run_deployment 2.0.1000 --diff -i ansible/inventory/prerelease
+### Examples with Tags
 
-# To see SSL debug output:
-./run_deployment 2.0.1000 -vvvv -i ansible/inventory/prerelease
 
-# To see debug output from each command:
-./run_deployment 2.0.1000 -vvv -i ansible/inventory/prerelease
+#### Deploy just bots
 
-# To deploy to just bots, use tags, '-t' (see playbook for tag names)
+```bash
 ./run_deployment 2.0.1000 -t bots -i ansible/inventory/prerelease
 ```
 
+#### Deploy NGINX and CertBot
+```bash
+./run_deployment 2.0.1000 -t nginx,certbot -i ansible/inventory/prerelease
+```
+
+
+## Example Prod Deployment
+ 
 Production deployment is only a matter of specifying the production inventory file.
 
-```
+```bash
 ./run_deployment 2.0.1000 -i ansible/inventory/production
 ```
 
--- a/infrastructure/README.md
+++ b/infrastructure/README.md
@@ -75,6 +75,7 @@ examples
 
 
 ### Typical Deployment Command
+
 ```bash
 ./run_deployment 2.0.1000 -i ansible/inventory/prerelease
 ```
@@ -181,8 +182,9 @@ ansible-vault encrypt --vault-password-file=vault_password ansible_ssh_key.ed255
 # Https Certificate Installation
 
 The 'certbot' role will:
-  - run lets-encrypt and create a publicly signed SSL cert.
-  - sets up a weekly renewal cronjob that will renew the SSL cert if it
+
+- run lets-encrypt and create a publicly signed SSL cert.
+- sets up a weekly renewal cronjob that will renew the SSL cert if it
     is within 30 days of expiry
 
 For each domain running SSH, a CAA DNS record needs to be created (one time):
--- a/infrastructure/README.md
+++ b/infrastructure/README.md
@@ -80,11 +80,12 @@ examples
 ```
 
 ### Useful flags
-  * -v: Verbose output
-  * -vvv: Debug output
-  * -vvvv: SSL Debug output
-  * --diff: Shows differences in updated templates and files
-  * -t : Tags
+
+  * `-v`: Verbose output
+  * `-vvv`: Debug output
+  * `-vvvv`: SSL Debug output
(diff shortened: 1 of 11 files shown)


infrastructure/ansible/roles/nginx/tasks/main.yml
```diff
@@ -4,52 +4,108 @@
...
   become: true
   systemd:
     name: nginx
     state: restarted
 
+- name: ensure nginx is started
+  become: true
+  systemd:
+    name: nginx
+    state: started
```
RoiEXLab: Doesn't the statement above already ensure the service is restarted? So IIRC this also verifies that the service is started if it wasn't before.

infrastructure/ansible/roles/certbot/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+  apt:
+    update_cache: yes
+    name:
+      - certbot
+      - python-certbot-nginx
+
+- name: look for certbot generated key file
+  become: true
+  stat:
+    path: /etc/letsencrypt/live/prerelease.triplea-game.org/cert.pem
```
RoiEXLab: I assume this path shouldn't be hardcoded?

infrastructure/ansible/roles/certbot/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+  register: has_certbot_pem
+
+- name: run certbot
+  become: true
+  command: "certbot --nginx -n -m tripleabuilderbot@gmail.com --agree-tos -d {{ inventory_hostname }}"
```
RoiEXLab: We don't have any mail forwarding for this email don't we?
I just got (2 days ago) an email from the certbot utility telling me that it failed to renew the MARTI HTTP certificates.
Turns out that there was (and still is, but I found a workaround) a DNSSEC issue with namecheap that fails for a nonexistent subdomain, which ultimately caused the check to fail.
Without this email we wouldn't have noticed until it's too late, so I'm convinced there's definitely some value in using an actual email here.

infrastructure/ansible/roles/nginx/tasks/main.yml
```diff
@@ -4,52 +4,108 @@
...
     port: "{{ item }}"
     proto: tcp
   with_items: "{{ nginx_allowed_ports }}"
 
-- name: relaunch nginx
+- name: restart nginx if config changed
+  when: |
+    (nginx_config_changed.changed) 
+      or (has_cert_key.stat.exists == false) 
+      or (has_cert.stat.exists == false)
```
RoiEXLab: How often is this file run?
Could it happen that the certificate runs out in-between runs?
For Marti and the forum I solved this problem by adding a post-hook that runs `sudo systemctl reload nginx` to ensure that nginx picks up the certificate correctly after refreshing it. It could be the case that certbot does this automatically using the nginx plugin (which is used), but when I initially configured everything there such a plugin didn't exist yer.

infrastructure/ansible/roles/certbot/tasks/main.yml
RoiEXLab: Oh and another thing: By adding `--rsa-key-size 4096` we get a stronger RSA key.
I use this for every other server we use certbot on.

infrastructure/ansible/roles/nginx/tasks/main.yml
RoiEXLab: I just double checked my assumption.
Looks like the nginx plugin takes care of automatically restarting the service, so no need to restart nginx "manually".
See https://community.letsencrypt.org/t/doest-certbot-automaticly-reload-nginx-on-successful-renew/40361/5

infrastructure/ansible/roles/certbot/tasks/main.yml
DanVanAtta: tripleabuilderbot@gmail.com  is a real email. If we have a dev-forwaring list, that would be good to use. I just sent a test email to: tripleadevelopers@gmail.com and got nothing - is that a working distro? Arguably we should also encode the email as well so that it does not get scraped and yield spam.

Added the key-size flag: 69a240eed; The command here is based on what we previously had in the README

infrastructure/ansible/roles/certbot/tasks/main.yml
DanVanAtta: Indeed, good catch, updated: 828cce124

infrastructure/ansible/roles/certbot/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+  when: has_certbot_pem.stat.exists == false
+
+- name: Add letsencrypt cronjob for cert renewal
+  become: true
+  cron:
+    name: letsencrypt_renewal
+    special_time: weekly
+    job: certbot renew
```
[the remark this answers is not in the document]
DanVanAtta: It raises a lot of questions if does:
- how does it know what kind of cron to use
- how does it know you actually want to auto-renew
- if the job does renew, why does the cert creation not mention it and also give instructions on how to renew

I'll see if I can verify if it does.

infrastructure/ansible/roles/nginx/tasks/main.yml
DanVanAtta: Note that the 'restart' now has a when clause on it for only when something has changed in configuration. That was extremely important otherwise on every build we would have disconnected all users when nginx restarted.

This 'started' command has no when clause. This will ensure the service is in a started state. This is really here to 'ensure' an expected state that nginx is running. Namely it is to repair the case where we by hand turn nginx off, or if it somehow terminates in some other manner. The restart should take effect on any initial install, so we would expect this 'started' to virtually never have an effect.

infrastructure/ansible/roles/nginx/tasks/main.yml
DanVanAtta: Is this file `tasks/main.yml`? If so, on every build. If 'this file' is the cert, if you scroll up in the same `tasks/main.yml` there is a stat command to check for `/etc/nginx/cert.crt` and key and generates if does not exist, so that is one-time. 

To explain this when clause, if someone deletes the 'crt' files and we recreate them, and even if 'sites-enabled/default' is not updated, we'll restart nginx. Otherwise, if 'sites-enabled/default' is modified, we'll restart nginx.

> Could it happen that the certificate runs out in-between runs?

This role/task is independent of the certbot certificates, it is for a 'base' nginx only and is unaware of certbot. AFAIK the generated certs `/etc/nginx/cert.crt`, do not expire as those are just keys.

> Looks like the nginx plugin takes care of automatically restarting the service, so no need to restart nginx "manually"

I think perhaps there is a disconnect. We are restarting here on configuration change or a change in the nginx keys, there is no certbot involved here yet.

DanVanAtta:
> It's great seeing the final steps for this, just a lot of questions from my side

As always, do appreciate the look-over and review. 

Setting up a fresh linode for prerelease, I realized pretty quickly the by-hand running of certbot and then checking in updated configuration changes was not very tenable. It's really significant/important that we have it all automated. This does get us in a good place where you can create a linode, which has a handy setting now to deploy a SSH-key to root authorized-keys on creation, and then once done in the linode UI there is no further configuration and the automation takes care of the rest. This should make creating new boxes and setting up prod to be easy and well documented.

DanVanAtta:
Just now checking on a clean box if certbot creates an auto-renew job, will be searching for any created cronjobs of sysctl type jobs. @RoiEXLab please have a final check-over if you would and let me know if there is anything else to address or any questions you may have.

infrastructure/ansible/roles/certbot/tasks/main.yml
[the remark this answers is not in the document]
RoiEXLab: > how does it know what kind of cron to use

Good question, can't answer that but there probably are some checks in place to get the correct one.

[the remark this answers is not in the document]
RoiEXLab: > how does it know you actually want to auto-renew

This is the default behaviour (the certificates are only valid for 90 days, so unless you only need TLS for less than 90 days you probably want to renew it eventually), but I think there's an option to disable it

>if the job does renew, why does the cert creation not mention

Hard to answer. I assume that the "bot" in certbot somewhat implies thae automation. Note that cerbot is not directly linked to letsencrypt. It's a separate utility that "just happens" to be created at the same time as letsencrypt and the ACME protocol and uses letsencrypt as its default CA, but it works with any other CA that supports the ACME protocol as well. So you can get a letsencrypt certificate without this utility.

> and also give instructions on how to renew

Well I mean renewing just applies the initial configuration again, that's what `certbot renew` does, just within a cronjob

infrastructure/ansible/roles/nginx/tasks/main.yml
RoiEXLab: Ah, ok makes sense.
Thanks for clarifying

DanVanAtta:
Well :hankey: 


>  Installer nginx\nObtaining a new certificate\nAn unexpected error occurred:\nThere were too many requests of a given type :: Error creating new order :: too many certificates already issued for exact set of domains: prerelease.triplea-game.org: see https://letsencrypt.org/docs/rate-limits


> The main limit is Certificates per Registered Domain (50 per week). A registered domain is, generally speaking, the part of

DanVanAtta:
@RoiEXLab checked in an update commenting out the cronjob block: 2fbc544c5; That'll give us time to see if we can find any type of renewal automation, if not, then we only need re-enable it, otherwise if we find a cronjob then we can delete it.

DanVanAtta:
To get around the rate limit, created a second domain 'prerelease2'  : )
Explicit cronjob is still temporarily disabled.

RoiEXLab:
FYI The official installation guide does mention the automatic renewal:
https://certbot.eff.org/lets-encrypt/ubuntubionic-nginx

RoiEXLab:
It also states that it creates one in one of the following locations:
```
    /etc/crontab/
    /etc/cron.*/*
    systemctl list-timers
```

DanVanAtta:
Cool, found this: `/etc/cron.d/certbot`

```
# /etc/cron.d/certbot: crontab entries for the certbot package
#
# Upstream recommends attempting renewal twice a day
#
# Eventually, this will be an opportunity to validate certificates
# haven't been revoked, etc.  Renewal will only occur if expiration
# is within 30 days.
#
# Important Note!  This cronjob will NOT be executed if you are
# running systemd as your init system.  If you are running systemd,
# the cronjob.timer function takes precedence over this cronjob.  For
# more details, see the systemd.timer manpage, or use systemctl show
# certbot.timer.
SHELL=/bin/sh
PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin

0 */12 * * * root test -x /usr/bin/certbot -a \! -d /run/systemd/system && perl -e 'sleep int(rand(43200))' && certbot -q renew
```

DanVanAtta:
Side-note, @RoiEXLab you have ready access to prerelease.triplea-game.org?

EG: ```ssh admin@prerelease.triplea-game.org```

DanVanAtta:
Removed commented out cronjob install: 0af29364e

infrastructure/README.md
```diff
@@ -165,47 +168,13 @@ ansible-vault encrypt --vault-password-file=vault_password ansible_ssh_key.ed255
...
-
-sudo certbot --nginx -m tripleabuilderbot@gmail.com --agree-tos
-```
+The 'certbot' role will:
+  - run lets-encrypt and create a publicly signed SSL cert.
+  - sets up a weekly renewal cronjob that will renew the SSL cert if it
+    is within 30 days of expiry
 
-Create CAA DNS records
+For each domain running SSH, a CAA DNS record needs to be created (one time):
```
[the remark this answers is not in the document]
RoiEXLab: Thoughts about the wording here?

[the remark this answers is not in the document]
DanVanAtta: Certainly getting into semantics :grin: 

I think explaining that we should but don't have to create such an entry will be too much. Terse, concise instructions that do not require decision making are valuable. We could add a link for background information potentially.

Hence for 'semantics', the 'need' is untrue as it's not required, but if we consider our security requirements to "need" such a cert for the improved security, then "need" is correct from that perspective.

So while not strictly needed, per the security standards we would expect, it is needed and should be done.

RoiEXLab:
@DanVanAtta Just checked, I do have access 👍

RoiEXLab:
FYI I went ahead and created a CAA record for prerelease2 for consistency

RoiEXLab:
TLS is looking pretty good so far:
https://www.ssllabs.com/ssltest/analyze.html?d=prerelease2.triplea%2dgame.org&s=2600%3a3c00%3a0%3a0%3af03c%3a92ff%3afee0%3a861e&latest

Issue #5480, opened 2019-11-14
Travis Master Branch Build Failed


tripleabuilderbot:
Latest master branch travis build [#16527](https://travis-ci.org/triplea-game/triplea/builds/612037696) failed.
No release artifacts will be generated until this is fixed.

tripleabuilderbot:
Build [#16527](https://travis-ci.org/triplea-game/triplea/builds/612037696) failed

tripleabuilderbot:
Build [#16527](https://travis-ci.org/triplea-game/triplea/builds/612037696) failed

tripleabuilderbot:
Build [#16527](https://travis-ci.org/triplea-game/triplea/builds/612037696) failed

RoiEXLab:
Oh: https://status.gradle.com/incidents/q2vq49fndxg0
Looks like a newer java 11 patch is required, TLS 1.0 is no longer supported

DanVanAtta:
We might be able to add a 'before-install' step to update java: https://travis-ci.community/t/old-java-version-in-bionic/4726

Simplest would be go to jdk12 for all builds, could be a decent temporary measure but would also be reason perhaps to migrate to java12 or java13 (not sure if good idea, maybe better to stay with LTS version).

RoiEXLab:
Whatever works to get the builds running again, I mean if you have some time to spend, you might want to give github CI a try, looks like Github Actions was finally released ^^

DanVanAtta:
> you might want to give github CI a try, looks like Github Actions was finally released ^^

Indeed, though would be a big migration, maybe even bigger than upgrading java version. We'd want to evaluate as well if that is the long term direction we want, pro's/cons vs travis and the ROI of migrating (it might not be worth it, we do have a working solution for now, even if confronted with this problem right now). 

I think the update java commands are perhaps the best path forward without doing a large ad-hoc migration (now).

DanVanAtta:
>  I mean if you have some time to spend

Side-note, we should be careful of this mentality. I'm about to go back to work full time, the hours I can dedicate to TripleA is going to be counted in minutes and hours soon, not days. We should also be really careful to be focused and use our time well. The amount we have to do, it's enough for a team of 10 to be busy for a year, we need to keep that in mind, failure is an option.

DanVanAtta:
[...]

PR #5899, opened 2020-02-03
Clean Up and Restructure Travis Build Scripts

DanVanAtta:
- **Re-structure Build Scripts**
  - Goal was to make it so .travis.yml contained very little to any logic at all for build commands.
    Second, by grouping build jobs together, the script structure and scope of which scripts were
    used for what is hopefully now more clear.
  - The new script configuration places build scripts into their own folders following this
    convention: '.travis/stages/<stage-name>/<job-name>/run'. Any additional support scripts
    used by a given job will be in the 'job' folder. For most part, any .travis.yml build job
    will just call a 'run' a script and nothing more
  - Scripts that are shared between multiple jobs are now located in '.travis/shared/'

- **Break up Build Jobs**
  - Results in a slightly faster build as we get better parallelism, and the slowest build is now
    slightly faster as it has fewer tasks to perform.
  - Error output is more focused, looking at which build job failed we can get a pretty good idea
    of what went wrong without looking at the logs
  - We get more feedback. Rather than knowing that just tests failed, we can see which parts of the
    build passed and which pieces failed. For example, if there is just a checkstyle violation,
    then we'll see just one failure for checkstyle and everything else would pass.

- **Clean up .travis.yml**
  - Build logic extracted to a 'run' script
  - Lists and maps written with a single line syntax where possible and clean.
  - Common operations, like "install: false", and "cache: .." are extracted to a top level
    where they become common for all jobs.

- **Remove java-11.0.5 install**
  - 11.0.2 is currently doing the job. We added the 11.0.5 install to mitigate a TLS versioning
    issue where connections to JCenter were being rejected due to our version of TLS being too low.
    Background info: https://github.com/triplea-game/triplea/issues/5480
    It appears the problem is currently mitigated or patched on the Travis side as 11.0.2 is working.
    We can restore the 11.0.5 install pretty readily if the need arises again, in the meantime it's
    not necessary (and adds unwanted complexity).

- **Move error reporting to a top-level task (out of build scripts) and be consistent**
  - With all build jobs extracted to folders and invoked on a single line in .travis.yml,
    we can more cleanly execute a build job fail handler by adding "|| $REPORT_FAIL" to
    the end of each run script. This means if any script fails, we'll call the report-fail handler
    to consistently have an issue be created whenever we see a build failure.
    Note: we filter PR builds by checking to see if needed env variables that are only available
    to a master branch build are present, if they are not, the 'report-fail' script is a no-op.

- **Move Succuss Detection to a final Stage**
  - We no longer require a specific script to detect if the build passed and report success. Instead
    there is now a third and final build stage that will report a success and close any 'Build failed'
    issues that may be open. This is in part to make it so that each build script in .travis/ can be
    more focused on just one task.

- **Updates to Prod Deployment - Just Install SSH Keys and Security updates**
  - Implicitly, decided prod deployments will be done 'by-hand'
  - Prod will be updated on every build, but it'll only have SSH keys deployed and security patches
    applied, no binaries or configuration will be changed.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done

- Verified branch build looks good
- Disabled conditional checks in .travis.yml to verify the 'deploy' stage builds looked good. They still died due to 'variable not defined' errors, but seemingly looked to be well configured FWIW.

Did not verify a build fail issue would be created on failure or closed on success. 

## Additional Review Notes

- Given some of the build steps are only done on master branch, I expect there may be some fast-follow iterations to fix minor problems that were not found during testing.
- The diff's are likely not easy to follow as much was moved:
   - The .travis.yml script was restructured, recommend to view it in its new state without any diff lines. Please provide feedback on structure, naming of stages, and if job and stage breakup makes sense.
- Please have a look at the new folder structure, check if that looks good.
- Build scripts are simplified/broken up, there is no new functionality or any removed. 

### Travis Jobs Before
Screenshot from 2020-02-02 21-04-59

### Travis Jobs After
Screenshot from 2020-02-02 21-05-07

###  Folder structure before:
Screenshot from 2020-02-02 21-10-03

###  Folder structure after:

Screenshot from 2020-02-02 21-13-37

### .travis.yml Before (sample)
Screenshot from 2020-02-02 21-16-56

### .travis.yml After (sample)
Screenshot from 2020-02-02 21-16-31


.travis.yml
```diff
@@ -1,188 +1,109 @@
...
-        - java --version
-      script:
-        - ./.travis/setup_gpg
-        ## Update product version to include build number.
-        ## EG: replace "2.0.0" to be "2.0.15555";
-        - sed -i "s/.0 *$/.$TRAVIS_BUILD_NUMBER/" game-core/src/main/resources/META-INF/triplea/product.properties
-        - ./.travis/do_release
+      script: [ $DEPLOY_STAGE/installers/run || $REPORT_FAIL ]
       deploy:
+        on: { tags: false, repo: triplea-game/triplea, branch: master }
```
RoiEXLab: Hmm, I think this mix of the yaml 1.1 syntax and the newer 1.2 syntax is a little bot weird here and there.
This line in particular doesn't add to readability IMO, but personal preference I guess

.travis.yml
```diff
@@ -1,188 +1,109 @@
...
-          - '.gradle'
-          - '~/jdk-11.0.5+10'
+    
+    ########################################################################
+    # Jobs are ordered tests first and then slower jobs before faster ones.
+  
+    - stage: Verify
+      name: "Tests: Integration Test"
+      addons: { postgresql: "10" }
+      script: [ $VERIFY_STAGE/integration-testing/run || $REPORT_FAIL ]
```
RoiEXLab: For all of the script parts: I'm pretty sure that travis doesn't care if a list or a single entry is given. After all the default value for this key should be `./gradlew check` IIRC

.travis.yml
```diff
@@ -1,188 +1,109 @@
...
+      addons: { apt: { packages: [ python3, python3-yaml ] } }
+      script: [ $DEPLOY_STAGE/update-maps-on-website/run || $REPORT_FAIL ]     
+
+    - stage: Deploy
+      name: "Ansible Deploy to Prerelease"
+      if: (branch = master) and (repo = 'triplea-game/triplea') and (type != 'pull_request')
+      language: python
+      python: "3.6"
+      addons: { apt: { packages: [ sshpass ] } }
+      install: [ pip install ansible ]
```
RoiEXLab: The same thing here, similar to the script thingy. It shouldn't be necessary to add list brackets here.

.travis.yml
DanVanAtta: Are you looking at this only from a web-browser, from a terminal, or from IDE?
In general, I found the shorter, the better for most of this, the multi-line elements lengthened the config file and stood out as most things are single line now.

How do you mean by mixing of the 1.1 vs 1.2 syntax, as far as I knew, the inline map syntax has always been equivalent.

.travis.yml
DanVanAtta: Hmm, not sure if this matters. If we wind up adding more, then it'll be less of a diff, on the other hand we should not be adding more lines as each item in 'script' is intentionally extracted to 'run'.

.travis.yml
RoiEXLab: Only the Web browser.
I get what you were trying to achieve, but I think in this particular case (also because there are not too many deployment blocks) it doesn't really add to readability.
But I could live with either way, just wanted to point it out.

DanVanAtta: Here is a contrast:

Screenshot from 2020-02-06 19-25-20

Screenshot from 2020-02-06 19-25-36

The multi-line leads to the most indentation and makes nearby indentation much more important. The values specified are somewhat noisy, but also makes the file quite a bit longer. One issue IMO we developed with the travis file is it became very long, too much to really understand easily.

.travis.yml
DanVanAtta: Took a second look at this, while the list syntax is not strictly necessary, it does do a good job to visually distinguish the script step from the other labels that are scalars.

PR #5903, opened 2020-02-07
Reduce/cleanup infrastructure files

DanVanAtta:
- Move various README content to wiki
- Delete 'status-check' scripts


DanVanAtta:
The removed documentation is just moved to wiki

PR #5904, opened 2020-02-07
Add forum server to production inventory

DanVanAtta:


DanVanAtta:
We probably want at least setup and any configuration to be checked in. As-is, recovery is unclear in case we need to recreate the server. Backing up the data would be good to do as well.

Regardless, the core management is the only thing being done on the prod inventory file, which is essentially installing SSH keys.

@RoiEXLab we need the common ansible ssh key to be installed in `/root/.ssh/authorized_keys`, that key should be on all of the other servers. Would you mind installing that?

RoiEXLab:
@DanVanAtta I copied `/root/.ssh/authorized_keys` from the prerelease server over to the forum server, I hope that's what you wanted.

Backups for the forum server are currently done via linodes' backup service, so we have a least some form of safety there.

PR #5915, opened 2020-02-08
Remove travis_terminate command from build script

DanVanAtta:
- travis_terminate is used to terminate travis build if any commands in a scripted
list fail (othewrise, even with a non-zero exit code, travis continues execution).
- This update removes a 'travis_terminate' which seemingly is not in scope from
within a scripted script:

```
.travis/stages/deploy/ansible-deploy-to-prerelease/run: line 12: travis_terminate: command not found
```

Given that we now only have a single command per script block, if the block exits with
a non-zero exit code, the build should be marked as failed anyways, negating the need
for a 'travis_terminate'


RoiEXLab:
Hmm now there's a different problem with the build

PR #5917, opened 2020-02-08
JavaFX local game launcher

RoiEXLab:
Starting a local Game works now from the JavaFX client.
The main reason this PR is "draft" is that I'd like to get some developer (and also community about the visuals) feedback before I try polishing everything.

## Functional Changes

[] New map or map update
[x] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done

## Screens Shots

Current State


DanVanAtta:
Good to see, give me a few days please @RoiEXLab to grok the changes. I presume we do plan to iterate quite a bit on the screens, layout, exact functionality, and further details like CSS and style. My impression that is that the core functionality and having a first MVP is really the key consideration here, and more should focus on general design/structure.

RoiEXLab:
> Good to see, give me a few days please

Sure, the main reason I created a PR is to urge myself to finally get this done. I had the "almost ready"-branch lying around for quite some time now and always found something else that felt more exciting to work on ^^

DanVanAtta:
I hear that, that last 20% can be really difficult to see through and be very grinding. I kinda feel that way about 2.0; hopefully we can get good ongoing testing feedback and have soft-launch be pretty soon.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
```diff
@@ -118,4 +123,33 @@ public static LocalLauncher create(
...
   }
+
+  public static LocalLauncher create(
+      final List<Entry<String, String>> roleMapping,
+      final Predicate<String> isDisabled,
+      final GameData gameData) {
+    final Predicate<Entry<String, String>> enabledPredicate =
+        entry -> !isDisabled.test(entry.getValue());
+    final PlayerListing playerListing =
+        new PlayerListing(
```
DanVanAtta: Is it worthwhile to create constructor over-loads to avoid the `null` args? We might be able to put more of the construction work in `PlayerListing` constructor as well to further simplify here.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
```diff
@@ -118,4 +123,33 @@ public static LocalLauncher create(
             null);
     return new LocalLauncher(gameSelectorModel, new PlainRandomSource(), pl, parent, launchAction);
   }
+
+  public static LocalLauncher create(
+      final List<Entry<String, String>> roleMapping,
+      final Predicate<String> isDisabled,
+      final GameData gameData) {
+    final Predicate<Entry<String, String>> enabledPredicate =
+        entry -> !isDisabled.test(entry.getValue());
```
DanVanAtta: The parameter coming in seems to be a logical negation - can we sanely change that to `isEnabled`?

Second, Is this lamba a place where we can use the `Predicates.not(..)`?

game-headed/src/main/java/org/triplea/game/client/ui/javafx/screens/RoleSelection.java
```diff
@@ -0,0 +1,192 @@
...
+  void cancelMapSelection() {
+    screenController.switchScreen(FxmlManager.MAP_SELECTION);
+  }
+
+  @FXML
+  void startGame() {
+    if (resourceModifierCheckbox.isSelected()) {
+      setupPuIncome();
+    }
+    final List<Entry<String, String>> flatMapping =
```
DanVanAtta: It looks like we have a decent amount of logic in the FXML class here. I was looking at the function reference "DISABLE_TEXT::equals", thinking that was logic in the XML class, but going up further, the `startGame` method itself has logic. This looks like a place where we are mixing display with back-end logic. 

I think we could have a really good fit for view-model, model-controller pattern here. If we had one class to represent the FXML components, I think we could decouple quite nicely and also make the logic test-able. 

EG:
```
roleSelectionViewModel.setResourceModifierSelected(resourceModifierCheckbox.isSelected());
roleSelectionViewModel.setPlayerRoles(roleForPlayers);
roleSelectionViewModel.startGame();
```

Have you considered using a view-model, any thoughts?

game-headed/src/main/resources/org/triplea/game/client/ui/javafx/fxml/RoleSelectionScreen.fxml
```diff
@@ -0,0 +1,52 @@
...
+            <Label text="Battle Orders" />
+         </children>
+      </HBox>
+      <ScrollPane>
+         <content>
+            <GridPane fx:id="factionGrid" hgap="5.0" vgap="5.0">
+              <columnConstraints>
+                <ColumnConstraints hgrow="SOMETIMES" />
+                <ColumnConstraints hgrow="SOMETIMES" />
+                  <ColumnConstraints hgrow="SOMETIMES" />
```
DanVanAtta: Any idea if xmllint or spotless works on fxml files?
If so, any objections to using either?

game-headed/src/main/java/org/triplea/game/client/parser/DefaultGameLoader.java
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  @Override
+  protected Task<GameData> createTask() {
+    return new Task<>() {
+      @Override
+      protected GameData call() throws GameParseException {
+        return gameChooserEntry
+            .getCompleteGameData()
+            .orElseThrow(() -> new GameParseException("Missing file!"));
```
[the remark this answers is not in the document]
RoiEXLab: I can't remember the exact details but I think the reason I chose `GameParseException` here is so that the exception type is consistent.
I thought that I had a try catch block somewhere where this would make sense, but we just have a generic error handler, so yeah any exception would do.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
RoiEXLab: I don't know if you noticed, but I changed the way how users disable nations in the new UI:
Instead of unchecking a checkbox to disable a player, there is a "Disable" option when choosing who controls the specific nation, which feels more natural IMO.
However `PlayerListing` is used to this checkbox based scheme so I have to convert the data representation so it matches the "old"/current scheme.

So I could add an overload constructor, but I think this would ultimately lead down a rabbit hole of potential improvements for the data scheme in general.

game-headed/src/main/java/org/triplea/game/client/parser/DefaultGameLoader.java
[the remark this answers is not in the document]
DanVanAtta: How about changing to `FileNotFound` or a `IllegalState`? `GameParseException` AFAIK is more about invalid syntax rather than missing game file.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
DanVanAtta: I did not quite notice that, there is quite a bit going on in this update and I'm afraid to say time is too limited to spend enough time to really properly review everything without further pointers/guidance to call important elements out.

An overload of `PlayerListing` is pretty limited, why would that lead to a rabbit hole? Just add only the overload, then the class here in question does not have to pass unused/null values.

game-headed/src/main/resources/org/triplea/game/client/ui/javafx/fxml/RoleSelectionScreen.fxml
DanVanAtta: This is a side-note for consideration, not to be done in this PR.

game-headed/src/main/java/org/triplea/game/client/ui/javafx/screens/RoleSelection.java
RoiEXLab: I feel like the line between what should be UI-only and what is logic is really blurry here.
Your suggestion sounds intruiging, but I'm not quite sure how that would look like for the cases where the view has to query the model for information, for example when pressing on any alliance button next to the control-type selection (AI/Human/Disabled).
Also I'm not quite happy with the disable text thingy either, I simply haven't found a good way to store the intermediate selection state and synchronize it with the UI, so I just try to convert the UI back to an "internal" repreentation we can work with.
Maybe that's just me and this isn't a problem at all, but if you can show me a good pattern that works both ways (from view to model and vice-versa) then I'd happily apply it everywhere.

game-headed/src/main/resources/org/triplea/game/client/ui/javafx/fxml/RoleSelectionScreen.fxml
RoiEXLab: Well FXML should be 100% valid XML AFAIK, but I'm using SceneBuilder for the UI (a WYSIWYG editor for the most part), so all of the XML is auto-generated anyways, so not much value in style checking that IMO.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
RoiEXLab: We could, but I'm pretty sure this would make the code longer and doesn't add to readability IMO

RoiEXLab:
Thanks for the review on both PRs @DanVanAtta I'll try to address the comments in the coming days 👍

game-headed/src/main/java/org/triplea/game/client/ui/javafx/screens/RoleSelection.java
DanVanAtta: Typically model interactions are mediated by the view model (afaik). Let's keep thinking about this, both read up on view-model pattern a bit. view-model is also agnostic about UI layer, which makes it really great for decoupling swing and potentially allowing more migration of UI. In essence it also allows for UI code to be far more pure, and hence better decoupling. I think there are some examples in the forum poster code where I'll try a view-model, it should give an example of the pattern as well, so let's keep this on our radar.

game-headed/src/main/resources/org/triplea/game/client/ui/javafx/fxml/RoleSelectionScreen.fxml
DanVanAtta: `WYSIWYG editor` for doing non-prototype work always makes me nervous : |

Anyways, when you get a chance, so others can jump in, please add some dev setup notes on which editor you are using and kindly provide a link to install it.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
DanVanAtta: Okay, it is a bit odd though as is, because:
1. enabled = !disabled
2. the predicate is not super clear, it's indirect, not clear why we would have a predicate or what the predicate really represents.

Perhaps confusing item (2) can be helped with naming. Instead of `enabledPredicate`, some (poor) suggestions, perhaps something like: `componentIsEnabled`, `componentIsNotDisabled`

game-headed/src/main/resources/org/triplea/game/client/ui/javafx/fxml/RoleSelectionScreen.fxml
RoiEXLab: Will do, it's pretty straightforward and IntelliJ integrates with it well.

I'd highly recommend trying it out. It makes dealing with details (like padding or visual effects like shadows etc.) easy to work with.
The main reason to use it is because there's no way to remember all of the available features, so having a couple of checkboxes to check and uncheck to see if it changes something comes in really handy most of the time.

Obviously it doesn't solve all problems, but it should allow non-coders to work on the UI if they want to bring in their own ideas

RoiEXLab: Obviously it's only useful for static content and to quickly prototype dynamic content (when you're not getting why it doesn't look like it's supposed to look)

game-headed/src/main/java/org/triplea/game/client/ui/javafx/screen/NavigationPane.java
```diff
@@ -54,13 +55,31 @@ public void switchScreen(final FxmlManager identifier) {
...
+    final Node screen = screens.get(identifier).getNode();
 
     if (root.getChildren().isEmpty()) {
       root.getChildren().add(screen);
     } else {
       root.getChildren().set(0, screen);
     }
+    final ControlledScreen<NavigationPane> controlledScreen = screens.get(identifier);
+    checkTypes(data, controlledScreen.getValidTypes());
+    controlledScreen.onShow(data);
```
RoiEXLab: TODO: Create a unit test for this check

game-headed/src/test/java/org/triplea/game/client/ui/javafx/screens/RoleSelectionTest.java
```diff
@@ -0,0 +1,149 @@
...
+    final ComboBox<String> comboBox = mock(ComboBox.class);
+    when(comboBox.getItems()).thenReturn(observableList);
+    roleSelection.setAllSelectorCheckbox(comboBox);
+
+    roleSelection.initialize();
+
+    assertThat(observableList, is(List.of(PlayerType.playerTypes())));
+  }
+
+  // TODO Add Test for onShow method
```
RoiEXLab: TODO: Add a Test case for the onShow method

RoiEXLab:
@DanVanAtta So I addressed the simple points and added a couple of test cases.
I still plan on adding the 2 test cases I mentioned in the diff as well as fixing the weird white border around the comboboxes.

About the view-model pattern, I'll try to implement something in that direction in the near future, but not as part of this PR. There are a handful of classes where we could apply such a pattern, so I'd really like to keep it consistent there.
I won't be having too much time in the coming weeks, but there are enough opportunities to do some coding for like half an hour every day, that's why I can't give some sort of ETA for the Pattern PR.

ron-murhammer:
@RoiEXLab I haven't tested this so just going by the screenshot. Overall, I like the direction. My one concern is readability of some of the text. I think we need to add more contrast so especially the black text is easier to read. Maybe do like a darkened or lightened box around the area where there is text/inputs?

DanVanAtta:
@ron-murhammer I think it's early to fine-tune the UI appearance, as commented earlier:

>  I presume we do plan to iterate quite a bit on the screens, layout, exact functionality, and further details like CSS and style. My impression that is that the core functionality and having a first MVP is really the key consideration here, and more should focus on general design/structure.

There area number of UI elements I'd like to revisit, there could be a lot of churn if we focus on it now.

RoiEXLab:
@DanVanAtta I'm done for now, I created some additional unit tests.
I'll follow up with a PR that removes the 2 PbF/PbEM buttons (as discussed in #5965) and then we can think about how to completely decouple the UI from the logic.

PR #5920, opened 2020-02-09
Simplify token parsing

RoiEXLab:
Follow-up to #5912

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/random/MartiDiceRoller.java
+++ b/game-core/src/main/java/games/strategy/engine/random/MartiDiceRoller.java
@@ -41,6 +41,8 @@ public final class MartiDiceRoller implements IRemoteDiceServer {
   private static final String DICE_ROLLER_PATH = "/MARTI.php";
 
   private final Pattern errorPattern = Pattern.compile("fatal error:(.*)!");
+  // Matches a comma separated list of integers like this:
+  // your dice are: 1,2,3 <p>
   private final Pattern dicePattern =
       Pattern.compile("your dice are:\\s*((?:\\d(?:,\\d+)*)?)\\s*<p>");
 
--- a/game-core/src/main/java/games/strategy/engine/random/MartiDiceRoller.java
+++ b/game-core/src/main/java/games/strategy/engine/random/MartiDiceRoller.java
@@ -5,7 +5,6 @@ import com.google.common.collect.ImmutableList;
 import games.strategy.engine.ClientContext;
 import games.strategy.engine.framework.system.HttpProxy;
 import java.io.IOException;
-import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
@@ -96,16 +95,15 @@ public final class MartiDiceRoller implements IRemoteDiceServer {
   }
 
   @Override
-  public int[] getDice(final String string, final int count)
-      throws IOException, InvocationTargetException {
+  public int[] getDice(final String string, final int count) throws DiceServerException {
     final Matcher errorMatcher = errorPattern.matcher(string);
     if (errorMatcher.find()) {
-      throw new InvocationTargetException(null, errorMatcher.group(1));
+      throw new DiceServerException(errorMatcher.group(1));
     }
 
     final Matcher diceMatcher = dicePattern.matcher(string);
     if (!diceMatcher.find()) {
-      throw new IOException("String '" + string + "' has an invalid format.");
+      throw new IllegalStateException("String '" + string + "' has an invalid format.");
     }
     return Splitter.on(',').omitEmptyStrings().splitToList(diceMatcher.group(1)).stream()
         .mapToInt(Integer::parseInt)
--- a/game-core/src/main/java/games/strategy/engine/random/MartiDiceRoller.java
+++ b/game-core/src/main/java/games/strategy/engine/random/MartiDiceRoller.java
@@ -9,11 +9,11 @@ import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
-import java.util.logging.Level;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.annotation.Nonnull;
 import lombok.Builder;
 import lombok.Getter;
-import lombok.extern.java.Log;
 import org.apache.http.HttpHost;
 import org.apache.http.HttpRequest;
 import org.apache.http.HttpResponse;
@@ -35,11 +35,14 @@ import org.apache.http.protocol.HttpContext;
 import org.apache.http.util.EntityUtils;
 
 /** A pbem dice roller that reads its configuration from a properties file. */
-@Log
 @Builder
 public final class MartiDiceRoller implements IRemoteDiceServer {
   private static final int MESSAGE_MAX_LENGTH = 200;
   private static final String DICE_ROLLER_PATH = "/MARTI.php";
+
+  private final Pattern errorPattern = Pattern.compile("fatal error:(.*)!");
+  private final Pattern dicePattern =
+      Pattern.compile("your dice are:\\s*((?:\\d(?:,\\d+)*)?)\\s*<p>");
 
   @Nonnull private final URI diceRollerUri;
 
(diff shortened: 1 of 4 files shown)


game-core/src/main/java/games/strategy/engine/random/MartiDiceRoller.java
```diff
@@ -35,12 +35,15 @@
...
 /** A pbem dice roller that reads its configuration from a properties file. */
-@Log
 @Builder
 public final class MartiDiceRoller implements IRemoteDiceServer {
   private static final int MESSAGE_MAX_LENGTH = 200;
   private static final String DICE_ROLLER_PATH = "/MARTI.php";
 
+  private final Pattern errorPattern = Pattern.compile("fatal error:(.*)!");
+  private final Pattern dicePattern =
+      Pattern.compile("your dice are:\\s*((?:\\d(?:,\\d+)*)?)\\s*<p>");
```
DanVanAtta: Any objection to adding a comment to explain this regex a bit and give an example of what it matches?
Can we simplify the regex, does it have to be this long or maybe strict? It is a bit scary looking and seems to be more specified than what we had. I'm not sure if we need it to be extremely "strict".

game-core/src/test/java/games/strategy/engine/random/MartiDiceRollerTest.java
```diff
@@ -0,0 +1,82 @@
...
+      MartiDiceRoller.builder()
+          .diceRollerUri(URI.create("http://uri.invalid"))
+          .ccAddress("")
+          .toAddress("")
+          .gameId("")
+          .build();
+
+  @Test
+  @SuppressWarnings("unchecked")
+  void successfulMessageExtraction() throws Exception {
```
DanVanAtta: nit, missing either display name and/or assertion comment to explain why we expect this test to pass. I'd lean to an assertion comment personally to spell it out a bit.

game-core/src/test/java/games/strategy/engine/random/MartiDiceRollerTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+  @Test
+  void unsuccessfulMessageExtraction() {
+    final Exception exception =
+        assertThrows(
+            InvocationTargetException.class, () -> martiDiceRoller.getDice(failureMessage, 0));
+    assertThat(
+        exception.getMessage(),
+        is(
+            " Emails [user@mail.invalid] are not registered."
```
DanVanAtta: nit, potentially over-specified test-case here. It's often not best to look for exact exception messages, instead if we check that key information is in the exception message than we get more flexibility. For example, we may want to change the message to be `Email 'user@mail.invalid' is not registered', which would fail this test. If we check instead that the marti URL and email address are present, then we are pretty sure the message is sufficient. 

Second, the marti URL being hardcode is potentially something that is likely to change at some point, is there a constant somewhere that represents the same?

game-core/src/test/java/games/strategy/engine/random/MartiDiceRollerTest.java
```diff
@@ -0,0 +1,82 @@
...
+                + " Please register them at https://dice.marti.triplea-game.org/register.php ."));
+    assertThat(exception.getCause(), is(nullValue()));
```
DanVanAtta: Why is this assertion useful? Is it needed for correctness that the exception thrown have no cause?

Side-note, it's really questionable already we are using exception for control-flow handling. The exception being thrown makes it hard to tell if we are seeing a case where the dice-roller is down, no network, or the totally expected case where a user is not yet registered.

game-core/src/main/java/games/strategy/engine/random/MartiDiceRoller.java
RoiEXLab: The "benefit" of this regex being rather strict is that it allows us to avoid any further exception handling down below.
So we could of course simply it quite a bit, but would then have to re-introduce a try catch block below that rethrows an exception.

But yeah adding a comment is a good idea

game-core/src/test/java/games/strategy/engine/random/MartiDiceRollerTest.java
RoiEXLab: I had mixed feelings when writing the test cases.
My number one goal was to ensure nothing breaks with this change. That's why I simply went ahead an just copy-pasted an actual response from MARTI (and changed the email afterwards) to ensure that 1. the tests work with the original code and 2. they work with my new code which so we can be rather sure that nothing breaks.

After that I wasn't sure if I should simplify the test to make it more readable, but also lose some of the safety guarantees

game-core/src/test/java/games/strategy/engine/random/MartiDiceRollerTest.java
RoiEXLab: Not really, it only sort-of ensures that this isn't some generic `InvocationTargetException` that is actually thrown for some legitimate reason.
I mean I'm convinced the designers of `InvocationTargetException` made the cause a mandatory argument for a reason in the constructor, in 99% of all cases there should be a cause

game-core/src/test/java/games/strategy/engine/random/MartiDiceRollerTest.java
DanVanAtta: That's a good approach in general, I think we can relax the assertions made in this line without losing that or over-simplifying.

game-core/src/test/java/games/strategy/engine/random/MartiDiceRollerTest.java
DanVanAtta: Hmm, isn't 99% of all cases a user not being registered? It seems like almost everyone hits that case and before this exception message *was* the thing to tell you to register. It was a common problem that people would not grok the error message and were just left confused how to make the system work.

Overall, I think the root problem is we are dealing with control-flow-by-exception-handling anti-pattern.

If we are verifying those criteria, we can add messages to the assertion to document the reasoning and perhaps check the instance of the exception as well.

RoiEXLab: Do you think it would help using an unchecked custom exception and throwing that and just catching a generic RuntimeException somewhere else?
In this case neither the exception message nor any exception causes would be relevant

DanVanAtta: My thinking, at a low level, there should be something that does the send test and returns a payload that has an error message or a success flag. Any exception would then be for cases where we can't connect or cannot send. If there is a difference in how we would recover from such a scenario, then different exceptions would be appropriate.

Catching unchecked exceptions, or throwing a custom exception and then doing a catch-all to catch any runtime exception does not seem super easy to grok. Exceptions are effectively glorified GOTO statements after-all. 

Getting back to this test case, I think the things to do, to make the test clear, would be:
1. add an assertion comment to explain why we are verifying the condition. You stated so much in response to my questions earlier: "it only sort-of ensures that this isn't some generic InvocationTargetException that is actually thrown for some legitimate reason."

2. Perhaps also verify the instance type of the exception is an `InvocationTargetException`

RoiEXLab: @DanVanAtta I re-did some of the Exceptions.
Basically explicit server error messages like "wrong/missing email" etc. now have their own exception type and the code has to explicitly catch that.
Semantic Errors like the server returning an invalid string that doesn't match the pattern is now treated as `IllegalStateException`, never explicitly caught and is supposed to be a likely programmer error.
All kind of network related exceptions are now treaded as such. As user-fixable errors and provides a link to the forum in case the error doesn't go away.

Let me know what you think, I tested all 3 scenarios (by changing the source code to invalid servers/URLs and entering non existing emails) and I believe the error pop-up seems appropriate in each case and provides the right amount of information to debug the issue.

game-core/src/main/java/games/strategy/engine/random/MartiDiceRoller.java
```diff
@@ -96,16 +95,15 @@ public String postRequest(
...
     final Matcher errorMatcher = errorPattern.matcher(string);
     if (errorMatcher.find()) {
-      throw new InvocationTargetException(null, errorMatcher.group(1));
+      throw new DiceServerException(errorMatcher.group(1));
     }
 
     final Matcher diceMatcher = dicePattern.matcher(string);
     if (!diceMatcher.find()) {
-      throw new IOException("String '" + string + "' has an invalid format.");
+      throw new IllegalStateException("String '" + string + "' has an invalid format.");
```
DanVanAtta: Nit, we should keep in mind the audience of exception messages. Things like "String" is geared towards a developer, would not otherwise make sense to an end user. They want to know what happened and what they should do about it if anything.

With that said, this exception type and message are fine for now. We may though want to consider creating some infrastructure that would suffix an error message with something like "This is an unexpected error and indicates a bug. Please report this to TripleA development"

DanVanAtta:
Thanks for making the updates :100: 
LGTM :+1:

PR #5936, opened 2020-02-15
Forum poster structure

DanVanAtta:
- Collapses children of NodeBbForumPoster into parent class. This is done by parameterizing the 'abstract' method implementations and injecting them into the constructor of the parent class. This is done via a 'builder parameter object' refactor. Following this we can further combine interface methods from IForumPoster into NodeBbForumPoster to reduce unnecessary indirection and remove implementation from the interface.

The changes are structured to be relatively reviewable commit-by-commit, can maybe be reviewed all at once as well.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done


DanVanAtta:
> It's just a "lucky" coincidence that both supported forums are using NodeBB at the moment, but forum software changes fast, so at some point we might have to introduce a different implementation.

The `IForumInterface` had some problems and it's impossible to tell if any such future implementations would work with the given interface.

Notably the interface is very polluted and had methods appended to it to make things work. It's IMO an example of a leaky/poor abstraction.

Looking at the methods:

```
  static IForumPoster newInstanceByName(final String name, final int topicId) {
```
This is bad as the interface was acting as a factory and was aware of all implementations and the `name` value that each implementation would correspond to. 

```
  static ImmutableSet<String> availablePosters() {
```
Again, the interface is aware of all implementations (this is not good, an interface is meant to decouple implementations from the interface spec).

```
  static boolean isClientSettingSetupValidForServer(final String server) {
```

Same thing about interface being aware of all implementations, but even worse the interface is now aware of the mechanism where each implementation loads its properties. Such a method prevents us from just adding a new implementation without updating the interface.

```
  String getTestMessage();
  CompletableFuture<String> postTurnSummary(String summary, String subject, Path savegame);
  String getDisplayName();
```

Are we sure that the above API will be corerct? Perhaps the `Path` save game should be a `byte[]` for another forum poster, perhaps there would be no summary and subject and instead a topic id, maybe not all forums would post a test message.

I don't know if there is a really great interface we can abstract to, or should at this point. We only have two examples, they are largely driven by configuration, the concrete class that implements the forum posting itself does present an interface. I'm thinking `postGame(String subject, File saveGame)` might be the right abstraction, though at this point it's premature abstraction IMO. Better to have simpler, easier to change code so we can modify it to fit future requirements than dig ourselves out from potentially inappropriate abstractions.

DanVanAtta:
Should also mention:
```
  String getDisplayName();
```

This was not a good interface method, there is nothing that requires it, this was just glue needed for UI code and eventually forum poster code to call the right implementation. It's pretty clearly a result of extracting an interface from an existing concrete class, it's there because it was part of a specific implementation.

DanVanAtta:
Sorry for long response that perhaps seemed harsh -  appreciate the feedback and consideration. TL;DR, the code is already quite tied to fairly specific implementations and simply extracting an interface at this junction, without it being clearly the right interface, is not necessarily the right thing to do and is very at risk of being a wrong abstraction.

RoiEXLab:
No worries, I really expected that answer.
I know that the interface had problems, especially the one that it knew its implementations too well, but at the same time I wanted to hear your unbiased opinion, that's why I intentionally left out the reason I suspected to be the reason for your actions.

PR #5950, opened 2020-02-17
Do not double handle map not found exception

DanVanAtta:
Currently if a map not found exception is raised when loading a game
we both prompt a user if they would like to download the map and
secondly we notify that the game failed to start and provide
an option to upload an error report.

This update removes the latter handling so that when a map is not
found, we simply notify and prompt the user to download the missing
map.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done

- using preferences, set maps folder to any directory not containing maps
- load a map
- observed double or single error prompt handling

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
@@ -78,6 +78,9 @@ public class LocalLauncher extends AbstractLauncher<ServerGame> {
       game.setRandomSource(randomSource);
       gameData.getGameLoader().startGame(game, gamePlayers, launchAction, null);
       return Optional.of(game);
+    } catch (final MapNotFoundException e) {
+      // The throwing method of MapNotFoundException notifies and prompts user to download the map.
+      return Optional.empty();
     } catch (final Exception ex) {
       log.log(Level.SEVERE, "Failed to start game", ex);
       return Optional.empty();


DanVanAtta:
In short, yes to both questions @RoiEXLab .

One nuance/point to be clear about:
- severe logging should be for unrecoverable exceptions
- severe logging prompts user to upload an error report, this should only be done for unexpected errors that caused a crash (hence, unrecoverable).
- last, it's really important we keep in mind we are dealing with a thick client when it comes to logging and not a web server. On a web server, all exceptions are typically logged to show that an error has happened (even if somehow the webpage gracefully degrades, you still want to know the of situation). On a thick client, any error popup will usually cause the user to restart the application, no matter what the error is. In essence, we need to be sure we handle errors as much as possible, and if not then we would want an error report to either give user a proper notification of an expected failure or simply just suppress the message because we handled the error in some way. EG: no network connection -> show user a prompt and disable network dependent buttons. EG: no password and can't post to forum -> show user a pop-up, and do not log anything. EG: can't post due to forum being down -> let the exception be thrown, severe log it, give user option to upload error report (or alternatively we handle this by doing retries and then we let user upload an error report).

tl;dr: the error handling in TripleA is a bit sub-par in that we see control-flow by error handling and do not handle basic "errors" (anything that is user driven is not an error, eg: missing password), or alternatively we tend to log as if the application were a web-server, logging is not very useful on a thick-client app.

Hopefully that wasn't too pedantic.. I'll see about moving the above logging & error handling notes to a wiki to give guidance for the future.

DanVanAtta:
Added quick notes on error handling/logging to wiki: https://github.com/triplea-game/triplea/wiki/error-handling-and-logging

PR #5952, opened 2020-02-17
Remove rule-book links/buttons in favor of user-guide

DanVanAtta:
The user-guide page on website has links to the rule book, linking directly
to the rule-book is slightly redundant. Making the situation worse, the rule-book
PDF is becoming deprecated in favor of an HTML version.

This update removes links to the rule-book, everywhere we have a rule-book
link or button, we also had a link to the website 'user-guide' which in turn
has links to the latest rule-book.

 

## Functional Changes

[] New map or map update
[] New Feature
[x] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done

Verified (Swing) main screen layout still looked good after removing "rule book" button.

### DIFF
--- a/game-headed/src/main/resources/org/triplea/game/client/ui/javafx/fxml/AboutInformation.fxml
+++ b/game-headed/src/main/resources/org/triplea/game/client/ui/javafx/fxml/AboutInformation.fxml
@@ -15,11 +15,6 @@
             <Insets top="10.0" />
          </VBox.margin>
       </Button>
-      <Button mnemonicParsing="false" onAction="#showRuleBook" styleClass="normal-font-size" text="%main.button.rule_book">
-         <VBox.margin>
-            <Insets top="10.0" />
-         </VBox.margin>
-      </Button>
       <Button mnemonicParsing="false" onAction="#back" text="%main.button.back">
          <VBox.margin>
             <Insets top="10.0" />
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/MetaSetupPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/MetaSetupPanel.java
@@ -31,7 +31,6 @@ public class MetaSetupPanel extends SetupPanel {
   private JButton connectToHostedGame;
   private JButton connectToLobby;
   private JButton enginePreferences;
-  private JButton ruleBook;
   private JButton userGuideButton;
 
   private final SetupPanelModel model;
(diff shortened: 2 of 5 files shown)


game-headed/src/main/resources/org/triplea/game/client/ui/javafx/fxml/AboutInformation.fxml
```diff
@@ -15,11 +15,6 @@
             <Insets top="10.0" />
          </VBox.margin>
       </Button>
-      <Button mnemonicParsing="false" onAction="#showRuleBook" styleClass="normal-font-size" text="%main.button.rule_book">
```
RoiEXLab: You aren't supposed to know this, but this lang string is referenced in the `.lang` file we provide for all the translations.
So currently there's an unused translation string there that should be removed.

DanVanAtta: Good call out.

It's a bit odd to see so many other properties that are unused:
Screenshot from 2020-02-17 16-38-37

RoiEXLab: Well they aren't unused, but IntelliJ is obviously only able to find the static usages.
All the greyed out entries are the engine properties that derive their translation key based on the enum name

DanVanAtta: Is that true for these properties as well?

```
settings.tab.ai=AI
settings.tab.battle_simulator=Battle Simulator
settings.tab.combat=Combat
settings.tab.game=Game
settings.tab.map_scrolling=Map Scrolling
settings.tab.network_proxy=Network Proxy
settings.tab.network_timeouts=Network Timeouts
settings.tab.folder_locations=Folders
settings.tab.testing=Testing
```

Using the enum name to index into a language property seems a bit counter-intuitive. I wonder if it would make sense to duplicate the enum name as a string for the sake of getting an an exlicit lang string. The decoupling of enum name  and language property is maybe not a bad thing for explicit linking.

RoiEXLab: > Is that true for these properties as well?

I think so.

By decoupling you mean that we add a field to the enum that holds the key right so we use a hardcoded string instead of generating one.
I'm not sure if intelliJ is smart enough to realise this is the lang string so it wouldn't solve the unused problem,  AFAIK it only checks fxml files for javafx.
I agree that it would make things more "refactor-proof" even though we can change both things already without breaking anything, if we remember to change both.
Regardless this could become a little bit tricky to achieve nevertheless. See, almost every enum maps to more than one javafx node. I tried to keep the key naming scheme consistent, but there are still a lot of strings to take care of

DanVanAtta: Yeah, by decoupling, essentially if we wanted to rename the enum, or the string property key, the two could be updated independent of one another. 

NBD at this point, was mostly spitballin'

PR #5953, opened 2020-02-18
Forum poster view model

DanVanAtta:
Key Changes:
1. Group PBEMSetupPanel with 'pbem' editors in the same PBEM package
2. Remove 'EditorPanel' base class by having children set their own gridbaglayout and moving setLabel utility method to SwingComponents
3. PBF forum selection is now always set, default the drop-down box to first available selection
4. Extract logic from PBF editor into a view-model

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done
- verified PBF 'test post button'
- verified PBF 'view post button'
- verified 'test and view' buttons on PBF enabled when topic id is made invalid/valid
- saved a PBF game with non-default settings, loaded same game back on PBF screen and verified settings were restored

## Additional Review Notes

Things to pay particular attention to:
- notice the mechanism where PBF view updates are sent to view model (PBF view constructor)
- notice the view model listener interface, please check documentation there; that it reads well and makes sense
- please double check synchronization on view model methods makes sense
- please double check view model test cases make sense
- PR is structured so that changes will be more bite-sized if reviewed commit-by-commit


game-core/src/main/java/games/strategy/engine/data/properties/GameProperties.java
```diff
@@ -99,7 +99,7 @@ public String get(final String key, final String defaultValue) {
     if (value == null) {
       return defaultValue;
     }
-    return (String) value;
+    return String.valueOf(value);
```
DanVanAtta: This adds flexibility if we want to load an integer property as a string, otherwise we get a cast exception

game-core/src/main/java/games/strategy/engine/framework/startup/mc/GameSelectorModel.java
```diff
@@ -28,7 +28,7 @@
  */
 @Log
 public class GameSelectorModel extends Observable {
-  @Getter private GameData gameData = null;
+  @Nullable @Getter private GameData gameData = null;
```
DanVanAtta: GameData may be nulllable, marking it allows us to get IDE warnings for potential NPE

game-core/src/main/java/games/strategy/engine/framework/startup/ui/SetupPanel.java
```diff
@@ -33,7 +33,7 @@ public void setPanelChangedListener(final Consumer<SetupPanel> listener) {
     this.listener = listener;
   }
 
-  void fireListener() {
+  public void fireListener() {
```
DanVanAtta: Raising scope now that PBEMSetupPanel is in a different package

game-core/src/main/java/games/strategy/engine/framework/startup/ui/editors/EditorPanel.java
```diff
@@ -1,34 +0,0 @@
...
- * <p>This class used to have a legitimate reason to be a superclass of editors, but making all
- * methods in this class static, and converting it to a utility class is probably a good thing to
- * do.
- */
-abstract class EditorPanel extends JPanel {
-  private static final long serialVersionUID = 8156959717037201321L;
-  private final Color labelColor = new JLabel().getForeground();
-
-  EditorPanel() {
-    super(new GridBagLayout());
```
DanVanAtta: Layout is moved to children classes, 'setLabelValid' is moved to SwingComponents. The two changes eliminates the need for this base-class.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/editors/ForumPosterEditor.java
```diff
@@ -1,296 +0,0 @@
...
-import javax.swing.JTextField;
-import javax.swing.SwingUtilities;
-import lombok.extern.java.Log;
-import org.triplea.java.TimeManager;
-import org.triplea.swing.DocumentListenerBuilder;
-import org.triplea.swing.ProgressWindow;
-
-/** A class for selecting which Forum poster to use. */
-@Log
-public class ForumPosterEditor extends EditorPanel {
```
DanVanAtta: Most of the code in this class is moved but not modified. There are enough changes that git thinks it is a new file post-move, see individual commits for better diff'ing of updates.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/editors/ForumPosterEditor.java
```diff
@@ -1,296 +0,0 @@
...
-    setupListeners();
-  }
-
-  private void checkFieldsAndNotify() {
-    areFieldsValid();
-    readyCallback.run();
-  }
-
-  /** Configures the listeners for the gui components. */
-  private void setupListeners() {
```
DanVanAtta: Listener setup is inlined to constructor to update viewmodel based on component state, otherwise all other methods in this class are moved to 'viewModel' and/or broken up further.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/pbem/DiceServerEditor.java
```diff
@@ -223,13 +225,12 @@ private void checkFieldsAndNotify() {
 
   public boolean areFieldsValid() {
     final boolean toValid =
-        setLabelValid(
-            !toAddress.getText().isEmpty() && PlayerEmailValidation.isValid(toAddress.getText()),
-            toLabel);
+        !toAddress.getText().isEmpty() && PlayerEmailValidation.isValid(toAddress.getText());
+    SwingComponents.highlightLabelIfNotValid(toValid, toLabel);
```
DanVanAtta: Mostly just an API update here, as `highlightLabelIfNotValid` no longer returns a value.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/pbem/ForumPosterEditor.java
```diff
@@ -0,0 +1,201 @@
...
+  private final JLabel topicIdLabel = new JLabel("Topic Id:");
+  private final JLabel forumLabel = new JLabel("Forums:");
+  private final JCheckBox attachSaveGameToSummary = new JCheckBox("Attach save game to summary");
+  private final JCheckBox alsoPostAfterCombatMove = new JCheckBox("Also Post After Combat Move");
+  private final JComboBox<String> forums = new JComboBox<>();
+
+  ForumPosterEditor(final ForumPosterEditorViewModel viewModel) {
+    super(new GridBagLayout());
+
+    viewModel.setView(this);
```
DanVanAtta: Most of the code in this class is original, note anywhere there is a call to `viewModel`, that will be new as well as the 'updateState' method from `implements ViewModelListener<ForumPosterEditorViewModel>`

game-core/src/main/java/games/strategy/engine/framework/startup/ui/pbem/ForumPosterEditor.java
```diff
@@ -0,0 +1,201 @@
...
+            0,
+            GridBagConstraints.WEST,
+            GridBagConstraints.NONE,
+            new Insets(0, 0, 0, 0),
+            0,
+            0));
+  }
+
+  @Override
+  public void viewModelChanged(final ForumPosterEditorViewModel forumPosterEditorViewModel) {
```
DanVanAtta: Notice that all other methods in this view class are removed, we just have a `viewModelChanged` method now.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/pbem/ForumPosterEditor.java
```diff
@@ -0,0 +1,201 @@
...
+            0,
+            GridBagConstraints.WEST,
+            GridBagConstraints.NONE,
+            new Insets(0, 0, bottomSpace, labelSpace),
+            0,
+            0));
+
+    viewModel.getForumSelectionOptions().forEach(forums::addItem);
+    forums.addActionListener(e -> viewModel.setForumSelection((String) forums.getSelectedItem()));
+    forums.setSelectedItem(viewModel.getForumSelection());
```
DanVanAtta: Important to note:
1. Initial state is set by getting the view model value
2. We add action listeners to components to update view model to have the latest value set in the UI.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/pbem/ForumPosterEditor.java
```diff
@@ -0,0 +1,201 @@
...
+            1,
+            1.0,
+            0,
+            GridBagConstraints.EAST,
+            GridBagConstraints.HORIZONTAL,
+            new Insets(0, 0, bottomSpace, 0),
+            0,
+            0));
+
+    viewPosts.addActionListener(e -> viewModel.viewForumButtonClicked());
```
DanVanAtta: View actions, ie: button clicks, are now just simple calls to the view-model.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/pbem/ForumPosterEditorViewModel.java
```diff
@@ -0,0 +1,113 @@
+package games.strategy.engine.framework.startup.ui.pbem;
```
DanVanAtta: The only real new logic in this class is to default the forum selection, otherwise all the logic here is extracted from the view class. Notice as well that the button clicks are not here either, those are more properly model methods and we delegate those actions to other components. In this way the view model is responsible for modeling view state, the logic behind view state updates, and for triggering model interactions and forwarding the needed data from view to the model (for example when doing a test post, we trigger the test post method call and we pass the forum selection and forum id).

game-core/src/main/java/games/strategy/engine/framework/startup/ui/pbem/PbemSetupPanel.java
```diff
@@ -44,6 +47,7 @@
   private final GameSelectorModel gameSelectorModel;
   private final DiceServerEditor diceServerEditor;
   private final ForumPosterEditor forumPosterEditor;
+  private final ForumPosterEditorViewModel forumPosterEditorViewModel;
```
DanVanAtta: This class is just updated to insantiate a view model, pass that to the PBF view, and then instead of interacting with the PBF view for 'are all fields valid', we query the view model instead.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/pbem/PbemSetupPanel.java
```diff
@@ -184,11 +200,27 @@ public boolean canGameStart() {
...
+        "Game Data must not be null when starting a game, "
+            + "this error indicates a programming bug that allowed for the start game button to be "
+            + "enabled without first valid game data being loaded. ");
     if (diceServerEditor.areFieldsValid()) {
       diceServerEditor.applyToGameProperties(data.getProperties());
     }
-    if (forumPosterEditor.areFieldsValid()) {
-      forumPosterEditor.applyToGameProperties(data.getProperties());
+    if (forumPosterEditorViewModel.areFieldsValid()) {
+      final GameProperties properties = data.getProperties();
```
DanVanAtta: Instead of having the view (or view model) set game properties, which is mixing of domains, we instead get data from the view model and this class is now responsible for binding game properties to the properties we have read from UI. This way the UI model is not as tied to the model.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/pbem/PbemSetupPanel.java
```diff
@@ -214,6 +246,12 @@ public void update(final Observable o, final Object arg) {
   /** Called when the user hits play. */
   @Override
   public Optional<ILauncher> getLauncher() {
+    Preconditions.checkNotNull(
+        gameSelectorModel.getGameData(),
```
DanVanAtta: The null assumption for `getGameData()` is a bit lax, this is an example of 'fail-fast' principle to ensure we have a good error message and secondly to document an assumption that the selector model game data should be set on game launch.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/pbem/test/post/SwingTestPostProgressDisplayFactory.java
```diff
@@ -0,0 +1,45 @@
+package games.strategy.engine.framework.startup.ui.pbem.test.post;
+
+import games.strategy.engine.framework.GameRunner;
+import java.util.function.Supplier;
+import javax.swing.JOptionPane;
+import javax.swing.SwingUtilities;
+import org.triplea.swing.ProgressWindow;
+
+public class SwingTestPostProgressDisplayFactory implements Supplier<TestPostProgressDisplay> {
```
DanVanAtta: This class is to decouple swing interactions from "TestPostProgressDisplay"

game-core/src/main/java/games/strategy/engine/framework/startup/ui/pbem/test/post/TestPostAction.java
```diff
@@ -0,0 +1,76 @@
...
+  @Override
+  public void accept(final String forumName, final Integer topicId) {
+    Preconditions.checkNotNull(forumName);
+    Preconditions.checkArgument(topicId > 0);
+
+    final TestPostProgressDisplay testPostProgressDisplay = testPostProgressDisplayFactory.get();
+
+    new Thread(
+            () -> {
+              File f = null;
```
DanVanAtta: All logic in this method is extracted from the view. Any interactions with swing are delegated to `TestPostProgressDisplay`

game-core/src/test/java/games/strategy/engine/framework/startup/ui/pbem/ForumPosterEditorViewModelTest.java
```diff
@@ -0,0 +1,239 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.triplea.java.ViewModelListener;
+
+@ExtendWith(MockitoExtension.class)
+class ForumPosterEditorViewModelTest {
```
DanVanAtta: This test is pretty important, deserves some good scrutiny.

java-extras/src/main/java/org/triplea/java/ViewModelListener.java
```diff
@@ -0,0 +1,54 @@
+package org.triplea.java;
+
+/**
```
DanVanAtta: The comment and example on this interface are important. Please scrutinize whether it all makes sense.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/pbem/ForumPosterEditorViewModel.java
```diff
@@ -25,12 +43,29 @@
...
+    this.forumSelection = forumSelection;
+    forumSelectionValid = !forumSelection.isBlank();
+    viewForumPostButtonEnabled = areFieldsValid();
+    testForumPostButtonEnabled = areFieldsValid();
+    readyCallback.run();
+  }
+
+  synchronized void setTopicId(final String topicId) {
+    this.topicId = topicId;
+    final Integer numericTopicId = Ints.tryParse(topicId);
```
RoiEXLab: Why the `Ints.tryParse` method?
I think I have never seen that before.

java-extras/src/main/java/org/triplea/java/ViewModelListener.java
```diff
@@ -0,0 +1,54 @@
...
+ *       userName.setText(viewModel.getUsername());
+ *       userIsEnabledCheckBox.setSelected(viewModel.getUserIsEnabled());
+ *    }
+ * }
+ * }</pre>
+ *
+ * @param <ViewModelT> The type of view model, view model listeners should generally be 1:1 to a
+ *     view model.
+ */
+public interface ViewModelListener<ViewModelT> {
```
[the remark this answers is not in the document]
DanVanAtta: Certainly a consideration, though we do not ever expect any implementors to actually be lambdas.

java-extras/src/main/java/org/triplea/java/ViewModelListener.java
```diff
@@ -0,0 +1,54 @@
...
+ *        userIsEnabledCheckBox.addActionListener(
+ *            checkBoxSelected -> viewModel.setUserIsEnabled(checkboxSelected);
+ *    }
+ *
+ *    @Override
+ *    void viewModelChanged(ViewModelT viewModel) {
+ *       userName.setText(viewModel.getUsername());
+ *       userIsEnabledCheckBox.setSelected(viewModel.getUserIsEnabled());
+ *    }
+ * }
```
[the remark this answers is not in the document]
DanVanAtta: Very much recommend the Fowler link on the pattern: https://martinfowler.com/eaaDev/PresentationModel.html

There are also some similarities to these other patterns: 
- https://martinfowler.com/eaaDev/SupervisingPresenter.html
- https://martinfowler.com/eaaDev/PassiveScreen.html

In essence, a data only class is not very useful, part of the goal is that the logic of the UI itself should  become testable. If we are only a data class, then it's yet some other model that is containing that logic. In essence, the view model is a data class and represents the dynamic data of a UI, second it is a thin to medium sized model of the view.

For example with `getPasswordStrength`, the view model would store a value "password:string" and "passwordStrength:enum". The view would be updated and would know how to render "passwordStrength = low|medium|high". The view model would be invoked by the view on password text updates. The view model would have essentially controller like logic to invoke a model that computes the password strength (perhaps a score from 1-100 that is then translated to the 'low|medium|high' values). The view model would then pass the password strength enum value to the view to be rendered.

I suppose at its simplest the view model is just a data class and a submit button would send that data to some sort of submit action.

Some of the goals is that we decouple from UI, one could conceivably swap out UI components entirely without needing to update the view model. Second, the logic of updating the view and actions from the view become testable, also said in another way, logic of how the view is updated becomes testable. Last, the view becomes a pretty stable rendering component, it loses pretty much all logic and control flow and is just focused on rendering a given state.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/editors/ForumPosterEditor.java
```diff
@@ -179,118 +177,19 @@ public ForumPosterEditor(final Runnable readyCallback) {
...
-    forums.setSelectedItem(properties.get(IForumPoster.NAME));
-    final Object topicId = properties.get(IForumPoster.TOPIC_ID);
-    topicIdField.setText(topicId == null ? "" : String.valueOf(topicId));
-    alsoPostAfterCombatMove.setSelected(properties.get(IForumPoster.POST_AFTER_COMBAT, false));
-    includeSaveGame.setSelected(properties.get(IForumPoster.INCLUDE_SAVEGAME, true));
+    final boolean isValid = forumPosterEditorViewModel.areFieldsValid();
+    setLabelValid(isValid, topicIdLabel);
+    setLabelValid(isValid, forumLabel);
+    viewPosts.setEnabled(isValid);
+    testForum.setEnabled(isValid);
```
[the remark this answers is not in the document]
DanVanAtta: There certainly are similarities :smile:

game-core/src/main/java/games/strategy/engine/framework/startup/ui/pbem/ForumPosterEditorViewModel.java
DanVanAtta: It is Guava, returns `null` if the string value is not an integer. This avoids using `IntegerParseException` to check validity. It seemed really annoying there is no `isInt(String)` method baked into java yet.

DanVanAtta:
@RoiEXLab I added a commit to remove an unnecessary update from view-model to view. The updates should generally be pretty linear and clear which methods on view model are being set by view and which other methods are being set by other models that then need to cascade down to the view.

I'm thinking about threading a bit and perhaps we should follow-up. My thoughts are any updates from view probably should be done off of the AWT EDT thread, and the updates in the view probably should always be done on a swing invokeLater thread. With luck that kind of change will mean we are pretty guaranteed to avoid any computation on the swing rendering thread and second we should likely avoid any situation where we update UI  on a wrong thread.

Issue #5856, opened 2020-01-20
PBEM/PBF 'Play' button is never enabled


DanVanAtta:
**How can the problem be recreated?**
1. Launch game
2. Select PBF/PBEM
3. Note, despite filling in all fields, that 'Play' button stays disabled

**When the problem is fixed, what should have happened instead?**
- Play button should be enabled when minimum conditions have been met
- We probably should update the play button tooltip when PBF/PBEM is entered to be specific about what should be filled in

**Which Engine Version are you using?** 
df5a2242e8fba95182fe8a9376da3db12b28a29c

**Screenshots** 
![Screenshot from 2020-01-19 16-38-32](https://user-images.githubusercontent.com/12397753/72691456-5f0a1800-3ada-11ea-957f-0a4094d3147f.png)

**Zipped Save Game**

RoiEXLab:
Have you configured an email server in your engine preferences?
I wanted to add an information dialog some time ago, but never actually did it

RoiEXLab:
Second thought: I think I wanted to make configuring the email setting in engine preferences optional, so that the engine would ask if no credentials are available.
Then there would be no reason to require the credentials in this screen in the first place.
I even tried this in the past, but failed at some point for some reason I forgot

DanVanAtta:
Is this a convenience item added? How much of the email server config needs to be set before the play button could be enabled?

RoiEXLab:
For email you need to configure a full SMTP config; Hostname, Port, security protocol, username password.
For PbF you just need to enter your username and password and specify the server you want to configure it for

DanVanAtta:
@RoiEXLab is there more background I can read on this? Why is SMTP server config a new requirement? 

Settings window was not intended to necessarily be the only place where settings are. They should generally be available where/when needed. If for example, you say "don't show this dialog again", then settings is a good place to turn it back on. Secondarily, to remove options from menus and consolidate the settings that were only available from the main loading screen to those that were only available in game (just because they were on different windows/screens).

This ticket certainly, at least IMO, represents something that must be fixed before 2.0 can be released. Specificially:
- play button being disabled needs to state what needs to be done to enable it
- a smpt server should not be required
- convenience settings/configuration for PBEM shoudl be available from that screen (in addition to settings window)

If we go to this length, it may be easy/reasonable enough to fix the PBEM/PBF multi-screen and split them into their own screens.

RoiEXLab:
[...]

PR #5954, opened 2020-02-19
Split main screen PBEM/PBF button to two

DanVanAtta:
Splits PBEM/PBF button into two and each into their own dedicated screen.

- Improves UX where we do not overload a single button for two purposes
- Should make simplifying the classes easier. Next steps will be aimed to moving required preferences settings to the same screen (https://github.com/triplea-game/triplea/issues/5856)

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[] Manual testing done

### Before
Screenshot from 2020-02-18 19-17-09

Screenshot from 2020-02-18 19-17-33

### After

Screenshot from 2020-02-18 19-17-00

Example of PBF screen, PBEM is similar, just contains one purpose specific tab.

Screenshot from 2020-02-18 19-17-18

## Additional Review Notes
- PbemSetupPanel is gratuitously duplicated. This duplication should not cause too many issues as we will be looking to replace such classes, and/or we'll have better opportunity to simplify a purpose specific panel rather than one that needs to be abstract. The goal of this update is more to accomplish the screen and button split before further updates take place (placing more value here on accomplishing the UX change before we do more under-the-hood work and enabling a better path for simplifications).

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/PbemSetupPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/PbemSetupPanel.java
@@ -42,7 +42,7 @@ import org.triplea.swing.jpanel.GridBagConstraintsFill;
  * panel, and we notify it through the notifyObservers()
  */
 public class PbemSetupPanel extends SetupPanel implements Observer {
-  private static final long serialVersionUID = 9006941131918034674L;
+  private static final long serialVersionUID = -4027051961383144244L;
   private final GameSelectorModel gameSelectorModel;
   private final DiceServerEditor diceServerEditor;
   private final EmailSenderEditor emailSenderEditor;
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/PbemSetupPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/PbemSetupPanel.java
@@ -2,7 +2,6 @@ package games.strategy.engine.framework.startup.ui.posted.game.pbem;
 
 import com.google.common.base.Preconditions;
 import games.strategy.engine.data.GameData;
-import games.strategy.engine.data.properties.GameProperties;
 import games.strategy.engine.framework.message.PlayerListing;
 import games.strategy.engine.framework.startup.launcher.ILauncher;
 import games.strategy.engine.framework.startup.launcher.LocalLauncher;
@@ -12,9 +11,6 @@ import games.strategy.engine.framework.startup.ui.PlayerSelectorRow;
 import games.strategy.engine.framework.startup.ui.PlayerType;
 import games.strategy.engine.framework.startup.ui.SetupPanel;
 import games.strategy.engine.framework.startup.ui.posted.game.DiceServerEditor;
-import games.strategy.engine.framework.startup.ui.posted.game.pbf.ForumPosterEditor;
-import games.strategy.engine.framework.startup.ui.posted.game.pbf.ForumPosterEditorViewModel;
-import games.strategy.engine.posted.game.pbf.IForumPoster;
 import games.strategy.engine.random.PbemDiceRoller;
 import java.awt.Dialog;
 import java.awt.Dimension;
@@ -49,8 +45,6 @@ public class PbemSetupPanel extends SetupPanel implements Observer {
   private static final long serialVersionUID = 9006941131918034674L;
   private final GameSelectorModel gameSelectorModel;
   private final DiceServerEditor diceServerEditor;
-  private final ForumPosterEditor forumPosterEditor;
-  private final ForumPosterEditorViewModel forumPosterEditorViewModel;
   private final EmailSenderEditor emailSenderEditor;
   private final List<PlayerSelectorRow> playerTypes = new ArrayList<>();
   private final JPanel localPlayerPanel = new JPanel();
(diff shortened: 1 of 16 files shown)


Cernelius:
I understood or believed that PBF was a subset of PBEM, as whenever you PBF, you also PBEM.

So, when you say PBEM, are you excluding PBF, instead?

This is what the current button is unclearly saying, in the moment you have "Start PBEM (Play-By-Email/Forum) Game", it seems like you are affirming that "Play-By-Forum" is a type of "PBEM" gameplay, though I disapprove the redundancy of showing "Play-By-Email" as a subset of "PBEM".

If my understanding is correct, then the button should have been something like "Start PBEM (Private/Forum) Game".

For example, in the forum, you can post in PBEM for searching both for PBF and not PBF opponents alike:
https://forums.triplea-game.org/topic/37/do-not-make-play-by-forum-games-in-this-section

Would this change make "off topic" searching for PBF opponents inside the PBEM forum section?

However if, as I assumed so far, PBF is a subset of PBEM, then it would be good having a word for defining a PBEM game that is not PBF.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/PbemSetupPanel.java
```diff
@@ -0,0 +1,245 @@
...
+import org.triplea.swing.jpanel.GridBagConstraintsAnchor;
+import org.triplea.swing.jpanel.GridBagConstraintsBuilder;
+import org.triplea.swing.jpanel.GridBagConstraintsFill;
+
+/**
+ * A panel for setting up Play by Email/Forum. This panel listens to the GameSelectionModel so it
+ * can refresh when a new game is selected or save game loaded The MainPanel also listens to this
+ * panel, and we notify it through the notifyObservers()
+ */
+public class PbemSetupPanel extends SetupPanel implements Observer {
```
RoiEXLab: While I appreciate the split between those 2 classes, (I believe at some point it was even possible to play both PbEM and PbF at the same time 🤷‍♂) I think simply copying over 80% of the code isn't really something I'd consider clean code.

Did you consider creating an abstract `PlayByX` class that implements the common functionality and leaves the details for each subclass to decide, simply so we don't end up with that much duplicate code.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/PbemSetupPanel.java
```diff
@@ -0,0 +1,245 @@
...
+  private static final long serialVersionUID = 9006941131918034674L;
```
RoiEXLab: This serialVersionUID is a duplicate of the other class, it should be changed so that it is unique, even though we don't really make use of this property.

DanVanAtta:
Agree the copy paste is not clean, but it is not meant to be. We are
looking to replace with javafx. Creating an abstraction now is likely
wasteful

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/PbemSetupPanel.java
DanVanAtta: Could be nice to make it unique, but not required: https://github.com/triplea-game/triplea/pull/5347#discussion_r334496182

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/PbemSetupPanel.java
DanVanAtta: To respond directly to this now that I have a few more moments, the additional comments section addressed this concern directly:
> PbemSetupPanel is gratuitously duplicated. This duplication should not cause too many issues as we will be looking to replace such classes, and/or we'll have better opportunity to simplify a purpose specific panel rather than one that needs to be abstract. The goal of this update is more to accomplish the screen and button split before further updates take place (placing more value here on accomplishing the UX change before we do more under-the-hood work and enabling a better path for simplifications).

> Did you consider creating an abstract PlayByX class that implements the common functionality and leaves the details for each subclass to decide, simply so we don't end up with that much duplicate code.

Yes, an abstract class based around what we have now could easily lead to more complexity, or alternatively we might invest in cleaning up UI code that has many issues when we are looking to replace/delete it entirely. 

I'd note, that DRY does not mean to do human pattern matching on code and eliminate all common patterns, but to condense ideas to be expressed in one place. For the perspective of a single screen, having similar if not duplicated UI code, given we have only two instances, and the only thing they share thus far is dice roller, the duplication is actually not that bad. 

Second, the code smell is just a smell right now, I don't think this will actually cause any problems or missed updates where an update that shoudl be applied to both instances would actually only be used in one.

Third point regarding the duplication, it could easily be better to inline the PBF and PBEM editor and then clean-up/refactor from there and leave only the dice editor as the common and extracted code.

DanVanAtta:
@RoiEXLab ball is still in your court on this one.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/PbemSetupPanel.java
RoiEXLab:
```suggestion
  private static final long serialVersionUID = -4027051961383144244L;
```

RoiEXLab:
@DanVanAtta I think it would  be nice to keep the serialversionUID unique, I created a suggestion for you. Feel free to merge with or without the change.
Have a look at #5965
I summarised my thoughts about the game navigation there, which would make those kind of setuppanels redundant,  but maybe you disagree.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/PbemSetupPanel.java
DanVanAtta: I don't think anyone would know this value is duplicated anywhere, except us for now.. If feeling important for the serialversionuid to be different, then the 0L and 1L values should arguably be updated too. I don't think there is a favorable ROI for even really worrying about this too much.

PR #5980, opened 2020-02-25
Add username and password fields to PBF selection screen

DanVanAtta:
The two fields are required in order to do a test posting or
to start a PBF game. This update places the two required fields
on the same screen where we have the 'play' button.

This update adds similar 'highlight field label red if invalid'
for the username and password field.

Of note, to avoid circular event callbacks between view and
view-model, the view updates its own state after firing an
event on the view-model. For example, the username field
sets the view-model username value, the view then queries
and updates to see if any labels or buttons are updated and
enabled. The view model could invoke the view to update all
fields, this can then trigger another callback though to the
view when fields are updated and action listeners fire to
update the view model which can then again trigger a view
refresh and so-on.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done
Verified:
-  fields are updated when selecting forum option
- play button is enabled/disabled when setting/unsetting fields
- view forum button works for both forum types and is enabled/disabled
- test post button is enabled/disabled as username/password/topic fields are all set/unset

### Before
Screenshot from 2020-02-24 22-50-32

### After
Screenshot from 2020-02-24 22-49-58

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbf/ForumPosterEditorViewModel.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbf/ForumPosterEditorViewModel.java
@@ -9,6 +9,7 @@ import games.strategy.engine.framework.startup.ui.posted.game.pbf.test.post.Swin
 import games.strategy.engine.framework.startup.ui.posted.game.pbf.test.post.TestPostAction;
 import games.strategy.engine.posted.game.pbf.IForumPoster;
 import games.strategy.engine.posted.game.pbf.NodeBbForumPoster;
+import games.strategy.triplea.settings.ClientSetting;
 import java.util.Optional;
 import java.util.function.BiConsumer;
 import lombok.Getter;
@@ -30,13 +31,12 @@ class ForumPosterEditorViewModel {
   private BiConsumer<String, Integer> viewForumPostAction =
       (forumName, topicId) -> NodeBbForumPoster.newInstanceByName(forumName, topicId).viewPosted();
 
-  @Getter private boolean viewForumPostButtonEnabled;
-  @Getter private boolean testForumPostButtonEnabled;
   private String forumSelection;
   @Getter private String topicId = "";
-  @Getter private boolean topicIdValid;
   @Setter @Getter private boolean attachSaveGameToSummary = true;
   @Setter @Getter private boolean alsoPostAfterCombatMove;
+  @Getter private String forumUsername;
+  @Getter private char[] forumPassword;
 
   ForumPosterEditorViewModel(final Runnable readyCallback) {
     this.readyCallback = readyCallback;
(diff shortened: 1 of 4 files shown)


game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbf/ForumPosterEditorViewModel.java
```diff
@@ -48,6 +48,28 @@
...
+        forumSelection.equals(NodeBbForumPoster.TRIPLEA_FORUM_DISPLAY_NAME)
+            ? ClientSetting.tripleaForumUsername
+            : ClientSetting.aaForumUsername;
+
+    usernameSetting.setValueAndFlush(username.toCharArray());
+    forumUsername = username;
+    readyCallback.run();
+  }
+
   @SuppressWarnings("Guava")
```
RoiEXLab: Just noticed this: It should be easily removable now that guava's `Predicates#not` can be replaced with [`Predicate#not`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/function/Predicate.html#not(java.util.function.Predicate)) from the standard library

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbf/ForumPosterEditorViewModel.java
```diff
@@ -60,8 +82,14 @@ synchronized void setForumSelection(final String forumSelection) {
...
-    viewForumPostButtonEnabled = areFieldsValid();
-    testForumPostButtonEnabled = areFieldsValid();
+    forumUsername =
+        this.forumSelection.equals(NodeBbForumPoster.TRIPLEA_FORUM_DISPLAY_NAME)
+            ? ClientSetting.tripleaForumUsername.getValue().map(String::valueOf).orElse("")
+            : ClientSetting.aaForumUsername.getValue().map(String::valueOf).orElse("");
+    forumPassword =
+        this.forumSelection.equals(NodeBbForumPoster.TRIPLEA_FORUM_DISPLAY_NAME)
+            ? ClientSetting.tripleaForumPassword.getValue().orElse(new char[] {})
+            : ClientSetting.aaForumPassword.getValue().orElse(new char[] {});
```
RoiEXLab:
```suggestion
            ? ClientSetting.tripleaForumPassword.getValue().orElse(new char[0])
            : ClientSetting.aaForumPassword.getValue().orElse(new char[0]);
```

game-core/src/test/java/games/strategy/engine/framework/startup/ui/posted/game/pbf/ForumPosterEditorViewModelTest.java
```diff
@@ -129,45 +131,49 @@ void viewForumPostNoOpOnInvalidData(final String forumSelection, final String to
...
+        // invalid topic IDs
+        Arguments.of("not a number", "user", new char[] {'a'}),
+        Arguments.of("", "user", new char[] {'a'}),
+        Arguments.of("0.0", "user", new char[] {'a'}),
+        Arguments.of("0", "user", new char[] {'a'}),
+        Arguments.of("-1", "user", new char[] {'a'}),
+        // missing user
+        Arguments.of("1", "", new char[] {'a'}),
+        // missing password
+        Arguments.of("1", "", new char[] {}));
```
RoiEXLab:
```suggestion
        Arguments.of("1", "", new char[0]));
```

But in this scenario I would be ok with keeping the initializer curly braces because of consistency

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbf/ForumPosterEditorViewModel.java
DanVanAtta: Is this just personal preference @RoiEXLab ?

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbf/ForumPosterEditorViewModel.java
DanVanAtta: Ah, indeed, thanks. The auto-correct for the warning was giving a nasty ".negation()"
Updated in: 668a0dfd1

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbf/ForumPosterEditorViewModel.java
RoiEXLab: Well yes, it's shorter and tbh I'm not really a fan of empty curly braces anywhere.
Just a weird thing of mine

RoiEXLab:
Thanks for doing this so quickly 👍

PR #5986, opened 2020-02-29
ClientMessenger refactoring

RoiEXLab:
What started as a simple follow-up to #5984 turned out to be a rather satisfying refactoring procedure.
If you're interested in the details feel free to look at the individual commits in order to better understand what's happening there's a lot of stuff going on here.

I did manually test those changes, but I must admit that I didn't check compatibility, but I'm 99% sure IClientMessenger is not shared over the network.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/network/ui/SetMapClientAction.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/network/ui/SetMapClientAction.java
@@ -33,7 +33,7 @@ public class SetMapClientAction extends AbstractAction {
   public void actionPerformed(final ActionEvent e) {
     if (availableGames.isEmpty()) {
       JOptionPane.showMessageDialog(
-              parent, "No available games", "No available games", JOptionPane.ERROR_MESSAGE);
+          parent, "No available games", "No available games", JOptionPane.ERROR_MESSAGE);
       return;
     }
     final DefaultComboBoxModel<String> model = new DefaultComboBoxModel<>();
--- a/game-core/src/main/java/games/strategy/engine/framework/network/ui/SetMapClientAction.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/network/ui/SetMapClientAction.java
@@ -31,17 +31,15 @@ public class SetMapClientAction extends AbstractAction {
 
   @Override
   public void actionPerformed(final ActionEvent e) {
+    if (availableGames.isEmpty()) {
+      JOptionPane.showMessageDialog(
+              parent, "No available games", "No available games", JOptionPane.ERROR_MESSAGE);
+      return;
+    }
     final DefaultComboBoxModel<String> model = new DefaultComboBoxModel<>();
     final JComboBox<String> combo = new JComboBox<>(model);
     model.addElement("");
-    for (final String game : availableGames) {
-      model.addElement(game);
-    }
-    if (model.getSize() <= 1) {
-      JOptionPane.showMessageDialog(
-          parent, "No available games", "No available games", JOptionPane.ERROR_MESSAGE);
-      return;
-    }
+    model.addAll(availableGames);
     final int selectedOption =
         JOptionPane.showConfirmDialog(
             parent, combo, "Change Game To: ", JOptionPane.OK_CANCEL_OPTION);
--- a/game-core/src/main/java/games/strategy/engine/framework/network/ui/SetMapClientAction.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/network/ui/SetMapClientAction.java
@@ -1,7 +1,6 @@
 package games.strategy.engine.framework.network.ui;
 
-import games.strategy.net.IClientMessenger;
-import games.strategy.net.INode;
+import games.strategy.engine.framework.startup.mc.IServerStartupRemote;
 import java.awt.Component;
 import java.awt.event.ActionEvent;
 import java.util.Collections;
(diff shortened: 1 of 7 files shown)


game-core/src/main/java/games/strategy/engine/framework/network/ui/SetMapClientAction.java
```diff
@@ -17,29 +16,28 @@
...
   public void actionPerformed(final ActionEvent e) {
-    final INode serverNode = clientMessenger.getServerNode();
     final DefaultComboBoxModel<String> model = new DefaultComboBoxModel<>();
     final JComboBox<String> combo = new JComboBox<>(model);
     model.addElement("");
     for (final String game : availableGames) {
       model.addElement(game);
     }
-    if (serverNode == null || model.getSize() <= 1) {
+    if (model.getSize() <= 1) {
```
RoiEXLab: This is a change I'm not sure is 100% safe, but it worked no problem so I didn't want to pass clientMessenger as well just for this null check

ron-murhammer:
Always nice to get rid of more of the reflection mess.

RoiEXLab:
@ron-murhammer Well the code is now gone, but the mechanism still exists. The only difference is that the code now no longer exists twice.

This PR made me thinking though:
It looks like all non-lobby traffic is now performed via the RMI-like interfaces `IRemote` and `IChannelSubscriber`.
So all the network method calls go through the `UnifiedInvocationHandler` at some point. This means we could easily replace this component with something different, more future-proof.

This would be a lot of work though, a similar amount of work like @DanVanAtta put into the http-server project probably.

So here's the sort-of compromise I've been thinking about: The code currently relies on some "magic" method ordering that is easily broken by simply adding or removing a method or just renaming it to something else.
What if we instead calculated that magic number and hardcoded it as part of an annotation and removed the whole magic ordering part.
This way refactoring interfaces would be "safer" in the future while keeping backwards-compatibility with the current code. Thoughts?

RoiEXLab:
By the way, I just hacked together some code to get the full list of "numbers".
You can have a look at it here:
https://gist.github.com/RoiEXLab/b63ad68a29438fc6152e509d6deb4add

There are more methods than I expected, but using some hacky automation script in combination with the auto-formatter should be able to do 90% of the heavy lifting and add appropriate annotations, so that after that the only thing left would be to change the implementation of `RemoteInterfaceHelper` to check the annotations rather than the "method order"

DanVanAtta:
IMO, most important to make things:
- compile time safe
- backward compatible. Particularly considering a case where you can add new fields and if they are unused by old clients, things still would work.

Long term/medium term, most important to open a route where we can easily forward messages. This is not a new idea, but I still think would be critical, we could attach a 'game-id' to messages and then have bots just be message routers removing the need for bot code to have anything other than a network layer.

DanVanAtta:
For the sake of discussion, I also wonder if it would be overall easier and better long term to convert the network layer to either HTTP and/or websocket. The argument for that is it would be better to go directly to a longer term, better solution for say 5 points of effort rather than spend 3 points for a medium term that makes the longer term effort still cost 5 points.

RoiEXLab:
>  The argument for that is it would be better to go directly to a longer term

That's why I was asking. With my idea we'd get backward compatibility, but not really any compile-time safety. After all annotations are meant to be used via reflection for the most part.

DanVanAtta:
Probably worth discussing this in more detail via issue so it's not on the bottom of a PR. I like the idea of annotations a good bit. We could consider making the remote name a property of the annotation. This would at least make it obvious which methods are remotes and allow us to rename and re-order methods.

RoiEXLab:
@DanVanAtta We don't even need the actual method name, only the "index" of the method is ever sent over the network.
So the list I pasted in my gist is all we need for network compatibility, perhaps we should "hardcode" the interface name as well, but that's a slightly different project

DanVanAtta:
Changing the lookup from number to name should not be too bad. IMO that would be a good update for maintainability and avoid the issue of having to 'fix' method numbers, having them be out of order, or swapping them. The bytes we save for number over name is trivial, probably all fits in the same IP frame anyways; and is a pittance compared to the sometimes unnecessary game data copies or copies of other data payloads that are sent over network.

In the gist, there would be annotations on all of the methods that could be called? Is it actually complete?

RoiEXLab:
> Changing the lookup from number to name should not be too bad

I agree that the overhead is completely fine, but it would break compatibility which I wouldn't actually go for until we come up with a real long-term solution. 

> In the gist, there would be annotations on all of the methods that could be called?

Yes, all of the methods listed there would need an annotation with the id next to them.

> Is it actually complete?

I hope so. The way I understand the code it should be, given that guava's classpath scan works as expected. The code you can find in the gist was used to generate this file.
`IRemote` and `IChannelSubscriber` are the only "proxy objects" (see java.lang.reflect.Proxy#newProxyInstance in the code) that are used with the UnifiedInvocationHandler. There are other places where proxy objects are created, but I don't think they're related to networking.
So my code simply checks if the class on the classpath is an interface itself and extends at least one of the 2 interfaces. If that's the case it iterates over all methods and applies the "get-number-algirthm" on each one which is then printed to stdout.

DanVanAtta:
2.0 is not a bad time to break compatibility. Otherwise we may want to shoot for a longer term solution, I would hope that bot network proxy could be a thing for 3.0. That and save-game compatibility fixes (making save-game be a non-serialized format) are perhaps two of the biggest/best legacy upgrades we can have, and avoids so much pain for compatibility testing that is all manual for now.

PR #5990, opened 2020-03-02
Fix password too short error message when creating an account

DanVanAtta:
Previous password validator assumed we were looking at a hashed value,
now we are looking at a plaintext value. The server validation of
length is still needed as there is no guarantee someone will use
our game-client when creating accounts.

This fix updates password validation on account creation to simply
check that the password is a minimum length.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done
- validated that the larger flow of creating an account works.


http-server/src/main/java/org/triplea/server/user/account/create/PasswordValidation.java
```diff
@@ -8,12 +8,12 @@
 /** We expect password to be sent to server as a hash. Simply check that it is not too short. */
 public class PasswordValidation implements Function<String, Optional<String>> {
   /** Expected MIN_LENGTH is length of a SHA512 hashed string. */
```
RoiEXLab: The Javadocs are no longer accurate

PR #5992, opened 2020-03-02
Hardcode network opcodes

RoiEXLab:
This is my proposal for the thing discussed in #5986 
~I have no Idea if this does actually work, but for some reason gradle + lombok doesn't seem to be working anymore so I can really test this :(~

Ran a local network game ~(where both parties were on this branch, so no compatibility checks there)~ and after some initial problems the current version seems to run fine without any problems. (I got an NPE in the UI code at some point, but it doesn't seem related)

Tell me what you think on the overall design choices.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done

## Additional Review Notes
 In case you're interested, here's the code I used to add most of the annotations:
```python3
#!/usr/bin/env python3

import re

def main():
  with open('path/to/Op-Codes.txt', 'r') as f:
    for line in f:
      if line:
        if line[0] != '\t':
          file_path = line.replace('.', '/').strip().split('$')[0] + '.java'
        elif ':' in line:
          method, opcode = line.split(':')
          method = method.strip()
          opcode = int(opcode.strip())

          split_method = method.split('(')
          args = [s.split('.')[-1] for s in split_method[1][:-1].split(', ') if s]

          method_name = split_method[0]

          pattern = '\\r?\\n.*?' + re.escape(method_name) + '\\([^)]*?' + '[^)]*?'.join(re.escape(s) for s in args) + '[^,]*\)'

          print(pattern)

          # Needs to run in triplea/game-core/src/main/java
          with open(file_path, 'r') as f1:
            content = f1.read()

          content = re.sub(pattern, lambda m: ('\n@RemoteActionCode(%d)' % opcode) + m.group(0), content)

          with open(file_path, 'w') as f2:
            f2.write(content)

if __name__ == '__main__':
  main()
```

Not perfect, but good enough.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/chat/IChatChannel.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/IChatChannel.java
@@ -5,12 +5,7 @@ import games.strategy.engine.message.RemoteActionCode;
 import org.triplea.domain.data.UserName;
 import org.triplea.http.client.lobby.chat.ChatParticipant;
 
-/**
- * Chat messages occur on this channel.
- *
- * <p>RMI warning: the ordering of methods cannot be changed, these methods will be invoked by
- * method order number
- */
+/** Chat messages occur on this channel. */
 public interface IChatChannel extends IChannelSubscriber {
   // we get the sender from MessageContext
   @RemoteActionCode(0)
--- a/game-core/src/main/java/games/strategy/engine/chat/IChatChannel.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/IChatChannel.java
@@ -12,18 +12,24 @@ import org.triplea.http.client.lobby.chat.ChatParticipant;
  */
 public interface IChatChannel extends IChannelSubscriber {
   // we get the sender from MessageContext
+@RemoteActionCode(0)
   void chatOccurred(String message);
 
+@RemoteActionCode(2)
   void slapOccurred(UserName userName);
 
+@RemoteActionCode(3)
   void speakerAdded(ChatParticipant chatParticipant);
 
+@RemoteActionCode(4)
   void speakerRemoved(UserName userName);
 
   // purely here to keep connections open and stop NATs and crap from thinking that our connection
   // is closed when it is
   // not.
+@RemoteActionCode(1)
   void ping();
 
+@RemoteActionCode(5)
   void statusChanged(UserName userName, String status);
 }
--- a/game-core/src/main/java/games/strategy/engine/chat/IChatChannel.java
+++ b/game-core/src/main/java/games/strategy/engine/chat/IChatChannel.java
@@ -1,6 +1,7 @@
 package games.strategy.engine.chat;
 
 import games.strategy.engine.message.IChannelSubscriber;
+import games.strategy.engine.message.RemoteActionCode;
 import org.triplea.domain.data.UserName;
 import org.triplea.http.client.lobby.chat.ChatParticipant;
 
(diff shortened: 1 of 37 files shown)


RoiEXLab:
@DanVanAtta I did some compatibility testing and apparently it looks like the interfaces can be renamed at will, they are already relying on a "hardcoded" string to identify the appropriate object.
`RemoteName.name` stores this particular string.
So with this change the only thing that can't be changed is the parameter signature of all the methods, but they can be renamed freely without having to fear that compatibility is going to get broken.

game-core/src/main/java/games/strategy/triplea/delegate/remote/IEditDelegate.java
```diff
@@ -6,37 +6,48 @@
...
   String addUnits(Territory t, Collection<Unit> units);
 
+  @RemoteActionCode(6)
   String changeTerritoryOwner(Territory t, GamePlayer player);
 
+  @RemoteActionCode(3)
   String changePUs(GamePlayer player, int pus);
 
-  String changeTechTokens(GamePlayer player, int tokens);
-
```
RoiEXLab: Now with those changes this unused method can safely be removed

game-core/src/main/java/games/strategy/engine/chat/IChatChannel.java
```diff
@@ -1,29 +1,31 @@
...
- * <p>RMI warning: the ordering of methods cannot be changed, these methods will be invoked by
- * method order number
- */
+/** Chat messages occur on this channel. */
 public interface IChatChannel extends IChannelSubscriber {
   // we get the sender from MessageContext
+  @RemoteActionCode(0)
   void chatOccurred(String message);
 
+  @RemoteActionCode(2)
```
DanVanAtta: Should we re-order the methods so that the numbers would be ascending? I think that might help make it clear there are no gaps and just be nicer when reading the methods.

game-core/src/test/java/games/strategy/engine/message/unifiedmessenger/EndPointTest.java
```diff
@@ -2,25 +2,26 @@
...
+    final EndPoint endPoint = new EndPoint("", TestInterface.class, false);
+    endPoint.addImplementor((TestInterface) () -> 2);
     final RemoteMethodCall call =
-        new RemoteMethodCall(
-            "",
-            "compare",
-            new Object[] {"", ""},
-            new Class<?>[] {Object.class, Object.class},
-            Comparator.class);
+        new RemoteMethodCall("", TestInterface.class.getMethod("dummy"), new Object[] {});
```
DanVanAtta: Looks like we are still bound by method name (and ordering must be preserved, but at least that is in annotation now). Is there a way we could put the method name in the annotation as well?

game-core/src/main/java/games/strategy/engine/message/RemoteActionCode.java
```diff
@@ -0,0 +1,19 @@
+package games.strategy.engine.message;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation that sets the "opcode" for any particular method that may be used over the network
```
DanVanAtta: I'd perhaps rephrase this slightly.
- opcode does not have a well defined meaning, perhaps better to say something along the lines of: "Annotation that defines a method number, methods are called by method number using the {@link RemoteMethodCall} mechanism}". In order to stay compatible over the network, annotated methods should be altered (including: renaming, changes in signature, or changes in the method ordering number)."

game-core/src/main/java/games/strategy/engine/chat/IChatChannel.java
RoiEXLab: We could, but it would be a tedious process + some classes skip a couple of numbers and if we ever plan to overload a method checkstyle would force us to keep the overloadoverloaded versions together.

We could add this as a "nice to have" task, but this is not something I'd like to address as part of this PR.

game-core/src/test/java/games/strategy/engine/message/unifiedmessenger/EndPointTest.java
RoiEXLab: We are only bound in this particular test-case.
The RemoteMethodCall class works via reflection after all, so we need to pass the method object to the constructor somehow.
In non-test code this conversion is done by the JVM:
We create a "proxy" object that implements a particular interface. Whenever we call a method on this proxy object, the invocationhandler we registered is called with the corresponding method object, so the non-test code doesn't have any references to the method names whatsoever

PR #5997, opened 2020-03-03
Reset value on failure

RoiEXLab:
For some reason I managed to mess up the encoded credentials on my machine, so I was repeatedly getting exception dialogs whenever I was opening the engine preferences dialog or was actually using the properties.
Adding this one line made the errors go away on my machine, but I'm not sure if it's desireable to have the value auto-reset if something goes wrong 🤷‍♂ 
So I'll leave it up to you to decide if this change is actually useful

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/settings/ClientSetting.java
+++ b/game-core/src/main/java/games/strategy/triplea/settings/ClientSetting.java
@@ -350,7 +350,7 @@ public abstract class ClientSetting<T> implements GameSetting<T> {
       return decodeValue(encodedValue);
     } catch (final ValueEncodingException e) {
       log.log(
-          Level.WARNING,
+          Level.INFO,
           String.format(
               "Failed to decode encoded value: '%s' in client setting '%s'", encodedValue, name),
           e);
--- a/game-core/src/main/java/games/strategy/triplea/settings/ClientSetting.java
+++ b/game-core/src/main/java/games/strategy/triplea/settings/ClientSetting.java
@@ -354,6 +354,7 @@ public abstract class ClientSetting<T> implements GameSetting<T> {
           String.format(
               "Failed to decode encoded value: '%s' in client setting '%s'", encodedValue, name),
           e);
+      resetValue();
       return getDefaultValue().orElse(null);
     }
   }


ron-murhammer:
@RoiEXLab Probably a good idea. Though it might be good to update the log message to specify that its being reset then.

RoiEXLab:
Hmm I'm not even sure if we should even print an error message at all and simply silently "fix" any problems

ron-murhammer:
@RoiEXLab I think you could make that argument. I'm pretty open either way just figured the message should line up and mention resetting if its going to show to the user.

RoiEXLab:
@DanVanAtta I'd be happy to hear your opinion on this

DanVanAtta:
My 2 cents this:
>           Level.WARNING,
          String.format(
              "Failed to decode encoded value: '%s' in client setting '%s'", encodedValue, name),
          e);

Would be better as an info log message. The user has no concern over what the encoded value is, that is only a developer concern. In addition to the info logging, a warning logging would be justified to note which field had its value lost, and a message something like:
"An incorrect value was detected in FIELD, it has been reset to its default value. If this problem happens frequently, please file a <link>bug</link>". 

It's worth considering how we handle the error further than that though, the warning/severe logging will give a user an option to report the stacktrace, we'll want to consider if we want that and if the stack trace will have enough information. It seems like we are also addressing a band-aid, this is a good fix, but there is a deeper problem out there as well and we are treating a symptom. It might be good to do a similar check after setting an encoded value to make sure it can be ready back (fail-early principle, AKA "fail at the time of error" principle).

RoiEXLab:
@DanVanAtta I lowered the log level.

> to make sure it can be ready back

I actually don't think this will fix the issue. I'm pretty sure such a check would only fail if the developer messes up and unit tests are better suited for this type of task.
I assume that this error was introduced for me simply because I switched between branches at some point where we changed something about the store or something and therefore the "wrong" thing got saved in an older version and the newer version was no longer able to read that correctly or the other way round.

DanVanAtta:
> I actually don't think this will fix the issue.

No, it would not. It would though move the time of error to when the incorrect value as written "fail-early principle"

>  I'm pretty sure such a check would only fail if the developer messes up and unit tests are better suited for this type of task.

Generally yes, yet we have seen this error in the wild somehow as well. The question is how have we messed up, AKA, what is the defect, AKA bug?

> we changed something about the store or something and therefore the "wrong" thing got saved in an older version and the newer version was no longer able to read that correctly or the other way round.

That would be a bona fide bug and would impact users. Not accounting for different versions causes plenty of issues, we should be making sure we do not have such problems. That scenario is something to avoid and/or handle gracefully (one way would be to keep both encoding's around, decode with old if there is a failure and then re-encode).

RoiEXLab:
I think we can agree on the fact that there is a possibility that we need to reset the value at some point. (Data corruption, programmer errors etc.).

So the scope of this PR is simply to "save" a user from an error message that can't be avoided that easily.

So the question is to merge or not to merge

ron-murhammer:
I'm fine moving forward with the reset and INFO level message.

PR #6001, opened 2020-03-04
Game runner extract main frame

DanVanAtta:
Series of refactors to help start fixing the entanglement of 'GameSelectorModel' with many other game selection components. The overall goal was to fix game loader error handling triggered from 'GameSelectorModel', to get there it is a long top-down road and is a first step.

In this update we start at GameRunner and extract and remove dependencies on MainFrame. This will let us start working out the interplay between 'GameSetupPanel' and 'GameSelectorModel'

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done
- checked game launch, various startup screens like PBF, background task runner still worked, started a game and checked quit still works.


game-headed-javafx/src/main/java/org/triplea/game/client/ui/javafx/TripleA.java
```diff
@@ -40,10 +37,6 @@ public void start(final Stage stage) {
     navigationPane.switchScreen(FxmlManager.MAIN_MENU_PANE);
 
     setupStage(stage, scene, loadedNode.getController());
-    // Don't invoke Swing if headless (for example in tests)
-    if (!GraphicsEnvironment.isHeadless()) {
-      SwingUtilities.invokeLater(GameRunner::newMainFrame);
-    }
```
DanVanAtta: @RoiEXLab this was removed as it does not seem to belong in the JavaFx componentry. This invocation seemed like an intermediary step to get java fx working. I'm hoping we can circle back and get this fixed correctly and not have swing dependencies in JavaFx, all in part of fixing up the game startup models.

I tried to test if this had an impact, and was getting an error launching this class:

> Error: JavaFX runtime components are missing, and are required to run this application

RoiEXLab: Did you launch this class via the new RunConfiguration?
There seems to be an issue with JavaFX where it can't find the components when the main class extends application and is not a module

DanVanAtta: Ah, thanks for pointing out the run configuration. :+1: 
The JavaFx variant launches both before and after this update. Admittedly that was not a very thorough regression check.

RoiEXLab: Try launching a game once.
The reason I introduced those 3 lines of code is because originally some swing code I invoke at some point actually expects this mainframe to exist, otherwise an NPE will be thrown.
The reason for the headless check was because we initially had some javafx integration tests that basically clicked through the screens, but this framework is only able to create dummy javafx components, so swing just crashed the test.

DanVanAtta:
I think we probably should fix this more deeply. Otherwise we get into
something of a catch 22 where the javafx call prevents cleaning code that
would make removing the javafx call possible.

RoiEXLab:
I agree, but the current state of the UI makes this very hard, almost everything is static for launching games and assumes that a "main frame" or at least something similar exists :/

DanVanAtta:
Ok.. the catch-22 is kinda of a problem. The point of javafx was in part we
were going to fix things, not just push problems along. This code for the
javafx ui to work is not long term, not the correct thing to do, and makes
the right thing even harder to fix. That the ui is in a bad state is a
major problem we're trying to solve, not codify.

RoiEXLab:
To be fair the code only needed to be there to be able to invoke swing components to fill the implementation gaps, once the swing code is replaced with JavaFX all of the "hacks" can be dropped

DanVanAtta:
Said another way from a different perspective, we shouldn't have
dependencies like this on swing, it needs to be fixed properly. That
probably starts with making the launch code better. I think javafx is too
eager here, getting swing out of the way is necessary or we will tie
ourselves in knots with hacks.

DanVanAtta:
That is part of my point, such dependencies prevent us from replacing the
swing code. Replacing just the ui is not going to end well if the models
are not fixed first

DanVanAtta:
Roi, I understand the point of this code now, ty. Though, it clearly a bad
deoendency as we should be able to elevate the swing ui to a top level sub
project that javafx cannot see. We can't spend time arguing, there is more
here to review, and you stated as well this is to fix an implementation gap
- it is temporary and is getting in the way of updates that would in turn
help fix this.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/EmailSenderEditor.java
```diff
@@ -181,8 +180,8 @@ private void testEmail() {
                 final int finalMessageType = messageType;
                 SwingUtilities.invokeLater(
                     () ->
-                        GameRunner.showMessageDialog(
-                            finalMessage, GameRunner.Title.of("Email Test"), finalMessageType));
+                        JOptionPane.showMessageDialog(
+                            null, finalMessage, "Email Test", finalMessageType));
```
RoiEXLab: One difference this is probably going to cause is that the dialogs now no longer share the icon with the main frame, this problem can be found in different places too though.
Also it might be no longer on top of the mainframe

game-core/src/main/java/games/strategy/engine/framework/ui/MainFrame.java
```diff
@@ -0,0 +1,65 @@
...
+  private MainFrame(
+      final SetupPanelModel setupPanelModel, final GameSelectorModel gameSelectorModel) {
+    mainFrame =
+        JFrameBuilder.builder()
+            .title("TripleA")
+            .windowClosedAction(GameRunner::exitGameIfFinished)
+            .build();
+    BackgroundTaskRunner.setMainFrame(mainFrame);
+
+    LookAndFeelSwingFrameListener.register(mainFrame);
```
RoiEXLab: I think those static actions should be moved up one call layer:
I'f this constructor is ever called tice this could result in weird problems

game-core/src/main/java/games/strategy/engine/framework/ui/background/BackgroundTaskRunner.java
```diff
@@ -11,17 +12,17 @@
...
+  private static JFrame mainFrame;
 
-  public BackgroundTaskRunner(final JFrame frame) {
-    checkNotNull(frame);
-
-    this.frame = frame;
+  public static void setMainFrame(final JFrame mainFrame) {
+    Preconditions.checkState(BackgroundTaskRunner.mainFrame == null);
+    BackgroundTaskRunner.mainFrame = mainFrame;
   }
```
RoiEXLab: Hmm I wonder if this really needs to be changed, would anything speak against the callers explicitly passing some sort of frame (not necessarily the main frame) and adding a deprecated getter for the singleton frame if needed?

game-core/src/main/java/games/strategy/engine/framework/startup/mc/SetupPanelModel.java
```diff
@@ -35,7 +35,7 @@
   protected SetupPanel panel = null;
 
   @Setter private Consumer<SetupPanel> panelChangeListener;
-  @Nonnull private final JFrame ui;
+  @Setter private JFrame ui;
```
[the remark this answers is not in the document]
DanVanAtta: Generally agree, though I think we can skip that step. SetupPanelModel should not even have a reference to the UI other than an interface for the 'setupPanel', and certainly should not have a reference to the main frame. The dialogs that are displayed from this model class that need a 'parent' reference all do not belong.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/EmailSenderEditor.java
DanVanAtta: The icon is significant, though the number of dialogs impacts is limited and they are not used that often. For now decided to favor dis-entangling things.

game-core/src/main/java/games/strategy/engine/framework/ui/MainFrame.java
DanVanAtta: The constructor is private, the factory method checks that the static instance is not yet set. One level higher is 'GameRunner', part of the goal is to remove anything UI or swing related out of GameRunner.

game-core/src/main/java/games/strategy/engine/framework/ui/background/BackgroundTaskRunner.java
DanVanAtta: Previously we always passed the main frame, so in part I'm not seeking to change too much at any one time. Passing other frames would be a really different type of change. The getter I'd like to avoid so that we have the 'mainFrame' pushing its reference and using 'setters' rather than other elements using 'getter' to get the reference (essentially an application of IOC, or the "we'll call you, don't call us" principle)

DanVanAtta:
Thanks for the turn-around on the review here @RoiEXLab  and feedback.

game-core/src/main/java/games/strategy/engine/framework/ui/MainFrame.java
RoiEXLab: By "one level higher" I was actually reffering to the static method inside this class rather than the constructor

Issue #5957, opened 2020-02-19
IOException - GameSelectorModel.load: games.strategy.util.Version


tripleabuilderbot:
## User Description
When i downloaded the new version, it will not let me play my saved games. I think there is an easy way to play them but am at a loss for the moment. Any assistance would be welcome.

## Log Message
Error loading game file: C:\Users\johnw\triplea\savedGames\fireandice free cities2.tsvg

## TripleA Version
2.0.18070

## Java Version
11.0.4

## Operating System
Windows 10

## Exception 
java.io.IOException: games.strategy.util.Version

## Stack Trace
games.strategy.engine.framework.GameDataManager.loadGame(GameDataManager.java:79)
games.strategy.engine.framework.GameDataManager.loadGame(GameDataManager.java:45)
games.strategy.engine.framework.startup.mc.GameSelectorModel.load(GameSelectorModel.java:83)
games.strategy.engine.framework.startup.ui.panels.main.game.selector.GameSelectorPanel.lambda$selectSavedGameFile$4(GameSelectorPanel.java:323)
games.strategy.engine.framework.ui.background.BackgroundTaskRunner$1.doInBackground(BackgroundTaskRunner.java:116)
java.desktop/javax.swing.SwingWorker$1.call(SwingWorker.java:304)
java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
java.desktop/javax.swing.SwingWorker.run(SwingWorker.java:343)
java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
java.base/java.lang.Thread.run(Thread.java:834)

RoiEXLab:
@DanVanAtta Ok, I don't think there's much we can do about that.

The user tried to load a 1.9 Savegame in 2.0.
Just how are we supposed to notify the user about that?

DanVanAtta:
Looks like a try/catch for the right error condition and then notify the user with a message indicating an incompatible game would be appropriate. This is a good error report to have IMO.

[...]

PR #6002, opened 2020-03-04
Incompatible or "File Not Found" Save Game Error Messaging Improvements

DanVanAtta:
## Overview

Improves error messaging when loading incompatible save games. Notably will give a specific error message when loading a 1.9 save game, addresses: https://github.com/triplea-game/triplea/issues/5957

The error handling was pretty involved and a bit confused. First step was to make it so that all error messages for bad version did not come up as a 'save game not found error', so first that was fixed. Following that, the old Version.java was restored so we could deserialize it and give a good error message that we are incompatible with 1.9 game saves. Of note, a 'TaskRunner' API was created to help the error handling and improves on the API provided by 'BackgroundTaskRunner'.

## Commits

commit 2bea667053e9d4e7940aa494668a23bfa541152c

    Handle save-game file not found case

commit 1af53b5d235b926a715281fc1eb070bab2859dca

    Show an error message when failing to load a save game

commit 40af5489657b7f85878b836e199894f32f7bf55b

    Resurrect Version.java from latest release (1.9.0.0.13066)

commit 4746b3017f4b95a902c13836bb8e2ce7bd7716d1

    Update old Version.java, deprecate it, suppress checkstyles, apply formatting

commit 4d124c258b6abf5dd26aafa0deb0892fe9bab281

    Give specific error message and download link for incompatible 1.9 save games

commit e42972525dda2fa0571ee9a32799adf913ab91f4

    Improve javadoc on new utility class: TaskRunner

commit 54883927248e671fcbdb217e3f286c3eba8ee429

    Remove unused "@Log"

commit a2e0d215132ab9c7d8db41ff087fe0972b0f490c

    Allow unmatched-tests custom check to be suppressed for specific class files via "SuppressWarnings" annotation

commit 3db89602f

    Refactor TaskRunner, shorten "run" method

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  #5957 
[] Other:   

## Testing

[x] Manual testing done

- Loaded various saves games:
  - an incompatible save game
  - a save game from an old version
  - a save game from a new version,
  - a compatiible save game (happy case)

- Verified error message links were clickable

- I added an infinite loop to 'TaskRunner' and verified that the wait dialog can be closed.

- Tested file not found case:
  - opened headed game, opened save game dialog, selected a file and then changed the file name to load to something that does not exist
  - started a local lobby with bot, joined the bot and selected a save game to load and mangled the file name to something that does not exist. I verified that latest master is a no-op under this condition.

# Screens Shots

(Note, seemingly on latest master @ 289b561d27cd5987263427d519a8acb7f2e95f05, all save game load errors come back as a "save game file not found" and the RC exception is logged in the background. I think this could potentially be an error dialog race condition where it depends on which logger statement triggers the error message dialog, or perhaps something more fundamental has changed since #5957 was reported. Regardless, this update improves the error handling to give a specific error message and the RC exception is printed in the error dialog).

### Before

Screenshot from 2020-03-04 00-31-06
Screenshot from 2020-03-04 00-31-15

### After
Screenshot from 2020-03-03 23-56-27

### Before
Screenshot from 2020-03-04 00-31-35
Screenshot from 2020-03-04 00-31-41

### After

Screenshot from 2020-03-03 23-56-55

### Before
Screenshot from 2020-03-04 00-32-15
Screenshot from 2020-03-04 00-32-22

### After

Screenshot from 2020-03-03 23-57-13


game-core/src/main/java/games/strategy/engine/framework/ui/background/TaskRunner.java
```diff
@@ -0,0 +1,96 @@
...
+  private SwingWorker<Void, Void> buildBackgroundJobRunnerWithWaitDialog(
+      final ThrowingRunnable<E> backgroundAction,
+      final AtomicReference<Throwable> exceptionRef,
+      final WaitDialog waitDialog) {
+    return new SwingWorker<>() {
+      @Override
+      protected Void doInBackground() {
+        try {
+          backgroundAction.run();
+        } catch (final Throwable e) {
```
RoiEXLab: It's a shame generic exceptions can't be caught here.
However I wouldn't catch throwable here, just the regular `Exception`, it can't even be specified as generic type anyways

game-core/src/main/java/org/triplea/game/server/HeadlessLaunchAction.java
```diff
@@ -31,8 +31,8 @@ public void handleGameInterruption(
       // then crashing out, then launching, etc.
       serverModel.setAllPlayersToNullNodes();
       final File f1 = getAutoSaveFileUtils().getHeadlessAutoSaveFile();
-      if (!f1.exists() || !gameSelectorModel.load(f1)) {
-        gameSelectorModel.resetGameDataToNull();
+      if (!f1.exists()) {
+        gameSelectorModel.load(f1);
```
RoiEXLab: I believe a small bug sneaked in here:
Load should only get invoked when the file does actually exist, not the other way round.
Previously in the if statement this was the case because the second or case was only ever executed if `!f1.exists()` was false i.e. `f1.exists()` was true.

game-core/src/main/java/games/strategy/engine/framework/GameRunner.java
```diff
@@ -153,7 +156,12 @@ private static void loadGame() {
     gameSelectorModel.loadDefaultGameSameThread();
     final String fileName = System.getProperty(TRIPLEA_GAME, "");
     if (!fileName.isEmpty() && new File(fileName).exists()) {
-      gameSelectorModel.load(new File(fileName));
+      try {
+        gameSelectorModel.load(new File(fileName));
+      } catch (final Exception e) {
+        log.log(
+            Level.SEVERE, "Error loading game file: " + new File(fileName).getAbsolutePath(), e);
```
RoiEXLab: `new File(fileName)` is called three times here, I'd appreciate if it was put into a single variable and used 3 times, even if that adds another if block

RoiEXLab:
@DanVanAtta I most of the changes in this PR are definitely improvements, however I really dislike the idea of keeping a second version class just to get a more detailed error message.
I feel like it shouldn't really matter why the loading of the savegame failed, it's almost always just some kind of compatibility issue, so we can tell them they just used the wrong version for this savegame and they should try using the version the savegame was created with.
I think this makes the important stuff clear enough while not going into too much detail.
A similar problem will occur whenever we change a class intentionally and the next patch fails to desiralize that version because suddenly a single field has changed. In this case users should get an explanation as well, but in this case we can't provide any extra detail which is fine too IMO.

game-core/src/main/java/org/triplea/game/server/HeadlessLaunchAction.java
DanVanAtta: You're right, that is a good catch.
Fixed in: 7846a1bfa

game-core/src/main/java/games/strategy/engine/framework/GameDataManager.java
```diff
@@ -59,16 +61,35 @@ public static GameData loadGame(final InputStream is) throws IOException {
...
+                    + "please visit: <a href=%s>%s</a>",
+                ClientContext.engineVersion(),
+                ((games.strategy.util.Version) version).getExactVersion(),
+                UrlConstants.OLD_DOWNLOADS_WEBSITE,
+                UrlConstants.OLD_DOWNLOADS_WEBSITE));
+
+      } else if (!(version instanceof Version)) {
+        throw new IOException(
+            "Incompatible engine version with save game, "
+                + "unable to determine version of the save game");
```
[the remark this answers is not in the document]
DanVanAtta: In short, the more detail we can get, the better.
Considering the reported issue: https://github.com/triplea-game/triplea/issues/5957, @ron-murhammer said: "Yeah we need a good error notification here as this will happen fairly often."

Without a good error message, I'm pretty positive a generic error message that does not really tell a user why the save game cannot be loaded would get reported as a bug. We do not have really any other way to tell a user to accurately that they need a 1.9 version. IMO, having the legacy class is a cost of the serialization technology used for save games. We introduced a feature regression when we removed the 'old jar' capability, keeping a legacy class around for a release is not a material cost to us.

Generally agree that we probably should encode the version as a string to keep the compatibility. Though, that is not what we have or had, it might be better to move directly to the project where we fix save game serialization and skip the intermediary step.

game-core/src/main/java/games/strategy/engine/framework/GameDataManager.java
```diff
@@ -59,16 +61,35 @@ public static GameData loadGame(final InputStream is) throws IOException {
...
+      } else if (!ClientContext.engineVersion().isCompatibleWithEngineVersion((Version) version)) {
+        throw new IOException(
+            String.format(
+                "Incompatible engine versions. We are: %s<br/>"
+                    + "Trying to load game created with: %s<br/>"
+                    + "To download the latest version of TripleA,<br/>"
+                    + "please visit: <a href=%s>%s</a>",
```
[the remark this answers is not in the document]
DanVanAtta: The swing HTML renderer is not exactly HTML5 compliant, regardless, those are not bad cleanups per-say, updated:
- d9b6893fd2030ea644e518d47a36a615e007516a
- af46b9f159eb96a6e77fedf198ecaa89790afcf30

game-core/src/main/java/games/strategy/engine/framework/startup/mc/GameSelectorModel.java
```diff
@@ -66,31 +67,25 @@ public void load(final GameChooserEntry entry) {
...
    *
-   * @return True if file parsing was successful and an internal {@code GameData} was set. Otherwise
-   *     returns false and internal {@code GameData} is null.
+   * @throws Exception If file parsing is successful and an internal {@code GameData} was set.
    */
-  public boolean load(final File file) {
-    if (!file.isFile()) {
-      return false;
-    }
+  public void load(final File file) throws Exception {
```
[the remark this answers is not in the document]
DanVanAtta: There is also:
```
  public static GameData parse(final String mapName, final InputStream stream)
      throws GameParseException, EngineVersionException {
```
Turns out if we break it out further, there would eventually be more of them.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameFileSelector.java
```diff
@@ -32,15 +36,26 @@ private GameFileSelector() {}
       fileDialog.setVisible(true);
       final String fileName = fileDialog.getFile();
       final String dirName = fileDialog.getDirectory();
-      return Optional.ofNullable(fileName).map(name -> new File(dirName, fileName));
+      return Optional.ofNullable(fileName)
+          .map(name -> new File(dirName, fileName))
+          .map(this::mapFileResult);
```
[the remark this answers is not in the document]
DanVanAtta: I think this direction, but not for this PR. The updates here already tested, would need to be re-tested again once this update is made. We can just as easily make this it's own follow-up and do its own testing for it. From another perspective, the update here is to also improve the existing handling, not necessarily change how the file is gathered.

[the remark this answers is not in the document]
DanVanAtta: I'm not sure if the 'findAny' would be very clear as well to future maintainers, having this logic be close to where we limit to one selection would probably be the thing to do.

game-core/src/main/java/games/strategy/engine/framework/ui/background/TaskRunner.java
DanVanAtta: This class should look very similar to `BackGroundTaskRunner`, it was modeled after it and really started off as a spawn of it. I remember debating this one and see why `BackGroundTaskRunner` had a `Throwable`, it is not needed here. I agree better to catch Exception, updated: ef1ecb994

game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorPanel.java
```diff
@@ -312,30 +318,41 @@ public void update(final Observable o, final Object arg) {
...
+                                DialogWithLinksParams.builder()
+                                    .title("Failed To Load Save Game")
+                                    .dialogType(DialogWithLinksTypes.ERROR)
+                                    .dialogText(
+                                        String.format(
+                                            "<html>Error: %s<br/><br/>"
+                                                + "If this is not expected, please "
+                                                + "file a <a href=%s>bug report</a><br/>"
+                                                + "and attach the error message above and the "
+                                                + "save game you are trying to load.",
```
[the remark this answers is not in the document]
DanVanAtta: Indeed, though, swing is not really a compliant HTML parser and we're not really rendering HTML pages. It would be inappropriate to have a body or head tag. The prefix of `<html>` is effectively a hack to tell swing to do html rendering. IIRC, swing supports only a subset of tags.

DanVanAtta:
@RoiEXLab  updated, thanks for the catches/recommendations.

I'm strongly opinionated on the 'old' version class, per:https://github.com/triplea-game/triplea/pull/6002#discussion_r387930633
- gives the user a clear error message and how to resolve, that is critical
- the old version class almost costs us nothing, just ignore it and focus on what needs to be done, we'll remove it eventually.

DanVanAtta:
> I feel like it shouldn't really matter why the loading of the savegame failed

To expound, I do disagree with that. IT does matter as a user would want to know how/why to fix it and not that TripleA is just broken. Best case we get a bug report from it. I'll emphasize that we have to keep in mind that most computer users are not anywhere close to power users like us, and that we are dealing with a thick client and the audience of error message is effectively your grandparents. All error messages need to say what went wrong, what's the implication to the user, and what the user need/should do to fix it. It is part of thick client development, if this were a web server, then I'd agree much more readily.

DanVanAtta:
> A similar problem will occur whenever we change a class intentionally and the next patch fails to desiralize that version because suddenly a single field has changed

If that happens, then we failed to update version numbers appropriately. Getting a bug report for that is just as well, as we would deserve it (it would be a configuration bug).

DanVanAtta:
@RoiEXLab ball is in your court here.

Re-reading your last comment, one question:

>  so we can tell them they just used the wrong version for this savegame and they should try using the version the savegame was created with.

How do you mean by "we can tell them?" Are you implying that the error messaging would tell a user that, or more if they open a support ticket?

RoiEXLab:
@DanVanAtta I'm just saying that I believe it would be sufficient to tell the user that whenever deserialization fails we could just tell them "you're savegame was saved with an older, incompatible version" without exactly specifying which one it is.

I just had an idea though:
`ObjectInputStream#readObject` throws a lot of different exception types.
`ClassNotFoundException` should be the one thrown without the legacy Version class there. If we ever catch this type of exception we could just notify the user their savegame was saved with a version <=1.9 
Which is the most likely reason.
If we want to be sure it actually is an older savegame, we could also check the exception message if it contains the `games.strategy.util.Version` class string or something.

All other kind of serialization errors when deserializing the first object from the stream is most likely a programming error and worth reporting.

In my opinion this is a pretty solid approach without keeping the class around.

I'd love to hear what you think about this.
In any case I'd ask you to remove the unused methods of the legacy version class, there's no need to implement Comparable and override the equals methods of they're not used anyways

DanVanAtta:
Classnotfound can be thrown later and for different reasons. Catching that
could mask over an actual problem.
Version meanwhile is fully accurate. I see no reason to pretty up a
deprecated class we will delete. It was not resurrected so we can maintain
it.

So you have any concern for having it? Is there a concrete problem you see
having it? I suggest we need to be pragmatic, this project is not here to
exercise code for the sake of code (right?)

RoiEXLab:
> Classnotfound can be thrown later and for different reasons. Catching that could mask over an actual problem.

Not if the try block is only around that first call of readObject.

You convinced me that it's ok for now to keep this class, but if a simple try catch would do the same thing then I think that's a better option.

DanVanAtta:
Sorry for difficult path to agreement, but we reached a concensus : )

The try-catch block around the first instance of checks I don't think will be quite enough. Just notably in the case where we break future compatibility unintentionally, notably changes to `Version.java` itself. As I can see it, having the old `Version.java` is just the most reliable way to get exact error detection.

@RoiEXLab everything else has been addressed? I agree having a legacy `Version.java` around is lame, I'm chalking it up to technical debt costs of the design of the system, it's not great, but it is what it is. With that said, is there anything more that should be done here?

game-core/src/main/java/games/strategy/engine/framework/GameRunner.java
RoiEXLab: This one

game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameFileSelector.java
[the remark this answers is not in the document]
RoiEXLab: This one in a future PR

game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameFileSelector.java
```diff
@@ -32,15 +36,26 @@ private GameFileSelector() {}
...
+      return Optional.of(fileChooser.getSelectedFile()).map(this::mapFileResult);
     }
     return Optional.empty();
   }
+
+  private File mapFileResult(final File file) {
+    if (file.exists()) {
+      return file;
+    } else {
+      fileDoesNotExistAction.accept(file);
```
RoiEXLab: It's a bit hard to wrap my head around this, this is a super weird side-effect, but ok.
Perhaps worth annotating this method with `@Nullable`

game-core/src/main/java/games/strategy/engine/framework/ui/background/TaskRunner.java
```diff
@@ -0,0 +1,96 @@
...
+        exceptionHandler.accept(exception);
+      }
+    } catch (final ClassCastException e) {
+      throw new AssertionError("Unexpected exception thrown", e);
+    }
+  }
+
+  private SwingWorker<Void, Void> buildBackgroundJobRunnerWithWaitDialog(
+      final ThrowingRunnable<E> backgroundAction,
+      final AtomicReference<Throwable> exceptionRef,
```
RoiEXLab: The `AtomicReference` is still Throwable instead of just exception

game-core/src/main/java/games/strategy/engine/framework/ui/background/TaskRunner.java
```diff
@@ -0,0 +1,96 @@
...
+    waitDialog.setVisible(true);
+
+    try {
+      @SuppressWarnings("unchecked")
+      final @Nullable E exception = (E) exceptionRef.get();
+      if (exception != null) {
+        exceptionHandler.accept(exception);
+      }
+    } catch (final ClassCastException e) {
+      throw new AssertionError("Unexpected exception thrown", e);
```
RoiEXLab: I'm 99% sure a `ClassCastException` can only ever be thrown inside the exception handler here, so in case we catch an unchecked Exception by accident causing the exceptionHandler to trip up, this exception isn't as impossible as it seems.

Thinking about it we might want to consider explicitly passing a class object here to correctly filter out the "unwanted" exception types 🤔

game-core/src/main/java/games/strategy/util/Version.java
```diff
@@ -0,0 +1,192 @@
...
+    return compareTo(other) < 0;
+  }
+
+  /**
+   * Returns a new version with the major, minor, and point versions from this instance and the
+   * specified micro version.
+   */
+  public Version withMicro(final int micro) {
+    return new Version(m_major, m_minor, m_point, micro);
+  }
```
RoiEXLab: This whole part as well as the constructors can be removed IMO, just to be sure no one actually tries to do something about this. 🤷‍♂ 

Just because it's deprecated doesn't mean people won't actually try to use it

RoiEXLab:
@DanVanAtta The first to comments are replies to the original comment.
Without this context it looks pretty confusing

game-core/src/main/java/games/strategy/engine/framework/ui/background/TaskRunner.java
DanVanAtta: > Thinking about it we might want to consider explicitly passing a class object here to correctly filter out the "unwanted" exception types thinking

This was an explicit decision. Background runner I found hard to use in the exception handling case and also turns out to be brittle (if the background task opens a swing window, because you can't close the modal wait dialog, you get a deadlock. The background task then only needs to log/throw an exception that triggers an error report dialog to crash the system). The exception API requiring an exception type was an unwelcome parameter to see, it's redundant to the generic type already. The trade-off for really specific error handling to API usage is not favorable here (otherwise background task runner could have been made fine to use).

game-core/src/main/java/games/strategy/engine/framework/ui/background/TaskRunner.java
DanVanAtta: Updated: 8e794f440

game-core/src/main/java/games/strategy/util/Version.java
DanVanAtta: Not willing to modify this.
1. Would like diffs to be minimal
2. Waste of time reviewing changes here and equally responding to any potential feedback, and I would be lost for words if we introduced any further problems and then had to spend yet more time on *that*. We need to value our time highly, and keep that in mind when considering our review bar and how much stuff we need to get done. Failures is an option, the community is shrinking, there are new alternatives for players being made, we need to get stuff done.

> Just because it's deprecated doesn't mean people won't actually try to use it

By that reasoning, folks might try to append to it as well. Hopefully the big strike-through makes it clear this is deprecated.

DanVanAtta:
@RoiEXLab updated. Getting a list of must-change items after a first review would have been helpful so I can know specifically what you thought was important vs less important possible improvements that you were just happening to point out.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameFileSelector.java
RoiEXLab: Well from just looking at the other method I assumed that `mapFileResult` was a "pure" function with no side-effects, so I was quite suprised to see the error handler there.
Maybe not "weird" but definitely unexpected.

game-core/src/main/java/games/strategy/engine/framework/GameRunner.java
RoiEXLab: The only reason I'm pointing this out is because this PR introduces another call to `new File(fileName)`.
So I was thinking that now that there's 3 of them adding a variable and use that3 times would finally be worth it, but nothing too crazy here. I can also do that in a follow-up PR. It's nothing special

DanVanAtta: It is your prerogative. IMO it could be more effective prioritization to focus on 2.0 or deeper refactoring that would remove game loading from `GameRunner`

RoiEXLab: Removing this bit of the code would be a good step into the right direction, but a hard one to achieve  unfortunately.

The "join headless host" code depends on it, so it can pass parameters to the newly spawned process 😅

game-core/src/main/java/games/strategy/engine/framework/ui/background/TaskRunner.java
RoiEXLab: My only problem with this is that we only use `AssertionError` for impossible conditions (like when UTF-8 is hardcoded as charset, but an exception is thrown, or when we do a switch and we checked all possible cases so the default case can never get reached). But this particular condition is very possible in practice, so I'd prefer using `IllegalStateException`

PR #6007, opened 2020-03-05
Expire games

DanVanAtta:
## Overview

Noticed that the mechanism to remove closed games no longer works ever since we went to websocket for game listing. This update fixes that a bit by adding a game listing removal listener that will notify clients if a game times out and does not issue a keep alive.

Still to do in future updates will be to send explicit game listing notices from game hosts that are shutting down back to the lobby.

## Commits

commit 25f3750245cac32841dabda58e1646e67195bf97

    Notify game client when lobby listings are expired from cache.
    
    Games are expired from cache if they are not renewed via keep-alive.
    Currently nothing notifies clients when such a game is removed and
    the client will think the game is still available. We previously
    would have polled for all available games and the client would
    determine when games have been removed. Then after switching to
    websockets and an expiring cache, there was nothing done to replace
    this notification mechanism.
    
    Getting a listener to be notified on cache expiry is a bit tricky and
    requires a cleanup job to be run. To extract these details, we
    create a 'Cache' interface and consolidate CacheUtils to this interface.
    Then we create an implementation that will invoke a removed item listener.
    This is done by creating a scheduled thread that periodically calls
    'cleanup' on the cache.

commit 1d2c34c25db927ec4a5ed45165a4cc33f3418923

    Add a number of contract tests for cache

commit c62312b1c8dfaa9524dedbd49bfffd5a0991aeb6

    Rename "stopTimer" method to "close"

commit 1777ff9ebf5a697b3475b9877c5684d7ce712569

    Remove mock cache from test

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

[x] Manual testing done


java-extras/build.gradle
```diff
@@ -1,5 +1,8 @@
 description = 'TripleA library for low-level helper APIs, ie: syntactic sugar'
 
 dependencies {
+    // https://mvnrepository.com/artifact/com.github.ben-manes.caffeine/caffeine
+    compile group: 'com.github.ben-manes.caffeine', name: 'caffeine', version: '2.8.1'
```
RoiEXLab: We usually use the single string notation where the different parts are separated with a colon + we have all the version numbers in the main build.gradle file

java-extras/src/main/java/org/triplea/java/cache/ExpiringAfterWriteCache.java
```diff
@@ -0,0 +1,90 @@
...
+      final long duration,
+      final TimeUnit timeUnit,
+      final Consumer<CacheEntry<IdT, ValueT>> removalListener) {
+    cache =
+        Caffeine.newBuilder()
+            .expireAfterWrite(duration, timeUnit)
+            .removalListener(
+                (key, value, cause) -> {
+                  if (cause == RemovalCause.EXPIRED || cause == RemovalCause.EXPLICIT) {
+                    removalListener.accept(new CacheEntry<>((IdT) key, (ValueT) value));
```
RoiEXLab: What about using BiConsumer instead?

java-extras/src/main/java/org/triplea/java/cache/ExpiringAfterWriteCache.java
```diff
@@ -0,0 +1,90 @@
...
+  @Override
+  public void put(final IdT id, final ValueT value) {
+    cache.put(id, value);
+  }
+
+  @Override
+  public Optional<ValueT> invalidate(final IdT id) {
+    final Optional<ValueT> value = get(id);
+    cache.invalidate(id);
+    value.ifPresent(valueT -> removalListener.accept(new CacheEntry<>(id, valueT)));
```
RoiEXLab: Maybe I'm missing something, but why is removalListener called explicitly here?
Shouldn't invalidate trigger the listener by itself already?

java-extras/src/main/java/org/triplea/java/cache/ExpiringAfterWriteCache.java
RoiEXLab: Just had a look at the caffeine documentation because I thought the explicit casting was weird.
You can avoid (I'm 99% sure) the unchecked warning by using
```java
.<IdT, ValueT>removalListener((key, value, cause) -> {});
```
Instead

java-extras/src/main/java/org/triplea/java/cache/ExpiringAfterWriteCache.java
```diff
@@ -0,0 +1,90 @@
...
+                  if (cause == RemovalCause.EXPIRED || cause == RemovalCause.EXPLICIT) {
+                    removalListener.accept(new CacheEntry<>((IdT) key, (ValueT) value));
+                  }
+                })
+            .build();
+
+    this.removalListener = removalListener;
+
+    cleanupTimer =
+        Timers.fixedRateTimer("cache-cleanup-" + Math.random())
```
RoiEXLab: what's the reason this task has a random id attached to it?

java-extras/src/main/java/org/triplea/java/cache/ExpiringAfterWriteCache.java
ben-manes: You could declare the types on the lambda parameters instead,
```java
.removalListener((IdT key, ValueT value, RemovalCause cause) -> {});
```

java-extras/src/main/java/org/triplea/java/cache/ExpiringAfterWriteCache.java
ben-manes: yes. Even if not, you can use `asMap().remove(id)` to remove the entry and obtain the old value atomically.

java-extras/src/main/java/org/triplea/java/cache/ExpiringAfterWriteCache.java
ben-manes: You can use `Caffeine.scheduler` to do this for you, too! If you are on JDK9+ then it can leverage the JVM-wide thread (internal to CompletableFuture), or else you can either supply a `ScheduledExecutorService` or your own implementation.

```java
Caffeine.newBuilder()
    .expireAfterWrite(duration, timeUnit)
    .scheduler(Scheduler.systemScheduler())
...
```

java-extras/src/main/java/org/triplea/java/cache/TtlCache.java
```diff
@@ -0,0 +1,49 @@
...
+  /**
+   * Places a new item in the cache if one existed with the same ID and replaces the existing value.
+   * The replaced item is returned, otherwise returns an empty.
+   */
+  default Optional<ValueT> replace(IdT id, ValueT newValue) {
+    final Optional<ValueT> value = get(id);
+    if (value.isPresent()) {
+      put(id, newValue);
+    }
+    return value;
```
ben-manes: Atomically,
```java
return cache.asMap().replace(id, newValue)
```

DanVanAtta:
> Why did you introduce Caffeine? It looks like a solid library, but I would've thought that guava provides enough functionality

@RoiEXLab, getting Guava to call the remove listener was a struggle. After spending some quality time trying to get that to work, I basically gave up on Guava and looked for other APIs/caches. The type inference with Guava was seemingly even more difficult. I also recall there is another tool, can't quite remember though exactly that was moving to Caffeine for cache config, I want to say Dropwizard 2.0.

java-extras/src/main/java/org/triplea/java/cache/TtlCache.java
DanVanAtta: Seems simpler, atomic is better - thank you for the suggestion! Updated: 08d724ea0

java-extras/src/main/java/org/triplea/java/cache/ExpiringAfterWriteCache.java
RoiEXLab: @DanVanAtta If it's only the tests, then we could make the test wait until `ForkJoinPool.commonPool()` processed all events (I had a look at the source code), or alternatively just pass a custom executor for the test cases which simply runs the specified `Runnable`

java-extras/src/main/java/org/triplea/java/cache/ExpiringAfterWriteCache.java
```diff
@@ -0,0 +1,90 @@
...
+  public Optional<ValueT> invalidate(final IdT id) {
+    final Optional<ValueT> value = get(id);
+    cache.invalidate(id);
+    value.ifPresent(valueT -> removalListener.accept(new CacheEntry<>(id, valueT)));
+    return value;
+  }
+
+  @Override
+  public Map<IdT, ValueT> asMap() {
+    return Map.copyOf(cache.asMap());
```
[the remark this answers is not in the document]
RoiEXLab: I think the way it's supposed to be done is by doing something like:
```java
Map<IdT, ValueT> map = CacheBuilder.<IdT, ValueT>newBuilder()
        .expireAfterWrite(1, TimeUnit.MILLISECONDS)
        .build()
        .asMap();
```
instead, but ok

java-extras/src/test/java/org/triplea/java/cache/ExpiringAfterWriteCacheTest.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    @Test
+    void putCanBeUsedToReplace() {
+      realCache.put("id", 0);
+
+      realCache.put("id", 1);
+
+      assertThat(realCache.get("id"), isPresentAndIs(1));
+      verify(cacheRemovalListener, never()).accept(any());
```
RoiEXLab: Thinking about my suggestion that passes a custom executor for the tests: This particular tests would gain some extra safety, because in case the removalListener is executed asynchronously, the tests might check too early missing this critical error.

java-extras/src/main/java/org/triplea/java/cache/ExpiringAfterWriteCache.java
DanVanAtta: Reasonable suggestions, though I think sticking with the explicit notification is okay:
- in our context we handle duplicate calls already (at least we should be!). Having a contract that allows for duplicate callbacks seems okay in general as well (and if not then there is room for other implementations to change that).
- the test is already a bit slow at 700ms. Doing a join could mean we wait for the scheduled cleanup job which could add a second
- a join in the test would turn the test from mostly if not entirely a black box test into a white box test. Tests that do not rely on internal implementations tend to be easier to understand and maintain

A custom executor would also fall under the category of test scaffolding, it's better minimized. At the end of the day we also get benefit from an immediate callback, which does make the UI seem more responsive as games are removed nearly immediately and not after a small delay.

java-extras/src/main/java/org/triplea/java/cache/ExpiringAfterWriteCache.java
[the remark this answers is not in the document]
DanVanAtta: Even that variant does not work, you get a type cast exception where `ConcurrentMap<Object,Object>` is not compatible. :grin:  
Screenshot from 2020-03-06 11-25-00

java-extras/src/test/java/org/triplea/java/cache/ExpiringAfterWriteCacheTest.java
DanVanAtta: Agree the test could miss the timing. In practice this won't be a problem, there is a trade-off here between a slow test and more thorough test. Adding a second delay would be the thing to do, and/or making the cache scheduler really aggressive.

I double checked with a timeout and many repeated tests, as expected the test stays solid which is the most important thing. At this point I'm happy leaving the test to be more behavior documentation/spec than full-proof.

java-extras/src/main/java/org/triplea/java/cache/ExpiringAfterWriteCache.java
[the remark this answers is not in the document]
ben-manes: Yes, the builder inference in both APIs is meant for the cache type. You have two options,
```java
Map<IdT, ValueT> map = CacheBuilder.newBuilder().<IdT, ValueT>build().asMap();

Cache<IdT, ValueT> cache = CacheBuilder.newBuilder().build();
Map<IdT, ValueT> map = cache.asMap();
```

java-extras/src/test/java/org/triplea/java/cache/ExpiringAfterWriteCacheTest.java
ben-manes: Another approach than specifying a direct executor is to use a testing library like [Awaitility](https://github.com/awaitility/awaitility/wiki/Usage).

DanVanAtta: Really what we are testing is the if condition here: https://github.com/triplea-game/triplea/pull/6007/files#diff-bd3a571488048aa02a100236fc5d8a77R33

```
            .removalListener(
                (IdT key, ValueT value, RemovalCause cause) -> {
                  if (cause == RemovalCause.EXPIRED || cause == RemovalCause.EXPLICIT) {
                    removalListener.accept(key, value);
                  }
```

Mockito 'verify' has a timeout function, so it would be straightforward to add the delay. The issue is that we are testing the absence of a call, so we'd always hit the timeout making this a very slow test.

IMO the best way to test probably would be to inject the filter logic, test the filter, and then trust that the library code does call the filter on remove listener.

I'm leaning towards keeping this test as is:
- injecting modules is at danger of over-engineering
- the ROI of a slow but more thorough test is not favorable. We're already close to testing pure library code

DanVanAtta:
PS: thank you @ben-manes for jumping to offer some good suggestions and tips, was a pleasant surprise  :+1:

ben-manes:
just as pleasant as finding someone adding my library and having a positive experience with it 😄

PR #6016, opened 2020-03-07
Extract file variable

RoiEXLab:
Follow-up to #6002

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/GameRunner.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/GameRunner.java
@@ -79,7 +79,7 @@ public final class GameRunner {
   public static void showMainFrame() {
     MainFrame.show();
     ProAi.gameOverClearCache();
-    loadGame();
+    loadGameIfSpecified();
     openMapDownloadWindowIfDownloadScheduled();
 
     if (System.getProperty(TRIPLEA_SERVER, "false").equals("true")) {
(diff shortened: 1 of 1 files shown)


DanVanAtta:
@RoiEXLab any thoughts to take care of the code climate warning?

game-core/src/main/java/games/strategy/engine/framework/GameRunner.java
```diff
@@ -90,19 +91,20 @@ public static void showMainFrame() {
     }
   }
 
-  private static void loadGame() {
+  private static void loadGameIfSpecified() {
     checkState(!SwingUtilities.isEventDispatchThread());
     gameSelectorModel.loadDefaultGameSameThread();
-    final String fileName = System.getProperty(TRIPLEA_GAME, "");
-    if (!fileName.isEmpty() && new File(fileName).exists()) {
+    final File saveGameFile = new File(System.getProperty(TRIPLEA_GAME, ""));
```
RoiEXLab: Note: The `File` constructor doesn't throw an exception when using an empty path and seems to return false for `new File("").exists()`, so the empty check is redundant.

This way the cognitive complexity should go down.

RoiEXLab:
@DanVanAtta Updated, tell me what you think

game-core/src/main/java/games/strategy/engine/framework/GameRunner.java
DanVanAtta: That's a good simplification :+1:

PR #6017, opened 2020-03-07
Use File array instead of constructing the file object

RoiEXLab:
Follow-up to #6002

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

I ran this code inside jshell to verify it actually returns the same thing. It did.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameFileSelector.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameFileSelector.java
@@ -36,6 +36,9 @@ public final class GameFileSelector {
       fileDialog.setDirectory(ClientSetting.saveGamesFolderPath.getValueOrThrow().toString());
       fileDialog.setFilenameFilter((dir, name) -> GameDataFileUtils.isCandidateFileName(name));
       fileDialog.setVisible(true);
+
+      // FileDialog.getFiles() always returns an array
+      // of 1 or 0 items
       return Arrays.stream(fileDialog.getFiles()).findAny().map(this::mapFileResult);
     }
 
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameFileSelector.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameFileSelector.java
@@ -7,6 +7,7 @@ import games.strategy.triplea.settings.ClientSetting;
 import java.awt.FileDialog;
 import java.awt.Frame;
 import java.io.File;
+import java.util.Arrays;
 import java.util.Optional;
 import java.util.function.Consumer;
 import javax.annotation.Nonnull;
@@ -35,11 +36,7 @@ public final class GameFileSelector {
       fileDialog.setDirectory(ClientSetting.saveGamesFolderPath.getValueOrThrow().toString());
       fileDialog.setFilenameFilter((dir, name) -> GameDataFileUtils.isCandidateFileName(name));
       fileDialog.setVisible(true);
-      final String fileName = fileDialog.getFile();
-      final String dirName = fileDialog.getDirectory();
-      return Optional.ofNullable(fileName)
-          .map(name -> new File(dirName, fileName))
-          .map(this::mapFileResult);
+      return Arrays.stream(fileDialog.getFiles()).findAny().map(this::mapFileResult);
     }
 
     // Non-Mac platforms should use the normal Swing JFileChooser
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameFileSelector.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameFileSelector.java
@@ -38,7 +38,7 @@ public final class GameFileSelector {
       fileDialog.setVisible(true);
 
       // FileDialog.getFiles() always returns an array
-      // of 1 or 0 items
+      // of 1 or 0 items, because FileDialog.multipleMode is false by default
       return Arrays.stream(fileDialog.getFiles()).findAny().map(this::mapFileResult);
     }
 


game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameFileSelector.java
```diff
@@ -35,11 +36,7 @@
       fileDialog.setDirectory(ClientSetting.saveGamesFolderPath.getValueOrThrow().toString());
       fileDialog.setFilenameFilter((dir, name) -> GameDataFileUtils.isCandidateFileName(name));
       fileDialog.setVisible(true);
-      final String fileName = fileDialog.getFile();
-      final String dirName = fileDialog.getDirectory();
-      return Optional.ofNullable(fileName)
-          .map(name -> new File(dirName, fileName))
-          .map(this::mapFileResult);
+      return Arrays.stream(fileDialog.getFiles()).findAny().map(this::mapFileResult);
```
DanVanAtta: I think this needs documentation to make explicit that we expect 'getFiles()' to have 0 or 1 elements. Second, did you test this out to be sure it works and you get 0 values back if you type in a file that DNE?

DanVanAtta: I wonder if this is a case where it's worth having Mac OS specific code? The lack of mac skinning would not ideal, but I don't think we can readily test this otherwise. If we can guarantee that a selected file exists or none is chosen, then we can probably simplify the callback logic and remove the 'save-game-does-not-exist' warning message (that would be dead code otherwise, as-is in this update, it's conditionally dead-code depending on the OS (which is really not ideal)).

RoiEXLab: > I think this needs documentation to make explicit that we expect 'getFiles()' to have 0 or 1 elements.

Good point.

> Second, did you test this out to be sure it works and you get 0 values back if you type in a file that DNE?

Yes, no doubt there.

> I wonder if this is a case where it's worth having Mac OS specific code?

If you ask me it's silly using a different kind of dialog in the code for a specific OS just to get the skinning right.
I'm not even sure if JFileDialog looks different with the MacOS L&F, I could check that some time next week or something.
In case it doesn't look different because the L&F improved: Great, we can remove the code in the whole codebase.
In case it does look different we can ask the mac folks in the community how they feel about it. In case they have strong feelings about keeping it, we could create our own wrapper class that provides a higher level API so we don't have double logic everywhere.

But those kind of questions are definitely out of scope for this PR.

DanVanAtta:
@RoiEXLab I disabled the mac check, I was not able to select multiple files, nor specify a custom file to load. The API change does not seem to really be necessary.

The selection dialog also looked better using the native on linux. I wonder if the Windows experience is also good and if you can specify (type) a file to load that way. If not, the mac loading logic is probably the logic all OS should use.

RoiEXLab:
> I was not able to select multiple files

Of course not, this option is not enabled.

> nor specify a custom file to load

What do you mean by that?

> The API change does not seem to really be necessary.

Not necessary, but it's the intended way of using it. `FileDialog#getDirectory()` and `FileDialog#getFile()` are just the counterparts to the setters of those fields, which are used to set the placeholder values for the dialog + `getFile()` obviously doesn't even work when you use `FileDialog.setMultipleMode(true)` and select multiple files.

DanVanAtta:
Interestingly on linux JFileChooser does not look very good at all.

## FileDialog
Screenshot from 2020-03-07 13-36-49

## JFileChooser with motif
Screenshot from 2020-03-07 13-39-35

## JFileChooser with nimbus
Screenshot from 2020-03-07 13-38-46

## JFileChooser with substance
Screenshot from 2020-03-07 13-39-35

DanVanAtta:
>>     nor specify a custom file to load

>What do you mean by that?

In JFileChooser you can type the file you want to load. That was an easy way to repro a file not found bug; I forget exactly how but there was a path where the file to load was pre-populated and was always missing the .tsvg, so I was able to hit that error quite naturally, let alone intentionally hitting it by typing a file that DNE.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameFileSelector.java
```diff
@@ -35,11 +36,10 @@
...
       fileDialog.setFilenameFilter((dir, name) -> GameDataFileUtils.isCandidateFileName(name));
       fileDialog.setVisible(true);
-      final String fileName = fileDialog.getFile();
-      final String dirName = fileDialog.getDirectory();
-      return Optional.ofNullable(fileName)
-          .map(name -> new File(dirName, fileName))
-          .map(this::mapFileResult);
+
+      // FileDialog.getFiles() always returns an array
+      // of 1 or 0 items
```
DanVanAtta: The comment does not explain why we expect this to be true, why force a maintainer do API research when we can just say "this is the setting that makes this so".

RoiEXLab:
> Interestingly on linux JFileChooser does not look very good at all.

The current version of Ubuntu uses GNOME right? I don't think any of the Linux Distros have a really good L&F implementation for swing components. On Windows it's really hard to spot the difference. They do look slightly different, but I wouldn't be able to tell you which one is the right one if I didn't know.

RoiEXLab:
Also swing seems to have a pure java implementation of file choosers, whereas awt seems to use the native dialogs all of the time.

I just double checked how JavaFX handles this issue (in the JavaFX client there are some engine preferences where you can set folders) and JavaFX seems to use the native dialogs again, at least on windows. Would you mind checking that?

DanVanAtta:
@RoiEXLab while I agree about scope creep, I tend to think the scope is perhaps too narrow here and the updates could be more impactful.


We have a few other usages of FileDialog that perhaps could use similar updates, eg:

```
  public static File getSaveGameLocation(final Frame frame) {
    // For some strange reason, the only way to get a Mac OS X native-style file dialog
    // is to use an AWT FileDialog instead of a Swing JDialog
    if (SystemProperties.isMac()) {
      final FileDialog fileDialog = new FileDialog(frame);
      fileDialog.setMode(FileDialog.SAVE);
      fileDialog.setDirectory(ClientSetting.saveGamesFolderPath.getValueOrThrow().toString());
      fileDialog.setFilenameFilter((dir, name) -> GameDataFileUtils.isCandidateFileName(name));
      fileDialog.setVisible(true);

      final String fileName = fileDialog.getFile();
      if (fileName == null) {
        return null;
      }

      // If the user selects a filename that already exists,
      // the AWT Dialog on Mac OS X will ask the user for confirmation
      return new File(fileDialog.getDirectory(), GameDataFileUtils.addExtensionIfAbsent(fileName));
    }
```

But if we are going to go to this kind of effort, IMO we probably would just as well remove `JFileChooser` usage and amortize the testing cost and very happily remove OS specific code.

All that is needed is to improve the comment here. 

@RoiEXLab what are your thoughts on following up for a more impactful change that has the same testing cost and reduces our OS specific code?

DanVanAtta:
@RoiEXLab is there somewhere on the JavaFX UI we have to load a save game??

RoiEXLab:
> @RoiEXLab is there somewhere on the JavaFX UI we have to load a save game??

Not yet, but plenty of open folder dialogs in the settings menu

> @RoiEXLab what are your thoughts on following up for a more impactful change that has the same testing cost and reduces our OS specific code?

Sounds like a good idea, but I won't really have the capacity for this in the coming week.

DanVanAtta:
@RoiEXLab I'm not sure it would require that much time. I also like the idea of improving our API usage in the few places we use it, everywhere and consistently. In part, a deeper change/fix IMO is merited, perhaps even consolidating our dialog finder usage behind a utility. I suggest that as we get better consistency, fix a problem more completely, and preserve and re-use coding, review and testing context and efforts.

DanVanAtta:
Yes, JavaFX looks to always use the native file chooser.

RoiEXLab:
> Yes, JavaFX looks to always use the native file chooser.

That's good to know so our problem will in any case eventually go away.

DanVanAtta:
Following up: https://github.com/triplea-game/triplea/pull/6019, #6019 removes JFileChooser in places where we select FileDialog for MacOS and JFileChooser for others.

PR #6019, opened 2020-03-07
Avoid using JFileChooser in preference for FileDialog

DanVanAtta:
FileDialog uses a native native file chooser. This has some benefits:
- looks better
- less error prone, removing the option to type a file name means a user
  can't specify a file that does not exist
- removes OS specific code meaning we can test this code once and not
  N times per OS.
- Using native file chooser dialog aligns with JavaFX, in essence
   it's a modernizing approach to avoid JFileChooser

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
- Verified that the AWT dialog gives a file-overwrite warning

### Before
Screenshot from 2020-03-07 13-38-46

### After
Screenshot from 2020-03-07 13-36-49

## Additional Review Notes

The only changes in this PR is to delete the OS specific if conditional and delete the 'else' branch that used JFileChooser.

Scope of this update is:
 -  to remove OS specific file dialog selection and always favor FileDialog

Out of scope is removing all usages of JFileChooser or updating FileDialog API usages

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/menubar/TripleAMenuBar.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/menubar/TripleAMenuBar.java
@@ -2,8 +2,6 @@ package games.strategy.triplea.ui.menubar;
 
 import games.strategy.engine.framework.GameDataFileUtils;
 import games.strategy.engine.framework.startup.ui.InGameLobbyWatcherWrapper;
-import games.strategy.engine.framework.system.SystemProperties;
-import games.strategy.engine.framework.ui.SaveGameFileChooser;
 import games.strategy.engine.lobby.client.ui.action.EditGameCommentAction;
 import games.strategy.engine.lobby.client.ui.action.RemoveGameFromLobbyAction;
 import games.strategy.triplea.settings.ClientSetting;
@@ -13,10 +11,8 @@ import java.awt.Frame;
 import java.awt.event.KeyEvent;
 import java.io.File;
 import java.util.Optional;
-import javax.swing.JFileChooser;
 import javax.swing.JMenu;
 import javax.swing.JMenuBar;
-import javax.swing.JOptionPane;
 
 /** The game client menu bar. */
 public final class TripleAMenuBar extends JMenuBar {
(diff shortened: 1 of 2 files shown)


game-core/src/main/java/games/strategy/triplea/ui/menubar/TripleAMenuBar.java
```diff
@@ -60,60 +56,19 @@ private void createLobbyMenu(final InGameLobbyWatcherWrapper watcher) {
...
-            "Sub directories are not allowed in the file name.  Please rename it.",
-            "Cancel?",
-            JOptionPane.DEFAULT_OPTION,
-            JOptionPane.WARNING_MESSAGE);
-        return null;
-      }
-    }
-    f = new File(f.getParent(), GameDataFileUtils.addExtensionIfAbsent(f.getName()));
-    // A small warning so users will not over-write a file
-    if (f.exists()) {
```
RoiEXLab: Hmm I believe we should have that warning on all systems

game-core/src/main/java/games/strategy/triplea/ui/menubar/TripleAMenuBar.java
```diff
@@ -60,60 +56,19 @@ private void createLobbyMenu(final InGameLobbyWatcherWrapper watcher) {
...
-    // Non-Mac platforms should use the normal Swing JFileChooser
-    final JFileChooser fileChooser = SaveGameFileChooser.getInstance();
-    final int selectedOption = fileChooser.showSaveDialog(frame);
-    if (selectedOption != JFileChooser.APPROVE_OPTION) {
+    final String fileName = fileDialog.getFile();
+    if (fileName == null) {
       return null;
     }
-    File f = fileChooser.getSelectedFile();
-    // disallow sub directories to be entered (in the form directory/name) for Windows boxes
```
RoiEXLab: Ok, this is a super weird thing I have never seen before.

DanVanAtta: Always quite a bit of 'interesting' code to find here and there  :grin:

game-core/src/main/java/games/strategy/triplea/ui/menubar/TripleAMenuBar.java
DanVanAtta: It is at least for sure on linux, it seems to be an OS default baked in already :+1:

RoiEXLab: Good to know, just double checked if this is the case on windows as well and it is 👍

game-core/src/main/java/games/strategy/triplea/ui/menubar/TripleAMenuBar.java
RoiEXLab: I mean at this point I shouldn't really be surprised anymore, but it happens again and again ^^

PR #6021, opened 2020-03-07
SelectionComponentFactory: Remvoe unused methods and remove unnecessary boxing

DanVanAtta:


game-core/src/main/java/games/strategy/triplea/settings/SelectionComponentFactory.java
```diff
@@ -164,7 +162,7 @@ public void save(final SaveContext context) {
 
       private Optional<Integer> parsePort(final String encodedPort) {
         try {
-          final Integer port = Integer.valueOf(encodedPort);
+          final int port = Integer.parseInt(encodedPort);
```
RoiEXLab: Actually this causes the int to be boxed 1 time: Optional.of converts the int back to an Integer object.

PR #6022, opened 2020-03-08
Update DocumentListenerBuilder to have a buffering callback API

DanVanAtta:
Forward looking change to add functionality to DocumentListenerBuilder
that will buffer text listener callbackss.

The text update callback will be scheduled to be executed only after
an initial delay after the first text update event. But, if another
text update comes along, the previously scheduled callback is canceled
and a new one with delay is scheduled.

This allows for essentially (more or less) one callback to be done
after a user is done typing text rather than rapid firing many callback
events.

Usages of the previous non-buffered API will be replaced with the new API
in upcoming PRs

 

## Testing
- As part of other changes, demo'd the update and verified how many callbacks were done and observed UI responsiveness. Without the buffer, initial text updates usually first 3-5 events and then many events as you type. With the buffer, you still get multiple events but it's not nearly as many and text input is less laggy.

### DIFF
--- a/swing-lib/src/main/java/org/triplea/swing/DocumentListenerBuilder.java
+++ b/swing-lib/src/main/java/org/triplea/swing/DocumentListenerBuilder.java
@@ -1,17 +1,84 @@
 package org.triplea.swing;
 
+import java.util.List;
+import java.util.Optional;
+import java.util.Timer;
+import java.util.TimerTask;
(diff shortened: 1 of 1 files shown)


DanVanAtta:
Closing, commit is contained in: https://github.com/triplea-game/triplea/pull/6026

Issue #6025, opened 2020-03-09
Lobby Hosted Game Are Not De-Listed When Closed


DanVanAtta:
**How can the problem be recreated?**

- connect to lobby
- host a game

Do any of:
1) click cancel
2) click quit
3) launch game, then quit ( 'File' > 'exit program')

- notice in all 3 cases that the lobby does not de-list the hosted game right away.

**When the problem is fixed, what should have happened instead?**
Hosted game should be removed from lobby list in timely manner.

**Which Engine Version are you using?** 
2.0 @ cd8847e (Sat Mar 7 20:00:37 2020 -0800)

prastle:
Actually they do close/de-list it just takes a minute or so to occur for some reason.  Hope this helps.  @DanVanAtta

RoiEXLab:
@DanVanAtta If @prastle's observation is true all the time, that means that we don't properly close the Websocket connection, and are waiting for it to timeout instead.

DanVanAtta:
Thanks for the input, I can elaborate a bit on this issue and am working on a fix. The problem was introduced when we went from http polling to websockets communication for game listing. Before https://github.com/triplea-game/triplea/pull/6007 "Expire games" was merged just a few days ago, games were not removed at all. I think previously we had a far smaller TTL and it appears we relied on that exclusively. Now the TTL is longer so the problem is more noticeable. The issue is more we are not closing the game out properly when a websocket connection is closed, rather than not closing the connection. IE: the websocket close is not connected to closing the game, it does not trigger notifications correctly.

PR #6038, opened 2020-03-11
Notify game removed

DanVanAtta:
## Overview 
Resolves: https://github.com/triplea-game/triplea/issues/6025 "Game listings not removed" by adding pro-active callbacks to the lobby to notify that a game has been closed. Otherwise we will rely on TTL expiration to remove a game if the lobby game fails to send a closed notification pro-actively.

## Commits

commit 06c6c4e5503eb23d9cfe44b26533e3b245bde245

    Shutdown lobby watcher on server game cancel

commit d39d2f72faffd11f1f45cf56fca3f09ab1bbb766

    Shutdown server if lobby staging screen is closed

commit a7a1aa338

    Add system exit callback to close servermodel if program is exited
    
    This ensures that we will remove any listed lobby games if we launch
    a server game and then exit the program. If we simply leave the program
    we should be sure to actually keep the game listed as we just change
    state from running to waiting.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  #6025 
[] Other:   

## Testing
Verified the 3 scenarios in #6025 were fixed and general smoke testing around quit functionality.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
@@ -415,6 +415,7 @@ public class ServerModel extends Observable implements IConnectionChangeListener
         final URI lobbyUri = URI.create(System.getProperty(LOBBY_URI));
         gameHostingResponse = GameHostingClient.newClient(lobbyUri).sendGameHostingRequest();
 
+        ExitStatus.SUCCESS.addExitAction(this::cancel);
         remoteActionsListener =
             WebsocketListenerFactory.newListener(
                 lobbyUri,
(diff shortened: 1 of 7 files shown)


game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
```diff
@@ -422,7 +426,6 @@ private void createServerMessenger(
                     .bannedPlayerListener(new PlayerDisconnectAction(serverMessenger, this::cancel))
                     .shutdownListener(
                         emptyStringMessage -> {
-                          cancel();
```
RoiEXLab: What's wrong with calling cancel here?

DanVanAtta:
Exit will invoke cancel for us now, so it would be a double call to cancel

PR #6054, opened 2020-03-16
Avoid tile mutex

RoiEXLab:
Follow-up to #6046 

Note that I haven't done any performance testing, I just verified the changes seem to work.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

I started a local game and the tiles loaded no problem.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
@@ -632,9 +632,8 @@ public class MapPanel extends ImageScrollerLargeView {
       final Collection<Tile> tileList = tileManager.getTiles(bounds);
       for (final Tile tile : tileList) {
         tile.drawImage(gameData, uiContext.getMapData());
-        final Image img = tile.getImage();
         g2d.drawImage(
-            img,
+            tile.getImage(),
             AffineTransform.getTranslateInstance(tile.getBounds().x, tile.getBounds().y),
             this);
       }
--- a/game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
@@ -658,8 +658,6 @@ public class MapPanel extends ImageScrollerLargeView {
     int y = getYOffset();
     final List<Tile> images = new ArrayList<>();
     final List<Tile> undrawnTiles = new ArrayList<>();
-    // make sure we use the same data for the entire paint
-    final GameData data = gameData;
     // if the map fits on screen, don't draw any overlap
     final boolean drawHorizontalOverlap = !fittingWidth && uiContext.getMapData().scrollWrapX();
     final boolean drawVerticalOverlap = !fittingHeight && uiContext.getMapData().scrollWrapY();
@@ -676,22 +674,22 @@ public class MapPanel extends ImageScrollerLargeView {
           final Rectangle2D.Double leftUpperBounds =
               new Rectangle2D.Double(
                   model.getMaxWidth() + (double) x, model.getMaxHeight() + (double) y, -x, -y);
-          drawTiles(g2d, images, data, leftUpperBounds, undrawnTiles);
+          drawTiles(g2d, images, leftUpperBounds, undrawnTiles);
         }
         final Rectangle2D.Double leftBounds =
             new Rectangle2D.Double(model.getMaxWidth() + (double) x, y, -x, getScaledHeight());
-        drawTiles(g2d, images, data, leftBounds, undrawnTiles);
+        drawTiles(g2d, images, leftBounds, undrawnTiles);
       }
       if (drawVerticalOverlap && y < 0) {
         final Rectangle2D.Double upperBounds =
             new Rectangle2D.Double(x, model.getMaxHeight() + (double) y, getScaledWidth(), -y);
-        drawTiles(g2d, images, data, upperBounds, undrawnTiles);
+        drawTiles(g2d, images, upperBounds, undrawnTiles);
       }
     }
     // handle non overlap
     final Rectangle2D.Double mainBounds =
         new Rectangle2D.Double(x, y, getScaledWidth(), getScaledHeight());
-    drawTiles(g2d, images, data, mainBounds, undrawnTiles);
+    drawTiles(g2d, images, mainBounds, undrawnTiles);
     if (routeDescription != null) {
       if (mouseShadowImage != null && routeDescription.getEnd() != null) {
         final AffineTransform t = new AffineTransform();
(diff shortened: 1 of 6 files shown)


game-core/src/main/java/games/strategy/triplea/ui/screen/Tile.java
```diff
@@ -15,80 +15,66 @@
...
 
-  private final Image image;
+  private Image image = Util.newImage(TILE_SIZE, TILE_SIZE, true);
   private final Rectangle bounds;
   private final Object mutex = new Object();
   private final Queue<IDrawable> contents = new PriorityQueue<>();
 
   Tile(final Rectangle bounds) {
     this.bounds = bounds;
-    image = Util.newImage((int) bounds.getWidth(), (int) bounds.getHeight(), true);
```
RoiEXLab: I just noticed "hardcoding" the bounds to TILE_SIZE is no longer neccessary in the final form of this PR 🤔

game-core/src/main/java/games/strategy/triplea/ui/screen/Tile.java
```diff
@@ -15,80 +15,66 @@
...
     g.setColor(Color.BLACK);
-    g.fill(new Rectangle(0, 0, TileManager.TILE_SIZE, TileManager.TILE_SIZE));
-    final Queue<IDrawable> queue = new PriorityQueue<>(contents);
-    while (!queue.isEmpty()) {
-      queue.remove().draw(bounds, data, g, mapData);
-      // Make sure we don't mess up other draws
-      g.setTransform(original);
+    g.fill(new Rectangle(0, 0, TILE_SIZE, TILE_SIZE));
+    synchronized (mutex) {
+      final Queue<IDrawable> queue = new PriorityQueue<>(contents);
```
RoiEXLab: As you can see the only thing that "mutex" does now is synchronize the access to `contents` and `isDirty` as a single operation.
Not sure if it would work without any unintended problems if we made contents a concurrent collection of some sort and removed the mutex. Currently calling `new PriorityQueue<>(contents)` is a constant time operation, changing the type of contents would make it `O(n * log n)`.
We could however use a concurrent set where the draw level is compared, and if even we compare the hashCodes or something in order not to "remove" "duplicate" tiles with the same draw level as others.

game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
```diff
@@ -631,16 +631,12 @@ public void drawMapImage(final Graphics g) {
...
-          if (img != null) {
-            g2d.drawImage(
-                img,
-                AffineTransform.getTranslateInstance(
-                    tile.getBounds().x - bounds.getX(), tile.getBounds().y - bounds.getY()),
-                this);
-          }
-        }
+        tile.drawImage(gameData, uiContext.getMapData());
+        final Image img = tile.getImage();
```
DanVanAtta: Does it make sense to have `drawImage` to also return the rendered image? 

EG:
```
final var image = tile.drawImage(gameData, uiContext.getMapData())
```

game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
```diff
@@ -746,13 +740,15 @@ public void paint(final Graphics g) {
     updateUndrawnTiles(undrawnTiles, 513);
     updateUndrawnTiles(undrawnTiles, 767);
     clearPendingDrawOperations();
+    // make sure we use the same data for the entire paint
+    final GameData data = gameData;
```
DanVanAtta: Kinda odd as `GameData` is quite mutable. I wonder if this would be more easily communicated if we extracted the logic that needs `GameData` to a method. Thoughts?

game-core/src/main/java/games/strategy/triplea/ui/screen/Tile.java
```diff
@@ -15,80 +15,66 @@
...
-   *
-   * @return the image we currently have.
-   */
-  public Image getRawImage() {
-    synchronized (mutex) {
-      return image;
-    }
+  /** This image may not reflect our current drawables. */
+  public Image getImage() {
+    return image;
```
DanVanAtta: Thoughts to replace this with `@Getter` and move and consolidate the comment to the `image` variable?

game-core/src/main/java/tools/image/TileImageBreaker.java
```diff
@@ -107,33 +107,29 @@ private void createMaps() throws IOException {
...
-      for (int y = 0; y * TileManager.TILE_SIZE < map.getHeight(null); y++) {
+    for (int x = 0; x * Tile.TILE_SIZE < map.getWidth(null); x++) {
+      for (int y = 0; y * Tile.TILE_SIZE < map.getHeight(null); y++) {
         final Rectangle bounds =
-            new Rectangle(
-                x * TileManager.TILE_SIZE,
-                y * TileManager.TILE_SIZE,
-                TileManager.TILE_SIZE,
-                TileManager.TILE_SIZE);
+            new Rectangle(x * Tile.TILE_SIZE, y * Tile.TILE_SIZE, Tile.TILE_SIZE, Tile.TILE_SIZE);
```
DanVanAtta: Minor, `Tile.TILE_SIZE` is a candidate for a static import, used quite a bit.

game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
RoiEXLab: IIRC the reason this variable was introduced is because we got NPEs on shutdown, when the gameData field was set to null mid-render.
This prevents this issue.

game-core/src/main/java/games/strategy/triplea/ui/screen/Tile.java
RoiEXLab: I'm not even sure if this method needs a comment tbh

game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
RoiEXLab: I thought about that, but by returning void it becomes obvious that this method is not just creating a new Image object, but rather altering the state of the Tile

RoiEXLab:
@DanVanAtta Updated

game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
DanVanAtta: If we sent the `gameData` reference as a method parameter, there would be no way for anything outside of the method to mutate that reference (value). We would surely guard against that scenario. Given this is also a long method, breaking up the dependencies on `gameData` could be good as well.

IMO that is a good option. If not, then we probably need more commentary/documentation to state we are defending against the `gameData` reference from not only being changed, but also potentially nulled out as part of a shutdown event.

game-core/src/main/java/games/strategy/triplea/ui/screen/Tile.java
DanVanAtta: If the comment is dropped or not, still holds that we could avoid the boiler plate getter as well. Any thoughts?

DanVanAtta: I see now that was done: https://github.com/triplea-game/triplea/pull/6054/commits/147182a50908b93429f4abee6b0c635b58a4f86f

Issue #6027, opened 2020-03-10
Bot using SLF4J no-op logging


DanVanAtta:
**How can the problem be recreated?**

On a system with a deployed bot, run:
```
sudo systemctl restart bot@01.service
sudo systemctl status bot@01.service
```

Notice the last few log messages:
```
Mar 10 03:33:10 li939-161 run_server[1305]: SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
Mar 10 03:33:10 li939-161 run_server[1305]: SLF4J: Defaulting to no-operation (NOP) logger implementation
Mar 10 03:33:10 li939-161 run_server[1305]: SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

**When the problem is fixed, what should have happened instead?**
Should get logging, either by providing a logging slf4j or removing the slf4j dep.

**Which Engine Version are you using?** 
triplea-game-headless-2.0.18289.jar

PR #6062, opened 2020-03-19
Remove websocket dependency

RoiEXLab:
When looking at #6027 it seems like the Java-WebSocket dependency introduced the SLF4J dependency according to gradle (`gradle game-headless:dependencies`).

So I decided to use the default JDK API for websockets instead. It works slightly different with some gotchas and more fine grained control, but I verified it actually works.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  #6027 (Haven't actually verified that, but it no longer has the dependency)
[] Other:   

## Testing

I verified I could join the lobby and send some chat messages (it uses the same code)

### DIFF
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
@@ -57,7 +57,7 @@ public class GenericWebSocketClient implements WebSocketConnectionListener {
             throwable -> {
               log.log(
                   Level.SEVERE, "Unexpected exception completing websocket connection", throwable);
-              return false;
+              return null;
             });
   }
 
@@ -95,7 +95,7 @@ public class GenericWebSocketClient implements WebSocketConnectionListener {
   }
 
   @Override
-  public void handleError(final Exception exception) {
-    log.log(Level.SEVERE, "Websocket error", exception);
+  public void handleError(final Throwable error) {
+    log.log(Level.SEVERE, "Websocket error", error);
   }
 }
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
@@ -51,14 +51,7 @@ public class GenericWebSocketClient implements WebSocketConnectionListener {
 
   public void registerListenerAndConnect(final Consumer<ServerMessageEnvelope> messageListener) {
     this.messageListener = messageListener;
-    client
-        .connect(this, errorHandler)
-        .exceptionally(
-            throwable -> {
-              log.log(
-                  Level.SEVERE, "Unexpected exception completing websocket connection", throwable);
-              return null;
-            });
+    client.connect(this, errorHandler);
   }
 
   /**
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
@@ -32,7 +32,7 @@ public class GenericWebSocketClient implements WebSocketConnectionListener {
   private Consumer<ServerMessageEnvelope> messageListener;
 
   public GenericWebSocketClient(final URI lobbyUri, final Consumer<String> errorHandler) {
-    this(new WebSocketConnection(swapHttpsToWssProtocol(lobbyUri)), errorHandler);
+    this(new WebSocketConnection(swapHttpToWsProtocol(lobbyUri)), errorHandler);
   }
 
   @VisibleForTesting
@@ -43,9 +43,9 @@ public class GenericWebSocketClient implements WebSocketConnectionListener {
   }
 
   @VisibleForTesting
-  static URI swapHttpsToWssProtocol(final URI uri) {
-    return uri.getScheme().equals("https")
-        ? URI.create(uri.toString().replace("https", "wss"))
+  static URI swapHttpToWsProtocol(final URI uri) {
+    return uri.getScheme().matches("^https?$")
+        ? URI.create(uri.toString().replaceFirst("^http", "ws"))
         : uri;
   }
 
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
@@ -157,6 +157,8 @@ class WebSocketConnection {
    * @param errorHandler Invoked if there is a failure to connect.
    * @throws IllegalStateException Thrown if connection is already open (eg: connect called twice).
    * @throws IllegalStateException Thrown if connection has been closed (ie: 'close()' was called)
+   * @return A {@link CompletableFuture} behaving like the {@link CompletableFuture} returned by
+   *     {@link WebSocket.Builder#buildAsync(URI, Listener)} in case of an error.
    */
   CompletableFuture<WebSocket> connect(
       final WebSocketConnectionListener listener, final Consumer<String> errorHandler) {
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
@@ -71,7 +71,7 @@ class WebSocketConnection {
   @Getter(
       value = AccessLevel.PACKAGE,
       onMethod_ = {@VisibleForTesting})
-  private final WebSocket.Listener webSocketListener = new WebSocketListener();
+  private final WebSocket.Listener internalListener = new InternalWebSocketListener();
 
   @Getter(
       value = AccessLevel.PACKAGE,
@@ -138,7 +138,7 @@ class WebSocketConnection {
     return httpClient
         .newWebSocketBuilder()
         .connectTimeout(Duration.ofMillis(DEFAULT_CONNECT_TIMEOUT_MILLIS))
-        .buildAsync(serverUri, webSocketListener);
+        .buildAsync(serverUri, internalListener);
   }
 
   /**
@@ -172,7 +172,7 @@ class WebSocketConnection {
   }
 
   @VisibleForTesting
-  class WebSocketListener implements Listener {
+  class InternalWebSocketListener implements Listener {
     private final StringBuilder textAccumulator = new StringBuilder();
 
     @Override
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
@@ -71,51 +71,7 @@ class WebSocketConnection {
   @Getter(
       value = AccessLevel.PACKAGE,
       onMethod_ = {@VisibleForTesting})
-  private final WebSocket.Listener webSocketListener =
-      new Listener() {
-        private final StringBuilder textAccumulator = new StringBuilder();
-
-        @Override
-        public void onOpen(final WebSocket webSocket) {
-          synchronized (queuedMessages) {
-            client = webSocket;
-            connectionIsOpen = true;
-            queuedMessages.forEach(
-                message ->
-                    client
-                        .sendText(message, true)
-                        .exceptionally(
-                            logWebSocketError(Level.SEVERE, "Failed to send queued text.")));
-            queuedMessages.clear();
-          }
-          webSocket.request(1);
-        }
-
-        @Override
-        public CompletionStage<?> onText(
-            final WebSocket webSocket, final CharSequence data, final boolean last) {
-          textAccumulator.append(data);
-          if (last) {
-            listener.messageReceived(textAccumulator.toString());
-            textAccumulator.setLength(0);
-          }
-          webSocket.request(1);
-          return null;
-        }
-
-        @Override
-        public CompletionStage<?> onClose(
-            final WebSocket webSocket, final int statusCode, final String reason) {
-          pingSender.cancel();
-          listener.connectionClosed(reason);
-          return null;
-        }
-
-        @Override
-        public void onError(final WebSocket webSocket, final Throwable error) {
-          listener.handleError(error);
-        }
-      };
+  private final WebSocket.Listener webSocketListener = new WebSocketListener();
 
   @Getter(
       value = AccessLevel.PACKAGE,
@@ -215,10 +171,48 @@ class WebSocketConnection {
     };
   }
 
-  private <T> Function<Throwable, T> invokeErrorListenerInExceptionCase() {
-    return throwable -> {
-      listener.handleError(throwable);
+  @VisibleForTesting
+  class WebSocketListener implements Listener {
+    private final StringBuilder textAccumulator = new StringBuilder();
+
+    @Override
+    public void onOpen(final WebSocket webSocket) {
+      synchronized (queuedMessages) {
+        client = webSocket;
+        connectionIsOpen = true;
+        queuedMessages.forEach(
+            message ->
+                client
+                    .sendText(message, true)
+                    .exceptionally(logWebSocketError(Level.SEVERE, "Failed to send queued text.")));
+        queuedMessages.clear();
+      }
+      webSocket.request(1);
+    }
+
+    @Override
+    public CompletionStage<?> onText(
+        final WebSocket webSocket, final CharSequence data, final boolean last) {
+      textAccumulator.append(data);
+      if (last) {
+        listener.messageReceived(textAccumulator.toString());
+        textAccumulator.setLength(0);
+      }
+      webSocket.request(1);
       return null;
-    };
+    }
+
+    @Override
+    public CompletionStage<?> onClose(
+        final WebSocket webSocket, final int statusCode, final String reason) {
+      pingSender.cancel();
+      listener.connectionClosed(reason);
+      return null;
+    }
+
+    @Override
+    public void onError(final WebSocket webSocket, final Throwable error) {
+      listener.handleError(error);
+    }
   }
 }
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
@@ -106,9 +106,6 @@ class WebSocketConnection {
         @Override
         public CompletionStage<?> onClose(
             final WebSocket webSocket, final int statusCode, final String reason) {
-          if (!reason.equals(CLIENT_DISCONNECT_MESSAGE)) {
-            log.severe("Connection to server closed: " + reason);
-          }
           pingSender.cancel();
           listener.connectionClosed(reason);
           return null;
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
@@ -187,17 +187,21 @@ class WebSocketConnection {
                     .exceptionally(logWebSocketError(Level.SEVERE, "Failed to send queued text.")));
         queuedMessages.clear();
       }
+      // Allow onText to be called at least once, WebSocketConnection is initialized
       webSocket.request(1);
     }
 
     @Override
     public CompletionStage<?> onText(
         final WebSocket webSocket, final CharSequence data, final boolean last) {
+      // No need to synchronize access, this listener is never called concurrently
+      // and always called in-order by the API
       textAccumulator.append(data);
       if (last) {
         listener.messageReceived(textAccumulator.toString());
         textAccumulator.setLength(0);
       }
+      // We're done processing, allow listener to be called again at least once
       webSocket.request(1);
       return null;
     }
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
@@ -97,7 +97,7 @@ class WebSocketConnection {
   /** Does an async close of the current websocket connection. */
   void close() {
     closed = true;
-    if (!client.isOutputClosed()) {
+    if (client != null && !client.isOutputClosed()) {
       client
           .sendClose(WebSocket.NORMAL_CLOSURE, CLIENT_DISCONNECT_MESSAGE)
           .exceptionally(logWebSocketError(Level.INFO, "Failed to close"));
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
@@ -3,17 +3,23 @@ package org.triplea.http.client.web.socket;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.WebSocket;
+import java.net.http.WebSocket.Listener;
+import java.nio.ByteBuffer;
+import java.time.Duration;
 import java.util.ArrayDeque;
 import java.util.Queue;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.logging.Level;
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.java.Log;
-import org.java_websocket.client.WebSocketClient;
-import org.java_websocket.handshake.ServerHandshake;
 import org.triplea.java.timer.ScheduledTimer;
 import org.triplea.java.timer.Timers;
 
@@ -31,6 +37,8 @@ import org.triplea.java.timer.Timers;
 @Log
 class WebSocketConnection {
   @VisibleForTesting static final int DEFAULT_CONNECT_TIMEOUT_MILLIS = 5000;
+
+  private static final String CLIENT_DISCONNECT_MESSAGE = "Client disconnect.";
 
   private WebSocketConnectionListener listener;
 
@@ -59,7 +67,7 @@ class WebSocketConnection {
   @Setter(
       value = AccessLevel.PACKAGE,
       onMethod_ = {@VisibleForTesting})
-  private WebSocketClient client;
+  private WebSocket client;
 
   @Getter(
       value = AccessLevel.PACKAGE,
@@ -68,44 +76,16 @@ class WebSocketConnection {
 
   WebSocketConnection(final URI serverUri) {
     this.serverUri = serverUri;
-    client =
-        new WebSocketClient(serverUri) {
-          @Override
-          public void onOpen(final ServerHandshake serverHandshake) {
-            synchronized (queuedMessages) {
-              connectionIsOpen = true;
-              queuedMessages.forEach(this::send);
-              queuedMessages.clear();
-            }
-          }
-
-          @Override
-          public void onMessage(final String message) {
-            listener.messageReceived(message);
-          }
-
-          @Override
-          public void onClose(final int code, final String reason, final boolean remote) {
-            if (remote) {
-              log.severe("Connection to server closed: " + reason);
-            }
-            pingSender.cancel();
-            listener.connectionClosed(reason);
-          }
-
-          @Override
-          public void onError(final Exception exception) {
-            listener.handleError(exception);
-          }
-        };
     pingSender =
         Timers.fixedRateTimer("websocket-ping-sender")
             .period(45, TimeUnit.SECONDS)
             .delay(45, TimeUnit.SECONDS)
             .task(
                 () -> {
-                  if (client.isOpen()) {
-                    client.sendPing();
+                  if (!client.isOutputClosed()) {
+                    client
+                        .sendPing(ByteBuffer.wrap(new byte[0]))
+                        .exceptionally(logWebSocketError("Failed to send ping."));
                   }
                 });
   }
@@ -113,7 +93,11 @@ class WebSocketConnection {
   /** Does an async close of the current websocket connection. */
   void close() {
     closed = true;
-    client.close();
+    if (!client.isOutputClosed()) {
+      client
+          .sendClose(WebSocket.NORMAL_CLOSURE, CLIENT_DISCONNECT_MESSAGE)
+          .exceptionally(logWebSocketError("Failed to close"));
+    }
   }
 
   /**
@@ -123,16 +107,16 @@ class WebSocketConnection {
    * @throws IllegalStateException Thrown if connection is already open (eg: connect called twice).
    * @throws IllegalStateException Thrown if connection has been closed (ie: 'close()' was called)
    */
-  CompletableFuture<Boolean> connect(
+  CompletableFuture<WebSocket> connect(
       final WebSocketConnectionListener listener, final Consumer<String> errorHandler) {
     this.listener = Preconditions.checkNotNull(listener);
-    Preconditions.checkState(!client.isOpen());
+    Preconditions.checkState(client == null);
     Preconditions.checkState(!closed);
 
     return connectAsync()
         .whenComplete(
-            (connected, throwable) -> {
-              if (connected && throwable == null) {
+            (webSocket, throwable) -> {
+              if (webSocket != null && throwable == null) {
                 pingSender.start();
               } else {
                 errorHandler.accept("Failed to connect to: " + serverUri);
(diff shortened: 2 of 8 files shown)


http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -49,17 +58,67 @@
...
+          }
+          webSocket.request(1);
+          return null;
+        }
+
+        @Override
+        public CompletionStage<?> onClose(
+            final WebSocket webSocket, final int statusCode, final String reason) {
+          if (!reason.equals(CLIENT_DISCONNECT_MESSAGE)) {
+            log.severe("Connection to server closed: " + reason);
```
DanVanAtta: The server will disconnect banned clients, this is probably not a case where we want to do logging. Likely we want to show an info message, probably just let the connection closed listener handle it.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -49,17 +58,67 @@
...
+                        .sendText(message, true)
+                        .exceptionally(logWebSocketError("Failed to send queued text.")));
+            queuedMessages.clear();
+          }
+          webSocket.request(1);
+        }
+
+        @Override
+        public CompletionStage<?> onText(
+            final WebSocket webSocket, final CharSequence data, final boolean last) {
```
DanVanAtta: I like that we continue to abstract away the underlying API, though `CharSequence` and `boolean last` seem to be not as nice of an API. I think this websocket library loses points here.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -68,52 +127,28 @@
...
             .period(45, TimeUnit.SECONDS)
             .delay(45, TimeUnit.SECONDS)
             .task(
                 () -> {
-                  if (client.isOpen()) {
-                    client.sendPing();
+                  if (!client.isOutputClosed()) {
+                    client
+                        .sendPing(ByteBuffer.wrap("Ping".getBytes(StandardCharsets.UTF_8)))
+                        .exceptionally(logWebSocketError("Failed to send ping."));
```
DanVanAtta: If the ping cannot be sent, would it maybe mean the connection is already closed?
Would we also want users to submit bug reports whenever this happens?

I suspect we may want to no-op if the ping cannot be sent.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -123,39 +158,28 @@ void close() {
...
-              }
-              completableFuture.complete(connected);
-            })
-        .start();
-    return completableFuture;
+  private CompletableFuture<WebSocket> connectAsync() {
+    return httpClient
+        .newWebSocketBuilder()
+        .connectTimeout(Duration.ofMillis(DEFAULT_CONNECT_TIMEOUT_MILLIS))
+        .buildAsync(serverUri, webSocketListener);
```
DanVanAtta: This builder is kinda nice, this library gains a point IMO here for this.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
RoiEXLab: Could potentially be the case if we close it between actually sending the ping and the if check, but the exception type will be IOException, so we'll have to check the outputstream again to be sure

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
RoiEXLab: Not a real fan of it either, but by looking at the documentation it seems that the API in theory allows you to do everything async, including message processing of messages that get split.

I actually believe we could ignore the last boolean. When debugging some issue I was having (due to wrong usage of the API) I realized that the data messages seem complete for the most part.

RoiEXLab:
> [...] I'm not sure if we really should be switching websocket library just for the sake of not adding a slf4j logging dependency or even ignoring that problem for now.

Actually the main motivation was to get rid of a dependency that has a potential replacement in the standard library. Even though in this particular case Java-WebSocket seems like a solid and well maintained library. Relying on the standard library just gives us this extra bit of safety that the API will receive bugfixes and get updated from time to time.
The fact that this also potentially gets rid of the warning is a nice side-effect IMO.

I agree that the standard lib API is a bit "closer to the metal" and has this little bit of extra overhead we don't really make use of. But at the same time I really like how it splits the sending part of the client and the receiving part of the client. If Java had a built-in wait to await futures like other languages do I believe this library would be a real pleasure to use, but I agree that the Futures do also add to the overhead I already mentioned, even though in theory this is a pretty solid approach for error hanling or synchronous request if your use-case would require it.

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
```diff
@@ -127,21 +180,13 @@ private void verifyPingerIsStarted() {
...
-    }
-
     @Test
     @DisplayName("Verify connect failing invokes error handler and pinger is not running")
-    void connectionFailure() throws Exception {
-      givenWebSocketConnects(false);
-
-      final boolean connected =
-          webSocketConnection.connect(webSocketConnectionListener, errorHandler).get();
+    void connectionFailure() {
```
DanVanAtta: I'm seeing this test fail when run from IDE:
```
org.opentest4j.AssertionFailedError: Expected java.util.concurrent.ExecutionException to be thrown, but nothing was thrown.

	at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:71)
	at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:37)
	at org.junit.jupiter.api.Assertions.assertThrows(Assertions.java:2952)
	at org.triplea.http.client.web.socket.WebSocketConnectionTest$SendMessageAndConnect.connectionFailure(WebSocketConnectionTest.java:186)
```
I think I'd quibble a bit more with the test case otherwise. The `@DisplayName` message is not really accurate anymore. It's also very notable that we can/need to rely on exception handling.

It's also notable as well that the method does not document the `ExecutionException` as being a possibility for somethign to be thrown:
```
   * @throws IllegalStateException Thrown if connection is already open (eg: connect called twice).
   * @throws IllegalStateException Thrown if connection has been closed (ie: 'close()' was called)
   */
```

Third, we pass an error handler to the method, yet we also have an exception that is possible to be thrown. Part of the purpose of the error handler is to invoke that instead of throwing an exception.

The local IDE failure does need to be sorted out, we can't have tests being non-deterministic and breaking when run from one context and not another. I suspect getting back to a boolean returning only method and not allowing an exception to be thrown, returning false and invoking the error handler would be more inline with the original method contract.

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
```diff
@@ -154,52 +199,36 @@ private void verifyErrorHandlerWasCalled() {
       verify(errorHandler).accept(any());
     }
 
-    @Test
-    @DisplayName(
-        "Verify connect failing with exception invokes error handler and pinger is not running")
```
DanVanAtta: Why is this test no longer applicable?

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
```diff
@@ -154,52 +199,36 @@ private void verifyErrorHandlerWasCalled() {
...
     @Test
     @DisplayName("Close will close the underlying socket and stops the pinger")
     void close() {
+      webSocketConnection.getWebSocketListener().onOpen(webSocket);
+      requiresCloseAction();
       webSocketConnection.close();
 
-      verify(webSocketClient).close();
+      verify(webSocket)
+          .sendClose(WebSocket.NORMAL_CLOSURE, WebSocketConnection.CLIENT_DISCONNECT_MESSAGE);
```
DanVanAtta: It does not seem to matter what the disconnect message is, or does it? We probably do not need to assert a specific message.

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
```diff
@@ -93,7 +128,16 @@ void queuedMessagesAreFlushedOnConnectionOpen() {
...
       webSocketConnection = new WebSocketConnection(INVALID_URI);
-      webSocketConnection.setClient(webSocketClient);
+    }
+
+    private void requiresSendTextAction() {
+      when(webSocket.sendText(any(), anyBoolean()))
+          .thenReturn(CompletableFuture.completedFuture(null));
+    }
+
+    private void requiresCloseAction() {
```
DanVanAtta: This method seems out of order, it is invoked much further down in the test, right?

Second, why does this method make sense, what does requiring a close action mean and have to do with a send close returning a nullable future?

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
DanVanAtta: > Would we also want users to submit bug reports whenever this happens?

My implication with this was that given the current handling, we are offering users to submit bug reports. Severe logging means "the application crashed, it should not have, the error is likely unrecoverable, the user should submit a bug report as the situation should never happen"

We should not be getting bug reports where our response is: "yeah, that's normal, we'll close this now."

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
```diff
@@ -43,9 +43,9 @@ public GenericWebSocketClient(final URI lobbyUri, final Consumer<String> errorHa
   }
 
   @VisibleForTesting
-  static URI swapHttpsToWssProtocol(final URI uri) {
-    return uri.getScheme().equals("https")
-        ? URI.create(uri.toString().replace("https", "wss"))
+  static URI swapHttpToWsProtocol(final URI uri) {
+    return uri.getScheme().matches("^https?$")
+        ? URI.create(uri.toString().replaceFirst("^http", "ws"))
```
DanVanAtta: Not sure if you're getting a bit over-paranoid with `replaceFirst`, the regex `^` really guarantees one match here. 

It seems like we are defending against the protocol from being `shttp`? Otherwise the protocol string being equal to https was surefire already.

Do we have to replace `http` protocol with `ws`?

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
```diff
@@ -57,7 +57,7 @@ public void registerListenerAndConnect(final Consumer<ServerMessageEnvelope> mes
             throwable -> {
               log.log(
                   Level.SEVERE, "Unexpected exception completing websocket connection", throwable);
-              return false;
+              return null;
```
DanVanAtta: This block is notable as the error handling is pretty screwed up to begin with. We have an error handler passed to the connect function, yet we still choose to have our own custom error handling on this out layer. Really the inner call `connect` should take care of the `exceptionally` case and invoke the error handler. As-is, it seems like the error-handler is a violation of DRY (expressing the same idea in multiple places) or arguably SRP.

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
```diff
@@ -52,39 +55,71 @@ void tearDown() {
 
     @Test
     void onMessage() {
-      webSocketConnection.getClient().onMessage(MESSAGE);
+      webSocketConnection.getWebSocketListener().onText(mock(WebSocket.class), MESSAGE, true);
       verify(webSocketConnectionListener).messageReceived(MESSAGE);
     }
 
+    @Test
+    void verifyListenerAccumulatesMessagesUntilLast() {
```
DanVanAtta: I think we may need some additional test cases that exercise concurrency with multiple, broken up messages. I'm concerned about that scenario.

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
```diff
@@ -52,39 +55,71 @@ void tearDown() {
 
     @Test
     void onMessage() {
-      webSocketConnection.getClient().onMessage(MESSAGE);
+      webSocketConnection.getWebSocketListener().onText(mock(WebSocket.class), MESSAGE, true);
```
DanVanAtta: Have you considered making the websocket listener a stand-alone class? This kind of test is a smell, (law of demeter violation). It we simpliy injected the listener implementation we might be able to decompose the test cases more thoroughly.

DanVanAtta:
Any updates or future plans for this @RoiEXLab ?

Summarizing open questions/issues:
-  how is concurrency and message splitting handled? I wonder if our existing implementation handles it, or if it actually is buggy in that case.
- need some attention to test cases

RoiEXLab:
@DanVanAtta I think I'll get to this PR today.

> how is concurrency and message splitting handled? I wonder if our existing implementation handles it, or if it actually is buggy in that case.

I think the API has a good solution for this, which allows multiple behaviours.
Look out for the `webSocket.request(1);` calls. This line tells the API "hey I'm ready to receive 1 more packet".
If this is the last call on every Listener packets are received in-order without concurrency problems.

If your API requires concurrent messages, you can do that as well by calling this line early in your listener or increasing the number from 1 to something else.

DanVanAtta:
Ok. Interesting. It does sound like the test cost is going to be somewhat
high. We're sure it is worth it?

DanVanAtta:
Keep in mind there will be concurrent incoming messages, testing that could
be tricky. It's pretty low level code as well, in effect we're writing our
own light weight wrapper library.

DanVanAtta:
Just occurred to me, how do you mean by listener packets @RoiEXLab ? 

This might get tricky. GenericWebsocket provides a pretty strong abstraction, users of that class should not know about JSON, packet ordering or packets being broken up. Essentially a user of GeneicWebsocket should say "send me messages that are of this type". GenericWebsocket then handles all other concerns, including cracking open the message envelope and casting the contents to the right type. If we need to have multiple listeners to aggregate all the messages back together, it sounds like it'll all be up to GenericWebsocket and to have multiple listeners. I'm not sure how clean that would work. Be careful of the [sunk cost fallacy](https://www.behavioraleconomics.com/resources/mini-encyclopedia-of-be/sunk-cost-fallacy/) here, needs to be worth the effort.

RoiEXLab:
> It does sound like the test cost is going to be somewhat
high. We're sure it is worth it?

I actually don't think it makes tests more complicated. If tests just call the listener without any multithreading nothing changes. Only if we want to create tests that sort of simulate the original API (for some weird reason?) things would get more complicated.

> Keep in mind there will be concurrent incoming messages, testing that could
be tricky. It's pretty low level code as well, [...]

Actually I wouldn't consider it low-level. From my understanding the java API is a wrapper around the low-level stuff, but without losing any of the fine-grained control.
For example AFAIK the JS WebSocket API in browsers doesn't even have ping/pong commands, simply because it's the browsers resposibility to keep the connection alive.
So I agree that we lose some convenience, but at the same time I'm far from calling like a couple of additional lines a dealbreaker.

> [...]  in effect we're writing our
own light weight wrapper library.

But isn't this what we're doing already by having a class abstracting away all the unecessary logic for our use-case and handling pings automatically? If we really want to avoid "lightweight frameworks", we could just pass the raw listener types to the "raw" API and call it a day.

> Just occurred to me, how do you mean by listener packets @RoiEXLab ?

I just had a look at the WebSocket specification, and it does seem that the way the WebsocketListener works is that it's called for every packet on the websocket level (not the TCP or even IP level).
WebsocketPackets can have variable length up to 2^(7+64) bytes in length if I understand the spec correctly, but can be split up into multiple packets (where the `last` boolean/bit comes into play).
I don't think we'll ever hit this limit, so we could probably just ignore this bit for our use case if we wanted to, depending on server-side implementation.
I didn't build it in because it was needed, just because it could potentially be needed (I haven't actually tested if we can ignore the boolean altogether, but I'm 90% sure we can).


TL;DR: I don't feel like this PR has a great downside. Yes, sorting out the current issues will be some work, but due to COVID-19 I currently have the capacity to do it if I'd actually take the time for it instead of watching another Episode of <Series> on <Streaming Service>

DanVanAtta:
> But isn't this what we're doing already by having a class abstracting away all the unecessary logic for our use-case and handling pings automatically? If we really want to avoid "lightweight frameworks", we could just pass the raw listener types to the "raw" API and call it a day.

I suppose it is debatable where the line is drawn. The generic websocket layer is meant to be mostly an example of [humble object pattern](https://ieftimov.com/post/tdd-humble-object/). In general wrapping dependencies and boundaries, creating an API that is both stable, friendly, and abstracts away the details of the underlying technology is a good thing. What's behind the API is ideally simple and mostly just uses a library without much logic underneath it. The checks we have for buffering messages IMO was a bit overboard, but it seems to work okay. I think part of the issue is that we're honoring too much a legacy code base and are building complicated things so we can plug them in to places that are have odd call-patterns and dependencies... Regardless, humble object : )

> TL;DR: I don't feel like this PR has a great downside.

No, not really, only if we introduce bugs somewhere or if the testing takes a really long runtime. I would like to see the abstraction of atomic messages be maintained, as that was the point behind the humble object and the original library to begin with.  

Time is extremely limited, so if we can get this done and move on to the tasks that support 2.0 to get that released finally and imminently, that would be the right focus.

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
RoiEXLab: Yes, because the JDK API requires the protocol to be wss, http doesn't work

ron-murhammer:
@DanVanAtta @RoiEXLab I didn't read all the discussion on this but can we move this forward, it looks like its been sitting for a few days?

DanVanAtta:
@ron-murhammer, been waiting for updates. The abstraction of the underlying websocket library is a bit different. We no longer get a string message but instead a charsequence that can be broken up over multiple messages. The 'GenericWebsocket' is fundamentally there to provide an abstraction over the websocket so that  you can register listeners and get a de-JSON'ified object back as a single message. We need to maintain this abstraction and second, I'm concerned/curious how the new library handles concurrency, websocket communication *will* be concurrent.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -68,52 +127,28 @@
...
         Timers.fixedRateTimer("websocket-ping-sender")
             .period(45, TimeUnit.SECONDS)
             .delay(45, TimeUnit.SECONDS)
             .task(
                 () -> {
-                  if (client.isOpen()) {
-                    client.sendPing();
+                  if (!client.isOutputClosed()) {
+                    client
+                        .sendPing(ByteBuffer.wrap("Ping".getBytes(StandardCharsets.UTF_8)))
```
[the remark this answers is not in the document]
RoiEXLab: We don't have to, I could as well use an empty array,
Quoting the websocket spec:

> A Ping frame MAY include "Application data".

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
RoiEXLab: It does no longer matter when we don't check for it.
(There's this check that logs an "exception" when the connection is closed server side, which you suggested should be removed)

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
RoiEXLab: For me the test case works fine, even when running it 10000 times.
The test basically checks if connecting to `wss://server.invalid` fails, which should always be the case because `.invalid` is a reserved invalid TLD unless you changed something in your host file.
The `ExecutionException` is not thrown by the method itself, it's thrown by `Future#get` to wrap the underlying IOException that is thrown inside the future. I update the documentation to clarify the future behaviour.

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
RoiEXLab: After my changes there was no longer an InterruptedException to be thrown so the "interrupted" and the "errored" test cases wer suddenly the same

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
RoiEXLab: The name is perhaps slightly misleading.
The idea is that tests that "require a close action"/run code that invokes sendClose at some point should call this method.

Suggestions for the name?

RoiEXLab: Nevermind, I'll just inline this function, it is only used once anyways

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
RoiEXLab: As I already mentioned, the JDK API won't invoke the listener again unless we call `webSocket.request(1 or higher)`.
From [the documentation](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/WebSocket.html):
>  A receive method is any of the onText, onBinary, onPing, onPong and onClose methods of Listener. WebSocket initiates a receive operation by invoking a receive method on the listener. The listener then must return a CompletionStage which completes once the operation has completed.
>
>To control receiving of messages, a WebSocket maintains an internal counter. This counter's value is a number of times the WebSocket has yet to invoke a receive method. While this counter is zero the WebSocket does not invoke receive methods. The counter is incremented by n when request(n) is called. The counter is decremented by one when the WebSocket invokes a receive method.

Also the [WebSocket.Listener documentation](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/WebSocket.Listener.html) states:
> The strict sequential order of invocations from WebSocket to Listener means, in particular, that the Listener's methods are treated as non-reentrant. This means that Listener implementations do not need to be concerned with possible recursion or the order in which they invoke WebSocket.request in relation to their processing logic.
>
>Careful attention may be required if a listener is associated with more than a single WebSocket. In this case invocations related to different instances of WebSocket may not be ordered and may even happen concurrently. 


So as long as we're not re-using a listener for multiple websockets we are fine

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
RoiEXLab: I did consider it, but the current implementation is tied too much to the WebSocketConnection class.
I cleaned up the test cases a bit though

RoiEXLab:
@DanVanAtta Updated, let me know what you think

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -153,23 +153,27 @@ void close() {
...
-              if (webSocket != null && throwable == null) {
-                pingSender.start();
-              } else {
-                errorHandler.accept("Failed to connect to: " + serverUri);
-              }
+        .thenRun(pingSender::start)
+        .exceptionally(
+            throwable -> {
+              errorHandler.accept("Failed to connect to: " + serverUri);
+              log.log(
```
DanVanAtta: The error handler should be taking care of any logging that is needed, whether that is a message to a user or a message to the logs.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -153,23 +153,27 @@ void close() {
...
-   *     {@link WebSocket.Builder#buildAsync(URI, Listener)} in case of an error.
    */
-  CompletableFuture<WebSocket> connect(
+  @SuppressWarnings("FutureReturnValueIgnored")
+  void connect(final WebSocketConnectionListener listener, final Consumer<String> errorHandler) {
+    connectInternal(listener, errorHandler);
+  }
+
+  @VisibleForTesting
+  CompletableFuture<Void> connectInternal(
```
DanVanAtta: Why are we returning a value here, just for testing maybe?

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
```diff
@@ -104,20 +142,27 @@ void tearDown() {
     @Test
     @DisplayName("Verify connect initiates connection and starts the pinger")
     void connectWillInitiateConnection() throws Exception {
-      givenWebSocketConnects(true);
-
-      final boolean connected =
-          webSocketConnection.connect(webSocketConnectionListener, errorHandler).get();
+      final HttpClient httpClient = mockHttpClient();
```
DanVanAtta: Why `mockHttpClient()` instead of using `@Mock HttpClient httpClient`?

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -49,17 +57,21 @@
...
+  private WebSocket client;
+
   @Getter(
       value = AccessLevel.PACKAGE,
       onMethod_ = {@VisibleForTesting})
-  @Setter(
-      value = AccessLevel.PACKAGE,
-      onMethod_ = {@VisibleForTesting})
-  private WebSocketClient client;
+  private final WebSocket.Listener webSocketListener = new WebSocketListener();
```
DanVanAtta: These two variables are named a bit too similarly, it's a bit difficult to keep them straight:
```
  private WebSocketConnectionListener listener;
  private final WebSocket.Listener webSocketListener = new WebSocketListener();
```

Can they be renamed reasonably or restructured perhaps?

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -68,52 +80,28 @@
...
+                        .exceptionally(logWebSocketError(Level.INFO, "Failed to send ping."));
                   }
                 });
   }
 
   /** Does an async close of the current websocket connection. */
   void close() {
     closed = true;
-    client.close();
+    if (client != null && !client.isOutputClosed()) {
```
DanVanAtta: Is it very burdensome to keep client non-nullable? Also wondering why it's nullable now.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -173,8 +156,63 @@ void sendMessage(final String message) {
...
+                    .exceptionally(logWebSocketError(Level.SEVERE, "Failed to send queued text.")));
+        queuedMessages.clear();
+      }
+      webSocket.request(1);
+    }
+
+    @Override
+    public CompletionStage<?> onText(
+        final WebSocket webSocket, final CharSequence data, final boolean last) {
+      textAccumulator.append(data);
```
DanVanAtta: It takes a lot of convincing that this is threadsafe. Messages coming in will be concurrent, ie: multiple chatters to server and those being handled on different threads then sending back messages concurrently to this listener. 

Perhaps a documentation reference and some commentary on why we are guaranteed to be threadsafe here would be useful for the future. A naive maintainer might look at this and decide we *must* have a sync lock or start to think the implementation is broken. It seems like an obvious problem if two 'onText' events come in at the same time and they are both not `last`, then we'll mix the messages together.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -173,8 +156,63 @@ void sendMessage(final String message) {
...
+      if (last) {
+        listener.messageReceived(textAccumulator.toString());
+        textAccumulator.setLength(0);
+      }
+      webSocket.request(1);
```
DanVanAtta: This line perhaps could use a comment, why are we requesting to the websocket, and why `1`?

DanVanAtta:
Want to be clear in the asks:
- more documentation
- double check automated testing is really thorough
- do some manual testing, join a local lobby, send chat messages and run the integration tests, you should see an some interaction in your local lobby with other clients connecting

If you're fully confident this is bomb-proof, then the above should probably do it.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
RoiEXLab: I'm actually not sure about this, but when testing the customHost thingy I got an NPE in this line, presumably because a connection never got opened (maybe concurrency related) and the connection was closed before that ever happened

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
RoiEXLab: I'm a little bit concerned we might lose the stacktrace in case an actual unintentional error occurs, thoughts?

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
RoiEXLab: Exactly, I wasn't sure how I could wait for the future to complete otherwise

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
RoiEXLab: 1. If that helps the code is almost 1:1 taken from an example implementation from the documentation: https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/WebSocket.html#request(long) so this is the way it's meant to be used.
2. I haven't actually thought of that and I see how it bothers you. I'll add some documentation which should hopefully clarify that.

RoiEXLab:
> double check automated testing is really thorough

Actually your smoke tests were useful for this PR, they caught an obvious error I had missed which prevented the headless launcher from starting, but yeah there's definitely room for improvement

http-clients/src/test/java/org/triplea/http/client/web/socket/WebSocketConnectionTest.java
RoiEXLab: AFAIK the mockito strict stups policy prevents declaring mocks that might not be used by all tests.
I'll create a nested test case to clean that up a bit

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
RoiEXLab: I renamed them to make the difference a little bit clearer.

RoiEXLab:
@DanVanAtta Updated, I did some manual testing with 4 local clients all writing messages, but unfortunately I wasn't able to get a local lobby up and running currently, my local docker installation doesn't like my virtualbox installation right now.
I should be able to launch docker inside my linux VM though, but this will take some time to setup as well.

> Are you feeling that this implementation is comletely bomb-proof? This is code I hope we never really need to maintain/revisit.

I'm pretty confident about the implementation. The added complexity actually adds to that confidence. With the usual listener patters you're alway unsure how they're actually called and implemented as a whole, but since it's up to you to "control" the call flow I'm actually pretty confident about this.
An example of this I had at work a while back: We were implementing some sort of periodic query service using some library I forgot. The idea was simple: Simply annotate a method and set a period and if it should be executed asynchronously or synchronously.
We wanted synchronous execution because it made things simpler, however I wondered what would happen if the method would take longer to execute than the time interval in between, would it skip some invocations, or just pile up a huge amount of tasks?
Turned out it was the latter, so we had to keep our custom logic because with busy traffic the possibility was real and could've actually happened.

For the maintenance part: I don't think we'll have to touch that part any time soon. Obviously code never ends up being "perfect", so it could always be the case that we want to see the tests refactored or something similar, but unless we want to send raw binary elements over websocket any time soon I don't think we'll need to touch this code again in the near future

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
RoiEXLab: I just tried to debug this issue and it's gone. Seems like the issue really was that the initial connection took too long, so that the connection was closed before it even really opened.

I'll add a comment clarifying that

RoiEXLab:
@DanVanAtta I was able to get a local lobby running inside my Ubuntu VM, and ran the integration tests that worked like a charm.

However I ran into some issues following the [instructions on the wiki](https://github.com/triplea-game/triplea/wiki/Local-Build-and-Testing).
 The postgres docker image startet and immediatly exited.
`docker logs` revealed this error:
`Error: Database is uninitialized and superuser password is not specified.` with some possible solutions.
I ended up running
```bash
docker run -d --env POSTGRES_HOST_AUTH_METHOD=trust --name=triplea-lobby-db -p 5432:5432 triplea/lobby-db
```
to allow all connections which worked fine for testing, but doesn't seem like a clean solution.

A couple of things I noticed:
`./launch_db` only works if the docker image was already built, which isn't the case if you have a clean docker installation, but is presented as a "all-in-one solution".
Also `./load_sample_data` assumes that `psql` is installed, which isn't always the case if you're just installing postgres via a docker container. Ideally this command would run inside the container.

DanVanAtta:
Thanks for the follow-up @RoiEXLab 

Re: `./launch_db`

- the build step is pretty slow, we probably could use better error checking to make sure that an image exists before trying to launch it.

Re: `./load_sample_data`

A check for psql would be good to give a clear error message and fail early. Ultimately we probably want a 'install-docker' or some sort of script similar to that. IMO that would be really good, arguably necessary, we should have it so that any dev can follow a README and/or execute some scripts from any environment to get up and running. Having to 'figure out' how to launch locally is tribal knowledge and just really annoying/time consuming of a process. It is very determental when local dev setup is not documented and is just something you should of course, just know how to do..

DanVanAtta:
In short, feel free to add better error checking to the DB scripts noted, add scripts that work for you, and do please update the README so that any future dev does not have to re-invent the local dev setup wheel.

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
DanVanAtta: I'm touching code close to this, can see about following up.

The 'severe' logging shows a pop-up to headed game clients, the same for the error handler. We'd wind up showing two error messages to a user along with any others. The stack trace is of some value, but only if a user is reporting the problem via error reporter (stack traces are not useful to users).

If we want to control whether we have a stack trace or not, the error handler perhaps should accept a nullable throwable.

DanVanAtta:
Just occurred to me.. @RoiEXLab would you mind double checking error handling?
- try to a connect to a lobby that DNE, verify you get a solid error message that probably should not allow for the error to be reported
- connect to lobby, then turn off the lobby server, verify we do not get a flood of error messages, probably none should offer the option to report the error.

In all cases the error message should be quite clear what happened and not cryptic "websocket ioexception occurred", but something like "Connection to lobby was lost." or "Connection to lobby could not be created"

RoiEXLab:
> try to a connect to a lobby that DNE, verify you get a solid error message that probably should not allow for the error to be reported

If the lobby doesn't exist in the first place (and it isn't just a non-working websocket) I get the usual error message that the post request failed.
However if I hardcode the URI to something invalid I get a stacktrace, but no error popup. This might need improvement.

> connect to lobby, then turn off the lobby server, verify we do not get a flood of error messages, probably none should offer the option to report the error.

This is a weird one. Closing the lobby simply kicked me out of the lobby screen, but without any message whatsoever

Issue #5654, opened 2019-12-04
AI and/or engine code is dominated by GameMap.getNeighborsValidatingCanals()


asvitkine:
AI and/or engine code is dominated by GameMap.getNeighborsValidatingCanals().

This is what shows up at the leaf of most function calls with significant CPU cost. Example:

![Screen Shot 2019-12-03 at 11 37 36 PM](https://user-images.githubusercontent.com/17648/70113116-f0eebc00-1625-11ea-9c76-2650bc2336cd.png)

We should investigate if this can be optimized. In particular, the underlying code uses streams API and lots of abstraction, which might be nice for readability but may not be great for performance.

At the very leafs, we also see the code querying attachments, which might be O(n) operations:

![Screen Shot 2019-12-03 at 11 42 32 PM](https://user-images.githubusercontent.com/17648/70113277-9f92fc80-1626-11ea-9979-6790adbf0e2f.png)

Perhaps there's some caching that can be done so that all attachments don't need to be scanned to find canals, for example.

DanVanAtta:
Good work pinpointing this @asvitkine, it's great research for focusing our efforts. With that said, I'll note this fun saying:
> It's better to optimize correct code than it is to correct optmized code.

So we should tread carefully, perhaps ideally starting by adding a test harnass before we engage in a lot of performance tuning. Overall the O(n) algorithm and lack of caching seem to be really good targets for better perf. Again, good research, thank you!  :+1:

RoiEXLab:
IIRC this code is used by the route finding algorithm as well.
Some time ago parts of this were "parallelized" using parallel streams which increased performance by a noticeable amount. This might explain the heavy CPU usage because it uses a lot of CPU cores.

Regardless I think caching values might help a lot here, I'm pretty sure that (without looking at the code) most of the time the values don't change at all. And in this case the overhead of streams shouldn't matter at all.

ron-murhammer:
As @RoiEXLab points out this is used for route finding and there are some comments I believe in the code pointing out that performance of that matters. This fires a lot in the AI as it calculates various unit move options and for the player every time they select units and move the cursor around.

DanVanAtta:
Moved this to a wiki page for more generalized tracking/discussion: https://github.com/triplea-game/triplea/wiki/Performance-Improvements

asvitkine:
> Moved this to a wiki page for more generalized tracking/discussion: https://github.com/triplea-game/triplea/wiki/Performance-Improvements

[...]

PR #6086, opened 2020-03-24
Make findNearbyEnemyCapitalsAndFactories() ~30x faster.

asvitkine:
On great war map with all AIs, after 1700 executions, the new version takes 152ms in aggregate while the previous implementation takes 4900ms.

The algorithm's functionality is kept the same with the exception of removing the max distance constant, which was only there as a performance optimization. It's not needed with the new algorithm.

 

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix: #5654
[] Other:   

## Testing

Tested by having a wrapper function that executed both the old and the new version of the function and compared that their outputs were the same and tracked their their total execution.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
@@ -1,12 +1,14 @@
 package games.strategy.triplea.ai.pro.util;
 
 import games.strategy.engine.data.GameData;
+import games.strategy.engine.data.GameMap;
 import games.strategy.engine.data.GamePlayer;
 import games.strategy.engine.data.Route;
 import games.strategy.engine.data.Territory;
 import games.strategy.triplea.ai.pro.ProData;
 import games.strategy.triplea.attachments.TerritoryAttachment;
 import games.strategy.triplea.delegate.Matches;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -15,13 +17,13 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import org.triplea.java.ObjectUtils;
 import org.triplea.java.collections.CollectionUtils;
 
 /** Pro AI battle utilities. */
 public final class ProTerritoryValueUtils {
 
   protected static final int MIN_FACTORY_CHECK_DISTANCE = 9;
-  private static final int MAX_FACTORY_CHECK_DISTANCE = 30;
 
   private ProTerritoryValueUtils() {}
 
@@ -430,18 +432,55 @@ public final class ProTerritoryValueUtils {
     return capitalOrFactoryValue / 100 + nearbyLandValue / 10;
   }
 
+  /**
+   * Finds enemy capitals / factors from a list that are "nearby" a given territory.
+   *
+   * <p>If any of the target territories exist within a distance of 9, returns the subset that do.
+   * Otherwise, proceeds to check the territories at each subsequent distance until at least one
+   * capital is found.
+   *
+   * <p>Note: This is an optimized version of a previous, much slower algorithm that has been
+   * designed to keep the original behavior, but tuned for speed.
+   *
+   * @param startTerritory The territory from where to start the search.
+   * @param enemyCapitalsAndFactories The territories to search for.
+   * @return Subset of enemyCapitalsAndFactories that were found.
+   */
   protected static Collection<Territory> findNearbyEnemyCapitalsAndFactories(
-      final Territory t, final Set<Territory> enemyCapitalsAndFactories) {
+      final Territory startTerritory, final Set<Territory> enemyCapitalsAndFactories) {
+    final GameMap map = startTerritory.getData().getMap();
+    // Use breadth first search to traverse territories, keeping track of which have already been
+    // visited and which territories from the target list have  been found.
+    final var found = new HashSet<Territory>();
+    final var visited = new HashSet<Territory>(List.of(startTerritory));
+    final var territoriesToCheck = new ArrayDeque<Territory>(List.of(startTerritory));
 
-    Set<Territory> nearbyEnemyCapitalsAndFactories = new HashSet<>();
-    for (int i = MIN_FACTORY_CHECK_DISTANCE; i <= MAX_FACTORY_CHECK_DISTANCE; i++) {
-      nearbyEnemyCapitalsAndFactories = t.getData().getMap().getNeighbors(t, i);
-      nearbyEnemyCapitalsAndFactories.retainAll(enemyCapitalsAndFactories);
-      if (!nearbyEnemyCapitalsAndFactories.isEmpty()) {
-        break;
+    // Since we process territories in order of distance, we can keep track of the last territory
+    // at the current distance that's in the territoriesToCheck queue. When we encounter it, we
+    // increment the distance and update lastTerritoryAtCurrentDistance.
+    int distance = 0;
+    Territory lastTerritoryAtCurrentDistance = startTerritory;
+    while (!territoriesToCheck.isEmpty()) {
+      final Territory node = territoriesToCheck.removeFirst();
+      for (final Territory neighbor : map.getNeighbors(node)) {
+        if (visited.add(neighbor)) {
+          territoriesToCheck.add(neighbor);
+          if (enemyCapitalsAndFactories.contains(neighbor)) {
+            found.add(neighbor);
+          }
+        }
+      }
+
+      // If we just processed the last territory at the current distance, increment the distance
+      // and set the territory at which we need to update it again to be the last one added.
+      if (ObjectUtils.referenceEquals(node, lastTerritoryAtCurrentDistance)) {
+        distance++;
+        if (distance >= MIN_FACTORY_CHECK_DISTANCE && !found.isEmpty()) {
+          break;
+        }
+        lastTerritoryAtCurrentDistance = territoriesToCheck.peekLast();
       }
     }
-
-    return nearbyEnemyCapitalsAndFactories;
+    return found;
   }
 }


asvitkine:
Re: The codeclimate warnings, I am tempted to leave as-is. The alternative is to have the logic split into multiple helpers, which I'm not sure would help readability in this case - would just result in more code and having to jump to look at more methods to understand the flow.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
```diff
@@ -429,18 +431,41 @@ private static double findWaterValue(
...
+            if (found.size() == enemyCapitalsAndFactories.size()) {
+              break;
+            }
+          }
+        }
       }
-    }
 
-    return nearbyEnemyCapitalsAndFactories;
+      if (node == nextDistanceAfter) {
```
RoiEXLab: Even if it saves some redundant overhead, it is good practice to use equals over reference equality.
Have a look at the errorprone explanation for this "warning"/error: https://github.com/google/error-prone/blob/master/docs/bugpattern/ReferenceEquality.md

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
```diff
@@ -429,18 +431,41 @@ private static double findWaterValue(
...
+        distance++;
```
RoiEXLab: I undertand how this distance is counted, pretty clever way to do it efficiently, however a comment explaining it would be nice in case we ever look at this code again (which will likely happen)

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
```diff
@@ -429,18 +431,41 @@ private static double findWaterValue(
...
   }
 
+  // 1700 oldT=4906 newT=227 newT2=166 newT3=152
+  @SuppressWarnings("ReferenceEquality")
   private static Set<Territory> findNearbyEnemyCapitalsAndFactories(
-      final Territory t, final Map<Territory, Double> enemyCapitalsAndFactoriesMap) {
-
-    Set<Territory> nearbyEnemyCapitalsAndFactories = new HashSet<>();
-    for (int i = MIN_FACTORY_CHECK_DISTANCE; i <= MAX_FACTORY_CHECK_DISTANCE; i++) {
-      nearbyEnemyCapitalsAndFactories = t.getData().getMap().getNeighbors(t, i);
```
RoiEXLab: I just had a look at the implementation of the getNeighbours method with a distance and it is bad but still used.
I'd really appreciate if you could have a look at that one as well in the near future.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
```diff
@@ -429,18 +431,41 @@ private static double findWaterValue(
...
-      nearbyEnemyCapitalsAndFactories.retainAll(enemyCapitalsAndFactoriesMap.keySet());
-      if (!nearbyEnemyCapitalsAndFactories.isEmpty()) {
-        break;
+      final Territory t, final Set<Territory> enemyCapitalsAndFactories) {
+    final GameMap map = t.getData().getMap();
+
+    int distance = 0;
+    Territory nextDistanceAfter = t;
+
+    final var found = new HashSet<Territory>();
```
DanVanAtta: IDK, I'd tend to agree with code climate on the length and complexity warning. The trick is not just to break up a method but to create the right abstractions. EG:

```
while (!queue.isEmpty()) {
   found.addAll(findAtDistance(distance, .... ));
   distance ++;
   if (distance > MAX_FACTORY_CHECK_DISTANCE
        || (distance >= MIN_FACTORY_CHECK_DISTANCE && !found.isEmpty())) {
       break;
    }
    nextDistanceAfter = queue.getLast();
}
return found;
```

I don't know if that quite does it, it's an art to find something that works right and does not just create indirection.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
```diff
@@ -429,18 +431,41 @@ private static double findWaterValue(
...
   private static Set<Territory> findNearbyEnemyCapitalsAndFactories(
-      final Territory t, final Map<Territory, Double> enemyCapitalsAndFactoriesMap) {
-
-    Set<Territory> nearbyEnemyCapitalsAndFactories = new HashSet<>();
-    for (int i = MIN_FACTORY_CHECK_DISTANCE; i <= MAX_FACTORY_CHECK_DISTANCE; i++) {
-      nearbyEnemyCapitalsAndFactories = t.getData().getMap().getNeighbors(t, i);
-      nearbyEnemyCapitalsAndFactories.retainAll(enemyCapitalsAndFactoriesMap.keySet());
-      if (!nearbyEnemyCapitalsAndFactories.isEmpty()) {
-        break;
+      final Territory t, final Set<Territory> enemyCapitalsAndFactories) {
```
DanVanAtta: Is `t` equivalent to the `startTerritory`, the 'basis' for the search? I'm wondering if the algorithm perhaps would be a bit clearer if 't' and 't2' were spelled out a bit more.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
```diff
@@ -429,18 +431,41 @@ private static double findWaterValue(
     return capitalOrFactoryValue / 100 + nearbyLandValue / 10;
   }
 
+  // 1700 oldT=4906 newT=227 newT2=166 newT3=152
```
DanVanAtta: Is this comment intended to be included?

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
asvitkine: Done.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
asvitkine: No, removed. Thanks!

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
asvitkine: Added a comment, thanks!

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
asvitkine: The logic here is intended to compute distance without significant extra overhead. I'd have to check how much equals() adds. If it's non-trivial, might be better to just track distance explicitly (by having a Node type that includes a Territory and a distance member).

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
asvitkine: I don't think that helps, findAtDistance() would require too many params which codeclimate will complain about.

The only way I see of "working around it" is to make a helper class and move all these variables to be members. Then, helper methods don't need to take these params because they'd be explicitly taking them as part of being on the object, so codeclimate would not complain.

I'm not sure this is something we should be optimizing for. The result would be instead of a 30-line method we'd have a 50-75 line class. Is that really a net win?

asvitkine: Put another way, would we be happy if every 30+ line method instead becomes its own class with a constructor and 2-5 other methods? Do we think that would be a better state?

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
asvitkine: I checked with .equals() and the result is a 20% slowdown. :(

ssoloff: We added `ObjectUtils#referenceEquals()` some time ago to act as documentation for the case when one _really_ wants to use reference equality.  The Javadocs for this method state:

```java
* <p>Use this method only when you really need to compare object references for equality. In
* almost all cases you should be using {@link Object#equals(Object)}. <strong>Using this method
* documents your intention that a reference equality check is required.</strong>
```

Its usage should prevent ErrorProne from triggering on this line.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
ron-murhammer: I'd say the complexity is ok here given this is a performance sensitive method. If anything I'd rather focus on adding some comments describing how it works as no matter how you chop it up it is going to be a bit complex.

DanVanAtta: I don't think that is necessarily the trade-off or decision. There are other ways..

DanVanAtta: > I don't think that helps, findAtDistance() would require too many params which codeclimate will complain about.

[Introduce Parameter Object](https://refactoring.com/catalog/introduceParameterObject.html) refactoring would solve that.

The big point is it's not necessarily about creating indirection, but abstractions and information hiding.

DanVanAtta:
Agree with @RoiEXLab for characterization test.

> However I'd really like to see a characterization test to be sure the implementations do the same., so the test passes with the old code as well as with the new code.


@asvitkine thoughts on adding that?

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
DanVanAtta: To expound slightly,  a data object with `@Builder` makes for a natural parameter object.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
```diff
@@ -429,18 +431,41 @@ private static double findWaterValue(
...
-    for (int i = MIN_FACTORY_CHECK_DISTANCE; i <= MAX_FACTORY_CHECK_DISTANCE; i++) {
-      nearbyEnemyCapitalsAndFactories = t.getData().getMap().getNeighbors(t, i);
-      nearbyEnemyCapitalsAndFactories.retainAll(enemyCapitalsAndFactoriesMap.keySet());
-      if (!nearbyEnemyCapitalsAndFactories.isEmpty()) {
-        break;
+      final Territory t, final Set<Territory> enemyCapitalsAndFactories) {
+    final GameMap map = t.getData().getMap();
+
+    int distance = 0;
+    Territory nextDistanceAfter = t;
```
DanVanAtta: Sorry to pile on, `nextDistanceAfter` is not very clear to me. I would almost expect the value to be a scalar, similar to distance.  Can we brainstorm some potential better variable names?

- `territoryAtNextDistanceToCheck`?

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
```diff
@@ -429,18 +431,41 @@ private static double findWaterValue(
...
+
+    final var found = new HashSet<Territory>();
+    final var visited = new HashSet<Territory>(List.of(t));
+    final var queue = new ArrayDeque<Territory>(List.of(t));
+    while (!queue.isEmpty()) {
```
DanVanAtta: nit, `queue` could perhaps be more descriptive as well. Is this the queue of territories to check? `territoriesToCheck` perhaps as a better name? It would make this boolean conditional perhaps a bit more clear.

asvitkine:
> Agree with @RoiEXLab for characterization test.
> 
> > However I'd really like to see a characterization test to be sure the implementations do the same., so the test passes with the old code as well as with the new code.
> 
> @asvitkine thoughts on adding that?

Looking at that.  Let me add the test in a separate PR that can be merged before this one so that it's clear that it works with the existing algorithm.

DanVanAtta:
I think the risk of merge conflict on this branch is likely low; sounds good @asvitkine  :+1:

asvitkine:
Sent #6092.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
asvitkine: Done.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
asvitkine: Went with lastTerritoryAtCurrentDistance.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
asvitkine: Thanks, used that now.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
asvitkine: One simplification I made is removing MAX_FACTORY_CHECK_DISTANCE.

This is technically a change from the previous implementation, but in practice I think the only reason the previous implementation had it is because such a condition was needed to make sure it terminated (the old algo just expanded distance and kept trying to find things).

It's not needed in this implementation and just adds extra complexity. In practice, 30 away is very far. A small map like revised seems to have every territory less than 9 territories away from any other. A larger map like Big World still is well within the 30 limit.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
DanVanAtta: Awesome, that is easier, thank you

ron-murhammer:
@RoiEXLab Correct, its defining the "nearby" range which the AI is looking for enemy capitals and factories. It was the easiest temporary solution to the performance challenges of the method. In theory, if the method is made fast enough, you could allow it to search the whole map or just have some larger max search distance.

ron-murhammer:
@asvitkine One comment on this. I realized you removed the max distance. I'm not sure if that could cause issues on very large maps that don't have factories? I don't think you'd get an endless loop since it tracks all the visited territories but it could search the entire map of thousands of territories. I guess the assumption is even if it does then its performance would be ok?

asvitkine:
> @asvitkine One comment on this. I realized you removed the max distance. I'm not sure if that could cause issues on very large maps that don't have factories? I don't think you'd get an endless loop since it tracks all the visited territories but it could search the entire map of thousands of territories. I guess the assumption is even if it does then its performance would be ok?

Yes, that's right. The new algorithm is fast and only visits each territory once, so it should be absolutely fine if it doesn't find anything. I'll update the PR description to mention it.

asvitkine:
Now rebased on top of my other PR.

PR #6091, opened 2020-03-26
Inline server side factory objects

DanVanAtta:
Refactor of http-server objects and how they are built.

To simplify and hide dependencies, delete single method factory classes
and move factory methods to the class that is being built. Second, avoid
building nested dependencies and add a factory build method to dependencies.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:


ron-murhammer:
@DanVanAtta Does this potentially have conflicts with the websocket PR? I see its touching some of the web socket classes.

http-server/src/main/java/org/triplea/http/ServerApplication.java
```diff
@@ -165,7 +160,7 @@ public void run(final AppConfig configuration, final Environment environment) {
         .getUserProperties()
         .put(PlayerConnectionWebSocket.GAME_LISTING_QUEUE_KEY, gameListingEventQueue);
 
-    final var messagingService = MessagingServiceFactory.build(jdbi, chatSessions, chatters);
+    final var messagingService = ChatMessagingService.build(jdbi, chatSessions, chatters);
```
ron-murhammer: Its not clear to me if the variable type changed here? Should the var name be updated to `chatMessagingService`?

http-server/src/main/java/org/triplea/modules/error/reporting/ErrorReportController.java
```diff
@@ -11,17 +12,38 @@
...
+  public static ErrorReportController build(final AppConfig configuration, final Jdbi jdbi) {
+    final GithubIssueClient githubIssueClient =
+        GithubIssueClient.builder()
+            .uri(AppConfig.GITHUB_WEB_SERVICE_API_URL)
+            .authToken(configuration.getGithubApiToken())
+            .githubOrg(AppConfig.GITHUB_ORG)
+            .githubRepo(configuration.getGithubRepo())
+            .build();
+
+    if (githubIssueClient.isTest()) {
```
ron-murhammer: Can you clarify what this block is for?

http-server/src/main/java/org/triplea/modules/game/ConnectivityController.java
```diff
@@ -26,6 +26,12 @@
 public class ConnectivityController extends HttpController {
   @Nonnull private final Predicate<InetSocketAddress> connectivityCheck;
 
+  public static ConnectivityController build() {
+    return ConnectivityController.builder() //
```
ron-murhammer: Is the `//` for formatting reasons?

DanVanAtta:
> @DanVanAtta Does this potentially have conflicts with the websocket PR? I see its touching some of the web socket classes.

No, but it conflicts with https://github.com/triplea-game/triplea/pull/6096, and that does conflict with the websocket technology change. As long as we keep the same abstraction  in GenericWebsocketListener, we'll be able to integrate the two.

ron-murhammer:
@DanVanAtta Ok. Well I think you and @RoiEXLab need to decide how to move forward on those then (I'm fine being the deciding vote if need be). If this doesn't conflict with either of them then feel free to merge it.

http-server/src/main/java/org/triplea/http/ServerApplication.java
DanVanAtta: It's not a super critical update as `ChatMessagingService` is still a 'messagingService'. `MessagingServiceFactory` was renamed as it was too generic, it really was 'chat' messages, and *not* things like network communication, internal events, queues, etc..

Let's keep an eye on this for the branch in #6096, updating this here would conflict with that, that is my only hesitation to not update it now.

The main thrust of this PR is that we are going from factory classes, to a factory method.

http-server/src/main/java/org/triplea/modules/error/reporting/ErrorReportController.java
DanVanAtta: It is moved code. I can't find where @RoiEXLab had a similar question, that answer would have been useful here.

In short it's a validation check of configuration to ensure we do not have an 'issue client' set to a 'test' mode with a prod configuration set. 

The `isTest` check is defined as:

```
  public boolean isTest() {
    return authToken.equalsIgnoreCase("test");
  }
```

In short we get a stubbing behavior when the authtoken used by the server to create github issues is set to 'test'.

http-server/src/main/java/org/triplea/modules/game/ConnectivityController.java
DanVanAtta: Indeed

PR #6092, opened 2020-03-26
Add test coverage for findNearbyEnemyCapitalsAndFactories.

asvitkine:
In preparation for a change in the implementation (in #6086), this adds test coverage for findNearbyEnemyCapitalsAndFactories() and improves its API.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[X] Other:  Test coverage

## Testing
That's what this is. :)

### DIFF
--- a/game-core/src/test/java/games/strategy/triplea/ai/pro/util/ProTerritoryValuesUtilsTest.java
+++ b/game-core/src/test/java/games/strategy/triplea/ai/pro/util/ProTerritoryValuesUtilsTest.java
@@ -1,0 +1,49 @@
+package games.strategy.triplea.ai.pro.util;
+
+import static games.strategy.triplea.delegate.GameDataTestUtil.territory;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsInAnyOrder;
+
+import games.strategy.engine.data.GameData;
+import games.strategy.engine.data.Territory;
+import games.strategy.triplea.xml.TestMapGameData;
+import java.util.Collection;
+import java.util.Set;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+public class ProTerritoryValuesUtilsTest {
+  private final GameData gameData = TestMapGameData.BIG_WORLD_1942.getGameData();
+  private final Territory swUsa = territory("Southwestern United States", gameData);
+  private final Territory england = territory("England", gameData);
+  private final Territory northJapan = territory("Northern Japan", gameData);
+  private final Territory moscow = territory("Moscow", gameData);
+  private final Territory newSouthWales = territory("New South Wales", gameData);
+
+  @Test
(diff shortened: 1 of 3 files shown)


game-core/src/test/java/games/strategy/triplea/ai/pro/util/ProTerritoryValuesUtilsTest.java
```diff
@@ -0,0 +1,49 @@
...
+public class ProTerritoryValuesUtilsTest {
+  private final GameData gameData = TestMapGameData.BIG_WORLD_1942.getGameData();
+  private final Territory swUsa = territory("Southwestern United States", gameData);
+  private final Territory england = territory("England", gameData);
+  private final Territory northJapan = territory("Northern Japan", gameData);
+  private final Territory moscow = territory("Moscow", gameData);
+  private final Territory newSouthWales = territory("New South Wales", gameData);
+
+  @Test
+  @DisplayName("all targets found when they're within distance of 9")
```
DanVanAtta: Looks good, just one question why distance '9'? Perhaps referencing the constant that controls this could make it clear?

IE: `when they're within min distance (ProTerritoryValuesUtils.MIN_FACTORY_CHECK_DISTANCE)`

Second, does `MAX_FACTORY_CHECK_DISTANCE` come into play anywhere? Is there a test case to be added where we don't find something that is outside of max distance?

asvitkine: I'm not covering the MAX_FACTORY_CHECK_DISTANCE case. I don't think we need to, I suspect it was there just to stop the algorithm, since otherwise it would just go indefinitely even when the map has no territories that are so far away.

In my other PR, I'm actually going to remove this check, since I don't think it's needed and allows us to simplify the code.

As for the magic number of 9, it would make the DisplayName not fit within the line length if we reference the constant. I thought we wanted this name to be short?

DanVanAtta: Ideally short, yes. More important to be clear, such that if the test ever breaks you can look at the test only and know why it was supposed to pass in the first place.

Perhaps you could use the 'assertThat' comment string to clarify, eg:

```
@DisplayName("all targets found when they're within min distance")
:
:
 assertThat(
    "swUsa and northJapan are within min distance: " + ProTerritoryValuesUtils.MIN_FACTORY_CHECK_DISTANCE,
   result, 
   containsInAnyOrder(swUsa, northJapan));
```

asvitkine: Done.

ron-murhammer:
@asvitkine This looks good to merge. Did you want to merge this PR first or the other one?

asvitkine:
Let's merge this first.

Issue #6037, opened 2020-03-11
Lobby exit leaves ghost chatter


DanVanAtta:
**How can the problem be recreated?**
- connect to a local lobby
- use 'file > exit'
- try to reconnect

Notice:
- get error message for name in use
- logging in with different name, previous name is listed as connected

**When the problem is fixed, what should have happened instead?**
- player should cleanly disconnect

**Which Engine Version are you using?** 
2.0 @ 49b4455cfe7187b06c6f300f11b01cc487fa07e7 (Mar 10, 2020)

**Notes**

Force killing the entire app removes the ghost chatters

DanVanAtta:
Fixed by: https://github.com/triplea-game/triplea/pull/6096

PR #6096, opened 2020-03-26
Websocket listeners

DanVanAtta:
This is a preview for a large update I have in the works (it will probably be too large to review).

## Overview

This update changes the websocket model of how we register listeners to handle messages on the client and server side. In short, instead of having groups of client side listeners with an enum for the message type to route messages, we'll instead have a proper message type object and register listeners one-by-one. Similarly for the server side, instead of having groups of event processors that procedurally switch over a message type and invoke a corresponding handler, we'll one-by-one register handlers (aka actions) by message type.

## Preview CR Objective

I'm opening this for potential early feedback:
 - I don't expect the final product to be very reviewable
 - better to get thoughts/opinions/improvements sooner rather than later.

## Goals for this update

The goals behind this update are:
- simplify websocket messaging model
- get it so we can handle 'session closed' events on server side. This way if a session is closed we can de-list a game or remove a player at that time. This would fix the 'ghost player' prerelease problem reported in: #6037. 

Also for example, after we moved from an http game polling model to websocket, we had a problem with 'ghost games' as well. This would fix that problem. For now we fix that problem by having the client explicitly notify the server of a game closed event, and failing that we kill games if they fail a keep-alive check. The issue is that when a session is closed for a listed game, we don't know which game that sessions corresponds to. So we can't close a game based on knowledge that a session has closed (the session is closing correctly, we just can't hook that into the proper game closed event, we are missing the link between session and game listing).

## Things to look at:

The client side is much more mature/baked, the server side I'm just starting to work on now.

### Client-Side

- New listener attachment pattern on the client side: 
  - https://github.com/triplea-game/triplea/pull/6096/files#diff-c69d4ca38981b4c559ee421d90dc8b0aR31
  - https://github.com/triplea-game/triplea/pull/6096/files#diff-15a83df1c9912b19760d54a63d90bb57R426
  - https://github.com/triplea-game/triplea/pull/6096/files#diff-f81fd321fafdadd4ebc6621243b9859dR49
- New mechanism for sending websocket messages (notably they are sent more directly now, simply call the 'send' method on websocket with a message, no longer need to go through a message factory):
  - https://github.com/triplea-game/triplea/pull/6096/files#diff-c3b1be9e35587056771b0707f28bbfa2R78
  - new mechanism for packaging, no longer have a factory but instead interface method: https://github.com/triplea-game/triplea/pull/6096/files#diff-0f3d511e45139de54a20587291e6a770R19
- "MessageListeners" deleted:
  - https://github.com/triplea-game/triplea/pull/6096/files#diff-bf1b19b5f3e872b5b36b4eaf24034611L3
- EnvelopeType enums deleted:
  - https://github.com/triplea-game/triplea/pull/6096/files#diff-254a7f8429782fa3a10bd6afcae73b78L5
- GenericWebsocket improved: https://github.com/triplea-game/triplea/pull/6096/files#diff-c2dcff4002a176fd301e661660272d6dR32
  - no longer need to create it with listeners, can attach listeners after creation
  - handles message routing to listeners
  - Server error is automatically handled: https://github.com/triplea-game/triplea/pull/6096/files#diff-c2dcff4002a176fd301e661660272d6dR72
- better interface and type safety enhancements:
  - Websocket interface: https://github.com/triplea-game/triplea/pull/6096/files#diff-ccf31ef7166ba143e5913b810ae1b665R7
  - MessageType class, enforces consistent type for listeners: https://github.com/triplea-game/triplea/pull/6096/files#diff-529cb331b4413f82f61fdf3752c17f2dR3
  - Messages have a type constant, eg:  https://github.com/triplea-game/triplea/pull/6096/files#diff-9829cc2bbf1d53f2c2e45b54e49585c7R14

### Server Side

- WebsocketMessageContext: https://github.com/triplea-game/triplea/pull/6096/files#diff-2443261337887ecd62307afbdeb6b4a5R10, this is passed to server-side listeners
- WebsocketMessagingBus, this will be wired to each of the websocket endpoints, we'll then use it to de-serialize JSON and route messages to listeners:
  - wiring to websocket: https://github.com/triplea-game/triplea/pull/6096/files#diff-0819d96a750721e8b9a6a6cb75022fd0R158

- ChatMessageReceived listener: 
   - Code: https://github.com/triplea-game/triplea/pull/6096/files#diff-a195d244539a7a4a64f4c5234a89f6b3R11
  - Registration: https://github.com/triplea-game/triplea/pull/6096/files#diff-47b676fc79f49ce6abfc7aa895858acbR100
  - Replaces: https://github.com/triplea-game/triplea/pull/6096/files#diff-1f76c806ffb6ee817dc6660fdf5c6c1cR80


DanVanAtta:
@ron-murhammer , @RoiEXLab any early feedback here? 
I'm interested if there are thoughts on the new messaging pattern and naming. If you guys think that the new pattern is scalable for replacing eventually all of the network communication between server and client (that is what we will be driving towards, websockets is the technology to do that and replace our hybrid RMI).

Might be easiest to give that feedback by checking this branch out and look over the new message types for 10 or 15 minutes and trace some of the callpaths.

RoiEXLab:
@DanVanAtta I started reviewing this, I should be able to comment on some things tomorrow, it's just a large diff

DanVanAtta:
Indeed and thanks @RoiEXLab 
I'm most interested in a strategic review. I don't think we can or even really should try to do a deep review of every line of code here to get everything perfect at a syntax level (to some extent, that is what codeclimate and our style conventions are for). It could all be easily mooted if we decide on a major design change.

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
```diff
@@ -19,74 +24,102 @@
...
   }
 
   @Override
   public void messageReceived(final String message) {
-    final ServerMessageEnvelope converted = gson.fromJson(message, ServerMessageEnvelope.class);
-    messageListener.accept(converted);
+    final MessageEnvelope converted = gson.fromJson(message, MessageEnvelope.class);
+
+    listeners.stream()
+        .filter(listener -> converted.messageTypeIs(listener.messageType))
```
RoiEXLab: Thoughts about using a MultiMap to get the `O(1)` runtime of this frequently used method?

game-core/src/main/java/games/strategy/engine/chat/LobbyChatTransmitter.java
```diff
@@ -21,30 +28,47 @@
...
+                message.getSlappingPlayer() + " slapped " + message.getSlappedPlayer());
+          }
+        });
+    playerToLobbyConnection.sendConnectToChatMessage();
   }
 
   @Override
   public Collection<ChatParticipant> connect() {
-    return playerToLobbyConnection.connect();
+    return List.of();
```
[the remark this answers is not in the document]
DanVanAtta: Certainly. Before we had more of a hybrid http and websocket model, sync and async. On the very first iteration we had a problem where if you initially connect to websocket, we would not know the chatter name. So what we did was the initial connection established a chatter, and we got the list of chatters back in a synchronized manner.

This is being updated so that we establish a websocket session first, but even with the websocket session you are not going to get any messages until you send a 'join' message. With that, the server will respond to 'join' messages with a websocket response containing the chatter list.

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
DanVanAtta: I did want a map, the listeners are of mixed type. If you can see how to solve that, it would make for cleaner logic probably. Otherwise, the mixed typing caused issues defining such a map.

FWIW, I don't think the iteration and lookup is necessarily a  problem. A recent saying I've come across that I found interesting, https://blog.cloudflare.com/when-bloom-filters-dont-bloom/, 
> "One colleague often says: "You can assume modern CPUs are infinitely fast. They run at infinite speed until they hit the memory wall"."

In this case, we are not going to have all listeners loaded from main memory, but still the point remains that a map is not necessarily going to be that much faster than sequential scanning on a pretty limited collection size.

DanVanAtta:
@RoiEXLab thanks for taking the look. I think we actually had as much type safety before, but a big difference is that listeners are no added one-by-one instead of in a group. Eventually this model should make it easier to convert some more of the messaging to websocket and tie chat users to their session (and games), and then when the session is closed we can trigger close and remove actions. Previously we could not do that as the association was not there due to the mix of sync and async.

http-clients/src/main/java/org/triplea/http/client/web/socket/GenericWebSocketClient.java
RoiEXLab: I agree that a map might not yield the desired performance benefits with this few listeners, but your answer about the types surprised me:
What would speak against code that looks something along the lines of this:
```java
private final Multimap<MessageType<?>, MessageListener> listeners = HashMultimap.create();

@Override
  public void messageReceived(final String message) {
    final MessageEnvelope converted = gson.fromJson(message, MessageEnvelope.class);

    listeners.get(converted.getMessageType()).forEach(listener ->
                listener.listener.accept(
                    converted.getPayload(listener.messageType.getPayloadType())));
```

DanVanAtta: Seems you'll lose type coherence, `getMessageType()` returns a string. `MessageListener` is also a generic type as well, it should match the generic type for `MessageType`
 
One also needs to account for the adding a listener as well under those constraints. The type safety we worked pretty hard for previously, clarity and correctness are important heuristics. We also need to think about system level issues, whether the incoming/outgoing message pattern is going to be clear and sustainable.

DanVanAtta:
@RoiEXLab @ron-murhammer 
I may have this building now. All that remains is add about 4 tests for the new server-side websocket message listeners. I went through and fixed all of the code climates that are reasonable to fix. To speak to the duplication warnings, they are not something that should be fixed. The patterns for websocket messages are just a bit boilerplate, abstracting things further is not really satisfying DRY but creating arguably unnecessary abstractions, ultimately I think it make it harder to understand. We also do want to keep messages to be pretty decoupled so we have flexibility for backward compatibility concerns.

With all that said, I think this is getting pretty dang close, some smoke testing looked good and the tests we have in place for http-server interaction are relatively comprehensive which gives confidence.

Once I've got the last few tests put in place, I'm planning to merge this. There are some things that could be broken up, but the bulk of the change is a near re-write/refresh of the messaging framework for websockets and I don't know how to do that incrementally in pieces that would still work.

DanVanAtta:
@ron-murhammer @RoiEXLab if you have time/desire, feel free to leave post-merge comments, I'll respond and/or address them in follow-ups.

game-core/src/main/java/games/strategy/engine/framework/startup/mc/SetupPanelModel.java
```diff
@@ -64,7 +64,9 @@ public ServerModel showServer() {
         this,
         ui,
         new HeadedLaunchAction(ui),
-        error -> SwingComponents.showError(null, "Connection problem", error));
+        error -> {
+          SwingComponents.showError(null, "Connection problem", error);
+        });
```
RoiEXLab, line 69: The curly braces are unecessary here

http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketProtocolSwapper.java
```diff
@@ -0,0 +1,13 @@
+package org.triplea.http.client.web.socket;
+
+import java.net.URI;
+import java.util.function.Function;
+
+class WebSocketProtocolSwapper implements Function<URI, URI> {
```
RoiEXLab, line 6: This class could even implement `UnaryOperator<URI>`
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/function/UnaryOperator.html

Issue #6023, opened 2020-03-09
PBEM & PBF Missing Help Texts


DanVanAtta:
**How can the problem be recreated?**
- launch game
- click PBEM
- notice there are no help buttons

**When the problem is fixed, what should have happened instead?**

Similar to version 1.9.0.0.7621, there should be a number of help boxes

### Dice Server Help Message

> Enter your your email address in the "To" field, and you opponents in in the "CC" field, you may enter multiple addresses each separated by a space. Visit https://dice.tripleawarclub.org to register your email addresses, this must be done to receive dice emails. If this is a league match you must enter a Game ID or Challenge ID, as requested by your tournament organizer

### Forum Poster Help Message

> Forum Poster
A forum poster can post the turn summary and save game to a forum at the end of each players turn. This allows two or more players to play a game where you don't have to be online at the same time.
The poster for each forum may be different so click help once you have selected a specific forum poster, to learn more.

### A&A.org Help Message

> Posts to www.AxisAndAllies.org
This poster is build for PBEM games via AxisAndAllies.org. Instructions:
Create a new Forum post in the Play Boardgames section http://www.axisandallies.org/forums/index.php?board=40.0
Copy the topic id from the URL displayed in the address bar of you browser. If the URL is http://www.axisandallies.org/forums/index.php?topic=25252.0, put this 5 digit number (25252) into the Topic ID field in TripleA
Put your login and password for the AxisAndAllies.org forum into the login and password fields
Click the Test Post button, to check that TripleA can reply to your forum post
Note: Your AxisAndAllies.org login and password are not stored as part of the save game, but they are stored encrypted in the local file system if you select the option to remember your credentials. You may have to enter your login and password again if you open the save game on another computer.

### forums.triplea-game.org

> Posts to forums.triplea-game.org
You can play PBEM/PBF games via forums.triplea-game.org. 
Instructions:
Create a new Forum post in the Play by Forum category https://forums.triplea-game.org/category/6/
Copy the topic id from the URL displayed in the address bar of you browser. If the URL is https://forums.triplea-game.org/topic/24/ put the topic number (24) into the Topic ID field in TripleA
Put your login and password for the forums.triplea-game.org forum into the login and password fields
Click the Test Post button, to check that TripleA can reply to your forum post
Note: Your forums.triplea-game.org login and password are not stored as part of the save game, but they are stored encrypted in the local file system if you select the option to remember your credentials. You may have to enter your login and password again if you open the save game on another computer.

[...]

PR #6109, opened 2020-03-29
Add Missing PBF/PBEM Help Messages

ron-murhammer:
Address #6023 

**Screenshots**
image

image

image


DanVanAtta:
Couple quick comments after demo'ing the updates, let me know what your thoughts are @ron-murhammer 

### Screenshot title
This screenshot compares old+new, couple things I note:
- line breaking would make the message a bit easier to read
- title of message box has changed "message" is a bit interesting of a title.
Screenshot from 2020-03-29 12-34-52

### Username vs login terminology
Nitpicking a bit this error message below:
- the messaging says "login" but the fields say "username", the two perhaps should be made consistent. 'Username' is the term used on the forum login page FWIW:
Screenshot from 2020-03-29 12-35-29

Screenshot from 2020-03-29 12-39-18

### PBEM, help text?

From a demo/ui perspective, I notice that the PBEM does not have a help text. Is that a follow-up change, or was it meant to be addressed here?
Screenshot from 2020-03-29 12-38-27
Screenshot from 2020-03-29 12-38-36

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/DiceServerEditor.java
```diff
@@ -36,6 +37,19 @@
               "<html>Opens email registration page to register with MARTI dice-roller.<br>"
                   + "Needs to be done once before MARTI dice server can be used.</html>")
           .build();
+  private final JButton helpButton =
+      new JButtonBuilder("Help")
+          .actionListener(
+              () ->
+                  JOptionPane.showMessageDialog(
```
DanVanAtta: There is a build for dialogs, it *might* help for setting the title. Something to consider.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/EmailSenderEditor.java
```diff
@@ -177,6 +187,82 @@ public void viewModelChanged(final EmailSenderEditorViewModel viewModel) {
           toggleFieldVisibility();
           emailProviderSelectionBox.setSelectedItem(viewModel.getSelectedProvider());
 
+          if (viewModel.isEmailProviderDisabled()) {
```
DanVanAtta: Design nit, the view model attempts to remove all logic from the view. This can be done by creating a method to fetch the help message directly from the view model, or secondarily to create a method to tell the view which help text to render. This way the view asks the view model which help text to render as opposed to the view calculating which help text to render.

EG:

```
 helpMessage.setText(viewModel.getEmailSenderHelpText());
```

Or:
```
switch(viewModel.getHelpTextToDisplay()) {
  case NO_EMAIL_PROVIDER_SELECTED:
      helpMessage.setText( ...);
      break;
   case GMAIL_EMAIL_PROVIDER:
     :
     : 
}
```

The above is different as currently view model is interpreting the meaning of `isEmailProviderDisabled` rather than being told directly what to render.

For a viewmodel pattern, the first variant of getting the help message is probably closest to the pattern. The switch statement with a dedicated flag condition would work too.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/EmailSenderEditorViewModel.java
```diff
@@ -16,6 +16,8 @@
 
 class EmailSenderEditorViewModel {
   static final String PROVIDER_DISABLED = "Disabled";
+  static final String GMAIL = "Gmail";
```
DanVanAtta: These values probably should not have been hardcoded to begin with. We do have constants for these values already though:
```
EmailProviderPreset.GMAIL.getName();
```

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbf/ForumPosterEditor.java
```diff
@@ -64,6 +73,43 @@
           SwingComponents.highlightLabelIfNotValid(viewModel.isForumUsernameValid(), usernameLabel);
           SwingComponents.highlightLabelIfNotValid(viewModel.isForumPasswordValid(), passwordLabel);
           testForum.setEnabled(viewModel.isTestForumPostButtonEnabled());
+          helpMessage.setText(
+              viewModel.getForumSelection().equals(NodeBbForumPoster.TRIPLEA_FORUM_DISPLAY_NAME)
```
DanVanAtta: Ditto on view model should be told what to render as opposed to computing what to render. This allows for the logic to be tested as well, you can append to the tests to verify that the right help message is provided under the correct conditions.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/EmailSenderEditorViewModel.java
ron-murhammer: Updated.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/DiceServerEditor.java
ron-murhammer: showMessageDialog actually has a method to specify title so just used that.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/EmailSenderEditor.java
ron-murhammer: Out of scope. Feel free to follow on adjusting it if you want.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbf/ForumPosterEditor.java
ron-murhammer: Updated to move the check to the view model.

ron-murhammer:
@DanVanAtta Updates made. I think you must have tested it before the PBEM help commit.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/EmailSenderEditor.java
DanVanAtta: I really don't think it is out of scope:
- adds something that we really need to do that should be done now.
- very much breaks pattern and consistency
- misses updates to test cases, adding more stuff to do later.

ViewModel is a pretty powerful pattern: https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93viewmodel, it was with quite some effort it was introduced to line up for the migration to javaFx where we would not need to update the views.

DanVanAtta:
@ron-murhammer please be sure to stay consistent with view-model pattern. It was not easy to introduce that, I do not want to have that lingering as another thing to do. We do not have many examples of view model, I would not want the example to become inconsistent, let alone needing fixing later.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/EmailSenderEditor.java
DanVanAtta: The update to get back to pattern is not very time consuming and changes the accessor methods added to the view model. The testing should be pretty quick as well, but we are talking 20 minutes of work. Opening up the view model as was done is breaking to the design pattern.

DanVanAtta:
Thinking about it for a bit more, adding tests is not gravy, it should be done:
- the view model is consistently testing. Losing coverage will be hard to detect and is not something anyone will want to come back and fix. Fixing that after the fact rather than now is intrinsically reduced ROI.
- our standards should not be different, I would have expected myself or anyone else to have updated the tests before calling this update done.


The two approaches of a direct flag or just passing the message to render will inform how the test is structured, whether you can just assert the right flag is returned or if you get the expected help message.

The original introduction of view model gave some useful reading links if you care about view-model:
- https://github.com/triplea-game/triplea/pull/5953#discussion_r381030037

As another link, M. Fowler actually walks through roughly exactly this kind of example of how to extract logic from a view and place it in a view-model: https://martinfowler.com/eaaDev/SeparatedPresentation.html

DanVanAtta:
Summarizing requested changes:
1. maintain view-model pattern to maintain pattern integrity and consistency
2. test the updates to view-model to maintain test coverage and consistency

DanVanAtta:
FWIW, at the bottom of: https://martinfowler.com/eaaDev/SeparatedPresentation.html, is a good example of logic extracted from the view.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/EmailSenderEditor.java
ron-murhammer: I really don't think there is much difference between your switch example and what is implemented. Either way the view has to interpret what to do based on either some form of enumeration/key like your example (getHelpTextToDisplay) or interpret whether to display certain text based on a few booleans. In the end, you always have to draw the line somewhere and try to keep the "business logic" out of the view but where that line is drawn can always be debated to some extent which can be shown by your 2 examples which take it to different degrees.

DanVanAtta: It's not that much different, perhaps there should be no objection then to fit squarely into the pattern?

The example we have here is really similar to: https://martinfowler.com/eaaDev/SeparatedPresentation.html

The 'boolean' parameter rather than an integer is less obvious of a parallel, but it's still state  the view interpreting state (logic) rather than being told what to do. In part, the pattern uses dedicated flags and switches to tell the view what to do. The same could be said about the different boolean parameters that control the various buttons. Many have the same logic condition behind them, yet the view model exposes those attributes as different boolean methods. In this way the view is totally unaware of the backend state, nor does it know that any state is coupled to anything else.

There are also two or three other helper methods for the 'isGmail' etc that could drop away if the view model were to be passed the text to be rendered.

DanVanAtta: If the view were to call: `getHelpTextToDisplay`, then it would have nothing to interpret. The switch is similar, but it's still different as one is inferring state, as opposed to being told what state to be in.

DanVanAtta: Like notice we could have done the same with:
```
    emailProviderSelectionBox.setSelectedItem(viewModel.getSelectedProvider());
```
That could have been a very similar 'if/else' block as well. It's not though to prevent the view from having to evaluate boolean conditions.  Let's do that consistently please.

DanVanAtta: > Either way the view has to interpret what to do based on either some form of enumeration/key like your example (getHelpTextToDisplay) or interpret whether to display certain text based on a few booleans

I don't think that fits. The enumeration/key would be dedicated and used in only one place in the view, and the enumeration would be provided to the view and not used internally by the view model. The enumeration/key would be a computed value, an output, it would not be an input used elsewhere. The output would be dedicated for the view component. In this example we are moving logic into the view.

It would also never be multiple booleans, the only boolean would be a single one. We really should have a string here to think about it, the state of the component is essentially a string. A view-model is the state of each component translated into a primitive. A button enabled = a single boolean, the text of an input field is a string. For a dynamic help text, it is a string. 

Think about it in this way, the view is pure presentation. We should be able to simply swap out the swing view for javafx and only need to update the component logic, the rest of the state logic should be able to stay the same. With javafx migration, we should be able to re-use exactly the view-model objects. It's argument for the help text to live in the view-model.

DanVanAtta: Piling on slightly, the switch and if statement are actually quite different.
The switch is iterating over a single value, it's a single value that controls the state of the UI. It is not the UI checking multiple parts of the backing state. 

A method to get the value of the text, is also a scalar. At the end of the day, the backend should give the front-end a single scalar to tell it how to render. Whether that scalar is a flag, or if it's the actual text, it's still a single value. It is not the view querying the view-model and doing logic.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/EmailSenderEditor.java
```diff
@@ -177,6 +190,82 @@ public void viewModelChanged(final EmailSenderEditorViewModel viewModel) {
...
+                "<html><p style='width: 400px;'>"
+                    + "Email sender<br/>"
+                    + "An email sender can email the turn summary and save game to multiple "
+                    + "recipients at the end of each players turn. This allows two or more "
+                    + "players to play a game where you don't have to be online at the same "
+                    + "time.<br/>"
+                    + "Each email sender may require custom configuration, to learn more "
+                    + "click help again after selecting a specific email sender."
+                    + "</p></html>");
+          } else if (viewModel.isEmailProviderGmail()) {
```
DanVanAtta: Building on the last thread, this is different from a switch statement on a single value.

```
        } else if (viewModel.isEmailProviderGmail()) {
```
That is a brand new boolean condition on the view model, the view should be able to know what to do based on a single value. In this case we have business logic in the view.

We should be able to get rid of all if statements and the method `isEmailProviderGmail`. The view should not be aware of a concept of 'gmail' or the different view providers, only that it gets data and it knows how to render it.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbf/ForumPosterEditorViewModel.java
```diff
@@ -116,6 +116,10 @@ synchronized boolean areFieldsValid() {
     return isTopicIdValid() && isForumUsernameValid() && isForumPasswordValid();
   }
 
+  boolean isForumProviderTripleA() {
```
DanVanAtta: This is perfect :+1:

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbf/ForumPosterEditor.java
```diff
@@ -64,6 +75,48 @@
           SwingComponents.highlightLabelIfNotValid(viewModel.isForumUsernameValid(), usernameLabel);
           SwingComponents.highlightLabelIfNotValid(viewModel.isForumPasswordValid(), passwordLabel);
           testForum.setEnabled(viewModel.isTestForumPostButtonEnabled());
+          helpMessage.setText(
+              viewModel.isForumProviderTripleA()
```
DanVanAtta: This is perfect :+1: 
The boolean is acting like a flag selecting between two values, looks good.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/EmailSenderEditorViewModel.java
```diff
@@ -100,10 +104,18 @@ boolean showEmailOptions() {
     return !isEmailProviderDisabled();
   }
 
-  private boolean isEmailProviderDisabled() {
+  protected boolean isEmailProviderDisabled() {
```
DanVanAtta: This method was meant to stay internal. I know that from the perspective as the author of this class. It's re-used as part of other computations, it is not intended to be used by the UI.

DanVanAtta: The genesis is this is actually just an extract method refactor, it's only here to avoid duplicate conditional checks elsewhere. A problem with exposing it to the UI is that it creates couplings between other logic checks where the value is de-duped. You couldn't change this value from `isEmailProviderDisabled` and impact only one UI control.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/EmailSenderEditorViewModel.java
```diff
@@ -100,10 +104,18 @@ boolean showEmailOptions() {
...
     return selectedProvider.equals(PROVIDER_DISABLED);
   }
 
+  protected boolean isEmailProviderGmail() {
+    return selectedProvider.equals(EmailProviderPreset.GMAIL.getName());
+  }
+
+  protected boolean isEmailProviderHotmail() {
```
DanVanAtta: These two methods use in combination with the `isEmailProviderDisabled` is where I see a problem. Combining these three into one method I would suggest is the way to go to avoid having any logic in the view and to fit very cleanly with the existing pattern.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/EmailSenderEditorViewModel.java
ron-murhammer: Reverted to private.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbem/EmailSenderEditorViewModel.java
ron-murhammer: Updated to a single method.

DanVanAtta:
Thank you @ron-murhammer

Issue #6048, opened 2020-03-14
Release 2.0 Sign-off List - Gathering Requirments


DanVanAtta:
Attempting to build a final sign-off list for 2.0 prerelease testing. Opening this issue to present an initial list and gather input for additional items that should be tested. Please add comments with additional items that should be included, the desired outcome is to have a mostly complete list of items that we should verify in the 2.0 prerelease.

-------------

### Lobby (L):

(L1) Chat:
L1.A) Chat messages are sent

(L2) Player listing:
L2.A) Moderators have (mod) tag
L2.B) Player listing right click menu shows a slap option
L3.C) Moderators only, shows a (working) disconnect and remove option in a right click menu
L4.D) Right clicking your own name does nothing

(L3) Slap:
L3.A) Slap messages are sent
L3.B) Slap message recipient sees "you have been slapped by X"
L3.C) Slap sender sees "you slapped X" 
L3.D) Everyone else sees "X slapped Y"

(L4) Banning:
L4.A) Banned players are removed from chat
L4.B) Banned players are removed from private host games
L4.C) Banned players are removed from bot games
L4.D) Games hosted by banned players are removed
L4.E) Banned players are unable to join lobby
L4.F) Banned players are unable to directly join hosted games
L4.G) Banned players are unable to directly join bot games
L4.H) On-ban, when a player is in lobby, a message is displayed to all users that a player was banned for violating lobby rules

(L5) Unbanning:
L5.A) Unbanning works, after unban player is able to join lobby and games

(L6) Username blocking:
L6.A) Username banning works
L6.B) What happens when we ban the username of a registered user?
L6.C) Users in lobby with a banned username are booted
L6.D) Users joining with a blocked name are unable to join with that name

(L7) Joining:
L7.A) registered users can join lobby multipe times and should see the same name used
L7.B) All players, including the joining player, should see "X has joined"
L7.C) joining with the same name that is already in lobby leads to a 'user has already joined' error message

(L8) Status Update
L8.A) Players can update their status, visible to everyone

(L9) Game Status
L9.A) Hosting a game posts game
L9.B) Changing game comments, player count changes, map changes, etc.. are reflected in lobby game status
L9.C) Launching game, stopping game, reflects properly in lobby as 'waiting', 'running', etc..
L9.D) Quitting a game delists the game from lobby

(L10) Player leaving lobby
L10.A) After leaving lobby a players name is removed from player list
L10.B) All players see a 'player has left' notification

(L11) Player disconnect
[...]

PR #6111, opened 2020-03-29
Fix connectivity check

RoiEXLab:
Fixes the connectivity issue mentioned in #6048 (presumably, haven't been able to test it so far should work though).

@DanVanAtta Have a look at the TODO comment in the test case I added.
For the test to work we need to register a lobby game, how would I do that inside this test case?

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:


http-server/src/test/java/org/triplea/modules/game/ConnectivityControllerTest.java
```diff
@@ -32,7 +34,9 @@ void checkConnectivityNegativeCase() {
   @Test
   void checkConnectivityPositiveCase() throws IOException {
     openSocket();
-    final boolean result = verifyEndpointReturningObject(client -> client.checkConnectivity(PORT));
+    // TODO Setup GameEntry so that the server correctly checks the connectivity.
```
DanVanAtta: The game entry and connectivity check happen effectively in parallel. There is no guarantee that the game entry will have been created. I think the IP to check needs to become an optional parameter to the connectivity check.

RoiEXLab: Well the main difference now is that we no longer pass an IP/port to this endpoint.
We just pass a GameID, which then checks the LobbyGame entry for this id and checks the endpoint passed there.

Maybe merging this endpoint with the game entry endpoint?

DanVanAtta: Hmm, I'd be concerned about a separation of concerns. It would be nice though to get that fast feedback when a game is posted that it cannot be accessed.

What do you think about creating a second connectivity check endpoint and invoke that one with a slightly different payload?

I didn't look at the full diff, @RoiEXLab , we're passing game id now? That might not be wise as there is no guarantee that create game *will* be invoked first, they happen on different threads.

DanVanAtta: @RoiEXLab having create game detect connectivity is pretty interesting. The error handling would be a bit more complex, as we can have multiple scenarios why a game could not be created. It certainly has the benefit though that a user does not wait the 5 or so seconds for the connectivity check to fire. It's worth investigating, and could lead to the removal of the connectivity check endpoint, which would be simpler.

DanVanAtta: @RoiEXLab thinking about it a bit more, another consideration - gameId is assigned by the serve when a game is created. You can't / don't pass a game id when creating a game. The server can also re-assign gameIds as well, or it is allowed to lose gameIds in which case a keep-alive will see a new gameId has been assigned.

FWIW, passing in IP address to a new connectivity check endpoint, or doing the check on game create seem like the two best options you may have.

RoiEXLab: Ok @DanVanAtta coming back at this PR I think I'll go with the merge post and verify step option. However I have one question:
How do I change `LobbyWatcherFeignClient#postGame(@HeaderMap Map<String, Object> headers, LobbyGame lobbyGame)` to return a more complex object than string to signalize that something went wrong with the verification?
Or should I just use a different HTTP status code for this?

DanVanAtta: Simply convert the return type to an object that can be converted to a JSON.

Could you walk me through briefly your reasons for re-using `postGame`?

It is a trade-off, FWIW this is the pro/con list I can think of:
Pro:
  - Can remove background thread that does connectivity check
  - Connectivity failures could become more responsive

Con:
  - There are at least two places where we post games. Both would need to handle the updated payload
  - The error code handling gets a bit tricky. First you need to be sure that the game id is returned null and you return an error code that indicates to the client the situation. This level of error checking / API contract is not the worst but it is intrinsically more complex
  - have to be careful of timing, the new game host server port may not yet be available until well after the 'post' call has completed. We'll need to find out if that was the reason for the connectivity check to have had the background thread and the very long delay before it is executed.
  - Break compatibility
  - Could maybe lead us down a pattern of having postGame return an even more complex return payload. Can be a slippery slope to violating  [interface segragation principle](https://en.wikipedia.org/wiki/Interface_segregation_principle).
  - Server side, every time a game is posted it'll always do a connectivity check, even if we do not want it. Clients reconnecting for example should not need the connectivity check. A callback from a 'postGame' endpoint could be considered an odd side effect.

Thinking about the timing issue now, and the last point of 'side-effects', and ease of update, I'd personally augment the connectivity check endpoint. I'm curious if you've different thoughts on this, if you see the pro list as much larger than I listed and perhaps the cons easier to solve than stated.

RoiEXLab:
@DanVanAtta I updated the code to perform the connectivity check inside the postGame request.
I haven't done any testing so far (I'm not even sure if the tests will pass, but everything should compile at least), but I'd like to hear your opinion on the code restructuring I had to do for this change.

I'm not 100% happy here and I'm sure you have a good suggestion on how to structure the code that handles a failed connectivity check.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/InGameLobbyWatcher.java
```diff
@@ -52,14 +53,16 @@ private InGameLobbyWatcher(
       final GameToLobbyConnection gameToLobbyConnection,
       final Consumer<String> errorReporter,
       final Consumer<String> reconnectionReporter,
-      @Nullable final InGameLobbyWatcher oldWatcher) {
+      @Nullable final InGameLobbyWatcher oldWatcher,
+      final Consumer<Boolean> connectivityHandler) {
```
DanVanAtta, line 57: I think we might be a bit too heavy on callbacks here. At this point it might be better to have a single response object, notably 'errorReporter' and it handle the connectivity check failing. This is moot though if we do not combine postGame

game-core/src/main/java/games/strategy/engine/framework/startup/ui/InGameLobbyWatcher.java
```diff
@@ -113,7 +117,12 @@ private InGameLobbyWatcher(
             .gameVersion("0")
             .build();
 
-    gameId = gameToLobbyConnection.postGame(gameDescription.toLobbyGame());
+    final PostGameResponse postGameResponse =
+        gameToLobbyConnection.postGame(gameDescription.toLobbyGame());
+
+    gameId = postGameResponse.getGameId();
+
+    connectivityHandler.accept(postGameResponse.isSuccessfulConnectivityCheck());
```
DanVanAtta, line 125: The connectivity handler perhaps should only be invoked on a failed connectivity check. I think it would create a much tighter abstraction rather than leaking details of everything that is involved with connecting. EG:

```
if(postGamREsponse.getErrorCode() == CONNECTIVITY_CHECK_FAILED) {
  connectivityHandler.accept( ... error message ... );
}
```

I'm also a bit concerned that the gameId is in an odd state in this case as well. FWIW this is where the code flow is probably simpler if we have the two be separate.

EG:

```
if(!connectvitiyCheck.verify( ..)) {
    connectivityHandler.accept( ... error message ... );
}

gameId = gameToLobbyConnection.postGame(gameDescription.toLobbyGame());

```

game-core/src/main/java/games/strategy/engine/framework/startup/ui/LobbyWatcherKeepAliveTask.java
```diff
@@ -33,7 +34,9 @@
   /** Call this to send a keep-alive request to server. */
   @Nonnull private final Predicate<String> keepAliveSender;
   /** Call this to re-post the current game, obtains a new game id. */
-  @Nonnull private final Supplier<String> gamePoster;
+  @Nonnull private final Supplier<PostGameResponse> gamePoster;
+  /** Call this to shutdown the client in case the server can't establish a connection. */
+  @Nonnull private final Consumer<Boolean> connectivityHandler;
```
DanVanAtta, line 39: This should not be needed, if we cannot create an initial connection then we should not be getting into the keep alive tasks (and those should be cancelled already by a failed connectivity check). The overload of keep alive is concerning, doubles the endpoint and I think is a bit of a stretch on SRP.

http-server/src/main/java/org/triplea/modules/game/listing/GameListing.java
```diff
@@ -71,18 +77,24 @@ public static GameListing build(final Jdbi jdbi, final WebSocketMessagingBus pla
...
     games.put(new GameId(apiKey, id), lobbyGame);
     playerMessagingBus.broadcastMessage(
         new LobbyGameUpdatedMessage(
             LobbyGameListing.builder().gameId(id).lobbyGame(lobbyGame).build()));
     log.info("Posted game: {}", id);
-    return id;
+
+    final boolean successfulCheck = connectivityCheck.test(getHostForGame(lobbyGame));
+    return PostGameResponse.builder()
+        .gameId(id)
```
DanVanAtta, line 95: I think this is getting a bit confused. The API returns a "here is your game id, but by the way, I cannot connect to your game!"

We are relying on client behavior to shut down and send a disconnect message to delist, otherwise we are waiting for the keep alive to time out. 

The server should probably not rely on client side implementation that heavily, if the connectivity check fails, we'd ideally not call the post game endpoint and/or not return a game id. 

If we do not return a game-id, which arguably is more cohesive, then the keep-alive endpoint contract becomes less consistent.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/LobbyWatcherKeepAliveTask.java
```diff
@@ -45,7 +48,9 @@ public void run() {
           reportReconnected();
         }
       } else {
-        gameId = gamePoster.get();
+        final PostGameResponse postGameResponse = gamePoster.get();
+        gameId = postGameResponse.getGameId();
+        connectivityHandler.accept(postGameResponse.isSuccessfulConnectivityCheck());
```
DanVanAtta, line 53: This can pose a major problem. Let's say a user disconnects from a network that has a direct connection to one that no longer does (which is about the only way you'd be able to create and post a game and then later have connectivity fail on a re-posting).

In that scenario the keep alive sender would no-op while disconnected, by design, but when we get a reconnection we would get a new game and then the connectivity check would fail. The connectivity handler would then crash the game that is then in-progress. That game crash would be quite severe, no option to save, the game just closes down on you with an error message.

One could simply not do the connectivity check here, which gets into my point a bit about interface segregation principle. I'm certainly leaning to decoupling the functionality so that we don't have dead code on certain call paths. Otherwise if the two functionalities remain coupled, then we need a different error handler implementation to not crash an already-in-progress game.

DanVanAtta: IF post-game can return a nullable gameId, the state here becomes a bit more complex as we now have potential for a null game id. That updates the test scenarios we need to account for the increased state complexity

DanVanAtta: Another item to consider with this code, we should not proceed to do another keep alive check if the connectivity check fails. The code here instead assumes the implementation of `connectivityHandler` will crash the game so that we do not proceed.

To keep the abstraction intact, we would probably want connectivity handler to be a predicate to return true/false if we should continue processing, but then again this code is already aware of the connectivity check result and probably could just no-op if the connectivity check is false. Messaging that well to a user is potentially a headache though.

RoiEXLab:
> Please revisit the discussion of pro/cons to this approach, why is this better than keeping the two endpoints separate?

Basically passing an IP address is not always feasible, the client doesn't know its public ip, so we'd have to make passing an ip optional, defaulting to the public IP, but still pass a port all of the time.

I think relying on gameId to verify the connection is still the right thing to do because all of the information is there on the server, it just has to access it and it sort-of ensures that the service isn't silently abused as a ddos bot.

There's also the concurrency issue you mentioned, the gameId might not exist when the connectivity check is performed, I'm not entirely convinced this is the case, but it's definitely a thing to consider.
However in my initial revision of this PR (which worked for me without any problems) I ran into the problem where I couldn't setup a unit test that had a pre-existing game id required for the test.

I currently feel like we're talking past each other on this topic, which is why I'd kindly ask you to checkout the first commit of this PR locally and have a look at the failing test class.
Maybe there's a simple solution there I just didn't realize and the whole discussion becomes redundant.

DanVanAtta:
@RoiEXLab good suggestion for us to back up and revisit. IMO the core problem is in this diff:
https://github.com/triplea-game/triplea/pull/6111/commits/389510b72b2c4f61d58ac058c2ba0900d5241199#diff-12c1e87164d775a4073116721ae0a814L40

Essentially the connectivity check is designed independent of the existence of a game. We could perhaps do a connectivity check when a player joins lobby, fire up a dummy server and see if it can be connected to. In that case we could disable the 'host' game with a hover text noting that they are not visible.

I think the "right"/easiest fix is to simply pass the IP+port combination that *should* be checked to the connectivity check endpoint.

RoiEXLab:
I have to admit that I currently don't have too much time to spare for triplea (at least for the next 2 weeks), so this might take a little longer than expected.

@DanVanAtta If you have an idea on what could be done here, I'd happily apply the changes, I just don't have the time to try some random changes until something works for now 😅

DanVanAtta:
Thanks for the heads up @RoiEXLab ; I might be able to implement something here. We may need to admit that the documentation and full roll out of this feature might be after the initial 2.0 launch. It's a conversation to be had when we'll break 2.0 compatibility, for some time after 2.0 we will have additional compatibles being released (far more frequently)

RoiEXLab:
@DanVanAtta Documentation is something I could write without too much effort.
Just tell me what kind of documentation you'd like to see and I'll see if I can fit it somewhere in my limited free time schedule ^^

DanVanAtta:
@RoiEXLab I think ideally we'd have the 'how to host' well organized as a user guide on the website and including a section on this option/feature. There may be some organization work to be done for it to be clean and easy to use. Finally, it would probably be good for the error message of the connectivity check to then refer a player to the user-guides for help on how to host.

Side-note: if you could do some CSS magic to put a border on the release note table on website, that would be pretty awesome as well :grin:

RoiEXLab:
👍 

> if you could do some CSS magic to put a border on the release note table on website, that would be pretty awesome as well 😁

Should be doable, but not today, please remind me of it in case I forget ^^

RoiEXLab:
@DanVanAtta Haven't had the time to create a PR, but here you go with a rough css scheme, just in case I don't get to create a PR soon-ish:

The CSS would looks as follows:
```css
.table-wrapper {
    width: 100%;
    border: 1px solid #c6c9cc;
    border-radius: 5px;
    padding: 10px 20px;
}

.table-wrapper table {
    border-collapse: collapse;
    width: 100%;
    text-align: left;
}

.table-wrapper table td {
    border-color: #c6c9cc;
    border-style: solid;
    border-width: 1px 0 0 0;
}
```

There are some adjustments that need to be made to the html though.
This means that we'll have to rename the file to .html instead of .md.
Basically we'll have to wrap the table inside a div like this:
```markdown
<div class="table-wrapper">
| Area  | Type  | Description  |
|---|---|---|
| Engine  | Feature | Automatic bug reporting, dialog to upload error logs appears when a game error occur  |
// etc.
</div>
```

In the end the table will look like this:
Table

Alternatively if you're not a fan of the border, or just don't want to modify any HTML this is the "base version" that doesn't require any adjustments to html:
Base Version

In this case you'd just remove the `.table-wrapper` part from the css (this will remove the first rule entirely)

DanVanAtta:
@RoiEXLab this is updated

- found a bug with the connectivity check warning dialog being closed on system.exit, that is fixed by moving the exist to be after the dialog is closed
- improved location of where the client launches the connectivity check to be right after server accepts connections
- improved testing some
- moved some functionality out of connectivity controller and into connectivity check itself
- fixed integration test, notably made it be a bit more naive and is happy getting a 400 back for game not found and does not worry about doing the reverse connection successfully. The controller integration test is more interested to know that the endpoint is available and can return any non-500 answer, rather than if it works logically (unit tests cover that).

I tested with the pre-release lobby and because the API has seemingly changed a bit I got a 400. I noticed that the lobby game did not close down. That is an outstanding problem. I think there is strong argument then for doing the connectivity check as part of the game-hosting-request that initially posts the game. In this way the lobby can avoid listing the game in the first place and the client does not have to do an explicit connectivity check. The only issue to be careful of is if the game-posting is done before the server is ready to accept incoming connections.

Looking at how the wiring was  re-done here to use game-id; it looks reasonable enough at this juncture. I have changed my opinion whether to make the connectivity check to be purely IP based, the game-id lookup has its benefits.

RoiEXLab:
@DanVanAtta The smoke test seems to be consistently failing because the application shuts itself down before it can be shutdown by the script

http-server/src/main/java/org/triplea/modules/game/ConnectivityCheck.java
```diff
@@ -1,27 +1,43 @@
...
+
+  /** Verifies if a game exists with a given API key and gameId. */
+  boolean gameExists(final ApiKey apiKey, final String gameId) {
+    return gameListing.getHostForGame(apiKey, gameId).isPresent();
+  }
 
-  ConnectivityCheck() {
-    this(Socket::new);
+  /** Checks if the lobby can create a connection to a given game. */
+  boolean canDoReverseConnect(final ApiKey apiKey, final String gameId) {
```
RoiEXLab, line 30: This is an excellent naming change 👍

http-server/src/main/java/org/triplea/modules/game/ConnectivityCheck.java
```diff
@@ -1,27 +1,43 @@
...
+        .orElse(false);
   }
 
-  @Override
-  public boolean test(final InetSocketAddress address) {
+  private boolean testConnectivityToAddress(final InetSocketAddress address) {
     try (Socket s = socketSupplier.get()) {
       s.connect(address, (int) TimeUnit.SECONDS.toMillis(10));
-      return true;
+      return s.isConnected();
```
RoiEXLab, line 40: That's an important details as well 👍

http-server/src/main/java/org/triplea/modules/game/ConnectivityController.java
```diff
@@ -40,9 +37,15 @@ public static ConnectivityController build(final GameListing gameListing) {
...
-    return connectivityCheck.test(
-        gameListing.getHostForGame(authenticatedUser.getApiKey(), gameId));
+
+    if (!connectivityCheck.gameExists(authenticatedUser.getApiKey(), gameId)) {
+      return Response.status(400).build();
+    }
+
+    return Response.ok()
+        .entity(connectivityCheck.canDoReverseConnect(authenticatedUser.getApiKey(), gameId))
+        .build();
```
RoiEXLab, line 49: This is something I have absolutely no experience with:
Basically the result of canDoReverseConnect is false whenever the game does not exist, but that means nothing.

I think I would've coded this in a way so canDoReverseConnect throws an exception whenever the game doesn't exist and map the exception to a http status code somehow.
Is this something Dropwizard is designed to do, or do you think it's better the way it currently is?

http-server/src/main/java/org/triplea/modules/game/ConnectivityController.java
```diff
@@ -40,9 +37,15 @@ public static ConnectivityController build(final GameListing gameListing) {
...
       rates = {@Rate(limit = 10, duration = 1, timeUnit = TimeUnit.MINUTES)})
   @RolesAllowed(UserRole.HOST)
-  public boolean checkConnectivity(
+  public Response checkConnectivity(
       @Auth final AuthenticatedUser authenticatedUser, final String gameId) {
-    return connectivityCheck.test(
-        gameListing.getHostForGame(authenticatedUser.getApiKey(), gameId));
+
+    if (!connectivityCheck.gameExists(authenticatedUser.getApiKey(), gameId)) {
+      return Response.status(400).build();
```
RoiEXLab, line 44: I'm always a bit suspicious when I see explicit 400 status codes, they always more or less hint on syntax errors.
I think 404 might be a better choice here actually.

See https://httpstatuses.com/400 and https://httpstatuses.com/404

RoiEXLab:
@DanVanAtta I created a guide btw.: https://forums.triplea-game.org/topic/1970/how-to-host-using-a-relay-service

DanVanAtta:
@RoiEXLab any chance I can convince you to move the write-up to the website?

I found that over the years there has been a LOT of duplicative and out of date documentation for tripleA. Forums posts were one of those. I also found that forum posts were quite bad for being out of date and were part of maintenance problems as there were quite a lot and needed a lot of links. My hope is all documentation can be hosted on the 'user-guide' on the website, and then linked from the both the game and forum.

https://triplea-game.org/user-guide/

There is currently a section of 'how to host', it seems incomplete (:wink: ) now without mention of 'relay service'.

DanVanAtta:
Props to the smoke test, it found a real problem :smiley: 

3093b22 fixes, I had 'host name' instead of 'host address' for the reverse connection address.

http-server/src/main/java/org/triplea/modules/game/ConnectivityController.java
DanVanAtta: 404 potentially has issues:
- 'resource' generally means the URL resource (right?)
- we won't be able to distinguish between an actual 404 and 404 due to bad data. For example, a common mistake is not register a controller with the ServerApplication. In such cases you get a 404 as the path you created is not actually listening.

You're right though about generic 400 being more syntax error @RoiEXLab .

422 - unprocessable entity seems to fit the bill.

DanVanAtta: Updated to 422: 27e3415469f2cd47c20d788387de73b7c1cab4f4

http-server/src/main/java/org/triplea/modules/game/ConnectivityController.java
DanVanAtta: There are exception mappers:

- https://github.com/triplea-game/triplea/blob/master/http-server/src/main/java/org/triplea/http/IllegalArgumentMapper.java
- https://github.com/triplea-game/triplea/blob/85825be5a9a6ac937fc43d599e2a53a432107f56/http-server/src/main/java/org/triplea/http/ServerApplication.java#L194

Some caution needs to be taken with them as they can go bad if too heavily used. Essentially control-flow-by-exception-handling anti-pattern. Signalling error cases directly via response objects and letting the controller layer or service/module layer decide how to handle is generally more direct (after-all, an exception is a glorified go-to, for better or worse).

The game-id not being there is quite a problem. We would suspect it's a client error, hence 4xx is reasonable. We would not want users to be able to craft a curl command to generate server side 5xx's on demand. 

`canDoReverseConnect`  not throwing is kinda useful as we allow its caller to decide if the situation is an error or not. It's the caller that has more context about whether the game-id is valid to begin with. It's a bit odd though as we ask the same module twice "is this valid", then "try this". It would probably be a bit more appropriate to have a service back the controller, have the server use a 'canDoReverseConnect' function and pass the function the game listing to use. I think that is probably getting to be a bit too over-engineered though.

DanVanAtta: >  or do you think it's better the way it currently is?

I was typing a direct response to this and considered return a flag more deeply, namely "Signalling error cases directly via response objects ". It turns out that becomes cleaner and resolve my concern where we are being very procedural with connectivity check, ie:
```
connectivity check, can you do this?
if yes, connectivity check, please try X
```
Rather than:
```
connectivity check, do X
```

The update is I think a bit cleaner in the end: 1f624bad5

DanVanAtta:
With luck the build will pass with the last update, marking as ready for review.

RoiEXLab:
@DanVanAtta We can move it to the website once we release a new stable.
For now I just wanted our "power users" to read it and potentially submit some feedback when I missed some details that might seem obvious to us developers.

DanVanAtta:
Sounds like  a plan @RoiEXLab :+1:

http-server/src/main/java/org/triplea/modules/game/ConnectivityController.java
```diff
@@ -38,11 +38,21 @@ public static ConnectivityController build() {
...
+
+    switch (result) {
+      case SUCCESS:
+        return Response.ok().entity(true).build();
+      case FAILED:
+        return Response.ok().entity(false).build();
+      case GAME_ID_NOT_FOUND:
+        return Response.status(HttpStatus.UNPROCESSABLE_ENTITY_422).build();
+      default:
+        throw new IllegalStateException("Switch case not handled: " + result);
```
RoiEXLab, line 55: Much better IMO 👍

DanVanAtta: Agree, particularly getting the logic out of the controller.

PR #6116, opened 2020-03-31
Fix BC Not At War Bug

ron-murhammer:
Addresses bug reported here: https://forums.triplea-game.org/topic/1836/release-2-0-prep-and-testing/3

The battle calc had a bug related to simulating battles where 2 nations aren't actually at war yet that is causing the odd AI behavior. You can actually see this yourself if you simulate a battle in global before 2 nations are at war with say 1 UK ship attacking a large Japanese fleet in the BC and you'll see that it shows a draw with all units destroyed. This is now fixed as it properly consider non-enemy (not at war) remaining defenders.

Tested
- Loading the given save game and the AI making more reasonable moves
- Various scenarios with the BC to ensure proper TUV results

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
@@ -260,6 +260,11 @@ public class MustFightBattle extends DependentBattle implements BattleStepString
     }
   }
 
+  /**
+   * Used by battle results to get the remaining attackers after the battle is completed. It
+   * includes remaining attackers, retreated attackers, and if attacker won/draw then any owned
+   * units left in the territory.
+   */
   @Override
   public List<Unit> getRemainingAttackingUnits() {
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -278,14 +278,17 @@ void addDependentUnits(final Map<Unit, Collection<Unit>> dependencies) {
 
   @Override
   public List<Unit> getRemainingDefendingUnits() {
-    final List<Unit> remaining = new ArrayList<>(defendingUnitsRetreated);
+    final Set<Unit> remaining = new HashSet<>(defendingUnitsRetreated);
+    remaining.addAll(defendingUnits);
```
DanVanAtta: This is a bit of a surprising update. This line seems like a really important item. I'm now not sure how this was even close to correct before.

ron-murhammer:
Added some method javadoc. Merging.

PR #6117, opened 2020-03-31
Update Unit Scroller - Single Row of Buttons

DanVanAtta:
Changes to the unit scroller:
- Put buttons on same row
- Scaled/shrank button images to 20x20
- Removed the unit flag and unit count
- Removed the unit highlight button and center on units button
- Decreased button horizontal gap spacing (now 2 pixels apart rather than 15)
- Fixed up layout so the component does not stretch to fill half of the vertical space of the action tab
- Put the component behind a collapse button, clicking the collapse button shows/hides the component
- Unit scroller is now always visible during combat and non-combat moves
- Made the territory text larger
- Moved the "units left to move count" to be part of the label for "Units To Move (#)"
- Moved the territory label to be below the units rather than above

Fixes:
- The check for if the scroller was visible skipped the computation of selecting
  the current players capitol. This meant the unit scroller was not pre-filled
  with a territory or units. An empty unit scroller is only really expected
  when all units have moved. The problem was introduced in another fix to avoid
  unnecessary computation when the unit scroller is not visible anyways. There
  are a number of other fixes that make that not as much of a problem. For now
  it is necessary to do computation in the unit scroller when it is not visible
  so that when it does become visible, the data in the unit scroller would be
  current.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

- Ran through a few turn phases to be sure that the unit scroller appeared and disappeared at expected times. 
- Exercised the unit scroller buttons, next/previous etc.. and verified behavior when unit movement was exhausted. 
- Verified that on a new player turn the players capitol is the 'current selected' territory and has units. 
- Verified that the units to move count updates/resets.

### Before

This is an example of the 'bug' that was fixed, on initial render there is no current selection.

Screenshot from 2020-03-31 00-40-46

Screenshot from 2020-03-31 00-40-59

### After
Screenshot from 2020-03-31 00-35-49
Screenshot from 2020-03-31 00-18-40

## Additional Review Notes

Forums feature thread: https://forums.triplea-game.org/topic/1433/screen-centering-cycling-around-map-ui-idea

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/AbstractMovePanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/AbstractMovePanel.java
@@ -4,7 +4,6 @@ import games.strategy.engine.data.GameData;
 import games.strategy.engine.data.GamePlayer;
 import games.strategy.engine.data.MoveDescription;
 import games.strategy.engine.data.Unit;
-import games.strategy.engine.framework.LocalPlayers;
 import games.strategy.engine.player.IPlayerBridge;
 import games.strategy.triplea.delegate.UndoableMove;
 import games.strategy.triplea.delegate.remote.IAbstractMoveDelegate;
@@ -59,7 +58,7 @@ abstract class AbstractMovePanel extends ActionPanel {
     }
   }
 
-  abstract Component getUnitScrollerPanel(LocalPlayers localPlayers);
+  abstract Component getUnitScrollerPanel();
 
   /*
    * sub-classes method for done handling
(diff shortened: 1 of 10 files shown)


DanVanAtta:
Re (1) The unused icons are in the assets repo. Need to fix up an SSH key config there first. But yes.

Re (2) Indeed I see what appears to be extra space. I think there may still be some padding there or could be the rendering. Making the avatar look better is a focus for follow-ups, will see if i can fix that.

PR #6137, opened 2020-04-05
Fix potential server-side timeout

RoiEXLab:
Follow up to #6104 should help debug this issue

## Functional Changes

[] New map or map update
[] New Feature
[] Feature update or enhancement
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

I verified I could start a local networked game without any problems

PR #6138, opened 2020-04-05
Simplify endpoint

RoiEXLab:
Some things I noticed when working on #6137

Note that there are small changes in behaviour: Sometimes where an exception would be thrown now a no-op is done instead.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

I verified I could start a local networked game

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/message/unifiedmessenger/EndPoint.java
+++ b/game-core/src/main/java/games/strategy/engine/message/unifiedmessenger/EndPoint.java
@@ -6,10 +6,12 @@ import games.strategy.engine.message.RemoteMethodCallResults;
 import games.strategy.net.INode;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.util.ArrayList;
 import java.util.List;
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.logging.Level;
+import java.util.stream.Collectors;
 import lombok.extern.java.Log;
 
 /**
@@ -24,10 +26,9 @@ class EndPoint {
   // the next number we can run
   private long currentRunnableNumber = 0;
   private final Object numberMutex = new Object();
-  private final Object implementorsMutex = new Object();
   private final String name;
   private final Class<?> remoteClass;
-  private final List<Object> implementors = new ArrayList<>();
+  private final Set<Object> implementors = new CopyOnWriteArraySet<>();
   private final boolean singleThreaded;
 
   EndPoint(final String name, final Class<?> remoteClass, final boolean singleThreaded) {
@@ -37,12 +38,10 @@ class EndPoint {
   }
 
   public Object getFirstImplementor() {
-    synchronized (implementorsMutex) {
-      if (implementors.size() != 1) {
-        throw new IllegalStateException("Invalid implementor count, " + implementors);
-      }
-      return implementors.get(0);
+    if (implementors.size() != 1) {
+      throw new IllegalStateException("Invalid implementor count, " + implementors);
     }
+    return implementors.iterator().next();
   }
 
   public long takeANumber() {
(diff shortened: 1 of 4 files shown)


game-core/src/main/java/games/strategy/engine/message/unifiedmessenger/EndPoint.java
```diff
@@ -24,10 +26,9 @@
   // the next number we can run
   private long currentRunnableNumber = 0;
   private final Object numberMutex = new Object();
-  private final Object implementorsMutex = new Object();
   private final String name;
   private final Class<?> remoteClass;
-  private final List<Object> implementors = new ArrayList<>();
+  private final Set<Object> implementors = new CopyOnWriteArraySet<>();
```
DanVanAtta, line 31: nit, I'd assign this to `Collection`. I've found that in cases where we pass `Set` to places that want a `List` parameter but could be a collection, we incur a collection copy unnecessarily.

EG:

```
foo(List.of(set));
void foo(List thisDoesNotHaveToBeList);

vs:

foo(setPassedAsCollection)
void foo(Collection anyCollectionWillDo);
```

game-core/src/main/java/games/strategy/engine/message/unifiedmessenger/EndPoint.java
```diff
@@ -91,19 +91,16 @@ public int getLocalImplementorCount() {
...
    */
   boolean removeImplementor(final Object implementor) {
-    synchronized (implementorsMutex) {
-      if (!implementors.remove(implementor)) {
-        throw new IllegalStateException(
-            "Not removed, impl:" + implementor + " have " + implementors);
-      }
-      return implementors.isEmpty();
+    if (!implementors.remove(implementor)) {
+      throw new IllegalStateException("Not removed, impl:" + implementor + " have " + implementors);
```
DanVanAtta, line 95: Side-comment (as this is existing behavior):

I've questioned a bit why the original code throws on this case.
Perhaps because we're trying to detect a bug where we did not remove an implementor and we got the name wrong?

If that is the motivation, then adding a remove listener is probably a better way to go, or having a 'removeAll' hook maybe.

game-core/src/main/java/games/strategy/engine/message/unifiedmessenger/EndPoint.java
```diff
@@ -119,16 +116,9 @@ boolean removeImplementor(final Object implementor) {
...
-    final List<Object> implementorsCopy;
-    synchronized (implementorsMutex) {
-      implementorsCopy = new ArrayList<>(implementors);
-    }
-    final List<RemoteMethodCallResults> results = new ArrayList<>(implementorsCopy.size());
-    for (final Object implementor : implementorsCopy) {
-      results.add(invokeSingle(call, implementor, messageOriginator));
-    }
-    return results;
+    return implementors.stream()
```
DanVanAtta, line 119: Important, is there *any* chance we will invoke this method while implementors are being removed? If so we could hit a concurrent modification exception. Have you considered creating an implementor collection copy? Since we are copying references of a couple dozen objects, it should be a negligibly cheap operation.

game-core/src/main/java/games/strategy/engine/message/unifiedmessenger/EndPoint.java
RoiEXLab: removeAll maybe, but I don't think all use cases always remove everything.
No idea why the exception is there either.

game-core/src/main/java/games/strategy/engine/message/unifiedmessenger/EndPoint.java
RoiEXLab: Unlikely we'll invoke both at the same time, but implementors is a thread safe implementation, which effectively creates a defensive copy whenever something is written to it so no worries there.

DanVanAtta: Cool, thanks :+1:

Issue #6105, opened 2020-03-27
Save Game Should prefill a suggested save name


DanVanAtta:
**How can the problem be recreated?**
1. From a new game
- start a new game
- single player
- save game
- notice that the save text field for file name is free-form and probably blank

Expected: some sort of prefilled file name, perhaps date (YYYY-mm-dd + map name)

2. From a load game
- load game
- save it
- notice the save field dialog is blank

Expected: save game file name should be prefilled to be the name of the game that was loaded

prastle:
:) Good Stuff

DanVanAtta:
this will be fixed by: https://github.com/triplea-game/triplea/pull/6143

beelee1:
so this still doesn't work quite as good as 13066. If you start a save, then make another save, it always shows the first save, instead of the new save. 

For example: I start with save game "test". I then save the game and name it "test 1". When I go to save again, it goes back to "test" instead of staying on "test 1".

On 19071

PR #6143, opened 2020-04-05
Default save game name

DanVanAtta:
## Commits
commit 9c2b6dbb0114a785bd5d12f2469baded41a1eeff

    Move save-game file chooser logic to SaveGameFileChooser

commit 5bee76ed6a6b6369dc62a0f5e19de7adc0f381be

    Add default save game name
    
    This is done by adding a gameData property that is the save game file name
    that we are loading from. The property is null if we are starting a new game.
    
    In the case wheere we are loading new game, instead of defaulting
    the save game file name to the name of the save game, we instead create
     a default name with this pattern "YYYY-MM-DD-map-name.tsvg"

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  https://github.com/triplea-game/triplea/issues/6105  
[] Other:   

## Testing
verified:
- default file name is set for new games
- save game file name matches the file name for load games
- verified the two points above hold for both host+client on network games

## Additional Review Notes

- It's technically a new feature that we create a default save game file name for new games, previously we only had a prefilled name for load games
- The 'SaveGameFileChooser' was unused, it was missed as part of a previous removal when we went from FileChooser to FileDialog recently.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/ui/SaveGameFileChooser.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/ui/SaveGameFileChooser.java
@@ -1,10 +1,13 @@
 package games.strategy.engine.framework.ui;
 
+import games.strategy.engine.data.GameData;
 import games.strategy.engine.framework.GameDataFileUtils;
 import games.strategy.triplea.settings.ClientSetting;
 import java.awt.FileDialog;
 import java.awt.Frame;
 import java.io.File;
+import java.time.ZonedDateTime;
+import java.util.StringJoiner;
 import javax.swing.JFileChooser;
 
 /** A file chooser for save games. Defaults to the user's configured save game folder. */
@@ -19,13 +22,14 @@ public final class SaveGameFileChooser extends JFileChooser {
    * @return The file to which the current game should be saved or {@code null} if the user
    *     cancelled the operation.
    */
-  public static File getSaveGameLocation(final Frame frame) {
+  public static File getSaveGameLocation(final Frame frame, final GameData gameData) {
     final FileDialog fileDialog = new FileDialog(frame);
     fileDialog.setMode(FileDialog.SAVE);
     fileDialog.setDirectory(ClientSetting.saveGamesFolderPath.getValueOrThrow().toString());
     fileDialog.setFilenameFilter((dir, name) -> GameDataFileUtils.isCandidateFileName(name));
-    fileDialog.setVisible(true);
+    fileDialog.setFile(getSaveGameName(gameData));
 
+    fileDialog.setVisible(true);
     final String fileName = fileDialog.getFile();
     if (fileName == null) {
       return null;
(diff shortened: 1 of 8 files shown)


game-core/src/main/java/games/strategy/engine/framework/ui/SaveGameFileChooser.java
```diff
@@ -1,48 +1,56 @@
...
-        return "Saved Games, *" + GameDataFileUtils.getExtension();
-      }
-    };
+  private static String formatGameName(final String gameName) {
+    final ZonedDateTime now = ZonedDateTime.now();
+    return new StringJoiner("-")
+        .add(String.valueOf(now.getYear()))
+        .add(String.valueOf(now.getMonthValue()))
+        .add(String.valueOf(now.getDayOfMonth()))
+        .add(gameName.replaceAll(" ", "-") + GameDataFileUtils.getExtension())
```
RoiEXLab, line 53: Thoughts about using https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/format/DateTimeFormatter.html#ISO_LOCAL_DATE for date formatting?

game-core/src/main/java/games/strategy/engine/framework/ui/SaveGameFileChooser.java
```diff
@@ -1,48 +1,56 @@
...
-    return instance;
-  }
 
-  private static void ensureDirectoryExists(final File f) {
-    if (!f.mkdirs() && !f.exists()) {
-      throw new IllegalStateException("Unable to create save game folder: " + f.getAbsolutePath());
-    }
+    // If the user selects a filename that already exists,
+    // the AWT Dialog will ask the user for confirmation
+    return new File(fileDialog.getDirectory(), GameDataFileUtils.addExtensionIfAbsent(fileName));
```
RoiEXLab, line 40: Remember our discussion about whether to use
```java
Arrays.stream(fileDialog.getFiles()).findAny().orElseThrow()
```
?
Because of the extension thingy it would be a little bit impractical to use this approach here, but without this detail it would save a couple lines of code.

game-core/src/main/java/games/strategy/engine/framework/ui/SaveGameFileChooser.java
DanVanAtta: It's better to use built-in items in general, but primitive and simple has it's virtue as well. IMO it's a wash here, there is nothing difficult to understand about this code. Meanwhile there is a format code that ties the year to the calendar week, which causes Dec-31 to flip to Jan-1, so there is some subtlety/caution to be used there.

game-core/src/main/java/games/strategy/engine/framework/ui/SaveGameFileChooser.java
DanVanAtta: I don't recall that discussion. It seems we are maybe picking on moved code. If there is a reasonable improvement, IMO follow-up with it.

Issue #5921, opened 2020-02-09
Email Credentials in 2.0.17997


panther2:
I might have missed that - but where are the login-credentials, entered in the "Play by Forum/Email" section of the "Engine Preferences" saved? And are they saved in an encrypted way?

RoiEXLab:
On windows, the passwords are stored in the registry, encrypted with a fixed key to avoid accidental leaks.
I'm not sure if you're able to set a master password for this.
However if you want "true security" you probably shouldn't store it there

DanVanAtta:
Any thoughts on prompting users for email password and never storing it? If the passwords are recoverable in plain-text, then they are recoverable in plain-text no matter how we do it.

We may want the same for the forums password. The convenience trade-off is more favorable for forum password. For me at least, probably most people, email password being compromised is as bad as it gets and results in actual financial ruin.

Second consideration/question, we seemingly have a perfectly good email server running on warclub, why not always use that for email?

panther2:
As long as all login details for forum and email including passwords are stored encrypted personally I can live with that, as I am using different logins/email-addresses/passwords for different purposes.

Generally I would welcome a solution being prompted for login details without storing them anywhere. Wasn't it like this in the past? I had to mark "remember me" to have those details saved, IIRC. 

Maybe this might be a solution:
In case there are no details entered in the "Engine Preferences" and the user chooses PBF/PBEM prompt him to enter the details when either testing to post or trying to post his turn. In a respective dialogue the user could be asked to "pemanently save" the details ("remember me") or to enter the details for a one-time puropse.

DanVanAtta:
While the credentials we store are not plain-text, TripleA can recover their plain-text value, which means any attacker who finds the non-plain-text values could similarly use the same code as TripleA uses to recover the non-plain-text value. 

We could prompt for password once per game session and then 'remember' password and store it in memory. If we store the password in persistent manner, @RoiEXLab mentioned needing a master password which would be the way the encryption would be unlocked in a way that you could not just do so with the TripleA software. 

In sum, to keep it secure, it seems we either need to:
1. prompt for password and remember it in memory only if authorized by user
2. (or) require a master password to unlock an encrypted password that is persisted to the system

Any thoughts/other suggestions anyone?

RoiEXLab:
We had a master password mechanism some time ago, but I don't remember what happened to it. Maybe @ssoloff does remember?

ssoloff:
[...]

PR #6147, opened 2020-04-06
Pbf pbem remember password

DanVanAtta:
## Summary

- Add username/password to PBEM screen

After this update the PBEM screen 'play' button can be enabled by filling in all forms found only on the PBEM screen.

- Password handling on PBEM and PBF are updated to have 'remember password' buttons that will blank out the password on game exit if not selected. 

## Commits

commit a15d422e6343cb1323c9c79eb33ee9dade78a848

    Update PBF & PBEM to have username+password and remember password

commit 3db45c1fddbf8e8236303b39f0ff76001ba523db

    Improve help text formatting

commit 17b146c6f96417aeccec5bd00eb5a6039fc39b30

    Update handling of forum password
    
    - Do not store the password as a variable in view model
    - This makes it consistent with how we handle email password
      in the PBEM view model.
    - Of note, we set only a dummy text in the password field
    - When password field is set, we will set the ClientSetting
      password value and then toggle a flag to indicate password
      has been set, but we will not store it as a variable.

commit 7a98d1bc705d52d3eeb0b7b6561e87c3ee77ff18

    Make help text fetch be consistent remove the only if statement in ForumPosterEditor constructor

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  https://github.com/triplea-game/triplea/issues/5921, https://github.com/triplea-game/triplea/issues/5856 
[] Other:   

## Testing
- test cover is pretty good, gives confidence on the backend
- play-tested the fields being set/unset to verify the label highlight and play button being enabled
- play-tested password remember being remembered and/or forgotton

### Before
Screenshot from 2020-04-05 20-10-48

### After
Screenshot from 2020-04-05 20-09-51

### Before
Screenshot from 2020-04-05 20-10-57

### After
Screenshot from 2020-04-05 20-06-31

### Before
Screenshot from 2020-04-05 20-11-10

### After
Screenshot from 2020-04-05 20-06-21


game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/HelpTexts.java
```diff
@@ -0,0 +1,135 @@
...
+          + "Put your username and password for the AxisAndAllies.org forum into the "
+          + "username and password fields<br/>"
+          + "Click the Test Post button, to check that TripleA can reply to your forum "
+          + "post<br/>"
+          + "<i>Note:</i> Your AxisAndAllies.org username and password are not stored as part "
+          + "of the save game, but they are stored encrypted in the local file system "
+          + "if you select the option to remember your credentials. You may have to "
+          + "enter your username and password again if you open the save game on "
+          + "another computer."
+          + "</p></html>";
```
RoiEXLab, line 53: This is no longer accurate, the axis & allies forum migrated to NodeBB quite some time ago, so the description should be more or less equivalent to our forum

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/HelpTexts.java
```diff
@@ -0,0 +1,135 @@
...
+          + "<b>Post:</b> The post of the smtp server (typically 25 for unencrypted, and 587 "
+          + "for TLS encrypted servers)<br/>"
+          + "Use TLS encryption: check this if yor server supports the TLS protocol for "
+          + "email encryption<br/>"
+          + "<i>Note:</i> Your SMTP login and password are not stored as part of the save game, "
+          + "but they are stored encrypted in the local file system if you select the "
+          + "option to remember your credentials.<br/>"
+          + "You may have to enter your login and password again if you open the save "
+          + "game on another computer."
+          + "</p></html>";
```
RoiEXLab, line 85: Also thoughts about moving this to a resource html file?
HTML in source code is alway kind of weird.

game-core/src/main/java/org/triplea/util/ExitStatus.java
```diff
@@ -14,10 +14,10 @@
...
   FAILURE(1);
 
+  private static final Collection<Runnable> exitActions = new HashSet<>();
   private final int status;
-  private final List<Runnable> exitActions = new ArrayList<>();
 
-  public void addExitAction(final Runnable runnable) {
+  public static void addExitAction(final Runnable runnable) {
     exitActions.add(runnable);
   }
```
RoiEXLab, line 22: Ok now there's no difference to https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Runtime.html#addShutdownHook(java.lang.Thread)
Thoghts about removing this code?

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/HelpTexts.java
DanVanAtta: The help text could use improvement, it has not been given much love, just mainly restored so far. No PR so far has really sought to change or improve it.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/HelpTexts.java
DanVanAtta: IMO it's overkill. A class with dedicated constants and lots of HTML IMO is fine, avoids complexity of making sure a resource file is in the right place. Eventually this is going to be behind a string identifier and presumably translated to multiple languages, I'd be concerned about code churn for anything short of that update.

game-core/src/main/java/org/triplea/util/ExitStatus.java
DanVanAtta: I didn't realize that API existed, good to learn about. It looks like we should indeed use that and remove our custom listener implementation.

PR #6153, opened 2020-04-07
Update unit scroller behavior when no moves left & fix alert button getting focus

DanVanAtta:
commit 95649dc667aa5277004f48cd6dfcad39a8be0028

    Update unit scroller behavior when no moves left
    
    1) When no moves left:
    - do not render the territory label
    - if skipping the last unit, getting to zero moves
      left, then remove the unit avatar rendering image
    
    2) Also, (cleanup) remove an unnecessary update to
       'movesLeft' in territory listener. Changing territory
       hover should not change the moves left and is not a
       necessary invocation.

commit 7910ed892f70d3fc3333e0b0caaffc5cf9d3cefa

    Remove focusable from 'alert' button
    
    Clicking the alert button repeatedly has no real effect, a user
    typically does not want to do this. After a first click, the button
    gains focus. *Then* pressing space will activate the button. This
    is confusing behavior as the next available unit is selected
    and it seems like we are 'skipping' units with the space button
    but we are really pressing 'wake-all' repeatedly.
    
    To avoid this confusing scenario, remove focusable from 'alert'
    button. This way if it is clicked, and then a user presses space,
    that will activate the 'skip' button.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

- some play testing

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/unit/scroller/UnitScroller.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/unit/scroller/UnitScroller.java
@@ -258,6 +258,7 @@ public class UnitScroller {
     final JButton wakeAllButton = new JButton(UnitScrollerIcon.WAKE_ALL.get());
     wakeAllButton.setToolTipText(WAKE_ALL_TOOLTIP);
     wakeAllButton.addActionListener(e -> wakeAllUnits());
+    wakeAllButton.setFocusable(false);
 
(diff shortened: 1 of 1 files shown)


DanVanAtta:
:smile: Nothing really coming to mind for a better name without getting super wordy. `movesRemaining` invites a problem where it seems like a list of moves rather than a count. So we then get to `movesRemainingCount`, which then leads to how do we rename the "updateMovesLeft()" method as well.

PR #6155, opened 2020-04-07
Hide units to place collapsible panel during placement

DanVanAtta:
## Summary
- Changes place panel to be rendered as it was in 1.9
- Updates placement collapsible panel to show placement unit count
- Removes "None" label from collapsible panel when zero units are purchased
- Swaps location of the unit scroller with placements collapsible panel

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
- play tested on napoleon
  - played through a couple players turns
- play tested on revised

## Screens Shots

Screenshot from 2020-04-07 00-07-38

Screenshot from 2020-04-07 00-10-44

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/PlacementUnitsCollapsiblePanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/PlacementUnitsCollapsiblePanel.java
@@ -1,0 +1,76 @@
+package games.strategy.triplea.ui;
+
+import games.strategy.engine.data.GameData;
+import games.strategy.engine.data.GameDataEvent;
+import games.strategy.engine.data.GamePlayer;
+import games.strategy.engine.data.GameStep;
+import games.strategy.triplea.util.UnitSeparator;
+import javax.swing.SwingUtilities;
+import lombok.Getter;
+import org.triplea.swing.CollapsiblePanel;
+
+class PlacementUnitsCollapsiblePanel {
+  private final SimpleUnitPanel unitsToPlacePanel;
+
+  @Getter private final CollapsiblePanel panel;
+  private final GameData gameData;
+
+  PlacementUnitsCollapsiblePanel(final GameData gameData, final UiContext uiContext) {
+    this.gameData = gameData;
+    unitsToPlacePanel =
+        new SimpleUnitPanel(
+            uiContext, SimpleUnitPanel.Style.SMALL_ICONS_WRAPPED_WITH_LABEL_WHEN_EMPTY);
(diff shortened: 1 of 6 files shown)


game-core/src/main/java/games/strategy/triplea/ui/PlacementUnitsCollapsiblePanel.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  private static boolean isInitializationStep(final GameStep gameStep) {
+    return gameStep.getPlayerId() == null;
+  }
+
+  private boolean stepIsAfterPurchaseAndBeforePlacement(final GameStep step) {
+    final GamePlayer currentPlayer = step.getPlayerId();
+
+    for (int i = gameData.getSequence().getStepIndex() - 1; i >= 0; i--) {
```
RoiEXLab, line 58: Is there any specific reason why we iterate backwards over the steps?

DanVanAtta: Yes, we're trying to see if given a current step, that within the same turn if a player has passed through a purchase phase. Some maps have a combat move before purchase, so we can't just check if we are in a combat move.

I'll note as well, there's still an assumption that there will be only one place phase, that it has a specific name too, all of which is already a bit brittle.

RoiEXLab: Ok, but why is the order important then?
Wouldn't it be sufficient to iterate from 0 to step index over the elements?

DanVanAtta: Perhaps important to keep in mind that steps is all steps across a full round of play, it has a length of around 70. If we go from 0, we will iterate more, but more importantly it's harder to know when we've dropped to the previously players turn.

asvitkine:
Some feedback after playing with this change recently:

I think not showing the panel during place units phase is not an improvement - especially for non-active players. It's very helpful to see what's left to place - especially when playing against AIs as there is not a lot of time to see the purchase (e.g. you don't see it on combat move or purchase phase while AI is thinking, so it's just during combat and non-combat moves which are now quite fast).

Could we make it show again for non-active players during placement? If I recall, I had made it update live as units are placed, so it's helpful to see when your opponents are placing units.

PR #6163, opened 2020-04-08
Revert "Fix potential server-side timeout"

DanVanAtta:
Reverts triplea-game/triplea#6137

Multiplayer game is crashing with:

```
Apr 07, 2020 7:02:19 PM games.strategy.engine.message.unifiedmessenger.UnifiedMessenger messageReceived
INFO: Missing latch. Did the method time out?
```

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/message/RemoteMethodCall.java
+++ b/game-core/src/main/java/games/strategy/engine/message/RemoteMethodCall.java
@@ -6,7 +6,6 @@ import java.io.IOException;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
 import java.lang.reflect.Method;
-import java.util.Arrays;
 import javax.annotation.Nullable;
 
 /** All the info necessary to describe a method call in one handy serializable package. */
@@ -95,12 +94,6 @@ public class RemoteMethodCall implements Externalizable {
   }
 
   private static String[] classesToString(final Class<?>[] classes, final Object[] args) {
-    if (args != null && classes.length != args.length) {
-      throw new IllegalArgumentException(
-          String.format(
-              "Classes and args arrays diff in length: %s, %s",
-              Arrays.toString(classes), Arrays.toString(args)));
-    }
     // as an optimization, if args[i].getClass == classes[i] then leave classes[i] as null
     // this will reduce the amount of info we write over the network in the common
     // case where the object is the same type as its arg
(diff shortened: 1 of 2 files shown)


DanVanAtta:
@RoiEXLab FYI

DanVanAtta:
Crash seems to have been time based, seems like the abort timer logic triggered at a bad time during normal play. I was advancing a multiplayer game to get to a placement phase to demo that issue and ran into this problem around the non-combat phase.

RoiEXLab:
@DanVanAtta I'm confused, the message you posted is just an info log, I thought only warnings and severe errors result in a dialog.
The simplest solution would be to decrease the logging level or remove the check altogether

RoiEXLab:
Regardless, the sever side changes should be fine. I'll redo them if you don't mind?

DanVanAtta:
Check if you can repro what I was seeing. Essentially playing against a bot
and after a timeout the game just crashed with an info message.

RoiEXLab:
@DanVanAtta Ah sorry for not clarifying:
The "crash" you're seeing is likely related to the client-side timeout of 20 seconds I built in.
If a network operation takes longer than 20 seconds, the latch is dropped and an exception is thrown.
If the method gets resolved after the timeout, the client notices that it does no longer has (or maybe never had in case of invalid data) a latch to release. In this case the error message is printed you mentioned.
So I completely removed all of the client side code changes (super easy because I simply had to cherry-pick commits) including the change that introduced this info message you reported. So it's impossible to trigger this message, simply because it doesn't exist anymore.

Here's my theory on why this issue occured and why I didn't catch it:
I know for a fact that the way the code flow for the individual players turns works is by using a lot of countdownlatches. So the code basically awaits a users action and then proceeds with the usual game logic, which is fine.
However we assumed that all network calls would trigger some code on the server resulting in a return value within 20 seconds, but what if the remote code itself waits for a latch/lock at some point, maybe for a remote players action?
In this case whenever a player takes longer than 20 seconds the timeout occurs and we get the problems you reported.
So why didn't I catch this issue? Remember my initial timeout of 1 minute? Obviously I never took longer than 1 minute to make some dummy moves just to check if there's no exception, I didn't re-check after lowering the timeout because what could possibly go wrong by reducing a timeout to 20 seconds?
But in the end this is just my theory, I haven't actually tried to prove it and I'm not planning to, so take my words with a grain of salt.

Regardless I checked my new branch on #6166 and everything seems to work fine there (which is expected), because all this branch changes is adding a pre-condition for the array problem we had and add a catch block to the handling part of the code which sends back any potential stacktrace properly.

DanVanAtta:
That sounds like a very plausible theory @RoiEXLab 
Hopefully in the medium term we'll be able to reduce the reliance on latches and have more explicit callbacks and timeouts.

Issue #6104, opened 2020-03-27
game froze clicking undo of chinese placements on 1941 v3


DanVanAtta:
**How can the problem be recreated?**
- Map = WWII 1941 v3
- multiplayer game on prerelease lobby on bot host (maybe pertinent?)
- placed chinese units
- clicked 'undo'
- game froze

**When the problem is fixed, what should have happened instead?**
- expecting no freezse

**Which Engine Version are you using?** 
prerelease 2.0.18509

```
"Reference Handler" #2 daemon prio=10 os_prio=0 cpu=113.50ms elapsed=7094.69s tid=0x00007f2ca8128800 nid=0xf9c waiting on condition  [0x00007f2c8c12b000]
   java.lang.Thread.State: RUNNABLE
	at java.lang.ref.Reference.waitForReferencePendingList(java.base@11.0.6/Native Method)
	at java.lang.ref.Reference.processPendingReferences(java.base@11.0.6/Reference.java:241)
	at java.lang.ref.Reference$ReferenceHandler.run(java.base@11.0.6/Reference.java:213)

   Locked ownable synchronizers:
	- None

"Finalizer" #3 daemon prio=8 os_prio=0 cpu=8.82ms elapsed=7094.69s tid=0x00007f2ca812a800 nid=0xf9d in Object.wait()  [0x00007f2c777fe000]
   java.lang.Thread.State: WAITING (on object monitor)
	at java.lang.Object.wait(java.base@11.0.6/Native Method)
	- waiting on <no object reference available>
	at java.lang.ref.ReferenceQueue.remove(java.base@11.0.6/ReferenceQueue.java:155)
	- waiting to re-lock in wait() <0x00000000802fc540> (a java.lang.ref.ReferenceQueue$Lock)
	at java.lang.ref.ReferenceQueue.remove(java.base@11.0.6/ReferenceQueue.java:176)
	at java.lang.ref.Finalizer$FinalizerThread.run(java.base@11.0.6/Finalizer.java:170)

   Locked ownable synchronizers:
	- None

"Signal Dispatcher" #4 daemon prio=9 os_prio=0 cpu=0.52ms elapsed=7094.67s tid=0x00007f2ca8130800 nid=0xf9e runnable  [0x0000000000000000]
   java.lang.Thread.State: RUNNABLE

   Locked ownable synchronizers:
	- None

[...]

PR #6170, opened 2020-04-10
Fix opcode mismatch

RoiEXLab:
Fix for #6104 and potentially other issues.
There is one thing that could potentially (in theory) still be an issue, but at this point it's highly unlikely it actually exists:
Interfaces that inherit their methods from `IRemote` and `IDelegate` don't specify an op-code for the methods defined in `IDelegate` and subinterfaces, just the ones specified in subinterfaces of `IRemote`. This decision was intentional and up until this point it does ot seem like delegate-only methods are ever invoked over the network, the client side would report this issue already.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  #6104
[] Other:   

## Testing

I verified #6104 could no longer be reproduced.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/remote/IAbstractPlaceDelegate.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/remote/IAbstractPlaceDelegate.java
@@ -55,4 +55,8 @@ public interface IAbstractPlaceDelegate extends IAbstractMoveDelegate<UndoablePl
    */
   @RemoteActionCode(9)
   Collection<Territory> getTerritoriesWhereAirCantLand();
+
+  @RemoteActionCode(17)
+  @Override
+  String undoMove(int moveIndex);
 }
--- a/game-core/src/main/java/games/strategy/triplea/delegate/remote/IAbstractPlaceDelegate.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/remote/IAbstractPlaceDelegate.java
@@ -19,6 +19,7 @@ public interface IAbstractPlaceDelegate extends IAbstractMoveDelegate<UndoablePl
   @RemoteActionCode(13)
   String placeUnits(Collection<Unit> units, Territory at, BidMode bidMode);
 
+  /** Convenience method for testing. Never called over the network. */
   default String placeUnits(final Collection<Unit> units, final Territory at) {
     return placeUnits(units, at, BidMode.NOT_BID);
   }
--- a/game-core/src/main/java/games/strategy/triplea/delegate/remote/IAbstractPlaceDelegate.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/remote/IAbstractPlaceDelegate.java
@@ -19,7 +19,6 @@ public interface IAbstractPlaceDelegate extends IAbstractMoveDelegate<UndoablePl
   @RemoteActionCode(13)
   String placeUnits(Collection<Unit> units, Territory at, BidMode bidMode);
 
-  @RemoteActionCode(12)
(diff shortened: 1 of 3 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/remote/IAbstractPlaceDelegate.java
```diff
@@ -19,7 +19,6 @@
   @RemoteActionCode(13)
   String placeUnits(Collection<Unit> units, Territory at, BidMode bidMode);
 
-  @RemoteActionCode(12)
```
RoiEXLab, line 22: Technically not needed, this method is only ever invoked in test code, so the annotation is redundant

ron-murhammer: Would it be a good idea to add a comment or test annotation to this?

RoiEXLab: Comment would be a good idea.

DanVanAtta:
@RoiEXLab thanks for the update to fix the undo problem. I'm wondering if can we be pretty certain we do not have an opcode mismatch problem anywhere else? Is there anyway to find those problems, have you double checked, and do we need to simply rely on manual testing to catch anything that maybe is still an issue?

RoiEXLab:
@DanVanAtta It's pretty simple to identify if there is a problem for a given interface, the list on my gist gives a complete overview of what the "op-codes" should be. Whereever 2 interfaces declare the same method (because of inheritance, which isn't part of the data, but can be checked manually which I did) and have different "op-codes" for it the problem exists.
The only real challenge is the sheer amount of methods that exist, which is the reason I have to automate some of that "identifying" for me, potentially introducing a point of failure.

If you [compare my list](https://github.com/triplea-game/triplea/issues/6104#issuecomment-612032593) with the methods that are actually changed in this PR, you'll notice that there are a lot of methods missing.
The reason for this is that 90% of the methods listed there are actually inherited from (sub)interfaces of `IDelegate`, which don't seem to be used at all over the network. At this point we would've noticed that we needed "op-codes" for them because there are a lot of methods for delegates that would instantly throw an exception, but we haven't seen any exceptions so far so that seems safe.
Another 5% of the methods just happen to have the same signature, but aren't actually related to each other, so no action required there, they already have different codes declared for them.
So what you end up seeing in this PR is the remaining 5% of the methods that could be problematic, after semi-manual review.

TL;DR: There is a risk we could've missed a method. but we can easily re-check if we are unsure.

DanVanAtta:
Anything that can give us full confidence everything works. Manual testing
gives a false sense of security, manual testing is just never going to be
comprehensive enough. If we have 1% of it wrong, we'll have a problem that
is going to be worthy of a rollback, it needs to be right.

RoiEXLab:
@DanVanAtta Thinking about it for a while and there might be a super easy way to check if everything is right.
If we transformed my gist with all the relevant opcodes a little bit so ever line has the format `class:method_name:arg1,arg2,...,argn:code` we could simply verify that all opcodes match (or are simply not present if removed on purpose) via reflection.
What do you think? Would it be worth baking that into a unit test?

DanVanAtta:
I'm not sure how that could be verified by reflection, sounds plausible and that you understand the verification better than I do :grin: 

A unit test to verify that network operations work would be good to have. I built a prototype branch a few nights ago that I'll try to get linked to this PR as soon as I switch computers here...

Issue #6173, opened 2020-04-10
The Bad Word function does not work.


prastle:
**How can the problem be recreated?**
  Access the admin toolbox and add a "bad word".
Rejoin the lobby as anonymous login using that bad word.  The name is allowed to join it should not be.

DanVanAtta:
Fixed in: https://github.com/triplea-game/triplea/pull/6177

PR #6177, opened 2020-04-11
Apply not nice username checks

DanVanAtta:
This update prevents user from taking names contain a 'bad-word' entry are listed in the banned-username table.

commit 6addd4ece79f842d278eea17299deab362db59c5

    Validate usernames do not match banned usernames or bad words
    
    This update:
    - adds name validation to login, (it was already present
      for account creation). Name validation asserts that
      a given name does not contain a bad word.
    - adds banned username check to name validation. This check
      verifies that a given name does not match a banned username.

commit 628a59a6b584e098582e7587ce9d92f80cb8480c

    Split username is already taken check from login
    
    If we check if a username is already taken when doing
    a registered login, we will see the name is always taken!
    This update removes the username is already taken check
    from username validation and moves the check to anonymous
    login and user registration.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix: https://github.com/triplea-game/triplea/issues/6173 
[] Other:   

## Testing

Tested locally when registration and login are blocked based on data in banned usernames table or bad words (and was not previously)

## Additional Review Notes

Would like to comment that the extensive automated spitfire tests really helped here. I completely missed, and arguably it was not completely obvious, that the username is taken check was going to break registered logins.

### DIFF
--- a/http-server/src/main/java/org/triplea/modules/user/account/login/LoginModule.java
+++ b/http-server/src/main/java/org/triplea/modules/user/account/login/LoginModule.java
@@ -17,6 +17,7 @@ import org.triplea.domain.data.UserName;
 import org.triplea.http.client.lobby.login.LobbyLoginResponse;
 import org.triplea.http.client.lobby.login.LoginRequest;
 import org.triplea.modules.chat.Chatters;
+import org.triplea.modules.user.account.NameValidation;
 import org.triplea.modules.user.account.login.authorizer.anonymous.AnonymousLoginFactory;
 import org.triplea.modules.user.account.login.authorizer.registered.RegisteredLogin;
 import org.triplea.modules.user.account.login.authorizer.temp.password.TempPasswordLogin;
@@ -29,6 +30,7 @@ class LoginModule {
   @Nonnull private Consumer<LoginRecord> accessLogUpdater;
   @Nonnull private final Function<LoginRecord, ApiKey> apiKeyGenerator;
   @Nonnull private final UserJdbiDao userJdbiDao;
+  @Nonnull private final Function<String, Optional<String>> nameValidation;
 
   public static LoginModule build(final Jdbi jdbi, final Chatters chatters) {
     return LoginModule.builder()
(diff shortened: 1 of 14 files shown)


http-server/src/main/java/org/triplea/modules/user/account/login/LoginModule.java
```diff
@@ -108,7 +116,6 @@ private ApiKey recordAnonymousLoginAndGenerateApiKey(
     return recordLoginAndGenerateApiKey(loginRequest, systemId, playerChatId, ip, false);
   }
 
-  // TODO: Project#12 also update access log
   @SuppressWarnings("unused")
```
RoiEXLab, line 119: Is this method still unused?

http-server/src/main/java/org/triplea/modules/user/account/login/authorizer/anonymous/AnonymousLoginFactory.java
```diff
@@ -1,21 +1,6 @@
...
-
-  public static Function<UserName, Optional<String>> build(
-      final Jdbi jdbi, final Chatters chatters) {
-    return AnonymousLogin.builder()
-        .chatters(chatters)
-        .userJdbiDao(jdbi.onDemand(UserJdbiDao.class))
-        .build();
-  }
-}
+public class AnonymousLoginFactory {}
```
RoiEXLab, line 6: I have the sneaking suspicion you forgot to remove this class 😉

http-server/src/main/java/org/triplea/modules/user/account/login/LoginModule.java
DanVanAtta: Looks like no, not sure how long the suppression was out of date for, looks like perhaps for some time.

http-server/src/main/java/org/triplea/modules/user/account/login/authorizer/anonymous/AnonymousLoginFactory.java
DanVanAtta: Twice even :grin: 
Was also missed earlier when we updated factory class pattern to be factory methods.

PR #4077, opened 2018-09-21
Performance improvements

RoiEXLab:
## Overview
Follow-up to #4068

## Functional Changes
None. Perhaps a little "fix" I'll point out in the diff.

## Manual Testing Performed
Route is drawn faster than before.

### Disclaimer
The values are not really consistent. I added a print statement to print the ms and went with the mouse in circles over the map while holding a unit.
So there's a little bit of randomness involved.
Also before "direct routes" were calculated outside of the RouteFinder, now they are calculated within this class, so this shifts the performance a little bit in favour of this PR as well.

Average route calculation time before #4068 : 9.402ms
Average route calculation time after this PR: 4.93545ms

PR #6187, opened 2020-04-11
Optimize findMaxLandMassSize() which can take 10% of AI time.

asvitkine:
Optimize findMaxLandMassSize() which can take 10% of AI time.
    
This change optimizes findMaxLandMassSize() by reducing the amount of work needed to do. The main difference is we now keep track of land territories already processed so that we don't need to start new searches from them.
    
This creates a new BreadthFirstSearch class that takes out the BFS implementation I implemented earlier for finding capitals, so that it can be re-used here.
    
It also fixes a correctness issue with finding the max land size - the previous search would only search up to a distance of 6. The new implementation does not have this limitation.
    
On Domination 1914 map, which has a lot of territories, this saves about 50 seconds of CPU time over a round of an all-AI game.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[X] Other: Optimization

## Testing
Tested manually on a game of Domination 1914 and profiled the game.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
+++ b/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
@@ -63,12 +63,11 @@ public class BreadthFirstSearch {
     int currentDistance = 0;
     Territory lastTerritoryAtCurrentDistance = territoriesToCheck.peekLast();
     while (!territoriesToCheck.isEmpty()) {
-      final Territory node = territoriesToCheck.removeFirst();
-      visitNeighbors(node);
+      final Territory territory = checkNextTerritory();
 
       // If we just processed the last territory at the current distance, increment the distance
       // and set the territory at which we need to update it again to be the last one added.
-      if (ObjectUtils.referenceEquals(node, lastTerritoryAtCurrentDistance)) {
+      if (ObjectUtils.referenceEquals(territory, lastTerritoryAtCurrentDistance)) {
         currentDistance++;
         if (!shouldContinueSearch(currentDistance)) {
           return;
@@ -78,7 +77,8 @@ public class BreadthFirstSearch {
     }
   }
 
-  private void visitNeighbors(final Territory territory) {
+  private Territory checkNextTerritory() {
+    final Territory territory = territoriesToCheck.removeFirst();
     // Note: We don't pass cond to getNeighbors() because that implementation is much slower.
     for (final Territory neighbor : map.getNeighbors(territory)) {
       if (cond.test(neighbor) && visited.add(neighbor)) {
@@ -86,5 +86,6 @@ public class BreadthFirstSearch {
         visit(neighbor);
       }
     }
+    return territory;
   }
 }
--- a/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
+++ b/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
@@ -1,0 +1,86 @@
+package games.strategy.engine.data.util;
+
+import games.strategy.engine.data.GameMap;
+import games.strategy.engine.data.Territory;
+import games.strategy.triplea.delegate.Matches;
+import java.util.ArrayDeque;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Predicate;
+import lombok.Getter;
+import org.triplea.java.ObjectUtils;
+
+/**
+ * Implements Breadth First Search (BFS) to traverse / find territories. Since the search criteria
+ * varies depending on the use case, the class is designed to be sub-classed, with methods visit()
+ * and shouldContinueSearch() that can be overridden to customize the behavior.
+ */
+public class BreadthFirstSearch {
+  private final GameMap map;
+  @Getter private final Set<Territory> visited;
+  private final ArrayDeque<Territory> territoriesToCheck;
+  private final Predicate<Territory> cond;
+
+  public BreadthFirstSearch(final Territory startTerritory, final Predicate<Territory> cond) {
+    this.map = startTerritory.getData().getMap();
+    this.visited = new HashSet<Territory>(List.of(startTerritory));
+    this.territoriesToCheck = new ArrayDeque<Territory>(List.of(startTerritory));
+    this.cond = cond;
+  }
+
+  public BreadthFirstSearch(final Territory startTerritory) {
+    this(startTerritory, Matches.always());
+  }
+
+  /**
+   * Called when a new territory is encountered. Can be overridden to provide custom search
+   * behavior.
+   *
+   * @param territory The new territory.
+   */
+  public void visit(final Territory territory) {}
+
+  /**
+   * Called after all territories within the specified distance have been searched. Can be
+   * overridden to terminate the search.
+   *
+   * @param distanceSearched The current distance searched
+   * @return Whether the search should continue.
+   */
+  public boolean shouldContinueSearch(final int distanceSearched) {
+    return true;
+  }
+
+  /**
+   * Performs the search. It will end when either all territories have been visited or
+   * shouldContinueSearch() returns false.
+   */
+  public void search() {
+    // Since we process territories in order of distance, we can keep track of the last territory
+    // at the current distance that's in the territoriesToCheck queue. When we encounter it, we
+    // increment the distance and update lastTerritoryAtCurrentDistance.
+    int currentDistance = 0;
+    Territory lastTerritoryAtCurrentDistance = territoriesToCheck.peekLast();
+    while (!territoriesToCheck.isEmpty()) {
+      final Territory node = territoriesToCheck.removeFirst();
+      // Note: We don't pass cond to getNeighbors() because that implementation is much slower.
+      for (final Territory neighbor : map.getNeighbors(node)) {
+        if (cond.test(neighbor) && visited.add(neighbor)) {
+          territoriesToCheck.add(neighbor);
+          visit(neighbor);
+        }
+      }
+
+      // If we just processed the last territory at the current distance, increment the distance
+      // and set the territory at which we need to update it again to be the last one added.
+      if (ObjectUtils.referenceEquals(node, lastTerritoryAtCurrentDistance)) {
+        currentDistance++;
+        if (!shouldContinueSearch(currentDistance)) {
+          return;
+        }
+        lastTerritoryAtCurrentDistance = territoriesToCheck.peekLast();
+      }
+    }
+  }
+}
--- a/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
+++ b/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
@@ -12,14 +12,13 @@ import org.triplea.java.ObjectUtils;
 
 /**
  * Implements Breadth First Search (BFS) to traverse / find territories. Since the search criteria
- * varies depending on the use case, the class is designed to be sub-classed, with methods visit()
- * and shouldContinueSearch() that can be overridden to customize the behavior.
+ * varies depending on the use case, the class is designed to take a Visitor object with methods
+ * visit() and shouldContinueSearch() for customizing the behavior.
  */
-public class BreadthFirstSearch {
+public final class BreadthFirstSearch {
   public abstract static class Visitor {
     /**
-     * Called when a new territory is encountered. Can be overridden to provide custom search
-     * behavior.
+     * Called when a new territory is encountered.
      *
      * @param territory The new territory.
      */
@@ -40,13 +39,19 @@ public class BreadthFirstSearch {
   private final GameMap map;
   private final Set<Territory> visited;
   private final ArrayDeque<Territory> territoriesToCheck;
-  private final Predicate<Territory> cond;
+  private final Predicate<Territory> neighborCondition;
 
-  public BreadthFirstSearch(final Territory startTerritory, final Predicate<Territory> cond) {
+  /**
+   * @param startTerritory The territory from where to start the search.
+   * @param neighborCondition Condition that neighboring territories must match to be considered
+   *     neighbors.
+   */
+  public BreadthFirstSearch(
+      final Territory startTerritory, final Predicate<Territory> neighborCondition) {
     this.map = startTerritory.getData().getMap();
     this.visited = new HashSet<>(List.of(startTerritory));
     this.territoriesToCheck = new ArrayDeque<>(List.of(startTerritory));
-    this.cond = cond;
+    this.neighborCondition = neighborCondition;
   }
 
   public BreadthFirstSearch(final Territory startTerritory) {
@@ -82,9 +87,9 @@ public class BreadthFirstSearch {
 
   private Territory checkNextTerritory(final Visitor visitor) {
     final Territory territory = territoriesToCheck.removeFirst();
-    // Note: We don't pass cond to getNeighbors() because that implementation is much slower.
+    // Note: The condition isn't passed to getNeighbors() because that implementation is very slow.
     for (final Territory neighbor : map.getNeighbors(territory)) {
-      if (cond.test(neighbor) && visited.add(neighbor)) {
+      if (neighborCondition.test(neighbor) && visited.add(neighbor)) {
         territoriesToCheck.add(neighbor);
         visitor.visit(neighbor);
       }
--- a/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
+++ b/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
@@ -64,13 +64,7 @@ public class BreadthFirstSearch {
     Territory lastTerritoryAtCurrentDistance = territoriesToCheck.peekLast();
     while (!territoriesToCheck.isEmpty()) {
       final Territory node = territoriesToCheck.removeFirst();
-      // Note: We don't pass cond to getNeighbors() because that implementation is much slower.
-      for (final Territory neighbor : map.getNeighbors(node)) {
-        if (cond.test(neighbor) && visited.add(neighbor)) {
-          territoriesToCheck.add(neighbor);
-          visit(neighbor);
-        }
-      }
+      visitNeighbors(node);
 
       // If we just processed the last territory at the current distance, increment the distance
       // and set the territory at which we need to update it again to be the last one added.
@@ -80,6 +74,16 @@ public class BreadthFirstSearch {
           return;
         }
         lastTerritoryAtCurrentDistance = territoriesToCheck.peekLast();
+      }
+    }
+  }
+
+  private void visitNeighbors(final Territory territory) {
+    // Note: We don't pass cond to getNeighbors() because that implementation is much slower.
+    for (final Territory neighbor : map.getNeighbors(territory)) {
+      if (cond.test(neighbor) && visited.add(neighbor)) {
+        territoriesToCheck.add(neighbor);
+        visit(neighbor);
       }
     }
   }
--- a/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
+++ b/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
@@ -23,7 +23,7 @@ public class BreadthFirstSearch {
      *
      * @param territory The new territory.
      */
-    public abstract void visit(final Territory territory);
+    public abstract void visit(Territory territory);
 
     /**
      * Called after all territories within the specified distance have been searched. Can be
--- a/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
+++ b/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
@@ -8,7 +8,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.function.Predicate;
-import lombok.Getter;
 import org.triplea.java.ObjectUtils;
 
 /**
@@ -17,8 +16,29 @@ import org.triplea.java.ObjectUtils;
  * and shouldContinueSearch() that can be overridden to customize the behavior.
  */
 public class BreadthFirstSearch {
+  public abstract static class Visitor {
+    /**
+     * Called when a new territory is encountered. Can be overridden to provide custom search
+     * behavior.
+     *
+     * @param territory The new territory.
+     */
+    public abstract void visit(final Territory territory);
+
+    /**
+     * Called after all territories within the specified distance have been searched. Can be
+     * overridden to terminate the search.
+     *
+     * @param distanceSearched The current distance searched
+     * @return Whether the search should continue.
+     */
+    public boolean shouldContinueSearch(final int distanceSearched) {
+      return true;
+    }
+  }
+
   private final GameMap map;
-  @Getter private final Set<Territory> visited;
+  private final Set<Territory> visited;
   private final ArrayDeque<Territory> territoriesToCheck;
   private final Predicate<Territory> cond;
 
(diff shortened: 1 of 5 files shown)


game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
```diff
@@ -0,0 +1,90 @@
...
+   */
+  public boolean shouldContinueSearch(final int distanceSearched) {
+    return true;
+  }
+
+  /**
+   * Performs the search. It will end when either all territories have been visited or
+   * shouldContinueSearch() returns false.
+   */
+  public void search() {
```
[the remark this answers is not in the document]
asvitkine: Don't see a good way of reducing further. Note: Overall it's still a reduction since the function where this code was originally had higher complexity.

game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
```diff
@@ -0,0 +1,91 @@
...
+          return;
+        }
+        lastTerritoryAtCurrentDistance = territoriesToCheck.peekLast();
+      }
+    }
+  }
+
+  private Territory checkNextTerritory() {
+    final Territory territory = territoriesToCheck.removeFirst();
+    // Note: We don't pass cond to getNeighbors() because that implementation is much slower.
```
RoiEXLab, line 82: Perhaps worth clarifying:
The reason the implementation is slower is because it tries to parallelize checking neighbours, which adds too much overhead for the ususally small amount of neighbours + it actually copies the collection instead of just returning it.

I'd suggest to clarify that a bit why it is much slower in this comment.
Maybe you can even drop the parallelism entirely, I'm not really convinced it actually provides a speed boost elsewhere.
I introduced this change in #4077 (apparently) because it yielded some performance gains for route-drawing, but I haven't really tested anything else as far as I can remember.

The `GameMap` class is a mess (I'm currently looking at it) and uses way too much recursion in combination with code constructs that aren't really supposed to be run recursively like this (Streams that are collected and then re-streamed in another layer add way too much overhead)

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
```diff
@@ -177,19 +176,31 @@ public static double findTerritoryAttackValue(
...
+
+      LandMassFinder(final Territory startTerritory) {
+        super(startTerritory, cond);
+      }
+
+      @Override
+      public void visit(final Territory territory) {
+        landMassSize++;
+      }
+    }
```
RoiEXLab, line 193: Ok, I didn't know you could nest classes inside methods 😮

RoiEXLab:
Overall I like the changes @asvitkine 👍 
Performance improvements by faster algorithms are always a good thing

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
```diff
@@ -177,19 +176,31 @@ public static double findTerritoryAttackValue(
...
-                        t, 6, ProMatches.territoryCanPotentiallyMoveLandUnits(player, data))
-                    .size();
-        if (landMassSize > maxLandMassSize) {
-          maxLandMassSize = landMassSize;
+      if (!t.isWater() && !fullVisited.contains(t)) {
+        final LandMassFinder finder = new LandMassFinder(t);
+        finder.search();
+        fullVisited.addAll(finder.getVisited());
+        if (finder.landMassSize > maxLandMassSize) {
+          maxLandMassSize = finder.landMassSize;
```
RoiEXLab, line 203: In java getters and setters are the preffered way of modifying fields, even on this small scale.
You can use lomboks `@Getter` and `@Setter` annotations if you want

game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
asvitkine: I haven't dug into why it's slower. I just measured things after my initial refactoring that did this and it was slower. I think we can rewrite that later, but didn't want to expand the scope of this change.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
DanVanAtta: @RoiEXLab I'm not positive that is always the case. Within a class for example using getter/setter can be brittle as any child overrides can unintentionally change many methods. Essentially java was designed to expect inheritance as a default, then effective java came along and basically said to disallow it by default or else explicitly design for it (complex).

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
```diff
@@ -448,39 +459,26 @@ private static double findWaterValue(
...
-    int distance = 0;
-    Territory lastTerritoryAtCurrentDistance = startTerritory;
-    while (!territoriesToCheck.isEmpty()) {
-      final Territory node = territoriesToCheck.removeFirst();
-      for (final Territory neighbor : map.getNeighbors(node)) {
-        if (visited.add(neighbor)) {
-          territoriesToCheck.add(neighbor);
-          if (enemyCapitalsAndFactories.contains(neighbor)) {
-            found.add(neighbor);
+    final BreadthFirstSearch bfs =
```
DanVanAtta, line 463: What do you think of using a var here in order to spell out the variable name instead? eg:
`var breadthFirstSearch =  new BreadthFirstSearch(startTerritory) {`

game-core/src/test/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtilsTest.java
```diff
@@ -62,4 +64,14 @@ void testFindNearbyEnemyCapitalsAndFactoriesNoneWithinDistance9() {
         result,
         containsInAnyOrder(northJapan));
   }
+
+  @Test
```
DanVanAtta, line 68: Good to see the test case, can we increase the number of tests and thoroughness? Are the edge cases tested?

game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
```diff
@@ -0,0 +1,91 @@
...
+    this(startTerritory, Matches.always());
+  }
+
+  /**
+   * Called when a new territory is encountered. Can be overridden to provide custom search
+   * behavior.
+   *
+   * @param territory The new territory.
+   */
+  public void visit(final Territory territory) {}
```
DanVanAtta, line 42: Can we make this method abstract? Overriding concrete methods can get pretty complex in a hurry. Another strategy that would avoid inheritance completely would be to pass in a constructor object that defines this behavior. EG:

```
private final Consumer<Territory> visitor;

  public void visit(final Territory territory) {
      visitor.accept(terrritory);
}
````

Looking at the other methods, it looks like we might be providing a lot of default behavior but any/all of it can be overriden. If this behavior were injected it could be more deterministic and easier to follow.  Essentially you would know have to know about instance types and the states of objects to know how they would function, WDYT?

File: game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java [hunk 8]
```diff
@@ -177,19 +176,31 @@ public static double findTerritoryAttackValue(
...
-                + data.getMap()
-                    .getNeighbors(
-                        t, 6, ProMatches.territoryCanPotentiallyMoveLandUnits(player, data))
-                    .size();
-        if (landMassSize > maxLandMassSize) {
-          maxLandMassSize = landMassSize;
+      if (!t.isWater() && !fullVisited.contains(t)) {
+        final LandMassFinder finder = new LandMassFinder(t);
+        finder.search();
+        fullVisited.addAll(finder.getVisited());
```
DanVanAtta, line 201: What do you think of having `finder.seach()` returning the `visited` list? This would avoid the void method call, and avoid the user of `finder` from knowing about the implementation detail that `finder.search` sets a value that can be returned by `getVisited` (arguably better/encapsulation)

DanVanAtta:
> Optimize findMaxLandMassSize() which can take 10% of AI time.

> On Domination 1914 map, which has a lot of territories, this saves about 50 seconds of CPU time over a round of an all-AI game.

@asvitkine does that imply you observed an AI time reduction from 500s to 450s?

game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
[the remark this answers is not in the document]
DanVanAtta: FWIW, this one was fixed:

> Method findNearbyEnemyCapitalsAndFactories has a Cognitive Complexity of 16 (exceeds 5 allowed). Consider refactoring.
Fixed

Going from 16 -> 6 is certainly the right direction :+1:

game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
RoiEXLab: No worries, just wanted to provide some extra information for documentation purposes.

game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.function.Predicate;
+import lombok.Getter;
+import org.triplea.java.ObjectUtils;
+
+/**
+ * Implements Breadth First Search (BFS) to traverse / find territories. Since the search criteria
+ * varies depending on the use case, the class is designed to be sub-classed, with methods visit()
+ * and shouldContinueSearch() that can be overridden to customize the behavior.
+ */
+public class BreadthFirstSearch {
```
RoiEXLab, line 19: Just a random thought I had.
This class would be super usefule to optimize `GameMap#getNeighbors(Set, Set, int, Predicate)`, which basically does the same thing here, just with less than optimal recursion.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
ron-murhammer: Yeah, not really sure how I feel about classes nested in a method either...

game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
DanVanAtta: Apologies @asvitkine if this is getting pedantic for you, just want to be sure the suggestion for how to avoid overriding concrete methods is somewhat clear so that you can respond or incorporate it.

Example:

```
class Parent {
  int computeSomething() {
       return 1;
   }
}

@Override
class Child extends Parent{
  int computeSomething() {
       return 10;
   }
}
```

In the above, there are really two behaviors, we can expand the inheritance tree to flatten the structure, ie instead of:
```
  P
    \
     C(1)
```

We get a graph that looks like this:
```
       P
     /   \
 C(1)   C(2)
```

That would look like this:

```
absract class Parent {
  abstract int computeSomething();
}

@Override
class Child1 extends Parent{
  int computeSomething() {
       return 1;
   }
}

@Override
class Child2 extends Parent{
  int computeSomething() {
       return 10;
   }
}
```

The other recommendation is to inject behavior via configuration, in which case there is no inheritance at all, eg:

```

@AllArgsBuilder
class Parent {
  private final Supplier<Integer> computation;
  int computeSomething() {
     return computation.get();
  }
}


class SomeClass {
   Parent child1 = new Parent(() -> 1);
   Parent child2 = new Parent(() -> 10);
}
```

DanVanAtta:
@asvitkine before pushing changes to this branch, please see if the CI not building this PR is potentially resolved by: https://github.com/triplea-game/triplea/pull/6200#issuecomment-612643835

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
asvitkine: In this case, it's accessing a what's essential a local variable. It can't be local to the method because anon or inner classes can't set primitives (i.e. can't have non-final variables). So in this case it's declared as a member of the inner class.

Another option is to declare an array or some other wrapper in the method itself as a local, e.g.

int maxLandMassSize[] = new int[1];

And have the inner class access maxLandMassSize[0], then it wouldn't need to be in the inner class (and the inner class could become anonymous).

But that seemed uglier, so I went with this.

By the way, if we structured the API without inheritance (e.g. the BFS class could take another interface instance to receive the new call backs), this doesn't solve this problem as the same problem exists. The extra object still needs to keep track of maxLandMassSize which is effectively something that's local to to the method and doesn't need a getter.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
asvitkine: Another option is to not expose visited() at all and have the caller's visit() method override keep track of this case.

game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
asvitkine: Yes, it's on my radar. I actually wrote the code for that, but it didn't help much with this optimization since the underlying algorithm was still bad (it was calling getNeighbors(, distance=6) on each territory). So I abandoned that to focus on the actual speed improvement which is an algorithm change.

However, when we have the new class in place, we can use it in more places like there. :)

game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
asvitkine: We could go with an interface or abstract class that has the two methods to override that are passed from the caller.  I agree that it tightens things up a bit more (i.e. the subclasses of that abstract can't do as much), so I can look at doing that if you prefer. 

Though it doesn't change things too much - the caller now just declares an anonymous class that extends the new abstract class/interface that has just the visit() and shouldContinueSearch() methods. Let me know if you prefer that approach.

My original thought was to keep it simpler to start with, so that as we apply this to more cases (like getNeighbours(distance) suggestion and possibly getRoute()), we can figure out what the API surface would need to be like to support them and evolve as necessary.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
asvitkine: It was to support tracking of landMassSize while keeping the logic self contained to the method. The other option is an anon class and have:

final int maxLandMassSize[] = new int[1];

in the method scope so that the inner class can write to maxLandMassSize[0] (since you can't write to an int. This seemed cleaner than that.

asvitkine:
> > Optimize findMaxLandMassSize() which can take 10% of AI time.
> 
> > On Domination 1914 map, which has a lot of territories, this saves about 50 seconds of CPU time over a round of an all-AI game.
> 
> @asvitkine does that imply you observed an AI time reduction from 500s to 450s?

Roughly, yes.

I just re-did the measurement with all three of my optimizations applied and I'm seeing a reduction of total round time for that saved game of 110 seconds (~560s -> ~440s) and the time spent by each battle calc worker from ~120 seconds to ~90 seconds.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
DanVanAtta: Would `AtomicReference` perhaps help? I think that can be templated to an integer, eg: `AtomicReference<Integer>`

asvitkine: It would work, but the atomicity is not needed and may be confusing to readers.

game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
DanVanAtta: > Though it doesn't change things too much - the caller now just declares an anonymous class that extends the new abstract class/interface that has just the visit() and shouldContinueSearch() methods rather than extending BFS. Let me know if you prefer that approach.

Assuming it can be done cleanly, I think that sounds good/preferable :+1: 

It sounds like that would avoid a concrete implementation from having an override. Such scenarios, concrete implementations having an override is usually where inheritance starts to become quite bad.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
DanVanAtta: That also sounds like a reasonably alternative. 

If I understand right, there would be no side-effects as the data state would be self-contained to the BFS implementation, do I have that right? If that can be done cleanly, it sounds good.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
DanVanAtta: `AtomicReference` AFAIK is preferred way to avoid the array of size 1:
`final int maxLandMassSize[] = new int[1];`

There might be other solutions to be had too..

asvitkine: Do you have a reference for why it's preferred?

It seems an antipattern to me. We're using something that was not meant for the purpose and has a semantic meaning (i.e. for use with multithreading) for something else. This both confuses the semantic meaning (a reader may think the code is multithreaded when it's not) and also can have performance implications.

Imho, the named inner class is the cleanest way to do without a dedicated MutableInt class (like exists in apache commons).

I searched a few sources to see if there's a reason to prefer AtomicInteger, but haven't found anything convincing:
https://www.techiedelight.com/pass-integer-reference-java/
https://stackoverflow.com/questions/319765/accessing-inner-anonymous-class-members

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
asvitkine: Done.

game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
asvitkine: Done.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
asvitkine: Done.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
DanVanAtta: Either way, `int[]` or `AtomicReference<Integer>` or `AtomicInteger`, for the sake of using a final value that is mutable, it's not the intended usage of any of those.

I know I've seen `AtomicReference` used for that before, don't have any examples stating it's a best practice (AFAIK none of it is necessarily a best practice though as we are hacking around a variable that needs to be final).

asvitkine: Well, the other approach was to have an inner class with a member int for, which is what my original change did. I would argue that is less hacky than the others, but then people didn't like a non-anomymous inner class inside a method.  ¯\_(ツ)_/¯

I went with int[] now.

DanVanAtta: Apologies as you're suffering from too many reviewers :grin: 
We've different tastes.
I would probably look for perhaps a 3rd option.
I respect avoiding code churn, `int[]` or `AtomicReference` are not that much different, not a big deal for me.

DanVanAtta:
@asvitkine 
```
> Task :game-core:checkstyleMain

[ant:checkstyle] [WARN] /home/travis/build/triplea-game/triplea/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java:26:32: Redundant 'final' modifier. [RedundantModifier]
```

game-core/src/test/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtilsTest.java
DanVanAtta: @asvitkine thoughts on if there are additional test cases to make sure the testing is thorough?
Perhaps the not found case, or the find at the zero distance, or cases where we should find but it's past max distance?

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtils.java
```diff
@@ -448,39 +456,25 @@ private static double findWaterValue(
...
+              @Override
+              public void visit(final Territory territory) {
+                if (enemyCapitalsAndFactories.contains(territory)) {
+                  found.add(territory);
+                }
+              }
+
+              @Override
+              public boolean shouldContinueSearch(final int distanceSearched) {
+                if (distanceSearched >= MIN_FACTORY_CHECK_DISTANCE && !found.isEmpty()) {
```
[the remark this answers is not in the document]
asvitkine: Done.

game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
```diff
@@ -0,0 +1,94 @@
...
+  public abstract static class Visitor {
+    /**
+     * Called when a new territory is encountered. Can be overridden to provide custom search
```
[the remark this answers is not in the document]
asvitkine: Done.

game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
```diff
@@ -0,0 +1,94 @@
...
+     */
+    public boolean shouldContinueSearch(final int distanceSearched) {
+      return true;
+    }
+  }
+
+  private final GameMap map;
+  private final Set<Territory> visited;
+  private final ArrayDeque<Territory> territoriesToCheck;
+  private final Predicate<Territory> cond;
```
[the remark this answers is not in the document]
asvitkine: Changed to `neighborCondition` and added Javadoc for the constructor param.

game-core/src/test/java/games/strategy/triplea/ai/pro/util/ProTerritoryValueUtilsTest.java
asvitkine: findMaxLandMassSize() doesn't really have parameters, so the only way to try different cases is with different maps.

I added cases for Revised and Minimap (which is a single small continent). For the latter, I added its XML file to test data. Let me know if you think this is OK.

game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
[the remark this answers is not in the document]
DanVanAtta: @asvitkine I'm not seeing the latest updates, do they need to be pushed perhaps?

asvitkine:
Done, sorry about that!

game-core/src/test/java/games/strategy/triplea/xml/TestMapGameData.java
```diff
@@ -38,7 +38,9 @@
 
   GAME_EXAMPLE("GameExample.xml"),
 
-  TWW("Total_World_War_Dec1941.xml");
+  TWW("Total_World_War_Dec1941.xml"),
+
+  MINIMAP("minimap.xml");
```
DanVanAtta, line 43: side-note, I'm thinking to propose a convention that we end enums with a trailing comma and semi on the next line. This way adding elements adds just a new line. 

In this case this would be:
```
MINIMAP("minimap.xml"),
;
```

Another side-comment, the tests that require full map data are not necessarily ideal. They involve lots of moving pieces, it's not a proper unit test if we use them. There are some additional drawbacks. Long story short, we want to in the short term see if we consolidate how many test XMLs are in use, use as few as possible, eventually get to a place where game rules can be tested ad-infinitum without an actual XML

DanVanAtta:
@asvitkine 

```
> Task :game-core:checkstyleTest

[ant:checkstyle] [WARN] /home/travis/build/triplea-game/triplea/game-core/build/resources/test/minimap.xml:526:1: Line contains a tab character. [FileTabCharacter]

[ant:checkstyle] [WARN] /home/travis/build/triplea-game/triplea/game-core/build/resources/test/minimap.xml:527:1: Line contains a tab character. [FileTabCharacter]

> Task :game-core:checkstyleTest FAILED

> Task :game-headed-javafx:compileJava

* What went wrong:

Execution failed for task ':game-core:checkstyleTest'.

> Checkstyle rule violations were found. See the report at: file:///home/travis/build/triplea-game/triplea/game-core/build/reports/checkstyle/test.html

  Checkstyle files with violations: 1

  Checkstyle violations by severity: [warning:2]
```

asvitkine:
Thanks, fixed.

DanVanAtta:
side note:

> * Try to reduce complexity to appease codeclimate

One should not appease codeclimate, the commit comment is actually not very useful. What is done to reduce complexity? What is the change made? Appeasing codeclimate is a means to an end. For example "Fix checkstyles" is not useful either, but "improveme javadoc, fix completness of javadoc, fix whitespace formatting" are more useful.

PR #6188, opened 2020-04-11
Add op code test

RoiEXLab:
Requested in https://github.com/triplea-game/triplea/pull/6170#issuecomment-612167449

Here's the python script used to generate the test set, feel free to review that as well.
```python3
#!/usr/bin/env python3

import sys

def main(argv):
  methods = {}
  current_type = None
  with open(argv[1]) as f:
    for line in f:
      if line and line[0] != '\t':
        current_type = line.strip()
        continue
      split_line = line.split(':')
      method = split_line[0].strip()
      op_code = int(split_line[1].strip())
      split_method = method.split('(')
      method_name = split_method[0]
      raw_method_params = split_method[1][:-1]
      method_params = raw_method_params.split(', ') if raw_method_params else []
      print(op_code, current_type, method_name, *method_params, sep=',')

if __name__ == '__main__':
  main(sys.argv)
```
The file used by this script can be found here: https://gist.github.com/RoiEXLab/b63ad68a29438fc6152e509d6deb4add#file-op-codes-txt

I had to make 4 kinds of manual adjustments to the test set:
1. The class ChatParticipant was moved to a new package, so I had to adjust the test set accordingly
2. The method `IEditDelegate#changeTechTokens(GamePlayer, int)` was removed in #5992 so I had to remove this one as well. (see ade5cd5 )
3. I removed all methods that are inherited from `IDelegate`, with the help of regular expressions. Please double check c5ed015 to be sure I didn't accidentally remove a method that wasn't supposed to be removed.
4. I removed one annotation in #6170 so the tests have been adjusted accordingly

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[x] Other:   Unit Tests

## Testing

Well... No manual testing ^^


game-core/src/test/resources/required-op-codes.txt
```diff
@@ -0,0 +1,134 @@
+3,games.strategy.engine.chat.IChatChannel,speakerAdded,org.triplea.domain.data.ChatParticipant
```
RoiEXLab, line 1: The package name of `org.triplea.domain.data.ChatParticipant` was changed by hand

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
```diff
@@ -0,0 +1,49 @@
...
+      final Class<?>[] args =
+          IntStream.range(offset + 2, arguments.size())
+              .mapToObj(i -> arguments.get(i, Class.class))
+              .toArray(Class<?>[]::new);
+      try {
+        return clazz.getMethod(methodName, args);
+      } catch (final NoSuchMethodException e) {
+        throw new ArgumentsAggregationException("Invalid method specified", e);
+      }
+    }
```
RoiEXLab, line 47: This is kind of overkill, but I just learned about this awesome JUnit feature today so I had to try it, and it keeps the actual test code really clean IMO

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @ParameterizedTest
+  @CsvFileSource(resources = "/required-op-codes.txt")
+  void verifyCorrectOpCode(
+      final int opCode, @AggregateWith(MethodAggregator.class) final Method method) {
+    var remoteActionCode = method.getAnnotation(RemoteActionCode.class);
+
+    assertThat("No annotation present for " + method, remoteActionCode, is(notNullValue()));
+
+    assertThat("Invalid value for " + method, remoteActionCode.value(), is(opCode));
```
RoiEXLab, line 27: The reason strings are not really required here, but for some reason my IntelliJ doesn't display the "correct" method name that includes the parameters toString info.

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.junit.jupiter.params.aggregator.ArgumentsAggregator;
+import org.junit.jupiter.params.provider.CsvFileSource;
+
+public class RemoteActionCodeTest {
+
+  @ParameterizedTest
+  @CsvFileSource(resources = "/required-op-codes.txt")
+  void verifyCorrectOpCode(
+      final int opCode, @AggregateWith(MethodAggregator.class) final Method method) {
+    final var remoteActionCode = method.getAnnotation(RemoteActionCode.class);
```
DanVanAtta, line 23: `var` is perhaps confusing here, I'm not offhand 100% sure what the return type is of `getAnnotation`. I think it is probably `Class`, but not fully evident. Perhaps if the variable name were `remoteActionCodeClass`, it would be clear and would not benefit from type information.

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+    assertThat("No annotation present for " + method, remoteActionCode, is(notNullValue()));
```
DanVanAtta, line 25: Why is this expected to be true? Can that be clarified in the assertion comment somehow perhaps?
Perhaps we can rephrase to say something like `"Expected an annotation to present on method:  "  + method`, it's generally clearer to explain what and why you were expected rather than to rephrase the error condition. EG: `result was not 4` vs `Expected 2+2 to be 4`

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
```diff
@@ -0,0 +1,49 @@
...
+    assertThat("Invalid value for " + method, remoteActionCode.value(), is(opCode));
+  }
+
+  static class MethodAggregator implements ArgumentsAggregator {
+    @Override
+    public Method aggregateArguments(
+        final ArgumentsAccessor arguments, final ParameterContext context)
+        throws ArgumentsAggregationException {
+      final int offset = context.getIndex();
+      final Class<?> clazz = arguments.get(offset, Class.class);
```
DanVanAtta, line 36: What kind of class is this expected to be? Is it an annotated one with the remote method number? Perhaps would be more clear as something like `RemoteMethodAnnotatedClass`

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+    assertThat("Invalid value for " + method, remoteActionCode.value(), is(opCode));
+  }
+
+  static class MethodAggregator implements ArgumentsAggregator {
+    @Override
+    public Method aggregateArguments(
+        final ArgumentsAccessor arguments, final ParameterContext context)
+        throws ArgumentsAggregationException {
+      final int offset = context.getIndex();
```
DanVanAtta, line 35: Can this offset variable be clarified a bit more, what are we offsetting from? It might be possible to rename the variable, or some inline commentary could help explain it.

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
```diff
@@ -0,0 +1,49 @@
...
+    final var remoteActionCode = method.getAnnotation(RemoteActionCode.class);
+
+    assertThat("No annotation present for " + method, remoteActionCode, is(notNullValue()));
+
+    assertThat("Invalid value for " + method, remoteActionCode.value(), is(opCode));
+  }
+
+  static class MethodAggregator implements ArgumentsAggregator {
+    @Override
+    public Method aggregateArguments(
```
DanVanAtta, line 32: It seems like a javadoc could really help here. Ideally it would set up expectations for reading the method to help guide understanding and/or make reading the method just not needed to know what we expect to get back when calling this method.

game-core/src/test/resources/required-op-codes.txt
DanVanAtta, line 1: Is there any way to document the structure and contents of this file and how it was generated? If we somehow added a new remote method, will it be easy for someone to update this (and to know to update it?)

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
RoiEXLab: Well `Method#getAnnotation(Class)` returns null if no annotation of the specified type is available on the method.
The tests set obviously only contains the methods that are being used over the network, so they are required to have this annotation.

Perhaps it would help if we said `Method %s not annotated with @RemoteActionCode`?

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
RoiEXLab: Classes are never annotated with this annotation, only methods of interfaces are.
The class in question is an arbitrary subinterface of `IRemote` or `IChannelSubscriber`, containing methods that are annotated with the annotation.
I originally wanted to call it `interface`, but that's a reserved keyword too.

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
RoiEXLab: It's a bit hard to explain. Basically `ArgumentsAccessor` has access to the Arguments that _could_ be passed to a specific method but the opcode (the first column in our csv file) has nothing to do with the actual method.
We could hardcode this offset to be 1 (so we ignore the first column of the csv) but because we already "consume" this parameter by explicitly declaring it in the test method I thought it would make sense to just ignore as many columns as there are parameters before Method.

Might be confusing, but I hope this helps a little bit:
```java
void verifyCorrectOpCode(/* parameter context index 0*/final int opCode, @AggregateWith(MethodAggregator.class) /* parameter context index 1*/final Method method) { /* Test code */}

// What argumentsAccessor holds, based off csv content:
arguments.getInt(0); // Op code
arguments.get(1, Class.class); // Subinterface of IRemote of IChannelSubscriber
arguments.getString(2); // method name

arguments.get(3, Class.class); // start method parameter types
.
.
.
arguments.get(arguments.size() - 1, Class.class); // end method parameter types
```

game-core/src/test/resources/required-op-codes.txt
RoiEXLab: If you're searching for an inline way to document this file there probably isn't a good one.
CSV is a format that doesn't support comments unfortunately.

But if we documented this file in the test case via a javadoc that would work.
Basically it's just a list of method descriptors and their appropriate methods.
Updating this list ist simple too, just add another line below that matches the scheme:
```
op-code:full-interface-name:method-name:method-param-type-0:...;method-param-type-n
```

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
DanVanAtta: > The tests set obviously only contains the methods that are being used over the network, so they are required to have this annotation.

I think that might depend on the maintainers context, revisiting this at first, that may not obvious until you spend some time investigating that. Telling a maintainer directly that to begin with saves time, makes code more scalable as this could be one of a dozen tests or tasks that a person has to go through.

The updated message example is an improvement, I like it :+1:

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
DanVanAtta: Perhaps rename to  `InterfaceWithRemoteMethods`?

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
DanVanAtta: Can we move this clarification into the code somehow? It's always very sad when you have to look at the date of a merge to find a PR to then read comments to understand the code you're looking at.

game-core/src/test/resources/required-op-codes.txt
DanVanAtta: I like the suggestion. I think we probably could use a `docs/` file somewhere to describe the remote method interface mechanism and perhaps give a check-list of things you need to update when adding a remote method. This way we preserve context of the moving pieces, some code examples even how the remote method invocation works and the differences between hub invoke, broadcast and point to point would all be great as well (perhaps those items can come later as we iterate on the documentation, for now probably most important to have a checklist of what you would need to touch if you added a remote method invocation so we can be a bit more confident that this test will be updated going forward)

RoiEXLab:
@DanVanAtta Updated, let me know what you think about the documentation

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
```diff
@@ -14,33 +14,53 @@
 import org.junit.jupiter.params.aggregator.ArgumentsAggregator;
 import org.junit.jupiter.params.provider.CsvFileSource;
 
+/**
+ * Test class that verifies that remote operation codes didn't change unexpectedly due to some
+ * random change.
```
DanVanAtta, line 19: nit, can probably omit "due to some random change", it's not 100% informative. You could perhaps give examples of what such changes would be, otherwise shorter and more information dense is perhaps better.

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
```diff
@@ -14,33 +14,53 @@
...
+ *
+ * <p>Please update {@code /required-op-codes.txt} accordingly in case the remote interfaces are
```
DanVanAtta: A similar kind of comment is probably good to have on the annotation itself, to tell a person who is exploring adding that annotation to look here to add it to a test.

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
```diff
@@ -14,33 +14,53 @@
...
+   * Helper class that aggregates the fields found in the csv of test data to a {@link Method}. In
+   * order for this to work as expected, the CSVs last n + 2 columns must follow the format:
+   *
+   * <p>{@code interface-name,method-name,method-param-type-0,...,method-param-type-n}
+   *
+   * <p>Note that due to the variable length of the type arguments
+   * {@code @AggregateWith(MethodAggregator.class)} can only ever be used on the last parameter of a
+   * test method to work properly, otherwise it might pickup the wrong arguments.
+   */
   static class MethodAggregator implements ArgumentsAggregator {
```
DanVanAtta: :+1:

DanVanAtta:
2 new comments @RoiEXLab , feel free to merge if and/or after they can be reasonably addressed.

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
RoiEXLab: I agree that "some random change" is kinda slang at this point.
What I wanted to express is that if you do some refactoring and mess with the interface structure, you might end up overriding another method unintentionally which could change the codes.
Or in case you change the codes on purpose, you need to consider updating the test cases as well.
Any suggestions on how to rephrase th sentence?

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
RoiEXLab: I'm always a bit hesitant to make references to tests in non-test code, but a small note won't hurt

game-core/src/test/java/games/strategy/engine/message/RemoteActionCodeTest.java
DanVanAtta:
```
 * Test class that verifies that remote operation codes didn't change unexpectedly. This can happen for a number of reasons, eg:
* - interface structure changed by overriding a method and changing the opcode that way
* - ...
```

> Or in case you change the codes on purpose, you need to consider updating the test cases as well.

That likely could be a  warning/comment on the annotation.

PR #6189, opened 2020-04-11
Make Territory.getUnits() return an unmodifiable list.

asvitkine:
Make Territory.getUnits() return an unmodifiable list.

Previously, it was returning a copy, which was actually quite expensive given how often this was called. On a Domination 1914 game, with all AIs, this took 100 seconds of CPU time (across different threads since it's used in battle sim) over a single round.

A few places that need a copy are updated to do the copy on their side.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[X] Other:  Optimization

## Testing

Ran an all-AI round without errors on Domination 1914 and on WW2 Global.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/AbstractPlaceDelegate.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/AbstractPlaceDelegate.java
@@ -1652,14 +1652,13 @@ public abstract class AbstractPlaceDelegate extends BaseTripleADelegate
     if (to == null) {
       return new ArrayList<>();
     }
-    final Collection<Unit> unitsInTo = to.getUnits();
     final Collection<Unit> unitsPlacedAlready = getAlreadyProduced(to);
     if (Matches.territoryIsWater().test(to)) {
       for (final Territory current : getAllProducers(to, player, null, true)) {
         unitsPlacedAlready.addAll(getAlreadyProduced(current));
       }
     }
-    final Collection<Unit> unitsAtStartOfTurnInTo = new ArrayList<>(unitsInTo);
+    final Collection<Unit> unitsAtStartOfTurnInTo = new ArrayList<>(to.getUnits());
     unitsAtStartOfTurnInTo.removeAll(unitsPlacedAlready);
     return unitsAtStartOfTurnInTo;
   }
@@ -1668,7 +1667,7 @@ public abstract class AbstractPlaceDelegate extends BaseTripleADelegate
     if (to == null) {
       return new ArrayList<>();
     }
-    final Collection<Unit> unitsInTo = to.getUnits();
+    final Collection<Unit> unitsInTo = new ArrayList<>(to.getUnits());
     final Collection<Unit> unitsAtStartOfStep = unitsAtStartOfStepInTerritory(to);
     unitsInTo.removeAll(unitsAtStartOfStep);
     return unitsInTo;
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
@@ -268,7 +268,7 @@ public class MustFightBattle extends DependentBattle implements BattleStepString
   @Override
   public List<Unit> getRemainingAttackingUnits() {
     final List<Unit> remaining = new ArrayList<>(attackingUnitsRetreated);
-    final Collection<Unit> unitsLeftInTerritory = battleSite.getUnits();
+    final Collection<Unit> unitsLeftInTerritory = new ArrayList<>(battleSite.getUnits());
     unitsLeftInTerritory.removeAll(killed);
     remaining.addAll(
         CollectionUtils.getMatches(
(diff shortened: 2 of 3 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/AbstractPlaceDelegate.java
```diff
@@ -1652,14 +1652,13 @@ private boolean getCanAllUnitsWithRequiresUnitsBePlacedCorrectly(
...
     }
-    final Collection<Unit> unitsInTo = to.getUnits();
     final Collection<Unit> unitsPlacedAlready = getAlreadyProduced(to);
     if (Matches.territoryIsWater().test(to)) {
       for (final Territory current : getAllProducers(to, player, null, true)) {
         unitsPlacedAlready.addAll(getAlreadyProduced(current));
       }
     }
-    final Collection<Unit> unitsAtStartOfTurnInTo = new ArrayList<>(unitsInTo);
+    final Collection<Unit> unitsAtStartOfTurnInTo = new ArrayList<>(to.getUnits());
```
DanVanAtta, line 1661: The existing code seems to have been kinda not optimal, returning a copy of a list to then create a copy of a list..

More fundamentally, if this is a heavily invoked code path, wouldn't the better datastructure here be a `Set` instead of `List` for faster item removal?

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -268,7 +268,7 @@ void addDependentUnits(final Map<Unit, Collection<Unit>> dependencies) {
   @Override
   public List<Unit> getRemainingAttackingUnits() {
     final List<Unit> remaining = new ArrayList<>(attackingUnitsRetreated);
-    final Collection<Unit> unitsLeftInTerritory = battleSite.getUnits();
+    final Collection<Unit> unitsLeftInTerritory = new ArrayList<>(battleSite.getUnits());
```
DanVanAtta, line 271: side-note, not to update for this PR: I think one problem we are running into is a lack of encapsulation. We would not have to worry as much about set mutability if `battleSite` perhaps just handled this, eg: `battleSite.getRemainingUnits`, instead it seems we have relatively procedural code and `battleSite` is more a struct, so we run into conflicts with not trying to mutate the state of an object that is fundamentally mutable state.

DanVanAtta:
Some test failures resulting, we all need to practice more care. TripleA seems like an easy code base because a lot of it is primitive/procedural, but it's brittle and coupled/entangled quite severely.
```
games.strategy.triplea.delegate.MoveDelegateTest > testAirCantLandInConquered() FAILED

    java.lang.UnsupportedOperationException

        at java.base/java.util.Collections$UnmodifiableCollection.removeAll(Collections.java:1071)
```

asvitkine:
Why wasn't that caught by a bot? Do we not run the tests on bots?

DanVanAtta:
The CI integration between travis-ci.org and travis-ci.com seems to be wonky and is not building your PRs correctly. I tried to re-install that integration, with luck that did the trick. In general any PR should have had `./verify` run locally and not use the PR CI as a first line of verification.

asvitkine:
Hmm, ./verify shows a bunch of unrelated failures.

org.triplea.modules.remote.actions.RemoteActionsControllerIntegrationTest >
userWasBanned() FAILED

    java.lang.RuntimeException: Could not create connection from
configuration. See documentation here:
https://github.com/database-rider/database-rider#jdbc-connection


114 tests completed, 114 failed

DanVanAtta:
@asvitkine you need to launch the docker DB:

```
cd lobby-db
./build_docker  # build is one time
./launch_db
```

Let us know if those commands do not work for you @asvitkine 

If not, and you're on windows, @RoiEXLab may have set up docker on windows already. Between the two of you would be really great to see some setup documentation be created.

I have on my radar to create a contributing.md file to explain these steps and expectations.

asvitkine:
Running ./build_docker tells me "docker: command not found"

Is there a way to just omit running those tests when running ./verify?

I don't really care for installing docker on my machine since I'm not
working on that part of the code.

tvleavitt:
Y'all should build a downloadable, pre-configured VM with everything on it,
no set up required, pulls updated config from the 'net as needed.

Thomas Leavitt

DanVanAtta:
@asvitkine the unit test coverage is around 20%, integration tests are not comprehensive but they do find problems that are otherwise not caught. TripleA is really brittle, it is actually *really* slow to development on because regressions are very easy to introduce and manual test burden is extremely high. You should be running integration tests even if you think your updates are "unrelated".

You'll need to install docker. Which OS are you running?

@tvleavitt we do have a vagrant file for running deployments locally. Though, I wouldn't want that for a development environment. For example, IDEA performance really degrades on a VM, we need the horsepower. The only thing we're missing is getting set up documentation for windows, the only dependency beyond java is docker.

asvitkine:
> @asvitkine you need to launch the docker DB:
> 
> ```
> cd lobby-db
> ./build_docker  # build is one time
> ./launch_db
> ```
> 
> Let us know if those commands do not work for you @asvitkine
> 
> If not, and you're on windows, @RoiEXLab may have set up docker on windows already. Between the two of you would be really great to see some setup documentation be created.
> 
> I have on my radar to create a contributing.md file to explain these steps and expectations.

$ ./build_docker

Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?

./run_docker 
Stopping docker...
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
xargs: illegal option -- -
usage: xargs [-0opt] [-E eofstr] [-I replstr [-R replacements]] [-J replstr]
             [-L number] [-n number [-x]] [-P maxprocs] [-s size]
             [utility [argument ...]]

(I'm on Mac.)

Seems these scripts don't really work on Mac? Can we have a way to exclude this test suite until someone who knows docker and these scripts (not me) can make them work on Mac?

DanVanAtta:
Let's continue the docker discussion thread in: https://github.com/triplea-game/triplea/issues/6230

Issue #6208, opened 2020-04-12
Frequent build issues on CI (dependency download failures)


DanVanAtta:
We're seeing pretty frequent build failures on PR builds or core branch builds.

Looks like they are usually problems downloading dependencies.

This issue is to track any discussion related to this. (follows-up on: https://github.com/triplea-game/triplea/issues/6205)

DanVanAtta:
There are a number of discussions in gradle about roughly this problem, for reference:
- https://github.com/gradle/gradle/issues/4629
- https://github.com/gradle/gradle/issues/3243
- https://github.com/gradle/gradle/issues/7355

DanVanAtta:
It looks like we are still seeing problems. I'm not sure the retry property is taking effect. We could build a retry directly into the CI script, though that seems like a very brute-force alternative.

PR #6209, opened 2020-04-12
Add a gradle properties file with more retries and default parallel build

DanVanAtta:
- sets default to parallel builds, we typically want to enable that
  for faster builds as a number of sub-projects can be built in parallel.
  Of note, any tests that could conflict should all be confined to
  a single sub-project, we should then be okay from test non-determinism

- increases gradle dependency retry count and increases back off time.
  With luck this will help reduce transient failures we see on a regular
  basis in CI builds downloading artifacts.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

Quick smoke-testing running a `./gradlew compileJava` and `test`

## Additional Review Notes

The retry increase hopefully will help with CI build issues we are seeing: https://github.com/triplea-game/triplea/issues/6208


gradle.propertiess
```diff
@@ -0,0 +1,11 @@
+org.gradle.caching=true
```
RoiEXLab, line 1: Uhhm I believe you mistyped the name of this file

DanVanAtta: thanks, good catch

RoiEXLab: Feel free to merge afterwards

PR #6212, opened 2020-04-12
Re-write a couple iterative operations in UnitCollections with stream operations

DanVanAtta:
## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
- checked that there is some automated coverage on the methods updated, would expect to see those fail if there were any issues in the update.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/UnitCollection.java
+++ b/game-core/src/main/java/games/strategy/engine/data/UnitCollection.java
@@ -1,5 +1,8 @@
 package games.strategy.engine.data;
 
+import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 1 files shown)


DanVanAtta:
Closing, this change is too risky. So much code assumes direct access to mutable collections, changing those collections to be unmodifiable is risky.

DanVanAtta:
Perhaps it was a 'toList' that failed.

All in all, low ROI for simplifying this code. We are better off getting the usages to not use UnitCollection where possible and make less reliance on procedural constructs.

PR #6216, opened 2020-04-12
Update units to place and units to move unit count text

DanVanAtta:
(1) Remove the unit count on units to place. It's pretty easy to
sum up, but it is not a consistent count with units to move.
Notably, if we sum up all units to place we get the count
in the title bar of the collapsible menu. But, if you sum up
the units to move visible in the collapsible menu, it is not
equal to the value in the title bar.

EG: units to place shows 4 units, title bar has 4 units, but
if you see 4 units in units to move, the title bar has
potentially and likely a different value which is the
sum of all units to move, not just those that are in the
current display.

(2) Update text of units to move from parens to colon, eg:
Units to Move (5) -> Units to Move: 5
This is to help make 'zero' easier to read.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

### Before
Screenshot from 2020-04-12 12-17-19

### After
Screenshot from 2020-04-12 12-10-46


DanVanAtta:
Closing, this change is too risky. So much code assumes direct access to mutable collections, changing those collections to be unmodifiable is risky.

PR #6217, opened 2020-04-12
Remove toggle arrow from collapsible menu

DanVanAtta:
The toggle arrow is defined by a unicode character, unfortunately
not all fonts have that unicode character mapped. In those cases
the toggle arrow will display as a square.

This update removes the toggle arrow. We could consider adding a
transparent GIF to represent the arrow, though we run into a number
of troubles:
- sizing
- foreground color contrasting against the background
- getting the arrow to look nice and be anti-aliased

Given that a user can visually tell when the collapsible menu
is expanded or collapsed, and learning that it is toggled by clicking
the menu button, it's not ideal but we can perhpas live without
the toggle arrows for now.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

### Before
Screenshot from 2020-04-12 13-10-25

### After
Screenshot from 2020-04-12 13-04-48

### DIFF
--- a/swing-lib/src/main/java/org/triplea/swing/CollapsiblePanel.java
+++ b/swing-lib/src/main/java/org/triplea/swing/CollapsiblePanel.java
@@ -15,20 +15,13 @@ import javax.swing.SwingUtilities;
 public class CollapsiblePanel extends JPanel {
   private static final long serialVersionUID = 1L;
 
-  private static final String COLLAPSED_INDICATOR = " ►";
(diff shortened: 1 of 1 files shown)


DanVanAtta:
I like that idea, will close this for the short term while trying that out.

DanVanAtta:
#6225 uses the `getFont` check, if not supported then we get the behavior of this branch where the arrow is not printed.

PR #6231, opened 2020-04-13
Make Territory.getUnits() return an unmodifiable list. (Reland#2)

asvitkine:
Make Territory.getUnits() return an unmodifiable list. (Reland#2)

Previously, it was returning a copy, which was actually quite expensive given how often this was called. On a Domination 1914 game, with all AIs, this took 100 seconds of CPU time (across different threads since it's used in battle sim) over a single round.

A few places that need a copy are updated to do the copy on their side.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[X] Other: Optimization

## Testing
Ran all tests under game-core via IntelliJ (1432 tests passed, no failures).

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/UnitCollection.java
+++ b/game-core/src/main/java/games/strategy/engine/data/UnitCollection.java
@@ -111,7 +111,7 @@ public class UnitCollection extends GameDataComponent implements Collection<Unit
   }
 
   public Collection<Unit> getUnits() {
-    return new ArrayList<>(units);
+    return Collections.unmodifiableList(units);
   }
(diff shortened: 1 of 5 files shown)


asvitkine:
@DanVanAtta Looks like Travis CI is working again for me and checks passed. Their system flagged my account and I had to reach out to them to get it fixed, which is now the case. I believe that was the cause of it not working before.

DanVanAtta:
Awesome, glad to hear that is now fixed :+1:

PR #6233, opened 2020-04-13
Cache unit support information to speed up AI.

asvitkine:
Cache unit support information to speed up AI. 

Computing this dominated the battle simulation code, which is also
used heavily in AI logic.

The cpu time (across threads) of DiceRoll.getSortedSupport() over
one round of an all-AI Domination 1914 game took 175 seconds prior
to this change. After this change, it was 14 seconds.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[X] Other:  Optimization

## Testing
- Ran unit tests from IntelliJ in game-core.
- Ran one round of an all-AI game on Domination 1914 without errors.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/UnitTypeList.java
+++ b/game-core/src/main/java/games/strategy/engine/data/UnitTypeList.java
@@ -16,7 +16,8 @@ public class UnitTypeList extends GameDataComponent implements Iterable<UnitType
   private static final long serialVersionUID = 9002927658524651749L;
 
   private final Map<String, UnitType> unitTypes = new LinkedHashMap<>();
-  private @Nullable Set<UnitSupportAttachment> supportRules;
+  // Cached support rules. Marked transient as the cache does not need to be part of saved games.
+  private transient @Nullable Set<UnitSupportAttachment> supportRules;
 
   public UnitTypeList(final GameData data) {
     super(data);
@@ -52,8 +53,7 @@ public class UnitTypeList extends GameDataComponent implements Iterable<UnitType
   public Set<UnitSupportAttachment> getSupportRules() {
     if (supportRules == null) {
       supportRules =
-          UnitSupportAttachment.get(getData())
-              .parallelStream()
+          UnitSupportAttachment.get(getData()).stream()
               .filter(usa -> (usa.getRoll() || usa.getStrength()))
               .collect(Collectors.toSet());
     }
--- a/game-core/src/main/java/games/strategy/engine/data/UnitTypeList.java
+++ b/game-core/src/main/java/games/strategy/engine/data/UnitTypeList.java
@@ -1,18 +1,22 @@
 package games.strategy.engine.data;
 
 import com.google.common.annotations.VisibleForTesting;
+import games.strategy.triplea.attachments.UnitSupportAttachment;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 
 /** A collection of unit types. */
 public class UnitTypeList extends GameDataComponent implements Iterable<UnitType> {
   private static final long serialVersionUID = 9002927658524651749L;
 
   private final Map<String, UnitType> unitTypes = new LinkedHashMap<>();
+  private @Nullable Set<UnitSupportAttachment> supportRules;
 
   public UnitTypeList(final GameData data) {
     super(data);
@@ -38,6 +42,22 @@ public class UnitTypeList extends GameDataComponent implements Iterable<UnitType
       types.add(type);
     }
     return types;
+  }
+
+  /**
+   * Returns the unit support rules for the unit types. Computed once and cached afterwards.
+   *
+   * @return The unit support rules.
+   */
+  public Set<UnitSupportAttachment> getSupportRules() {
+    if (supportRules == null) {
+      supportRules =
+          UnitSupportAttachment.get(getData())
+              .parallelStream()
+              .filter(usa -> (usa.getRoll() || usa.getStrength()))
+              .collect(Collectors.toSet());
+    }
+    return supportRules;
   }
 
   public int size() {
--- a/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
@@ -997,14 +997,9 @@ public class DiceRoll implements Externalizable {
       final GameData data,
       final boolean defence,
       final boolean allies) {
-    final Set<UnitSupportAttachment> rules =
-        UnitSupportAttachment.get(data)
-            .parallelStream()
-            .filter(usa -> (usa.getRoll() || usa.getStrength()))
-            .collect(Collectors.toSet());
     getSupport(
         unitsGivingTheSupport,
-        rules,
+        data.getUnitTypeList().getSupportRules(),
         supportsAvailable,
         supportLeft,
         supportUnitsLeft,


game-core/src/main/java/games/strategy/engine/data/UnitTypeList.java
```diff
@@ -40,6 +44,22 @@ public UnitType getUnitType(final String name) {
...
+  /**
+   * Returns the unit support rules for the unit types. Computed once and cached afterwards.
+   *
+   * @return The unit support rules.
+   */
+  public Set<UnitSupportAttachment> getSupportRules() {
+    if (supportRules == null) {
+      supportRules =
+          UnitSupportAttachment.get(getData())
+              .parallelStream()
```
DanVanAtta, line 56: Did you happen to measure if there is a performance boost of using parallel stream here instead of stream? `parallel` stream is oddly often slower than stream.

game-core/src/main/java/games/strategy/engine/data/UnitTypeList.java
```diff
@@ -40,6 +44,22 @@ public UnitType getUnitType(final String name) {
     return types;
   }
 
+  /**
+   * Returns the unit support rules for the unit types. Computed once and cached afterwards.
+   *
+   * @return The unit support rules.
+   */
+  public Set<UnitSupportAttachment> getSupportRules() {
```
DanVanAtta, line 52: side-note, the method consuming this is overspecified
```
  private static void getSupport(
      final Collection<Unit> unitsGivingTheSupport,
      final Set<UnitSupportAttachment> rules,
      :
```

With the above, a consequence is if you have a `List` of rules, or any other non-Set collection, then you need to do a needless copy into a set. That method works fine when the collection type is changed to a Collection, ie:
```
  private static void getSupport(
      final Collection<Unit> unitsGivingTheSupport,
      final Collection<UnitSupportAttachment> rules,
     :
```

game-core/src/main/java/games/strategy/engine/data/UnitTypeList.java
```diff
@@ -40,6 +44,22 @@ public UnitType getUnitType(final String name) {
...
+    if (supportRules == null) {
+      supportRules =
+          UnitSupportAttachment.get(getData())
+              .parallelStream()
+              .filter(usa -> (usa.getRoll() || usa.getStrength()))
```
DanVanAtta, line 57: Are you sure that `usa`, unit support attachment `getStrength` will never change?

game-core/src/main/java/games/strategy/engine/data/UnitTypeList.java
asvitkine: I didn't try it, but I don't think it would make a big difference. The main issue was this was being called *a lot* and this change makes it get called just once, so it goes from taking a long time to taking no time at all (the remaining work getSortedSupport() is the result of the other calls).

DanVanAtta:
> The cpu time (across threads) of DiceRoll.getSortedSupport() over
one round of an all-AI Domination 1914 game took 175 seconds prior
to this change. After this change, it was 14 seconds.


What's the total AI runtime? Is the 175 seconds just in this one method, or across the whole round? 90% reduction is quite dramatic.

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
```diff
@@ -997,14 +997,9 @@ public static void getSortedSupport(
...
       final boolean allies) {
-    final Set<UnitSupportAttachment> rules =
-        UnitSupportAttachment.get(data)
-            .parallelStream()
-            .filter(usa -> (usa.getRoll() || usa.getStrength()))
-            .collect(Collectors.toSet());
     getSupport(
         unitsGivingTheSupport,
-        rules,
+        data.getUnitTypeList().getSupportRules(),
```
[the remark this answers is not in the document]
asvitkine: The rational is to cache it on the per GameData instance. Originally I had it in GameData, but GameData was already doing a lot of different things and had an abstraction for its UnitTypeList, so it seems like a natural place to put it.

game-core/src/main/java/games/strategy/engine/data/UnitTypeList.java
asvitkine: I am not actually. If it can change, then there should be logic to invalidate the cached info.

Do you know who the right person would be to give a definitive answer?

asvitkine:
> > The cpu time (across threads) of DiceRoll.getSortedSupport() over
> > one round of an all-AI Domination 1914 game took 175 seconds prior
> > to this change. After this change, it was 14 seconds.
> 
> What's the total AI runtime? Is the 175 seconds just in this one method, or across the whole round and all methods? 90% reduction is quite dramatic.

In this case, it was 560 seconds before all my three optimizations for total time for that round. But this specific code is run in the battle simulation context on multiple threads, so the 175 seconds was split across 8 worker threads on my machine (so presumably contributing roughly 20 seconds to the 560 second total). Although on a machine with fewer cores, it would likely contribute more.

game-core/src/main/java/games/strategy/engine/data/UnitTypeList.java
DanVanAtta: I'd recommend giving it a quick measure if you can. If not, blindly replacing with a standard stream.

Small tasks run in parallel often are actually much slower, it's surprising (I think the reasoning is the overhead for thread coordination is a lot, streams are not yet very optimized by the JDK, meanwhile single-threaded iteration is highly optimized still). Essentially it's to the point that one should not use parallel stream unless it has been benchmarked.

Thoughts?

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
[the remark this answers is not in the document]
DanVanAtta: This might be a damned if you do, damned if you don't kind of situation.

With the move, we're going to be assuming that `UnitTypeList` is a singleton, which is arguably a bit assumption. `UnitTypeList` is also not much of a data structure, I think it's a candidate for us to inline all of its methods to where they are used and removed.

`Support` rules and the existing support code totally do not belong in `DiceRoll` either. Though, reaching out to another class to grab data to be used in a local computation is also a code smell as well.

I think a third approach, extracting the support method and the support cache to a new class so that it does not live in either DiceRoll nor UnitTypeList is probably the way to go. I can see that being too much for the scope of this update.

I'd be most comfortable if we had the support rules stay local to `DiceRoll` for now, or going with the expanded scope route. 

One concern is for serialization, as a `GameDataComponent` the `UnitTypeList` will be serialized to disk. That means after this update, any save game will *require* the support set field to be present, moving it later will break save-game compatibility. I think that would perhaps be the biggest reasoning to avoid a cached object in `GameDataComponent`, so that transient data is not persisted to save game files.

game-core/src/main/java/games/strategy/engine/data/UnitTypeList.java
DanVanAtta: It looks like it will not change after a game data has been loaded:

UnitSupportAttachment.java:
```
        .put("strength", MutableProperty.ofReadOnly(this::getStrength))
```
```
  public static <T> MutableProperty<T> ofReadOnly(final Supplier<T> getter) {
    return of(noSetter(), noStringSetter(), getter, noResetter());
  }
```

```
    <attachment name="supportAttachmentFortificationbritishex" attachTo="britishFortification" javaClass="games.strategy.triplea.attachments.UnitSupportAttachment" type="unitType">
      <option name="unitType" value="britishInfantry:britishMarine:britishCombatEngineer:britishAlpineInfantry:britishParatrooper:russianInfantry:russianMarine:russianCombatEngineer:russianAlpineInfantry:russianParatrooper:americanInfantry:americanMarine:americanCombatEngineer:americanAlpineInfantry:americanParatrooper:chineseInfantry:chineseCombatEngineer:chineseAlpineInfantry"/>
      <option name="side" value="defence"/>
      <option name="bonus" value="2"/>
      <option name="bonusType" value="FortBonus"/>
      <option name="impArtTech" value="false"/>
      <option name="players" value="ExiledAllies"/>
    </attachment>
```

Setting a breakpoint on `getStrength` to double check when it is loaded could help definitely answer the question.

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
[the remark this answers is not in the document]
asvitkine: The cache has to be per-game data, though.

So we either need to hang it off GameData (does that mean it will be part of save games?) or we hold a map keyed by GameData which is extra complexity and either a memory leak or we need a way to clear that cache otherwise.

game-core/src/main/java/games/strategy/engine/data/UnitTypeList.java
asvitkine: Tried it out and using parallelStream() is indeed slower (about 3x slower - 0.9ms vs. 0.3ms).

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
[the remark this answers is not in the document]
asvitkine: Also, my current implementation doesn't actually assume UnitTypeList is a singleton.

If in the future, this is called on other UnitTypeLists, everything works fine. They cache their associated support rules and everything works fine.

Given that any solution where the cache doesn't hang off GameData somehow results in a memory leak or needing to do lifetime management (e.g. when a game is closed, to go clear out these caches that would be keyed by GameData), I don't think it's a good idea to keep the cache separate.

For save game serialization, we can mark the field transient. WDYT?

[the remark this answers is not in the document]
DanVanAtta: Anything that extends `GameDataComponent` will become part of save games.

Making it transient could do the trick indeed. Just please be sure to double check that it nulls out when you save and load and that the cache is refreshed after loading.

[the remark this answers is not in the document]
asvitkine: Thanks! I did the test and confirmed that loading the save game results in the field being null.

game-core/src/main/java/games/strategy/engine/data/UnitTypeList.java
asvitkine: I think it's a bit of a philosophical question - that applies to not just here. Would it be valid to repeat the same UnitSupportAttachment rule multiple times in the list? If logically that does not make sense, then to me it seems valid to use a Set to codify that in the API rather than a Collection which doesn't carry that guarantee when the function should not expect the same instance repeated in the collection.

DanVanAtta: It's a matter of contract. A contract that accepts says "you must give me a set, otherwise there will be errors in how I do computation". Part of it is letting go and trusting the implementation that is passed is what the caller intends. If the caller is happy passing you duplicates, and you can process duplicates, then it is an issue with the implementation. If non-duplicates are needed for correctness, then it makes sense to require a set (at which point it's very justified to force callers to convert their list to a set). If duplicates are just inefficient, then it's on the caller for not doing the cleanup first, and you may be incurring worse efficiencies by forcing a collection copy. At the end of the day it's a trust that the implementation type satisfies any contract that you need. If you need unique items, then the contract should be set, if you need ordering, then list, if you really don't care (but set would be more efficient), then use collection and hope that callers give you a set (or handle it in the processing).

DanVanAtta: From another perspective, it's simply the caller that often decides if duplicates are allowed or not. They instantiate the data, they likely use it in other places/ways - they probably are passing you a set. Otherwise, set should be required as an input when it is required for correctness, it enforces a contract. Needless enforcement between list and set often results in a lots of unnecessary data copy. When done well, using abstract types, often the concrete type is defined in just one place, can be changed, and everything just works with minimal to no copying of collection data.

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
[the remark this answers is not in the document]
DanVanAtta: Cool, thanks for checking :+1:

DanVanAtta:
Thanks for working on this @asvitkine :100: 

If you ever get hungry for more performance tuning, any slow test is a good candidate too :grin: 

One fine day we'll remove the need to do game data copy for battle calc, which should give some of the best possible gains for AI. Wrangling the game data object is a big task, we need to change how it is serialized (we've been thinking to use a text format instead of serialized java objects). Fundamentally for battle calc we need to determine exactly which objects it needs and only copy those and not the whole game data object, which likely will require a lot of decoupling work.

Issue #6181, opened 2020-04-11
"Shift+d" needs to be remapped to "ctrl+enter"


DanVanAtta:
We agreed that 'shift+d' for 'done' button shortcut is not super intuitive.

This is a task placeholder to remap that hotkey to "ctrl+enter"

DanVanAtta:
Task completed in: https://github.com/triplea-game/triplea/pull/6234

PR #6234, opened 2020-04-13
Clean up keybindings code and remap done shortcut to "ctrl+enter"

DanVanAtta:
Refactoring:
- Moves keybindings utilities methods from SwingComponents to SwingKeyBindings
- Helps combine keybinding helpers, they had become scattered and duplicative
- Creates a single enum for 'ButtonDownModifier' (eg: shift, control)
- Expands enumerated items in 'KeyCode'
- Consistently uses addKeyBindings method across all usages, should help
  in tracking down and finding keybindings.

Feature updadting:
- Changes 'shift+d' hotkey that was bound to the 'done' button to be 'ctrl+enter'

## Functional Changes

[] New map or map update
[] New Feature
[x] Feature update or enhancement: https://github.com/triplea-game/triplea/issues/6181
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
Did some quick smoke testing on a single player game and verified new "ctrl+enter" key binding works.

## Additional Review Notes

The keybinding helper methods in SwingComponents becamse a bit of a mess. Extracting them to a single class helped to unify and clean them up


swing-lib/src/main/java/org/triplea/swing/key/binding/SwingKeyBinding.java
```diff
@@ -0,0 +1,108 @@
...
+        action);
+    addKeyMapping(
+        (JComponent) component.getContentPane(),
+        KeyCombination.of(key, ButtonDownMask.META),
+        action);
+  }
+
+  private static void addKeyMapping(
+      final JComponent component, final KeyCode keyCode, final Runnable action) {
+    addKeyMapping(component, KeyCombination.of(keyCode, ButtonDownMask.NONE), action);
```
RoiEXLab, line 71: What do you think about making KeyCombination and KeyCode implement the same interface?
This way you woldn't need to convert one to the other and were able to nest things in theory

DanVanAtta:
> Key bindings are actually one of my suspects when it comes to the memory leak issue, because they are never removed and have references to lambda expressions that might have hidden references to enclosing objects.

It's the swing components that will have references to the action maps. If the swing components are never GC, then the action maps will not be either. If the action maps are not cleaned up, it's not the key bindings that are the problem but the thing that is keeping a reference to the swing component. It would be making a memory leak worse, but should not be the cause.

swing-lib/src/main/java/org/triplea/swing/key/binding/SwingKeyBinding.java
DanVanAtta: Not a  bad thought given the common method. It does not help that much though as we need to convert `KeyCode` into `KeyCombination`.

DanVanAtta: Going along those lines, `KeyCode` actually does not need the `toKeyStroke` at all and can be removed.

DanVanAtta:
Couple quick updates here @RoiEXLab ; would you mind double checking those?

swing-lib/src/main/java/org/triplea/swing/key/binding/SwingKeyBinding.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  private static void addKeyMapping(
+      final JComponent component, final KeyCombination keyCombination, final Runnable action) {
+    final String keyBindingIdentifier = UUID.randomUUID().toString();
+
+    final AtomicBoolean enabled = new AtomicBoolean(true);
+
+    if (keyCombination.getButtonDownMask() == ButtonDownMask.NONE
+        || keyCombination.getButtonDownMask() == ButtonDownMask.SHIFT) {
```
RoiEXLab, line 81: I was just about to complain that this might not always work, but then I reliazed that using this abstraction that only allows a single key to be down this is completely fine

Issue #6222, opened 2020-04-12
Player Ban is no longer showing disconnect pop-up


DanVanAtta:
**How can the problem be recreated?**
- join lobby
- get banned
- notice you get disconnected and there is no pop-up telling you were banned (session disconnect reason).

**When the problem is fixed, what should have happened instead?**
- when session is unexpectedly disconnected, we should be printing the disconnect reason in a dialog to the user (which will tell them they were banned)

**Which Engine Version are you using?** 
2.0.18xxx

DanVanAtta:
Fixed in: https://github.com/triplea-game/triplea/pull/6236

PR #6236, opened 2020-04-13
Add distinct callback for websocket terminated

DanVanAtta:
Currently we have only one callback whenever a websocket
is closed. This update creates a distinction between
when a websocket is 'terminated' (closed by server)
vs 'closed' (standard close by client).

The 'closed' callback remains just a runnable and continues
to be just for client side cleanup. The 'terminated' callback
is invoked with the disconnect error reason, which for the
case of bans will be the ban message from the server.

If the server gives no message, then we default that message
to a generic "the server disconnected" message.

After this update users are given a notification when they
are disconnected due to ban, before they were given no notification.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix: https://github.com/triplea-game/triplea/issues/6222 
[] Other:   

## Testing
- verified before there was no message when a user is banned
- verified when banned there is a disconnect message
- verified when server is shutdown there is a disconnect message
- verified leaving lobby (normal disconnect) does not yield a disconnect message

## Screens Shots

Message when disconnected due to ban:
Screenshot from 2020-04-13 12-34-17

Message when disconnected due to server shutdown:
Screenshot from 2020-04-13 12-38-00

### DIFF
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
@@ -38,7 +38,7 @@ import org.triplea.java.timer.Timers;
 class WebSocketConnection {
   @VisibleForTesting static final int DEFAULT_CONNECT_TIMEOUT_MILLIS = 5000;
 
-  private static final String CLIENT_DISCONNECT_MESSAGE = "Client disconnect.";
+  @VisibleForTesting static final String CLIENT_DISCONNECT_MESSAGE = "Client disconnect.";
 
   private WebSocketConnectionListener listener;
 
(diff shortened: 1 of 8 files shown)


http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
```diff
@@ -205,7 +205,11 @@ public void onOpen(final WebSocket webSocket) {
     public CompletionStage<?> onClose(
         final WebSocket webSocket, final int statusCode, final String reason) {
       pingSender.cancel();
-      listener.connectionClosed(reason);
+      if (reason.equals(WebSocketConnection.CLIENT_DISCONNECT_MESSAGE)) {
+        listener.connectionClosed();
+      } else {
+        listener.connectionTerminated(reason.isBlank() ? "Server disconnected" : reason);
+      }
```
RoiEXLab, line 212: I don't know what the server side of the websocket implementation is capable of, but I think it would be a better idea to use a custom status code  for closing (it looks like codes 4000 - 4999 are reserved for application use) and make decisions based on that.

DanVanAtta: Why better?

status code for http calls can certainly give more information. There are cases where the payload of a 200 is an error message, ie: the message and service are all okay, everything happened as expected, but something resulted in an error.

In this case though, the message string is quite exact, keeps it quite simple. (For example, a close from server very well could be a 200, which would look like the same if the client closed)

Issue #6248, opened 2020-04-14
Notifications are hidden by Battle window


ZjelcoP:
**How can the problem be recreated?**

-2.0.18816
-Win 10
-Play the Tutorial until you get to battles.

Playing the Tutorial there are several notifications.
The battle window now hides the notifications. You are unable to read them, the battle window always stays on top. Only way to proceed is to Space bar through the notifications without reading them.

Notifications should have priority over battle window.

Also when completing Combat moves and there is only one battle this battle immediately starts. You no longer have the option to start it yourselves. This may be intentional but I'm not sure I like it and it definitely doesn't help the Tutorial.

![afbeelding](https://user-images.githubusercontent.com/13902879/79275010-b7687000-7ea5-11ea-8e8e-f99662047210.png)

Cheers Zjelco

beelee1:
bummer was it hoping it was fixed. Sounds like #6132 . If u move window is the other one behind it ?

DanVanAtta:
@beelee1 have you been able to reproduce the problem of #6132? This might be a different issue as the notification windows are not spawned by the battle proper.

DanVanAtta:
Crud, I'm not seeing this problem on my computer.

@ZjelcoP could you double check with the very very latest just in case I had the version number wrong and this is actually fixed.

When you move the window that is behind the battle window, does it stat to display on top of the battle window, or does it *always* appear behind the battle window?

DanVanAtta:
I'm trying a slightly 'blind' fix for this with: https://github.com/triplea-game/triplea/pull/6251

Not being able to repro really hurts the chances it'll get properly fixed. :crossed_fingers:

ZjelcoP:
Issue hasn't changed with 2.0.18827

-Battle window stays on top even if you move the notification around.
-The Battle window cannot be moved while a notification is active.
-A small size notification is impossible to view (fully hidden by battle window) and can only be closed by space bar.
-A lager notification can be grabbed by the edge and moved to a corner to view, however battle window will stay on top.

Also the second issue is starting to bug me more:
_Also when completing Combat moves and there is only one battle this battle immediately starts. You no longer have the option to start it yourselves. This may be intentional but I'm not sure I like it and it definitely doesn't help the Tutorial._

I like to have a local view of the map before starting a battle. It now feels like everyone is in position and just when you want to give the order to attack you find everyone already rushed off.
I can see why it might have been put in for sake of efficiency but please make it optional under combat settings.

PR #6251, opened 2020-04-15
Battle display as dialog - remove always on top setting

DanVanAtta:
## Changes

commit c36cdc0711dd992009755fd24fa8982a2b4ec304

    BattleWindow from JFrame to JDialog
    
    - Remove always on top property of the battle window
    On some OS this seems to always win out over any other
    notifications that are supposed to appear on top
    of the battle window
    
    - The conversion to JDialog will help ensure that the
    battle window stays on top of the map and does not
    get hidden behind it. This should have the same effect
    as always set on top being set to true.

commit ab6379439ad57dc9de6fdd69eef4534136a08ec6

    Rename BattleFrame to BattleWindow as it is no longer a frame

commit 801a58713996827eeef4292d95b2b4f18d9e0ee5

    Remove PBEM.setFocusWindow
    
    It appears to still work even with battle panel shown in a dialog
    instead of in a frame.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  https://github.com/triplea-game/triplea/issues/6248 
[] Other:   

## Testing
- Did some smoke testing on tutorial and with PBEM dice roller enabled to verify dialogs still appear on top of battle window, and also that battle window stays on top of the map.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/random/PbemDiceRoller.java
+++ b/game-core/src/main/java/games/strategy/engine/random/PbemDiceRoller.java
@@ -28,21 +28,10 @@ import org.triplea.util.ExitStatus;
  * error we wait until we get a good roll before returning.
  */
 public class PbemDiceRoller implements IRandomSource {
-  private static Frame focusWindow;
-
   private final IRemoteDiceServer remoteDiceServer;
 
   public PbemDiceRoller(final IRemoteDiceServer diceServer) {
     remoteDiceServer = diceServer;
-  }
-
-  /**
-   * If the game has multiple frames, allows the UI to set what frame should be the parent of the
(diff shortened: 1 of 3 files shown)


game-core/src/main/java/games/strategy/engine/random/PbemDiceRoller.java
```diff
@@ -74,9 +63,6 @@ public int getRandom(final int max, final String annotation) {
   }
 
   private static Frame getFocusedFrame() {
```
RoiEXLab, line 65: I'm really questioning if this method should exist

PR #6261, opened 2020-04-16
Fix use of checkArgument().

asvitkine:
The code in PlainRandomSource was passing a String.format() result to checkArgument() instead of letting checkArgument() do the formatting. The latter is preferred because then the formatting will only be done if the precondition fails. The random number generation is pretty hot (used in combat in battle simulations) and this was showing up in profiles (e.g. 30s CPU time over a 1914 Domination round)

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[X] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
None


DanVanAtta:
:+1:

PR #6262, opened 2020-04-16
Speed up sortByStrength() by precomputing strengths.

asvitkine:
Speed up sortByStrength() by precomputing strengths.

The computation for getting unit strength was dominating this function. With this change an execution of this function over two AI players turns on Domination 1914 goes from 39s of CPU time to 6s.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[X] Other: Optimization

## Testing
Ran several turns of Domination 1914.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
@@ -1,6 +1,8 @@
 package games.strategy.triplea.delegate;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.HashBasedTable;
+import com.google.common.collect.Table;
 import games.strategy.engine.data.GameData;
 import games.strategy.engine.data.GamePlayer;
 import games.strategy.engine.data.Territory;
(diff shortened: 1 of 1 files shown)


asvitkine:
> For performance/refactoring improvements, would be good to see any kind of characterization/functionality tests be put in place so we can be more confident there are no issues (it also helps review, as review only needs to focus on what changed and that testing is sufficient, and not be looking for problems).
> 
> Do you know why `getDefense` and `getAttack` are so expensive?

They're not *so* expensive, but it's called a lot during sort. But the underlying issue is looking up they look attach bonuses from tech:

Screen Shot 2020-04-16 at 12 19 47 AM

DanVanAtta:
Does that mean tech could alter the strength values @asvitkine ? Does there need to be a cache invalidation when tech changes? Or is tech post-computed from these values? You've piqued my curiosity.

asvitkine:
The cache is for the duration of the call. It's not persisted.

DanVanAtta:
cool :+1: 
Thanks for clarifying

PR #6271, opened 2020-04-17
Fix removing disabled units with battle AA

ron-murhammer:
Addresses issue 1 here: https://forums.triplea-game.org/topic/1031/world-of-war-heroes-official-thread/110

AA rolls were done before disabled units (bombing damage greater than `maxOperationalDamage`) are removed from combat. They should just be removed before the battle even starts rather than after AA rolls.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
@@ -576,6 +576,7 @@ public class MustFightBattle extends DependentBattle implements BattleStepString
   @Override
   public void fight(final IDelegateBridge bridge) {
     removeUnitsThatNoLongerExist();
+    removeDisabledUnits();
     if (stack.isExecuting()) {
       final IDisplay display = bridge.getDisplayChannelBroadcaster();
       display.showBattle(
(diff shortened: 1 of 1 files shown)


DanVanAtta:
@ron-murhammer it looks like the PR comment is useful and perhaps would do well to be in the commit history.  I think it should be kept in mind that comments made in a PR are only for the reviewer and are of extremely transitory value. IMO ideally everything in the PR overview is auto-filled from the commit comment and a PR author then only needs to check a box and explain the testing that they have then done.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -576,6 +576,7 @@ private static void removeFromDependents(
   @Override
   public void fight(final IDelegateBridge bridge) {
     removeUnitsThatNoLongerExist();
```
DanVanAtta, line 578: The pattern of mutating void method calls is unfortunate. The reasoning/logic behind the ordering is not captured, makes it really brittle as re-ordering lines or adding functionality can cause unexpected problems.

It's also not clear as well why the ordering is as it is, we also run into issues as well for reading the functions as the void calls you have to trace down to see what is happening. Much of this perhaps could just as well be inlined and/or made functional so the control flow is captured at a high level. Regardless, it's just kinda sad to see this anti-pattern. There are some interesting tools out there that would perhaps catch this, those tools permute lines and re-run tests to see if any file due to line permutations and then gives you a score of methods that fail highlighting brittlness.

DanVanAtta:
I updated the squash commit message to have the overview included here.

PR #6276, opened 2020-04-17
Casualty selector breakup

DanVanAtta:
## Overview

- Breaking up casualty selector to new classes
- Very few actual code changes, should be most reviewable commit-by-commit
- Added some new tests cases for casualty selector inner workings.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
- did some smoke testing while verifying various casualty scenarios on different maps

## Additional Review Notes

- please suggest any additional new test cases that would be really good to add to the two new casualty selector tests (f825494)


game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtyOrderOfLosses.java
```diff
@@ -276,13 +297,13 @@ static void clearOolCache() {
...
+              + parameters.battlesite.getName()
               + "|"
-              + defending
+              + parameters.defending
               + "|"
-              + amphibious
+              + parameters.amphibious
               + "|"
               + targetsHashCode
               + "|"
```
RoiEXLab, line 310: I'm seeing this code snippet an odd amount of times.
Thoughts about extracting this to a helper method?

java-extras/src/main/java/org/triplea/java/collections/IntegerMap.java
```diff
@@ -30,6 +30,10 @@ public IntegerMap(final IntegerMap<T> integerMap) {
     this(integerMap.mapValues);
   }
 
+  public static <X> IntegerMap<X> of(final Map<X, Integer> map) {
+    return new IntegerMap<>(map);
+  }
+
```
RoiEXLab, line 36: What is this method for and why isn't the constructor enough?

game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtyOrderOfLosses.java
DanVanAtta: There are more areas of copy / paste than this too, code climate does a pretty decent job pointing them out.

This update is just preliminary, intentionally not modifying too much of the existing code yet. It needs to be done iteratively. I'm hoping as well that this hashing function can be removed as part of one those iterations.

java-extras/src/main/java/org/triplea/java/collections/IntegerMap.java
DanVanAtta: Convenience and avoids the diamond notation:
eg:
```
new IntegerMap<>(Map.of(...));
```
vs:
```
IntegerMap.of(Map.of(...))
```

Guava does that type of think all over the place, type inference in java is much smarter about static calls than it is about `new`

Ultimately I think `IntegerMap` is possibly a pre-Java 1.5 artifact from before there were even generics, we should consider replacing `IntegerMap` with `Map<T, Integer>`.

DanVanAtta:
@RoiEXLab did you happen to spot any interesting test cases that could be added?

@ron-murhammer curious if you had a chance to look at the tests in this PR and could recommend adding additional interesting OOL tests.

ron-murhammer:
@DanVanAtta Anything with multiple different support attachments that aren't just +1 to 1 unit would be good. I think 270bc elephants are a good example of that. Also something with subs/destroyers where strength is the same between taking a destroyer and non-destroyer against subs and it should take the non-destroyer.

java-extras/src/main/java/org/triplea/java/collections/IntegerMap.java
RoiEXLab: @DanVanAtta I should've clarified:
I don't think I've seen this method being used inside of this PR.
So I suspect it's unused

DanVanAtta: FWIW - test cases make use of it:
https://github.com/triplea-game/triplea/pull/6276/files#diff-98b2c58d739b161b3aa5cc607fe5b6baR42

PR #6277, opened 2020-04-17
Code climate increase thresholds and add sonar

DanVanAtta:
commit b1abccabd109ce609950987798c69039d79074fb

    Increate codeclimate thresholds
    
    Cognitive complexity of 7 seems to be a more reasonable
    threshold where methods are actually quite difficult. We have
    had some examples where 6 and 7 are hard to reduce down further.
    
    Method line count of 25 is almost about right, breaking 30 starts
    to be more clear cut and reduces noise at that threshold.

commit 15cc8196bc060ed668d8cc3e1fcca367cfde8728

    Add sonar plugin to code climate, should enable sonar analysis checks

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Problem fix:  
[] Other:

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -27,3 +27,6 @@ plugins:
   # https://docs.codeclimate.com/docs/shellcheck
   shellcheck:
     enabled: true
+  sonar-java:
+    enabled: true
+
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -4,12 +4,12 @@ version: "2"
 # https://docs.codeclimate.com/docs/maintainability#section-checks
 # https://docs.codeclimate.com/docs/advanced-configuration
 checks:
-  - method-complexity: 
-      config:
-        threshold: 7
-  - method-lines:
-      config:
-        threshold: 30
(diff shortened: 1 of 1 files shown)


tvleavitt:
Did you notice that the CodeCov report here refers to different code?
"MustFightBattle.java", etc.

Thomas Leavitt

RoiEXLab:
Code climate doesn't seem to happy with this change

DanVanAtta:
@tvleavitt I can't say exactly why the test coverage appears changed. I would first suspect it is due to something that is effectively an integration test running that is probably executing large amounts of code but not actually testing it. Regrettably the state of automated test is very primitive, low test coverage, and a decent number of integration tests that are not focused.

To fix we'd need to mark the integration test that is providing coverage as an integ test so it is excluded, we have arhitecture work to do to break things up so we can test our modules in isolation.

`MustFightBattle` is one area where we have probably the most technical debt, the class has just grown and grown with many features over years and is akin to a god-class.

PR #6280, opened 2020-04-18
Exclude gradlew script from codeclimate checks & remove newly added sonar check

DanVanAtta:
## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Problem fix:  
[] Other:


DanVanAtta:
Checking out the new sonar violations, looks like sonar is not seeing the lombok generated code and is throwing a lot of false positives (a number of issues it is raising are real, so this is unfortunate).

PR #6289, opened 2020-04-18
Speed up code in UnitsHitChange.

asvitkine:
Previously, this code would need to look at all units in all territories to figure out which territories were affected, so that they can be notified. This was quite slow, taking 254 sec of CPU time over one round of all-AI Domination 1914 (across all threads - this was mostly affected battle simulation which was split across 8 threads on my machine).

This change explicitly passes the list of affected territories to the change object, eliminating that extra work. On the same Domination 1914 all-AI round, UnitsHitsChange.perform() now only takes 200 ms (instead of 254 sec).

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[X] Other:   Optimization

## Testing
Played one round of all-AI Domination 1914.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/UnitHitsChange.java
+++ b/game-core/src/main/java/games/strategy/engine/data/UnitHitsChange.java
@@ -1,7 +1,6 @@
 package games.strategy.engine.data;
 
-import java.util.Collections;
-import java.util.Set;
+import java.util.Collection;
 import org.triplea.java.collections.IntegerMap;
 
 /** A game data change that captures the damage done to a collection of units. */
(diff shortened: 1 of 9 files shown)


DanVanAtta:
@asvitkine unrelated to this PR, would you mind writing up a quick wiki page on your methodology for performance measurements?

I think there would be some very important benefits:
- it'd allow us to reproduce the data you have presented
- others could do similar in the future when the same code needs to be updated again
- would generally a good learning tool, it seems you have the process pretty well refined.

asvitkine:
> @asvitkine unrelated to this PR, would you mind writing up a quick wiki page on your methodology for performance measurements?
> 
> I think there would be some very important benefits:
> 
> * it'd allow us to reproduce the data you have presented
> * others could do similar in the future when the same code needs to be updated again
> * would generally a good learning tool, it seems you have the process pretty well refined.

Sure, I can take a stab at it.

game-core/src/main/java/games/strategy/engine/data/UnitHitsChange.java
```diff
@@ -10,35 +9,42 @@
...
   }
 
   @Override
   protected void perform(final GameData data) {
     for (final Unit item : hits.keySet()) {
       item.setHits(hits.getInt(item));
     }
-    final Set<Unit> units = hits.keySet();
-    for (final Territory element : data.getMap().getTerritories()) {
-      if (!Collections.disjoint(element.getUnits(), units)) {
```
DanVanAtta, line 34: Is this disjoint check just not needed in the end?

asvitkine: That's right. The old code was using it to determine if the affected units are in the territory in question. Now, we track that directly, so this is removed. (This call was what was so slow in the first place.)

asvitkine:
> I took a look and have just the one question. Generally looks okay 👍
> 
> If you would, please also double check a network game as `BattleDelegate` is potentially network code (can be tested done by hosting locally and connecting to the same game with a second client).

Checked, seems to be working fine. :)

asvitkine:
> @asvitkine unrelated to this PR, would you mind writing up a quick wiki page on your methodology for performance measurements?
> 
> I think there would be some very important benefits:
> 
> * it'd allow us to reproduce the data you have presented
> * others could do similar in the future when the same code needs to be updated again
> * would generally a good learning tool, it seems you have the process pretty well refined.

Added it here:

https://github.com/triplea-game/triplea/wiki/Performance-Improvements

DanVanAtta:
Thank you @asvitkine :partying_face:

PR #6293, opened 2020-04-18
Merge TripleAUnit into Unit

DanVanAtta:
This update simplifies a lot of casting and 'fixes' our data model with Unit by merging TripleAUnit into Unit. TripleAUnit was split and inherited from Unit in order to support other game types. We long have dropped additional game types which means we have a inheritance tree where all Units are effectively TripleAUnit. To get at methods in TripleAUnit, we very frequently blindly cast Unit objects into TripleAUnit, hence they are really one and the same.

This update merges the methods of TripleAUnit into Unit and then removes the TripleAUnit class.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
- did some smoke testing watching an all AI game on single player and a local hosted game


game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
```diff
@@ -2033,14 +2026,13 @@ private Matches() {}
   }
 
   static Predicate<Unit> unitHasWhenCombatDamagedEffect(final String filterForEffect) {
-    return u -> {
-      if (!unitHasWhenCombatDamagedEffect().test(u)) {
+    return unit -> {
+      if (!unitHasWhenCombatDamagedEffect().test(unit)) {
         return false;
       }
```
RoiEXLab, line 2032:
```suggestion
    return unitHasWhenCombatDamagedEffect().and(unit -> {
```

For style points

game-core/src/main/java/games/strategy/triplea/delegate/MoveValidator.java
```diff
@@ -1100,7 +1098,7 @@ public static BigDecimal getMaxMovement(final Collection<Unit> units) {
     }
     BigDecimal max = BigDecimal.ZERO;
     for (final Unit unit : units) {
-      final BigDecimal left = TripleAUnit.get(unit).getMovementLeft();
+      final BigDecimal left = unit.getMovementLeft();
```
RoiEXLab, line 1101: Off-Topic:

I really don't quite get why we're using BigDecimal for movement.
I'm pretty sure `double` would be precise enough for TripleA

game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
```diff
@@ -953,19 +953,19 @@ private static void sortUnitsToRemove(final List<Unit> units) {
 
   private static Comparator<Unit> getRemovableUnitsOrder() {
     return Comparator.comparing(
-        TripleAUnit::get,
+        Function.identity(),
```
RoiEXLab, line 956: Well in this case there's no point to use the surrounding `Comparator.comparing`, just return the lambda directly I'd say

game-core/src/test/java/games/strategy/triplea/delegate/MoveValidatorTest.java
```diff
@@ -58,7 +57,7 @@ void testGetLeastMovement() {
     final Collection<Unit> collection = bomber.create(1, british);
     assertEquals(new BigDecimal(6), MoveValidator.getLeastMovement(collection));
     final Object[] objs = collection.toArray();
-    ((TripleAUnit) objs[0]).setAlreadyMoved(BigDecimal.ONE);
+    ((Unit) objs[0]).setAlreadyMoved(BigDecimal.ONE);
```
RoiEXLab, line 60:
```suggestion
    collection.iterator().next().setAlreadyMoved(BigDecimal.ONE);
```

And remove the line above

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProTransportUtils.java
```diff
@@ -244,12 +243,12 @@ public static int findUnitsTransportCost(final List<Unit> units) {
...
     if (Matches.unitIsLandTransportWithoutCapacity().test(unit)) {
       units.sort(
-          Comparator.<Unit>comparingInt(u -> TripleAUnit.get(u).getMovementLeft().intValue())
+          Comparator.<Unit>comparingInt(u -> u.getMovementLeft().intValue())
               .thenComparing(getDecreasingAttackComparator(player)));
       results.add(units.get(0));
     } else {
       units.sort(
-          Comparator.<Unit>comparingInt(u -> TripleAUnit.get(u).getMovementLeft().intValue())
+          Comparator.<Unit>comparingInt(u -> u.getMovementLeft().intValue())
```
RoiEXLab, line 251: @ron-murhammer You might have any insight here.
Is it important that movement left is rounded to an  int before comparing units?

Otherwise it would be simpler writing something like
```
Comparator.comparing(Unit::getMovementLeft)
```

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMatches.java
```diff
@@ -561,14 +560,14 @@ private ProMatches() {}
...
-        .and(Matches.unitIsTransporting());
+        .and(Unit::isTransporting);
   }
 
   public static Predicate<Unit> unitIsOwnedAndMatchesTypeAndNotTransporting(
       final GamePlayer player, final UnitType unitType) {
     return Matches.unitIsOwnedBy(player)
         .and(Matches.unitIsOfType(unitType))
-        .and(Matches.unitIsTransporting().negate());
+        .and(Unit::isTransporting);
```
RoiEXLab, line 570: I think this change was unintended

game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
DanVanAtta: A number of these I noticed. Tried to keep the scope as limited as possible to avoid getting too deep and mixing too many different kinds of changes in.

game-core/src/test/java/games/strategy/triplea/delegate/MoveValidatorTest.java
DanVanAtta: Going to shy away from this update, there are many more updates we can do. This line is only swapping the casted type, I do not want to do more than just that (it opens up the flood gates, in which case such a giant is unlikely to be accurate, ever be finished, and would not be reviewable)

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMatches.java
DanVanAtta: This is a good catch @RoiEXLab :+1: 
An example of me going slightly too far in simplifications.

game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
DanVanAtta: This update resulted in:
```
Step 'google-java-format' found problem in 'game-core/src/main/java/games/strategy/triplea/delegate/Matches.java':
```

The inline suggestion is nice, though I think we probably should reserve it for quick typos. This PR is really risky for scope creep, there are tons of follow on simplifications that should be done.

DanVanAtta: Needed a missing close paren.

I like the change, but it's not for this PR. I don't like how the indentation is updated and increases the diff pretty drastically. Collapsing methods from Matches into TripleAPlayer was arguably too far of an update as well. With that the focus is really just on the one change at a time.

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMatches.java
DanVanAtta: Fixed in: 278da46b9

game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
DanVanAtta: Agree, updated in: a82b9830a

game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProMatches.java
DanVanAtta: Actually fixed in: 7d7174db1

RoiEXLab:
@DanVanAtta I'd appreciate if you could follow-up with my suggestions in another PR

DanVanAtta:
> @DanVanAtta I'd appreciate if you could follow-up with my suggestions in another PR

Don't mean to be disrespectful, but I don't think it's a reasonable ask. It's important to remember PR's have specific scopes for reasons. I could easily point out *many* additional improvements to other code changes and it's then unreasonable to say "do these things which I think should be done" when the other person may not have the time or priority.

The scope of change here was limited and had a specific goal. Going outside of that goal is risky. We also get into subjectivity about what else should be updated. It's one thing to say that existing patterns/classes become broken, it's another to say, in this update we are only replacing this class with that, and the fact we can then simplify later is what comes later.

It also gets to be very subjective, maybe some of these next steps we are going to delete code, in which case we are improving code that is going to be deleted. Generally if an item is out of scope from the goal of a PR, it is out of scope and you can follow-up post-merge. A reviewer should keep in mind what the intended goal/purpose of a change is and review just that.

From another perspective, I am repeating myself; had I chased down every loose end, this would have been either: a) never finished, b) reverted due to some odd bug, c) un-reviewable. Had I left some loose ends, then those could be other "by the way, while you're here" kinds of changes.

RoiEXLab:
@DanVanAtta In general I'd agree, but in this particular case it's to more or less purely syntactical changes I'd consider safe to do.
I understand your point of view though, so if you don't feel like creating such a PR just tell me and I'll create it instead. It shouldn't take longer than a couple of minutes for neither of us.

DanVanAtta:
Feel free to go for them @RoiEXLab

Issue #6310, opened 2020-04-22
RuntimeException - HeadedGameRunner.lambda$initializeClientSettingAndLogging$0: Exception on remote


tripleabuilderbot:
## Log Message
Exception on remote

## TripleA Version
2.0.18907

## Java Version
11.0.6

## Operating System
Linux

## Exception 
java.lang.RuntimeException: Exception on remote

## Stack Trace
games.strategy.engine.message.UnifiedInvocationHandler.invoke(UnifiedInvocationHandler.java:58)
com.sun.proxy.$Proxy15.placeUnits(Unknown Source)
java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
java.base/java.lang.reflect.Method.invoke(Method.java:566)
games.strategy.engine.player.DefaultPlayerBridge$GameOverInvocationHandler.invoke(DefaultPlayerBridge.java:152)
com.sun.proxy.$Proxy15.placeUnits(Unknown Source)
games.strategy.triplea.TripleAPlayer.place(TripleAPlayer.java:557)
games.strategy.triplea.TripleAPlayer.start(TripleAPlayer.java:150)
games.strategy.engine.framework.ServerGame.waitForPlayerToFinishStep(ServerGame.java:538)
games.strategy.engine.framework.ServerGame.runStep(ServerGame.java:408)
games.strategy.engine.framework.ServerGame.startGame(ServerGame.java:297)
games.strategy.engine.framework.startup.launcher.LocalLauncher.launchInternal(LocalLauncher.java:61)
games.strategy.engine.framework.startup.launcher.LocalLauncher.launchInternal(LocalLauncher.java:34)
games.strategy.engine.framework.startup.launcher.AbstractLauncher.lambda$launch$0(AbstractLauncher.java:16)
java.base/java.lang.Thread.run(Thread.java:834)

PR #6314, opened 2020-04-22
Fix stacktrace

RoiEXLab:
Noticed in #6310 the stacktrace only contains the main execption, not the causes

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

None

### DIFF
--- a/game-core/src/main/java/org/triplea/debug/error/reporting/StackTraceErrorReportFormatter.java
+++ b/game-core/src/main/java/org/triplea/debug/error/reporting/StackTraceErrorReportFormatter.java
@@ -93,7 +93,7 @@ class StackTraceErrorReportFormatter implements BiFunction<String, LogRecord, Er
 
   private static String throwableToString(final Throwable e) {
     final var outputStream = new ByteArrayOutputStream();
-    try (PrintWriter printWriter = new PrintWriter(outputStream)) {
+    try (PrintWriter printWriter = new PrintWriter(outputStream, false, StandardCharsets.UTF_8)) {
       e.printStackTrace(printWriter);
     }
     return "## Exception \n"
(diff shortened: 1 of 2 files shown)


game-core/src/main/java/org/triplea/debug/error/reporting/StackTraceErrorReportFormatter.java
```diff
@@ -91,18 +92,15 @@ private String buildBody(@Nullable final String userDescription, final LogRecord
   }
 
   private static String throwableToString(final Throwable e) {
+    final var outputStream = new ByteArrayOutputStream();
+    try (PrintWriter printWriter = new PrintWriter(outputStream, false, StandardCharsets.UTF_8)) {
```
DanVanAtta, line 96: - How do you know this works @RoiEXLab ? Are you going off of Stack Overflow, are there other examples in the code, or is this a pattern you just know?
- Can you simulate an exception going through this and give an example? Perhaps better yet, evaluate this to a strategy object and then test it. (EG:  this method becomes a `Function`, then testing the function you can verify that key words from a nested exception are all presented/printed)

RoiEXLab: 1. It's just a pattern I know. Basically I just looked at the overloads of `Throwable#printStackTrace`, checked the `PrintWriter` constructor and added the `ByteArrayOutputStream` that would store it in a string for me.
2. There are already test cases in place that check for the existence of some lines:
https://github.com/triplea-game/triplea/blob/b133fd9387cc6d4295e6f2d71d193e2ff8e8b977/game-core/src/test/java/org/triplea/debug/error/reporting/StackTraceErrorReportFormatterTest.java#L137-L150
3. I just realized I should maybe extend this test case so it verifies the causes are also checked

DanVanAtta: Thanks, good work @RoiEXLab

Issue #6322, opened 2020-04-24
Units killed on first round of combat can appear twice in casualty view


asvitkine:
### How can the problem be recreated?
1. Load attached save game:
[buggy.tsvg.zip](https://github.com/triplea-game/triplea/files/4526209/buggy.tsvg.zip)
2. Do battle in Changchun
3. Observe that on first round, enemy casualties are duplicated (for example, in the screenshot below, we see the 4 conscripts appear twice)

<img width="912" alt="Screen Shot 2020-04-23 at 10 02 14 PM" src="https://user-images.githubusercontent.com/17648/80167328-1e74db80-85ae-11ea-90fd-5cedd6f85270.png">

### What is an expected fix?

It should appear once.

The issue seems to be that the same unit casualties are added both in the constructor of BattleDisplay and as a result of casualtyNotification().

As an aside, the save game has another oddity - loading it immediately bring up a battle, which also seems unexpected/undesired to me.

### Which Engine Version are you using? 
2.0.18871

PR #6323, opened 2020-04-24
Fix issue where casualties may be duplicated in battle panel.

asvitkine:
It was caused by collections that were later modified being
passed to async code and by the time the async Swing code ran,
those collections no longer had the same content as when the
call took place.

Also cleans up some params that were nullable to not be.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix:  https://github.com/triplea-game/triplea/issues/6322
[] Other:   

## Testing
Manual testing of problem.


DanVanAtta:
Thanks for bringing this fix @asvitkine , really appreciate it :100:

PR #6328, opened 2020-04-25
Simplify ConcurrentBattleCalculator code

RoiEXLab:
This PR tries to simplify the code within `ConcurrentBattleCalculator` by removing some heuristics and using more "standard" approaches in general.

I ran some benchmarks and the time seems to be about equal (which is to be expected because the approach didn't really change)

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

I verified BattleCalc works as usual and let the Hard AI play a game of classic WW2, which worked without any problems.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/GameDataUtils.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/GameDataUtils.java
@@ -28,6 +28,20 @@ public final class GameDataUtils {
     return dataCopy;
   }
 
+  public static byte[] serializeGameDataWithoutHistory(final GameData data) {
+    final History temp = data.getHistory();
+    data.resetHistory();
+    final byte[] bytes;
+    try {
+      bytes = IoUtils.writeToMemory(os -> GameDataManager.saveGame(os, data, false));
+    } catch (final IOException e) {
+      throw new RuntimeException("Failed to serialize GameData", e);
+    } finally {
+      data.setHistory(temp);
+    }
+    return bytes;
+  }
+
   public static GameData cloneGameData(final GameData data) {
     return cloneGameData(data, false);
   }
--- a/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
+++ b/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
@@ -11,7 +11,6 @@ import games.strategy.engine.data.Unit;
 import games.strategy.engine.framework.GameDataManager;
 import games.strategy.engine.framework.GameDataUtils;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Set;
@@ -20,6 +19,8 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import javax.annotation.concurrent.ThreadSafe;
 import lombok.extern.java.Log;
 import org.triplea.io.IoUtils;
@@ -108,24 +109,25 @@ public class ConcurrentBattleCalculator implements IBattleCalculator {
     Preconditions.checkState(!isShutDown, "ConcurrentBattleCalculator is already shut down");
     Preconditions.checkState(bytes.length != 0, "Data has not been set yet.");
     final long start = System.currentTimeMillis();
-    final List<Future<AggregateResults>> results = new ArrayList<>();
     final int runsPerWorker = runCount / MAX_THREADS;
+    final List<Future<AggregateResults>> results;
     synchronized (mutex) {
-      int remainingRuns = runCount;
-      while (remainingRuns > 0) {
-        final int individualRemaining = Math.min(remainingRuns, runsPerWorker);
-        remainingRuns -= runsPerWorker;
-        results.add(
-            createBattleCalcWorker(
-                attacker,
-                defender,
-                location,
-                attacking,
-                defending,
-                bombarding,
-                territoryEffects,
-                individualRemaining));
-      }
+      results =
+          IntStream.range(0, MAX_THREADS)
+              .map(index -> index == 0 ? runCount % MAX_THREADS : 0)
+              .map(runs -> runs + runsPerWorker)
+              .mapToObj(
+                  individualRemaining ->
+                      createBattleCalcWorker(
+                          attacker,
+                          defender,
+                          location,
+                          attacking,
+                          defending,
+                          bombarding,
+                          territoryEffects,
+                          individualRemaining))
+              .collect(Collectors.toList());
     }
     final AggregateResults result = aggregateResults(results, runsPerWorker);
     result.setTime(System.currentTimeMillis() - start);
--- a/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
+++ b/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
@@ -141,7 +141,7 @@ public class ConcurrentBattleCalculator implements IBattleCalculator {
       final Collection<Unit> bombarding,
       final Collection<TerritoryEffect> territoryEffects,
       final int runs) {
-    BattleCalculator calculator = new BattleCalculator();
+    final BattleCalculator calculator = new BattleCalculator();
     calculator.setKeepOneAttackingLandUnit(keepOneAttackingLandUnit);
     calculator.setAmphibious(amphibious);
     calculator.setRetreatAfterRound(retreatAfterRound);
@@ -174,12 +174,12 @@ public class ConcurrentBattleCalculator implements IBattleCalculator {
   private static AggregateResults aggregateResults(
       final List<Future<AggregateResults>> results, final int runsPerWorker) {
     final AggregateResults result = new AggregateResults(runsPerWorker);
-    for (Future<AggregateResults> future : results) {
+    for (final Future<AggregateResults> future : results) {
       try {
         result.addResults(future.get().getResults());
-      } catch (InterruptedException e) {
+      } catch (final InterruptedException e) {
         Thread.currentThread().interrupt();
-      } catch (ExecutionException e) {
+      } catch (final ExecutionException e) {
         throw new IllegalStateException("Exception from worker", e);
       }
     }
--- a/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
+++ b/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
@@ -22,7 +22,6 @@ import java.util.concurrent.Future;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import javax.annotation.concurrent.ThreadSafe;
-import lombok.extern.java.Log;
 import org.triplea.io.IoUtils;
 
 /**
@@ -30,7 +29,6 @@ import org.triplea.io.IoUtils;
  * the run count across these workers. This is mainly to be used by AIs since they call the
  * OddsCalculator a lot.
  */
-@Log
 @ThreadSafe
 public class ConcurrentBattleCalculator implements IBattleCalculator {
   private static final int MAX_THREADS = Math.max(1, Runtime.getRuntime().availableProcessors());
--- a/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
+++ b/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
@@ -137,11 +137,6 @@ public class ConcurrentBattleCalculator implements IBattleCalculator {
   }
 
   @Override
-  public boolean getIsReady() {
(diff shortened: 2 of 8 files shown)


game-core/src/main/java/games/strategy/engine/framework/GameDataUtils.java
```diff
@@ -28,6 +28,20 @@ public static GameData cloneGameDataWithoutHistory(
     return dataCopy;
   }
 
+  public static byte[] serializeGameDataWithoutHistory(final GameData data) {
```
DanVanAtta, line 31: This is an interesting way to reduce the game data serialization size. I've been thinking we'll want to eventually have a param object for calculations that has a copy of *just* the data it needs. In the meantime this is interesting to move us in that direction.

game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
```diff
@@ -70,393 +64,182 @@ public ConcurrentBattleCalculator(final String threadNamePrefix) {
...
-      // should make sure that all workers have their game data set before we can call calculate and
-      // other things
-      isDataSet = true;
-      dataLoadedAction.run();
-    }
-    // allow setting new data to take place if it is waiting on us
-    latchWorkerThreadsCreation.countDown();
-    // allow calcing and other stuff to go ahead
-    latchSetData.countDown();
+    return MAX_THREADS;
```
DanVanAtta, line 82: side-note, it seems questionable to return MAX_THREADs and to also use that same constant in this class. It looks like this 'get' is only used for display purposes. When multi-threading of the battle calc was brand new, made more sense. Now I'm more thinking it's an implementation detail that the calc is multi-threaded and we don't really need to tell the user that (it's not important to them, only that the calc completes quickly and accurately).  CC: @ron-murhammer 

In that case, if we remove that, we could remove this interface method and implementation.

game-core/src/main/java/games/strategy/engine/framework/GameDataUtils.java
RoiEXLab: In an ideal world this "copy" process would be super cheap to do, so that we don't have to roll back changes manually after each simulation run, but could just create a new copy instead.

The benefit of this intermediate solution is that there's absolutely no locking involved once we have the byte array.
This way all threads can create their own instance of the GameData independent of each other.

game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
RoiEXLab: Good Point. It sounds like something #6329 should include, so I'd rather make a change there and merge this PR as-is.

This count is currently used for the loading screen where it says "Calculating Odds (X Threads)" or something similar. I think displaying the thread count is a good explanation/excuse for the high CPU usage the "BattleCalculator" uses. (My Laptop has "only" 4 cores, so it's definitely noticeable that everything gets slower once the simulation runs).
That's why I wonder: Do you think it's a reasonable solution to create a class that provides the amount of available threads and make the UI "assume" that the actual implementation is using the same amount of cores?

RoiEXLab: In some sense we're already relying on java to actually use 4 different threads for our battlecalc (there's no guarantee the threads will all be run side by side at the same time)

DanVanAtta: Personally, the "X threads" is interesting, but IMO more for devs. A good percentage of users will not know what a thread even is.

DanVanAtta: > Good Point. It sounds like something #6329 should include, so I'd rather make a change there and merge this PR as-is.

Indeed, to be explicit, any comment I prefix with "side-note" is really just a general comment; My general intent is the item is something that should *not* be addressed in this PR, but is worth a mention maybe and a follow-up.

asvitkine:
Out of curiosity, did you do any sanity checks to measure that there's no performance hit from this? I just started a new game of Domination 1914 and first turn of Germany Hard AI is taking a very long time and it seems it's all in game copying code in battle calc. Not sure if a regression - but I don't remember it being so long before, but I could check.

RoiEXLab:
@asvitkine On my machine the execution time of a single simulation run was about the same.

Basically 2 things changed:
1. Deserializing the GameData is now done within the threads of the threadpool,only a single serialization is performed on the main thread that schedules. This could make a difference if the AI runs the calculation multiple times without changing the GameData in-between. Previously we kept the workers in memory with a copy of the whole GameData each, now we only store a single copy.
2. We always create as many task as the CPU can handle in cores. Previously we had some heuristic that estimated how many GameData clones the system could handle. So it could be the case that it actually creates more threads than before on your machine, therefore increasing the total execution time

asvitkine:
I did a test, adding some code to fireGameDataEvent() to print time deltas in MS on step transitions.

Here's what it looks like on first turn of Revised on eb21301be0bc431503704dd176d63260d18065f1 (prior to the change):

```
Step GameStep:russianTech delegate:tech player:PlayerId named:Russians displayName:null delta 55
Step GameStep:russianTechActivation delegate:tech_activation player:PlayerId named:Russians displayName:null delta 301
Step GameStep:russianPurchase delegate:purchase player:PlayerId named:Russians displayName:null delta 4
Step GameStep:russianCombatMove delegate:move player:PlayerId named:Russians displayName:null delta 6151
Step GameStep:russianNonCombatMove delegate:move player:PlayerId named:Russians displayName:Non Combat Move delta 1704
Step GameStep:russianCombatMove delegate:move player:PlayerId named:Russians displayName:null delta 568
Step GameStep:russianBattle delegate:battle player:PlayerId named:Russians displayName:null delta 2687
Step GameStep:russianNonCombatMove delegate:move player:PlayerId named:Russians displayName:Non Combat Move delta 1714
Step GameStep:russianPlace delegate:place player:PlayerId named:Russians displayName:null delta 5186
Step GameStep:russianEndTurn delegate:endTurn player:PlayerId named:Russians displayName:null delta 1220
Step GameStep:germanTech delegate:tech player:PlayerId named:Germans displayName:null delta 35
Step GameStep:germanTechActivation delegate:tech_activation player:PlayerId named:Germans displayName:null delta 2
Step GameStep:germanPurchase delegate:purchase player:PlayerId named:Germans displayName:null delta 0
Step GameStep:germanCombatMove delegate:move player:PlayerId named:Germans displayName:null delta 3550
Step GameStep:germanBattle delegate:battle player:PlayerId named:Germans displayName:null delta 82
Step GameStep:germanNonCombatMove delegate:move player:PlayerId named:Germans displayName:Non Combat Move delta 1365
Step GameStep:germanCombatMove delegate:move player:PlayerId named:Germans displayName:null delta 483
Step GameStep:germanBattle delegate:battle player:PlayerId named:Germans displayName:null delta 5836
Step GameStep:germanNonCombatMove delegate:move player:PlayerId named:Germans displayName:Non Combat Move delta 3296
Step GameStep:germanPlace delegate:place player:PlayerId named:Germans displayName:null delta 7149
Step GameStep:germanEndTurn delegate:endTurn player:PlayerId named:Germans displayName:null delta 1620
Step GameStep:britishTech delegate:tech player:PlayerId named:British displayName:null delta 31
Step GameStep:britishTechActivation delegate:tech_activation player:PlayerId named:British displayName:null delta 13
Step GameStep:britishPurchase delegate:purchase player:PlayerId named:British displayName:null delta 14
```

Here it is on 9ee2d1b298eb8e5031098f9ae8f0aad55263feb8 (with the change):

```
Step GameStep:russianTech delegate:tech player:PlayerId named:Russians displayName:null delta 47
Step GameStep:russianTechActivation delegate:tech_activation player:PlayerId named:Russians displayName:null delta 304
Step GameStep:russianPurchase delegate:purchase player:PlayerId named:Russians displayName:null delta 41
Step GameStep:russianCombatMove delegate:move player:PlayerId named:Russians displayName:null delta 8859
Step GameStep:russianNonCombatMove delegate:move player:PlayerId named:Russians displayName:Non Combat Move delta 3166
Step GameStep:russianCombatMove delegate:move player:PlayerId named:Russians displayName:null delta 803
Step GameStep:russianBattle delegate:battle player:PlayerId named:Russians displayName:null delta 2521
Step GameStep:russianNonCombatMove delegate:move player:PlayerId named:Russians displayName:Non Combat Move delta 1700
Step GameStep:russianPlace delegate:place player:PlayerId named:Russians displayName:null delta 6661
Step GameStep:russianEndTurn delegate:endTurn player:PlayerId named:Russians displayName:null delta 1230
Step GameStep:germanTech delegate:tech player:PlayerId named:Germans displayName:null delta 32
Step GameStep:germanTechActivation delegate:tech_activation player:PlayerId named:Germans displayName:null delta 2
Step GameStep:germanPurchase delegate:purchase player:PlayerId named:Germans displayName:null delta 1
Step GameStep:germanCombatMove delegate:move player:PlayerId named:Germans displayName:null delta 9449
Step GameStep:germanBattle delegate:battle player:PlayerId named:Germans displayName:null delta 241
Step GameStep:germanNonCombatMove delegate:move player:PlayerId named:Germans displayName:Non Combat Move delta 2510
Step GameStep:germanCombatMove delegate:move player:PlayerId named:Germans displayName:null delta 542
Step GameStep:germanBattle delegate:battle player:PlayerId named:Germans displayName:null delta 4954
Step GameStep:germanNonCombatMove delegate:move player:PlayerId named:Germans displayName:Non Combat Move delta 3696
Step GameStep:germanPlace delegate:place player:PlayerId named:Germans displayName:null delta 8764
Step GameStep:germanEndTurn delegate:endTurn player:PlayerId named:Germans displayName:null delta 1626
Step GameStep:britishTech delegate:tech player:PlayerId named:British displayName:null delta 27
Step GameStep:britishTechActivation delegate:tech_activation player:PlayerId named:British displayName:null delta 1
Step GameStep:britishPurchase delegate:purchase player:PlayerId named:British displayName:null delta 1
```

Eyeballing these, it does seem like there's a regression. This on REVISED, where everything is pretty fast. On Domination 1914, I was seeing something like 10 minutes for the first Germans turn. I haven't done the test with the measurements, but definitely seems very slow compared to what I remembered before.

This is what I used to measure, which is kind of hacky. Not sure if there's a better way:

```
diff --git a/game-core/src/main/java/games/strategy/engine/data/GameData.java b/game-core/src/main/java/games/strategy/engine/data/GameData.java
index d670beed1..7990a024f 100644
--- a/game-core/src/main/java/games/strategy/engine/data/GameData.java
+++ b/game-core/src/main/java/games/strategy/engine/data/GameData.java
@@ -259,7 +259,22 @@ public class GameData implements Serializable {
     territoryListeners.forEach(territoryListener -> territoryListener.ownerChanged(t));
   }
 
+  static long oldT = System.currentTimeMillis();
+  static String oldDisplayName = "";
+
   public void fireGameDataEvent(final GameDataEvent event) {
+
+    final GameStep step = getSequence().getStep();
+    if (step != null
+        && step.getDelegate() != null
+        && step.getDisplayName() != null
+        && !oldDisplayName.equals(step.getDisplayName())) {
+      long t = System.currentTimeMillis();
+      System.out.println("Step " + step + " delta " + (t - oldT));
+      oldT = t;
+      oldDisplayName = step.getDisplayName();
+    }
+
     gameDataEventListeners.accept(event);
   }
 
```

asvitkine:
Here's a save game that demonstrates the extreme regression on Domination 1914. Just load it up and let AI do its turn as Turkey. It's very fast prior to this change and really slow after. Try it yourself.

[dom_arabia_1.tsvg.zip](https://github.com/triplea-game/triplea/files/4549311/dom_arabia_1.tsvg.zip)

Given this, I think we should revert this to avoid the huge perf regression. :(

RoiEXLab:
I mean we could sort-of revert the change I mentioned, so we continue to clone the GameData in the beginning, that should fix any potential performance regression.
Note that #6329 changes the Battle calc interface completely, so we should wait until that PR is merged

asvitkine:
Sure, if there's a way we can keep most of the benefits of the clean up and
just revert the cloning logic, that would be great!

DanVanAtta:
Ya'll have any idea why the cloning logic would be slower?

Really hate to think we're in this kind of situation with optimized code. "Better to optimize correct code than it is to correct optimized code"

"Simplicity is prerequisite for reliability." - Dijkstra

"How do we convince people that in programming simplicity and clarity —in short: what mathematicians call "elegance"— are not a dispensable luxury, but a crucial matter that decides between success and failure? " - Djikstra

DanVanAtta:
@asvitkine FWIW there is a utility class checked in to do ad-hoc measurements: [PerfTimer.java](https://github.com/triplea-game/triplea/blob/master/game-core/src/main/java/org/triplea/performance/PerfTimer.java)

RoiEXLab:
> Ya'll have any idea why the cloning logic would be slower?

Basically previously `IBattleCalculator#setGameData` did the expensive cloning and now this cloning is done as part of `IBattleCalculator#calculate`. My (apparently false) assumption was that for every calculate setGameData is called once, but it seems that this is not necessarily the case on certain maps.
It would be nice if @asvitkine could confirm my theory though, but besides that I didn't really change anything that could affect performance so badly.

RoiEXLab:
@asvitkine #6351

PR #6329, opened 2020-04-25
Simplify IBattleCalculator interface

RoiEXLab:
Builds on top of #6328
I realized that in a lot of places the interface declaration was really only used on a specific type, so I decided to reduce the `IBattleCalculator` interface to its minimum.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

~No testing so far. If #6328 changes in any way, I'll have to redo that anyways.~
I verified the Hard AI could play a game of classic WW2 3rd


game-core/src/main/java/games/strategy/triplea/ai/pro/logging/ProLogSettings.java
```diff
@@ -57,7 +57,7 @@ static void saveSettings(final ProLogSettings settings) {
                   outputStream.writeObject(settings);
                 }
               });
-      final Preferences prefs = Preferences.userNodeForPackage(ProAi.class);
```
RoiEXLab, line 60: Note to future self:
This change was unintended

game-core/src/main/java/games/strategy/triplea/ui/menubar/DebugMenu.java
```diff
@@ -18,10 +18,10 @@
     setMnemonic(KeyEvent.VK_D);
 
     final Set<Player> players = frame.getLocalPlayers().getLocalPlayers();
-    final boolean areThereProAIs = players.stream().anyMatch(ProAi.class::isInstance);
+    final boolean areThereProAIs = players.stream().anyMatch(AbstractProAi.class::isInstance);
     if (areThereProAIs) {
-      ProAi.initialize(frame);
-      add(SwingAction.of("Show Hard AI Logs", ProAi::showSettingsWindow))
+      AbstractProAi.initialize(frame);
+      add(SwingAction.of("Show Hard AI Logs", AbstractProAi::showSettingsWindow))
```
RoiEXLab, line 24: Interesting observation.
"Fast AI" is a subclass of the new AbstractProAi, previously just ProAi.
So it would fall under the same category. Is this intended?

RoiEXLab:
@DanVanAtta Added your suggestion from #6328 
Suggestions welcome

game-core/src/main/java/games/strategy/triplea/ai/pro/AbstractProAi.java
```diff
@@ -0,0 +1,513 @@
...
+
+  @Override
+  public void stopGame() {
+    super.stopGame(); // absolutely MUST call super.stopGame() first
+    calc.stop();
+  }
+
+  @Override
+  public PlayerType getPlayerType() {
+    return PlayerType.PRO_AI;
```
DanVanAtta, line 99: Should this remain as an abstract method?
At first glance it seems like Fast AI would want to return its own player type.

game-core/src/main/java/games/strategy/triplea/ai/pro/AbstractProAi.java
```diff
@@ -0,0 +1,513 @@
...
+  }
+
+  public ProOddsCalculator getCalc() {
+    return calc;
+  }
+
+  public static void initialize(final TripleAFrame frame) {
+    ProLogUi.initialize(frame);
```
DanVanAtta, line 107: Side-note, `ProLogXX` should probably be specific to the ProAI; it seems we have a more generic tool and it should instead be the `AiLogUi` and/or `AiLogger`

game-core/src/main/java/games/strategy/triplea/ai/pro/AbstractProAi.java
```diff
@@ -0,0 +1,513 @@
...
+
+  private void initializeData() {
+    proData.initialize(this);
+  }
+
+  public void setStoredStrafingTerritories(final List<Territory> strafingTerritories) {
+    storedStrafingTerritories = strafingTerritories;
+  }
+
+  protected abstract void prepareData(GameData data);
```
DanVanAtta, line 124: This method could perhaps use a javadoc to explain what we would expect implementations to do, and to define `prepareData`

game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -54,7 +54,7 @@
   private boolean isDefensive;
   private boolean isBombing;
 
-  ProCombatMoveAi(final ProAi ai) {
+  ProCombatMoveAi(final AbstractProAi ai) {
```
DanVanAtta, line 57: nit, we probably should be shooting for the dependency to be on an interface where it so happens that `AbstractProAi` implements that interface. To get there we may have to think a bit more about the modelling of this one AI and which interfaces it makes sense to have.

To another extent, we should perhaps also consider 'interface segregation principal' and how many interfaces we would need so that a dependency like this is only given an interface where it actually uses everything that is passed to it.

game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
```diff
@@ -69,20 +68,13 @@ public ConcurrentBattleCalculator(final String threadNamePrefix) {
     this.dataLoadedAction = dataLoadedAction;
   }
 
-  @Override
   public void setGameData(final GameData data) {
     synchronized (mutex) {
```
DanVanAtta, line 72: Question/side-comment: if we are making a copy of `GameData` for the battle calculator, do you know why we need to lock it?

game-core/src/main/java/games/strategy/triplea/ui/menubar/DebugMenu.java
DanVanAtta: I would suspect so. It would kinda seem Fast AI is not really a true AI but a variant of ProAI. Given that, I would personally look to see if we can inject behavior to a ProAI class.

EG:
```
FastAI extends ProAi{
   FastAI() {
      super(ProAiConfig.builder()
             .battleCalcImplementation(new FastAiBattleCalc())
             .build());
   }
}

LaboriousAi extends ProAi{
   LaboriousAiI() {
      super(ProAiConfig.builder()
             .battleCalcImplementation(new LaboriousBattleCalc())
             .build());
   }
}
```

Or even better, do the above as constructor args.

game-core/src/main/java/games/strategy/triplea/ai/pro/AbstractProAi.java
RoiEXLab: Right, wanted to make this change, but forgot to do so

game-core/src/main/java/games/strategy/triplea/ai/pro/AbstractProAi.java
RoiEXLab: Maybe @ron-murhammer has any insight?

game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
RoiEXLab: I added a javadoc to the mutex field that should clarify this question.
Basically the mutex is not for thread safety, but rather for thread consistency.
Without this mutex it could happen that the gamedata is changed when we're actively deserializing the gamedata to create `BattleCalculator` workers, so the first x workers have the old data and the last y workers use the newer data which would screw up the results

RoiEXLab:
> How much is left to do in draft PR before this is ready to PR?

I don't have anything planned.
The only reason I made this PR draft is to prevent someone from accidentally merging it before #6328 so the only thing that's left here is to verify I don't get any exceptions when running a normal game with Hard AI

game-core/src/main/java/games/strategy/triplea/ai/pro/AbstractProAi.java
ron-murhammer: I believe the log UI initialize is used for both the pro and fast ai.

RoiEXLab:
@DanVanAtta Updated, let me know if you think this PR needs another iteration

RoiEXLab:
@DanVanAtta I'm going to merge this in order to make room for a fix to #6350/#6349 and the performance regression mentioned by @asvitkine 
Given the performance regression, a partial revert of  #6328 seems justified now (unless someone wants to go through the hassle of restoring original behaviour with less code), I'd just like to keep the current interface structure which should be doable. by removing some `@Override` annotations and adjusting method signatures slightly.

Looking at the diff of #6328 I realized a couple of minor differences in behaviour that could explain the performance regression as well: For example the concurrent calc does cancel its operations whenever a new GameData is being set, so that might lead to longer times if multiple calls to setGameData and calculate are performed in quick succession.

I don't know if I'll be able to submit a PR before monday, but I'll try nevertheless

RoiEXLab:
Okay, this was easier than expected #6351

PR #6334, opened 2020-04-26
Upload lobby game chat messages to lobby & store in DB

DanVanAtta:
- Adds a table to store lobby game chat messages
- Adds a DAO to insert those messages
- Adds an upload module that will verify a given api-key and game-id
  are valid on each upload request, if so we insert a chat message
- Adds an endpoint controller to receive upload requests
- Adds a new http client to upload chat messages to the upload endpoint.
- Updates ServerModel to chec if we have a game-to-lobby-connection, if
  so then we add a new chat listener that will use a background thread
  to upload messages to lobby

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
- launched a local lobby
  - hosted a game on that lobby, verified chat messages are uploaded into database
  - disconnected lobby, verified that failure to upload errors were silently handled (though we do print an info message about the failure)
- connected a bot to a local lobby, joined, and verified messages are uploaded into database
- hosted a local network game, verified that it does not try to upload messages

I did notice on the disconnect case we retry a number of times to do the upload. I think we'll want to follow up and tune the retry conditions and how many we do. Perhaps to just a single retry after a fixed back-off.


http-server/src/main/java/org/triplea/db/dao/lobby/games/LobbyGameDao.java
```diff
@@ -0,0 +1,60 @@
...
+    Postconditions.assertState(
+        insertCount == 1, "Failed to insert lobby game: " + lobbyGameListing);
+  }
+
+  @SqlUpdate(
+      "insert into lobby_game(host_name, game_id, game_hosting_api_key_id) "
+          + "values ("
+          + "  :hostName,"
+          + "  :gameId,"
+          + "  (select id from game_hosting_api_key where key = :apiKey))")
```
RoiEXLab, line 35: Ok this formatting confused me for a second

http-server/src/main/java/org/triplea/db/dao/lobby/games/LobbyGameDao.java
```diff
@@ -0,0 +1,60 @@
...
+
+  default void insertLobbyGame(ApiKey apiKey, LobbyGameListing lobbyGameListing) {
+    final String hashedkey = new ApiKeyHasher().apply(apiKey);
+    final int insertCount =
+        insertLobbyGame(
+            lobbyGameListing.getLobbyGame().getHostName(), //
+            lobbyGameListing.getGameId(),
+            hashedkey);
+    Postconditions.assertState(
+        insertCount == 1, "Failed to insert lobby game: " + lobbyGameListing);
```
RoiEXLab, line 27: Is this to check if `(select id from game_hosting_api_key where key = :apiKey)` did actually return a single entry?

Might be worth adding a comment to explain this.
I don't think any other place in the code actually checks the insert count

http-server/src/main/java/org/triplea/modules/game/lobby/watcher/ChatUploadModule.java
```diff
@@ -0,0 +1,40 @@
...
+class ChatUploadModule {
+  private final LobbyGameDao lobbyGameDao;
+  private final BiPredicate<ApiKey, String> gameIdValidator;
+
+  static ChatUploadModule build(final Jdbi jdbi, final GameListing gameListing) {
+    return new ChatUploadModule(
+        jdbi.onDemand(LobbyGameDao.class), gameListing::isValidApiKeyAndGameId);
+  }
+
+  public boolean upload(final ChatMessageUpload chatMessageUpload) {
```
RoiEXLab, line 32: I just noticed that the return value is used to potentially display  a warning message.
Why did you decide it's better to do the warning outside?

http-server/src/main/java/org/triplea/db/dao/lobby/games/LobbyGameDao.java
DanVanAtta: The string concats do make it kinda difficult. One 'values' arg per line is not too uncommon though.

http-server/src/main/java/org/triplea/db/dao/lobby/games/LobbyGameDao.java
DanVanAtta: If the subselect returns 0 elements we'll get a non-null constraint violation.

> I don't think any other place in the code actually checks the insert count

We should have at least a few. When adding a good number before I actually removed them so that I could PR the `postconditions` class at the same time.. Slightly interesting history; in general we should be checking insert count and not just blindly trusting it succeeded.

http-server/src/main/java/org/triplea/modules/game/lobby/watcher/ChatUploadModule.java
DanVanAtta: Separation of responsibility I suppose. This class has trouble knowing if the 'silent fail' is actually an error condition or not.  Beyond that, we don't have all the information we'd want to log, notably the IP of who sent us this request as this condition is perhaps likely to be an attacker.

PR #6344, opened 2020-04-28
Is instant matcher

DanVanAtta:
## Commits

commit c420bd1e7bd325923d278e1ac0bcab9ca1de7610

    Add IsInstant matcher to facilitate matching instant values
    
    - Rather than parsing an instant value from a formatted string
    and comparing against a target instant, this matcher allows for
    an instant to be matched against a numerically specified instant,
    eg: 'assertThat(Instant.now(), isInstant(2020, 2, 20, 23, 59, 0));'

commit 5a4f7e6d962542b18cb9e73cbecc642e7083ff2e

    Convert tests to use 'isInstant(...)' in preference to 'is(Instant.parse(...))'

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:

### DIFF
--- a/test-common/src/main/java/org/triplea/test/common/IsInstant.java
+++ b/test-common/src/main/java/org/triplea/test/common/IsInstant.java
@@ -3,10 +3,10 @@ package org.triplea.test.common;
 import static org.hamcrest.core.Is.is;
 import static org.hamcrest.core.IsEqual.equalTo;
 
-import com.google.common.base.Preconditions;
 import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
 import lombok.experimental.UtilityClass;
-import org.apache.commons.lang3.StringUtils;
 import org.hamcrest.Matcher;
 
 @UtilityClass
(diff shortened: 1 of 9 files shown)


test-common/src/main/java/org/triplea/test/common/IsInstant.java
```diff
@@ -0,0 +1,59 @@
...
+        equalTo(
+            Instant.parse(
+                String.format(
+                    "%s-%s-%sT%s:%s:%sZ",
+                    StringUtils.leftPad(String.valueOf(year), 4, '0'),
+                    StringUtils.leftPad(String.valueOf(month), 2, '0'),
+                    StringUtils.leftPad(String.valueOf(day), 2, '0'),
+                    StringUtils.leftPad(String.valueOf(hour), 2, '0'),
+                    StringUtils.leftPad(String.valueOf(minute), 2, '0'),
+                    StringUtils.leftPad(String.valueOf(second), 2, '0')))));
```
RoiEXLab, line 57: Thoughts about using https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/LocalDateTime.html#of(int,int,int,int,int,int)
Instead and converting that to an instant?

This way you wouldn't need the preconditions and custom parse logic.

Something like
```java
var ldt = LocalDateTime.of(year, month, day, hour, minute, second);
var instant = ldt.toInstant(ZoneOffset.UTC);
```

DanVanAtta: Good suggestion :+1: 
9c34b42

Issue #6349, opened 2020-04-29
IllegalStateException - HeadedGameRunner.lambda$initializeClientSettingAndLogging$0: Data has not been set yet.


tripleabuilderbot:
## User Description
I dont know what this error is about, but it happened while starting up a map and as the AI was doing placments of held units.

## Log Message
Data has not been set yet.

## TripleA Version
2.0.19003

## Java Version
11.0.6

## Operating System
Windows 10

## Exception 
java.lang.IllegalStateException: Data has not been set yet.

## Stack Trace
```
java.lang.IllegalStateException: Data has not been set yet.
	at com.google.common.base.Preconditions.checkState(Preconditions.java:508)
	at games.strategy.triplea.odds.calculator.ConcurrentBattleCalculator.calculate(ConcurrentBattleCalculator.java:108)
	at games.strategy.triplea.ai.pro.util.ProOddsCalculator.callBattleCalc(ProOddsCalculator.java:232)
	at games.strategy.triplea.ai.pro.util.ProOddsCalculator.callBattleCalc(ProOddsCalculator.java:205)
	at games.strategy.triplea.ai.pro.util.ProOddsCalculator.calculateBattleResults(ProOddsCalculator.java:138)
	at games.strategy.triplea.ai.pro.util.ProOddsCalculator.calculateBattleResults(ProOddsCalculator.java:121)
	at games.strategy.triplea.ai.pro.ProPurchaseAi.prioritizeTerritoriesToDefend(ProPurchaseAi.java:557)
	at games.strategy.triplea.ai.pro.ProPurchaseAi.place(ProPurchaseAi.java:471)
	at games.strategy.triplea.ai.pro.ProAi.place(ProAi.java:284)
	at games.strategy.triplea.ai.AbstractAi.start(AbstractAi.java:523)
	at games.strategy.engine.framework.ServerGame.waitForPlayerToFinishStep(ServerGame.java:538)
	at games.strategy.engine.framework.ServerGame.runStep(ServerGame.java:408)
	at games.strategy.engine.framework.ServerGame.startGame(ServerGame.java:297)
	at games.strategy.engine.framework.startup.launcher.LocalLauncher.launchInternal(LocalLauncher.java:61)
	at games.strategy.engine.framework.startup.launcher.LocalLauncher.launchInternal(LocalLauncher.java:34)
	at games.strategy.engine.framework.startup.launcher.AbstractLauncher.lambda$launch$0(AbstractLauncher.java:16)
[...]

PR #6351, opened 2020-04-29
Partial revert of battle calc simplifications

RoiEXLab:
This is a partial revert of #6328 which should fix the performance regression reported by @asvitkine and the issue reported in #6349

@asvitkine It would be nice if you could verify this actually solves the regression for you. It should, but there's also a slight chance it doesn't.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  #6349 https://github.com/triplea-game/triplea/pull/6328#issuecomment-620933476
[] Other:

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
+++ b/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
@@ -247,9 +247,6 @@ public class ConcurrentBattleCalculator implements IBattleCalculator {
           // causing it to abort with null data
           return new AggregateResults(0);
         }
-        if (!worker.getIsReady()) {
-          throw new IllegalStateException("Called calculate before setting calculate data!");
-        }
         final int currentWorkedRunCount = (runCount <= 0 ? 0 : workerRunCount);
         if (currentWorkedRunCount > 0) {
           totalRunCount += currentWorkedRunCount;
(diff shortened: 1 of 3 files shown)


game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
```diff
@@ -95,21 +234,34 @@ public AggregateResults calculate(
...
-                  individualRemaining ->
-                      createBattleCalcWorker(
+    synchronized (mutexCalcIsRunning) {
+      awaitLatch();
+      final long start = System.currentTimeMillis();
+      // Create worker thread pool and start all workers
+      int totalRunCount = 0;
+      int runCount = initialRunCount;
+      final int workerNum = workers.size();
+      final int workerRunCount = Math.max(1, (runCount / Math.max(1, workerNum)));
```
RoiEXLab, line 246: This code here changed slightly from the original implementation, basically we compute all of the run counts here now.
Basically it does the same as my IntStream implementation to the left with some added exit conditions

asvitkine:
I patched this in at current tip of tree and compared with eb21301 which was before the regression and tried my saved game.

I'm seeing an even faster performance now than before - under 30s whereas closer to 45s before. (And with the regression it was several minutes I think.)

So this seems great!

RoiEXLab:
@asvitkine That's cool to hear, the restructured interface allowed me to remove some unused code and intermediate steps (such as double synchronization on the same object), that could be one of the reasons for the increased performance if the code is called very often

asvitkine:
Actually, maybe it's too good to be true? I tried the Odds calculator in game and it gave 0% odds for something that shouldn't be 0. So something is broken?

RoiEXLab:
@asvitkine You're right, I accidentally dropped one line where a boolean flag was set. It doesn't make sense anymore so I can just remove it. Please check the latest commit

asvitkine:
Tested the new patch. 45s - so similar to before the regression, just a bit faster.

PR #6354, opened 2020-04-30
Add lobby right-click menu for moderators: "Show Player Information"

DanVanAtta:
This update adds an equivalent right-click menu feature that was in 1.9
where moderators can click chatter names in lobby and select 'show player
information'.

The player information panel shows a copy-pastable summary of the
players name, IP and system-id and currently two tabs: aliases & bans.

The aliases are obtained from a lobby query and database fetch where we
search the 'access log' table for the player IPs or system-ids matching
the current user and return matching records.

We fetch player aliases limited to the last 'n' days, bans we will
fetch all that we have on record.

Change listing:
- add UI component to show player information
- add two tab UI components to show player ban and alias information in a table
- add a client query to ModeratorChatClient to fetch player information
- add a controller endpoint to service this query, restriced to moderator access
- add a player information module on server side to aggregate and fetch
  player information from database.
- add a player information DAO with JDBI return data types to query database
  for bans and aliases.

## Functional Changes

[] New map or map update
[x] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

- tested on local lobby

## Screens Shots
Screenshot from 2020-04-29 22-25-44

Screenshot from 2020-04-29 22-15-06

Screenshot from 2020-04-29 22-14-58

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/ui/action/player/info/ShowPlayerInformationAction.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/ui/action/player/info/ShowPlayerInformationAction.java
@@ -1,0 +1,46 @@
+package games.strategy.engine.lobby.client.ui.action.player.info;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+import java.util.logging.Level;
+import javax.annotation.Nonnull;
+import javax.swing.Action;
+import javax.swing.JFrame;
+import lombok.Builder;
+import lombok.extern.java.Log;
+import org.triplea.domain.data.PlayerChatId;
+import org.triplea.http.client.lobby.moderator.PlayerSummaryForModerator;
+import org.triplea.http.client.web.socket.client.connections.PlayerToLobbyConnection;
+import org.triplea.swing.SwingAction;
+
+/**
+ * When this action is taken (by a moderator), we will fetch from server information about the
+ * clicked on player and display it in a tabbed dialog.
+ */
+@Builder
+@Log
+public class ShowPlayerInformationAction {
+  @Nonnull private final JFrame parent;
+  @Nonnull private final PlayerToLobbyConnection playerToLobbyConnection;
+  @Nonnull private final PlayerChatId playerChatId;
+  private Function<PlayerSummaryForModerator, String> dataFormatter;
+
+  public Action toSwingAction() {
+    return SwingAction.of(
+        "Show Player Information",
+        () ->
+            CompletableFuture.runAsync(this::fetchPlayerInfoAndShowDisplay)
+                .exceptionally(this::logFetchError));
+  }
+
+  private void fetchPlayerInfoAndShowDisplay() {
+    final var playerSummaryForModerator =
+        playerToLobbyConnection.fetchPlayerInformation(playerChatId);
+    PlayerInformationPopup.showPopup(parent, playerSummaryForModerator);
+  }
+
+  private Void logFetchError(final Throwable throwable) {
+    log.log(Level.SEVERE, "Error fetching player information", throwable);
+    return null;
+  }
+}
--- a/http-server/src/main/java/org/triplea/db/dao/moderator/player/info/PlayerInfoForModeratorDao.java
+++ b/http-server/src/main/java/org/triplea/db/dao/moderator/player/info/PlayerInfoForModeratorDao.java
@@ -1,0 +1,48 @@
+package org.triplea.db.dao.moderator.player.info;
+
+import java.util.List;
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+
+/**
+ * DAO used to lookup player correlation information for moderators. Answers questions such as
+ *
+ * <ul>
+ *   <li>"how many times and was this player banned?"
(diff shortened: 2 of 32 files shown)


test-common/build.gradle
```diff
@@ -3,4 +3,5 @@ description = 'Test utility library, generic test utilities useful for TripleA p
 dependencies {
     implementation "org.hamcrest:java-hamcrest:$hamcrestVersion"
     implementation "org.junit.jupiter:junit-jupiter-api:$junitJupiterVersion"
+    implementation "org.apache.commons:commons-lang3:$apacheCommonsLangVersion"
```
RoiEXLab, line 6: I don't think this dependency is used in this PR

java-extras/src/main/java/org/triplea/java/DateTimeFormatterUtil.java
```diff
@@ -0,0 +1,29 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.util.Locale;
+import lombok.experimental.UtilityClass;
+
+/** Utility class for date-time formatting operations. */
+@UtilityClass
+public class DateTimeFormatterUtil {
```
RoiEXLab, line 12: We already have TimeManager.java which has similar functionality.
Does it make sense to merge those 2 classes?

http-server/src/main/java/org/triplea/db/dao/moderator/player/info/PlayerInfoForModeratorDao.java
```diff
@@ -0,0 +1,48 @@
...
+          + "    max(access_time) as accessTime"
+          + "  from access_log"
+          + "  where "
+          + "    access_time > (now() - '14 day'::interval)"
+          + "    and ("
+          + "      ip = :ip::inet"
+          + "      or system_id = :systemId"
+          + "    )"
+          + "  group by name, ip, systemId"
+          + "  order by accessTime desc")
```
RoiEXLab, line 30: Side-Note: One day we'll have [Text Blocks](https://openjdk.java.net/jeps/378) and the world will be a better place ^^

game-core/src/main/java/games/strategy/engine/lobby/client/ui/action/player/info/ShowPlayerInformationAction.java
```diff
@@ -0,0 +1,46 @@
...
+  @Nonnull private final PlayerToLobbyConnection playerToLobbyConnection;
+  @Nonnull private final PlayerChatId playerChatId;
+  private Function<PlayerSummaryForModerator, String> dataFormatter;
+
+  public Action toSwingAction() {
+    return SwingAction.of(
+        "Show Player Information",
+        () ->
+            CompletableFuture.runAsync(this::fetchPlayerInfoAndShowDisplay)
+                .exceptionally(this::logFetchError));
```
RoiEXLab, line 33: Side-Note: This is potentially something we could establish instead of creating "real threads" left and right, we could try to use CompletableFutures instead.
It'll probably decrease the footprint of TripleA a little bit

prastle:
Woot! Woot! :)

prastle:
Off topic Sorry but since it affects the  toolbox I will mention it here.  Is it possible to add a search function that will  look for id#s and ip #s?  if we just type in an interesting one we notice in lobby?  Or is that all corelated with the alias tab?

test-common/build.gradle
DanVanAtta: Thanks - removed in: 7bb93e7ba

java-extras/src/main/java/org/triplea/java/DateTimeFormatterUtil.java
DanVanAtta: Makes sense to combine them, yes.
`TimeManager` has a naming issue IMO. `Manager` is a bit of a smell in general, kingdom of nouns type deal and is generally not very meaningful. In this context it's not really managing state of any type. A final strike to `TimeManager` is I would not have thought to look for formatting tools in that class.

`DateTimeUtils` perhaps could be a better unified class name. Thoughts?

http-server/src/main/java/org/triplea/db/dao/moderator/player/info/PlayerInfoForModeratorDao.java
DanVanAtta: :+1: 
Though TBH, I'm not sure when. We need the next LTS to come out and for us to be willing to migrate to it. I do think avoiding the constants between queries and the JDBI types was an improvement; that is something I think we should do going forward and see about retrofitting.

game-core/src/main/java/games/strategy/engine/lobby/client/ui/action/player/info/ShowPlayerInformationAction.java
DanVanAtta: Agree, it is a nice API as well.

java-extras/src/main/java/org/triplea/java/DateTimeFormatterUtil.java
RoiEXLab: Go with `DateTimeUtils` 🤷

PR #6356, opened 2020-04-30
Simplify logging logic

RoiEXLab:
Follow-up to #6351

Note that there are several ways to "fix" this custom logic, but I think simply throwing the first exception is a valid solution

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

None

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
+++ b/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
@@ -10,11 +10,7 @@ import games.strategy.engine.data.Unit;
 import games.strategy.engine.framework.GameDataUtils;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
```diff
@@ -262,52 +255,20 @@ public AggregateResults calculate(
...
-            e = entry.iterator().next();
-            log.log(
-                Level.SEVERE,
-                entry.size() + " Battle results workers aborted by exception",
-                e.getCause());
-          }
-        }
-        if (e != null) {
-          throw new IllegalStateException(e.getCause());
+          throw new IllegalStateException("Battle results workers aborted by exception", e);
```
DanVanAtta, line 271: nit, we probably should avoid the word "exception" from being displayed to users. It makes enough sense, but it's not the typical parlance of a player, they're not developers. It is a odd that the audience for most exceptions/error message is not developers but players.

PR #6369, opened 2020-05-02
Consider left over run count for BattleCalc

RoiEXLab:
This change makes the "run count distribution" consider the fact that some integers are not cleanly dividable by the ampunmt of threads on your machine this is what happens on my machine using current master:
Screenshot
As you can see even I told it to run 1337 times, it only ran 1332 times because `floor(1337/12) * 12 = 1332`

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

I verified I could run the battle calc as usual

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
+++ b/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
@@ -16,7 +16,6 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 import org.triplea.java.Interruptibles;
@@ -229,7 +228,7 @@ public class ConcurrentBattleCalculator implements IBattleCalculator {
     synchronized (mutexCalcIsRunning) {
       awaitLatch();
       final long start = System.currentTimeMillis();
-      final AtomicBoolean overflowRunsAvailable = new AtomicBoolean(true);
+      final AtomicInteger overflowRunsAvailable = new AtomicInteger(runCount % workers.size());
       // Create worker thread pool and start all workers
       final int runsPerWorker = runCount / workers.size();
       final List<Future<AggregateResults>> list =
@@ -249,9 +248,7 @@ public class ConcurrentBattleCalculator implements IBattleCalculator {
                                   retreatWhenOnlyAirLeft,
                                   // Ensure that we always achieve the target run count even if
                                   // the number is not dividable by workers.size()
-                                  (overflowRunsAvailable.getAndSet(false)
-                                          ? runCount % workers.size()
-                                          : 0)
+                                  (overflowRunsAvailable.getAndDecrement() > 0 ? 1 : 0)
                                       + runsPerWorker)))
               .collect(Collectors.toList());
       if (!getIsReady()) {
--- a/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
+++ b/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
@@ -8,7 +8,6 @@ import games.strategy.engine.data.Territory;
 import games.strategy.engine.data.TerritoryEffect;
 import games.strategy.engine.data.Unit;
 import games.strategy.engine.framework.GameDataUtils;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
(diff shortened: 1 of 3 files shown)


game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
```diff
@@ -223,44 +224,43 @@ public AggregateResults calculate(
...
+                                  // the number is not dividable by workers.size()
+                                  (overflowRunsAvailable.getAndSet(false)
+                                          ? runCount % workers.size()
+                                          : 0)
+                                      + runsPerWorker)))
+              .collect(Collectors.toList());
+      if (!getIsReady()) {
+        // we could have attempted to set a new game data, while the old one was still being set,
+        // causing it to abort with null data
+        return new AggregateResults(0);
```
RoiEXLab, line 260: tbh I'm not quite sure if we actually need this.
If something is wrong with the data we should get an exception at some point, but IMO this doesn't add anything to potential error handling

game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
```diff
@@ -223,44 +223,41 @@ public AggregateResults calculate(
...
+                                  defender,
+                                  location,
+                                  attacking,
+                                  defending,
+                                  bombarding,
+                                  territoryEffects,
+                                  retreatWhenOnlyAirLeft,
+                                  // Ensure that we always achieve the target run count even if
+                                  // the number is not dividable by workers.size()
+                                  (overflowRunsAvailable.getAndDecrement() > 0 ? 1 : 0)
```
DanVanAtta, line 251: Question: I wonder if there is a way to make this a bit faster/easier to understand?

Suggestion: Have you considered extracting this bit of logic to a method perhaps? In that case the bit of logic could be labelled and perhaps be commented.

game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
```diff
@@ -223,44 +223,41 @@ public AggregateResults calculate(
...
+                                      + runsPerWorker)))
+              .collect(Collectors.toList());
```
DanVanAtta, line 253: suggestion: What are your thoughts to add a postcondition here to check that sum of aggregated results is equal to the run count?

game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
RoiEXLab: I have thought about extracting the logic into a method, but you'd still have to pass an `AtomicInteger` instance to the method because it has to be shared for the same call.

However I think extracting this "distribution" to an object and adding unit tests for that could be an elegant abstraction.
I won't have the time to do this today, this will have to wait a couple of days.

game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
RoiEXLab: Not quite sure if that would make sense.
I mean in the end it doesn't really matter if we get exactly the number of runs we required or just something that's very similar. The only "real" reason that would cause a significant difference in runs would be if a worker got an exception, which should be caught by our error handling later anyways.

So in the end if I make the "distribution logic" its own stateful object that distributes run counts evenly we could just add unit tests for that and that should make any potential post-condition redundant

game-core/src/main/java/games/strategy/triplea/odds/calculator/RunCountDistributor.java
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Returns the next run-count.
+   *
+   * @throws IllegalStateException If this method was called more often than provided level of
+   *     parallelism.
+   */
+  int nextRunCount() {
+    final int leftoverRuns = this.leftoverRuns.getAndDecrement();
+    if (leftoverRuns <= targetLeftover) {
+      throw new IllegalStateException(
```
RoiEXLab, line 40: I'm actually not sure if we need this strict behaviour.
If we drop this exception we could also drop the field `targetLeftover`

RoiEXLab:
@DanVanAtta Updated, I moved the distribution logic into its own class

game-core/src/test/java/games/strategy/triplea/odds/calculator/RunCountDistributorTest.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.stream.IntStream;
+import org.junit.jupiter.api.Test;
+
+public class RunCountDistributorTest {
+
+  @Test
+  void verifyEvenDistributionWhenCleanlyDividable() {
+    final var runCountDistributor = new RunCountDistributor(20, 4);
+
+    assertThat(runCountDistributor.nextRunCount(), is(equalTo(5)));
```
DanVanAtta, line 17: Minor suggestion: Might be good to have a display name to say we are spawning 4 threads and each should have 5. It's not immediately clear why there is repetition.

DanVanAtta:
:+1:

Issue #6316, opened 2020-04-23
Commit Comment Standards


DanVanAtta:
I feel like we could use a standard for commit comments. Do we agree?

Commit comments will ideally make our history useful. If you look at `git log`, the last 'n' changes should hopefully all make sense and not be noise. It'd also be great for the comment to tell you enough where you don't need to see which files have changed.

In this case, the following commit comments examples would be insufficient:

- "fix bug"
- "fix transport bug"
- "formatting"
- etc...

A standard should indicate what we expect to be in a commit comment. It is a philosophical opinion on how useful commit history is in general. Some feel it's not useful, so the comment barely matters, others feel that it is potentially useful and for a very long time after the fact.

My 2 cents is I do not want any important context in a PR that is not in a commit comment. Having to look at PR discussions to understand code I think is quite bad; that's a minimum I'd certainly want to avoid. 

How do others feel?

RoiEXLab:
I agree that some comit messages are less then ideal, but often times I don't really feel like an auto-formatting commit needs extra description

DanVanAtta:
Here is an example:

```
  Simplify code (#6299)
   
    * Simplify code
    
    * Auto-Formatting
```
To understand what was simplified you need to look at the PR or the diff. Having to go to some other system to understand this is a rule of thumb for something to avoid. If you have 30 commits, and each one you need to go look up in another tool, you are not going to be efficient. It's evidence that we're failing to capture context in the git history where it arguably belongs and is the reason for it. Said another way, instead of using git history what it is for, we're using github PRs for that purpose.

Specifically, the messages not in commit but in a PR are for the reviewer and only the reviewer, at the time of review. The messages in the commit is the stuff that lives on with the code and is for all future developers trying to understand the code. The rule of thumb basically comes back in, if you can't understand the code through the commit history, IMO the tool of commit history is not being used well and instead we're using our review tool in its place.

With that said, we do have a large number of quite good quality commit messages.

It would be great IMO to get more consistent about it. (ESLint)[https://github.com/eslint/eslint] is a fantastic example of a well done open-source project. For example they have a very spelled out [commit standard](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-eslint)

RoiEXLab:
I have thought about this in the meantime and I realised a similar thing.
I think most of the time I just want to make a commit without thinking about what this commit actually changes, but I never come back to actually clarify what actually changed.

[...]

PR #6373, opened 2020-05-03
createsUnitsList can be zero

ron-murhammer:
Address https://forums.triplea-game.org/topic/1945/trigger-createsunitslist-count-from-1-to-0

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
@@ -1920,8 +1920,8 @@ public class UnitAttachment extends DefaultAttachment {
           "createsUnitsList: No unit called:" + unitTypeToProduce + thisErrorMsg());
     }
     final int n = getInt(s[0]);
-    if (n < 1) {
-      throw new GameParseException("createsUnitsList must have positive values" + thisErrorMsg());
+    if (n < 0) {
+      throw new GameParseException("createsUnitsList cannot have negative values" + thisErrorMsg());
     }
     createsUnitsList.put(ut, n);
   }


ron-murhammer:
Merging as very minor change that can be tested by the map maker.

FrostionAAA:
This change seems to work. I don’t know if it is intentionally, but I would have imagined that the generated tooltip info would not state the fact that a unit does not produce another unit anymore. I would have expected the text “1xBattle-Chicken” to just disappear, not to become “0xBattle-Chicken”

example

I guess it is a consequence of the unit actually still having the unit attachment <option name="createsUnitsList" value="0: Battle-Chicken"/>. But if I am not mistaken, not all unit options with value 0 or false are shown? Right? So it could also be possible to hide this one?

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
```diff
@@ -1920,8 +1920,8 @@ private void setCreatesUnitsList(final String value) throws GameParseException {
           "createsUnitsList: No unit called:" + unitTypeToProduce + thisErrorMsg());
     }
     final int n = getInt(s[0]);
-    if (n < 1) {
-      throw new GameParseException("createsUnitsList must have positive values" + thisErrorMsg());
+    if (n < 0) {
+      throw new GameParseException("createsUnitsList cannot have negative values" + thisErrorMsg());
```
DanVanAtta, line 1924: nit, missing the trailing space before 'thisErrorMsg()'

suggestion: we might have a better error message if we print out the `s` array to give a person more context to where the error is, particularly if there are many create units lists.

DanVanAtta:
@ron-murhammer the comment on which feature is being addressed is useful, but IMO it would be more useful in the commit history. If there is no review, then any PR commentary is of zero value - the PR overview and comments are for the reviewer.

DanVanAtta:
An even better commit comment would have stated that this change was to support examples such as:
```
<option name="unitProperty" value="createsUnitsList" count="0:Battle-Chicken"/>
```

IMO we should decide if we want our commit history to be a useful tool or not to bother trying -  that discussion is in: https://github.com/triplea-game/triplea/issues/6316

PR #6382, opened 2020-05-04
Github issue client cleanup & Add "Error Report" label to uploaded issues

DanVanAtta:
1. Feature: Update automatic issue creation to add a label "Error Report"
   to issues created via the error report functionality. The labels will
   server to more easily identify and organize errors that were uploaded
   from the game client.

   Implementation changes:
   To attach labels we split the create error report object, prevoiusly
   the error report object sent to the HTTP server matched the one we
   sent to github. After splitting the error report request object,
   the 'createIssueRequest' object we send to github to create issues
   can have labels added to it.

2. Cleanup: Update github issue client to be consistent with a well
   named feign client class.

3. Cleanup: Make github issue client upload 'isTest' behavior be internal
   to the client. Before it was the user of the client class that checked
   if the client was in test mode and stubbed a response, now the client
   will transparently return a stubbed response when in test mode.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
- updated configuration.yml to have a personal access token
- set the game client to use a local lobby server
- launched a local lobby server
- updated game runner to throw an exception after the UI is rendered
- verified error report pop-up appeared and uploaded an error report.
- verified error report issue contains expected label: https://github.com/triplea-game/test/issues/58


http-server/src/main/java/org/triplea/modules/error/reporting/ErrorReportController.java
```diff
@@ -27,15 +27,18 @@
...
             .uri(AppConfig.GITHUB_WEB_SERVICE_API_URL)
             .authToken(configuration.getGithubApiToken())
             .githubOrg(AppConfig.GITHUB_ORG)
             .githubRepo(configuration.getGithubRepo())
+            .isTest(isTest)
             .build();
 
-    if (githubIssueClient.isTest()) {
+    if (isTest) {
       Preconditions.checkState(!configuration.isProd());
```
RoiEXLab, line 42: What about checking `!isTest || !configuration.isProd()` or `!(isTest && configuration.isProd())` instead of the if block?

http-clients/src/main/java/org/triplea/http/client/github/issues/CreateIssueRequest.java
```diff
@@ -0,0 +1,28 @@
...
+  private String title;
+  private String body;
+  private String[] labels;
+
+  public String getTitle() {
+    return title == null ? null : Ascii.truncate(title, 125, "...");
+  }
+
+  public String getBody() {
+    return body == null ? null : Ascii.truncate(body, 20000, "...");
```
RoiEXLab, line 26: Is 20000 a github limit?

DanVanAtta: no, it's somewhat arbitrary. This request object is lifted from the error upload request object. Before we had the same request object between client, server, and github. So this limit is just carried forward from that previous object. The fact that they match now I'm not sure is actually important. It still might be better to document that this is an arbitrarily large number.

http-server/src/main/java/org/triplea/modules/error/reporting/ErrorReportController.java
DanVanAtta: I was/am tempted, though I'm not sure it's actually simpler to reason about.

RoiEXLab: Same, it's kind of a weird thing to check though

Issue #6338, opened 2020-04-26
NPE for image when starting local game


RoiEXLab:
I originally just wanted to report this exception using the built in report feature, but this one froze on me for whatever reason.

```
java.lang.NullPointerException
	at java.base/java.util.Objects.requireNonNull(Objects.java:221)
	at java.base/java.util.Optional.<init>(Optional.java:107)
	at java.base/java.util.Optional.of(Optional.java:120)
	at games.strategy.triplea.image.UnitImageFactory.getImage(UnitImageFactory.java:124)
	at games.strategy.triplea.ui.screen.UnitsDrawer.draw(UnitsDrawer.java:116)
	at games.strategy.triplea.ui.screen.Tile.draw(Tile.java:70)
	at games.strategy.triplea.ui.screen.Tile.drawImage(Tile.java:55)
	at games.strategy.triplea.ui.panels.map.MapPanel.lambda$paint$7(MapPanel.java:749)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

Seems like a concurrency problem to me.

DanVanAtta:
Marking as a problem, assuming for now this is an existing problem unless we can show this is new as of 2.0

DanVanAtta:
https://github.com/triplea-game/triplea/pull/6385 has a likely fix.

Identified a TOCTOU if we clear images cache after checking if an image exists and then trying to load it. The only other alternative is we are storing a null image into cache which does not look to be the case.

PR #6385, opened 2020-05-04
Fix unit factory toctou

DanVanAtta:
Should address: https://github.com/triplea-game/triplea/issues/6338

## Commits

commit 9122685518435cf9ef66c98bba588a69c4a1cf3b

    Simplify, avoid assigning to Optional type, instead use 'map' operation

commit fff83890bf707c3cfa4c638fc5d46a889d5b7e1b

    Fix TOCTOU when getting images from an image cache.
    
    There exists a 'clear cache' method that can possibly remove
    the data in 'images'. If a method call interleaving is just right
    then we have a TOCTOU problem where we can detect an image exists,
    clear it, then try to retrieve it resulting in a NPE.
    
    This is fixed by getting the image and checking if that is null,
    if not we use the image we retrieved.

commit 5da21d60eceea0029563baad07adeca6cd5880e1

    Remove the need to clear image cache in UnitImageFactory by using a new instance
    
    - Change 'setResourceLoader' to instead to be a constructor. If we use UnitImageFactory
    before we have called 'setResourceLoader', we would get a NPE. This implies that it
    is safe to convert 'setResourceLoader' to be a constructor.
    
    - When changing scale factor, instead return a new UnitImageFactory that has a fresh
    cache.
    
    With the two changes above we no longer need to clear cache.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing
- did a quick smoke test loading a game and changed unit scale.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
+++ b/game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
@@ -120,26 +120,27 @@ public class UnitImageFactory {
       final UnitType type, final GamePlayer player, final boolean damaged, final boolean disabled) {
     final String baseName = getBaseImageName(type, player, damaged, disabled);
     final String fullName = baseName + player.getName();
+
     if (images.containsKey(fullName)) {
       return Optional.of(images.get(fullName));
     }
-    final Optional<Image> image = getTransformedImage(baseName, player, type);
-    if (image.isEmpty()) {
-      return Optional.empty();
-    }
-    final Image baseImage = image.get();
-
-    // We want to scale units according to the given scale factor.
-    // We use smooth scaling since the images are cached to allow to take our time in doing the
-    // scaling.
-    // Image observer is null, since the image should have been guaranteed to be loaded.
-    final int width = (int) (baseImage.getWidth(null) * scaleFactor);
-    final int height = (int) (baseImage.getHeight(null) * scaleFactor);
-    final Image scaledImage = baseImage.getScaledInstance(width, height, Image.SCALE_SMOOTH);
-    // Ensure the scaling is completed.
-    Util.ensureImageLoaded(scaledImage);
-    images.put(fullName, scaledImage);
-    return Optional.of(scaledImage);
+    getTransformedImage(baseName, player, type)
+        .map(
+            baseImage -> {
+              // We want to scale units according to the given scale factor.
+              // We use smooth scaling since the images are cached to allow to take our time in
+              // doing the
+              // scaling.
+              // Image observer is null, since the image should have been guaranteed to be loaded.
+              final int width = (int) (baseImage.getWidth(null) * scaleFactor);
+              final int height = (int) (baseImage.getHeight(null) * scaleFactor);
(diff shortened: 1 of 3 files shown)


game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
```diff
@@ -120,26 +106,29 @@ public Image getImage(final UnitCategory unit) {
...
-    // scaling.
-    // Image observer is null, since the image should have been guaranteed to be loaded.
-    final int width = (int) (baseImage.getWidth(null) * scaleFactor);
-    final int height = (int) (baseImage.getHeight(null) * scaleFactor);
-    final Image scaledImage = baseImage.getScaledInstance(width, height, Image.SCALE_SMOOTH);
-    // Ensure the scaling is completed.
-    Util.ensureImageLoaded(scaledImage);
-    images.put(fullName, scaledImage);
-    return Optional.of(scaledImage);
+    return Optional.ofNullable(images.get(fullName))
```
RoiEXLab, line 110: Actually I'd go for `Optional.ofNullable(images.computeIfAbsent(fullName, k -> {/* The lambda below*/}))`.

However this change would no longer attempt to recompute an image if `getTransformedImage(baseName, player, type)` returned an empty optional, so maybe we need something like this instead:
```java
Optional.ofNullable(images.compute(fullName, (key, image) -> Optional.ofNullable(image).or(
/* lambda below*/
).orElse(null)));
```

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
```diff
@@ -55,30 +55,22 @@
...
   }
 
   /** Set the unitScaling factor. */
-  public void setScaleFactor(final double scaleFactor) {
-    if (this.scaleFactor != scaleFactor) {
-      this.scaleFactor = scaleFactor;
-      clearImageCache();
-    }
+  public UnitImageFactory withScaleFactor(final double scaleFactor) {
+    return this.scaleFactor == scaleFactor
```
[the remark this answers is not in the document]
DanVanAtta: Turns out yes, but we need to remove the excess constructors first. Also of interest, the class was using static state and re-setting it in constructor, that state can be made local and final as well once the constructors were removed.

e6b26852b294108a5c5b839b54c70896d7c8e09b

game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
DanVanAtta: Extracting the computation function to another method helps (arguably the thing to do to avoid too much inline code), and `computeIfAbsent` does avoid the optional `or`.

Updated: 1fdf91d4b

DanVanAtta: Interesting landed quickly into concurrent modification errors, reverted the update: cc3e02409

PR #6390, opened 2020-05-04
Update ShowDialogWithLinks to create message dialog on EDT

DanVanAtta:
Avoids component created off of EDT exception.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:

### DIFF
--- a/swing-lib/src/main/java/org/triplea/swing/SwingComponents.java
+++ b/swing-lib/src/main/java/org/triplea/swing/SwingComponents.java
@@ -406,11 +406,16 @@ public final class SwingComponents {
 
     // parentComponent == null to avoid pop-up from appearing behind other windows
     final Component parentComponent = null;
-    JOptionPane.showMessageDialog(
-        parentComponent,
-        messageToShow,
-        params.title,
-        Optional.ofNullable(params.dialogType).orElse(DialogWithLinksTypes.INFO).optionPaneFlag);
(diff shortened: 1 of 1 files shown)


swing-lib/src/main/java/org/triplea/swing/SwingComponents.java
```diff
@@ -406,11 +406,16 @@ public static void showDialogWithLinks(final DialogWithLinksParams params) {
 
     // parentComponent == null to avoid pop-up from appearing behind other windows
     final Component parentComponent = null;
```
RoiEXLab, line 408: If you ask me we could inline this null value and keep the comment

DanVanAtta: I somewhat wonder if the comment is actually correct and if previously the wrong parent component was simply being used. It could be that using a null parent component is always 'safer' even if the pop-up is not centered.

The assigning to a variable for a null value for the sake of adding a label is a bit old style compared to using a builder or something similar. In IDEA with the parameters being labelled it's far less useful, though still debatable as some would argue that the code should be understandable even if you are not using an IDE (which means you can infer what the variables are by looking at the usage and not have to look at the method definition).

Issue #6366, opened 2020-05-02
NullPointerException - HeadedGameRunner.lambda$initializeClientSettingAndLogging$0


tripleabuilderbot:
## TripleA Version
2.0.18898

## Java Version
11.0.6

## Operating System
Windows 10

## Exception 
java.lang.NullPointerException

## Stack Trace
games.strategy.triplea.ui.TripleAFrame.lambda$showGame$58(TripleAFrame.java:2266)
java.desktop/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:313)
java.desktop/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:770)
java.desktop/java.awt.EventQueue$4.run(EventQueue.java:721)
java.desktop/java.awt.EventQueue$4.run(EventQueue.java:715)
java.base/java.security.AccessController.doPrivileged(Native Method)
java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:85)
java.desktop/java.awt.EventQueue.dispatchEvent(EventQueue.java:740)
java.desktop/java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:203)
java.desktop/java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:124)
java.desktop/java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:113)
java.desktop/java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:109)
java.desktop/java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:101)
java.desktop/java.awt.EventDispatchThread.run(EventDispatchThread.java:90)

PR #6392, opened 2020-05-05
Remove unnecessary call to historyPanel.gotoEnd

DanVanAtta:
- The call to 'gotoEnd' is unnecessary as in the very next line we null
out the history panel. The 'gotoEnd' does not change how we show the current
game, when re-rendering the history panel we will be at the last node without
this call.

- The 'gotoEnd' can result in a NPE, reported in:
https://github.com/triplea-game/triplea/issues/6366

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
@@ -2262,7 +2262,6 @@ public final class TripleAFrame extends JFrame implements QuitHandler {
               historySyncher.deactivate();
               historySyncher = null;
             }
-            historyPanel.goToEnd();
             historyPanel = null;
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
```diff
@@ -2262,7 +2262,6 @@ private void showGame() {
               historySyncher.deactivate();
               historySyncher = null;
             }
-            historyPanel.goToEnd();
```
RoiEXLab, line 2265: I'm concerned this call might have some required unintended side-effect on the underlying GameData which might break things, but ok 🤷

Issue #6279, opened 2020-04-17
In WWII Revised, deselecting units with right click, or ESC key, does not work


swurzer121:
**How can the problem be recreated?**
In Revised, right click on any unit, then right click, or hit ESC key
This used to deselect the unit.  It no longer does so.

**When the problem is fixed, what should have happened instead?**

**Which Engine Version are you using?** 

**Screenshots** 

**Zipped Save Game**

<!--
Please zip and post a save game to this bug report!

[2020-4-17-World-War-II-Revised_deselect_unit_bug.zip](https://github.com/triplea-game/triplea/files/4494958/2020-4-17-World-War-II-Revised_deselect_unit_bug.zip)

DanVanAtta:
This update should help the right click case: https://github.com/triplea-game/triplea/pull/6393

It might not be that right click is broken in 2.0, 1.9.0.0.13066 also had a problem.

If you do not move the map via right click drag, right click deselect should work. But, if you right click drag the map, then select a unit, you'll need to right click twice to deselect the unit. This was a bug in 1.9.0.0.13066, #6393 should fix this so if you select a unit and then right click - the unit will be deselected.

Cernelius:
Oh, I've been annoyed by this for a rather long time but I've never been really sure what was that. Initially I thought it was just me moving the cursor by like 1 pixel unintentionally while right clicking, thus having the drag behaviour instead of the deselect one.

DanVanAtta:
Yeah, if there were ever a way for the game to gaslight you... :grin: 

https://github.com/triplea-game/triplea/pull/6394 will fix the escape key behavior in 2.0 and have it working again.

As fixes are now in flight, will close. Please re-open or comment if there is more to this issue than what either #6393 or #6934 fixes.

swurzer121:
👍

PR #6394, opened 2020-05-05
Fix escape key, should deselect units

DanVanAtta:
Escape key listener to deselect units needs to be attached to move panel
root pane, not the root move panel itself. A recent consolidation
of listeners broke this: a9f88fc

Addresses escape key listener not working, reported in: https://github.com/triplea-game/triplea/issues/6279

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing
- did a quick check that after selecting units 'escape' key will deselect them

## Additional Review Notes

Checked 1.9.0.0.13066 code and this was the only place where the root pane was added as the listener target.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/AbstractMovePanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/AbstractMovePanel.java
@@ -284,7 +284,7 @@ public abstract class AbstractMovePanel extends ActionPanel {
           }
           listening = true;
           if (getRootPane() != null) {
-            SwingKeyBinding.addKeyBinding(this, KeyCode.ESCAPE, this::cancelMove);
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/triplea/ui/AbstractMovePanel.java
```diff
@@ -284,7 +284,7 @@ protected final void setUp(final IPlayerBridge bridge) {
           }
           listening = true;
           if (getRootPane() != null) {
-            SwingKeyBinding.addKeyBinding(this, KeyCode.ESCAPE, this::cancelMove);
+            SwingKeyBinding.addKeyBinding(getRootPane(), KeyCode.ESCAPE, this::cancelMove);
```
RoiEXLab, line 287: This null check is making way more sense now 🤔

PR #6406, opened 2020-05-05
Update triplea_maps.yaml

mongoose1130:
## Functional Changes

[X] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2066,7 +2066,7 @@
 - mapName: Twelve-Clans
   mapCategory: EXPERIMENTAL
   url: https://github.com/triplea-maps/twelve-clans/archive/master.zip
-  version: 1.0
+  version: 1
   img: https://raw.githubusercontent.com/triplea-maps/twelve-clans/master/description/MapDesc.png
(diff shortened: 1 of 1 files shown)


triplea_maps.yaml
```diff
@@ -2063,3 +2063,31 @@
     <br>
     <br> This is the Charles de Gaulle 1939 Mod for the "Global 40" game.
     <br>
+- mapName: Twelve-Clans
+  mapCategory: EXPERIMENTAL
+  url: https://github.com/triplea-maps/twelve-clans/archive/master.zip
+  version: 1.0
```
ron-murhammer, line 2069: I think the version here needs to be "1" instead of "1.0". The numbers in this file don't necessarily correspond to the map's actual version and are just used by the TripleA engine to determine when players should be prompted that there is an update for the map. And yes, I realize its a bit confusing :)

mongoose1130: OK

PR #6414, opened 2020-05-06
Moderator Mute (Front-End)

DanVanAtta:
- Add front-end implementation for moderator mute ability.
  Mutes are capped to 60 minutes.
- Add no-op backend controller endpoint to do mutes
- Restructure BanDuration dialog to be more generic and
  rename to 'ActionDuration'
- Allow for the action duration time unit and max unit to
  be specified per instance
- Restructure Action duration dialog to return an optional
  duration instead of automatically running an action.
- Remove the 'permanent' time-unit (it defaulted to a very
  large number), instead moderators can do the same and
  just ban for 10,000 months or something similar.
- Add a confirmation prompt to both player mute and
  player ban actions.

## Functional Changes

[] New map or map update
[x] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
- verified ban from both right click menu option and toolbox
- verified mute selection and sends request to backend (and then no-ops)

## Screens Shots
Screenshot from 2020-05-05 21-47-45


RoiEXLab:
@DanVanAtta I'll have a look at this Pr tomorrow.
Also some merge conflicts showed up

PR #6421, opened 2020-05-06
Fix display of Can't Target info in tooltips.

asvitkine:
The list of unit types was previously incorrectly formatted.
Also makes the getter get called only once, since the getter
does work in generating the list.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix:  
[] Other:   

## Testing
Manual.

### Before
Screen Shot 2020-05-06 at 6 01 44 PM

### After
Screen Shot 2020-05-06 at 6 01 15 PM

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
@@ -3592,7 +3592,7 @@ public class UnitAttachment extends DefaultAttachment {
   }
 
   private void addLabeledUnitTypes(
-      String label,
+      final String label,
       final Collection<UnitType> unitTypes,
       final List<Tuple<String, String>> tuples) {
     if (!unitTypes.isEmpty()) {
--- a/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
@@ -3366,16 +3366,20 @@ public class UnitAttachment extends DefaultAttachment {
     if (getCanBeMovedThroughByEnemies()) {
       tuples.add(Tuple.of("Can Be Moved Through By Enemies", ""));
     }
-    if (!getCanNotTarget().isEmpty()) {
-      if (getCanNotTarget().size() <= 4) {
-        tuples.add(Tuple.of("Can't Target: ", getCanNotTarget().toString()));
+    final Collection<UnitType> cannotTarget = getCanNotTarget();
+    if (!cannotTarget.isEmpty()) {
+      if (cannotTarget.size() <= 4) {
+        tuples.add(Tuple.of("Can't Target", MyFormatter.defaultNamedToTextList(cannotTarget)));
       } else {
         tuples.add(Tuple.of("Can't Target Some Units", ""));
       }
     }
-    if (!getCanNotBeTargetedBy().isEmpty()) {
-      if (getCanNotBeTargetedBy().size() <= 4) {
-        tuples.add(Tuple.of("Can't Be Targeted By: ", getCanNotBeTargetedBy().toString()));
+    final Collection<UnitType> cannotBeTargettedBy = getCanNotBeTargetedBy();
+    if (!cannotBeTargettedBy.isEmpty()) {
+      if (cannotBeTargettedBy.size() <= 4) {
+        tuples.add(
+            Tuple.of(
+                "Can't Be Targeted By", MyFormatter.defaultNamedToTextList(cannotBeTargettedBy)));
       } else {
         tuples.add(Tuple.of("Can't Be Targeted By Some Units", ""));
       }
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
```diff
@@ -3366,16 +3366,20 @@ public String toStringShortAndOnlyImportantDifferences(final GamePlayer player)
     if (getCanBeMovedThroughByEnemies()) {
       tuples.add(Tuple.of("Can Be Moved Through By Enemies", ""));
     }
-    if (!getCanNotTarget().isEmpty()) {
-      if (getCanNotTarget().size() <= 4) {
-        tuples.add(Tuple.of("Can't Target: ", getCanNotTarget().toString()));
+    final Collection<UnitType> cannotTarget = getCanNotTarget();
+    if (!cannotTarget.isEmpty()) {
```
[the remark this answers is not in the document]
asvitkine: Made a helper function.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
```diff
@@ -3366,16 +3366,20 @@ public String toStringShortAndOnlyImportantDifferences(final GamePlayer player)
...
+        tuples.add(Tuple.of("Can't Target", MyFormatter.defaultNamedToTextList(cannotTarget)));
       } else {
         tuples.add(Tuple.of("Can't Target Some Units", ""));
       }
     }
-    if (!getCanNotBeTargetedBy().isEmpty()) {
-      if (getCanNotBeTargetedBy().size() <= 4) {
-        tuples.add(Tuple.of("Can't Be Targeted By: ", getCanNotBeTargetedBy().toString()));
+    final Collection<UnitType> cannotBeTargettedBy = getCanNotBeTargetedBy();
+    if (!cannotBeTargettedBy.isEmpty()) {
```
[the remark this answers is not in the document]
asvitkine: Made a helper function.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
```diff
@@ -3607,6 +3591,19 @@ public String toStringShortAndOnlyImportantDifferences(final GamePlayer player)
     return result.toString();
   }
 
+  private void addLabeledUnitTypes(
```
DanVanAtta, line 3594: What are your thoughts to have this method return a `Tuple` rather than use side-effects?

https://github.com/triplea-game/triplea/wiki/Java-Code-Conventions#do-not-mutate-parameters-avoid-out-and-avoid-inout-parameters-5489

Cernelius:
Of course, since trench is 0/0 and nothing can support it, it doesn't matter what it can target. On top of it, it would not matter anyways, since a trench can never be in battle against another trench, because they cannot move, nor hit a gas, since the gas would suicide before the trench can roll, on defence (where the trench is not a suicide unit).

Just a side note. I guess it would be not worthwhile trying to have the automatic tooltips inferring all such things on its own.

Cernelius:
Still, I can see a noob asking himself why the hell the program is telling him that a trench cannot target another trench or a gas, when that is totally irrelevant anyway, but I guess not a big deal.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
asvitkine: It would duplicate more code because we'd need to move the if statement out of the function - either on the return value, by making the function return a null or optional tuple, or on the input param - which would then require an extra local if we don't want to recompute it twice.

So I prefer keeping it as is. Plus there's other functions already following this pattern.

asvitkine:
> Of course, since trench is 0/0 and nothing can support it, it doesn't matter what it can target. On top of it, it would not matter anyways, since a trench can never be in battle against another trench, because they cannot move, nor hit a gas, since the gas would suicide before the trench can roll, on defence (where the trench is not a suicide unit).
> 
> Just a side note. I guess it would be not worthwhile trying to have the automatic tooltips inferring all such things on its own.

That's a separate thing with the map. I wonder if this property even needs to exist in this case or if trench would just be specified to not have any attack power it would be sufficient? Anyway, that's orthogonal to making the tooltip look correct in cases where this is set (which can be on other maps too with more legitimate use cases).

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
DanVanAtta: The indirection is problematic as well; side-effects and no-op methods that are 200 lines below where you are looking can make things also difficult to follow. Notably to understand what those two lines above do you need to head down the page and then back up.

We've the code convention established to declare essentially that existing examples are anti-patterns and not something we want to build on or continue.

Be what it may, there are  a number of problems here, I'd agree there are trade-offs to be made.

PR #6423, opened 2020-05-07
DiceRoll - Support sort extracted

DanVanAtta:
Two refactors, one to extract sort logic, second to combine nearly duplicate logic. Perhaps easier to review commit-by-commit.

## Commits

commit 97b634ab6e828246705c1e0523c574c882d05881

    Extract support attachment sort to its own class and inline usages
    
    1. Create a new class 'SupportRuleSort' implementing comparator to be the new
       home for unit attachment support sorting logic (in DiceRoll.java)
    2. Move sorting logic from DiceRoll.java 'SupportRuleSort'
    3. Inline calls to 'SupportRuleSort'

commit 5a6dee7b2af954bfd2128d84084a2c748d94284a

    Combine duplicated getSortedSupport logic
    
    By accepting a list of rules to 'getSortedSupport', we can nearly
    combine 'getSortedAASupport' with 'getSortedSupport'.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
@@ -19,6 +19,7 @@ import games.strategy.triplea.attachments.UnitSupportAttachment;
 import games.strategy.triplea.delegate.Die.DieType;
 import games.strategy.triplea.delegate.battle.AirBattle;
 import games.strategy.triplea.delegate.battle.IBattle;
+import games.strategy.triplea.delegate.power.calculator.SupportRuleSort;
 import games.strategy.triplea.formatter.MyFormatter;
 import java.io.Externalizable;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
```diff
@@ -1182,149 +1190,6 @@ public static void sortByStrength(final List<Unit> units, final boolean defendin
...
-    sortSupportRules(
-        support,
-        defense,
-        friendly,
-        UnitSupportAttachment::getRoll,
-        UnitSupportAttachment::getStrength);
-  }
-
-  private static void sortSupportRules(
-      final Set<List<UnitSupportAttachment>> support,
```
DanVanAtta, line 1210: Of note' support' here is not extracted to the new calls as the sort was applied to this. Instead we'll have the callers do the sort and extract the rest of this function and variables to the new class (see first commit).

PR #6433, opened 2020-05-08
Update single-player staging screen alliances button - click to play alliance

DanVanAtta:
Alliances buttons can now be clicked to play that alilance

Changes the behavior of the alliances button on the single player
game staging screen to set the human player to play all factions
of a given alliance.

Previously the button would replicate the current select player
in the row to each faction of the alliance, this behavior was
not very intuitive nor terribly useful.

Secondarily, the tooltip text of 'click to play alliance' button
was set on game load, in which case it always read
"click to [alliance] to human" which could be incorrect if the
current row had been changed. Now the text is more generic
 and always accurate "click to play [alliance]"

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
- manual

### Before
Mouse hover over 'allies' button:
Screenshot from 2020-05-07 23-57-01

### After
Mouse hover over 'allies' button:
Screenshot from 2020-05-07 23-58-31

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerSelectorRow.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerSelectorRow.java
@@ -89,7 +89,8 @@ public class PlayerSelectorRow implements PlayerCountrySelection {
     if (!playerAlliances.contains(playerName)) {
       final String alliancesLabelText = playerAlliances.toString();
       alliances = new JButton(alliancesLabelText);
-      alliances.setToolTipText("Click to play " + alliancesLabelText);
+      alliances.setToolTipText(
+          "Set all " + alliancesLabelText + " to " + PlayerType.HUMAN_PLAYER.getLabel());
       alliances.addActionListener(
           e ->
               playerRows.stream()
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerSelectorRow.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerSelectorRow.java
@@ -89,17 +89,15 @@ public class PlayerSelectorRow implements PlayerCountrySelection {
     if (!playerAlliances.contains(playerName)) {
       final String alliancesLabelText = playerAlliances.toString();
       alliances = new JButton(alliancesLabelText);
-      alliances.setToolTipText(
-          "Set all " + alliancesLabelText + " to " + playerTypes.getSelectedItem().toString());
+      alliances.setToolTipText("Click to play " + alliancesLabelText);
       alliances.addActionListener(
-          e -> {
-            final String currentType = playerTypes.getSelectedItem().toString();
-            playerRows.stream()
-                .filter(
-                    row ->
-                        row.alliances != null && row.alliances.getText().equals(alliancesLabelText))
-                .forEach(row -> row.setPlayerType(currentType));
-          });
+          e ->
+              playerRows.stream()
+                  .filter(
+                      row ->
+                          row.alliances != null
+                              && row.alliances.getText().equals(alliancesLabelText))
+                  .forEach(row -> row.setPlayerType(PlayerType.HUMAN_PLAYER.getLabel())));
     }
 
     incomePercentage =
--- a/game-headed-javafx/src/main/java/org/triplea/game/client/ui/javafx/screens/RoleSelection.java
+++ b/game-headed-javafx/src/main/java/org/triplea/game/client/ui/javafx/screens/RoleSelection.java
@@ -142,14 +142,13 @@ public class RoleSelection implements ControlledScreen<ScreenController<FxmlMana
     final var faction = new Button(playerAlliances.toString());
     faction.setOnAction(
         e -> {
-          final int targetIndex = controllingPlayer.getSelectionModel().getSelectedIndex();
+          controllingPlayer.getSelectionModel().select(PlayerType.HUMAN_PLAYER.getLabel());
           playerAlliances.stream()
               .map(gameData.getAllianceTracker()::getPlayersInAlliance)
               .flatMap(Collection::stream)
               .map(roleForPlayers::get)
-              .filter(stringComboBox -> stringComboBox.getItems().size() > targetIndex)
               .map(ComboBox::getSelectionModel)
-              .forEach(selectionModel -> selectionModel.select(targetIndex));
+              .forEach(selectionModel -> selectionModel.select(PlayerType.HUMAN_PLAYER.getLabel()));
         });
     return faction;
   }


RoiEXLab:
Looks like the build had a temporary problem with rate limits:
```
org.triplea.modules.game.listing.GameListingControllerTest > bootGame() FAILED
    org.triplea.http.client.HttpInteractionException: Status 429 reading GameListingFeignClient#bootGame(Map,String)
    Reason: Too Many Requests
        at org.triplea.http.client.HttpClient.lambda$errorDecoder$1(HttpClient.java:94)
        at java.base/java.util.Optional.orElseGet(Optional.java:369)
        at org.triplea.http.client.HttpClient.errorDecoder(HttpClient.java:94)
        at feign.SynchronousMethodHandler.executeAndDecode(SynchronousMethodHandler.java:151)
        at feign.SynchronousMethodHandler.invoke(SynchronousMethodHandler.java:80)
        at feign.ReflectiveFeign$FeignInvocationHandler.invoke(ReflectiveFeign.java:103)
        at org.triplea.http.client.lobby.game.lobby.watcher.$Proxy105.bootGame(Unknown Source)
        at org.triplea.http.client.lobby.game.lobby.watcher.GameListingClient.bootGame(GameListingClient.java:37)
        at org.triplea.modules.game.listing.GameListingControllerTest.lambda$bootGame$0(GameListingControllerTest.java:33)
        at org.triplea.modules.http.ProtectedEndpointTest.verifyEndpoint(ProtectedEndpointTest.java:56)
        at org.triplea.modules.game.listing.GameListingControllerTest.bootGame(GameListingControllerTest.java:33)
```

ron-murhammer:
@DanVanAtta Not sure about this one. Have you review this concept with folks in the lobby or forum? Being able to set all the nations in an alliance to a selected AI is pretty useful I think.

DanVanAtta:
@ron-murhammer what are your concerns?

I've heard/had and experienced the existing behavior is counter-intuitive and not terribly useful. I think a forum route would delay and multiply the effort required (explaining a feature 3 times is a bit much, particularly when we have a tendency to not understand fully what is being proposed and to get off topic so easily - this is also a change to something that did not originally go through that process (did set-all-to-AI go through a forum post first and how long did it take if so?).

>  Being able to set all the nations in an alliance to a selected AI is pretty useful I think.

You can still do that by selecting all to be a specific AI and then selecting the alliance you want to play. It basically makes for a 2 click operation. I'm quite positive nobody really knew how to use this before.

More broadly the staging screen being so different between single and multiplayer is very inconsistent and not something I realized had happened until after the fact when reviewing the set-all-to feature and the alliance buttons. Having a 'click-to-play' makes sense for multiplayer but does not make sense as a "copy current row to alliance.

For multiplayer this fixes a couple problems:
- it's slow and error prone to deselect an alliance
- it's error prone to select a single alliance, it's pretty common to have to restart a game because of incorrect selections.

RoiEXLab:
@DanVanAtta btw. The JavaFX staging screen is currently replicating this behaviour, so if you are able to convince @ron-murhammer it would be nice to change that one as well.

Honestly I'm fine with either solution. I'd agree that I didn't understand the buttons for like a year, but once you get used to them they are convenient at times

Cernelius:
One way or the other, clicking on what is a description to interact with a number of types sharing said description is not going to be very intuitive. Now that I think about it, I don't think I've ever actually used it, because it just never comes to my mind I have this option available, when I'm actually in the process of selecting stuff.

Maybe this would be rather better done with a per-alliance option between the "Set All To" and the Names. For example, you would, then, have:
```
Set All To:
Set All Allies To:
Set All Axis To:

Name
Russians:
Germans:
British:
...
```
This would also get rid of the need of having a tooltip at all. In general, UI needing additional tooltips to be understood are not as good as UI you can understand without tooltips.

Also, personally, I would move all the "Set All" under all power names and, in this case, I also think that having the "set all" would be redundant.

So, something like this:
```
Name
Russians:
Germans:
British:
Italians:
Chinese:
Japanese:
Americans:

Allies:
Axis:

All:
```
Off topic, I guess, but I really don't like the "Name" and "Type" labels either. I would have them, respectively, as "Faction Name" and "Player Type", when not hosting, and "Player Name", when hosting (and you have a number of different "human" players), or something like that.

So, it would be more something like:
```
Faction
Russians:
Germans:
British:
Italians:
Chinese:
Japanese:
Americans:

Side
Allies:
Axis:

All:
```
Finally, the selection thing should go back empty when not all the factions are sharing the same. For example, now the "Set All To" is empty at start, but, if I first select "Hard (AI)" for it and, then, set 1 player to something else, it still says "Hard (AI)", while it should rather go back saying nothing, as soon as I set 1 player to something else.

Not sure how much of what I said is on or off topic. You can decide.

DanVanAtta:
> One way or the other, clicking on what is a description to interact with a number of types sharing said description is not going to be very intuitive. Now that I think about it, I don't think I've ever actually used it, because it just never comes to my mind I have this option available, when I'm actually in the process of selecting stuff.

@Cernelius I didn't quite understand which scenario you are talking about, before this update (the current behavor), or after?

It sounds a bit like you're suggesting an additional drop down, I'm not sure how that could be cleanly done without potentially adding an excessive number of controls to the screen.

DanVanAtta:
@Cernelius I'm on board with grouping the players by alliance, but we're actively avoiding major work to these UI screens in favor of the JavaFX update. Not sure if you recall about 2 years ago there was a major re-write that did this but we decided to jump to JavaFX. The updates here are to make things intuitive (useful), and consistent between single player and multi.

Cernelius:
>> One way or the other, clicking on what is a description to interact with a number of types sharing said description is not going to be very intuitive. Now that I think about it, I don't think I've ever actually used it, because it just never comes to my mind I have this option available, when I'm actually in the process of selecting stuff.
> 
> @Cernelius I didn't quite understand which scenario you are talking about, before this update (the current behavor), or after?
> It sounds a bit like you're suggesting an additional drop down, I'm not sure how that could be cleanly done without potentially adding an excessive number of controls to the screen.

Both. I cannot recall having ever used it when actually going to play (not meaning it is not useful, just I think I always forget this feature exists).

DanVanAtta:
Gotchya. I hope after this update it'll be more useful and natural. Whenever I remember this feature exists it usually causes more anguish than benefit. It just takes far too much thought and setup to make it useful compared to some quick manual clicks.

Cernelius:
Anyway, if you would click on the Alliance name to assign that faction to yourself, then I think it would be consistent the most if you would also be able to click on the faction name to assign the faction to yourself (as an alternative to clicking on the drop-down selection right to it). For bots, this would mean you can get rid of the "Play" buttons (removing that whole column), and have the player either clicking on the faction names or on the alliance names only.

Mostly I think the inconsistency, as said, it is the fact that you click on the alliance name to assign all the faction of that alliance but you don't click on the faction name to assign that faction. I would say either both or none should work the same way.

Side note, when I say "faction" I mean the "name" column of what are the "players". Just using this terminology, instead, as I think the official one TripleA uses is too confusing. I hope everyone understands what I'm talking about.

Finally, also "Alliance" is not a god name. While most TripleA games have fixed alliances, this is not true in general, because you may have games in which you can change alliances (this is referenced in multiple points, like the archetype allied and some politics properties). This is why I've used "Side", instead of "Alliance", previously. Anyway, I guess this is off topic.

DanVanAtta:
That's an interesting idea to make the 'faction' name clickable. That *is* a different update compared to what is here though, nonetheless an interesting idea.. Incrementalism is incredibl essential when updating code. FWIW in theory rapid smaller iterations accomplishes more in the aggregate than does fewer and larger updates.

In the meantime I do want to 'fix' the alliance button, the current behavior is not my original intent behind it and even I found it to not be very useful (particularly when you can just set-all).

ron-murhammer:
@DanVanAtta Does this update impact PBF/PBEM as well? If so I'm not sure that "click to play" makes sense since you'll have "human" nations that you aren't yourself playing.

DanVanAtta:
> If so I'm not sure that "click to play" makes sense since you'll have "human" nations that you aren't yourself playing.

Can you give an example @ron-murhammer ; could you clarify how the previous behavior made more sense compared to using the 'set-all-to'?

A lot of what it comes down to is three things:
- the previous behavior was non-intuitive and thus likely not used (ever). I even had trouble using it, which is a really bad sign.
- setting an alliance to current is not that useful when we have 'set-all-to', the two features we essentially developed in-flight at about the same time and did not take the other one into account. You may recall the UI re-write proposal that had this feature.
- having a 'set-all-to-current' would not work for multiplayer. A client is limited by what the host releases and there is no quick way for a host to release a given alliance.

DanVanAtta:
>  If so I'm not sure that "click to play" makes sense since you'll have "human" nations that you aren't yourself playing.

Typically those nations remain as human, or will be whatever preset comes from the PBEM save-game. In any case the alliance selection is not relevent and in both cases would set the alliance from human, to human. So there is no difference before or after. If on the other hand there are AI players, just as the same for a single player a person is still in a better position as you can use 'set-all-to' and then click the alliance(s) you want to set back to human.

DanVanAtta:
@ron-murhammer want to call out that there are no new buttons in this update, hopefully that is clear from the overview. The user flow is also more intuitive and easier. For example, PBEM, if all entities are human, then there is no difference. If wanting to set one side to a specific AI, you can use 'set-all-to' and the select the alliance you want. That is easier and more intuitive compared to using 'set-all-to', setting one entity of an alliance to human and then clicking the alliance button. I can't think of any case where that kind of scenario is necessarily easier. If you want mixed-AIs, using 'set-all-to' and then individually selecting is still going to be easier/more efficient.

It's also notable to consider how multiplayer selection works and how the current behavior is just not useful/applicable. I'd recommend demo'ing the multiplayer PR and this one as well.

With that said, I'd like to move forward here so we can get possible feedback and further play-testing (and generally have this off my radar and move forward with other code investigations and not be concerned with merge conflicts).

ron-murhammer:
@DanVanAtta I guess I'd vote against the change unless you've had other users say it would be more useful this way. I'll leave it up to @RoiEXLab to be the tie breaker as I don't think its a big deal changing or leaving this as either way its really not that intuitive of a feature that I think very few people use.

DanVanAtta:
The multiplayer case is significant and useful. Relaunching games due to
click error is common. It's odd you argued against the alliance button and
now would rather keep it the same.

With more intuitive behavior I'd expect this button to be used in
combination with set all to and mostly expect this to be used in all 1v1
multiplayer games

Given th ed previous behavior is not intuitive, I introduced this and now
fault, having s commitee to agree with the complaints does not seem
efficient

DanVanAtta:
@RoiEXLab ball is in your court. Let me know if I can clarify anything or how this update fits in with #6437

DanVanAtta:
Keeping the tooltip the same as what it would be for multiplayer I think was probably over-zealous consistency, the singleplayer and multiplayer tooltip do not need to be the same necessarily (though the general gist of what the buttons do should be the same). The previous tooltip was buggy to not update, but beyond that it was actually okay.

I updated the tooltip back to what it was,  bd6e873b6c981f93404af737ff045b6808266a7c, which helps in a few ways as it states more directly what is happening.

RoiEXLab:
@DanVanAtta @ron-murhammer 

I must say @DanVanAtta is a damn good negotiator and he definitely has some good points. I didn't find those buttons very intuitive either when they were introduced (also they were annoying to re-implement when writing the JavaFX screen).
I'm not entirely convinced the "new" usage is ideal either, but the argument for consistency is a good one here. So I'd argue in favour for @DanVanAtta here.
My only requirement: Please update the JavaFX ui as well, otherwise we might end up having the same discussion in the future.

side-note: I noticed that I haven't touched the JavaFX code in a while. I have some free time to spend currently, so after I "fix" the 2FA for PbF problem I might be able to get back to it.

DanVanAtta:
@RoiEXLab JavaFX UI updated. My 2 cents, the javaFX UI needs quite a lot of work, there will be many discussions to be had. I'm looking forward to collaborating with you on it; once 2.0 is out we should focus on getting that done. I think there is a long list of improvements and tweaks to be had.

RoiEXLab:
👍 I agree it needs a lot of improvements. I think the current screen switching system is fine, but besides that I haven't really found a good guideline to follow.

I have looked at your change and it looks good. Feel free to merge

PR #6454, opened 2020-05-09
Moderator toolbox use long and format date-time values

DanVanAtta:
## Summary
- improve date time formatter to handle null value and be defaulted to UTC by test
- change instances where JSON payload has an instant when serving data to moderator toolbox, instead use epoch milli timestamp.
- format date displays in the moderator toolbox instead of using `Instant#toString`

## Commits

commit d553c51129d6fbfb501a8718eb1146dcd70d0945

    Use @BindList in ModeratorDao and inline column names
    
    - Use a @BindList annotation to parameterize user role lookup
    - Inline column names in JDBI mapper and remove unit test for JDBI mapper
      The unit test is redundant, when we test the DAO, we cover the mapper.

commit 1c615f3ac193abf26a237459c07295d1cffd05ff

    Date-time formatter allow format without timezone and improve test control

commit 04c4298a614d3447b4be10d971fa469c45ac48b6

    Use 'epochMilli' (long) in JSON payload instead of 'Instant'

commit c2a517651e67701a6be7429b30fa5137554fce74

    Update data-time-formatter to handle null values

commit 245a201d6fdf7111a7ececebfead91045b7f15c9

    Convert all moderator toolbox date interchange format from Instant to Long (epoch millis) & format

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:


RoiEXLab:
Also some merge conflicts showed up

DanVanAtta:
Instant and epoch milli do have equivalencies in terms of what data they represent. The difference in JSON should be on the order of something like: `date: instant { value: 'some-value' }` compared to `date: epoch-milli`

The driving principle is to keep inter-change formats and cross-system dependencies to be as 'primitive' as possible. If for example we depended on Joda-type values to be transmitted, then we would have had work to do to upgrade to Instant. Epoch milli as a long will never "go out of style", so it's more flexible and has the least dependencies. Essentially, between systems, strong types is bad. The boundary of a system should seek to convert these weak types into strong types (do validation in the process), then internally use as strong of types as possible (for example avoiding stringly types). So for example `GameId` is a wrapper around string, better to transmit that as a string but then convert and use a `GameId` within any system that has received or broadcast it. This way any changes in types do not cross system-to-system thresholds.

DanVanAtta:
I'm afraid I got long winded, at the risk of typing too much, the rule of thumb:
- use as strong of types as possible internal to a system
- use as primitive types as possible as dependencies between systems
- convert from primitive types to strong types at the boundaries of a system

Instant wraps a date value, same thing as for example with 'SystemId' wrapping a string. In case either needs to change, if we use the primitive value then we have more flexibility.

DanVanAtta:
@RoiEXLab say for example we decide to create a semantic around ban_expiry and then create a `BanExpiry` object. If we use instants for data transmission, we'll be forced to either break compatibility or create shimming code to needless convert from one type to another to keep compatibility.

Same thing with 'gameId' and 'systemId', we do not transmit the JSON objects but the primitives. This update keeps things a more consistent in that perspective and helps avoid encoding elements as JSON objects rather than as JSON primitives.

DanVanAtta:
One can also consider the other direction, let's say PlayerChatId becomes a value object, if we consistently transmit value objects then we need to update all payloads to transmit the value object which breaks compatibility. Instead if our pattern is to use value objects at the boundaries of a system, then there is no need to update. `Instant` is essentially a value object around an epoch timestamp (hence the analogies).

From another perspective, we should try to keep JSON payloads as simple as possible, which means minimizing the number of JSON objects being transmitted.

Issue #6448, opened 2020-05-09
2.0.19274: MovableUnitsFilter#addMustMoveWith:247 - NullPointerException


tripleabuilderbot:
## TripleA Version
2.0.19274

## Java Version
11.0.6

## Operating System
Mac OS X

## Exception 
java.lang.NullPointerException

## Stack Trace
```
java.lang.NullPointerException
	at java.base/java.util.AbstractCollection.addAll(AbstractCollection.java:351)
	at games.strategy.triplea.ui.panel.move.MovableUnitsFilter.addMustMoveWith(MovableUnitsFilter.java:247)
	at games.strategy.triplea.ui.panel.move.MovableUnitsFilter.validateMoveWithDependents(MovableUnitsFilter.java:222)
	at games.strategy.triplea.ui.panel.move.MovableUnitsFilter.filterUnitsThatCanMove(MovableUnitsFilter.java:120)
	at games.strategy.triplea.ui.panel.move.MovePanel.updateUnitsThatCanMoveOnRoute(MovePanel.java:1129)
	at games.strategy.triplea.ui.panel.move.MovePanel$3.mouseMoved(MovePanel.java:705)
	at games.strategy.triplea.ui.panels.map.MapPanel.notifyMouseMoved(MapPanel.java:498)
	at games.strategy.triplea.ui.panels.map.MapPanel.updateMouseHoverState(MapPanel.java:318)
	at games.strategy.triplea.ui.panels.map.MapPanel$4.mouseMoved(MapPanel.java:266)
	at java.desktop/java.awt.AWTEventMulticaster.mouseMoved(AWTEventMulticaster.java:338)
	at java.desktop/java.awt.Component.processMouseMotionEvent(Component.java:6677)
	at java.desktop/javax.swing.JComponent.processMouseMotionEvent(JComponent.java:3360)
	at java.desktop/java.awt.Component.processEvent(Component.java:6401)
	at java.desktop/java.awt.Container.processEvent(Container.java:2263)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:5008)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2321)
	at java.desktop/java.awt.Component.dispatchEvent(Component.java:4840)
	at java.desktop/java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4918)
	at java.desktop/java.awt.LightweightDispatcher.processMouseEvent(Container.java:4560)
[...]

PR #6462, opened 2020-05-12
Handle null must-move-with-unit value

DanVanAtta:
This update replaces a 'getOrDefault' call to to convert any
returned null value to be an empty list.

If we place a null value into a map explicitly, then
a 'getOrDefault(key, default)' call will return the null value
and not the default.

Addresses NPE reported in:
https://github.com/triplea-game/triplea/issues/6448

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/data/MustMoveWithDetails.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/data/MustMoveWithDetails.java
@@ -5,6 +5,7 @@ import java.io.Serializable;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 /**
  * A response to a must move query. Returns a mapping of unit -> collection of units. Units that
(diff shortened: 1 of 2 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/data/MustMoveWithDetails.java
```diff
@@ -29,6 +30,6 @@ public MustMoveWithDetails(final Map<Unit, Collection<Unit>> mapping) {
   }
 
   public Collection<Unit> getMustMoveWithForUnit(final Unit unit) {
-    return mapping.getOrDefault(unit, List.of());
```
RoiEXLab, line 32: Hmm think it would be better if we just put en empty list there instead of null in the first place, but we're just passing the map somewhere else, so yeah...

DanVanAtta:
I agree the better fix is to ensure non-null values. The data structure
here has very little encapsulation,  perhaps removing the map get
operations and adding a null check could tell us where/when that is
happening.. clearly opted for just a logic fix for now as nulls seemingly
are expected.

PR #6466, opened 2020-05-13
Mute complete

DanVanAtta:
## Overview

Finishes up backend for moderator-mute functionality. After this update mutes are functional. A message is printed in lobby when a player is muted, avoids the need for a moderator to tell a player they are about to be muted.

Mutes are kept in-memory  to reduce implementation and  maintenance cost. Mutes are IP-based, this prevents a player from simply logging back in to remove a mute.

## Commits

commit 0ade3278b4bbfdd8443d3c2169fd7331f2a725db

    Add TimeManager convenience utility to get an Instant at UTC by year/month/day/hour/minute

commit 9ba6a3add4d9bf7ad3a1f18eeb9dba62176b2f88

    Suppress muted-chatter messages, send them a message that they are muted

commit 3bc3a2b484a536f21fa88f909ca35c6a45c3a42a

    Rename class: TimeManager->DateTimeUtil

commit b3566e4c6b8fabb9188672e3dea568272397faf1

    Move method 'toDateString' from DateTimeUtil to DateTimeFormatterUtil

commit d11b4b0575c3c91057a8ff77e0846038d50b31ba

    Update server mute controller to send mute requests to chatters backend data model

commit 641b79d733de44e0f597124644ed793abb3074a7

    Convert mutes from player-chat-id based to IP address based
    
    - Requires attaching IP addresses to chatter sessions so we can identify
      chatters by IP.

commit 87e6db938277a9f868dc3eaf08e3a45efd992f51

    Send broadcast message to lobby informating everyone of mute.

commit 060da64cfffe04e66e81a6d9ef66b14c0734fede

    Remove right click menu mute confirmation prompt, lobby broadcast message is confirmation.

## Functional Changes

[] New map or map update
[x] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
- launched a local lobby
- connected with two clients, one being moderator (test:test)
- muted the other player
- verified: 
  - message is broadcasted when mute occurs
  - muted player messages are dropped
  - if logging back in, still muted
  - player muted messages are responded (to only them) that they are muted
  - mute is removed after expiry and muted player is able to chat again
 

## Screens Shots

Screenshot from 2020-05-12 22-37-26


http-server/src/main/java/org/triplea/modules/chat/Chatters.java
```diff
@@ -97,4 +107,73 @@ public boolean disconnectPlayerSessions(final UserName userName, final String di
...
+    return Optional.ofNullable(playerMutes.get(inetAddress))
+        .map(
+            muteInstant -> {
+              if (muteInstant.isAfter(clock.instant())) {
+                return muteInstant;
+              } else {
+                playerMutes.remove(inetAddress);
+                return null;
+              }
+            });
```
RoiEXLab, line 130: This sounds like a good candidate for [Map#computeIfPresent](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Map.html#computeIfPresent(K,java.util.function.BiFunction)):
```java
return Optional.ofNullable(playerMutes.computeIfPresent(inetAddress, (key, muteInstant) -> {
  if (muteInstant.isAfter(clock.instant())) {
    return muteInstant;
  }
  return null;
}));
```
Note that I haven't actually tested this code, but it should have the exact same behaviour if the javadocs are accurate

http-server/src/main/java/org/triplea/modules/chat/event/processing/PlayerIsMutedMessage.java
```diff
@@ -0,0 +1,26 @@
...
+class PlayerIsMutedMessage {
+
+  @VisibleForTesting
+  static final BiFunction<Clock, Instant, String> muteDurationRemainingToString =
+      (clock, muteExpiry) -> {
+        final long minutes = Duration.between(clock.instant(), muteExpiry).toMinutes();
+        return minutes > 0
+            ? minutes + " minutes"
+            : Duration.between(clock.instant(), muteExpiry).toSeconds() + " seconds";
+      };
```
RoiEXLab, line 20: Thoughts on using a simple static method instead of a lambda edxpression?

http-server/src/main/java/org/triplea/web/socket/InetExtractor.java
```diff
@@ -15,11 +15,16 @@
 
   @SuppressWarnings("UnstableApiUsage")
   public static InetAddress extract(final Map<String, Object> userSession) {
-    // expected format '/127.0.0.1:42840'
-    final String ipString = String.valueOf(userSession.get(IP_ADDRESS_KEY)).substring(1);
+    // expected format '/127.0.0.1:42840' or (for test-cases) '127.0.0.1'
```
RoiEXLab, line 18: I don't think it's a good thing to get different values from test cases that we make special cases for in non-test code.

http-server/src/test/java/org/triplea/modules/chat/event/processing/PlayerIsMutedMessageTest.java
```diff
@@ -0,0 +1,43 @@
...
+  @Test
+  @DisplayName("Verify an example mute message calculation with 10 minutes remaining")
+  void verifyTimeDurationComputation() {
+    final Instant banExpiry = CURRENT_TIME.plus(10, ChronoUnit.MINUTES);
+
+    final String result =
+        PlayerIsMutedMessage.muteDurationRemainingToString.apply(
+            Clock.fixed(CURRENT_TIME, ZoneOffset.UTC), banExpiry);
+
+    assertThat(result, is("10 minutes"));
```
RoiEXLab, line 29: If for some reason this test will take longer than a minute to execute we might run into a failure 😛

http-server/src/test/java/org/triplea/modules/chat/event/processing/PlayerIsMutedMessageTest.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  @Test
+  @DisplayName("Verify an example mute message calculation with seconds remaining")
+  void verifyTimeDurationComputationWithSecondsRemaining() {
+    final Instant banExpiry = CURRENT_TIME.plus(20, ChronoUnit.SECONDS);
+
+    final String result =
+        PlayerIsMutedMessage.muteDurationRemainingToString.apply(
+            Clock.fixed(CURRENT_TIME, ZoneOffset.UTC), banExpiry);
```
RoiEXLab, line 39: 1 second is a tight time window.
Let's hope we never run into any issues with that. 🙏

java-extras/src/main/java/org/triplea/java/TimeManager.java
```diff
@@ -1,38 +0,0 @@
-package org.triplea.java;
```
RoiEXLab, line 1: Thanks for merging the classes 👍

http-server/src/main/java/org/triplea/modules/chat/Chatters.java
```diff
@@ -24,6 +32,8 @@
   @Getter(value = AccessLevel.PACKAGE, onMethod_ = @VisibleForTesting)
   private final Map<String, ChatterSession> participants = new ConcurrentHashMap<>();
 
+  private final Map<InetAddress, Instant> playerMutes = new HashMap<>();
```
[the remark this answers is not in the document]
DanVanAtta: I'm not sure it really helps, could be potentially cargo-culting to make it concurrent without a need.

In essence, the concurrency problem is not important enough to properly solve. For example, let's say we get a concurrent chat and mute at the same time. The ordering of which one will land first can be left non-deterministic.  In theory we'd have to create a lock whenever a request enters to mute and block any chat messages until that lock is clear. Simply ensuring concurrent access to the map therefore is just not enough to get proper ordering, and even then if you take an even larger perspective then network latency comes into play. On another perspective, if multiple moderators mute at the same time, one will overwrite each other, that overwrite will be non-deterministic whether we use a concurrent map or not.

http-server/src/main/java/org/triplea/modules/chat/Chatters.java
DanVanAtta: Does that remove expired mutes? The final return null seems like we might be setting a null value, it would be better though to remove the entire entry.

http-server/src/main/java/org/triplea/modules/chat/event/processing/PlayerIsMutedMessage.java
DanVanAtta: I think this was done to be able to override and mock the function, looks like that relic. I'll look into the test cases, this probably can be a 'simple' method. On the other hand, static `@VisibleForTesting` methods are a code smell, likely better to make this be injected behavior.

http-server/src/main/java/org/triplea/web/socket/InetExtractor.java
DanVanAtta: Generally agree. I'm on the fence as it's a huge PITA for test code to inject values in the right format and creates a lot of coupling to a specific format. Probably better to create a test utility method for less coupling; So for sure this is hack, but the cost of the hack vs ROI of better test utility method is not clear.

http-server/src/test/java/org/triplea/modules/chat/event/processing/PlayerIsMutedMessageTest.java
DanVanAtta: Note the presence of the fixed clocks: `Clock.fixed(CURRENT_TIME, ZoneOffset.UTC)`; that should keep this fully deterministic.

http-server/src/test/java/org/triplea/modules/chat/event/processing/PlayerIsMutedMessageTest.java
DanVanAtta: Ditto on fixed clocks, the chrono never changes.

java-extras/src/test/java/org/triplea/java/DateTimeFormatterUtilTest.java
```diff
@@ -35,4 +38,12 @@ void verifyFormattingNoTimeZone() {
             DEC_FIRST_EPOCH_MILLIS, FormatOption.WITHOUT_TIMEZONE);
     assertThat(result, is("2000-12-1 15:59"));
   }
+
+  @Test
+  void toDateString() {
+    assertThat(
+        DateTimeFormatterUtil.toDateString(
+            LocalDateTime.ofInstant(JAN_FIRST_INSTANT, ZoneOffset.UTC)),
+        is("Wed. Jan. 01 14:30:00 UTC 2020"));
```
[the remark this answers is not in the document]
DanVanAtta: Ah, this is a bug of me moving this test from DateTimeUtil to this class. The DateTimeUtil has a capability to inject locale, that was done in the other test but not in this one.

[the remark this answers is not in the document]
DanVanAtta: Turns out, we need to set the locale explicitly with the formatter for it to be used.

0694e3c1b

http-server/src/main/java/org/triplea/modules/chat/event/processing/PlayerIsMutedMessage.java
DanVanAtta: Improved this a bit by making the bi-function to be a static inner class: b8ef3f663

http-server/src/main/java/org/triplea/modules/chat/Chatters.java
RoiEXLab: According to the Javadoc I linked it does.

http-server/src/test/java/org/triplea/modules/chat/event/processing/PlayerIsMutedMessageTest.java
RoiEXLab: Ah, missed that detail

http-server/src/main/java/org/triplea/modules/chat/Chatters.java
DanVanAtta: Cool, I did not know that. We can also use a ternary here as well to simplify further. 

Updated: 53b4f70e2

http-server/src/main/java/org/triplea/modules/chat/event/processing/PlayerIsMutedMessage.java
```diff
@@ -4,23 +4,34 @@
...
 
-  String build(final Instant muteExpiry) {
-    return "You have been muted, expiring in: "
-        + muteDurationRemainingToString.apply(Clock.systemUTC(), muteExpiry);
+    @Override
+    public String apply(final Instant muteExpiry) {
+      final long minutes = Duration.between(clock.instant(), muteExpiry).toMinutes();
+      return minutes > 0
+          ? minutes + " minutes"
+          : Duration.between(clock.instant(), muteExpiry).toSeconds() + " seconds";
```
RoiEXLab, line 34: side-note: We could store the result of `Duration.between(clock.instant(), muteExpiry)` in a variable instead of potentially calculating it twice, but just a minor detail

http-server/src/main/java/org/triplea/modules/chat/Chatters.java
RoiEXLab: I didn't know that until yesterday either, but I finally found a use-case for this method ^^

PR #6473, opened 2020-05-15
Improve map not found error messaging

DanVanAtta:
Tell the user the name of the map that could not be found
and the paths on file system that were searched.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix: error messaging improvement 
[] Other:   

### Before

> SEVERE: null
games.strategy.engine.framework.startup.launcher.MapNotFoundException
	at games.strategy.triplea.ResourceLoader.getMapResourceLoader(ResourceLoader.java:87)
	at org.triplea.util.LocalizeHtml.localizeImgLinksInHtml(LocalizeHtml.java:45)
	at games.strategy.engine.framework.ui.GameChooser.updateInfoPanel(GameChooser.java:166)
	at games.strategy.engine.framework.ui.GameChooser.lambda$setupListeners$4(GameChooser.java:195)

### After

> SEVERE: Could not find map: steampunk
Not found, searched these locations:
/home/dan/triplea/downloadedMaps/steampunk/map
/home/dan/triplea/downloadedMaps/steampunk
/home/dan/triplea/downloadedMaps/steampunk-master/map
/home/dan/triplea/downloadedMaps/steampunk-master
/home/dan/triplea/downloadedMaps/steampunk.zip
/home/dan/triplea/downloadedMaps/steampunk-master.zip
/home/dan/triplea/downloadedMaps/steampunk.zip

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/MapNotFoundException.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/MapNotFoundException.java
@@ -11,13 +11,15 @@ import java.util.stream.Collectors;
 public class MapNotFoundException extends IllegalStateException {
   private static final long serialVersionUID = -1027460394367073991L;
 
-  public MapNotFoundException() {}
-
   public MapNotFoundException(final String mapName, final List<File> candidatePaths) {
     super(
         "Could not find map: "
             + mapName
-            + "\nNot found, searched these locations:\n"
+            + "\nTypically this will be because the map is not downloaded."
+            + "\n\nIf you are *sure* you have the map, double check that the"
+            + "\ngame XML can be found under one of the locations listed below"
+            + "\nand the 'mapName' XML attribute in the XML file has the correct name."
+            + "\nSearched these locations:\n"
             + candidatePaths.stream().map(File::getAbsolutePath).collect(Collectors.joining("\n")));
   }
 }
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/MapNotFoundException.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/MapNotFoundException.java
@@ -1,5 +1,9 @@
 package games.strategy.engine.framework.startup.launcher;
 
+import java.io.File;
+import java.util.List;
+import java.util.stream.Collectors;
+
 /**
  * An unchecked exception indicating the map associated with a save game was not found and must be
  * installed before game play can begin.
@@ -8,4 +12,12 @@ public class MapNotFoundException extends IllegalStateException {
   private static final long serialVersionUID = -1027460394367073991L;
 
   public MapNotFoundException() {}
+
+  public MapNotFoundException(final String mapName, final List<File> candidatePaths) {
+    super(
+        "Could not find map: "
+            + mapName
+            + "\nNot found, searched these locations:\n"
+            + candidatePaths.stream().map(File::getAbsolutePath).collect(Collectors.joining("\n")));
+  }
 }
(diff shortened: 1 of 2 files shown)


Cernelius:
> SEVERE: Could not find map: steampunk
> Not found, searched these locations:
> /home/dan/triplea/downloadedMaps/steampunk/map
> /home/dan/triplea/downloadedMaps/steampunk
> /home/dan/triplea/downloadedMaps/steampunk-master/map
> /home/dan/triplea/downloadedMaps/steampunk-master
> /home/dan/triplea/downloadedMaps/steampunk.zip
> /home/dan/triplea/downloadedMaps/steampunk-master.zip
> /home/dan/triplea/downloadedMaps/steampunk.zip

Why are there two of the same instance of:
`/home/dan/triplea/downloadedMaps/steampunk.zip`
?

And why are there 4 non-zipped instances, but only 3 zipped ones?

Is it missing the ability to read zipped folders having a path like:
`/home/dan/triplea/downloadedMaps/steampunk/map`
or
`/home/dan/triplea/downloadedMaps/steampunk-master/map`
or both?

I know I could test this (and I guess I'll do it), but here I'm mostly pointing out that it may feel weird on the receiver getting two instances of the same thing (the 5th and 7th one read the same, if they are not) and 4 non-zipped and 3 zipped alternatives, like if one of the 4 will work only non-zipped.

DanVanAtta:
> Why are there two of the same instance of:
/home/dan/triplea/downloadedMaps/steampunk.zip
?

DanVanAtta:
> Why are there two of the same instance of:
/home/dan/triplea/downloadedMaps/steampunk.zip


Either a copy/paste issue or the lookup logic is buggy to look at the same location multiple times. Sorry, I do not have the repro case readily available. Either way fixing that is beyond the scope of this update and exposing that bug is far less harmful compared to leaving someone in the dark about why their map is not working (the game engine goes on to ask a person to download the map, which then also fails, which is actually another bug as the engine should be smart enough to see if that map exists for download).

> And why are there 4 non-zipped instances, but only 3 zipped ones?

That's the search code.. It looks at a variety of locations. It's a bit of a mess actually there are so many variants.

>  Is it missing the ability to read zipped folders having a path like:
/home/dan/triplea/downloadedMaps/steampunk/map

A zip file like: `/home/dan/triplea/downloadedMaps/steampunk/map/seampunk.zip` does not really make sense. But yes, the game engine would not look for a zip at that path. Ditto for the -master variant.

> I know I could test this (and I guess I'll do it), but here I'm mostly pointing out that it may feel weird on the receiver getting two instances of the same thing (the 5th and 7th one read the same, if they are not) and 4 non-zipped and 3 zipped alternatives, like if one of the 4 will work only non-zipped.

As weird as the paths are, those are the paths that are valid. I think it's better than nothing and was the information I needed to debug why `steampunk_ultimate` was not loading.. The fact it's weird is a different problem compared to us not showing anything other than `null` and `map missing, would you like to download it?` (which does not help if you know you are working with a local XML that you just selected from the game list).

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/MapNotFoundException.java
```diff
@@ -1,11 +1,25 @@
...
+  public MapNotFoundException(final String mapName, final List<File> candidatePaths) {
+    super(
+        "Could not find map: "
+            + mapName
+            + "\nTypically this will be because the map is not downloaded."
+            + "\n\nIf you are *sure* you have the map, double check that the"
+            + "\ngame XML can be found under one of the locations listed below"
+            + "\nand the 'mapName' XML attribute in the XML file has the correct name."
+            + "\nSearched these locations:\n"
+            + candidatePaths.stream().map(File::getAbsolutePath).collect(Collectors.joining("\n")));
```
RoiEXLab, line 23:
```suggestion
            + candidatePaths.stream().map(File::getAbsolutePath).distinct().collect(Collectors.joining("\n")));
```
If you want to "fix" the issue mentioned by @Cernelius you could just add this call for now 🤷

DanVanAtta: That's a bit of a band-aid fix as we are still searching duplicate paths. `candidatePaths` should not return duplicates in the first place.

Issue #6453, opened 2020-05-09
moderator last access is missing in toolbox


DanVanAtta:
### How can the problem be recreated?

- login as a registered user
- root cause: notice the lobby_user table 'last_login' column is not being populated

### What is an expected fix?

Ideal:
- Remove last_login column from lobby_user table
- Add a nullable FK from access_log table to lobby_user
- Query last login of a user by joining lobby_user to access_log and finding max of login-time

Alternative (band-aid):
- update last login time on successful login. This is a band-aid fix as we incur an extra DB query and the FK column is more useful and would make this update throw-away.

PR #6491, opened 2020-05-19
Add lobby_user FK to access_log

DanVanAtta:
## Overview

Lobby_user table last_login column was not being updated. This update drops that column and update access_log to have a FK column to lobby_user. We can then use the access_log table to determine last login time. 

## Commits

commit 485d7433583706166c759813f4e4428900c08aff

    Update access_log table, replace registered column with FK to lobby_user
    
    Instead of adding a 'true/false' value to a registered column
    in access_log, we can use a nullable foreign key to the lobby_user
    table. We'll insert a null FK value if the user is anonymous,
    otherwise we'll insert into the access_log table their lobby_user.id

commit 3df952b9a5e2837300e1dacb9084dca081f875f8

    Drop last_login from lobby_user and replace with access_log.access_time
    
    - The value in lobby_user is buggy and is not being updated.
    - Access log table now has a FK to the lobby_user table which
      give us the last login time. To know the last login time of a user
      we can join lobby_user with access_log and select the max
      access_time to know their last login.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix: https://github.com/triplea-game/triplea/issues/6453 "moderator last access is missing in toolbox"
[] Other:   

## Testing
- verified in a local lobby
- automated test coverage is good

### Before

Screenshot from 2020-05-18 22-35-57

### After

(Notice 'last_login' is now populated)
Screenshot from 2020-05-18 22-17-43

### DIFF
--- a/http-server/src/main/java/org/triplea/db/dao/moderator/ModeratorUserDaoData.java
+++ b/http-server/src/main/java/org/triplea/db/dao/moderator/ModeratorUserDaoData.java
@@ -14,9 +14,6 @@ import org.triplea.db.TimestampMapper;
 @NoArgsConstructor
 @Getter
 public class ModeratorUserDaoData {
-  public static final String USERNAME_COLUMN = "username";
-  public static final String LAST_LOGIN_COLUMN = "last_login";
-
   private String username;
   private Instant lastLogin;
 
@@ -25,7 +22,7 @@ public class ModeratorUserDaoData {
     return (rs, ctx) ->
         ModeratorUserDaoData.builder()
             .username(rs.getString("username"))
-            .lastLogin(TimestampMapper.map(rs, "last_login"))
+            .lastLogin(TimestampMapper.map(rs, "access_time"))
             .build();
   }
 }
--- a/http-server/src/main/java/org/triplea/db/dao/user/UserJdbiDao.java
+++ b/http-server/src/main/java/org/triplea/db/dao/user/UserJdbiDao.java
@@ -15,9 +15,6 @@ public interface UserJdbiDao {
   @SqlQuery("select bcrypt_password from lobby_user where username = :username")
   Optional<String> getPassword(@Bind("username") String username);
 
-  @SqlUpdate("update lobby_user set last_login = now() where username = :username")
-  int updateLastLoginTime(@Bind("username") String username);
-
   @SqlUpdate(
       "update lobby_user set password = null, bcrypt_password = :newPassword where id = :userId")
   int updatePassword(@Bind("userId") int userId, @Bind("newPassword") String newPassword);
--- a/http-server/src/main/java/org/triplea/modules/user/account/login/AccessLogUpdater.java
+++ b/http-server/src/main/java/org/triplea/modules/user/account/login/AccessLogUpdater.java
@@ -21,15 +21,10 @@ class AccessLogUpdater implements Consumer<LoginRecord> {
   @Override
   public void accept(final LoginRecord loginRecord) {
     final int updateCount =
-        loginRecord.isRegistered()
-            ? accessLogDao.insertRegisteredUserRecord(
-                loginRecord.getUserName().getValue(),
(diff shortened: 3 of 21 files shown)


http-server/src/main/java/org/triplea/db/dao/moderator/ModeratorUserDaoData.java
```diff
@@ -14,9 +14,6 @@
 @NoArgsConstructor
 @Getter
 public class ModeratorUserDaoData {
-  public static final String USERNAME_COLUMN = "username";
-  public static final String LAST_LOGIN_COLUMN = "last_login";
```
DanVanAtta, line 18: These two constants were left-over from previous cleanups, not used.

http-server/src/main/java/org/triplea/db/dao/user/UserJdbiDao.java
```diff
@@ -15,9 +15,6 @@
   @SqlQuery("select bcrypt_password from lobby_user where username = :username")
   Optional<String> getPassword(@Bind("username") String username);
 
-  @SqlUpdate("update lobby_user set last_login = now() where username = :username")
-  int updateLastLoginTime(@Bind("username") String username);
```
DanVanAtta, line 19: Nothing called this update method, it can be freely removed. It's a bug that we never invoked it. The fix here is a deeper one to get this same value from the access_log table rather than maintaining the last login in two places.

http-server/src/main/java/org/triplea/modules/user/account/login/AccessLogUpdater.java
```diff
@@ -21,15 +21,10 @@ public static AccessLogUpdater build(final Jdbi jdbi) {
...
-        loginRecord.isRegistered()
-            ? accessLogDao.insertRegisteredUserRecord(
-                loginRecord.getUserName().getValue(),
-                loginRecord.getIp(),
-                loginRecord.getSystemId().getValue())
-            : accessLogDao.insertAnonymousUserRecord(
-                loginRecord.getUserName().getValue(),
-                loginRecord.getIp(),
-                loginRecord.getSystemId().getValue());
+        accessLogDao.insertUserAccessRecord(
```
DanVanAtta, line 24: `insertUserAccessRecord` can now figure out on its own if a user is registered or not by selecting from the lobby_user table.

lobby-db/sample_data.sql
```diff
@@ -24,12 +24,12 @@ values (1000, 'test', 'email@email.com', (select id from user_role where name =
...
-insert into access_log(access_time, username, ip, system_id, registered)
-values (now() - interval '1 days', 'user1', '1.1.1.1', 'system-id1', false),
-       (now() - interval '2 days', 'user2', '1.1.1.2', 'system-id2', false),
-       (now() - interval '3 days', 'user3', '1.1.1.3', 'system-id3', true),
-       (now() - interval '4 days', 'user1', '1.1.1.2', 'system-id4', false),
-       (now() - interval '5 days', 'user2', '1.1.1.4', 'system-id5', false);
+insert into access_log(access_time, username, ip, system_id, lobby_user_id)
+values (now() - interval '1 days', 'user1', '1.1.1.1', 'system-id1', null),
+       (now() - interval '2 days', 'user2', '1.1.1.2', 'system-id2', null),
+       (now() - interval '3 days', 'user3', '1.1.1.3', 'system-id3', null),
```
RoiEXLab, line 30: Why is this one null when it was true before?

DanVanAtta: The 'true' was to give a sampling of data, it was easy to do. Now it takes a bit more work to FK it to an existing user row. If you log in as a moderator, you will be registered, it will be recorded in the access log, so that will just as well give an example of a registered user logging in and how that looks in the access log tab.

PR #6500, opened 2020-05-20
Add property 'canRetreatOnStalemate' and allow v3 transport vs transport retreat

DanVanAtta:
- Update stalemate battle to check if attacker can retreat, and allow retreat if so.
- Add a stalemate battle check that first looks for a new 'canRetreatOnStalemate'
  property. If any unit has the property set, we will use the property value. If the
  property value has a mix of 'true' & 'false' any 'false' values will prevail.
  Otherwise, if no property, we will fallback to checking if we are in a 
  V3 transport vs transport situation which should allow retreat. We do 
  the fallback to avoid updating all maps to have an explicit property and still
  have a rule fix.
- Finally, add the 'canRetreatOnStalemate' property that can be added as a unit
  attachment option. If it is not specified then we default to not looking at it
  and will only check if we have V3 transport vs transport.

Resolves: https://github.com/triplea-game/triplea/issues/2367

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing
On 1941 v3, set up a cruiser & transport vs cruiser & transport battle (gives good odds both cruisers will hit leaving transport vs transport).
- verified that without any properties set that a transport vs transport is offered retreat. 
- verified that setting 'canRetreatOnStalemate=false' to a transport removes the retreat prompt

## Additional Review Notes

This is a better fix of v3 transport vs transport rules compared to the one initially proposed in: https://github.com/triplea-game/triplea/pull/6033, and adds a property that can be used to customize this behavior on other maps as desired.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
@@ -75,6 +75,7 @@ public class UnitAttachment extends DefaultAttachment {
   private int attackRolls = 1;
   private int defenseRolls = 1;
   private boolean chooseBestRoll = false;
+  private Boolean canRetreatOnStalemate;
 
   // sub/destroyer related
   private boolean canEvade = false;
@@ -2610,6 +2611,22 @@ public class UnitAttachment extends DefaultAttachment {
     tuv = -1;
   }
 
+  public void setCanRetreatOnStalemate(final boolean value) {
+    canRetreatOnStalemate = value;
+  }
+
+  public void setCanRetreatOnStalemate(final String value) {
+    canRetreatOnStalemate = getBool(value);
+  }
+
+  public Boolean getCanRetreatOnStalemate() {
+    return canRetreatOnStalemate;
+  }
+
+  public void resetCanRetreatOnStalemate() {
+    canRetreatOnStalemate = null;
+  }
+
   /**
    * Returns the maximum number of units of the specified type that can be placed in the specified
    * territory according to the specified stacking limit (movement, attack, or placement).
@@ -4363,6 +4380,11 @@ public class UnitAttachment extends DefaultAttachment {
             "isAAmovement",
             MutableProperty.<Boolean>ofWriteOnly(this::setIsAaMovement, this::setIsAaMovement))
         .put("isTwoHit", MutableProperty.<Boolean>ofWriteOnly(this::setIsTwoHit, this::setIsTwoHit))
+        .put(
+            "canRetreatOnStalemate",
+            MutableProperty.of(
+                this::setCanRetreatOnStalemate, this::setCanRetreatOnStalemate,
+                this::getCanRetreatOnStalemate, this::resetCanRetreatOnStalemate))
         .build();
   }
 }
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
@@ -38,6 +38,7 @@ import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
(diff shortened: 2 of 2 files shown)


DanVanAtta:
@RoiEXLab, @Cernelius  any thoughts on this update?

Cernelius:
What "stalemate", or similar, cases does this apply to:
1- Both sides have 0 power.
2- One or both sides have 1+ power, but no side having positive power is able to target any units of the other side.
3- The battle reaches the maximum number of rounds.
4- Only the attacker has 1+ power (this is currently handled that the attacker can retreat, which is a wrong behaviour, but what would happen, if fixed, is that the battle ends immediately and all defending war units are automatically destroyed (would this option allow choosing to retreat, instead?)).
5- Only the attacker is able to target all or some opposite units (here, to be consistent with point 4, what would make sense happening is that the battle ends immediately and all defending units that can be targeted are automatically destroyed).

Also, how about in any of the aforementioned cases, there are still AA attacks, so the two sides can still actually target each other?

I tend to think this property should have been done after finalizing and fixing whatever else it may apply to, but as long as you can foresee the application of it, I guess it's fine either way.

Cernelius:
To make a practical example of cases 1 and 4, let's take D1914NML.

We have two cases (to keep it simple, let's say this happened during the course of the first battle round of a battle that started with also 1 defending infantry taken as casualty):
1- 1 conscript is alone against 1 trench and no other defending units.
2- 1 infantry is alone against 1 trench and no other defending units.

The infantry is the usual unit of all basic games, while the conscript is an infantry having attack value 0.

What correctly happens in the first case is that the conscript is stalemated. What I understand this option, assigned to the conscript, would give is the ability, for the conscript, to choose whether to stalemate or retreat.

What currently happens in the second case is that the infantry can choose whether to win the battle (autodestroying the trench) or retreat, while, by correct rules, it should not have the option to retreat. If that other issue would be fixed, would this option allow for getting back the ability to retreat?

Cernelius:
This also on the assumption that @ron-murhammer will add this option to the conscripts of D1914NML. Am I right?

DanVanAtta:
What "stalemate", or similar, cases does this apply to:
1- Both sides have 0 power


> I tend to think this property should have been done after finalizing and fixing whatever else it may apply to, but as long as you can foresee the application of it, I guess it's fine either way.

My initial reaction would have been to do it that way. When we hashed out the potential fix for this it is clear we want this behavior to be configurable. For the sake of limiting round-trips and iterations, it's included here for this to be a complete fix and to put the issue to bed.



> What correctly happens in the first case is that the conscript is stalemated. What I understand this option, assigned to the conscript, would give is the ability, for the conscript, to choose whether to stalemate or retreat.

Yes, if you add a `canRetreatOnStalemate` option to the conscriptthen it will be given an option to retreat on stalemate.




> What currently happens in the second case is that the infantry can choose whether to win the battle 

The infantry has a non-zero attack power, it's not technically a stalemate, the update here does not apply (does not change the behavior one way or another).


> This also on the assumption that @ron-murhammer will add this option to the conscripts of D1914NML. Am I right?


@ron-murhammer is pretty swamped these days with other responsibilities. If he does not add the option, then presumably I would do so if the relevant map discussion thread comes to concensus.

Cernelius:
I strongly believe this should apply at the very least to the first 3 cases I made, namely:
>1- Both sides have 0 power.
2- One or both sides have 1+ power, but no side having positive power is able to target any units of the other side.
3- The battle reaches the maximum number of rounds.

Both sides having power 0 is conceptually really the same as both sides being unable to target each other (if @panther2 can confirm).

Max rounds is another king of stalemate, but it is still a stalemate, and called as such by TripleA, in multiple places. If you are positive not to include it, then I urge you to change the name of this option, as `canRetreatOnStalemate` is definitely saying that I can retreat instead of getting stalemated, in every case.

However, I think it would be great if this option apply to all the first 3 cases I made. Especially the ability to retreat upon reaching the maximum number of combat rounds would be a great option to have (it would also allow having retreat and 1 combat round, which is currently impossible).

DanVanAtta:
@Cernelius since we're now talking about expanding the feature, would you be okay if we extracted the remaining requirements conversation to a new thread? It's much easier to make small additions then it is to get everything done all at once.

Cernelius:
The risk I see here is that something is labelled as `canRetreatOnStalemate`, while it is not applying to all cases of stalemate. So, the naming would be inconsistent until expanded. On the other hand, if named more precisely, then it would have to be renamed if expanded.

panther2:
@Cernelius 
> Both sides having power 0 is conceptually really the same as both sides being unable to target each other (if @panther2 can confirm).

Actually there is a conceptual difference:
- Both sides have 0 power (Transport stalemate): When the battle ends the attacker either remains in the contested sea zone  or is (exceptionally) granted a retreat option.
- Only units that cannot hit each other remain: The battle ends, there is no retreat option. The attacker stays in the contested sea zone.

Cernelius:
> @Cernelius
> 
> Both sides having power 0 is conceptually really the same as both sides being unable to target each other (if @panther2 can confirm).
> 
> Actually there is a conceptual difference:
> 
> Both sides have 0 power (Transport stalemate): The battle ends. The attacker either remains in the contested sea zone  or is (exceptionally) granted a retreat option.
> Only units that cannot hit each other remain: The battle ends, there is no retreat option. The units remain in the contested sea zone.

I meant aside from the special retreat ability of transports against transports (that has to apply only to the 0 power case, since everything can otherwise hit transports).

In practice, you can compare these two situations, instead:
- Attack 0 carriers alone remain against transports alone.
- Only units that cannot hit each other remain.

Isn't this conceptually the same?

Also, here we need some level of generalization, if we want to make something that is sound for TripleA in general. You can think about it in terms of saying if I make a variant in which battleships would have attack and defence 0, would that be the same as battleships being unable to target other battleships, if you have a battleship only battle?

In my opinion, ending up in a situation in which both sides have power 0 should give exactly the same results as ending up in a situation in which both sides are unable to target each other. Still in my opinion, the exceptional behaviour of transports is not to be seen specifically related to either, meaning that, if I would assign the option at this PR to a unit, that option should be effective in either case.

Finally, again, if one doesn't want to apply this option to any stalemate case, then the option should be called something else than `canRetreatOnStalemate`.

---

@panther2 Do you have any opinion about the combat rounds time-out stalemate?

Cernelius:
@DanVanAtta Also, am I right assuming the 0 power case is indifferently applied also to the case of having positive strength but 0 rolls?

DanVanAtta:
Any changes to the code means it all needs to be re-tested. That is a cost of 30 minutes. The more time goes by, the more context switch cost there is to simply remember "How do I set this up", "where is the save", etc.. This is in part why I say it's better to do incremental/small changes, it's less redundant work in testing. The biggest cost in TripleA is the extensive and redundant manual testing we have to do with just about every code change. That is not to mention the context cost of just remembering what was updated, where, and the risk of even more work coming from merge conflicts that almost nobody likes. We need to stay focused in PRs to describing whether the proposed updates are improvements and not extrapolate to future updates.

With that said, unless there can be objection to the current update, then I'll move forward and we can continue discussing further stalemate scenarios out-of-band from a code update. I'll note that there is a feature to "reference comment in a new issue" that makes it easy to pull any comment out into its own issue.

panther2:
@Cernelius 

I am answering here, as this part of the discussion has not been transferred to the new issue:

>In practice, you can compare these two situations, instead:
>
>Attack 0 carriers alone remain against transports alone.
>Only units that cannot hit each other remain.
>Isn't this conceptually the same?

In case you won't grant a retreat option to either of them yes, I agree.

>In my opinion, ending up in a situation in which both sides have power 0 should give exactly the same results as ending up in a situation in which both sides are unable to target each other. 

Yes, the battle ends in any case without any retreat option (apart from the exceptional Transport stalemate).

Not sure if I have fully covered your questions now. Maybe I have still not understood what you are after... in that case please ask again.

PR #6501, opened 2020-05-20
Battle Tree Estimator

trevan:
This adds a new battle estimator and two AIs that use it.  One of the AIs (BattleTree (AI)) uses the new estimator while the other AI (BattleTree Compare (AI)) runs both the new estimator and the existing concurrent estimator and compares the results.  Both of the AIs are hidden behind the beta flag.

I'm calling the battle estimator "BattleTree" because it generates the tree of possible results.  It uses a depth-first search to calculate the probabilities of the results.  In the forum, I had called it Minimax, but it really isn't doing that.

## Functional Changes

[] New map or map update
[x] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:  

## Testing
I've added new tests to check that the probabilities are calculated correctly.

I've manually run a few battles using 1940 Global using the BattleTree Comp AI and the results look good.

I've also added a test called BattleCalculatorComparisonTests.comparePerformanceOfCalculators.  I used this test for performance testing (using Async Profiler).  This test doesn't do any asserts so it really isn't a true test.  I'd like to keep it around somewhere but I'm not sure where.

### Duplicate code
I copied several functions from MustFightBattle.  They are:
 * MustFightBattle::addCheckEndBattleAndRetreatingSteps (one of its steps)
 * MustFightBattle::checkForUnitsThatCanRollLeft
 * MustFightBattle::checkUndefendedTransports
 * MustFightBattle::submergeSubsVsOnlyAir
 * MustFightBattle::updateOffensiveAaUnits

The existing functions are instance methods of MustFightBattle so I'm not able to utilize them from within BattleStep.  Some also will call methods on bridges or update instance data in MustFightBattle.  I'd like to move them to helper classes so that the code can be shared.

### Missing battle types
I know that the calculator doesn't handle:
 * bombarding
 * sub battle order  - it does attacker subs, then defender subs
 * amphibious battles - not sure if I actually need to worry about this but I haven't checked

### Calculator comparation
The BattleTree Comp AI compares the win percentage between the ConcurrentBattleCalculator and the BattleTreeCalculator.  It allows a 5% deviation but I'm not sure if that is good enough.  If I run the ConcurrentBattleCalculator with a runCount of 20,000, there is very few deviations in Global 1940.  But if it uses the default runCount (which is determined by the number of units in the battle), there are a lot of deviations.  These deviations are a false positive because the ConcurrentBattleCalculator didn't run enough battle simulations.  But increasing the runCount to 20,000 will make the Comp AI extremely slow and will also hide the cases where ConcurrentBattleCalculator is faster.  I'd like to make it dynamic somehow so that beta testers can run either situation.


Cernelius:
How about units with AA shots after the first round of combat (that may be advisable taking out after other units with greater regular power)?

DanVanAtta:
@trevan Good to see the PR :+1: 
4K'ish net lines of code is quite a lot!

For expectations, I'd suspect we're only 30-60% of the way through this process before merge - there is potentially quite a bit more to do.

With that said, I think we should discuss a "merge-plan" to get this landed before we get too deep into anything. Some areas to discuss:
- code climate's
- unit testing
- code re-use
- AI performance & scaling (does this algorithm work well if you have max units selected in battle calculator? If not, at what point does it break down?)
- AI performance in terms of game-play (can this AI hold it's own against Fast AI?)
- Questions on future direction. EG: What tie-in's to future improvements & enchancements should we realistically consider? I've been thining a decision tree for LL should be really simple and strikingly fast - is this code in a good position for such an enhancement? Are there any other enhancements that we are pretty sure we'd want to do?
- Design/structural review of the code
- Detailed review
- Potential preliminary/break-up PRs to pave the way for a new AI. For example, what areas are there that we can carve off into other PRs so that this one could become more or less, just the new AI.

@trevan what are your thoughts on any of the above points and how you'd like to see this land? Also, could you give us a sense of your coding background and if breaking up a PR, unit testing, and/or further work to for example address code-climates seem reasonable/easy-enough, or if you'd like for us to collaborate on the coding and help with that?

trevan:
> How about units with AA shots after the first round of combat (that may be advisable taking out after other units with greater regular power)?

@Cernelius , what do you mean?  The tree currently tries AA guns every single round.  If you are talking about selecting these units as an early causality, I'd rather use the existing causality calculator and do any improvements in its selection.

trevan:
> @trevan Good to see the PR 👍
> 4K'ish net lines of code is quite a lot!

@DanVanAtta Probably half of that is tests or copied code.  The tests is a lot of lines but a lot of it is really similar.  I could probably create a helper function that would shrink the count of those lines.

> 
> For expectations, I'd suspect we're only 30-60% of the way through this process before merge - there is potentially quite a bit more to do.
> 
> With that said, I think we should discuss a "merge-plan" to get this landed before we get too deep into anything. Some areas to discuss:
> 
>     * code climate's
I'm fine with trying to fix what code climate caught.  Is there a way I can have IntelliJ do the code climate checks locally?  Otherwise, I'll have to fix them once I've pushed.
> 
>     * unit testing
I've added quite a bit of unit tests for this.  I do feel like the tests are brittle since they are checking that the results equal a specific number.  I'd like to somehow create a test that runs every iteration of a battle and then check that number against the BattleTree results.  Then I can ensure that the expected value is correct.
> 
>     * code re-use
> 
>     * AI performance & scaling (does this algorithm work well if you have max units selected in battle calculator? If not, at what point does it break down?)

Max units will definitely be a problem.  The biggest calculation I've done is 30 units of 5 different types on both sides for Global 1940 and 3 units of 15 different types on both sides for Total World War.  The Global 1940 took about 8 seconds and the Total World War took about 30 seconds.

Can it get faster?  Yes.  I've been using Async Profiler (https://github.com/jvm-profiling-tools/async-profiler) to see where the slow points are and have been chipping away at them.  I just recently figured out a performance win that halved the times.  I just pushed those changes.

It is also possible to not try every single possible branch.  I already exclude branches that have a probability of <.5% and could increase that more.  It just adds more uncertainty to the results.

> 
>     * AI performance in terms of game-play (can this AI hold it's own against Fast AI?)

I would expect it to hold its own against ProAI.  I just did a manual test of playing Global 1940 with the Comp AI.  It printed out when it had win percentages different between ConcurrentBattleCalculator and the BattleTreeCalculator.  In most of the differences, the BattleTree Calculator was actually more accurate.

I'd need to see it in action on other maps as well to see how the rule differences affect it.  One area I know that I haven't figured it out is the order of subs.  So if a map allows defender subs to attack first, BattleTree will calculate the win percentage in correctly.

> 
>     * Design/structural review of the code
> 
>     * Detailed review
> 
>     * Potential preliminary/break-up PRs to pave the way for a new AI. For example, what areas are there that we can carve off into other PRs so that this one could become more or less, just the new AI.

I copied a bunch of code out of MustFightBattle as I mentioned in the description.  I think a separate PR where that code is moved to a shared location can be carved off of this.  As for other pieces to carve out, I can't think of any.
> 
> 
> @trevan what are your thoughts on any of the above points and how you'd like to see this land? Also, could you give us a sense of your coding background and if breaking up a PR, unit testing, and/or further work to for example address code-climates seem reasonable/easy-enough, or if you'd like for us to collaborate on the coding and help with that?

I've pretty adept at coding.  I can break up a PR, edit git commits, build and run unit tests, etc.  I'm not great with Java UI coding, though.  Most of my testing of the changes have been through the unit test framework.  I only just recently plumbed in the AI code so that I could see it actually play in the UI.

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
```diff
@@ -0,0 +1,1216 @@
...
+        outcomes.compute(
+            new StepUnits(currentUnits, hitProbability * probability),
+            (key, value) -> {
+              if (value == null) {
+                return key;
+              } else {
+                value.addProbability(hitProbability * probability);
+              }
+              return value;
+            });
```
RoiEXLab, line 505: This is something I don't quite get.
Using this compute function the key would always be the same object as the value right?
So why bother using a map at all?

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
```diff
@@ -0,0 +1,1216 @@
...
+
+      rollsByDicePower.compute(
+          power,
+          (key, prevRolls) -> {
+            if (prevRolls == null) {
+              return rolls;
+            } else {
+              return prevRolls + rolls;
+            }
+          });
```
RoiEXLab, line 629: compute is nice and all, but I think using a classical approach makes htis more readable:
```suggestion
      rollsByDicePower.put(power, rollsByDicePower.getOrDefault(power, 0) + rolls);
```

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
```diff
@@ -0,0 +1,1216 @@
...
+                    battleSite,
+                    territoryEffects,
+                    parameters.isAmphibious,
+                    parameters.amphibiousLandAttackers),
+                parameters.getData());
+        if (attackPower == 0 && defensePower == 0) {
+          tieProbability = 1;
+          hasResult = true;
+        }
+         */
```
RoiEXLab, line 715: Commented out code should be remove in the end

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
```diff
@@ -0,0 +1,1216 @@
...
+        // an example scenario would be two units fighting
+        // each round, there is a possibility that no one will hit
+        // so the units never change
+
+        final double[] siblingResults = new double[5];
+        siblingResults[0] = cachedBattleStep.winProbability;
+        siblingResults[1] = cachedBattleStep.loseProbability;
+        siblingResults[2] = cachedBattleStep.tieProbability;
+        siblingResults[3] = cachedBattleStep.averageRounds;
+        siblingResults[4] = 1.0;
```
RoiEXLab, line 766: Instead of using a double array, why not create a dedicated type to hold all of these values?

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
```diff
@@ -0,0 +1,1216 @@
...
+    final Map<Integer, double[]> calculatedProbabilityCache =
+        this.calculatedProbabilityCache.computeIfAbsent(
+            rollData.maxDiceSides,
+            key -> new HashMap<>(rollData.maxDiceSides * rollData.totalRolls));
+
+    final int[] diceAvailableGroupedByPower = new int[maxPower + 1];
+    for (int dicePowerIndex = 0; dicePowerIndex < dicePowersList.size(); dicePowerIndex++) {
+      final int dicePower = dicePowersList.get(dicePowerIndex);
+
+      dicePowers[dicePowerIndex] = dicePower;
```
RoiEXLab, line 859: You can use `final int[] dicePowers = dicePowersList.toArray(new int[0]);` to convert a List to an array.
Because it's an arraylist implementation copying the whole array might be slightly faster than iterating

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
```diff
@@ -0,0 +1,1216 @@
...
+        diceAvailableGroupedByPower,
+        calculatedProbabilityCache);
+
+    final List<Double> probabilitiesList = new ArrayList<>();
+    for (final double probability : probabilities) {
+      probabilitiesList.add(probability);
+    }
+
+    return probabilitiesList;
+  }
```
RoiEXLab, line 880:
```suggestion
    return Arrays.stream(probabilities).boxed().collect(Collectors.toList());
  }
```

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
```diff
@@ -0,0 +1,1216 @@
...
+            + String.format("%.3f", loseProbability)
+            + " t:"
+            + String.format("%.3f", tieProbability)
+            + " b:"
+            + String.format("%.3f", badProbability)
+            + " [r:"
+            + round
+            + "]";
+
+    return out;
```
RoiEXLab, line 1214: 2 Things:
1. No need to create a local variable out here, you can just return directly.
2. I thinkit's better to just wrap everything inside a single String.format

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
```diff
@@ -0,0 +1,1216 @@
...
+      case ATTACKER:
+        type = "AT";
+        break;
+      case DEFENDER:
+        type = "DE";
+        break;
+      default:
+        type = "??";
+        break;
+    }
```
RoiEXLab, line 1183: This might be something that could get built into the Type enum.
Just give this enum a "string representation field" and call the getter here.

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleTreeCalculator.java
```diff
@@ -0,0 +1,124 @@
...
+      attackingOrderOfLoss.sort(
+          (unit1, unit2) ->
+              Boolean.compare(
+                  Matches.unitIsTransport().and(Matches.unitIsSea()).test(unit1),
+                  Matches.unitIsTransport().and(Matches.unitIsSea()).test(unit2)));
+      defendingOrderOfLoss.sort(
+          (unit1, unit2) ->
+              Boolean.compare(
+                  Matches.unitIsTransport().and(Matches.unitIsSea()).test(unit1),
+                  Matches.unitIsTransport().and(Matches.unitIsSea()).test(unit2)));
```
RoiEXLab, line 100: `Comparator.comparing(Matches.unitIsTransport().and(Matches.unitIsSea())::test)`
might even work already

game-core/src/main/java/games/strategy/triplea/ai/tree/StepUnits.java
```diff
@@ -0,0 +1,446 @@
...
+      enemyUnitsChances.put(unit, 1.0);
+    }
+    for (final Unit unit : retreatedEnemy) {
+      enemyUnitsChances.put(unit, 1.0);
+    }
+  }
+
+  void updateUnitChances(final StepUnits child, final double probability) {
+    for (final Map.Entry<Unit, Double> entry : child.enemyUnitsChances.entrySet()) {
+      friendlyUnitsChances.compute(
```
RoiEXLab, line 342: This case should never really occur if you ask me, Double should never be null, so I'd rather throw an exception here

game-core/src/main/java/games/strategy/triplea/ai/tree/StepUnits.java
```diff
@@ -0,0 +1,446 @@
...
+              return value + entry.getValue() * probability;
+            }
+          });
+    }
+    for (final Map.Entry<Unit, Double> entry : child.friendlyUnitsChances.entrySet()) {
+      enemyUnitsChances.compute(
+          entry.getKey(),
+          (key, value) -> {
+            if (value == null) {
+              return entry.getValue() * probability;
```
RoiEXLab, line 357: Same here

game-core/src/main/java/games/strategy/triplea/ai/tree/StepUnits.java
```diff
@@ -0,0 +1,446 @@
...
+        + ", enemyInjured="
+        + enemyWaitingToDieBits
+        + ", enemiesCount="
+        + enemyUnits.size()
+        + ", enemyHitPoints ="
+        + enemyHitPoints
+        + ", probability="
+        + probability
+        + '}';
+  }
```
RoiEXLab, line 445: String.format could help here as well

game-core/src/test/java/games/strategy/triplea/ai/BattleCalculatorComparisonTests.java
```diff
@@ -0,0 +1,405 @@
...
+      final Collection<Unit> attackingUnits = new ArrayList<>();
+      attackingUnits.addAll(INFANTRY.create(1, BRITISH));
+      attackingUnits.addAll(ARTILLERY.create(1, BRITISH));
+      final Collection<Unit> defendingUnits = new ArrayList<>();
+      defendingUnits.addAll(INFANTRY.create(2, GERMAN));
+      compareCalculators(attackingUnits, defendingUnits, FRANCE);
+      return;
+    }
+
+    if (true) {
```
RoiEXLab, line 171: if (true) is a smell.

If you simply want to re-use variable names, you can just leave the curly braces and remove the if statement, or even better create a separate helper method

game-core/src/test/java/games/strategy/triplea/ai/BattleCalculatorComparisonTests.java
```diff
@@ -0,0 +1,405 @@
...
+            TuvUtils.getCostsForTuv(defendingUnits.iterator().next().getOwner(), data));
+    output +=
+        " Avg Tuv Units Left: "
+            + df.format(tuvUnitsLeft.getFirst())
+            + ","
+            + df.format(tuvUnitsLeft.getSecond());
+
+    System.out.println(output);
+    System.out.println("Avg Attacking Units: " + result.getAverageAttackingUnitsRemaining());
+    System.out.println("Avg Defending Units: " + result.getAverageDefendingUnitsRemaining());
```
RoiEXLab, line 403: System.out is nice for debugging, but in the end you should use a logger instead

game-core/src/test/java/games/strategy/triplea/ai/tree/BattleStepTest.java
```diff
@@ -0,0 +1,940 @@
...
+    final Map<Integer, Integer> diceGrouped = getDiceGrouped(attackingOrderOfLoss, List.of());
+    final StepUnits attackingUnits =
+        new StepUnits(attackingOrderOfLoss, BRITISH, List.of(), GERMAN);
+    final BattleStep root = new BattleStep(attackingUnits, BRITISH, 0, createParameters());
+    final List<Double> hitProbability =
+        root.calculateHitProbabilities(BattleStep.RollData.of(2, diceGrouped, diceSides));
+
+    assertEquals(0.4166, hitProbability.get(0), 0.0001);
+    assertEquals(0.5, hitProbability.get(1), 0.0001);
+    assertEquals(0.0833, hitProbability.get(2), 0.0001);
```
RoiEXLab, line 148: I assume those are the probabilities for win, lose, draw?

If the List was a dedicated type instead, this would be clearer

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
```diff
@@ -0,0 +1,1216 @@
...
+        case SUB_DEFENDER:
+          return ATTACKER;
+        case ATTACKER:
+          return DEFENDER;
+        case DEFENDER:
+          return AA_ATTACKER;
+        default:
+          Preconditions.checkArgument(false, "A type has been forgotten in swapSides.");
+          return AA_ATTACKER;
+      }
```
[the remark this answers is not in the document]
trevan: Unfortunately, since the order of sub attacks is configurable, that way won't work.  Once I've finished getting subs to work, it won't be a simple switch statement.

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
trevan: The probability isn't part of the equality check.  So it allows summing up the probabilities if the other parts are equal.

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
```diff
@@ -0,0 +1,1216 @@
...
+  static class Parameters {
+    GameData data;
+    Territory location;
+    Collection<TerritoryEffect> territoryEffects;
+    // TODO: Determine amphibious battles
+    boolean isAmphibious = false;
+    Collection<Unit> amphibiousLandAttackers = List.of();
+  }
+
+  // Taken from MustFightBattle::updateOffensiveAaUnits
```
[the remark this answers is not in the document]
trevan: I'd like to as well.  How would you prefer?  Should I make the function in MustFightBattle static?  Or create a static class?  Or something else?

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
trevan: Good point.  This function has changed a lot as I've added/removed things to help with debugging so it used to have additional variables and logic but those have been removed and I can simplify it further.

game-core/src/main/java/games/strategy/triplea/ai/tree/StepUnits.java
trevan: It can occur and does fairly often.  It happens when two child branches have different units hit.   This is merging the branch results so one branch will have a unit that has a probability for a unit and the other won't.  enemyUnitsChances/friendlyUnitsChances only has entries if there is a chance the unit might survive.

game-core/src/test/java/games/strategy/triplea/ai/BattleCalculatorComparisonTests.java
trevan: I'd like to know where this entire class should go.  It isn't really a test but a way to do performance runs.  I think the helper function at the bottom (that does the comparisons of the calculators) needs to be somewhere so that in the future, it is easy to redo performance checks.

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
RoiEXLab: I assumed that already, thanks for clarifying that.
But my question still stands (even though I should've been more clear): Why use Map<StepUnit, StepUnit> over something like Set<StepUnit> or even Map<List<Unit>, Double>?

game-core/src/main/java/games/strategy/triplea/ai/tree/StepUnits.java
RoiEXLab: Ah, nevermind.
I thought you were updating the same map you were reading from, which isn't the case 👍

game-core/src/test/java/games/strategy/triplea/ai/tree/BattleStepTest.java
trevan: These are probabilities of hits.  The first line is 0 hits (out of 2 dice), the second is 1 hit, the third is 2 hits.  hitProbability can grow depending on the number of dice rolled.

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
trevan: Good point.

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
```diff
@@ -0,0 +1,1216 @@
...
+import org.triplea.java.collections.CollectionUtils;
+import org.triplea.java.collections.IntegerMap;
+import org.triplea.util.Triple;
+import org.triplea.util.Tuple;
+
+public class BattleStep {
+
+  // enable this to keep children in memory for analysis
+  // watch out for running out of memory, though
+  private static final boolean DEBUG = false;
```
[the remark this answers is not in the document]
trevan: While doing debugging of the code, I have forgotten to either allow the children to remain in memory (causing a restart of the debugging process) or forgotten to clear out the children (causing a OOM failure).  I'd prefer to leave this in somehow so that it is easy to turn on/off the children.  There are currently only two places to remember but in the future, it won't be easy to remember those places.

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
trevan: Agreed.  But this is in code copied from MustFightBattle and I wanted to indicate why that else statement was commented out.  Once the duplicate code from MustFightBattle has been put in a common place, this will no longer be needed.

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
trevan: I've changed it to Map<StepUnit, Double>

game-core/src/test/java/games/strategy/triplea/ai/BattleCalculatorComparisonTests.java
trevan: The `if(true)` was so I could turn on/off different comparison tests easily.

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
trevan: I actually like the compute syntax.  I can change it, though if you want.

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
trevan: I already have to loop through the array to calculate `diceAvailableGroupedByPower` so doing two things in the same loop vs two different loops is better in this situation.  This function is called a lot.

trevan:
Here's a high level overview:

BattleTreeCalculator implements the IBattleCalculator interface.  It takes in the battle parameters and creates a BattleStep.  This BattleStep is the root of a tree of all possible outcomes in the battle.

In BattleStep#calculateBattle, it will calculate what are all the possible dice rolls.  For example, if you have two units that succeed at 1/6, then it will calculate the probability of 0 hits, 1 hit, or 2 hits.  These probabilities are calculated using a binomial distribution function (https://en.wikipedia.org/wiki/Binomial_distribution).

It will then loop through those possible dice rolls and descend into a new BattleStep which has the results of the battle (0 hits would have no units killed, 1 hit would have 1 unit killed, and 2 hits would have 2 units killed) and is a new step in the battle (AA Attacker, AA Defender, Sub Attacker, Sub Defender, Attacker, Defender, and (soon) Bombardment).

This continues until it reaches the end of a battle (BattleStep#checkEndOfBattle and BattleStep#checkEndOfBranch).  It will store the result (win, tie, lose) and which units survived.  As it goes back up the tree, these results are multiplied by the probability of the branch.

trevan:
@DanVanAtta I think a "game-plan" for merging this would be to:

 * Figure out how the duplicate code should be handled.  I can start a PR on that and go with helper static classes.  Is that ok?
 * Figure out what to do with `BattleCalculatorComparisonTests`.  It is where I do profiling and printing out comparisons of the different calculators.  It has no asserts and shouldn't be a test.  I'd like to keep that ability to do profiling and calculator comparisons and have that committed.  But I don't know how that would best fit into the code base.  Any ideas?
 * Decide on whether the known battle issues need to be fixed before this can be merged.  These are the issues around bombardment and subs.

Once those are decided and done, I think it should be ready for merging.

DanVanAtta:
@trevan that sounds like a reasonable game plan. 


Moving code to utility classes as a first step sounds good. Here are some recent examples of that:
- https://github.com/triplea-game/triplea/blob/master/game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtyOrderOfLosses.java
- https://github.com/triplea-game/triplea/blob/master/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AvailableSupportCalculator.java
- https://github.com/triplea-game/triplea/blob/36b923652ddb65bcb5c579ee53f60de27261db9a/game-core/src/main/java/games/strategy/engine/lobby/client/ui/LobbyFrame.java#L119
- https://github.com/triplea-game/triplea/blob/36b923652ddb65bcb5c579ee53f60de27261db9a/game-core/src/main/java/games/strategy/engine/lobby/client/ui/action/MutePlayerAction.java#L19

When doing the move hopefully we can also create some better package grouping. You may have noticed that for example all battle classes are in a single mega package `triplea.delegate`. It'd be more ideal to have 3-12 classes per class, ideally with just one or two having public interfaces and the rest having mostly package protected interfaces to support the class that exposes a public API.

The above first two examples has an example of parameter object (you probably noticed that when using the casualty order of losses), that is a decent way to avoid the very long parameter lists and avoid the code climate warnings about too many parameters.

The last two examples are basically static methods where the variables have been made into class variables, a `@Builder` is used and we have this pattern:
```
UtilityClassWithSomeFunction.builder()
    .parameter(param1)
    .otherParameter(param2)
    .build()
    .staticMethodCallNowWithNoArgs();
```

RE: compatibility concerns to watch out for when moving classes

Pre-2.0 launch, which is pretty imminent, you are quite free to move most anything. Post 2.0, we'll want save-games to remain compatible, regrettably that means anything that is saved as part of a `GameData` object (ie: all game data fields and their children), are all serialized into a save game file and hence moving the packages of those files will cause incompatibility. It's an important project for us to migrate save-games to a different format so we are not coupled to class files, their locations and variables, but like many things it's yet to be done.

If we are moving only static methods around, happily there is no compatibility concern.

Long story shorter, re-arranging any of the existing code so that it can be re-used in this PR and cutting those as preliminary PRs is a good move. I'd suggest to go further and try to cut anything that is not strictly needed for the new AI  to be its own PR.

DanVanAtta:
Re: `BattleCalculatorComparisonTests`

The comparisons TBH seem to be mostly one-time. We should consider this further, my default reaction is YAGNI though, unless the tests have asserts then they should be dropped. Perhaps there can be some parts of it that we can transform into a long-lived test, let's see what we can do on this one (there is plenty to do in the meantime).


> Decide on whether the known battle issues need to be fixed before this can be merged. These are the issues around bombardment and subs.

As an experimental AI for now, I'd say no and we can iterate on it. IMO the most important items first is to get everything cleanly landed, then improve. Though this also depends no your willingness to follow-up. If the experimental AI will just be in the maintainers hands, then it should become complete before it is "handed over." If you're willing to iterate, then the 'beta' feature flag is very much there so we can merge more incrementally and you do not need to get everything done on a first PR.

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleTreeCalculator.java
```diff
@@ -0,0 +1,135 @@
...
+                        .defending(false)
+                        .build())
+                .amphibiousLandAttackers(
+                    amphibiousLandAttackers == null ? List.of() : amphibiousLandAttackers)
+                .battlesite(location)
+                .costs(TuvUtils.getCostsForTuv(attacker, data))
+                .data(data)
+                .build());
+
+    final List<Unit> defendingOrderOfLoss =
```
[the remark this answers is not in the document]
asvitkine: Since the block of code is very similar to the one above, I think you should make a helper function for it with parameters for the things that are different.

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleTreeResults.java
```diff
@@ -0,0 +1,97 @@
...
+      final IntegerMap<UnitType> attackerCostsForTuv,
+      final IntegerMap<UnitType> defenderCostsForTuv) {
+    double attackerTuv = 0;
+    double defenderTuv = 0;
+    int count = 0;
+    for (int i = 1; i <= 20; i++) {
+      count++;
+      attackerTuv +=
+          TuvUtils.getTuv(
+              getAverageAttackingUnitsRemaining((double) i / 20.0), attackerCostsForTuv);
```
asvitkine, line 67: Can you make a local variable for i / 20.0 with a name that documents what it is? Otherwise it's a bit opaque to figure out what's being passed.

game-core/src/test/java/games/strategy/triplea/ai/BattleCalculatorComparisonTests.java
```diff
@@ -0,0 +1,395 @@
...
+import games.strategy.triplea.xml.TestMapGameData;
+import java.text.DecimalFormat;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.triplea.java.collections.CollectionUtils;
+import org.triplea.util.Tuple;
+
+class BattleCalculatorComparisonTests {
```
asvitkine, line 32: Add a comment?

game-core/src/test/java/games/strategy/triplea/ai/BattleCalculatorComparisonTests.java
asvitkine: Does it make sense for these to be separate test methods? Then you can run individual ones from the IDE as needed.

game-core/src/test/java/games/strategy/triplea/ai/BattleCalculatorComparisonTests.java
```diff
@@ -0,0 +1,395 @@
...
+              .create(quantity, BRITAIN));
+      attackingUnits.addAll(
+          dataTww.getUnitTypeList().getUnitType("britishTacticalBomber").create(quantity, BRITAIN));
+      attackingUnits.addAll(
+          dataTww
+              .getUnitTypeList()
+              .getUnitType("britishAdvancedTacticalBomber")
+              .create(quantity, BRITAIN));
+      attackingUnits.addAll(
+          dataTww.getUnitTypeList().getUnitType("britishAntiAirGun").create(quantity, BRITAIN));
```
asvitkine, line 108: Perhaps you can make a helper for this logic that just gets passed the unit type, country and quantity? To avoid the repetitiveness and extra wrapping.

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
```diff
@@ -0,0 +1,1221 @@
...
+      getTargetGroupFightOutcomes(
+          outcomes,
+          groupsAndTargets,
+          groupsAndTargets.size() - 1,
+          aliveOrInjuredUnits.swapSides(),
+          aliveOrInjuredUnits.getType(),
+          returnFire,
+          true,
+          1.0);
+    } else {
```
asvitkine, line 358: If you invert the if and just return, then the code in the above block could just be moved out (no need for else).

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
```diff
@@ -0,0 +1,1221 @@
...
+          outcomes,
+          aaGroupsAndTargets,
+          aaGroupsAndTargets.size() - 1,
+          aliveOrInjuredUnits.swapSides(),
+          aliveOrInjuredUnits.getType(),
+          MustFightBattle.ReturnFire.ALL,
+          allowMultipleHitsPerUnit,
+          1.0);
+    } else {
+      return List.of(new StepUnits(aliveOrInjuredUnits.swapSides(), 1.0));
```
asvitkine, line 431: Same comment here.

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
```diff
@@ -0,0 +1,1221 @@
...
+      } else {
+        currentUnits.killFriendly(targetUnit);
+      }
+      if (isSuicideOnHit) {
+        currentUnits.hitEnemy(firingGroup.get(totalHits - 1));
+      }
+    }
+
+    final double hitProbability = 1.0 - totalProbability;
+    if (groupsAndTargetsIndex > 0) {
```
asvitkine, line 530: This logic seems very similar to line 496. Can you make a helper function with whatever params that differ?

game-core/src/main/java/games/strategy/triplea/ai/tree/BattleStep.java
```diff
@@ -0,0 +1,1221 @@
...
+    }
+  }
+
+  /**
+   * Figures out all the probabilities of rolling dice with a specific set of powers and amounts.
+   *
+   * @param rollData The details about the dice available to roll
+   * @return List of possible probabilities from rolling 0 dice to rolling rollData.totalRolls})
+   */
+  List<Double> calculateHitProbabilities(final RollData rollData) {
```
asvitkine, line 851: I think this should be a helper class to simplify the state.

Basically, you would create the object, which will do the initialization in the ctor (all the code of the function) and then call calculateHitProbabilities() on it, which would recurse as necessary. Then, most of the params of calculateHitProbabilitiesWorker() could be removed and could instead be variables on the object.

It will also allow you to make BattleStep shorter by removing this complexity from it.

trevan:
@DanVanAtta I created https://github.com/triplea-game/triplea/pull/6520 that has the new utility classes that the battletree will use to reduce duplicate code.

trevan:
@DanVanAtta should I close this PR for now?

PR #6517, opened 2020-05-23
Add a separate setting for AI combat step pause duration.

asvitkine:
Add a separate setting for AI combat step pause duration.

Motivation:
  - During combat, it can be useful to have a bit more time
    to see the dice rolls.
  - On the other hand, while AI is moving units, not as much
    time is needed.

So this change introduces a new setting with a higher default and
also reduces the existing setting (since I suspect the higher
default was to accomodate the time needed to view combat dice
rolls).

I believe these new default settings will actually make games
against AIs faster as there's much more time spent by AIs moving
units than doing dice rolls.

Also removes some unnecessary pause() calls from DoNothingAI,
as there's no reason to pause when the AI isn't doing anything.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
Manual testing.

### After

Screen Shot 2020-05-23 at 9 30 45 AM

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ai/weak/DoesNothingAi.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/weak/DoesNothingAi.java
@@ -41,23 +41,18 @@ public class DoesNothingAi extends AbstractAi {
       new WeakAi(this.getName())
           .purchase(purchaseForBid, pusToSpend, purchaseDelegate, data, player);
     }
-    pause();
   }
 
   @Override
   protected void tech(
-      final ITechDelegate techDelegate, final GameData data, final GamePlayer player) {
-    pause();
-  }
+      final ITechDelegate techDelegate, final GameData data, final GamePlayer player) {}
 
   @Override
   protected void move(
       final boolean nonCombat,
       final IMoveDelegate moveDel,
       final GameData data,
-      final GamePlayer player) {
-    pause();
-  }
+      final GamePlayer player) {}
 
   @Override
   protected void place(
@@ -69,13 +64,10 @@ public class DoesNothingAi extends AbstractAi {
     if (!player.getUnitCollection().isEmpty()) {
       new WeakAi(this.getName()).place(placeForBid, placeDelegate, data, player);
     }
-    pause();
   }
 
   @Override
-  public void politicalActions() {
-    pause();
-  }
+  public void politicalActions() {}
 
   @Override
   protected void endTurn(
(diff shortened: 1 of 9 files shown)


game-core/src/main/java/games/strategy/triplea/ai/weak/DoesNothingAi.java
```diff
@@ -41,23 +41,18 @@ protected void purchase(
       new WeakAi(this.getName())
           .purchase(purchaseForBid, pusToSpend, purchaseDelegate, data, player);
     }
-    pause();
   }
 
   @Override
   protected void tech(
-      final ITechDelegate techDelegate, final GameData data, final GamePlayer player) {
-    pause();
```
DanVanAtta, line 50: The AI never researches tech? I'm just guessing, is there another reason to remove the tech delegate pause here? I'm also curious about the politics action pause removal as well

asvitkine: This is the donorhingai

DanVanAtta: Hehe, took a moment to decode "donorhingai", kept reading that as "donor" "hingai"

Cool, thanks for pointing that out, makes sense :+1:

FrostionAAA:
@asvitkine I like this change a lot. I have long thought about asking for some more pause during the AI's combat. So this is great! But there are two things I think could (maybe be improved).

**1:**
I like that one can see, both on map and in right side panel, the combat rounds progress. But it feels and looks best during the last of the combat engagement of the AI player, as the last place to have a combat going on will be followed by a small “pause” after combat (I think because the camera is fixed on the last combat territory and the AI is doing its NCM calculations.) This small pause lets the observer see the result of the battle. Right now it seems that the conclusions of the not-last battles are finished so abruptly that the observer cannot really see the outcome.

Maybe the whole AI player battle sequence could be even better followed, if the observer got to have a split second pause after all the battles, so that the observer got to see the territory turn to the winner, see last man/units standing or see the attacker retreat.

**2:**
As nearly no other text displayed is bold/fat type, it looks a bit out of place that the text info accompanying the dice pictures are bold/fat. Maybe it would look OK if only the “Total Hits: X” headline was bold, but the other text above the dice was plain. Or maybe if all the text was plain, just like most other text in the tabs? (Maybe you could make and show some examples pictures?)

Issue #6484, opened 2020-05-17
Hosting a game in pre-release lobby NPE's on lobby watcher


DanVanAtta:
### How can the problem be recreated?

- launch local lobby
- host game, observer it can NPE in 'InGameLobbyWatcherWrapper#executeConnectivityCheck'

### What is an expected fix?

- connectivity check should be executed successfully

### (Optional) Additional information

Unless someone volunteers, I'll take this one up. I think a good solution would be to remove the explicit connectivity check endpoint and bundle it with the 'post-game' endpoint. Post-game would need to return a composite object with the connectivity result.

prastle:
Explain in English since I head no problems hosting ?

DanVanAtta:
@prastle there might have been a race condition to trigger this. I was able to repro this with a local lobby consistently and did assume the problem would carry forward to prerelease. One key difference between the two environments (prerelease and local) is latency. Regardless this problem would be fixed with #6519 as we've changed the design of when/how the lobby verifies internet connectivity to hosts (which auto-shuts down the host if it is not available on internet).

PR #6519, opened 2020-05-23
Do reverse-connection check of new game hosts when posting games

DanVanAtta:
This update combines the connectivity check endpoint with the post-game
endpoint. This avoids any timing issues and seeks to simplify
the game posting flow. Instead of a client doing two round trips
to the server to first post a game and then ask for a connectivity
check, it is done as one now.

For implementation updates, we update the response from the game-posting
request to include both game-id (nullable) and the result of the
connectivity check. After posting a game we now should check
the connectivity result and if okay, then proceed, otherwise
execute the previous behavior when the connectivity check failed
(which is to show the "this is how to set up port forwarding" message
and shutdown).

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix: Addresses: https://github.com/triplea-game/triplea/issues/6484 
[] Other:   

## Testing
Verified could connect to a local lobby and host a game.

## Additional Review Notes

This gets us super close to a 2.0 RC. I'd like to have this get in and add a single or double retry to websocket connections and then consider that next build a RC.


game-core/src/main/java/games/strategy/engine/framework/startup/WatcherThreadMessaging.java
```diff
@@ -12,14 +12,31 @@
...
-  void serverNotAvailableHandler(String message);
+  String COMPUTER_NOT_REACHABLE_ERROR_MESSAGE =
+      "Your computer is not reachable from the internet.\n"
+          + "Please make sure your Firewall allows incoming connections (hosting) "
+          + "for TripleA.\n"
+          + "(The firewall exception must be updated every time a new version of "
+          + "TripleA comes out.)\n"
+          + "And that your Router is configured to send TCP traffic the correct port "
+          + " to your local ip address.\n"
+          + "See 'How To Host...' in the help menu, at the top of the lobby "
```
RoiEXLab, line 23: What is the "How to host" button currently doing?
We might want to mention the possibility of using a relay service in the future

DanVanAtta: Good question, this is something to update. We no longer have a 'how to host' menu option. The reason it was removed is that the game was linking to many forum posts, some/many of which were getting out of date. Instead of having that type of situation, instead all help content should be linked to the user guide and then we maintain just one link to the user guide.

DanVanAtta: If the 'how to host' page is updated to mention a relay service, then we'll have that done.

PR #6520, opened 2020-05-24
Shared code for Battle Tree Estimator

trevan:
This moves all the code that Battle Tree uses into utility classes.  I've tested these changes with the battle tree branch to make sure that the new utility classes work for both the MustFightBattle and the battle tree.

There shouldn't be any functional changes.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
I opened up Global 1940 and Total World War in TripleA and run a few scenarios in the battle calculator.  I tested aa gun fights, sub fights, bombardments, and undefended transports.  The results were similar.


trevan:
Oh, I forgot I was going to do the builder pattern.  I'll push a change to use that.  I'll also work on the code climate issues.

trevan:
The 3 complexity issues from codeclimate are in code that I barely touched.  Do you want me to update that code as well?
I'm not sure how to fix the duplicate code issues from codeclimate without making the code harder to read.  Thoughts?

Also, I don't know how to run codeclimate locally or in my own repo.  Is there a way to do that?

game-core/src/main/java/games/strategy/triplea/delegate/battle/subs/Subs.java
```diff
@@ -0,0 +1,91 @@
...
+    if (defendingSubsSneakAttack(gameData) && !defendingSubsFireWithAllDefenders) {
+      return FireOrder.DEF_BEFORE_REGULAR;
+    }
+    return FireOrder.DEF_WITH_REGULAR;
+  }
+
+  public boolean defenderSubsFireFirst(
+      final Collection<Unit> attackingUnits,
+      final Collection<Unit> defendingUnits,
+      final GameData gameData) {
```
asvitkine, line 40: Would it make sense to make GameData a ctor param so all these methods don't have to take it as a param?

game-core/src/main/java/games/strategy/triplea/delegate/battle/grouptarget/FiringGroup.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import lombok.Value;
+import org.triplea.java.collections.CollectionUtils;
+
+@Value(staticConstructor = "of")
+public class FiringGroup {
+  Collection<Unit> firingGroup;
```
asvitkine, line 16: Should these be private & final?

game-core/src/main/java/games/strategy/triplea/delegate/battle/grouptarget/FiringGroup.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import lombok.Value;
+import org.triplea.java.collections.CollectionUtils;
+
+@Value(staticConstructor = "of")
+public class FiringGroup {
```
asvitkine, line 15: I think it would be helpful to have a short comment explaining what this represents exactly. How are the units grouped?

game-core/src/main/java/games/strategy/triplea/delegate/battle/grouptarget/Aa.java
```diff
@@ -0,0 +1,70 @@
...
+    final Set<UnitType> validTargetTypes =
+        UnitAttachment.get(firingGroup.iterator().next().getType()).getTargetsAa(gameData);
+    final Set<UnitType> airborneTypesTargeted =
+        defending
+            ? TechAbilityAttachment.getAirborneTargettedByAa(hitPlayer, gameData).get(aaType)
+            : new HashSet<>();
+    return CollectionUtils.getMatches(
+        attackableUnits,
+        Matches.unitIsOfTypes(validTargetTypes)
+            .or(Matches.unitIsAirborne().and(Matches.unitIsOfTypes(airborneTypesTargeted))));
```
asvitkine, line 68: This last part will evaluate false if defending is false.

Perhaps split it up to make the logic clearer?

e.g.
```
Predicate<Unit> predicate = Matches.unitIsOfTypes(validTargetTypes);
if (defending) {
  final Set<UnitType> airborneTypesTargeted = ...;
  predicate = predicate.or(Matches.unitIsAirborne().and(Matches.unitIsOfTypes(airborneTypesTargeted))));
}
```
return CollectionUtils.getMatches(attackableUnits, predicate);

game-core/src/main/java/games/strategy/triplea/delegate/battle/grouptarget/Aa.java
```diff
@@ -0,0 +1,70 @@
...
+  private final boolean defending;
+  private @NonNull final GameData gameData;
+
+  public List<FiringGroup> getFiringGroups() {
+    final List<FiringGroup> aaGroupsAndTargets = new ArrayList<>();
+
+    for (final String aaType : aaTypes) {
+      final Collection<Unit> aaTypeUnits =
+          CollectionUtils.getMatches(aaUnits, Matches.unitIsAaOfTypeAa(aaType));
+      getGroupsAndTargets(aaGroupsAndTargets, aaType, aaTypeUnits);
```
asvitkine, line 35: This function should have a different name since it has a set effect and not just a getter.

Maybe updateGroupsAndTargets()?

game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtySelector.java
```diff
@@ -346,6 +341,29 @@ private static void killAmphibiousFirst(final List<Unit> killed, final Collectio
     return Tuple.of(defaultCasualtySelection, sorted);
   }
 
+  public static List<Unit> getSortedUnits(
+      final Collection<Unit> targetsToPickFrom,
```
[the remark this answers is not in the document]
asvitkine: Does it make sense to add a helper class to hold these parameters? Is the same list of params passed around to many other functions too?

game-core/src/main/java/games/strategy/triplea/delegate/battle/grouptarget/FiringGroup.java
trevan: @Value makes them private & final.

game-core/src/main/java/games/strategy/triplea/delegate/battle/subs/Subs.java
trevan: I had left Subs as a utility class so all of the functions are static.  I could instantiate Subs once during MustFightBattle's ctor with gameData and then use the instance each time.  Would that be ok?

game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtySelector.java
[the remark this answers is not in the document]
trevan: I can't find any other place besides here and the battletree calculator that would use this same list.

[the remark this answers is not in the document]
trevan: I'm not even sure if I should have created this function since it just creates a parameter object.  And I believe the parameter class was created so the function wouldn't have so many arguments.  So making a new function to create the parameter object kind of defeats the purpose of the parameter object.

game-core/src/main/java/games/strategy/triplea/delegate/battle/subs/Subs.java
asvitkine: Ah, I hadn't seen that before. It's very confusing to not have the static keyword on a method and have something else cause it to be static. Is there a consensus for the project to use that? If not, I'd prefer the methods be marked as static explicitly.

game-core/src/main/java/games/strategy/triplea/delegate/battle/grouptarget/FiringGroup.java
asvitkine: Ah, same thing! Very confusing when you're not familiar with it.

game-core/src/main/java/games/strategy/triplea/delegate/battle/subs/Subs.java
trevan: I went through the code and found a few other classes that had the lombok utility class annotation and they still had the static modifier.  So, I'll add the static/final/etc modifiers to be explicit about it.

game-core/src/main/java/games/strategy/triplea/delegate/battle/firinggroups/FiringGroup.java
```diff
@@ -0,0 +1,80 @@
...
+ * Breaks list of units into groups of non suicide on hit units and each type of suicide on hit
+ * units since each type of suicide on hit units need to roll separately to know which ones get
+ * hits.
+ */
+@Value(staticConstructor = "of")
+public class FiringGroup {
+  private final @NonNull Collection<Unit> firingGroup;
+  private final @NonNull Collection<Unit> validTargets;
+  private final boolean suicide;
+  // type is currently only used by Aa firing groups
```
asvitkine, line 25: Should it be an enum? If not, can the comment document possible values at least?

game-core/src/main/java/games/strategy/triplea/delegate/battle/subs/Subs.java
```diff
@@ -0,0 +1,91 @@
...
+    return returnFireAgainstAttackingSubs(attackingUnits, defendingUnits, gameData)
+            == ReturnFire.ALL
+        && returnFireAgainstDefendingSubs(attackingUnits, defendingUnits, gameData)
+            == ReturnFire.NONE;
+  }
+
+  public static ReturnFire returnFireAgainstAttackingSubs(
+      final Collection<Unit> attackingUnits,
+      final Collection<Unit> defendingUnits,
+      final GameData gameData) {
```
asvitkine, line 50: What do you think of making GameData the first param in these cases? It seems a bit more natural to me.

game-core/src/main/java/games/strategy/triplea/delegate/battle/FireAa.java
```diff
@@ -92,84 +89,77 @@
...
+      final IExecutable rollDice =
+          new IExecutable() {
+            private static final long serialVersionUID = 6435935558879109347L;
+
+            @Override
+            public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
+              validTargets.removeAll(casualtiesSoFar);
+              if (!validTargets.isEmpty()) {
+                dice =
+                    DiceRoll.rollAa(
```
asvitkine, line 117: Can this take FiringGroup to reduce number of params? (since validTargets, firingGroup, etc can just be taken from that param)

game-core/src/main/java/games/strategy/triplea/delegate/battle/FireAa.java
```diff
@@ -92,84 +89,77 @@
...
-                          allFriendlyUnitsAliveOrWaitingToDie,
-                          bridge,
-                          battleSite,
-                          defending);
-                  if (!headless) {
-                    SoundUtils.playFireBattleAa(firingPlayer, aaType, dice.getHits() > 0, bridge);
-                  }
+    for (final FiringGroup groupAndTarget : groupsAndTargets) {
+      final Collection<Unit> validTargets = groupAndTarget.getValidTargets();
+      final Collection<Unit> firingGroup = groupAndTarget.getFiringGroup();
```
asvitkine, line 105: It's a bit confusing to have a FiringGroup have a member called firingGroup and a getter for it.

Could it be named differently? e.g. getFiringUnits()?

game-core/src/main/java/games/strategy/triplea/delegate/battle/FireAa.java
```diff
@@ -92,84 +89,77 @@
...
-                      selectCasualties(validTargets, firingGroup, bridge, aaType);
-                  battle.markDamaged(details.getDamaged(), bridge);
-                  casualties = details;
-                  casualtiesSoFar.addAll(details.getKilled());
-                }
+            @Override
+            public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
+              if (!validTargets.isEmpty()) {
+                final CasualtyDetails details =
+                    selectCasualties(validTargets, firingGroup, bridge, aaType);
```
asvitkine, line 139: This could also take FiringGroup directly. Then it just need two params (FiringGroup and bridge).

game-core/src/main/java/games/strategy/triplea/delegate/battle/end/NoUnitsWithRolls.java
```diff
@@ -0,0 +1,69 @@
...
+  private final boolean hasRetreatTerritories;
+  private @NonNull final Collection<Unit> attackingUnits;
+  private @NonNull final Collection<Unit> defendingUnits;
+  private @NonNull final Territory battleSite;
+
+  public Collection<Unit> check() {
+    if (isRetreatPossible() || attackingUnits.isEmpty() || defendingUnits.isEmpty()) {
+      return List.of();
+    }
+    final Collection<Unit> units;
```
asvitkine, line 25: This seems a bit strange. Can we simply not eliminate isAttacker boolean and have the caller pass units/enemyUnits in different args when it wants it one way or the other?

e.g. when using this builder:

```
            .attackingUnits(isAttacker ? attackingUnits : defendingUnits)
            .defendingUnits(isAttacker ? defendingUnits : attackingUnits)
```

Then, the code in the builder can be simplified.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1360,10 +1316,16 @@ private void fireNavalBombardment(final IDelegateBridge bridge) {
       }
       final List<Unit> allEnemyUnitsAliveOrWaitingToDie = new ArrayList<>(defendingUnits);
       allEnemyUnitsAliveOrWaitingToDie.addAll(defendingWaitingToDie);
+      final List<FiringGroup> firingGroups =
+          Bombard.builder()
+              .firingUnits(bombard)
+              .attackableUnits(attacked)
+              .defending(false)
+              .build()
+              .getFiringGroups();
```
asvitkine, line 1325: Is there a place where build() isn't immediately followed by getFiringGroups()? If not, should we just make build() return them? Or make buildFiringGroups()?

game-core/src/main/java/games/strategy/triplea/delegate/battle/firinggroups/FiringGroup.java
```diff
@@ -0,0 +1,80 @@
...
+    // add the non suicide hit group last
+    buildNonSuicideFiringGroup(getValidTargets, type, nonSuicideFiringGroup, result);
+    return result;
+  }
+
+  private static void buildSuicideFiringGroups(
+      final Function<Collection<Unit>, Collection<Unit>> getValidTargets,
+      final String type,
+      final Map<UnitType, Collection<Unit>> suicideUnitsByType,
+      final List<FiringGroup> result) {
```
DanVanAtta, line 58: I notice that `List<FiringGroup>` is an out parameter. Would you object to converting this to be a return value? 

https://github.com/triplea-game/triplea/wiki/Java-Code-Conventions#do-not-mutate-parameters-avoid-out-and-avoid-inout-parameters-5489

The updated signature of this function would then be:
```
 private static Collection<FiringGroup> buildSuicideFiringGroups(
      final Function<Collection<Unit>, Collection<Unit>> getValidTargets,
      final String type,
      final Map<UnitType, Collection<Unit>> suicideUnitsByType)
```

(side-note, the collection type would cascade back, if we need indexed access then `List` would be an appropriate return type)

game-core/src/main/java/games/strategy/triplea/delegate/battle/firinggroups/FiringGroup.java
```diff
@@ -0,0 +1,80 @@
...
+  // type is currently only used by Aa firing groups
+  private final @NonNull String type;
+
+  static List<FiringGroup> newFiringUnitGroups(
+      final Collection<Unit> units,
+      final Function<Collection<Unit>, Collection<Unit>> getValidTargets,
+      final String type) {
+
+    // split the units by suicideOnHit and unitType (if suicideOnHit)
+    final Map<UnitType, Collection<Unit>> suicideUnitsByType = new HashMap<>();
```
DanVanAtta, line 34: Any thoughts to using guava multimap? It's effectively the same data structure as a `Map<X, Collection<Y>>` and simplifies adding elements and retrieving.

That would look like this:
```
    final Multimap<UnitType, Unit> suicideUnitsByType = HashMultimap.create();
    final Collection<Unit> nonSuicideFiringGroup = new ArrayList<>();
    for (final Unit unit : units) {
      final UnitType unitType = unit.getType();
      if (UnitAttachment.get(unitType).getIsSuicideOnHit()) {
        suicideUnitsByType.put(unitType, unit);  // << note the computeIfAbsent is removed
      } else {
        nonSuicideFiringGroup.add(unit);
      }
    }
```
Then when accessing later:
```
       for (final Collection<Unit> suicideFiringGroup : suicideUnitsByType.asMap().values()) {
```

game-core/src/main/java/games/strategy/triplea/delegate/battle/firinggroups/FiringGroup.java
```diff
@@ -0,0 +1,80 @@
...
+      final UnitType unitType = unit.getType();
+      if (UnitAttachment.get(unitType).getIsSuicideOnHit()) {
+        suicideUnitsByType.computeIfAbsent(unitType, key -> new ArrayList<>()).add(unit);
+      } else {
+        nonSuicideFiringGroup.add(unit);
+      }
+    }
+
+    final List<FiringGroup> result = new ArrayList<>();
+    // add all of the suicide hit groups first
```
DanVanAtta, line 46: Why must the suicide hit groups be computed first? I'm curious why there is an ordering dependency.

game-core/src/main/java/games/strategy/triplea/delegate/battle/FireAa.java
```diff
@@ -92,84 +89,77 @@
...
-            CollectionUtils.getMatches(
-                attackableUnits,
-                Matches.unitIsOfTypes(validTargetTypes)
-                    .or(
-                        Matches.unitIsAirborne()
-                            .and(Matches.unitIsOfTypes(airborneTypesTargeted))));
-        final IExecutable rollDice =
-            new IExecutable() {
-              private static final long serialVersionUID = 6435935558879109347L;
+    final List<FiringGroup> groupsAndTargets =
```
DanVanAtta, line 92: What do you think about making the `isSuicide` first dependency explicit and a condition established by the caller rather than as part of an (unenforced) contract?

The problem I can see here potentially is that it's not clear that `isSuicide` units would be first, you have to dig into the `Aa.builder()`  to know that which makes it less obvious what the code is doing from a caller perspective.

EG:
```
    // Loop through each type of AA and break into firing groups based on suicideOnHit
    final List<FiringGroup> groupsAndTargets =
        Aa.builder() //
            .aaUnits(firingUnits) //
            .aaTypes(aaTypes) //
            .hitPlayer(hitPlayer) //
            .attackableUnits(attackableUnits) //
            .defending(defending) //
            .gameData(bridge.getData()) //
            .build() //
            .getFiringGroups() //
            .stream()
            .sorted(Comparator.comparing(FiringGroup::isSuicide))
```
(You'll need to double check please that sorts isSuicide first, otherwise a 'reverse' or negation needs to be thrown in)

An alternative would be to document the contract of the return value:
```
Aa.builder()
   :
   :
   .getFiringGroupsWithSuicideFirst();
```

Personally I would let unit testing drive the decision. If it's easer to just unit test `Aa#getFiringGroups` without the sort, then I'd sort here. If it's easy to test, then I'd perhaps update the return method name.

game-core/src/main/java/games/strategy/triplea/delegate/battle/end/NoMoreUnits.java
```diff
@@ -0,0 +1,77 @@
...
+import games.strategy.triplea.delegate.Matches;
+import java.util.Collection;
+import java.util.List;
+import java.util.function.Predicate;
+import lombok.Builder;
+import lombok.NonNull;
+import org.triplea.java.collections.CollectionUtils;
+
+@Builder
+public class NoMoreUnits {
```
DanVanAtta, line 17: This class could potentially benefit from a Javadoc. Is this a computation class to determine when there is no more units who won? It takes some reverse engineering to see what this does and why it would be used. A good javadoc would helpful give an overview of purpose and functionality to then guide a reader to more quickly understand the implementation. Please also consider adding javadocs to any other classes.

game-core/src/main/java/games/strategy/triplea/delegate/battle/end/NoMoreUnits.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public enum Winner {
+    ATTACKER,
+    DEFENDER,
+    DEFENDER_TURN_1,
+    UNDETERMINED
```
DanVanAtta, line 23: `UNDETERMINED` was not super obvious to me at first. I wondered if perhaps it meant an error happened, or if the battle could not determine a winner because the battle is "weird" (like air vs sub that cannot be hit). Would `NONE` maybe be a more clear name here with the same meaning? If not, are there other cases where `UNDETERMINED` is still a better fit?

game-core/src/main/java/games/strategy/triplea/delegate/battle/end/NoMoreUnits.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  private @NonNull final GameData gameData;
+  private @NonNull final GamePlayer attacker;
+  private @NonNull final Collection<Unit> attackingUnits;
+  private @NonNull final Collection<Unit> defendingUnits;
+  private @NonNull final Territory battleSite;
+  private final int round;
```
DanVanAtta, line 31: To force this value to be set with a builder, it would need to be an `Integer`, then you could use `@Nonnull` with it and enforce that it'll be specified when constructing otherwise this can silently default to zero.  Is it intentional to allow this to have a default?

game-core/src/main/java/games/strategy/triplea/delegate/battle/subs/Subs.java
trevan: I had noticed that many of the existing methods have GameData as the last parameter.  So I was following that convention.  But I can move it to the first parameter.  I don't have an opinion either way.

game-core/src/main/java/games/strategy/triplea/delegate/battle/end/NoMoreUnits.java
```diff
@@ -0,0 +1,77 @@
...
+  private @NonNull final Runnable checkUndefendedTransports;
+  private @NonNull final Runnable checkForUnitsThatCanRollLeft;
+
+  public Winner check() {
+    final Predicate<Unit> notInfrastructure = Matches.unitIsNotInfrastructure();
+    if (CollectionUtils.getMatches(attackingUnits, notInfrastructure).isEmpty()) {
```
DanVanAtta, line 37: You can use an `anyMatch` here to make this more performant, ie:
```
  public Winner check() {
    if(attackingUnits.stream().anyMatch(Matches.unitIsNotInfrastructure())) {
      return noAttackingUnits();
    } else if(defendingUnits.stream().anyMatch(Matches.unitIsNotInfrastructure())) {
      return noDefendingUnits();
    }
    return Winner.UNDETERMINED;
  }
```

With the update above iteration stops when you find the first unit that matches rather than iterating through all units to then check if the resulting collection is empty.

game-core/src/main/java/games/strategy/triplea/delegate/battle/firinggroups/FiringGroup.java
trevan: As far as I can tell, this comes out of the game xml.  It comes out of UnitAttachment#getAllOfTypeAas.  So I think the possible values are user defined.

game-core/src/main/java/games/strategy/triplea/delegate/battle/end/NoMoreUnits.java
```diff
@@ -0,0 +1,77 @@
...
+              .and(Matches.isUnitAllied(attacker, gameData));
+      final List<Unit> alliedTransports =
+          CollectionUtils.getMatches(battleSite.getUnits(), matchAllied);
+      // If no transports, just end the battle
+      if (alliedTransports.isEmpty()) {
+        result = Winner.DEFENDER;
+      } else if (round <= 1) {
+        result = Winner.DEFENDER_TURN_1;
+      } else {
+        result = Winner.DEFENDER;
```
DanVanAtta, line 63: Is this correct to return `Winner.DEFENDER` here?

If so, then the matches for transport could be skipped and you'd only need to check if we are in round 1. It seems like this should be `UNDETERMINED`.

game-core/src/main/java/games/strategy/triplea/delegate/battle/FireAa.java
trevan: I don't feel comfortable making a change to DiceRoll with this push.  DiceRoll#rollAa is also called from DiceRoll#rollSbrOrFlyOverAa so I'd have to change that as well.  And then I think having DiceRoll#rollAa using FiringGroup while DiceRoll#rollDice doesn't would feel weird.

DanVanAtta:
PS: I hit my review limit for today, eyes are glazing over. Regrettably I mostly reviewed details not high level concerns. I will try to give this another look tomorrow or the day after. In the meantime if the comments from @asvitkine are all addressed and he has finished his review, and if a decent number of tests are added, I'd be happy to see this PR being merged without another look. Otherwise I'll try to do another review shortly in the coming days.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: @DanVanAtta had left a comment in the Battle Tree Estimator pull request that said that the code base had this pattern for utility classes.  See his comment at https://github.com/triplea-game/triplea/pull/6501#issuecomment-632436501.

trevan:
> Added some review feedback. The breakup is a good direction as we get to more, easier to understand modules -> bite-sized is good.
> 
> A good chunk of this logic is not testable. What do you think about attempting to unit test some of these functions @trevan ? Refactoring code without unit tests is also called re-writing code, sadly it means we have lots of potential for introducing defects. Unit testing is most useful for ensuring good code design and reducing defects density. I think those would be excellent benefits at this point that would very well justify the return on investment and help guarantee there are no bugs in the resulting code. What are your thoughts?

I had actually started building tests to make sure the different types of battles were handled and then I came across a bunch of acceptance tests (such as PactOfSteel2Test, WW2V3Year42Test, and WW2V3Year41Test).  After looking through those, it appeared that all the code I as changing was covered fairly well.  But I could do unit tests, the only problem is that some of the new utility classes need specific parameters in GameData and I haven't yet found some examples of mocking/stubbing GameData to create those specific parameters.

game-core/src/main/java/games/strategy/triplea/delegate/battle/end/NoMoreUnits.java
trevan: That is what the original code had.  Are you sure that the transport check is no longer needed?  I don't see why.

game-core/src/main/java/games/strategy/triplea/delegate/battle/end/NoMoreUnits.java
trevan: I was just copying the original code.  I'll change it.

game-core/src/main/java/games/strategy/triplea/delegate/battle/end/NoMoreUnits.java
trevan: No, I wasn't aware that was what builder would do.  I'll fix that.

game-core/src/main/java/games/strategy/triplea/delegate/battle/end/NoMoreUnits.java
trevan: `NONE` is fine.  Just needed a flag to indicate that the winner isn't known yet.

game-core/src/main/java/games/strategy/triplea/delegate/battle/firinggroups/FiringGroup.java
trevan: No idea.  It is what the original code did.  That comment is from the original code.

game-core/src/main/java/games/strategy/triplea/delegate/battle/FireAa.java
trevan: I'm not sure of the benefit of doing this.  The only caller in the current code that uses this apparently needs suicide first and the new caller in battle tree calculator would want to follow the same order.  So I'm not sure any caller would want the order to be different.

game-core/src/main/java/games/strategy/triplea/delegate/battle/firinggroups/Aa.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Predicate;
+import lombok.Builder;
+import lombok.NonNull;
+import org.triplea.java.collections.CollectionUtils;
+
+@Builder
+public class Aa {
```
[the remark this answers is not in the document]
trevan: Sure.  I thought since it was inside of the "firinggroups" package, that would indicate that it was a firing group.  Calling it `AaFiringGroup` felt a little redundant.  But I can change its name.

game-core/src/main/java/games/strategy/triplea/delegate/battle/firinggroups/FiringGroup.java
trevan: The original code used the Map.  If you want me to switch it to guava multimap, I have no problem with that.

game-core/src/main/java/games/strategy/triplea/delegate/battle/end/NoUnitsWithRolls.java
trevan: `isAttacker` is still needed for the call to `Matches#unitIsSupporterOrHasCombatAbility`.  But I like the idea of flipping the units in the call to the builder instead of inside of `check`.

game-core/src/main/java/games/strategy/triplea/delegate/battle/end/NoMoreUnits.java
trevan: oh, now I see what you mean.  You're right.  It does look like it might be a bug but the original code is 
```
                if (alliedTransports.isEmpty()) {
                  endBattle(bridge);
                  defenderWins(bridge);
                } else if (round <= 1) {
                  attackingUnits =
                      CollectionUtils.getMatches(
                          battleSite.getUnits(), Matches.unitIsOwnedBy(attacker));
                } else {
                  endBattle(bridge);
                  defenderWins(bridge);
                }
```
I went through the history and that `else` statement has been around for a long time.  It was added in 2009 at https://github.com/triplea-game/triplea/commit/aa693839b8115e71a5b50ade06783a9657e10cdf.

I can simplify the if statements.

game-core/src/main/java/games/strategy/triplea/delegate/battle/grouptarget/FiringGroup.java
DanVanAtta: IDEA even generates unnecessary 'private & final' modifiers for `@Value`. IMO this is one where it's okay to leave off, some familiarity with lombok is perhaps required. Making methods explicit on `@UtilityClass` IMO is okay (more helpful as you might not always often see the annotation on the same screen as the method)

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
DanVanAtta: Indeed. The 'swingAction' variant I thought at the time was pretty nice. Thinking back to it, I just don't know which of these two is better:
```
UtilClass.methodCall(Parameters.builder()
    .param1(...)
    .param2(...)
    .build());
```

or:
```
UtilClass.builder()
    .param1(...)
    .param2(...)
    .build()
    .voidMethodCall();
```

This might be worth a code convention issue to discuss when/where to use either or if to prefer one. I can see the first being more useful if there are multiple methods that could be invoked, the latter is a more typical parameter object pattern. I'd guess that the latter is more idiomatic.

game-core/src/main/java/games/strategy/triplea/delegate/battle/firinggroups/Aa.java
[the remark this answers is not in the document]
DanVanAtta: Since it's public it can be used outside of the package, makes it less clear at the point of use. Though, I'd general agree that reducing the redundant nouns/specifiers is a good thing though.

game-core/src/main/java/games/strategy/triplea/delegate/battle/firinggroups/FiringGroup.java
DanVanAtta: Is this mostly just a straight code move? If we're doing more than simply moving code, then treating it as new code is relatively appropriate (it's arguably in effect re-written/new).

game-core/src/main/java/games/strategy/triplea/delegate/battle/FireAa.java
DanVanAtta: That would be argument then for updating the name of the method to be explicit about the ordering. Otherwise it's not clear there is a contract and potentially a dependency on the ordering, essentially an undocumented behavior.

game-core/src/main/java/games/strategy/triplea/delegate/battle/end/NoMoreUnits.java
DanVanAtta: On second look, the equivalent if is something like:
```
!alliedTransports.isEmpty() && round <=1
```

Perhaps we need the transport checking still. Though it's still pretty curious all in all.. I can't say I fully grok why transports are being checked.

DanVanAtta:
>  But I could do unit tests, the only problem is that some of the new utility classes need specific parameters in GameData and I haven't yet found some examples of mocking/stubbing GameData to create those specific parameters.

DanVanAtta:
>  But I could do unit tests, the only problem is that some of the new utility classes need specific parameters in GameData and I haven't yet found some examples of mocking/stubbing GameData to create those specific parameters.

The acceptance tests are a bit of the integration-trap testing anti-pattern where that was the majority of testing done. Part of the reason is that mocking `GameData` is often relatively painful. The long term hope is to decouple map parsing from the acceptance tests, build out more unit tests, and decouple game-data from being passed as reference to virtually everything. Eventually as well we'd like for game-data persistence to be text based, not serialization, which would want/require game data to not be a property of everything..

The efforts are appreciated, as a group the test coverage is far better than where it was, but still a lot of cleaning & testing and paying down of the technical debt need to be done.

game-core/src/main/java/games/strategy/triplea/delegate/battle/firinggroups/FiringGroup.java
trevan: It is mostly a straight code move with changes added that are requested by either you or @asvitkine.  I've also made changes to resolve the code climate complexities.  It is probably about half rewritten at this point.

game-core/src/main/java/games/strategy/triplea/delegate/battle/end/NoMoreUnits.java
trevan: I've gone through the git history and I think the alliedTransport check is a bug.  The code used to also do the undefended transport check and when that was refactored, this alliedTransport part forgotten.

I came up with a situation where this would happen in a game and played it out and got a weird state in the game.  The situation is where your ally has your land units in it and is next to a country that has a factory.  On the enemies turn, a ship is produced in the same sea zone.  When the production is after combat, the two ships (enemy ship and allied transport) stay around.  On your turn, you move a ship into the zone and initiate a combat.  If your ship dies in the first round, you get an option to retreat because this code added your transported land units to the attacking units and they aren't dead.  If you retreat, your transported land units are moved by themselves.  I believe that is a bug.

I think the alliedTransport check should be removed.

game-core/src/main/java/games/strategy/triplea/delegate/battle/firinggroups/FiringGroup.java
DanVanAtta: I also wondered if an enum made sense. We'd need a mapping of string values to enum values so the two could be mapped (and leave allowance for an 'other' category). Ultimately though I think that is an update too far, we should be focused on just moving code for this iteration. The clean-ups and changes are probably even a step too far for this iteration (but those are already started, so we should finish; to be clear, simply moving code is probably a good first PR, then changing, cleaning up is a good follow up. We do run into a conflict that is endemic in TripleA where you simply need far too many iterations to make all changes bite-sized. Strategic layering of commits can help)

game-core/src/main/java/games/strategy/triplea/delegate/battle/end/NoMoreUnits.java
DanVanAtta: Oddly that is good news, fixes a problem and is much simpler. It might be worthwhile to hold off on that specific update and follow up on it in a new a dedicated PR to that change. It is up to you (keeping PRs atomic, focused is generally a very good thing. Makes for easier review and if rolled back then not as much is caught up in the rollback. Generally changes happen faster with many small PRs compared to fewer large ones).

trevan:
I'm still working on this.  I'm building out the unit tests but the FiringGroups is taking some time since it needs suicide units and there really aren't any in the existing test gamedatas.  So I'm trying to build a new gamedata that will have more suicide units and aa units for testing it.

game-core/src/main/java/games/strategy/triplea/delegate/battle/firinggroups/FiringGroup.java
trevan: I don't think it can be enum.  This is a way to group the aa units in a map.  A map could have one aa group or it could have 30 aa groups.  This name is also presented to the user in the battle dialog when each group fires.

What could be done is to grab the group name in FireAa after the FiringGroups have been established instead of here during the creation of the FiringGroups.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: Should I change anything?  Use a Parameters builder?

game-core/src/main/java/games/strategy/triplea/delegate/battle/end/NoMoreUnits.java
trevan: I made the change.  The battletree calculator will be simpler with this fix.

game-core/src/main/java/games/strategy/triplea/delegate/battle/FireAa.java
trevan: After building unit tests, I discovered that suicide units aren't always at the front.  It first groups the firingUnits by what they can/can not target (think subs and airplanes) and then it splits those groups by isSuicide.  See RegularFiringGroupTest#getFiringGroupsWithNonTargetableUnitsAndSuicideUnits.

game-core/src/main/java/games/strategy/triplea/delegate/battle/firing/group/TargetGroup.java
```diff
@@ -1,4 +1,4 @@
-package games.strategy.triplea.delegate.battle;
+package games.strategy.triplea.delegate.battle.firing.group;
```
trevan, line 1: This class was only being used by RegularFiringGroup so I moved it into the firing.group package and made the methods be package visible instead of public.

game-core/src/test/java/games/strategy/triplea/delegate/battle/end/DetectWinnerTest.java
```diff
@@ -0,0 +1,116 @@
...
+      "Verify attacker wins if there are no defenders left "
+          + "with TRANSPORT_CASUALTIES_RESTRICTED == false")
+  void noDefendersLeftAndTransportCasualtiesUnrestricted() {
+    final List<Unit> attackers = GameDataTestUtil.armour(GLOBAL_1940_GAME_DATA).create(1, BRITISH);
+    final List<Unit> defenders = List.of();
+
+    final Runnable removeUndefendedTransports = mock(Runnable.class);
+    final Runnable removeUnitsWithNoRollsLeft = mock(Runnable.class);
+
+    GLOBAL_1940_GAME_DATA.getProperties().set(TRANSPORT_CASUALTIES_RESTRICTED, false);
```
trevan, line 98: Is there a better way to change the gameData properties for a test?

game-core/src/test/resources/firing_groups_test.xml
```diff
@@ -0,0 +1,580 @@
+<?xml version="1.0"?>
```
trevan, line 1: Is it ok to create a new map file for tests?  Does this map look ok?

trevan:
After this is finished, I think I'd like to take another stab at the subs class.  I had originally only pulled out what was needed for the battle tree calculator but I'd like to unify the subs code in MustFightBattle#determineStepStrings with the new Subs utility class.

I'd also like to try and move the FiringGroups to be more central to MustFightBattle.  It currently grabs the aa units and uses them for some checks and then later uses them to build the firinggroups.  I think it should build the aa firing groups (and others) at the beginning of a round and use those through out the class (and maybe other classes, such as DiceRoll).

trevan:
@DanVanAtta and @asvitkine should I close this and split it up?  There's basically 3 main parts:
* Code to detect fight order, retreat conditions, and "dead" units
* Code to detect win/lose condition
* Code to create firing groups

I've been looking at the MustFightBattle#determineStepStrings and it has the first.  So if I create a PR just to refactor that function, I should have what I need for the first part.
I could then create a smaller PR just for the win/lose condition.
And I could create another PR that uses the firing groups in a more complete way.  Right now, this PR was just trying to move the firing group creations to a helper class but I think MustFightBattle could use firing groups more extensively.

Thoughts?

asvitkine:
+1 for making smaller PRs. Thanks!

Issue #6521, opened 2020-05-24
Concurrent Modification Exception in 19485


beelee1:
### How can the problem be recreated?

Map is Global 40 House Rules with Canada.

When Bombing raid is undertaken this error is thrown.

![Screenshot from 2020-05-24 02-48-05](https://user-images.githubusercontent.com/24382682/82749036-11453b00-9d6c-11ea-9f02-f2b87d6232f5.png)

### What is an expected fix?

Not sure but it works in 13066

### Which Engine Version are you using? 

19485

### (Optional) Additional information
here is save with 130666 where it works. Max scramble to sz 110 and sz 111. Pick Factory as one target and airfield as next. Note it will say "units may die" or something because it is a A0 D0 unit but it is made for SBR only. Ignore the warning and continue. 

/home/greg/Desktop/13066 vs 19485.zip

here is pre 19485 where the error is thrown:

/home/greg/Desktop/19485vs13066 SBR Error.zip

My suspicion is something must have changed where the "units will die" prompt that was able to be ignored is now no longer able too ? idk ? I thought they would just die then so maybe something different. Anyway it basically breaks the map so...

beelee1:
[19485vs13066 SBR Error.zip](https://github.com/triplea-game/triplea/files/4673163/19485vs13066.SBR.Error.zip)

[13066vs19485.zip](https://github.com/triplea-game/triplea/files/4673166/13066vs19485.zip)

RoiEXLab:
Thanks to @asvitkine this should be fixed in the latest pre-release that should be ready in the next couple of minutes

beelee1:
Thanks Guys :)

PR #6522, opened 2020-05-24
Fix concurrent modification exception by iterating list copy.

asvitkine:
Also uses List.copyOf() in similar cases and removes some other
list copies where the return value is never modified.

Fixes: https://github.com/triplea-game/triplea/issues/6521

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix:  https://github.com/triplea-game/triplea/issues/6521
[] Other:   

## Testing

Manual testing with save game from https://github.com/triplea-game/triplea/issues/6521.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/AbstractPlaceDelegate.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/AbstractPlaceDelegate.java
@@ -812,7 +812,7 @@ public abstract class AbstractPlaceDelegate extends BaseTripleADelegate
       return "Too many constructions in " + to.getName();
     }
     final List<Territory> capitalsListOwned =
-        new ArrayList<>(TerritoryAttachment.getAllCurrentlyOwnedCapitals(player, getData()));
+        TerritoryAttachment.getAllCurrentlyOwnedCapitals(player, getData());
     if (!capitalsListOwned.contains(to) && isPlacementInCapitalRestricted(player)) {
       return "Cannot place these units outside of the capital";
     }
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleTracker.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleTracker.java
@@ -185,7 +185,7 @@ public class BattleTracker implements Serializable {
   }
 
   void clearFinishedBattles(final IDelegateBridge bridge) {
-    for (final IBattle battle : new ArrayList<>(pendingBattles)) {
+    for (final IBattle battle : List.copyOf(pendingBattles)) {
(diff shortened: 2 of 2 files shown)


DanVanAtta:
How confident can we be that we have fixed all of these?

asvitkine:
> How confident can we be that we have fixed all of these?

We can never be confident, but if they happen we'll hear about them since it will cause an error. I suspect there won't be much more since people (including myself) have been playing prereleases pretty successfully without running into problems.

Sure, new maps can uncover code paths that are rare / previously not exercised, but that's already the case for the existing code anyway as we can see from the bug reports from warcraft map.

DanVanAtta:
I suspect we're over the hump. Getting to 100% confidence would be a goal, it's very unfortunate to have an engine crash because of this. At some point it would have been worth biting the bullet and keeping mutable collections, at this point I'd guess not. I think if we were to do this over, I would have kept all mutable collections and only converted the ones where there was test coverage. A lack of test coverage remains one of the largest problems TripleA (the fact that most code is effectively untestable is a related and equally significant problem).

tvleavitt:
Have we thought about generating a map or maps which systematically
exercise(s) every available documented feature?

The way I've always done Q/A testing is to imagine the most obnoxious and
oblivious end user possible, the guy who won't read directions, who copies
and pastes multi-paragraph Unicode into input fields, types 1 or 792 when
asked his age, etc. and see what happens. Then modify my code, design and
error messaging appropriately.

When I ran my web hosting start up, we gave our users a very crippled form
of Perl that had no loops (among other limitations designed to limit load
and the effects of bad programming). We called it the "Form Processor" and
assumed it would be a slightly more flexible version of the various
"Form2Email" cgi scripts then circulating (in 1994).

They instead built discussion forum managers, shopping carts and photo
galleries with it (among other things). We just shook our heads in
amazement.

Thomas Leavitt

DanVanAtta:
I like the anecdote, sounds right out of daily WTF. The issue is really existing code. It's a balance of mutability and modifiability and converting collection types to be unmodifiable. We've code that at worst passes a direct reference to an underlying collection. Changes to that returned value changes the original object. Second to that, we copy such datastructures so that changes to the collection do not change the original object (but in some cases this is a noticeable performance penalty). Last, we return an modifiable copy of a collection. The problem with the latter is that there are many code paths from resulting return values and if any such paths modify the collection, then we'll hit an error. If we go from case (1), to case (3) and see an error, we are seeing a potential area where an underlying object is unexpectedly modified, so removing these by converting to (3) is good. On the downside, it presumes we can know and verify all such code paths for it to be 'safe'.

PR #6526, opened 2020-05-25
Websocket async connection single retry

DanVanAtta:
## Overview
Adds a single retry attempt to establish websocket connection.  With a local lobby have observed connection failures that succeed on a subsequent retry.

## Commits

commit 76bccdeefac7683d9d44d6eca58777ee33463970

    Delegate error messaging to error handling.
    
    Do not call logger when we fail to connect, allow 'errorHandler'
    to control all error messaging and logging in such cases.

commit f5f5c8d7f42f7a7554e3979a3e525be79f34b9a0

    Merge ping-send starter with connect method

commit 7578ec2b46da77de2f2271dc7f453d8a3468e70c

    Add websocket connection retry
    
    - Single retry connection attempt
    - Fixed backoff of 1 second, long enough to give a decent
      pause between attempts but not so long a user will notice
      too excessive of a delay.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:

### DIFF
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/WebSocketConnection.java
@@ -20,6 +20,7 @@ import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.java.Log;
+import org.triplea.java.Interruptibles;
 import org.triplea.java.timer.ScheduledTimer;
 import org.triplea.java.timer.Timers;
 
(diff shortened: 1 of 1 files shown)


RoiEXLab:
Actually I'd go even further and suggest retrying with exponential backoff

DanVanAtta:
Two retries could be worthwhile. We lack data though on their efficacy and notably that we'd expect a 2nd or 3rd retry to have a better chance than the first retry. Anecdotally a single retry can help.

PR #6529, opened 2020-05-26
Improve some battle history logging code.

asvitkine:
Improve some battle history logging code.

Removes code duplication and unnecessary complexity and moves unrelated logic (marking attacking transports, to a different function).

Also removes an unnecessary list copy in Route.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[X] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

Looked at history after attacking. Tried with two defending countries.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
@@ -13,6 +13,7 @@ import games.strategy.engine.data.UnitType;
 import games.strategy.engine.data.changefactory.ChangeFactory;
 import games.strategy.engine.delegate.IDelegateBridge;
 import games.strategy.engine.display.IDisplay;
+import games.strategy.engine.history.IDelegateHistoryWriter;
 import games.strategy.triplea.Properties;
 import games.strategy.triplea.UnitUtils;
 import games.strategy.triplea.attachments.TechAbilityAttachment;
@@ -34,7 +35,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
@@ -601,6 +601,7 @@ public class MustFightBattle extends DependentBattle implements BattleStepString
     }
     bridge.getHistoryWriter().startEvent("Battle in " + battleSite, battleSite);
     removeAirNoLongerInTerritory();
+    markAttackingTransports(bridge);
     writeUnitsToHistory(bridge);
     if (CollectionUtils.getMatches(attackingUnits, Matches.unitIsNotInfrastructure()).isEmpty()) {
       endBattle(bridge);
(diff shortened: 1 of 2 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -668,92 +669,71 @@ private void removeAirNoLongerInTerritory() {
...
-      final GamePlayer current = defendersIter.next();
-      final String delim;
-      if (defendersIter.hasNext()) {
-        delim = "; ";
-      } else {
-        delim = "";
+    return allies;
+  }
+
+  private void addPlayerCombatHistoryText(
```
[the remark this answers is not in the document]
asvitkine: This is OK because the following two existing findings are fixed:

```
Method writeUnitsToHistory has a Cognitive Complexity of 26 (exceeds 7 allowed). Consider refactoring.
Method writeUnitsToHistory has 78 lines of code (exceeds 30 allowed). Consider refactoring.
```

https://codeclimate.com/github/triplea-game/triplea/pull/6529

DanVanAtta:
@asvitkine have you done any smoke testing into this?

asvitkine:
> @asvitkine have you done any smoke testing into this?

Sorry, forgot to write fill in the Testing section. Did manual testing with looking at history after an attack. The code is also not so large so can also just be read for the old vs. new versions.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -668,92 +669,71 @@ private void removeAirNoLongerInTerritory() {
...
+    final Collection<GamePlayer> attackers = findAllies(playersWithUnits, attacker);
+    addPlayerCombatHistoryText(attackers, attackingUnits, true, bridge.getHistoryWriter());
+    final Collection<GamePlayer> defenders = findAllies(playersWithUnits, defender);
+    addPlayerCombatHistoryText(defenders, defendingUnits, false, bridge.getHistoryWriter());
+  }
+
+  private static Collection<GamePlayer> findAllies(
+      final Collection<GamePlayer> candidatePlayers, final GamePlayer player) {
+    final var relationshipTracker = player.getData().getRelationshipTracker();
+    final Collection<GamePlayer> allies = new ArrayList<>();
```
DanVanAtta, line 705: FWIW, this block could be re-written as follows:
```
return candidatePlayers.stream()
    .filter(current -> current.equals(player) || relationshipTracker.isAllied(player, current)
```

I somewhat feel like relationship tracker probably should own this logic though. EG:
```
relationshipTracker.getAllies(player, candidatePlayers);
```

DanVanAtta: There's certainly more places where we can simplify or 'fix' responsibilities and change logic ownership, but I recognize we are mostly moving code here.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -601,6 +601,7 @@ public void fight(final IDelegateBridge bridge) {
     }
     bridge.getHistoryWriter().startEvent("Battle in " + battleSite, battleSite);
     removeAirNoLongerInTerritory();
+    markAttackingTransports(bridge);
```
DanVanAtta, line 604: This could potentially be slightly simpler written as:
```
    if (!headless) {
      markAttackingTransports(bridge);
      writeUnitsToHistory(bridge);
    }
```

Essentially we're bringing the no-op method check to the point of the caller instead of in the method. This way if you are looking at the higher level flow, you can more easily follow the branching logic. As-is, with a series of void methods, to some extent they almost may as well be inlined with demarcating comments, a reader has to jump through each one to understand what is happening. Hence, bringing the logic flow out of the lower level methods allows a reader to see the code flow at a higher level without jumping to each method.

PR #6533, opened 2020-05-26
Game chooser UI window simplify

DanVanAtta:
A series of simplifications to improve the UI code behind the game-chooser window

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
- manually verified game chooser UI looks the same before and after.

### DIFF
--- a/swing-lib/src/main/java/org/triplea/swing/JLabelBuilder.java
+++ b/swing-lib/src/main/java/org/triplea/swing/JLabelBuilder.java
@@ -33,8 +33,13 @@ public class JLabelBuilder {
   private Integer borderSize;
   private int biggerFont;
   private Color foregroundColor;
+  private boolean bold;
 
   public JLabelBuilder() {}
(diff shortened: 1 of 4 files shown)


swing-lib/src/main/java/org/triplea/swing/JLabelBuilder.java
```diff
@@ -33,9 +33,14 @@
   private Integer borderSize;
   private int biggerFont;
   private Color foregroundColor;
+  private boolean bold;
 
   public JLabelBuilder() {}
 
+  public JLabelBuilder(final String text) {
+    this.text = text;
+  }
```
RoiEXLab, line 42: I think this constructor is a bit inconsistent, it somewhat breaks the builder pattern IMO

Issue #6525, opened 2020-05-25
2.0.19500: GameParser#newGameParseException:119 - GameParseException


tripleabuilderbot:
## User Description
When loading Feudal Japan map with the Prerelease.
Error does not happen with 1.9.

## Log Message
Could not parse:file:/Users/shadowknight/triplea/downloadedMaps/feudal_japan-master/map/games/FeudalJapan.xml

## TripleA Version
2.0.19500

## Java Version
11.0.6

## Operating System
Mac OS X

## Exception 
games.strategy.engine.data.GameParseException: map name: 'file:/Users/shadowknight/triplea/downloadedMaps/feudal_japan-master/map/games/FeudalJapan.xml', game name: 'Feudal Japan', Missing property definition for option 'costPU' in attachment 'politicalActionAttachmentAsaiAsakuraWar'

## Stack Trace
```
games.strategy.engine.data.GameParseException: map name: 'file:/Users/shadowknight/triplea/downloadedMaps/feudal_japan-master/map/games/FeudalJapan.xml', game name: 'Feudal Japan', Missing property definition for option 'costPU' in attachment 'politicalActionAttachmentAsaiAsakuraWar'
	at games.strategy.engine.data.GameParser.newGameParseException(GameParser.java:119)
	at games.strategy.engine.data.GameParser.newGameParseException(GameParser.java:112)
	at games.strategy.engine.data.GameParser.lambda$setOptions$12(GameParser.java:1379)
	at java.base/java.util.Optional.orElseThrow(Optional.java:408)
	at games.strategy.engine.data.GameParser.setOptions(GameParser.java:1377)
	at games.strategy.engine.data.GameParser.parseAttachment(GameParser.java:1323)
	at games.strategy.engine.data.GameParser.parseAttachments(GameParser.java:1237)
	at games.strategy.engine.data.GameParser.parseMapDetails(GameParser.java:213)
	at games.strategy.engine.data.GameParser.parse(GameParser.java:107)
	at games.strategy.engine.data.GameParser.parse(GameParser.java:99)
	at games.strategy.engine.data.GameParser.parse(GameParser.java:85)
	at games.strategy.engine.framework.ui.DefaultGameChooserEntry.getCompleteGameData(DefaultGameChooserEntry.java:63)
	at games.strategy.engine.framework.ui.DefaultGameChooserEntry.fullyParseGameData(DefaultGameChooserEntry.java:47)
[...]

PR #6534, opened 2020-05-27
Smoke-testing, add test that all current maps can be parsed

DanVanAtta:
- Add list of all current map XMLs (map-list.txt)
- Updates smoke-testing/build.gradle to download all XMLs from map-list.txt
- Adds a test to smoke-testing that loads each XML and verifies
  each one can be parsed.

Goal: this test is to ensure we do not have any map parsing regressions
and that we can continue parsing any existing XMLs. This will help
ensure that the engine is not updated to break parsing and secondarily
that maps available for download will not fail on parsing.

Note: we can potentially see a problem if a map becomes broken in repository.
If that happens then we can comment out the map XML in map-list.txt until
it is fixed. On the other hand if we have a non-backward compatible update
we can do similar while updating maps (this highlights the problem we have
with non-backward compatible map updates).

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[x] Other: increased test-coverage  

## Additional Review Notes

- This check will help ensure we proactively detect map parsing regressions (https://github.com/triplea-game/triplea/issues/6525)
- We can also be sure that all maps are still working correctly in prep for 2.0 release


RoiEXLab:
Also travis failed

RoiEXLab:
Good point, I'll try to do it tomorrow if I don't forget, it's already getting pretty late here

PR #6544, opened 2020-05-29
Update 0 vs 0 stalemate check to take into account dice rolls

DanVanAtta:
Instead of only looking at just attack power for determining
if both defense and offense have zero power, also look at the
number of dice rolls available and consider the battle
a stalemate if both sides either have no rolls or no power.
This is done simply by multiplying the power and rolls together,
if either is zero the product will be zero.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
- none

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
@@ -885,6 +885,11 @@ public class DiceRoll implements Externalizable {
     @Nonnull Integer totalPower;
     @Nonnull Integer totalRolls;
 
+    /** Returns the product of power and dice rolls. */
+    public int getEffectivePower() {
+      return totalPower * totalRolls;
+    }
+
     public TotalPowerAndTotalRolls subtractPower(final int powerToSubtract) {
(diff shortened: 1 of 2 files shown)


Cernelius:
Just to make sure, how about if a side has two units, one of which has positive strength but 0 rolls and the other one has 0 strength and positive rolls? Will this be seen as if the side is powerless? Asking because, under the definition of "both sides either have no rolls or no power", even if we consider that being an inclusive "or" (meaning "both sides either have no rolls or no power or neither") this case would be literally false. It would be true under the definition "every unit of either side has no rolls or no strength or neither" (I know the "or" is always inclusive in programming, but it is not so in English), which is how I believe it should be.

Is this looking at the factual situation (after applying supports, marine bonuses...)?

RoiEXLab:
Ah sorry @Cernelius, didn't see your comment there.
Please keep on discussing

DanVanAtta:
@Cernelius If i understand your question correctly, then yes. With this update we take into account number of rolls, 0 power is defined as either rolling at 0 (zero strength) or having zero rolls (or both).

PR #6546, opened 2020-05-30
Don't trim issue body

RoiEXLab:
This prevents the report client from trimming away the important stack trace information whenever we get something like a StackOverflow as an error like in #6545

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

None.

### DIFF
--- a/http-clients/src/main/java/org/triplea/http/client/error/report/ErrorReportRequest.java
+++ b/http-clients/src/main/java/org/triplea/http/client/error/report/ErrorReportRequest.java
@@ -1,9 +1,11 @@
 package org.triplea.http.client.error.report;
 
 import com.google.common.base.Ascii;
+import javax.annotation.Nullable;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.EqualsAndHashCode;
+import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.ToString;
 import org.triplea.http.client.github.issues.GithubIssueClient;
@@ -15,16 +17,10 @@ import org.triplea.http.client.github.issues.GithubIssueClient;
 @NoArgsConstructor
 @AllArgsConstructor
 public class ErrorReportRequest {
-  private String title;
-  private String body;
+  @Nullable private String title;
+  @Getter @Nullable private String body;
 
   public String getTitle() {
     return title == null ? null : Ascii.truncate(title, GithubIssueClient.TITLE_MAX_LENGTH, "...");
-  }
-
-  public String getBody() {
-    return body == null
-        ? null
-        : Ascii.truncate(body, GithubIssueClient.REPORT_BODY_MAX_LENGTH, "...");
   }
 }


RoiEXLab:
Also Note to myself:
We should increase the default Stack size in the installer

RoiEXLab:
@djensen47 The build consistently fails trying to reach https://www.axisandallies.org/forums
Any chance the travis ci servers are blocked from the forum due to high connection frequency?

DanVanAtta:
Having no limit seems risky. As a general rule it's important to place an upper bound on anything, but particularly on anything that is transmitted over network. 

Should we instead increase the limit size to the actual max that github issue's would allow?

Is a stack trace that is extremely long and probably redundant likely to be that useful? Would it be worthwhile to instead do some redundant line de-duping? 

As a third alternative, could we print the top 'n' of each caused by and get the same information while still keeping a total upper bound on the size of the stack trace?

RoiEXLab:
@DanVanAtta Valid point, and I'm not very happy with it either.
The problem is that it is hard to tell which part of the stacktrace is actually important.
For a Stackoverflow the important part is on the bottom of the stack (how did we end up here?) which is the part being cropped out currently.
For a NumberFormatException the important bit is the middle of the stacktrace, almost the top (we want to find out where the non-int value is actually coming from).
For a NPE the most important bit is most likely the top of the stacktrace, except when it isn't because null is being passed from class to class without any precondition.

TL;DR I don't think there's a good way to crop a stacktrace in a way that works for every use case. We  could potentially add a special case for StackOverflowErrors where we simply crop from the bottom but I'm not sure where the limits should be and how to handle potentially nested errors.

That's why I think this is the most pragmatic solution. Thoughts?

DanVanAtta:
The options I see (perhaps there are more), are:
A) we determine how to create a gist via API. If we can do that, we can upload stack traces that are over length X to the gist.
B) determine the real issue upload limit. Then truncate the "lower middle" of a stack trace to ensure we have at least the first N lines and the last M lines such that N+M = max.
C) same as B, but do the truncation to the middle of each stack trace in each caused-by. We can view the stack trace as just one giant list, or as a series of lists with different caused by exceptions. So we can 'middle-truncate' just once or to each caused by. This options strikes me as a bit complex.
D) Simply increase the upload limit to the max that github allows and do truncation at that limit. I'd opt for option B over this, though there is little harm in increasing the existing arbitrary limit.

DanVanAtta:
@RoiEXLab it looks like the max comment body length is: 65536

tvleavitt:
Re: option A.

"The Gist API provides up to one megabyte of content for each file in the
gist."

https://developer.github.com/v3/gists/

Thomas Leavitt's

DanVanAtta:
In the meantime, until we land on a more permanent fix, I opened: https://github.com/triplea-game/triplea/pull/6678, to truncate issue body at the github max length. Personally I'm leaning towards the option of posting long stack traces to a gist.

PR #6548, opened 2020-05-30
Fix link checker

RoiEXLab:
This update fixes the link checker script to work with websites that don't offer HTTP 2.

Going to self merge so #6546 and  #6547 can be built.

@djensen47 Update to my question in https://github.com/triplea-game/triplea/pull/6546#issuecomment-636390516 :
Looks like our script had a bug. Any chance there was a downgrade to HTTP/1.1 from HTTP/2 recently for https://www.axisandallies.org/forums?
Because it sure looks like it, otherwise I don't have an explanation for this.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

Manually verified the regular expressions work as expected

### DIFF
--- a/.travis/stages/verify/check-links-and-yaml-syntax/check-links
+++ b/.travis/stages/verify/check-links-and-yaml-syntax/check-links
@@ -15,7 +15,7 @@ fi
 
 function checkUri() {
   local -r uri="$1"
-  curl -s --head "$uri" | head -n 1  | grep -qE "HTTP/(2|1.1) (200( OK)?|302 Found)" \
+  curl -s --head "$uri" | head -n 1  | grep -qE "HTTP/(2 200|1.1 (200 OK|302 Found))" \
     || reportBrokenUri "$uri"
 }
 
--- a/.travis/stages/verify/check-links-and-yaml-syntax/check-links
+++ b/.travis/stages/verify/check-links-and-yaml-syntax/check-links
@@ -6,6 +6,7 @@ URI_FILE=$(find . -type f -path "*/src/main/java/*" -name "UrlConstants.java")
 green="\e[32m"
 end="\e[0m"
 FAILURE_PREFIX="Failed with Code: "
+IGNORE_LIST=()
 
 if [ -z "$URI_FILE" ]; then
   echo "Error could not find file 'UrlConstants.java'"
@@ -19,21 +20,28 @@ function checkUri() {
   local output
   if ! output="$FAILURE_PREFIX$(curl -Lsf -o /dev/null -w "%{http_code}" "$uri")"; then
     reportBrokenUri "$uri" "$output"
+    FAILURE=1
   fi
 }
 
 function reportBrokenUri() {
-  FAILURE=1 
   local -r red="\e[38;5;196m"
   local -r blue="\e[38;5;75m"
   local -r error_reason=${2//${FAILURE_PREFIX}000/Connection Failure!}
-  echo -e "Error: ${red}${1}${end} ${error_reason}${blue}${end}"
+  echo -e "Error: ${red}${1}${end} ${blue}${error_reason}${end}"
+}
+
+function checkIsOnIgnoreList() {
+  for ignored_uri in "${IGNORE_LIST[@]}"; do
+    [[ "$1" = "$ignored_uri" ]] && return 0
+  done
+  return 1
 }
 
 FAILURE=0
 
 while read -r uri; do
-  checkUri "$uri";
+  checkIsOnIgnoreList "$uri" || checkUri "$uri"
 done < <(sed 's/^.*"\(.*\)".*$/\1/' "$URI_FILE" | grep "http")
 
 if [ "$FAILURE" == 0 ]; then
--- a/.travis/stages/verify/check-links-and-yaml-syntax/check-links
+++ b/.travis/stages/verify/check-links-and-yaml-syntax/check-links
@@ -5,6 +5,7 @@ set -eu
 URI_FILE=$(find . -type f -path "*/src/main/java/*" -name "UrlConstants.java")
 green="\e[32m"
 end="\e[0m"
+FAILURE_PREFIX="Failed with Code: "
 
 if [ -z "$URI_FILE" ]; then
   echo "Error could not find file 'UrlConstants.java'"
@@ -15,17 +16,18 @@ fi
 
 function checkUri() {
   local -r uri="$1"
-  curl -s --head "$uri" | head -n 1  | grep -qE "HTTP/(2 200|1.1 (200 OK|302 Found))" \
-    || reportBrokenUri "$uri"
+  if ! output="$FAILURE_PREFIX$(curl -Lsf -o /dev/null -w "%{http_code}" "$uri")"; then
+    reportBrokenUri "$uri" "$output"
+  fi
 }
 
 function reportBrokenUri() {
   FAILURE=1 
   local -r red="\e[38;5;196m"
   local -r blue="\e[38;5;75m"
-  echo -en "Error: ${red}$uri${end} ${blue}"
-  (curl -s --head "$uri" || echo "Failed to connect!") | head -n 1 | sed -E 's!HTTP/(2|1.1)!!'
-  echo -en "${end}"
+  echo -en "Error: ${red}$1${end} ${blue}"
+  echo -en "${2//${FAILURE_PREFIX}000/Connection Failure!}"
+  echo -e "${end}"
 }
 
 FAILURE=0
--- a/.travis/stages/verify/check-links-and-yaml-syntax/check-links
+++ b/.travis/stages/verify/check-links-and-yaml-syntax/check-links
@@ -24,7 +24,7 @@ function reportBrokenUri() {
   local -r red="\e[38;5;196m"
   local -r blue="\e[38;5;75m"
   echo -en "Error: ${red}$uri${end} ${blue}"
-  curl -s --head "$uri" | head -n 1 | sed -E 's!HTTP/(2|1.1)!!'
+  (curl -s --head "$uri" || echo "Failed to connect!") | head -n 1 | sed -E 's!HTTP/(2|1.1)!!'
   echo -en "${end}"
 }
 
--- a/.travis/stages/verify/check-links-and-yaml-syntax/check-links
+++ b/.travis/stages/verify/check-links-and-yaml-syntax/check-links
@@ -20,6 +20,8 @@ function checkUri() {
   if ! code=$(curl -Lsf -o /dev/null -w "%{http_code}" "$uri"); then
     reportBrokenUri "$uri" "$code"
     FAILURE=1
+  else
+    echo "Successfully checked '$uri', returned $code"
   fi
 }
 
@@ -32,7 +34,11 @@ function reportBrokenUri() {
 
 function checkIsOnIgnoreList() {
   for ignored_uri in "${IGNORE_LIST[@]}"; do
-    [[ "$1" = "$ignored_uri" ]] && return 0
+    if [[ "$1" = "$ignored_uri" ]]
+    then
+      echo "Ignoring '$1'"
+      return 0
+    fi
   done
   return 1
 }
--- a/.travis/stages/verify/check-links-and-yaml-syntax/check-links
+++ b/.travis/stages/verify/check-links-and-yaml-syntax/check-links
@@ -15,7 +15,7 @@ fi
 
 function checkUri() {
   local -r uri="$1"
-  curl -s --head "$uri" | head -n 1  | grep -qE  "HTTP/(2|1.1) (200|302) (OK|Found)" \
+  curl -s --head "$uri" | head -n 1  | grep -qE "HTTP/(2|1.1) (200 OK|302 Found)" \
     || reportBrokenUri "$uri"
 }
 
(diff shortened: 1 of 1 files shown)


RoiEXLab:
Looks like there are multiple issues after all.
The script still can't reach https://www.axisandallies.org/forums for an unknown reason

DanVanAtta:
Check links is a brittle part of the build. This is something we should make more sturdy.

There could be a need to just do more debugging, perhaps print out the output from curl in its entirety, maybe there is an error message that is being suppressed.

RoiEXLab:
@DanVanAtta FYI there are 2 notable changes here:
1. It now works when the request only returns a HTTP 1.1 response, grep and sed
2. In case curl fails establishing a connection entirely (which is the case here), we now print a "failed to connect" error string instead of just nothing.

I think this should help at least a little bit. Also perhaps worth noting: Yesterday I learned that HTTP 2 no longer sends the keywords for status codes (I.e. 200 OK, 404 NOT FOUND, etc.) with the request, which is the reason curl doesn't print it either. So we could potentially just check for the status code and ignore everything else.

.travis/stages/verify/check-links-and-yaml-syntax/check-links
```diff
@@ -15,7 +15,7 @@ fi
 
 function checkUri() {
   local -r uri="$1"
-  curl -s --head "$uri" | head -n 1  | grep -qE  "HTTP/2 200|(200|302) (OK|Found)" \
+  curl -s --head "$uri" | head -n 1  | grep -qE "HTTP/(2 200|1.1 (200 OK|302 Found))" \
     || reportBrokenUri "$uri"
```
DanVanAtta, line 19: I think positive confirmation would be helpful to know what we are checking and are seeing. I'd also recommend we try and keep it simple. Perhaps there is no need to even check the HTTP type, just status code and the optional last word.

WDYT of something like this?

```
  echo -n "$uri -- "  
  local -r curlResult=$(curl -s --head "$uri" | head -n 1)
  echo "$curlResult"
  echo "$curlResult" | grep -qE "200( OK)?|302( Found)?" \
    || reportBrokenUri "$uri"
```

DanVanAtta:
I pushed and merged a change to disable the failure trigger from link checker. It's temporary to keep the build working.

.travis/stages/verify/check-links-and-yaml-syntax/check-links
```diff
@@ -15,28 +17,31 @@ fi
...
   local -r blue="\e[38;5;75m"
-  echo -en "Error: ${red}$uri${end} ${blue}"
-  curl -s --head "$uri" | head -n 1 | sed 's|HTTP/[12][.01]?||'
-  echo -en "${end}"
+  local -r error_reason=${2//${FAILURE_PREFIX}000/Connection Failure!}
+  echo -e "Error: ${red}${1}${end} ${blue}${error_reason}${end}"
+}
+
+function checkIsOnIgnoreList() {
+  for ignored_uri in "${IGNORE_LIST[@]}"; do
```
RoiEXLab, line 35: Interesting: If `IGNORE_LIST` is empty, this line throws an error on bash 4.3 and prior.
From the [changelog](https://tiswww.case.edu/php/chet/bash/CHANGES):
> a.  Using ${a[@]} or ${a[*]} with an array without any assigned elements when
    the nounset option is enabled no longer throws an unbound variable error.

Which is the reason this build failed: https://travis-ci.com/github/triplea-game/triplea/jobs/342279779

.travis/stages/verify/check-links-and-yaml-syntax/check-links
```diff
@@ -15,28 +17,31 @@ fi
...
+    FAILURE=1
+  fi
 }
 
 function reportBrokenUri() {
-  FAILURE=1 
   local -r red="\e[38;5;196m"
   local -r blue="\e[38;5;75m"
-  echo -en "Error: ${red}$uri${end} ${blue}"
-  curl -s --head "$uri" | head -n 1 | sed 's|HTTP/[12][.01]?||'
```
RoiEXLab, line 32: What surprised me here is that `$uri` is not actually the parameter passed to this function, (which was unused before my change), but instead the global variable that is defined below at `while read -r uri; do`.

I tried limiting the scope of this loop, but without success, so `$uri` could still  be used here.
I wonder if it's better to use the global uri everywhere, or just ignore it an pass it everywhere explicitly 🤔

.travis/stages/verify/check-links-and-yaml-syntax/check-links
```diff
@@ -15,28 +17,31 @@ fi
...
-  echo -en "${end}"
+  local -r error_reason=${2//${FAILURE_PREFIX}000/Connection Failure!}
```
RoiEXLab, line 30: This replacement expression replaces `Failed With Code: 000` with `Connection Failure`, which would be printed otherwise if curl was unable to reach the server at all

.travis/stages/verify/check-links-and-yaml-syntax/check-links
```diff
@@ -15,28 +17,31 @@ fi
...
   local -r uri="$1"
-  
-  local -r curlResult=$(curl -s --head "$uri" | head -n 1)
-  echo "$uri -- $curlResult"  
-  # Temporarily disabled. We're unable to connect to A&A.org from Travis.
-
-  # echo $curlResult |  grep -qE  "HTTP/2 200|(200|302) (OK|Found)" \
-  #  || reportBrokenUri "$uri"
+  local output
+  if ! output="$FAILURE_PREFIX$(curl -Lsf -o /dev/null -w "%{http_code}" "$uri")"; then
```
RoiEXLab, line 21: I think this is a lot more robust than the code we had before.
The command prints the actual HTTP status code purely for logging purposes.
`-L` makes sure to follow redirects.
`-f` makes curl return with non-zero exit codes in case the status code is 4xx or 5xx or something like that, whatever it considers to be an error
`-o /dev/null` just redirects the HTTP body i.e. the HTML to void so it won't land in stdout
`-w "%{http_code}"` makes curl print the status code in stdout if it exists, or "000" otherwise

RoiEXLab:
@DanVanAtta I think I was able to come up with a robust system.
Let me know what you think

DanVanAtta:
I see the output from the check links is:
```
.travis/stages/verify/check-links-and-yaml-syntax/check-links

Links are valid
```

I think we could use output of:
1. Which links were checked?
2. What were the status codes?


If we see a future failure, we can then check to the previous build to verify the previous status value. What do you think of adding such 'positive' confirmation?

.travis/stages/verify/check-links-and-yaml-syntax/check-links
DanVanAtta: `-f` and `-w` are really good flags, good work :+1: 

I think a good chunk of this comment probably needs to land in the code so that this reasoning/logic explanation can be carried forward.

DanVanAtta: Minor (consider fix if feasible): I also wonder if breaking up the assignment and status code check could potentially make things simpler, it seems we are quite overloaded on one line. Would it look gnarly to break it up to multiple lines?

.travis/stages/verify/check-links-and-yaml-syntax/check-links
DanVanAtta: Fortunate bug, I think it was probably intended to assign it to a local function variable and then use that variable. The variables name happened to match so it worked :disappointed:

.travis/stages/verify/check-links-and-yaml-syntax/check-links
```diff
@@ -5,6 +5,8 @@ set -eu
 URI_FILE=$(find . -type f -path "*/src/main/java/*" -name "UrlConstants.java")
 green="\e[32m"
 end="\e[0m"
+FAILURE_PREFIX="Failed with Code: "
+IGNORE_LIST=("https://www.axisandallies.org/forums")
```
DanVanAtta, line 9: Would really love more insight about why A&A.org is failing from travis, have you been able to do any more research/investigation into this? Having a full status code and perhaps full output from curl, if not manually done, then printed by the script would seem helpful.

It's hard to know if the previous failure message was accurate and was not masking something a bit more benign.

.travis/stages/verify/check-links-and-yaml-syntax/check-links
DanVanAtta: I'd recommend, if possible, replacing the '000' with 'Connection Failure' and then prefixing the 'Failed with Code' otherwise. Matching this much of a string seems indirect to me. I also think some commentary in-script would go quite a long way to explain that.

.travis/stages/verify/check-links-and-yaml-syntax/check-links
RoiEXLab: No Idea, I haven't been able to reach @djensen47 via GitHub.
My best guess atm is that the builds might exceeded some sort of threshold to we get blocked at a firewall level

.travis/stages/verify/check-links-and-yaml-syntax/check-links
RoiEXLab: The primary reason I made it like this is because shellheck suggested it.

I see your point but I'm not actually sure it would make things better tbh

.travis/stages/verify/check-links-and-yaml-syntax/check-links
RoiEXLab: I thought this way round is pretty convenient from a code perspective because it avoids the need for an else clause, but I think I can write something self-explanatory there

RoiEXLab:
@DanVanAtta Updated, please tell me what you thin of the changes

.travis/stages/verify/check-links-and-yaml-syntax/check-links
```diff
@@ -15,28 +16,37 @@ fi
...
-  echo -en "Error: ${red}$uri${end} ${blue}"
-  curl -s --head "$uri" | head -n 1 | sed 's|HTTP/[12][.01]?||'
-  echo -en "${end}"
+  local -r error_reason=$([[ $2 -ne 0 ]] && echo "Failed with Code: $2" || echo "Connection Failure!")
+  echo -e "Error: ${red}${1}${end} ${blue}${error_reason}${end}"
+}
+
+function checkIsOnIgnoreList() {
+  for ignored_uri in "${IGNORE_LIST[@]}"; do
+    if [[ "$1" = "$ignored_uri" ]]
```
DanVanAtta, line 37: Nit, the other scripts have a convention to assign parameters to variables in functions. Thus any parameter passed to a function receives a name, you see off-the-bat as well how many parameters are expected to be passed to a function. It also means that a variable like `$1` is only ever referenced once in a function and that is only to assign it to another variable that then gives it a name.

This was undocumented, I added a note about this convention here: https://github.com/triplea-game/triplea/wiki/Shell-Script-Conventions

.travis/stages/verify/check-links-and-yaml-syntax/check-links
DanVanAtta: A comment in the code with an issue reference would be good documentation and something for someone to check if there is ever follow-up on this.

I'm otherwise surprised the request volume was ever noticeable enough for us to run into this situation. I would think the build verification check, being one-time, would have been a pittance compared to a single user browsing A&A.org

.travis/stages/verify/check-links-and-yaml-syntax/check-links
```diff
@@ -15,28 +16,37 @@ fi
...
-  # Temporarily disabled. We're unable to connect to A&A.org from Travis.
-
-  # echo $curlResult |  grep -qE  "HTTP/2 200|(200|302) (OK|Found)" \
-  #  || reportBrokenUri "$uri"
+  local code
+  if ! code=$(curl -Lsf -o /dev/null -w "%{http_code}" "$uri"); then
+    reportBrokenUri "$uri" "$code"
+    FAILURE=1
+  else
+    echo "Successfully checked '$uri', returned $code"
```
DanVanAtta, line 24: :+1:

.travis/stages/verify/check-links-and-yaml-syntax/check-links
```diff
@@ -15,28 +16,37 @@ fi
...
+    fi
+  done
+  return 1
 }
 
 FAILURE=0
 
 while read -r uri; do
-  checkUri "$uri";
+  checkIsOnIgnoreList "$uri" || checkUri "$uri"
```
DanVanAtta, line 49: :+1:

.travis/stages/verify/check-links-and-yaml-syntax/check-links
```diff
@@ -15,28 +16,37 @@ fi
...
-  echo -en "${end}"
+  local -r error_reason=$([[ $2 -ne 0 ]] && echo "Failed with Code: $2" || echo "Connection Failure!")
+  echo -e "Error: ${red}${1}${end} ${blue}${error_reason}${end}"
+}
+
+function checkIsOnIgnoreList() {
+  for ignored_uri in "${IGNORE_LIST[@]}"; do
+    if [[ "$1" = "$ignored_uri" ]]
+    then
+      echo "Ignoring '$1'"
```
DanVanAtta, line 39: :+1:

PR #6549, opened 2020-05-31
MustFightBattle#determineStepStrings unittests

trevan:
This moves `MustFightBattle#determineStepStrings` into a helper class called `BattleSteps`.  I also moved several methods that were used by `determineStepStrings` and other parts of `MustFightBattle` into utility classes named `RetreatChecks` and `SubChecks`.

I didn't change any of the logic in any of the functions that were moved.  This will NOT pass the codeclimate checks.

I wrote unit tests for the new class `BattleSteps`.  It has 100% code coverage.  I believe I handle all of the possible code paths.  While working on the unittests, I found a few situations that I feel are bugs and I left a TODO there.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[X] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
I performed no manually tests.

## Additional Review Notes
I've written at https://forums.triplea-game.org/topic/2011/refactoring-mustfightbattle some of the rational behind this change.

The unit test file is the majority of this PR.  It is >2000 lines.


game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
```diff
@@ -0,0 +1,242 @@
...
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import lombok.Builder;
+import lombok.NonNull;
+import org.triplea.java.collections.CollectionUtils;
+
+/** Get the steps that will occurr in the battle */
+@Builder
+public class BattleSteps implements BattleStepStrings {
```
DanVanAtta, line 25: Would you mind describing the reasoning to implement BattleStepStrings?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
```diff
@@ -0,0 +1,242 @@
...
+  final @NonNull GameData gameData;
+  final @NonNull Collection<Unit> bombardingUnits;
+  final @NonNull Function<Collection<Unit>, Collection<Unit>> getDependentUnits;
+  final @NonNull Boolean isBattleSiteWater;
+  final @NonNull Boolean isAmphibious;
+  final @NonNull Supplier<Collection<Territory>> getAttackerRetreatTerritories;
+  final @NonNull BiFunction<GamePlayer, Collection<Unit>, Collection<Territory>>
+      getEmptyOrFriendlySeaNeighbors;
+
+  public List<String> get() {
```
DanVanAtta, line 48: `get()` is the method for a `Supplier`, should this class implement  `Supplier<List<String>>`?

Otherwise, seeing the usage, I would have thought that this class were a supplier.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
```diff
@@ -0,0 +1,242 @@
...
+import java.util.function.Supplier;
+import lombok.Builder;
+import lombok.NonNull;
+import org.triplea.java.collections.CollectionUtils;
+
+/** Get the steps that will occurr in the battle */
+@Builder
+public class BattleSteps implements BattleStepStrings {
+
+  final @NonNull Boolean canFireOffensiveAa;
```
DanVanAtta, line 27: Have you consider decomposing this class further?  It would likely simplify the testing substantially as well as this class itself.

To illustrate, the main 'get' method of this class would look something like this:
```
final List<String> steps = new ArrayList<>();
steps.addAll(aaBattleStepBuilder.apply(battleStepParameters);
steps.addAll(firstRunBattleStepBuilder.apply(battleStepParameters);
steps.addAll(submergeSubsBattleStepBuilder.apply(battleStepParameters);
:
:
```

Implied in the above, the class properties here would be extracted to a value object that would hold all of the parameters.

To enable test you would want to be able to inject implementations. That could be done by making the class have the following structure:


```
@Builder
public class BattleSteps {
  @Nonnull private final Function<BattleStepParameters, List<String>> aaBattleStepBuilder;
  :
  :

  // This method call it from test code but do nothing with it, just verify you can invoke this and 
  // and have no errors, that everything can be constructed, that will be an integration test.
  static newBattleStepBuilder() {
     return builder()
           .aaBattleStepBuilder(new AaBattleStepBuilder())
           :
           .build();
  }

} 
```

In the above, you can use test code to call `builder()` and inject mock implementations. Then when testing `get` all you need to do is to verify that you invoked each battle step builder, have each one return a stubbed value and then be sure that all of the stubbed values are present on the returned list.

EG:
```

@Mock
private Function<BattleStepParameters, List<String>> aaBatttleStepBuilder;

private BattleSteps battleSteps;

@Mock
private BattleStepParameters battleStepParameters;

@BeforeEach
void setup() {
  battleSteps = BattleSteps.builder()
     .aaBattleStepBuilder(aaBattleStepBuilder)
    :
    :
    .build();
}

@Test
void get() {
when(aaBattleStepBuilder.apply(battleStepParameters)).thenReturn("aaBattleStep"));
:
:

List<String> battleSteps = battleSteps.get(battleStepParameters);

// using explicit static imports as an example, favor statically importing the matchers below
assertThat(battleSteps, IsCollectionWithSize.hasSize( x ));
assertThat(battleSteps, IsIterativeWithItems.hasItems(
   "aaBattleSteps",
   :
   :
   ));
}
```

Finally from there, you can implement each set of battle steps one-by-one and test them each individually.

WDYT?

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/BattleStepsTest.java
```diff
@@ -0,0 +1,2199 @@
...
+  void setupMocks() {
+    when(mockGameData.getProperties()).thenReturn(mockGameProperties);
+  }
+
+  private void players() {
+    when(attacker.getName()).thenReturn("mockAttacker");
+    when(defender.getName()).thenReturn("mockDefender");
+  }
+
+  private void unit1() {
```
DanVanAtta, line 100: Explicit input/output of test setup methods can help make for more explicit unit tests. This method for example could possibly be changed to:

```
private Unit givenUnitWithAttachment(UnitAttachment unitAttachment) {
  Unit unit = mock(Unit.class);
  UnitType unitType = mock(UnitType.class);
  when(unit.getType).thenReturn(unitType);
  when(unitType.getAttachment(anyString()).thenReturn(unitAttachment);
  return unit;
}
```

Then a test, for at least this, can be a bit more self-contained, eg:

```
    players();
    attackerRetreat();
    Unit unit1 = givenUnitWithAttachment(mockUnit1Attachment);
    unit2();
    final List<String> steps =
        getStepBuilder()
            .attackingUnits(List.of(unit1))
            .defendingUnits(List.of(unit2))
            .isBattleSiteWater(false)
            .build()
            .get();
```
In the above, it's pretty clear the usage of 'unitAttachment' is a side-effect somewhere. Following the above pattern should eventually have unitAttachment become a parameter or a return value of some setup method, so at least you could see where it is configured (which should make the test a bit more self-explanatory and self-contained).


WDYT of this approach?

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/BattleStepsTest.java
```diff
@@ -0,0 +1,2199 @@
...
+
+  @Mock GameData mockGameData;
+  @Mock GameProperties mockGameProperties;
+  @Mock Function<Collection<Unit>, Collection<Unit>> mockGetDependentUnits;
+  @Mock Supplier<Collection<Territory>> getAttackerRetreatTerritories;
+
+  @Mock
+  BiFunction<GamePlayer, Collection<Unit>, Collection<Territory>> getEmptyOrFriendlySeaNeighbors;
+
+  @Mock Territory mockBattleSite;
```
DanVanAtta, line 73: nit: I used to personally have a style of prefixing mocks with `mock`, I think that style is a bit outdated now and the preference is to omit the `mock` prefix on mock variable names. IE: this variable would be just `battleSite`.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/BattleStepsTest.java
```diff
@@ -0,0 +1,2199 @@
...
+    when(unit1.getType()).thenReturn(mockUnit1Type);
+    when(mockUnit1Type.getAttachment(anyString())).thenReturn(mockUnit1Attachment);
+  }
+
+  private void unit2() {
+    when(unit2.getType()).thenReturn(mockUnit2Type);
+    when(mockUnit2Type.getAttachment(anyString())).thenReturn(mockUnit2Attachment);
+  }
+
+  private void unit3() {
```
DanVanAtta, line 110: The helper setup methods are good things :+1: 
Explicit input/output for them would not be bad though, as mentioned in another comment.
Another way to do the explicit input/output, for this method, as an example, woudl be:

```
private static void givneUnitHasAttachment(Unit unit, UnitAttachment unitAttachment) {
  UnitAttachment unitAttachment = mock(UnitAttachment.class);
   when(unit.getType).thenReturn(unitAttachment);
   when(unitAttachment.getAttachment(anyString()).thenReturn(unitAttachment);
}
```

Then as usage:
```
givenUnitHasAttachment(unit3, mockUnit3Attachment);
```
In the above you can see explicitly which variables are being configured.


These should complement well, how tests are done, for example:
```
Unit aaUnit = givenUnitWithAttachment(unitAttachment1);
when(unitAttachment.getTypeAa()).thenReturn("AntiAirGun");

Unit notAaUnit = givenUnitWithAttachment(unitAttachment2);
when(unitAttachment.getTypeAa()).thenReturn("");
```
In the above, later when you use `aaUnit` and `notAaUnit` it'll be a bit more explicit how they were configured and why we have expectations for one unit or another to be in an output list.


If you go the route of breaking up BattleStep, then it'll potentially start to make sense to have even more involved helper methods for the test, ie:
```
Unit aaUnit = givenUnitWithAaType("AntiAirGun");
Unit notAaUnit = givenUnitWithAaType("");
:
:
```

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
trevan: Sure.  Almost all of the constants in that class are used by `determineStepStrings`.  Since I moved that code into a new class, I thought I would just implement it just like MustFightBattle.  I'm not sure why it was implemented by MustFightBattle vs importing the constants directly and just wanted to follow the existing convention.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
trevan: No, this isn't supposed to be a supplier.  I'm just bad with names.

DanVanAtta:
Looking at usages of battle step strings, I see there is code to determine battle steps that is potentially scattered, eg: `StrategicBombingBattle` has some logic for step strings. Any thoughts if that will get looped in?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
trevan: Oh, I absolutely plan to reconstruct this class.  This PR is just the first step.  I wanted a small PR where everything was unittested.  Then I can create further PRs where I start to reconstruct it.

Have you seen my ideas on the forum - https://forums.triplea-game.org/topic/2011/refactoring-mustfightbattle?  I'd like to combine this code with the step executable code.  Each step would be its own separate class, like you mentioned.

trevan:
> Looking at usages of battle step strings, I see there is code to determine battle steps that is potentially scattered, eg: `StrategicBombingBattle` has some logic for step strings. Any thoughts if that will get looped in?

Yes, it appears that each of the battle classes have logic to calculate their step strings and logic to calculate and run the actual steps.  I'd like to unify how it works so that the steps are self contained and then each battle would have a "step generator" that would build out the list of steps and in the correct order.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/BattleStepsTest.java
trevan: Interesting idea.  I just recently tried out a different approach where I did something like:

```
private Unit createUnit() {
   Unit unit = mock(Unit.class);
   UnitType unitType = mock(UnitType.class);
   when(unit.getType).thenReturn(unitType);
   when(unitType.getAttachment(anyString()).thenReturn(unitAttachment);
}

**Comment by trevan** [comment]
> Left a number of comments to potentially help improve things further. Let me know what you think @trevan. IMO it would be fine to merge as-is and then follow-up, or follow-up in this same PR.

@DanVanAtta , I removed the `mock` prefix in the names.  I'd like to merge it now so that I can start trying to refactor the class.  I like your ideas about the mock objects and I'm going to try and utilize them.  I'm thinking of creating a `MockUnit` class that has these helper methods `newUnit`, `givenUnitWithAaType`, etc.  Then I can share this code between the tests as I work on combining the steps logic.

**Comment by DanVanAtta** [comment]
Be a bit careful with too much test scaffolding @trevan , I'd recommend to keep any helper's local to the package of tests that use them. Most tests should be different enough that the common helpers may not help as much as one would think (and then it might appear that we are just dumping all mock unit expectations there, which is not great as tests are really ideally self-contained).

File: game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/BattleStepsTest.java (hunk 4 above)
Comment (DanVanAtta) [reply game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/BattleStepsTest.java:100]: The only criticism I would have for that would be this line:
```
unit.getAttachment("");
```

It takes some guess work why the empty string returns something useful. At this juncture, IMO the simplest would be to have a helper method that would allow you to write something like: 
```
Unit aaUnit = givenUnitWithAaType("AntiAirGun");
```

trevan:
Should I create a tracker bug so that others can see where I'm at in the refactoring process, @DanVanAtta ?

DanVanAtta:
If you wish @trevan , my inclination is to suggest to not worry about it and instead keep a personal backlog somewhere. If there's anything to discuss, an issue/forum post or a draft PR are both good things though.

PR #6551, opened 2020-05-31
Initial unittests for MustFightBattle#getBattleExecutables

trevan:
Gives all of the step executables a class name so that they can be accessed from unittests.

Adds tests for most of the executable steps.  Some of the steps are not tested because they have the exact same checks in BattleSteps.  Since I plan to merge the two steps code together, I didn't want to duplicate the tests at this point.

This should be the last really large commit.  After this, I'll be working on combining the logic for each step and moving it into their own step classes.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:


game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
```diff
@@ -0,0 +1,378 @@
...
+          mockGameData.performChange(change);
+          return null;
+        })
+        .when(delegateBridge)
+        .addChange(any());
+
+    step.execute(null, delegateBridge);
+
+    verify(mockGameData, times(1)).performChange(ChangeFactory.EMPTY_CHANGE);
+    // TODO: Somehow get the stack and check that no new execs were added to it.
```
trevan, line 331: The `FireNavalBombardment` step adds more items to the stack.  I'd like to test that the items were either added or not added but the stack is not visible for testing nor is the `deque` inside of the stack.

I could either:
* make the stack and its deque visible to testing
* change the step code to use the stack that is currently passed in to each of the steps instead of the instance version of the stack

I prefer the later because then I can mock it but the former has less change to the code.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
```diff
@@ -0,0 +1,378 @@
...
+        List.of(),
+        defender,
+        List.of()
+    );
+    final List<IExecutable> execs = battle.getBattleExecutables(true);
+
+    final int index = getIndex(execs, MustFightBattle.FireNavalBombardment.class);
+    final IExecutable step = execs.get(index);
+
+    final IDelegateBridge delegateBridge = mock(IDelegateBridge.class);
```
trevan, line 362: I tried using `MockDelegateBridge#newDelegateBridge` but the strict stubbing check in Mockito makes it not usable.  Mockito complains if a stub was added that wasn't used and `newDelegateBridge` adds a lot of stubs that these individual tests don't use.

I could make turn on the `lenient` mode but I worry about those repercussions.  Ideas on how to simplify the mock setups?

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -69,6 +69,256 @@
     PARTIAL_AMPHIB
   }
 
+  /**
+   * An action representing attacking aa guns firing during a battle.
+   *
+   * <p>NOTE: This type exists solely for tests to interrogate the execution stack looking for an
+   * action of this type.
+   */
+  public static abstract class FireOffensiveAaGuns implements IExecutable {
```
DanVanAtta, line 78: Marker interfaces/classes are usually an anti-pattern. Are each of these steps associated with a specific battle step string?

It seems like this is perhaps more fitting to an enum. EG:
```
@AllArgsConstructor
enum BattleStepPhase {
  FIRE_OFFENSIVE_AA_GUNS("Fire Offensive AA"),
  :
  :
  ;

  @Getter
  private final String label;
}
```

At that point the battle step list can become a list of `BattleStepPhase` and then just convert it to a list of strings at time of display, eg:
```
  battleStepPhases.stream()
     .map(BattleStepPhase::getLabel)
```

With a list of `BattleStepPhase`, you can query that the list has the expected enum element.

To tie it in with the executable, a helper object probably would be needed and you would have a list of those instead. EG:

```
@Value
@Builder
class BattleStep {
   BattleStepPhase battleStepPhase;
   Runnable executable;
}
```

Then you'd have `List<BattleStep>`, and you could query which battle phase it is, get the label from that, and have a reference to the actions to be run. WDYT?

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
DanVanAtta: Have you considered extracting the creation of the battle stack? 

If you had an object that is injected via constructor injection, you could mock it. Then you'd only need to test that you called the method and it would return an arbitrary list of battle execution steps. Then in turn you could test the usage of that class that creates the battle step stack and assert the stack has the right size and right contents in the right conditions.

It's always a bit of a code smell when you have to test via:
```
when(...).thenReturn(...)

doVoidMethodCall(..)

verify(object).someMethod(..);
```

The above is testing for side-effects, it can usually be re-arranged with a strategy object that returns a behavior or a value.

In other words, have you considered making the stack and its deque be the output of another class, a strategy class? You could then inject & mock that strategy class and then test the stack more directly.

trevan: I have thought of that.  I'm just trying to do that in pieces.  I want to first get unittests in so that as I make the changes, I don't break something accidentally.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: This is basically what I have in mind.  But you keep saying do small PRs :)

It isn't quite an enum.  The order of steps is dynamic depending on the units and depending on the state of the battle.  And a "step" can have zero or more "step strings".  Also a "step string" might be shown but the "step" actually won't run because of changes to the units during the course of the battle.

This PR is just go get started with unittests in the code base and to work out any issues around those unittests.  I don't want to make any structural changes at this point for fear of breaking something without realizing it.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
DanVanAtta: I think in part the issue is with the code that is being tested. It is doing too much, nearly everything, so we are left stubbing too many items and testing via `verify` and asserting side effects. 

An approach to explore, extract functionality from a top level into strategy objects. The strategy objects are then ideally nearly functional and require very few mocks to tests. The high level then contains most of the control-flow logic, easily mocked, and you mainly just need to assert interactions between objects at that level.

For example, if another object, a strategy object, were to compute the bombard steps, then you would not need to test for it here as that computation would be mocked. You'd only need to test that the mock object is invoked in the right conditions.

trevan: So would you say it is fine to go with the current approach (complex mock setups) to ensure everything is tested?  And then as the code is reconstructed with helper functions, classes, etc, the mocks can get simplified.  I'd be fine with doing that.

I don't want to try and extract the functionality at this point because it is scattered all over the place.

trevan:
@DanVanAtta I've finished with adding tests.  This should be ok to merge.  The only changes to actual code is to make things visible and accessible to the unit tests.

I also reordered a few of the tests in BattleSteps.  The tests for the executable steps and for BattleSteps are very similar.  I plan to merge all of the logic together after this is merged.  I'll be doing the merging in small PRs.  About one for each step.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
```diff
@@ -0,0 +1,1671 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.triplea.java.collections.IntegerMap;
+import org.triplea.sound.ISound;
+import org.triplea.test.common.Integration;
+import org.triplea.util.Tuple;
+
+@ExtendWith(MockitoExtension.class)
+@Integration
```
trevan, line 76: I marked this `@Integration` because I didn't want to put these tests in the original MustFightBattleTest.java file.  The checks complained about having a test class that isn't connected to a non-test class unless I put `@Integration` on it.  I can move the tests to the other file if this is a problem.

Once I've refactored steps, these tests will be moved to the individual step test file so it is just temporary.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
DanVanAtta: It's a chicken-and-egg problem. The code was not written with tests in mind, it is hardly testable, to make it testable, it needs to be re-written, to re-write it - it needs tests first.

A technique for working with this kind of code is to apply high level characterization and integration tests to test large modules. Re-write the inner ones carefully piece by piece and add unit tests to them. Combine with some manual testing and rinse-wash-repeat until it's all well decomposed. That can be a difficult process as the original code can have cyclical dependencies, poor modularity (or no modularity at all), etc..

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
DanVanAtta: Understood, there is a trade-off.

These classes being present only for tests is not ideal:
```
  public abstract static class AttackerRetreatPlanes implements IExecutable {
    private static final long serialVersionUID = 6828071620685651390L;
  }
```

I can understand that being an intermediary step for lack of better solutions without going into deeper changes. I wonder though if the unit tests are going to be re-written just as well if you break up the code. If when re-writing the prod code you also have to re-write the tests, then it's still a re-write nonetheless. Unit tests written after-the-fact can suffer from verifying (or testing) that a bug exists! IE: the code could be incorrect, and there can be a test that asserts the code functions as-is, bugs included. 

If you break up the battle steps to their own classes, they will lose that abstract class and then the test will break because it's asserting for the wrong abstract class.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1127,6 +1379,8 @@ private void fireNavalBombardment(final IDelegateBridge bridge) {
                 attacker, battleSite, gameData));
     // bombarding units can't move after bombarding
     if (!headless) {
+      // TODO: StepRefactor: Why is a change always added even if there are no units?
```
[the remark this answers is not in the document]
DanVanAtta: We've had some disucssions on TODO. I can provide links if desired. Our net conclusion is that TODOs wind up being confusing and are often never done. It's perhaps better to keep a personal backlog for such questions rather than inject them into the code where it's still only really useful for the author anyways (in which case just having a personal backlog is just-as-well).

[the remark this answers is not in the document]
DanVanAtta: No follow-up required.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -72,6 +72,256 @@
...
+    private static final long serialVersionUID = 994041503171919922L;
+  }
+
+  /**
+   * An action representing standard attacking fire.
+   *
+   * <p>NOTE: This type exists solely for tests to interrogate the execution stack looking for an
+   * action of this type.
+   */
+  public abstract static class StandardAttackersFire implements IExecutable {
```
DanVanAtta, line 211: If we later convert the battle steps to be something else, modules with injected functionality and step strings, we will need to remove these abstract classes. At the very least we will want to remove them so we avoid prod-code that is here only to support tests (a code smell).

Is there argument for not updating `IExecutable` to return a step-name enum value? This way if you have a list of `IExecutable`, you could call a `IExecutable#getStepName` method and check to see if it is an expected value, eg: `IExecutable.getStepName() == BattleStepName.STANDARD_ATTACKERS_FIRE`.

If you then restructure the steps to be individual classes and then compose them together to get the full battle sequence, I suspect any unit test checking the ordering of the steps would remain intact. WDYT?

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
DanVanAtta: `@SuppressWarnings("UnmatchedTest")` Should do the trick directly

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: This naming of `IExecutables` was there originally for the sub first strike instances.  It is used by the RevisedTest and has been around since 2009.  I was just following that mechanism.

As for changing `IExecutables`, I don't like that idea.  They are used in a lot of places (battles, movements).  A grep shows 48 `new IExecutable` and another 7 places that implement `IExecutable` (ignoring the ones in `MustFightBattle`).  I'd have to edit a lot of places and I only need the step names in a few of the places.  So I'd either have to come up with a bunch of names for all of the places or put a placeholder in for the places that the tests don't care about.

I've pushed a change that deleted many of these test classes because they weren't being used for tests.  That leaves 12 of them, with two of them being around since 2009.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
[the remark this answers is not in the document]
trevan: I had found the entry in the wiki at https://github.com/triplea-game/triplea/wiki/Java-Code-Conventions#todo-comments-attach-a-tracking-token-5249.  I used "StepRefactor" is my 'grep token'.

DanVanAtta:
I was mistaken about the suppression for this case, it was another check where that worked. In this case the same suppression should work too, I updated the orphaned test check to include a check for suppression: https://github.com/triplea-game/triplea/pull/6604

After #6604 is merged, please merge this PR to latest @trevan and the build check should then be fixed.

DanVanAtta:
@trevan have you checked if this breaks save-game compatibility? Are IExecutables saved as part of `GameData` serialization?

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
```diff
@@ -0,0 +1,1670 @@
...
+    lenient().when(mockRelationshipTracker.isAtWar(attacker, defender)).thenReturn(true);
+    lenient().when(mockRelationshipTracker.isAllied(attacker, attacker)).thenReturn(true);
+    lenient().when(mockRelationshipTracker.isAllied(defender, defender)).thenReturn(true);
+    lenient().when(mockRelationshipTracker.isAllied(defender, attacker)).thenReturn(false);
+    lenient().when(mockRelationshipTracker.isAllied(attacker, defender)).thenReturn(false);
+
+    return new MustFightBattle(battleSite, attacker, gameData, battleTracker);
+  }
+
+  private void assertThatStepIsMissing(
```
DanVanAtta, line 129: nit, we use 'depth-first-method-ordering'. I'm not sure if it's always the cleanest for tests. The ordering essentially is that methods should be defined as soon after first usage as possible. IE:
```
@Test
void foo() {
  helper1();
  helper2();
  // stuff
}

void helper1() { ... }

void helper2() { ... }

@Test
void foo2() {
   helper1();
   helper2();
  :
  :
}

```

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
```diff
@@ -0,0 +1,1670 @@
...
+  void firstStrikeOrderAttHasDestroyerDefHasDestroyerWW2v2TrueSneakAttackTrueFalse() {
+    assertThatFirstStrikeStepOrder(
+        givenFirstStrikeBattleSetup(true, true, true, true, true),
+        List.of(
+            FirstStrikeBattleStep.ATTACKER,
+            FirstStrikeBattleStep.DEFENDER,
+            FirstStrikeBattleStep.STANDARD));
+  }
+
+  private Tuple<MustFightBattle, List<IExecutable>> givenFirstStrikeBattleSetup(
```
DanVanAtta, line 1120: We're trying to kill `Tuple`, please avoid it's usage. Typically we recommend to just use a value object instead. Lombok makes this pretty easy. EG:

```
@Value
class MustFightBattleWithExecutable {
  private MustFightBattle mustFightBattle;
  private IExecutable executable;
}
```

Second, the usage is quite hard to follow, ie:
```
  givenFirstStrikeBattleSetup(true, true, true, true, true),
```

Do you have any object to extracting the parameters to a value object?

EG:
```
givenBattleSetup(BattleSetupParameters.builder()
   .firstStrike(true)
   .attackerHasDestroyer(true)
   .defenderHasDestroyer(true)
   .ww2v2Rules(true)
   .build());
```

One nice benefit here is that any booleans that are false can be omitted and you can use the default 'false' value for them.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
```diff
@@ -0,0 +1,1670 @@
...
+      final boolean attackerDestroyer,
+      final boolean defenderDestroyer,
+      final boolean ww2v2,
+      final boolean defendingSubsSneakAttack,
+      final boolean ignoreDefendingSubsSneakAttack) {
+    final MustFightBattle battle = spy(newBattle(WATER));
+    lenient().doNothing().when(battle).firstStrikeAttackersFire(any());
```
DanVanAtta, line 1127: Which cases invoke this lenient and which other ones do not? Is the lenient a sign that we could use multiple setup methods? I'm curious about this as lenient mocks can be a testing smell, perhaps an overly generic test scaffolding.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
```diff
@@ -0,0 +1,1670 @@
...
+      when(gameProperties.get(DEFENDING_SUBS_SNEAK_ATTACK, false))
+          .thenReturn(defendingSubsSneakAttack);
+    }
+
+    final Unit attackerUnit = attackerDestroyer ? givenUnitDestroyer() : givenUnit();
+    final Unit defenderUnit = defenderDestroyer ? givenUnitDestroyer() : givenUnit();
+
+    battle.setUnits(
+        List.of(defenderUnit), List.of(attackerUnit), List.of(), List.of(), defender, List.of());
+    final List<IExecutable> execs = battle.getBattleExecutables(true);
```
DanVanAtta, line 1143: The IExecutable comes from a battle, do we really need the tuple? Could we just get the second value by calling `battle.getBattleExecutables(true)` directly?

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
```diff
@@ -0,0 +1,1670 @@
...
+    return Tuple.of(battle, execs);
+  }
+
+  private enum FirstStrikeBattleStep {
+    ATTACKER,
+    DEFENDER,
+    STANDARD,
+  }
+
+  private void assertThatFirstStrikeStepOrder(
```
DanVanAtta, line 1154: nit, something to watch for, if you can mark a method as static, please do so. Only ask for this to convey more information, static coupling is bad, while marking methods as static conveys useful information.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
```diff
@@ -0,0 +1,1670 @@
...
+        getIndex(execs, MustFightBattle.FirstStrikeDefendersFire.class));
+    indices.put(
+        FirstStrikeBattleStep.STANDARD,
+        getIndex(execs, MustFightBattle.StandardAttackersFire.class));
+
+    assertThat(indices.get(stepOrder.get(0)), lessThan(indices.get(stepOrder.get(1))));
+    assertThat(indices.get(stepOrder.get(1)), lessThan(indices.get(stepOrder.get(2))));
+  }
+
+  private void assertThatFirstStrikeReturnFireIs(
```
DanVanAtta, line 1176: A few nits here:

1. The boolean parameter is a bit hard to grok. Specifically "is attacker = true" is not necessarily clear until you look at the implementation in detail.

2. We are not really asserting anything in this method. I initially wondered if a matcher could replace this method, but instead we are doing a mockito verify.

What do you think of renaming to "verify" instead of "assert" as the method prefix?

I'd also suggest the boolean parameter be removed simply by creating additional methods that can be called directly to then call the underlying method, ie:
```
  private void verifyAttackersFireWithFirstStrike(
      final Tuple<MustFightBattle, List<IExecutable>> battleTuple,
      final MustFightBattle.ReturnFire returnFire) {

    verifyThatFirstStrikeReturnFireIs(battleTuple, returnFire, true);
  }
```
WDYT? Is there a cleaner way to get rid of the boolean parameter from the test cases?

trevan:
> @trevan have you checked if this breaks save-game compatibility? Are IExecutables saved as part of `GameData` serialization?

I have loaded master, played a game, then saved at the start of the battle round, in the middle of a sea battle and a land battle, and then after the battle round.  I've then loaded each of those saves in this branch and everything loaded fine and looked correct.  Is there other things that I need to check for compatibility?

And I believe the current IExecutables in the stack are saved as part of the `GameData` serialization.  I haven't added or removed any of the IExecutables so they should deserialize/serialize without any issues.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
```diff
@@ -0,0 +1,1670 @@
...
+        givenFirstStrikeBattleSetup(true, true, true, true, true),
+        MustFightBattle.ReturnFire.ALL,
+        true);
+  }
+
+  @Test
+  @DisplayName(
+      "When attacker has a destroyer, defender has a destroyer, WW2v2 is true, "
+          + "and DEFENDING_SUBS_SNEAK_ATTACK is either, then defender has return fire all")
+  void firstStrikeDefenderReturnFireAttHasDestroyerDefHasDestroyerWW2v2TrueSneakAttackTrueFalse() {
```
DanVanAtta, line 1215: Good use of the display name, though the test method name I think maybe became a bit too formulaic.

For example: `SneakAttackTrueFalse` does not really mean that much. I'd recommend looking at the method names and trying to state things more directly.  IE: The morpheus principle of naming, "Stop trying to hit me and hit me!"

Suggestion:
```
ww2v2DefenderHasReturnFireAllWhenBothAttackerAndDefenderHaveDestroyers()
```

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
```diff
@@ -0,0 +1,1670 @@
...
+    assertThatFirstStrikeReturnFireIs(
+        givenFirstStrikeBattleSetup(true, false, false, false, false),
+        MustFightBattle.ReturnFire.ALL,
+        false);
+  }
+
+  @Test
+  @DisplayName(
+      "When attacker has no destroyer, defender has a destroyer, WW2v2 is true, "
+          + "and DEFENDING_SUBS_SNEAK_ATTACK is either, then attacker>defender>standard")
```
DanVanAtta, line 1400: nit, if the value of `and DEFENDING_SUBS_SNEAK_ATTACK is either`, you can probably just omit it from the display name. I think it would only be important if this test were parameterized and you actually test both the `DEFENDING_SUBS_SNEAK_ATTACK == true` and `DEFENDING_SUBS_SNEAK_ATTACK == false` cases.

Have you considered actually parameterizing this test to verify both of those cases? EG:

```
@ParameterizedTest
@ValueSource(booleans = { true, false })
void testMethodName(boolean defendingSubsSneakAttack) {
  // ....
}
```

If the test is parameterized, there is some auto-generation of test case names, check how they read in your IDE output and perhaps the display name can be shortened.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
```diff
@@ -0,0 +1,1670 @@
...
+        MustFightBattle.ReturnFire.ALL,
+        false);
+  }
+
+  @Test
+  @DisplayName(
+      "When attacker has a destroyer, defender has a destroyer, WW2v2 is false, "
+          + "and DEFENDING_SUBS_SNEAK_ATTACK is false, then attacker>standard>defender")
+  void firstStrikeOrderAttHasDestroyerDefHasDestroyerWW2v2FalseSneakAttackFalse() {
+    assertThatFirstStrikeStepOrder(
```
DanVanAtta, line 1262: CONSIDER THIS COMMENT FIRST!

After reading through most of the tests, the pattern of testing I think has become more clear. Have you considered parameterizing the whole thing?

If you had a value object for the battle setup, then it should become pretty easy!

Example:
```
@ParameterizedTest
@MethodSource
void verifyFirstStrikeBattleScenarios( 
     BattleSetupParameters setup, MustFightBattle returnFire, boolean attackersExpectedToHaveFirstStrike) {

   verifyFirstStrikebattle(
        setup,  returnFire,  attackersExpectedToHaveFirstStrike);
}


static List<Arguments> verifyFirstStrikeBattleScenarios() {
  return List.of(
     Arguments.of(BattleSetupParameters.builder()....build(), MustFightBattle.ReturnFire.ALL, false),
     :
     :
     );
}
```

DanVanAtta:
@trevan the compatibility check is simply if you can load an older save into a newer branch. I'm a bit surprised but it looks like that might have been okay (?). For the record, the save-game compatibility is a bane for us. Having been in a non-compatible mode for a long time now has really been useful, now that we are looking to keep compatibility as we approach 2.0, that problem is coming back in full force. I still think fixing save-game serialization is one of the highest project priorities we have so we can at least do simple things like add private fields, change methods, etc... without breaking save game compatibility in a way that is unrecoverable. We are thinking to do JSON serialization instead of java object serialization.

@trevan I think any further changes can be done as iterative PRs. Please merget to latest to fix the build and we'll iterate on this in additional PRs. I'll trust you'll keep working on most all areas mentioned. As you mentioned, you're trying to keep things bite-sized, that approach should help unblock those efforts and split them into multiple parallel efforts rather than a monolithic PR.

DanVanAtta:
I'll go ahead and take care of the merge to latest @trevan

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
trevan: Yes, I had thought of parameterizing it.  I'll do that next time.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
trevan: I don't remember why I had the tuple.  The tests were once a bit different so I probably had a reason and then after refactoring them, the reason disappeared but I still had the tuple legacy.  It can be simplified.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
trevan: The cases where `attacker == true` in `assertThatFirstStrikeReturnFireIs` will call the `firstStrikeAttackersFire` and when `attacker == false` it will call `firstStrikeDefendersFire`

PR #6552, opened 2020-05-31
Update to include WW2 Path to Victory

regularkyd:
Update to include WW2 Path to Victory

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -421,7 +421,7 @@
     <br>Game design by alkexr. Visual style by Hepps. Contributions by Frostion and redrum. Thanks for the members of the TripleA community for their feedback and ideas, who enabled turning this into a truly great map.
     <br>This map is the continuation of the (deprecated) Large Middle Earth map. Currently in balance testing period.
     <br>Version 2.0.7, last update 2018.10.08. for engine 1.9.0.0.11000+
-  mapName: WW2 Path to Victory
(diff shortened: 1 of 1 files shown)


ron-murhammer:
@RoiEXLab Looks like there is some issue with the link checker from another open issue? Can you take a look at this to see if it looks ok and merge it so the map maker can test the download?

RoiEXLab:
@ron-murhammer Kind of. There is an issue with the script, but travis isn't able to reach the a&a forum for some reason, which causes the problem. @DanVanAtta suggested to just disable this check for now, but I don't plan to do this before tomorrow (it's currently 2am here)

ron-murhammer:
Ok, can we just merge this PR though? Just wanted a second set of eyes on the yaml changes.

RoiEXLab:
We could but it would cause the master build to fail, so it would be better to just comment in the line in the bash script that does the link checking

DanVanAtta:
Disabled the link check for in-game links, merged latest to this PR. The link check then failed but due to a typo in the YAML. I fixed that just now, should be good to merge once build checks pass.

DanVanAtta:
Congrats @regularkyd , the map should be live for download now from in-game.

PR #6556, opened 2020-06-01
Update Flyway deployment

DanVanAtta:
- Instead of changing user to flyway (brittle if not already root),
  instead use explicit ownership commands to change ownership
  of flyway download artifacts.
- Inline deployment zip variable (to simplify)

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

Verified with a vagrant modified to Ubuntu 20.04
```
cd infrastructure
vagrant up
./run_ansible_vagrant
```

### DIFF
--- a/infrastructure/ansible/roles/flyway/tasks/main.yml
+++ b/infrastructure/ansible/roles/flyway/tasks/main.yml
@@ -7,43 +7,41 @@
 
 - name: Download flyway CLI
   become: true
-  become_user: flyway
   get_url:
     url: "{{ flyway_download_location }}"
-    dest: "{{ flyway_downloaded_file_location }}"
+    dest: "/home/flyway/flyway-cli.tar.gz"
     mode: "644"
     checksum: "md5:57f496acc6399fe30ee1fd957545abf1"
+    owner: flyway
+    group: flyway
 
 - name: extract flyway
   become: true
-  become_user: flyway
   unarchive:
     copy: no
-    src: "{{ flyway_downloaded_file_location }}"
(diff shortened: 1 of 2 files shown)


infrastructure/ansible/roles/flyway/tasks/main.yml
```diff
@@ -7,47 +7,46 @@
...
     checksum: "md5:57f496acc6399fe30ee1fd957545abf1"
+    owner: flyway
+    group: flyway
 
 - name: extract flyway
   become: true
-  become_user: flyway
   unarchive:
     copy: no
-    src: "{{ flyway_downloaded_file_location }}"
```
RoiEXLab, line 22: Curious to know why you decided inlining this variable would be a better choice

DanVanAtta: I think it was a bit over-zeolous using of variables. Reducing this level of duplication is not worth very much, and it was more confusing to know where the file was being created which made it harder to diagnose permission errors writing to that location. There is also about zero chance we would want to override it, so that in combination with the little de-duping value and that it's used in close proximity; it's a case where a variable does not really help IMO. I think to another extent *key* configurations are better as variables as well, things that are duplicated a lot or do have overrides. To another extent keeping the variables to a shorter list is a good thing, this one is more of an intermediate value as well and the path could almost just as well be `/tmp` (so it's not the most valuable of abstractions).

PR #6566, opened 2020-06-01
Execute 'updateMoves' async

DanVanAtta:
Changes bridge code to be run async when fetching moves for
the abstract-move-panel. The invoked code downstream
will already update UI in a swing background thread, which implies
sending this code to be async should be safe as we already
execute the final update in a 'scheduled-later' thread.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Testing

- verified in a single player could do combat moves and view moves in history accurately and undo some moves. (Blocking operation on EDT warning was triggered via single player game, tested that scenario)

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/AbstractMovePanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/AbstractMovePanel.java
@@ -14,6 +14,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.logging.Level;
 import javax.swing.Box;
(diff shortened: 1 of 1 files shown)


DanVanAtta:
Agree @RoiEXLab ; the warning has been really effective. I kinda wish the exception was specific as it can cause a bit of noise, but it's probably better to just fix the underlying problem. I feel really similar about the error reporting which has lead to a number of fixes. It's something to feel good about, fewer cases where the game can just freeze, or crash. I don't want to jynx anything, so far 2.0 is looking like it could be one of the more solid releases yet.

PR #6573, opened 2020-06-02
Simplify: Combine if statements

DanVanAtta:
- Resolves various Codacy violations
- Please double check for any 'or' conditions in particular and that if blocks are combined correctly
- PR is broken up into multiple to avoid review overload
- Please review without whitespace changes, only changes to if statement blocks

https://github.com/triplea-game/triplea/pull/6573/files?w=1

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/ObjectivePanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/ObjectivePanel.java
@@ -196,17 +196,16 @@ public class ObjectivePanel extends AbstractStatPanel {
 
         // find which section
         boolean found = false;
-        if (sections.containsKey(player.getName())) {
-          if (sections.get(player.getName()).contains(key.get(1))) {
-            final Map<ICondition, String> map = statsObjectiveUnsorted.get(player.getName());
-            if (map == null) {
-              throw new IllegalStateException(
-                  "objective.properties group has nothing: " + player.getName());
-            }
-            map.put(condition, value);
-            statsObjectiveUnsorted.put(player.getName(), map);
-            found = true;
+        if (sections.containsKey(player.getName())
+            && sections.get(player.getName()).contains(key.get(1))) {
+          final Map<ICondition, String> map = statsObjectiveUnsorted.get(player.getName());
+          if (map == null) {
+            throw new IllegalStateException(
+                "objective.properties group has nothing: " + player.getName());
           }
+          map.put(condition, value);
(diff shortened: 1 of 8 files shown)


game-core/src/main/java/games/strategy/triplea/ui/ObjectivePanel.java
```diff
@@ -196,17 +196,16 @@ private void setObjectiveStats() {
...
-            final Map<ICondition, String> map = statsObjectiveUnsorted.get(player.getName());
-            if (map == null) {
-              throw new IllegalStateException(
-                  "objective.properties group has nothing: " + player.getName());
-            }
-            map.put(condition, value);
-            statsObjectiveUnsorted.put(player.getName(), map);
-            found = true;
+        if (sections.containsKey(player.getName())
+            && sections.get(player.getName()).contains(key.get(1))) {
```
RoiEXLab, line 200: Just a thought:
```java
sections.getOrDefault(player.getName(), List.of()).contains(key.get(1))
```

would be a shorter variant of the if statement, I haven't convinced myself that it's clearly better though

game-core/src/main/java/games/strategy/triplea/ui/ObjectivePanel.java
```diff
@@ -196,17 +196,16 @@ private void setObjectiveStats() {
...
+          final Map<ICondition, String> map = statsObjectiveUnsorted.get(player.getName());
+          if (map == null) {
+            throw new IllegalStateException(
+                "objective.properties group has nothing: " + player.getName());
           }
+          map.put(condition, value);
+          statsObjectiveUnsorted.put(player.getName(), map);
```
RoiEXLab, line 207: This put operation is redundant btw if I'm not missing something obvious

game-core/src/main/java/games/strategy/triplea/ui/ObjectivePanel.java
DanVanAtta: A more radical change introduces unwanted risk and perhaps should arguably be tested. I'm not saying I disagree, only that such a change should be done as it's own PR where we make that one change or a set of similar changes and ensure it is adequately tested.

DanVanAtta: A multimap might be what we want. Having a person reading code realize that the later execution is a no-op because we default to an empty list is a bit indirect. It is arguably simpler than the series of if statements, I agree that is debatable.

Issue #6469, opened 2020-05-14
NullPointerException - HeadedGameRunner.lambda$initializeClientSettingAndLogging$0


tripleabuilderbot:
## User Description
running auto placement finder and received error

## TripleA Version
2.0.19022

## Java Version
11.0.6

## Operating System
Windows 10

## Exception 
java.lang.NullPointerException

## Stack Trace
```
java.lang.NullPointerException
	at games.strategy.triplea.ui.mapdata.MapData.getContainedTerritoryPolygons(MapData.java:522)
	at tools.image.AutoPlacementFinder.calculate(AutoPlacementFinder.java:269)
	at tools.image.AutoPlacementFinder.runInternal(AutoPlacementFinder.java:115)
	at tools.image.AutoPlacementFinder.run(AutoPlacementFinder.java:83)
	at tools.map.making.MapCreator.runUtility(MapCreator.java:365)
	at tools.map.making.MapCreator.lambda$createPart2Panel$11(MapCreator.java:301)
	at org.triplea.swing.SwingAction$1.actionPerformed(SwingAction.java:60)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1967)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2308)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:405)
	at java.desktop/javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:262)
	at java.desktop/javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:279)
	at org.pushingpixels.substance.internal.utils.RolloverButtonListener.mouseReleased(RolloverButtonListener.java:108)
	at java.desktop/java.awt.Component.processMouseEvent(Component.java:6632)
	at java.desktop/javax.swing.JComponent.processMouseEvent(JComponent.java:3342)
	at java.desktop/java.awt.Component.processEvent(Component.java:6397)
	at java.desktop/java.awt.Container.processEvent(Container.java:2263)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:5008)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2321)
	at java.desktop/java.awt.Component.dispatchEvent(Component.java:4840)
[...]

PR #6580, opened 2020-06-02
Fix NPE in auto-placement-finder when finding islands

DanVanAtta:
The contract of MapData#getContainedTerritoryPolygons (in javadoc)
states the parameter should be a sea territory. The caller passes
all territories to this method which then results in a NPE in
the method. To fix, relax the contract and return an empty collection
if the parameter is not a sea territory.

Reported in: https://github.com/triplea-game/triplea/issues/6469

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix: #6469 
[] Other:   

## Testing
Repo by:
- unzip a downloaded map
- launch game
- launch map maker tools
- find the auto-placement finder
- enter the name of the unzipped map folder name, eg: "napoleonic_empires-master"
- select 'yes' on all prompts
- observe a NPE (consistent)

After fix verified that the auto-placement-finder could run to completion.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
@@ -515,14 +515,18 @@ public class MapData {
   /**
    * Returns the polygons for any territories contained within a given territory.
    *
-   * @param territoryName Name of the territory, expected to be a sea territory, where we will be
-   *     checking for any contained 'island' territories.
+   * @param territoryName Name of the territory, where we will be checking for any contained
+   *     'island' territories.
+   * @return Empty if the parameter is not a sea territory or contains no 'islands', otherwise
+   *     returns any 'islands' contained within the given territory.
    */
   public Set<Polygon> getContainedTerritoryPolygons(final String territoryName) {
-    return contains.get(territoryName).stream()
-        .map(this::getPolygons)
-        .flatMap(Collection::stream)
-        .collect(Collectors.toSet());
+    return contains.get(territoryName) == null
+        ? Set.of()
+        : contains.get(territoryName).stream()
+            .map(this::getPolygons)
+            .flatMap(Collection::stream)
+            .collect(Collectors.toSet());
   }
 
   public void verify(final GameData data) {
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
```diff
@@ -515,14 +515,18 @@ public Color impassableColor() {
...
    *
-   * @param territoryName Name of the territory, expected to be a sea territory, where we will be
-   *     checking for any contained 'island' territories.
+   * @param territoryName Name of the territory, where we will be checking for any contained
+   *     'island' territories.
+   * @return Empty if the parameter is not a sea territory or contains no 'islands', otherwise
+   *     returns any 'islands' contained within the given territory.
    */
   public Set<Polygon> getContainedTerritoryPolygons(final String territoryName) {
-    return contains.get(territoryName).stream()
```
RoiEXLab, line 522: What about `contains.getOrDefault(territoryName, Set.of())`?

DanVanAtta: I'm really on the fence. I like that it is more direct, but the later operations are what give me trouble. Specifically I don't like how it's kind of a 'hack' where a reader has to see "okay, in this case the set is empty, so we then get no-op, no-op, no-op, okay, we returned an empty set". Having it be more direct I think is more clear. IE: "if this does not contain, then we return empty set, otherwise we take the set, transform it, and return that".

DanVanAtta: It is between:
```
    return !contains.containsKey(territoryName)
        ? Set.of()
        : contains.get(territoryName).stream()
            .map(this::getPolygons)
            .flatMap(Collection::stream)
            .collect(Collectors.toSet());
```

Or the getDefault with no-ops.

DanVanAtta: Went with the 'getOrDefault' approach: 0edd2bc50

The `!contains` was also problematic as well.

PR #6582, opened 2020-06-02
Performance cleanup: Replace string .trim().length() == 0 checks with isBlank()

DanVanAtta:
Per Codacy, it's more efficient to use an 'isBlank' check:
> String.trim().length() == 0 (or String.trim().isEmpty() for the same reason) is an inefficient way to check if a String is really blank, as it creates a new String object just to check its size.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:   

## Additional Review Notes
Please pay extra attention that I got the "not" is blank checks correct and did not flip any conditions.

### DIFF
--- a/.github/workflows/auto-format.yml
+++ b/.github/workflows/auto-format.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
     - name: Set up JDK 11
       uses: actions/setup-java@v1
       with:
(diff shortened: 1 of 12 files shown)


DanVanAtta:
Thank you for fixing @RoiEXLab  :+1:

RoiEXLab:
No problem, turns out the newer version of the commit action was changed to avoid this exact problem.
From what I read it turned out they used to use `find` to locate all of the files to be added, but now they just use built-in git functionality instead, which ignores untracked files if not specified explicitly.
So all I had to do is bump the version

PR #6585, opened 2020-06-02
Update triplea_maps.yaml

mongoose1130:
Update 1.3 for Twelve-Clans map

## Functional Changes

[] New map or map update
[X] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2102,7 +2102,7 @@
     <br> This is the Charles de Gaulle 1939 Mod for the "Global 40" game.
     <br>
 - mapName: Twelve-Clans
-  mapCategory: GOOD
+  mapCategory: EXPERIMENTAL
   url: https://github.com/triplea-maps/twelve-clans/archive/master.zip
   version: 4
   img: https://raw.githubusercontent.com/triplea-maps/twelve-clans/master/description/MapDesc.png
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2102,13 +2102,13 @@
     <br> This is the Charles de Gaulle 1939 Mod for the "Global 40" game.
     <br>
 - mapName: Twelve-Clans
-  mapCategory: EXPERIMENTAL
+  mapCategory: GOOD
   url: https://github.com/triplea-maps/twelve-clans/archive/master.zip
-  version: 3
+  version: 4
   img: https://raw.githubusercontent.com/triplea-maps/twelve-clans/master/description/MapDesc.png
   description: |
     <br>by Michael Hoover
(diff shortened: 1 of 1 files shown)


mongoose1130:
@RoiEXLab No problem, I wasn't sure about it either

Cernelius:
@RoiEXLab I see this as an inherently contradictory, or even partially broken, process, overall.

In the moment one updates a map, since the update is immediately available for download, in substitution to what was offered before, it makes the only sense (to me) that the update to the `triplea_maps.yaml` is merged as soon as possible (perfection would be that the yaml update and the map update are both merged exactly at the same moment), as, meanwhile, people will fail to be informed that the map has been updated or, if they update it meanwhile, will, then, be told that they must update a map they have already at its currently available state.

If one would have to judge upon a request of upgrade or downgrade for the category the map belongs to, which has to take some time, easily several weeks (to decide how good or bad a map is), then, that would work only if the process would have the map itself not updated until the moment in which the request is accepted, wouldn't it? But this is not how it works, isn't it?

Also, if I own a map in High Quality and, without even pushing any `triplea_maps.yaml`, I just modify it into the worst quality map ever, that is becoming available immediately to everyone, isn't it? If it is, in the moment part of the process is immediate and dependent on the map owner only, shouldn't its necessary complement be immediate too and sharing the same dependencies?

Or am I missing or misunderstanding something in the process?

RoiEXLab:
I wouldn't call it contradictory, but I'd agree on the broken part:
Basically incrementing the number only notifies clients that there's a newer version available, but it actually becomes available as soon as you update your master branch on the repository, which is less than ideal.
I once tried to "freeze" the map version listed (which is possible using because of git), but unfortunately it's not backwards compatible at all, we'd have to change the map zip layout again which is already a mess (you can have a folder, a zip file, a zip file according to github etc.) which can all interfere with each other.
We could of course implement this all and add a new property to maps that has this frozen version and keep the current one for backwards compatibility, but this still wouldn't solve the compatibility problem we have. Old version would still try to upgrade to map versions that are completely incompatible.
We could also solve this problem eventually by manually tracking the changes did to each map and figuring out which engine versions are compatible somehow but then there are maps like TWW where in a single zip package there are 2 variants of the same map but for different versions and we haven't even talked about map skins yet.

IIRC @DanVanAtta and I eventually agreed that the way maps currently work is fundamentally broken. The long term goal is that eventually maps will be managed by the "lobby"/Spitfire-server software so that map makers have a convenient way to publish their maps for the community to try out and we don't have this inconvenient process anymore, but right now that's just a general idea

Issue #6101, opened 2020-03-27
Air vs subs - do not offer option to submerge/remain if submerge is automatic


DanVanAtta:
**How can the problem be recreated?**
- send fighters vs a navy containing at least one sub
- notice there is an option to submerge or remain
- subs will always automatically submerge

**When the problem is fixed, what should have happened instead?**
- do not offer choice for subs to submerge, just submerge them.

panther2:
Why do they even have to submerge? In all rulesets where they can't hit air units and cannot be hit by air units they don't even need to submerge.

Or is this just the program logic to keep them off the battle?

Cernelius:
Yeah, assuming that by "send fighters" you mean "**send fighters and nothing else into a hostile sea zone without units of yours already in there**", as @panther2 says, it's not like submarines submerge automatically, it is they cannot be hit by air and they cannot hit air, thus they remain there doing nothing, if not submerged, which they don't have to and it's irrelevant (if it happens that only air remain in offence and only submarines remain in defence, then the battle ends immediately, for both sides having no possible targets, **practically just like the 0 vs 0 power case**, in this case, if the submarines never decided to submerge).

PR #6593, opened 2020-06-04
Move SubmergeVsAir and AirVsNonSubs steps to individual BattleStep classes

trevan:
This shows the new BattleStep classes and refactoring three steps.

Each BattleStep has 0 or more names and 0 or 1 executables.  Of the three steps that have been converted, one has a name and an executable, the other two have no executables.

Each BattleStep has a `valid` function that checks if the step is needed according to the battle state.  This is used when determining if the names should be grabbed, if the executable should be added to the stack, or if the executable should be executed.

I've created two new interfaces: `BattleActions` and `BattleState`.  The `MustFightBattle` implements both of them.  The `BattleActions` will only encompass methods that interact with the `IBridgeDelegate`.  `BattleState` will only encompass methods that interact with internal data structures of `MustFightBattle`.  These interfaces will grow as more steps are converted.  Once all the steps are converted, it might be good to look into splitting the logic behind these interfaces into separate classes instead of `MustFightBattle`.  The interfaces will also help in converting the other battle classes.

I haven't yet built a factory class that generates the steps in the correct order.  I'm not sure how to implement it with only some of the steps converted.  I'll build the factory class once I've converted more steps.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
No manual testing

## Additional Review Notes

I'm also fixing what I think is two bugs.

In the original code, the step `SUBMERGE_SUBS_VS_AIR_ONLY` was shown in the battle UI after the subs fired step.  But the actual step was executed before the subs fired.  So you would have a battle ui that said "Subs fire", "Remove Sub Casualties", "Subs Submerge vs Air".  I think the order is wrong.  Also, the "subs fire", "remove sub casualties" probably shouldn't show up but that's another step that I'll work on.

Also, it only showed the `SUBMERGE_SUBS_VS_AIR_ONLY` if the attacker was all airplanes and the defender was subs.  But if the attacker was all subs and the defender was all airplanes, you'd still submerge.  You just wouldn't see that in the battle ui.  I changed that as well.


trevan:
I've added comments and made a few changes.  Biggest change is the new battle step classes are moved to subpackages.  I had started making a few more battle steps on a different branch and the number of classes in the steps package was getting a little large.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
```diff
@@ -0,0 +1,99 @@
...
+  protected abstract void execute(ExecutionStack stack, IDelegateBridge bridge);
+
+  /**
+   * This is used to break up the battle into separate atomic pieces. If there is a network error,
+   * or some other unfortunate event, then we need to keep track of what pieces we have executed,
+   * and what is left to do. Each atomic step is in its own BattleAtomic with the definition of
+   * atomic is that either:
+   *
+   * <ol>
+   *   <li>The code does not call to an IDisplay, IPlayer, or IRandomSource
```
trevan, line 77: I copied this comment from `MustFightBattle#getBattleExecutables` but I think the comment might be outdated.  I think it should reference `IDelegateBridge` instead of `IDisplay, IPlayer, or IRandomSource`.  The only way I see steps can talk to `IDisplay, IPlayer, or IRandomSource` is by going through `IDelegateBridge`.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
```diff
@@ -44,9 +55,14 @@
   final @NonNull Supplier<Collection<Territory>> getAttackerRetreatTerritories;
   final @NonNull BiFunction<GamePlayer, Collection<Unit>, Collection<Territory>>
       getEmptyOrFriendlySeaNeighbors;
+  final @NonNull BattleActions battleActions;
 
   public List<String> get() {
 
+    final BattleStep submergeSubsVsOnlyAir = new SubmergeSubsVsOnlyAirStep(this, battleActions);
+    final BattleStep airAttackVsNonSubs = new AirAttackVsNonSubsStep(this, battleActions);
+    final BattleStep airDefendVsNonSubs = new AirDefendVsNonSubsStep(this, battleActions);
```
trevan, line 64: As I mentioned in the first comment, these manual instantiations of the steps will be moved into a step factory that builds the required steps and puts them in the correct order.  But I can't build that factory until I have a few more steps converted.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
```diff
@@ -1603,68 +1602,4 @@ void firstStrikeDefenderReturnFireAttNoDestroyerDefNoDestroyerWW2v2FalseSneakAtt
         MustFightBattle.ReturnFire.ALL,
         false);
   }
-
-  @Test
-  @DisplayName("Verify attacking canNotBeTargetedByAll can submerge if all is air")
```
trevan, line 1608: These two tests are now being done in `SubmergeSubsVsOnlyAirStepTest#testSubmerging`.  I considered leaving these tests in till after this is merged just to show that the logic is still correct.  Should I put them back in or is it ok to delete them now?

trevan:
This breaks save game compatibility.  It is because I moved the SubmergeSubsVsOnlyAir IExecutable from MustFightBattle to SubmergeSubsVsOnlyAirStep.

trevan:
I was able to figure out a way to handle saved game compatibility.

panther2:
Just want to point to #6101. Is that related?

trevan:
> Just want to point to #6101. Is that related?

It is kind of related.  The step that causes the auto-submerge is the same step that I'm converting.  The step to query the submerge is a different step that I haven't yet converted.

This change won't fix that issue, though.  The only thing that is changing (besides the code refactoring) is for the name of the step to show up in the UI when the step actually happens.  It currently shows the name in the wrong location and it doesn't always show the name even though the step will happen.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * The current state of the battle
+   *
+   * <p>Note: This is current mutable so the underlying data can change.
+   */
+  protected final BattleState battleState;
+
+  /** Actions that can occur in a battle that require interaction with {@link IDelegateBridge} */
+  protected final BattleActions battleActions;
```
DanVanAtta, line 36: Inheritance is a very strong form of coupling, using `protected` variables is also potentially very painful as well when considering inheritance hierarchies. In some cases as well some implementations may not need all of the protected variables.

What are your thoughts to convert this class to an interface? In such a case, it seems like we only need to move these two variables to the child classes. The level of code duplication is pretty negligable since they already have the all arg constructors and in turn each of the child classes would be much more strongly encapsulated. WDYT, any concerns with that suggestion?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
```diff
@@ -0,0 +1,88 @@
...
+
+  public abstract BattleAtomic getExecutable();
+
+  public abstract List<String> getNames();
```
DanVanAtta, line 40: A javadoc on this method could be of some use. It's a bit odd to think that a step would have multiple names. It'd be nice to know what it means to have multiple names for a step and secondarily if this is for display purposes only.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
```diff
@@ -0,0 +1,88 @@
...
+   * Determine if this step should run based on the request
+   *
+   * @return true if valid
+   */
+  public abstract boolean valid();
+
+  /**
+   * Executes the step
+   *
+   * <p>This is called by the BattleAtomic and {@link #valid} has already been checked
```
DanVanAtta, line 52: Typically methods should not be aware of who is calling them. It's also a design smell when we require one method to have been called before another (and furthermore when that method is determining state that dictates whether you can and should call another method).

That could potentially be fixed by returning an object to then be executed. EG:

```
protected Optional<BiConsumer<ExecutionStack, IDelegateBridge>> getBattleStepExecutable();
```

Then usage below of:
```
      final BattleStep executingStep = BattleStep.this;
      if (executingStep.valid()) {
        executingStep.execute(stack, bridge);
      }
```

would be transformed into:
```
       getBattleStepExecutable().ifPresent(stepExecutable -> stepExecutable.accept(stack, bridge));
  }
```

It also appears that `getName()` has a dependency on `isValid()` as well. In those cases if `isValid()` is false then presumably `getNames()` could return an empty collection.

WDYT?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
```diff
@@ -0,0 +1,88 @@
...
+   * atomic is that either:
+   *
+   * <ol>
+   *   <li>The code does not call to an IDisplay, IPlayer, or IRandomSource
+   *   <li>If the code calls to an IDisplay, IPlayer, IRandomSource, and an exception is called from
+   *       one of those methods, the exception will be propagated out of execute() and the execute
+   *       method can be called again.
+   * </ol>
+   */
+  public abstract class BattleAtomic implements IExecutable {
```
DanVanAtta, line 72: Is this an inner class only because of `IExecutable`?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
DanVanAtta: Probably might as well update it so it makes sense. Won't make any more sense to anyone else at this point, and you've about as much context into this as anyone would have after the work you've done here. So if it does not make sense to you now, it probably won't make sense to anyone else (in which case the documentation is not very useful).

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/air/AirVsNonSubsStep.java
```diff
@@ -0,0 +1,33 @@
...
+/** Air can not attack subs unless a destroyer is present */
+public abstract class AirVsNonSubsStep extends BattleStep {
+
+  public AirVsNonSubsStep(final BattleState battleState, final BattleActions battleActions) {
+    super(battleState, battleActions);
+  }
+
+  @Override
+  public BattleAtomic getExecutable() {
+    return null;
```
DanVanAtta, line 21: nit, if the return value is nullable, at the very least the method should be marked as `@Nullable`. Perhaps better though if this returns `Optional<BattleAtomic>` so that any caller won't mistakenly think the return value is non-nullable and run into a NPE.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/sub/SubmergeSubsVsOnlyAirStep.java
```diff
@@ -0,0 +1,72 @@
...
+  @Override
+  public BattleAtomic getExecutable() {
+    return new BattleAtomic() {
+      private static final long serialVersionUID = 99990L;
+    };
+  }
+
+  @Override
+  public List<String> getNames() {
+    return List.of(SUBMERGE_SUBS_VS_AIR_ONLY);
```
DanVanAtta, line 37: If we always return 0 or 1 element, would `Optional<String>` be a better return API? In that case there would not be any confusion that we can have more than one element and it'd be clear we'd have either zero or one.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MockBattleState.java
```diff
@@ -0,0 +1,34 @@
...
+
+  final @NonNull Collection<Unit> attackingUnits;
+  final @NonNull Collection<Unit> defendingUnits;
+
+  public static MockBattleState.MockBattleStateBuilder givenBattleState() {
+    return MockBattleState.builder().attackingUnits(List.of()).defendingUnits(List.of());
+  }
+
+  @Override
+  public Collection<Unit> getAttackingUnits() {
```
DanVanAtta, line 26: Minor: I noticed you used lombok getters on another class, why not here too?

game-core/src/test/java/games/strategy/triplea/delegate/battle/MockBattleState.java
```diff
@@ -0,0 +1,34 @@
...
+ * <p>{@link #givenBattleState()} will return a builder with everything defaulted and the test can
+ * override the specific items needed.
+ */
+@Builder
+public class MockBattleState implements BattleState {
+
+  final @NonNull Collection<Unit> attackingUnits;
+  final @NonNull Collection<Unit> defendingUnits;
+
+  public static MockBattleState.MockBattleStateBuilder givenBattleState() {
```
DanVanAtta, line 21: It looks like all usages of this method always define attacking unit and defending units, why not just use the builder directly?

IE: instead of: ```MockBattleState.givenBattleState()```, use instead: ```MockBattleState.builder()```

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
```diff
@@ -31,8 +37,13 @@
   final @NonNull GamePlayer defender;
   final @NonNull Collection<Unit> offensiveAa;
   final @NonNull Collection<Unit> defendingAa;
+
+  @Getter(onMethod = @__({@Override}))
   final @NonNull Collection<Unit> attackingUnits;
```
DanVanAtta, line 42: Attacking and defending units being together pretty much always looks like the [data clump](https://en.wikipedia.org/wiki/Data_clump) code smell. It appears that `BattleState` could instead become a first class object and be a relatively simple value object. 

It might make sense to rename `BattleState` to `UnitsInBattle` or `BattleUnits`, or `BattleParticipants` as we are not capturing the full battle state but just the units in battle.


If serialization were not a concern, then we could replace both attacking and defending units with a `BattleState`. To fix the data clump code smell we could:
- add a transient `BattleState` object. Whenever we assign a battle state object then we write the corresponding values to attacking and defending units. This way we would get serialization and any usages of the data would be through battle state. It might be possible as well to override the serialization write-object method and to serialize attacking units and defending units through a battle state. Serialization proxy might be a way to do that, though the proxy itself could break compatibility alone.

Fundamentally `BattleState` is not really an interface, it's more data that can be present in an object and I suspect it makes sense to simply have a reference to it.

WDYT?

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
```diff
@@ -0,0 +1,12 @@
+package games.strategy.triplea.delegate.battle;
+
+import games.strategy.engine.data.Unit;
+import java.util.Collection;
+
+/** Exposes the battle state and allows updates to it */
```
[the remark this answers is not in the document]
trevan: I plan to add those.  My plan is to add the existing methods from `MustFightBattle` that are needed in the steps.  As I convert each step, I'll be adding more and more methods to the two interfaces.  Some of these methods will be mutator methods.  I don't plan on renaming them at that point, though.

Then, I plan on splitting up methods that have functionality that should be in `BattleActions` (IDelegateBridge interaction) and `BattleState`.  An example of this is `MustFightBattle#queryRetreat`.

Once I have the methods split between the two interfaces, I'll look into refactoring them so that they have a targeted focus.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/sub/SubmergeSubsVsOnlyAirStep.java
trevan: No, there are steps that will have more than 1.  An example is the step that fires aa guns.  The step will return "attacker fires guns", "defender selects casualties", and "defender removes casualties".

Now, it is possible after further refactoring, that step could be split up into three separate steps.  But I'm not sure at this point.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MockBattleState.java
trevan: Because some of the steps will require more items added.  An example is the aa guns.  They don't care about offensive/defensiveUnits but use the offensive/defensive aa.

I'm actually thinking that all the parameters for `BattleState` is going to end up looking like the parameters to `BattleSteps`.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
trevan: I agree.  It's what I'd like to get to.  Once the `BattleState` interface has been fully built, I'll then know what is all the parts of `MustFightBattle` that is part of the state and I'd like to make a value object that holds that data.  I'd also like to convert the attacking/defending units into `FiringGroups` (from my older PR that I closed).  Then it doesn't need separate attacking/defending units, attacking/defending aa, and bombarding units.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
trevan: No, it needs a reference to the BattleStep that is enclosing it so that it can then call `valid` on it and `execute`.

I could change it to have a instance variable that is the BattleStep and each step has to construct it and pass in itself.  The `execute` method on `BattleStep` will need to made public as well.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
trevan: The explanation about name is on the class.  I'll move it down to here.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
trevan: Before the conversion to use `MustFightBattle` as the `BattleState` there was more of a reason.  I can't think of a good reason to keep it inherited so I'll change it to an interface..

trevan:
@DanVanAtta I think some of your comments are mostly caused because this is only converting a few steps.  I didn't go through the process of figuring out exactly all of the methods and such that are needed in `BattleState` and `BattleActions` because they would be unused at this point in time.  I wanted to keep this small with just a few steps to show it in action.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MockBattleState.java
trevan: It's because I'm not accustomed to using lombok.  So, I'll remember to use it sometimes but then revert to my usual ways of not using it.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
trevan: `getBattleStepExecutable` actually returns the `BattleAtomic`.  So calling `getBattleStepExecutable` inside of the `BattleAtomic` will not work since it will just go into an infinite loop.

I see what you are trying to do but I don't see how that helps.  You are just moving the `valid()` call to another method which is still making `execute()` be dependent on `valid()`.  I had made `execute()` a protected method so that the dependency on `valid()` was controlled by the class and callers wouldn't have to worry about it.  It was an internal implementation.  Though, with your idea of turning this class into an interface, that internal implementation becomes public, so it complicates it.  I guess implementations of BattleStep will just have to ensure that `execute()` checks `isValid()` beforehand.

As for `getNames()`, I can see why you say it has a dependency.  The only caller of it is always checking `isValid()`.  I can envision other callers not caring if it is valid or not but I currently don't envision creating those callers so I can move `isValid()` inside of each of the `getNames()`.

trevan:
@DanVanAtta I've pushed changes from your feedback.  By moving the `valid()` calls into `getNames()` and `execute()`, I realized I no longer needed the `BattleAtomic` class and that the `BattleStep` could just implement`IExecutable` itself.  So `getExecutable()` is now gone and `execute()` is public per the `IExecutable` contract.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/air/AirAttackVsNonSubsStep.java
```diff
@@ -2,18 +2,20 @@
...
 public class AirAttackVsNonSubsStep extends AirVsNonSubsStep {
-  public AirAttackVsNonSubsStep(final BattleState battleState, final BattleActions battleActions) {
-    super(battleState, battleActions);
+  public AirAttackVsNonSubsStep(final BattleState battleState) {
+    super(battleState);
   }
 
   @Override
   public List<String> getNames() {
+    if (!valid()) {
```
DanVanAtta, line 16: nit, can use a ternary here to be a bit more concise. It's generally a bit easier (lower cognitive load) for if statements to lead with the 'true' value.

IE:
```
return valid()
   ? List.of(AIR_ATTACK_NON_SUBS)
   : List.of();
```

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/air/AirDefendVsNonSubsStepTest.java
```diff
@@ -6,67 +6,59 @@
...
-            .attackingUnits(List.of(defender1, defender2))
-            .defendingUnits(List.of(attacker))
-            .build();
-    final AirAttackVsNonSubsStep underTest = new AirAttackVsNonSubsStep(battleState);
-    assertThat(underTest.valid(), is(false));
+  static List<Arguments> testWhatIsValid() {
+    return List.of(
+        Arguments.of(
+            "Defender has air units and no destroyers vs Attacker subs",
+            givenBattleState()
```
DanVanAtta, line 43: The givenBattleState is still odd, it probably should at least return a builder().

You can make list values non-file and use @Builder.Default to initialize them to empty lists.
Also consider using `@Singular` which I think will default the value to an empty list if not specified. Using `@Singular` on the attacking and defending units properties, you could instead write:
```
MockBattleState.builder()
   .attackingUnit(givenUnitCanNotBeTargetedBy(mock(UnitType.class)))
   .defendingUnit(givenUnit())
   .defendingUnit(givenUnitIsAir())
   .build();
```

game-core/src/test/java/games/strategy/triplea/delegate/battle/MockBattleState.java
DanVanAtta: My point is that you don't need to use a builder to init the values to empty lists and then return that builder to overwrite the values. You can use builder directly I would think. Look into using `@Singular` on those: https://projectlombok.org/features/Builder#singular

game-core/src/test/java/games/strategy/triplea/delegate/battle/MockBattleState.java
```diff
@@ -0,0 +1,28 @@
...
+import lombok.NonNull;
+
+/**
+ * Simple implementation of BattleState for tests to use
+ *
+ * <p>{@link #givenBattleState()} will return a builder with everything defaulted and the test can
+ * override the specific items needed.
+ */
+@Builder
+public class MockBattleState implements BattleState {
```
DanVanAtta, line 17: Nit, this is not really a Mock. A stub is something that returns the same value (the stub controls the return value), a mock is something that returns the value you tell it to (test controls the return value), and a 'fake' is a simplified implementation (aka, a test double).

In this case it looks like we basically have a real implementation. It's tempting to call this a test-fake, but given the API is so simple it looks to be full-blown. 

This makes me wonder if we should have a `DefaultBattleState` on the `BattleState` interface itself.  Essentially in `BattleState` declaring a public static class `DefaultBattleState` that would be this class. You've mentioned though you expect this class to grow and perhaps then we'll want to use a true-mock.

In the meantime the naming is a bit misleading and I think would be useful to fix. Perhaps we could call this a `DefaultBattleState`, or a `SimpleBattleState`.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/air/AirAttackVsNonSubsStepTest.java
```diff
@@ -0,0 +1,64 @@
...
+      assertThat(underTest.getNames(), hasSize(0));
+    }
+  }
+
+  static List<Arguments> testWhatIsValid() {
+    return List.of(
+        Arguments.of(
+            "Attacker has air units and no destroyers vs Defender subs",
+            givenBattleState()
+                .attackingUnits(List.of(givenUnit(), givenUnitIsAir()))
```
DanVanAtta, line 44: Nit, most given methods should take a parameter. Something like:
```
givenUnitWithAttackPower(2)
```
If there is no parameter, and the return value is very generic, then something like the following would be typical (note the 'any' word):
```
givenAnyUnit()
```

Otherwise the given should probably spell out what is special about the thing that is being returned, you've done that in the example of: `givenUnitIsAir()`

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/BattleStepsTest.java
```diff
@@ -139,6 +145,19 @@ public static Unit givenUnitAttackerFirstStrikeCanNotBeTargetedBy(final UnitType
...
+
+  public static Unit givenUnitCanNotBeTargetedBy(final UnitType otherType) {
+    final Tuple<Unit, UnitAttachment> unitAndAttachment = newUnitAndAttachment();
+    when(unitAndAttachment.getSecond().getCanNotBeTargetedBy()).thenReturn(Set.of(otherType));
+    return unitAndAttachment.getFirst();
+  }
+
   public static Unit givenUnitDefenderFirstStrike() {
     final UnitType canNotTargetType = mock(UnitType.class);
     final Tuple<Unit, UnitAttachment> unitAndAttachment = newUnitAndAttachment();
```
DanVanAtta, line 163: Side note, if you could follow-up and remove the `Tuple` return value from `newUnitAndAttachment`, it would be appreciated. We really do want to delete that class outright in favor of value objects.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/air/AirAttackVsNonSubsStepTest.java
```diff
@@ -0,0 +1,64 @@
...
+                .defendingUnits(List.of(givenUnitCanNotBeTargetedBy(mock(UnitType.class))))
```
DanVanAtta, line 45: Any thoughts to simplify this helper method from:
```
givenUnitCanNotBeTargetedBy(mock(UnitType.class))
```
To:
```
givenUnitCanNotBeTargetedByAnyUnit();
```

Or am I misunder

DanVanAtta:
@trevan please also have a look at the Codacy violations, you should likely be able to address 2 of the 4:
- Document empty method body
- Consider simply returning the value vs storing it in local variable 'builder'

game-core/src/test/java/games/strategy/triplea/delegate/battle/MockBattleState.java
trevan: I'm trying to build a faked `BattleState` in a fluent manner where the test methods can easily modify the object to get it to the state they need.

With this mechanism, I could have one test that does
```
givenBattleState().attackingUnits(units).defendingUnits(units2).build();
```
while another test could do
```
givenBattleState().bombardingUnits(units).build();
```
and another could do
```
givenBattleState().bombardingUnits(units).battleSite(territory).build();
```

That's why I'm returning the `MockBattleStateBuilder` class instead of the `BattleStateBuilder` class.  I could add `@Builder(toBuilder = true)` and then do
```
givenBattleState().toBuilder().bombardingUnits(units).battleSite(territory).build();
```
but that felt like extra work that I could skip.

I'd like to use `mock()` but I don't see how I can do that in a fluent manner.  Instead, I'd have to do
```
BattleState battleState = givenBattleState();
when(battleState.attackingUnits()).thenReturn(units);
when(battleState.defendingUnits()).thenReturn(units2);
```

I could call it a fake because it will not be a full blown implementation.  Once I start adding in the updater methods, I'll not be implementing them.  The tests will just spy on the BattleState and check that the methods were called correctly.  And if I ever come across a step that updates the state and then reads from the updated state, I'll have rethink this :)  I haven't yet seen one but I haven't fully investigated every single step yet.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/air/AirDefendVsNonSubsStepTest.java
trevan: Thanks for the pointer to `@Singular`

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/air/AirAttackVsNonSubsStepTest.java
trevan: I'm renaming it to `givenUnitCanNotBeTargetedByRandomUnit()`.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/air/AirDefendVsNonSubsStepTest.java
trevan: `@Singular` is causing the `Error Prone static analysis plugin` to throw an exception.  All you need to do is put `@Singular` on one of the fields in `MockBattleState` and then run it.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MockBattleState.java
trevan: I've renamed the class to `FakeBattleState` and the given method to `givenBattleStateBuilder()`.

DanVanAtta: Understood on the fluent API, that is cool :+1: 

One word of caution/note, spying on state changes is generally a test code smell. It's generally better if one module computes the state change and returns the new state value. Then have another module that uses the first one to do the state mutation.

IE:
```

BiFunction<GameData, BattleState, Integer> computationStrategy;
BattleState battleState;

void foo(GameData gameData, BattleState battleState) {
  int newValue = computationStrategy.apply(gameData, battleState);
  battleState.setSomeValue(newValue);
}
```

In this case the test for the above is quite trivial as the computation function can be injected. In TripleA so far the cleanest way I know to do that is to use a static build function, ie:
```
@Builder
class ImperativeShell {
  @Nonnull private final BattleState battleState;
  @Nonnull private final GameData gameData;
  @Nonnull private final BiFunction<BattleState, GAmeDAta, Integer> computationFunction;

   public static ImperativeShell build(GameData gameData, BattleState battleState) {
         return builder()
                      .battleState(battleState)
                      .gameData(gameData)
                      .computationFunction(new RealComputationFunction())
                      .build();
   }
}
```

Then in test you can use the builder() directly to inject mocks.

At the end of this you can then test the 'RealComputationFunction' directly and assert it returns the expected values without having to spy for side-effects. It turns assertions that look like this:

```
computation.apply(battleState);

verify(battleState).unitRemoved(unit);
```

Into:
```
Unit result = computation.apply(battleState);

assertThat(result, is(someExpectedUnit));
```

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/air/AirDefendVsNonSubsStepTest.java
DanVanAtta: Ah yeah, error prone and singular do not play well.  I think the error prone analysis that fails needs to be suppressed for classes that use singular. Perhaps turning the rule off is a better move. IMO `@Singular` can be really useful, error prone is arguably getting in our way with it.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MockBattleState.java
DanVanAtta: Nota bene, if the objects to inject are all unique minus the generic types (recall that in java generic types are dropped during compilation time), then `@InjectMocks` can make a test even easier and automatically link mocks to an `@RequiredsArgsConstructor` (which is created automatically when using `@Builder`), ie:
```
class ImperativeShellTest {
  @Mock private BattleState battleState;
  @Mock private GameData gameData;
  @Mock private BiFunction<BattleState, GAmeDAta, Integer> computationFunction;
   
   @InjectMocks
   private ImperativeShell imperativeShell;
```

PR #6601, opened 2020-06-06
Skip smoke-testing 'test' step when running unit tests

DanVanAtta:
1. We have a dedicated build stage for running the 'tests' in the smoke-testing
  project, it is redundant to run them in the 'unit test' stage. The separation
  is there for performance, so we do not slow down unit tests and to make
  debugging easier in case there is a test failure (with isolated stages,
  if one fails, there are fewer things being run and so there are fewer
  reasons to see a failure)

2. Minor, combine gradle steps together.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:

### DIFF
--- a/.travis/stages/verify/unit-testing/run
+++ b/.travis/stages/verify/unit-testing/run
@@ -3,10 +3,8 @@
 set -eEux
 
 # Run Tests
-./gradlew --parallel test
+./gradlew -x :smoke-testing:test -parallel test jacocoTestReport
 
-# Generate reports
-./gradlew --parallel jacocoTestReport
 
 # upload coverage report - https://github.com/codecov/example-gradle
 bash <(curl -s https://codecov.io/bash)


.travis/stages/verify/unit-testing/run
```diff
@@ -3,10 +3,8 @@
 set -eEux
 
 # Run Tests
-./gradlew --parallel test
```
RoiEXLab, line 6: Does it work if `--parallel` just is written `-parallel` now?

DanVanAtta: FWIW, from the gradle help text:
> --parallel                Build projects in parallel. Gradle will attempt to determine the optimal number of executor threads to use.

A pseudo general convention is flags that are fully spelled out get double dashes with single dash word delimiters, eg: `--first-second-word`, while 'short flags' that are single letters are single dash, eg: `-p`

PR #6602, opened 2020-06-06
Convert link check to only do a HEAD request

DanVanAtta:
If all we need is status code, we can simply do just an HTTP HEAD request
and avoid the heavier call that receives the full HTTP body payload
of an HTTP GET request.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:

### DIFF
--- a/.travis/stages/verify/check-links-and-yaml-syntax/check-links
+++ b/.travis/stages/verify/check-links-and-yaml-syntax/check-links
@@ -17,7 +17,7 @@ fi
 function checkUri() {
   local -r uri="$1"
   local code
-  if ! code=$(curl -Lsf -o /dev/null -w "%{http_code}" "$uri"); then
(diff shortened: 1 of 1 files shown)


DanVanAtta:
Interestingly this fails build. I think the HEAD request is returning a non-zero and then our script is stalling. Perhaps we are grepping on an empy input and are waiting for STDIN. 

@RoiEXLab I figured you might be interested to see this. If you'd like to pick this up so that we can do a HEAD request only for link checking, please feel free. I might be able to return to this in a week or so.

RoiEXLab:
The Problem with HEAD requests although they should behave just like GET is that they often don't.
So if a server on our list is unable to handle a HEAD request correctly we get a change in behaviour, I might be able to have a look at it, but I won't promise anything

DanVanAtta:
No worries then @RoiEXLab , this is not the most important task for us.

PR #6604, opened 2020-06-06
Allow orphaned test check to skip based on @SuppressWarnings("UnmatchedTest")

DanVanAtta:


.travis/stages/verify/find-unused-or-unmatched-tests/find-orphaned-tests
```diff
@@ -56,6 +57,7 @@ if [ "$status" -eq 1 ]; then
   echo " - renaming the test file to end with 'IntegrationTest'"
   echo " - adding an @Integration annotation to the test"
   echo " - adding an @ExtendedUnitTest annotation to the test"
+  echo " - adding an @SuppressWarngins(\"UnmatchedTest\") to the test"
```
trevan, line 60: Looks like you have a typo here.

DanVanAtta: Good catch! Certainly would be useful to have something that is easily copy/pasted.

Fixed in: cb12baa36

PR #6608, opened 2020-06-06
Server options password as char array

DanVanAtta:
## Commits

commit 1a310bae7753843d7c68bbfc51d4fe24e4ed62bc

    Convert String.trim().isEmpty() check to a 'isBlank()'

commit 9ea89afc6d46d54b8e43e5dd2f713adbbe2090b3

    Use char[] for password in ServerOptions to reduce conversions to String
    
    Reasons:
    1. We get fewer conversions to strings, yields simpler code
    2. While we do not require passwords to be handled as char[], we should
    try to avoid using Strings for password where conveniently possible.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
- launched a locally hosted game with a password.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
@@ -351,7 +351,10 @@ public class ServerModel extends Observable implements IConnectionChangeListener
           ServerConnectionProps.builder()
               .name(System.getProperty(TRIPLEA_NAME))
               .port(Integer.parseInt(System.getProperty(TRIPLEA_PORT)))
-              .password(System.getProperty(SERVER_PASSWORD))
+              .password(
+                  Optional.ofNullable(System.getProperty(SERVER_PASSWORD))
+                      .map(String::toCharArray)
+                      .orElse(null))
               .build());
     }
     final UserName userName = UserName.of(ClientSetting.playerName.getValueOrThrow());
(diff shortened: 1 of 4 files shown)


game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
```diff
@@ -351,7 +351,10 @@ private void gameDataChanged() {
           ServerConnectionProps.builder()
               .name(System.getProperty(TRIPLEA_NAME))
               .port(Integer.parseInt(System.getProperty(TRIPLEA_PORT)))
-              .password(System.getProperty(SERVER_PASSWORD))
+              .password(
+                  Optional.ofNullable(System.getProperty(SERVER_PASSWORD))
+                      .map(String::toCharArray)
+                      .orElse(null))
```
RoiEXLab, line 357: Not sure if the benefit is that great if we don't ever clear the char array afterwards

DanVanAtta: Agree. On one hand GC should kick in and at least destroy the array and at some point the memory would be overwritten.

For the most part it's a "use char[] where convenient", so don't go out of our way to use String. In this case overall I think/hope there is less conversion to string, ultimately simpler. To another extent, because we are reading password in from CLI args, coming in as a `String[]`, we are pretty boned in terms of the password being stored as a `String`

Issue #6607, opened 2020-06-06
Travis Master Branch Build Failed


tripleabuilderbot:
Travis master branch build [#19758](https://travis-ci.com/triplea-game/triplea/builds/170116968) failed.     Build stage: prerelease deployment

tripleabuilderbot:
Build [#19758](https://travis-ci.com/triplea-game/triplea/builds/170116968) failed on stage: end of script failure catch block

tripleabuilderbot:
Build [#19762](https://travis-ci.com/triplea-game/triplea/builds/170117403) failed on stage: build installers

tripleabuilderbot:
Build [#19762](https://travis-ci.com/triplea-game/triplea/builds/170117403) failed on stage: end of script failure catch block

tripleabuilderbot:
Build [#19768](https://travis-ci.com/triplea-game/triplea/builds/170117701) failed on stage: prod2 deployment

tripleabuilderbot:
Build [#19768](https://travis-ci.com/triplea-game/triplea/builds/170117701) failed on stage: end of script failure catch block

tripleabuilderbot:
Build [#19780](https://travis-ci.com/triplea-game/triplea/builds/170120083) failed on stage: prerelease deployment

tripleabuilderbot:
Build [#19780](https://travis-ci.com/triplea-game/triplea/builds/170120083) failed on stage: end of script failure catch block

tripleabuilderbot:
Build [#19794](https://travis-ci.com/triplea-game/triplea/builds/170121164) failed on stage: prod2 deployment

tripleabuilderbot:
Build [#19794](https://travis-ci.com/triplea-game/triplea/builds/170121164) failed on stage: end of script failure catch block

tripleabuilderbot:
Build [#19800](https://travis-ci.com/triplea-game/triplea/builds/170124870) failed on stage: prerelease deployment

tripleabuilderbot:
Build [#19800](https://travis-ci.com/triplea-game/triplea/builds/170124870) failed on stage: end of script failure catch block

tripleabuilderbot:
Build [#19812](https://travis-ci.com/triplea-game/triplea/builds/170125398) succeeded.

PR #6609, opened 2020-06-06
Remove end of script travis failure catch block

DanVanAtta:
So far it looks like we will consistently report a failure if any step
fails. This update removes a travis catch block that triggers on failure
and also reported a failure. This should simplify the messaging we get
in any 'master branch build failed' issue and list a failure just once.

A risk from this update is if we get a failure in an unexpected place
that is not explicitly handled.

-->

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Problem fix:  
[] Other:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -111,7 +111,3 @@ jobs:
       if: (branch = master) and (repo = 'triplea-game/triplea') and (type != 'pull_request')
       script: [ ".travis/shared/report-build-status SUCCESS" ] 
 
-after_failure:
-  - echo "================ Build step 'after_failure' =================" > /dev/null
-  - $REPORT_FAIL "end of script failure catch block"
-


tvleavitt:
Can we not tell if a failure has already been reported, and suppress the
duplicative message if it has? What's the process for detecting if the
build is silently failing?

I agree that the waterfall of messages is unproductive.

Thomas Leavitt

DanVanAtta:
You can see an example of the "duplicative" messages here: https://github.com/triplea-game/triplea/issues/6607  (just notice there is a message for the section that failed and a generic "end of script" block that catches the failure).

If we don't catch a failure in the travis config script then we can get a false positive. Come to think of it, the deployment of artifacts to releases is more or less built-in, I think that might be the place where we could see such a false positive.

If you check the [travis.yml config file](https://github.com/triplea-game/triplea/blob/master/.travis.yml), you can see where we catch failures. Notice the `|| $REPORT_FAIL` that suffixes each command.

DanVanAtta:
@tvleavitt to answer more directly:

> Can we not tell if a failure has already been reported, and suppress the
duplicative message if it has?

The report failure script will look for an open issue and append to it instead of creating a new issue. 

> What's the process for detecting if the build is silently failing?

There basically is none to some extent. Travis sends us an email for a broken build that can be hard to notice. The issue gives us a main location to discuss build failures and to communicate more broadly about build failures.

Issue #6618, opened 2020-06-07
2.0.19579: MoveValidator#addToMapping:1579 - UnsupportedOperationException


tripleabuilderbot:
## User Description
Playing round 4 on America's turn, and when I click on 7 Sea Zone I get the error. "UnsupportedOperationException".

## TripleA Version
2.0.19579

## Java Version
11.0.6

## Operating System
Windows 10

## Exception 
java.lang.UnsupportedOperationException

## Stack Trace
```
java.lang.UnsupportedOperationException
	at java.base/java.util.Collections$UnmodifiableCollection.addAll(Collections.java:1070)
	at games.strategy.triplea.delegate.MoveValidator.addToMapping(MoveValidator.java:1579)
	at games.strategy.triplea.delegate.MoveValidator.mustMoveWith(MoveValidator.java:1559)
	at games.strategy.triplea.delegate.MoveValidator.getMustMoveWith(MoveValidator.java:1540)
	at games.strategy.triplea.ui.panel.move.MovePanel.setFirstSelectedTerritory(MovePanel.java:1450)
	at games.strategy.triplea.ui.panel.move.MovePanel$1.selectUnitsToMove(MovePanel.java:234)
	at games.strategy.triplea.ui.panel.move.MovePanel$1.unitsSelected(MovePanel.java:143)
	at games.strategy.triplea.ui.panels.map.MapPanel.notifyUnitSelected(MapPanel.java:519)
	at games.strategy.triplea.ui.panels.map.MapPanel$3.mouseReleased(MapPanel.java:230)
	at java.desktop/java.awt.AWTEventMulticaster.mouseReleased(AWTEventMulticaster.java:298)
	at java.desktop/java.awt.Component.processMouseEvent(Component.java:6632)
	at java.desktop/javax.swing.JComponent.processMouseEvent(JComponent.java:3342)
	at java.desktop/java.awt.Component.processEvent(Component.java:6397)
	at java.desktop/java.awt.Container.processEvent(Container.java:2263)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:5008)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2321)
	at java.desktop/java.awt.Component.dispatchEvent(Component.java:4840)
	at java.desktop/java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4918)
	at java.desktop/java.awt.LightweightDispatcher.processMouseEvent(Container.java:4547)
[...]

PR #6626, opened 2020-06-07
Fix exception w/ UnmodifiableCollecation and simplify code.

asvitkine:
Fix exception w/ UnmodifiableCollecation and simplify code.

We create an ArrayList<>() before calling addAll() on it.

Fixes https://github.com/triplea-game/triplea/issues/6618

Also simplifies the code:
  - The mapping.isEmpty() case is moved into addToMapping().
  - The !newMapping.isEmpty() check is removed - iterating over an empty map is fine.
  - The new HashMap<>() for newMapping calls are removed, as they're not needed (we just iterate over the temporary, so no need to copy it.)

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix:  https://github.com/triplea-game/triplea/issues/6618
[] Other:   

## Testing
Sanity testing of an all Hard AI game.


DanVanAtta:
:+1:

Issue #6632, opened 2020-06-07
2.0 Bots are Running out of Memory (Map Parsing Related)


DanVanAtta:
Bots are exhausting heap space on launch. Looking at the stack trace below it looks to be map related. The way map XMLs are structured perhaps have hit a breaking point for data storage inefficiency : (

In the short term we need to determine which map is triggering this and exclude it from bots until we determine a way to fix the root cause.

```
Jun 07 21:04:05 li1116-193 systemd[1]: Started TripleA Bot 01.
Jun 07 21:08:22 li1116-193 run_server[324823]: Exception in thread "main" java.lang.OutOfMemoryError: Java heap space
Jun 07 21:08:22 li1116-193 run_server[324823]:         at java.base/java.util.HashMap.newNode(HashMap.java:1797)
Jun 07 21:08:22 li1116-193 run_server[324823]:         at java.base/java.util.HashMap.putVal(HashMap.java:637)
Jun 07 21:08:22 li1116-193 run_server[324823]:         at java.base/java.util.HashMap.put(HashMap.java:607)
Jun 07 21:08:22 li1116-193 run_server[324823]:         at java.base/java.util.HashSet.add(HashSet.java:220)
Jun 07 21:08:22 li1116-193 run_server[324823]:         at games.strategy.triplea.attachments.TriggerAttachment.getTriggers(TriggerAttachment.java:187)
Jun 07 21:08:22 li1116-193 run_server[324823]:         at games.strategy.triplea.attachments.TriggerAttachment.setActivateTrigger(TriggerAttachment.java:358)
Jun 07 21:08:22 li1116-193 run_server[324823]:         at games.strategy.triplea.attachments.TriggerAttachment$$Lambda$1148/0x0000000840235c40.accept(Unknown Source)
Jun 07 21:08:22 li1116-193 run_server[324823]:         at games.strategy.engine.data.MutableProperty.setStringValue(MutableProperty.java:58)
Jun 07 21:08:22 li1116-193 run_server[324823]:         at games.strategy.engine.data.MutableProperty.setValue(MutableProperty.java:85)
Jun 07 21:08:22 li1116-193 run_server[324823]:         at games.strategy.engine.data.GameParser.setOptions(GameParser.java:1383)
Jun 07 21:08:22 li1116-193 run_server[324823]:         at games.strategy.engine.data.GameParser.parseAttachment(GameParser.java:1323)
[...]

PR #6634, opened 2020-06-08
Fix bots to do shallow parsing of zipped maps

DanVanAtta:
This update changes:
1. bots to do 'shallow' parsing of zipped map files.
   A shallow parse fetches only a limited amount of data.
2. Add logging to bots to print which map is being loaded.

Bots do 'shallow' parsing of expanded maps (folder) but were doing
full parsing of zipped maps when fetching the list of available games.
Full parsing is slower and evaluates the full Map XML converting
all nodes into a GameData object. This is unnecessary work when
we only need the game name and version from the XML.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix: https://github.com/triplea-game/triplea/issues/6632 2.0 Bots are Running out of Memory (Map Parsing Related)  
[] Other:   

## Testing
- checked locally, map load speed is faster.

### DIFF
--- a/game-core/src/main/java/org/triplea/game/server/AvailableGames.java
+++ b/game-core/src/main/java/org/triplea/game/server/AvailableGames.java
@@ -57,6 +57,7 @@ final class AvailableGames {
         .parallelStream()
         .forEach(
             map -> {
+              log.info("Loading map: " + map);
               if (map.isDirectory()) {
                 populateFromDirectory(
                     map,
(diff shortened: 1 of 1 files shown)


DanVanAtta:
merging now in order to deploy this update to bots; with luck this will fix OOM errors.

RoiEXLab:
I'm surprised this actually works, but I'm not complaining

PR #6646, opened 2020-06-08
Add travis-retry to check-links & compatibility check tasks

DanVanAtta:
The check-links & compatibility checks task can see random failures.
This update adds 'travis-retry' which will automatically retry up to 3 times.

Retry documentation is at:
https://docs.travis-ci.com/user/common-build-problems/#travis_retry

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Problem fix:  
[] Other:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ sudo: false
 language: java
 jdk: openjdk11
 cache: { directories: [ ".gradle", "~/.gradle/caches" ] }
-install: skip
+install: skip travis_retry
 
 env:
   global:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,7 +4,7 @@ sudo: false
 language: java
 jdk: openjdk11
 cache: { directories: [ ".gradle", "~/.gradle/caches" ] }
-install: skip
+install: skip travis_retry
```
RoiEXLab, line 7: AFAIK "skip" is just a placeholder term to skip the install task, so it doesn't make sense to keep it here

RoiEXLab: +From the documentation it seems that this is a built-in command no need to run it inside the install step

DanVanAtta: True enough @RoiEXLab 
I must have been tired or something when doing this, I was thinking this was a list of dependencies to install and that 'skip' was one of them somewhere..  :grin: 

Yeah, 'install' block is a script to run, makes sense for this to have error'ed out.

PR #6647, opened 2020-06-08
Convert Offensive/Defensive AA Fire Steps

trevan:
This converts the Offensive/Defensive AA Fire steps to the new BattleStep.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
I created a save file before a battle with AA guns, during AA casualty selection, after AA casualty selection, and after the battle.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
@@ -382,13 +382,11 @@ public class MustFightBattle extends DependentBattle
 
   @Override
   public List<Unit> getOffensiveAa() {
-    updateOffensiveAaUnits();
     return offensiveAa;
   }
 
   @Override
   public List<Unit> getDefendingAa() {
-    updateDefendingAaUnits();
     return defendingAa;
   }
 
@@ -874,6 +872,12 @@ public class MustFightBattle extends DependentBattle
 
   @VisibleForTesting
   public List<String> determineStepStrings(final boolean showFirstRun) {
+    if (offensiveAa == null) {
+      updateOffensiveAaUnits();
+    }
+    if (defendingAa == null) {
+      updateDefendingAaUnits();
+    }
     return BattleSteps.builder()
         .showFirstRun(showFirstRun)
         .attacker(attacker)
@@ -1086,6 +1090,12 @@ public class MustFightBattle extends DependentBattle
   }
 
   private void addFightStartSteps(final boolean firstRun, final List<IExecutable> steps) {
+    if (offensiveAa == null) {
+      updateOffensiveAaUnits();
+    }
+    if (defendingAa == null) {
+      updateDefendingAaUnits();
+    }
     final boolean offensiveAa = canFireOffensiveAa();
     final boolean defendingAa = canFireDefendingAa();
     final BattleStep offensiveAaStep = new OffensiveAaFire(this, this);
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
@@ -1103,13 +1103,11 @@ public class MustFightBattle extends DependentBattle
     if (offensiveAaStep.valid()) {
       steps.add(offensiveAaStep);
     }
-    // see Save Game Compatibility Note on getBattleExecutables
     new IExecutable() {
       private static final long serialVersionUID = 3802352588499530533L;
 
       @Override
       public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
-        // if this gets deserialized, then forward the work to the new BattleStep
         final BattleStep offensiveAaStep =
             new OffensiveAaFire(MustFightBattle.this, MustFightBattle.this);
         offensiveAaStep.execute(stack, bridge);
@@ -1118,13 +1116,11 @@ public class MustFightBattle extends DependentBattle
     if (defensiveAaStep.valid()) {
       steps.add(defensiveAaStep);
     }
-    // see Save Game Compatibility Note on getBattleExecutables
     new IExecutable() {
       private static final long serialVersionUID = -1370090785540214199L;
 
       @Override
       public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
-        // if this gets deserialized, then forward the work to the new BattleStep
         final BattleStep defensiveAaStep =
             new DefensiveAaFire(MustFightBattle.this, MustFightBattle.this);
         defensiveAaStep.execute(stack, bridge);
@@ -1491,13 +1487,11 @@ public class MustFightBattle extends DependentBattle
 
     final BattleStep submergeSubsVsOnlyAir = new SubmergeSubsVsOnlyAirStep(this, this);
     steps.add(submergeSubsVsOnlyAir);
-    // see Save Game Compatibility Note on getBattleExecutables
     new IExecutable() {
       private static final long serialVersionUID = 99990L;
 
       @Override
       public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
-        // if this gets deserialized, then forward the work to the new BattleStep
         final BattleStep submergeSubsVsOnlyAir =
             new SubmergeSubsVsOnlyAirStep(MustFightBattle.this, MustFightBattle.this);
         submergeSubsVsOnlyAir.execute(stack, bridge);
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
@@ -31,6 +31,8 @@ import games.strategy.triplea.delegate.battle.steps.BattleStep;
 import games.strategy.triplea.delegate.battle.steps.BattleSteps;
 import games.strategy.triplea.delegate.battle.steps.RetreatChecks;
 import games.strategy.triplea.delegate.battle.steps.SubsChecks;
+import games.strategy.triplea.delegate.battle.steps.fire.aa.DefensiveAaFire;
+import games.strategy.triplea.delegate.battle.steps.fire.aa.OffensiveAaFire;
 import games.strategy.triplea.delegate.battle.steps.retreat.sub.SubmergeSubsVsOnlyAirStep;
 import games.strategy.triplea.delegate.data.BattleRecord;
 import games.strategy.triplea.formatter.MyFormatter;
@@ -73,26 +75,6 @@ public class MustFightBattle extends DependentBattle
     SUBS,
     PLANES,
     PARTIAL_AMPHIB
-  }
-
-  /**
-   * An action representing attacking aa guns firing during a battle.
-   *
-   * <p>NOTE: This type exists solely for tests to interrogate the execution stack looking for an
-   * action of this type.
-   */
-  public abstract static class FireOffensiveAaGuns implements IExecutable {
-    private static final long serialVersionUID = 7266708569436973099L;
-  }
-
-  /**
-   * An action representing defending aa guns firing during a battle.
-   *
-   * <p>NOTE: This type exists solely for tests to interrogate the execution stack looking for an
-   * action of this type.
-   */
-  public abstract static class FireDefensiveAaGuns implements IExecutable {
-    private static final long serialVersionUID = 2124868665883519949L;
   }
 
   /**
@@ -396,6 +378,18 @@ public class MustFightBattle extends DependentBattle
               Matches.unitIsOwnedBy(defender).or(Matches.enemyUnit(attacker, gameData))));
     }
     return new ArrayList<>(remaining);
+  }
+
+  @Override
+  public List<Unit> getOffensiveAa() {
+    updateOffensiveAaUnits();
+    return offensiveAa;
+  }
+
+  @Override
+  public List<Unit> getDefendingAa() {
+    updateDefendingAaUnits();
+    return defendingAa;
   }
 
   /**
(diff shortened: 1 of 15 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -398,6 +380,18 @@ void addDependentUnits(final Map<Unit, Collection<Unit>> dependencies) {
     return new ArrayList<>(remaining);
   }
 
+  @Override
+  public List<Unit> getOffensiveAa() {
+    updateOffensiveAaUnits();
```
DanVanAtta, line 385: This side effect is striking since getter methods tend to not mutate state. It could also be surprising if one caller were to use this class, and then we add another caller that then calls `getOffensiveAa` and *that* would change the state in a way the first caller might not be expecting.

Making the side-effect of setting the value of `offensiveAa` generally helps. IE:
```
offensiveAa = computeRemainingOffensiveAaUnits();
return offensiveAa;
```

That is still odd though, which makes me wonder if instead this problem would be solved by removing all side effects and instead calling the method something like `computeRemainingOffensiveAaUnits()`. In that case, any place where we call the getter and need those side effects, then we call the method instead. Perhaps this would lead to the `offensiveAa` variable to no longer be needed at all.

What are your thoughts?

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1086,28 +1088,38 @@ private void pushFightLoopOnStack(final boolean firstRun) {
...
+      private static final long serialVersionUID = 3802352588499530533L;
 
-            @Override
-            public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
-              fireDefensiveAaGuns();
-            }
-          });
+      @Override
+      public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
+        // if this gets deserialized, then forward the work to the new BattleStep
```
DanVanAtta, line 1102: Given we already have the detailed comment above about save-game compatibility, do we very much need the 'if this gets' comment? 

I suspect we could delete it. The 'new' BattleStep mention is also very notable as it makes sense when looking at this diff, but a future maintainer looking at the post-merged code will not necessarily know have context to see the difference between 'new' and 'old'.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1086,28 +1088,38 @@ private void pushFightLoopOnStack(final boolean firstRun) {
...
+        final BattleStep offensiveAaStep =
+            new OffensiveAaFire(MustFightBattle.this, MustFightBattle.this);
+        offensiveAaStep.execute(stack, bridge);
+      }
+    };
+    if (defensiveAaStep.valid()) {
+      steps.add(defensiveAaStep);
     }
+    // see Save Game Compatibility Note on getBattleExecutables
```
DanVanAtta, line 1111: Ditto, we may not need this comment. Someone modifying a method should be aware of the method contract, which implies they would have at least read the method javadoc. Removing this comment could reduce future maintenance burden (less stuff to keep in sync, code and comments need to always be verified and kept in sync).

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/aa/DefensiveAaFireTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+  @Mock ExecutionStack executionStack;
+  @Mock IDelegateBridge delegateBridge;
+  @Mock BattleActions battleActions;
+
+  @ParameterizedTest(name = "[{index}] {0} is {2}")
+  @MethodSource
+  void testWhatIsValid(
+      final String displayName, final BattleState battleState, final boolean expected) {
+    final DefensiveAaFire underTest = new DefensiveAaFire(battleState, battleActions);
```
DanVanAtta, line 36: nit, it's a novel contention to name the 'test object' as `underTest`. Other conventions are `sut`, `testObj`, etc.. 

For the most part I think our latest convention is just to use 'natural naming'. IE: `DefensiveAaFire defensiveAaFire`. A benefit of this naming is you can see an example of how code would read when actually used. IE:
```
defensiveAaFire.isValid()
```
vs
```
underTest.isValid()
```
The latter does not give as good of an indication if methods are well named. As well `underTest` is not necessarily easy to follow in test.

Do you have any strong objection to renaming the test variable?

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/aa/DefensiveAaFireTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class DefensiveAaFireTest {
+
+  @Mock ExecutionStack executionStack;
+  @Mock IDelegateBridge delegateBridge;
+  @Mock BattleActions battleActions;
+
+  @ParameterizedTest(name = "[{index}] {0} is {2}")
```
DanVanAtta, line 32: Have a second look at the parameterization if you would. I think it's a clear win if we have ballpark 5 or more variations. With just two, particularly with boolean logic, it might be more straight forward just enumerating the different test cases with individual test blocks. I think in big part parameterized tests should not have `if` blocks, that indicates really that you have multiple test methods (essentially one is testing too many things at once), secondarily data that changes how a test method behaves I think is arguably a test code-smell. WDYT?

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/aa/OffensiveAaFireTest.java
```diff
@@ -0,0 +1,66 @@
...
+@ExtendWith(MockitoExtension.class)
+class OffensiveAaFireTest {
+
+  @Mock ExecutionStack executionStack;
+  @Mock IDelegateBridge delegateBridge;
+  @Mock BattleActions battleActions;
+
+  @ParameterizedTest(name = "[{index}] {0} is {2}")
+  @MethodSource
+  void testWhatIsValid(
```
DanVanAtta, line 34: What do you think of re-grouping this test case with a nested test class to improve the test method naming?

EG:
```
@Nested
class IsValid {
   
   @Test
    void notValidWithNoOffensiveAa() {

    }

    @Test
    void validWithOffensiveAa() {

     }

}

```

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/aa/OffensiveAaFire.java
```diff
@@ -0,0 +1,40 @@
...
+public class OffensiveAaFire extends AaFireAndCasualtyStep {
+  private static final long serialVersionUID = 5843852442617511691L;
+
+  public OffensiveAaFire(final BattleState battleState, final BattleActions battleActions) {
+    super(battleState, battleActions);
+  }
+
+  @Override
+  public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
+    if (valid()) {
```
DanVanAtta, line 21: Will we call `execute` of this action if `valid()` were ever false? Asked in another way, do we only call `execute` when `valid()` is true?

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/aa/OffensiveAaFireTest.java
```diff
@@ -0,0 +1,66 @@
...
+        Arguments.of(
+            "No Offensive Aa", givenBattleStateBuilder().offensiveAa(List.of()).build(), false),
+        Arguments.of(
+            "Some Offensive Aa",
+            givenBattleStateBuilder().offensiveAa(List.of(givenUnitWithTypeAa())).build(),
+            true));
+  }
+
+  @Test
+  void testFiringAaGuns() {
```
DanVanAtta, line 56: nit: `test` prefix is a bit of an old convention (IIRC Junit3 required test methods to begin with 'test'). Latest convention is to focus on describing what is being verified. IE: `verifyFireOffensiveAaGunActionInvoked`. The example I think also demonstrates a question of "under which conditions should that be true?"

Should there be a second test case, or do we legitimately again assume that `isValid()` is invoked first? EG:

```
@Nested
class ExecuteOffensiveAaFire {
   @Test
    @DisplayName("Execute will not invoke fire Offensive Aa Guns when no offensive AA Exist")
   void noOffensiveAa {

   }

   @Test
    @DisplayName("Execute will invoke fire Offensive Aa Guns When Exists Offensive Aa")
   void hasOffensiveAa() {

   }
}
```

(side-note, reading the display name, and or a un-camel-cased list of the test methods and display names in a test class should practically form a complete spec for the class under test,  eg:

- not valid with no offensive Aa
- valid with offensive aa
- execute will not invoke fire Offensive Aa Guns when no offensive AA exist
- execute will invoke fire Offensive Aa Guns When Exists Offensive Aa
)

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: I agree it is surprising.  I had actually introduced a bug when I first started this because I assumed that `offensiveAa` was the value I wanted but it just so happened that `updateOffensiveAaUnits` was called earlier in the caller tree.  When I did some refactoring, my new code was called before `updateOffensiveAaUnits` and so I had the wrong set of aa units compared to the old code.

Unfortunately, `updateOffensiveAaUnits` actually updates 2 private values: `offensiveAa` and `offensiveAaTypes`.  And, as far as I can tell, they both need to stay in sync.

What I could do is to find the place early enough in the call chain where `offensiveAa` and `offensiveAaTypes` need to be initialized.  And then call `updateOffensiveAaUnits` there.  Then `getOffensiveAa` would just return the value.

Another method that has this side effect of updating offensiveAa is `canFireOffensiveAa`, by the way.  I hope to have remove that method when I convert the aa casualty removal step.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: My only fear is that these old steps are inside of methods that are called by `getBattleExecutables`.  So if you go to the top of just this method, you won't see the note about save compatibility.  But I'll remove the note.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/aa/DefensiveAaFireTest.java
trevan: I can rename it.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/aa/OffensiveAaFire.java
trevan: For steps in general, we will call `execute` when `valid()` is false.  For this specific step, I don't think so.  So I could remove it here.  But I've been trying to give each step the same "structure" to help reduce my cognitive load.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/aa/DefensiveAaFireTest.java
trevan: I could get rid of the `if` statement by adding another argument that has the number of names to expect.  So I don't think `if` is that big of a code smell.  I didn't add an extra argument because it was easy to determine what the number of names is based on the expected value.

But having multiple `asserts` that are tangentially related (valid and names) is more of a code smell in doing to much in the test.  I'll split them up.

`AirAttackVsNonSubsStepTest`, `AirDefendVsNonSubsStepTest`, and `SubmergeSubsVsOnlyAirStepTest` are all like this.  Should I change them as well?

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/aa/OffensiveAaFireTest.java
trevan: I'm fine with it.  I don't have a preference any way.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: Going over the code more, I think by calling `updateOffensiveAaUnits` every time `getOffensiveAa` is called might actually introduce a bug.  I think the `offensiveAa` variable should only be updated at the start of each battle round and then stay constant through the round.  So using a `computeRemainingOffensiveAaUnits()` might also introduce a bug.

I've explicitly called `updateOffensiveAaUnits` in all the places where it was implicitly being called originally so now it should be a little more obvious that it is being updated there.  At some future point, we can refactor it to be less brittle.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/aa/OffensiveAaFire.java
DanVanAtta: If you expect `valid()` to always be true, perhaps better to assert it being true:
```
Preconditions.assertState(valid())
```
This way:
1. It becomes a documented expectation and readers can learn of this assumption
2. If the condition is violated while we are doing any kind of demo/testing, we'll have found a bug

PR #6671, opened 2020-06-10
Cleanup: Use interface type instead of concrete type

DanVanAtta:
Resolves codacy code style warning:
> Avoid using implementation types like 'LinkedHashMap'; use the interface instead

Second, minor simplificaton in MoveBatcherTest to unnecessary list wrapping.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:


game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
```diff
@@ -82,7 +82,7 @@ private void addMove(final MoveDescription newMove) {
    * @param sequences The sequences to try to merge into the sequence.
    */
   private static void mergeSequences(
-      final ArrayList<MoveDescription> sequence, final List<ArrayList<MoveDescription>> sequences) {
+      final List<MoveDescription> sequence, final List<List<MoveDescription>> sequences) {
     for (final var otherSequence : sequences) {
       boolean merge = (otherSequence.size() == sequence.size());
       for (int i = 0; merge && i < sequence.size(); i++) {
```
asvitkine, line 88: Note: One of the reasons for choosing ArrayList as params when I wrote this method is because it relies on the List being efficient at random access (i.e. passing a LinkedList would make this highly inefficient).

If we don't want to express such concepts in function signatures, should we make a project rule that LinkedList or more generally lists that don't implement RandomAccess should never be used? Otherwise, loosening signatures loses the extra type checking at compile time that a mistake isn't made.

We could add a runtime check for instanceof RandomAccess, but that's strictly worse than a compile check (extra overhead, error has to be found at runtime, etc).

There's not much reason to use LinkedList so I am OK if we want to take the stance that it just shouldn't be used.

DanVanAtta: The conventional style in Java for better or worse is to use interface types for collections and to strongly avoid implementation types. This is even a default rule in Codacy:
> Avoid using implementation types like 'LinkedHashMap'; use the interface instead

I suspect such a static check exists in other static analysis tools and that Codacy is not alone.

In practice, if you use a more generic interface type, you can avoid needless collection copies. For example, if someone specifies a list all the time, even when not needed and a collection would do, then if you have a set, then you in a position where you needless need to copy the set elements into a list. From a practical perspective, this is the most significant, overly specific implementation types can cause headaches for the API clients and even result in unnecessary data copying.

For example, if someone specified a linked list, and you needed to convert from an array list. The performance gain of a linked list will probably be mooted if every time you have to create a data copy to get the right data type. In general, if you need random access, then use a `List`, otherwise a `Collection`. It's up to the data type that is being stored to then define uniqueness whether you use a `Set` or a `List` (or if the implementation will be wrong if there are duplicate elements, then a `Set` interface is appropriate).

If a method truly needs a specific implementation type, then it probably should be that method that does the data copy to the more efficient data type rather than leaving it to the caller. The name of the game is to provide flexible and easy to use APIs and to avoid others from having to read implementation code. IE: they can "just use" the method and not worry about surprising side-effects or other implementation details.

To another extent, and in practice, keeping types loose allows for different data structures to more easily be used. If everything is a collection, then the specific data type is only defined when you instantiate it. In which case you have freedom to choose whatever type you want, a change from a List to a Set is a change of a single word. Sometimes that avoids having to touch code in dozens or hundreds of places to reduce unnecessary typing.

From a theoretical perspective, one should code to an API, not an implementation. 

From yet another perspsective, performance changes should only be done on clean and tested code that is then measured before and after any performance tweak. It should also be done in context of the overall performance so you avoid optimizing something that is only responsible for 10% of the total performance (instead focus on the 90%). For example, if the list only ever 100 elements large, iterating over it 1000 times with O(n) access vs O(1) will likely be a difference of nano-seconds. One should choose appropriate data structures, but focus on simple code that for examples avoids calling a method an extra 10 times because it is simple and you can see that the 10 subsequent calls are not needed. 

In TripleA, we have a call to get default casualty selection, which does a call to get unit power stats (the heavy part to compute supports), but we actually only cache results for one of those calls. If we could simplify that code so that the results were re-used instead of recomputed a second time with caching, we'd have faster code with less memory and would correct cases where we fail to invalidate the cache.

Last point, it's conventional to use HashMap, HashSet, and ArrayList. Using any other implementation generally needs a special case. And again, if it's really important for a method, that method should measure performance and convert to the needed data structure. 

The above is admittedly not a direct answer, but in the end using concrete types is not conventional and generally unnecesssary, and can cause harm either from an API usage point of view or from requiring excessive burden on the clients of API.

asvitkine:
On Mon, Jun 15, 2020, 11:27 PM Dan Van Atta <notifications@github.com>
wrote:

> *@DanVanAtta* commented on this pull request.
> ------------------------------
>
> In
> game-core/src/main/java/games/strategy/triplea/ai/pro/util/MoveBatcher.java
> <https://github.com/triplea-game/triplea/pull/6671#discussion_r440564389>:
>
> > @@ -82,7 +82,7 @@ private void addMove(final MoveDescription newMove) {
>     * @param sequences The sequences to try to merge into the sequence.
>     */
>    private static void mergeSequences(
> -      final ArrayList<MoveDescription> sequence, final List<ArrayList<MoveDescription>> sequences) {
> +      final List<MoveDescription> sequence, final List<List<MoveDescription>> sequences) {
>      for (final var otherSequence : sequences) {
>        boolean merge = (otherSequence.size() == sequence.size());
>        for (int i = 0; merge && i < sequence.size(); i++) {
>
> The conventional style in Java for better or worse is to use interface
> types for collections and to strongly avoid implementation types. This is
> even a default rule in Codacy:
>
> Avoid using implementation types like 'LinkedHashMap'; use the interface
> instead
>
> I suspect such a static check exists in other static analysis tools and
> that Codacy is not alone.
>
> In practice, if you use a more generic interface type, you can avoid
> needless collection copies. For example, if someone specifies a list all
> the time, even when not needed and a collection would do, then if you have
> a set, then you in a position where you needless need to copy the set
> elements into a list. From a practical perspective, this is the most
> significant, overly specific implementation types can cause headaches for
> the API clients and even result in unnecessary data copying.
>
> For example, if someone specified a linked list, and you needed to convert
> from an array list. The performance gain of a linked list will probably be
> mooted if every time you have to create a data copy to get the right data
> type. In general, if you need random access, then use a List, otherwise a
> Collection. It's up to the data type that is being stored to then define
> uniqueness whether you use a Set or a List (or if the implementation will
> be wrong if there are duplicate elements, then a Set interface is
> appropriate).
>
Copying a collection is a O(n) operation. I believe this algorithm is
O(n^2) if a LinkedList is used, so it would actually be more efficient to
copy the elements into an ArrayList for any large ish values of n.

Anyway, it doesn't look like we use LinkedList anywhere in the codebase.
Can we just ban it so that we can assume that all Lists are efficient at
random access?


If a method truly needs a specific implementation type, then it probably

DanVanAtta:
Banning LinkedList seems draconian. I would agree we should avoid it unless we really need it. In such cases I think the scope of its usage would likely be pretty limited. Specifically we'd pass the instance to a method that still accepts an API of `List` and would add a note to the instantiation of why we chose the `LinkedList`. 

Part of my point is that a O(n^2) vs O(n) method of small collection is immaterial. A list of 10 elements requiring 1ns vs 5ns simply does not matter. In general performance comes from simple design first. It's not exceptionally common where it's the algorithm. If you can simplify a design to the point where some code is obviously not needed, having no code is infinitely faster than any code you could write.

The rule of thumb I go by is "optimize what you need to, when you need to, and after you have cleaned the code". Second rule of thumb: "working, clean, fast", in that order. I'd often take code that results in a 10% increase in runtime from 1s to 1.1s  if it results in code that is far cleaner and puts us on a path where we can simplify the code to be 0.1s in runtime. Sometimes a step back is needed for two steps forward.

On another point, a bit biased, I also think the whiteboard interviews have been one of the single most harmful things to the industry and software development. Those interviews have trained us to look for fast algorithms first and foremost when often it does not matter and instead we should prefer dumb and easy to maintain algorithms and simpler design that in the end leads to a faster, easier to maintain system. I believe simple is fast.

"Simplicity is prerequisite for reliability."  - https://en.wikiquote.org/wiki/Edsger_W._Dijkstra

FWIW, I highly recommend this blog: https://blog.codinghorror.com/the-infinite-space-between-words/, it's an interesting read and shares an interesting perspective (and it's funny how many places where we reach over the network needless, resulting in delays of a few hundred milliseconds). 

With all that said, given we have benchmarked this code and found some bottlenecks, the perf improvements are merited. We have aligned the right data structure, so ultimately there is no problem.

PR #6677, opened 2020-06-10
Disable codeclimate check to start lists at beginning of line

DanVanAtta:
This code climate rule conflicts with codacy, this update is to prefer
the codacy check that lists are indented.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[x] Configuration Change
[] Problem fix:  
[] Other:   

## Additional Review Notes

Resolves code climate violations triggered in: https://github.com/triplea-game/triplea/pull/6668

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -24,6 +24,11 @@ plugins:
   # https://docs.codeclimate.com/docs/markdownlint
   markdownlint:
     enabled: true
+    checks:
+      MD006:
+        # Disable "Consider starting bulleted lists at the beginning of the line"
+        enabled: false
+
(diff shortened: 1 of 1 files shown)


DanVanAtta:
Do you have a strong preference @RoiEXLab ?
I've a slight preference for the indentation, I think that is how we had it originally and changed it for the sake of codeclimate.

RoiEXLab:
Not really, but now when I think about it:
I think I put indentations there whenever I use bullet points (unordered) but I don't when I use an ordered list.
But honestly I don't care too much I'm fine with both

PR #6686, opened 2020-06-11
Initial tests for MustFightBattle#getAttackerRetreatTerritories

trevan:
Adds tests for `MustFightBattle#getAttackerRetreatTerritories`.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Additional Review Notes
This is a WIP because I'd like feedback on the code convention of dealing with testing this type of method.

### DIFF
--- a/game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleTest.java
+++ b/game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleTest.java
@@ -1,5 +1,8 @@
 package games.strategy.triplea.delegate.battle;
 
+import static games.strategy.triplea.Constants.LAND_BATTLE_ROUNDS;
+import static games.strategy.triplea.Constants.RETREATING_UNITS_REMAIN_IN_PLACE;
+import static games.strategy.triplea.Constants.UNIT_ATTACHMENT_NAME;
 import static games.strategy.triplea.delegate.GameDataTestUtil.addTo;
 import static games.strategy.triplea.delegate.GameDataTestUtil.battleDelegate;
 import static games.strategy.triplea.delegate.GameDataTestUtil.move;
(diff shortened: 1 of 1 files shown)


game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleTest.java
```diff
@@ -74,4 +96,293 @@ void testFightWithBothZeroStrength() {
...
+      final Map<Territory, Collection<Unit>> attackingFromMap,
+      final List<Unit> attackingUnits,
+      final Territory battleSite,
+      final BattleTracker battleTracker,
+      final GameData gameData,
+      final boolean headless,
+      final Collection<Territory> expected) {
+
+    final MustFightBattle battle =
+        new MustFightBattle(battleSite, attacker, gameData, battleTracker);
```
trevan, line 114: Instantiating a `MustFightBattle` is not an easy thing.  It's code isn't really set up for Unit testing.  Most of the mocks that exist are just to work on the code that is called in the constructor.

One idea is to create a mock and then use mockito's `thenCallRealMethod` to call it:
```
MustFightBattle battle = mock(MustFightBattle.class);
battle.attackingUnits = attackingUnits;
... more setup ...
when(battle.getAttackerRetreatTerritories()).thenCallRealMethod();
Collection<Territory> retreatTerritories = battle.getAttackerRetreatTerritories();
```

I haven't tried that but I think it would work.

Another idea is to move this method to a helper class (call it `RetreatTerritories` and put both offensive and defensive retreat logic there) and then call the help class from the original `getAttackerRetreatTerritories`.  Then the test can work on that class instead of `MustFightBattle`.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleTest.java
```diff
@@ -74,4 +96,293 @@ void testFightWithBothZeroStrength() {
...
+              battleSite,
+              mock(BattleTracker.class),
+              MockGameData.builder().isAtWar(attacker, defender).landBattleRounds().build(),
+              false,
+              List.of(attackFrom)));
+    }
+    return arguments;
+  }
+
+  static class MockGameData {
```
trevan, line 235: What do you think of these "mock" builder classes?  Building out the complex objects that are needed for these tests (which are more like an integration test instead of a unit test) can be a little complex with mockito.  Instead of doing a bunch of
```
when(mockObject.function()).thenReturn(something);
when(mockObject.otherFunction()).thenReturn(somethingElse);
```
I'd rather do something more fluent.

But GameData, Unit, Territory, etc do not have a fluent builder api so I can't utilize mockito's fluent apis and even then I'm not sure I'd be able to use them.

So I played around with these builder classes.  It generates a mock object and then gives me methods that I can fluently call to modify the object.  In the end, I have a finished mock that can be used.

I could replace this with a ton of `givenGameData...` methods that deal with the different permutations but that feels like it will become unwieldy with all of the possible permutations.

PR #6696, opened 2020-06-15
Replace BattleStep#valid with BattleStep#getOrder

trevan:
I discovered that some steps have different `valid` conditions depending on if it is called at the beginning of a round during `getNames` or later during `execute`.  An example is the offensive sub submerge/withdraw.  During `getNames`, it shouldn't care about the presence of a destroyer because the destroyer could be killed later.  But during `execute`, it must care about the presence of a destroyer.

So, this removes the `valid` method from the `BattleStep` interface.

But since I still need to know if a step should be visible or not, I've added a `getOrder` method.  If a step should be skipped, then it can return the enum `NOT_APPLICABLE`.  Otherwise, it should return the order it should be run in.  All of the existing steps that have been converted only return `SKIP` or one other enum.  But other steps will return different enums depending on their situation (eg. subs can submerge at the beginning of the round or at the end).

Once all the steps have been converted, then I should be able to get the step names with code like:
```
List<BattleStep> steps = new ArrayList<>();
steps.add(new OffensiveAaFire(this, battleActions));
steps.add(new DefensiveAaFire(this, battleActions));
steps.add(new ...);
steps.add(new ...);

List<String> stepNames = steps.stream()
        .filter(step -> step.getOrder() != BattleStep.Order.NOT_APPLICABLE)
        .sorted(Comparator.comparing(BattleStep::getOrder))
        .flatMap(step -> step.getNames().stream())
```

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
@@ -71,9 +71,12 @@ public class BattleSteps implements BattleStepStrings, BattleState {
     final BattleStep airDefendVsNonSubs = new AirDefendVsNonSubsStep(this);
 
     final List<String> steps = new ArrayList<>();
-    steps.addAll(offensiveAaStep.getNames());
-
-    steps.addAll(defensiveAaStep.getNames());
+    if (offensiveAaStep.getOrder() != BattleStep.Order.SKIP) {
+      steps.addAll(offensiveAaStep.getNames());
+    }
+    if (defensiveAaStep.getOrder() != BattleStep.Order.SKIP) {
+      steps.addAll(defensiveAaStep.getNames());
+    }
 
     if (showFirstRun) {
       if (!isBattleSiteWater && !bombardingUnits.isEmpty()) {
@@ -109,7 +112,9 @@ public class BattleSteps implements BattleStepStrings, BattleState {
             || defendingUnits.stream().anyMatch(Matches.unitIsTransport()))) {
       steps.add(REMOVE_UNESCORTED_TRANSPORTS);
     }
-    steps.addAll(submergeSubsVsOnlyAir.getNames());
+    if (submergeSubsVsOnlyAir.getOrder() != BattleStep.Order.SKIP) {
+      steps.addAll(submergeSubsVsOnlyAir.getNames());
+    }
 
     final boolean defenderSubsFireFirst =
         SubsChecks.defenderSubsFireFirst(attackingUnits, defendingUnits, gameData);
@@ -161,7 +166,9 @@ public class BattleSteps implements BattleStepStrings, BattleState {
       steps.add(REMOVE_SNEAK_ATTACK_CASUALTIES);
     }
 
-    steps.addAll(airAttackVsNonSubs.getNames());
+    if (airAttackVsNonSubs.getOrder() != BattleStep.Order.SKIP) {
+      steps.addAll(airAttackVsNonSubs.getNames());
+    }
 
     if (attackingUnits.stream().anyMatch(Matches.unitIsFirstStrike().negate())) {
       steps.add(attacker.getName() + FIRE);
@@ -178,7 +185,9 @@ public class BattleSteps implements BattleStepStrings, BattleState {
       steps.add(defender.getName() + FIRST_STRIKE_UNITS_FIRE);
       steps.add(attacker.getName() + SELECT_FIRST_STRIKE_CASUALTIES);
     }
-    steps.addAll(airDefendVsNonSubs.getNames());
+    if (airDefendVsNonSubs.getOrder() != BattleStep.Order.SKIP) {
+      steps.addAll(airDefendVsNonSubs.getNames());
+    }
     if (defendingUnits.stream().anyMatch(Matches.unitIsFirstStrike().negate())) {
       steps.add(defender.getName() + FIRE);
       steps.add(attacker.getName() + SELECT_CASUALTIES);
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
@@ -71,12 +71,8 @@ public class BattleSteps implements BattleStepStrings, BattleState {
     final BattleStep airDefendVsNonSubs = new AirDefendVsNonSubsStep(this);
 
     final List<String> steps = new ArrayList<>();
-    if (offensiveAaStep.getOrder() != BattleStep.Order.NOT_APPLICABLE) {
-      steps.addAll(offensiveAaStep.getNames());
-    }
-    if (defensiveAaStep.getOrder() != BattleStep.Order.NOT_APPLICABLE) {
-      steps.addAll(defensiveAaStep.getNames());
-    }
+    steps.addAll(offensiveAaStep.getNames());
+    steps.addAll(defensiveAaStep.getNames());
 
     if (showFirstRun) {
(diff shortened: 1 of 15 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
```diff
@@ -71,9 +71,12 @@
     final BattleStep airDefendVsNonSubs = new AirDefendVsNonSubsStep(this);
 
     final List<String> steps = new ArrayList<>();
-    steps.addAll(offensiveAaStep.getNames());
-
-    steps.addAll(defensiveAaStep.getNames());
+    if (offensiveAaStep.getOrder() != BattleStep.Order.NOT_APPLICABLE) {
+      steps.addAll(offensiveAaStep.getNames());
```
DanVanAtta, line 75: I think there might be something of a feature envy code smell here. Notably we query an object for its state, then depending on that state we query the same object again for more state. The feature envy is on `BattleSteps` with respect to the state of `offensiveAaStep`. In practical sense this indicates a lack of data encapsulation.

This could potentially be fixed by moving the `getOrder()` concern to `offensiveAaStep`. In such a case there would be no conditional in `BattleSteps`. In other words, `offensiveAaStep` would evaluate its own order, if not applicable it would return an empty `getNames()` list, otherwise it would return some values.

WDYT?

trevan: I agree with the code smell but that is because this is an intermediate state.  There are steps that will return different values for `getOrder` depending on their situation.  Once all of the steps are converted, I can grab the list of steps with code like:
```
List<BattleStep> steps = new ArrayList<>();
steps.add(new OffensiveAaFire(this, battleActions));
steps.add(new DefensiveAaFire(this, battleActions));
steps.add(new ...);
steps.add(new ...);

List<String> stepNames = steps.stream()
        .filter(step -> step.getOrder() != BattleStep.Order.NOT_APPLICABLE)
        .sorted(Comparator.comparing(BattleStep::getOrder))
        .flatMap(step -> step.getNames().stream())
```

Another idea, is that instead of allowing a step to have different positions in the battle order, I could just create duplicate steps.  They could inherit from the same parent class where all the main logic is and then override a method that determines whether `getName` or `executable` should be run.

Then the code to get the steps would look more like:
```
List<BattleStep> steps = new ArrayList<>();
steps.add(New OffensiveAaFire(this, battleActions));
steps.add(new SubRetreatBeforeBattle(this, battleActions));
steps.add(...<some other step>);
steps.add(new SubRetreatAfterBattle(this, battleActions));
steps.add(...<other steps>);

List<String> stepNames = steps.stream()
        .flatMap(step -> step.getNames().stream())
```

In that case, both `SubRetreatBeforeBattle` and `SubRetreatAfterBattle` would extend `SubRetreat` but the first class would check for `getSubRetreatBeforeBattle() == true` while the second class would check if `getSubRetreatBeforeBattle() == false`.

What would you prefer?  More classes or less classes?  Or maybe you have another idea on how to handle steps that don't have a static position in the order but can change depending on the state at the beginning of the round.

DanVanAtta: @trevan 
I was thinking this:
```

List<String> stepNames = steps.stream()
        .sorted(Comparator.comparing(BattleStep::getOrder))
        .flatMap(step -> step.getNames().stream())
```

If you make BattleStep implement Comparable, perhaps could get even go simpler:

```
public interface BattleStep extends IExecutable, Comparable<BattleStep> {
 :
 :
  @Override
  default int compareTo(BattleStep rhs) {
    return Integer.compare(getOrder().ordinal(), rhs.getOrder().ordinal());
  }
```

```
List<String> stepNames = steps.stream()
        .sorted()
        .flatMap(step -> step.getNames().stream())
```

trevan: @DanVanAtta so no pre filtering?  I'll try that out.  The existing code does do pre-filtering on the IExecutables that are added and I was trying to allow that with the `NOT_APPLICABLE` order but that might just be a relic.

DanVanAtta: Basically the filtering would be internalized to each BattleStep. Each one would be responsible to know if it applies and if so generating any necessary step names.

trevan: @DanVanAtta , I'm not just talking about step names but the executable part as well.  I'm worried about the lines such as:

https://github.com/triplea-game/triplea/blob/f524e9819a292c6c3bfde535c7f4d90bb94088ca/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java#L1105
https://github.com/triplea-game/triplea/blob/f524e9819a292c6c3bfde535c7f4d90bb94088ca/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java#L1118
https://github.com/triplea-game/triplea/blob/f524e9819a292c6c3bfde535c7f4d90bb94088ca/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java#L1131
https://github.com/triplea-game/triplea/blob/f524e9819a292c6c3bfde535c7f4d90bb94088ca/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java#L1475

I'm not entirely sure why the check for those steps is outside of the executable.  A check outside of the executable happens at the start of the round while the checks inside of the executable happen when the step actually happens.  So the checks are being done on different data.  Most of the executables have all their checks inside of them or the checks outside of them are just determining the order of the step.

It might be that I can move all of those checks inside of the executable so that they run when the step occurrs but I'm not positive at this moment in time and so the `getOrder() != NOT_APPLICABLE` was mainly for those steps.  I guess I can remove the NOT_APPLICABLE order and when I get to converting those steps, figure out if I need NOT_APPLICABLE or not.

trevan: Oh, in addition, by not adding the step to the stack, they don't get saved in saved games.  That is another thing that I have to look into when I convert them.  By not having the `getOrder() != NOT_APPLICABLE` check, all of the steps will be added to the saved game, including ones that were previously not added.

trevan: @DanVanAtta I've made the change.  I've also tested a scenario that would originally have had just offensive AA steps in the save and no defensive AA steps in the save.  I created a save with master and with this branch and then loaded both saves in this branch.  It loaded fine and it looks like it played out the rest of the round without issues.

DanVanAtta: Interesting context, thank you. Sounds as though we should err on the side of caution.

DanVanAtta:
@trevan this is ready for a last pass of review? I hope to have some time tomorrow to finish up review.

trevan:
> @trevan this is ready for a last pass of review? I hope to have some time tomorrow to finish up review.

@DanVanAtta yes, it should be ready.

I'm leaving `getOrder()` around instead of doing your idea of `Comparable`.  I like the idea of `Comparable` and so might switch to that once all the steps have migrated.

I did remove the `NOT_APPLICABLE` enum value and will revisit that in the future if I come across a step that I need to not add to the executable stack.

DanVanAtta:
:+1:

DanVanAtta:
Thanks for working on this @trevan  :100:

PR #6713, opened 2020-06-18
Compute loop constant once

DanVanAtta:
Small change to a for loop to compute the terminal condition just once.
'getHitPoints' is a call to attachments which does casting, it is not
a constant value (so its better to compute it once rather than on each
loop iteration).

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/UnitChooser.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/UnitChooser.java
@@ -418,7 +418,9 @@ public final class UnitChooser extends JPanel {
       defaultHits = new ArrayList<>(Math.max(1, category.getHitPoints() - category.getDamaged()));
       final int numUnits = category.getUnits().size();
       int hitsUsedSoFar = 0;
-      for (int i = 0; i < Math.max(1, category.getHitPoints() - category.getDamaged()); i++) {
+      for (int i = 0, m = Math.max(1, category.getHitPoints() - category.getDamaged());
+          i < m;
+          i++) {
         // TODO: check if default value includes damaged points or not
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/triplea/ui/UnitChooser.java
```diff
@@ -418,7 +418,9 @@ void disableMax() {
       defaultHits = new ArrayList<>(Math.max(1, category.getHitPoints() - category.getDamaged()));
       final int numUnits = category.getUnits().size();
       int hitsUsedSoFar = 0;
-      for (int i = 0; i < Math.max(1, category.getHitPoints() - category.getDamaged()); i++) {
+      for (int i = 0, m = Math.max(1, category.getHitPoints() - category.getDamaged());
+          i < m;
+          i++) {
```
RoiEXLab, line 423: Actually I think it would be better to just move initialisation and deklaration out of the loop, up one line

DanVanAtta: IE, convert the for loop to a while loop?

RoiEXLab: Not really, just the constant variable 'm'

DanVanAtta: That has a problem where the `m` gets increased scoping compared to what's needed.

FWIW, computing loop termination conditions like this used to be quite idiomatic. PMD used to recommend it. This performance improvement I think just became less common with Java 1.5 and for-each loops as there are usually better ways to iterate and iterating with a non-constant loop condition is less common.

PR #6719, opened 2020-06-19
Create OffensiveSubsRetreat step

trevan:
This converts the offensive subs retreat step into a BattleStep.

This step can happen before or after a battle, depending on the setting "Sub Retreat Before Battle".

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing
Saved a game with subs and loaded it.

## Additional Review Notes
This unifies the logic for offensive subs withdrawal/submerge.  But because of the unification, some changes are going to occur.  I have limited the changes to how the UI displays the step string.

* If `Sub Retreat Before Battle` was true or if`Sub Retreat Before Battle` was false and `Submersible Subs` was false, then it would show the step string even if there was no territories to retreat to.  It now checks it always.
* It `Sub Retreat Before Battle` was false and `Submersible Subs` was false, then it would not show the step string if there was a destroyer there.  But if the destroyer died before the step actually happened, then it would trigger an error.  It now only checks for a destroyer if `Sub Retreat Before Battle` is true.


game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
```diff
@@ -51,24 +55,49 @@
...
 
+  final @NonNull Boolean isOver;
+
+  @Override
+  public Collection<Territory> getAttackerRetreatTerritories() {
+    return getAttackerRetreatTerritories.get();
+  }
+
+  @Override
+  public boolean isOver() {
```
trevan, line 84: I couldn't use `@Getter` because isOver is a `Boolean` and not a `boolean`.  `@Getter` will create a `is` function only for `boolean`.  I could change isOver to be a `boolean` and then I can't have the `@NonNull`.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
```diff
@@ -185,22 +214,8 @@
     // remove casualties
     steps.add(REMOVE_CASUALTIES);
     // retreat attacking subs
-    if (attackingUnits.stream().anyMatch(Matches.unitCanEvade())) {
-      if (Properties.getSubmersibleSubs(gameData)) {
-        // TODO: BUG? Should the presence of destroyers be checked?
```
trevan, line 190: I figured out why destroyers weren't being checked and it was correct.  The else after this (where it calls canAttackerRetreatSubs) shouldn't have been checking for destroyers either.  I was able to trigger an error in the UI in that else case.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreat.java
```diff
@@ -0,0 +1,93 @@
...
+
+  protected final BattleActions battleActions;
+
+  @Override
+  public List<String> getNames() {
+    if (!isEvaderPresent() || !isRetreatPossible()) {
+      return List.of();
+    }
+
+    if (getOrder() == SUB_OFFENSIVE_RETREAT_BEFORE_BATTLE && isDestroyerPresent()) {
```
trevan, line 35: I'm not sure about keeping this logic.  It only works as long as nothing can kill the destroyer before the sub has a chance to retreat.  Since the sub can retreat at the beginning of the battle, the only thing that fires before its retreat is AA guns.  And if a map has an AA gun that can target destroyers, then an error could occur.

Would you expect AA guns that can target destroyers?  I could either remove this logic or change it to go through all of the units, figure out if there are any AA guns and determine if they can target all of the destroyers in the defenders side.  I lean towards removing the logic.  It'll mean that the withdraw/submerge step will show up more often in the UI then it did previously.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
```diff
@@ -448,240 +444,6 @@ void offensiveAndDefensiveAaFire() {
         is(2));
   }
 
-  @Test
```
trevan, line 451: All of these tests have been moved to the OffensiveSubsRetreat tests.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/sub/OffensiveSubsRetreatTest.java
```diff
@@ -0,0 +1,196 @@
...
+        Arguments.of(
+            "No retreat territories and can not submerge",
+            givenBattleStateBuilder()
+                .gameData(
+                    MockGameData.givenGameData().withTransportCasualtiesRestricted(false).build())
+                .attackerRetreatTerritories(List.of())
+                .build()));
+  }
+
+  static class MockGameData {
```
trevan, line 159: What do you think of this class?  I was wanting a fluent way to generate a mock GameData.

trevan:
I checked the Codacy report.  The new issue it found of "combining nested if statements" is valid and I plan to fix that when I convert the defensive subs retreat step.  I can fix the Codacy issue now or completely remove it when I do the conversion.  Which would you prefer?

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreatTest.java
```diff
@@ -0,0 +1,195 @@
...
+@ExtendWith(MockitoExtension.class)
+class OffensiveSubsRetreatTest {
+
+  @Mock ExecutionStack executionStack;
+  @Mock IDelegateBridge delegateBridge;
+  @Mock BattleActions battleActions;
+
+  @ParameterizedTest(name = "[{index}] {0}")
+  @MethodSource
+  void retreatHappens(final String displayName, final BattleState battleState) {
```
DanVanAtta, line 43: nit, if you have to parameterize the display name, I think we might be parameterizing 'wrong'

The display name should likely be a constant for any test, the only thing that is changing are just values under test.

For example, if you are testing a 'sine' function, the input and expected outputs are parameterized, but in all cases we are testing a sine function. In part as well, JUnit will automatically print out the test parameters as part of the display.

To resolve, a comment on the test data could be sufficient. You could also consider passing in as parameters the components for a `Battlestate` and then constructing the battle state in the test.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreatTest.java
```diff
@@ -0,0 +1,195 @@
...
+  static List<Arguments> retreatDoesNotHappen() {
+    return List.of(
+        Arguments.of(
+            "Battle is over",
+            givenBattleStateBuilder()
+                .attackingUnits(List.of(givenUnitCanEvade()))
+                .over(true)
+                .build()),
+        Arguments.of(
+            "Defending Destroyer exists",
```
DanVanAtta, line 107: I think I might be missing something. A defending destroyer prevents submerge, why does it prevent retreat?

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/sub/OffensiveSubsRetreatTest.java
```diff
@@ -0,0 +1,196 @@
...
+    private final GameData gameData;
+    private final GameProperties gameProperties;
+    private boolean propertiesSetup = false;
+
+    private MockGameData() {
+      gameData = mock(GameData.class);
```
DanVanAtta: Can this mock instantiation be inlined and a constructor avoided?

I think our general convention is to favor inlining declarations and only declare what you must in a constructor.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/sub/OffensiveSubsRetreatTest.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    MockGameData withTransportCasualtiesRestricted(final boolean value) {
+      setupProperties();
+      when(gameProperties.get(TRANSPORT_CASUALTIES_RESTRICTED, false)).thenReturn(value);
+      return this;
+    }
+
+    private void setupProperties() {
+      if (!propertiesSetup) {
```
DanVanAtta: Is this if check here to avoid strict mocking checks?
It might be a case where we can use lenient. IMO using lenient is going to be better than having lazy instantiation of a mock expectation.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/sub/OffensiveSubsRetreatTest.java
```diff
@@ -0,0 +1,196 @@
...
+
+    static MockGameData givenGameData() {
+      return new MockGameData();
+    }
+
+    GameData build() {
+      return gameData;
+    }
+
+    MockGameData withTransportCasualtiesRestricted(final boolean value) {
```
DanVanAtta: I like this method :+1: 

I tend to think this indicates we should aggregate more of this kind of logic in `GameData` itself so you could have method on game data to turn this property on.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreat.java
DanVanAtta: It seems a bit questionable either way. Not sure I have a strong opinion to remove it. I'd opt personally to leave as-is for now. If changed, would be good to have it be a change on its own and a complete analysis of what we're updating (how often we see the display when we don't need to).

I think the real fix is to make the concepts more generic and check for those. 

IE: `firstStrikeUnit` and `firstStrikeCounterUnit`. You then have a 'first strike' label present and it's neutralized if you have a counter unit. We also need to keep in mind that the rulesets can vary as well and that we may need to make whether that happens or not completely configurable.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreatTest.java
trevan: I tried your later idea (passing in parameters) but I couldn't build out a decent display name with the passed in parameters.

This does fit parameterizing, but the parameters are extremely complicated so building a human visible display name out of it is not workable.  I'll probably flip it so that the `retreatHappens` is just a helper `then` function and split out the parameterization into separate methods that call this `then` function.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreatTest.java
trevan: The special sub submerge/retreat step is separate from the main retreat step.  This special step will either allow submerging or retreating depending on a gameData parameter (`getSubmersibleSubs`).

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/sub/OffensiveSubsRetreatTest.java
trevan: Yes, it is to avoid strict mocking checks.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreat.java
trevan: Even if the concepts become more generic, the problem still exists.  The main reasons are:
* A battle has up to 3 places where casualties are removed and can't fire back.  This is after the AA battle, the sub battle, and bombardment.
* The collection of steps to show to the user happen at the very beginning of a round before any steps actually occur.

So if there is a `firstStrikeUnit` and a `firstStrikeCounterUnit`, you would normally think that the `firstStrikeWithdrawStep` would not happen since there is a counter unit.  But if the AA part happens before the `firstStrikeWithdrawStep` and the `firstStrikeCounterUnit` is killed, it is removed from the battle and no longer prevents the `firstStrikeWithdrawStep`.  So you would actually need to show the `firstStrikeWithdrawStep`.

But I'll leave the check here since it was how it was originally.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/sub/OffensiveSubsRetreatTest.java
DanVanAtta: Do you have objection to using a lenient? It seems like a pretty good case for it. Avoiding state in test scaffolding is generally a very good thing.

trevan: I'm fine with lenient.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreat.java
DanVanAtta: If a step is valid at the beginning of combat, and then is removed, that could be fine. Perhaps nicer to visualize that it no longer applied. Overall, AFAIK, the number one requested feature for combat phases is the ability to "walk back in time" by clicking the phases to see what happened during that phase - very similar to how the history mode works.

FWIW, keeping as-is for now seems prudent, keeping this just to a refactor update.  :+1:

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreatTest.java
trevan: Codacity doesn't like what I did.  It is complaining that none of the test methods have `assert` or `fail`.  I can "fix" this by inlining the `then` methods or moving the `verify` calls from the `then` methods to the test methods.

Or, I could move back to the parameterized code and add a comment as you originally suggested.

@DanVanAtta , thoughts?

trevan: I inlined the methods.  The `then` method was only three lines long.

trevan:
Codacy is flagging a place where the if statements can be combined.  That will be reworked when I convert the defensive subs retreat step so I'd rather do it then and not as part of this pr.

PR #6725, opened 2020-06-20
Try #2 to fix NPE in MustFightBattle.

asvitkine:
This change reverts: https://github.com/triplea-game/triplea/commit/b67eb52c3d30f43550ddbaf02f0ea67a63564a67

As the underlying cause was not that player was null, but that
player.getGameData() was null. Instead, it changes the code to
use the GameData from the battle, as was the case before the
refactoring in: https://github.com/triplea-game/triplea/commit/321b579556b78f0b2b77de4e11aa4dac9540a7de

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix:  https://github.com/triplea-game/triplea/issues/6724
[] Other:


RoiEXLab:
Thanks for looking into this 👍

PR #6727, opened 2020-06-21
Fix PbF Name

### LINKED ISSUE #6717: PBF savegame gets cryptic name
Play any PBF game, at the end of the turn do the turn post.

### What is an expected fix?
As in the past the name of the savegame should contain the thread-number and the turn number of and with the power that just performed.

### Which Engine Version are you using?
2.0.20080

### (Optional) Additional information

I have not tested PBF for some time now, so I cannot tell when this behaviour began.
Look at the savegame here: 

https://www.axisandallies.org/forums/post/1410278

It is named 
triplea16780680437380566071.tsvg

instead of something like
triplea-20781-Ger1.tsvg

The new naming is not intuitive and does not give any clue to which game it belongs.
What is the idea behind that? What have I maybe missed?

Relevant for #5025, IMHO.

RoiEXLab:
Should fix #6717

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  #6717 
[] Other:   

## Testing

Haven't actually tested the changes but they should do the trick

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbf/test/post/TestPostAction.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbf/test/post/TestPostAction.java
@@ -3,6 +3,7 @@ package games.strategy.engine.framework.startup.ui.posted.game.pbf.test.post;
 import com.google.common.base.Preconditions;
 import games.strategy.engine.ClientContext;
 import games.strategy.engine.posted.game.pbf.NodeBbForumPoster;
+import games.strategy.engine.posted.game.pbf.NodeBbForumPoster.SaveGameParameter;
 import java.awt.Graphics;
 import java.awt.image.BufferedImage;
 import java.io.File;
@@ -53,7 +54,12 @@ public class TestPostAction implements BiConsumer<String, Integer> {
                           + ", time: "
                           + DateTimeUtil.getLocalizedTime(),
                       "Testing Forum poster",
-                      f != null ? f.toPath() : null);
+                      f != null
+                          ? SaveGameParameter.builder()
+                              .path(f.toPath())
+                              .displayName("Test.jpg")
+                              .build()
+                          : null);
               testPostProgressDisplay.close();
               try {
                 // now that we have a result, marshall it back unto the swing thread
--- a/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbForumPoster.java
+++ b/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbForumPoster.java
@@ -13,6 +13,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import lombok.Builder;
 import org.apache.http.NameValuePair;
 import org.apache.http.client.entity.UrlEncodedFormEntity;
@@ -54,6 +55,12 @@ public class NodeBbForumPoster {
     @Nonnull private final char[] username;
     @Nonnull private final char[] password;
     @Nonnull private final String forumUrl;
+  }
+
+  @Builder
+  public static class SaveGameParameter {
+    @Nonnull private final Path path;
+    @Nonnull private final String displayName;
   }
 
   private NodeBbForumPoster(final ForumPostingParameters forumPostingParameters) {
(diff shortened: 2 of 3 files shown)


game-core/src/main/java/games/strategy/engine/framework/startup/ui/posted/game/pbf/test/post/TestPostAction.java
```diff
@@ -53,7 +54,12 @@ public void accept(final String forumName, final Integer topicId) {
                           + ", time: "
                           + DateTimeUtil.getLocalizedTime(),
                       "Testing Forum poster",
-                      f != null ? f.toPath() : null);
+                      f != null
+                          ? SaveGameParameter.builder()
+                              .path(f.toPath())
+                              .displayName("Test.jpg")
```
RoiEXLab, line 60: Fun Fact:
I changed it back to the original name: https://github.com/triplea-game/triplea/blame/a0d65198458522e08abc9fba36543a1494cb2058/game-core/src/main/java/games/strategy/engine/framework/startup/ui/editors/ForumPosterEditor.java#L139

game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbForumPoster.java
```diff
@@ -106,12 +113,12 @@ public static NodeBbForumPoster newInstanceByName(final String name, final int t
    * @return true if the post was successful
    */
   public CompletableFuture<String> postTurnSummary(
-      final String summary, final String title, final Path path) {
+      final String summary, final String title, @Nullable final SaveGameParameter saveGame) {
```
DanVanAtta, line 116: I'm not sure if parameter object is the right pattern here. I suspect factory methods might be more appropriate.

We have two invocations of this method:

```
                .postTurnSummary(
                    (gameNameAndInfo + "\n\n" + turnSummary),
                    "TripleA " + title + ": " + currentPlayer.getName() + " round " + roundNumber,
                    SaveGameParameter.builder()
                        .path(saveGameFile.toPath())
                        .displayName(saveGameName)
                        .build());
```
```
                      "Test summary from TripleA, engine version: "
                          + ClientContext.engineVersion()
                          + ", time: "
                          + DateTimeUtil.getLocalizedTime(),
                      "Testing Forum poster",
                      f != null
                          ? SaveGameParameter.builder()
                              .path(f.toPath())
                              .displayName("Test.jpg")
                              .build()
                          : null);
```

Having the nullable parameter value is a big code smell. 



I think the two APIs probably should be something more like:

```
              final CompletableFuture<String> future =
                  poster.postTestSummary();
```

And: 
```
              final CompletableFuture<String> future =
                  poster.post(gameData, historyLog, saveGameFile);
```

-----------

The odd thing about a parameter object here is that it's not capturing all of the parameters. It's really just a value object.

It would also be cool to see some of this logic extracted yet further and then tested so we can avoid such a regression going forward.

RoiEXLab: I see your point. The reason I decided to implement it like this is to ensure that we don't end up in a state where for some reason one parameter is null and the other isn't

DanVanAtta: Makes sense, any thoughts to rename the object? Any thoughts on how we might be able to test this? I suspect the test decision would drive most of the rest of the choices.

Side-note, given we have save game in so many places, it's really quite the true violation of DRY.

DanVanAtta:
We could use the code fix sooner.

@RoiEXLab could you follow up with renaming the parameter object since it is a value object (not a parameter object). Second, please consider testing this update.

DanVanAtta:
Squash merged, updated the commit comment "fix PBF name" and added a description to it.

DanVanAtta:
:hankey: 
Should have read this more carefully first: "Haven't actually tested the changes but they should do the trick"

While the honesty is spot-on, we need to do smoke testing and have unit tests for all updates...

RoiEXLab:
Some context for this change: I squeezed the changes into a time slot where I should've been doing other things 😅
That's why I really didn't have the time to check if the changes actually work

panther2:
@RoiEXLab 
While working on a new issue I realized that a new savegames appears as "triplea_20781_1rus.tsvg", 

Looks good 🙂

PR #6750, opened 2020-06-23
Fix empty map selection rendering bug

DanVanAtta:
Fixes a problem where map selection screen renders blank.
This is caused by bad handling of no-map selected case
where an early return statement was introduced. Instead
the code should go on to render the map selection screen,
this update fixes that.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:


DanVanAtta:
Merging now, break fix.This is an important bug that causes map selection rendering to be blank. We can trigger this if there is a bad map in the list (or really just a 1.9 map that is not compatible with 2.0 - if that map is the default selected map we will trigger this problem).

PR #6753, opened 2020-06-23
Restore deprecated map property costPu

DanVanAtta:
Some players might have ancient maps, the error from the missing property
is non-obvious to recover from. Furthermore, there is misleading error
messaging that the map is not present and a prompt to download, which
might not even result in a fix and then another error when the
game becomes the default selection.

This problem speaks to being able to version maps for when they are
not forward compatible and to be able to download engine specific versions.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  
[] Other:

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/attachments/AbstractUserActionAttachment.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/AbstractUserActionAttachment.java
@@ -9,6 +9,7 @@ import games.strategy.engine.data.MutableProperty;
 import games.strategy.engine.data.Resource;
 import games.strategy.engine.data.changefactory.ChangeFactory;
 import games.strategy.engine.delegate.IDelegateBridge;
+import games.strategy.triplea.Constants;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


DanVanAtta:
Will merge once build completes so we can get an updated release out the door.

Cernelius:
I disagree. In my opinion, the tendency of game-makers of keeping using deprecated options, thus making more and more games having them after they have been deprecated, is a much stronger factor than the issue of not being able still to play maps having such options.

A good example (amongst others) is the "trigger", instead of "conditions", option, that I keep seeing used in new games despite having been implicitly deprecated since about a decade!

Cernelius:
And I'm quite sure one or more of such new games will be eventually added to the repository, thus invalidating mine and @ron-murhammer work in making sure no officially available games have the `costPu` option (assuming I didn't miss any one).

Cernelius:
Though I would not be surprised if, in the interval of time between when me and @ron-murhammer updated all listed downloadable games having the `costPu` option and the recent 2.0 first release, one or more other games having the `costPu` option have made it to the repository (being tested with 1.9). I suppose we can update those too, if this option is kept out of the program, now on.

DanVanAtta:
Cernel, it's not a matter of map makers using the deprecated property. You
have to take a bigger picture.

 If someone does not update their maps, has this property and it's their
default map, they get a bad error. The engine does not handle non-forward
compatible changes well. I simulated this and in combination with another
NPE it made the entire game unplayable.

DanVanAtta:
I think in part there is an assumption that players do update their maps.
Interestingly, the map that triggered this was just an old copy.

PR #6783, opened 2020-06-24
Update headless game logging when cannot load a map

DanVanAtta:
Print the list of available maps when a map cannot be loaded.
Useful to debug why a deployed production bot cannot load
from save game while local running bots with the same version
and same base map zip have no problem.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:


DanVanAtta:
Merging now so this can be deployed to a test bot in production.

PR #6821, opened 2020-06-24
Fix white lines on map with certain zoom levels & scroll offsets.

asvitkine:
Switch to nearest neighbor interpolation for drawing tiles which
does not exhibit the problem. Bicubic also didn't have the
problem, but it was extremely slow (2ms per tile as opposed to
0.5ms with bilinear and 0.3ms with nearest neighbor).

Also expands PerfTimer with more functionality:
  - Showing running average.
  - Passing a frequency param to print stats every N calls.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix
[] Other:   

## Testing
Manually tested on GoT map w/ 66% zoom with scroll offset being forced-set via the following (to easily repro):
```
diff --git a/game-core/src/main/java/games/strategy/ui/ImageScrollModel.java b/game-core/src/main/java/games/strategy/ui/ImageScrollModel.java
index b762bed62..1e62a7559 100644
--- a/game-core/src/main/java/games/strategy/ui/ImageScrollModel.java
+++ b/game-core/src/main/java/games/strategy/ui/ImageScrollModel.java
@@ -131,7 +131,10 @@ public class ImageScrollModel {
   void set(final int x, final int y) {
     this.x = x;
     this.y = y;
+    this.x = 1458;
+    this.y = 2947;
     enforceBounds();
     updateListeners();
   }
 }
```

The above confirmed the fix (no white line).

For performance, I measured using PerfTimer around the following call:
```
-      g.drawImage(
-          tile.getImage(),
-          AffineTransform.getTranslateInstance(tile.getBounds().x, tile.getBounds().y),
-          this);
+      var timer = org.triplea.performance.PerfTimer.startTimer("Draw Image", 100);
+      g.drawImage(tile.getImage(), tile.getBounds().x, tile.getBounds().y, this);
+      timer.close();
```

## Screens Shots

Before:
Screen Shot 2020-06-24 at 5 37 09 PM

After:
Screen Shot 2020-06-24 at 5 36 17 PM

## Release Note

FIX|White lines appearing on map at certain zoom levels & scroll offsets.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
@@ -644,8 +644,14 @@ public class MapPanel extends ImageScrollerLargeView {
   public void paint(final Graphics g) {
     final Graphics2D g2d = (Graphics2D) g;
     g2d.setRenderingHint(RenderingHints.KEY_RENDERING, RenderingHints.VALUE_RENDER_SPEED);
+    // Note: Using nearest neighbor interpolation produces the best results compared to
+    // bilinear and bicubic. Bilinear has problems with sometimes producing white lines at
+    // certain zoom levels and scroll offsets. Bicubic is extremely slow (2ms per tile drawn).
+    // Nearest neighbor does not appear to have the white line problem and is actually the
+    // fastest.
     g2d.setRenderingHint(
-        RenderingHints.KEY_INTERPOLATION, RenderingHints.VALUE_INTERPOLATION_BILINEAR);
+        RenderingHints.KEY_INTERPOLATION, RenderingHints.VALUE_INTERPOLATION_NEAREST_NEIGHBOR);
+
     super.paint(g2d);
     g2d.scale(scale, scale);
     g2d.clip(new Rectangle2D.Double(0, 0, getImageWidth(), getImageHeight()));
@@ -808,10 +814,7 @@ public class MapPanel extends ImageScrollerLargeView {
       } else {
         images.add(tile);
       }
-      g.drawImage(
-          tile.getImage(),
-          AffineTransform.getTranslateInstance(tile.getBounds().x, tile.getBounds().y),
-          this);
+      g.drawImage(tile.getImage(), tile.getBounds().x, tile.getBounds().y, this);
     }
     g.translate(bounds.getX(), bounds.getY());
   }
--- a/game-core/src/main/java/org/triplea/performance/PerfTimer.java
+++ b/game-core/src/main/java/org/triplea/performance/PerfTimer.java
@@ -23,14 +23,16 @@ import lombok.extern.java.Log;
 @Log
 public class PerfTimer implements Closeable {
 
-  private static final PerfTimer DISABLED_TIMER = new PerfTimer("disabled");
   private static final Map<String, AtomicLong> runningTotal = new HashMap<>();
-  final String title;
+  private static final Map<String, AtomicLong> runningCount = new HashMap<>();
+  private final String title;
   private final long startMillis;
+  private final int reportingFrequency;
 
-  private PerfTimer(final String title) {
+  private PerfTimer(final String title, int reportingFrequency) {
     this.title = title;
-    this.startMillis = System.nanoTime();
+    this.reportingFrequency = reportingFrequency;
+    this.startMillis = this.reportingFrequency > 0 ? System.nanoTime() : 0;
   }
 
   private long stopTimer() {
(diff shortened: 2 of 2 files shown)


DanVanAtta:


```diff

Complexity increasing per file
==============================
- game-core/src/main/java/org/triplea/performance/PerfTimer.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=5753815&bid=18835535)

DanVanAtta:


```diff

Complexity increasing per file
==============================
- game-core/src/main/java/org/triplea/performance/PerfTimer.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=5753815&bid=18835535)

Cernelius:
I really believe the developers in general should make a council and generally decide with what approximation TripleA is going to go, for the time being, when any zoom is used (on whatever), as that is better being something remaining fairly stable overtime (especially for maps meant to be frequently zoomed).

I strongly disagree with going with nearest neighbor over bilinear or bicubic, as I cannot imagine performance can be bad enough to justify the use of such a sub-optimal approximation.

And, yes, the matter has been already analyzed in the past. I guess we could go through the same process every couple of years, for several hours every time, and the results will be rediscovering:
>some maps where nearest neighbour looked really bad.

asvitkine:
To be clear, the actual content of the tiles is still rendered via bicubic interpolation. It's just the tile images themselves that get rendered with nearest neighbour.

Bilinear was clearly not working correctly since it was resulting in random white lines per my screenshot. So this is a bugfix. (The performance aspect is orthogonal and I was mostly just checking this didn't regress it - bicubic did, so it's good I verified since that's what I tried first.)

Cernelius:
Wouldn't it be better to find what is causing the white lines, and fix that, instead of changing to a worse rendering, because that happens to fix the issue (if I'm actually understanding the matter)?

Cernelius:
> To be clear, the actual content of the tiles is still rendered via bicubic interpolation. It's just the tile images themselves that get rendered with nearest neighbour.

In my mind, the general zoom feature should work on the final image offered to the viewer, like if you take a screenshot and, then, shrink it. Meaning first putting the various elements together, then merge all these "layers" into a single image, then zoom that (for example, the unit image zoom should be applied before applying the general zoom).

DanVanAtta:
@asvitkine beyond fixing the white lines, are there any other visual differences that you were able to observe? If not, and if you could do some smoke checking (check a few other maps at a few different zooms, Napoleonic Empires, TWW, Revised, Civil War and perhaps LOTR as examples), then I personally have no objection to this at all and consider this a straight bug-fix.

With that said, thank and I really appreciate the investigation and care to check the performance was not negatively impacted :+1: 

------------

@Cernelius in which cases would you expect or have you observed "worse rendering?" 

I do agree that zoom should probably work quite a bit differently. The change you're describing is more akin to completely remodelling a bathroom where here we've just fixed a leaky pipe.

> I really believe the developers in general should make a council 

We don't really have a council. We post github issues to discuss generic questions and to get opinions from everyone, and we go with concensus based on that. Most any major change will go through that process. In a way I suppose that is a council. Otherwise if push comes to shove then Ron is the decision maker followed by me if he is not available. We try to do most things by consensus though and via the discussion issues.

asvitkine:
Looking closer at my screenshots, there are visual differences as a result of this change. The text that's printed on the map background does end up looking worse. But I couldn't tell this normally - only when I take my screenshots and zoom them in an image editor and the switch between the two. During actual gameplay, the difference is not perceptible to me.

Here's what it looks like when zooming in at the details - you can see how thick the single-pixel white line to see how much I had to zoom in after to see it. These are the same images I posted originally, so I'm curious if anyone can tell the difference from them at their regular size without having to zoom in.

Screen Shot 2020-06-25 at 9 16 00 AM
Screen Shot 2020-06-25 at 9 16 06 AM

As for finding a different way to fix it? I think it would be great, but I currently don't know of how that would work. As far as I could tell from my debugging, the issue really is with the bilinear interpolation implementation in Java2D. The coordinates we're passing for the tile bounds up line up correctly - it's just the interpolation is not working. But I can look more...

Cernelius:
I'm not having white lines in anything I'm playing recently, but, having to choose, I'd go with white lines and bilinear over no white lines and nearest neighbour.

Cernelius:
I wouldn't be surprised if this is actually a map issue, and, like, the heigh and wideness have been defined incorrectly with respect to the tiles or something (no idea). This was one of the most bugged maps ever added to the repository (which says something), until relatively recently, when someone fixed it. If the white lines are happening only in this map, the problem is almost currently ineffectual (due to the map's popularity).

I would be curious so far how many developers have opened the map and tested it, because I've tested it some and never managed to hit any white lines at all, the zoom working perfectly, as of TripleA 2.0.20234. So I'm really wondering how many other users are whitnessing the white lines that @asvitkine is reporting.

asvitkine:
I'm going to see if some other approach can't be used to resolve this.

I noticed this happens at 50% zoom when the scroll offset is at an odd number (so boundaries between the tiles math out to fractional pixels). There are things we can do without changing the rendering mode to try to mitigate - for example, not allow scroll offsets that are fractional pixels (based on zoom level). That might help.

We could also have the tiles' content be at the zoom level, rather than at the non-zoomed resolution. I suspect it will fix the issue for the 50% zoom but probably not in the general case.

Going to try the scroll offset approach.

In the meantime, I've split the perftimer changes from this PR into a separate one.

asvitkine:
Hmm, the approach of avoiding fractional pixel boundaries doesn't work for the general case.
For example, at 66% zoom, offset 20 maps to 13.2, which if we round to 13 ends up being 19.69 when we convert back.

Which we can either round up to 20 (essentially no effect) or round down to 19 (but then 19 results in 12.54 which results in 18.18 if we round up)...

So basically the approach I was using only works for 50% in any reasonable way. So either we could special case 50% (which does fix the issue), or we go back to the drawing board...

DanVanAtta:
Let's continue any discussion in a problem issue please. I'd like to clear this PR off of the queue so we can keep focus on getting any in-flight changes reviewed and hopefully merged.

@asvitkine your efforts here are appreciated, I look forward to more discussion and getting this rendering issue fixed.

Cernelius:
Why is this PR being referenced in the changelog?

https://triplea-game.org/release_notes/

Has the "white lines appearing on map at certain zoom levels & scroll offsets" problem been actually fixed? Is now the zoom using nearest neighbour instead of bilinear?

DanVanAtta:
Good catch @Cernelius , it's a problem in the release note generating script to not filter un-merged PRs.

Issue #6774, opened 2020-06-23
[Urgent] Bots unable to load WW_II_Revised saves


DanVanAtta:
### How can the problem be recreated?

- start a game on bot server, ww_II_Revised
- save
- try loading it

If reproduced, game load should do roughly nothing and observe the following in logs:
```
Jun 23 19:56:22 li1116-193 run_server[2853509]: INFO: Changed to game map: World War II Revised
Jun 23 19:56:26 li1116-193 run_server[2853509]: INFO: Game mapName not in available games listing: world_war_ii_revised
```

### If playing a prerelease, which version are you using?

Game Version:

### If playing a prerelease, does this happen on the latest release?

- [ ] yes
- [ ] no

DanVanAtta:
Custom deployed v2.0.20193 to a bot in prod2 and could not reproduce this problem...   :anguished:

DanVanAtta:
This will perhaps simply be fixed by deploying latest to all bots. I learned in this we should have confidence/faith in the prerelease bot and be sure to use that for our "what happens with the latest version" testing that we want to be actually deployed out. FWIW, local, prerelease, and the test server in prod all on latest were consistently behaved.

The 'thing' that is impacting bots on 2.0.20183 causing save game to not load is mysterious. This is also a place where error handling from bot to user could be better, the save game fails to load silently.

DanVanAtta:
Deployed a 2.0.20193 to prod bot 02; hit the same error!

DanVanAtta:
With some more debugging, the situation is expected. Why, hard to say.

The list of available games does not include "world_war_ii_revised", but we do parse the map zip for it:

> Jun 24 06:48:01 li1232-106 run_server[2989014]: INFO: Loading map: /home/bot/maps/world_war_ii_revised-master.zip

> INFO: Done loading maps, availableMapFolderOrZipNames count: 137, contents: [age_of_tribes-master, greyhawk-master, ww2v3_11n-master, charles_de_gau
lle_1939-master, iron_war_europe-master, command_and_conquer-master, operation_market_garden-master, battle_for_arda-master, middle_east-master, ww2v3_variants-master, sengoku_jidai-master, new_wo
rld_order-master, twilight_imperium-master, d-day-master, fallen_empire-master, world_war_ii_revised_variations-master, diplomacy-master, arnhem-master, caribbean_trade_war-master, big_world-maste
[...]

PR #6834, opened 2020-06-25
Headless map loading

DanVanAtta:
Bug fix in: https://github.com/triplea-game/triplea/commit/33fd4e565bf7cab7e9bb65d39d85c5fd81fb3a8d

Bots keep a list of both map zips and map folders, and a map of game names to the URL of the XML (whether in a directory or a zip). The first list is used only when loading save games. We can instead lookup save games by their name which uses the map. We *only* keep the first list of zips and map folders for save game lookup.

This update moves lookup of savegames from folder or zip name, to be by game name. First, this allows us to simplify quite a bit as we no longer need to keep track of folder and zip names. Second, this is a major bug fix as the list of folders and saves is not reliable. On prerelease, local, hand deployed servers, we see consistently we get all map zips and folder names loaded, but all production2 bots are consistently dropping the 'world_war_ii-master' map zip name. But, both prerelease and production2 bots do agree on the list of parsed game names. Hence, by looking up save games by parsed game name, we expect to fix the problem where world_war_ii_revised game saves cannot be loaded. Those saves cannot be loaded because the WWII zip name is missing from our list of available zips. Why exactly the folder is missing is still unknown.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix:  https://github.com/triplea-game/triplea/issues/6774
[] Other:   

## Testing

Verified on a local bot:
- could select and start a game
- could load an autosave and start a game
- could load a WWII revised save game and start it

## Release Note

// No release note as changes are all for headless bot, users do not need to download an updated version to pick up this fix. Will merge and deploy this to a few sample bots to verify fix.


DanVanAtta:
@RoiEXLab , @trevan please do a post-merge review of this if you would.
I'm going to merge shortly to get this built and then deploy to a select bot to see if we finally can get a bug fix out for WWII revised save games not loading. The problem has really stumped me. I've been able to determine that the list of maps and zips is missing just WWII, and just on prod2 bots. Pre-release and locally hosted bots do not have this problem, frustrating yet, I deployed a new bot server to bot2 that had a single game instance and it also did not have a problem.

PR #6842, opened 2020-06-25
Indicate if the missing image is for a damaged or disabled unit

trevan:
This adds some additional text to the 'MISSING UNIT IMAGE' error message to indicate if the image should be a damaged and/or disabled image.  That should help in figuring out what image is actually missing.  It also might help in figuring out if the game is requesting these images more in 2.0 vs 1.9.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix
[] Other:   

## Release Note

CHANGE|Missing Unit Image error now indicates if image should be a damaged and/or disabled image

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/screen/UnitsDrawer.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/screen/UnitsDrawer.java
@@ -112,21 +112,21 @@ public class UnitsDrawer extends AbstractDrawable {
     }
     final GamePlayer owner = data.getPlayerList().getPlayerId(playerName);
     final boolean damagedImage = damaged > 0 || bombingUnitDamage > 0;
-    final boolean disabledImage = disabled;
     final Optional<Image> img =
-        uiContext.getUnitImageFactory().getImage(type, owner, damagedImage, disabledImage);
+        uiContext.getUnitImageFactory().getImage(type, owner, damagedImage, disabled);
 
     if (img.isEmpty() && !uiContext.isShutDown()) {
-      final List<String> imageQualifiers = new ArrayList<>();
-      if (damagedImage) {
-        imageQualifiers.add("damaged ");
-      }
-      if (disabledImage) {
-        imageQualifiers.add("disabled ");
+      final String imageQualifier;
+      if (disabled) {
+        imageQualifier = "disabled ";
+      } else if (damagedImage) {
+        imageQualifier = "damaged ";
+      } else {
+        imageQualifier = "";
       }
       log.severe(
           "MISSING UNIT IMAGE (won't be displayed): "
-              + String.join("and ", imageQualifiers)
+              + imageQualifier
               + type.getName()
               + " owned by "
               + owner.getName()
--- a/game-core/src/main/java/games/strategy/triplea/ui/screen/UnitsDrawer.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/screen/UnitsDrawer.java
@@ -19,6 +19,7 @@ import java.awt.Image;
 import java.awt.Point;
 import java.awt.Rectangle;
 import java.awt.RenderingHints;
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/triplea/ui/screen/UnitsDrawer.java
```diff
@@ -110,14 +111,22 @@ public void draw(
       throw new IllegalStateException("Type not found:" + unitType);
     }
     final GamePlayer owner = data.getPlayerList().getPlayerId(playerName);
+    final boolean damagedImage = damaged > 0 || bombingUnitDamage > 0;
+    final boolean disabledImage = disabled;
```
DanVanAtta, line 115: nit, what's the motivation to re-assign `disabled` to a new variable?

game-core/src/main/java/games/strategy/triplea/ui/screen/UnitsDrawer.java
```diff
@@ -110,14 +111,22 @@ public void draw(
...
-            .getImage(type, owner, damaged > 0 || bombingUnitDamage > 0, disabled);
+        uiContext.getUnitImageFactory().getImage(type, owner, damagedImage, disabledImage);
 
     if (img.isEmpty() && !uiContext.isShutDown()) {
+      final List<String> imageQualifiers = new ArrayList<>();
+      if (damagedImage) {
+        imageQualifiers.add("damaged ");
+      }
+      if (disabledImage) {
+        imageQualifiers.add("disabled ");
```
DanVanAtta, line 125: We can have, or even require sometimes a disabled & damaged image? I kinda thought they were mutually exclusive. If not, it implies a pretty large number of image permutations that is probably not typically included (which would be kinda bad news).

game-core/src/main/java/games/strategy/triplea/ui/screen/UnitsDrawer.java
trevan: Attempt to ensure that the same values passed in the getImage call would be used in the log message.

game-core/src/main/java/games/strategy/triplea/ui/screen/UnitsDrawer.java
trevan: Ah, looks like I missed the "else" in the UnitImageFactory.  I somehow misread it as two `if` statements instead of a `if else if`.

game-core/src/main/java/games/strategy/triplea/ui/screen/UnitsDrawer.java
DanVanAtta: Is it not thread safe?
If we're trying to protect against concurrency, what's to keep the value from changing before this line? (in which case, unless I'm mistaken, the log message is also incorrect)

trevan: It should be thread safe.

This was to protect a developer from changing the condition for the call and not changing the condition for the log call.  But I've removed the disabledImage variable and now just use disabled.

game-core/src/main/java/games/strategy/triplea/ui/screen/UnitsDrawer.java
```diff
@@ -110,14 +110,22 @@ public void draw(
...
+      final String imageQualifier;
+      if (disabled) {
+        imageQualifier = "disabled ";
+      } else if (damagedImage) {
+        imageQualifier = "damaged ";
+      } else {
+        imageQualifier = "";
+      }
       log.severe(
           "MISSING UNIT IMAGE (won't be displayed): "
```
RoiEXLab, line 127: I think this error message shouldn't be all caps tbh

PR #6852, opened 2020-06-25
Fix UnsupportedOperationException trying to modify a collection.

asvitkine:
Fix UnsupportedOperationException trying to modify a collection.

Make a copy of the collection when calling remove().

Fixes: https://github.com/triplea-game/triplea/issues/6818

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix https://github.com/triplea-game/triplea/issues/6818
[] Other:   

## Testing

None. Found the issue via the stack trace from the error report.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/AbstractBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/AbstractBattle.java
@@ -104,7 +104,7 @@ abstract class AbstractBattle implements IBattle {
     }
     final var unitsCopy = new ArrayList<>(units);
     unitsCopy.addAll(dependentUnits);
-    return unitsCopy;
+    return Collections.unmodifiableList(unitsCopy);
   }
 
   void addDependentTransportingUnits(final Collection<Unit> units) {
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/AbstractBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/AbstractBattle.java
@@ -95,13 +95,13 @@ abstract class AbstractBattle implements IBattle {
         dependentUnits.addAll(dependent);
       }
     }
-    return Collections.unmodifiableList(dependentUnits);
+    return Collections.unmodifiableCollection(dependentUnits);
   }
 
-  protected Collection<Unit> getWithDependents(final Collection<Unit> units) {
+  protected Collection<Unit> getUnitsWithDependents(final Collection<Unit> units) {
     final Collection<Unit> dependentUnits = getDependentUnits(units);
     if (dependentUnits.isEmpty()) {
-      return units;
+      return Collections.unmodifiableCollection(units);
     }
     return ImmutableList.copyOf(Iterables.concat(units, dependentUnits));
   }
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/AbstractBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/AbstractBattle.java
@@ -86,7 +86,7 @@ abstract class AbstractBattle implements IBattle {
   }
 
   @Override
-  public List<Unit> getDependentUnits(final Collection<Unit> units) {
+  public Collection<Unit> getDependentUnits(final Collection<Unit> units) {
     final List<Unit> dependentUnits = new ArrayList<>();
     for (final Unit unit : units) {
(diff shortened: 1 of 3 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -639,7 +639,7 @@ private void removeFromNonCombatLandings(
       if (landedTerritory == null) {
         throw new IllegalStateException("not unloaded?:" + units);
       }
-      remove(lost, bridge, landedTerritory, false);
+      remove(new ArrayList<>(lost), bridge, landedTerritory, false);
```
DanVanAtta, line 642: I'm not sure if this is the best fix. Creating a copy probably should be the responsibility of the method, not the caller. Usually we would want APIs to be impossible to use and not have arcane contracts where just by using the API you break it.

It's also a bit odd as it's difficult to see what the side effect would be from either the parameters or the method name (yikes, the code here really could have used some improvements..)

What's the cost to moving the copy to the method? Failing that, perhaps it should be commented here why we create a copy, but that leaves unresolved that we have a *very* brittle API that can break just by using it in a very reasonable way.

asvitkine: Done. I made a helper function which is also reused from another remove() implementation that avoids the extra copy if there are no dependent units.

game-core/src/main/java/games/strategy/triplea/delegate/battle/AbstractBattle.java
```diff
@@ -97,6 +98,14 @@
     return Collections.unmodifiableList(dependentUnits);
   }
 
+  protected Collection<Unit> getWithDependents(final Collection<Unit> units) {
+    final Collection<Unit> dependentUnits = getDependentUnits(units);
+    if (dependentUnits.isEmpty()) {
+      return units;
+    }
+    return ImmutableList.copyOf(Iterables.concat(units, dependentUnits));
```
DanVanAtta, line 106: Does `concat` of a non-empty collection with an empty collection return just the non-empty collection? Perhaps we should just let the API do that for us?

In, part, it's strange that if you pass an mutable list to `getWithDependents` that is empty, you get a mutable list back from this method, but passing a non-empty mutable list returns an immutable collection. Generally a returned collection should always be mutable or immutable and not depend on its parameter.

DanVanAtta: `getWithDependents` is notable to me as a method name. It makes me wonder 'get what?'

`getUnitWithDependents` makes perhaps more sense. Though, if we have a one-liner method, perhaps just in-lining this in the two places its used is just as well? WDYT?

asvitkine: > Does `concat` of a non-empty collection with an empty collection return just the non-empty collection? Perhaps we should just let the API do that for us?

It returns an iterable. Creating a list of it does require a copy.

> 
> In, part, it's strange that if you pass an mutable list to `getWithDependents` that is empty, you get a mutable list back from this method, but passing a non-empty mutable list returns an immutable collection. Generally a returned collection should always be mutable or immutable and not depend on its parameter.

We can wrap units in an immutable wrapper for consistency.

asvitkine: Renamed to `getUnitsWithDependents` and made the return value unmodifiable.

DanVanAtta: Sorry to not have been clear in my question:

When calling:
```ImmutableList.copyOf(Iterables.concat(units, dependentUnits))```

Does it matter if `dependentUnits` is empty? Wouldn't we still get the same result?

DanVanAtta: Side-note, I'm not sure we should be completely zeolous to wrap collections in unmodifiable variants. IMO it's far more important for the contents to be immutable than a collection.

We were concerned about performance of unnecessary collection copies, seems we would be falling into that potential problem more and more. One thing to do is for us to simply stop mutating collections though and focus on instead of copying collections, to instead do stream operations on them which creates a new collection for us. This way we avoid both mutating collections and copying them explicitly as part of return or input arguments.

PR #6872, opened 2020-06-26
Use Streams in a functional way to create map

RoiEXLab:
Follow-up to #6834

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix
[] Other:

### DIFF
--- a/game-core/src/main/java/org/triplea/game/server/AvailableGames.java
+++ b/game-core/src/main/java/org/triplea/game/server/AvailableGames.java
@@ -10,12 +10,13 @@ import java.io.InputStream;
 import java.net.URI;
 import java.net.URL;
 import java.net.URLClassLoader;
-import java.util.Collections;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.logging.Level;
+import java.util.stream.Collectors;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/org/triplea/game/server/AvailableGames.java
```diff
@@ -35,19 +36,22 @@
...
-                availableGames.putAll(getGamesFromDirectory(map));
-              } else if (map.isFile() && map.getName().toLowerCase().endsWith(ZIP_EXTENSION)) {
-                availableGames.putAll(getGamesFromZip(map));
-              }
-            });
+    availableGames =
+        FileUtils.listFiles(ClientFileSystemHelper.getUserMapsFolder())
+            .parallelStream()
+            .<Map<String, URI>>map(
+                map -> {
```
DanVanAtta, line 43: nit, have you considered extracting this to a function?

Often streams are best when there is no or little body (eg; try to convert most of the blocks into method references).

Beyond that, looks fine, feel free to update and merge. (I'll be AFK for a day or two).

DanVanAtta: @RoiEXLab I"m not sure how much we really moved the needle on this one by getting rid of the `forEach`, in some ways the extra syntax introduced makes it perhaps only very marginal of an improvement:

```

            .<Map<String, URI>>map(
        .forEach(
                map -> {
	
```

I do think the big improvement would be to convert the lambda blocks to be method references. Though, at the end of the day having tests drive the design is probably even better.

RoiEXLab: Ah, sorry for not replying, I wanted to extract it to a method (would also get rid of the explicit type parameters), but I simply just haven't got to it yet.

Will follow up soon-ish

PR #6885, opened 2020-06-27
Only check for FirstStrike when deciding to show the FirstStrike step

### LINKED ISSUE #6878: 2.0.20258: BattleStepsPanel#setStep - Could not find step name:Spanish select first strike casualties
## Log Message
Could not find step name:Spanish select first strike casualties

## TripleA Version
2.0.20258

## Java Version
11.0.6

## Operating System
Linux

trevan:
Fixes #6878

To duplicate, you need a map that has an `isSub` unit but no air units.  Then you attack the `isSub` unit with an `isDestroyer` unit.  The `Carribbean Trade War` map has this scenario.  Attack a Sloop with a Battleship and this error will occur.

You can also duplicate it by having an `isFirstStrike` unit that has an empty `canNotTarget`.  I don't know of any maps like that, though.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix
[] Other:   

## Testing
Attacked a Sloop with a Battleship in the Caribbean Trade War.

## Release Note

FIX|Attacking a sub unit with a destroyer unit in a map with no air units will no longer throw an error.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
@@ -201,8 +201,7 @@ public class BattleSteps implements BattleStepStrings, BattleState {
     // also, ww2v3/global rules, defending subs without sneak attack fire with all defenders
     final Collection<Unit> defendingUnitsAliveAndDamaged = new ArrayList<>(defendingUnits);
     defendingUnitsAliveAndDamaged.addAll(defendingWaitingToDie);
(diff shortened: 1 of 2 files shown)


trevan:
I did a git history dive and it looks like this was a typo during the sub attribute separation.  This function had two sub attribute conversions: `isSub` to `isFirstStrike` and `isSub` to `canNotTargetAll`.  Before and after this line change, the conversions were `isSub` to `canNotTargetAll` so it looks like this was an accident.

PR #6487, opened 2020-05-17
Store forum token instead of password

RoiEXLab:
This is what I was able to come up with from #6452
This PR is a WIP, but I wanted to hear your thoughts on the UI decisions.

Basically the user never really gets to see the token and we only store the username for display purposes, so the user can tell their settings have some sort of credentials stored.
We also store a uid to be able to properly revoke the token again.

Whenever a user clicks "save" now in engine preferences, we make a request to the proper forum and create a new token. There's no UI mechanism in NodeBB for this outside of the admin panel. Tokens are revoked if
1. A User creates a new token aka changes their username in the UI and clicks save
2. The Setting is cleared.

This has 2 benefits:
1. It works with 2FA which is the main goal of this PR
2. We no longer have to store any forum password, just a random token which is huge for security (maybe we want to consider explicitly clearing stored passwords to prevent them from floating around in the registry?)

Open question: How should we implement this on the "on screen"/not-engine-preferences login fields? Same UI like the engine preferences or just a pre-filled token field?

## Functional Changes

[] New map or map update
[] New Feature
[x] Feature update or enhancement
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix:  
[] Other:   

## Testing

I verified basic posting works with this new approach.

## Release Note

Feature|Support for 2FA for PbF<br>
Enhancement|PbF now stores login tokens instead of passwords<br>
Fix|Usernames with spaces work for PbF

PR #6925, opened 2020-06-29
Begin extracting nodeBB forum http interface

DanVanAtta:
Starting with the 'getUserId' API, extract from NodeBbForumPoster
the HTTP client code. This will allow us to add a wiremock
test and then verify that we properly encode the URL for userId
lookups. This will then lead to a bug fix where the URL formed
by user names that have spaces in them are not properly URL encoded.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix
[] Other:   

@RoiEXLab opening as a draft for any feedback, coordination and planning. To be able to actually test the URL encoding problem we need to have a test in place. I suppose there is argument for doing a blind fix and manually testing..

The direction I think we should move is to extract the HTTP client to its own interface and let the forum poster code execute the client actions async. It already looks like it would be simpler to move to feign and would reduce nearly all of the JSON posting code. Since we need to do manual testing, it's perhaps a good time do that now and then ensure our automated testing matches the results from manual curls.  This would get us mostly out of the manual testing trap.

Arguably we should move to feign probably sooner rather than later as well. If we complete the token work, I don't think we'll want to touch this code again, particularly as it'll then be somewhat complex and would require a round of manual testing that needs a lot of context (and frankly I don't think we'll be up for it unless we really need to, and then again the argument for a blind fix would just be even stronger).

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/posted/game/pbf/ForumPostingException.java
+++ b/game-core/src/main/java/games/strategy/engine/posted/game/pbf/ForumPostingException.java
@@ -1,0 +1,8 @@
+package games.strategy.engine.posted.game.pbf;
+
+public class ForumPostingException extends RuntimeException {
+
+  public ForumPostingException(final Exception e) {
+    super("Failed to communicate with forum: " + e.getMessage(), e);
+  }
+}
--- a/game-core/src/main/java/games/strategy/engine/posted/game/pbf/ForumClient.java
+++ b/game-core/src/main/java/games/strategy/engine/posted/game/pbf/ForumClient.java
@@ -1,0 +1,22 @@
+package games.strategy.engine.posted.game.pbf;
+
+import javax.annotation.Nullable;
+import lombok.Builder;
+import lombok.Value;
+
+public interface ForumClient {
+
+  /**
+   * Returns a user id by user name, if user name is not found then returns an empty optional.
+   *
+   * @throws ForumPostingException Thrown if there are any errors communicating with forum.
+   */
+  GetUserIdResult getUserId(String userName);
+
+  @Builder
+  @Value
+  class GetUserIdResult {
+    @Nullable String errorMessage;
+    @Nullable Integer userId;
+  }
+}
--- a/game-core/src/main/java/games/strategy/engine/posted/game/pbf/HttpForumClient.java
+++ b/game-core/src/main/java/games/strategy/engine/posted/game/pbf/HttpForumClient.java
@@ -1,0 +1,61 @@
+package games.strategy.engine.posted.game.pbf;
+
+import games.strategy.engine.framework.system.HttpProxy;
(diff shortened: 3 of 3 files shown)


game-core/src/main/java/games/strategy/engine/posted/game/pbf/ForumPostingException.java
```diff
@@ -0,0 +1,8 @@
+package games.strategy.engine.posted.game.pbf;
+
+public class ForumPostingException extends RuntimeException {
```
DanVanAtta, line 3: Codacy found an issue: [Classes implementing Serializable should set a serialVersionUID](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=5774013)

DanVanAtta:


```diff

Issues
======
- Added 1
           

Coverage increased per file
===========================
+ game-core/src/main/java/games/strategy/net/nio/Decoder.java  2
         

Complexity increasing per file
==============================
- game-core/src/main/java/games/strategy/engine/posted/game/pbf/HttpForumClient.java  4
- game-core/src/main/java/games/strategy/engine/posted/game/pbf/ForumPostingException.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=5774013&bid=18898625)

DanVanAtta:
@RoiEXLab have you had any thoughts on direction? If we convert the nodeBB interaction to feign first we can then test it, consolidate technology (consistency!), and then convert to using tokens having tested code. If we skip that then this probably can be closed, but I suspect if we skip the migration and go directly to tokens we will never come back and convert to a unified HTTP language nor get the testing in place.

DanVanAtta:
@RoiEXLab bump

RoiEXLab:
Ah sorry for the delayed reponse, "real-life" is super time consuming this month, but I will have holidays soon-ish so I will be able to work a little bit more on TripleA again.

Anyways regarding the proposed changes:
I would agree that if we want up to test our code feign is the way to go. It would definitely allow us to "simulate" a forum Server when testing even when we don't have an actual one.
However during the creation of #6487 I realized that the actual networking code is the smallest problem I ran into, the most problems were encountered when I was trying to route the launching code to correctly fetch a new token.
That's why I doubt that migrating to feign will allow us to add good tests before we don't "fix" the game launching code 😞 

That's why I think #6487 should be merged to allow for 2FA, even if that would mean that there's no pressing reason to migrate right now.
In case that helps: I know for a fact that the guys at NodeBB are planning to make fundamental changes to the write-api we're currently using, which will eventually land in core. So sooner or later we'll have to actually touch the "real network logic" code anyways, which will provide a good (in my opinion better, because it can be timed with upgrading the forum) oppurtunity to migrate.

PR #6926, opened 2020-06-29
Compute maps folder location rather than storing as client setting

DanVanAtta:
This update removes client setting storage of default maps folder
and instead computes the location when it is not overriden.
The default maps folder location is not a value the user would configure
and hence is not appropriate to be in client settings.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix
[] Other:   

It seems possible that storing the map folder as a client setting could cause problems if we have users install TripleA to a new location and delete their old. I'll do some follow-up testing to see if this is also a problem fix. In the meantime the update to not store map folder is IMO correct. The client settings should generally not be so aware of implementation details of like how to find a default folder path and should generally try to have primitive values.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/ClientFileSystemHelper.java
+++ b/game-core/src/main/java/games/strategy/engine/ClientFileSystemHelper.java
@@ -105,7 +105,7 @@ public final class ClientFileSystemHelper {
    *     installations. Users can override this location in settings.
    */
   public static File getUserMapsFolder() {
-    return getUserMapsFolder(ClientFileSystemHelper::getRootFolder);
+    return getUserMapsFolder(ClientFileSystemHelper::getUserRootFolder);
   }
 
   @VisibleForTesting
--- a/game-core/src/main/java/games/strategy/engine/ClientFileSystemHelper.java
+++ b/game-core/src/main/java/games/strategy/engine/ClientFileSystemHelper.java
@@ -3,12 +3,12 @@ package games.strategy.engine;
 import com.google.common.annotations.VisibleForTesting;
 import games.strategy.engine.framework.system.SystemProperties;
 import games.strategy.triplea.settings.ClientSetting;
-import games.strategy.triplea.settings.GameSetting;
 import java.io.File;
 import java.io.IOException;
 import java.net.URISyntaxException;
 import java.nio.file.Path;
 import java.security.CodeSource;
+import java.util.function.Supplier;
 import javax.annotation.Nullable;
 import lombok.extern.java.Log;
 import org.triplea.game.ApplicationContext;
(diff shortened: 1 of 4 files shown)


game-core/src/main/java/games/strategy/engine/ClientFileSystemHelper.java
```diff
@@ -105,25 +105,22 @@ public static File getUserRootFolder() {
...
-        getUserMapsFolder(ClientSetting.userMapsFolderPath, ClientSetting.mapFolderOverride)
-            .toFile();
-    if (!mapsFolder.exists()) {
-      mapsFolder.mkdirs();
-    }
-    if (!mapsFolder.exists()) {
-      log.severe("Error, downloaded maps folder does not exist: " + mapsFolder.getAbsolutePath());
-    }
-    return mapsFolder;
+    return getUserMapsFolder(ClientFileSystemHelper::getRootFolder);
```
RoiEXLab, line 108: `getRootFolder()` is a little bit weird tbh.
On a clean system the triplea folder will never be inside "documents", only if you manually create this folder it will end up storing it there indefinitely.

game-core/src/main/java/games/strategy/engine/ClientFileSystemHelper.java
```diff
@@ -105,25 +105,22 @@ public static File getUserRootFolder() {
...
-      final GameSetting<Path> currentUserMapsFolderSetting,
-      final GameSetting<Path> overrideUserMapsFolderSetting) {
-    return overrideUserMapsFolderSetting.isSet()
-        ? overrideUserMapsFolderSetting.getValueOrThrow()
-        : currentUserMapsFolderSetting.getValueOrThrow();
+  static File getUserMapsFolder(final Supplier<File> userHomeRootFolderSupplier) {
+    final File mapsFolder =
+        ClientSetting.mapFolderOverride
+            .getValue()
+            .map(Path::toFile)
```
RoiEXLab, line 116: Some day we can migrate all of the APIs to use `java.nio.Path` ^^

DanVanAtta:
> I don't think I neccessarily agree that the folder shouldn't be configurable, 

The folder for maps is still configurable both via settings and via a command line parameter (the latter useful for bots).

The weird thing is that we stored the default location in client settings and used the settings mechanism to return a computed path. This is perhaps us over-using the client settings and being a bit over-eager with it.

DanVanAtta:
Sigh on the test failure:

> MapDownloadListTest > testAvailable() FAILED
    org.triplea.util.ServiceNotAvailableException: No service available of type 'org.triplea.game.ApplicationContext'

The service discovery is not trivial to understand.. I respect what's happening, but it's not easy. We need to perhaps buckle down on getting rid of the instantiation of a new game process (which would let us post TripleA to Steam!)

PR #6935, opened 2020-06-29
Extract lambda to method

RoiEXLab:
Follow up to #6872

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix
[] Other:

### DIFF
--- a/game-core/src/main/java/org/triplea/game/server/AvailableGames.java
+++ b/game-core/src/main/java/org/triplea/game/server/AvailableGames.java
@@ -39,16 +39,7 @@ final class AvailableGames {
     availableGames =
         FileUtils.listFiles(ClientFileSystemHelper.getUserMapsFolder())
             .parallelStream()
-            .<Map<String, URI>>map(
-                map -> {
-                  log.info("Loading map: " + map);
-                  if (map.isDirectory()) {
-                    return getGamesFromDirectory(map);
-                  } else if (map.isFile() && map.getName().toLowerCase().endsWith(ZIP_EXTENSION)) {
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/org/triplea/game/server/AvailableGames.java
```diff
@@ -58,6 +49,16 @@
             availableGames.keySet().size(), availableGames.keySet()));
   }
 
+  private static Map<String, URI> getGames(final File map) {
+    log.info("Loading map: " + map);
+    if (map.isDirectory()) {
+      return getGamesFromDirectory(map);
+    } else if (map.isFile() && map.getName().toLowerCase().endsWith(ZIP_EXTENSION)) {
+      return getGamesFromZip(map);
+    }
```
DanVanAtta, line 58: code-style side-note, I think we probably *should* be using 'else', it can prevent errors and secondly  it visually makes the logic block look more like a logic block.

If one were to be really consistent here the `else if` would be converted to an `if` statement as well.

game-core/src/main/java/org/triplea/game/server/AvailableGames.java
```diff
@@ -58,6 +49,16 @@
...
+    return Map.of();
```
DanVanAtta, line 59: side-note: I think this original error handling is extremely questionable. Since this is a bot server, we should not be in a no-op case, this indicates a likely configuration error. Returning a no-op is probably a silent error.

PR #6978, opened 2020-07-01
Fix lobby chat line breaking words

DanVanAtta:
Line breaking in words is caused by auto-insert of new lines
at a max line width. This does not account for word boundaries.
We can instead have long lines and rely on JTextArea to do the
line breaking at word boundaries. Longer lines is better than
breaks at odd locations.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix
[] Other:   

## Testing
Checked rendering of long lines in a local lobby

## Release Note

FIX|Fix lobby chat to break on word boundaries

### DIFF
--- a/http-clients/src/main/java/org/triplea/http/client/web/socket/messages/envelopes/chat/ChatReceivedMessage.java
+++ b/http-clients/src/main/java/org/triplea/http/client/web/socket/messages/envelopes/chat/ChatReceivedMessage.java
@@ -1,9 +1,6 @@
 package org.triplea.http.client.web.socket.messages.envelopes.chat;
 
 import com.google.common.base.Ascii;
-import com.google.common.base.Joiner;
-import com.google.common.base.Splitter;
-import java.util.Optional;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import org.triplea.domain.data.UserName;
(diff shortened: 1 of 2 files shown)


Cernelius:
Also on my screen the line ends at about 1/3 of the screen dimension, which means it uses less than 1/2 of the usual chat view. I believe it currently goes to a new line after reaching 80 characters. It would be better for me if it goes to a new line after reaching about 200. Of course, I assume this is not the same for everyone.

What I'm saying is that in my experience the line is cut way too soon, leaving more than half of the chat view blank and pointless.

DanVanAtta:


```diff

Complexity increasing per file
==============================
- http-clients/src/main/java/org/triplea/http/client/web/socket/messages/envelopes/chat/ChatReceivedMessage.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=5783241&bid=18933471)

DanVanAtta:
@Cernelius long lines that wrap are hard to read. Whitespace is not always bad.

With that said, it's better to have good line breaking than to have short lines, which is the trade-off being taken here (which you would be favorable to as this re-introduces long lines).

PR #6984, opened 2020-07-01
De dupe error uploads & Add Map Name to Error Uploads

DanVanAtta:
## Commits

commit 6af5f46092f247f9677cf879073f8e83c0b58919

    De-Dupe Error Report Uploads
    
    Prevent redundant error message uploads by first checking
    with the server if an error report already exists.
    If the error report exists, then we will show a prompt
    to the user and give them a link to the existing error report.
    
    Change Summary:
    - Add columns in error report history table to track the
      title of the error report, reporting game version,
      and the issue link that was created.
    - Add APIs to error reporting DAO to get the link of
      an existing error report by title and game version.
    - Add server API to check if an error report exists
    - Update client so that after clicking 'upload error report',
      we first query the server if an error report exists.
      If one exists, we show the link to the user, othewise
      we show the user the upload form.
    
    Note: we use both title and game version for de-duplication.
      This may lead to some duplication as version number
      changes, but in case we do not fix a problem or if
      the title matches but is actually different in a different
      version we will allow the error report upload.

commit 8ed128a76e98949e7e5b0e438fb43d2da6a5bc0f

    Add field to error report upload for map name

commit e9da7eb10e0e26e6fad48bc0539a0be5d09545ff

    Do not print Exception in its own block (redundant)
    
    Exception is always at the head of the stack trace which
    we already print in the issue body. This update removes
    the redundant 'exception' block.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix
[] Other:   

## Testing
Add an explicit throw exception at the end of gameRunner launch to test out against a local server. Created an error report and then reported and verified that there was a de-dupe message.

### De-Dupe Sequence

Existing screen prompt to upload:
Screenshot from 2020-07-01 00-32-23

Clicking upload, get the de-dupe message:
Screenshot from 2020-07-01 00-32-30

### New Map Name Field
Notice at the bottom of the dialog there is a 'map name' field:
Screenshot from 2020-07-01 01-05-59

Notice in the preview there is a 'map name' section:
Screenshot from 2020-07-01 01-06-09

## Release Note

UPDATE|Error uploads will now recognize duplicates and give you a link to an existing error upload if there is one. Added 'map-name' as an error upload field

### DIFF
--- a/game-core/src/main/java/org/triplea/debug/error/reporting/StackTraceErrorReportFormatter.java
+++ b/game-core/src/main/java/org/triplea/debug/error/reporting/StackTraceErrorReportFormatter.java
@@ -170,11 +170,6 @@ class StackTraceErrorReportFormatter
     try (PrintWriter printWriter = new PrintWriter(outputStream, false, StandardCharsets.UTF_8)) {
       e.printStackTrace(printWriter);
     }
-    return "## Exception \n"
-        + e.getClass().getName()
-        + Optional.ofNullable(e.getMessage()).map(msg -> ": " + msg).orElse("")
-        + "\n\n## Stack Trace\n```\n"
-        + outputStream.toString(StandardCharsets.UTF_8)
-        + "\n```\n\n";
+    return "## Stack Trace\n```\n" + outputStream.toString(StandardCharsets.UTF_8) + "\n```\n\n";
   }
 }
--- a/game-core/src/main/java/org/triplea/debug/error/reporting/StackTraceErrorReportFormatter.java
+++ b/game-core/src/main/java/org/triplea/debug/error/reporting/StackTraceErrorReportFormatter.java
@@ -33,8 +33,9 @@ class StackTraceErrorReportFormatter implements BiFunction<String, LogRecord, Er
   @Override
   public ErrorReportRequest apply(final String userDescription, final LogRecord logRecord) {
     return ErrorReportRequest.builder()
-        .title(versionSupplier.get() + ": " + createTitle(logRecord))
+        .title(createTitle(logRecord))
         .body(buildBody(userDescription, logRecord))
+        .gameVersion(versionSupplier.get())
         .build();
   }
 
@@ -44,7 +45,7 @@ class StackTraceErrorReportFormatter implements BiFunction<String, LogRecord, Er
    * logging or exception, and we expect for there to always either be at least a log message or an
    * exception.
    */
-  private static String createTitle(final LogRecord logRecord) {
+  static String createTitle(final LogRecord logRecord) {
     // if we have a stack trace with a triplea class in it, use that for the title;
     // EG: org.triplea.TripleaClass.method:[lineNumber] - [exception]; Caused by: [exception cause]
     return extractTripleAClassAndLine(logRecord)
--- a/game-core/src/main/java/org/triplea/debug/error/reporting/StackTraceErrorReportFormatter.java
+++ b/game-core/src/main/java/org/triplea/debug/error/reporting/StackTraceErrorReportFormatter.java
@@ -9,7 +9,7 @@ import java.io.PrintWriter;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Optional;
-import java.util.function.BiFunction;
+import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.logging.LogRecord;
 import javax.annotation.Nullable;
@@ -22,7 +22,8 @@ import org.triplea.http.client.error.report.ErrorReportRequest;
  * ErrorReport} object that we can send to the HTTP server.
  */
 @AllArgsConstructor(access = AccessLevel.PACKAGE, onConstructor_ = @VisibleForTesting)
-class StackTraceErrorReportFormatter implements BiFunction<String, LogRecord, ErrorReportRequest> {
+class StackTraceErrorReportFormatter
+    implements Function<ErrorReportRequestParams, ErrorReportRequest> {
 
   private final Supplier<String> versionSupplier;
 
@@ -31,10 +32,14 @@ class StackTraceErrorReportFormatter implements BiFunction<String, LogRecord, Er
   }
 
   @Override
-  public ErrorReportRequest apply(final String userDescription, final LogRecord logRecord) {
+  public ErrorReportRequest apply(final ErrorReportRequestParams errorReportRequestParams) {
     return ErrorReportRequest.builder()
-        .title(createTitle(logRecord))
-        .body(buildBody(userDescription, logRecord))
+        .title(createTitle(errorReportRequestParams.getLogRecord()))
+        .body(
+            buildBody(
+                errorReportRequestParams.getUserDescription(),
+                errorReportRequestParams.getMapName(),
+                errorReportRequestParams.getLogRecord()))
         .gameVersion(versionSupplier.get())
         .build();
   }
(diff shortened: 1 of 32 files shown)


DanVanAtta:


```diff

Coverage increased per file
===========================
+ http-server/src/main/java/org/triplea/modules/error/reporting/CreateIssueStrategy.java  17
+ http-clients/src/main/java/org/triplea/http/client/error/report/ErrorReportRequest.java  5
         

Coverage decreased per file
===========================
- game-core/src/main/java/games/strategy/triplea/odds/calculator/DummyPlayer.java  -8
- game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java  -1
- game-core/src/main/java/games/strategy/triplea/Properties.java  -1
         

Complexity increasing per file
==============================
- http-server/src/main/java/org/triplea/modules/error/reporting/CanUploadErrorReportStrategy.java  1
- http-server/src/test/java/org/triplea/modules/error/reporting/CanUploadErrorReportStrategyTest.java  1
- game-core/src/main/java/org/triplea/debug/error/reporting/CanNotUploadSwingView.java  1
- game-core/src/main/java/org/triplea/debug/error/reporting/UploadDecisionModule.java  2
- http-server/src/main/java/org/triplea/modules/error/reporting/ErrorReportController.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=5784874&bid=18938765)

DanVanAtta:


```diff

Complexity increasing per file
==============================
- http-server/src/main/java/org/triplea/modules/error/reporting/ErrorReportController.java  2
- http-server/src/main/java/org/triplea/modules/error/reporting/CanUploadErrorReportStrategy.java  1
- http-server/src/test/java/org/triplea/modules/error/reporting/CanUploadErrorReportStrategyTest.java  1
- game-core/src/main/java/org/triplea/debug/error/reporting/UploadDecisionModule.java  2
- game-core/src/main/java/org/triplea/debug/error/reporting/CanNotUploadSwingView.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=5784874&bid=18938765)

Cernelius:
I think it should say "game name", as that is what the players have visible the most, it is what they select and it should give a more specific insight, for maps having two or more games. For example, if a player is playing "WAW 1940", it makes the most sense that it should tell such game name, instead of delving into finding what is the map name for that game, and posting "world_at_war". On top of this, the map names have the issue that are usually defined differently (upper cases additions and underscores removals) when you are prompted to download the maps, so the game name is consistent the most.

So, I strongly advice changing "which map are you playing" with "which game are you playing" and "Map" with "Game" in the report.

DanVanAtta:
Let's bike shed the map vs game naming terminology in: https://github.com/triplea-game/triplea/issues/6994

Cernelius:
My comment was actually **not about discussing the terminology at all** (and it is not really making sense to me as its own issue, to the point that I wouldn't even know what to say in the same, and I doubt anyone else will). It was about what element to prompt the user to reference, however it is called. Obviously, I called them as I believe they are currently called.

I was just saying that I would rather ask the user to tell which game it is rather than to tell which map it is (for the reasons I've explained). My comment would be the same, besides terminology, if map and game would be called some other way (so no terminology at all).

Anyway, it's just a suggestion. I would agree in most cases telling the map or telling the game would not make a difference, and probably most users will tell either one or the other regardless of which one they've been asked for.

DanVanAtta:
Players do not know, nor should they, what the distinction is between a map and a game. To them it's essentially the same thing. For map makers, it's an important difference, though there are historical reasons that make less sense now.

I'd like to bring that conversation to a new issue as we can start to bike-shed around it. The terminology of map vs game is *not* important compared to the fact there are database changes here, the overall structure of the code, and many other really important topics that could have the game crashing if done wrong. Swapping out one word for another can be done super easily later. The topic of the terminology goes deeper, if we invest our energy here discussing that we could be missing a much larger problem while focused on a point that is more interesting but much less significant or risky ([bike-shedding](http://bikeshed.com/))

game-core/src/main/java/org/triplea/debug/error/reporting/StackTraceErrorReportFormatter.java
```diff
@@ -158,11 +170,6 @@ private static String throwableToString(final Throwable e) {
     try (PrintWriter printWriter = new PrintWriter(outputStream, false, StandardCharsets.UTF_8)) {
       e.printStackTrace(printWriter);
     }
-    return "## Exception \n"
-        + e.getClass().getName()
-        + Optional.ofNullable(e.getMessage()).map(msg -> ": " + msg).orElse("")
-        + "\n\n## Stack Trace\n```\n"
-        + outputStream.toString(StandardCharsets.UTF_8)
-        + "\n```\n\n";
+    return "## Stack Trace\n```\n" + outputStream.toString(StandardCharsets.UTF_8) + "\n```\n\n";
```
RoiEXLab, line 173: Something worth considering:
As we all know the report body is truncated at X characters. This means that the trailing `\n```\n\n` will be cut off if the stacktrace is too long.

Perhaps we want to truncate the stracktrace so we don't ever hit the other limit

http-server/src/main/java/org/triplea/db/dao/error/reporting/ErrorReportingDao.java
```diff
@@ -23,4 +26,12 @@
    */
   @SqlUpdate("delete from error_report_history where date_created < :purgeSinceDate")
   void purgeOld(@Bind("purgeSinceDate") Instant purgeSinceDate);
+
+  @SqlQuery(
+      "select created_issue_link"
+          + "  from error_report_history"
+          + "  where report_title = :reportTitle "
+          + "    and game_version = :gameVersion")
```
RoiEXLab, line 34: I wonder if a check for equality is the right move here.
If we fix an issue, but people are still using an unpatched version, they shouldn't be able to create an issue IMO.
Thoughts?

DanVanAtta: That'll be a problem if you go back in time. So we can certainly get 'n' issues up until a problem is fixed. That is a risk. If we fail to fix an issue, it's more risky if we suppress it. False-positives are not great,  false-negatives are worse.

Does that make sense?

game-core/src/main/java/org/triplea/debug/error/reporting/StackTraceErrorReportFormatter.java
DanVanAtta: If you're missing the trailing ` ``` `, you'll still get the code format.

If we truncate only the stack trace, you'll need to calculate how much the rest of the content is taking and do a subtraction. 

It then seems just easier to truncate the whole thing knowing that the stack trace is at the end.

http-server/src/main/java/org/triplea/db/dao/error/reporting/ErrorReportingDao.java
DanVanAtta: @RoiEXLab how do you mean by an unpatched version? 

Let's say a problem is reported for 'v2', and someone is using a 'v3' and we only have a fix in 'v4', we can get a duplicate report fro both 'v2' and 'v3', is that the scenario you are considering?

IMO it's okay:
- the stack traces could change between v2 and v3 along with line numbers. Having that be different is slightly helpful.
- we still go from 'n' reports being linear to the number of versions instead of number of occurrances.
- we avoid a situation where we think we fixed the problem in 'v4', but perhaps we did not. If we de-dupe on title only, then a closed issue is theoretically definitive even if the problem is re-introduced or never fixed in the first place.

In part we're being clever about detecting when problems are fixed by just de-duping by title and version. If we suppressed all future version reports we'd need to know when a problem is fixed. As-is, a person looks at the bug report, sees it's closed and hopefully we'll comment which version has a fix. 

Going forward pre-release hopefully will get back to being used for preview and testing and not early access like it was. In part the early access was due to map features and because map versioning is deficient when it comes to forward & backward compatibility. Removing the need for early access implies we've some work to improve how maps are handled.

DanVanAtta:
@RoiEXLab I'm going to merge this so I can cut a 2.1 tonight. That gives some time tomorrow to respond to any issues. I'm happy to respond to any further feedback post-merge.

http-server/src/main/java/org/triplea/db/dao/error/reporting/ErrorReportingDao.java
RoiEXLab: >  is that the scenario you are considering

Yes, that's exactly what I meant

game-core/src/main/java/org/triplea/debug/error/reporting/StackTraceErrorReportFormatter.java
RoiEXLab: True, just wanted to through the idea out there

Issue #6504, opened 2020-05-21
Revised - Useless submerge subs prompt


DanVanAtta:
### How can the problem be recreated?
- Map: WWII Revised
- set up a combat of a sub vs a transport
- fight a round of combat where both the subs and transport miss
- notice: there is a retreat or submerge prompt, click submerge
- notice: there is a retreat or remain prompt

### What is an expected fix?

Clicking 'retreat' sends the sub back to the territory from whence it came, the submerge prompt seems to have no effect and probably should not be displayed in such cases.

### (Optional) Additional information

This is an example scenario set up for this situation:
![Screenshot from 2020-05-20 19-03-40](https://user-images.githubusercontent.com/12397753/82515275-b6da7f00-9acc-11ea-921c-6f47af4294a1.png)

[test-revised-vs-solo-transport.tsvg.zip](https://github.com/triplea-game/triplea/files/4660239/test-revised-vs-solo-transport.tsvg.zip)

Save game is captured @ SHA: edafc921297149a7ace2f41a1f8b15694a9c2a25 (May 20, 2020)

DanVanAtta:
@panther2 would you mind having a look at this scenario, could you confirm that indeed a submerge prompt is incorrect and there should only be a retreat prompt?

panther2:
@DanVanAtta I have recreated the battle using a newer pre-release - and I get these windows after the dice have been rolled for both sides for the first time:
![1_Retreat_by_submerge](https://user-images.githubusercontent.com/20812775/82579895-bb676c00-9b8e-11ea-8407-020fe88418a3.PNG)
and klicking "Remain" I get:
![2_Retreat_by Retreat](https://user-images.githubusercontent.com/20812775/82579950-cfab6900-9b8e-11ea-9731-5d28b64bd8b2.PNG)

DanVanAtta:
Does clicking 'remain' actually lead the sub to submerging?
I honestly had not considered the game engine would do that, 'Remain' typically means "remain in combat" in every other context. It seems that the "remain" option would actually be better worded as the proper submerge option.

It sounds like we probably should have one dialog with four options:
- Submerge
- Retreat
- Remain
- Ask me Later

In this way it is clear submerge ends the battle and remain retains the meaning it does in other battle prompts (stays consistent). WDYT @panther2?

Cernelius:
> It sounds like we probably should have one dialog with four options:
> 
> Submerge
> Retreat
> Remain
> Ask me Later

[...]

Issue #4821, opened 2019-04-28
Submarine submerge and retreat issues


panther2:
### Engine version
Tested with 1.10.14614

I am very short of time at the moment so this is more or less an announcement of issues that have to be deeper investigated for all versions. While testing https://github.com/triplea-game/triplea/pull/4820 and taking a quick glance on some other versions I noticed this:

First issue:
When the attacker chooses to submerge TripleA alway submerges **all** attacking submarines.
Instead the player should be able to choose how many of his subs should submerge.
Also the defender should be able to select how many of his subs should submerge,

Second issue:
When the attacker chooses to retreat submerged subs are retreated together with the surface ships. Instead the submerged ships should remain in the contested seazone.

I am sorry for this "unprofessional" report, but as I said, I don't have much time. So any help in investigating and testing is highly welcome.

CC: @ron-murhammer

ron-murhammer:
@panther2 Interesting. I didn't realize you could choose to submerge only specific subs. I figured it was like retreating units which is all or nothing. Are there certain scenarios you can think of that you'd want to submerge only some subs?

panther2:
@ron-murhammer Yes there are, Sometimes you want to leave (one or more) submerged submarines in the contested seazone. When they resurface the opponents have to deal with them in any way. 
At the same time you maybe want to retreat your (non submerged) subs together with the fleet to an adjacent space.

Older rulesets state the decision on a single submarine basis. Newer rulesets explicitly state "submerge any number of subs". No ruleset says "all subs". 
Also submerged subs should not been moved with the retreating fleet.

When I have some more time, I will check the engine map version by map version (ruleset by ruleset) concerning those issues. I just can't do it these days.

trevan:
I have a fix for the second issue in #7007

PR #7007, opened 2020-07-02
Don't retreat submerged units

trevan:
This fixes the second issue listed in #4821 as well as the issue in #6504 where it asks you to retreat if you've already submerged all of your subs and you don't have any other units.

It doesn't change the dialogs that pop up.  I was just trying to fix the second issue in #4821 where the sub was retreated with the rest of the units even though it has been submerged. During testing, I noticed that it partially "fixed" #6504 because the engine doesn't see any units to retreat, so it doesn't ask to retreat.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix
[] Other:   

## Testing
Played with sub only battles and sub with other units.  I submerged and retreated the units to see where things ended up.

## Release Note

FIX|Submerged subs don't retreat when other units retreat


DanVanAtta:
:+1:

PR #7073, opened 2020-07-08
Make warnings non-reportable

DanVanAtta:
Add a check for logging level in our global logging handler. If the log
level is warning then disable the 'upload button'. Same, if the level
is a warning also add a generic statement giving the user a link to
report the problem.

This update allows us to notify users of general errors that have happened
that are outside of TripleA's control and are something that we can
tell the user how to fix in the warning message. This way we can
use a 'warning' logging to report this to the user without
constructing a full swing dialog with the same text.

Events logged at a 'severe' level will continue to have an 'upload' button
enabled and there is no change to the 'severe' logging behavior.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix
[] Other:   

## Testing
Injected a logging statement into GameRunner main method to verify dialog behavior for warning and severe levels.

## Screens Shots

**Before**
Screenshot from 2020-07-07 21-02-46

**After**
Screenshot from 2020-07-07 20-59-32

Of note, this fix is not for us to blindly convert severe logging into warning to make it non-reportable. We should be careful that any warning logging actually tells a user how to fix the problem (eg: "unavailable, try again", "you are a probably already hosting a game on this port, please choose another port")

### DIFF
--- a/game-core/src/main/java/org/triplea/debug/ErrorMessage.java
+++ b/game-core/src/main/java/org/triplea/debug/ErrorMessage.java
@@ -7,17 +7,18 @@ import java.awt.event.WindowAdapter;
 import java.awt.event.WindowEvent;
 import java.net.URI;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.logging.Level;
 import java.util.logging.LogManager;
 import java.util.logging.LogRecord;
 import javax.swing.JButton;
 import javax.swing.JFrame;
-import javax.swing.JLabel;
 import javax.swing.SwingUtilities;
+import javax.swing.border.EmptyBorder;
 import org.triplea.debug.error.reporting.UploadDecisionModule;
 import org.triplea.http.client.error.report.ErrorReportClient;
 import org.triplea.live.servers.LiveServersFetcher;
 import org.triplea.swing.JButtonBuilder;
-import org.triplea.swing.JLabelBuilder;
+import org.triplea.swing.JEditorPaneWithClickableLinks;
 import org.triplea.swing.jpanel.JPanelBuilder;
 
 /**
@@ -38,7 +39,7 @@ public enum ErrorMessage {
 
   private static final String DEFAULT_LOGGER = "";
   private final JFrame windowReference = new JFrame("TripleA Error");
-  private final JLabel errorMessage = JLabelBuilder.builder().errorIcon().iconTextGap(10).build();
+  private final JEditorPaneWithClickableLinks errorMessage = new JEditorPaneWithClickableLinks("");
   private final AtomicBoolean isVisible = new AtomicBoolean(false);
   private volatile boolean enableErrorPopup = false;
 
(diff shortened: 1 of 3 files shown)


game-core/src/main/java/org/triplea/debug/ErrorMessage.java
```diff
@@ -130,7 +132,12 @@ private void setUploadRecord(final LogRecord logRecord) {
   }
 
   private void activateUploadErrorReportButton(final URI lobbyUri, final LogRecord logRecord) {
-    INSTANCE.uploadButton.setVisible(true);
+    // Set upload button to be visible only if the logging is greater than a 'warning'.
+    // Warning logging should tell a user how to correct the error and it should be something
+    // that is 'normal' (but maybe not happy, like no internet) and something they can fix
+    // (or perhaps something that we simply can never fix). Hence, for warning we want to inform
+    // the user that the problem happened, how to fix it, and let them create a bug report manually.
+    INSTANCE.uploadButton.setVisible(logRecord.getLevel().intValue() > Level.WARNING.intValue());
```
RoiEXLab, line 140: I actually think this problem is too complex to represent it in a single severity level.
Let's say a map maker is trying to build a new map and wants to test it.
Turns out they got the format wrong and therefore the engine is unable to parse the map. I would still consider this a severe error, not a warning because it is an exception that prevents an entire action from happening, but it shouldn't be reportable.
However let's say $User has their maps saved as a folder somewhere and the write and read permissions are messed up. In this case the error is still severe and it's a clear case for user help, but the system can't communicate that at all.

I must admit that I don't have a better idea than this, it's probably the best short-term approach we can go with, but we might want to change the system to do some heuristics if the error is user related (local IOException, FileNotFound etc.) or it has a real code implication (NullPointerException etc.) in the long run.
Of course there are edgecases as well: Is a StackOverflowError a regression or a low-memory user issue?
But I think it's better than this approach

DanVanAtta:
Given it's a thick client, the 'log' is not really useful. It's all about
the user perspective and either self recovery in face of errors or a game
crash.

For example, many players will restart their game if they get any message
at all,and things like Ioexception are meaningless to them.

I suppose bottom line,  logging in a thick client is not actually useful.
Users will not see the severity level, info logging has no real purpose,
and any raw error message is meaningless to a user

game-core/src/main/java/org/triplea/debug/ErrorMessage.java
DanVanAtta: @RoiEXLab see my response to the PR, but secondarily I'd suggest warning is still okay. The game did not crash. IMO `Severe` is the game is left in a dead state, warning is basically an action is not going to be available. INFO probably is not really worth using (pointless). I really can't emphasize enough how the logging perspective for a thick client is different compared to say a web server.

PR #7088, opened 2020-07-10
Headless: Ignore duplicate game entries

DanVanAtta:
If we see a duplicate game entry, instead of a bot server crashing,
this will update will log the duplicate game entry and ignore it.

Mitigates problem: https://github.com/triplea-game/triplea/issues/7005

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix
[] Other:   

## Testing
- Downloaded both d-day maps, launched a headless bot locally and confirmed could repro the error.

## Release Note

Instead of creating a `Map` directly from a stream, we instantiate the `Map` first and then do a `forEach` to add games. If we see any duplicates then we log them and do not add them to the map.

### DIFF
--- a/game-core/src/main/java/org/triplea/game/server/AvailableGames.java
+++ b/game-core/src/main/java/org/triplea/game/server/AvailableGames.java
@@ -16,7 +16,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.logging.Level;
-import java.util.stream.Collectors;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 import javax.annotation.Nonnull;
@@ -36,13 +35,23 @@ final class AvailableGames {
   private final Map<String, URI> availableGames;
 
   AvailableGames() {
-    availableGames =
-        FileUtils.listFiles(ClientFileSystemHelper.getUserMapsFolder())
-            .parallelStream()
-            .map(AvailableGames::getGames)
-            .map(Map::entrySet)
-            .flatMap(Collection::stream)
-            .collect(Collectors.toUnmodifiableMap(Map.Entry::getKey, Map.Entry::getValue));
+    availableGames = new HashMap<>();
+    FileUtils.listFiles(ClientFileSystemHelper.getUserMapsFolder())
+        .parallelStream()
+        .map(AvailableGames::getGames)
+        .map(Map::entrySet)
+        .flatMap(Collection::stream)
+        .forEach(
+            gameEntry -> {
+              if (!availableGames.containsKey(gameEntry.getKey())) {
+                availableGames.put(gameEntry.getKey(), gameEntry.getValue());
+              } else {
+                log.warning(
+                    String.format(
+                        "DUPLICATE GAME ENTRY! Ignoring game entry: %s, " + "existing value is: %s",
+                        gameEntry, availableGames.get(gameEntry.getKey())));
+              }
+            });
     log.info(
         String.format(
             "Done loading maps, " + "availableGames count: %s, contents: %s",


tvleavitt:
I'd categorize this as a workaround, not a fix, because the duplicate maps could have different content. Discussing more in #7089

game-core/src/main/java/org/triplea/game/server/AvailableGames.java
```diff
@@ -36,13 +35,23 @@
...
+            gameEntry -> {
+              if (!availableGames.containsKey(gameEntry.getKey())) {
+                availableGames.put(gameEntry.getKey(), gameEntry.getValue());
+              } else {
+                log.warning(
+                    String.format(
+                        "DUPLICATE GAME ENTRY! Ignoring game entry: %s, " + "existing value is: %s",
+                        gameEntry, availableGames.get(gameEntry.getKey())));
+              }
+            });
```
RoiEXLab, line 54: While I don't think I can make a qualified decision if this kind of duplicate handling is the best approach to this, I want to point out that there's an overload of the `Collectors#toUnmodifianleMap` method that accepts a merge function:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/Collectors.html#toUnmodifiableMap(java.util.function.Function,java.util.function.Function,java.util.function.BinaryOperator)
The default implementation seems to be this one:

https://github.com/openjdk/jdk/blob/0a38584c10c1b0d971271f47934e4003bde7d986/src/java.base/share/classes/java/util/stream/Collectors.java#L148-L159

Which throws exactly the exception we are experiencing.
By providing this custom function we can keep the functional style + we maintain the unmodifiable map, which is no longer defensively copied or wrapped somehow for thread safety (the reason why unmodifiability would be an important thing to keep)

DanVanAtta: The merge function looks really attractive. I just gave it a try, but it looks like we won't have access to the 'key' in conflict.

The merge function essentially allows `(oldValue, newValue) -> deconflictedValue`. While we can figure out the conflicting game name based on the URI alone, it certainly is more direct if we can log the game key that has the duplicate value.

DanVanAtta:
@tvleavitt agree this is not a 'fix', the terminology I used there is very loose. It's only a 'fix' in that the bot server won't crash on startup, "mitigation" is more accurate of a term.

game-core/src/main/java/org/triplea/game/server/AvailableGames.java
RoiEXLab: Ah, that's something I didn't think of. Unfortunate.

Even though this PR is already merged I have some concerns.
This is a parallel Stream, so we could end up running into concurrency issues because the now unsynchronized map doesn't have a wrapper for this kind of protection.
Also somewhere down the line we give this exact map away without any sort of write protection, which might introduce weird bugs if we modify it somewhere by accident.

I was able to come up with 4 potential solutions to this:
1. Just wrap it in a Collections.synchronizedMap package and return an unmodifiable collection wrapper int the API (I forgot where this was, but I removed it in my PR), the easiest solution but might hurt parallel performance and adds unecessary abstraction.
2. Do it in 2 steps. We could use the streams and everything to put all of the entry objects in an unmodifiable List and use a normal for-each loop to fill a map, and all we store is the unmodifiable wrapper to this map. Easy solution, but still not optimal use of streams IMO.
3. Same as 2. but instead of using for-each we use the same collector as before and iterate over all elements until we find the correct key in case we find a conflicting value. Still weird, but works as well.
4. Provide a custom collector. Instead of using the default `Collectors#toUnmodifiableMap` we could provide our own "implementation". The `Collector` interface offers some handy static functions for this.
This would add some complexity:
```java
Collector.of(
  HashMap::new,
  (map, entry) -> {
    // Alternative approach would be to use Map#containsKey
    if (map.put(entry.getKey(), entry.getValue()) != null) {
      log.log(Level.INFO, "Duplicate key: " + entry.getKey());
    }
  },
  (map1, map2) -> {
    for (final var entry : m2.entrySet()) {
      // Variant two, as described earlier
      if (map1.containsKey(entry.getKey())) {
        log.log(Level.INFO, "Duplicate key: " + entry.getKey());
      }
      map1.put(entry.getKey(), entry.getValue())
    }
    return m1;
  },
  Collections::unmodifiableMap,
  // Optional but for the sake of completeness
  Collector.Characteristics.UNORDERED);
```

The only thing I don't like about approach 4 is that it duplicates the detection logic, but this could be solved by extracting it to a private static helper method, so it would look like this:
```java
Collector.of(
  HashMap::new,
  (map, entry) -> putAndWarn(map, entry.getKey(), entry.getValue()),
  (map1, map2) -> {
    for (final var entry : map2.entrySet()) {
      putAndWarn(map1, entry.getKey(), entry.getValue());
    }
    return map1;
  },
  Collections::unmodifiableMap,
  // Optional but for the sake of completeness
  Collector.Characteristics.UNORDERED);
```

What do you think? I leaning towards option 4 simply because I think it's the most consistent way to do it.

tvleavitt:
Why go through all this when simply including including the leading path or
source zip guarantees uniqueness, you can't have two files with the same
name in the same directory? This would eliminate the need to do any
comparisons at all.

Thomas Leavitt

DanVanAtta:
@RoiEXLab we probably could/should remove the parallel stream. The bots start up much faster now that we do shallow parsing, that was a *much* bigger win compared to parallel processing. Parallel processing also puts more memory pressure on the bots as well as the XMLs are loaded more than one at a time. Furthermore, we only have 1 CPU on the bots.

@tvleavitt The mapping in question is a mapping of game name to the full path in the map zip or map folder. This is the code for how a bot finds the full path of an XML file given the game-name.

Fundamentally there was/is an assumption that all game-names would have a unique name to them. Without that holding we have other issues. We could list duplicate game-names in the map selection for bots and it would then be unclear to players which one they are selecting. The headed-client essentially does this, but at least you get the map description before you select the game.

RoiEXLab:
@DanVanAtta I think we should definitely do some benchmarking before actually removing the parallel part.
It has some overhead, but depending on where the threshold is this might still make a big difference.

DanVanAtta:
@RoiEXLab benchmarking is not going to be that practical because you have to do it on a linode. Furthermore, we got more gains from going to shallow parse than we did from parallel (even with multiple cores). If we get an easier time parsing maps, waiting an extra 10-30 seconds or so for bots to boot up is okay. They used to take minutes when we were doing deep parse, it's under a minute now.

DanVanAtta:
For comparison, with 8 cores and a dozen maps:

Deep parse, stream: 4400ms
Deep parse, parallel: 3100ms

Shallow parse, stream: 2400ms
Shallow parse, parallel: 1400ms

With a single core, the difference is likely to be zero (or parallel could even be worse given the threading overhead). We used to do 'deep parsing' for quite some time. Hence, we'll be in good enough shape by going to non-parallel.

DanVanAtta:
Crud.. accidently pushed to master :disappointed: : 

@RoiEXLab I made the processing here single threaded: https://github.com/triplea-game/triplea/commit/2caa37811b62a044d8f60b9372e74b55975f9729

RoiEXLab:
I already commented on the commit, but just wanted to repeat it again: https://github.com/triplea-game/triplea/commit/2caa37811b62a044d8f60b9372e74b55975f9729#r40534355

Basically you pushed the non-shallow version of the code.

RoiEXLab:
Also I think I recall the reason we did in-depth parsing for a long time (take it with a grain of salt though, might be false memories):

I think there's this potential, slight chance that there's an issue with deep parsing that doesn't occur in shallow parsing.

From the javadoc of `AvailableGames`:
> We make sure we can parse them all, but we don't keep them in memory.

PR #7101, opened 2020-07-11
Save game compatibility testing

DanVanAtta:
Adds a test that will download a sample of various save games and verify that they can be loaded. This test is to help ensure that we maintain save-game compatibility. Once we migrate to 3.x and beyond, we will have to empty the list and/or add new save games for 3.x
 
Important note, **this testing is not full proof**. For example we can still rename fields in `GameData` or any object serialized as part of `GameData` and it will be initialized to null when loaded. In such a case we'd only see a NPE when any corresponding code is exercised and not at time of game-data load.

## Commits

commit 87fc00d80c725692f1040513f3ef4beaaf9f66ba

    Extract test file listing utility to its own class

commit f579419632d8a5b8d119de7f884c873d8a9bbb3e

    Improve file listing test utility robustness, print folder name if not found and improve variable naming

commit dfaf707a54ba2ebab3a79f5e8c3df2d00592e210

    Add a test that will load save-games from test/resources/save-games and tries to load them.
    
    This test serves as a compatibility check.

commit 3b54c08e84d351d80a72fc4609f9dae47aac2a62

    Add task to download save-game files from assets & add list of save-game files to be downloaded

commit 527d42c2cce9e06b44d95078e3a21c37371eed47

    Improve compatibility test, assert various gameData properties are not null after load
    
    If we change the name of an item in GameData, it will not cause a loading exception but we will get an unexpected null value. This update makes the compatibility check more robust (but still not fool-proof)

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix
[x] Other: test infrastructure  

## Testing

Verified the following scenarios failed the test:
- loaded a 1.9 save game
- changed  'GameData' SerialVersionUUID 
- changed a field name in Version.java
- changed a field name in `GameData`

### DIFF
--- a/smoke-testing/src/test/java/games/strategy/engine/data/TestDataFileLister.java
+++ b/smoke-testing/src/test/java/games/strategy/engine/data/TestDataFileLister.java
@@ -7,6 +7,7 @@ import java.util.Collection;
 import java.util.Comparator;
 import java.util.stream.Collectors;
 import lombok.experimental.UtilityClass;
+import org.triplea.java.Postconditions;
 
 @UtilityClass
 class TestDataFileLister {
@@ -24,11 +25,11 @@ class TestDataFileLister {
   }
 
   private static File findFilesInFolder(final String folderName) {
-    final File mapXmls = Paths.get("src", "test", "resources", folderName).toFile();
-    if (mapXmls.exists()) {
-      return mapXmls;
+    File resourcesFolder = Paths.get("src", "test", "resources", folderName).toFile();
+    if (!resourcesFolder.exists()) {
+      resourcesFolder = Paths.get("smoke-testing", "src", "test", "resources", folderName).toFile();
     }
-
-    return Paths.get("smoke-testing", "src", "test", "resources", folderName).toFile();
+    Postconditions.assertState(resourcesFolder.exists(), "Could not find folder: " + folderName);
+    return resourcesFolder;
   }
 }
--- a/smoke-testing/src/test/java/games/strategy/engine/data/TestDataFileLister.java
+++ b/smoke-testing/src/test/java/games/strategy/engine/data/TestDataFileLister.java
@@ -1,0 +1,34 @@
+package games.strategy.engine.data;
+
+import java.io.File;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.stream.Collectors;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+class TestDataFileLister {
+
+  /**
+   * Returns a list of files located in folder like: 'test/resources/{folder-name}'.
+   *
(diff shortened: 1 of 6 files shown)


DanVanAtta:


```diff

Complexity increasing per file
==============================
- smoke-testing/src/test/java/games/strategy/engine/data/GameSaveCompatibilityTest.java  1
- smoke-testing/src/test/java/games/strategy/engine/data/TestDataFileLister.java  2
         

Complexity decreasing per file
==============================
+ smoke-testing/src/test/java/games/strategy/engine/data/ParseGameXmlsTest.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=5837150&bid=19113165)

smoke-testing/src/test/java/games/strategy/engine/data/TestDataFileLister.java
```diff
@@ -0,0 +1,35 @@
...
+   * Returns a list of files located in folder like: 'test/resources/{folder-name}'.
+   *
+   * @param folderName Name of the folder in 'test/resources'
+   * @return List of all files located in specified folder.
+   */
+  Collection<File> listFilesInTestResourcesDirectory(final String folderName) {
+    return Arrays.stream(findFilesInFolder(folderName).listFiles())
+        .sorted(Comparator.comparing(File::getName))
+        .collect(Collectors.toList());
+  }
```
RoiEXLab, line 25: Might be overkill, but you might want to consider creating a custom [ArgumentsProvider](https://junit.org/junit5/docs/current/api/org.junit.jupiter.params/org/junit/jupiter/params/provider/ArgumentsProvider.html).
If you combine this with a meta annotation like this:
```java
@Target({ANNOTATION_TYPE,METHOD})
@Retention(RUNTIME)
@ArgumentsSource(CustomArgumentSourceThatFindsDirectories.class)
public @interface DirectorySource {
  String folderName;
}
```

The `ArgumentsProvider` could then read out this annotation and provide the folders based on the name.
In my opinion a nice use of the API JUnit provides, but as mentioned already potentially overkill for just 2 methods.
Would avoid the repeated method source code though

DanVanAtta: Interesting idea. I was not aware of the custom data source.
IMO we'll probably want around 3 to 5 examples before that would really makes sense.

RoiEXLab: Yeah, at this point it's probably too much boilerplate to be worth it.
The JUnit API is really well designed in my opinion. It's a little bit heavy on reflection use in my opinion, but a lot of the code is designed around annotation inheritance, which makes it very extensible and customizable.
But there's definitely a hurdle to get the hang of it by observing how the junit implements csvsource for example

PR #7108, opened 2020-07-11
Remove HeadlessUiContext.java, not used

DanVanAtta:
We instantiate the object, call setters on it, but then do
nothing further with the local object which then goes out of
scope. The object is a no-op, the setter calls do not have
any needed side-effects, hence we can just delete the class.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix
[] Other:   

## Testing
- checked could launch a map on a local headless instance


DanVanAtta:
> Not sure if this should be part of this PR or a follow-up PR.

We certainly go down a rabbit hole when we do that. This PR is 'complete' for its intent, generally smaller will be better for faster review iterations.

DanVanAtta:
Notably, the HeadlessLauncher does not use any args, it is also a no-op implementation itself. I think that is something we should look at it, more code that could be dropped. 

I'd agree UIContext probably does not need an interface at this point (the interface is certainly a bit of a hodge). I suspect the genesis of the interface is Veqryn scrambling at 1am to convert all headed classes that would have to be modified to interfaces to then drop a headless implementation in its place. The split of headed and headless was certainly done on-the-rush (we were in a bit of a pinch at the time, so understandable)

PR #7118, opened 2020-07-11
Restore parseShallow, partial revert of: 2caa37811b62a04

DanVanAtta:
## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix
[] Other:


DanVanAtta:
side-note, to disallow direct pushes we need to require one approval. I suppose that is the reason we have not turned on non-PR pushes.

Issue #6475, opened 2020-05-15
Error (why must a player have this image in his unit image directory?)


FrostionAAA:
This is such a strange engine requirement that I can’t think it anything else than a flaw in the engine. Look at this situation, where a player nation kills enemies of another player:
![Unavngivet](https://user-images.githubusercontent.com/15268177/82071294-34ba1700-96d6-11ea-99cd-3f0f57dc0a08.png)

The facts in regards to this situation’s map files:
- The _Trolls_ player can in no way build the Kobold unit, so it does not have that image in its _Trolls_ unit image directory.
- The Kobold unit image is only located in the _units_ root dir.
- The Loot bag unit image is only located in the _units_ root dir.

So the unit is killed and the engine looks for the Kobolds/Lootbag images it in the Trolls dir?

```
<attachment name="unitAttachment" attachTo="Kobold-Miners" javaClass="games.strategy.triplea.attachments.UnitAttachment" type="unitType">
<option name="defense" value="2"/>
<option name="whenHitPointsDamagedChangesInto" value="1:false:Loot-Level-1"/>
</attachment>

<attachment name="unitAttachment" attachTo="Loot-Level-1" javaClass="games.strategy.triplea.attachments.UnitAttachment" type="unitType">
<option name="isInfrastructure" value="true"/>
<option name="createsResourcesList" value="10:XP"/>
<option name="createsResourcesList" value="1:Gold"/>
</attachment>
```

I guess the engine image display has a problem with this kind of unit kill -> change -> capture?

By the way, there is no other consequences of the “missing unit image”, other than no image is displayed in history-mode  when reviewing the battle.
![Unavngivet2](https://user-images.githubusercontent.com/15268177/82072266-b9596500-96d7-11ea-9d4a-eddd09c38839.png)

Can this issue be addressed?

Johnny1900:
I noticed the same error while playing Domination 1914.

Cernelius:
I don't think the "Trolls" player has anything to do with this. The error message is searching the unit for the "Kobolds" player.

FrostionAAA:
@Cernelius The history screenshot is of a Troll player flag (Green) and 3 "Kobold-Miners" units that has no unit image in Troll unit directory and therefore is missing.

The generic flag unit image of the Kobold player is the cross with cranium. This is only show as the flag image before the territory is captured.

Cernelius:
[...]

PR #7158, opened 2020-07-13
If a unit is dead, don't categorize it

### LINKED ISSUE #7117: 2.1.20473: UnitsDrawer#draw - MISSING UNIT IMAGE (won't be displayed): Wraith (Wraith_hit) owned by Ghosts in The Coil
## User Description
This happened while playing my warcraft map. Strangly the engine askes for a _hit image of a unit that does only have 1 HP? This is strange //FrostionAAA

## Log Message
MISSING UNIT IMAGE (won't be displayed): Wraith (Wraith_hit) owned by Ghosts in The Coil

## TripleA Version
2.1.20473

## Java Version
11.0.6

## Operating System
Windows 10

trevan:
Fixes #7117

This changes the UnitSeparator code to ignore dead units.  I don't see a reason for why UnitSeparator should worry about dead units since they are generally not in the territory.

The problem that it is fixing is that during a battle, when an attacker dies and a defender dies, if the defender will change into a new unit, a change request is made to set the hit points to 0 instead of just removing the unit.  I believe this change request is needed to trigger the unit to change into the other unit.

Because the change request sets the hit points to 0, the map will sort the units and include these dead units.  If a redraw is triggered (which I think always happens if there is a dead attacker unit, but I'm not sure), then it will try and draw these dead units.

By excluding the dead units during the categorizing/sorting, the map will not try and draw them.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix
[] Other:   

## Testing

I tested the save from the bug with a breakpoint at the `continue` line.  I verified that the line was reached and that the error didn't occur afterwards.

I also ran a all AI match in Warcraft War Heroes with a breakpoint on the `continue` line to see if things acted weirdly after it was triggered.  I didn't see anything obviously wrong.

## Release Note

FIX|Missing unit image in Warcraft War Heroes for dead units shouldn't happen anymore

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/util/UnitSeparator.java
+++ b/game-core/src/main/java/games/strategy/triplea/util/UnitSeparator.java
@@ -101,6 +101,10 @@ public class UnitSeparator {
       categories = new LinkedHashMap<>();
     }
     for (final Unit current : units) {
+      if (current.getUnitAttachment().getHitPoints() - current.getHits() == 0) {
+        continue;
+      }
+
       BigDecimal unitMovement = new BigDecimal(-1);
       if (categorizeMovement
           || (categorizeTrnMovement && Matches.unitIsTransport().test(current))) {
--- a/game-core/src/test/java/games/strategy/triplea/util/UnitSeparatorTest.java
+++ b/game-core/src/test/java/games/strategy/triplea/util/UnitSeparatorTest.java
@@ -76,6 +76,26 @@ class UnitSeparatorTest {
     assertEquals(expected, categories);
   }
 
+  @Test
+  void ignoreKilledUnitsInSorting() {
+    final GameData data = TestMapGameData.TWW.getGameData();
+    final Territory northernGermany = territory("Northern Germany", data);
+    northernGermany.getUnitCollection().clear();
+    final GamePlayer italians = GameDataTestUtil.italy(data);
+    final List<Unit> units =
+        new ArrayList<>(GameDataTestUtil.italianInfantry(data).create(1, italians));
+    units.get(0).setHits(1);
+    final GamePlayer british = GameDataTestUtil.britain(data);
+    units.addAll(new ArrayList<>(GameDataTestUtil.britishInfantry(data).create(1, british)));
+    GameDataTestUtil.addTo(northernGermany, units);
+    when(mockMapData.shouldDrawUnit(ArgumentMatchers.anyString())).thenReturn(true);
+    final List<UnitCategory> categories =
+        UnitSeparator.getSortedUnitCategories(northernGermany, mockMapData);
+    final List<UnitCategory> expected = new ArrayList<>();
+    expected.add(newUnitCategory("britishInfantry", british, data));
+    assertEquals(expected, categories);
+  }
+
   private static UnitCategory newUnitCategory(
       final String unitName, final GamePlayer player, final GameData data) {
     return new UnitCategory(new UnitType(unitName, data), player);


Cernelius:
Wouldn't this fix
https://github.com/triplea-game/triplea/issues/6475
too?

DanVanAtta:


```diff

Complexity increasing per file
==============================
- game-core/src/main/java/games/strategy/triplea/util/UnitSeparator.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=5844055&bid=19134434)

trevan:
> Wouldn't this fix
> #6475
> too?

@Cernelius , it might but that issue was already marked as closed.

Cernelius:
> > Wouldn't this fix
> > #6475
> > too?
> 
> @Cernelius , it might but that issue was already marked as closed.

As I've written there, wrongly so, in my opinion.

trevan:
Another possible fix is to re-order the actions.  Right now, the order is:
1. Set damage on units that will change
2. Remove dead units

If that is swapped to:
1. Remove dead units
2. Set damage on units that will change

Then the unit gets removed from the territory and it won't be drawn.  But I don't know how to verify that it doesn't break something else.  I did a small test and nothing appeared to be broken but I don't know to be confident that nothing is broken.

Cernelius:
> Another possible fix is to re-order the actions. Right now, the order is:
> 
> 1. Set damage on units that will change
> 2. Remove dead units
> 
> If that is swapped to:
> 
> 1. Remove dead units
> 2. Set damage on units that will change
> 
> Then the unit gets removed from the territory and it won't be drawn. But I don't know how to verify that it doesn't break something else. I did a small test and nothing appeared to be broken but I don't know to be confident that nothing is broken.

Would this still allow a fully damaged (reduced to 0 hitpoints left) unit to change?

trevan:
> > Another possible fix is to re-order the actions. Right now, the order is:
> > 
> > 1. Set damage on units that will change
> > 2. Remove dead units
> > 
> > If that is swapped to:
> > 
> > 1. Remove dead units
> > 2. Set damage on units that will change
> > 
> > Then the unit gets removed from the territory and it won't be drawn. But I don't know how to verify that it doesn't break something else. I did a small test and nothing appeared to be broken but I don't know to be confident that nothing is broken.
> 
> Would this still allow a fully damaged (reduced to 0 hitpoints left) unit to change?

It should.  In the test I did, the Wraith still changed.

Cernelius:
According to my tests, on an unreleased map, at least for 2.0, if you would have a 1 hitpoint "wall" unit that changes into a "wall_fallen" when damaged/killed/removed (which is the same situation, as the unit has only 1 hitpoint), you need to provide all the following images, otherwise the program will trow an error for missing images:

`wall`
`wall_hit`
`wall_fallen`
`wall_fallen_hit`

These are the codes for the unit attachments:
```
                <attachment name="unitAttachment" attachTo="wall" javaClass="games.strategy.triplea.attachments.UnitAttachment" type="unitType">
                         <option name="movement" value="0"/>
                         <option name="canNotMoveDuringCombatMove" value="true"/>
                         <option name="hitPoints" value="1"/>
                         <option name="whenHitPointsDamagedChangesInto" value="1:true:wall_fallen"/>
                         <option name="requiresUnits" value="metropolis"/>
                </attachment>
				
                <attachment name="unitAttachment" attachTo="wall_fallen" javaClass="games.strategy.triplea.attachments.UnitAttachment" type="unitType">
                         <option name="movement" value="0"/>
                         <option name="canNotMoveDuringCombatMove" value="true"/>
                         <option name="destroyedWhenCapturedBy" value="barbarians"/>
                         <option name="hitPoints" value="2"/>
                         <option name="whenHitPointsRepairedChangesInto" value="0:true:wall"/>
                         <option name="requiresUnits" value="metropolis"/>
                </attachment>
```
As I said, I find this unnecessary, and suggest the engine only requiring "wall" and "wall_fallen_hit" (not "wall_hit" and "wall_fallen"). The reason is that the "wall_hit" image should really never display, as the "wall" has only 1 hitpoint, while the "wall_fallen" image should really never display, as when the "wall_fallen" is repaired it turns into the "wall". So the "wall" actually never truly exists in a damaged state, while the "wall_fallen" never truly exists an an undamaged state, yet you need images for both.

I intended to discuss this matter with @ron-murhammer (who created this feature), but I didn't get a response.

game-core/src/main/java/games/strategy/triplea/util/UnitSeparator.java
```diff
@@ -101,6 +101,10 @@ private UnitSeparator() {}
       categories = new LinkedHashMap<>();
     }
     for (final Unit current : units) {
+      if (current.getUnitAttachment().getHitPoints() - current.getHits() == 0) {
```
RoiEXLab, line 104:
```suggestion
      if (current.getUnitAttachment().getHitPoints() == current.getHits()) {
```

trevan:
I'm closing this.  I think the re-ordering of the operations would be a better fix.

PR #7217, opened 2020-07-21
Change dead units after they've been removed from the battle site

### LINKED ISSUE #7117: 2.1.20473: UnitsDrawer#draw - MISSING UNIT IMAGE (won't be displayed): Wraith (Wraith_hit) owned by Ghosts in The Coil

trevan:
Fixes #7117

This moves the code that changes dead units after the code that removes the dead units.

The bug is that the map is redrawing with the units before they've been changed.  By clearing out the dead units first, the map won't redraw them.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[X] Problem fix
[] Other:   

## Testing

I played the save game in the referenced issue.  I verified that the history of the game still has the same order.  I also used break points to make sure things were changing as expected.

## Release Note

FIX|Missing unit error message for Warcraft War Heroes when a unit dies and is turned into XP should no longer occur.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
@@ -568,6 +568,15 @@ public class MustFightBattle extends DependentBattle
     }
     final Collection<Unit> killed = getUnitsWithDependents(killedUnits);
 
+    // Remove units
+    final Change killedChange = ChangeFactory.removeUnits(battleSite, killed);
+    this.killed.addAll(killed);
+    killedDuringCurrentRound.addAll(killed);
+    final String transcriptText =
+        MyFormatter.unitsToText(killed) + " lost in " + battleSite.getName();
+    bridge.getHistoryWriter().addChildToEvent(transcriptText, new ArrayList<>(killed));
+    bridge.addChange(killedChange);
+
     // Set max damage for any units that will change into another unit
     final IntegerMap<Unit> lethallyDamagedMap = new IntegerMap<>();
     for (final Unit unit :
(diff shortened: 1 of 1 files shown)


Cernelius:
So, let's see if I'm understanding the change correctly.

---

Before this change, if I had a unit "a" that has 1 hitpoint and changes into a unit "b" when damaged, I needed the following image names:
`a`
`a_hit`
`b`

After this change, I won't need to have the `a_hit` image.

---

Before this change, if I had a unit "a" that has 2 hitpoints and changes into a unit "b" when damaged, but not transferring the damage, I needed the following image names:
`a`
`a_hit`
`b`

After this change, will I need to have the `a_hit` image or not?

---

Before this change, if I had a unit "a" that has 2 hitpoints and changes into a unit "b" when damaged, and also transferring the damage, I needed the following image names:
`a`
`a_hit`
`b_hit`
`b`

After this change, will I need to have the `a_hit` image or not?
If the unit "b" can never be repaired (thus never really exist in its undamaged state), will I need the `b` image anyway or not?

---

What are the plans about documenting the matter for mapmakers to know which images they need to provide, in every case?

trevan:
> So, let's see if I'm understanding the change correctly.
> 
> Before this change, if I had a unit "a" that has 1 hitpoint and changes into a unit "b" when damaged, I needed the following image names:
> `a`
> `a_hit`
> `b`
> 
> After this change, I won't need to have the `a_hit` image.
> 
> Before this change, if I had a unit "a" that has 2 hitpoints and changes into a unit "b" when damaged, but not transferring the damage, I needed the following image names:
> `a`
> `a_hit`
> `b`
> 
> After this change, will I need to have the `a_hit` image or not?
> 
> Before this change, if I had a unit "a" that has 2 hitpoints and changes into a unit "b" when damaged, and also transferring the damage, I needed the following image names:
> `a`
> `a_hit`
> `b_hit`
> `b`
> 
> After this change, will I need to have the `a_hit` image or not?
> If the unit "b" can never be repaired (thus never really exist in its undamaged state), will I need the `b` image anyway or not?
> 
> What are the plans about documenting the matter for mapmakers to know which images they need to provide, in every case?

I don't know how to answer your questions.  This just prevents the error from occurring in a certain situation.  I bet there are other places that require the a_hit/b_hit images.  I'm not trying to change what images are required or not.  Just in this specific situation, the image doesn't appear to be needed so it shouldn't be requested.

trevan:
I created #7222 to add additional logging since it looks like there is at least one other place where the _hit image is being requested and it isn't finding it.  But that location isn't logging anything so we have no insight into how often that occurs.

Issue #5394, opened 2019-11-01
Documenting Everything Set By The Property "WW2V2" (1.9.0.0.13066)


Cernelius:
This is not much for myself, as I think I know everything a non-developer can know about this, already, but to cover what is maybe the biggest documentation black hole in pos2.

I don't think it should be assumed that each and every mapmaker is supposed to be already perfectly knowledgeable about the Classic and Revised referring rule sets, upon starting game-making (maybe just modding), and, even if this would be the case, it cannot be assumed TripleA being exactly and unmistakeably in line with such rule sets (and I can surely tell it isn't, albeit for arguably marginal matters).

This is all you get in pos2, as now:

https://github.com/triplea-maps/the_pact_of_steel/blob/master/map/games/pact_of_steel_2.xml
```
    <!-- ww2v2 rules override many individual rules to create revised rules

		<property name="WW2V2" value="false" editable="false">

				<boolean/>

		</property> -->
```
(by the way, why is this property even commented out in pos2, instead of being just there as false?)

What I think it should be there, instead, is:
1. "To create revised rules" from what? Not from nothing for sure, as TripleA has some rules also if you don't set this option true. If those individual rules, that are overridden by the property, are actually creating a non-Revised rule set of any kind, what that rule set actually is should be clarified, even if only by calling it simply "WW2V1". Side note, from that comment I don't believe it is even clear what are the "ww2v2 rules" and what are the "revised rules", that I believe are just two different ways to call the same rule set (again, something that is supposed to be a learning tool should not assume everyone to be well learned in even obscure matters). I don't know if TripleA is the only game engine in the world that doesn't even clarify what its default rule set is, as there is only one obscure mention of some "ww2v1 map", in pos2.
2. The list of all behaviours that this property changes directly (and are not specifically offered by other properties) (this is the most obscure item, that non-developers can currently know by testing only, which is overly time consuming, as you would need testing literally every single behaviour in which Classic and Revised may differ).
3. The list of all properties set true by this general property (this is something that a perfectly knowledgeable mapmaker can at least already learn, by reading all the properties and spotting those that create behaviours consistent with Revised but not with Classic; or should I said it could, would TripleA be always consistent...).
[...]

PR #7227, opened 2020-07-23
Track the isSub/isSuicide combo properties separately from the individual props

### LINKED ISSUE #7177: Planes don't attack subs
Global 40 House Rules with Canada

Follow the prompts in the saved games. 13066 allows planes to shoot at subs but 20365 doesn't.

### Do you have any ideas for an expected fix?
See how 13066 did it.

### Attach a Save Game

[13066 planes do attack subs.zip](https://github.com/triplea-game/triplea/files/4928323/13066.planes.do.attack.subs.zip)

[20234 planes don't attack subs.zip](https://github.com/triplea-game/triplea/files/4928325/20234.planes.don.t.attack.subs.zip)

Game Version:
2.1.20365

- [] yes
- [] no

I turned on map option "Option Redesign" and turned off "Air Attack Sub Restricted". Scramble one ftr from Scotland. You may need to run more than once if sub is killed before ftr would shoot at it.

Trying to add the "Regression" label but haven't been able to figure it out.

trevan:
Fixes #7177

With the change from isSub and isSuicide to the separated properties, the engine now checks the gameProperties only once when the save is loaded or the game map is loaded.  If the user then edits the gameProperties through the UI and changes sub or suicide related properties, they will not take affect.

This adds back the isSub and the isSuicide properties.  The separate properties are no longer set during setIsSub and setIsSuicide.  Instead, each of the properties will check if isSub or isSuicide is set.  And if a gameProperty is needed, it will check it then.  This way, the edited gameProperty will be seen when the game starts.

I've verified that this is save compatible with older saves in 2.x.  Those saves will not be fixed by this change.  The user will have to restart the game to see this fix.

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[x] Problem fix
[] Other:   

## Testing

I've played a few sub games with a few different properties.  It looks like the properties are acting correctly and that user editable properties work.

## Release Note

FIX|Disabling "Air Attack Sub Restricted" in Map Options will correctly allow airplanes to hit subs.
REGRESSION|Loading a saved games from a previous version for maps that have triggers to change isSub and isSuiced (such as Total World War 2.8 and 3.0, and HexGlobeFFA_w_TWW) may stop changing those properties during the triggers.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
@@ -226,6 +226,10 @@ public class UnitAttachment extends DefaultAttachment {
   // Manually set TUV
   private int tuv = -1;
 
+  // combo properties
+  private boolean isSub = false;
+  private boolean isSuicide = false;
+
   public UnitAttachment(final String name, final Attachable attachable, final GameData gameData) {
     super(name, attachable, gameData);
   }
@@ -629,13 +633,10 @@ public class UnitAttachment extends DefaultAttachment {
   }
 
   private void setIsSub(final Boolean s) {
-    setCanEvade(s);
-    setIsFirstStrike(s);
-    setCanMoveThroughEnemies(s && Properties.getSubmersibleSubs(getData()));
-    setCanBeMovedThroughByEnemies(s && Properties.getIgnoreSubInMovement(getData()));
+    isSub = true;
     if (s) {
       canNotTarget = null;
-      canNotBeTargetedBy = Properties.getAirAttackSubRestricted(getData()) ? null : new HashSet<>();
+      canNotBeTargetedBy = null;
     } else {
       resetCanNotTarget();
       resetCanNotBeTargetedBy();
@@ -647,7 +648,7 @@ public class UnitAttachment extends DefaultAttachment {
   }
 
   public boolean getCanEvade() {
-    return canEvade;
+    return canEvade || isSub;
   }
 
   private void setIsFirstStrike(final Boolean s) {
@@ -655,7 +656,7 @@ public class UnitAttachment extends DefaultAttachment {
   }
 
   public boolean getIsFirstStrike() {
-    return isFirstStrike;
+    return isFirstStrike || isSub || isSuicide;
   }
 
   private void setCanMoveThroughEnemies(final Boolean s) {
@@ -663,7 +664,7 @@ public class UnitAttachment extends DefaultAttachment {
   }
 
   public boolean getCanMoveThroughEnemies() {
-    return canMoveThroughEnemies;
+    return canMoveThroughEnemies || (isSub && Properties.getSubmersibleSubs(getData()));
   }
 
   private void setCanBeMovedThroughByEnemies(final Boolean s) {
@@ -671,7 +672,7 @@ public class UnitAttachment extends DefaultAttachment {
   }
 
   public boolean getCanBeMovedThroughByEnemies() {
-    return canBeMovedThroughByEnemies;
+    return canBeMovedThroughByEnemies || (isSub && Properties.getIgnoreSubInMovement(getData()));
   }
 
   private void setCanNotTarget(final String value) throws GameParseException {
@@ -698,7 +699,7 @@ public class UnitAttachment extends DefaultAttachment {
   public Set<UnitType> getCanNotTarget() {
     if (canNotTarget == null) {
       final Predicate<UnitType> unitTypeMatch =
-          (isSuicideOnAttack && isFirstStrike)
+          (getIsSuicideOnAttack() && getIsFirstStrike())
               ? Matches.unitTypeIsSuicideOnAttack().or(Matches.unitTypeIsSuicideOnDefense())
               : Matches.unitTypeIsAir();
       canNotTarget =
@@ -732,9 +733,11 @@ public class UnitAttachment extends DefaultAttachment {
   public Set<UnitType> getCanNotBeTargetedBy() {
     if (canNotBeTargetedBy == null) {
       canNotBeTargetedBy =
-          new HashSet<>(
-              CollectionUtils.getMatches(
-                  getData().getUnitTypeList().getAllUnitTypes(), Matches.unitTypeIsAir()));
+          Properties.getAirAttackSubRestricted(getData())
+              ? new HashSet<>(
+                  CollectionUtils.getMatches(
+                      getData().getUnitTypeList().getAllUnitTypes(), Matches.unitTypeIsAir()))
+              : new HashSet<>();
     }
     return canNotBeTargetedBy;
   }
@@ -1552,7 +1555,7 @@ public class UnitAttachment extends DefaultAttachment {
         defense
             + TechAbilityAttachment.getDefenseBonus(
                 (UnitType) this.getAttachedTo(), player, getData());
-    if (defenseValue > 0 && isFirstStrike && TechTracker.hasSuperSubs(player)) {
+    if (defenseValue > 0 && getIsFirstStrike() && TechTracker.hasSuperSubs(player)) {
       final int bonus = Properties.getSuperSubDefenseBonus(getData());
       defenseValue += bonus;
     }
@@ -1764,16 +1767,8 @@ public class UnitAttachment extends DefaultAttachment {
   }
 
   @Deprecated
-  private void setIsSuicide(final String s) {
-    setIsSuicide(getBool(s));
-  }
-
-  @Deprecated
   private void setIsSuicide(final Boolean s) {
-    setIsSuicideOnAttack(s);
-    // Global property controlled whether isSuicide units would suicide on defense
-    setIsSuicideOnDefense(s && !Properties.getDefendingSuicideAndMunitionUnitsDoNotFire(getData()));
-    setIsFirstStrike(s);
+    isSuicide = true;
     if (s) {
       canNotTarget = null;
     } else {
@@ -1783,7 +1778,7 @@ public class UnitAttachment extends DefaultAttachment {
 
   @Deprecated
   public boolean getIsSuicide() {
-    return isSuicideOnAttack || isSuicideOnDefense;
+    return isSuicide || isSuicideOnAttack || isSuicideOnDefense;
   }
 
   private void setIsSuicideOnAttack(final Boolean s) {
@@ -1791,7 +1786,7 @@ public class UnitAttachment extends DefaultAttachment {
   }
 
   public boolean getIsSuicideOnAttack() {
-    return isSuicideOnAttack;
+    return isSuicideOnAttack || isSuicide;
   }
 
   private void setIsSuicideOnDefense(final Boolean s) {
@@ -1799,7 +1794,9 @@ public class UnitAttachment extends DefaultAttachment {
   }
 
   public boolean getIsSuicideOnDefense() {
-    return isSuicideOnDefense;
+    return isSuicideOnDefense
+        // Global property controlled whether isSuicide units would suicide on defense
+        || (isSuicide && !Properties.getDefendingSuicideAndMunitionUnitsDoNotFire(getData()));
   }
 
   private void setIsSuicideOnHit(final String s) {
--- a/game-core/src/test/java/games/strategy/triplea/attachments/UnitAttachmentTest.java
+++ b/game-core/src/test/java/games/strategy/triplea/attachments/UnitAttachmentTest.java
@@ -1,9 +1,19 @@
 package games.strategy.triplea.attachments;
 
+import static games.strategy.triplea.Constants.AIR_ATTACK_SUB_RESTRICTED;
+import static games.strategy.triplea.Constants.IGNORE_SUB_IN_MOVEMENT;
+import static games.strategy.triplea.Constants.SUBMERSIBLE_SUBS;
+import static games.strategy.triplea.Constants.UNIT_ATTACHMENT_NAME;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.lenient;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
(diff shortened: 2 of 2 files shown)


game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
```diff
@@ -629,13 +633,10 @@ private void setIsSub(final String s) {
...
   private void setIsSub(final Boolean s) {
-    setCanEvade(s);
-    setIsFirstStrike(s);
-    setCanMoveThroughEnemies(s && Properties.getSubmersibleSubs(getData()));
-    setCanBeMovedThroughByEnemies(s && Properties.getIgnoreSubInMovement(getData()));
+    isSub = true;
     if (s) {
       canNotTarget = null;
-      canNotBeTargetedBy = Properties.getAirAttackSubRestricted(getData()) ? null : new HashSet<>();
+      canNotBeTargetedBy = null;
```
DanVanAtta, line 639: Codacy found an issue: [Assigning an Object to null is a code smell.  Consider refactoring.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=5897929)

DanVanAtta:


```diff

Issues
======
+ Solved 1
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=5897929&bid=19311411)

Cernelius:
I believe this is the proper way to handle this, as I've already detailed in here:
https://forums.triplea-game.org/topic/328/unit-option-can-submerge-hide-for-land-units-partisan-guerrilla-spy-diplomat-munition/44?page=3

Of course, I foresaw that the issue at hand (that this PR aims to fix) would have eventually happened, likely a few weeks or months after the release of 2.0 (as it did). As I said, I don't believe that the inability of changing properties mid game is an important item, so, please, rather refer to the other matters I raised at the linked topic.

CC: @ron-murhammer 

---

@trevan I'm curious if and how this may influence adding/setting `isSub` and `isSuicide` with triggers (within any unit attachment) during the course of the game. When 2.0 was released, the situation was that it was not (any longer) possible to do such a thing. Is this PR impacting on this matter in any ways?

Cernelius:
> This adds back the isSub and the isSuicide properties. The separate properties are no longer set during setIsSub and setIsSuicide. Instead, each of the properties will check if isSub or isSuicide is set. And if a gameProperty is needed, it will check it then. This way, the edited gameProperty will be seen when the game starts.

What I understand it will be the program behaviour after the merging of this PR is that in a game in which "isSub" is true for one or more units, each of the units will be non-targetable by air units if "Air Attack Sub Restricted" is true (no matter the setting of this property upon starting the game and at any previous point during the course of the game).

However, what will happen if, instead, the same unit has no "isSub" option defined but has "canNotBeTargetedBy" set to all air units types?

What I understand it will happen is that the "Air Attack Sub Restricted" acts as a behavioural switch for the "isSub" option only, but it is irrelevant for any of its atomized options (the "isSub" option i a shortcut of).

If my understanding is correct, I believe this would be not very good a behaviour, as it would give the "isSub" option an ability not covered by any of the options it is a shortcut of. A shortcut should be just a shortcut, not something that allows achieving anything not otherwise possibly achievable.

Meaning that I believe it is advisable that, if the "Air Attack Sub Restricted" property causes "isSub" units to be non-targetable by air units, then the same property should also be required to have the "canNotBeTargetedBy" option being effective (meaning that if the "Air Attack Sub Restricted" property is false, then all "canNotBeTargetedBy" options are ineffective (of course, this would be very confusing, since the "canNotBeTargetedBy" may not have only and all air units in list and may be assigned to a non-submersible unit, so the property should be renamed or deprecated)).

trevan:
> @trevan I'm curious if and how this may influence adding/setting `isSub` and `isSuicide` with triggers (within any unit attachment) during the course of the game. When 2.0 was released, the situation was that it was not (any longer) possible to do such a thing. Is this PR impacting on this matter in any ways?

I forgot about testing triggers.  This will break saved games for those maps.  I don't know how to handle saves for those maps since I can't look at the gameData properties as they aren't available during the deserialization of the UnitAttachment and I can't deduce if the UnitAttachment should be isSub or not based on the saved properties since a non-isSub unit can look like an isSub.

I don't know how to fix this issue and still handle old saves.  I'm closing this since it is an invalid fix.  Someone else who knows the code better will need to figure out a solution.

trevan:
> > This adds back the isSub and the isSuicide properties. The separate properties are no longer set during setIsSub and setIsSuicide. Instead, each of the properties will check if isSub or isSuicide is set. And if a gameProperty is needed, it will check it then. This way, the edited gameProperty will be seen when the game starts.
> 
> What I understand it will be the program behaviour after the merging of this PR is that in a game in which "isSub" is true for one or more units, each of the units will be non-targetable by air units if "Air Attack Sub Restricted" is true (no matter the setting of this property upon starting the game and at any previous point during the course of the game).
> 
> However, what will happen if, instead, the same unit has no "isSub" option defined but has "canNotBeTargetedBy" set to all air units types?
> 
> What I understand it will happen is that the "Air Attack Sub Restricted" acts as a behavioural switch for the "isSub" option only, but it is irrelevant for any of its atomized options (the "isSub" option is a shortcut of).
> 
> If my understanding is correct, I believe this would be not very good a behaviour, as it would give the "isSub" option an ability not covered by any of the options it is a shortcut of. A shortcut should be just a shortcut, not something that allows achieving anything not otherwise possibly achievable.
> 
> Meaning that I believe it is advisable that, if the "Air Attack Sub Restricted" property causes "isSub" units to be non-targetable by air units, then the same property should also be required to have the "canNotBeTargetedBy" option being effective (meaning that if the "Air Attack Sub Restricted" property is false, then all "canNotBeTargetedBy" options are ineffective (of course, this would be very confusing, since the "canNotBeTargetedBy" may not have only and all air units in list and may be assigned to a non-submersible unit, so the property should be renamed or deprecated)).

@Cernelius, That is tangentially related to this issue so it should be in its own separate issue so that it doesn't get lost.

Cernelius:
> > @trevan I'm curious if and how this may influence adding/setting `isSub` and `isSuicide` with triggers (within any unit attachment) during the course of the game. When 2.0 was released, the situation was that it was not (any longer) possible to do such a thing. Is this PR impacting on this matter in any ways?
> 
> I forgot about testing triggers. This will break saved games for those maps. I don't know how to handle saves for those maps since I can't look at the gameData properties as they aren't available during the deserialization of the UnitAttachment and I can't deduce if the UnitAttachment should be isSub or not based on the saved properties since a non-isSub unit can look like an isSub.
> 
> I don't know how to fix this issue and still handle old saves. I'm closing this since it is an invalid fix. Someone else who knows the code better will need to figure out a solution.

Dang! I believe your reworking is itself way more important than any save-game compatibility, especially considering that this would be a problem for only a very few users (which can still play with the old engine).

My suggestion, do this anyway, and just be sure to upgrade the TripleA version to 2.2 when you make a new release, giving the list of games that will lose save-game compatibility.

On this matter, I think it would be good to use the second digit for  partial compatibility losses like this one, reserving a third digit for completely compatible new releases (first digit will break compatibility, second digit may break compatibility, third digit will not break compatibility).

Cernelius:
> > > This adds back the isSub and the isSuicide properties. The separate properties are no longer set during setIsSub and setIsSuicide. Instead, each of the properties will check if isSub or isSuicide is set. And if a gameProperty is needed, it will check it then. This way, the edited gameProperty will be seen when the game starts.
> > 
> > 
> > What I understand it will be the program behaviour after the merging of this PR is that in a game in which "isSub" is true for one or more units, each of the units will be non-targetable by air units if "Air Attack Sub Restricted" is true (no matter the setting of this property upon starting the game and at any previous point during the course of the game).
> > However, what will happen if, instead, the same unit has no "isSub" option defined but has "canNotBeTargetedBy" set to all air units types?
> > What I understand it will happen is that the "Air Attack Sub Restricted" acts as a behavioural switch for the "isSub" option only, but it is irrelevant for any of its atomized options (the "isSub" option is a shortcut of).
> > If my understanding is correct, I believe this would be not very good a behaviour, as it would give the "isSub" option an ability not covered by any of the options it is a shortcut of. A shortcut should be just a shortcut, not something that allows achieving anything not otherwise possibly achievable.
> > Meaning that I believe it is advisable that, if the "Air Attack Sub Restricted" property causes "isSub" units to be non-targetable by air units, then the same property should also be required to have the "canNotBeTargetedBy" option being effective (meaning that if the "Air Attack Sub Restricted" property is false, then all "canNotBeTargetedBy" options are ineffective (of course, this would be very confusing, since the "canNotBeTargetedBy" may not have only and all air units in list and may be assigned to a non-submersible unit, so the property should be renamed or deprecated)).
> 
> @Cernelius, That is tangentially related to this issue so it should be in its own separate issue so that it doesn't get lost.

Can you answer me if I got everything right and, in this case, assure me I'm not misunderstanding anything?

In my opinion, such an issue (to consider the matter more comprehensively) should have been opened before making this PR.

trevan:
> > > @trevan I'm curious if and how this may influence adding/setting `isSub` and `isSuicide` with triggers (within any unit attachment) during the course of the game. When 2.0 was released, the situation was that it was not (any longer) possible to do such a thing. Is this PR impacting on this matter in any ways?
> > 
> > 
> > I forgot about testing triggers. This will break saved games for those maps. I don't know how to handle saves for those maps since I can't look at the gameData properties as they aren't available during the deserialization of the UnitAttachment and I can't deduce if the UnitAttachment should be isSub or not based on the saved properties since a non-isSub unit can look like an isSub.
> > I don't know how to fix this issue and still handle old saves. I'm closing this since it is an invalid fix. Someone else who knows the code better will need to figure out a solution.
> 
> Dang! I believe your reworking is itself way more important than any save-game compatibility, especially considering that this would be a problem for only a very few users (which can still play with the old engine).
> 
> My suggestion, do this anyway, and just be sure to upgrade the TripleA version to 2.2 when you make a new release, giving the list of games that will lose save-game compatibility.
> 
> On this matter, I think it would be good to use the second digit for partial compatibility losses like this one, reserving a third digit for completely compatible new releases (first digit will break compatibility, second digit may break compatibility, third digit will not break compatibility).

I think only @DanVanAtta or @RoiEXLab can make the call on breaking save compatibility.  The save game would load and it would play.  But the triggers would no longer work correctly.  They would be turning on/off `isSub` but the original atomic properties (`canEvade`, etc) would not be affected unless they had been disabled previously.

trevan:
> Can you answer me if I got everything right and, in this case, assure me I'm not misunderstanding anything?
> 
> In my opinion, such an issue (to consider the matter more comprehensively) should have been opened before making this PR.

I haven't looked that in depth in how this functionality works.  I'm just trying to ensure that when you change a map property through the UI, the changed property actually takes affect.  As to how the property works and interacts with pieces, I'm not that well versed.

Go ahead and create the issue if you feel it needs to be discussed more in depth.

Cernelius:
> > Can you answer me if I got everything right and, in this case, assure me I'm not misunderstanding anything?
> > In my opinion, such an issue (to consider the matter more comprehensively) should have been opened before making this PR.
> 
> I haven't looked that in depth in how this functionality works. I'm just trying to ensure that when you change a map property through the UI, the changed property actually takes affect. As to how the property works and interacts with pieces, I'm not that well versed.
> 
> Go ahead and create the issue if you feel it needs to be discussed more in depth.

Not being a developer (thus being unable to read through the code), I need to be confirmed on a very least to two points, before seriously considering opening anything:

1. The "Air Attack Sub Restricted" property causes units that are "isSub" to be non-targetable by air units (what this PR would have changed is just to make this true in any case, instead of relating it to the state of the property upon starting the game only).
2. The "Air Attack Sub Restricted" property does nothing to units that are both "canNotBeTargetedBy" and non-"isSub" (now and regardless of this PR being merge).

As a side note, since "isSub" is a shortcut (possibly setting "canNotBeTargetedBy" to all air units), I also wonder what would happen to units having both of the "isSub" and "canNotBeTargetedBy" options set.

trevan:
> > > Can you answer me if I got everything right and, in this case, assure me I'm not misunderstanding anything?
> > > In my opinion, such an issue (to consider the matter more comprehensively) should have been opened before making this PR.
> > 
> > 
> > I haven't looked that in depth in how this functionality works. I'm just trying to ensure that when you change a map property through the UI, the changed property actually takes affect. As to how the property works and interacts with pieces, I'm not that well versed.
> > Go ahead and create the issue if you feel it needs to be discussed more in depth.
> 
> Not being a developer (thus being unable to read through the code), I need to be confirmed on at least two points, before seriously considering opening anything:
> 
>     1. The "Air Attack Sub Restricted" property causes units that are "isSub" to be non-targetable by air units (what this PR would have changed is just to make this true in any case, instead of relating it to the state of the property upon starting the game only).

If a unit has `isSub` and the game has "Air Attack Sub Restricted" to true, then it gets `canNotBeTargetedBy` filled with all of the air units in the game.  If "Air Attack Sub Restricted" is false, then it has an empty `canNotBeTargetedBy`.  This PR doesn't change that functionality.  It only ensures that if "Air Attack Sub Restricted" is changed at any point (such as before the game starts through the "Map Options" UI button), then the current value is used instead of the hardcoded value in the XML.

> 
>     2. The "Air Attack Sub Restricted" property does nothing to units that are both "canNotBeTargetedBy" and non-"isSub" (now and regardless of this PR being merged).
> 
Yes.  "Air Attack Sub Restricted" is only used if the unit is `isSub`.

> 
> As a side note, since "isSub" is a shortcut (possibly setting "canNotBeTargetedBy" to all air units), I also wonder what would happen to units having both of the "isSub" and "canNotBeTargetedBy" options set.

I haven't tested this scenario, but the code looks like the setting of `canNotBeTargetedBy` will override the `isSub` value.  But it is possible that the `isSub` might override the `canNotBeTargetedBy` depending on the order of the properties in the XML.
```
<option name="isSub" value="true" />
<option name="canNotBeTargetedBy" value="something" />
```
and
```
<option name="canNotBeTargetedBy" value="something" />
<option name="isSub" value="true" />
```
might function differently just because the order is different.

Cernelius:
> If a unit has `isSub` and the game has "Air Attack Sub Restricted" to true, then it gets `canNotBeTargetedBy` filled with all of the air units in the game. If "Air Attack Sub Restricted" is false, then it has an empty `canNotBeTargetedBy`. This PR doesn't change that functionality. It only ensures that if "Air Attack Sub Restricted" is changed at any point (such as before the game starts through the "Map Options" UI button), then the current value is used instead of the hardcoded value in the XML.

Thanks for the clarification. In this case, while somewhat an improvement from current, I'm not feeling as much positive any longer about this PR. Still setting (thus burning in the save-game) the option based on the property state at the start of the game, then possibly ignoring such a setting thereafter looks to me like a workaround that fixes a problem while keeping the roots of it in place, putting a substantially duplicate process on top of a different one, for the same matter.

trevan:
> > If a unit has `isSub` and the game has "Air Attack Sub Restricted" to true, then it gets `canNotBeTargetedBy` filled with all of the air units in the game. If "Air Attack Sub Restricted" is false, then it has an empty `canNotBeTargetedBy`. This PR doesn't change that functionality. It only ensures that if "Air Attack Sub Restricted" is changed at any point (such as before the game starts through the "Map Options" UI button), then the current value is used instead of the hardcoded value in the XML.
> 
> Thanks for the clarification. In this case, while somewhat an improvement from current, I'm not feeling as much positive any longer about this PR. Still setting (thus burning in the save-game) the option based on the property state at the start of the game, then possibly ignoring such a setting thereafter, looks to me like a workaround that fixes a problem, while keeping the roots of it in place, putting a substantially duplicate process on top of a different one, for the same matter.

@Cernelius , I don't understand what you mean by that.  This isn't a work around to fix a problem or "burning in the save-game" or anything like that.  There are 4 properties that are affected by this: "Air Attack Sub Restricted", "Submersible Subs", "Ignore Sub In Movement", and "Defending Suicide and Munition Units Do Not Fire".  Right now, if you load a map with an `isSub` unit and has one of those options marked as editabled, then changing the value through the "Map Options" ui will not have any affect.

DanVanAtta:
Long thread, just a quick comment, if we broke save-game compatibility it would be a `3.0` rather than a `2.2`

trevan:
> Long thread, just a quick comment, if we broke save-game compatibility it would be a `3.0` rather than a `2.2`

Would it be considered breaking save-game compatibility if the save-game can load and be played but functionality will be incorrect?  For this PR, the game will load, the `isSub` triggers will just not work correctly.

DanVanAtta:
It could probably be argued both ways. A discussion issue with a pro/con
list is perhaps a good way to move forward.

trevan:
> It could probably be argued both ways. A discussion issue with a pro/con list is perhaps a good way to move forward.

I created #7237 to discuss this.

trevan:
I've reopened it per the discussion in #7237.

trevan:
Should I put something in the release notes about how this might break some saved games?

DanVanAtta:
@trevan yes, and potentially list the maps that would be impacted to make it easy/clear for users.

Cernelius:
A correction: While I still believe my understanding of the available documentation is correct, I just did some testing, and it appears that the property setting upon starting the game doesn't matter. What matters is only the setting defined in the xml, no matter if changed via "Map Options" (as far as the traditional terminology goes, wrong label, as it should say "Game Options"), before starting the game.

I tested it by giving a sea unit the option:
`<option name="isSub" value="true"/>`

And adding the property:
```
                <property name="Submersible Subs" value="false" editable="true">
                        <boolean/>
                </property>
```
No matter whether or not I set the property true (via "Map Options"), the unit was never having the "canMoveThroughEnemies" true behaviour.

---

As far as I recall, I believe this is not the behaviour wanted by the original creator of the "canMoveThroughEnemies" option (@ron-murhammer ). What I understood, during past discussions, was that the "isSub" shortcut was meant to set or not to set the option true depending on the "Submersible Subs" setting at the start of the game, not disregarding any use customization made before that point, as well (what I understood was that the user would have been unable to customize the behaviour after the start of the game, not unable to customize it at all, as it seems it is the case, instead). Again, as far as I recall.

trevan:
@Cernelius , are you saying that is the current behavior?  Or the behavior with this fix?  Because that is what I'm fixing.  If you tested with this branch and got that behavior, then I'll need to look into it.

Cernelius:
> @Cernelius , are you saying that is the current behavior? Or the behavior with this fix? Because that is what I'm fixing. If you tested with this branch and got that behavior, then I'll need to look into it.

I'm talking about the current behaviour (without this fix). At least from my limited testing, it appears that the current behaviour is following the setting in the game file (xml) regardless of any user customization. Meaning that, differently from what I previously said, the problem doesn't appear to be only that you cannot change the behaviour during the course of the game, but that you cannot customize it at all (also before starting the game).

trevan:
> > @Cernelius , are you saying that is the current behavior? Or the behavior with this fix? Because that is what I'm fixing. If you tested with this branch and got that behavior, then I'll need to look into it.
> 
> I'm talking about the current behaviour (without this fix). At least from my limited testing, it appears that the current behaviour is following the setting in the game file (xml) regardless of any user customization. Meaning that, differently from what I previously said, the problem doesn't appear to be only that you cannot change the behaviour during the course of the game, but that you cannot customize it at all (also before starting the game).

Yes, that is what I've been saying.  This should fix that so that you can customize it before you start the game.

Cernelius:
> > > @Cernelius , are you saying that is the current behavior? Or the behavior with this fix? Because that is what I'm fixing. If you tested with this branch and got that behavior, then I'll need to look into it.
> > 
> > 
> > I'm talking about the current behaviour (without this fix). At least from my limited testing, it appears that the current behaviour is following the setting in the game file (xml) regardless of any user customization. Meaning that, differently from what I previously said, the problem doesn't appear to be only that you cannot change the behaviour during the course of the game, but that you cannot customize it at all (also before starting the game).
> 
> Yes, that is what I've been saying. This should fix that so that you can customize it before you start the game.

Only before or also during the course of the game? Normally you should be able to take a save game, reload it, change settings and continue the game with the new settings.

For example, if the "Air Attack Sub Restricted" is set false and editable in the game file (xml), I should be able to enable it before starting the game, so that air won't be able to target submarines. However, then, I should be able to save the game, reload it, disable this option and continue the game with the submarines henceforth targetable by air.

trevan:
> > > > @Cernelius , are you saying that is the current behavior? Or the behavior with this fix? Because that is what I'm fixing. If you tested with this branch and got that behavior, then I'll need to look into it.
> > > 
> > > 
> > > I'm talking about the current behaviour (without this fix). At least from my limited testing, it appears that the current behaviour is following the setting in the game file (xml) regardless of any user customization. Meaning that, differently from what I previously said, the problem doesn't appear to be only that you cannot change the behaviour during the course of the game, but that you cannot customize it at all (also before starting the game).
> > 
> > 
> > Yes, that is what I've been saying. This should fix that so that you can customize it before you start the game.
> 
> Only before or also during the course of the game? Normally you should be able to take a save game, reload it, change settings and continue the game with the new settings.
> 
> For example, if the "Air Attack Sub Restricted" is set false and editable in the game file (xml), I should be able to enable it before starting the game, so that air won't be able to target submarines. However, then, I should be able to save the game, reload it, disable this option and continue the game with the submarines henceforth targetable by air.

It should do both.  Instead of checking the property only when you load the map, it checks the property each time is needed.

Cernelius:
Sounds good, even though I think it would make a better sense for any property to act only on the actual options, and every short-cut being no more than that. Sadly, currently the "isSub" and "isSuicide" are not really short-cuts, as you need them to use some properties for user customization (since the options they set are unable to).

I mean that rather than having "isSub" setting "canMoveThroughEnemies" true if "Submersible Subs" is true, I would advice "isSub" being a true short-cut for always setting "canMoveThroughEnemies" (and the other options) true, then having "canMoveThroughEnemies" active only as long as "Submersible Subs" is true. This way, the "isSub" would be truly a short-cut, as it would allow doing nothing more than what you can do by setting its components directly. This is what I told when the "isSub" was originally atomized (but was rejected).

Otherwise, I tend to suggest deprecating every property that is acting on short-cuts directly, if, going on, it is preferred, for example, not having the "canMoveThroughEnemies" options requiring the "Submersible Subs" property being true to be effective (which is actually not literally clear a correlation).

Opinions?

trevan:
> Sounds good, even though I think it would make a better sense for any property to act only on the actual options, and every short-cut being no more than that. Sadly, currently the "isSub" and "isSuicide" are not really short-cuts, as you need them to use some properties for user customization (since the options they set are unable to).
> 
> I mean that rather than having "isSub" setting "canMoveThroughEnemies" true if "Submersible Subs" is true, I would advice "isSub" being a true short-cut for always setting "canMoveThroughEnemies" (and the other options) true, then having "canMoveThroughEnemies" active only as long as "Submersible Subs" is true. This way, the "isSub" would be truly a short-cut, as it would allow doing nothing more than what you can do by setting its components directly (since also when you set the "canMoveThroughEnemies" option true directly in the game file (without using "isSub"), you would need to have the "Submersible Subs" property true for the option to be effective). This is what I told when the "isSub" was originally atomized (but was rejected).
> 
> Otherwise, I tend to suggest deprecating every property that is acting on short-cuts directly if, going on, it is preferred, for example, not having the "canMoveThroughEnemies" options requiring the "Submersible Subs" property being true to be effective (which is actually not literally clear a correlation).
> 
> Opinions?

I'm not going to change the existing usage of the properties in this PR.  I think that deserves its own PR so that people can debate the type of questions you are asking.

But as for my opinion, I would be very confused if my helicopter (which has canMoveThroughEnemies) is affected by the property "Submersible Subs".

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
```diff
@@ -226,6 +226,10 @@
   // Manually set TUV
   private int tuv = -1;
 
+  // combo properties
```
DanVanAtta, line 229: A 'combo' property is not well defined, what does this mean? Is this comment adding more than the questions it raises? I would suggest either explaining this more or just omitting the comment.

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
```diff
@@ -647,31 +648,31 @@ private void setCanEvade(final Boolean s) {
...
+    return canEvade || isSub;
   }
 
   private void setIsFirstStrike(final Boolean s) {
     isFirstStrike = s;
   }
 
   public boolean getIsFirstStrike() {
-    return isFirstStrike;
+    return isFirstStrike || isSub || isSuicide;
```
DanVanAtta, line 659: This is sad to see of course as options become super complex. What happens when I set 'isFirstStrike = false' and 'isSub = true' (well, it is first strike even though the option is set to off). I think that examples shows we lack flexibility, you cannot have a non-first strike sub unit or non-first strike suicide unit. 

Is the other option for us instead of the 'combo' property would be to update all maps? I'm not sure if there is really a good alternative to the code update proposed, I'm more thinking out loud and wondering what other options we might have.

DanVanAtta:
@trevan I updated the release note:
- only one release note per PR is supported
- the regression is rare enough and so nuanced I think it's going to be noise to the average player. If anyone reports a save-game having an issue, we can take up the problem then and just deal with it as it arrives. Perhaps it won't even be noticed..

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
trevan: Yes, the "combo" attributes are going to cause problems.  Before this code change, if you had
```
      <option name="isFirstStrike" value="false"/>
      <option name="isSub" value="true"/>
```
then your unit would have isFirstStrike == true.
But if you had
```
      <option name="isSub" value="true"/>
      <option name="isFirstStrike" value="false"/>
```
then your unit would have isFirstStrike == false.

With this change, at least the order doesn't matter anymore.  isFirstStrike will always be true.

If you want a non-first strike sub unit, you can make that.  You just can't use `isSub`.  You have to set the atomic properties yourself.  Same with a non-first strike suicide unit.

Another option, I think, is to add to the documentation that the individual atomic properties of the combo properties can't be overridden.  If they want to have custom units, then they need to use the atomic properties instead.  The combo properties are there just to simplify the creation of basic units.

If we want atomic properties to be able to override combo properties, then we'll need some sort of flag to indicate that the atomic property is actually set.  Because setting a property to false is currently the same thing as not setting the property at all.  But I think we should focus the documentation on moving towards atomic properties only and just mark all of the combo properties as deprecated.  Then, somehow, indicate to the map maker that the property shouldn't be used.

trevan:
> @trevan sorry for delay in getting some feedback to you on this. Would our only alternative to a combo property be to update all existing maps (with any of the notably properties in this PR)?

Updating the existing maps away from the combo property won't fix this issue.  Part of this issue is being able to turn on/off the map property "Air Attack Sub Restricted" and have it affect the unit.  The atomic properties don't care about the map property.  We would have to change the atomic properties to follow the map properties (which is what @Cernelius is advocating above).

Cernelius:
> > Sounds good, even though I think it would make a better sense for any property to act only on the actual options, and every short-cut being no more than that. Sadly, currently the "isSub" and "isSuicide" are not really short-cuts, as you need them to use some properties for user customization (since the options they set are unable to).
> > I mean that rather than having "isSub" setting "canMoveThroughEnemies" true if "Submersible Subs" is true, I would advice "isSub" being a true short-cut for always setting "canMoveThroughEnemies" (and the other options) true, then having "canMoveThroughEnemies" active only as long as "Submersible Subs" is true. This way, the "isSub" would be truly a short-cut, as it would allow doing nothing more than what you can do by setting its components directly (since also when you set the "canMoveThroughEnemies" option true directly in the game file (without using "isSub"), you would need to have the "Submersible Subs" property true for the option to be effective). This is what I told when the "isSub" was originally atomized (but was rejected).
> > Otherwise, I tend to suggest deprecating every property that is acting on short-cuts directly if, going on, it is preferred, for example, not having the "canMoveThroughEnemies" options requiring the "Submersible Subs" property being true to be effective (which is actually not literally clear a correlation).
> > Opinions?
> 
> I'm not going to change the existing usage of the properties in this PR. I think that deserves its own PR so that people can debate the type of questions you are asking.
> 
> But as for my opinion, I would be very confused if my helicopter (which has canMoveThroughEnemies) is affected by the property "Submersible Subs".

"Submersible Subs" is a badly named property anyway, to start with. What that property appears doing is turning the "submarines" behaviour from v1 (Classic) to v2 (Revised) rules (problems aside). Being able to submerge being only one item (likely the most important one).

In my opinion, the proper way to address such confusion is by renaming the property itself (deprecating the current name) or, better, making this property into what it should have been since ever: a short-cut only property for setting a number of other properties, one for every change from v1 to v2 it is causing.

In our case, one of the sub-properties that the "Submbersible Subs" short-cut property will set true may be called as "Land and Naval Units May Move Through Enemies" and, in turn, this property will be required being true for every "canMoveThroughEnemies" unit option to be effective.

This way, the "canMoveThroughEnemies" option is part of the "isSub" option (also setting other options) and the "Land and Naval Units May Move Through Enemies" property is part of the "Submersible Subs" property (also setting other properties).

In my opinion, short-cuts should be nothing else but that, and this strange process, we are going along with, of having shortcuts being necessary elements for user customization (meaning in case you want to let the user be able to opt) should never exist.

DanVanAtta:
Shortcut properties will generally always cause an issue. If you set a shortcut property, and set one of the shortcutted properties to a contrary value, then which one wins? I do like in this update that odering does not matter as much and to fix that you just enumerate the non-shortcut properties explicitly. 

In general, having properties be 1:1 I think is just going to be the cleanest. A shortcut is saving perhaps 2 or 3 lines per unit that has the shortcut, so in a XML file that can be thousands of lines, we're talking at best a dozen lines. I think the shortcut property was probably introduced as a concept when we realized that one property was controlling many options and we wanted those options individually. Instead of removing the shortcut, we added the individual properties and then defined the shortcut property in terms of the others. If that is the evolution, then really a shortcut property is a legacy property.

Cernelius:
> Shortcut properties will generally always cause an issue. If you set a shortcut property, and set one of the shortcutted properties to a contrary value, then which one wins?

In my opinion, having both a short-cut and one or more of anything the short-cut is setting is bad coding. The mapmaker should be informed never to do that.

In my opinion, I certainly believe this should work the same way in all cases. The way it currently works, for the record, is that priority goes to the option set after, for all options that cannot have multiple instances on the same unit.

I would personally prefer the behaviour being the game failing to parse if a same unit has multiple instances of options that cannot have multiple instances. This would, in turn, imply that any unit having both a short-cut option and one or more single-instance options that are set by it should fail to parse (thus being unavailable to every user), but adopting such a behaviour at this point will likely cause a huge number of (in my opinion, badly coded) games to fail to parse.

>I do like in this update that odering does not matter as much and to fix that you just enumerate the non-shortcut properties explicitly.

This is bad, because it is non-coherent. If the order matters for something, it should matter the same way for everything. If the order doesn't matter for something, it shouldn't matter for anything.

---

Short-cuts themselves are great. The best example of short-cuts are those that (problems aside) allow you to set true all properties making a rules set, which currently means only this, for example:
```
    <property name="WW2V2" value="true" editable="false">
      <boolean/>
    </property>
```
If the above short-cut property would not exist, any mapmaker wanting simply to have the v2 rules set would be forced to take care to set every relevant property correctly, which can only cause an increase in the likehood of setting something wrong or missing to set something or both.

DanVanAtta:
> In my opinion, having both a short-cut and one or more of anything the short-cut is setting is bad coding. The mapmaker should be informed never to do that.

Wouldn't it be even better if the XML only allowed for good coding?

> The way it currently works, for the record, is that priority goes to the option set after, for all options that cannot have multiple instances on the same unit.

This is not guaranteed. XML is not order-dependent. You cannot guarantee that any XML parser will respect ordering of tags. While it 'happens to work', it could break in the future depending on the XML parser. It's bad XML design to require ordering, notably for this reason.

>  If the order matters for something, it should matter the same way for everything.

Perhaps, but again, XML ordering dependency is fundamentally broken. It is not reliable. Anything that does rely on ordering should be considered as potentially buggy (it could stop working at any moment as soon as the bug is exposed). Re-ordering lines would be one way to do that. What it comes down to is the XML spec is a bit naive and not the best in some places.


> Re: WW2v2

That's not actually a shortcut. There are behaviors that depend exclusively on WW2v2 being set. You cannot completely emulate that property by setting other properties, it is its own property. The fact that other properties interact with WW2v2 is more the game engine being a mess.

Cernelius:
> I think the shortcut property was probably introduced as a concept when we realized that one property was controlling many options and we wanted those options individually. Instead of removing the shortcut, we added the individual properties and then defined the shortcut property in terms of the others. If that is the evolution, then really a shortcut property is a legacy property.

This is certainly (not just probably) the case for both the "isSub" and the "isSuicide" options. It may as well be the case for every short-cut currently in existance. Short-cuts that set only 1 option (for example, "isInfantry") are obviously old versions (the old name deemed bad enough to require a new one). Still, this doesn't mean that the short-cut is merely a legacy. Personally, if we would be adding all options the "isSub" is setting, I would suggest having a short-cut like "isSub". The legacy element is more in the wording of the short-cut itself, that goes back to the times when only sea units were allowed to have it, and they were supposed to be "submarines".

I can easily see some mapmaker setting something wrong or missing to set something, or both, while attempting to have a unit behaving like the "submarine" units of some rules set, with no "isSub" short-cut option to use.

DanVanAtta:
I'll grant there is convenience and simplicity in a shortcut, though it makes it complicated as soon as you get out of that best-case scenario and want for example a non-first strike sub. Having subs be a standard 2 or 3 options and then copy/pasted I'm sure is very manageable.

DanVanAtta:
>  The legacy element is more in the wording of the short-cut itself, that goes back to the times when only sea units were allowed to have it, and they were supposed to be "submarines".

Yeah, I think this is really similar to specifying multiple behaviors with the single option. Unit options I suspect are going to always be best enumerated one by one. This way you don't have the craziness of an air unit being "isSub". If we called each unit option by what the option is, then it's hard to think of when it would get out of date. 'FirstStrke' should always be, 'FirstStrike', and that helps since you can then have very sensible first strike air units.

Cernelius:
> > Re: WW2v2
> 
> That's not actually a shortcut. There are behaviors that depend exclusively on WW2v2 being set. You cannot completely emulate that property by setting other properties, it is its own property. The fact that other properties interact with WW2v2 is more the game engine being a mess.

Yes, I'm aware of this and I disagree with this. I think `WW2v2` should be merely a short-cut (as I've said in an issue more specific to the matter). Anyway, I think it's pretty clear that `WW2v2` is mainly a short-cut. I didn't enter into details merely to try to stay focused on the matter at hand.

This is the off-topic issue, where I would rather discuss such matters (as I said, I believe those behaviours you are mentioning are not even documented anywhere at all: Good luck mapmaking that if you are not a developer!):
https://github.com/triplea-game/triplea/issues/5394

DanVanAtta:
Ah, interesting you raised an issue @Cernelius 

Disagreeing/agreeing with WW2v2 being a short cut is not quite the point. I would agree though that it *should* be a shortcut (and at that point I'd like to see it removed!). IMO WW2v2 should be a setting only when speciyfing game rules when starting a game.  The checkbox would essentially disable all other settings and give you a default set of options. IMO it would be good to then be able to unselect it and override any options you would want. Similar for V3, so you could play with a specific rule set, or you could specify them individually. At the end of the day the engine is only then aware of the individual rules/options (so the engine would not even be aware of "V3", it would only care "do naval bombard casualties fire back?" (which could be turned on/off by however manner).

The way `WW2v2` is used in the code is a mess.. Which leads to exactly "Good luck mapmaking that if you are not a developer!", but even then "Good luck if you are a developer!"

Cernelius:
> Ah, interesting you raised an issue @Cernelius
> 
> Disagreeing/agreeing with WW2v2 being a short cut is not quite the point. I would agree though that it _should_ be a shortcut (and at that point I'd like to see it removed!).

For the removed part, as I've already argued against that in the issue, with other two developers, I'll just summarize the main two reasons why having a rules set short-cut is highly advantageous:
1- A lot of mapmakers have scarce knowledge of the differences between rules set and just want to pick one they are familiar with (usually, v2), especially in the moment the TripleA default is mostly the rather backward and unpopular "Classic": Having to set every property manually will easily result in missing something or setting something wrong, or both (and, with custom maps (not having a full rulebook, most likely), is, then, not necessarily easy to be sure whether something was originally wanted or not).
2- Some differences between the rules set are currently not supported by TripleA, with individual properties. Therefore, if anyone would add a missing property that applies only to some rules set, having a short-cut property for setting a whole rules-set allows fixing existent maps automatically (for example, if you add a missing property to support a currently unsupported v2 behaviour (as an example amongst many, a developer implements the call-off landings rules of Revised, which are currently completely missing), all games having the `WW2v2` property true should in theory benefit from the enhancement (keeping with the previous example, you will be able to call off landings when playing World At War, from that version of TripleA onwards)).

Please, refer to this issue for a more in dept analysis:
https://github.com/triplea-game/triplea/issues/5394

---
I might agree with not having rules set wide short-cut properties if TripleA would already perfectly support all rules set (no properties missing) and would have a flawless, very good and complete map-making program, in which you would choose a rules set and the game file (XML or whatever) would be autogenerated accordingly.

PR #7265, opened 2020-07-27
Add an initial schema for maps-db

DanVanAtta:
- Contains enough of a schema that we can model the triplea_maps.yaml file

## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[] Code Cleanup or refactor
[] Configuration Change
[] Problem fix
[] Other:   

Open to discussion on this, working draft PR. Schema is important!!

### DIFF
--- a/maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
+++ b/maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
@@ -1,0 +1,33 @@
+comment on database maps_db is 'Database used to store, track and serve map information';
+
+create table map (
+  id serial primary key,
+  name varchar(256) not null unique check(length(name) > 3),
+  date_created timestamptz not null default now(),
+  date_updated timestamptz not null default now()
+);
+
+create table category (
+  id serial primary key,
+  name varchar(32) unique not null check (length(name) > 2),
+  date_created timestamptz not null default now()
+);
+
+insert into category (name) values
+  ('BEST'),
+  ('GOOD'),
+  ('EXPERIMENTAL');
+
+
+create table map_version (
+  id serial primary key,
+  map_id integer not null references map(id),
+  version integer not null check (version > 0),
+  url varchar(256) not null check (url like 'http%'),
+  thumbnail varchar(256) not null check (thumbnail like 'http%'),
+  category_id integer not null references category(id),
+  description varchar(4096),
+  date_created timestamptz not null default now(),
+  date_updated timestamptz not null default now()
+);
+
--- a/maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
+++ b/maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
@@ -23,11 +23,15 @@ create table map_version (
   id serial primary key,
   map_id integer not null references map(id),
   version integer not null check (version > 0),
-  url varchar(256) not null check (url like 'http%'),
-  thumbnail varchar(256) not null check (thumbnail like 'http%'),
+  url varchar(256) not null unique check (url like 'http%'),
+  thumbnail varchar(256) check (thumbnail is null or thumbnail like 'http%'),
   category_id integer not null references category(id),
   description varchar(4096),
   date_created timestamptz not null default now(),
   date_updated timestamptz not null default now()
 );
+
+create unique index map_version_uniq_version
+on map_version (map_id, version);
+
 


DanVanAtta:
@RoiEXLab  / @trevan please have a look and let me know if you think of anything schema-wise here (potential problems, etc)

trevan:
> @RoiEXLab / @trevan please have a look and let me know if you think of anything schema-wise here (potential problems, etc)

It looks fine to me.

maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
```diff
@@ -0,0 +1,33 @@
+comment on database maps_db is 'Database used to store, track and serve map information';
+
+create table map (
+  id serial primary key,
+  name varchar(256) not null unique check(length(name) > 3),
+  date_created timestamptz not null default now(),
```
RoiEXLab, line 6: Not sure we need a creation date, but I guess it doesn't hurt

maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
```diff
@@ -0,0 +1,33 @@
...
+  ('GOOD'),
+  ('EXPERIMENTAL');
+
+
+create table map_version (
+  id serial primary key,
+  map_id integer not null references map(id),
+  version integer not null check (version > 0),
+  url varchar(256) not null check (url like 'http%'),
+  thumbnail varchar(256) not null check (thumbnail like 'http%'),
```
RoiEXLab, line 27: I don't think Map Thumbnails are a requirement right now.
So the not null constraint might be wrong, I could be wrong though

maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
```diff
@@ -0,0 +1,33 @@
...
+  category_id integer not null references category(id),
+  description varchar(4096),
+  date_created timestamptz not null default now(),
+  date_updated timestamptz not null default now()
```
RoiEXLab, line 31: Ok, I don't quite get what those timestamps are supposed to represent.
Doesn't the map table store the "last updated" date already?

maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
```diff
@@ -0,0 +1,33 @@
...
+insert into category (name) values
+  ('BEST'),
+  ('GOOD'),
+  ('EXPERIMENTAL');
+
+
+create table map_version (
+  id serial primary key,
+  map_id integer not null references map(id),
+  version integer not null check (version > 0),
```
RoiEXLab, line 25: Sooner or later we might want to revisit this too

maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
```diff
@@ -0,0 +1,33 @@
...
+create table map (
+  id serial primary key,
+  name varchar(256) not null unique check(length(name) > 3),
+  date_created timestamptz not null default now(),
+  date_updated timestamptz not null default now()
+);
+
+create table category (
+  id serial primary key,
+  name varchar(32) unique not null check (length(name) > 2),
```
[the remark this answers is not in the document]
DanVanAtta: Adding as many constraints as you can is a good thing when it comes to the DB. A category with an empty or 1 character name is not going to be correct.

The principle at play here is "database as fortress", and trying to assert that all data in database is as valid as possible:
https://www.oreilly.com/library/view/97-things-every/9780596800611/ch23.html

maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
DanVanAtta: Generally all rows should have both date_created and date_updated. Missing date_updated implies immutability. TBD how we handle deleted or removed maps. The value here is if a version were to go away when we could see when the map was originally uploaded.

maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
DanVanAtta: This will be a good thing to check, you could be right.

I wonder if there should be room to update the map thumbnail. Immutable DB rows is a good thing, perhaps a lookup table would be in order.

maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
DanVanAtta: The first version of a map and the map date_created likely are going to always match. The map table is the umbrella for all of the different map versions that can be uploaded.

A new map version being uploaded gets a new row and hence a date_created. If we change the category, then date_updated would change.

maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
DanVanAtta: I'm pretty sure we do not want semver. Checksumming the map and then making map versions immutable would solve all semver problems. Simply then if a game client is missing the map that is uniquely identified, it can download it and you then have a new map version on your system.

DanVanAtta: w.r.t map semver, how is a map maker to know what is a compatible map change and what is not? We _do not_ want map makers to have to have to continue to be 'mini-developers' knowing implementation details of how the code works. Currently they must know far too much about the internals that it just creates a really high barrier to entry. If I change for example a unit stat and upload a new version, the game engine should just do the right thing without me having to ask a dev if that is semver version increase, and to which number. WE then get into further issues with how the engine should handle such semver version numbers, it just is overly complex I think

RoiEXLab: I just read your post on the forum where you explained a couple of things.
I was more thinking about making use of SQL to determine the latest version.
There's no need to increment a version id if we could just look at the "creation date" and/or the compatible engine versions to determine a suited version to download.

maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
RoiEXLab: Ok, but wouldn't `map.date_created` and `map.date_updated` always be the "oldest"/newest entry of the corresponsing values in `map_version`?
I'm just thinking that there's some redundancy involved here

maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
RoiEXLab: I think I found one:
https://triplea-game.org/map/big-world/

The way the websites handles a missing image is by using a fallback image.

I assume by 'room to update the map thumbnail' you mean within the same version, i.e. in order to update the thumbnail a new version needs to be pushed right?
In this case I'm not quite sure, both approaches have their pros and cons.
If you mean in general I wouldn't agree.

maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
DanVanAtta: `map.date_updated` would be updated if the name of a map is changed. It would not be updated if the `map_version` table is touched. `map.date_created` might very well correspond to the 'oldest' entry of `date_created` in `map_version`. I think at that point if you want to answer a query like "get me the list of maps by order of creation date", doing a window query on map_version is something of a pain. One would also want a created & updated date on `map` so that you can know if the name has been modified since creation.

maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
DanVanAtta: I'm not convinced that the game engine semver is appropriate when it comes to map compatibility. For example, if we add a new property to maps and deprecate an old one, are we then making a major version update? 

I'm pretty sure we will likely be moving map parsing logic into a sub-project that is below game-headed. At that point the map parsing engine is no longer aware of the 'product.version' at all. 

Last, game engine version does not capture the range of compatibility for a map. We essentially could have a case where  map is compatible with 2.3, 2.4, 3.0, but not 3.1.

We should also consider that there can be multiple versions of a map suitable for a game engine, but perhaps we don't want the latest. The map itself could be updated to use latest properties (for example a map that does work with 2.0 by using legacy properties but is then updated to use 2.0 specific properties). 

I'm suspecting that we'll eventually create a "map version", IE: a specification value, similar to "HTML 4.0" that specifies the valid properties that can be defined in the map. Today we re-use `version` like crazy, which I think is over-use of that concept.

The version number here is in part needed just so we can transition from `triplea_maps.yaml` and have similar functionality. The game clients should be able to read data from either the server or the `triplea_maps.yaml` while we transition.

maps-db/src/main/resources/db/migration/V1.00.01__create_map_index_table.sql
DanVanAtta: If the thumbnail only is updated, I was thinking we'd have a mutable row and it would be an update statement.
If we want immutable rows, then we need a second table that has the listing of thumbnails. Thinking about it a bit more, IMO mutable rows here is fine.

> The way the websites handles a missing image is by using a fallback image.

DanVanAtta: > The way the websites handles a missing image is by using a fallback image.

I'm curious what kind of traffic and usage the 'maps' section is getting (?)

I'm not sure if we should continue maintaining it.. The download link is extra functionality, the browsing of maps is totally redundant to what we have in-game. I'd like to see evidence that 'maps' on the website is having an impact (I could see that being a good and proper thing if this were a web-app, in which case that is how you would look through maps to play them perhaps, but for us...)

DanVanAtta: Removed the thumbnail non-null constraint and added some additional UK constraints:  16d148b1363198440df7913e476f71665c6a73e2

DanVanAtta:
If all looks good, I think this would be ready for merge. I did want us to spend a moment on it as changing schema after it's deployed once does add a few steps (plus, schema is something that is best to spend some time thinking about and do well to avoid larger future problems).

PR #7307, opened 2020-08-01
Simplify, combine Unit constructors, remove unused constructor

DanVanAtta:
## Functional Changes

[] New map or map update
[] New Feature
[] Feature Removal
[x] Code Cleanup or refactor
[] Configuration Change
[] Problem fix
[] Other:

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/Unit.java
+++ b/game-core/src/main/java/games/strategy/engine/data/Unit.java
@@ -96,18 +96,10 @@ public class Unit extends GameDataComponent implements DynamicallyModifiable {
   private boolean chargedFlatFuelCost = false;
 
   /** Creates new Unit. Owner can be null. */
-  public Unit(final UnitType type, final GamePlayer owner, final GameData data) {
-    this(type, owner, data, UUID.randomUUID());
-  }
-
-  public Unit(final UnitType type, final GamePlayer owner, final GameData data, final UUID id) {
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/engine/data/Unit.java
```diff
@@ -96,18 +96,10 @@
...
-
-    checkNotNull(type);
-    checkNotNull(id);
-
-    this.type = type;
-    this.id = id;
+    this.type = checkNotNull(type);
+    this.id = UUID.randomUUID();
 
     setOwner(owner);
```
DanVanAtta, line 104: side-comment: @trevan I somewhat wonder if `originalowner` should be set in the constructor so that it is always non-null. It does look like `owner` should be non-null as it defaults to `null_player_id`. Hence, it's unclear if serialization has something to do with the nullity of unit owners.

trevan: `originalOwner` looks to be used in extremely few places (basically, ownership of factories) so making it non-null is probably a good idea.  I would assume it should default to the value of `owner`.  But I don't see how that affects how `owner` becomes null.

DanVanAtta: >  But I don't see how that affects how owner becomes null.

It doesn't, at least not directly that I know of. If there is somewhere that uses original owner in place of owner, then perhaps..

Issue #7311, opened 2020-08-01
Travis Master Branch Build Failed


tripleabuilderbot:
Travis master branch build [#20759](https://travis-ci.com/triplea-game/triplea/builds/178133294) failed.     Build stage: prerelease deployment

tripleabuilderbot:
Build [#20761](https://travis-ci.com/triplea-game/triplea/builds/178140759) failed on stage: prerelease deployment

tripleabuilderbot:
Build [#20762](https://travis-ci.com/triplea-game/triplea/builds/178150011) failed on stage: prerelease deployment

tripleabuilderbot:
Build [#20763](https://travis-ci.com/triplea-game/triplea/builds/178150030) failed on stage: prerelease deployment

tripleabuilderbot:
Build [#20765](https://travis-ci.com/triplea-game/triplea/builds/178150377) failed on stage: prerelease deployment

tripleabuilderbot:
Build [#20770](https://travis-ci.com/triplea-game/triplea/builds/178207506) failed on stage: prerelease deployment

tripleabuilderbot:
Build [#20773](https://travis-ci.com/triplea-game/triplea/builds/178226278) failed on stage: prerelease deployment

tripleabuilderbot:
Build [#20777](https://travis-ci.com/triplea-game/triplea/builds/178354356) failed on stage: prod2 deployment

tripleabuilderbot:
Build [#20778](https://travis-ci.com/triplea-game/triplea/builds/178356812) succeeded.

PR #7318, opened 2020-08-03
Bump bot version to 2.1.20765

DanVanAtta:
As usual, changes would take effect after a bot restart.
This latest version has a bug fix for the "host is already launching"
that is impacting numerous bots.

### DIFF
--- a/infrastructure/ansible/group_vars/prod2_bot.yml
+++ b/infrastructure/ansible/group_vars/prod2_bot.yml
@@ -1,2 +1,2 @@
-version: "2.1.20567"
+version: "2.1.20765"
 journalctl_applications: ["bot@01", "bot@02", "bot@03", "bot@04", "bot@05"]


Cernelius:
> As usual, changes would take effect after a bot restart.
> This latest version has a bug fix for the "host is already launching"
> that is impacting numerous bots.
> 
> ## Testing
> ## Screens Shots
> ## Additional Notes to Reviewer
> ## Release Note

By bot restart you also mean using the moderator "shutdown" feature?

Cernelius:
Well, anyway, I just restarted them all using "shutdown", since that reset the "started" column. Now all bots span from 09:08 to 09:12 for my timezone. Nobody was playing, thus nothing lost if useless.

prastle:
I restarted all bot servers with Linode.  Since I suspect they need a server reboot for this to take affect not single bot restarts.  I was unable to get London or Frankfort since they were both being used.

DanVanAtta:
Watch this issue: https://github.com/triplea-game/triplea/issues/7311

After that is cleared up we'll have an actual deployment to prod.

Re: "shutdown" @Cernelius , the application does not know anything about OS configuration, that the OS is configured to restart any stopped bots. If we were to remove that, or execute a 'shutdown' of a non-bot hosts, you'll see it's a true shut down. You can even experiment on a game you've hosted yourself and run a 'shutdown', it is a pretty hard stop to the server.

Re: linode reboot: @prastle , while I appreciate the attempts to get things sorted out, linode restarts are never necessary.

PR #7347, opened 2020-08-10
Bump version to 2.2 from 2.1

DanVanAtta:

### DIFF
--- a/game-core/src/main/resources/META-INF/triplea/product.properties
+++ b/game-core/src/main/resources/META-INF/triplea/product.properties
@@ -1,1 +1,1 @@
-version = 2.1.0
+version = 2.2.0


RoiEXLab:
Feel free to merge whenever you think is a good time

DanVanAtta:
@RoiEXLab FWIW -  there was no real time dependency here, this just updates the version on build artifacts.

Cernelius:
Should automated hosts be using 2.2 too?

I'm mostly worried that if someone releases a game with the
`<triplea minimumVersion="2.2"/>`
it won't be available in such hosts.

DanVanAtta:
We'll know if someone does. Hopefully map makers would use the actual min
version required and not latest

Cernelius:
I'm a bit surprised to hear that mapmakers are supposed to test and find out what is the earliest compatible version of TripleA for their maps, setting it in the "minimumVersion". I assumed, instead, that mapmakers would simply have a "minimumVersion" equal to the latest TripleA version.

I don't feel mapmakers having to backtrack and find out what is the earliest compatible version for their maps is feasible, for a number of reasons:
- That would increase the testing required on maps, to next to zero value (the value would be merely that someone could play the game using a released version prior to the current one).
- It has the negative value that (especially when someone reports something you never experienced in your playtesting) you'll have to wonder someone might have played your map with an older version than the release you tested it for (having a higher version would save you having problem reports from items already fixed).
- Since games and maps have usually a lot of options, properties and other elements used, and I believe nowhere is fully documented, for each specific element, what is the current earliest version of TripleA required for using it fully, that would imply that the only way mapmakers can tentatively know what is the earliest compatible version is testing everything, then setting the version equal to the one for the element having the highest one.
- Since some elements in TripleA (in particular, all properties) are not validated, it is very hard to test them, as the only way a non-developer can do it is by testing the behaviour (to see whether the property, or whatever, is actually doing anything at all or it is being simply ignored as inexistent (since the program doesn't tell you when it does that)).
- Even when validated, sometimes there are elements that work only from some TripleA version, while being loaded also on previous versions (so, would the minimum version be the minimum version that allows the game actually to work as playtested or the minimum version that allows the game to load without errors?).
- Sometimes older versions support the same items, but problems may have been fixed, so mapmakers may prefer to set a version after such fixes, even though the element was previously supported (with a wrong, or anyway different, behaviour).

---
An example is a new feature (like a new unit option) being added to TripleA 2.3. When TripleA 2.5 is out, a mapmaker makes a new map. Should the mapmaker really go opening and testing all previous TripleA releases until finding that the earliest one working is 2.3 or can he just save itself the trouble and document the "minimumVersion" as 2.5, if that is the only one it used to playtest the game?

---
Let's make an example: I have a game that has these unit attachments:
```
                <attachment name="unitAttachment" attachTo="wall" javaClass="games.strategy.triplea.attachments.UnitAttachment" type="unitType">
                         <option name="movement" value="0"/>
                         <option name="canNotMoveDuringCombatMove" value="true"/>
                         <option name="hitPoints" value="1"/>
                         <option name="whenHitPointsDamagedChangesInto" value="1:true:wall_fallen"/>
                         <option name="requiresUnits" value="metropolis"/>
                </attachment>
				
                <attachment name="unitAttachment" attachTo="wall_fallen" javaClass="games.strategy.triplea.attachments.UnitAttachment" type="unitType">
                         <option name="movement" value="0"/>
                         <option name="canNotMoveDuringCombatMove" value="true"/>
                         <option name="destroyedWhenCapturedBy" value="barbarians"/>
                         <option name="hitPoints" value="2"/>
                         <option name="whenHitPointsRepairedChangesInto" value="0:true:wall"/>
                         <option name="requiresUnits" value="metropolis"/>
                </attachment>
```

These attachments make it so the "wall" unit changes to "wall_fallen" and back, under some conditions.

The facts are:
- This game loads with TripleA 1.9, but the "wall" is never turned into "wall_fallen", because the "whenHitPointsDamagedChangesInto" is supported in 1.9 but it does nothing if the unit has already reached 0 hitpoints remaining.
- This game loads with TripleA 1.9, but it doesn't with older versions of 1.9, and they are actually all 1.9.0.0 (I don't know what is the 5th digit that is the earliest one able to load the game, and certainly I don't want to bother using time finding this out).
- I don't know whether TripleA 1.10 is the same as 1.9 or 2.0 or yet something different, on this matter, and I don't want to sink time testing it (minor matter, as 1.10 never existed as a release).
- While I believe (but, obviously, I cannot be sure) that everything will work for any 2 version of TripleA to date, game wise, map (original skin) wise all TripleA versions before a fix happened at some time during 2 (namely this one: https://github.com/triplea-game/triplea/pull/7217) require me to add an additional image, that is not needed any longer after that fix (so, either I add an image that is not being used by the latest TripleA or I set the minum version to assure that the game is playable only by TripleA versions after that fix, even though the game itself is loading fine since the earliest 2.0).

So, in this case, assuming I'm a new mapmaker and I made the map now and playtested it only with Triplea 2.2 (why would I use anything but the latest release?), am I supposed to set the "minimumVersion" at 2.2 (the only one I actually tested), at 2.1 (the earliest released version after the https://github.com/triplea-game/triplea/pull/7217 fix (not necessary game wise, but allowing me to not have an image in the map)), at 2.0 (the earliest released version for which the "wall" unit actually turns into "wall_fallen") or at 1.9 (a version for which the game will load and play without errors, except that the "wall" will never turn into "wall_fallen") (the game will work only for recent 1.9.0.0 versions, while very old 1.9.0.0 ones will not work)?

I thought the answer to this question was that the mapmaker would set the minimum version at simply the latest release (2.2) or at whatever version of TripleA under which it playtested the map in its finished form. Am I, instead, being told that a good mapmaker should somehow "use the actual min version required and not latest" (hard to say if, in the example, that is 2.1, 2.0 or 1.9.0.0, in my opinion), despite all the matters I detailed (and maybe more), both in terms of time loss and risk?

DanVanAtta:
I skimmed your response @Cernelius , but as a quick reply, it's not that complicated. Essentially they should use `2.0` unless they know they are using a feature that is introduced in a later version (which would like be a `3.0` considering a new map feature that cannot be read by old game engines is a non-backward compatible feature which is the definition of a major version upgrade)

Issue #7226, opened 2020-07-22
2.1.20365: AddUnits#lambda$buildUnitOwnerMap$0:45 - NullPointerException


tripleabuilderbot:
## User Description
I've been playing by playing a local game by making a move, saving, and sending the save through email to two other players. Only recently when I've been trying to open either of their save games I'm getting an "Error: null" popup.
The other two players can both open each others saves fine, and open my saves no problem.
We're on round 13 New World Order.

## Log Message
Error loading game file: /Users/johncrimmins/Desktop/7-21-20 I see Paris.tsvg

## TripleA Version
2.1.20365

## Java Version
11.0.6

## Operating System
Mac OS X

## Stack Trace
```
java.lang.NullPointerException
	at games.strategy.engine.data.changefactory.AddUnits.lambda$buildUnitOwnerMap$0(AddUnits.java:45)
	at java.base/java.util.stream.Collectors.lambda$uniqKeysMapAccumulator$1(Collectors.java:178)
	at java.base/java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1654)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
	at games.strategy.engine.data.changefactory.AddUnits.buildUnitOwnerMap(AddUnits.java:45)
	at games.strategy.engine.data.changefactory.AddUnits.readObject(AddUnits.java:85)
	at jdk.internal.reflect.GeneratedMethodAccessor13.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at java.base/java.io.ObjectStreamClass.invokeReadObject(ObjectStreamClass.java:1160)
[...]

PR #7366, opened 2020-08-14
Fix: Don't create the owner map in readObject as the owner might not yet be deserialized

trevan:
I think this is a better fix for #7226

I went through the code trying to understand why the owner wasn't available during deserialization and I realized that there can be a deserialization chain that goes Unit -> Owner -> ... -> AddUnits -> Unit.  In this situation, the Owner nor the Unit have been fully deserialized so during the `readObject` call in AddUnits, Unit will have a null owner.

And in the case where `unitOwnerMap` is null, the value of `buildUnitOwnerMap` was identical to `units`.  So instead of having a `readObject` to check for the the null case, I just moved that check inside of the code path.

As far as I can tell, the unit will never have a null owner while the game is actually running.  It can only have a null owner during deserialization.

## Testing
Played several games, loaded several saved games, stepped through the code with breakpoints

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/changefactory/AddUnits.java
+++ b/game-core/src/main/java/games/strategy/engine/data/changefactory/AddUnits.java
@@ -25,7 +25,7 @@ public class AddUnits extends Change {
    * performed. To ensure that the newly created units have the correct ownership, their original
    * owners are stored in this separate map.
    */
-  private Map<UUID, String> unitOwnerMap;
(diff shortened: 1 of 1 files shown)


DanVanAtta:


```diff

Complexity increasing per file
==============================
- game-core/src/main/java/games/strategy/engine/data/changefactory/AddUnits.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6013283&bid=19712440)

PR #7369, opened 2020-08-14
Update triplea_maps.yaml

mongoose1130:
adding my new map "Labyrinth"

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2143,3 +2143,27 @@
     <br>Map is in BETA, so expect errors while playing!
     <br>
     <br>
+- mapName: Labyrinth
+  mapCategory: EXPERIMENTAL
+  url: https://github.com/triplea-maps/labyrinth/archive/master.zip
+  version: 1
+  img: https://raw.githubusercontent.com/triplea-maps/labyrinth/master/description/MapThumbnail.png
+  description: |
+    <br>by Michael Hoover
+    <br>Version 1.0 for TripleA 2.0
+    <br>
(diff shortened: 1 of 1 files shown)


mongoose1130:
Not sure why there was a failed build here, all I did was add some text to the yaml. Help anyone? What do I do now?

beelee1:
double check your steps. This is a good guide:

https://docs.google.com/document/d/1FfF7N0srp9QG0_if5D-c1d1Aa1QTttdhxgm1GBh3pI4/edit

mongoose1130:
That's the guide I use. After looking over everything I believe it may be because I forgot to make the repo public before I transferred ownership to triplea-maps 😰

I'm looking for the option to do that but I don't think I have it anymore. Would someone be awesome and make it public for me?

beelee1:
yea I looked but couldn't find the repo. Probably have to ask one of the big dogs. Dan or Roi :) Sure they'll be cycling through stuff pretty soon :)

DanVanAtta:
This is odd, looks right. Trying to curl the download URL, it does give me a 404. Clicking the download link in browser, I get a `codeload` URL instead. I would have expected the CURL response to have been a 302 redirect, not a 404. I think we'll have to check back in a couple days to see if anything changes. This could portend a bad situation though where you'll need to create a release get a valid download URL.

mongoose1130:
I just figured we needed the repo to be made public (I forgot to do that before transferring ownership). Please let me know if there is anything I can do at this point. I'd really like to get my map out there!

PR #7472, opened 2020-08-27
Use DropwizardServerExtension to clean up data after test execution

DanVanAtta:
This update adds functionality to the extensions 'afterAll' block
to execute the SQL statements in a test/resources/db-cleanup.sql
file if it exists.

This update allows for most 'cleanBefore' tags to be removed and
removes non-determinism of DbRider tests where certain test
execution orderings can result in a FK constraint violation error.

Motivation & Benefit: DBRider tests will clear tables before inserting
data. One problem is that the table clearing logic is only aware of
the tables in that test. If other tables have a FK to one of these
cleared tables, there will be a constraint violation error. Worse yet,
this situation depends on the ordering of when tests are executed.

By deleting data using SQL from a file that is aware of all tables
and the correct order to delete data, we can execute this to ensure
that when any new tests run that there won't be any foreign key
reference violations when setting up new data.

### DIFF
--- a/dropwizard-common-test/src/main/java/org/triplea/dropwizard/test/DropwizardServerExtension.java
+++ b/dropwizard-common-test/src/main/java/org/triplea/dropwizard/test/DropwizardServerExtension.java
@@ -5,11 +5,15 @@ import io.dropwizard.Configuration;
 import io.dropwizard.db.DataSourceFactory;
 import io.dropwizard.testing.DropwizardTestSupport;
 import java.net.URI;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Collection;
 import lombok.extern.slf4j.Slf4j;
 import org.jdbi.v3.core.Jdbi;
 import org.jdbi.v3.core.mapper.RowMapperFactory;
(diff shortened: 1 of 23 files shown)


RoiEXLab (approved):
Sorry that I didn't review the other PRs in time.
Had a brief look on  them and they all look fine 👍

DanVanAtta:
@RoiEXLab no problem. Just please be sure to remove the 'post merge review' label if you do get a chance to go over them. It's always an epic challenge for everyone to have their free time to line up and manage iterative PRs/development. I think there probably is a place for post-merge if we are feeling pretty confident and are good about coming back to address feedback on already merged items.

DanVanAtta:
PS: the JUnit extension to inject JDBI, DAOs and do the DB cleanup I found to be quite slick, I was thinking this was something you'd particular enjoy seeing :smile:

RoiEXLab:
> Just please be sure to remove the 'post merge review' label if you do get a chance to go over them.

Ah didn't realize this, I will think about this next time!

> PS: the JUnit extension to inject JDBI, DAOs and do the DB cleanup I found to be quite slick, I was thinking this was something you'd particular enjoy seeing 😄

Yeah, the JUnit extensions feel really good to use ^^

PR #7487, opened 2020-08-30
A NonCombatMovementBonus Tech

WCSumpton:
This is a NonCombatMovementBonus Tech that uses TechAbilityAttachment to give bonus movement to units that move during the non-combat movement phase. The bonus can be positive or negative, and goes through the UnitAttachment Movement check to insure the effected units do not receive less then 0 movement. 

## Testing

Have tested with land units and works as desired. Wondering if it should only be limited to land and sea units?

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
@@ -16,6 +16,7 @@ import games.strategy.engine.data.Unit;
 import games.strategy.engine.data.UnitType;
 import games.strategy.triplea.Constants;
 import games.strategy.triplea.Properties;
+import games.strategy.triplea.delegate.GameStepPropertiesHelper;
 import games.strategy.triplea.delegate.Matches;
 import games.strategy.triplea.delegate.TechTracker;
 import games.strategy.triplea.delegate.TerritoryEffectHelper;
@@ -1479,9 +1480,13 @@ public class UnitAttachment extends DefaultAttachment {
   }
 
   public int getMovement(final GamePlayer player) {
+    final int nonCombatMoveBonus = GameStepPropertiesHelper.isNonCombatMove(getData(),true) ?
+          TechAbilityAttachment.getNonCombatMovementBonus(
+              (UnitType) this.getAttachedTo(), player, getData()) : 0;
     return Math.max(
         0,
         movement
+            + nonCombatMoveBonus
             + TechAbilityAttachment.getMovementBonus(
                 (UnitType) this.getAttachedTo(), player, getData()));
   }
--- a/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
@@ -1480,9 +1480,11 @@ public class UnitAttachment extends DefaultAttachment {
   }
 
   public int getMovement(final GamePlayer player) {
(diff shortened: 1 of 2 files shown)


DanVanAtta:
Cool to see the update @WCSumpton 

NB: The pact of steel map should be updated after this update to give an example.

It would be good to see a basic integration test around this feature be added as well.  There should also be a pact of steel map in the project as well (or something really similar) where you can also add an example of this attachment and then can verify that a given unit has bonus movement. Please let me know if that sounds reasonable enough and if you could use any further guidance on getting that set up.

There is a failing test to take care of, I think it might be this (it's 1 test in game-core that is failing on us here):

```
 MapPropertyTest > ValidateTest > shouldReturnFalseWhenValueIsMapAndValuesAreCompatibleButValueTypeIsNotSupported() STANDARD_ERROR

    Aug 30, 2020 2:26:04 AM games.strategy.engine.data.properties.MapProperty validate

    WARNING: Validation failed: cannot instantiate MapProperty with value type: java.lang.Long

MapPropertyTest > ValidateTest > shouldReturnFalseWhenValueIsMapAndContainsNullValue() STANDARD_ERROR

    Aug 30, 2020 2:26:04 AM games.strategy.engine.data.properties.MapProperty validate

    WARNING: Validation failed: cannot instantiate MapProperty with value type: <null>

```

I should be able to take a deeper look into this PR, I hope, as early as tomorrow morning or evening.

WCSumpton:
@DanVanAtta 
Thank you for reviewing this. I have run some testing with the following results:
Land and Sea units which did not move during CM or Attack were able to move during NCM with the bonus.
Air units, because they move during both CM and NCM were able to move during NCM with the bonus, even after they were involved in a battle.
I have made changes to prevent the Air units from getting the bonus, but the number of changes necessary (the unit would need to be passed to getMovement), seemed excessive. More changes, more chances for errors. So I left that check out.
I would be happy to include this into POS2, with a note about using it with isAir units.

The error seems to stem from 'GameStepPropertiesHelper.isNonCombatMove(getData(), true)' which may return a null value. IntelliJ suggest adding 'assert getData() != null;'. But I have not seen this problem with any testing I've done.

Cheers...

game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
```diff
@@ -1479,9 +1480,15 @@ private int getMovement() {
...
   public int getMovement(final GamePlayer player) {
+    final int nonCombatMoveBonus =
+        GameStepPropertiesHelper.isNonCombatMove(getData(), true)
+            ? TechAbilityAttachment.getNonCombatMovementBonus(
+                (UnitType) this.getAttachedTo(), player, getData())
+            : 0;
     return Math.max(
         0,
         movement
+            + nonCombatMoveBonus
```
DanVanAtta, line 1491: Do I understand right that the bonus is not additive? IE: if the tech says movement is '1', it's not +1 but actually set to a movement of one? If I understand that right, could you walk me through the reasoning why the bonus should not be additive?

DanVanAtta:
I'm curious if there is a way we can extend this property to all units in a sane manner. If not, if there is a way to make it more clear this property is only for specific unit types.

TBH this is a case where the current XML is being stretched, such configuration is not really right for an 'option' tag, but so it is...

@WCSumpton any thoughts on building an acceptance/integration test into the code to ensure this property is working end-to-end?

DanVanAtta:
@WCSumpton checking in, I assume you just haven't had time to come back to this PR? Are the next steps clear, any questions/concerns/considerations?

PR #7488, opened 2020-08-30
Speed up db rider tests

DanVanAtta:
## Commits 

commit d7fc3653d9bfe27af5fecceccc7873c4924274da

    Update DropWizardExtension: clean after each test & init JDBI once  & fail if server already started
    
    1. Re-creating DB connection takes some time, if we already have a DB
    connection (ie: jdbi != null), then do not re-created it. This is
    a matter of adding an if null check around the JDBI instantiation.
    
    2. Clean DB after each test, this helps avoid FK violations between
    tests within a test class. The execution of the clean-up is quite
    fast, this makes it easier to write valid tests without slowing
    things down noticably.

    3. To improve error handling and error messaging, configure the drop wizard
    extension server startup (starts server when tests execute) to fail if
    the server is already running. Given we no longer manually start a server
    in CI, this is okay.

commit d7fc3653d9bfe27af5fecceccc7873c4924274da

    Speed up Tests: Update all @DataSet to have 'useSequenceFiltering = false'
    
    This is a big boost in performance for DBRider tests that has more than a
    few tables. SequenceFiltering checks the right order in which to insert tables,
    presumably by attempting permutations. By not doing this, we see pretty significant
    test execution speed-ups.
    
    A downside to this udate is that DbRider will typically fail if we define
    multiple tables in a single YML. The order of tables in YML does not determine
    the order of creation in DbRider. To fix this, we can put each database table
    in its own YML and include multiple YML files per @Dataset. The order of included
    YMLs is deterministic relative to table creation and in this way we can
    ensure tables are created in the right order.

**Benchmarks**

Running all tests of the server subprojects in IDEA:

maps_server:  1.8s ->  0.9ms
lobby_server : 124s -> **9.7s**

Running integration test in Travis:  7m 14s ->  **3min 44s**


lobby-server/src/test/java/org/triplea/db/dao/access/log/AccessLogDaoTest.java
```diff
@@ -96,22 +101,16 @@ void searchForIp() {
...
-    accessLogDao.insertUserAccessRecord("registered", "127.0.0.20", "registered-system-id");
-  }
-
-  @Test
-  @DataSet(cleanBefore = true, value = "access_log/empty_data.yml")
-  @ExpectedDataSet("access_log/insert_anonymous_after.yml")
-  void insertAnonymousAccessLog() {
+  @DataSet(cleanBefore = true, value = EMPTY_ACCESS_LOG, useSequenceFiltering = false)
+  @ExpectedDataSet(value = "access_log/access_log_post_insert.yml", orderBy = "username")
+  void insertAccessLogRecords() {
```
DanVanAtta, line 112: Codacy found an issue: [JUnit tests should include assert() or fail()](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6092439)

DanVanAtta:


```diff

Issues
======
+ Solved 2
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6092439&bid=19947919)

DanVanAtta:


```diff

Issues
======
+ Solved 2
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6092439&bid=19947919)

lobby-server/src/test/java/org/triplea/db/dao/access/log/AccessLogDaoTest.java
DanVanAtta, line 112: Codacy found an issue: [JUnit tests should include assert() or fail()](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6092439)

DanVanAtta:


```diff

Issues
======
+ Solved 2
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6092439&bid=19947919)

lobby-server/src/test/java/org/triplea/db/dao/access/log/AccessLogDaoTest.java
DanVanAtta, line 112: Codacy found an issue: [JUnit tests should include assert() or fail()](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6092439)

dropwizard-common-test/src/main/java/org/triplea/dropwizard/test/DropwizardServerExtension.java
```diff
@@ -60,25 +60,24 @@
...
-      final String localUri = "http://localhost:" + support.getLocalPort();
-      serverUri = URI.create(localUri);
-      log.info("Local server URL set to: {}", localUri);
-    } catch (final RuntimeException e) {
-      log.warn("Ignoring setup error, server already started: {}", e.getMessage());
     }
+
+    final String localUri = "http://localhost:" + support.getLocalPort();
+    serverUri = URI.create(localUri);
+    log.info("Local server URL set to: {}", localUri);
```
RoiEXLab, line 76: Some thoughts here:
It seems that JUnit does not make any guarantees when an Extension is instantiated + there is no explicit closing of the DB connection in this Extension.

I did a little bit of research and from what I was able to tell here's how it's supposed to be done:
Instead of using static fields, the `ExtensionContext` provides a store to store intermediate values.
This would also allow to make use of the `ExtensionContext.Store.CloseableResource` interface to do the automated cleanup. Something like
```java

@Override
void beforeAll(ExtensionContext context) {
  context.getStore(ExtensionContext.Namespace.GLOBAL).getOrComputeIfAbsent("JDBI", () -> new JdbiWrapper(/* create jdbi*/), JdbiWrapper .class).getJdbi();;
}

@AllArgsConstructor
@Getter
static class JdbiWrapper implements ExtensionContext.Store.CloseableResource {

  private final Jdbi jdbi;

  @Override
  void close() {
    // cleanup of jdbi
  }
}
```

Thoughts?

lobby-server/src/test/java/org/triplea/db/dao/access/log/AccessLogDaoTest.java
DanVanAtta: I disabled the rule yesterday after this PR ran. In general I'd agree with the rule, for example an "AssertNotThrows" is a good thing to have even if there are not other assertions, but in our case we are testing via the expected data set. Going forward this should not be flagged though. I believe/hope you've the same permissions @RoiEXLab to turn off codacy patterns as well. To do that you click through to the codacy analysis and there is a settings option to ignore and/or suppress patterns.

dropwizard-common-test/src/main/java/org/triplea/dropwizard/test/DropwizardServerExtension.java
DanVanAtta: We used to do a 'suppport.close()' as well and that is removed.

The static field is needed as it turns out otherwise the support server and jdbi connection are all killed after the 'afterAll' phase. You'll get a server starting up anew for every test.

Re: closing of connection

Is it really needed? When the tests execute, if the connection is closed then, we're good, right? The tests closing are pretty much a `System.exit`, after which the connections should not survive. If we created many connections, then perhaps they would stack up, but we only have one. If we run all tests and somehow you get the server tests before others, having the connection be left open while those tests run does not seem like a big issue. Overall, the cleanup happens when the tests exit, is there really reason to do so explicitly?

Re: scope

This is quite arguably existing code at this point, the change here is not adding new connections/state. The conversion of the db-cleanup to after-each is to make the tests more reliable. To avoid having many iterative PRs the jdbi optimization is included as otherwise that would conflict with the db-cleanup update.

RoiEXLab: > The static field is needed as it turns out otherwise the support server and jdbi connection are all killed after the 'afterAll' phase. 

Is that the case even with `ExtensionContext.Store`?
I thought that the store has a longer lifespan than the individual tests and the store was the intended way of storing "state" between tests. Maybe I'm missing something though.

> The tests closing are pretty much a `System.exit`

Fair point, but not the cleanest approach in my opinion.
I mean on the same level one could argue that memory leaks in test code is fine because the tests never run for a long amount of time, but I just feel like some mechanism should take care of that.

RoiEXLab:
Side-Note: I was wondering why you deleted a comment from me, until I realized that you were likely just deleting the false positive warning, causing my reply to it getting deleted as well 😄

dropwizard-common-test/src/main/java/org/triplea/dropwizard/test/DropwizardServerExtension.java
DanVanAtta: Looking at this a bit more:
- I'm not seeing a why to even close a JDBI, I'm not sure it truly needs it. This query can be used to check open connections:
```  select sum(numbackends) from pg_stat_database;```
While executing tests this goes up to around 11 and then when the tests are done it goes back down. It appears that the connections are getting cleaned up without any need to do so explicitly.

- The ContextExtension namespace is only good for the lifecycle of the extension. Specifically this is one test file where you have a 'beforeAll' and an 'afterAll' being executed. By making the support and JDBI variables static we ensure that we only have one copy. The 'support.before' is what launches the server and it's a no-op if the server already exists. If we were to tie those to each unit test file, we might see issues with unable to bind to port and things would generally get quite slow. The cold start of tests is about a second, since we no longer need to create the server we pay that once. The JDBI connection costs a sizable fraction of a second.

DanVanAtta:
@RoiEXLab codacy is interacting through my user; removing the codacy check pattern it went ahead and deleted the initial comment. I was wondering the same and if I had a cat walk across the keyboard or something  :grin:

DanVanAtta:
(It's actually a TODO on my plate to re-create the codacy link using the service account user and not mine).

dropwizard-common-test/src/main/java/org/triplea/dropwizard/test/DropwizardServerExtension.java
DanVanAtta: > Is that the case even with ExtensionContext.Store?
I thought that the store has a longer lifespan than the individual tests and the store was the intended way of storing "state" between tests. Maybe I'm missing something though.

I did an experiment where I checked a value from a store in a 'beforeAll', set a value in a store and checked the value again. I found that that when we hit the 'beforeAll' the stored value is always initially unset.

So I think it depends on how you define individual tests. For the context of a store, that means the "@Test" methods, not the test file. Once the 'afterAll' is executed, it seems the store is wiped away. AFAIK the extension is instantiated on each test class that has it, hence the beforeAll and afterAll are executed as part of each test files life cycle.

> I mean on the same level one could argue that memory leaks in test code is fine because the tests never run for a long amount of time, but I just feel like some mechanism should take care of that.

We should avoid hyperbole, but one could make the same argument that you should set all of your variables to null to explicitly close them. It seems that the 'support' is doing its job of closing when the tests exit. I don't think there is a memory leak here at all. If you re-run the tests, you'll have all memory reclaimed in-between. 

DB connections can be leaked and remain open. `Tomee` is an example of a server notorious for that and after enough restarts your DB crashes because you exhaust all available connections. It's not the case for us though, DB connections are being closed and the server is being stopped.

Do we need the server to be stopped after a sup-project is done being tested? No.
Can we stop a server after the sub-project is done being tested? I'm not sure if we can, it requires the test runner to know that there are more test files to be executed. `AfterAll` is just the scope of one test file. 

I don't think what we're doing here is actually all that dirty at all even. `static` variables can lead to some dangerous patterns, but this seems like an exact fit for what we want here.

DanVanAtta: > Fair point, but not the cleanest approach in my opinion.
I mean on the same level one could argue that memory leaks in test code is fine because the tests never run for a long amount of time, but I just feel like some mechanism should take care of that.

If memory or connections are still allocated after a System.exit, that is a leak and would be bad for any kind of code. In our case we're intentionally leaving the server and DB connection open so that the next test file does not need to re-establish it. Given it's all closed up on System.exit, I would not call this a memory leak, it's intentionally left open between test (files).

RoiEXLab:
> @RoiEXLab codacy is interacting through my user; removing the codacy check pattern it went ahead and deleted the initial comment. I was wondering the same and if I had a cat walk across the keyboard or something 😁

Even better 😄

dropwizard-common-test/src/main/java/org/triplea/dropwizard/test/DropwizardServerExtension.java
RoiEXLab: > I did an experiment where I checked a value from a store in a 'beforeAll', set a value in a store and checked the value again. I found that that when we hit the 'beforeAll' the stored value is always initially unset.

Ok, in this case I'm a bit disappointed that there's no "real" mechanism to allocate a "global" resource and deallocate it within junit. Feels like something that's a very common thing to require.

Go ahead then 👍

PR #7497, opened 2020-08-30
Convert sub steps

trevan:
This converts the sub battle steps into the step classes.

One thing I don't like about this is that the new step classes require additional parameters (`order` and `returnFire`).  Both of them are required because of how the sub steps are determined.  They are determined at the beginning of the round with the initial BattleState.  By the time the steps will actually run, the BattleState may be different enough to cause the step to act differently.  An example of this is a battle where the `sub` is joined with a unit that can target a `destroyer` during the `aa` round and is attacking a group of units that include a `destroyer`.  At the start of the round, the presence of the `destroyer` affects when the `sub` will fire (if it is a defensive sub) and affects its `returnFire`.  If the `destroyer` is destroyed by the `aa` gun, then when the sub step executes, the `destroyer` is no longer present and so the `order` and the `returnFire` will be different.

The dead units are kept around in the `killedDuringCurrentRound` class member of `MustFightBattle` but it doesn't split the units between offensive and defensive units.  So I would need to figure out if the `destroyer` is on the opposing side or not and I couldn`t find any obvious code that would do that for me.

At some point, I'd like to make a change to how the battle steps are shown in the UI and how sub casualties are removed.  If the battle steps in the UI could be updated during the round, and if the sub casualties are moved to a dedicated step instead of tracked by the `returnFire` variable, I think I can make it work without these two variables.

## Testing

I loaded saves that include beginning of sub battles and in the middle of sub battles.  I discovered a save compatibility issue that is affecting master as well (see the `returnFireAgainstAttackingSubs` and `returnFireAgainstDefendingSubs` variables).


DanVanAtta:


```diff

Complexity increasing per file
==============================
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/FirstStrikeStepOrder.java  5
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/OffensiveFirstStrike.java  2
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/OffensiveFirstStrikeTest.java  1
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/FirstStrikeStepOrderTest.java  7
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/DefensiveFirstStrike.java  3
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/DefensiveFirstStrikeTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6093604&bid=19951532)

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/DefensiveFirstStrike.java
```diff
@@ -0,0 +1,80 @@
...
+
+  protected final BattleState battleState;
+
+  protected final BattleActions battleActions;
+
+  protected final FirstStrikeStepOrder order;
+
+  protected final ReturnFire returnFire;
+
+  public DefensiveFirstStrike(
```
DanVanAtta, line 28: Have you considered using Lombok @AllArgsConstructor to replace this constructor? The annotation should do the same thing.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
```diff
@@ -28,7 +28,11 @@
     SUB_OFFENSIVE_RETREAT_BEFORE_BATTLE,
     SUB_DEFENSIVE_RETREAT_BEFORE_BATTLE,
     SUBMERGE_SUBS_VS_ONLY_AIR,
+    FIRST_STRIKE_DEFENSIVE_SNEAK_ATTACK,
+    FIRST_STRIKE_OFFENSIVE,
+    FIRST_STRIKE_DEFENSIVE,
```
DanVanAtta, line 33: How is FIRST_STRIKE_DEFENSIVE different from FIRST_STRIKE_DEFENSIVE_NO_SNEAK_ATTACK? Perhaps a comment on the enum value would be useful here to describe the intent behind the distinction.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1534,15 +1544,20 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
...
 
-            @Override
-            public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
-              firstStrikeDefendersFire(firstStrikeOrder.getDefender().getReturnFire());
-            }
-          });
+        @Override
+        public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
+          new DefensiveFirstStrike(
+                  MustFightBattle.this,
```
DanVanAtta, line 1554: Seeing the same parameter twice is a bit difficult to grok. Have you considered using a builder for this constructor? In that case the parameters would be named and then more self-documenting.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1479,40 +1472,57 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
...
+                  MustFightBattle.this,
+                  // The order doesn't matter at this point
```
DanVanAtta, line 1520: I'm not sure this comment makes sense, what is "this point" relative to? Secondly, the ordering of what exactly, the step order?

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
DanVanAtta: Alternatively, have you considered perhaps having a second constructor that is simplified, or using static builder methods (factory pattern)?

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
```diff
@@ -446,8 +451,19 @@ private MustFightBattle givenFirstStrikeBattleSetup(
       final boolean ww2v2,
       final boolean defendingSubsSneakAttack) {
     final MustFightBattle battle = spy(newBattle(WATER));
-    lenient().doNothing().when(battle).firstStrikeAttackersFire(any());
-    lenient().doNothing().when(battle).firstStrikeDefendersFire(any());
+    lenient()
+        .doNothing()
```
DanVanAtta, line 455: Is this  'lenient' 'doNothing' expectation required?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
trevan: I should probably rename the enums.  A defensive sub can fire before the offensive subs fire (FIRST_STRIKE_DEFENSIVE_SNEAK_ATTACK), with the offensive subs (FIRST_STRIKE_DEFENSIVE), or with the other regular units (FIRST_STRIKE_DEFENSIVE_NO_SNEAK_ATTACK).  Any ideas on a name that is easier to understand?  I had thought "FIRST_STRIKE_DEFENSIVE_NO_FIRST_STRIKE" but that felt confusing as well.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: The parameters are only used twice because currently the BattleAction and BattleState is tied to the same class.

I plan to have a factory class that creates all of the steps, sorts them by order, and then returns the items.  I can't use that yet because the steps aren't all converted so they would be out of order.  The backwards compatible classes, though, will always be a little weird.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
DanVanAtta: First, are these enum values persisted as part of `gameData`, are we free to rename the existing ones? 
If not, creating a shim to allow for flexibility perhaps would be good to do sooner before we lock ourselves in.


Perhaps differentiating sneak attack and first strike and make them more mutually exclusive. In this case the order is:
- sneak attack
- first strike
- attack

DanVanAtta:


```diff

Complexity increasing per file
==============================
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/OffensiveFirstStrikeTest.java  1
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/DefensiveFirstStrike.java  3
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/FirstStrikeStepOrder.java  5
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/OffensiveFirstStrike.java  2
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/DefensiveFirstStrikeTest.java  1
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/FirstStrikeStepOrderTest.java  7
         

Complexity decreasing per file
==============================
+ game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6093604&bid=19951532)

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: This is talking about the next parameter.  I was trying to explain why I could pick a "random" `order`.  To construct the class, I need an `order` but at this point in the code, I can pick one of several.  I would prefer to use `firstStrikeOrder.getDefender()` but `firstStrikeOrder` doesn't exist in an old save and a null point exception would be thrown.

So, maybe I could change the comment to something like
```
firstStrikeOrder.getDefender() is the correct value but that isn't available in an old save so one of the possible FirstStrikeStepOrder.FirstStrikeResult has to be choosen
```

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
trevan: `BattleStep` enums shouldn't be persisted.  They are never used as an object field.

I'm not exactly sure what you mean?  Do you mean to rename the enums as "FIRST_STRIKE_DEFENSIVE_SNEAK_ATTACK", "FIRST_STRIKE_DEFENSIVE_FIRST_STRIKE", and "FIRST_STRIKE_DEFENSIVE_ATTACK"?

trevan:
I'm going to try and rework this a little so that the `returnFire` is handled differently.

trevan:
@DanVanAtta , are casualties from subs and naval bombardment supposed to be able to "change into" other units when they die?  Because the current code looks like it doesn't do that.

When a unit is hit by a sub or naval bombardment, the `removeCasualties` method is called.  If `returnFire` is NONE, then the units are immediately removed (by calling `remove`).  If `returnfire` is SUBS, then the non FirstStrike units are immediately removed.  But it looks like that `clearWaitingToDieAndDamagedChangesInto` is where units are changed.  And that is only called at the end of the aa phase and the full round.  And it only works on units that are in the `attackingWaitingToDie` and `defendingWaitingToDie` collections.

The `remove` method does create a `UnitsHitChange` with all of the units that can change when they are killed but I can't see how that will actually trigger the changing code.  I only see the code to change units in the `damagedChangeInto` which is only called by `clearWaitingToDieAndDamagedChangesInto`.

I'm asking this because I want to move units killed by subs (with `returnFire` NONE and SUBS) to the `damagedChangesInto` collection and then call `clearWaitingToDieAndDamagedChangesInto`.  And I worry that if I do that, I'll change an expected behavior where sub casualties shouldn't change into units.

DanVanAtta:
I suspect naval bombard and subs casualties not following 'damaged-change-into' is an oversight (a bug of omission). 

cc: @Cernelius  , please check out the previous comment, any thoughts?

trevan:
Here's another situation that I'm not sure of.  If the `WW2V2` option is enabled, then the subs always fire before the rest of the units.  The presence of a destroyer determines whether the casualties of the sub can return fire or not.

The situation I'm unsure of is when a sub and a destroyer attack a sub and a non-destroyer/sub (like a carrier or a battleship).  If the attacking sub hits, then if it hits the carrier, the carrier is sunk and can not fire back.  But if it hits the defending sub, that sub CAN fire back.

So, in `WW2V2` mode, should subs always be able to return fire?  Even if their "sneak attack" is negated because of the presence of the destroyer?

I don't know which rule set `WW2V2` is supposed to indicate but if I found https://axisallies.com/rules/axis-allies-rules-1942-2nd-edition.pdf and https://axisallies.com/rules/axis-allies-rules-classic-2nd-edition.pdf.  And in both of those 2nd edition rule sets, it would appear that this is a bug.

trevan:
Ok, I found https://axisallies.com/wp-content/uploads/2014/12/axis-allies-rules-revised.pdf which I think is `WW2V2`.  It does look like a defending sub does get to fire even if it is hit by the attacking sub.  Apparently, subs always get to fire.  It is only after both sides fire, then the casualties are removed depending on the presence of the destroyers.

beelee1:
@trevan A sub can't return fire if it was killed by a "aaGun" attack. That happens before "regular" combat though, so might not be pertinent to what you're doing.

trevan:
> @trevan A sub can't return fire if it was killed by a "aaGun" attack. That happens before "regular" combat though, so might not be pertinent to what you're doing.

@beelee1 , yeah, that is before the sub steps so I'm not looking at that.

DanVanAtta:
@beelee1 thanks for jumping in :smile: 
@Cernelius , @panther2 , would ya'll mind also weighing in? @trevan raised some interesting questions here.

trevan:
It does look like the "damaged-changed-into" functionality is working for sub casualties.  I haven't yet found a map that I can see it in action, but the `remove` method adds all killed units to the `killedDuringCurrentRound` and that collection is used by `clearWaitingToDieAndDamagedChangesInto` to transform the units.  It wasn't exactly clear the first few times of reading the code, though.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/ClearFirstStrikeCasualtiesTest.java
```diff
@@ -1,433 +1,419 @@
...
-    defendingUnits.add(givenAnyUnit());
-    if (parameters.contains(BattleStateVariation.HAS_DEFENDING_DESTROYER)) {
-      defendingUnits.add(givenUnitDestroyer());
-    }
-    if (parameters.contains(BattleStateVariation.HAS_DEFENDING_FIRST_STRIKE)) {
-      defendingUnits.add(givenUnitFirstStrike());
-    }
+  @ParameterizedTest
+  @MethodSource
+  void getStep(
```
DanVanAtta, line 36: Codacy found an issue: [The method 'getStep(List, boolean, boolean, boolean)' has an NPath complexity of 256, current threshold is 200](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6093604)

DanVanAtta:


```diff

Issues
======
- Added 2
           

Complexity increasing per file
==============================
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/BattleStateBuilder.java  6
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/ClearFirstStrikeCasualties.java  12
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/ClearFirstStrikeCasualtiesTest.java  15
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/DefensiveFirstStrike.java  4
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/DefensiveFirstStrikeTest.java  1
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/OffensiveFirstStrike.java  4
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/OffensiveFirstStrikeTest.java  1
         

Complexity decreasing per file
==============================
+ game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java  -4
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6093604&bid=19951532)

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/ClearFirstStrikeCasualties.java
```diff
@@ -0,0 +1,126 @@
...
+    return steps;
+  }
+
+  @Override
+  public Order getOrder() {
+    return Order.FIRST_STRIKE_REMOVE_CASUALTIES;
+  }
+
+  @Override
+  public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
```
DanVanAtta, line 84: Codacy found an issue: [The method 'execute(ExecutionStack, IDelegateBridge)' has an NPath complexity of 288, current threshold is 200](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6093604)

DanVanAtta:


```diff

Complexity increasing per file
==============================
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/BattleStateBuilder.java  6
- game-core/src/test/java/games/strategy/triplea/delegate/battle/FakeBattleState.java  2
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/DefensiveFirstStrike.java  4
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/DefensiveFirstStrikeTest.java  1
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/OffensiveFirstStrikeTest.java  1
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/OffensiveFirstStrike.java  4
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/ClearFirstStrikeCasualties.java  6
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/ClearFirstStrikeCasualtiesTest.java  9
         

Complexity decreasing per file
==============================
+ game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java  -4
         

Clones added
============
- game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java  1
- game-core/src/test/java/games/strategy/triplea/delegate/battle/FakeBattleState.java  1
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6093604&bid=19951532)

DanVanAtta:


```diff

Complexity increasing per file
==============================
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/ClearFirstStrikeCasualties.java  6
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/BattleStateBuilder.java  6
- game-core/src/test/java/games/strategy/triplea/delegate/battle/FakeBattleState.java  2
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/DefensiveFirstStrike.java  4
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/OffensiveFirstStrikeTest.java  1
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/DefensiveFirstStrikeTest.java  1
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/OffensiveFirstStrike.java  4
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/ClearFirstStrikeCasualtiesTest.java  9
         

Complexity decreasing per file
==============================
+ game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java  -4
         

Clones added
============
- game-core/src/test/java/games/strategy/triplea/delegate/battle/FakeBattleState.java  1
- game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java  1
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6093604&bid=19951532)

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
```diff
@@ -5,10 +5,16 @@
 import games.strategy.engine.data.Territory;
 import games.strategy.engine.data.Unit;
 import java.util.Collection;
+import java.util.EnumSet;
 
 /** Exposes the battle state and allows updates to it */
 public interface BattleState {
 
+  enum Side {
```
trevan, line 13: This was added to reduce `code complexity`.  I'm not exactly sure if this is the best name for it or if it should stay.

game-core/src/test/java/games/strategy/triplea/delegate/battle/MustFightBattleExecutablesTest.java
```diff
@@ -499,47 +495,6 @@ private void assertThatFirstStrikeStepOrder(
...
-    } else {
-      verify(battle).firstStrikeDefendersFire(returnFire);
-    }
-  }
-
-  @Test
-  @DisplayName(
-      "When attacker has a destroyer, defender has a destroyer, WW2v2 is true, "
-          + "and DEFENDING_SUBS_SNEAK_ATTACK is either, then attacker has return fire all")
-  void firstStrikeAttackerReturnFireAttHasDestroyerDefHasDestroyerWW2v2TrueSneakAttackTrueFalse() {
```
trevan, line 529: All of these tests are replaced with the tests in the three FirstStrike step classes.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/BattleStepsTest.java
```diff
@@ -122,8 +123,8 @@ public static UnitAndAttachment newUnitAndAttachment() {
     final Unit unit = mock(Unit.class);
     final UnitType unitType = mock(UnitType.class);
     final UnitAttachment unitAttachment = mock(UnitAttachment.class);
-    when(unit.getType()).thenReturn(unitType);
-    when(unitType.getAttachment(UNIT_ATTACHMENT_NAME)).thenReturn(unitAttachment);
+    lenient().when(unit.getType()).thenReturn(unitType);
```
trevan, line 126: Because of the change in how the sub steps are figured out, these mocks are being called sometimes when they aren't needed.  But to work around that would require making the test methods more complicated.  I felt adding lenient() to it was ok.

trevan:
I've added a new FirstStrikeCasualty step.  Instead of determining what the `ReturnFire` is and removing the casualties based on that, it will now clear the casualties list after the FirstStrike have made their sneak attack.  This allowed me to change the code to just have two sets of FirstStrike steps: one set for the sneak attack and one set for the regular attack.  The offense/defense steps will determine if they should be sneak or regular.

The WW2V2 required a bit of work in the casualty step.  Instead of just removing everybody in the `waitingToDie` collections, it will figure out which side has a "true" sneak attack and remove from that `waitingToDie` collection.  The other `waitingToDie` collection will stay around for the rest of the round and then be cleared up at the end of the round.

One odd thing is the two `ReturnFire` variables inside of `MustFightBattle#addFightSteps`.  The code sets them to `ReturnFire.ALL` but they are only needed for save compatibility.  These variables are in the saved game file and when the saved game is loaded, the value of those variables is updated.  So, in a saved game, the value could be `ALL`, `SUBS`, or `NONE`.  I've checked this.  I put a big comment above them to try and explain why they are there and why their value is hardcoded.

DanVanAtta:


```diff

Complexity increasing per file
==============================
- game-core/src/test/java/games/strategy/triplea/delegate/battle/FakeBattleState.java  2
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/DefensiveFirstStrikeTest.java  1
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/ClearFirstStrikeCasualtiesTest.java  9
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/OffensiveFirstStrikeTest.java  1
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/DefensiveFirstStrike.java  4
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/BattleStateBuilder.java  6
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/ClearFirstStrikeCasualties.java  6
         

Complexity decreasing per file
==============================
+ game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java  -4
         

Clones added
============
- game-core/src/test/java/games/strategy/triplea/delegate/battle/FakeBattleState.java  1
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java  1
- game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6093604&bid=19951532)

DanVanAtta:


```diff

Complexity increasing per file
==============================
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/OffensiveFirstStrikeTest.java  1
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/DefensiveFirstStrikeTest.java  1
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/OffensiveFirstStrike.java  4
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/ClearFirstStrikeCasualties.java  6
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/BattleStateBuilder.java  6
- game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/ClearFirstStrikeCasualtiesTest.java  9
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/DefensiveFirstStrike.java  4
- game-core/src/test/java/games/strategy/triplea/delegate/battle/FakeBattleState.java  2
         

Complexity decreasing per file
==============================
+ game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java  -4
         

Clones added
============
- game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java  1
- game-core/src/test/java/games/strategy/triplea/delegate/battle/FakeBattleState.java  1
- game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6093604&bid=19951532)

panther2:
>Ok, I found https://axisallies.com/wp-content/uploads/2014/12/axis-allies-rules-revised.pdf which I think is WW2V2. It does look like a defending sub does get to fire even if it is hit by the attacking sub. Apparently, subs always get to fire. It is only after both sides fire, then the casualties are removed depending on the presence of the destroyers.

Indeed in this edition subs always fire during the opening step. It is the presence of a destroyer that enables the sub's casualty to fire back.

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleActions.java
```diff
@@ -35,4 +38,27 @@ void landParatroopers(
...
+
+  void remove(
+      Collection<Unit> killedUnits,
+      IDelegateBridge bridge,
+      Territory battleSite,
+      Boolean defenderDying);
+
+  void damagedChangeInto(
+      GamePlayer player,
+      Collection<Unit> units,
```
DanVanAtta, line 61: We're baking in an in/out parameter here and this value is modified.
What would you think if we returned instead the collection of transformed units? That would imply that the caller would be the one doing the `IDelagetBridge` work.

DanVanAtta (commented):
Some comments about the new API @trevan , sorry it took so long to get back to this. Please @ mention me when this is good for another look and/or responses.

trevan:
@DanVanAtta , those are all good ideas.  But all of those apis are existing apis and some of them are used by other parts of battle steps.  If you want me to make changes to the api now, I'll do it but I'd prefer to wait till after all the battle steps have been converted.

DanVanAtta:
@trevan IMO better to come back to it sooner; it's a tall order to revisit such details after a longer amount of time. I'm open for those updates to be done in upcoming PR ("fast-follow") if you're willing rather than done in this PR. WDYT @trevan ?  Is there another reason I'm missing to wait for further conversions, is that going to be easier or just more you'd rather stay focused on that?

DanVanAtta:
Considering all the APIs are just extracted, I'll leave it up to you @trevan when it's revisted. My preference is to do so sooner while we have context and to make sure it gets done (and less things to track and have on your radar), but up to you. Will go ahead and merge this now, thank you for working on this @trevan   :+1:

trevan:
@DanVanAtta 

> Is there another reason I'm missing to wait for further conversions, is that going to be easier or just more you'd rather stay focused on that?

I mostly want to stay focused on the step conversion.  But getting all of the step conversions done first should make it easier to clean up the api.  Because then I'll have a better idea on how each part works and can clean up the api.

PR #7503, opened 2020-08-31
Refactor: move method getNeighborsValidatingCanals to RouteFinder

DanVanAtta:
Reduces game-core dependencies in GameMap

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/RouteFinder.java
+++ b/game-core/src/main/java/games/strategy/engine/data/RouteFinder.java
@@ -3,6 +3,7 @@ package games.strategy.engine.data;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import games.strategy.triplea.delegate.TerritoryEffectHelper;
+import games.strategy.triplea.delegate.move.validation.MoveValidator;
 import java.math.BigDecimal;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


DanVanAtta:


```diff

Complexity decreasing per file
==============================
+ game-core/src/test/java/games/strategy/engine/data/RouteFinderTest.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6094046&bid=19952883)

game-core/src/main/java/games/strategy/engine/data/RouteFinder.java
```diff
@@ -16,27 +17,29 @@
 import java.util.Set;
 import java.util.function.Function;
 import java.util.function.Predicate;
+import javax.annotation.Nullable;
+import lombok.AccessLevel;
+import lombok.RequiredArgsConstructor;
 
+@RequiredArgsConstructor(access = AccessLevel.PACKAGE, onConstructor_ = @VisibleForTesting)
```
RoiEXLab, line 24: As far as I can tell this constructor is only used by this class, so it might as well be private, without the annotation.

game-core/src/main/java/games/strategy/engine/data/RouteFinder.java
```diff
@@ -16,27 +17,29 @@
...
 class RouteFinder {
 
+  private final MoveValidator moveValidator;
```
RoiEXLab, line 27: I'm not entirely convinced this field is really necessary.
The `GameMap` object is still there, so why not create this object on the fly inside of `getNeighborsValidatingCanals`?

game-core/src/main/java/games/strategy/engine/data/RouteFinder.java
DanVanAtta: Yeah, this one went in circles. I was expecting to need to be able to control `MoveValidator` for test, though it appears it's not covered by the test cases and therefore injecting a mock is not necessary.

game-core/src/main/java/games/strategy/engine/data/RouteFinder.java
DanVanAtta: Creating an object on the fly is a violation of IOC. It would be equivalent to a static method call, which would be an example of static coupling. The intent was to get a mock in there so that this class could be tested, 'map' is already mocked and I pretty much assumed the move validator path was going to need the same.

In the short term, raising the dependency from a method where there is in effect a static dependency to a constructor injection can help us decouple this class from 'gameData' (which in turn comes from gameMap). This class is already very odd as the predicate member probably should *not* be a constructor arg..

DanVanAtta:


```diff

Complexity decreasing per file
==============================
+ game-core/src/test/java/games/strategy/engine/data/RouteFinderTest.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6094046&bid=19952883)

PR #7506, opened 2020-09-01
Draft review of new XML parser

DanVanAtta:
Opening a draft for early feedback on a new XML parser that uses the Stax XML processing API.

Stax essentially holds a cursor and you stream through XML asking for next elements and you get an event flag to let you know what kind of data you can read. I wrote a generic builder around this so you can add handlers for child tags and attribute and body text to then set that data to back to value objects.

The really important concept beyond the new parser is a new sub-project for map-data that has straight POJOs representing the XML data. Eventually there will be another layer above that constructs these POJO objects into a "MapState" (the name "MapData" is already in use) which could then be used to create a "GameData". 

The testing of the parsing is also pretty interesting I think as well since we can create very minimal XML snippets and verify that it maps to the correct POJO. I believe there is potentially a lot of room for error handling and data validation.

A benefit from this approach is that we have doc type validation turned off and we can relatively seemlessly add in the defaults that doctype is providing. One blocker for old & new maps is when doctype needs to or has changed - not using it is a goal. 

Finally though, and I want to emphasize the reasoning behind this approach, once we have a XML mapped to POJO objects the assembly into a game data can be done in logic that is easy to control. With luck this will give us both a lot of flexibility for additions & variations as well as better read performance.


map-data/src/main/java/org/triplea/map/reader/MapElementReader.java
```diff
@@ -0,0 +1,66 @@
...
+
+  private static GameTag doRead(final XMLStreamReader streamReader) throws XMLStreamException {
+
+    final int eventType = streamReader.next();
+
+    switch (eventType) {
+      case XMLStreamReader.START_ELEMENT:
+        final String elementName = streamReader.getLocalName();
+
+        switch (elementName) {
```
DanVanAtta, line 48: Codacy found an issue: [Switch statements should have a default label](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6102557)

DanVanAtta:


```diff

Issues
======
- Added 4
           

Complexity increasing per file
==============================
- map-data/src/main/java/org/triplea/map/reader/MapElementReader.java  4
- map-data/src/main/java/org/triplea/map/data/elements/AttachmentListTag.java  1
- map-data/src/main/java/org/triplea/map/data/elements/GameTag.java  1
- map-data/src/main/java/org/triplea/map/data/elements/DiceSidesTag.java  1
- map-data/src/main/java/org/triplea/map/reader/XmlParser.java  14
- map-data/src/main/java/org/triplea/map/data/elements/TripleaTag.java  1
- map-data/src/test/java/org/triplea/map/reader/AttachmentListTagReadingTest.java  1
- map-data/src/test/java/org/triplea/map/reader/DiceSidesReadingTest.java  1
- map-data/src/test/java/org/triplea/map/reader/InfoTagReadingTest.java  1
- map-data/src/test/java/org/triplea/map/reader/XmlReaderTestUtils.java  3
- map-data/src/main/java/org/triplea/map/reader/MapReadResult.java  3
- map-data/src/main/java/org/triplea/map/data/elements/InfoTag.java  1
- map-data/src/test/java/org/triplea/map/reader/TripleaTagTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6102557&bid=19976423)

map-data/src/main/java/org/triplea/map/reader/XmlParser.java
```diff
@@ -0,0 +1,137 @@
...
+          if (attributeHandler != null) {
+            String attributeValue = streamReader.getAttributeValue(index);
+            attributeHandler.accept(attributeValue);
+          }
+        }
+      }
+
+      while (streamReader.hasNext() && !endTagReached) {
+        final int event = streamReader.next();
+        switch (event) {
```
DanVanAtta, line 111: Codacy found an issue: [Switch statements should have a default label](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6102557)

map-data/src/test/java/org/triplea/map/reader/XmlReaderTestUtils.java
```diff
@@ -0,0 +1,30 @@
...
+
+  static GameTag parseMapXml(final String xmlFileName) {
+    try (InputStream stream = openFile(xmlFileName)) {
+      final MapReadResult readResult = MapElementReader.readXml(xmlFileName, stream);
+
+      assertThat(readResult.getErrorMessages(), is(empty()));
+      assertThat(readResult.getGameTag(), isPresent());
+      return readResult.getGameTag().orElseThrow();
+    } catch (final IOException e) {
+      throw new RuntimeException("Failed to parse (expecting a valid XML file): " + xmlFileName, e);
```
DanVanAtta, line 27: Codacy found an issue: [Avoid throwing raw exception types.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6102557)

map-data/src/main/java/org/triplea/map/reader/MapElementReader.java
```diff
@@ -0,0 +1,66 @@
...
+    } finally {
+      streamReader.close();
+    }
+  }
+
+  private static GameTag doRead(final XMLStreamReader streamReader) throws XMLStreamException {
+
+    final int eventType = streamReader.next();
+
+    switch (eventType) {
```
DanVanAtta, line 44: Codacy found an issue: [Switch statements should have a default label](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6102557)

Cernelius:
I suggest testing this on the "War of the Relics" game to assure that won't be slower than it currently is (the map itself is very unpopular, but gives a good extreme example to check).

On this regard, I've already noticed a worsening in performances, by the way: If I open the game with 2.2.20868, it takes 9 seconds to load, while if I open it with 1.9.0.0.13066 it takes only 5 seconds.

I'm not able reliably to test on games that take very little time to load, so I wonder if the developers are aware of this worsening of parsing, or at least loading, time since the latest 1.9.0.0 release?

After testing at standard settings for 2.2.20868, meaning
```
-Xmx2048M
-Xms2048M
-Xss1250K
```
I've also changed the "vmoptions" settings for 2.2.20868 to
```
-Xmx4096M
-Xms4096M
-Xss256K
```
and this lowered the loading time to 7 or 8 seconds (instead of 9).

DanVanAtta:
@Cernelius one should look at the number of garbage collections that happened. The second saving might be because there were fewer GC's, but when one does happen you might find it is exponentially longer. This is what makes memory tuning not straight forward, more memory is not necessarily better.

But I appreciate the feedback nonethless. The update/direction here is being designed to perform well but it not be more performant in the end. Performance is not the goal. It'll also be a while before we can do any kind of benchmarking, if we ever do, as these updates will come in waves and won't be complete for some time.

Really the goal here is flexibility, to allow us to specify XML with more variations and more easily and get us to a place as well where there is less map data in 'gameData' and put us in a place perhaps where save games can simply be a list of changes from an initial map data (which then makes save game compatibility a **far** easier problem to solve). For developers it's even better as save game compatibility prevents the most basic of updates, it is extremely constraining and burdensome to keep testing and designing for it.

To finish discuss performance, it's not clear how much of a hit there will be when we recombine java objects together post-parsing and if that'll actually be faster. The current parsing engine does build a full model of the XML DOM and dies on the WoW map. We do want to get XSS back to a default value as it does seem there is a large increase in memory crashes since we did that. The STAX parser has promise as it's about as light weight as you can get for parsing XML, we very much should *not* see the stack overflow errors when reading an XML. (But again, the main goal here is better code structure and flexibility, decouple testing so that we have better test coverage, simpler and more modular design, and get to a place where we can more easily support a more expressive XML).

map-data/src/main/java/org/triplea/map/reader/XmlParser.java
```diff
@@ -0,0 +1,154 @@
...
+              final String text = streamReader.getText();
+              log.info("Characters Event reached: " + text);
+              bodyHandler.accept(text);
+            }
+            break;
+          case XMLStreamReader.END_ELEMENT:
+            final String endTagName = streamReader.getLocalName();
+            log.info("end tag reached: " + endTagName);
+            if (endTagName.equals(currentTag)) {
+              endTagReached = true;
```
RoiEXLab, line 147: Nitpick: In this case I'd actually go for a return statement or use labels to break the while loop

map-data/src/main/java/org/triplea/map/reader/XmlParser.java
```diff
@@ -0,0 +1,154 @@
...
+      boolean endTagReached = false;
+
+      log.info("Processing current tag: " + currentTag);
+
+      if (streamReader.getEventType() == XMLStreamReader.START_ELEMENT) {
+        for (int i = 0, n = streamReader.getAttributeCount(); i < n; i++) {
+          final int index = i;
+
+          final String attributeName = streamReader.getAttributeLocalName(index);
+          final Consumer<String> attributeHandler = attributeHandlers.get(attributeName);
```
RoiEXLab, line 112: Design consideration:
Currently the attributes are all processed separately, however I think it would be better to aggregate all of the attributes at this point into a hashmap and pass this map to the handler.
This way it becomes easier to check if all of the semantically important attributes are present (even though I have to admit I don't know how much work the dtd does here for us)

RoiEXLab (commented):
I like the direction where this is going, the hierarchical approach adds some abstraction to digest, but once the scheme is clear it makes things very readable.
I'm not entirely convinced by the way attributes are handled (I think a simple HashMap would be sufficient for parsing), but at this point I have no idea how you're planning to map the parser objects into actual game logic objects, so maybe this approach is far better.

map-data/src/main/java/org/triplea/map/reader/XmlParser.java
DanVanAtta: Side-note, it's really interesting that it looks like body elements are never used, it's all attributes..

So, sketching out using a map to return the attributes, it would be something like this:

```
        XmlParser.tag(TAG_NAME)
            .addAttributeHandler(attributes -> {
              unitType = attributeMap.get("unitType");
              territory = attributeMap.get("territory");
            })
            .parse(streamReader);
```

vs: 

```
        XmlParser.tag(TAG_NAME)
            .addAttributeHandler("unitType", value -> unitType = value)
            .addAttributeHandler("territory", value -> territory = value)
            .parse(streamReader);
```


The prevailing responsibility of the `map.data.elements` classes is to capture exactly what is in the XML and to just be java objects representing the XML data. The 'dtd' today introduces defaults, it's a question whether these POJOs should add those default values in or not.

Eventually these objects will be passed up a layer to `game-core` and assembled into a GameData. This will probably take place in `GameParser` where instead of parsing the raw XML it'll invoke the `map-data` parser and get back a `Game` object which it can then traverse and assemble into a game data.

A next step after that will be to store the `Game` tag data as a `MapData` object in `GameData`, make that immutable and not part of the save game. This offloads a lot of data from `GameData` and we would persist much less. Then from there the idea is a `GameData` would lose more information and instead be a `MapData` and a list of change objects. If you then replay all of the change objects onto the map data you would then get the latest game state. In this way we only need to persist which map generated the original `MapData` and then persist the series of change objects (so no more java serialization).

Which comes back to, should this layer be responsible for adding in default values (in essence do the work of the DTD which adds defaults)?  WDYT @RoiEXLab ? I'm honestly not sure at this point, I can see it argued both ways where maybe we're not quite exactly modeling the data we get from the map and are abstracting the detail of default values, on the other hand there is merit to exactly modeling the XML. I think when we get to a point where `MapData` is not persisted with save games it will be more important. But even then, the translation of a `MapData` to a `GameState`, AKA `GameData` will still be done; all of question of whether it should be done in game-core or in map-data. The other consideration I suppose is when we write map data. Would we want the map data to add in defaults that game-core does not request to be written?

Also of note though, the SAX parser does not look at the DTD at all. The API for parsing here is super-low level. I saw some benchmarks showing it's actually slower than JAXB, but uses the least amount of memory. Since there is no DOM model created, we just pass through the XML with a cursor, that makes sense it would be low memory. On the plus side that should hopefully really help with our current problems where we do see memory issues loading maps. With a cleaner model though perhaps we'll be able to get faster parser, but not guaranteed and really not the goal. 

Something else to consider is a migration path, where we'll probably be allowing a large "OR" of conditions where we would read data. Let's say for example:
```
<territory name="Poland" />
```

And:
```
<territory>Poland</territory>
```

In theory a person could also write:
```
<territory name="Poland">Poland</territory>
```

The latter is going to be non-deterministic which would win, but the point is the parser is looking at multiple locations for the same data. This is a bit more profound perhaps if we start nesting tags in different ways. In that latter case it would be the job of the GameData assembler to choose the right data if presented with multiple options. For example, something like:
```
<relationships>
  <relationship>
         <type name="Peace" />
         <playerRelationship p1="Polish" p2="Austrians" />
  </relationship>
</relationships>
```

Today the above is modeled as two top level tags:
```
<relationships>
   <relationship type="Peace" player1="Polish" player2="Austrians" />
</relationships>

<relationshipTypes>
   <relationshipType name="Peace" />
</relationshipTypes>
```

Hence, the map-data parser is going to have to look at both places and would return both if both are present (again leaving it to the game-data assembler to decide which one to prefer if any). This does make me think it's more cohesive if the map data reader really does no more than model the existing XML and allow the assembler to then make any further decisions about defaulting missing values.

I would also emphasize that the map-data-reader does not have any semantic knowledge of the data it reads. For example the property keys we read are important, to the map-data-reader it's just a set of property tags and it does not care what's in them. Similar for any special encoded values where we have for example colon delimited player names. I think to this end it's probably more reason for defaulting of missing values to happen in the assembler. Of course it'll then be up to the assembler to do any semantic validation (for example if an attribute value is a wrong value).

DanVanAtta:
@RoiEXLab I went with your suggestion to return a simple map rather than have a handler:  f88b91a 

I've been thinking that code for mapping attributes and tags became really boilerplate and is something that annotations should be able to drive. At the risk of re-inventing DOM-less Jaxb, I wired in some annotations! :smiling_imp: 

All of the XMLParser handler code is behind the scenes and all of the data classes are simply marked up with three tags:
```
@Attribute
@Tag
@TagList(Type.class)
```

So now this PR has two new sub-projects:
- (A) map-data: mostly POJOs with annotations on them
- (B) xml-reader: generic reader, hides the parsing details and has annotations and the mapper class which is responsible for reading class metadata via relfection and then setting up the child tag and attribute handlers.

DanVanAtta:


```diff

Issues
======
- Added 3
           

Complexity increasing per file
==============================
- map-data/src/main/java/org/triplea/map/reader/MapReadResult.java  3
- map-data/src/test/java/org/triplea/map/reader/InfoReadingTest.java  1
- map-data/src/test/java/org/triplea/map/reader/GamePlayReadingTest.java  1
- map-data/src/test/java/org/triplea/map/reader/TripleaTest.java  1
- map-data/src/test/java/org/triplea/map/reader/XmlReaderTestUtils.java  3
- xml-reader/src/test/java/org/triplea/generic/xml/reader/XmlMapperTest.java  1
- map-data/src/test/java/org/triplea/map/reader/AttachmentListReadingTest.java  1
- xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlParser.java  12
- xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlMapper.java  9
- map-data/src/test/java/org/triplea/map/reader/InitializeReadingTest.java  1
- map-data/src/main/java/org/triplea/map/reader/MapElementReader.java  2
- map-data/src/test/java/org/triplea/map/reader/DiceSidesReadingTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6102557&bid=19976423)

map-data/src/test/java/org/triplea/map/reader/XmlReaderTestUtils.java
```diff
@@ -0,0 +1,30 @@
...
+
+  static Game parseMapXml(final String xmlFileName) {
+    try (InputStream stream = openFile(xmlFileName)) {
+      final MapReadResult readResult = MapElementReader.readXml(xmlFileName, stream);
+
+      assertThat(readResult.getErrorMessages(), is(empty()));
+      assertThat(readResult.getGame(), isPresent());
+      return readResult.getGame().orElseThrow();
+    } catch (final IOException e) {
+      throw new RuntimeException("Failed to parse (expecting a valid XML file): " + xmlFileName, e);
```
DanVanAtta, line 27: Codacy found an issue: [Avoid throwing raw exception types.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6102557)

xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlMapper.java
```diff
@@ -0,0 +1,70 @@
...
+      if (hasNestedTags) {
+        tagParser.parse(xmlStreamReader);
+      }
+
+      return instance;
+    } catch (final IllegalAccessException
+        | NoSuchMethodException
+        | InstantiationException
+        | InvocationTargetException e) {
+      throw new RuntimeException("Class: " + pojo.getCanonicalName(), e);
```
DanVanAtta, line 67: Codacy found an issue: [Avoid throwing raw exception types.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6102557)

xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlParser.java
```diff
@@ -0,0 +1,111 @@
...
+      boolean endTagReached = false;
+
+      if (log.isLoggable(Level.FINE)) {
+        log.fine("Processing tag: " + currentTag);
+      }
+
+      while (streamReader.hasNext() && !endTagReached) {
+        log.info("event type: " + streamReader.getEventType());
+        final int event = streamReader.next();
+        switch (event) {
```
DanVanAtta, line 80: Codacy found an issue: [Switch statements should have a default label](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6102557)

DanVanAtta:
For some more context on how the annotations work, the field name drives the expected tag name. The matching is case-insenstiive. There is a generic XML sample in the xml-reader subproject and a unit test that goes through how the annotations should be set up.

I still need to be able to handle body content. I think maybe the only place we do anything with body content is for game notes.. So it needs to be handled. That will likely land in a follow-up PR. I'd like to iterate a bit more on the design, though this might be there, clean this up and likely submit for actual PR so an initial iteration can land. Then next steps will be to fill out the XML data model objects, add support for body content, and then move on to game data assembly and remove the old XML parser. At that point things will be really interesting.

DanVanAtta:
Also, for more context, it turned out the attribute handler was able to go away completely. When a class is instantiated the stream reader is at a location it can read attributes directly. So we can simply do:
```
myVariable = xmlStreamReader.getAttribute(null, "attributeName")
```

That then removed the attribute handling code from the XML Parser. It only then needs to worry about tag handling. The code that remained in each class then was moved to XmlMapper which did all that in a generic way. So the XmlMapper starts by mapping attribute values, then it sets up any child tag handling.

DanVanAtta:


```diff

Issues
======
- Added 3
           

Complexity increasing per file
==============================
- map-data/src/test/java/org/triplea/map/reader/InfoReadingTest.java  1
- map-data/src/main/java/org/triplea/map/reader/MapElementReader.java  2
- map-data/src/test/java/org/triplea/map/reader/DiceSidesReadingTest.java  1
- map-data/src/test/java/org/triplea/map/reader/XmlReaderTestUtils.java  3
- map-data/src/test/java/org/triplea/map/reader/AttachmentListReadingTest.java  1
- xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlParser.java  11
- xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlMapper.java  9
- map-data/src/test/java/org/triplea/map/reader/InitializeReadingTest.java  1
- map-data/src/test/java/org/triplea/map/reader/TripleaTest.java  1
- map-data/src/test/java/org/triplea/map/reader/GamePlayReadingTest.java  1
- xml-reader/src/test/java/org/triplea/generic/xml/reader/XmlMapperTest.java  1
- map-data/src/main/java/org/triplea/map/reader/MapReadResult.java  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6102557&bid=19976423)

xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlMapper.java
DanVanAtta, line 67: Codacy found an issue: [Avoid throwing raw exception types.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6102557)

map-data/src/test/java/org/triplea/map/reader/XmlReaderTestUtils.java
DanVanAtta, line 27: Codacy found an issue: [Avoid throwing raw exception types.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6102557)

xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlParser.java
```diff
@@ -0,0 +1,109 @@
...
+
+    public void parse(final XMLStreamReader streamReader) throws XMLStreamException {
+      if (log.isLoggable(Level.FINE)) {
+        log.fine("Processing tag: " + currentTag);
+      }
+
+      while (streamReader.hasNext()) {
+        log.info("event type: " + streamReader.getEventType());
+        final int event = streamReader.next();
+        switch (event) {
```
DanVanAtta, line 78: Codacy found an issue: [Switch statements should have a default label](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6102557)

DanVanAtta:


```diff

Issues
======
- Added 3
           

Complexity increasing per file
==============================
- xml-reader/src/test/java/org/triplea/generic/xml/reader/XmlMapperTest.java  1
- map-data/src/test/java/org/triplea/map/reader/DiceSidesReadingTest.java  1
- map-data/src/test/java/org/triplea/map/reader/TripleaTest.java  1
- map-data/src/main/java/org/triplea/map/reader/MapReadResult.java  3
- xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlMapper.java  9
- map-data/src/main/java/org/triplea/map/reader/MapElementReader.java  2
- map-data/src/test/java/org/triplea/map/reader/GamePlayReadingTest.java  1
- map-data/src/test/java/org/triplea/map/reader/InfoReadingTest.java  1
- map-data/src/test/java/org/triplea/map/reader/XmlReaderTestUtils.java  3
- xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlParser.java  11
- map-data/src/test/java/org/triplea/map/reader/AttachmentListReadingTest.java  1
- map-data/src/test/java/org/triplea/map/reader/InitializeReadingTest.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6102557&bid=19976423)

xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlMapper.java
```diff
@@ -0,0 +1,70 @@
...
```
DanVanAtta, line 67: Codacy found an issue: [Avoid throwing raw exception types.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6102557)

xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlParser.java
```diff
@@ -0,0 +1,109 @@
...
```
DanVanAtta, line 78: Codacy found an issue: [Switch statements should have a default label](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6102557)

map-data/src/test/java/org/triplea/map/reader/XmlReaderTestUtils.java
DanVanAtta, line 27: Codacy found an issue: [Avoid throwing raw exception types.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6102557)

xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlMapper.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+        if (field.getAnnotation(Tag.class) != null) {
+          field.setAccessible(true);
+          tagParser.childTagHandler(
+              field.getName(), () -> field.set(instance, mapXmlToClass(field.getType())));
+          hasNestedTags = true;
+        } else if (field.getAnnotation(TagList.class) != null) {
+          field.setAccessible(true);
+          if (field.getType() == List.class) {
+            final List tagList = new ArrayList();
```
RoiEXLab, line 46: Ahhh raw types 😰 
```suggestion
            final List<Object> tagList = new ArrayList<>();
```

Another potential approach would be to make the tagParser thingy part of a dedicated method that uses a generic parameter something like this, which makes it clear that the list and the field are supposed to be the same generic type, but doesn't make the code more readable IMO.

```java
@SuppressWarnings("unchecked")
<T> List<?> populateList() throws Exception {
  final List<T> tagList = new ArrayList<>();
  final Class<T> listType = (Class<T>) field.getAnnotation(TagList.class).value();
  tagParser.childTagHandler(listType.getSimpleName(), () -> tagList.add(mapXmlToClass(listType)));
  return tagList;
}
```

xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlParser.java
```diff
@@ -0,0 +1,109 @@
...
+  @AllArgsConstructor(access = AccessLevel.PRIVATE)
+  public static class GenericParser {
+    private final String currentTag;
+    private final Map<String, ThrowingRunnable> childTagHandlers;
+    private Consumer<String> bodyHandler;
+
+    public void parse(final XMLStreamReader streamReader) throws XMLStreamException {
+      if (log.isLoggable(Level.FINE)) {
+        log.fine("Processing tag: " + currentTag);
+      }
```
RoiEXLab, line 73: Does this if statement make a difference?
AFAIK log.fine is no-op when this level is not loggable

xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlParser.java
```diff
@@ -0,0 +1,109 @@
...
+import org.triplea.java.function.ThrowingRunnable;
+
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+@Log
+class XmlParser {
+  @Nonnull private final String tagName;
+
+  public static XmlParser tag(final String tagName) {
+    return new XmlParser(tagName.toUpperCase());
+  }
```
RoiEXLab, line 23: I don't think this method is used anymore?

RoiEXLab (commented):
Ok, didn't expect to see an annotation-driven approach here, but it's super clean for modeling 👍

xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlParser.java
DanVanAtta: The if statement is to avoid evaluation of arguments.

Say our logging method is:
```
void fine(String statementToLog) {
   if(log.isLoggable(this.currentLogLevel()) {
          // do some logging
   }
}
```

If you invoke that with:
```
log.fine("debug statement: " + expensiveToString.toString());
```

Then in the above you'll evaluate: `"debug statement: " + expensiveToString.toString()` to then pass that to the logging statement which then on-ops. So by wrapping the statement with an if you are avoiding the parameter evaluation. When a method is called, first the parameters are evaluated and then those are passed to the method.

Since we have the concatenation of two strings, we are not saving much at all. Our default log level is only info, so this is further not super useful to have..

xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlParser.java
```diff
@@ -0,0 +1,109 @@
...
+    private final String currentTag;
+    private final Map<String, ThrowingRunnable> childTagHandlers;
+    private Consumer<String> bodyHandler;
+
+    public void parse(final XMLStreamReader streamReader) throws XMLStreamException {
+      if (log.isLoggable(Level.FINE)) {
+        log.fine("Processing tag: " + currentTag);
+      }
+
+      while (streamReader.hasNext()) {
```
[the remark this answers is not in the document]
DanVanAtta: It might be because the `next()` is not always clear, there are additional "next" methods available. Overall the API is pretty damn low level. I kinda wish that XmlStreamReader implemented auto-closeable..

xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlParser.java
DanVanAtta: No, artifact of this still being rough and a draft. Now that XmlParser is effectively an internal class, I simplified it a good bit. Having the fluent-API is no longer that useful since it's only used from the one class now.

The reasoning for the fluent-API is to avoid adding tag handles and body handlers together. If an XML has nested tags, then it should not have body content and vice versa. Adding both kinds of handlers is generally an error and hence the reason for an API that avoided that configuration.

xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlMapper.java
DanVanAtta: I'd wonder if the class cast might cause issues based on the XML data and make error messaging difficult at all. To another extent, raw typing is going to be somewhat unavoidable. Not a bad suggestion though. I'm about ready to do some polishing of the XML-reader module and will see if I can incorporate this.

DanVanAtta:
@RoiEXLab pushed up an update that mostly completes the effort  this first effort  : )
I think the design iterations landed in a good place. I wonder a lot whether we should re-use JAXB annotations which could make it a complete two way street to write the XML back. Otherwise we may see we have both our annotations and JAXB on the data as well. I am planning to add JAXB for the writing side of this. 

For parsing having our own custom parser seems beneficial though:
- STAX does not create a DOM, so this is a lower memory usage approach over JAXB
- We gain a lot of flexibility. I'm thinking adding a "names" attribute to the `@Attribute` and `@Tag` annotations coudl be powerful, for example we could add "attatchment" as an alternative spelling and suddenly start supporting the 1.8 maps again.
- The parsing code *is* pretty minimal,it's under 200 lines in total, surprisingly minimal.

DanVanAtta:
"re-use JAXB annotations", to clarify, rather than looking for our "@Tag" annotation in reflection, we'd look for Jaxb's "@XmlElement" instead. It might be a bit confusing though as it would look like we're using Jaxb for both writing *and* reading XML. The intent is to use STAX for reading, and Jaxb for writing.

xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlParser.java
ssoloff: I believe you can use the overload [`Logger#fine(Supplier<String>)`](https://docs.oracle.com/en/java/javase/11/docs/api/java.logging/java/util/logging/Logger.html#fine(java.util.function.Supplier)) to avoid building the message when the current log level is above `FINE`.

xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlMapper.java
RoiEXLab: I mean the first example should work too and gets rid of the compiler warning

xml-reader/src/main/java/org/triplea/generic/xml/reader/XmlParser.java
DanVanAtta: Good to know that's been built-in now @ssoloff, having to wrap log statements with if conditions always seems kinda gross.

DanVanAtta:
Some interesting performance results. The STAX parser with the code presented here parsed war of relics in about 9s. For comparison game parser in game-core is about 7s. I removed all of the logger statements and the StAX based parser is a blazing 900ms.

DanVanAtta:
Having gotten a bit further, the game data assembly is looking to be the long pull of game parsing, not the XML reading and DOM creation. In part that is good news meaning performance improvements are within our control. I also noticed that even with the StAX reader we still see a stack-overflow on the WoW map without XSS set to a higher level. (I kinda wonder if it is due to nested variables, but hard to say, but importantly it's not the XML DOM creation that is doing that but instead it is our post-processing).

DanVanAtta:
Got this update submitted as carved off pieces and https://github.com/triplea-game/triplea/pull/7558 would be the last one. 

The early feedback was very appreciated, thank you

PR #7518, opened 2020-09-02
Remove unused XML tag from DTD: 'remove-connection'

DanVanAtta:
It looks like the tag is only defined in game.dtd
- Could not find any code to handle the tag
- Could not find any maps using the tag

REMOVE|Remove unused and unsupported game XML tag 'removeConnection'


RoiEXLab (approved):

RoiEXLab:
The DTD is always up to a surprise ^^

PR #7552, opened 2020-09-05
Reduce state in GameChooser & Apply Strangler to Shallow Parsing

DanVanAtta:
1. Stop keeping tracking of a fully parsed vs shallow parsed game data in game chooser.
  Instead we'll return a fully parsed game data when requested and otherwise
  we keep track of the shallow parsed data. It turns out the shallow parsed data is
  just the game name.

2. Extract shallow parsing to its own module so that later we can swap out
  the implementation (strangler pattern). As part of this we can simplify the API
  of shallow parsing since we only need the game data name.

3. Reduce severity level for invalid maps from 'SEVERE' to 'WARN'. Bad map
  data we do want reported to users but they are not actionable bug reports
  for us generally (users who are working on new maps would expect to see problems).

This update changes error handling behavior such that if a map has a parse error we do
not deselect it from the game chooser. This is a bit forward looking as eventually
a shallow parse will not reveal any errors anyways. It would make sense to get
a deep parse only when selecting a game map and seeing errors then.

This update does reduce game performance in the map selection as we do a game parse
whenever clicking any new game selection items. We'll need to revisit this to
see if and how that can and/or should be improved later.


DanVanAtta:


```diff

Complexity increasing per file
==============================
- game-core/src/main/java/games/strategy/engine/data/gameparser/ShallowGameParser.java  1
- game-core/src/main/java/games/strategy/engine/framework/ui/GameChooser.java  1
         

Complexity decreasing per file
==============================
+ game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6131815&bid=20053945)

RoiEXLab (commented):
I must admit that I don't quite understand the new code completely, I assume that's mostly because the "old" code wasn't that clear to understand either.
It's a good thing that the "shallow" and "deep" parsing differences are reduced, but I'm not enterely sure where state is kept and where the code follows a more reactive approach

game-core/src/main/java/games/strategy/engine/framework/ui/DefaultGameChooserEntry.java
```diff
@@ -4,126 +4,67 @@
...
 
 @Log
-final class DefaultGameChooserEntry implements GameChooserEntry {
-  private final URI url;
-  @Nonnull private GameData gameData;
-  private boolean gameDataFullyLoaded = false;
-  private final String gameNameAndMapNameProperty;
+@EqualsAndHashCode(of = "uri")
+@Getter
+public class DefaultGameChooserEntry implements Comparable<DefaultGameChooserEntry> {
```
RoiEXLab, line 21: Also a consideration:
Now that `GameChooserEntry` is no longer a thing this class could get renamed

RoiEXLab:
Also a question @DanVanAtta is the "Post Merge Review" Label supposed to be removed whenever a post merge review is created, or should it stay until the last few points are resolved?

game-core/src/main/java/games/strategy/engine/framework/ui/DefaultGameChooserEntry.java
DanVanAtta: Indeed. I've been waiting on the rename for a moment for a few reasons:
- renames are really tough on review as the file often will look new. File renaming and package moves is perhaps the biggest downside of doing reviews, reviews do discourage those operations because the diff gets so messed up
- a rename done on its own would conflict with work that is continuing on from this point. A post-merge really indicates that I needed to incorporate these updates to keep going in a clean manner. The merge conflicts have been working out pretty okay so far, but I have been encountering some even with a pretty quick cadence of merges. So this is to say that there is already future work that would make a rename not nice as it would conflict badly with that, so it's a matter of finding the time to fit in
- hasn't necessarily been that important per-say. The name 'default game chooser entry' is not too horrible necessarily.
- not clear if this class has a long term future or would be refactored away to another form or just reduced down to a map. It is now only a tuple class that holds a string and a URI, really TBD if it'll survive

DanVanAtta:
@RoiEXLab I think 'post merge review' should mean that there has not been a look at it at all or more review needs to be done to complete the review. Perhaps if we think there needs to be follow-up, then another label should be used. We probably should have a meta conversation finally about what items should get follow up and how to agree on that. Generally we've worked it out pre-merge, but I would emphasize we need to be strategic with our time, priorities, and respect the scope of an intended change. I also wonder if for doing lots of 'post-merge' items, if some things we are just as well doing ourselves as a reviewer and then posting it for PR. The latter has some appeal as that could move the ball the most quickly and avoids round-trips.

DanVanAtta:
Let's more the meta conversation to: https://github.com/triplea-game/triplea/issues/7625 @RoiEXLab ; it is best to keep PRs conversations about the code that changed.

PR #7576, opened 2020-09-08
Convert standard battle steps

trevan:
This converts the standard battle steps.

## Testing

Loaded old save games and played a few battles to make sure the right units were rolling and dying.


game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1553,15 +1547,15 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
       steps.add(offensiveFirstStrike);
     }
     // Attacker fire remaining units
-    steps.add(
-        new StandardAttackersFire() {
-          private static final long serialVersionUID = 99994L;
+    steps.add(offensiveStandard);
+    new IExecutable() {
+      private static final long serialVersionUID = 99994L;
```
DanVanAtta, line 1552: nit, while the serialVersionUID do not need to be unique across classes, this might throw a few programmers and typically they are UUID's.

I wonder if we're at the point for creating a helper to create an `IExecutable for us? It really is just a `BiConsumer` interface, eg:

```
IExecutable {


   static IExecutable newExecutable(BiConsumer<ExecutionStack, DelegateBridge> action) {

      return new IExecutable() {

          @Override
            action.accept(stack, bridge);
          }
      }
   }
}

```


That would make the current code nicer. Though, I'm starting to think our biggest gains is by removing the `IDelegateBridge` from the interface if possible. IMO it's important that we are making such a future decoupling easier rather than more difficult.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1576,15 +1570,15 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
       }
     };
 
-    steps.add(
-        new IExecutable() {
-          private static final long serialVersionUID = 1560702114917865290L;
+    steps.add(defensiveStandard);
```
DanVanAtta, line 1573: nit, can we use a different word than `standard`? I don't know it's well enough defined to have a lot of meaning. `DefensiveRoll` would perhaps be just as meaningful. `standard` raises the question, what is non-standard?

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1687,32 +1681,6 @@ private void checkForUnitsThatCanRollLeft(final IDelegateBridge bridge, final bo
     }
   }
 
-  private void standardAttackersFire() {
-    findTargetGroupsAndFire(
```
DanVanAtta, line 1691: Have you considered inlining these methods 'standardAttackersFire' and 'standardDefendersFire'? WDYT about that idea? Otherwise, can you clarify the motivation or benefit for having them?

As-is, from a casual perspective, we have a one-line method (the IExecutable) that uses none of the passed in args, which invokes a single use method that is itself a one-liner. Usually such patterns are candidates for inlining to reduce indirection. Perhaps I'm missing something (?)  (an answer to the above questions might point out what that is)

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
```diff
@@ -185,20 +189,14 @@ public boolean isAmphibious() {
       steps.addAll(offensiveFirstStrike.getNames());
     }
     steps.addAll(airAttackVsNonSubs.getNames());
+    steps.addAll(offensiveStandard.getNames());
 
-    if (attackingUnits.stream().anyMatch(Matches.unitIsFirstStrike().negate())) {
-      steps.add(attacker.getName() + FIRE);
```
DanVanAtta, line 190: Side-note, it seems like this class knows quite a lot:
- how to form step names (EG:  `attacker.getName() + FIRE`
- how to extract step names (EG: `steps.addAll(offensiveFirstStrike.getNames());`)
- the ordering of steps
- if to offer certain steps (evaluates retreat rules, eg: line 215 (if (RetreatChecks.canAttackerRetreat(...)

Any thoughts about this long term and if we're heading in a direction to fix this?

DanVanAtta, line 190: Some elements have multiple steps, eg: ` steps.addAll(submergeSubsVsOnlyAir.getNames());`, can we use the same pattern here for `attacker`? EG: `attacker.getNames()`? This way we can avoid forming the 'fire & select casualties' here.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/standard/DefensiveStandardTest.java
```diff
@@ -0,0 +1,89 @@
...
+      final DefensiveStandard defensiveStandard =
+          new DefensiveStandard(
+              givenBattleStateBuilder().defendingAa(List.of()).gameData(gameData).build(),
+              battleActions);
+
+      defensiveStandard.execute(executionStack, delegateBridge);
+
+      verify(battleActions)
+          .findTargetGroupsAndFire(
+              eq(MustFightBattle.ReturnFire.ALL),
```
DanVanAtta, line 78: Any thoughts if we can have something return this value to us?
Perhaps we can create a module that calculates the next step and returns that as a value. IN such a case if we inject that we can then mock and control it, in which case the `findTarggsGroupsAndFire` parameters is not terribly interesting since it mostly comes from mocks, but ultimately testable. In turn, that module can be tested directly with a very functional style and few mocks and particularly no checking that a void method call was called with the right arguments. Thoughts?

DanVanAtta (commented):
Took a first review pass @trevan , perhaps about 50% reviewed. Left some comments, interested what your thoughts might be and a bit curious what you're planning for future updates as well.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -108,16 +110,6 @@
     private static final long serialVersionUID = 1369227461759133105L;
   }
 
-  /**
-   * An action representing standard attacking fire.
-   *
-   * <p>NOTE: This type exists solely for tests to interrogate the execution stack looking for an
```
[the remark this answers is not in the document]
trevan: This class has been deleted.  The tests that depend on the class types are slowly going away.  They are being replaced with the `Order` enums.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
trevan: Once all of the steps are converted, it will just create the steps, then order them, and then map them to their names.

See `MustFightBattle#addFightStartSteps` at https://github.com/triplea-game/triplea/blob/1cd72b85200ec7e9de2890aad83a034332b1ffb4/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java#L1051 for a rough example of what it would look like.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: I'm confused.  I did inline this in the OffensiveStandard and DefensiveStandard class.  It looks like you might have gotten the deleted lines and the added lines backwards.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: All of the other steps refer to their "type": FirstStrike, AA, NavalBombardment, etc.  I agree that `standard` is probably not the best, but I don't think `roll` is even better.  I could do `units` but then it might be confusing since this isn't all of the defensive units, only the ones that haven't fired yet.  I've seen the code also refer to this group as `regular`.  I'm not sure what to call them.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: I can't change this serial number.  It will break save games.

All of the apis that require the `IDelegateBridge` are being added to the `BattleActions` interface.  Once all of the steps have been converted, I plan to go through the apis and shrink them as much as possible.  I'd like to get to the point when all the api does is call something on `IDelegateBridge`.  At that point, we can look into moving `BattleActions` from `MustFightBattle` to `IDelegateBridge` or changing the `IDelegateBridge` api itself.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/standard/DefensiveStandardTest.java
trevan: I'm not sure what you are asking.

trevan:
@DanVanAtta , I've replied to most of your comments.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/standard/DefensiveStandardTest.java
DanVanAtta: I'll be direct:
- mocks are testing code smell
- verifying calls to void methods on mocks is an even bigger smell (you're verifying that a side-effect has occurred in very coupled manner).

To simplify and illustrate what we have:
```
StrategyObject mockStrategy = mock(StrategyObject.class);
ClassUnderTest sut = new ClassUnderTest(mockStrategy);

sut.voidCall("some arg");

verify(mockStrategy).sideEffect(eq("some computation"));
```

The above I'd go as far to say is a testing anti-pattern:
- highly coupled to implementation. Change the prod code, even if you just change API you use, you break the test. Not because the feature is broken, but because the implementation is changed.
- it bakes in the void method calls, which are side-effects into the code deeper and tests that we have side-effects invoked
- if you ever mock an object, one must in turn that mocked object with all of the expected parameters to assert it actually returns the stubbed responses. Otherwise who is to say that the actual object behaves that way, in which case the test is not testing anything. For example, if I mock a DB layer to return 5 results from a select, I need to take the real DB and assert that given that select I get 5 results. Otherwise if the real DB throws a syntax error or otherwise, and all I have done is mocked it, then I'll have a bug that the test papered-over and was not sufficient enough to test (so either use a real object or create another test to verify the behavior that was mocked in the orignal test).

To fix the above, one can often transform the object under test to be a function and then have the caller pass the result to the void method call. So in production code, our original example is this:

```
new ClassUnderTest(new StrategyObject()).voidCall("some arg");
```

Instead you can transform that into:

```
var result = ClassUnderTest().computeSomething("some arg");
new StrategyObject().sideEffect(result);
```

Now, you can test `ClassUnderTest` without mocks and you can still test `StrategyObject` just as you would have before.

If things are *really* coupled then a half step that helps is:

```

ClassUnderTest(new ArgTransformFunction(), new StrategyObject()  ).voidCall("some arg");

// where ArgTransformFunction = "some arg" -> "some computation"
``` 

With that latter the test for ClassUnderTest is completely with mocks. Which is good as all the variation is under your control and you can still test "new ArgTransformFunction()" and "new StrategyObject()" independently. In these cases you are roughly able to remove the verify invocations.

Now, how difficult it would be to do such a transformation is a question. The arg conversion strategy can often be injected as a '@VisibleForTest' constructor and this way you move the interesting code under test to its own unit that you can test in isolation without mocks.

---------------


Beyond the above, something I was thinking about, I'd recommend building some harness or characterization tests. I suspect tests that would help most in the restructuring would have setups like "given a sub, and a v2 transport", assert that the executable stack has "first-strike attack and defender return fire".

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
DanVanAtta: hah, true enough : ) 

Sorry about that.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
DanVanAtta: Perhaps simply then: 'defensive' without the qualifier?

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
DanVanAtta: Oh.. understood on the serial version ID.

I'm thinking IDelegateBridge probably should not be related to `MustFightBattle` at all and the callers should do the coordination. between the two.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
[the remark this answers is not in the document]
DanVanAtta: Gah, my mistake again, the 'night mode' in my browser made the deleted lines look like additions. I see now this is removed.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/standard/DefensiveStandardTest.java
trevan: Understood.  I'll work on this once I've got all the battle steps converted.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: Yes, `IDelegateBridge` shouldn't be related to `MustFightBattle`.  I'm saying that the interface `BattleActions` might fit better on `IDelegateBridge` once all of the apis have been cleaned up.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: The class names would be just "Defensive"?  That feels a little weird to me.

I went through a few rule books and I noticed that "general" was used a few times.  So maybe defensiveGeneral/offensiveGeneral?  Or "units" since the rule books would say "attacking units", even though it had mentioned subs, etc beforehand.

DanVanAtta: Interesting thoughts, as another suggestion:  `DefensiveUnitsReturnFire`?

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/standard/DefensiveStandardTest.java
DanVanAtta: Updating the unit test to be nicer, I'm happy letting things evolve first. In the meantime, is there a reasonable way in which this test could fail? What type of programming errors are possible for that to happen? Have you considered deleting this test?

My 2 cents, such characterization tests are probably going to keep these updates much safer for you. You'll have a lot more confidence that you have not introduced any bugs and because you'll have test coverage you won't have to add unit tests at every step of the way. It's always risky doing the latter approach as it's hard to know if you are unit testing buggy code and thereby asserting defective behavior.

DanVanAtta: Phrased another way, what exactly are we testing here? Is this a valuable test? Will it be deleted the first time we try to any kind of refactoring and is a failure more likely to indicate that the code changed rather than is wrong?

DanVanAtta: For example, there is another test with this:
```
      verify(battleActions)
          .findTargetGroupsAndFire(
              eq(MustFightBattle.ReturnFire.ALL),
              any(),
              eq(false),
              any(),
              any(),
              anyCollection(),
              anyCollection(),
              anyCollection(),
              anyCollection());
    }
```

If both of these fail, one would be left wondering "why was this a valid assertion in the first place?" Why was this test correct, and what was it testing? DisplayName on test and comments in the assertion and in verifications woud help with that.

Tests are great, but a double edged sword as they can become slow and a point of maintenance. When a test fails, the first questions are: "what was this testing?" and "why was this test valid?". If those questions take a lot of time to determine, and it's just one of a dozen tests to fix, then there is substantial cost to change code. Hence, it's very important that tests are overly verbose in what they are testing and why we expect them to be correct and pass. This way someone trying to fix the test can easily know if the test is still valid and what to look for if the test fails

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/standard/OffensiveStandardTest.java
```diff
@@ -0,0 +1,74 @@
...
+      final OffensiveStandard offensiveStandard = new OffensiveStandard(battleState, battleActions);
+      assertThat(offensiveStandard.getNames(), hasSize(2));
+    }
+
+    @Test
+    void hasNoNamesIfStandardUnitIsNotAvailable() {
+      final BattleState battleState =
+          givenBattleStateBuilder().attackingUnits(List.of(givenUnitFirstStrike())).build();
+      final OffensiveStandard offensiveStandard = new OffensiveStandard(battleState, battleActions);
+      assertThat(offensiveStandard.getNames(), hasSize(0));
```
DanVanAtta, line 47: nit, there is a `IsEmptyCollection.empty()` matcher, often used with the is matcher:
```
assertThat(offensiveStandard.getNames(), is(empty());
```

Have you considered using the matcher? If the test fails that matcher will print the contents of the collection rather than reporting simply that the size was not zero.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreatTest.java
```diff
@@ -325,53 +316,4 @@ void retreatDoesNotHappenWhenCanNotSubmergeAndNoRetreatTerritories() {
 
     verify(battleActions, never()).queryRetreat(anyBoolean(), any(), any(), any());
   }
-
-  public static class MockGameData {
```
DanVanAtta, line 329: Good to see a lot of custom mocking removed :+1:

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/standard/DefensiveStandardTest.java
DanVanAtta: @trevan I'm still thinking about this example, I'm not necessarily convinced one way or another if it should remain or go.
The fact there are any matchers for most of the items we are testing is quite good to indicate which values we are really testing. A comment next to the 'false' perhaps could be most of what we need to make it clear why we expect that 'false' value to be there. Maybe ditto with the 'ReturnFire.ALL' enum.

trevan: @DanVanAtta , I've just deleted the tests.  The really important part that needs to be tested is the matcher but I can't figure out how to match that.  I'll be refactoring the findTargetGroupsAndFire at some point in the future and then the tests will be better.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: I've renamed them to be `DefensiveGeneral` and `OffensiveGeneral` and they are in the new `general` package.

DanVanAtta: I think I've a slight preference for the names "OffensiveFire" and "DefensiveFire". The qualified variants of that I think distinguish those, somewhat concerned that 'general' has the same issues as standard.

NBD though.

DanVanAtta (approved):
One more part of `MustFightBattle` down  :+1: 
N to go ;)

PR #7580, opened 2020-09-09
Suppress error-prone java.lang clash warning for XML types

DanVanAtta:
Regrettably we have XML tags named "String" and "Number", we cannot
change the tag name. This update suppresses the error prone warning
to avoid compilation warning messages that we cannot readily fixed.

### DIFF
--- a/map-data/src/main/java/org/triplea/map/data/elements/PropertyList.java
+++ b/map-data/src/main/java/org/triplea/map/data/elements/PropertyList.java
@@ -33,11 +33,14 @@ public class PropertyList {
       @BodyText private java.lang.String data;
     }
 
+    @SuppressWarnings("JavaLangClash")
     public static class Boolean {}
 
+    @SuppressWarnings("JavaLangClash")
     public static class String {}
 
     @Getter
+    @SuppressWarnings("JavaLangClash")
     public static class Number {
       @Attribute private int min;
       @Attribute private int max;


map-data/src/main/java/org/triplea/map/data/elements/PropertyList.java
```diff
@@ -33,11 +33,14 @@
       @BodyText private java.lang.String data;
     }
 
+    @SuppressWarnings("JavaLangClash")
```
RoiEXLab, line 36: Thoughts on introducing a tag to override the class name?

DanVanAtta: AFAIK tags cannot inherit tags. Is there actually a way to do that? If so, having the XML annotation to go ahead and add suppress warnings for 'unused' would be good. The lang clash is very limited, I would not expect for us to see more of it and if we introduce more lang clashes it's something that IMO we should very well fix. These tags being named after lang classes is really not ideal.

RoiEXLab: I was more thinking of a `@CustomName("String")` tag, so the class can be called whatever it wants to but if this annotation is present it will "override" the class name to use the custom string instead

DanVanAtta:
Perhaps allow for aliases in the current tags? We could even then easily
support those old attachment tags :grin:

It's a good idea to use that here

PR #7584, opened 2020-09-09
Update: PerfTimer use throwing functions and unify timing method names to 'time'

DanVanAtta:
Having throwing functions makes using PerfTimer easier, the inner code executed
by the function can throw without needing to be explicitly wrapped with a
new try/catch.

### DIFF
--- a/java-extras/src/main/java/org/triplea/performance/PerfTimer.java
+++ b/java-extras/src/main/java/org/triplea/performance/PerfTimer.java
@@ -47,11 +47,6 @@ public class PerfTimer implements Closeable {
     }
   }
 
-  @SuppressWarnings("unused")
-  public static PerfTimer time(final String title) {
-    return time(title, 1);
-  }
-
   /**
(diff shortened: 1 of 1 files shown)


java-extras/src/main/java/org/triplea/performance/PerfTimer.java
```diff
@@ -61,10 +57,37 @@ public static PerfTimer startTimer(final String title) {
...
+  public static void time(final String title, final ThrowingRunnable<?> functionToTime) {
+    try (PerfTimer timer = time(title)) {
+      try {
+        functionToTime.run();
+      } catch (final Throwable throwable) {
+        throw new IllegalStateException(
+            "Unexpected throwable in timed method: " + throwable.getMessage(), throwable);
+      }
+    }
+  }
```
RoiEXLab, line 89: In order to re-use code I would've done something like this:
```java
time(() -> {
  functionToTime.run();
  return null;
})
```

PR #7585, opened 2020-09-09
Update: Improve map parsing time - more efficient player attachment lookup

DanVanAtta:
This update improves game parsing speed by increasing the efficiency of
attachment processing. This update converts the attachment processing algorithm,
used on full-game-parse, from a runtime of O(p * a), where p is number of players
and 'a' is the number of attachments per player to O(p).

The core of this update is taking advantage that we map attachments by their
name and we can do a O(1) lookup by name. The previous algorithm iterated over
every attachment looking an attachment with a given name.

For full map parsing, here are some example performance numbers before
and after this update:
- War of the Relics: 4751 ms -> 1569.8 ms
- World at war:      3200 ms -> 1102.7 ms
- World of Warcraft: 2783 ms -> 1147.1 ms

## Testing

- manual tested while getting performance numbers

## Release Note

UPDATE|Improved map loading performance (about 3x times faster)

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/attachments/AbstractTriggerAttachment.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/AbstractTriggerAttachment.java
@@ -51,16 +51,11 @@ public abstract class AbstractTriggerAttachment extends AbstractConditionsAttach
    */
   public static CompositeChange triggerSetUsedForThisRound(final GamePlayer player) {
     final CompositeChange change = new CompositeChange();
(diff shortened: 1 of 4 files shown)


Cernelius:
I'll certainly test this once it is merged (sadly, I don't know how to get the exact times, so I just look at a clock for the time the game takes to be available), also with one or more games not available in the repository.

DanVanAtta:
Hopefully will be a noticeable improvement @Cernelius  :smiley:

PR #7596, opened 2020-09-10
Update: convert game parser to use xml-reader and map-data (StAX based parser)

DanVanAtta:
- Changes GameParser over to use xml-reader & map-data
 This conversion allows for more flexibility when parsing as we can more
 easily get the same different from different locations.

- Game parsing no longer requires or uses a DTD
 This gives us flexibility going forward if change tags, an old or new map
 that does not conform to such a standard would not get a DTD failure
 and we can help migrate the compatibility in code. DTD is still
 useful for validation and we should see about building a map validation
 tool that would use it.

- GameParsing becomes simpler, it now assembles a GameData instead of
 parsing XML. The details of XML parsing are removed and things like
 checking for child nodes is removed and we get native java types
 that are already converted to for example integers or booleans.

## Testing
Did some smoke testing to verify various games would launch. It looks like there was incidentally pretty good test coverage as we tend to parse full maps to do other testing (so there is quite a lot of map parsing in tests in addition to the 'validation' kicking off after parsing *and* the compatibility check parsing over a hundred more XMLs).

Performance change was not measured to be significant. Sometimes faster, sometimes slower, looks like performance is more dependent perhaps on disk caching. 

## Release Note

UPDATE|Map XML parser re-written to allow for more flexible XML structure. The XML DTD is no longer checked or used.


game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
```diff
@@ -395,149 +371,118 @@ private void parseResources(final Element root) throws GameParseException {
...
-      final String defaultType = current.getAttribute("defaultType");
-      final boolean isHidden = current.getAttribute("isHidden").equals("true");
-      final GamePlayer newPlayer =
-          new GamePlayer(name, isOptional, canBeDisabled, defaultType, isHidden, data);
-      playerList.addPlayerId(newPlayer);
-    }
-  }
-
-  private void parseAlliances(final Element root) throws GameParseException {
+  private void parsePlayerList(final org.triplea.map.data.elements.PlayerList playerListData) {
```
DanVanAtta, line 394: Codacy found an issue: [Unnecessary use of fully qualified name 'org.triplea.map.data.elements.PlayerList' due to existing import 'org.triplea.map.data.elements.PlayerList'](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6154389)

DanVanAtta:
One map is failing to parse in compatibility check, the error is in the CDATA block of that map. I suspect this might be due to an old JDK version, trying to get a newer JDK installed and it is fighting me on it. I can't reproduce the parse issue locally, I think if we ensure that install4j has a recent version of Java11 that we'll be okay. I'm tempted to remove the map from the download list or even to modify it such that `<br>` tags will be self-enclosed as `<br/>`

DanVanAtta:
Resolved the issue by updating the map that had issues to use `<br/>` instead of `<br>`. Certainly a bit of a band-aid fix, though it does seem like the problem is very likely due to old java version in travis given the lack of local repro.

game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
```diff
@@ -395,149 +371,118 @@ private void parseResources(final Element root) throws GameParseException {
...
+            currentPlayer, enemyPlayer, relationshipTypeList.getDefaultWarRelationship());
       }
     }
   }
 
-  private void parseRelationInitialize(final List<Element> relations) throws GameParseException {
-    if (!relations.isEmpty()) {
+  private void parseRelationInitialize(final Initialize.RelationshipInitialize relations)
+      throws GameParseException {
+    if (!relations.getRelationships().isEmpty()) {
```
RoiEXLab, line 447: I wonder where this check comes from. If the list is empty the for loop shouldn't iterate a single time and getRelationShipTracker doesn't look like it would throw an exception

PR #7602, opened 2020-09-10
Fix: Disable audio if there is no audio device

DanVanAtta:
Addresses: https://github.com/triplea-game/triplea/issues/7546

## Release Note

FIX|Disable sound options instead of showing an error message if there is no sound on installed system.

### DIFF
--- a/game-core/src/main/java/org/triplea/sound/ClipPlayer.java
+++ b/game-core/src/main/java/org/triplea/sound/ClipPlayer.java
@@ -6,7 +6,6 @@ import games.strategy.engine.framework.GameRunner;
 import games.strategy.triplea.ResourceLoader;
 import games.strategy.triplea.settings.ClientSetting;
 import java.io.File;
-import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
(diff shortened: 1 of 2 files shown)


game-core/src/main/java/org/triplea/sound/ClipPlayer.java
```diff
@@ -134,6 +139,18 @@ private ClipPlayer(final ResourceLoader resourceLoader) {
...
 
+  private static AudioDevice createAudio() {
+    try {
+      return FactoryRegistry.systemRegistry().createAudioDevice();
+    } catch (final JavaLayerException e) {
+      log.log(
+          Level.INFO,
+          "Unable to create audio device, is there audio on the system? " + e.getMessage(),
+          e);
+      return null;
```
RoiEXLab, line 150: Not sure if I consider this a satisfying solution but better than a weird exception.

DanVanAtta: Not all systems are going to have audio at all, if that is the case disabling audio features and adding tooltips stating "you don't have audio, so we disabled audio" I think is about the best we can do. The 'info' level will be background and users will typically never see it, the tooltip being displayed thus is quite important.

Can you think of any thing more comprehensive that we could do, or do you have other thoughts on how we could or shoudl handle this? @RoiEXLab 

We had a series of such error reports in a row, so it seems like the user was getting spammed with error messages every time the game tried to play a sound.

RoiEXLab: Nah, I was mostly thinking about the API. I just think it's weird that an exception is thrown to control the flow

DanVanAtta: It's a little weird to see an exception handled and  'null' returned  rather than 'optional' returned too , speaking of API :grin: 

I think I'm starting to come to a pattern where an optional return value means we have either an exception that is handled (no logging needed, it is all already done) and/or a nulllable return value; in those cases getting an optional back has seemed quite clean (so no logging when you get an empty optional back, just assume that it's all been handled already).

PR #7606, opened 2020-09-10
Convert unprotected units step

trevan:
This converts the step that auto-kills undefended transports and "unprotected units" (the original name was "units that can't roll").

In addition to converting the step, I made two functional changes:

1. In the detection for undefended units, it didn't actually check if the enemy units were enemy units.  It just assumed that any unit in the battlesite with attack must be the enemy.  I added an `Matches.enemyUnit`.

2. The logic to auto-kill the undefended transports and "unprotected units" was in two places but used the "Transport Casualties Restricted" property slightly differently.  I made both places use the new step.  This means that now they both use the property the same.   I'm fairly certain this will actually not affect anything.  The second case only happens if the non-infrastructure units are 0 and it looks like the "unprotected units" auto-kill only kills non-infrastructure units (see RemoveUnprotectedUnitsTest#actionForDefenderIfNoRetreatAndUnitsAndCanNotRollAndOtherSideCanButUnitsAreInfra for a test around that).

## Testing

Opened up some save games.  Played a game where the transports should be auto-killed and ensured they were.


game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
```diff
@@ -168,13 +181,7 @@ public boolean isAmphibious() {
...
       steps.addAll(defensiveSubsSubmerge.getNames());
     }
-    // See if there any unescorted transports
-    if (isBattleSiteWater
-        && Properties.getTransportCasualtiesRestricted(gameData)
-        && (attackingUnits.stream().anyMatch(Matches.unitIsTransport())
-            || defendingUnits.stream().anyMatch(Matches.unitIsTransport()))) {
-      steps.add(REMOVE_UNESCORTED_TRANSPORTS);
-    }
+    steps.addAll(removeUndefendedUnits.getNames());
```
DanVanAtta, line 184: Is it inconsistent to have 'removeUndefendedUnits' and 'removeUnprotectedUnits'?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
```diff
@@ -0,0 +1,173 @@
...
+import java.util.function.Predicate;
+import lombok.AllArgsConstructor;
+import org.triplea.java.collections.CollectionUtils;
+
+@AllArgsConstructor
+public class RemoveUnprotectedUnits implements BattleStep {
+
+  private static final long serialVersionUID = 4357860848979564096L;
+
+  protected final BattleState battleState;
```
DanVanAtta, line 29: Do these class variables have to be 'protected', can they be private?

Having child classes modify parent state is a bad form of coupling. It makes program flow very hard to reason about as you have to know the program state to know what will execute and how. Having class variables be uniformly private helps with this and avoiding inheritance.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
```diff
@@ -0,0 +1,173 @@
...
+
+  protected final BattleActions battleActions;
+
+  @Override
+  public List<String> getNames() {
+    final List<String> steps = new ArrayList<>();
+    if (battleState.getBattleSite().isWater()
```
DanVanAtta, line 36: Nit: Have you considered using a ternary here? Avoiding intermediate variables is often a good thing.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
```diff
@@ -0,0 +1,173 @@
...
+    return Order.REMOVE_UNPROTECTED_UNITS;
+  }
+
+  @Override
+  public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
+    checkAndRemoveUnits(bridge, BattleState.Side.DEFENSE);
+    checkAndRemoveUnits(bridge, BattleState.Side.OFFENSE);
+  }
+
+  public void checkAndRemoveUnits(final IDelegateBridge bridge, final BattleState.Side side) {
```
DanVanAtta, line 56: This being a potential no-op if kinda concerning. I'm also wondering exactly what is meant by 'check', makes for a difficult public API.  Will this method ever be invoked if `(Properties.getTransportCasualtiesRestricted(battleState.getGameData()))` is false? Can we perhaps make it so that this step is never added to begin with? In that case you could assert that state to be true here

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
```diff
@@ -0,0 +1,173 @@
...
+  private void checkUndefendedTransports(
+      final IDelegateBridge bridge, final BattleState.Side side) {
+    final GamePlayer player =
+        side == BattleState.Side.OFFENSE ? battleState.getAttacker() : battleState.getDefender();
+    // if we are the attacker, we can retreat instead of dying
+    if (attackerHasRetreat(side)) {
+      return;
+    }
+    final List<Unit> alliedTransports = getAlliedTransports(player);
+    // If no transports, just return
```
DanVanAtta, line 72: nit, obvious comment, mostly restates the code. IMO this one can be removed.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
```diff
@@ -0,0 +1,173 @@
...
+      battleActions.remove(
+          unprotectedUnits, bridge, battleState.getBattleSite(), side == BattleState.Side.DEFENSE);
+    }
+  }
+
+  private boolean areFightingOrSupportingUnitsLeft(final BattleState.Side side) {
+    final boolean hasUnitsThatCanRollLeft;
+    hasUnitsThatCanRollLeft =
+        battleState.getUnits(EnumSet.of(side)).stream()
+            .anyMatch(
```
DanVanAtta, line 148: nit, deeply nested functional chain. Consider decomposing this a bit with helper methods. EG:

```
return battleState.getUnits(EnumSet.of(side)).stream()
  .anyMatch(unitIsActive(BattleState.side.OFFENSE));
```

I believe you can also get rid of the intermediate variable as well and just return the result directly as well.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
```diff
@@ -0,0 +1,173 @@
...
+                getActiveUnits()
+                    .and(
+                        Matches.unitIsSupporterOrHasCombatAbility(
+                            side == BattleState.Side.OFFENSE)));
+    return hasUnitsThatCanRollLeft;
+  }
+
+  private Predicate<Unit> getActiveUnits() {
```
DanVanAtta, line 156: This is perhaps too much information for a step class to have, to know exactly when a unit is active. I suspect having `unit` have a method for this would maybe be more appropriate, but failing that, would it make sense to create a `Matches` for this condition?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
```diff
@@ -0,0 +1,173 @@
...
+                : Matches.unitIsLand().negate());
+  }
+
+  private Collection<Unit> getUnprotectedUnits(final BattleState.Side side) {
+    final Collection<Unit> unitsToKill;
+    unitsToKill =
+        CollectionUtils.getMatches(
+            battleState.getUnits(EnumSet.of(side)),
+            getActiveUnits().and(Matches.unitIsNotInfrastructure()));
+    return unitsToKill;
```
DanVanAtta, line 171: Ditto on direct return, can avoid the intermediate variable.

game-core/src/test/java/games/strategy/triplea/delegate/battle/FakeBattleState.java
```diff
@@ -82,6 +82,18 @@
     return dependentUnits;
   }
 
+  @Override
+  public Collection<Unit> getUnits(final EnumSet<Side> sides) {
+    final Collection<Unit> units = new ArrayList<>();
```
DanVanAtta, line 87: Seeing this same code 3 times is raising real questions :grin: 

Why does FakeBattleState need to have this implementation? Can we avoid it and perhaps use a true mock? This is kinda concerning code as our test code has implementation that arguably should have a test of its own.

Are there options where we can remove this method and/or class?

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnitsTest.java
```diff
@@ -0,0 +1,508 @@
...
+  @Nested
+  class GetNames {
+
+    @Test
+    void noNameIfNotWater() {
+      when(battleSite.isWater()).thenReturn(false);
+      final BattleState battleState = givenBattleStateBuilder().battleSite(battleSite).build();
+      final RemoveUnprotectedUnits removeUnprotectedUnits =
+          new RemoveUnprotectedUnits(battleState, battleActions);
+      assertThat(removeUnprotectedUnits.getNames(), hasSize(0));
```
DanVanAtta, line 57: nit, favor using the IsEmptyCollection matcher instead of checking size == 0 for collections.

DanVanAtta (commented):
Got most of the way through this PR, wouldn't mind having one more pass at it though.
Left a few comments/questions, nothing too major. It is good to see code extracted and put under test.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
trevan: I changed the name of this class multiple times.  I thought I had updated all of the places with the latest name but it looks like I missed one :)

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
trevan: They can be private.  Leaving them as protected just makes accessing them easier in the child classes.  This will just change the code from `this.battleState.getUnits()` to `this.getBattleState().getUnits()` so I'm not sure how useful it is to make them private.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
trevan: I don't prefer ternary operations when the condition is multiple lines.  I'll get rid of the intermediate variables by just returning a new List in each return.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
trevan: Yes, it will be invoked if `getTransportCasualtiesRestricted` is false.  This method `checkAndRemoveUnits` is called from another step (see https://github.com/triplea-game/triplea/pull/7606/files#diff-009c223dfe7c716d07dc90f27899c933R1651).

The "check" is talking about the inner checks, not the `getTransportCasualtiesRestricted`.  It is checking for which units need to be removed.  Sometimes, units should be removed but most of the time, none are.

And the way that steps are now being created, they will always be added.  So we have to do all of the checks within the `execute` method or methods that it calls.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
trevan: Oops, ignore my comment.  I was thinking that this class had children classes but it doesn't.  I'll change them to private.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
trevan: Moving to `Matches` is a good idea but I'm not sure about the name.  The original code called this match `notSubmergedAndType`.  I didn't like that name and thought it looked more like active units.  But if it is going to be in `Matches` it probably needs a better name.

game-core/src/test/java/games/strategy/triplea/delegate/battle/FakeBattleState.java
trevan: I created FakeBattleState so that I didn't have to mock as many things.  Instead of having to specify all the required states, I can just use this real object that has default values and then override those defaults as needed.

I can't remove the method because it is part of the `BattleState` interface and this implements that interface.

As for code in 3 places, it is looking more and more like I'll be deleting `BattleSteps`.  There's 2-5 more steps that need to be converted but then the code for `BattleSteps` will basically be a list of objects that are sorted.  At that point, I'm thinking I just need a helper class to build the list of objects and then `MustFightBattle` can grab those objects, sort them, and then map them to their step names.  That will remove one of the `BattleState` implementations.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
DanVanAtta: Fair enough, could we rename this perhaps to something more direct like "removeUnprotectedUnits" and leave the fact that perhaps no units will be removed as an implementation detail. The API already does not expose any status flags that something was removed and we are purely modifying internal state of the class. Come to think about it, this is actually a code smell, some other class knows when this should be called, knows that this class needs to mutate its state rather than that being managed by this class itself. I think that is a bigger and more important issue, but perhaps a bit much to tackle now, I'd be happy if we can document the side-effects that happen and perhaps improve the name of the method.

game-core/src/test/java/games/strategy/triplea/delegate/battle/FakeBattleState.java
DanVanAtta: I'll certainly say that mocks are a testing smell and need to be used strategically. Mocking input can be good (though still it's better to use value objects for input),  injecting mock implementations is generally bad.

Fakes are certainly a class of test object of their own, similar to stubs, but often one would want a very simplified implementation and not a real implementation. I think we probably would want to get rid of `FakeBattleState` in the medium term. I think we can see how this evolves knowing that there are still some smells lingering around.

DanVanAtta:
@trevan I'll try to take another (Final) review pass either later tonight or tomorrow. We're caged in here in the US northwest by wildfire smoke, on the plus side it means I'll have time available for Triplea.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
trevan: The removal of unprotected units and undefended transports is done in two different places.  I didn't want to duplicate the code so I have the code in this step and the other step calls this code as well.  I could create a helper class that does all this work as well and then have both steps call that helper class, if you'd like.

I'll rename it, though.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
trevan: I've created a new `Matches.unitIsActiveInTerritory`

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
```diff
@@ -21,10 +29,12 @@
 
   Collection<Unit> getAttackingUnits();
 
-  Collection<Unit> getAttackingWaitingToDie();
-
   Collection<Unit> getDefendingUnits();
 
+  Collection<Unit> getUnits(EnumSet<Side> sides);
```
DanVanAtta, line 34: nit, any consideration to make the API generic and allow for `EnumSet` to passed in as an implementation detail?
For example, if I had a `Set<Side>`, should I really be forced to convert it to an `EnumSet` to call this method? Is the runtime difference of forcing a collection copy enough that the `EnumSet` is that much more efficient?

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
```diff
@@ -12,7 +12,15 @@
 
   enum Side {
     OFFENSE,
-    DEFENSE,
+    DEFENSE;
+
+    public Side getOpposite() {
+      if (this == OFFENSE) {
```
DanVanAtta, line 18: nit, have you considered using a ternary here?

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
DanVanAtta: Perhaps even better, perhaps a vararg for this? In such a case a caller would not have to create a collection at all. WDYT?

DanVanAtta: For yet more consideration, any thought to making this a default method?

It looks like these methods would give you the data access needed:
```

  @Override
  public Collection<Unit> getDefendingUnits() {
    return Collections.unmodifiableCollection(defendingUnits);
  }

  @Override
  public Collection<Unit> getRemainingAttackingUnits() {
    return Collections.unmodifiableCollection(attackingUnits);
  }
```

DanVanAtta: Still more questions.. Should the above two methods be replaced by usages of this `getUnits(...)` method? Reducing the API surface of such a class would be a good thing, particularly removal of simple getter API methods.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
DanVanAtta: Fair enough, something to consider though. Having a method lead with `return` is  a powerful clue to a reader. Otherwise with an 'if' and 'otherwise' block here, it's arguably cognitively more complex. Specifically until you've scanned the whole method do you realize that all it is is a condition and a return statement based on that (and the simplest form of that is a ternary which leads with the keyword 'return' to help tell a reader that information immediately)

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
```diff
@@ -0,0 +1,156 @@
...
+        Matches.unitIsNotLand()
+            .and(Matches.enemyUnit(player, battleState.getGameData()))
+            .and(Matches.unitIsSubmerged().negate())
+            .and(Matches.unitCanAttack(player));
+    return CollectionUtils.getMatches(battleState.getBattleSite().getUnits(), enemyUnitsMatch);
+  }
+
+  private void checkUnprotectedUnits(final IDelegateBridge bridge, final BattleState.Side side) {
+    // if we are the attacker, we can retreat instead of dying
+    if (attackerHasRetreat(side)) {
```
DanVanAtta, line 123: What do you think about something that is more like:
```
if(!attackerCanRetreat(side)) {
   removeUnprotectedUnits(side);
}

```

Having control flow at the outer layers is generally a good thing. A series of void method calls is not a great abstraction, it's cognitively equivalent to adding comments and inlining the methods. A reader being able to skip sections of code because they are tracing control flow is on the other hand very valuable beyond demonstrating the flow of control at the outset. That pattern is called "[imperative shell, functional core](https://www.destroyallsoftware.com/screencasts/catalog/functional-core-imperative-shell)". 

I also suggest that as it took quite a bit of 'groking' to determine that this method is largely just a lot of conditionals then we do the remove operation. I think there is also a mixing of layers of abstraction code smell here too, eg:
```
- high level boolean check
- low level boolean check
- low level boolean check
- high level fetch of unit collection
- low level removal of units
```

With the above you keep the high level boolean check and the control flow is extracted, and you can also potentially inline `getUnprotectedUnits` into `removeUnprotectedUnits` which I think then makes a lot of sense and keeps the abstractions at a consistent level (and there are just as many methods, but each one is more focused in what they do).

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/MockGameData.java
```diff
@@ -29,6 +33,18 @@ public GameData build() {
     return gameData;
   }
 
+  public MockGameData withAlliedRelationship(
+      final GamePlayer player1, final GamePlayer player2, final boolean value) {
+    when(relationshipTracker.isAllied(player1, player2)).thenReturn(value);
+    return this;
+  }
+
+  public MockGameData withWarRelationship(
```
DanVanAtta, line 42: At a first glance this is a bit confusing to have a boolean parameter that negates the whole thing. For consideration, boolean parameters automatically violate SRP, the method by definition then does multiple things (so long as the parameter is used).

Perhaps the underlying code should get the relationship type instead of using the 'isAllied' and 'isAtWar' methods?

In that case you could have `withRelationship(Player1, Player2, RelationshipType)`

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnitsTest.java
```diff
@@ -0,0 +1,510 @@
...
+  @Mock IDelegateBridge delegateBridge;
+  @Mock BattleActions battleActions;
+  @Mock GamePlayer attacker;
+  @Mock GamePlayer defender;
+  @Mock Territory battleSite;
+
+  @Nested
+  class GetNames {
+
+    @Test
```
DanVanAtta, line 53: A @DisplayName on this one could be useful IMO to explain why there is no name expected if land. It would be good to see this made clear up front why we expect this situation to be true. I mention that because without context, it's not clear why 'unprotected units' is an invalid step for land, 'land' can have protected or unprotected units, hence there is context here into the production code needed to understand the test, which is a test code maintenance scaling problem.

DanVanAtta: This makes me wonder if the rule should be named more directly "RemoveUnprotectedCasualtyRestrictedTransports", WDYT?

DanVanAtta (commented):
Sorry to do an ongoing review, I think I'm about 60-80% done with review. Certainly getting a bit finer grained. I'll pick up again in a few hours or tomorrow.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnitsTest.java
trevan: Do you mean rename the step?  Or the test?

And I agree that the `getNames` condition on water vs land is weird but that is the relic of the original code.  I'm not confident in changing at this point.  I actually think that this step should be split in 2 steps, one for the undefended transports and one for the unprotected units.  But with the discussion we had in the forum about the "auto-kill" logic, I think this step will need to be reworked completely.  So, I'd rather leave the original logic alone and test it.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/MockGameData.java
trevan: I would have to make the mocks lenient.  Because making a player be at war with another requires that the allied be false and the war true.  But both of those mocked calls will not be executed during the test and the test will fail.

Also, this is really just a wrapper around the `isAtWar` and `isAllied` methods to make give it a fluent api.  So, if you mocked the relationship outside of this wrapper, you'd still have the same problems you mentioned.

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
trevan: I hadn't thought of the vararg idea but I like it.  I'll make that change.

I've thought of replacing the `getAttackingUnits` and `getDefendingUnits` with just `getUnits(...)` but didn't want to make that change in this PR.  I've got one more step conversion PR in the process and then after that one, I'll do a clean up PR.  There's some tests that are now redundant that I can remove and the steps creation in `MustFightBattle#addFightSteps` will be ready for simplification.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
trevan: I had basically copied the original code and extracted out larger pieces so that it was more readable but you are right, it is still hard to read.  So, I've restructured the code to follow the [return early pattern](https://medium.com/swlh/return-early-pattern-3d18a41bba8).

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
```diff
@@ -51,20 +51,27 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
   }
 
   public void removeUnprotectedUnits(final IDelegateBridge bridge, final BattleState.Side side) {
-    if (Properties.getTransportCasualtiesRestricted(battleState.getGameData())) {
-      checkUndefendedTransports(bridge, side);
-      checkUnprotectedUnits(bridge, side);
+    if (!Properties.getTransportCasualtiesRestricted(battleState.getGameData())) {
+      return;
     }
+    // if we are the attacker, we can retreat instead of dying
```
DanVanAtta, line 57: Personally I would not have preserved this comment, it makes sense with some context but otherwise is actually not that helpful and is arguably more cheeky than informative. It does leave me with more questions than answers.

Is there any reason otherwise to not combine the two if statements?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
```diff
@@ -51,20 +51,27 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
...
+    if (attackerHasRetreat(side)) {
+      return;
+    }
+    checkUndefendedTransports(bridge, side);
```
DanVanAtta, line 61: Having the early return in this method is much better :+1: 

Early return is a pattern I like, I do prefer:
```
void method()  {
  if(something) {
     return;
  }
}
```

Over:

```
void method() {
  if(!something) {
    // do stuff

  }
  return;

}

```

But, if a method is going to be a no-op, I generally prefer for it just not to be invoked. It's always a bit frustrating to read through code to realize that method calls are no-ops. Methods generally are called so that something can happen. For example, if a method no-ops because of null input, presumably the caller called it for something to happen, did the caller realize that they even provided a null input? IN such a case, I see something like:

```
void method(Sringg arg) {
  if(arg == null) {
     return; // << this is a silent failure, a BUG. Did the caller know they provided a null arg and were just doing a no-op?
  }
}
```

DanVanAtta (approved):

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnitsTest.java
DanVanAtta: I was more suggesting to rename the  class under test.

PR #1077, opened 2016-08-17
XML Property Updates from #1033 + XML GameProperty cleanup

DanVanAtta:
Turns out non-String versions of the methods marked @GameProperty are not called. This is for sure true for Boolean/Integer params when there is a competing String version. These were cleaned up and hence the high deletion count. 

Otherwise most of the items from #1033 were done minus naval bomard rename which requires flipping the game engine behavior, also not done was the SBR rename which caused a test break.

After this is done we'll need to apply similar updates to the maps.

The following are done:

-------
4)Units repair at end of round->Units Repair Hits End Turn
5)Units repair at beginning of round->Units Repair Hits Start Turn
6)option name="turns"->option name="rounds"
7A)victoryCity": stop supporting boolean only for this option
7B)option name="victoryCity" value="true"->option name="victoryCity" value="1"
7B)option name="victoryCity" value="false"->option name="victoryCity" value="0"
8A)"takeUnitControl"
8B)"giveUnitControl": stop supporting boolean only for this option
9A)"occupiedTerrOf"
9B)"occupiedTerrOf"->"originalOwner"
10A)"isParatroop"
10B)"isParatroop"->"isAirTransportable"
11A)"isMechanized"
11B)"isMechanized"->"isInfantry"
12A)"isTwoHit"
12B)"isTwoHit" value="true"->"hitPoints" value="2"
12B)"isTwoHit" value="false"->"hitPoints" value="1"
13A)option name="conditionType" value="XOR"
13B)option name="conditionType" value="XOR"->option name="conditionType" value="1"

-----

For reference, scripted commands:

```
find . -name "*xml"  -type f 2> /dev/null | xargs sed -i 's/occupiedTerrOf/originalOwner/'
find . -name "*xml"  -type f 2> /dev/null | xargs sed -i 's/ictoryCity" value="false"/ictoryCity" value="0"/'
find . -name "*xml"  -type f 2> /dev/null | xargs sed -i 's/ictoryCity" value="true"/ictoryCity" value="1"/'
find . -name "*xml"  -type f 2> /dev/null | xargs sed -i 's/isTwoHit" value="false"/hitPoints" value="1"/'
find . -name "*xml"  -type f 2> /dev/null | xargs sed -i 's/isTwoHit" value="true"/hitPoints" value="2"/'
find . -name "*xml"  -type f 2> /dev/null | xargs sed -i '/option name="giveUnitControl" value="true/d'
[...]

PR #7630, opened 2020-09-12
Update doc: "Upgrade from 1.8.0.9 to 1.9.0.0.3266" (formatting)

DanVanAtta:

### DIFF
--- a/docs/map-making/upgrading-1-8-maps.md
+++ b/docs/map-making/upgrading-1-8-maps.md
@@ -13,48 +13,38 @@ Note: Some items are not validated; thus, if your game plays on 1.9.0.0.3266, th
 ### Not already deprecated xml changes:
 _(you normally need to change these only)_
 
-`attatchment`:`attachment`
-`Attatchment`:`Attachment`
-`attatchTo=`:`attachTo=`
-`="isImpassible"`:`="isImpassable"`
-`="conditionType" value="XOR"`:`="conditionType" value="1"`
-`="turns":="rounds"`
-`="Battleships repair at end of round"`:`="Units Repair Hits End Turn"`
-`="Battleships repair at beginning of round"`:`="Units Repair Hits Start Turn"`
-`="Naval Bombard Casualties Return Fire Restricted"`:`="Naval Bombard Casualties Return Fire"`
+- `attatchment`:`attachment`
+- `Attatchment`:`Attachment`
+- `attatchTo=`:`attachTo=`
+- `="isImpassible"`:`="isImpassable"`
+- `="conditionType" value="XOR"`:`="conditionType" value="1"`
+- `="turns":="rounds"`
+- `="Battleships repair at end of round"`:`="Units Repair Hits End Turn"`
+- `="Battleships repair at beginning of round"`:`="Units Repair Hits Start Turn"`
+- `="Naval Bombard Casualties Return Fire Restricted"`:`="Naval Bombard Casualties Return Fire"`
 
 ### Special code xml changes for the above changes:
 _(you probably don't have these)_
 
-`value="conditionType" count="XOR"`:`value="conditionType" count="1"`
-`value="conditionType" count="-reset-XOR"`:`value="conditionType" count="-reset-1"`
+- `value="conditionType" count="XOR"`:`value="conditionType" count="1"`
+- `value="conditionType" count="-reset-XOR"`:`value="conditionType" count="-reset-1"`
 
 ### Already deprecated but effective xml changes:
 _(you shouldn't have had these in your games already)_
 
-`="isTwoHit" value="true"`:`="hitPoints" value="2"`
-`="isTwoHit" value="false"`:`="hitPoints" value="1"`
-`="occupiedTerrOf"`:`="originalOwner"`
-`="victoryCity" value="true"`:`="victoryCity" value="1"`
-`="victoryCity" value="false"`:`="victoryCity" value="0"`
-
-### Already deprecated and ineffective xml deletions, for all the respective attachments:
-_(you shouldn't have had these in your games already and, if you did, they were doing nothing)_
-
-`="isParatroop"`:*
-`="isMechanized"`:*
-`="takeUnitControl"`:*
-`="giveUnitControl" value="true"`:*
-`="giveUnitControl" value="false"`:*
-*Delete the whole `<option/>`
+- `="isTwoHit" value="true"`:`="hitPoints" value="2"`
+- `="isTwoHit" value="false"`:`="hitPoints" value="1"`
+- `="occupiedTerrOf"`:`="originalOwner"`
+- `="victoryCity" value="true"`:`="victoryCity" value="1"`
+- `="victoryCity" value="false"`:`="victoryCity" value="0"`
 
 ### map.properties changes:
 
-`color.Impassible=`:`color.Impassable=`
+- `color.Impassible=`:`color.Impassable=`
 
 ### sounds.properties changes:
 
-`.wav`:`.mp3`
+- `.wav`:`.mp3`
 
 ### Non-code changes:
 
@@ -63,26 +53,32 @@ Change all sounds inside "/sounds" from "wav" to "mp3" format.
 **Procedural alternatives:**
 
 For a more defined change, instead of the aforementioned:
-`attatchment`:`attachment`
+- `attatchment`:`attachment`
 you can change:
-`attatchmentList`:`attachmentList`
-`<attatchment`:`<attachment`
-`attatchment>`:`attachment>`
-`="games.strategy.triplea.attatchments`:`="games.strategy.triplea.attachments`
+- `attatchmentList`:`attachmentList`
+- `<attatchment`:`<attachment`
+- `attatchment>`:`attachment>`
+- `="games.strategy.triplea.attatchments`:`="games.strategy.triplea.attachments`
 
 For a more defined change, instead of the aforementioned:
-`Attatchment`:`Attachment`
+- `Attatchment`:`Attachment`
 you can change:
-`="techAttatchment`:`="techAttachment`
-`="unitAttatchment`:`="unitAttachment`
-`="territoryAttatchment`:`="territoryAttachment`
-`="canalAttatchment`:`="canalAttachment`
-`="rulesAttatchment`:`="rulesAttachment`
-`="playerAttatchment`:`="playerAttachment`
+- `="techAttatchment`:`="techAttachment`
+- `="unitAttatchment`:`="unitAttachment`
+- `="territoryAttatchment`:`="territoryAttachment`
+- `="canalAttatchment`:`="canalAttachment`
+- `="rulesAttatchment`:`="rulesAttachment`
+- `="playerAttatchment`:`="playerAttachment`
 
 **Note:**
 
 The following xml codes, as being properties, are not validated (if you leave them, the game will play, but they will set nothing in the engine) (it is suggested you double check you don't have them anymore):
-`="Battleships repair at end of round"`
-`="Battleships repair at beginning of round"`
-`="Naval Bombard Casualties Return Fire Restricted"`
+- `="Battleships repair at end of round"`
+- `="Battleships repair at beginning of round"`
+- `="Naval Bombard Casualties Return Fire Restricted"`
+- `="isParatroop"`:*
+- `="isMechanized"`:*
+- `="takeUnitControl"`:*
+- `="giveUnitControl" value="true"`:*
+- `="giveUnitControl" value="false"`:*
+*Delete the whole `<option/>`


Cernelius:
This pull request has messed up the file, that is now giving wrong information, as far as I know.

The problem is that you have deleted the whole "Already deprecated and ineffective xml deletions, for all the respective attachments" and added them to the (unnecessary) "Note" section.

For the future, I suggest such changes being reviewed by a developer other than the one making the changes.

DanVanAtta:
Feel free to maintain this doc @Cernelius and make suggestions to it. There have been people asking for 1.8 upgrades, referred to this document and it was not helpful.The formatting in particular made it difficult to read and follow.

The notes section stated "delete these", and another section said "delete these because they do nothing." That seems redundant.

DanVanAtta:
> I suggest such changes being reviewed by a developer other than the one making the changes.

Ideally, but we don't have the capacity. It's the reality. I'll impress on you that there things that NEED to get done, and it's FAR more than we can humanly do. We cannot spend weeks on every little thing, it's not a fast enough pace for what we need. So feel free to jump in where you can in a helpful and constructive way to move us forward.

docs/map-making/upgrading-1-8-maps.md
```diff
@@ -13,48 +13,38 @@ Note: Some items are not validated; thus, if your game plays on 1.9.0.0.3266, th
...
+- `="Battleships repair at end of round"`:`="Units Repair Hits End Turn"`
+- `="Battleships repair at beginning of round"`:`="Units Repair Hits Start Turn"`
+- `="Naval Bombard Casualties Return Fire Restricted"`:`="Naval Bombard Casualties Return Fire"`
 
 ### Special code xml changes for the above changes:
 _(you probably don't have these)_
 
-`value="conditionType" count="XOR"`:`value="conditionType" count="1"`
-`value="conditionType" count="-reset-XOR"`:`value="conditionType" count="-reset-1"`
+- `value="conditionType" count="XOR"`:`value="conditionType" count="1"`
```
DanVanAtta, line 29: @Cernelius , as far as I can tell this was never done and "1" is not an equivalent value. Do you have background for why this line item was added? Do you know of any maps where this update was actually made? It looks like the code still only handles 'xor'.

DanVanAtta:
@Cernelius now that we can support 1.8 maps, there is no longer a need to upgrade them. They can happily continue working and we can spend our time elsewhere. I'd argue upgrading an existing map hosted in a some unknown repository for it to continue working for only the sake of having the most pleasing XML is busy-work. If there is no other benefit, if the map can continue working without being modified, that's a win.

Do we still then really need this document? Would we be better off on focusing on making it so new maps can be easily made and have the cleanest instructions possible for that? Do we really need to burden people who just want to make a map with telling them all of the legacy options they should not use, vs focusing on telling them how to make a map with the latest?

Cernelius:
> Feel free to maintain this doc @Cernelius and make suggestions to it. There have been people asking for 1.8 upgrades, referred to this document and it was not helpful.The formatting in particular made it difficult to read and follow.
> 
> The notes section stated "delete these", and another section said "delete these because they do nothing." That seems redundant.

Can you go back to the previous state? That's the change I would make.

The "Already deprecated and ineffective xml deletions, for all the respective attachments" lists all the items that were doing nothing if present, and I understand have been completely deleted from the program, so now must be removed (they should have already removed before, like everything that was already deprecated).

The "Note" is merely a list of items that are not validated, so the program will not tell you if you didn't update them or failed to update them correctly (and, actually, all of them are not to remove, but to change, and are items of the "Not already deprecated xml changes" section). The fact that I've written "it is suggested you double check you don't have them anymore" doesn't imply that are things you need to delete. It just means that you should not to have them anymore (because you substituted them with something else or with nothing). Practically, the "Note" section is warning the maker that if its game runs that doesn't necesarily mean it updated it correctly, pointing at the things that the program fails to validate.

DanVanAtta:
> The "Already deprecated and ineffective xml deletions, for all the respective attachments" lists all the items that were doing nothing if present, and I understand have been completely deleted from the program, so now must be removed (they should have already removed before, like everything that was already deprecated).

Cool, though I wonder if we should have a different document to state this. In the short term, can this be stated in a more clear way in this document (made less wordy and more direct. As is, it's not clear how that applies to 1.8->1.9 migration and why it's included here - I think it's this document trying to do "too much").

> Practically, the "Note" section is warning the maker that if its game runs that doesn't necesarily mean it updated it correctly, pointing at the things that the program fails to validate.

This is good context, though again, I think we need a second document stating:
- Here is old, here is what the new should be.


This document I think is best if it strictly states what you *have* to do to convert a 1.8 to a 1.9 map. Though again, I'm not sure if we really need this anymore since 1.8 maps will no longer be required to be updated. So perhaps it's best to transform this into a document of "here are a bunch of legacy options, here is what they changed into"

DanVanAtta:
@Cernelius are you willing to take a crack at fixing the formatting of the document and could you submit a PR for that? I think that is an easy first step to get this document to at least be useful again.

Cernelius:
> @Cernelius are you willing to take a crack at fixing the formatting of the document and could you submit a PR for that? I think that is an easy first step to get this document to at least be useful again.

I don't feel it needs to be changed, but I'm open to review changes to assure the substance is the same, under my understanding. Meaning that I feel it is good enough, but if someone rewords or reformats it it can be better.

docs/map-making/upgrading-1-8-maps.md
Cernelius: I remember you told me that the "XOR" is not anymore supported. If the "XOR" is still supported and it is not deprecated, then all the items about converting "XOR" to "1" should be removed from the document.

The reason why I suggested to remove the "XOR" is that "XOR" means "1", as long as the conditions are 2 or more (I don't know what's going to happen if you have "XOR" and 1 or 0 conditions (logically, you should never have "XOR" with less than 2 conditions), but I assume it is still the same as having "1"). Therefore, the "XOR" value is completely useless, as you can use "1" instead of it. I also suppose not everyone knows what "XOR" means, especially since in English "or" is normally supposed to be exclusive ("You can have this or that for 1 pound" doesn't mean you can have both).

DanVanAtta:
The fact I changed this I would present to you as evidence it does need to be changed. I try not to spend times on things just for the pure fun of it. More to the point, someone asked about 1.8 to 1.9 conversions, this document WAS NOT HELPFUL. It's useless in its current form. Starting with the fact that when you look at the rendered version of it, the lack of formatting makes it nonsensical. Beyond that, ti's not well enough focus to be clear and readily understood. Please have a look at how it renders, at the very least the formatting needs to be fixed:

https://github.com/triplea-game/triplea/blob/master/docs/map-making/upgrading-1-8-maps.md

DanVanAtta:
@Cernelius I"m sorry to be a bit sharp with you. It's one thing to point out problems that need to be resolved, but it's another to say "no" and "I liked the old problems because they were not problems for me." I think you're ignoring the existing problems, to the point where this document for example makes sense to you, but otherwise is a cost with no benefit to everyone else. In such a case, please understand the existing problems and see how we can work together to fix them. If you can contribute directly to the solution, we are short on labor in this project, those efforts are sincerely appreciated. Otherwise the simple "no, I don't like this", is just not helpful.

I'm pointing to a problem called the "normalcy of the absurd". Where something that is outright absurd becomes normal because you get used to it. It's a problem that a new person would see and go "wow, this is insane", but when you deal with it enough it just becomes normal and part of what you deal with everyday and then just seems normal.

Does it seem normal that it's extremely difficult to know what should change, to what, and what should be delete by simply reading this document? https://github.com/triplea-game/triplea/blob/master/docs/map-making/upgrading-1-8-maps.md

When I read it, I was left wondering what should change to what and was not sure what should have been deleted or could remain or what. In the end, for a document of what's needed to go from 1.8 to 1.9, we're not there.

DanVanAtta:
For example, to know which features were needed to support 1.8, this document was not helpful. I had to go to the original issue where we discussed what was changing to what, and used that. So we need some work here..

docs/map-making/upgrading-1-8-maps.md
DanVanAtta: Thanks for the response here. Yeah, I think removing the XOR mention for now is best until we can clarify it. Agree that it's a classic problem that the commonly used 'or' is not a logical or, and is actually a logical 'xor'.

I was trying to track down the '1' value that was mentioned here and noticed that we still had code to handle xor.

DanVanAtta:
@Cernelius what do you think:

1. Should we improve this document and just try to make it clearly and simply focused on the 1.8 to 1.9 transition?

2. Should we convert this document into a  more generic legend of deprecated properties and what they should be instead? This would be almost more a glossary of all old properties and what they should changed into.

I think (2) would perhaps have more value compared to (1), WDYT?

docs/map-making/upgrading-1-8-maps.md
Cernelius: So, can you confirm that the "XOR" has never been removed, thus the information about the fact that the "XOR" needed to be changed to "1" has always been wrong.

If so, can we at least deprecate the "XOR" in favour of "1"?

Cernelius:
> @Cernelius what do you think:
> 
> 1. Should we improve this document and just try to make it clearly and simply focused on the 1.8 to 1.9 transition?
> 2. Should we convert this document into a  more generic legend of deprecated properties and what they should be instead? This would be almost more a glossary of all old properties and what they should changed into.
> 
> I think (2) would perhaps have more value compared to (1), WDYT?

I've no plans, nor the knowledge, to update that document to a later version. If nobody else is doing it, the document can only be kept as a legacy document from 1.8 to 1.9. A new document can be made for anything more recent. I believe the document is missing items that were made after the specified new version, anyways.

**In any case, it would be good to assure that the document is at least correct and exaustive for the specified versions.**

In particular, there is also still this matter, that was never addressed, as far as I know (with special reference to the "SBR Affects Unit Production" property, that I believe was removed but missing documentation as such):
#3462

I believe, but I'm not sure (so, I cannot update until I'm clarified), that you removed the "SBR Affects Unit Production" property, as I suggested, but you didn't tell me that you did, so this matter was never documented (basically the opposite of what it seems it happened to the "XOR" option, that would need to be removed from the documentation, instead, if you didn't remove it).

DanVanAtta:
This was renamed:     "SBR Affects Unit Production"->"Damage From Bombing Done To Units Instead Of Territories"

I can't see where that was merged, so the backstory and communication I cannot track down to tell you.

For XOR, it looks like it was out of the range of the items mentioned as completed in #3462, so perhaps it was a communication confusion. If it still makes sense to change it, a 'developer > maps' posting in forums would be a good place to create a thread about it. I'm not really sure what a numeric value means for a condition, so I've certainly questions about what and why it would be changing.

----------


I think it would make sense to have one document describing legacy properties. It was important to know specifically how to upgrade from 1.8 to 1.9, but if 1.8 maps will work anyways, then nobody will wonder "how do I get this map to work again?" Otherwise, having one place, a sort of master release notes for map makers, to see what is old and what they should do instead I think would be really useful. If we were to make that latter document, it would be duplicative of the 1.8 transition notes, so I wonder if we would just as well make a deprecate property gloassary. I also wonder a bit too, if we change the 1.9 properties, and there are new values that should be used, then it might lead to a situation where you have to follow multiple documents to get to a final result.

Cernelius:
> This was renamed: "SBR Affects Unit Production"->"Damage From Bombing Done To Units Instead Of Territories"

I don't think so. The property "Damage From Bombing Done To Units Instead Of Territories" already existed. I guess you are looking at something beforehand.

To clarify, before the changes, I believe the "SBR Affects Unit Production" property was setting the "Damage From Bombing Done To Units Instead Of Territories" true if true. I suggested removing the "SBR Affects Unit Production" completely from the engine. Was that done between 1.8.0.9 to 1.9.0.0.3266? The specific question is if both of these are true:

- In 1.8.0.9 the "SBR Affects Unit Production" sets the "Damage From Bombing Done To Units Instead Of Territories" true if true.
- In 1.9.0.0.3266 there is no "SBR Affects Unit Production" property (anywhere in the program).

DanVanAtta:
"SBR Affects Unit Production" no longer exists in the code base.


commit ebeece4a5df97deb24fdf2f3a6f07c714824ccdc
Author: Dan Van Atta <dhvatta@gmail.com>
Date:   Wed Aug 17 03:07:25 2016 -0700

    "SBR Affects Unit Production"->"Damage From Bombing Done To Units Instead Of Territories"


I don't know which PR merged that.

> Unless you mean that it was renamed in all maps having that code,

I presume this happened.


-----------

> The specific question is if both of these are true:

At this point @Cernelius,  I do think we should focus on a document of how to update maps to the latest. Nobody cares to upgrade a map from 1.8 to an early version of 1.9 where the map still would not work. IN other words, instructions of how to get a non-working map to a more recent version of a non-working map does not seem very useful (historical yes, useful, no).

Cernelius:
> At this point @Cernelius, I do think we should focus on a document of how to update maps to the latest. Nobody cares to upgrade a map from 1.8 to an early version of 1.9 where the map still would not work. IN other words, instructions of how to get a non-working map to a more recent version of a non-working map does not seem very useful (historical yes, useful, no).

Assuming that means there is no documentation about that, already, that's up to you or a developer. I recall there have been a series of additional changes (for example, "isInfantry" to "isLandTransportable" and many others) after 1.9.0.0.3266, made mostly by @ron-murhammer. I don't know how you can track them all down to document them in an exaustive (missing nothing!) document like the one I made for 1.9.0.0.3266. I believe some or most of them are currently documented in pos2 as deprecated, but I think there have been other changes I don't remember.

I would still keep the 1.8.0.9 and 1.9.0.0.3266 as a legacy documentation.

Cernelius:
> "SBR Affects Unit Production" no longer exists in the code base.

Good to know, but, if we want to be sure to be correct, I would also need to be sure that it doesn't exist in TripleA 1.9.0.0.3266 (if it was removed after that, then it would not be part of those changes). I'm not sure, because I suggested to remove it, but I recall you didn't list amongst the thing you removed. There might be a chance it might have been removed some time after.

Cernelius:
Once this is confirmed, I will proceed updating the "1.8.0.9 to 1.9.0.0.3266" documentation by removing the "XOR" things and adding the "SBR Affects Unit Production" thing. Hopefully, that should make the documentation finally correct an exaustive.

docs/map-making/upgrading-1-8-maps.md
DanVanAtta: XOR Is still there.

What would "1" mean as a condition type?

DanVanAtta:
@Cernelius perhaps we should move this conversation to an issue if there is more to confirm? PR discussions tend to be hard to find, particularly for someone looking for background information on a topic. PR discussions ought to generally stay focused on what has changed in that PR, any auxilierary topics taken to an issue (or forums) to broaden it up.

With that said, I think removing the XOR portion would stay accurate. 

I am leaning towards having one document with deprecated options. An historical and duplicative document is of questionable value.

DanVanAtta:
With the above said, yeah, whatever documentation we have, it should be accurate (and really current, which is my concern about *3266 properties possibly being changed).

DanVanAtta:
Let's consider this scenario, we keep documenting how to upgrade, in which case you go to the 1.8->3266 document, then the 3266->7058 document, then the 7058->1.10, then the 1.10->2.0 document, and then maybe more, and at the end of that and after tracing all of the transitive changes and aggregating it all, you know what the final properties should be. I think it's quite compelling to have one document here. (But again, getting the current documentation to be accurate is a first order task).

Cernelius:
> With that said, I think removing the XOR portion would stay accurate.

Ok, then now we know that "XOR" was not removed between 1.8.0.9 and 1.9.0.0.3266 (and all this time we told people that they needed to change "XOR" to "1", while they didn't).

Can it be quickly checked whether or not "SBR Affects Unit Production" was removed between 1.8.0.9 and 1.9.0.0.3266? Like seeing what is the latest TripleA having it.

docs/map-making/upgrading-1-8-maps.md
Cernelius: A numeral condition type means that exactly that number of conditions must be true (no more, no less). That is why having "1" is the same as having "XOR".

DanVanAtta: Thanks for the clarification @Cernelius 

For the purpose of this conversation, it seems like the XOR mention should be removed from this doc since as far as I can tell it has not been updated.

If we were to update it, it's probably a conversation to move to forums. IMO we could probably do even better than changing the value from condition type to condition count.

DanVanAtta:
1.9.0.0.3266 does not have "SBR..." referenced in the code.

"Damage From Bombing Done To Units Instead Of Territories" is referenced in the code quite a bit in 1.9.0.0.3266

Cernelius:
Actually, I still would like a clarification about the "XOR".

I found this:
https://github.com/triplea-game/triplea/pull/1077

In that, there is the mention "Remove XOR option type, remove nonString arg @GameProperty methods", linking to:

https://github.com/triplea-game/triplea/pull/1077/commits/c0463068649216742abd98765835e9bcc203a8a3

So, how is the "XOR" still around, if it was removed? Was it maybe only partially removed. Any chances that it might be broken?

If it was removed between 1.8.0.9 and 1.9.0.0.3266, but added back after 1.9.0.0.3266, it should still figure in these (legacy) changes (even though this is likely not going to be helpful right now).

**So, can you take a look at that and confirm that the "XOR" is actually to be deleted from the list of changes between 1.8.0.9 and 1.9.0.0.3266?**

Since you already confirmed it, I'll delete it, but I would like to be sure based on the links I provided. Substantially, the matter would be not to confirm the "XOR" is part of the current program, but whether or not it is part of TripleA 1.9.0.0.3266.

DanVanAtta:
It'd probably be best to test the condition out.

It was based on this code that it appeared to still be supported: https://github.com/triplea-game/triplea/commit/c0463068649216742abd98765835e9bcc203a8a3#diff-cbdad94f09ebd961ea55daa912e71e26R125

Though @Cernelius, good sleuthing work, the snippet you found is the removal of the handler code. If we do have an XOR it looks like it would fall through and then error out as the value is not numeric. That would imply any map with XOR and the condition were evaluated, that the game would crash. I think that probably speaks to very few maps having the option.

RE: "1.8.0.9 and 1.9.0.0.3266"

I'll emphasize that we probably should not focus on just the now purely historical 1.8 to 1.9 transition and limit to documenting the history of the one-time decision regardless whether that is correct now. Documentation should be current, right? Do we really want map makers to have to refer to 'N' documents to know how to get maps up to latest? Said another way, is there anyone who actually wants to upgrade their map to *just* 3266 and not the latest? What is the point of upgrading the map to an older version that is still broken.

I've been thinking about this some and I'd like to make this document more generic. It will be useful for upcoming XML changes. To capture when things happen, perhaps we can get the XML translations into a table and have a column noting which version had the change in it. Then you could know also when something happened in addition to having a single place to then know how to upgrade a map.

Cernelius:
> It'd probably be best to test the condition out.
> 
> It was based on this code that it appeared to still be supported: [c046306#diff-cbdad94f09ebd961ea55daa912e71e26R125](https://github.com/triplea-game/triplea/commit/c0463068649216742abd98765835e9bcc203a8a3#diff-cbdad94f09ebd961ea55daa912e71e26R125)
> 
> Though @Cernelius, good sleuthing work, the snippet you found is the removal of the handler code. So while the property could still be set, so the game would survive map parse, it does look like it would crash if the condition were evaluated. I think that probably speaks to very few maps having the option.

So, I just want to know exactly, since I'm only updating the 1.8.0.9 to 1.9.0.0.3266 information, should they tell to change "XOR" to "1" or should they not? I need to know this.

Cernelius:
> I've been thinking about this some and I'd like to make this document more generic. It will be useful for upcoming XML changes. To capture when things happen, perhaps we can get the XML translations into a table and have a column noting which version had the change in it. Then you could know also when something happened in addition to having a single place to then know how to upgrade a map.

As I said, documenting (correctly) is always good, but I, instead, suggest keeping the legacy document archived and making a new document from something to something. You are most likely not going to document all changes from TripleA 1.0, so whatever document is anyway going to be a "from X to Y" thing (I guess you'll pick 1.8.0.9 for the "X"), that will be limited backwards (unless you really document all changes from the initial TripleA) and likely forward too (since you cannot be sure yourself or other developers will keep on updating it).

DanVanAtta:
An archived legacy document is actually less than useful @Cernelius . It makes the useful documentation harder to find, it also clouds search results when you search for key words. 

>  You are most likely not going to document all changes from TripleA 1.0, so whatever document is anyway going to be a "from X to Y" thing

I agree, but having one document is of valuable. Though you're arguing for the oppostive that is even worse. If someone wants to go from X->Z, having two documents that are "X->Y" and "Y->Z" is worse. Why not have one document that can be maintained and have simply "X->Z"?


You're ignoring this statement @Cernelius 

 >Map makers care about getting their maps to latest, right? Does anyone really care or want to get their maps just to 1.9.0.0.3266

Upgrading a map to only 1.9.0.0.3266 is *useless*.

DanVanAtta:
> So, I just want to know exactly, since I'm only updating the 1.8.0.9 to 1.9.0.0.3266 information, should they tell to change "XOR" to "1" or should they not? I need to know this.

Probably, but I don't think it's changed since 1.9.0.0.3266. The way to really know is to fire up a game with both configurations and see if there is a difference when the condition is evaluated and see if XOR actually crashes the game.

Cernelius:
I was just suggesting keeping the 1.8.0.9 to 1.9.0.0.3266 somewhere out of the way (a legacy folder). Having a 1.8.0.9 to current would be good and arguably mandatory (for the future, one could add the rule that PR cannot be merged until after updating it, if need be).

DanVanAtta:
Legacy documentation hides useful documentation. It comes up in search results, which makes current documentation harder to keep up to date. Beyond that, legacy is legacy, it's not useful. All data, text, code, is an active cost. It costs time to filter from search results, to organize, to read, etc.. 

If we want to archive the document in a legacy way, put it in a zip file, attach it to an issue, then close the issue. We did that with the JavaFX code, that's how you do it. The text that has been committed will also forever be in history and can always be retreived.

>  Having a 1.8.0.9 to current would be good and arguably mandatory

Yeah, I'm saying we do our best to make one document that is "old" -> "new". If we have multiple versions, then we have duplication. For example, let's say there is a 2.0 -> 3.0, now you have to document those updates in two places or you have to have people dig through all of the legacy documentation to find each migration guide. It's much easier to have one than N

DanVanAtta:
@Cernelius the section "### Changes of already deprecated game (XML) file items:" is already not consistent with a conversion guide, as that is reaching to before 1.8

Cernelius:
> @Cernelius the section "### Changes of already deprecated game (XML) file items:" is already not consistent with a conversion guide, as that is reaching to before 1.8

That just splits the items between those that were already deprecated in 1.8.0.9 and those that were not (under my understanding). It doesn't necessarily reach before 1.8.0.9, as it says nothing whether the deprecated items were already deprecated before 1.8.0.9 or has been deprecated since the release of 1.8.0.9.

DanVanAtta:
@Cernelius I'm confused. When reading " that were already deprecated in 1.8.0.9", that says to me they were deprecated before 1.8.0.9. Hence my point, this document is not just the 1.8 to 1.9 migration.

Putting that to the side, pedantics aside, future changes to XML need a place to be documented. You haven't responded to my half dozen points about why multiple documents would be less than ideal. Please respond directly to those if you really think we should have multiple documents for every XML migration that occurs (and please consider how many of these we could have going forward, particularly if the pace of XML changes picks up).

Cernelius:
> @Cernelius I'm confused. When reading " that were already deprecated in 1.8.0.9", that says to me they were deprecated before 1.8.0.9. Hence my point, this document is not just the 1.8 to 1.9 migration.
> 
> Putting that to the side, pedantics aside, future changes to XML need a place to be documented. You haven't responded to my half dozen points about why multiple documents would be less than ideal. Please respond directly to those if you really think we should have multiple documents for every XML migration that occurs (and please consider how many of these we could have going forward, particularly if the pace of XML changes picks up).

I've a hard time understanding what you don't understand, since you did the changes.

"Were already deprecated in 1.8.0.9" means that. It doesn't say anything whether or not they were were deprecated in 1.8.0.8 or whatever.

I'll try to explain again.

A bunch of things that existed in 1.8.0.9 were removed from the programs. Some of those things were deprecated items, some others were not. So you have some items that went from being deprecated to being removed and some non-deprecated items that were removed. The deprecated items are further split into effective and ineffective. I hope now it is clear.

Cernelius:
> Putting that to the side, pedantics aside, future changes to XML need a place to be documented. You haven't responded to my half dozen points about why multiple documents would be less than ideal. Please respond directly to those if you really think we should have multiple documents for every XML migration that occurs (and please consider how many of these we could have going forward, particularly if the pace of XML changes picks up).

I'm not against you having a single document that tells how to upgrade maps since TripleA 1.0 (or whatever point you wish to start), if you think that can be workable. Sometimes it happens that someone still wants to upgrade some old maps made before Triplea 1.3 or something.

DanVanAtta:
That clarifies it, thanks @Cernelius .

PR #7635, opened 2020-09-12
Update upgrading-1-8-maps.md

Cernelius:

### DIFF
--- a/docs/map-making/upgrading-1-8-maps.md
+++ b/docs/map-making/upgrading-1-8-maps.md
@@ -1,88 +1,147 @@
 A list of (case sensitive) changes, to upgrade a map from a TripleA version to another one.
 
 ## Upgrade from 1.8.0.9 to 1.9.0.0.3266
-(old:new)
+
 Note: The following list covers all basic upgrade cases for maps, but it may not upgrade "Property" options (playerProperty, unitProperty...) of already deprecated items.
+
 Note: Before applying the changes below (assuming you do it with a simple replace), you should make sure to not have unnecessary spaces inside angle brackets (chevrons).
-Note: Some items are not validated; thus, if your game plays on 1.9.0.0.3266, this doesn't necessarily mean that you correctly upgraded it.
 
-### Set:
+Note: Some items are not validated; thus, if your game plays on 1.9.0.0.3266, this doesn't necessarily mean that you correctly upgraded it (for more information, see the "warning" section).
 
-`<triplea minimumVersion="1.9.0.0"/>`
+### New XML settings:
 
-### Not already deprecated xml changes:
-_(you normally need to change these only)_
+Within every game file (XML), inside the "games" folder inside the main folder of the map, change all occurrences of the code before the colon with the code after the colon.
+
+`<triplea minimumVersion="*"/>`:`<triplea minimumVersion="1.9.0.0"/>`
+
+(The "*" means whatever)
+
+### Not already deprecated XML changes:
+
+_(these are the main changes on game file items that were not deprecated, so you normally need to change these only, within the game file)_
+
+Within every game file (XML), inside the "games" folder inside the main folder of the map, change all occurrences of the code before the colon with the code after the colon.
 
 `attatchment`:`attachment`
+
 `Attatchment`:`Attachment`
+
 `attatchTo=`:`attachTo=`
+
 `="isImpassible"`:`="isImpassable"`
+
 `="conditionType" value="XOR"`:`="conditionType" value="1"`
-`="turns":="rounds"`
+
+`="turns"`:`="rounds"`
+
 `="Battleships repair at end of round"`:`="Units Repair Hits End Turn"`
+
 `="Battleships repair at beginning of round"`:`="Units Repair Hits Start Turn"`
+
 `="Naval Bombard Casualties Return Fire Restricted"`:`="Naval Bombard Casualties Return Fire"`
 
-### Special code xml changes for the above changes:
-_(you probably don't have these)_
+### Special code XML changes for the above changes:
+
+_(these are cases of advanced coding, so you probably don't have these)_
+
+Within every game file (XML), inside the "games" folder inside the main folder of the map, change all occurrences of the code before the colon with the code after the colon.
 
 `value="conditionType" count="XOR"`:`value="conditionType" count="1"`
+
 `value="conditionType" count="-reset-XOR"`:`value="conditionType" count="-reset-1"`
 
-### Already deprecated but effective xml changes:
-_(you shouldn't have had these in your games already)_
+### Already deprecated but effective XML changes:
+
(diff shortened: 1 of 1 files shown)


docs/map-making/upgrading-1-8-maps.md
```diff
@@ -1,88 +1,147 @@
...
+### Not already deprecated XML changes:
 
-`<triplea minimumVersion="1.9.0.0"/>`
+_(these are the main changes on game file items that were not deprecated, so you normally need to change these only, within the game file)_
 
-### Not already deprecated xml changes:
-_(you normally need to change these only)_
+Within every game file (XML), inside the "games" folder inside the main folder of the map, change all occurrences of the code before the colon with the code after the colon.
 
 `attatchment`:`attachment`
```
DanVanAtta, line 25: pre-pending items with a dash or asterisk will create a list, I think that might be the formatting looked for. Adding in the extra line breaks works too, though not as compact.

docs/map-making/upgrading-1-8-maps.md
```diff
@@ -1,88 +1,147 @@
...
+
 `="Naval Bombard Casualties Return Fire Restricted"`:`="Naval Bombard Casualties Return Fire"`
 
-### Special code xml changes for the above changes:
-_(you probably don't have these)_
+### Special code XML changes for the above changes:
+
+_(these are cases of advanced coding, so you probably don't have these)_
+
+Within every game file (XML), inside the "games" folder inside the main folder of the map, change all occurrences of the code before the colon with the code after the colon.
```
DanVanAtta, line 47: I wonder if we can state this just once somewhere at the beginning of the document. Keeping things short and concise is valuable and makes it take less time to use a document (which makes it more useful).

docs/map-making/upgrading-1-8-maps.md
```diff
@@ -1,88 +1,147 @@
...
+
 `="occupiedTerrOf"`:`="originalOwner"`
+
 `="victoryCity" value="true"`:`="victoryCity" value="1"`
+
 `="victoryCity" value="false"`:`="victoryCity" value="0"`
 
-### Already deprecated and ineffective xml deletions, for all the respective attachments:
-_(you shouldn't have had these in your games already and, if you did, they were doing nothing)_
+### Already deprecated and ineffective XML deletions, for all the respective attachments:
```
DanVanAtta, line 69: " ineffective XML deletions, " is a bit odd. I'm not actually sure what that means.

docs/map-making/upgrading-1-8-maps.md
```diff
@@ -1,88 +1,147 @@
...
+
+_(these items were already deprecated and were also ineffective, so you shouldn't have had these in your games already and, if you did, they were doing nothing)_
+
+Within every game file (XML), inside the "games" folder inside the main folder of the map, delete the whole `<option/>` containing the code.
```
DanVanAtta, line 73: The requirement for XML to be in a games folder is being loosened. It might be just as well to say "for every game file (XML)", that might be enough. Removing the extra, unnecessary specification, or just keeping it to a minimum can sometimes help keep things from getting out of date.

DanVanAtta: Perhaps you can simply say:


Delete these items (they were doing nothing):
`<option name="isParatroop" value="..."/>`

I wonder if just giving the full XML example might make it more easier and reduce the amount of pre-cursor text to explain what they are referring to.

DanVanAtta: In such a case, this document could be perhaps a bit simpler and just more explicit:

```
## Change these attributes

- <option name="isParatroop" value="..."/> -> <option name="isAirTransportable" value="..." />

## Delete these options (they do not do anything)

- `<option name="..." value="..." />`
```

DanVanAtta (approved):
Some commentary & feedback for you to consider. Thank you though for working on the formatting and clarity, it helps :+1:

docs/map-making/upgrading-1-8-maps.md
Cernelius: Not really. There are some items that are not game file (XML) codes and that are not changes (you could present them with an empty "new" version, to mean that they need to be removed (substituted with nothing), but that would be unclear).

For the format of "old" and "new" set apart by a colon ("change all occurrences of the code before the colon with the code after the colon"), that could be specified once (and that is what I was doing before this set of changes), but then I feel like the phrase where you tell the specific stuff is hardly meaningful.

docs/map-making/upgrading-1-8-maps.md
Cernelius: Well, it is the code being deleted that it was deprecated and ineffective. I guess it could be explained better, but this is the same compact format I used in the other places for titles. I would need to make all titles longer, I think.

Would you change "Already deprecated and ineffective XML deletions, for all the respective attachments" with "Deletions of already deprecated and ineffective XML items, for all the respective attachments"? If I do this, I would also change most other tiles, making them clearer but less compact. No problem.

docs/map-making/upgrading-1-8-maps.md
Cernelius: For now, this document is substantially a legacy document for upgrading from 1.8.0.9 to 1.9.0.0.3266, and has never been updated past 1.9.0.0.3266 (I believe it misses several items that were made after that version, but I don't remember what). So the wording should accord to that.

---

Apparently, I'm still not being clear enough. If you change "isParatroop" to "isAirTransportable" you are NOT updating the map: You are changing it. As far as I know, the "isParatroop" option was doing nothing. I assume it was supported doing nothing just to avoid the game crashing if present.

Feel free to propose changes to how I explained the items at the "already deprecated and ineffective XML deletions, for all the respective attachments" section. I'm not sure how to improve it, while keeping it fairly concise. This is the explanation I gave:

>(these items were already deprecated and were also ineffective, so you shouldn't have had these in your games already and, if you did, they were doing nothing)
>
>Within every game file (XML), inside the "games" folder inside the main folder of the map, delete the whole <option/> containing the code.

As far as I know, having "isParatroop" equals having nothing. It doesn't equal having a same valued "isAirTransportable" (unless the value is false, of course).

DanVanAtta: It looks like isParatroop did have an effect, it was only made into a no-op after we converted the maps to have 'isAirTransportable'

DanVanAtta: But the property was still left in as a value though to avoid game crashes; and at that point is made into a no-op, seemingly a messy transition there..

DanVanAtta: Hmm.. looking deeper, the code that did the handling of the property might have lead to a no-op later. This is twisty stuff, it's hard to say what it actually did without investigating for quite some time some really old code... So I'm not sure if it did anything or not.

PR #7637, opened 2020-09-13
Update: Allow unlimited default player colors & add a default for 'color.Impassable'

DanVanAtta:
commit 2ab58896ff192dc4ff1f15ad4b919d69601ae018

    Rename 'DefaultColors' -> 'PlayerColors'

commit 04b7551b3025a8139a70182f68f7be8db7ebb627

    Add a utility to construct colors 'ColorUtils'

commit a1b3d0239352f61367c833c68004c9775eef0e12

    Allow unlimited default player colors, default the impassable player color, and ensure stable colors
    
    - Cache playerColors such that default player colors are stable. Requesting the color
      of a player will always return the same color.
    
    - Generate default player colors if we exhaust the default player color list. Once
      we exceed the colors on the list, start to generate colors randomly.
    
    - Generate default (random) player colors in a stable manner, always generate
      the same random default player color per player name (use a fixed random seed
      based on player name).

commit c4e1d366bedbda8169d0ac89edd7ab1ecbeca200

    Clean up  ColorUtils methods.

## Release Note
It's not uncommon for map-makers to run out default player colors and run into errors. This update resolves that problem by generating additional default colors:
- https://forums.triplea-game.org/topic/1946/no-such-element-exception
- https://forums.triplea-game.org/topic/2201/map-not-loading-correctly
- https://forums.triplea-game.org/topic/2189/error-loading-my-new-ish-map

UPDATE|Generate an unlimited number of default player colors when player colors are not specified in map.properties. Use default color '#DEB887' for the 'impassable' player color if not specified in map.properties.

### DIFF
--- a/java-extras/src/main/java/org/triplea/java/ColorUtils.java
+++ b/java-extras/src/main/java/org/triplea/java/ColorUtils.java
@@ -20,7 +20,11 @@ public class ColorUtils {
       return new Color(Integer.decode("0x" + colorString));
     } catch (final NumberFormatException nfe) {
       throw new IllegalArgumentException(
-          "Colors must be 6 digit hex numbers, eg FF0011, not: " + colorString);
+          "Colors must be 6 digit hex numbers, eg FF0011, not: "
+              + colorString
+              + ", "
+              + nfe.getMessage(),
+          nfe);
     }
   }
 
--- a/java-extras/src/main/java/org/triplea/java/ColorUtils.java
+++ b/java-extras/src/main/java/org/triplea/java/ColorUtils.java
@@ -28,11 +28,6 @@ public class ColorUtils {
     }
   }
 
-  /** Returns a randomly generated color. */
-  public Color randomColor() {
-    return randomColor(123L);
-  }
-
   /** Returns a randomly generated color using a fixed random seed. */
   public Color randomColor(final long randomSeed) {
     final Random random = new Random(randomSeed);
--- a/java-extras/src/main/java/org/triplea/java/ColorUtils.java
+++ b/java-extras/src/main/java/org/triplea/java/ColorUtils.java
@@ -24,13 +24,14 @@ public class ColorUtils {
     }
   }
 
+  /** Returns a randomly generated color. */
+  public Color randomColor() {
+    return randomColor(123L);
+  }
+
+  /** Returns a randomly generated color using a fixed random seed. */
   public Color randomColor(final long randomSeed) {
     final Random random = new Random(randomSeed);
     return Color.getHSBColor(random.nextFloat(), random.nextFloat(), random.nextFloat());
-  }
-
-  /** Returns a randomly generated color. */
-  public Color randomColor() {
-    return Color.getHSBColor((float) Math.random(), (float) Math.random(), (float) Math.random());
   }
 }
--- a/java-extras/src/main/java/org/triplea/java/ColorUtils.java
+++ b/java-extras/src/main/java/org/triplea/java/ColorUtils.java
@@ -1,0 +1,20 @@
+package org.triplea.java;
+
+import com.google.common.base.Preconditions;
+import java.awt.Color;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class ColorUtils {
+  public Color fromHexString(final String colorString) {
+    Preconditions.checkArgument(
+        colorString.length() == 6,
+        "Colors must be 6 digit hex numbers, eg FF0011, not: " + colorString);
+    try {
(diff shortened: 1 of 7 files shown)


java-extras/src/main/java/org/triplea/java/ColorUtils.java
```diff
@@ -0,0 +1,37 @@
...
+   * @param colorString EG: 00FF00, FF00FF, 000000
+   */
+  public Color fromHexString(final String colorString) {
+    Preconditions.checkArgument(
+        colorString.length() == 6,
+        "Colors must be 6 digit hex numbers, eg FF0011, not: " + colorString);
+    try {
+      return new Color(Integer.decode("0x" + colorString));
+    } catch (final NumberFormatException nfe) {
+      throw new IllegalArgumentException(
```
DanVanAtta, line 22: Codacy found an issue: [New exception is thrown in catch block, original stack trace may be lost](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6170150)

java-extras/src/main/java/org/triplea/java/ColorUtils.java
```diff
@@ -0,0 +1,41 @@
...
+          "Colors must be 6 digit hex numbers, eg FF0011, not: "
+              + colorString
+              + ", "
+              + nfe.getMessage(),
+          nfe);
+    }
+  }
+
+  /** Returns a randomly generated color. */
+  public Color randomColor() {
```
RoiEXLab, line 32: The name might be a bit misleading, it's more like an arbitrary number, also it doesn't look like this method is used at all.

java-extras/src/main/java/org/triplea/java/ColorUtils.java
```diff
@@ -0,0 +1,41 @@
...
+    return randomColor(123L);
+  }
+
+  /** Returns a randomly generated color using a fixed random seed. */
+  public Color randomColor(final long randomSeed) {
+    final Random random = new Random(randomSeed);
+    return Color.getHSBColor(random.nextFloat(), random.nextFloat(), random.nextFloat());
+  }
```
RoiEXLab, line 40: That's a really good idea 👍

java-extras/src/main/java/org/triplea/java/ColorUtils.java
DanVanAtta: It was used, then I thought it made sense to keep the colors "sticky" so that games would be consistent. At a library/API level, I wouldn't want someone to use the `randomColor(long randomSeed)` API unless that is really what they wanted. I'm concerned someone would see that and either send a random seed because they had to or not use this API. My hope is that if you have a no-arg API, that would fit what you need. I don't think this is enough code to quite fit YAGNI given the library context, so I'm leaning we clean it up later if it's not used after a few months.

DanVanAtta: In what sense is the name misleading, I'm not quite sure I see that. Could you help me understand that perspective?

java-extras/src/main/java/org/triplea/java/ColorUtils.java
DanVanAtta: Thanks :smile:

java-extras/src/main/java/org/triplea/java/ColorUtils.java
RoiEXLab: If I read "random" I expect a random result i.e. the result differs every time.
I think you can't tell that this method always returns the same arbitrary color from the name and javadoc.

DanVanAtta: Ah, but be careful, the sequence of results will be the same if you restart the game, but that sequence will be random. If for example you request random colors in a different order for player names, the result would depend on the ordering of player names.

The other API is deterministic by player name because we use the player name hashcode for the random seed which will give us a fixed sequence for that player name (and we only grab the first element from that sequence).

Is it still your opinion that the method name should change or be clarified?

RoiEXLab: I could be missing something, but right now this method has "123" hardcoded as the seed, so if it were ever to be called the color would always be the exact same

DanVanAtta: Ah yeah, it would be exactly the same every time, that's a bug : )
I was thinking that the random generator would be new'ed up once rather than each time.

DanVanAtta: Removed in: dc9d6b0d9

PR #7643, opened 2020-09-14
Remove 'isBombing' boolean, instead infer it from BattleType

DanVanAtta:
This update adds flags to 'BattleType' that can be used to directly
infer if a battle type is bombing or an air battle. This has a
relation to the 'isBombing' flag that is passed around into methods,
so rather than using methods like "boolean isBombing, BattleType type",
we can replace that with "type.isBombing, type".

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/AbstractProAi.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/AbstractProAi.java
@@ -441,7 +441,7 @@ public abstract class AbstractProAi extends AbstractAi {
     final GamePlayer player = this.getGamePlayer();
     final BattleDelegate delegate = DelegateFinder.battleDelegate(data);
     final IBattle battle =
-        delegate.getBattleTracker().getPendingBattle(scrambleTo, false, BattleType.NORMAL);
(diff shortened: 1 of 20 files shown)


trevan:
[MiddleSubBattle1.tsvg.zip](https://github.com/triplea-game/triplea/files/5215672/MiddleSubBattle1.tsvg.zip)
If you load the attached save, you'll see oddities in the battle listing.

Screen Shot 2020-09-13 at 7 44 17 PM

To replicate, create a save using master in the combat phase with at least 2 battles.  Then load that save in this branch.

DanVanAtta:
Thanks for checking this out. That change is coming from the tostring of
battletype. It'll need to be fixed (perhaps changed to "toDisplayText" to
make it clear what it's for)

RoiEXLab (approved):
👍

RoiEXLab:
The build is failing due to pmd violations

trevan:
@DanVanAtta , just found another string oddity.  If you have a dependent battle (such as an amphibious attack after a naval battle) and you try to run the amphibious attack first, it says "Must complete IBattle.BattlType.NORMAL(type=Battle)"

PR #7650, opened 2020-09-14
Update map upgrade doc

DanVanAtta:
cc/ @Cernelius 

commit 83d9e40f63133ff050e7c728dc2bb3c657f01e2a

    Update: Formatting - 1.8 map upgrade doc
    
    - line breaks
    - convert 'colon' delimter to double arrow ">>"
    - use lists

commit 65d44b46a351af8c31a5d25488aa24eb2c6ec250

    Update: Add title, minor content pruning

### DIFF
--- a/docs/map-making/upgrading-1-8-maps.md
+++ b/docs/map-making/upgrading-1-8-maps.md
@@ -1,4 +1,4 @@
-# Upgrading Legacy Map Options & Properties
+# Upgrading Legacy Maps
 
 A list of (case sensitive) changes to upgrade legacy TripleA Game XML properties and options
 to the most recent.
@@ -34,7 +34,7 @@ Within every game (XML) file, inside the "games" folder inside the main folder o
 - `attatchment` >> `attachment`
 - `Attatchment` >> `Attachment`
 - `attatchTo=` >> `attachTo=`
-- `="isImpassible"`>> `="isImpassable"`
+- `="isImpassible"` >> `="isImpassable"`
 - `="conditionType" value="XOR"` >> `="conditionType" value="1"`
 - `="turns"` >> `="rounds"`
 - `="Battleships repair at end of round"` >> `="Units Repair Hits End Turn"`
@@ -78,7 +78,7 @@ Within every game file (XML), inside the "games" folder inside the main folder o
 - `="victoryCity" count="false"` >> `="victoryCity" count="0"`
 - `="victoryCity" count="-reset-false"` >> `="victoryCity" count="-reset-0"`
 
-### Deletions
+### XML Deletions
 
 _(these items were already deprecated and did nothing)_
 
--- a/docs/map-making/upgrading-1-8-maps.md
+++ b/docs/map-making/upgrading-1-8-maps.md
@@ -1,8 +1,10 @@
-A list of (case sensitive) changes, to upgrade a map from a TripleA version to another one.
+# Upgrading Legacy Map Options & Properties
 
+A list of (case sensitive) changes to upgrade legacy TripleA Game XML properties and options
+to the most recent.
+ 
 When changes are required, all occurrences of the code before the `>>` must be substituted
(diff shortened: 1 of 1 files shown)


DanVanAtta:
@Cernelius given the upcoming deprecations of XML elements that will have no effect, I"m pretty eager to get this document to a place where we can expand on it to include those. I notice that we already have a good bit of "these items were deleted before 1.8, so you should not have them". I'm convinced we want a single document to as best as possible document XML migrations.

docs/map-making/upgrading-1-8-maps.md
```diff
@@ -1,14 +1,22 @@
 A list of (case sensitive) changes, to upgrade a map from a TripleA version to another one.
 
-When changes are required, all occurrences of the code before the colon must be substituted with the code after the colon (if there is no code after the colon, the code must be substituted with nothing (deleted)). In any code, the "\*" symbol means whatever (the change or deletion is to be applied to every given code having whatever character or characters in place of the "\*").
+When changes are required, all occurrences of the code before the `>>` must be substituted
+with the code after the `>>` (if there is no code after the `>>`, the code must be substituted
+with nothing (deleted)). In any code, the "\*" symbol means whatever (the change or deletion is
+to be applied to every given code having whatever character or characters in place of the "\*").
```
Cernelius, line 9: You have removed all cases in which "there is no code after the `>>`", to mean deletions, so that's now a pointless thing to say.

docs/map-making/upgrading-1-8-maps.md
```diff
@@ -18,158 +26,128 @@ Within every game (XML) file, inside the "games" folder inside the main folder o
...
-
-`="Battleships repair at beginning of round"`:`="Units Repair Hits Start Turn"`
-
-`="Naval Bombard Casualties Return Fire Restricted"`:`="Naval Bombard Casualties Return Fire"`
-
-`="SBR Affects Unit Production"`:`="Damage From Bombing Done To Units Instead Of Territories"`
+- `attatchment` >> `attachment`
+- `Attatchment` >> `Attachment`
+- `attatchTo=` >> `attachTo=`
+- `="isImpassible"`>> `="isImpassable"`
```
Cernelius, line 37: Here the format is missing spacing before the ">>" (compared to the rest).

docs/map-making/upgrading-1-8-maps.md
```diff
@@ -1,14 +1,22 @@
 A list of (case sensitive) changes, to upgrade a map from a TripleA version to another one.
 
-When changes are required, all occurrences of the code before the colon must be substituted with the code after the colon (if there is no code after the colon, the code must be substituted with nothing (deleted)). In any code, the "\*" symbol means whatever (the change or deletion is to be applied to every given code having whatever character or characters in place of the "\*").
+When changes are required, all occurrences of the code before the `>>` must be substituted
```
Cernelius, line 6: I preferred ":", instead of ">>". I'm not sure if ">>" is meaningful for reasons I'm not aware. It just makes me thing to "skip forward". Alternatively to the ":", I might have chosen the "|", as symbol. Just stating my personal preference.

RoiEXLab (approved):
No problem from my side, leaving this open to discuss this with @Cernelius

docs/map-making/upgrading-1-8-maps.md
DanVanAtta: The `:` was really hard to see. The double ">>" is an arrow. A single ">" looks like an inequality sign, and an actual arrow "->" rendered kinda weird.

docs/map-making/upgrading-1-8-maps.md
DanVanAtta: Quite the eagle eye :eagle:   :smile:  :+1:

DanVanAtta:
@Cernelius I"m going to merge these updates after having done some follow-up, thanks for the review and feedback. I'm planning another iteration that could address your larger points.

PR #7651, opened 2020-09-14
Update: Automatically infer game XML property 'types'

DanVanAtta:
This update makes the `<String/>` and `<Boolean/>` child
nodes of `<Property>` to no longer be read. Instead we now
always infer the value type from the property value attribute
(eg: `<property value=...>`).

For legacy purposes we keep the `<Number min="..." max="..."/>` tag
as a fallback so existing maps would continue to work.

Forum discussion:
https://forums.triplea-game.org/topic/2265/proposal-automatic-property-type-discovery-based-on-value

### DIFF
--- a/map-data/src/main/java/org/triplea/map/data/elements/PropertyList.java
+++ b/map-data/src/main/java/org/triplea/map/data/elements/PropertyList.java
@@ -4,6 +4,7 @@ import java.util.List;
 import lombok.Getter;
 import org.triplea.generic.xml.reader.annotations.Attribute;
 import org.triplea.generic.xml.reader.annotations.BodyText;
+import org.triplea.generic.xml.reader.annotations.LegacyXml;
 import org.triplea.generic.xml.reader.annotations.Tag;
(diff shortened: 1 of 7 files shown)


map-data/src/main/java/org/triplea/map/data/elements/PropertyList.java
```diff
@@ -14,33 +15,26 @@
...
+    @Attribute private String value;
+    @Attribute private Integer min;
+    @Attribute private Integer max;
 
     @Tag private Property.Value valueProperty;
-    @Tag private Property.Boolean booleanProperty;
-    @Tag private Property.String stringProperty;
-    @Tag private Property.Number numberProperty;
+
+    @Tag(names = "Number")
```
RoiEXLab, line 27: 👌

RoiEXLab (approved):

RoiEXLab:
@DanVanAtta The markdown formatter in your initial post mistakenly stripped the xml tags because it thought you were trying to inject HTML, I updated your comment accordingly

Issue #4688, opened 2019-02-16
Re-introduction of a retreat rule issue


panther2:
### Engine version
1.9.0.0.13066
1.10.14019

### Map name
ww2-global_1940

### Can you describe how to trigger the error? (eg: what sequence of actions will recreate it?)
Referring to https://github.com/triplea-game/triplea/issues/1270, an issue that had already been solved is back again.

TripleA should not offer the attacker's option to retreat, if after a round of combat only a defending AAA-gun is left. In the attached savegames (1.9/1.10) the remaining infantry is incorrectly offered a retreat option when facing only a "defending" AAA gun.

### Instead of this error, what should have happened?

The AAA gun is autodestroyed, the territory is taken by the enemy. There is nothing to retreat from.

### Any additional information that may help

@simon33-2 mentioned in https://www.axisandallies.org/forums/post/1258083 
> What about the retreat rules? Issue #1270. #2080 was reverted due to performance problems in #2544. So there’s still problems - if you attack and only have a transport or an AA Gun defending, you are still permitted a retreat.

While the transport scenario seemed to work fine in my test scenario, I could reproduce the AA Gun issue as mentioned above:

[savedGames.zip](https://github.com/triplea-game/triplea/files/2871875/savedGames.zip)

simon33-2:
Just pointing out there is a typo above where you say "destroyer" instead of "transport"

Also, there's a good improvement here potentially, that when you send a sub to kill an undefended transport, you shouldn't get a prompt to submerge the sub like you do in 1.9.

There's actually two ways of achieving this, I was thinking you'd need an incompatible release way (1) is to change the order so that undefended transports are removed before the submerge prompt (2) is to remove the prompt when only defenseless units are defending. 

Actually (2) may be better because in (1) you would still get the submerge prompt which would be annoying.

panther2:
Fixed that, thank you @simon33-2 .

ron-murhammer:
@panther2 It appears the check for submerging subs is done before the check for undefended transports. Would flipping the order of those 2 checks solve the issue? Or would that create a different problem?

panther2:
@redrum 
In case there are undefended transports the battle ends with the transports being destroyed and the subs remaining. So there shouldn't be an offer for the subs to submerge/retreat at all. If that was the (only) result of your idea it would be fine.

Regarding possible different problems: Is there anything else behind the check for undefended transports?

Cernelius:
[...]

PR #7660, opened 2020-09-15
Convert attacker retreat steps

trevan:
This converts the attacker retreat steps into a single battle step.

## Testing

Tested partial amphibious retreats and regular retreats.  Loaded an old save that was in the middle of a partial amphibious retreat.


game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1661,45 +1667,29 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
...
-                    attackingUnits, gameData, isAmphibious)) {
-              attackerRetreatNonAmphibUnits(bridge);
-            }
-          }
-        });
-    steps.add(
-        new IExecutable() {
-          private static final long serialVersionUID = 669349383898975048L;
+      @Override
+      public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {}
```
DanVanAtta, line 1676: Codacy found an issue: [Document empty method body](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6178813)

panther2:
@trevan 
Does that address #4688 and #4821, too?

trevan:
> @trevan
> Does that address #4688 and #4821, too?

@panther2 , no, this doesn't change any functionality of the game.  It is just refactoring the code.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1661,45 +1667,33 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
...
-              attackerRetreatNonAmphibUnits(bridge);
-            }
-          }
-        });
-    steps.add(
-        new IExecutable() {
-          private static final long serialVersionUID = 669349383898975048L;
+      @Override
+      public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
+        // Intentionally left blank
```
DanVanAtta, line 1677: Some background on why this is intentional could be useful. Why is this intentionally left blank?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  @Override
+  public Order getOrder() {
+    return Order.OFFENSIVE_GENERAL_RETREAT;
+  }
+
+  @Override
+  public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
+    retreatUnits(bridge);
```
DanVanAtta, line 50: It's really odd for `retreatUnits` to be the implementation of this interface method, and also called externally directly. Can you speak to why that is @trevan ?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
```diff
@@ -0,0 +1,129 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Value;
+import org.triplea.java.collections.CollectionUtils;
+
+@AllArgsConstructor
+public class OffensiveGeneralRetreat implements BattleStep {
+
+  final BattleState battleState;
+
+  final BattleActions battleActions;
```
DanVanAtta, line 30: Can these variables be made private?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
```diff
@@ -0,0 +1,129 @@
...
+
+  @Override
+  public List<String> getNames() {
+    if (canAttackerRetreat()
```
DanVanAtta, line 34: Nitpicky - Even if a multi-line, I still think this would be cleaner as a ternary. I think with context, it seems easier, but I would suggest that maintaining/reading the code, it's easier to grok as a ternary. Perhaps the boolean can be extracted to a helper method to make it a simple tenary. WDYT?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
```diff
@@ -0,0 +1,129 @@
...
+    return Order.OFFENSIVE_GENERAL_RETREAT;
+  }
+
+  @Override
+  public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
+    retreatUnits(bridge);
+  }
+
+  public void retreatUnits(final IDelegateBridge bridge) {
+    if (battleState.isOver()) {
```
DanVanAtta, line 54: Does this condition happen often? TBH, it seems a bit like a bug. If the battle is over, the caller perhaps did not realize that and should not have made this call to begin with.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
```diff
@@ -0,0 +1,129 @@
...
+        battleState.isAmphibious());
+  }
+
+  private boolean canAttackerRetreatSeaPlanes() {
+    return battleState.getBattleSite().isWater()
+        && battleState.getUnits(BattleState.Side.OFFENSE).stream().anyMatch(Matches.unitIsAir());
+  }
+
+  private boolean canAttackerRetreatPartialAmphib() {
+    if (Properties.getPartialAmphibiousRetreat(battleState.getGameData())) {
```
DanVanAtta, line 108: Have you consider pushing this conditional to the callers? I think in both cases it would be the calling code easier to understand, it avoids having to jump around the class file to understand those locations. Furthermore, without this if statement the method body would be a simple stream & any match statement. Not that it matters too much, there would be a small performance boost as you wouldn't need to compute the full set of land units but could halt on the first match.

DanVanAtta:
Did a first pass, looks pretty solid @trevan ; Will try to come back to this tomorrow to finish the review.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: For some reason, the retreat steps were broken up into three separate IExecutables even though only one of them would ever run.  I merged all of them into one step and so two of the old IExecutables no longer have anything to do.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
trevan: The same logic (retreat) is called in other steps.  I either had the choice of creating a static helper class to store the logic or to allow the other steps to instantiate and call this method directly.

I would prefer that the other steps didn't need this logic and hope that I might be able to fix it but that will involve changing the logic of the steps and might even change the game play.  I'd like to make that change in a separate PR.

DanVanAtta: Fair enough. Extracting logic for re-use would seemingly be okay. Passing an instance to a constructor can help avoid static coupling.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
trevan: Yes.  How the stack work is that at the beginning of the round, all of the IExecutables/BattleSteps are created and added to the stack.  Then, it goes through each of the items in the step.  There isn't currently a way to short circuit the stack when the battle is over.  So, this step already is in the stack and will execute even if the battle finished in a previous step.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
trevan: I think it is better if the caller doesn't have to know what properties enable/disable partial amphibious retreat.

I've made the stream/anyMatch change, though.

DanVanAtta: Hmm, it's a private method though.

Generally it's better to get the conditional logic done at a higher layer. So you have something that looks like this:

```
public void method() {
  if(a || b) {
        doFoo();
         if(a1) {
               doBar()
          }
  } else if(c) {
      doZed();
  }
}
```

vs:

```
public void method() {
   doFoo();
   doBar();
   doZed();
}

private void doFoo() {    
    if(!a && !b && c) {
        doZed();
    } else if(!a && !b) {
        return;
    }
    doBar();
}

private void doBar() {
    if(!a1) {
        return;
    }
}
```

Notice that above the control flow is pushed down and it's very hard to follow what the logic is.

The idea is 3 part:
1. Avoid lists of void method calls
2. Choose where the control flow will be, have the method with control flow just do mainly that and the things it invokes should be relatively functional.
3. Push control flow to higher layers, avoid doing control flow in private methods That indicates you perhaps want a strategy object to handle the behavior.

DanVanAtta: I'm not insisting any changes be made here, but I don't agree that having control flow at a lower layer like this makes it easier. A reader of the code needs to jump to this method anyways and it's cognitively complex to keep track of control flow throughout  'n' layers of code.

DanVanAtta: Expouding on "needs to jump to this method anyways"

Let's say the reader knows that only "a" is true. In this case the following method:

```
public void method() {
  if(a || b) {
        doFoo();
         if(a1) {
               doBar()
          }
  } else if(c) {
      doZed();
  }
}
```

Requires a reader to only read `doFoo`. On the other hand, with the series of void call, a reader needs to go through each method, hence why I suggest often they may as well be inlined, as you have to read it all anyways. The fact you can't skip code (as in this example where you can skip 'doBar' and 'doZed'), is a break-down of information hiding. My 2 cents.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
DanVanAtta: Thanks for the context. I don't find that terribly satisfying, but I would accept that wiring a 'Predicate' for battle is over to the execution stack at this juncture is potentially not easy and out of scope of this update.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
DanVanAtta: Can you update the code with a comment or a 'remove on major release' annotation? This would help us to know what to do with this going forward.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
```diff
@@ -20,24 +19,26 @@
...
-  final BattleState battleState;
+  private final BattleState battleState;
 
-  final BattleActions battleActions;
+  private final BattleActions battleActions;
 
   @Override
   public List<String> getNames() {
-    if (canAttackerRetreat()
+    return isRetreatPossible()
```
DanVanAtta, line 32: :+1: 

This looks good, thanks!

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
```diff
@@ -105,18 +106,13 @@ private boolean canAttackerRetreatSeaPlanes() {
...
-        }
-      }
+    if (!Properties.getPartialAmphibiousRetreat(battleState.getGameData())) {
+      return false;
     }
-    return false;
+    // Only include land units when checking for allow amphibious retreat
+    return battleState.getUnits(BattleState.Side.OFFENSE).stream()
+        .filter(Matches.unitIsLand())
+        .anyMatch(unit -> !unit.getWasAmphibious());
```
DanVanAtta, line 115: nitpicky, note - you can use Predicates here to write this as a method references:
```
.anyMatch(Predicates.not(Unit::getWasAmphibious));
```

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
trevan: I think I understand what you are saying but this entire method is control flow.  So it is like:
```
public void method() {
  if (isAllowed()) {
    doFoo();
  }
}

private boolean isAllowed() {
  return a || (b && c) || d;
}
```
vs
```
public void method() {
  if (a || (b && c) || d) {
    doFoo();
  }
}
```

You just recommended that I do the former in `getNames()` (with the the new `isRetreatPossible()` method).  How is `canAttackerRetreatPartialAmphib()` any different?

Or is it because both `canAttackerRetreatPartialAmphib()` and `canAttackerRetreatAmphibPlanes()` checks for `Properties.getPartialAmphibiousRetreat` and so you want to move that call out of them?

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: I added the annotation to all of the `execute` methods of the old `IExecutables`.  I also add a larger comment to indicate why it is blank.

trevan:
The build failure is unrelated to my changes: `Exception in thread "main" javax.net.ssl.SSLProtocolException: Connection reset`

DanVanAtta:
marking for further review later, in the meantime I'm pretty sure there is nothing here that we can't simply follow up on later.

PR #7662, opened 2020-09-15
Map upgrade doc changes

DanVanAtta:

### DIFF
--- a/docs/map-making/how-to-upgrade-legacy-map-features.md
+++ b/docs/map-making/how-to-upgrade-legacy-map-features.md
@@ -1,35 +1,26 @@
 # How To Upgrade Legacy Map Features
 
-A list of (case sensitive) changes to upgrade legacy TripleA Game XML properties and options
-to the most recent.
+A list of changes to upgrade legacy TripleA Game XML properties and options to the most recent.
  
-When changes are required, all occurrences of the code before the `>>` must be substituted
-with the code after the `>>`. In any code, the "\*" symbol means whatever (the change or deletion is
-to be applied to every given code having whatever character or characters in place of the "\*").
-
-## Upgrade from 1.8.0.9 to 1.9.0.0.3266
-
-Note: The following list covers all basic upgrade cases for maps, but it may not upgrade "Property"
-options (playerProperty, unitProperty...) of ineffective items you must delete (such changes have
-been omitted from this document for brevity).
-
 Note: Before applying the changes below (assuming you do it with a simple replace), you should make
 sure to not have unnecessary spaces inside angle brackets (chevrons).
 
-Note: Some items are not validated. Thus, if the game plays without errors using TripleA 1.9.0.0.3266,
+Note: Some items are not validated. Thus, if the game plays without errors using a latest,
 this doesn't necessarily mean that you correctly upgraded it (for more information, see the "warning"
 section).
+
+
+## XML Changes
+
+These changes need to happen in your game XML files.
 
 ### New game (XML) file settings:
 
 `<triplea minimumVersion="*"/>`:`<triplea minimumVersion="1.9.0.0"/>`
 
-### Changes of not already deprecated game (XML) file items:
+### Changes:
 
-_(these are the main changes on game file items that were not deprecated, so you normally need to change
-these only, within the game file)_
-
-Within every game (XML) file, inside the "games" folder inside the main folder of the map, change:
+All occurrences of the code before the `>>` must be substituted with the code after the `>>`.
 
 - `attatchment` >> `attachment`
 - `Attatchment` >> `Attachment`
@@ -41,42 +32,19 @@ Within every game (XML) file, inside the "games" folder inside the main folder o
 - `="Battleships repair at beginning of round"` >> `="Units Repair Hits Start Turn"`
 - `="Naval Bombard Casualties Return Fire Restricted"` >> `="Naval Bombard Casualties Return Fire"`
 - `="SBR Affects Unit Production"` >> `="Damage From Bombing Done To Units Instead Of Territories"`
-
-### Changes of game (XML) file special code for the above items:
-
-_(these are cases of advanced coding, so you probably don't have these within the game file)_
-
-Within every game file (XML), inside the "games" folder inside the main folder of the map, change:
-
 - `="conditionType" count="XOR"` >> `="conditionType" count="1"`
 - `="conditionType" count="-reset-XOR"` >> `="conditionType" count="-reset-1"`
-
-### Changes of already deprecated game (XML) file items:
-
-_(these items were already deprecated, so you shouldn't have had these within the game file already)_
-
-Within every game (XML) file, inside the "games" folder inside the main folder of the map, change:
-
-- `="isTwoHit" value="true"` >> `="hitPoints" value="2"`
-- `="isTwoHit" value="false"` >> `="hitPoints" value="1"`
-- `="occupiedTerrOf"` >> `="originalOwner"`
-- `="victoryCity" value="true"` >> `="victoryCity" value="1"`
-- `="victoryCity" value="false"` >> `="victoryCity" value="0"`
-
-### Changes of game (XML) file special code for the above items:
-
-_(these are cases of advanced coding, so you probably don't have these within the game file)_
-
-Within every game file (XML), inside the "games" folder inside the main folder of the map, change:
-
 - `="isTwoHit" count="true"` >> `="hitPoints" count="2"`
+- `="isTwoHit" value="false"` >> `="hitPoints" value="1"`
 - `="isTwoHit" count="-reset-true"` >> `="hitPoints" count="-reset-2"`
 - `="isTwoHit" count="false"` >> `="hitPoints" count="1"`
 - `="isTwoHit" count="-reset-false"` >> `="hitPoints" count="-reset-1"`
+- `="occupiedTerrOf"` >> `="originalOwner"`
 - `="victoryCity" count="true"` >> `="victoryCity" count="1"`
 - `="victoryCity" count="-reset-true"` >> `="victoryCity" count="-reset-1"`
 - `="victoryCity" count="false"` >> `="victoryCity" count="0"`
 - `="victoryCity" count="-reset-false"` >> `="victoryCity" count="-reset-0"`
+
 
 ### XML Deletions
 
@@ -88,21 +56,21 @@ _(these items were already deprecated and did nothing)_
 - `<option name="giveUnitControl" value="true"/>`
 - `<option name="giveUnitControl" value="false"/>`
 
-### Changes of map (properties) file items:
+## map.properties Changes
 
 Within every "map" properties file, inside the main folder of the map, change:
 
 - `color.Impassible=` >> `color.Impassable=`
 
-### Changes of sounds (properties) file items:
-
-Within every "sounds" properties file, inside the main folder of the map, change:
-
-- `.wav` >> `.mp3`
-
-### Changes of non-code items:
+## Sound file changes
 
 Change all sounds inside the "sounds" main folder of the map from "wav" to "mp3" format.
+128Kbps and 168Kbps are the preferred bit rates.
+
+
+## More XML Changes
+
+### Changes of non-code items:
 
 ---
 
@@ -135,7 +103,8 @@ you can change:
 - `="rulesAttatchment` >> `="rulesAttachment`
 - `="playerAttatchment` >> `="playerAttachment`
 
-**Warning:**
+
+## Warnings
 
 The following xml codes, as being properties, are not validated (if you leave them, the game will play, but they will
 set nothing in the engine) (it is suggested you double check you don't have them within any game files, after
--- a/docs/map-making/how-to-upgrade-legacy-map-features.md
+++ b/docs/map-making/how-to-upgrade-legacy-map-features.md
@@ -46,14 +46,7 @@ All occurrences of the code before the `>>` must be substituted with the code af
 - `="victoryCity" count="-reset-false"` >> `="victoryCity" count="-reset-0"`
 
 
-### Changes of non-code items:
-
----
-
-_All that follows is supposed to be unnecessary additional information: If you correctly already made all that is
-described above, all that follows is irrelevant and can be ignored._
-
-**Procedural alternatives:**
+### **Procedural alternatives:**
 
 For a more defined change, instead of the aforementioned:
 
@@ -66,19 +59,12 @@ you can change:
 - `attatchment>` >> `attachment>`
 - `="games.strategy.triplea.attatchments` >> `="games.strategy.triplea.attachments`
 
-For a more defined change, instead of the aforementioned:
-
-- `Attatchment` >> `Attachment`
-
-you can change:
-
 - `="techAttatchment` >> `="techAttachment`
 - `="unitAttatchment` >> `="unitAttachment`
 - `="territoryAttatchment` >> `="territoryAttachment`
 - `="canalAttatchment` >> `="canalAttachment`
 - `="rulesAttatchment` >> `="rulesAttachment`
 - `="playerAttatchment` >> `="playerAttachment`
-
 
 
 ### XML Deletions
--- a/docs/map-making/how-to-upgrade-legacy-map-features.md
+++ b/docs/map-making/how-to-upgrade-legacy-map-features.md
@@ -46,30 +46,6 @@ All occurrences of the code before the `>>` must be substituted with the code af
 - `="victoryCity" count="-reset-false"` >> `="victoryCity" count="-reset-0"`
 
 
-### XML Deletions
-
-_(these items were already deprecated and did nothing)_
-
-- `<option name="isParatroop"*/>`
-- `<option name="isMechanized"*/>`
-- `<option name="takeUnitControl"*/>`
-- `<option name="giveUnitControl" value="true"/>`
-- `<option name="giveUnitControl" value="false"/>`
-
-## map.properties Changes
-
-Within every "map" properties file, inside the main folder of the map, change:
-
-- `color.Impassible=` >> `color.Impassable=`
-
-## Sound file changes
-
-Change all sounds inside the "sounds" main folder of the map from "wav" to "mp3" format.
-128Kbps and 168Kbps are the preferred bit rates.
-
-
-## More XML Changes
-
 ### Changes of non-code items:
 
 ---
@@ -104,16 +80,39 @@ you can change:
 - `="playerAttatchment` >> `="playerAttachment`
 
 
+
+### XML Deletions
+
+_(these items were already deprecated and did nothing)_
+
+- `<option name="isParatroop"*/>`
+- `<option name="isMechanized"*/>`
+- `<option name="takeUnitControl"*/>`
+- `<option name="giveUnitControl" value="true"/>`
+- `<option name="giveUnitControl" value="false"/>`
+
+## map.properties Changes
+
+Within every "map" properties file, inside the main folder of the map, change:
+
+- `color.Impassible=` >> `color.Impassable=`
+
+## Sound file changes
+
+Change all sounds inside the "sounds" main folder of the map from "wav" to "mp3" format.
+128Kbps and 168Kbps are the preferred bit rates.
+
+
 ## Warnings
 
-The following xml codes, as being properties, are not validated (if you leave them, the game will play, but they will
-set nothing in the engine) (it is suggested you double check you don't have them within any game files, after
-upgrading the map):
+The following xml codes, as being properties, are not validated (if you leave them, the game will play,
+but they will set nothing in the engine) (it is suggested you double check you don't have them within
+any game files, after upgrading the map):
 
 - `="Battleships repair at end of round"`
 - `="Battleships repair at beginning of round"`
 - `="Naval Bombard Casualties Return Fire Restricted"`
 - `="SBR Affects Unit Production"`
 
-While all options named "takeUnitControl" must be deleted, any option named "giveUnitControl" must be deleted only if
-its value is "true" or "false".
+While all options named "takeUnitControl" must be deleted, any option named "giveUnitControl" must be
+deleted only if its value is "true" or "false".


RoiEXLab (approved):

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -1,35 +1,26 @@
 # How To Upgrade Legacy Map Features
 
-A list of (case sensitive) changes to upgrade legacy TripleA Game XML properties and options
-to the most recent.
+A list of changes to upgrade legacy TripleA Game XML properties and options to the most recent.
```
Cernelius, line 3: As I said, here it is not clear whether the mentioned properties are the properties within the game file or the properties files of the map. Assuming that "TripleA Game XML properties and options" means the properties and options within the game (XML) file, this is not correct, as this "list of changes" is not only about that. That also would imply that this list of changes is useless for upgrading mapskins, while it actually contains information critical for such maps too.

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -1,35 +1,26 @@
...
  
-When changes are required, all occurrences of the code before the `>>` must be substituted
-with the code after the `>>`. In any code, the "\*" symbol means whatever (the change or deletion is
-to be applied to every given code having whatever character or characters in place of the "\*").
```
Cernelius, line 8: I assume you assume that the "*" symbol is itself meaningful enough for everyone that there is no need to explain it means something like "zero or more of any characters". If so, I tend not to agree. I believe that a code like
`<triplea minimumVersion="*"/>`:`<triplea minimumVersion="1.9.0.0"/>`
should better have an explanation about the fact that the original "minimumVersion" value can be whatever.

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -1,35 +1,26 @@
...
-
-## Upgrade from 1.8.0.9 to 1.9.0.0.3266
```
Cernelius, line 10: If you completely remove every reference of the TripleA versions from which and to which to upgrade, I assume this can only mean this is for upgrading every map feature. Meaning that someone having a map made for TripleA 1.0 should be able to upgrade it to TripleA 2.2 by following the instruction in this document. Even in this case, I would make explicit that the starting point is TripleA 1.0. In any case, since, as far as I can tell, this document is still seemingly about 1.8.0.9 to 1.9.0.0.3266 upgrades only, I would keep such reference until it is correct or, at the very least, clarifying that this document is an incomplete "work-in-progress", making very clear that people should not currently expect it giving full information for upgrading to the latest whatever map they may have (because I can see it is certainly not able to assure that).

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -1,35 +1,26 @@
...
-
-Note: The following list covers all basic upgrade cases for maps, but it may not upgrade "Property"
-options (playerProperty, unitProperty...) of ineffective items you must delete (such changes have
-been omitted from this document for brevity).
```
Cernelius, line 14: This should have not been removed. Since the document is not informing how to upgrade such items, I think it is good having a mention telling that such items are not covered by the information given. For example, if one has a map with a trigger changing the value of "giveUnitControl" from false to true, that item will not be touched, by following the information at this documentation literally (it would be needed manually to delete the whole trigger, since the option is useless both at the "true" and at the "false" values, thus the trigger is useless too). I decided not to provided such information as the case is remote, and merely noticing that "Property" options are not touched (to avoid confusion, I'll point out that property options are not properties (the term "property" in TripleA means at least three different and largely unrelated things, as far as I know)).

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -41,43 +32,20 @@ Within every game (XML) file, inside the "games" folder inside the main folder o
 - `="Battleships repair at beginning of round"` >> `="Units Repair Hits Start Turn"`
 - `="Naval Bombard Casualties Return Fire Restricted"` >> `="Naval Bombard Casualties Return Fire"`
 - `="SBR Affects Unit Production"` >> `="Damage From Bombing Done To Units Instead Of Territories"`
-
-### Changes of game (XML) file special code for the above items:
```
Cernelius, line 45: The reason why I created two "special code" sections is that such changes touch items that are very rarely found in games. Practically, only games having triggers changing values of options may need making such changes. Maybe the "so you probably don't have these within the game file" note could have been clarified by explaining that games with no triggers don't need having these changes applied to them (even though games with triggers are unlikely to have them too). Consolidating all in a single section is a reasonable preference, but this means having a lot of changes code that is very rarely useful, thus my preference of splitting between common and uncommon changes to make.

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -41,43 +32,20 @@ Within every game (XML) file, inside the "games" folder inside the main folder o
...
 - `="conditionType" count="XOR"` >> `="conditionType" count="1"`
 - `="conditionType" count="-reset-XOR"` >> `="conditionType" count="-reset-1"`
-
-### Changes of already deprecated game (XML) file items:
-
-_(these items were already deprecated, so you shouldn't have had these within the game file already)_
-
-Within every game (XML) file, inside the "games" folder inside the main folder of the map, change:
-
-- `="isTwoHit" value="true"` >> `="hitPoints" value="2"`
```
Cernelius, line 60: The new document is completely missing this item. I believe it has been removed by mistake, thus needs to be added back.

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -41,43 +32,20 @@ Within every game (XML) file, inside the "games" folder inside the main folder o
...
-- `="isTwoHit" value="false"` >> `="hitPoints" value="1"`
-- `="occupiedTerrOf"` >> `="originalOwner"`
-- `="victoryCity" value="true"` >> `="victoryCity" value="1"`
```
Cernelius, line 63: The new document is completely missing this item. I believe it has been removed by mistake, thus needs to be added back.

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -41,43 +32,20 @@ Within every game (XML) file, inside the "games" folder inside the main folder o
...
-- `="victoryCity" value="false"` >> `="victoryCity" value="0"`
```
Cernelius, line 64: The new document is completely missing this item. I believe it has been removed by mistake, thus needs to be added back.

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -41,43 +32,20 @@ Within every game (XML) file, inside the "games" folder inside the main folder o
...
-
-### Changes of game (XML) file special code for the above items:
-
-_(these are cases of advanced coding, so you probably don't have these within the game file)_
-
-Within every game file (XML), inside the "games" folder inside the main folder of the map, change:
-
 - `="isTwoHit" count="true"` >> `="hitPoints" count="2"`
+- `="isTwoHit" value="false"` >> `="hitPoints" value="1"`
```
Cernelius, line 38: This item, or any item, should better not stay in the middle of the "reset" and "non-reset" version of the code for turning the "isTwoHit" option value to true (the "reset" is a redundant coding, since the option accepts only one value, but it cannot be exclude a game might have it).

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -41,43 +32,20 @@ Within every game (XML) file, inside the "games" folder inside the main folder o
...
+- `="occupiedTerrOf"` >> `="originalOwner"`
 - `="victoryCity" count="true"` >> `="victoryCity" count="1"`
 - `="victoryCity" count="-reset-true"` >> `="victoryCity" count="-reset-1"`
 - `="victoryCity" count="false"` >> `="victoryCity" count="0"`
 - `="victoryCity" count="-reset-false"` >> `="victoryCity" count="-reset-0"`
 
+
 ### XML Deletions
 
 _(these items were already deprecated and did nothing)_
```
Cernelius, line 72: You previously removed a "were already deprecated" reference. I suggest either removing all or none of them. Also saying that "did nothing" is meaningless in the moment it is not documented when (this was supposed to mean "did nothing in 1.8.0.9", that being the referenced starting point).

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -88,21 +56,21 @@ _(these items were already deprecated and did nothing)_
...
 - `<option name="giveUnitControl" value="false"/>`
 
-### Changes of map (properties) file items:
+## map.properties Changes
 
 Within every "map" properties file, inside the main folder of the map, change:
 
 - `color.Impassible=` >> `color.Impassable=`
 
-### Changes of sounds (properties) file items:
```
Cernelius, line 97: This whole section has been removed without a valid substutite, so it is now a missing item, thus needs to be added back. To clarify, what this section was doing, was telling that if having a code like
`game_start=classical/game_start/game_start_02_gladiator_entrance.wav`
within the "sounds" properties file, the code has to be changed to
`game_start=classical/game_start/game_start_02_gladiator_entrance.mp3`.
This is not even necessarily related to the "sounds" folder, as, in the example, we are referencing code in the TripleA "assets" folder.

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -88,21 +56,21 @@ _(these items were already deprecated and did nothing)_
...
+## Sound file changes
 
-Within every "sounds" properties file, inside the main folder of the map, change:
+Change all sounds inside the "sounds" main folder of the map from "wav" to "mp3" format.
+128Kbps and 168Kbps are the preferred bit rates.
 
-- `.wav` >> `.mp3`
 
-### Changes of non-code items:
+## More XML Changes
```
Cernelius: Since, as explained, all changes that follow are not additional changes, but merely "unnecessary additional information", I believe this "More XML Changes" entry is confusing or even wrongly descriptive (since from this point onwards there are no more changes of any kind, just alternative ones and additional information).

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -88,21 +56,21 @@ _(these items were already deprecated and did nothing)_
...
 
-Change all sounds inside the "sounds" main folder of the map from "wav" to "mp3" format.
+### Changes of non-code items:
```
Cernelius: This line used to be before the information about changing the format of the sound files. The "non-code" items mentioned were only the needed change of substituting all "wav" files with "mp3" ones, which is now detailed before this line. I believe everything else are code changes. So this line does not make any sense to me (I don't get what are these "non-code" items, now). Moreover, since a section break follows, it seems to me that this is even saying the "non-code items" part is empty.

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -66,12 +59,6 @@ you can change:
 - `attatchment>` >> `attachment>`
 - `="games.strategy.triplea.attatchments` >> `="games.strategy.triplea.attachments`
 
-For a more defined change, instead of the aforementioned:
-
-- `Attatchment` >> `Attachment`
```
Cernelius, line 127: Assuming that all changes are supposed to be case sensitive (something that used to be explicitly stated) changing "attatchment" to "attachment" is a different change from changing "Attatchment" to "Attachment", so what follows are not procedural alternatives to changing "attatchment" to "attachment", thus the information now given is wrong.

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -66,12 +59,6 @@ you can change:
...
-
-you can change:
-
 - `="techAttatchment` >> `="techAttachment`
```
Cernelius, line 62: This is now given as an alternative of "`attatchment` >> `attachment`", which is wrong, since it is, instead, an alternative of "`Attatchment` >> `Attachment`". The same problem exist for all the five following lines.

DanVanAtta: This is an existing line @Cernelius ; can you clarify how that is "wrong" , can you give an example where following the instructions a reasonable person would make a mistake?

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: The 2.x engine is actually becoming case insenstive. I think the assumption is that the changes are going to be case sensitive because the casing is specified.

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: The existing was not sufficient that because you can't just change the file extension. There is now a mention that you need to re-encode the wav files. Mentioning that the files are referenced in the sound properties file would probably be good.

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: Please check again, this is present in the latest:

> ="victoryCity" count="true" >> ="victoryCity" count="1"
="victoryCity" count="-reset-true" >> ="victoryCity" count="-reset-1"
="victoryCity" count="false" >> ="victoryCity" count="0"
="victoryCity" count="-reset-false" >> ="victoryCity" count="-reset-0"

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: Side-note about review, it's not cool to send too many review comments. Try to group them up when possible and when there is a lot of commentary drop the least important comments.

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: Ditto, please check again:

> ="isTwoHit" count="true" >> ="hitPoints" count="2"
="isTwoHit" value="false" >> ="hitPoints" value="1"
="isTwoHit" count="-reset-true" >> ="hitPoints" count="-reset-2"
="isTwoHit" count="false" >> ="hitPoints" count="1"

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: The distinction is really not that important and adds a lot of verbiage for the sake of "yeah, you have to change these too, but you probably don't have them.". It's not that important of a distinction. What we're looking for in this list is a s simple as possible,  "change this, into that".

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: Hmm, the sentence needs work.  For example, the following sentence does not make sense:   "but it may not upgrade "Property"options " . 

We do have a section of "delete" these items. Perhaps we're only saying that this document is not complete? Beyond that, practically speaking, I'm not sure how this distinction is actually important to someone upgrading their map from 1.8 to later. Are you saying this document is, by design, not even complete for that?

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: "legacy map options and features" are legacy, it's not necessarily important to give a starting point. If someone finds something not in this list, then either it won't work or it's not legacy. If it does not work, hopefully we'll get a forum post about it to help.

>  I would keep such reference until it is correct or, at the very least, clarifying that this document is an incomplete "work-in-progress"

A warning that this might not be complete is perhaps not bad. I'd suggest against labelling it as a WIP, it'll be that way for forever (it is a living document).

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: We can replace the "*" with "..." if you think it might be confusing. I don't think a reasonable map maker is going to be confused by seeing that XML example.

I'm not sure if the min version should ever have been set to 1.9.0.0 without the build number. I dont' think this is worth spending more time on right now as the min version might be removed from XML

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: There is only one kind of XML file, it's clear that any XML examples are going to be for the game files. We are pretty explicit otherwise about the other files that need changing. We need to balance reasonable confusion with keeping this document concise. 

>  That also would imply that this list of changes is useless for upgrading mapskins

I don't necessarily draw that implication, but it could be worth clarifying perhaps. It probably is going to be obvious though that anyone upgrading a mapskin will skip over the XML parts.

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: The biggest consumer of this document so far, probably has been me. Considering I used it to port legacy game features to non-legacy in the game engine. Though, I found the document to not be useful and had to go back to the 10 page conversation and to great pains parse out what had been changed. Really I want a simple list of:
`a -> b`, and beyond that an absolute minimum of explanatory text and other corner cases that are not important.

I state the above also after the observation that someone explicitly asking to upgrade maps was directed to this document and it was not useful for them. Hence, the motivation to keep it simple and a straight forward change log of old to new. It may even turn out to be the case that this becomes just a reference document as well if the engine can automatically upgrade maps for map-makers.

So with that said, could you clarify why the ordering is important and/or how the before and after instruction needs improvement? Can you illustrate an example where this would steer someone wrong?

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: That context was not clear, this section perhaps should then instead give what these attributes should have been. I see now there was an assumption that maps being upgraded to 1.8.0.9 were done in a sloppy manner and instead of replacing legacy attributes only new attributes were added. That is a slightly confusing assumption I think, hence if we just give what the values should be, we can leave it to map makers to de-dupe the options (but duplicate options won't hurt the engine regardless so it's more important then we give the correct upgrade)

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: I've not tested it recently, but I believe that in the past only the first three digits were accounted (so "1.9.0.0" would be actually the same as writing "1.9.0"). I would be surprised if minimum version 1.9.0.0.x would actually exclude lower 1.9.0.0 versions.
Of course, it should have had a 5th digit, since not all 1.9.0.0 versions are fine, but the problem is that I believe the program was not reading it (I'm not sure, so here I would need a developer to verify and document if there is a digit limit and, if so, what it is).

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: Rather a warning that it is incomplete (though it might be happen to be complete for some map needs).

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: Yes, that is what I'm saying and what that part is saying. I think it's clear enough that it is incomplete: "it may not upgrade "Property" options (playerProperty, unitProperty...) of ineffective items you must delete".

Cernelius: That section tells to delete the options in the attachments, but it is not necessarily good enough if you happen to have the same options referenced in triggers (that is having triggers that change the options). Since this is a remote possibility, and it is particularly complex to tell how to handle it (it may or may not be the only thing that the trigger is doing), especially for a deletion case, I've left it out.

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: There is no "`="isTwoHit" value="true"` >> `="hitPoints" value="2"`" in what you are quoting. I assume you are confusing it with "`="isTwoHit" count="true"` >> `="hitPoints" count="2"`" (both are supposed to be present).

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: There is no "`="victoryCity" value="true"` >> `="victoryCity" value="1"`" in what you are quoting.

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: I guess I could have said that the same problem applies to the next line, in my previous comment, but I don't think that is the correct way to do it. They are two different items and lines, and they may have not been one next to the other, either.

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: The order is not needed, as you can make the changes in any order, but I think what is similar the most is better to stay close. This principle appears to be generally followed in the document, as well. For the changes at hand, I suggest the following order, if we have a single section:
- `="isTwoHit" value="true"` >> `="hitPoints" value="2"`
- `="isTwoHit" value="false"` >> `="hitPoints" value="1"`
- `="isTwoHit" count="true"` >> `="hitPoints" count="2"`
- `="isTwoHit" count="-reset-true"` >> `="hitPoints" count="-reset-2"`
- `="isTwoHit" count="false"` >> `="hitPoints" count="1"`
- `="isTwoHit" count="-reset-false"` >> `="hitPoints" count="-reset-1"`

And so on.
As I said, we could mix up the above in whatever order and intermix any of them with other code changes, but I think that would be unpolished. It is simply better for what changes the same option to be next to each other.

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: The existing section was only telling to open the "sounds" properties file and change any occurrence of ".wav" to ".mp3". Then there was another section telling to change all files within the "sounds" folder from "wav" to "mp3". I'm not seeing how those two sections were not fully sufficient (and they are independent, as you can have a map with "wav" references in the "sounds" properties file but no "sounds" folder (the example I made is "270BC", in High Quality) as well as a map with "sounds" in folder and no specific references to any of them in the "sounds" properties file (you can reference whole folders and subfolders, instead of specific sound files, and that would not need any upgrade, as far as I know)).

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: I would certainly advice against that, but this is off topic. Thanks for the information.

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: One needs to keep in mind that review is a social activity. The goal is not perfection, it is not just about the code/documentation & the text on page, you are interacting with another human. It's akin to criticizing every single article of clothing that is someone is wearing (even if each has a problem, the level of criticism conveys it's own message even if well intentioned). I'm calling this out as I fell into this same trap when first doing reviews and ignored the social aspect and had the mentality "we're going to make this the best damn thing and I'm going to call out every thing that I can see that can be improved.".

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: Here I'm just specifying the matter I noticed above, for this and the following 5 lines. This line is an alternative change for the general "`Attatchment` >> `Attachment`", while now it is given as the same for "`attatchment` >> `attachment`".

Assuming the case matters, this is wrong, because, if I start from `="techAttatchment` and change "`attatchment` >> `attachment`" either nothing happens (if my search was case sensitive) or the code is changed to `="techattachment` (instead of `="techAttachment`) (if my search was case insensitive).

It is not the line itself that is wrong. The fact that it is given as an alternative of "`attatchment` >> `attachment`" is wrong.

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: Well, at least in the moment that now this document is not specifying any particular version, I believe it is obvious that saying "were already deprecated and did nothing" is meaningless, as it is really referring to nothing. Assuming they are currently deprecated and do nothing, there is most likely some version of TripleA in the past where they weren't. A line like this has any sense only as long as the document itself is referencing a TripleA version from which to upgrade (where these items would have been already deprecated and useless). Strictly speaking, one may argue that, in the original document, they were no longer deprecated in the final TripleA version, since they went from being deprecated to being completely removed (something that I understant now has been effectively reverted, going back to the 1.8.0.9 situation).

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: Pointing out that I advise against what I'm saying now (to keep if friendly for persons not able to handle advanced search and replace), the above 6 lines of codes could have been exchanged for the following two (assuming "\*" means "zero or more of any"):
- `="isTwoHit"*="*true"` >> `="hitPoints"*="*2"`
- `="isTwoHit"*="*false"` >> `="hitPoints"*="*1"`

PR #7682, opened 2020-09-17
Documentation map changelog

DanVanAtta:
commit 01d83a1a9e604026d230ea17bf60559f389b1800

    Documentation: legacy map upgrade edits
    
    - Re-title the page, prep for the page to be named "map-change-log.md"
    - Clarify that changes need to be done in sound.properties
    - Remove warnings section, all the properties mentioned are already in
      the change list
    - Add missing variations of 'victoryCity' and 'isTwoHit' options

commit 7f6a2fc538ec1553fbd713fb6e242cccc2c6c64c

    Add notes to delete property children: boolean, string, and how to change 'number' property

- I'm planning to rename the document shortly to 'map-features-change-log.md'. For now keeping the original name to keep the diff intact. I'm thinking the name change will be desirable as the actual 'how to upgrade' will soon evolve. Specifically instead of changing a lot of text in XML, players can simply load a 1.8 map in 2.3 and then re-export the XML. Hence this document is now more geared as a change log (and hence why the the mention about minimum version is completely removed).

### DIFF
--- a/docs/map-making/how-to-upgrade-legacy-map-features.md
+++ b/docs/map-making/how-to-upgrade-legacy-map-features.md
@@ -28,8 +28,8 @@ Caution, this list might not be complete, please let us know if there is anythin
 - `="victoryCity" count="-reset-true"` >> `="victoryCity" count="-reset-1"`
 - `="victoryCity" count="false"` >> `="victoryCity" count="0"`
 - `="victoryCity" count="true"` >> `="victoryCity" count="1"`
-- `="victoryCity" value="false"` >> `="victoryCity" count="0"`
-- `="victoryCity" value="true"` >> `="victoryCity" count="1"`
+- `="victoryCity" value="false"` >> `="victoryCity" value="0"`
+- `="victoryCity" value="true"` >> `="victoryCity" value="1"`
 - `attatchment` >> `attachment`
 - `Attatchment` >> `Attachment`
 - `attatchTo=` >> `attachTo=`
--- a/docs/map-making/how-to-upgrade-legacy-map-features.md
+++ b/docs/map-making/how-to-upgrade-legacy-map-features.md
@@ -47,12 +47,6 @@ Caution, this list might not be complete, please let us know if there is anythin
 
 ### XML Deletions
 
-Delete these options entirely:
-
-- `<option name="takeUnitControl"`
-- `<option name="giveUnitControl"`
-
-
 Delete these tags (found under `<property>`)
 
 - `<boolean/>`
--- a/docs/map-making/how-to-upgrade-legacy-map-features.md
+++ b/docs/map-making/how-to-upgrade-legacy-map-features.md
@@ -1,26 +1,10 @@
-# How To Upgrade Legacy Map Features
+# Map Change Log
 
-A list of changes to upgrade legacy TripleA Game XML properties and options to the most recent.
+A list of changes to map files that have occurred over time.
+Caution, this list might not be complete, please let us know if there is anything missing.
  
-Note: Before applying the changes below (assuming you do it with a simple replace), you should make
-sure to not have unnecessary spaces inside angle brackets (chevrons).
-
-Note: Some items are not validated. Thus, if the game plays without errors using a latest,
-this doesn't necessarily mean that you correctly upgraded it (for more information, see the "warning"
-section).
-
 
 ## XML Changes
-
-These changes need to happen in your game XML files.
-
-### New game (XML) file settings:
-
-`<triplea minimumVersion="*"/>`:`<triplea minimumVersion="1.9.0.0"/>`
-
-### Changes:
-
-All occurrences of the code before the `>>` must be substituted with the code after the `>>`.
 
 - `attatchment` >> `attachment`
 - `Attatchment` >> `Attachment`
@@ -35,70 +19,36 @@ All occurrences of the code before the `>>` must be substituted with the code af
 - `="conditionType" count="XOR"` >> `="conditionType" count="1"`
 - `="conditionType" count="-reset-XOR"` >> `="conditionType" count="-reset-1"`
 - `="isTwoHit" count="true"` >> `="hitPoints" count="2"`
-- `="isTwoHit" value="false"` >> `="hitPoints" value="1"`
-- `="isTwoHit" count="-reset-true"` >> `="hitPoints" count="-reset-2"`
 - `="isTwoHit" count="false"` >> `="hitPoints" count="1"`
+- `="isTwoHit" value="true"` >> `="hitPoints" value="2"`
+- `="isTwoHit" value="false"` >> `="hitPoints" count="1"`
+- `="isTwoHit" count="-reset-true"` >> `="hitPoints" count="-reset-2"`
 - `="isTwoHit" count="-reset-false"` >> `="hitPoints" count="-reset-1"`
 - `="occupiedTerrOf"` >> `="originalOwner"`
 - `="victoryCity" count="true"` >> `="victoryCity" count="1"`
-- `="victoryCity" count="-reset-true"` >> `="victoryCity" count="-reset-1"`
 - `="victoryCity" count="false"` >> `="victoryCity" count="0"`
+- `="victoryCity" value="true"` >> `="victoryCity" count="1"`
+- `="victoryCity" value="false"` >> `="victoryCity" count="0"`
+- `="victoryCity" count="-reset-true"` >> `="victoryCity" count="-reset-1"`
 - `="victoryCity" count="-reset-false"` >> `="victoryCity" count="-reset-0"`
-
-
-### **Procedural alternatives:**
-
-For a more defined change, instead of the aforementioned:
-
-- `attatchment` >> `attachment`
-
-you can change:
-
-- `attatchmentList` >> `attachmentList`
-- `<attatchment` >> `<attachment`
-- `attatchment>` >> `attachment>`
-- `="games.strategy.triplea.attatchments` >> `="games.strategy.triplea.attachments`
-
-- `="techAttatchment` >> `="techAttachment`
-- `="unitAttatchment` >> `="unitAttachment`
-- `="territoryAttatchment` >> `="territoryAttachment`
-- `="canalAttatchment` >> `="canalAttachment`
-- `="rulesAttatchment` >> `="rulesAttachment`
-- `="playerAttatchment` >> `="playerAttachment`
+- `<option name="isParatroop"` >> `<option name="isAirTransportable"`
+- `<option name="isMechanized"` >> `<optiona name="isLandTransportable"`
 
 
 ### XML Deletions
 
-_(these items were already deprecated and did nothing)_
+Delete these options entirely:
 
-- `<option name="isParatroop"*/>`
-- `<option name="isMechanized"*/>`
-- `<option name="takeUnitControl"*/>`
-- `<option name="giveUnitControl" value="true"/>`
-- `<option name="giveUnitControl" value="false"/>`
+- `<option name="takeUnitControl"`
+- `<option name="giveUnitControl"`
 
-## map.properties Changes
-
-Within every "map" properties file, inside the main folder of the map, change:
+## map.properties changes
 
 - `color.Impassible=` >> `color.Impassable=`
 
-## Sound file changes
+## sound.properties and sound file changes
 
-Change all sounds inside the "sounds" main folder of the map from "wav" to "mp3" format.
-128Kbps and 168Kbps are the preferred bit rates.
+- .wav files need to be re-encoded as mp3 (128 or 168Kbps are preferred bit rates)
+- all file names in sound.properties need to be updated to match the new '.mp3' suffix
+  on sound files.
 
-
-## Warnings
-
-The following xml codes, as being properties, are not validated (if you leave them, the game will play,
-but they will set nothing in the engine) (it is suggested you double check you don't have them within
-any game files, after upgrading the map):
-
-- `="Battleships repair at end of round"`
-- `="Battleships repair at beginning of round"`
-- `="Naval Bombard Casualties Return Fire Restricted"`
-- `="SBR Affects Unit Production"`
-
-While all options named "takeUnitControl" must be deleted, any option named "giveUnitControl" must be
-deleted only if its value is "true" or "false".
--- a/docs/map-making/how-to-upgrade-legacy-map-features.md
+++ b/docs/map-making/how-to-upgrade-legacy-map-features.md
@@ -6,7 +6,7 @@ Caution, this list might not be complete, please let us know if there is anythin
 
 ## XML Changes
 
-- `<option name="isMechanized"` >> `<optiona name="isLandTransportable"`
+- `<option name="isMechanized"` >> `<option name="isLandTransportable"`
 - `<option name="isParatroop"` >> `<option name="isAirTransportable"`
 - `="Battleships repair at beginning of round"` >> `="Units Repair Hits Start Turn"`
 - `="Battleships repair at end of round"` >> `="Units Repair Hits End Turn"`
--- a/docs/map-making/how-to-upgrade-legacy-map-features.md
+++ b/docs/map-making/how-to-upgrade-legacy-map-features.md
@@ -1,4 +1,4 @@
-# Map Change Log
+# Map Features Change Log
 
 A list of changes to map files that have occurred over time.
 Caution, this list might not be complete, please let us know if there is anything missing.
@@ -35,12 +35,30 @@ Caution, this list might not be complete, please let us know if there is anythin
 - `<option name="isMechanized"` >> `<optiona name="isLandTransportable"`
 
 
+```
+<property name="..." value="...">
+   <number min="..." max="..."/>
+</property>
+
+>>
+
+<property name="..." value="..." min="..." max="..."/>
+```
+
 ### XML Deletions
 
 Delete these options entirely:
 
 - `<option name="takeUnitControl"`
 - `<option name="giveUnitControl"`
+
+
+Delete these tags (found under `<property>`)
+
+- `<boolean/>`
+- `<string/>`
+
+
 
 ## map.properties changes
 
--- a/docs/map-making/how-to-upgrade-legacy-map-features.md
+++ b/docs/map-making/how-to-upgrade-legacy-map-features.md
@@ -6,33 +6,33 @@ Caution, this list might not be complete, please let us know if there is anythin
 
 ## XML Changes
 
+- `<option name="isMechanized"` >> `<optiona name="isLandTransportable"`
+- `<option name="isParatroop"` >> `<option name="isAirTransportable"`
+- `="Battleships repair at beginning of round"` >> `="Units Repair Hits Start Turn"`
+- `="Battleships repair at end of round"` >> `="Units Repair Hits End Turn"`
+- `="conditionType" count="-reset-XOR"` >> `="conditionType" count="-reset-1"`
+- `="conditionType" count="XOR"` >> `="conditionType" count="1"`
+- `="conditionType" value="XOR"` >> `="conditionType" value="1"`
+- `="isImpassible"` >> `="isImpassable"`
+- `="isTwoHit" count="-reset-false"` >> `="hitPoints" count="-reset-1"`
+- `="isTwoHit" count="-reset-true"` >> `="hitPoints" count="-reset-2"`
+- `="isTwoHit" count="false"` >> `="hitPoints" count="1"`
+- `="isTwoHit" count="true"` >> `="hitPoints" count="2"`
+- `="isTwoHit" value="false"` >> `="hitPoints" count="1"`
+- `="isTwoHit" value="true"` >> `="hitPoints" value="2"`
+- `="Naval Bombard Casualties Return Fire Restricted"` >> `="Naval Bombard Casualties Return Fire"`
+- `="occupiedTerrOf"` >> `="originalOwner"`
+- `="SBR Affects Unit Production"` >> `="Damage From Bombing Done To Units Instead Of Territories"`
+- `="turns"` >> `="rounds"`
+- `="victoryCity" count="-reset-false"` >> `="victoryCity" count="-reset-0"`
+- `="victoryCity" count="-reset-true"` >> `="victoryCity" count="-reset-1"`
+- `="victoryCity" count="false"` >> `="victoryCity" count="0"`
+- `="victoryCity" count="true"` >> `="victoryCity" count="1"`
+- `="victoryCity" value="false"` >> `="victoryCity" count="0"`
+- `="victoryCity" value="true"` >> `="victoryCity" count="1"`
 - `attatchment` >> `attachment`
 - `Attatchment` >> `Attachment`
 - `attatchTo=` >> `attachTo=`
-- `="isImpassible"` >> `="isImpassable"`
-- `="conditionType" value="XOR"` >> `="conditionType" value="1"`
-- `="turns"` >> `="rounds"`
-- `="Battleships repair at end of round"` >> `="Units Repair Hits End Turn"`
-- `="Battleships repair at beginning of round"` >> `="Units Repair Hits Start Turn"`
-- `="Naval Bombard Casualties Return Fire Restricted"` >> `="Naval Bombard Casualties Return Fire"`
-- `="SBR Affects Unit Production"` >> `="Damage From Bombing Done To Units Instead Of Territories"`
-- `="conditionType" count="XOR"` >> `="conditionType" count="1"`
-- `="conditionType" count="-reset-XOR"` >> `="conditionType" count="-reset-1"`
-- `="isTwoHit" count="true"` >> `="hitPoints" count="2"`
-- `="isTwoHit" count="false"` >> `="hitPoints" count="1"`
-- `="isTwoHit" value="true"` >> `="hitPoints" value="2"`
-- `="isTwoHit" value="false"` >> `="hitPoints" count="1"`
-- `="isTwoHit" count="-reset-true"` >> `="hitPoints" count="-reset-2"`
-- `="isTwoHit" count="-reset-false"` >> `="hitPoints" count="-reset-1"`
-- `="occupiedTerrOf"` >> `="originalOwner"`
-- `="victoryCity" count="true"` >> `="victoryCity" count="1"`
-- `="victoryCity" count="false"` >> `="victoryCity" count="0"`
-- `="victoryCity" value="true"` >> `="victoryCity" count="1"`
-- `="victoryCity" value="false"` >> `="victoryCity" count="0"`
-- `="victoryCity" count="-reset-true"` >> `="victoryCity" count="-reset-1"`
-- `="victoryCity" count="-reset-false"` >> `="victoryCity" count="-reset-0"`
-- `<option name="isParatroop"` >> `<option name="isAirTransportable"`
-- `<option name="isMechanized"` >> `<optiona name="isLandTransportable"`
 
 
 ```


docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -35,70 +19,36 @@ All occurrences of the code before the `>>` must be substituted with the code af
 - `="conditionType" count="XOR"` >> `="conditionType" count="1"`
```
Cernelius, line 19: To keep everything listed similarly, I would move this and the following line after the line
- `="conditionType" value="XOR"` >> `="conditionType" value="1"` (I would prefer making two different lists, but this is not related to this update).

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -35,70 +19,36 @@ All occurrences of the code before the `>>` must be substituted with the code af
...
 - `="conditionType" count="-reset-XOR"` >> `="conditionType" count="-reset-1"`
 - `="isTwoHit" count="true"` >> `="hitPoints" count="2"`
-- `="isTwoHit" value="false"` >> `="hitPoints" value="1"`
-- `="isTwoHit" count="-reset-true"` >> `="hitPoints" count="-reset-2"`
 - `="isTwoHit" count="false"` >> `="hitPoints" count="1"`
+- `="isTwoHit" value="true"` >> `="hitPoints" value="2"`
```
Cernelius, line 23: I would have this and the following line before the other "isTwoHit" related lines, since this line and the following one are the two most important ones (by far). Practically this suggestion would change this part of the listing to this order:
- `="isTwoHit" value="true"` >> `="hitPoints" value="2"`
- `="isTwoHit" value="false"` >> `="hitPoints" count="1"`
- `="isTwoHit" count="true"` >> `="hitPoints" count="2"`
- `="isTwoHit" count="false"` >> `="hitPoints" count="1"`
- `="isTwoHit" count="-reset-true"` >> `="hitPoints" count="-reset-2"`
- `="isTwoHit" count="-reset-false"` >> `="hitPoints" count="-reset-1"`

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -35,70 +19,36 @@ All occurrences of the code before the `>>` must be substituted with the code af
...
+- `="isTwoHit" value="false"` >> `="hitPoints" count="1"`
+- `="isTwoHit" count="-reset-true"` >> `="hitPoints" count="-reset-2"`
 - `="isTwoHit" count="-reset-false"` >> `="hitPoints" count="-reset-1"`
 - `="occupiedTerrOf"` >> `="originalOwner"`
 - `="victoryCity" count="true"` >> `="victoryCity" count="1"`
-- `="victoryCity" count="-reset-true"` >> `="victoryCity" count="-reset-1"`
 - `="victoryCity" count="false"` >> `="victoryCity" count="0"`
+- `="victoryCity" value="true"` >> `="victoryCity" count="1"`
```
Cernelius, line 30: This is wrong. You are changing "value" to "count".

Also, as per the above comment, I would list the "value" ones before the "count" ones (the value ones are more important as they act on the final attachments, while the other ones are for triggers changing them (rare case)).
This would be the listing order, with the (hopefully) correct code:
- `="victoryCity" value="true"` >> `="victoryCity" value="1"`
- `="victoryCity" value="false"` >> `="victoryCity" value="0"`
- `="victoryCity" count="true"` >> `="victoryCity" count="1"`
- `="victoryCity" count="false"` >> `="victoryCity" count="0"`
- `="victoryCity" count="-reset-true"` >> `="victoryCity" count="-reset-1"`
- `="victoryCity" count="-reset-false"` >> `="victoryCity" count="-reset-0"`

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -35,70 +19,36 @@ All occurrences of the code before the `>>` must be substituted with the code af
...
+- `="victoryCity" value="false"` >> `="victoryCity" count="0"`
```
Cernelius, line 31: This is wrong (you are changing "value" to "count") and covered by my previous comment.

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -35,70 +19,36 @@ All occurrences of the code before the `>>` must be substituted with the code af
...
-- `attatchment>` >> `attachment>`
-- `="games.strategy.triplea.attatchments` >> `="games.strategy.triplea.attachments`
-
-- `="techAttatchment` >> `="techAttachment`
-- `="unitAttatchment` >> `="unitAttachment`
-- `="territoryAttatchment` >> `="territoryAttachment`
-- `="canalAttatchment` >> `="canalAttachment`
-- `="rulesAttatchment` >> `="rulesAttachment`
-- `="playerAttatchment` >> `="playerAttachment`
+- `<option name="isParatroop"` >> `<option name="isAirTransportable"`
```
Cernelius, line 34: Since this is not any longer specifically about upgrading from maps made for TripleA 1.8.0.9, I guess sometime in the past this option was doing something, which may or may not be (exactly) the same that "isAirTransportable" is doing. However, this needs to be verified to be actually the case (we cannot just assume "isAirTransportable" is merely a name change of "isParatroop", because they sound being about the same thing, unless this assumption is documented, which I'm not aware).

Moreover, it may be that there are games in which you have "isParatroop" set to "false" in a unit attachment after "isAirTransportable" set to true (garbage code, which I can tell is generally frequent in TripleA games). In this case, if in the TripleA for which the game was made the "isParatroop" was ineffective, you are substantially bugging the game, changing the behaviour of "isAirTransportable" from "true" to "false" (unburying bad coding).

In theory, to be exact, here you would need to find out the range of TripleA versions for which "isParatroop" was doing nothing, and telling that upgrading from those need to have the option deleted, to be sure to keep the behaviour unchanged (the upgrader would need to check the minimum version of the game, and delete or otherwise, depending on it).

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -35,70 +19,36 @@ All occurrences of the code before the `>>` must be substituted with the code af
...
+- `<option name="isMechanized"` >> `<optiona name="isLandTransportable"`
```
Cernelius, line 35: All that I said in my previous comment applies here too, changing "isParatroop" with "isMechanized" and "isAirTransportable" with "isLandTransportable".

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -35,70 +19,36 @@ All occurrences of the code before the `>>` must be substituted with the code af
...
-- `<option name="isParatroop"*/>`
-- `<option name="isMechanized"*/>`
-- `<option name="takeUnitControl"*/>`
-- `<option name="giveUnitControl" value="true"/>`
-- `<option name="giveUnitControl" value="false"/>`
+Delete these options entirely:
 
-## map.properties Changes
+- `<option name="takeUnitControl"`
+- `<option name="giveUnitControl"`
```
Cernelius, line 53: As I said in a section deleted at this update "any option named "giveUnitControl" must be deleted only if its value is "true" or "false".

If this is not true, this needs to be indagated, as I believe "giveUnitControl" is currently documented as an effective and non-deprecated option. If so, telling to delete any occurrence of such option would be a very bad and damaging information.

docs/map-making/how-to-upgrade-legacy-map-features.md
```diff
@@ -35,70 +19,36 @@ All occurrences of the code before the `>>` must be substituted with the code af
...
 
-- `="Battleships repair at end of round"`
-- `="Battleships repair at beginning of round"`
-- `="Naval Bombard Casualties Return Fire Restricted"`
-- `="SBR Affects Unit Production"`
+- .wav files need to be re-encoded as mp3 (128 or 168Kbps are preferred bit rates)
+- all file names in sound.properties need to be updated to match the new '.mp3' suffix
+  on sound files.
 
-While all options named "takeUnitControl" must be deleted, any option named "giveUnitControl" must be
```
Cernelius, line 103: To limit the risk of mistakenly remove a good "giveUnitControl" option, I think this warning (comprising the next line) is valuable even if the referenced item would be correct after this update.

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: I suppose it's a bit arbitrary how to order the list. I appreciate the suggestion though for the ordering to make more sense.

I pushed an update to alpha-sort the list: 81b5d1c48

Since it is a bit arbitrary, that at least gives a consistent ordering and keeps related elements close together.

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: Ugh, I didn't realize that VC option can be either value or count. IMO it should just be one or the other. Thanks for spotting this.

Updated in: c20682143

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: Updated in: c206821

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: I guess that's acceptable but, sadly, this means that the much less important "value" ones come after the "count" ones. I guess this is a minor matter.

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: @Cernelius I don't think we can assume that if a map is at version 1.8 that it was correctly upgraded to 1.8. Specifically, if we see 'isParatroop', does that mean the new property was simply added, or does it mean it was never added?

IMO it's better to upgrade and allow the map maker to spot and remove duplicate options if then present. That handles the potential 'true' / 'false' case which would already have been a potential problem.

Here is the diff of when 'isAirTransportable' was added: 

commit d56265ad9bef86651ccbcd310f3964b78188d5b0
Author: comradekev <comradekev@yahoo.com>
Date:   Thu Jun 3 19:36:19 2010 +0000

    * fix bug 2998846 kamikaze aircraft not working * add isAirTransport(able) to support air dropping units * fix bug 2994004 air landing on newly produced carrier in multi-national SZ disappears * feature request (2799840) add tech tokens to status tab

    git-svn-id: https://svn.code.sf.net/p/triplea/code/trunk/triplea@2785 a4c711aa-641f-0410-879d-9b464e52b682

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: They cannot, as options. However, generally speaking, when you use triggers to change options, you define the name as value and the value as count (so you cannot use the count but in its value-integrated form (having the count being the first element specified in "value", which, in this case, would be the first element specified in "count", since the "count" is the "value" (it actually doesn't matter in this specific case, as the "victoryCity" option happens not to have a "count" at all (you can specify a value, never a count, for it)))).

Since I feel what I just said is not very clear, I'll just redirect to the official information:
https://github.com/triplea-maps/the_pact_of_steel/blob/master/map/games/pact_of_steel_2.xml#L3397

As a warning, such "properties" have nothing to do with what is normally identified as "property", within the game XML (and of course, nothing to do with the other things identified as "properties" too outside of it).

The "value" ones are for the options, the "count" ones are for triggers that are changing such options during the course of the game (most likely, there is not a single game doing that, I guess).

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: Ditto, I am pretty confident this is the correct update to be done.


commit 67caac77e8b3da9ac759c0f2283806b89e68f70d
Author: Ron <Ron@RMURHAMM-W10E.fios-router.home>
Date:   Thu Dec 7 17:20:08 2017 -0600

    Add isLandTransportable property and some refactoring

+  // isInfantry is DEPRECATED, use isLandTransportable

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: I guess there's no need to worry too much about these things, since this is a mapmaker aid, but certainly I believe that, if we were ourselves making mass-changes, mass changing all "isParatroop" to "isAirTransportable", after many years in which the original has been ineffective, would be very risky.

It may be interesting to see a summary of all instances of "isParatroop" that have been removed in the repository, back then when the mass-deletion was made (I'm assuming it was).

I'll still point out that we should not assume that "isParatroop" used to do the same that "isAirTransportable" is doing just because the name seems to be about the same thing. We shouldn't really infer anything from the name itself, as that is merely a label. So such a change really needs confirmation that "isAirTransportable" is indeed merely a renamed version of "isParatroop". If it is not, then we would be changing, not upgrading, also games made when the "isParatroop" option was non deprecated.

All this based on the assumption that what I know and documented is right, since I've not tested any of this, at least recently.

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: My second comment at my previous comment still applies here too, the same way.

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: Why was this removal mentioned in the original notes?  I'm not aware we removed the 'true' 'false' to do nothing in 1.8 or 1.9

DanVanAtta: I think it's better we just remove it for now. Chances are good the map will crash with incorrect values which would mean it would get noticed. 

Updated: 131e6407a

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: I don't think this reduces the risk of mistakenly removing a valid option. If someone follows this guide exactly, then it won't be an issue, right? But furthermore, this is becoming a pure change log for information purposes, not to be followed with find/replacements. Specifically if we can load 1.8 maps in 2.3, then the right process will be to load the map and re-export it.

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: Given the time of when 'isAirTransportable' was introduced, I'm pretty sure it was not mass replaced. It was left to individual map makers to make the update and instead their maps were broken by leaving in a useless property that while the game still ran, the 'isParatroop' no longer did anything. So map makers would have updated it or their maps were changed.


>  So such a change really needs confirmation that "isAirTransportable" is indeed merely a renamed version of "isParatroop". If it is not, then we would be changing, not upgrading, also games made when the "isParatroop" option was non-deprecated.

Check the referenced commit, it replaces isParatroop. I would guess if we see 'isParatroop', it's more likely that it was never replaced rather than was replaced but left in. I find that a far more likely scenario. Beyond that, if isParatroop is in the map, then 'isAirTransportable' was the desired behavior. A map not having 'isAirTransportable' would then not be functioning correctly. I'd reject the statement that if a map made it to 1.8 and someone continued to overlook that, then 'oh well', the map was never supposed to have 'isParatroop' to begin with.

DanVanAtta: > I guess there's no need to worry too much about these things, since this is a mapmaker aid, but certainly I believe that, if we were ourselves making mass-changes, mass changing all "isParatroop" to "isAirTransportable", after many years in which the original has been ineffective, would be very risky.

I think the days of mass changes are behind us. The migration is breaking to old engines and the benefit is non-existent now that 2.3 and beyond will be able to handle legacy values.

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: I don't believe "isInfantry" has anything to do with this matter, beside being the old name for "isLandTransportable". I'm certain that "isLandTransportable" didn't yet exist when "isInfantry" was non-deprecated and "isMechanized" was.

If "isMechanized" is merely an old name for "isInfantry", then "isInfantry" would be the intermediate rename between "isMechanized" and "isLandTransportable", but I believe this is not actually the case, as I recall that "isInfantry" used to be an option applying more generally (I think it might have been the reference used for paratroopers too, at some point, but I'm not sure about this) and I don't even know which one existed before, in TripleA, between "isMechanized" and "isInfantry". I vaguely recall that the application of "isInfantry" varied overtime, until being restricted to, then superseded by, what "isLandTransportable" is currently doing, but I don't have the details.

My suggestion is not caring about the history of "isInfantry". What is actually needed is to assure that the current behaviour of "isLandTransportable" is exactly the same as the pre-deprecation behaviour of "isMechanized", if it is really wanted to suggest such a change (instead of telling to delete "isMechanized").

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: Absolutely not. As fas as I know, "giveUnitControl" is a currently valid option that should not necessarily be deleted.

What my notes were saying is that "giveUnitControl" was doing nothing if set at value "true" or "false" and that, after the removals, should be deleted if set at those values (only).

If my notes were wrong, a developer would need to prove it by looking at the code or testing. However, I can reference the official game information, that currently have "giveUnitControl" as a non-deprecated option:
https://github.com/triplea-maps/the_pact_of_steel/blob/master/map/games/pact_of_steel_2.xml#L4021

Therefore, unless proved wrong, my information that "takeUnitControl" should be always deleted (as an option), while "giveUnitControl" should be deleted (as an option) only if set "true" of "false" should be correct and valuable information. Removing all reference to such options would be, in this case, a major step backward in making this document less incomplete.

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: IIRC 'isInfantry' implied 'isMechanized' and a few other properties and that is why it was removed.

'isMechanized' is seemingly properly replaced by 'isLandTransportable'. For example, I found this check for it a unit can be loaded when the 'mechanized infantry' tech is discovered:

```
    if (getIsLandTransportable() && playerHasMechInf(player)) {
      tuples.add(Tuple.of("Can be Land Transported", ""));
    }
```

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: > What my notes were saying is that "giveUnitControl" was doing nothing if set at value "true" or "false" and that, after the removals, should be deleted if set at those values (only).

Why would it have been set to those values? Was it doing something previously? If so, I'm not aware of that changing in 1.9, was it change in 1.9? 

Are you saying "always" deleted assuming someone upgraded to 1.8 correctly? Personally, I am _NOT_ assuming maps were always correctly upgraded to a 1.8 and just because you see a 'does nothing' property in 1.8 that it should be removed rather than replaced by it's correct replacement.

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: I recall that "isInfantry" was renamed to "isLandTransportable" merely to have a clearer name. When that happened, TripleA was past the point in which I believe you completely removed "isMechanized" from the program (under my suggestion).

I believe what you quote may have nothing to do with the matter at hand. "MechInf" is referring to a technology a player may have, while "isMechanized" is a unit option. Unless proved by other means, I don't think that similar names assume relation (we have cases they do not, like "kamikaze"). Even if it is likely, in this case, that they are related, this still is not assuring that the non-deprecated behaviour of a "isMechanized" true unit would be the same as the current behaviour of a "isLandTransportable" true unit.

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: Right or wrong (and, to the extent of my knowlege, it has always been right), what my documentation was telling is to:
- Always delete the "takeUnitControl" option.
- Delete the "giveUnitControl" option only if its value is "true" or "false".

I think this is clear enough.

This is based on the information (which now it is an assumption) that such items to delete were doing nothing, of course.

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: I'm pretty sure 'isInfantry' was not renamed, but 'isMechanized' was. 'isinfantry' granted more options other than just land transportability.

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: Indeed, dangerous assumption. You're assuming that someone:
1. was upgrading their map
2. saw a deprecated property
3. knew it was supposed to be updated
4. Added the new property but did not remove the old

IMO it's MUCH more likely that a map was simply just *NEVER* updated and so the property still remains.

When we consider the risk of the two scenarios:
- in the first you have unclean XML and possibly conflicting values. If a map maker follows this guide then they'll see they have duplicate values (hopefully, but the proximity will be close to each other) and will resolve).
- in the second you have a map that is just not working correctly.

I'll emphasize this is a change log document now to describe what map makers should do *now* given whatever map they are seeing. This goes back to my point about the lack of utility to upgrade to only 1.9.x and also how that does not serve an ongoing purpose of where to document changes (which then if this is a historical document, we then have duplication, legacy issues and just essentially more documentation laying around that makes it all harder to keep up to date).

DanVanAtta: >  Otherwise, for example in this case, if I'm correct, you would need to have one upgrade information set for when the option was doing one thing and another one for when the option was doing nothing.

I don't think so. If the new properties would lead to duplication, we can leave it to a map maker to see that duplication. Even if they do not spot it, the duplication causes no harm. If the values were conflicting, then the map had issues to begin with.

> Whether or not the map was correctly coded in 1.8.0.9 I believe it is beside the point.

That actually is very much to the point. If we say "delete" this, and the map then changes in a way that is hard to notice, that's bad. Compare that to the risk of something not being upgraded more properly (likely), or someone not realizing that they are looking at a 1.7 map thinking it is a 1.8 map, then suddenly our directions steer them wrong.

DanVanAtta: I'm not going to add these back in:

```

    Always delete the "takeUnitControl" option.
    Delete the "giveUnitControl" option only if its value is "true" or "false".
```

We need to know what the options are supposed to be changed into and let a map maker take care of potential duplication (which is relatively harmless if there is a duplicated value).

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: I'm pretty confident that the noted changes are correct. Could certainly be wrong, it's an overly complicated mess of a code-base (that does great things, but messy under the covers). If we find out otherwise we can fix it. In the meantime I suspect nobody is going to use this guide for quite some time and perhaps if they can bring it up. For practical purposes this guide is actually most useful to me right now for coding up how legacy options are to be ported forward.

DanVanAtta:
Thanks for bearing with me @Cernelius ; we've improved this. Feel free to suggest future edits, I am hijacking this document to be a master changelog as that is truly needed and is more useful than one single historical  document.

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: "isInfantry" was not renamed? I'm as sure as I can be that @ron-murhammer renamed "isInfantry" to "isLandTransportable" and I still believe this has nothing to do with this matter. If the "isInfantry" to "isLandTransportable" would have been never made, I think now we would be at the same point, just that "isLandTransportable" would be still called "isInfantry". Then, we would be just discussing whether or not to suggest changing "isMechanized" to "isInfantry", exactly like now we are discussing the same for "isMechanized" to "isLandTransportable".

If @ron-murhammer can confirm what I'm saying, otherwise someone would need to look at the code, I assume.

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: Well, the original documentation was meant for upgrading maps from 1.8.0.9, while keeping exactly the same behaviour they had, when played with that version of TripleA. If now you are setting different standards, that is a change of direction. Here I'm just highlighting risks. However, if it can be confirmed that, at some time in TripleA history, these options were changed to become ineffective (and this was still the case when 1.8.0.9 came out), I think it would be the only consistent thing to do to turn them ineffective still. It is the same matter as if an option's behaviour was changed in the past. Maybe there is a map still coded on the assumption that the old behaviour applies, but you cannot go after such things, reliably.

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: No, isInfantry was not renamed as it was triggering multiple behaviors, including 'isLandTransportable'. "isInfantry' implies 'isLandTransportable', but not the other way around.

docs/map-making/how-to-upgrade-legacy-map-features.md
DanVanAtta: Please see my other comments, it's more risky to assume that someone did a bad upgrade to 1.8.0.9 and left in needless properties after replacing rather than the risk of they simply never upgraded at all.

If we need to assume a perfect upgrade to 1.8.0.9, then we need to have that documentation, which we don't. Otherwise how can you ever be confident, even if someone claims that a given map was upgraded, that it was done in a correct manner?

DanVanAtta: > Well, the original documentation was meant for upgrading maps from 1.8.0.9

I am changing the scope, in sum for three reasons:
- that is a risky assumption (see previous comment)
- not useful (there is no point to upgrade to just 1.9.0.0.3xxx and there is no guarantee that a map is actually perfectly on 1.8.0.9)
- fails to serve an ongoing need

docs/map-making/how-to-upgrade-legacy-map-features.md
Cernelius: I clearly remember that "isInfantry" was renamed to "isLandTransportable", while another option named "isInfantry" was added for compatibility. The "isInfantry" after the addition of "isLandTransportable" would do nothing but setting "isLandTransportable" to its same value.

Meaning that I believe the only reason why "isInfantry" still exists is to support old maps, just like if you would update them all changing "isInfantry" to "isLandTransportable".

Immediately before the addition of "isLandTransportable", I recall "isInfantry" was doing nothing more than what "isLandTransportable" currently does ("isInfantry" was probably doing other things further in the past, before 1.8.0.9).

>isInfantry was not renamed as it was triggering multiple behaviors, including 'isLandTransportable'.

I believe that, from the addition of "isLandTransportable" onwards, "isInfantry" does nothing but setting "isLandTransportable". So I'm not aware that now or at any point in time "isInfantry" has ever both set "isLandTransportable" and done anything else, at the same time.

>"isInfantry' implies 'isLandTransportable', but not the other way around.

As far as I know, this is true only because "isInfantry" does nothing but setting "isLandTransportable" to its same value. If "isInfantry" currently does anything else (making "isLandTransportable" a subset of "isInfantry"), this should be documented, I assume.

@ron-murhammer Since I recall you are behind the "isLandTransportable" addition, can you help clarifying the matter here?

DanVanAtta: It could be the case we were in a partial state where 'isInfantry' was made to be a subset of its original functionality. I would be a bit surprised AFAIK the only mass migrations of maps was done when we went to 1.9.

Currently 'isInfantry' appears nowhere in the codebase except in the legacy property mapper.

DanVanAtta: 'isInfantry' is missing from the list actually. We are beyond the scope of this update @Cernelius . It's best practice to keep conversations in a PR to the content in the PR. If we want clarification on this, let's discuss in an issue so that others can find it later and not have to search random PRs for the convo.

Cernelius: > 'isInfantry' is missing from the list actually. We are beyond the scope of this update @Cernelius . It's best practice to keep conversations in a PR to the content in the PR. If we want clarification on this, let's discuss in an issue so that others can find it later and not have to search random PRs for the convo.

This is exactly what I said:

>I don't believe "isInfantry" has anything to do with this matter, beside being the old name for "isLandTransportable". I'm certain that "isLandTransportable" didn't yet exist when "isInfantry" was non-deprecated and "isMechanized" was.

We should have really never even mentioned "isInfantry", as the matter here appears to be exclusively whether to suggest to delete "isMechanized" as an option or to change "isMechanized" to "isLandTransportable". I believe this is dependent exclusively on what these two options did or do. You added "isInfantry" to the matter, as an element to take into consideration.

PR #7688, opened 2020-09-17
Convert general casualty removal

trevan:

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
@@ -34,6 +34,7 @@ import games.strategy.triplea.delegate.battle.steps.BattleStep;
 import games.strategy.triplea.delegate.battle.steps.BattleSteps;
 import games.strategy.triplea.delegate.battle.steps.RetreatChecks;
 import games.strategy.triplea.delegate.battle.steps.change.ClearAaCasualties;
+import games.strategy.triplea.delegate.battle.steps.change.ClearGeneralCasualties;
 import games.strategy.triplea.delegate.battle.steps.change.LandParatroopers;
 import games.strategy.triplea.delegate.battle.steps.change.MarkNoMovementLeft;
 import games.strategy.triplea.delegate.battle.steps.change.RemoveNonCombatants;
@@ -1587,16 +1588,19 @@ public class MustFightBattle extends DependentBattle
     final BattleStep defensiveSubsRetreat = new DefensiveSubsRetreat(this, this);
     final BattleStep removeGeneralSuicide = new RemoveGeneralSuicide(this, this);
     final BattleStep offensiveGeneralRetreat = new OffensiveGeneralRetreat(this, this);
+    final BattleStep clearGeneralCasualties = new ClearGeneralCasualties(this, this);
 
-    steps.add(
-        new IExecutable() {
-          private static final long serialVersionUID = 8611067962952500496L;
+    steps.add(clearGeneralCasualties);
+    new IExecutable() {
+      private static final long serialVersionUID = 8611067962952500496L;
 
-          @Override
-          public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
-            clearWaitingToDieAndDamagedChangesInto(bridge);
-          }
-        });
+      @Override
+      @RemoveOnNextMajorRelease
+      public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
+        new ClearGeneralCasualties(MustFightBattle.this, MustFightBattle.this)
+            .execute(stack, bridge);
+      }
+    };
     steps.add(removeGeneralSuicide);
     new IExecutable() {
       private static final long serialVersionUID = 6387198382888361848L;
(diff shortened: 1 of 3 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1587,16 +1588,19 @@ private void addCheckEndBattleAndRetreatingSteps(final List<IExecutable> steps)
...
+    new IExecutable() {
+      private static final long serialVersionUID = 8611067962952500496L;
 
-          @Override
-          public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
-            clearWaitingToDieAndDamagedChangesInto(bridge);
-          }
-        });
+      @Override
+      @RemoveOnNextMajorRelease
```
DanVanAtta, line 1598: We're going to be inline this on next major vs outright removal, right?
If so, this might help us in the future to comment that so we don't think we can *just* delete it:
```
RemoveOnNextMajorRelease("inline IExecutable body and remove IExecutable")
```

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
```diff
@@ -58,7 +59,10 @@
         new ClearFirstStrikeCasualties(battleState, battleActions);
     final BattleStep offensiveStandard = new OffensiveGeneral(battleState, battleActions);
     final BattleStep defensiveStandard = new DefensiveGeneral(battleState, battleActions);
-    final BattleStep generalRetreat = new OffensiveGeneralRetreat(battleState, battleActions);
+    final BattleStep offensiveGeneralRetreat =
+        new OffensiveGeneralRetreat(battleState, battleActions);
+    final BattleStep clearGeneralCasualties =
```
DanVanAtta, line 64: Any thoughts to rename 'clear' to 'remove'? They are mostly synonymous, IMO 'remove' is just slightly more descriptive.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/ClearGeneralCasualties.java
```diff
@@ -0,0 +1,35 @@
...
+import games.strategy.triplea.delegate.battle.BattleState;
+import games.strategy.triplea.delegate.battle.steps.BattleStep;
+import java.util.List;
+import lombok.AllArgsConstructor;
+
+@AllArgsConstructor
+public class ClearGeneralCasualties implements BattleStep {
+  private static final long serialVersionUID = 995118832242624142L;
+
+  protected final BattleState battleState;
```
DanVanAtta, line 17: Why protected on these variables? Can we mark them as private?

> Effective Java Item 17: Design and document for inheritance or else prohibit it
https://medium.com/@rufuszh90/effective-java-item-17-design-and-document-for-inheritance-or-else-prohibit-it-be6041719fbc

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: No, it should be removed in the next major version.  These IExecutables are completely replaced by the `BattleStep` classes that I'm building.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
trevan: Good idea except we already have `ClearAaCasualties` and `ClearFirstStrikeCasualties` which do the same thing for the other battle phases.  I'd like to keep their names similar.  And since they are stored in the save game, we would have to either just duplicate the class or do some sort of deserialization magic to handle the rename.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
DanVanAtta: :+1:

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleSteps.java
DanVanAtta: Fair enough :+1:

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/ClearGeneralCasualties.java
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  @Override
+  public Order getOrder() {
+    return Order.GENERAL_REMOVE_CASUALTIES;
+  }
+
+  @Override
+  public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
+    battleActions.clearWaitingToDieAndDamagedChangesInto(bridge);
```
DanVanAtta, line 33: My first comment on this seems to have been lost in github : /

I'm wondering/concerned if this class is too thin and a bit indirect. Can we move the logic of `MustFightBattle` here and have the other references in `MustFightBattle` and elsewhere call this class? Even if we move that logic here as a static method, at least it would have the one, isolated home and presumably could be independently tested. Of most significant though we'd be unburdening `MustfightBattle` that is already doing too much.

trevan: Yes, I plan to look into doing something like that now that most of the battle steps are converted.  You'll start seeing PR where I'm changing the `BattleAction` and `BattleState` apis and pull more of the code into the battlesteps or some sort of helper.

DanVanAtta: :+1:

DanVanAtta (approved):

PR #7689, opened 2020-09-17
Add a serialVersionUID to OffensiveGeneralRetreat

trevan:
I noticed I had forgotten to hardcode the serialVersionUID.  I calculated the auto-generated value and this hardcodes it.

## Testing

Save a battle with master and then loaded it in this branch to make sure it unserializes the object.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
@@ -25,6 +25,8 @@ import lombok.Value;
 @AllArgsConstructor
 public class OffensiveGeneralRetreat implements BattleStep {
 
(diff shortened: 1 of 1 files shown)


DanVanAtta:
Appreciate the due diligence around testing :+1: 
We do have a save-game compatibility test,  perhaps we can augment that? I think it has a save that is during battle (if not, likely should be added). With that said, never can be too careful.

trevan:
Where can I see the saves that are used in the compatibility test?

DanVanAtta:
@trevan , please see below the list:
- https://github.com/triplea-game/triplea/blob/master/smoke-testing/save-game-list.txt

How they are downloaded:
- https://github.com/triplea-game/triplea/blob/master/smoke-testing/build.gradle#L39

Usage in test:
- https://github.com/triplea-game/triplea/blob/master/smoke-testing/src/test/java/games/strategy/engine/data/GameSaveCompatibilityTest.java#L20

PR #7692, opened 2020-09-18
Convert end game checks battle step

trevan:
This is the last real battle step to convert.  The remaining one after this one starts a new round.  Once this is merged, I'll have a separate PR to clean up the step creation and ordering.

## Testing

I've played a few battles to make sure it determines winners correctly still.


game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/CheckGeneralBattleEnd.java
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  @Override
+  public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
+    if (hasSideLost(BattleState.Side.OFFENSE)) {
+      battleActions.endBattle(bridge);
+      battleActions.defenderWins(bridge);
+
+    } else if (hasSideLost(BattleState.Side.DEFENSE)) {
+      removeUnprotectedUnits.accept(bridge, BattleState.Side.DEFENSE);
```
trevan, line 50: This step is, currently, the only step that needs access to logic from another battle step.  I'm undecided in how best to access that logic.  I've three thoughts:

1. Instantiate the steps and call the public method right when it is needed (aka `new RemoveUnprotectedUnits(battleState, battleActions).removeUnprotectedUnits(bridge, BattleState.Side.DEFENSE)`)
2. Pass in the the step to the constructor of `CheckGeneralBattleEnd` and then call the public method when it is needed.
3. Pass in a lambda of the public method and call the lambda.

I've currently coded up #3.

I don't like #2 because it will cause the other battle steps to be saved with this battle step and would make future changes harder to do.

I've tried #1 and it makes testing a little harder.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/CheckGeneralBattleEnd.java
```diff
@@ -0,0 +1,130 @@
...
+    if (hasSideLost(BattleState.Side.OFFENSE)) {
+      battleActions.endBattle(bridge);
+      battleActions.defenderWins(bridge);
+
+    } else if (hasSideLost(BattleState.Side.DEFENSE)) {
+      battleActions.endBattle(bridge);
+      battleActions.attackerWins(bridge);
+
+    } else if (isStalemate()) {
+      if (!canAttackerRetreatInStalemate()) {
```
DanVanAtta, line 56: Codacy found an issue: [These nested if statements could be combined](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6199427)

trevan:
I've reworked the logic.  Now, instead of running the "RemoveUnprotectedUnits" and the "OffensiveGeneralRetreat" as part of the "CheckGeneralBattleEnd", they run as separate steps.

In the old logic for "RemoveUnprotectedUnits", the unprotected units were only removed if the defender had lost.  Running it always shouldn't have any change.
* If the attacker had lost, then the units would still have been destroyed.
* In a stalemate, units shouldn't be destroyed when "RemoveUnprotectedUnits" is called since there is nothing to kill the other units.

In the old logic for "OffensiveGeneralRetreat", it would check if a retreat is available and then call the retreat before ending the game.  Now, it doesn't end the game and let the regular retreat steps happen.  But before the defender can retreat, it adds a new battle step that checks if it is a stalemate and then it ends the game.

Unfortunately, I had to create a "CheckGeneralBattleEndOld" (any ideas on the name?) that is only used for save compatibility reasons.  This has to do the original logic because the steps are part of the saved game.

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleActions.java
```diff
@@ -39,6 +39,10 @@ void landParatroopers(
 
   void attackerWins(IDelegateBridge bridge);
 
+  void defenderWins(IDelegateBridge bridge);
+
+  void nobodyWins(IDelegateBridge bridge);
```
DanVanAtta, line 44: The "xxWins" seems to be an enumeration of the ' WhoWon' enum. WDYT if we parameterized the API instead?

EG:
```
battleComplete(IDelegateBridge bridge,  WhoWon battleResult);
```

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
```diff
@@ -35,6 +40,8 @@ public Side getOpposite() {
 
   Collection<Unit> getBombardingUnits();
 
+  Collection<Unit> getAmphibiousLandAttackers();
```
DanVanAtta, line 43: While thinking about parameterizing APIs, should we parameterize this as well? Should this even be an API method come to think of it?

What if the caller were to do the filtering it needed on its own leaving the API to be simply "getAttackers()".

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/CheckGeneralBattleEnd.java
```diff
@@ -0,0 +1,128 @@
...
+      battleActions.nobodyWins(bridge);
+    }
+  }
+
+  protected boolean hasSideLost(final BattleState.Side side) {
+    return battleState.getUnits(side).stream().noneMatch(Matches.unitIsNotInfrastructure());
+  }
+
+  protected boolean isStalemate() {
+    return (battleState.getMaxBattleRounds() > 0
```
DanVanAtta, line 66: I think we might be seeing a data clump code-smell here with getMaxBattleRounds & getBattleRounds().

Extract object is the refactoring pattern to fix that, in this case that would be a `BattleRound` object.

Then for example this code would be:
```
return battleState.getBattleRound().hasMoreRounds() || (getPower ...)
```

WDYT?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnits.java
```diff
@@ -52,6 +53,7 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
     removeUnprotectedUnits(bridge, BattleState.Side.OFFENSE);
   }
 
+  @RemoveOnNextMajorRelease("This doesn't need to be public in the next major release")
```
DanVanAtta, line 56: :+1: 

This looks good. We've had issues of marking changes that needed to happen on the next major release and have missed some or come across them and were then unsure of what to change exactly. It looks like this annotation is working out, I'm much more confident on next major we'll catch & fix these types of things.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnitsGeneral.java
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  @Override
+  public List<String> getNames() {
+    return List.of();
+  }
+
+  @Override
+  public Order getOrder() {
+    return Order.REMOVE_UNPROTECTED_UNITS_GENERAL;
```
DanVanAtta, line 20: It does not look like there is much implementation here and this is mostly a configuration class.

What do you think of creating a more generic variant of this class? It looks like all we would need to do it to parameterize the 'getOrder()' value as a constructor arg, and then we could simply return it. Can we potentially reduce the inheritance tree even and have the super class do some of this work, particularly if it had a constructor that accepted the `getNames()` and `getOrder()` values? Since those are constants, it seems like we probably could new-up the parent directly.

DanVanAtta: Looking at this a bit more, I do think we should avoid the inheritance tree and see if we can flatten this. A rule of thumb is to never override methods that have implementations. It generally makes code hard to understand, test, and inheritance is the strongest form of coupling. Instead just make the parent method abstract, convert the parent class into two, one that is abstract and a second that has the original implementation from the parent. Then this class would become a second child.

What that would look like is:
parent-hierarchy


But, the differences between the child classes are just constants, configuration, not behavior. In this case I'd flatten it more so you would just new up the BattleStep via a factory type of pattern and inject the right configurations and some behavior for the execution, EG:

```
BattleStep RemoveUnprotectedUnits = BattleStepFactory.builder()
    .names( ... )
    .order( ...)
    .executable( removeUnprotectedUnitsExecutable(bridge))
    .build()
    .toBattleStep();
```

WDYT?

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleActions.java
trevan: Yup, one of the things I plan to work on after this is done.

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
trevan: I thought about doing that as part of this PR but discovered that the `amphibiousLandAttackers` list was being generated in a manner that didn't lend to allowing the caller to generate.  I need to do more investigation and see why it is being done that way and if I can change it without causing problems.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/CheckGeneralBattleEnd.java
trevan: `getBattleRound()` is already used and has an `int` return value.  So I used `getBattleRoundState()` unless you have a better name idea.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnitsGeneral.java
trevan: The problem is save compatibility.  `RemoveUnprotectedUnits` is part of the saved file and so it has to be able to deserialize correctly.  I can add a parent to it but I can't move the two private variables.  That would mean that I would need to add abstract `getBattleState` and `getBattleActions` methods to the parent class and then have the two child classes overload them to make the variables available to the parent methods.  That just feels weird.

As for the builder idea, that means that now the configuration is part of the saved file and it makes it even harder to change code in the future.  I would much rather keep it in classes with as little class state as possible.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/CheckGeneralBattleEnd.java
DanVanAtta: I'm suggesting that `getBattleRound()` should return a `BattleRound` object. I''m not really suggesting to rename. The code smell is that we are always using both 'max rounds' and 'current rounds' together, that is the data clump (https://martinfowler.com/bliki/DataClump.html)

trevan: I understand that.  But `getBattleRound()` already exists outside of the `BattleState` interface and is used by other classes that aren't battle steps.  So, to prevent having to change those classes, I modified `BattleState` to not expose `getBattleRound()` but to instead expose `getBattleRoundState()` and that returns the `BattleRound` object.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/RemoveUnprotectedUnitsGeneral.java
DanVanAtta: I kinda suspect there might be some ways around the serialization issue, though perhaps not.. It is of course frustrating that serialization concerns really impacts our code design (and in this case, forces inheritance. I wonder a bit how many "Java inheritance now considered harmful" articles there are by now).

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/CheckGeneralBattleEnd.java
DanVanAtta: Ah, understood. I wonder then if we actually have a true data clump if those other locations only need the current round.

PR #7693, opened 2020-09-18
Improve available game loading speed

DanVanAtta:
- Speeds up the button "select maps"
- Speeds up bot startup time

(1) We swap ZipInputStream (slow) for ZipFile (fast) when
    searching for XML files within zip files. This is
    the majority of the performance improvement and is
    a magnitude faster.

(2) Preload available games at game launch. By the time a
    user has clicked the 'Select Maps' button, chances
    are really good we'll have pre-cached the list of
    available games and the list would then open without
    delay.

Addresses the performance problem noted in: https://github.com/triplea-game/triplea/issues/7656

## Release Note

UPDATE|Improve performance of the 'available maps' button and startup time for bot servers


tvleavitt:
Referenced wrong issue here. The performance issue was noted in: #7656

DanVanAtta:
Thanks for the correction @tvleavitt , I updated  the OP.

PR #7713, opened 2020-09-19
Use BattleStep.getAll instead of creating each one individually

trevan:
All of the steps except for the final "re-loop" step is done.  So I've changed everything to just use `BattleStep.getAll()` and no longer construct the steps individually.

I also added some `@RemoveOnNextMajorRelease` comments to indicate changes that should be done later.  Maybe there should be a `@ChangeOnNextMajorRelease` annotation?

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
@@ -8,6 +8,7 @@ import games.strategy.engine.data.Unit;
 import java.util.Collection;
 import java.util.UUID;
 import lombok.Value;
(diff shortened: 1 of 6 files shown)


DanVanAtta:
@trevan :

> Task :game-core:checkstyleMain

>  [ant:checkstyle] [WARN] /home/travis/build/triplea-game/triplea/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/CheckStalemateBattleEnd.java:11: Line is longer than 100 characters (found 101). [LineLength]

Adding a 'ChangeOnNextMajorRelease' makes sense. The annotations are there to serve us, so whatever keeps it easy and keeps context.

DanVanAtta (approved):
LGTM

PR #7723, opened 2020-09-21
Refactor the BattleState/BattleActions retreat api

trevan:
This removes the `submergeUnits` and `queryRetreat` methods from the `BattleActions` and moves all of the retreat/submerge code in `MustFightBattle` to the individual steps or helper utility classes.

Also, part of this involved changing the `attackerWins`, `defenderWins`, and `nobodyWins` methods to always call `endBattle` and to have just one `endBattle(WhoWon)` exposed in `BattleActions`.  I plan to rework the `endBattle` method as well so that it is split better between `BattleState` and `BattleActions` (it is currently on `BattleState` but assumes that it has access to `BattleState` instead of being explicit).

There should only be one functional change: defending subs can now submerge if `SUBMARINES_DEFENDING_MAY_SUBMERGE_OR_RETREAT` is true and they have no other retreat territories.  Previously, defending subs would be able to submerge only if there was at least one other retreat territory.

## Testing

I've tested the following retreat scenarios:
* Defensive subs that can't retreat
* Defensive subs that can submerge
* Defensive subs that can submerge and retreat to other territories
* Offensive subs submerging by themselves
* Offensive subs submerging before other units retreat
* Offensive subs retreating with other units
* Partial Amphibious assault with non-amphibious land units and air units
* Partial Amphibious assault with air units
* Naval battles that have dependent amphibious assault battles
* Naval battles that have allied air units as cargo
* Naval battles that have allied air units participating

## Release Note

FIX|Defending subs can now submerge if "Submarines Defending May Submerge Or Retreat" is true even if there are no other retreat territories.


game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/retreat/DefensiveSubsRetreatTest.java
```diff
@@ -118,41 +129,200 @@ void hasNoNamesWhenCanNotSubmergeAndNoRetreatTerritories() {
...
+
+      defensiveSubsRetreat.execute(executionStack, delegateBridge);
+
+      verify(battleState).retreatUnits(BattleState.Side.DEFENSE, retreatingUnits);
+    }
+
+    @Test
+    void retreatHappensWhenHasNoRetreatTerritoriesButDefendingIsSubmersible() {
+      final GameData gameData =
+          MockGameData.givenGameData()
```
DanVanAtta, line 183: Codacy found an issue: [Unnecessary use of fully qualified name 'MockGameData.givenGameData' due to existing static import 'games.strategy.triplea.delegate.battle.steps.MockGameData.givenGameData'](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6208307)

game-core/src/test/java/games/strategy/triplea/delegate/battle/FakeBattleState.java
```diff
@@ -140,6 +153,9 @@ public void clearWaitingToDie(final Side... sides) {
     }
   }
 
+  @Override
+  public void retreatUnits(final Side side, final Collection<Unit> units) {}
```
DanVanAtta, line 157: Codacy found an issue: [Document empty method body](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6208307)

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/retreat/DefensiveSubsRetreatTest.java
```diff
@@ -118,41 +129,200 @@ void hasNoNamesWhenCanNotSubmergeAndNoRetreatTerritories() {
...
+      when(delegateBridge.getSoundChannelBroadcaster()).thenReturn(sound);
+      when(delegateBridge.getHistoryWriter()).thenReturn(historyWriter);
+      when(defender.getName()).thenReturn("defender");
+      when(battleSite.getUnitCollection()).thenReturn(battleSiteCollection);
+      when(battleSiteCollection.getHolder()).thenReturn(battleSite);
+    }
+
+    @Test
+    void retreatHappensWhenNotSubmersibleButHasRetreatTerritories() {
+      final GameData gameData = MockGameData.givenGameData().build();
```
DanVanAtta, line 238: Codacy found an issue: [Unnecessary use of fully qualified name 'MockGameData.givenGameData' due to existing static import 'games.strategy.triplea.delegate.battle.steps.MockGameData.givenGameData'](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6208307)

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/retreat/DefensiveSubsRetreatTest.java
```diff
@@ -118,41 +129,200 @@ void hasNoNamesWhenCanNotSubmergeAndNoRetreatTerritories() {
...
+    public void setupMocks() {
+      when(delegateBridge.getDisplayChannelBroadcaster()).thenReturn(display);
+      when(delegateBridge.getSoundChannelBroadcaster()).thenReturn(sound);
+      when(delegateBridge.getHistoryWriter()).thenReturn(historyWriter);
+      when(defender.getName()).thenReturn("defender");
+    }
+
+    @Test
+    void retreatHappensWhenHasNoRetreatTerritoriesButIsSubmersible() {
+      final GameData gameData = MockGameData.givenGameData().withSubmersibleSubs(true).build();
```
DanVanAtta, line 149: Codacy found an issue: [Unnecessary use of fully qualified name 'MockGameData.givenGameData' due to existing static import 'games.strategy.triplea.delegate.battle.steps.MockGameData.givenGameData'](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6208307)

Cernelius:
I wonder if destroyers (I mean units having `<option name="isDestroyer" value="true"/>` in the attachment) can make the submarines unable to submerge, in this case.

Since (besides optionally for non-IB 3rd edition) there are no such destroyers in the basic games using that property (no destroyers at all normally in non-IB 3rd edition, and destroyers units that are destroyers in name only in IB 3rd edition), I think good arguments can be made both in favour and against.

I tend to think that destroyers (still meaning units having `<option name="isDestroyer" value="true"/>`, regardless whether or not they are actually named "destroyer") should be able to negate this defensive-only submerge too, but this is merely a preference.

Basically, to clarify, the matter is that this property was added in v1.III, while the actual anti-sub destroyers were introduced only in v2, which removes the property in favour of "Submersible Subs", thus the two things never meet each other in the basic games, beside custom options (v1.III Iron Blitz has units called destroyers that aren't destroyers).

trevan:
@Cernelius , currently, the presence of an `isDestroyer` unit prevents defending subs from submerging or retreating.  This doesn't change that behavior.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
```diff
@@ -61,71 +97,101 @@ public void retreatUnits(final IDelegateBridge bridge) {
...
-  private boolean canAttackerRetreatAmphibPlanes() {
-    final GameData gameData = battleState.getGameData();
-    return (Properties.getWW2V2(gameData)
-            || Properties.getAttackerRetreatPlanes(gameData)
-            || Properties.getPartialAmphibiousRetreat(gameData))
-        && battleState.getUnits(BattleState.Side.OFFENSE).stream().anyMatch(Matches.unitIsAir());
+  private Change retreatOtherTerritoryUnits(
+      final IDelegateBridge bridge,
+      final Territory retreatTo,
+      final CompositeChange change,
```
DanVanAtta, line 186: Codacy found an issue: [Avoid unused method parameters such as 'change'.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6208307)

trevan:
I won't be able to fix the last remaining codeclimate issue.  I'll do a follow up PR where I work on merging the getUnits and related methods.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
```diff
@@ -61,71 +97,97 @@ public void retreatUnits(final IDelegateBridge bridge) {
...
-  }
+    final CompositeChange change = new CompositeChange();
 
-  private boolean canAttackerRetreatSeaPlanes() {
-    return battleState.getBattleSite().isWater()
-        && battleState.getUnits(BattleState.Side.OFFENSE).stream().anyMatch(Matches.unitIsAir());
+    change.add(retreater.extraRetreatChange(retreatTo, retreatUnits));
+
+    final Map<Retreater.RetreatLocation, Collection<Unit>> retreatingUnitMap =
+        retreater.splitRetreatUnits(retreatUnits);
```
trevan, line 146: I've thought about having the retreaters have a `getChanges()` method that returns a list of change objects.  These change objects would encapsulate the `retreatSameTerritoryUnits`, `retreatOtherTerritoryUnits`, and `extraRetreatChange` logic.  But I didn't think the existing `Change` objects fit this.  The `HistoryNode` objects are kind of a good fit since these "changes" already create history nodes but `HistoryNode` doesn't seem to have actions to perform.

The api of this "change" object would be something like:
```
interface BattleChange {
  void writeHistoryNode();
  Change perform();
}
```
And `perform` would build the actual `Change` objects as well as modify the `BattleState` (if needed).

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/EvaderRetreat.java
```diff
@@ -0,0 +1,143 @@
...
+    BattleActions battleActions;
+    BattleState.Side side;
+    IDelegateBridge bridge;
+    Collection<Territory> possibleRetreatSites;
+    Collection<Unit> units;
+    String step;
+  }
+
+  public static void retreatUnits(final Parameters parameters) {
+    final BattleState battleState = parameters.battleState;
```
DanVanAtta, line 34: Why are all of the parameters being aliased (IE: re-assigned) to new variables up front here? The alternative would be to just use the parameter values inline.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/EvaderRetreat.java
```diff
@@ -0,0 +1,143 @@
...
+    if (retreatTo == null) {
+      return;
+    }
+    playSound(battleState, bridge, retreatingPlayer, units);
+    if (battleState.getBattleSite().equals(retreatTo)) {
+      submergeEvaders(battleState, battleActions, units, side, bridge);
+      broadcastRetreat(bridge, retreatingPlayer, step, " submerges subs");
+    } else {
+
+      final CompositeChange change = new CompositeChange();
```
DanVanAtta, line 64: nit, mixture of abstractions code smell here. The other methods are high level, then this bit of functionality is very low level. Recommend to extract to a method so we can keep a consistent level of abstraction, eG: "updateHistoryWithUnitsMoved(..)"

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/EvaderRetreat.java
```diff
@@ -0,0 +1,143 @@
...
+      battleState.retreatUnits(side, units);
+
+      addHistoryRetreat(bridge, units, " retreated to " + retreatTo.getName());
+      notifyRetreat(battleState, battleActions, units, side, bridge);
+      broadcastRetreat(
+          bridge, retreatingPlayer, step, " retreats", " retreats subs to " + retreatTo.getName());
+    }
+  }
+
+  private static void playSound(
```
DanVanAtta, line 76: A method for a single line method call seems a overkill. Is the 'isHeadless' check truly needed? Even if it is, what do you think about inlining this?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/EvaderRetreat.java
```diff
@@ -0,0 +1,143 @@
...
+    broadcastRetreat(bridge, retreatingPlayer, step, messageShortSuffix, messageShortSuffix);
+  }
+
+  private static void broadcastRetreat(
+      final IDelegateBridge bridge,
+      final GamePlayer retreatingPlayer,
+      final String step,
+      final String messageShortSuffix,
+      final String messageLongSuffix) {
+    final String messageShort = retreatingPlayer.getName() + messageShortSuffix;
```
DanVanAtta, line 137: I think this method is here to avoid doing this same string concatenation in two places. IMO the indirection with overloaded method args is more harmful than duplicating two lines of code. WDYT of inlining this broadcast method @trevan ?

DanVanAtta (commented):
Reviewed about 30%, still quite a bit to go. Let a few comments that you could potentially start working on @trevan . As always thank you for the efforts to help improve things.

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
```diff
@@ -72,4 +86,8 @@ public boolean isFirstRound() {
   Collection<Territory> getAttackerRetreatTerritories();
 
   Collection<Unit> getDependentUnits(Collection<Unit> units);
+
+  Collection<Unit> getTransportDependents(Collection<Unit> units);
+
+  Collection<IBattle> getDependentBattles();
```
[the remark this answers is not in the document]
trevan: I'm not yet sure where `BattleActions` is headed.  I first thought I'd use it to contain all of the delegate logic but now that I've actually refactored a method that uses the delegate, I'm finding it not as useful.  `BattleState` might morph with `IBattle` or `AbstractBattle` once I start working on other battle classes.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -315,6 +319,57 @@ public BattleRound getBattleRoundState() {
...
+      final boolean submerge,
+      final String text) {
+    final Territory retreatTo =
+        remotePlayer.retreatQuery(
+            battleState.getBattleId(),
+            submerge,
+            battleState.getBattleSite(),
+            availableTerritories,
+            text);
+    if (retreatTo != null && !availableTerritories.contains(retreatTo)) {
```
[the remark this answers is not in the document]
trevan: This error handling was added back in 2005.  So it has been around for a while.  It has been a severe log since 2018.  I haven't seen the error show up in github so it is possible it never happens.

Returning null will just have the retreat not happen.  The game will continue as if the user never had a retreat option.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/CheckGeneralBattleEndOld.java
```diff
@@ -4,6 +4,7 @@
 import games.strategy.triplea.delegate.ExecutionStack;
```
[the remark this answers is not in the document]
trevan: The `Old` was to differentiate it from the non `Old` class.  Unfortunately, since they are part of the save game, we can't change their names :(

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/EvaderRetreat.java
trevan: I initially didn't have the parameters object but codeclimate complained about too many parameters.  So I switched to the object but didn't want to add `parameters.` in front of all of the places.  Just laziness.  If you want `parameters.` inlined, I'll do it.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/CheckGeneralBattleEndOld.java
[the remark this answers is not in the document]
DanVanAtta: This code is new since the last release (right?) -  sparingly, you can break prerelease save games. The latest release is where we need to absolutely need to guarantee compatibility unless we agree to do a major version release.  It's somewhat assumed that those that stumble upon a prerelease know what they are doing and compatibility is not 100% guaranteed.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/EvaderRetreat.java
DanVanAtta: In the places where it makes sense, perhaps most of them, please do.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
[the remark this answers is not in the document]
DanVanAtta: I wonder a bit if it is simply just dead code. Given the seeming remote call, one could picture a badly behaving client that would trigger this.

It is a shame that the diff marks don't show more clearly what is moved vs new.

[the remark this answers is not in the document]
trevan: It might be.  Looking at the original commit (https://github.com/triplea-game/triplea/commit/94dcbc456ee416c732facb9e512450d43ee98531), the message mentions that the messaging framework is still buggy.

Also, there was a situation in the original code where the selected territory wasn't part of the possible territories.  This happened when a submerge was offered.  But that code was changed sometime ago so that the possible territories included the submerge territory.

Do you want me to change it to an exception or just drop the code?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/CheckGeneralBattleEndOld.java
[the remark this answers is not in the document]
trevan: Yes, the CheckGeneralBattleEnd* classes were added in 2.3.21665 which I believe is a prerelease.  Also, I think I'm mistaken on the `Old` class.  It shouldn't be included in any save.  The `CheckGeneralBattleEnd` class, though, will be.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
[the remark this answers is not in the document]
DanVanAtta: Probably let's follow up with this, but soon as we have spent time on it and it's better to resolve it sooner than later.
IMO if we are pretty confident it is dead code, then drop it. If we think it's only possible if there is a buggy game client (like the server is okay, but the other person is running some sort of hacked up code), then perhaps the current handling is actually more or less okay.. It's hard to really know what to do with this one.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/CheckGeneralBattleEndOld.java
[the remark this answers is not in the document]
DanVanAtta: Any objects then to removing the 'General' word from the class names?

IMO this is just as descriptive of a name `CheckBattleEnd` as is: `CheckGeneralBattleEnd`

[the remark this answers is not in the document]
trevan: `General` stands for the "general phase" (that's what the rule books generally call it).  I would expect to add a `CheckFirstStrikeBattleEnd` and a `CheckAaBattleEnd` at some point.  So I'd like to keep `General` around.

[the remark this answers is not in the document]
DanVanAtta: In this context, the word `General` IMO simply adds nothing.

Seeing these two would be fine by me:
- CheckFirstStrikeBattleEnd
- CheckBattleEnd

I'll agree to disagree with you here. When you've lost context on this code and come back to it in a few weeks or months, perhaps in the back of your mind, check if you still have the same opinion and if the word 'general' adds more information or is actually more noise.

DanVanAtta:
@trevan ball I think is in my court to finish review on this. I think I should have some time later tonight to take another look.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/EvaderRetreat.java
```diff
@@ -0,0 +1,141 @@
...
+import lombok.Builder;
+import lombok.experimental.UtilityClass;
+import org.triplea.sound.SoundUtils;
+
+@UtilityClass
+public class EvaderRetreat {
+
+  @Builder(toBuilder = true)
+  public static class Parameters {
+    BattleState battleState;
```
DanVanAtta, line 24: nit, are the parameter values mutable? Can these be marked as final? It's perhaps also worthwhile to mark the non-null ones as `@Nonnull` so that they are checked when the builder is constructed. A downside of lombok builder is that it's not compile time safe, required args can be omitted, so generally it's important to at least get runtime validation with the `@Nonnull` annotation.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
```diff
@@ -61,71 +97,95 @@ public void retreatUnits(final IDelegateBridge bridge) {
...
+            getName(),
+            battleState.getAttacker());
   }
 
-  private boolean canAttackerRetreatPartialAmphib() {
-    if (!Properties.getPartialAmphibiousRetreat(battleState.getGameData())) {
-      return false;
+  private void retreatSameTerritoryUnits(
+      final IDelegateBridge bridge, final Collection<Unit> retreatingUnits) {
+    if (retreatingUnits != null && !retreatingUnits.isEmpty()) {
```
DanVanAtta, line 171: IMO the control flow would be better at the caller level. This method is a no-op otherwise an is just a void call from the caller. It provides some abstraction but a maintainer needs to jump around in the code to see that this is a potential no-op and when it would be a no-op. It also looks like the null check is dead-code.

I suspect some details are a bit hidden that probably could be exposed to the caller that would make the code easier to follow. For example, instead of this:

```
  @Override
  public Map<RetreatLocation, Collection<Unit>> splitRetreatUnits(
      final Collection<Unit> retreatUnits) {
    return Map.of(RetreatLocation.SAME_TERRITORY, retreatUnits);
  }
```

Perhaps the API should be:

```
Collection<Unit> splitRetreatUnits(
     RetreatLocation,
      final Collection<Unit> retreatUnits)
```

In this way it'd be more obvious that if we are calling a non-nullable method, then we'd always have a key value for 'retreatingUnits' and one would then be very comfortable removing the null check.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
```diff
@@ -61,71 +97,95 @@ public void retreatUnits(final IDelegateBridge bridge) {
...
-    final GameData gameData = battleState.getGameData();
-    return (Properties.getWW2V2(gameData)
-            || Properties.getAttackerRetreatPlanes(gameData)
-            || Properties.getPartialAmphibiousRetreat(gameData))
-        && battleState.getUnits(BattleState.Side.OFFENSE).stream().anyMatch(Matches.unitIsAir());
+  private Change retreatOtherTerritoryUnits(
+      final IDelegateBridge bridge,
+      final Territory retreatTo,
+      final Collection<Unit> retreatingUnits) {
+    if (retreatingUnits != null && !retreatingUnits.isEmpty()) {
```
DanVanAtta, line 182: ditto on burying control flow in private methods, this should probably be pushed up a layer so it's obvious to a caller that this is a no-op method in certain conditions. Arguably it's a bug whenever a no-op method is invoked (Did the caller know they would get a no-op, if so, why would they call a no-op method? If the caller did not know, is the caller in an unknown state and computing garbage data?)

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreat.java
```diff
@@ -59,13 +66,36 @@ public Order getOrder() {
...
+
+    final Collection<Unit> unitsToRetreat =
+        CollectionUtils.getMatches(
+            battleState.getUnits(BattleState.Side.OFFENSE), Matches.unitCanEvade());
+    if (unitsToRetreat.isEmpty()) {
+      return;
+    }
+
+    final Collection<Territory> retreatTerritories;
+    if (Properties.getSubmersibleSubs(battleState.getGameData())) {
```
DanVanAtta, line 83: nit, good place for a ternary here. Any thoughts or objections to that?

DanVanAtta: With a ternary done, one could even consider inlining and you'd remove the intermediate variable as well.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/Retreater.java
```diff
@@ -0,0 +1,32 @@
+package games.strategy.triplea.delegate.battle.steps.retreat;
+
+import games.strategy.engine.data.Change;
+import games.strategy.engine.data.Territory;
+import games.strategy.engine.data.Unit;
+import games.strategy.triplea.delegate.battle.MustFightBattle;
+import java.util.Collection;
+import java.util.Map;
+
+interface Retreater {
```
DanVanAtta, line 10: I think a javadoc would do well here to explain what this abstraction is representing exactly. WDYT?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/Retreater.java
```diff
@@ -0,0 +1,32 @@
...
+  enum RetreatLocation {
+    SAME_TERRITORY,
+    OTHER_TERRITORY;
+  }
+
+  Collection<Unit> getRetreatUnits();
+
+  Collection<Territory> getPossibleRetreatSites(Collection<Unit> retreatUnits);
+
+  String getQueryText();
```
DanVanAtta, line 21: This seems to be a UI concern, I'm not sure if it belongs in this API as then any implementation is forced to mix UI and logic. Perhaps the caller could do a lookup of query text based on the retreat type?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/Retreater.java
```diff
@@ -0,0 +1,32 @@
...
+
+  MustFightBattle.RetreatType getRetreatType();
+
+  Map<RetreatLocation, Collection<Unit>> splitRetreatUnits(Collection<Unit> retreatUnits);
+
+  Change extraRetreatChange(Territory retreatTo, Collection<Unit> retreatUnits);
+
+  String getShortBroadcastSuffix();
```
DanVanAtta, line 29: Ditto on UI concerns present in the API. Perhaps the caller can take care of this based on the retreat type.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/Retreater.java
```diff
@@ -0,0 +1,32 @@
...
+
+  Collection<Territory> getPossibleRetreatSites(Collection<Unit> retreatUnits);
+
+  String getQueryText();
+
+  MustFightBattle.RetreatType getRetreatType();
+
+  Map<RetreatLocation, Collection<Unit>> splitRetreatUnits(Collection<Unit> retreatUnits);
+
+  Change extraRetreatChange(Territory retreatTo, Collection<Unit> retreatUnits);
```
DanVanAtta, line 27: I wonder if we are mixing responsibilities with this "extraRetreatChange" API. The retreat code probably should not know much about creating a change object. 

Suggestion: have the caller check the battle type, if amphib then the change is 'empty' or do not even add it to the change stack, otherwise inline the implementation code into the caller.

DanVanAtta (commented):
Perhaps 70% done reviewing now, will need one more pass to look at some final parts of the changes and the test code.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/EvaderRetreat.java
```diff
@@ -0,0 +1,141 @@
...
+import games.strategy.triplea.delegate.battle.MustFightBattle;
+import games.strategy.triplea.formatter.MyFormatter;
+import java.util.ArrayList;
+import java.util.Collection;
+import lombok.Builder;
+import lombok.experimental.UtilityClass;
+import org.triplea.sound.SoundUtils;
+
+@UtilityClass
+public class EvaderRetreat {
```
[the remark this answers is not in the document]
trevan: `canEvade` units.  But I'll add a javadoc.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
trevan: Codeclimate complained about too much complexity in the `retreat` method so I created these helper methods to move the `if` statements away.

I, personally, don't see a problem with calling a no-op method when using abstractions.  In this case, the `Retreater` classes may or may not have these types of retreating instances.  The caller doesn't need to care.  It just needs to ask the `Retreater` for the data (which might be empty or null) and the helper method does the checks.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/Retreater.java
trevan: These methods are a "lookup of query text based on the retreat type".  That's what this API is for: handling a specific retreat type with its related change objects and UI text.  Something has to keep the two together.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/Retreater.java
trevan: I mentioned when I first created this PR about turning these into "change objects".  I don't think it fits with any of the existing "change" interfaces (`Change`, `HistoryNode` are what I've found) because it either can't invert (such as `Change`) or it isn't just a history text (such as `HistoryNode`).  Maybe at some point, we might want to allow it to invert so being an instance of `Change` would be best.  WDYT?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
DanVanAtta: In this case it hurts a bit as you lose context of why retreating units can be null (it can't). It makes it seem like this is called from multiple places. One needs to be careful about void method calls. They are generally poor abstractions. 

For example, what if someone moves where this invoked or adds something in between where the method is called and where these variables are set. 

Another scenario, what if you need an 'else' statement, in this case the abstraction breaks down completely since you are doing two different things. I think it is analgous to this:

```
   doPrinting(Connection c);
}

void doPrinting(Connection c) {
     if(c.connected) {
          sendToFax();
     }
}
```

vs:
```

if(c.connected()) {
  sendToFax();
}
```

And with an if/else branch:

```
   doPrinting(Connection c);
}

private void doPrinting(Connection c) {
     if(c.connected) {
          sendToFax();
     } else {
          sendToCopier();
    }
}
```

In the above, you can see it is just as well to have the if/else statement at the higher level. I think it also comes back to trying to favor having control flow in one place, at a high level, then keep the private methods that are invoked to be functional. 

Last comment, when a no-op method, or a method is carved off with control-flow, it often means the abstraction is cut at a slightly awkward level. The method is not quite control flow, not quite functional. Either the method should do more control flow and invoke other methods, or it should be just the body of the if statement. So  I believe from that it is a code smell.

DanVanAtta: I do think this is more cohesive and actually simpler since you (AFAIK) you don't have to worry about the empty change:

```
    final var retreatingUnits = retreatingUnitMap.get(Retreater.RetreatLocation.OTHER_TERRITORY);
    if (retreatingUnits != null && !retreatingUnits.isEmpty()) {
      change.add(retreatOtherTerritoryUnits(bridge, retreatTo, retreatingUnits));
    }

    bridge.addChange(change);

    if (battleState.getUnits(BattleState.Side.OFFENSE).isEmpty()) {
      battleActions.endBattle(IBattle.WhoWon.DEFENDER, bridge);
    } else {
      bridge.getDisplayChannelBroadcaster().notifyRetreat(battleState.getBattleId(), retreatUnits);
    }
```

DanVanAtta: Even more, the if condition of the private method is duplicated, so the control flow is actually kind loopy to understand. We can simplify by combining, eg:

```
    final Map<Retreater.RetreatLocation, Collection<Unit>> retreatingUnitMap =
        retreater.splitRetreatUnits(retreatUnits);
    final var retreatingUnits = retreatingUnitMap.get(Retreater.RetreatLocation.OTHER_TERRITORY);
    if (retreatingUnits != null && !retreatingUnits.isEmpty()) {
      retreatSameTerritoryUnits(
          bridge, retreatingUnitMap.get(Retreater.RetreatLocation.SAME_TERRITORY));
      change.add(retreatOtherTerritoryUnits(bridge, retreatTo, retreatingUnits));
    }

    bridge.addChange(change);

    if (battleState.getUnits(BattleState.Side.OFFENSE).isEmpty()) {
      battleActions.endBattle(IBattle.WhoWon.DEFENDER, bridge);
    } else {
      bridge.getDisplayChannelBroadcaster().notifyRetreat(battleState.getBattleId(), retreatUnits);
    }
```

DanVanAtta: To fix the complexity of this method, I would do the above, but also make the 'retreat' method actually do a retreat. It violates the "yoda" principle: "Do or do not, there is no try".

Specifically here:
```
    if (retreatTo == null) {
      return;
    }
```

In this case we have a 'retreat' method that is only conditionally doing a retreat.  If the method says 'retreat', I would expect it to do an unconditional retreat.

In this case I would do similar and push the control flow up a level so that this method has a SRP to it and does "retreat". In that case it then becomes the retreat sequence and no longer some of the logic if a retreat should be initiated *and* the actual sequence.

EG:
```
    if (retreater != null) {
      final Collection<Unit> retreatUnits = retreater.getRetreatUnits();
      final Collection<Territory> possibleRetreatSites =
          retreater.getPossibleRetreatSites(retreatUnits);
      final String text = retreater.getQueryText();
      bridge.getDisplayChannelBroadcaster().gotoBattleStep(battleState.getBattleId(), getName());
      final Territory retreatTo =
          battleActions.queryRetreatTerritory(
              battleState, bridge, battleState.getAttacker(), possibleRetreatSites, text);
      if (retreatTo != null) {
        retreat(bridge, retreater, retreatTo);
      }
    }
```

This makes the 'retreatUnits' method more worthwhile, more control flow is pushed up. Now if you get to 'retreat', you actually know that is executed and is not yet more control flow and potentially a no-op.

It makes the retreat method much more cohesive too IMHO:
```
  private void retreat(final IDelegateBridge bridge, final Retreater retreater, final Territory retreatTo) {
    final Collection<Unit> retreatUnits = retreater.getRetreatUnits();

    SoundUtils.playRetreatType(
        battleState.getAttacker(), retreatUnits, retreater.getRetreatType(), bridge);

    final CompositeChange change = new CompositeChange();
   :
   :
```

I would probably go on to rewrite the retreat method as follows:

```
  private void retreat(final IDelegateBridge bridge, final Retreater retreater, final Territory retreatTo) {
    final Collection<Unit> retreatUnits = retreater.getRetreatUnits();

    SoundUtils.playRetreatType(
        battleState.getAttacker(), retreatUnits, retreater.getRetreatType(), bridge);

    bridge.addChange(computeRetreatChange(...));
    
    if (battleState.getUnits(BattleState.Side.OFFENSE).isEmpty()) {
      battleActions.endBattle(IBattle.WhoWon.DEFENDER, bridge);
    } else {
      bridge.getDisplayChannelBroadcaster().notifyRetreat(battleState.getBattleId(), retreatUnits);
    }

    bridge
        .getDisplayChannelBroadcaster()
        .notifyRetreat(
            battleState.getAttacker().getName() + retreater.getShortBroadcastSuffix(),
            battleState.getAttacker().getName() + retreater.getLongBroadcastSuffix(retreatTo),
            getName(),
            battleState.getAttacker());
  }
```

I think I woudl stop there for rewriting that method since you really need to then pivot functionality out of the bridge and I think it hits diminishing returns.. 

IDK if this was overly pedantic for you, hopefully helpful maybe for some ideas, WDYT @trevan ?

trevan: I've reworked it.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/Retreater.java
DanVanAtta: If invert becomes a big problem, then we can probably compensate that by replaying the n-1 events. I think eventually we'll have a base-state that will be fixed that would allow us to do that. Overall I think the history probably should not be displaying every single change and changes would probably become more granular.

I think there is some groundwork that needs to be done before we can get there, so I wouldn't force objects to be 'change' objects just yet unless it really fits.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/Retreater.java
DanVanAtta: The UI text element seems to make this interface very implementation specific and even coupled to the UI. If the query text is a function of the retreat type, and that is only used in one place, IMO it's a stronger interface if it's made more focused and loses the UI concern. The methods in this interface are generally all related to model, hence it's a bit out of place for the UI parts.

trevan: I'd prefer having the logic around the different types of retreats together but I've gone ahead and removed the text strings methods.  The caller now has switch statements to figure out which text to use.

DanVanAtta: I'm getting into over-engineering territory, IMO that implies just having multiple interfaces.
`Retreater` & `RetreaterUiLabels`

Implementations can then implement both interfaces, the interfaces themselves would be cohesive/whole, and you can pass around the right interface type depending on your need, if you need the UI portion or if you need the business logic portion. I recognize that we probably have the two concerns mixed in well enough though that we'd either see one interface inherit from the other or both interface types being passed around, both of which would be somewhat ugly.

I appreciate you making that update though, thank you.

DanVanAtta (approved):
Changes look good @trevan , thank you.

Couple merge conflicts came up, good to merge once resolve.

PR #7740, opened 2020-09-23
Replace amphibiousLandAttackers list with checks for wasAmphibious

trevan:
As far as I can tell, the `amphibiousLandAttackers` should function almost identically to the `wasAmphibious` property.  It looks like `wasAmphibious` is only added when a unit is unloaded from a sea based transport.  I tested with land, sea, and air based transports in Iron War to check.

In the battle calculator, it would just set all of the attackers as `amphibiousLandAttackers`, so I replicated that behavior by setting `wasAmphibious` to all of the units.  It appears to work correctly.

There might be some small difference in behavior but I can't find one yet.  If there is a way to move a land unit from a sea territory without using a sea transport, then that would could cause a difference.

## Testing

Tested with land, sea, and air transports to make sure `wasAmphibious` is set only during sea transports.
Tested that the amphibious checkbox in the battle calculator updates `wasAmphibious` correctly.
Ran an amphibious battle and ensured that things were calculated correctly during the battle.


game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculator.java
```diff
@@ -103,11 +103,25 @@ public AggregateResults calculate(
         final MustFightBattle battle =
             new MustFightBattle(location2, attacker2, gameData, battleTracker);
         battle.setHeadless(true);
+        if (amphibious) {
+          attackingUnits.forEach(
+              unit -> {
+                unit.getProperty(Unit.UNLOADED_AMPHIBIOUS)
```
trevan, line 109: I have this code duplicated several times.  If you have any suggestions on how to shrink it, I'd love to know.

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
```diff
@@ -803,7 +797,7 @@ public static DiceRoll rollNDice(
                 UnitSupportAttachment::getStrength);
       } else {
         strength = ua.getAttack(unit.getOwner());
-        if (ua.getIsMarine() != 0 && isAmphibiousBattle && amphibiousLandAttackers.contains(unit)) {
```
DanVanAtta, line 806: Have you tested if marines are being given an amphib assault bonus on standard land battles? I'd double check that looking at this update. Have you run across where 'getWasAmphibious' is set to true (could you point me there)?

game-core/src/main/java/games/strategy/triplea/delegate/battle/AbstractBattle.java
```diff
@@ -57,7 +57,6 @@
 
   List<Unit> attackingUnits = new ArrayList<>();
   List<Unit> defendingUnits = new ArrayList<>();
-  List<Unit> amphibiousLandAttackers = new ArrayList<>();
```
DanVanAtta, line 60: This is a suspect I believe for save game compatibility issues. Have you double checked that this field is not serialized?

game-core/src/main/java/games/strategy/triplea/delegate/battle/AirBattle.java
```diff
@@ -705,7 +705,6 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
                       defendingUnits,
                       attackingUnits,
                       false,
-                      new ArrayList<>(),
```
DanVanAtta, line 708: Does the list need to be modifiable? If not, `List.of()` is the conventional preference for empty lists.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -145,14 +145,13 @@ public void setUnits(
       final Collection<Unit> defending,
       final Collection<Unit> attacking,
       final Collection<Unit> bombarding,
-      final Collection<Unit> amphibious,
+      final boolean amphibious,
```
DanVanAtta, line 148: WDYT of using an enum here to represent the different battle states? It would be purely for a more descriptive API

DanVanAtta (commented):
Thank you for the PR. I'm about half way through reviewing. I'll try to finish up review tomorrow or the day after.

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
trevan: `getWasAmphibious` is only set in `TransportTracker#unloadTransportChange` and that is when a transport is unloaded.  I just ran a game where I did an amphibious assault and then in the next turn used the same units to attack a standard land battle.  `wasAmphibious` was true for the first battle but false for the second battle.  I've also tested battles that don't have any amphibious assault and `wasAmphibious` is false there as well.

game-core/src/main/java/games/strategy/triplea/delegate/battle/AbstractBattle.java
trevan: Oddly enough, it didn't prevent loading an old save game.  But I'm adding them back in.

game-core/src/main/java/games/strategy/triplea/delegate/battle/AirBattle.java
trevan: This list was removed.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: Such as `BattleType`?  You recently modified it for the different air battles.  Maybe add `AMPHIBIOUS` there as well?

game-core/src/main/java/games/strategy/triplea/delegate/battle/Fire.java
```diff
@@ -46,7 +47,9 @@
   private final Collection<Unit> allFriendlyUnitsNotIncludingWaitingToDie;
   private final Collection<Unit> allEnemyUnitsNotIncludingWaitingToDie;
   private final boolean isAmphibious;
-  private final Collection<Unit> amphibiousLandAttackers;
+
+  @RemoveOnNextMajorRelease("amphibiousLandAttackers is no longer used")
+  private final Collection<Unit> amphibiousLandAttackers = List.of();
```
DanVanAtta, line 52: Codacy found an issue: [Avoid unused private fields such as 'amphibiousLandAttackers'.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6221536)

game-core/src/main/java/games/strategy/triplea/delegate/battle/FireAa.java
```diff
@@ -47,7 +48,10 @@
   private final List<Unit> allFriendlyUnitsAliveOrWaitingToDie;
   private final List<Unit> allEnemyUnitsAliveOrWaitingToDie;
   private final boolean isAmphibious;
-  private final Collection<Unit> amphibiousLandAttackers;
+
+  @RemoveOnNextMajorRelease("amphibiousLandAttackers is no longer used")
+  private final Collection<Unit> amphibiousLandAttackers = List.of();
```
DanVanAtta, line 53: Codacy found an issue: [Avoid unused private fields such as 'amphibiousLandAttackers'.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6221536)

game-core/src/main/java/games/strategy/engine/display/IDisplay.java
```diff
@@ -54,6 +55,8 @@ void reportMessageToPlayers(
    * @param attacker - PlayerId of attacker
    * @param defender - PlayerId of defender
    */
+  @RemoveOnNextMajorRelease(
+      "Remove isAmphibious, amphibiousLandAttackers, dependentUnits, and battleTitle")
   @RemoteActionCode(12)
   void showBattle(
```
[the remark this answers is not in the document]
trevan: This is only for showing the battle display.  It actually makes a copy of all of these parameters (see `BattlePanel#showBatle`).

It looks like the actual battle takes place on the client that is attacking.  The other clients have the battle display and this is initially populated with the data from the attacking client and then updated through calls from the attacking client.

This method looks like it could use the `BattleState` instead of the individual items.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
DanVanAtta: That is not a bad idea. I think there are two options:
1. a new enum with `AMPHIBIOUS` and `NON_AMPHIBIOUS`
2. add an entry for battle type

There pros & cons to either approach.

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
DanVanAtta: Interesting, that implies/confirms at one point a new way to track amphibious units was simply built without regard to the existing mechanism. This is then a very good example of SRP / DRY being violated. 

It might be worthwhile to walk through this code execution in a debugger to make sure it all looks good at that level. It sounds like the test cases you've ran through have been pretty comprehensive.

trevan: Yeah, I went through it with a debugger to make sure the value was what I expected.  If you have two units attacking and one was amphibious but the other goes by land, only the amphibious unit has `wasAmphibious`.

Looking at the history, it appears that `amphibiousLandAttackers` was first created to track if a battle was amphibious.  Then, when marine bonuses were added, the `wasAmphibious` was added but it was used oddly (it kind of determined if a battle was amphibious but wasn't used to check if the unit should get a marine bonus).  Then that code had concurrent issues because it talked to the battle tracker.  It was finally removed and replaced by plumbing `amphibiousLandAttackers` through.

I think this new way (using `wasAmphibious`) should work correctly and it should be simpler.

DanVanAtta: :+1:

trevan:
@DanVanAtta , is there a way to pass the pmd check for the unused private variables?  Maybe some sort of annotation?

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: I looked into what the `amphibious` was used for and after tracking down its usage, I don't think it is actually needed.  I think that 'amphibious' can be determined by looping through the units and seeing if one of them has a `wasAmphibious`.

I wasn't completely certain because of a few uses of `AbstractBattle#isAmphibious` but I felt that the use here could be removed.  So, I didn't do any enum.

game-core/src/main/java/games/strategy/triplea/delegate/battle/Fire.java
```diff
@@ -45,8 +46,12 @@
...
   private final Collection<Unit> allFriendlyUnitsNotIncludingWaitingToDie;
   private final Collection<Unit> allEnemyUnitsNotIncludingWaitingToDie;
-  private final boolean isAmphibious;
-  private final Collection<Unit> amphibiousLandAttackers;
+
+  @RemoveOnNextMajorRelease("amphibiousLandAttackers is no longer used")
+  private final boolean isAmphibious = false;
+
+  @RemoveOnNextMajorRelease("amphibiousLandAttackers is no longer used")
+  private final Collection<Unit> amphibiousLandAttackers = List.of();
```
DanVanAtta, line 54: Codacy found an issue: [Avoid unused private fields such as 'amphibiousLandAttackers'.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6221536)

game-core/src/main/java/games/strategy/triplea/delegate/battle/FireAa.java
```diff
@@ -46,8 +47,13 @@
...
   private final List<Unit> allFriendlyUnitsAliveOrWaitingToDie;
   private final List<Unit> allEnemyUnitsAliveOrWaitingToDie;
-  private final boolean isAmphibious;
-  private final Collection<Unit> amphibiousLandAttackers;
+
+  @RemoveOnNextMajorRelease("amphibiousLandAttackers is no longer used")
+  private final boolean isAmphibious = false;
+
+  @RemoveOnNextMajorRelease("amphibiousLandAttackers is no longer used")
+  private final Collection<Unit> amphibiousLandAttackers = List.of();
```
DanVanAtta, line 55: Codacy found an issue: [Avoid unused private fields such as 'amphibiousLandAttackers'.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6221536)

game-core/src/main/java/games/strategy/triplea/delegate/battle/Fire.java
```diff
@@ -45,8 +46,12 @@
   private final Collection<Unit> allFriendlyUnitsAliveOrWaitingToDie;
   private final Collection<Unit> allFriendlyUnitsNotIncludingWaitingToDie;
   private final Collection<Unit> allEnemyUnitsNotIncludingWaitingToDie;
-  private final boolean isAmphibious;
-  private final Collection<Unit> amphibiousLandAttackers;
+
+  @RemoveOnNextMajorRelease("amphibiousLandAttackers is no longer used")
+  private final boolean isAmphibious = false;
```
DanVanAtta, line 51: Codacy found an issue: [Avoid unused private fields such as 'isAmphibious'.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6221536)

game-core/src/main/java/games/strategy/triplea/delegate/battle/FireAa.java
```diff
@@ -46,8 +47,13 @@
   private final Collection<TerritoryEffect> territoryEffects;
   private final List<Unit> allFriendlyUnitsAliveOrWaitingToDie;
   private final List<Unit> allEnemyUnitsAliveOrWaitingToDie;
-  private final boolean isAmphibious;
-  private final Collection<Unit> amphibiousLandAttackers;
+
+  @RemoveOnNextMajorRelease("amphibiousLandAttackers is no longer used")
+  private final boolean isAmphibious = false;
```
DanVanAtta, line 52: Codacy found an issue: [Avoid unused private fields such as 'isAmphibious'.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6221536)

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -718,17 +709,13 @@ public void fight(final IDelegateBridge bridge) {
...
+        List.of());
     display.listBattleSteps(battleId, stepStrings);
     if (!headless) {
       // take the casualties with least movement first
-      if (isAmphibious()) {
-        CasualtySortingUtil.sortAmphib(attackingUnits, amphibiousLandAttackers);
-      } else {
-        CasualtySortingUtil.sortPreBattle(attackingUnits);
-      }
+      CasualtySortingUtil.sortPreBattle(attackingUnits);
```
trevan, line 718: I can't figure out why these sorts (`sortPreBattle` and `sortAmphib`) happen here.  I tested battles with them removed and things seemed to work the same.  I think that these methods probably existed before the more complex casualty sorter was built and no one got around to removing them.  I worried about removing them as part of this PR but I think they can be removed.

game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtySelector.java
```diff
@@ -308,11 +298,9 @@ private static void killAmphibiousFirst(final List<Unit> killed, final Collectio
                 .combatModifiers(
                     CombatModifiers.builder()
                         .territoryEffects(territoryEffects)
-                        .amphibious(amphibious)
+                        .amphibious(targetsToPickFrom.stream().anyMatch(Unit::getWasAmphibious))
```
trevan, line 301: I first looked at not checking for `amphibious` here and instead doing the `getWasAmphibious` check inside of the sorter.  But the sorter only looks at one unit of each type so if a battle had an amphibious marine and a non-amphibious marine, the results could be different depending on which unit was seen first.

I think the sorter needs to be looked at because, right now, if you have a partial amphibious attack (some marines are amphibious and others aren't), it sorts them as if all of the marines are amphibious.

game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtySortingUtil.java
```diff
@@ -10,46 +10,24 @@
...
-    if (amphibiousLandAttackers.contains(u1) && !amphibiousLandAttackers.contains(u2)) {
-      return -1;
-    } else if (amphibiousLandAttackers.contains(u2) && !amphibiousLandAttackers.contains(u1)) {
-      return 1;
-    }
-    final int m1 = UnitAttachment.get(u1.getType()).getIsMarine();
-    final int m2 = UnitAttachment.get(u2.getType()).getIsMarine();
-    return m2 - m1;
-  }
+  public static Comparator<Unit> compareMarines() {
```
trevan, line 21: Before I changed this class, I wrote some unit tests to make sure I understood the logic.  And I got unexpected results.  So, I "fixed" the class to match the tests.  Take a look at the tests to see if maybe I have an incorrect assumption on how this is supposed to sort units.

Also, as I mentioned earlier, I'm not sure this sort code is even needed.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
DanVanAtta: The casualty sorter I believe assumes units are provided in an already sorted ordering.

DanVanAtta:
> @DanVanAtta , is there a way to pass the pmd check for the unused private variables? Maybe some sort of annotation?

Yes,  `SuppressWarnings("unused")` should do it. You may have to play around though and look for the exact error from PMD and use that token instead `./gradles pmdMain` to check locally.

Cernelius:
For your information, at least for v3 and following rules sets, it is a wrong behaviour of TripleA to force assigning hits to sea borne units of the same type first: You should be always able to select the specific unit to take out, amongst the ones that can be targeted (of course, when two or more units are the same, it makes sense for TripleA to make you choose only the type and the number of that type, to simplify).

For example, if I attack a territory with 2 sea borne infantry units and 2 other infantry and take 2 hits, TripleA obliges me to take out the 2 sea borne infantry units. Instead, I should be offered these options:
- Take out the 2 sea borne units.
- Take out the 2 land borne units.
- Take out 1 sea borne unit and 1 land borne unit.

If you are wondering if there are cases in which you want to take out land borne units first, yes there are (even without accounting possible marine bonuses).

game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtySelector.java
DanVanAtta: That would lead to an over-evaluation of those units powers, likely an outright defect.

game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculator.java
DanVanAtta: I suspect `attackingUnits` would need to become a datatype. In this case you could put this logic behind a method. I don't think I'd advise to do that just yet, the amount of data copying could actually become significant. BattleCalculator is regrettably a giant mess *AND* performance sensitive, a horrible combination and place to be..

DanVanAtta:
@Cernelius that seems like a valid problem, I used the 'reference in a new issue' to create an issue from your last comment here: https://github.com/triplea-game/triplea/issues/7756

game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtySelector.java
trevan: Do you want me to try to fix this as part of this PR?  It isn't a new issue so this change isn't making it any worse.

DanVanAtta: new PR, bug fixes are best done as focused PRs. Ideally a test created to demonstrate the problem and then fix is relatively surgical.

One should also consider the blast radius of a roll-back. If something is wrong and there is a rollback, it's a shame to have other helpful changes also be rolled bakc.

game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculator.java
DanVanAtta: Another thought, the API of the 'setProperty' is wonky with the exception type. If there were type safe overloads of the method without a throws, then the repetition would not be much of a problem at all.

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/CasualtySortingUtilTest.java
```diff
@@ -0,0 +1,159 @@
...
+    when(unitAttachment3.getIsMarine()).thenReturn(2);
+    when(unitType3.getAttachment(UNIT_ATTACHMENT_NAME)).thenReturn(unitAttachment3);
+    final Unit unitWith2Marine = spy(new Unit(unitType3, player, gameData));
+    when(unitWith2Marine.getWasAmphibious()).thenReturn(true);
+
+    final List<Unit> units = Arrays.asList(unitWith2Marine, unitWith1Marine, unitWithNoMarine);
+
+    CasualtySortingUtil.sortPreBattle(units);
+
+    assertThat(units, is(List.of(unitWithNoMarine, unitWith1Marine, unitWith2Marine)));
```
DanVanAtta, line 102: nit, have you considered using the collection matcher? IE: `IsCollectionContaining`

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/CasualtySortingUtilTest.java
```diff
@@ -0,0 +1,159 @@
...
+    final Unit unitWith3MarineButNotAmphibious = spy(new Unit(unitType4, player, gameData));
+    when(unitWith3MarineButNotAmphibious.getWasAmphibious()).thenReturn(false);
+
+    final List<Unit> units =
+        Arrays.asList(
+            unitWith2Marine, unitWith1Marine, unitWithNoMarine, unitWith3MarineButNotAmphibious);
+
+    CasualtySortingUtil.sortPreBattle(units);
+
+    assertThat(
```
DanVanAtta, line 150: nit, a message on the assertion is desirably for most assertions to explain why we expect an exact condition to be true. Spelling out why we expect a test to be correct is important for maintenance

DanVanAtta (approved):
LGTM

PR #7749, opened 2020-09-24
Structure lobby login window so that it can be re-used

DanVanAtta:
To be able to do a lobby login prompt from other screens,
we need to decouple the successful login action from the
login screen.

This update restructures responsibilities so that the
caller of the login screen can inject the successful
login behavior.

Additionally, this update adds a flag that can disable
unregistered login. The reasoning is if we are requiring
a login to enable a functionality, then we need
the user to have or to create an account.

## Testing
- Did some smoke testing to make sure the lobby login window still looked good and worked.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/lobby/client/login/LoginPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/lobby/client/login/LoginPanel.java
@@ -54,20 +54,11 @@ final class LoginPanel extends JPanel {
   private final JButton cancel = new JButton("Cancel");
   private final JLabel passwordLabel = new JLabel("Password:");
 
-  LoginPanel() {
-    initializeComponents();
-    layoutComponents();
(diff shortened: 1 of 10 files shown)


game-core/src/main/java/games/strategy/engine/lobby/client/login/LoginPanel.java
```diff
@@ -54,20 +54,11 @@
...
 
-  LoginPanel() {
-    initializeComponents();
-    layoutComponents();
-    setupListeners();
-    updateComponents();
-  }
-
-  private void initializeComponents() {
+  LoginPanel(final LoginMode loginMode) {
```
DanVanAtta, line 57: Codacy found an issue: [Avoid really long methods.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6228179)

PR #7765, opened 2020-09-26
warcraft air transport issue

trevan:
In the AA firing phase, it will now exclude any units that are currently being transported.

## Testing

Ran a game with air transported units and ensured that they were not fired on until after they were landed.

## Release Note

FIX|Air Transported units in no longer cause battle errors in Warcraft: War Heroes

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/TransportTracker.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/TransportTracker.java
@@ -48,16 +48,6 @@ public class TransportTracker {
     return transport.getTransporting();
   }
 
-  /**
-   * @return Unmodifiable collection of units that the given transport is transporting.
-   * @deprecated Invoke unit.getTransporting(Collection) directly
-   */
-  @Deprecated
-  public static List<Unit> transporting(
-      final Unit transport, final Collection<Unit> transportedUnitsPossible) {
-    return transport.getTransporting(transportedUnitsPossible);
-  }
-
   /** @return Unmodifiable map of transport -> collection of transported units. */
   public static Map<Unit, Collection<Unit>> transporting(final Collection<Unit> units) {
     return transporting(units, TransportTracker::transporting);
@@ -83,11 +73,11 @@ public class TransportTracker {
   /**
    * Returns a map of transport -> collection of transported units. This method is identical to
    * {@link #transporting(Collection)} except that it considers all elements in {@code units} as the
-   * possible units to transport (see {@link #transporting(Unit, Collection)}).
+   * possible units to transport
    */
   public static Map<Unit, Collection<Unit>> transportingWithAllPossibleUnits(
       final Collection<Unit> units) {
-    return transporting(units, transport -> transporting(transport, units));
+    return transporting(units, transport -> transport.getTransporting(units));
   }
 
   public static boolean isTransporting(final Unit transport) {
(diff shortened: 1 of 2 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/TransportTracker.java
```diff
@@ -48,16 +48,6 @@ private static void assertTransport(final Unit u) {
     return transport.getTransporting();
   }
 
-  /**
-   * @return Unmodifiable collection of units that the given transport is transporting.
-   * @deprecated Invoke unit.getTransporting(Collection) directly
-   */
-  @Deprecated
-  public static List<Unit> transporting(
```
trevan, line 56: This method was only used in one place so I just inlined it.

game-core/src/main/java/games/strategy/triplea/delegate/battle/FireAa.java
```diff
@@ -78,7 +78,9 @@
       final List<Unit> allEnemyUnitsAliveOrWaitingToDie,
       final List<String> aaTypes) {
     this.attackableUnits =
-        CollectionUtils.getMatches(attackableUnits, Matches.unitIsNotInfrastructure());
+        CollectionUtils.getMatches(
+            attackableUnits,
+            Matches.unitIsNotInfrastructure().and(Matches.unitIsBeingTransported().negate()));
```
trevan, line 83: This is the actual fix.

DanVanAtta:
@trevan , Could you weigh in whether it's feasible to add a targetted test for this functionality?

DanVanAtta (approved):

game-core/src/main/java/games/strategy/triplea/delegate/TransportTracker.java
```diff
@@ -139,17 +129,10 @@ public static Change unloadAirTransportChange(
       return change;
     }
     assertTransport(transport);
-    if (!transport.getTransporting().contains(unit)) {
-      throw new IllegalStateException(
-          "Not being carried, unit:" + unit + " transport:" + transport);
```
trevan, line 144: This is the exception being thrown.  I removed it because it really is red-herring.  The code `transport.getTransporting()` goes through all the territories on the map to find the one territory that has the transport.  It then loops through all of the units in that territory and finds all the units that have `Unit#getTransportedBy` equal to the transport.  And then, this method just checks if the `unit` is in that list.

But if you look a few lines up (line 127), it grabbed the transport by calling `Unit#getTransportedBy`.  So `transport.getTransporting().contains(unit)` will always return true except for the odd situation of when the unit and its transporter are in different territories.  And in this case, the unit wasn't in any territory because it had been removed.

This check was originally added when transports kept a list of the units they were transporting and back then this check made sense.  But transports have changed and it now is redundant.

DanVanAtta:
Second, @trevan are you confident that transported air units, in this case, if the air transport were hit by AA that the transported units are also removed? I suspect such a change might be warrant a PR, though it's certainly an important element of correctness for the game rules.

trevan:
> @trevan , Could you weigh in whether it's feasible to add a targetted test for this functionality?

I don't think of any the maps that are available for tests can replicate the situation.  So, we would either need to create a new map or work out all of the mocks needed to run a battle with this specific situation.  I would worry that with all of the mocks, it will make this test very brittle.

trevan:
> Second, @trevan are you confident that transported air units, in this case, if the air transport were hit by AA that the transported units are also removed? I suspect such a change might be warrant a PR, though it's certainly an important element of correctness for the game rules.

I just ran a Iron War with an air transported unit attacking territory with an AA gun.  The AA gun shot down the airplane and the transported unit was removed and couldn't fire back.

Checking the code `MustFightBattle#remove` is grabbing dependent units and removing them so I'm pretty confident that transported units die when their transport dies.

DanVanAtta:
@trevan I was thinking something along the lines of creating a module that would be a data object of the units participating in a combat, it would be run through a function and the output would be the units eligible for target by AA. We can then test that given a set of units, some being transported, that we get the expected units as the output of that function. Then the logic that does a series of calculations, we would at least use that function in that place to determine the units eligible for AA fire.

trevan:
@DanVanAtta , the `FireAa` class has logic to calculate "firing groups".  I think if that was turned into a separate module and the transported check was moved there, then we could build tests around that.

I was planning on reworking the "firing group" calculation since similar logic is done in `FireAa`, `Fire` (with a slight difference between subs and non-subs), and the bombard code.  Mind if I wait till then before working on unit tests for this situation?

DanVanAtta:
@trevan was just curious for your thoughts and if it seemed feasible. Really it could maybe be as simple as extracting a few lines to a static method somewhere and then invoking that static method (keep it simple).

DanVanAtta:
please be sure to close the related bug reports if you would @trevan , thank you for fixing this, it's a good one to have fixed :+1:

PR #7767, opened 2020-09-26
Fix: Casualty OOL handles amphibious and non-amphibious units in same battle

trevan:
The casualty OOL now checks the wasAmphibious instead of just assuming that all units are amphibious if the battle is amphibious.

## Release Note

FIX|Non-amphibious units are sorted lower than similar amphibious units in AI casualty selection.


game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/CasualtyOrderOfLossesTestOnBigWorldV3.java
```diff
@@ -165,4 +161,115 @@ void infantryAndArtillery() {
...
+                .build());
+
+    assertThat(result, hasSize(4));
+    assertThat(
+        "Non amphibious marine only has attack of 2 since it doesn't get marine bonus",
+        result.get(0),
+        is(attackingUnits.get(3)));
+    assertThat(result.get(1), is(attackingUnits.get(1)));
+    assertThat("Amphibious marine has attack of 3", result.get(2), is(attackingUnits.get(2)));
+    assertThat(result.get(3), is(attackingUnits.get(0)));
```
trevan, line 210: I don't know why the armor is the last unit.  This test has the same units as the `improvedArtillery` test except that one of the marines isn't amphibious.  I expected the tank to be below the amphibious marine like in the `improvedArtillery` but for some reason it is above.  I'm not entirely sure if this is correct behavior or some bug.

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/CasualtyOrderOfLossesTestOnBigWorldV3.java
```diff
@@ -165,4 +161,115 @@ void infantryAndArtillery() {
...
+                .combatModifiers(
+                    CombatModifiers.builder().defending(false).territoryEffects(List.of()).build())
+                .player(testData.british)
+                .enemyUnits(List.of())
+                .battlesite(testData.france)
+                .costs(testData.costMap)
+                .data(testData.gameData)
+                .build());
+
+    assertThat(result2, hasSize(3));
```
trevan, line 270: I'd really like to test that the cache was used and that the rest of the code was skipped, but it is all static based so I can't `spy` on method calls to see what is called or not.  I manually stepped through the code for this test to verify that the cache was being used.

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/CasualtyOrderOfLossesTestOnBigWorldV3.java
DanVanAtta: ## 1st Casualty

The first casualty is clear to remove the non-amphib.

## 2nd Casualty

This leaves the follow powers (where the +1 is support bonus):

```
tank      3
marine    3 +1
artillery 2
```

f the artillery is taken, leaves total of 6
```
tank 3
marine 3
```

if the marine is taken, leaves total of 5
```
tank 3
artillery 2
```

if the tank is taken, leaves total of 6
```
marine 3+1
artillery 2
```

Removing the marine is a bad choice. This leaves a choice between artillery and tank with the same remaining power. Tank is more expensive and has better defensive power, so the choice is to remove artillery.

## 3rd Casualty

We are left with this choice:
```
tank 3
marine 3
```
Tank is more expensive and has better defensive value, so marine is the choice

## 4th casualty

Trivial case as we are left with only the tank.


## Final ordering

marine, artillery, marine, tank

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/CasualtyOrderOfLossesTestOnBigWorldV3.java
DanVanAtta: IMO that is good enough. Though I *think* the cache is actually only used once. All of this code is called to determine where units are to be placed on the combat board, and then later we go through this path with cache when computing supports.

I took a deep look at calculating supports some time ago but the performance was not super easy to match. If you're ever interested I could go over you where I left and what I think is a simple algorithm for this.

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/CasualtyOrderOfLossesTestOnBigWorldV3.java
```diff
@@ -165,4 +161,115 @@ void infantryAndArtillery() {
...
+  @Test
+  void nonAmphibiousMarineWithAmphibiousAssault() {
+    testData.addTech(new ImprovedArtillerySupportAdvance(testData.gameData));
+
+    final List<Unit> amphibUnits = new ArrayList<>();
+    amphibUnits.addAll(testData.tank(1));
+    amphibUnits.addAll(testData.artillery(1));
+    amphibUnits.addAll(testData.marine(1));
+
+    amphibUnits.forEach(
```
DanVanAtta, line 174: Would it be just as well to do set the property for just the marine rather than all units? That would remove the forEach loop

game-core/src/main/java/games/strategy/triplea/delegate/battle/UnitBattleComparator.java
```diff
@@ -48,7 +47,7 @@ public UnitBattleComparator(
       final Collection<TerritoryEffect> territoryEffects,
       final GameData data,
       final boolean bonus) {
-    this(defending, costs, territoryEffects, data, bonus, false, false);
+    this(defending, costs, territoryEffects, data, bonus, false);
   }
 
   public UnitBattleComparator(
```
DanVanAtta, line 53: Complete side note, I think we got an error report stating that this comparator is invalid. The error message was "contract violated". If the following are units, { u1, u2, u3}, and if {u1 < u2}, and {u2 < u3}, then the following should be true {u1 < u3}. Apparently it was not and we got an exception for that.

game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtyOrderOfLosses.java
```diff
@@ -239,29 +237,34 @@ void clearOolCache() {
...
+
+  static String computeOolCacheKey(
+      final Parameters parameters, final List<AmphibType> targetTypes) {
     return parameters.player.getName()
         + "|"
         + parameters.battlesite.getName()
         + "|"
         + parameters.combatModifiers.isDefending()
         + "|"
-        + parameters.combatModifiers.isAmphibious()
```
DanVanAtta, line 263: I do suspect the cache will be incorrect here. If we fight a battle with non-amphib units, then later those units are amphib, the cache will have a value when we would have preferred to have had a cache miss.

There were already some problems here, and this code is certainly not fully correct, *very* complicated, *and* performance optimized (a really bad combination). 

Any thoughts @trevan ? I suspect that this is a consideration we are building in that just breaks this cache. We would not want a new cache value for every permutation of possibly amphib unit, that really blows it up in size.

DanVanAtta: It's possibly an option to just allow the cache to be incorrect for amphib units.. though not ideal.

Cernelius:
I'm not sure what's up here and I've not tested recently. However, I can say that, in all my testing in the past, as far as I remember, the program always took sea borne units first as casualty (amongst the same type of units). I just did a quick test with Napoleonic Empires and I confirm that the stronger offloaded marines are taken before the weaker non-offloaded ones.

I suggest assuming that all units are amphibious if it doesn't make any possible difference (namely, amphibious partial retreat is false (v1 and v2 rules sets) and the unit type to have hits assigned has no marine bonuses or they are equal to 0).

A possible challenge of this, that should be documented in the code, is if, in the future, other features that may make a difference will be added (also, the ones I mentioned are the only ones I'm aware of).

I also know that you can choose in v3 and following, but I don't actually know how it works for marines in v1.iii (I assume you can choose, but you really always want to take the weaker non-offloaded ones first).

game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtyOrderOfLosses.java
trevan: The cache key includes whether the unit type is amphibious or not.  A non-amphibious marine would create a different cache key than an amphibious marine.  The `AmphibType` keeps track of amphibious or not.

trevan: I'll change the cache key to also check if there is a marine bonus.  If a unit type doesn't have a marine bonus, then there is no need to care if the unit is amphibious or not.

DanVanAtta: I'm afraid you'll need to check the cache performance before and after and make sure the size of the cache doesn't blow up too much. Maps where this is an issue are world at war or NWO where you can have very large numbers of units.

trevan:
> I'm not sure what's up here and I've not tested recently. However, I can say that, in all my testing in the past, as far as I remember, the program always took sea borne units first as casualty (amongst the same type of units). I just did a quick test with Napoleonic Empires and I confirm that the stronger offloaded marines are taken before the weaker non-offloaded ones.
> 
> I suggest assuming that all units are amphibious if it doesn't make any possible difference (namely, amphibious partial retreat is false (v1 and v2 rules sets) and the unit type to have hits assigned has no marine bonuses or they are equal to 0).
> 
> A possible challenge of this, that should be documented in the code, is if, in the future, other features that may make a difference will be added (also, the ones I mentioned are the only ones I'm aware of).
> 
> I also know that you can choose in v3 and following, but I don't actually know how it works for marines in v1.iii (I assume you can choose, but you really always want to take the weaker non-offloaded ones first).

@Cernelius , the code to kill amphibious units first is done in a different place.  The method is actually called `CasualtySelector#killAmphibiousFirst`.  It appears to only kill amphibious units first if "Partial Amphibious Retreat" is allowed.  This PR shouldn't affect that functionality (whether it is broken or not).  This PR will affect which units are "auto-selected" in the UI but once the user has selected their units, the `CasualtySelector#killAmphibiousFirst` code will then try and kill any amphibious units out of the unit types that were selected.

game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtyOrderOfLosses.java
trevan: I ran one round of world of war with hard ai.  I printed out the size of the ool cache right before it was cleared.  In case you don't know, the ool cache is cleared after each step, such as combat, production, non-combat move, etc.

The numbers were roughly the same.  In some cases, this new code had more entries and in other cases it had less.

I think the main reason for that is the old code added the `amphibiousLandAttackers` to the cache key.  This new code doesn't use that and instead uses `getWasAmphibious` but it is roughly the same affect.

DanVanAtta: My recollection is faulty, I though it was kept around for longer. That kinda makes it less useful than I had thought.. I really regret I did not stick longer with the rewrite of this code, much of it was deleted... Ah well, one day maybe.

I was thinking for a test to do something like a 300 vs 300 unit battle. We get a cache entry for each unit in the battle per side, but it's no different really as the amphib entry does not really blow that up. The WaW case is significant around round 10-20 of that map, a save game at that location is best, the army stacks get to be pretty outragous with outragous bship stacks (like 30-40 ships, similar number of carriers etc)...

trevan: I ran WaW with fast ai until I got to the middle of round 10.  I save it and then ran hard ai for a little under a round with the old code and the new code.  I summed up the maximum number of cache entries before each clearing.  With the old code, there were 66,495 cache entries.  With the new code, there were 69,573 cache entries.  So it does add more cache entries but not significantly.

I ran two large scale battles with WaW that I created using edit mode.  Both battles had 100 armored cars, artillery, elites, infantry, marines, mech. infantry, mot. infantry, and tanks.  One battle had a 50/50 split between amphibious and not.  The other battle had 51/49 split between amphibious and not.  I ran the 50/50 battle first.  WaW is low luck so the outcomes of the battles should be fairly deterministic.  With the old code, there were 3566 cache entries after both battles were finished.  With the new code, there were 2199 cache entries.

DanVanAtta: Thank you for the detailed check

DanVanAtta (approved):

PR #7782, opened 2020-09-26
Simplify BattleState api by combining methods

trevan:
This reduces the number of methods exposed in `BattleState`.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
@@ -33,13 +33,11 @@ public interface BattleState {
   enum UnitBattleStatus {
     // units that are either undamaged or damaged
     ALIVE,
-    // either ALIVE or CASUALTY
-    ACTIVE,
     // units that are dead but can still act
     // (such as defending units that haven't had their turn to return fire)
     CASUALTY,
     // units that are no longer in the game
-    DEAD,
+    REMOVED_CASUALTY,
   }
 
   @Value(staticConstructor = "of")
@@ -69,6 +67,12 @@ public interface BattleState {
   UUID getBattleId();
 
   Collection<Unit> getUnits(UnitBattleStatus status, Side... sides);
+
+  /**
+   * Allow callers to request units from two different states. The overloaded method is required
+   * because Java doesn't support multiple varargs in the same method
+   */
+  Collection<Unit> getUnits(UnitBattleStatus status1, UnitBattleStatus status2, Side... sides);
 
   void clearWaitingToDie(Side... sides);
 
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
@@ -6,6 +6,8 @@ import games.strategy.engine.data.Territory;
 import games.strategy.engine.data.TerritoryEffect;
 import games.strategy.engine.data.Unit;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.EnumSet;
 import java.util.UUID;
 import lombok.Getter;
 import lombok.Value;
@@ -40,6 +42,22 @@ public interface BattleState {
     REMOVED_CASUALTY,
   }
 
+  class UnitBattleFilter {
+    public static UnitBattleFilter ACTIVE =
+        new UnitBattleFilter(UnitBattleStatus.ALIVE, UnitBattleStatus.CASUALTY);
+    public static UnitBattleFilter ALIVE = new UnitBattleFilter(UnitBattleStatus.ALIVE);
+    public static UnitBattleFilter CASUALTY = new UnitBattleFilter(UnitBattleStatus.CASUALTY);
+    public static UnitBattleFilter REMOVED_CASUALTY =
+        new UnitBattleFilter(UnitBattleStatus.REMOVED_CASUALTY);
+
+    @Getter private final EnumSet<UnitBattleStatus> filter;
+
+    UnitBattleFilter(final UnitBattleStatus... status) {
+      this.filter = EnumSet.noneOf(UnitBattleStatus.class);
+      Collections.addAll(this.filter, status);
+    }
+  }
+
   @Value(staticConstructor = "of")
   class BattleStatus {
     int round;
@@ -66,13 +84,7 @@ public interface BattleState {
   @ChangeOnNextMajorRelease("Use a BattleId class instead of UUID")
   UUID getBattleId();
 
-  Collection<Unit> getUnits(UnitBattleStatus status, Side... sides);
-
-  /**
-   * Allow callers to request units from two different states. The overloaded method is required
-   * because Java doesn't support multiple varargs in the same method
-   */
-  Collection<Unit> getUnits(UnitBattleStatus status1, UnitBattleStatus status2, Side... sides);
+  Collection<Unit> filterUnits(UnitBattleFilter status, Side... sides);
 
   void clearWaitingToDie(Side... sides);
 
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
@@ -43,11 +43,11 @@ public interface BattleState {
   }
 
   class UnitBattleFilter {
-    public static UnitBattleFilter ACTIVE =
+    public static final UnitBattleFilter ACTIVE =
         new UnitBattleFilter(UnitBattleStatus.ALIVE, UnitBattleStatus.CASUALTY);
-    public static UnitBattleFilter ALIVE = new UnitBattleFilter(UnitBattleStatus.ALIVE);
-    public static UnitBattleFilter CASUALTY = new UnitBattleFilter(UnitBattleStatus.CASUALTY);
-    public static UnitBattleFilter REMOVED_CASUALTY =
+    public static final UnitBattleFilter ALIVE = new UnitBattleFilter(UnitBattleStatus.ALIVE);
+    public static final UnitBattleFilter CASUALTY = new UnitBattleFilter(UnitBattleStatus.CASUALTY);
+    public static final UnitBattleFilter REMOVED_CASUALTY =
         new UnitBattleFilter(UnitBattleStatus.REMOVED_CASUALTY);
 
     @Getter private final EnumSet<UnitBattleStatus> filter;
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
@@ -30,14 +30,15 @@ public interface BattleState {
     }
   }
 
-  enum UnitsStatus {
-    // units that have not been hit
+  enum UnitBattleStatus {
+    // units that are either undamaged or damaged
     ALIVE,
-    // both ALIVE and CASUALTY
+    // either ALIVE or CASUALTY
     ACTIVE,
-    // units that have been hit but not yet cleared
+    // units that are dead but can still act
+    // (such as defending units that haven't had their turn to return fire)
     CASUALTY,
-    // units that are dead
+    // units that are no longer in the game
     DEAD,
   }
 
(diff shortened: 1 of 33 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -363,7 +365,29 @@ public BattleRound getBattleRoundState() {
   }
 
   @Override
-  public Collection<Unit> getUnits(final Side... sides) {
+  public GamePlayer getPlayer(final Side side) {
+    return side == OFFENSE ? getAttacker() : getDefender();
+  }
+
+  @Override
+  public Collection<Unit> getUnits(final UnitsStatus status, final Side... sides) {
```
DanVanAtta, line 373: Do I understand right that we're doing more of a filter operation here rather than a straight get? If so, what would you think of naming this method 'filterUnits'?  Or 'getUnitsByStatus', or 'filterUnitsBySTatus'

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
```diff
@@ -30,10 +30,24 @@ public Side getOpposite() {
     }
   }
 
+  enum UnitsStatus {
+    // units that have not been hit
+    ALIVE,
+    // both ALIVE and CASUALTY
+    ACTIVE,
```
DanVanAtta, line 37: I'm a bit confused by 'active' being alive and casualty. The confusing part is if they are alive and casualty, then wouldn't they be 'casualty'? 

The enum maybe would be more descriptive as well called `UnitBattleStatus` as it's seemingly the units status relative to the current battle. 

The 'active' I think could use more clarity

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
```diff
@@ -30,10 +30,24 @@ public Side getOpposite() {
...
+    // units that have been hit but not yet cleared
+    CASUALTY,
+    // units that are dead
+    DEAD,
```
DanVanAtta, line 41: Would it make sense to call this "removed" and use that terminology in the comments above as well?

I notice: " but not yet cleared", which would imply this is a "CLEARED" status. "CLEARED" could mean a few other things, "REMOVED" on the other hand I believe would be unambiguous. 

Arguably, 'dead' and 'casualty' can mean the same thing, a slight rename would keep that distinction crystal clear.

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
```diff
@@ -65,20 +77,10 @@ public boolean isFirstRound() {
 
   Collection<Unit> getBombardingUnits();
 
-  Collection<Unit> getKilled();
```
DanVanAtta, line 68: Nice, many fewer methods here, good work  :+1:

game-core/src/test/java/games/strategy/triplea/delegate/battle/FakeBattleState.java
```diff
@@ -96,12 +94,34 @@
...
   }
 
   @Override
-  public Collection<Unit> getUnits(final Side... sides) {
+  public GamePlayer getPlayer(final Side side) {
+    return side == OFFENSE ? attacker : defender;
+  }
+
+  @Override
+  public Collection<Unit> getUnits(final UnitsStatus status, final Side... sides) {
```
DanVanAtta, line 107: The exact implementation in the fake as we have in the real is concerning. At some point this is not even a fake but just another battle-state implementation.  I don't know if there is a good solution for this problem that would fit within the scope of this PR. However, let's keep an eye on this as it seems to be a growing problem for the fake battle state.

DanVanAtta (commented):
Few comments/discussions on naming to work through, beyond that LGTM

game-core/src/test/java/games/strategy/triplea/delegate/battle/FakeBattleState.java
trevan: Yeah, it is turning into a real implementation.  I've started to think that it might be better to pull it out as a real implementation and change `MustFightBattle` to hold a transient instance of it.  Then instead of passing `MustFightBattle` around as a `BattleState`, it would instead pass this implementation.

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
trevan: Yeah, terminology is a little confusing.   Basically, there are 3 states a unit can have:

1. Unit has been removed from the battle
2. Unit is on the casualty strip but can still fire back
3. Unit is none of the above

The 2nd and 3rd case are generally requested together (such as when figuring out who can fire) so I added a 4th state to combine the 2nd and 3rd case.

So, the names I picked were:

1. DEAD
2. CASUALTY
3. ALIVE
4. ACTIVE (the union of CASUALTY and ALIVE)

Calling DEAD as REMOVED might add some confusion since retreated units are removed but they aren't dead.

I see the confusion where the term casualty generally includes both dead and wounded.  I thought of WOUNDED but a two-hit unit with only one hit is "wounded" but it wouldn't be in that group.  I choose CASUALTY since the rules generally call it the "casualty strip".

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
trevan: `filterUnits` sounds like it should take a collection of units and return a filtered subset.  This is more of a `get`.  I also don't think adding `ByStatus` to it is needed since the first parameter is the status so it already indicates you are getting by status.  But I could change it to `getUnitsByStatus`.

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
DanVanAtta: Perhaps: `ACTIVE, CASUALTY, REMOVED_CASUALTY`

IMO it would be good to refer to RETREATED rather than REMOVED, but perhaps beyond the point.

The union of the two other states is kinda odd.  Is it actually an intersection? Can a unit have multiple states? To resolve this, perhaps the callers should query for either CASUALTY or ALIVE and then it's just on the caller to query for either state?

trevan: I've updated the comments because the original comment said "ALIVE and CASUALTY" when it probably should have said "or".  It is a union of the two states, not an intersection.  A unit can only have one state.

The "ACTIVE" state is there to help ensure that code doesn't forget to check the CASUALTY state.  Right now, I think there are a few cases where it is forgetting.  I expect ACTIVE to be the normal state that callers will use and ALIVE and CASUALTY will be the special cases.

RETREATED doesn't exist yet but that is just because there is no current code that uses `BattleState` and that needs the retreated units.  It probably will exist at some point.

Is `REMOVED_CASUALTY` supposed to replace `DEAD`?  Because the `DEAD` state really are dead units.  They are no longer in the game.

DanVanAtta: > Is REMOVED_CASUALTY supposed to replace DEAD? Because the DEAD state really are dead units. They are no longer in the game.

Yes, I believe the two are quite synonymous, a removed casualty is removed from the game. The idea is to try and bake into the names a relationship so we can see the transition from the naming alone. 'casualty' -> 'removed-casualty'. 

-----

The union still strikes me as odd. Typically states and even enum values are well to be mutually exclusive. It's kinda confusing as the union is both states, so it could be confusing to think that you are one enum value when you are really another. For example, take "A", "B", and "C" is the union of "A" and "B". If someone is looking at code dealing with only "A" and "B", it could be confusing to get a "C" state back and or know that if a unit transitions to state "B" then it's a conditional transition and the unit might have to go to state "C".

Any thoughts to having callers to send the union? We have a vararg method already, I think one would only need to replace "ALIVE" to be its constituent components. That would make it clear at a caller level we are doing an 'or' between two states.

trevan: I thought of doing a method but java doesn't allow two varargs in the same method.  I've just pushed a change where I created an overloaded getUnits.  And I've removed ACTIVE.

But, maybe the confusion is that it is named "states".  You initially mentioned calling it `filterUnits`.  Maybe it should be changed to that.  What about `UnitBattleFilter` or something?  Because ACTIVE isn't a "state".  It is a filter that is the boolean OR of ALIVE and CASUALTY.  A unit is either ALIVE or CASUALTY.  And since I expect most callers to want both ALIVE and CASUALTY units, I wanted to make it simple for callers to just request a pre-built "OR" filter, called ACTIVE.

trevan: Here's another idea.  What about keeping the `UnitBattleStatus` and add a `UnitBattleFilter` that is defined like:

```
  enum UnitBattleFilter {
    ACTIVE(EnumSet.of(UnitBattleStatus.ALIVE, UnitBattleStatus.CASUALTY)),
    ALIVE(EnumSet.of(UnitBattleStatus.ALIVE)),
    CASUALTY(EnumSet.of(UnitBattleStatus.CASUALTY));

    EnumSet<UnitBattleStatus> status;
    UnitBattleFilter(final EnumSet<UnitBattleStatus> status) {
      this.status = status;
    }
  }
```

Then `getUnits` (or `filterUnits`) would take a `UnitBattleFilter`.

DanVanAtta: Interesting idea. My only concern is the amount of scaffolding/indirection, but we're still in a very reasonable realm here.

You might be able to absorb the filter method into the `UnitBattleFilter` method as well.

Some really rough code for that to give an idea:
```
public class UnitBattleFilter implements UnaryFunction<Unit> {
    ACTIVE = new UnitBattleFilter(UnitBattleStatus.ALIVE, UnitBattleStatus.CASUALTY));
   
   private final List<UnitBattleStatus> states;

    private    UnitBattleFilter(UnitBattleStatus ... states)  {
  }
```

trevan: I don't think I can absorb the filter method into the `UnitBattleFilter`.  The class will need to access the raw unit lists (`attackingUnits`, `defendingUnits`, etc.) from `MustFightBattle` so I'd have to make public methods for access to those lists and that takes us back to having a bunch of public methods.

trevan: Oh, I guess if the unit knew which status it was, then the unary function would work.  But unfortunately, the units don't actually know what their status is.  Only `MustFightBattle` does at this time.

DanVanAtta: I've been wondering if unit attachment should be collapsed into units. That's a big change. The filter enum seems like a good compromise though in the meantime.

trevan: That's a big change.  But even unit attachment doesn't store the battle status.  There's nothing on the unit nor anything attached to it (type, attachment, etc) that has the battle status.  It is solely managed by internal lists in `MustFightBattle`.  Which can mean that you can cause a bug where a unit is both alive, casualty, and dead at the same time.  You just have to add the unit to all three lists :)

That's kind of what caused the paratrooper issue.  The paratrooper was dead and so it was in the killed list.  But it was a dependent, so it was in the dependent list as well.  And air dependents are always considered alive.  So the paratrooper was both dead and alive :)

DanVanAtta:
@trevan some checkstyles came up:

>[ant:checkstyle] [WARN] /home/travis/build/triplea-game/triplea/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java:46:36: Static variable name 'ACTIVE' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9]*$'. [StaticVariableName]

> [ant:checkstyle] [WARN] /home/travis/build/triplea-game/triplea/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java:48:36: Static variable name 'ALIVE' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9]*$'. [StaticVariableName]

> [ant:checkstyle] [WARN] /home/travis/build/triplea-game/triplea/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java:49:36: Static variable name 'CASUALTY' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9]*$'. [StaticVariableName]

> [ant:checkstyle] [WARN] /home/travis/build/triplea-game/triplea/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java:50:36: Static variable name 'REMOVED_CASUALTY' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9]*$'. [StaticVariableName]

PR #7792, opened 2020-09-27
Use BattleState.Side to simplify code in SubmergeSubsVsOnlyAir

trevan:
This is mostly code simplification using the BattleState.Side enum.  I also added a check to ensure that there are units to submerge instead of trying to submerge an empty set.


DanVanAtta (approved):
Nice work :+1:

PR #7796, opened 2020-09-27
Move the LandParatrooper and MarkNoMovement logic to their steps

trevan:
This removes two methods from the BattleActions api and moves the logic to the individual battle steps.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
@@ -102,6 +102,8 @@ public interface BattleState {
 
   Collection<Unit> getDependentUnits(Collection<Unit> units);
 
+  void removeDependentUnits(Collection<Unit> units);
+
   Collection<Unit> getTransportDependents(Collection<Unit> units);
 
   Collection<IBattle> getDependentBattles();
--- a/game-core/src/test/java/games/strategy/engine/data/CompositeChangeMatcher.java
+++ b/game-core/src/test/java/games/strategy/engine/data/CompositeChangeMatcher.java
@@ -1,0 +1,42 @@
+package games.strategy.engine.data;
+
+import java.util.Arrays;
+import java.util.List;
+import lombok.AllArgsConstructor;
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+
+@AllArgsConstructor
+public class CompositeChangeMatcher extends ChangeMatcher<CompositeChange> {
+
+  private final List<ChangeMatcher<?>> changeMatchers;
+
+  @Override
+  protected boolean matchesSafely(final CompositeChange compositeChange) {
+    final List<Change> changes = compositeChange.getChanges();
+
+    if (changes.size() != changeMatchers.size()) {
+      return false;
+    }
+
+    for (int i = 0; i < changeMatchers.size(); i++) {
+      if (!changeMatchers.get(i).matches(changes.get(i))) {
+        return false;
+      }
+    }
+
+    return true;
+  }
+
+  @Override
+  public void describeTo(final Description description) {
+    description.appendText("CompositeChange <[");
+    changeMatchers.forEach(change -> change.describeTo(description));
+    description.appendText("]>");
+  }
+
+  public static Matcher<CompositeChange> compositeChangeContains(
+      final ChangeMatcher<?>... changes) {
+    return new CompositeChangeMatcher(Arrays.asList(changes));
+  }
+}
--- a/game-core/src/test/java/games/strategy/engine/data/CompositeChangeMatcher.java
+++ b/game-core/src/test/java/games/strategy/engine/data/CompositeChangeMatcher.java
@@ -6,6 +6,15 @@ import lombok.AllArgsConstructor;
 import org.hamcrest.Description;
 import org.hamcrest.Matcher;
 
+/**
+ * Matches all changes of a {@link CompositeChange} object
+ *
+ * <p>Each of the changes are matched against the requested matchers in the same order and will fail
+ * if the changes are not the same size or in the same order.
+ *
+ * <p>Example usage: assertThat(change, compositeChangeContains(propertyChange(property, newValue,
+ * oldValue)));
+ */
 @AllArgsConstructor
 public class CompositeChangeMatcher extends ChangeMatcher<CompositeChange> {
 
--- a/game-core/src/test/java/games/strategy/engine/data/changefactory/ObjectPropertyChangeMatcher.java
+++ b/game-core/src/test/java/games/strategy/engine/data/changefactory/ObjectPropertyChangeMatcher.java
@@ -1,0 +1,34 @@
+package games.strategy.engine.data.changefactory;
+
+import games.strategy.engine.data.ChangeMatcher;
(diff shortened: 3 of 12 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
```diff
@@ -102,6 +102,8 @@ public boolean isFirstRound() {
 
   Collection<Unit> getDependentUnits(Collection<Unit> units);
 
+  void removeDependentUnits(Collection<Unit> units);
```
trevan, line 105: There's now 3 methods dealing with dependents (`getDependentUnits`, `removeDependentUnits`, and `getTransportDependents`).  I'm looking into if I can simplify the api around dependent units.  I expect `getTransportDependents` and `getDependentUnits` can be combined somehow.  I also think that they might be combined with `filterUnits` by adding a new status `TRANSPORTING`.  But that will be a different PR once.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/change/MarkNoMovementLeftTest.java
```diff
@@ -22,24 +35,84 @@
...
+            .battleRound(1)
+            .headless(false)
+            .attackingUnits(List.of(givenNonAirUnitWithMovementLeft(BigDecimal.ONE)))
+            .build();
+    final MarkNoMovementLeft markNoMovementLeft =
+        new MarkNoMovementLeft(battleState, battleActions);
+
+    markNoMovementLeft.execute(executionStack, delegateBridge);
+
+    verify(delegateBridge).addChange(any());
```
DanVanAtta, line 73: The change object is the interesting thing, right? The structure and data of the change object we do expect to be in a certain way, that defines correctness? For example, if it were null, I think this test might still pass.

A captor could be used to grab the change object and then verify it. We are then pretty tied to implementation and we have another code smell where the code could change, still be correct, and the tests are breaking because they were testing implementation and not behavior.

So, captor is an option, another option is to have the `execute` method return the change object. In that case the test code would get a reference to it directly. IF we are limited by compatibility concerns (we might not be), but if we are, then the existing code could be converted to a facade and invokes a method that is visible to testing that does return a change object. 

WDYT?

trevan: I changed the code to use `any(Change.class)`.  That prevents null.

As for actually verifying the change object, I think that might get really tied to the implementation.  This particular change object is a `CompositeChange` with multiple `ObjectPropertyChange` changes in it.  So the test would have to check that it is a `CompositeChange` and then grab the inner changes, cast them to `ObjectPropertyChange`, and then verify that they are setting the correct property (`ALREADY_MOVED`) and that the new value and the old value (which currently have no getters) are what is expected.  Is there already a change matcher somewhere in the code base that would do this?

Another idea is to just compare the `toString` result of the change object.  Then the code would be something like:
```
    verify(delegateBridge).addChange(argThat(hasToString("CompositeChange <[Property change, unit:null owned by null property:alreadyMoved newValue:1 oldValue:0]>")));
```

DanVanAtta: Indeed, it sounds like the check is if the expected property and perhaps expected value are present anywhere in the change object. A matcher would be pretty slick since you could do an iterative search through the change object. 'toString' is a bit brittle as you rely on the the toString method, but it would work too. IMO the matcher is probably the more robust way to go. AFAIK there is no such matcher.

trevan: I've made a custom matcher for `ObjectPropertyChange` and `CompositeChange`.  As more `Change` objects are used, the matcher can be expanded.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/change/MarkNoMovementLeftTest.java
```diff
@@ -71,7 +73,12 @@ void nonAirWithMovementLeftAreMarkedAsMoved() {
 
     markNoMovementLeft.execute(executionStack, delegateBridge);
 
-    verify(delegateBridge).addChange(any(Change.class));
+    verify(delegateBridge)
+        .addChange(
+            argThat(
+                CompositeChangeMatcher.compositeChangeContains(
+                    ObjectPropertyChangeMatcher.propertyChange(
+                        "alreadyMoved", BigDecimal.ONE, BigDecimal.ZERO))));
```
DanVanAtta, line 81: nit, would you mind extracting 'altreadyMoved' to a constant? Is that possible? The idea there being the test will automatically update with the constant and it makes it clear as well where that value is coming from for any maintainers that are potentially looking at a failing test.

Second nit, by convention tests use static imports for most everything matcher related (I don't always like that one, but it is a relatively consistent convention)

game-core/src/test/java/games/strategy/engine/data/changefactory/ObjectPropertyChangeMatcher.java
```diff
@@ -0,0 +1,34 @@
...
+  private final String property;
+
+  private final Object newValue;
+
+  private final Object oldValue;
+
+  @Override
+  protected boolean matchesSafely(final ObjectPropertyChange item) {
+    return item.getProperty().equals(property)
+        && item.getNewValue().equals(newValue)
```
DanVanAtta, line 19: Any nullity concerns here?

What if the new value or old value were null, and we were matching for a null value?

game-core/src/test/java/games/strategy/engine/data/CompositeChangeMatcher.java
```diff
@@ -0,0 +1,42 @@
+package games.strategy.engine.data;
+
+import java.util.Arrays;
+import java.util.List;
+import lombok.AllArgsConstructor;
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+
+@AllArgsConstructor
+public class CompositeChangeMatcher extends ChangeMatcher<CompositeChange> {
```
DanVanAtta, line 10: When we get checkstyle upgraded to 8.3x, type javadocs will I think become required again. Probably best to try and add them proactively so we don't have to do as much to migrate.

Beyond checkstyle, a javadoc here with some example usages I think would be really helpful. Some sample code on how to use this would make it really easy if someone were want to pick this up, they could copy/paste it and modify it to fit their needs.

DanVanAtta:
@trevan , a checkstyle failed the build:

```
[ant:checkstyle] [WARN] /home/travis/build/triplea-game/triplea/game-core/src/test/java/games/strategy/engine/data/ChangeMatcher.java:6: Type Javadoc comment is missing @param <T> tag. [JavadocType]
```

The above is what makes me really not like our current checkstyle situation. The 'JavadocType' check is broken, unless you have a javadoc and you have a generic class, then you are *required* to describe the '@param <T>'. When we update to check style 8.36, I think it'll be a strong consideration whether to turn off the rule for javadoc'ing the generic type arg. Sometimes helpful, often not too though.

DanVanAtta (approved):
:+1:

Issue #6599, opened 2020-06-06
Infrastructure as Code - Ansible control for Forum


DanVanAtta:
It might be time for us to put forum software into ansible so that the configuration can be documented. 

Motivating reasons:
1. I don't know how the forum is deployed, what components are part of it, if there is a NGINX fronting it or not. If I do not know this, others will not either.
2. Upgrade process is opaque, maintenance operations just happen without notification or documentation of what is changed when.
3. No disaster recovery beyond restoring from linode
4. Infrastructure as code is a philosophy that says nobody should touch the running servers. You modify servers by modifying the checked in code and that manages your servers. This keeps a running history and audit trail of configuration and ensures documentation and transparency. 

@RoiEXLab would you be comfortable to start developing an ansible role that could set up a forum server? To start being able to control the nodeBB version and/or any NGINX server in front of the server would be excellent. Later on the list would be to be able to maintain and automate any database tasks.

RoiEXLab:
It's actually a little tougher than one might think. I'm not happy with the current state and I think a deployment redesign would be a far better choice actually.

The main problem I encounter every time when updating (which is also the reason I do it so rarely) comes down to the permission system.
Basically the basic idea is that every file related to nodebb is owned by the `nodebb` user which is a "virtual user", without home directory.
Now the problem with that is the npm utility.
See the basic update procedure is as follows (in theory):
1. Stop the NodeBB server if it's running (nginx should stay alive in order to display a nicer error page)
2. Checkout the most recent release branch (they create a new branch for every minor release in semver terms, something like `v1.13.x`), or simply pull if it's just a patch version change.
3. Run `./nodebb upgrade` (I think at some point you need to confirm that you want to update the installed plugins as well), which is basically a superset of `./gradlew build` which forces npm to download the latest dependencies and updates the database accordingly as well.
4. Start NodeBB again (we manage it via systemd btw, so `sudo systemctl start nodebb`)

Now the Problems occur in step 3.
If I run the upgrade command as my own user it will fail because my own user doesn't have any permissions inside the nodebb owned folders, also any created files would belong to my user instead of nodebb.
Using sudo doesn't help unfortuntely because nodebb doesn't run it's build scripts (to build native binaries for js libraries that require them) with root privileges (for a good reason), so they still fail.
[...]

PR #7799, opened 2020-09-27
Add preliminary Dockerfile for Node-BB

DanVanAtta:
- Allows for a container to be built
- Getting errors on run

Posting as a draft, help on this branch is welcome.


DanVanAtta:
Run error looks like this:
```
~/work/triplea/node-bb$ docker run ab7455bf1234
internal/modules/cjs/loader.js:968
  throw err;
  ^

Error: Cannot find module '/usr/src/app/nodebb'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:965:15)
    at Function.Module._load (internal/modules/cjs/loader.js:841:27)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
internal/modules/cjs/loader.js:968
  throw err;
  ^

Error: Cannot find module '/usr/src/app/nodebb'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:965:15)
    at Function.Module._load (internal/modules/cjs/loader.js:841:27)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```

RoiEXLab:
@DanVanAtta I believe the error is that the NodeBB source code is missing 😅 
So obviously it can't find a module that doesn't exist, simply because there's no source to begin with.  So a download step will probably fix the error already

node-bb/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    
+COPY . /usr/src/app
+
+ENV NODE_ENV=production \
+    daemon=false \
+    silent=false
+
+EXPOSE 4567
+
+CMD node ./nodebb build ;  node ./nodebb start
```
RoiEXLab, line 22: I'm not 100% sure if "nodebb start" (works without specifying `node` because of shebang) works as expected inside docker-
IIRC `./nodebb start` just spins up a daemon service and terminates itself soon after.
The current systemd script runs `/usr/bin/env node loader.js`, which spins up the process directly.

DanVanAtta: This docker file is verbatim from: https://hub.docker.com/r/nodebb/docker/

Your guess is as good as mine if there is anything missing.

DanVanAtta:
For context, I have not done too much here, but it's a start I suppose. I don't necessarily intend on following up too much if at all TBH. Ideally someone else would take this over.

So far:
- the package.json from the forum server is added
- a docker file is copied in from: https://hub.docker.com/r/nodebb/docker/

bacrossland:
I can help but I need to understand what your goal is. Are you trying to run a customized copy of NodeBB source code or just the current released version of NodeBB (with configurations for your site) in a Docker container?

DanVanAtta:
@bacrossland the help would be really appreciated!

The goal is to get as similar of a configuration as a our current forum. https://github.com/triplea-game/triplea/issues/6599 has some background on the steps @RoiEXLab has done to deploy it.

The larger goal is we want a docker container to simulate our forum and how that would interact with new dependencies. For example, if we were to add a new authentication module to the nodeBB forum and then use that for authentication, we would want the container available for local development to test & verify that configuration.

I think this will also be a case where we probably deploy the docker container to production and run docker in production. This initially seems like a good fit for something done in docker, tested locally, and then deployed as a docker container.

bacrossland:
Got it. I just need to read up a bit on where NodeBB stores it's configurations and customizations then I'll update this PR with a new Dockerfile.

bacrossland:
@DanVanAtta can you get in the package-lock.json as well?

bacrossland:
Scratch that. Don't check in the package-lock.json. I was going to use it to change the Docker build to an npm ci so it built only the using the locked packages. The problem with that is it forces future development on the forums into a longer process flow that would require updating the lock file that would be included in the Docker build for time you wanted to change a package. Leaving it out allows for a simple dev cycle of update package.json then run Docker build. Rinse repeat as needed.

bacrossland:
@DanVanAtta and @RoiEXLab I put in PR #7906 which contains my contribution to this PR.

Issue #4133, opened 2018-10-03
Incorrect handling of AA-fire and/or AA-fire casualty selection


panther2:
### Engine version
irrelevant

### Map name

wwII_v1, v2, v3, v3, v4, v5 , wwII_"any" (1940)

### Can you describe how to trigger the error? (eg: what sequence of actions will recreate it?)

Simply create situations that lead to AA-fire and resulting casualty selection with more than one plane of every type involved.

In brief this is what is incorrect:

v1: 
- needs to be investigated by someone more familiar with the ruleset, maybe similar to v2

v2: 
- incorrect timing of AA fire (different rules for OOB and LHTR apply, but both versions are affected)
- AA gun does not fire against groups of air units (if possible)
- attacker cannot choose casualties

v3:
- AA gun does not fire against groups of air units (if possible)
- attacker cannot choose casualties (casualty apparently is taken out on a random basis)

v4:
- AA gun does not fire against groups of air units (if possible)
- attacker cannot choose casualties (casualty apparently is taken out on a random basis)

v5:
- attacker cannot choose casualties (casualty apparently is taken out ~~on a random basis~~ following a somehow systematic "take-out error"), tested with factory-built-in AA gun against a bunch of bombers)

wwII_"any" (1940)
- attacker cannot choose casualties (casualty apparently is taken out ~~on a random basis~~ following a somehow systematic "take-out error") (see v5)

### Instead of this error, what should have happened?

In general (v1-v4) AA guns should fire against group of units if possible (e.g. against fighters as a group, against bombers as a group...). The attacker should be able to select casualties among those groups.
In v5 and 1940-games the attacker should be able to choose the casualties totally free (as long as no special rules apply).

In v2 (likely v1, too) all movement has to be plotted out before any AA fire is resolved.
Also we have AA-fire during CM, Combat and NCM phases in OOB-rules while in LHTR AA-fire takes place only during Combat.

### Any additional information that may help

The background, rules and clarifications are elaborated on in detail here:
https://forums.triplea-game.org/topic/989/open-aa-revised-bug-aa-fire-casualty-selection-issues

Regarding the _somehow systematic "take-out error"_, see especially 
https://forums.triplea-game.org/post/26839

~~(Don't be irritated by the tile of that thread, actually this is far from being minor ;-) )~~

This issue serves more or less as an initial collection of bugs.
I am happy to provide more detailled information in separate issues of course, in case someone is willing to fix them.

DanVanAtta:
[...]

PR #7823, opened 2020-10-03
Refactor roll dice, select casualties, and remove casualties steps

trevan:
This is a pretty large change but it is focused on just the `Fire` and `FireAa` classes and all helping classes and methods.

The original code would call a few methods on `MustFightBattle` (`fireOffensiveAaGuns`, `fireDefensiveAaGuns`, `fireNavalBombardment`, and `findTargetGroupsAndFire`) and those methods would instantiate `FireAa` (for the first two methods) and `Fire` (for the last two methods).  The two fire classes would then instantiate three IExecutables that would "roll the dice", "select the casualties", and "remove the casualties" (though it only changes the units status to `CASUALTY` and NOT to `REMOVED_CASUALTY` so the name might be a little confusing.).

I went through both `FireAa` and `Fire` and took out the common functionality and created three new step classes `RollDice`, `SelectCasualties`, and `MarkCasualties`.  The `FireRoundState` tracks the state between the three steps.  This state was previously tracked as class members of `Fire` and FireAa`.

The differences between aa, bombardment, and "normal" fire is now managed by a set of `FiringGroupFilter`s and two sets of BiFunctions.  The implementations of `FiringGroupFilter` (`FiringGroupFilterBombard`, `FiringGroupFilterGeneral`, and `FiringGroupFilterAa`) generate the different firing/targeting groups (such as AA groups, Subs vs non air, Air vs non subs, etc).  The two sets of BiFunctions handle the differences in dice roll and casualty selection.

With these changes, the Battle UI step names will be different.  Previously, each AA firing group had a "fire", "select", and "remove" line while the bombardment group only had a "fire" and "select" line and the other groups were rolled up in a single "fire" and "select" line.  Here's an example of what it originally looked like:

BombardmentPre

Here's what it will now look like:

BombardmentPost

Notice that there is a "fire", "select", and "remove" line for all of the firing groups.

Another change is how "air vs subs" is displayed in the Battle UI.  Previously, a single line was shown "Air defend non subs" or "Air attack non subs".  Here's an example:

AirVsSubsPre

Now, that line is no longer going to be shown and instead each of the firing groups get their own set of "fire", "select", and "remove".  I've added special coding to detect the air vs subs to give it a better name.  Here's what it looks like now:

AirVsSubsPost

In addition to the new names, the Battle UI will actually highlight them.  So when there is different firing groups, each time the dice is rolled, it will highlight which firing group is doing it.  It previously worked this way only for AA firing groups.  If there were multiple other groups, it would just highlight the same "<Player> fire" line or "<Player> select casualties" until all the groups were done.

## Testing

I've run a Hard AI run of WWII 1940 world for 12 rounds.  No errors were found.

I ran multiple of my old saves to verify compatibility.  There is potentially some issues if some of the new step names don't match the old step names.  I've fixed most of them but there might be some really rare cases.  If these issues do happen, the UI will just say that "<step name> wasn't found" and continue.  And this will only affect the current round of the save game.  So it should be very minor.

I ran AA flyover battles, strategic bombing battles, amphibious, land, sea, and sub battles.

I think the "Remove" lines need to be renamed or maybe even removed.  The units are only removed after all of the AA firing groups have finished, after all of the first strike units have finished, and after all of the other units have finished.  There is currently a generic "Remove casualties" that is shown at the end of the first strike groups and the other groups.  There isn't one after the AA groups.  I think the "Remove casualties" line should be after all of the AA groups and that the other "remove" lines should be renamed to something else.  I'm not exactly what to call it though.

## Release Note

CHANGE|Strings in Battle UI have been updated to be more specific


game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/DefensiveFirstStrikeTest.java
```diff
@@ -128,4 +114,27 @@ void getStep(final List<BattleStateVariation> parameters, final BattleStep.Order
...
+                givenAnyUnit())
+            .stream()
+            .filter(DefensiveFirstStrike.FIRING_UNIT_PREDICATE.apply(gameData))
+            .collect(Collectors.toList());
+
+    assertThat(units, hasSize(2));
+    assertThat(
+        "There should be only first strike units",
+        units.stream().allMatch(Matches.unitIsFirstStrike()),
+        Matchers.is(true));
```
DanVanAtta, line 138: Codacy found an issue: [Unnecessary use of fully qualified name 'Matchers.is' due to existing static import 'org.hamcrest.Matchers.is'](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6277681)

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroupTest.java
```diff
@@ -0,0 +1,156 @@
...
+    unitType.addAttachment(UNIT_ATTACHMENT_NAME, unitAttachment);
+    unitAttachment
+        .getProperty("isSuicideOnHit")
+        .ifPresent(
+            property -> {
+              try {
+                property.setValue(suicideOnHit);
+              } catch (final MutableProperty.InvalidValueException e) {
+                // should not happen
+                System.out.println(e.getMessage());
```
DanVanAtta, line 55: Codacy found an issue: [System.out.println is used](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6277681)

Cernelius:
**Can you please bring back the ability to target infrastructures with AA?** That is a change that I guess happened accidentaly in a previous refactor like this one.

For example, an air unit getting shot in a territory with an aaGun, no matter if the air unit is an infrastructure.

---

Also, for you information, the current behaviour of the engine is wrong on several points:

In v2, naval bombardment happens at the same time as AA fire (the program has AA fire happening first).

In v3, naval bombardment happens before AA fire (the program has AA fire happening first).

In v1 (the program default), naval bombardment casualties are able to fire back, meaning that there is not the removal of bombard casualties immediately after selecting them, and they are removed in the normal remove step (the program has a property called "Naval Bombard Casualties Return Fire" for having the v1 (as well as v3 and later) behaviour, but how it should work, instead, is that the "Return Fire" behaviour happens as default and you have a property that removes bombardment casualties immediately if true and it is automatically set true if the property "WW2V2" is true).

CC: @panther2

trevan:
@Cernelius , can you give me a save where that type of battle is ready to start?  That will ensure that I'm actually testing the correct scenario.

For your v2 comment, you say that naval bombardment should happen at the same time as the AA fire.  How exactly would that affect the actual behavior of the game?  Should the naval bombardment and AA casualties be removed only after both of them have fired?  Or is there some other behavioral change.

For your v3 comment, is there any rule set where naval bombardment should happen after AA?  Should I just switch the order of naval bombardment with AA?

For your v1 comment, I don't understand what you mean.  You seem to be describing what the game is already doing so I'm not sure what type of changes you are asking for.

DanVanAtta:
> That is a change that I guess happened accidentaly in a previous refactor like this one.

To nitpick terminology, refactor means changing the structure of well tested code without updating the tests. Because the code is well tested the behavior remains the same. Otherwise, it's a rewrite, most of what we've been doing has been rewrites and very little refactoring.

@Cernelius please raise such create a bug report when you notice such problems. The less time passing between a problem being introduced and noticed tends to reduce the cost for a fix.

---------


@trevan it's going to take perhaps up to two weeks to review this PR and perhaps another week or two to then work through any updates. I really recommend this be broken up. WDYT?

For ideas of how you can do that:
- all new classes can be submitted individually in parallel
- tests submitted on their own
- once the above is merged then incrementally kill code and replace with invocation to the new classes that were previously unused

You might also be able to make use of the beta feature flag such that we use new code flows.

I realize this is going to take a lot of effort on your part (speaking from personal experience here). As-is, I think we should target this to be around 4-6 PRs, perhaps more if we can really parallelize the new code and submit it on its own.

Cernelius:
> @Cernelius , can you give me a save where that type of battle is ready to start? That will ensure that I'm actually testing the correct scenario.

I've just made a game modification of "pos2", the only difference being that "air_transport" is an infrastructure. In this savegame, Germans should lose 1 air_transport:
[2020-10-3-Pact-of-Steel-2-Cernel(1).zip](https://github.com/triplea-game/triplea/files/5321130/2020-10-3-Pact-of-Steel-2-Cernel.1.zip)

> For your v2 comment, you say that naval bombardment should happen at the same time as the AA fire. How exactly would that affect the actual behavior of the game? Should the naval bombardment and AA casualties be removed only after both of them have fired? Or is there some other behavioral change.

This is a tough call, and one that I'm not sure I feel like doing myself. In v2, naval bombardment and AA fire happen during the same step, and also the selection of their casualties happen during that step (while their removal happen on an immediately subsequent step). However, the selection of AA fire casualties hardly matters, as they are already rolled per type (in the basic games, you roll separately for fighters and bombers, so you substantially have two separate AA fire situations, one for the fighters and one for the bombers, happening at the same time). Despite my clarification (that I hope and believe it is correct), I actually tend to suggest keeping the current behaviour as is for v2, except only having the (already selected) casualties removed on an immediate subsequent step, after the step in which the last opening fire casualties are selected (whichever they are). In practice, this would be the step order I suggest (not necessarily the only possible or even correct one, since the matter is not very clear), for v2 only:
- AA Fire (split per type, and further split if some of the types are air transporting (but this doesn't matter for basic v2, since there are no paratroopers in such games), though I believe TripleA incorrectly rolls one dice per aircraft (instead of a number of dice per type equal to the total number of aircraft of that type, as it should), but this is yet another problem).
- AA Fire casualties selection (intra type, though I believe TripleA incorrectly handles it as purely random (assigning 1 dice per aircraft and rolling it), but this is yet another problem).
- Offshore Bombardment (here there is yet another problem, since in v2 this should not happen if the embattled territory has only air units defending).
- Offshore Bombardment casualties selection (here there is yet another problem, since in v2 air units should not be selectable here).
- Offending Submarines Fire.
- Offending Submarines Fire casualties selection (meaning that the defender selects them).
- Defending Submarines Fire.
- Defending Submarines Fire casualties selection (meaning that the offender selects them).
- AA Fire/Offshore Bombardment/Submarines Fire casualties removal (Submarines Fire casualties are not removed at this point if the same side has 1 ore more destroyers (this raises the matter about what should happen if all of the destroyers are removed now because of non-submarines opening fire, which can never happen in the basic games (for example, if you have a custom game in which there is offensive AA Fire that can target destroyers))).
- Combat infrastructures are removed from the battle and returned to the zone of the battle (unless they have been already removed as casualties, which can never happen in the basic games, as the only combat infrastructure is the AA Gun) (since TripleA supports infrastructures firing normally and also over multiple rounds of combat, as well as offending infrastructures (that arguably should be able to retreat with the rest), such combat infrastructures should not be removed from the battle, but just made impossible to be targeted during each combat round from this point onwards (I think it is sensible the most to assume that the basic games remove all infrastructures at this point just because, in these games, there would be no point keeping them in the battle any longer, but TripleA should not remove them, otherwise)).

Moreover, Rockets fire also happens at the same time as AA Fire and Naval Bombardment, but I tend to think it is fine as long as it is anywhere in the middle of them or immediately before or after any one of them.

For more information about the incidental issues of AA fire casualties selection, refer to this issue (and preferably ask @panther2, if you need any clarification):
https://github.com/triplea-game/triplea/issues/4133

> For your v3 comment, is there any rule set where naval bombardment should happen after AA? Should I just switch the order of naval bombardment with AA?

I'm almost sure the current behaviour is correct for v1, as far as making the AA Fire before the Naval Bombardment goes. This may be correct or advisable for v2 too, as I said, even though I'm not as sure.

> For your v1 comment, I don't understand what you mean. You seem to be describing what the game is already doing so I'm not sure what type of changes you are asking for.

What I believe the program is doing is that Naval Bombardment casualties are removed immediately after selecting them unless 
"Naval Bombard Casualties Return Fire" is true, in which case they are removed when casualties from normal (offensive) fire are removed (that is, if this property is true, an infantry killed by a battleship is removed as the same time as an infantry killed by an offending infantry).

What I'm as sure as I can be that the program should be doing is:

For v1 and v3, as well as every game after v3, the behaviour currently given when "Naval Bombard Casualties Return Fire" is true should be the default (needing no property being true), in keeping with v1 being the default.

For v2 only, there should be a property that, when true, casualties from Naval Bombardment are removed immediately after all opening fire actions are resolved (which means this should happen immediately after the last of AA Fire, Rocket Fire, Naval Bombardment, Submarines Fire, whichever it is (for the basic games, this substantially equals removing casualties from Naval Bombardment immediately as they are selected)).

Making this change should fix all v1 Rules games, that currently are missing the fire back from naval bombardment, if they are missing having the "Naval Bombard Casualties Return Fire" true. I suggest this property being deprecated and rendered ineffective (or entirely removed from the program). All games having the "Naval Bombard Casualties Return Fire" set false should probably be updated with the new property, unless they already have the property "WW2V2" set true (as this property should set the new property true automatically).

The new property (for the v2 only no fire back) can be called something like "Naval Bombardment Casualties Are Removed Immediately" (which would imply immediately after everything happening at the same time, so if a unit able to first strike is hit by naval bombardment (this can never happen in the basic games), it would still do its first strike fire, on that round of combat).

---

@panther2 Can you please check and possibly confirm everything I said at this post and the one before?

Cernelius:
> @Cernelius please raise such create a bug report when you notice such problems. The less time passing between a problem being introduced and noticed tends to reduce the cost for a fix.

The problem is that TripleA has so many problems and wrong behaviours that, if I would report everything I know, that would be really a massive amount of issues, so what I do is making a selection myself and trying to focus on what I believe is most important (and, even in that case, the majority of times I report something the problem is not fixed). So, while you are correct in theory, in practice the matter for me is to find a balance between the effort of documenting and discussing the issues and my perceived chance at seeing it actually fixed (I would just never report anything if I would believe that such chance is 0%).

Aside from this, there is a chance that at least some of the issues I raised in here I've already reported, maybe in one or more closed issues (I cannot remember everything).

DanVanAtta:
@Cernelius, I respect the work you do. You raise a whole new topic. Though, now we have different participants with different concerns that all need to read everything that everyone has written. This is not efficient. 

Please move the topics that have been raised into their own issues. PRs are for discussing the specific code that has changed. Instead of raising related topics in a PR, create an issue and link to the PR.

DanVanAtta:
@Cernelius I really urge you to use the "reference in a new issue" so that your comments do not get lost.

Please consider the goal of discussion in the different kinds of documents..
- In PRs, the goal of a discussion is for changes to the already changed lines of code and to within the intent of the change. This update here is to change *nothing* but the internal structure. Asking for any kind of a functional change is automatically not going to be done. When this PR is closed, that conversation is then lost.
- In issues, the goal is to either, state a bug report, or b) discussion such that other task reports are created. The end goal is a are issues that are each under 200 words each concisely stating a problem and what the desired fix is.

The discussion started here is immensely valuable, but it falls squarely into the second category. Let's continue the meta-conversation of what is appropriate where in: https://github.com/triplea-game/triplea/issues/7825

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
```diff
@@ -945,15 +953,14 @@ private static DiceRoll rollDiceLowLuck(
       final boolean defending,
       final GamePlayer player,
       final IDelegateBridge bridge,
-      final IBattle battle,
+      final Territory location,
```
DanVanAtta, line 956: Converting this one parameter is perhaps of a good single PR that could be split off.

DanVanAtta:
I'm torn, there is valuable conversation and valuable code changes here. From personal experience I know how not fun it is to redo changes in a more incremental manner. Though, my comment in: https://github.com/triplea-game/triplea/pull/7823#issuecomment-703052696 stands. This needs to be broken up to be effectively reviewed.

DanVanAtta:
Some back of the envelope math:
- a typical reviewer can do 500 LOC / session
- a typical session is 30 - 60 minutes, beyond that and review effectiveness drops off
- assuming one review session per day, best case, for 5500 changes lines, that is 11 session. That means best case, it's almost two weeks for this have a first pass review done.
- Worst case, this is a hobby, we get time only every 3rd night perhaps, which triples the time, in that case the first pass review of this is done around November 7th after spending a cumulative 11 hours reviewing.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/SelectNormalCasualties.java
```diff
@@ -0,0 +1,169 @@
...
+          TargetUnits.of(new ArrayList<>(step.getFiringGroup().getTargetUnits()), List.of());
+    }
+    return targetUnits;
+  }
+
+  /**
+   * The maximum number of hits that this collection of units can sustain, taking into account units
+   * with two hits, and accounting for existing damage.
+   */
+  private static int getMaxHits(final Collection<Unit> units) {
```
DanVanAtta, line 112: This function move is a good candidate for a dedicated PR.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/SelectNormalCasualties.java
```diff
@@ -0,0 +1,169 @@
...
+  /**
+   * Limit the number of transports to hitsLeftForTransports per ally
+   *
+   * @param restrictedTransports All the transports available to be selected
+   * @param hitsLeftForTransports The number of transports per ally to return
+   * @return List of transports that is equal to a maximum of hitsLeftForTransports * ally count
+   */
+  private List<Unit> limitTransportsToSelect(
+      final Collection<Unit> restrictedTransports, final int hitsLeftForTransports) {
+    final Map<GamePlayer, Collection<Unit>> alliedHitPlayer = new HashMap<>();
```
DanVanAtta, line 130: Have you considered using Guava multimap here? That is essentially exactly the data structure of a `"map<A, Collection<B>>"`. It should simplify as you won't have to do the compute if absent, you could just add units.

EG:
```
restrictedTransports.forEach(unit -> alliedhitPlayer.add(unit.getOwner(), unit));
```

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/SelectNormalCasualties.java
```diff
@@ -0,0 +1,169 @@
...
+        alliedHitPlayer.entrySet()) {
+      transportsToSelect.addAll(
+          playerTransports.getValue().stream()
+              .limit(hitsLeftForTransports)
+              .collect(Collectors.toList()));
+    }
+    return transportsToSelect;
+  }
+
+  static class Select {
```
DanVanAtta, line 145: While this class is short, could it be renamed to be more descriptive?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/aa/AaFireAndCasualtyStep.java
```diff
@@ -26,25 +33,75 @@
...
+  private List<BattleStep> getSteps() {
+    return FireStepsBuilder.buildSteps(
+        FireStepsBuilder.Parameters.builder()
+            .battleState(battleState)
+            .battleActions(battleActions)
+            .firingGroupFilter(FiringGroupFilterAa.of(getSide()))
+            .side(getSide())
+            .returnFire(MustFightBattle.ReturnFire.ALL)
+            .roll(new Roll())
+            .selectCasualties(new Casualties())
```
DanVanAtta, line 59: Is select casualties an in/out parameter? You probably can use `@builder.default` on 'select casualties' to give it a default value, and avoid passing a new instance.

Similar question about the `new roll()`

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/aa/OffensiveAaFire.java
```diff
@@ -1,20 +1,16 @@
...
-import games.strategy.engine.data.GamePlayer;
-import games.strategy.engine.data.Unit;
-import games.strategy.engine.delegate.IDelegateBridge;
-import games.strategy.triplea.delegate.ExecutionStack;
 import games.strategy.triplea.delegate.battle.BattleActions;
 import games.strategy.triplea.delegate.battle.BattleState;
-import java.util.Collection;
 
 /** Offensive Aa units can fire and the player can select their casualties */
 public class OffensiveAaFire extends AaFireAndCasualtyStep {
```
DanVanAtta, line 9: This class got quite a bit shorter, nice work :+1:

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/DefensiveFirstStrike.java
```diff
@@ -99,15 +110,27 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
...
-        true,
-        battleState.getPlayer(DEFENSE),
-        Matches.unitIsFirstStrikeOnDefense(battleState.getGameData()),
-        battleState.filterUnits(ALIVE, DEFENSE),
-        battleState.filterUnits(CASUALTY, DEFENSE),
-        battleState.filterUnits(ALIVE, OFFENSE),
-        battleState.filterUnits(CASUALTY, OFFENSE));
+    final List<BattleStep> steps = getSteps();
+
+    // steps go in reverse order on the stack
```
DanVanAtta, line 115: This seems familiar ;)
Perhaps the `FireStepsBuilder` should know this and just return a stack itself?

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/aa/FiringGroupFilterAaTest.java
```diff
@@ -0,0 +1,351 @@
...
+            .apply(
+                givenBattleStateBuilder()
+                    .gameData(givenGameData().withWarRelationship(defender, attacker, true).build())
+                    .attacker(attacker)
+                    .defender(defender)
+                    .attackingUnits(List.of(fireUnit))
+                    .defendingUnits(List.of(targetUnit))
+                    .build());
+
+    assertThat(firingGroups, hasSize(1));
```
DanVanAtta, line 60: Why should the firing group have a size of one?

"Have pity on the reader"

I forget who wrote that, I think it was Fowler perhaps, but have pity on the reader of code and make the things that are obvious, *really* obvious. A reader does not have context and is thinking about several concerns. An assertion comment string here could go a long way to explain why this assertion should be a true statement. EG:
```
assertThat("combat AA can taget any unit, given one opposing unit, expect that unit to be in the firing group", firingGroups, hasSize(1))
```

Reading this test, that is my best guess why this assertion is valid. Beyond that, I'd have to start debugging and/or reverse engineering the code. If there were any changes that caused this test to break, that is what a developer would have to do, to fix this one assertion, of this one test. I emphasize this for the importance of making it very easy for maintainers to read tests and understand why the test is valid. A mixture of `@DisplayText` and assertion comments should be favored to help explain this. When there is one assertion, a `@DisplayText` is sufficient, when there are multiple assertions then generally assertion comments should start coming into play. (For the record this is a convention that should be followed, to make tests easy to understand for maintainers: 
- https://github.com/triplea-game/triplea/blob/master/docs/development/code-conventions/java-test-code-conventions.md#favor-adding-displayname-to-each-test-case-5525
- https://github.com/triplea-game/triplea/blob/master/docs/development/code-conventions/java-test-code-conventions.md#use-assertthat-string-parameter-to-describe-why-we-assert-a-condition-5527)

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/aa/FiringGroupFilterAaTest.java
```diff
@@ -0,0 +1,351 @@
...
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class FiringGroupFilterAaTest {
+
+  @Mock GamePlayer attacker;
+  @Mock GamePlayer defender;
+
+  @Test
+  void basicFiringGroup() {
```
DanVanAtta, line 44: May want to spell out what you mean by 'basic'. The meaning of 'basic' is context dependent, without context, I would not know a 'basic' firing group from one that is not 'basic'

DanVanAtta (commented):
@trevan I did some low level reviewing to do a first initial quick pass to give you an idea of things that could perhaps be worked on and to get those items out of the way. I'd need to fire this up in an IDE to really get a sense of the overall strategic refactor that is happening here (something that I should do anyways). Perhaps some overview notes can help outline that direction? 

60 minutes into this PR, I'd say it's about 15% reviewed. I'm not going to be able to do another review until next weekend. Perhaps it's time we had a discussion to give you write permissions so that you could do things like method moves with a post-merge-review label.

DanVanAtta:
@trevan we need to discuss strategies for getting your efforts merged in. Look for a PM in forums from me.

@Cernelius sorry to be rude here and close this PR, but it's cutting to the chase. Please use "reference in a new issue" for your discussion comments and lets continue those conversations.

PR #7828, opened 2020-10-03
Create an object to track state between fire rounds

trevan:
This is a piece of #7823.  In `FireAa` and `Fire`, they track the state between each of the "fire", "select", and "remove" steps by using member variables.  Moving to individual steps require a separate state object.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundState.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundState.java
@@ -1,0 +1,14 @@
+package games.strategy.triplea.delegate.battle.steps.fire;
+
+import games.strategy.triplea.delegate.DiceRoll;
+import games.strategy.triplea.delegate.data.CasualtyDetails;
+import lombok.Data;
+
+/** Tracks the state through a Fire round (dice roll, select casualties, remove casualties) */
+@Data
+public class FireRoundState {
+
+  private DiceRoll dice;
+
+  private CasualtyDetails casualties;
+}


game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundState.java
```diff
@@ -0,0 +1,14 @@
+package games.strategy.triplea.delegate.battle.steps.fire;
+
+import games.strategy.triplea.delegate.DiceRoll;
+import games.strategy.triplea.delegate.data.CasualtyDetails;
+import lombok.Data;
+
+/** Tracks the state through a Fire round (dice roll, select casualties, remove casualties) */
+@Data
```
DanVanAtta, line 8: I'm noticing this is a mutable object. 

Would you agree it's better to have mutable references to immutable objects than it is to have an final reference to a mutable object?

Can we instead have this be a `@Value`?

trevan: I agree that it is better to have mutable references to immutable objects.  And that is basically what this is.  It is a collection of mutable references to two mostly immutable objects.  DiceRoll appears to be completely immutable and could probably be marked as `@Value` and CasualtyDetails looks like it could be made completely immutable and my changes only use it in an immutable manner.

And this can't be a `@Value` because the three steps need to access and update these references.  The first one will update the dice roll.  The second will use the dice roll to update the casualty details.  And the third uses both the dice roll and the casualty details.

DanVanAtta: Instead of accessing and updating this object, what if a new one were created and that new value then passed forward? Is this a good use-case for the lombok `toBuilder()` where you can create a mutated copy of an object?

trevan: I'd love to do that.  But IExecutables in the stack don't share anything.  You can't pass things between them.

DanVanAtta: :cry:

PR #7829, opened 2020-10-03
Create FiringGroup classes to rework how battles determine firing and targeted units

trevan:
This is a piece of #7823.  Currently, the firing and targeting unit detection is done in several disjointed places: `MustFightBattle#fireNavalBombardment`, `MustFightBattle#findTargetGroupsAndFire`, and `FireAa`.

This extracts the logic into functional classes (`FiringGroupFilter`) that take a `BattleState` and turn the active/alive units into a list of `FiringGroup`s.  The `FiringGroup` keeps track of which units are firing on their specific targets.

There is a `FiringGroupFilter` for bombardment, aa, and other units.  These functional classes contain the same logic you can find in `MustFightBattle#fireNavalBombardment`, `MustFightBattle#findTargetGroupsAndFire`, and `FireAa`.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleStepStrings.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleStepStrings.java
@@ -9,11 +9,15 @@ public interface BattleStepStrings {
   String CASUALTIES_SUFFIX = " casualties";
   String LAND_PARATROOPS = "Land Paratroopers";
   String NAVAL_BOMBARDMENT = "Naval bombardment";
+  String NAVAL_BOMBARD = "naval bombardment";
   String SELECT_NAVAL_BOMBARDMENT_CASUALTIES = "Select naval bombardment casualties";
   String REMOVE_UNESCORTED_TRANSPORTS = "Remove unescorted transports";
   String SUBMERGE_SUBS_VS_AIR_ONLY = "Submerge subs against only air units";
   String AIR_ATTACK_NON_SUBS = "Air attack non subs";
   String AIR_DEFEND_NON_SUBS = "Air defend non subs";
+  String AIR_FIRE_NON_SUBS = "air vs non subs";
+  String UNITS = "units";
+  String FIRST_STRIKE_UNITS = "first strike units";
   String REMOVE_SNEAK_ATTACK_CASUALTIES = "Remove sneak attack casualties";
   String FIRE = " fire";
   String FIRST_STRIKE_UNITS_FIRE = " first strike units fire";
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
@@ -1,0 +1,143 @@
+package games.strategy.triplea.delegate.battle.steps.fire;
+
+import games.strategy.engine.data.Unit;
+import games.strategy.engine.data.UnitType;
+import games.strategy.triplea.delegate.Matches;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import lombok.Value;
+import org.triplea.java.collections.CollectionUtils;
+
+/**
+ * Group of units that are firing on targets
+ *
+ * <p>If the group is suicideOnHit, then all of the units should have the same unit type
+ */
+@Value
+public class FiringGroup {
+  String displayName;
+  String groupName;
+  Collection<Unit> firingUnits;
+  Collection<Unit> targetUnits;
+  boolean suicideOnHit;
+
+  public Collection<Unit> getTargetUnits() {
+    return Collections.unmodifiableCollection(targetUnits);
+  }
+
+  public Collection<Unit> getFiringUnits() {
+    return Collections.unmodifiableCollection(firingUnits);
+  }
+
+  /**
+   * Keeps alive units around
+   *
+   * <p>Units unfortunately don't track their own status so the list of targets needs to be updated
+   * as the battle progresses.
+   *
+   * @param aliveUnits Units that are still alive
+   */
+  public void retainAliveTargets(final Collection<Unit> aliveUnits) {
+    targetUnits.retainAll(aliveUnits);
+  }
+
+  /**
+   * Splits up the firingUnits by suicideOnHit status and groups them by unit type
+   *
+   * @param name Name of the firing units
+   * @param firingUnits Collection of units that are firing
+   * @param targetUnits Collection of units that are being hit
+   * @return List of FiringGroup
+   */
+  public static List<FiringGroup> groupBySuicideOnHit(
+      final String name, final Collection<Unit> firingUnits, final Collection<Unit> targetUnits) {
+    final List<FiringGroup> groups = new ArrayList<>();
+    final List<Collection<Unit>> separatedBySuicide = separateSuicideOnHit(firingUnits);
+
+    // ensure each firing group has a unique name by adding prefixes
+    // if the firing groups have different types of suicide units
+    if (separatedBySuicide.size() == 1) {
+      groups.add(
+          new FiringGroup(
+              name,
+              name,
+              separatedBySuicide.get(0),
+              targetUnits,
+              Matches.unitIsSuicideOnHit().test(separatedBySuicide.get(0).iterator().next())));
+
+    } else if (separatedBySuicide.size() == 2
+        && Matches.unitIsSuicideOnHit().test(separatedBySuicide.get(0).iterator().next())
+            != Matches.unitIsSuicideOnHit().test(separatedBySuicide.get(1).iterator().next())) {
+      groups.addAll(
+          generateFiringGroupsWithOneSuicideAndOneNonSuicide(
+              name, targetUnits, separatedBySuicide));
+
+    } else {
+      groups.addAll(generateFiringGroups(name, targetUnits, separatedBySuicide));
+    }
+    return groups;
+  }
+
+  /**
+   * Separate the suicide on hit units from the others and group them by their type. The suicide on
+   * hit units need to fire separately so that they can be removed if they hit.
+   */
+  private static List<Collection<Unit>> separateSuicideOnHit(final Collection<Unit> units) {
+
+    final Map<UnitType, Collection<Unit>> map = new HashMap<>();
+    for (final Unit unit : CollectionUtils.getMatches(units, Matches.unitIsSuicideOnHit())) {
+      final UnitType type = unit.getType();
+      if (map.containsKey(type)) {
+        map.get(type).add(unit);
+      } else {
+        final Collection<Unit> unitList = new ArrayList<>();
+        unitList.add(unit);
+        map.put(type, unitList);
+      }
+    }
+
+    final List<Collection<Unit>> result = new ArrayList<>(map.values());
+    final Collection<Unit> remainingUnits =
+        CollectionUtils.getMatches(units, Matches.unitIsSuicideOnHit().negate());
+    if (!remainingUnits.isEmpty()) {
+      result.add(remainingUnits);
+    }
+    return result;
+  }
+
+  /** Handle the case where there are only two groups and one is suicide and the other is not */
+  private static List<FiringGroup> generateFiringGroupsWithOneSuicideAndOneNonSuicide(
+      final String name,
+      final Collection<Unit> targetUnits,
+      final List<Collection<Unit>> separatedBySuicide) {
+    final List<FiringGroup> groups = new ArrayList<>();
+    for (final Collection<Unit> newFiringUnits : separatedBySuicide) {
+      final boolean isSuicideOnHit =
+          Matches.unitIsSuicideOnHit().test(newFiringUnits.iterator().next());
+      final String nameWithSuffix = name + (isSuicideOnHit ? " suicide" : "");
+      groups.add(
+          new FiringGroup(nameWithSuffix, name, newFiringUnits, targetUnits, isSuicideOnHit));
+    }
+    return groups;
+  }
+
+  private static List<FiringGroup> generateFiringGroups(
+      final String name,
+      final Collection<Unit> targetUnits,
+      final List<Collection<Unit>> separatedBySuicide) {
+    final List<FiringGroup> groups = new ArrayList<>();
+    for (final Collection<Unit> newFiringUnits : separatedBySuicide) {
+      final Unit firstUnit = newFiringUnits.iterator().next();
+      final boolean isSuicideOnHit = Matches.unitIsSuicideOnHit().test(firstUnit);
+      final String nameWithSuffix =
+          name + (isSuicideOnHit ? " suicide " + firstUnit.getType().getName() : "");
+      groups.add(
+          new FiringGroup(nameWithSuffix, name, newFiringUnits, targetUnits, isSuicideOnHit));
+    }
+    return groups;
+  }
+}
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
@@ -5,6 +5,7 @@ import com.google.common.collect.Multimap;
 import games.strategy.engine.data.Unit;
 import games.strategy.engine.data.UnitType;
 import games.strategy.triplea.delegate.Matches;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
@@ -15,12 +16,30 @@ import org.triplea.java.collections.CollectionUtils;
 /**
  * Group of units that are firing on targets
  *
- * <p>If the group is suicideOnHit, then all of the units should have the same unit type
+ * <p>If the group is suicideOnHit, then all of the units must have the same unit type. This is
+ * because suicideOnHit units need to die when they hit a unit. If a firing group has multiple unit
+ * types, then it is harder to determine which unit type hit and so must commit suicide. By forcing
+ * the firing group to have all of the same unit type, it the units to commit suicide can be easily
+ * figured out since all of the units are the same.
  */
 @Value
 public class FiringGroup {
 
-  private static final UnitType NON_SUICIDE_MULTIMAP_KEY = new UnitType("nonsuicide", null);
+  @Value(staticConstructor = "of")
+  private static class SuicideAndNonSuicide {
+    Multimap<UnitType, Unit> suicideGroups;
+    Collection<Unit> nonSuicideGroup;
+
+    Collection<Collection<Unit>> values() {
+      final Collection<Collection<Unit>> values = new ArrayList<>(suicideGroups.asMap().values());
+      values.add(nonSuicideGroup);
+      return values;
+    }
+
+    int groupCount() {
+      return suicideGroups.keySet().size() + (nonSuicideGroup.isEmpty() ? 0 : 1);
+    }
+  }
 
   String displayName;
   String groupName;
@@ -49,30 +68,35 @@ public class FiringGroup {
   }
 
   /**
-   * Keeps alive units around
+   * Retain targeted units for future firing
    *
    * <p>Units unfortunately don't track their own status so the list of targets needs to be updated
    * as the battle progresses.
    *
-   * @param aliveUnits Units that are still alive
+   * <p>By the time this firing group gets a chance to fire, a previous firing group might have made
+   * the targets un-targetable (such as killing them).
+   *
+   * <p>The caller will be tracking the unit status and so can call this with the list of targetable
+   * units to ensure that this firing group will only target units that are available to be fired
+   * at.
+   *
+   * @param aliveUnits Units that need to be retained as targets
    */
   public void retainAliveTargets(final Collection<Unit> aliveUnits) {
     targetUnits.retainAll(aliveUnits);
   }
 
   /**
-   * Splits up the firingUnits by suicideOnHit status and groups them by unit type
+   * Groups the suicideOnHit firingUnits by unit type and groups the non-suicideOnHit firingUnits
+   * all together.
    *
-   * @param name Name of the firing units
-   * @param firingUnits Collection of units that are firing
-   * @param targetUnits Collection of units that are being hit
-   * @return List of FiringGroup
+   * <p>See the class documentation on why suicideOnHit matters.
    */
   public static List<FiringGroup> groupBySuicideOnHit(
       final String name, final Collection<Unit> firingUnits, final Collection<Unit> targetUnits) {
-    final Multimap<UnitType, Unit> separatedBySuicide = separateSuicideOnHit(firingUnits);
+    final SuicideAndNonSuicide separatedBySuicide = separateSuicideOnHit(firingUnits);
 
-    return separatedBySuicide.asMap().values().stream()
+    return separatedBySuicide.values().stream()
         .map(
             units ->
                 new FiringGroup(
@@ -80,11 +104,7 @@ public class FiringGroup {
         .collect(Collectors.toList());
   }
 
-  /**
-   * Separate the suicide on hit units from the others and group them by their type. The suicide on
-   * hit units need to fire separately so that they can be removed if they hit.
-   */
-  private static Multimap<UnitType, Unit> separateSuicideOnHit(final Collection<Unit> units) {
+  private static SuicideAndNonSuicide separateSuicideOnHit(final Collection<Unit> units) {
 
     final Multimap<UnitType, Unit> map = ArrayListMultimap.create();
     for (final Unit unit : CollectionUtils.getMatches(units, Matches.unitIsSuicideOnHit())) {
@@ -94,23 +114,20 @@ public class FiringGroup {
 
     final Collection<Unit> remainingUnits =
         CollectionUtils.getMatches(units, Matches.unitIsSuicideOnHit().negate());
-    if (!remainingUnits.isEmpty()) {
-      map.putAll(NON_SUICIDE_MULTIMAP_KEY, remainingUnits);
-    }
-    return map;
+    return SuicideAndNonSuicide.of(map, remainingUnits);
   }
 
   private static String generateName(
       final String originalName,
       final Collection<Unit> firingUnits,
-      final Multimap<UnitType, Unit> separatedBySuicide) {
+      final SuicideAndNonSuicide separatedBySuicide) {
 
-    if (separatedBySuicide.keySet().size() == 1) {
+    if (separatedBySuicide.groupCount() == 1) {
       // there is only one firing group so no need to give unique suffices
       return originalName;
 
-    } else if (separatedBySuicide.keySet().size() == 2
-        && separatedBySuicide.containsKey(NON_SUICIDE_MULTIMAP_KEY)
+    } else if (separatedBySuicide.groupCount() == 2
+        && !separatedBySuicide.nonSuicideGroup.isEmpty()
         && firingUnits.stream().allMatch(Matches.unitIsSuicideOnHit())) {
       // there are two firing groups and one of them is non suicide. So the suicide group
       // doesn't need to have its type name added to it.
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
@@ -7,7 +7,6 @@ import games.strategy.engine.data.UnitType;
 import games.strategy.triplea.delegate.Matches;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 import lombok.Value;
@@ -60,11 +59,11 @@ public class FiringGroup {
   }
 
   public Collection<Unit> getTargetUnits() {
-    return Collections.unmodifiableCollection(targetUnits);
+    return targetUnits;
   }
 
   public Collection<Unit> getFiringUnits() {
-    return Collections.unmodifiableCollection(firingUnits);
+    return firingUnits;
   }
 
   /**
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
@@ -37,7 +37,7 @@ public class FiringGroup {
     this.groupName = groupName;
     this.firingUnits = firingUnits;
     this.targetUnits = targetUnits;
-    this.suicideOnHit = this.firingUnits.stream().anyMatch(Matches.unitIsSuicideOnHit());
+    this.suicideOnHit = this.firingUnits.stream().allMatch(Matches.unitIsSuicideOnHit());
   }
 
   public Collection<Unit> getTargetUnits() {
@@ -106,19 +106,20 @@ public class FiringGroup {
       final Multimap<UnitType, Unit> separatedBySuicide) {
 
     if (separatedBySuicide.keySet().size() == 1) {
+      // there is only one firing group so no need to give unique suffices
       return originalName;
 
     } else if (separatedBySuicide.keySet().size() == 2
-        && separatedBySuicide.containsKey(NON_SUICIDE_MULTIMAP_KEY)) {
-      // special case where one firing group is suicideOnHit and the other is not
-      return firingUnits.stream().allMatch(Matches.unitIsSuicideOnHit())
-          ? originalName + " suicide"
-          : originalName;
-
-    } else {
-      return firingUnits.stream().allMatch(Matches.unitIsSuicideOnHit())
-          ? originalName + " suicide " + firingUnits.iterator().next().getType().getName()
-          : originalName;
+        && separatedBySuicide.containsKey(NON_SUICIDE_MULTIMAP_KEY)
+        && firingUnits.stream().allMatch(Matches.unitIsSuicideOnHit())) {
+      // there are two firing groups and one of them is non suicide. So the suicide group
+      // doesn't need to have its type name added to it.
+      return originalName + " suicide";
     }
+
+    // add a suffix to suicide groups that includes their type name to differentiate them
+    return firingUnits.stream().allMatch(Matches.unitIsSuicideOnHit())
+        ? originalName + " suicide " + firingUnits.iterator().next().getType().getName()
+        : originalName;
   }
 }
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
@@ -24,6 +24,12 @@ import org.triplea.java.collections.CollectionUtils;
 @Value
 public class FiringGroup {
 
+  String displayName;
+  String groupName;
+  Collection<Unit> firingUnits;
+  Collection<Unit> targetUnits;
+  boolean suicideOnHit;
+
   @Value(staticConstructor = "of")
   private static class SuicideAndNonSuicide {
     Multimap<UnitType, Unit> suicideGroups;
@@ -39,12 +45,6 @@ public class FiringGroup {
       return suicideGroups.keySet().size() + (nonSuicideGroup.isEmpty() ? 0 : 1);
     }
   }
-
-  String displayName;
-  String groupName;
-  Collection<Unit> firingUnits;
-  Collection<Unit> targetUnits;
-  boolean suicideOnHit;
 
   private FiringGroup(
       final String displayName,
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
@@ -1,14 +1,14 @@
 package games.strategy.triplea.delegate.battle.steps.fire;
 
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
 import games.strategy.engine.data.Unit;
 import games.strategy.engine.data.UnitType;
 import games.strategy.triplea.delegate.Matches;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
+import java.util.stream.Collectors;
 import lombok.Value;
 import org.triplea.java.collections.CollectionUtils;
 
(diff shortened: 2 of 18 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import lombok.Value;
+import org.triplea.java.collections.CollectionUtils;
+
+/**
+ * Group of units that are firing on targets
+ *
+ * <p>If the group is suicideOnHit, then all of the units should have the same unit type
```
DanVanAtta, line 18: This seems to imply a different object semantic, perhaps even a different object. WDYT of something like `SuicideOnHitFiringGroup`? Would it simplify the eventual usages?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
```diff
@@ -0,0 +1,143 @@
...
+   * @param firingUnits Collection of units that are firing
+   * @param targetUnits Collection of units that are being hit
+   * @return List of FiringGroup
+   */
+  public static List<FiringGroup> groupBySuicideOnHit(
+      final String name, final Collection<Unit> firingUnits, final Collection<Unit> targetUnits) {
+    final List<FiringGroup> groups = new ArrayList<>();
+    final List<Collection<Unit>> separatedBySuicide = separateSuicideOnHit(firingUnits);
+
+    // ensure each firing group has a unique name by adding prefixes
```
DanVanAtta, line 61: Where is the prefix added? It's an interesting comment, though scanning down it's not immediately obvious where/what this is refererring to exactly.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+    return result;
+  }
+
+  /** Handle the case where there are only two groups and one is suicide and the other is not */
+  private static List<FiringGroup> generateFiringGroupsWithOneSuicideAndOneNonSuicide(
+      final String name,
+      final Collection<Unit> targetUnits,
+      final List<Collection<Unit>> separatedBySuicide) {
+    final List<FiringGroup> groups = new ArrayList<>();
```
DanVanAtta, line 117: WDYT of having a constructor do the work for the `isSuidideOnHit` check? 

In such a case this method would be reduced to:
```
    return separatedBySuicide.stream()
        .map(newFiringUnits -> new FiringGroup(name, newFiringUnits, targetUnits))
```

That makes it short enough it perhaps is worth inlining.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
```diff
@@ -0,0 +1,143 @@
...
+      result.add(remainingUnits);
+    }
+    return result;
+  }
+
+  /** Handle the case where there are only two groups and one is suicide and the other is not */
+  private static List<FiringGroup> generateFiringGroupsWithOneSuicideAndOneNonSuicide(
+      final String name,
+      final Collection<Unit> targetUnits,
+      final List<Collection<Unit>> separatedBySuicide) {
```
DanVanAtta, line 116: This param is interesting `List<Collection<Unit>>`. I wonder a bit what the ordering of the list is. If not ordered, then should it be just a `Collection`? That makes me wonder further if this would be better as a `Map<?, Collection<Unit>>`

For map key, if an enum would perhaps do well, something like `FiringGroupType { SUICIDE_ON_HIT, STANDARD }`

WDYT?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
```diff
@@ -0,0 +1,143 @@
...
+    final List<FiringGroup> groups = new ArrayList<>();
+    for (final Collection<Unit> newFiringUnits : separatedBySuicide) {
+      final boolean isSuicideOnHit =
+          Matches.unitIsSuicideOnHit().test(newFiringUnits.iterator().next());
```
DanVanAtta, line 120: Will we always be guaranteed to have a first unit in: `newFiringUnits.iterator().next()`?
I do not know, maybe there is some code earlier that guarantees this, however it seems like a pretty strong precondition that is not necessarily documented (hopefully enforced). If we convert the incoming list to a map, that would resolve this problem as you would not need to check any of the unit elements to determine if they are 'suicide on hit', you could retrieve them directly from the input map.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroupFilter.java
```diff
@@ -0,0 +1,8 @@
+package games.strategy.triplea.delegate.battle.steps.fire;
+
+import games.strategy.triplea.delegate.battle.BattleState;
+import java.util.List;
+import java.util.function.Function;
+
+/** Converts the BattleState into a list of FiringGroups */
+public interface FiringGroupFilter extends Function<BattleState, List<FiringGroup>> {}
```
DanVanAtta, line 8: Couple questions:
- Could you explain why an interface "alias" is helpful?
- Is the output list ordered?

A filter will typically exclude or include items. My presumption is that this is more of a sieve, that we are not losing any elements but instead are grouping them. Is that correct? If so, perhaps this would be better named as a `Splitter`, or as a `Sorter`.  OTOH - If we do not need an interface alias, then the naming problem would drop away.

DanVanAtta: For consideration, [Effective Java 3](https://github.com/muhdkhokhar/test/blob/master/Joshua%20Bloch%20-%20Effective%20Java%20(3rd)%20-%202018.pdf), page 200:


Screenshot from 2020-10-03 12-34-53

Screenshot from 2020-10-03 12-36-30

Screenshot from 2020-10-03 12-34-35

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
DanVanAtta: If there were a map of firing group by type, then FiringGroup would not need the `suicideOnHit` boolean.

DanVanAtta (commented):
There is quite a lot in this PR, perhaps about 30% reviewed

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
trevan: The prefixes are added inside of the callees `generateFiringGroupsWithOneSuicideAndOneNonSuicide` and `generateFiringGroups`.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
trevan: Collection is fine.  The order doesn't matter.

A map won't work because there can be more than one suicideOnHit collection.  This specific method adds a simplified prefix for the special case where there is only one suicideOnHit collection and one non suicideOnHit collection (I tried to express that in the method name).  The logic to figure out which one is the suicideOnHit and which one is not the suicideOnHit is inside of this method.  If I move that logic out of this method, then this method becomes basically nothing and would be inlined.  But doing that will increase the complexity of the caller and codeclimate will complain.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
trevan: No, it wouldn't simplify any eventual usage.  This flag is only used to determine whether to kill the units after they fire.  We have to have a separate group so that we can easily detect which units hit and so which ones can be removed.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
trevan: You'll notice that the isSuicideOnHit is used to determine the suffix.  And just because it is a suicideOnHit doesn't mean it will have the same suffix.  If you look at the next method `generateFiringGroups`, it also checks the suicideOnHit but it uses a different suffix.

Here's some examples of firing unit groups and what their suffices would be:

One (doesn't matter if it is suicide or not) firing group -> No suffix
One suicide firing group and one non suicide firing group -> The suicide group has " suicide" suffix and the non suicide firing group has no suffix
Two suicide firing groups -> They will both have " suicide <unitType.name>" suffix.
Two suicide firing groups and one non suicide firing group -> The two suicide groups will have the suffix " suicide <unitType.name>" and the non suicide firing group will not have a suffix.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
DanVanAtta: Cool, though that seems to be a detail of those methods. At this level, it's not obvious that is happening, therefore this comment seems out of place. One rule of thumb, any method should be understandable by simply reading it and not having to also read the methods that it invokes. One can see if that rule is violated then there is a scaling problem to understand a given method if you have to read all of the called methods, and all of the methods that are in turn called to the n'th degree.

trevan: It was there to explain why the if/else if/else.  The only reason for them is to generate suffices.  If I didn't care about having unique suffices, I would just do a simple
```
return separatedBySuicide.stream()
        .map(newFiringUnits -> new FiringGroup(name, newFiringUnits, targetUnits))
```

But I've been thinking about your comments here and I think if I instead change the above to:
```
return separatedBySuicide.stream()
        .map(newFiringUnits -> new FiringGroup(generateName(name, newFiringUnits, separatedBySuicide), newFiringUnits, targetUnits))
```

Then the logic for the unique names is more obviously contained to just dealing with the string instead of generating firing groups.

trevan:
I've reworked the FiringGroup.  I now use Guava's Multimap and moved the name generation to a more obvious name generation method.  I think that should resolve your feedback.

I also renamed all of the Filter classes to "Splitter" and removed the empty base interface.  And I did a few minor tweaks to `FiringGroupSplitterGeneral`: fixed some variable name and definitions, made the `canFire` variable a final variable.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroupSplitterBombard.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Create naval bombardment firing groups
+ *
+ * <p>The firing groups are separated by isSuicideOnHit
+ */
+@Value(staticConstructor = "of")
+public class FiringGroupSplitterBombard implements Function<BattleState, List<FiringGroup>> {
+
+  @Override
+  public List<FiringGroup> apply(final BattleState battleState) {
```
trevan, line 25: The original logic that this is based off can be found at `MustFightBattle#fireNavalBombardment` (https://github.com/triplea-game/triplea/blob/e160d341f6c60ecf5297aad9ab26ac5a524dc2dd/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java#L1462) and `MustFightBattle#fire` (https://github.com/triplea-game/triplea/blob/e160d341f6c60ecf5297aad9ab26ac5a524dc2dd/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java#L1497).

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/aa/FiringGroupSplitterAa.java
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * <p>The firing groups are separated by typeAa and isSuicideOnHit
+ */
+@Value(staticConstructor = "of")
+public class FiringGroupSplitterAa implements Function<BattleState, List<FiringGroup>> {
+
+  BattleState.Side side;
+
+  @Override
+  public List<FiringGroup> apply(final BattleState battleState) {
```
trevan, line 34: The original logic that this is based off of is in `MustFightBattle#updateOffensiveAaUnits` (https://github.com/triplea-game/triplea/blob/e160d341f6c60ecf5297aad9ab26ac5a524dc2dd/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java#L928), `MustFightBattle#updateDefensiveAaUnits` (https://github.com/triplea-game/triplea/blob/e160d341f6c60ecf5297aad9ab26ac5a524dc2dd/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java#L949), `FireAa#FireAa` (https://github.com/triplea-game/triplea/blob/e160d341f6c60ecf5297aad9ab26ac5a524dc2dd/game-core/src/main/java/games/strategy/triplea/delegate/battle/FireAa.java#L66), and `FireAa#execute` (https://github.com/triplea-game/triplea/blob/e160d341f6c60ecf5297aad9ab26ac5a524dc2dd/game-core/src/main/java/games/strategy/triplea/delegate/battle/FireAa.java#L100).

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/general/FiringGroupSplitterGeneral.java
```diff
@@ -0,0 +1,135 @@
...
+
+  BattleState.Side side;
+
+  Predicate<Unit> firingUnitPredicate;
+
+  /** Name displayed in the Battle UI */
+  String groupName;
+
+  @Override
+  public List<FiringGroup> apply(final BattleState battleState) {
```
trevan, line 41: The original logic that this is based off of is in `MustFightBattle#findTargetGroupsAndFire` (https://github.com/triplea-game/triplea/blob/e160d341f6c60ecf5297aad9ab26ac5a524dc2dd/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java#L1638) and `MustFightBattle#fire` (https://github.com/triplea-game/triplea/blob/e160d341f6c60ecf5297aad9ab26ac5a524dc2dd/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java#L1497)

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/general/FiringGroupSplitterGeneral.java
```diff
@@ -0,0 +1,135 @@
...
+    if (targetGroups.size() == 1) {
+      firingGroups.addAll(buildFiringGroups(groupName, canFire, enemyUnits, targetGroups.get(0)));
+    } else {
+      // General firing groups don't have individual names so find commonly used groups and
+      // give them unique names
+      final List<TargetGroup> airVsSubGroups =
+          targetGroups.stream()
+              .filter(this.filterAirVsSubTargetGroups(enemyUnits))
+              .collect(Collectors.toList());
+      generateNamedGroups(AIR_FIRE_NON_SUBS, firingGroups, airVsSubGroups, canFire, enemyUnits);
```
trevan, line 74: This special named group will replace the "Air defend non subs" or "Air attack non subs" battle steps.  I mentioned this and posted images of what it will look like in the original PR #7823

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
DanVanAtta: >  But doing that will increase the complexity of the caller and codeclimate will complain.

codeclimate is not perfect. There are times when the actual victory is from a fundamental change rather than a surface level. In other words, sometimes right right fix to codeclimate issues are not simple, surface level fixes. If a codeclimate warning is in the way of what is actually simpler, then codeclimate should be ignored in that instance.

DanVanAtta: The encoding of the 'prefix' does strike me as complicated.

What are the different types of `suicideOnHit` collection? Perhaps we are talking instead a more complicated data structure, something like: `Map<UnitType, Map<SuicideOnHitFlag, Collection<Unit>>`. Generally, such a data structure is a candidate to be converted to a first class object. One would init it by passing a collection of units, perhaps calling a factory method, and the methods on the object would be queries to return units of a given type.  

I think there could be some advantages to that:
- readily tested
- extracts a good bit of detail
- readily mocked, could be useful for testing other logic that relies on units with certain characteristics

trevan: I've made changes to how this works.  Take a look at the new code.

DanVanAtta:
I was just opening this to review!
@trevan any plans to re-open? Breaking it up further?

trevan:
Oops, I was cleaning up branches and I deleted this one by mistake.

DanVanAtta:
All good, happily very reversible : )

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
```diff
@@ -0,0 +1,125 @@
...
+  private FiringGroup(
+      final String displayName,
+      final String groupName,
+      final Collection<Unit> firingUnits,
+      final Collection<Unit> targetUnits) {
+    this.displayName = displayName;
+    this.groupName = groupName;
+    this.firingUnits = firingUnits;
+    this.targetUnits = targetUnits;
+    this.suicideOnHit = this.firingUnits.stream().allMatch(Matches.unitIsSuicideOnHit());
```
DanVanAtta, line 40: It's not obvious  offhand why every unit should be 'unitIsSuicideOnHit' for the whole group to be marked as such. I'm just wondering why it's not any unit, or why it's not tracked at a per-unit level even. Perhaps a comment on the variable could explain it's role in the class.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  public Collection<Unit> getTargetUnits() {
+    return Collections.unmodifiableCollection(targetUnits);
```
DanVanAtta, line 44: I wonder if callers could just avoid mutating the return value and/or copy as needed. If we can skip the defensive copy, we can have a performance improvement and simpler code in this class as a simpler getter would do.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  public Collection<Unit> getFiringUnits() {
+    return Collections.unmodifiableCollection(firingUnits);
+  }
+
+  /**
+   * Keeps alive units around
```
DanVanAtta, line 52: The concept of being kept around is not very well defined. Would you mind making this comment more specific, more defined?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
```diff
@@ -0,0 +1,125 @@
...
+   *
+   * <p>Units unfortunately don't track their own status so the list of targets needs to be updated
```
DanVanAtta, line 54: I think this comment is more of an implementation note. Ideally this class is designed so that it is generic and has a role, a function, and is not aware of limitations around it. IF this class is responsible for tracking units that are still alive, perhaps that can just be cleanly documented in the class javadoc? AS-is, it's a bit odd to see this note, I'm not sure if it really answers questions, seems like if you have context then it makes more sense but otherwise is a bit confusing.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
```diff
@@ -0,0 +1,125 @@
...
+    targetUnits.retainAll(aliveUnits);
+  }
+
+  /**
+   * Splits up the firingUnits by suicideOnHit status and groups them by unit type
+   *
+   * @param name Name of the firing units
+   * @param firingUnits Collection of units that are firing
+   * @param targetUnits Collection of units that are being hit
+   * @return List of FiringGroup
```
DanVanAtta, line 69: nit, not sure this @return javadoc adds much, restates the code. We are not publishing library code, anyone reading the javadoc will likely be reading the javadoc in the class file at the same time. I think this `@return` can perhaps be removed or otherwise should describe any invariants that would be true of the return value. For example, it is a list, what is the ordering of the list? Are there constraints on the size of the list. Also, will the firigin group be a collection of the firing units, or the target units, or both?

DanVanAtta: I am a bit confused though, `suicideOnHit` already implies a specific unit type, right? If we are grouping by unit type, then don't we automatically also group by `suicideOnHit` too?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Group of units that are firing on targets
+ *
+ * <p>If the group is suicideOnHit, then all of the units should have the same unit type
+ */
+@Value
+public class FiringGroup {
+
+  private static final UnitType NON_SUICIDE_MULTIMAP_KEY = new UnitType("nonsuicide", null);
```
DanVanAtta, line 23: Creating a fake unit type seems to contradict the unit type datatype (SRP violation). The method that uses this I think probably would do better to return a first class data structure, something like:

```

class SuicideOnHitSortedFiringGroups {
     Map<UnitType, FiringGroup> suicideOnHitFiringGroups;
     Collection<Unit> nonSuicideOnHitFiringGroups;;
}
```

WDYT?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroupSplitterBombard.java
```diff
@@ -0,0 +1,37 @@
...
+    final Collection<Unit> canFire = battleState.getBombardingUnits();
```
DanVanAtta, line 26: nit, would 'bombarindUnits' be a more descriptive name here? Perhaps even consider inlining it to avoid the intermediate variable?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/aa/FiringGroupSplitterAa.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
+import lombok.Value;
+import org.triplea.java.collections.CollectionUtils;
+
+/**
+ * Creates AA and Targeted Hit firing groups
```
DanVanAtta, line 24: For the splitter class, it would be nice to see some more detail in the javadoc to explain what characteristics would be true of the output `List<FiringGroup>`. I'm not really sure what "Targeted Hit firing groups" are.

Perhaps we can/should add 'targeted hit' to the glossary?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/aa/FiringGroupSplitterAa.java
```diff
@@ -0,0 +1,80 @@
...
+                battleState.getGameData()));
+
+    final List<String> aaTypes = UnitAttachment.getAllOfTypeAas(aaUnits);
+
+    final Collection<Unit> validTargetUnits =
+        CollectionUtils.getMatches(
+            battleState.filterUnits(ALIVE, side.getOpposite()),
+            Matches.unitIsNotInfrastructure().and(Matches.unitIsBeingTransported().negate()));
+
+    final List<FiringGroup> firingGroups = new ArrayList<>();
```
DanVanAtta, line 60: The logic in here is really heavy. Do you have enough context where you could readily add some comments to guide a reader in understanding what this code is doing?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/general/FiringGroupSplitterGeneral.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import lombok.Value;
+import org.triplea.java.PredicateBuilder;
+import org.triplea.java.collections.CollectionUtils;
+
+/**
+ * Create firing groups with units that match {@link #firingUnitPredicate}
+ *
+ * <p>The firing groups are separated by canNotTarget, canNotBeTargetedBy, and isSuicideOnHit
```
DanVanAtta, line 28: It's a bit odd to have a generic predicate and for this class to know how the firing groups are going to be split.

Is the predicate acting as a filter to remove units from the firing groups, or is the predicate more acting at the firing group level?

The firing predicate is a bit concerning as I wonder if we are too abstract here, too complex. Could the caller apply the predicate rather than passing it in? I'm wondering if a caller would have to know a great deal about this implementation to be able to reasonably pass a good predicate to this class.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/general/FiringGroupSplitterGeneralTest.java
```diff
@@ -0,0 +1,357 @@
...
+                givenBattleStateBuilder()
+                    .gameData(givenGameData().withAlliedAirIndependent(true).build())
+                    .attacker(attacker)
+                    .defender(defender)
+                    .attackingUnits(List.of(fireUnit))
+                    .defendingUnits(List.of(targetUnit))
+                    .build());
+
+    assertThat(firingGroups, hasSize(1));
+    assertThat(firingGroups.get(0).getDisplayName(), is(UNITS));
```
DanVanAtta, line 59: It seems we are testing multiple behaviors at the same time here. Is the expected display name of units simply because that is the splitter parameter? Have you considered adding a set of tests to only test display name behavior?

DanVanAtta (commented):
Probably about 70% of the way through. Some heavy lifting in here it seems.. Good to see the code being worked and tested :+1:

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/general/FiringGroupSplitterGeneralTest.java
trevan: No, the expected display name will change.  The last two tests in this file test for changed display names: `twoFiringGroupsWithCanNotTarget` and `twoGroupsWhenAirAndSeaVsSub`.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
trevan: Something to hope for :)

trevan:
I've pushed a change with more documentation.  Hopefully that helps in explaining things.

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleStepStrings.java
```diff
@@ -9,11 +9,15 @@
   String CASUALTIES_SUFFIX = " casualties";
   String LAND_PARATROOPS = "Land Paratroopers";
   String NAVAL_BOMBARDMENT = "Naval bombardment";
+  String NAVAL_BOMBARD = "naval bombardment";
```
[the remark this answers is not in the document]
DanVanAtta: Hmm, okay. Though that seems to be overly context dependent. I wonder if it would be better to do these with built-in patterns and use them with string format. Alternatively maybe as method invocations even rather than contexts. The problem I see is that the usage is very specific, while this class is independent, it turns out to have a level of coupling to where it is used.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
DanVanAtta: The copies can add up to be significant. Can we do better than hoping? The java8 streaming API is particularly helpful for operating on returned collections as the collection is not mutated.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/aa/FiringGroupSplitterAa.java
DanVanAtta: The terms are broadly not well defined. TripleA was created before the concept of Domain-Driven-Design was more mainstream, as such instead of defining and using domain terminology, brand new terminology was invented (EG: "Node").

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/aa/FiringGroupSplitterAa.java
DanVanAtta: Perhaps add a comment about the high level operations that are happening, what the code is roughly going to do.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/general/FiringGroupSplitterGeneralTest.java
DanVanAtta: I see, what I'm more pointing to is that we are testing multiple/all aspects of firingGroups per test rather than testing on behavior at a time. For example, it's considered a best practice for tests to have only one reason for why they would fail. If for example the firing group size were to change *or* if the displayName were to change, then this test would fail. At a more concrete level, isolating the behaviors makes it easier to see which parameters control which outputs.

Hence I'm suggesting to pivot the tests to group and isolate the tests that verify display name, size, etc.. and do so independently.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
trevan: The return value of this method is passed off to other methods and those pass off to other methods.  So the code path isn't easy to follow.

But I can't remember why exactly I added unmodifiableCollection.  I thought it was because that is how the original code was but I can't find any evidence of that now.  So, I've removed the unmodifiableCollection part.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
```diff
@@ -0,0 +1,141 @@
...
+      values.add(nonSuicideGroup);
+      return values;
+    }
+
+    int groupCount() {
+      return suicideGroups.keySet().size() + (nonSuicideGroup.isEmpty() ? 0 : 1);
+    }
+  }
+
+  String displayName;
```
DanVanAtta, line 43: Codacy found an issue: [Fields should be declared at the top of the class, before any method declarations, constructors, initializers or inner classes.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6278561)

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
```diff
@@ -0,0 +1,141 @@
...
+  String groupName;
+  Collection<Unit> firingUnits;
+  Collection<Unit> targetUnits;
+  boolean suicideOnHit;
```
DanVanAtta, line 47: Codacy found an issue: [Fields should be declared at the top of the class, before any method declarations, constructors, initializers or inner classes.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6278561)

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
```diff
@@ -0,0 +1,141 @@
...
+      return values;
+    }
+
+    int groupCount() {
+      return suicideGroups.keySet().size() + (nonSuicideGroup.isEmpty() ? 0 : 1);
+    }
+  }
+
+  String displayName;
+  String groupName;
```
DanVanAtta, line 44: Codacy found an issue: [Fields should be declared at the top of the class, before any method declarations, constructors, initializers or inner classes.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6278561)

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
```diff
@@ -0,0 +1,141 @@
...
+  Collection<Unit> firingUnits;
+  Collection<Unit> targetUnits;
```
DanVanAtta, line 46: Codacy found an issue: [Fields should be declared at the top of the class, before any method declarations, constructors, initializers or inner classes.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6278561)

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    int groupCount() {
+      return suicideGroups.keySet().size() + (nonSuicideGroup.isEmpty() ? 0 : 1);
+    }
+  }
+
+  String displayName;
+  String groupName;
+  Collection<Unit> firingUnits;
```
DanVanAtta, line 45: Codacy found an issue: [Fields should be declared at the top of the class, before any method declarations, constructors, initializers or inner classes.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6278561)

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/general/FiringGroupSplitterGeneralTest.java
trevan: You'll notice that the majority of the tests don't check for the display name.  Only this one and the other two I mentioned.  I understand that the test should generally only have one reason why they should fail.  And I believe this test is structured that way.

DanVanAtta: Can you clarify "test should generally only have one reason why they should fail. And I believe this test is structured that way."

From an initial look, it seems like if the firing group size changes this test would fail, or if the display name changes it would fail. The size of the firing group and display name are independent of one another, which makes me think that if one of two independent things were to either change, then this test method would fail for either reason. Are the two values, the firing group size and the display name actually linked?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/general/FiringGroupSplitterGeneral.java
DanVanAtta: Here is one thought, the list of split predicates is pretty well defined. What if we passed an enum value to specify the split type? This would I think be pretty beneficial as it would push complexity into this module, simplifying the callers. The cohesion of this class would be stronger, to understand what it's doing you would not have to look at the predicates being passed in, which is fundamentally a better and more useful module. Furthermore, by passing the right enum value, you get the right behavior, this really abstracts away implementation details from the callers. Otherwise, with a predicate a caller (or a maintainer, or someone looking to use or understand this code) has to have pretty deep knowledge of how the predicate is used.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/general/FiringGroupSplitterGeneralTest.java
trevan: This test is validating what a one firing group result should look like.  And a single firing group result will always have the same name.  This is a display name and the only reason to change the display name that is passed in is if there are multiple groups that need to be differentiated.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/aa/FiringGroupSplitterAa.java
trevan: Check out the comments that I added to the code.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroup.java
trevan: I've added comments about the suicideOnHit.  See if that works.

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleStepStrings.java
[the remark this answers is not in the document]
trevan: All of these strings are only used for the Battle UI in the step list on the left hand side.  I think a probably better thing would be to use some sort of localization library and have these strings put in the language files.  I don't know if that is in any future plans.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/general/FiringGroupSplitterGeneral.java
trevan: Do you mean the "filter type"?  Because the predicate is for filtering the units, it isn't for splitting the units.  But having an enum for the three different filters sounds like an ok idea.

trevan: I've changed the predicate to enums that determine the predicate.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/general/FiringGroupSplitterGeneralTest.java
```diff
@@ -86,13 +88,60 @@ void oneWaitingToDieFiringUnitVsOneTargetableUnitMakesOneFiringGroup() {
...
+  }
+
+  @Test
+  void defensiveNormalTypeExcludesFirstStrike() {
+    final Unit targetUnit = givenAnyUnit();
+    final Unit fireUnit = givenAnyUnit();
+    final Unit fireUnit2 = givenUnitFirstStrike();
+
+    final List<FiringGroup> firingGroups =
+        FiringGroupSplitterGeneral.of(OFFENSE, FiringGroupSplitterGeneral.Type.DEFENSIVE_NORMAL, "")
```
DanVanAtta, line 118: The empty string arg seems very similar to a 'null' arg or a boolean parameter. IN general those kinds of a args are not self-explanatory and make it difficult for someone to read and use an API. Notably, you can't just use the API, you have to look up which values are allowed to then realize empty string is the way to go. One would wonder, "can I use null here"? In any case, it makes you look up the implementation which is a heuristic for a difficult to use API. Similar for those reading, they'll be unsure why an empty string is needed.

Generally the fix is to have method overloads and have an API that omits the unnecessary parameter.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/general/FiringGroupSplitterGeneralTest.java
```diff
@@ -210,7 +283,7 @@ void excludeSuicideOnAttackTargetsIfDefense() {
     final Unit fireUnit = givenAnyUnit();
 
     final List<FiringGroup> firingGroups =
-        FiringGroupSplitterGeneral.of(DEFENSE, unit -> true, "")
```
DanVanAtta: Looking at these usage examples, and in the prod code, it looks like the update to a flag instead of passing a predicate was a strong update, it is a much nicer looking API now :+1:

DanVanAtta (approved):

PR #7830, opened 2020-10-03
DiceRoll.rollDice only needs the territory out of the battle

trevan:
This is a piece of #7823.  I noticed that DiceRoll.rollDice only grabs the territory of of the battle object.  So instead of passing the entire battle object around, I changed it to only pass the territory.

I also noticed that the units passed into DiceRoll.rollDice can be a `Collection` instead of a `List` so I changed that as well.


DanVanAtta (approved):
:+1:

PR #7831, opened 2020-10-03
DiceRoll.getAnnotation only needs territory and round from the battle

trevan:
This is a piece of #7823.  DiceRoll.getAnnotation doesn't need the full battle object.  It only needs the territory and round.  And the only caller of DiceRoll.getAnnotation (`Fire`) is guaranteed to have a battle object so the null check isn't needed.


DanVanAtta (approved):
:+1:

PR #7832, opened 2020-10-03
Improve AA Battle UI text

trevan:
This is a piece of #7823. 

I noticed during testing that the string shown in the battle ui when you need to select AA fire casualties was really strange.  It says "Select X casualties from aa fire in Y select X casualties.".  I tracked down where the redundant "select X casualties" and found that during the casualty selection, if the battle id is not null, it will always append "select X casualties".

So, I changed `AaCasualtySelector#getAaCasualties` to take a text string and then modified the callers to build a better string for their use case.  `AaInMoveUtil` keeps the old string because it has a battle id of null.  The other two callers (`FireAa` and `StrategicBombingRaidBattle`) now specify which aa type is firing and their string works with the auto-appended "select X casualties".  I basically followed how callers of `CasualtySelector#selectCasualties` construct the string.


game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelector.java
```diff
@@ -37,6 +37,7 @@ public static CasualtyDetails getAaCasualties(
       final Collection<Unit> allFriendlyUnits,
       final Collection<Unit> defendingAa,
       final Collection<Unit> allEnemyUnits,
+      final String text,
```
DanVanAtta, line 40: I don't know if pushing this complexity to the caller is a good move. The caller will have to know the full implementation details of this method to know how this text is being used.

I think for example one smell is the value of parameter is a composition of other parameters:
```
            "Select "
                + dice.getHits()
                + " casualties from "
                + currentTypeAa
                + " fire in "
                + territory.getName(),
```

- Perhaps we can fix this cleanly with a configuration flag? 

To another extent, the root cause seems to be the overall structure here and arguably a lack of abstractions altogether. I think this is a good example of a "routine", in other words a procedural method. The deeper fix here is likely to be quite the rabbit hole.

trevan: As I mentioned in the description, the "battleid" is basically used as a configuration flag.  It controls whether the entire text is used as is or if it will append ", select X casualties".

We are limited on what changes we can make because this goes through a RemoteAction.  I looked at a possibility of change and one idea is to only pass the `currentTypeAa` (or similar for non aa) through the text parameter.  In the final callees (`BattleDisplay#getCasualties` and `BattlePanel#getCasualtiesAa`), they can get access to the territory and the dice.  So at that point, they can generate the final string.

One potential issue is that `AirBattle.AttackersFire#execute` and `AirBattle.AttackersFire#execute` use a different string format: "Defenders Fire" and "Attackers Fire".  By making this change, they would now say something like "Hits from attackers" and "Hits from defenders".

DanVanAtta: TBH I really cringe by passing a UI specific text to this API, but I recognize the code we're working with is tangled into knots. Are there any other reasonable options, or is this really the least worst?

trevan: Considering we can't add additional parameters or change the types of the parameters, I think either leaving it as is or making the change to only passing the `currentTypeAa` and similar is what we have available.

DanVanAtta: :+1:

PR #7834, opened 2020-10-03
Fix: When AI builds factories, set remainingConstruction, not remainingUnitProduction

### LINKED ISSUE #7819: 2.3.21625 AI Factory/Town building in version TripleA v2.3+
I have a map I have been working on, it is War of the Lance with fixed links and my xml, a Fantasy map.

It is easy for me to use this map to test Factory/Town building, as I know it well.

By turn 5 the Imperial faction should of built another Town/Factory because they have spare PUs to do so.

Attached are two outputs from the Hard AI logs;
One from version TripleA 1.9.0.0.13066 and one from version 2.3.21625

v1.9 line 508 **Best factory unit: Town**
v2.3 line 615 should say Best factory unit: Town, **but it is missing**

Both versions have the line above
507 and 614 **Try to purchase factory for territory: Heartlund**

Im guessing, it would appear it is a logic problem or the routine is missing?

.
Finally in my testing I have noticed that ver 2.3 AI is **significantly** faster than 1.9, **thanks Devs!!!**

[1.9-WotL-Imp-T05.txt](https://github.com/triplea-game/triplea/files/5317558/1.9-WotL-Imp-T05.txt)
[2.3-WotL-Imp-T05.txt](https://github.com/triplea-game/triplea/files/5317559/2.3-WotL-Imp-T05.txt)

trevan:
During a reworking of the AI code (#5842), the method `ProPurchaseValidationUtils#removeInvalidPurchaseOptions` had its "remainingUnitProduction" parameter split to track "remainingConstructions" separately.  When the AI tried to buy a factory, it's code was not updated and it set "remainingUnitProduction" to 1 instead of "remainingConstructions".  This would cause the AI to always believe there wasn't enough space to build factories and so it would never build factories.

Fixes #7819

## Testing

Using the save game from #7819, I stepped through the code and was able to see factories constructed now.

## Release Note

FIX|Hard AI purchases factories/constructions again

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
@@ -1345,9 +1345,11 @@ class ProPurchaseAi {
           startOfTurnData,
           purchaseOptionsForTerritory,
           resourceTracker,
-          1,
+          0,
           new ArrayList<>(),
-          purchaseTerritories);
+          purchaseTerritories,
+          1,
+          maxTerritory);
       resourceTracker.clearTempPurchases();
 
       // Determine most expensive factory option (currently doesn't buy mobile factories)


tvleavitt:
In my opinion, this is a game-breaking bug for any game where AIs are involved (every single local single-player game), and thus justifies a new release, or some kind of hotfix to the current release (don't think we have a mechanism for that).

Looking at #5842, I note that all the testing done was manual, which points up the severe limitations of that as an approach. Clearly, this needed automated testing.

DanVanAtta:
This is a very significant fix @trevan , good work :+1:

DanVanAtta:
@tvleavitt  FWIW, 'hotfixes' are generally for releases that we un-release due to something that is game crashing. The window for such fixes is about a week or two following a release (ideally less). I agree that this patch is addressing a very major bug. 2.3 is close to being in a releasable state, i suspect it'll be launched in a week or two. Given that, the pain to download 2.3 with a bug fix vs a hotfix is about equal. Therefore I think it would be better to push the patch in the next release and have players upgrade to 2.3 once rather than pick up a patched 2.2.

tvleavitt:
What's needed to release 2.3? I assume that will actually be 2.4 if we're pursuing a stable/non-stable release process. Is integrating the PR @trevan is systematically breaking up into smaller chunks a requirement? Some of it is actually visible to the end-user. Is getting that done in the suggested timeframe a reasonable expectation?

DanVanAtta:
:grin: despite my best efforts to communicate the release model, we've left the stable/non-stable process perhaps about 5 years ago. The emphasis is for more testing per change and not a big ball of testing before release. We got into that anti-pattern when 2.0 was released and spent 2 months fixing problems. 1.8 had a problem where we spent 1 month fixing problems after release. We are getting better at it, but ultimately all changes to the code should be solid, the code should stay stable.

AFAIK 2.3 needs:
- a fix for the error reporting, that seems to have regressed
- some generalized smoke testing and pre-release regression testing

That is not necessarily that much. The changes @trevan  is in process of working on will likely straddle 2.3 to 2.4. For example, there is a new mechanism to download maps from our map server that was introduced in 2.1. That has not been released yet though as it's not turned on, but the code is turned on and if you turn on 'use beta features' you'll probably get an error downloading maps because the production maps server is not set up. That is an example, but if we keep code stable then it demonstrates that we can release at relatively arbitrary times because any code that is merged is solid, and any features that are in progress are turned off or otherwise hidden leaving only the fully complete and always stable code active.

DanVanAtta:
>  but ultimately all changes to the code should be solid, the code should stay stable.

With that said, we are still stuck doing a ton of manual testing at every change, and before every release. Better testing is a big priority because it's one of the areas that will most increase our efficiency, so much manual testing could be skipped if we automate more of it.

tvleavitt:
@DanVanAtta This is a meta-discussion, but I've already opened multiple discussion issues today, LOL. Is there a place to continue this? I want to help with the "smoke testing" and pre-release regression testing... is there any documented process around this, or do I just fiddle with the latest release until I see something and maybe run a game with all players set to Fast AI? Can't really do anything on this little MBA I'm working from this weekend, but I have a pretty decent desktop where I'm going to be all next week.

DanVanAtta:
@tvleavitt mostly it's fiddling around
We have a doc on testing, though it's not super complete: https://github.com/triplea-game/triplea/blob/master/docs/qa-testing.md

PR #7857, opened 2020-10-07
Move TotalPowerAndTotalRolls to power.calculator and mark helper classes as package only

trevan:
I noticed that the classes in power.calculator were just helper classes for TotalPowerAndTotalRolls.  So I moved it there and marked many of those classes as package visible instead of public.

I found some tests specific to TotalPowerAndTotalRolls in the DiceRoll tests and moved them to a separate test file.

I also moved a method to fit the code style of declaring methods immediately after they are used.

This change has no behavioral changes.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProBattleUtils.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProBattleUtils.java
@@ -15,7 +15,7 @@ import games.strategy.triplea.delegate.Matches;
 import games.strategy.triplea.delegate.TerritoryEffectHelper;
 import games.strategy.triplea.delegate.battle.UnitBattleComparator;
 import games.strategy.triplea.delegate.battle.casualty.CasualtyUtil;
-import games.strategy.triplea.delegate.dice.TotalPowerAndTotalRolls;
+import games.strategy.triplea.delegate.power.calculator.TotalPowerAndTotalRolls;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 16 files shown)


DanVanAtta:
@trevan you may want to check out this draft PR that never was merged:
- https://github.com/triplea-game/triplea/pull/6376

The PR got bogged down in details but hit a really good performance level *without* any caching. That performance IIRC was about 2x slower than the heavily cached version we have today and also IIRC was slightly more accurate.

The idea of the PR was:
- group all units in a battle by type
- remove one unit for each type and compute the remaining total effective power without that one unit
- choose the unit to remove based on the one that leaves the maximum effective power after removal
- decide to break ties based on additional heuristics, notably things like resource cost

If we could get the right datastructure, I think it was called a "unitGroup" as an input to a battle, the performance would have been much better. The calc code is structures kinda poorly and wound up re-computing the unit-groups repeatedly and consume a lot of CPU just doing that. I think the next level optimization would be to remove "N" units at a time, if you can remove for example 3 of 5 infantry before support levels are changed, then you could say "remove 3 infantry" and then skip 2 whole iterations.

Ultimately, I think the above algo is the right way to do it, you may notice in the tests there are comments stating that some of the current selections are actually buggy (not optimal choices). The code to calculate support is a major head-ache and is deceptively complex.

trevan:
> @trevan you may want to check out this draft PR that never was merged:
> 
>     * #6376

Is #6268 the issue where all the discussion was held?

DanVanAtta:
@trevan indeed, yes.

PR #7867, opened 2020-10-09
Add some battle terms to the glossary

trevan:
This just adds some terms to the glossary.


docs/glossary.md
```diff
@@ -11,3 +11,19 @@
...
+  - If an opposing side has an `isDestroyer` unit, this phase is skipped.
+  - WW2V2 always has this phase. The casualties will not be removed at the end of the phase if an `isDestroyer` unit is present on the opposing side.
+  - `Defending Subs Sneak Attack` game property controls whether the defender gets to fire in the phase.
+- **Targeted Attack Phase**: This phase is one of the first battle phases. Units determine their strength and roll based on their AA attributes (such as `attackAA` and `offensiveAttackAA`). If a unit has AA attributes, they can fire at this time. The `typeAA` and `targetsAA` attributes determine what units are targeted during this attack. At the end of the phase, all casualties are removed from the game or transformed into other units which do not participate in the battle. Units that fire during this phase can fire during later phases if they have normal strength and roll attributes.
```
DanVanAtta, line 24: I would have thought this was the 'anti-aircraft fire' phase. I suspect most players would not know what the 'targeted attack phase' is. Checking if players use this language naturally is a test if the domain terminology fits. In this case I don't think it necessarily does.

docs/glossary.md
```diff
@@ -11,3 +11,19 @@
 - **Game**: (AKA Game XML) refers to a specific XML within a map
 - **Map**: collection of files, eg: images, polygons text, properties, and at least one game XML
 
+## Battle Terms:
+- **Roll**: How many dice can the unit roll.
+- **Strength**: The number on the dice that determines whether a hit was made. Any number less than or equal to the number is a hit.
+- **Round**: A battle round includes both sides firing at each other. A battle can consist of multiple rounds. During and at the end of each round, casualties are removed from the game and can not participate anymore.
+- **Phase**: A battle phase is subpart of a battle round where specific rules are in play that determine what the units can do. Some phases allow the units to retreat. Other phases allow the units to fire at each other.
+- **General Phase**: This phase is the default phase for units to fire. The attacker fires first and then the defender retaliates. Any casualties that the attacker hits can retaliate. At the end of the phase, all casualties are removed from the game or transformed into other units which do not participate in the battle.
```
DanVanAtta, line 19: @panther2 , @Cernelius , can you weigh in if "general phase" is a term you would use and if this definition fits your understanding of that battle phase?

docs/glossary.md
```diff
@@ -11,3 +11,19 @@
...
+  - `Defending Subs Sneak Attack` game property controls whether the defender gets to fire in the phase.
+- **Targeted Attack Phase**: This phase is one of the first battle phases. Units determine their strength and roll based on their AA attributes (such as `attackAA` and `offensiveAttackAA`). If a unit has AA attributes, they can fire at this time. The `typeAA` and `targetsAA` attributes determine what units are targeted during this attack. At the end of the phase, all casualties are removed from the game or transformed into other units which do not participate in the battle. Units that fire during this phase can fire during later phases if they have normal strength and roll attributes.
+- **Retreat Phase**: This phase allows the attacker to retreat from battle. All units retreat together.
+  - Amphibious assaults only allow retreating of planes if `Attacker Retreat Planes` property is set or land based attackers and airplanes if `Partial Amphibious Retreat` is set.
+- **Can Evade Phase**: This phase allows units with `canEvade` to retreat or submerge (determined by `Submersible Subs` property).
```
DanVanAtta, line 27: Would 'Can evade phase' be better named the 'evasion phase'? 
Do we expose "can evade" as a label to players? I think "submerge subs" is the term most players would recognize. For other maps with non-sub units having 'canEvade', then calling it the 'evasion phase' would make sense (in this case we would have aliases, one is just more specific if you only have subs with 'canEvade').

docs/glossary.md
trevan: I have seen people refer to it as "targeted attacks" and as "aa attacks".  The most recent case was in the warcraft war heroes thread.  I assumed that the new terminology has been "targeted" since that includes more units than "aa".  But I really don't know.

docs/glossary.md
trevan: Is this glossary for developers (including map makers) and players?  I was under the impression this is more for developers.

I agree about showing the name "submerge/retreat subs" or "evade(?) units" depending on the situation.

docs/glossary.md
panther2: I would rather call it "General Combat Phase" or only "General Combat"

docs/glossary.md
panther2: Submerged units usually don't retreat. Also we have rulesets where the decision to submerge is made before rolling dice - and not during the Press Attack/Retreat step of a battle.
So this is not necessarily one and the same phase.

docs/glossary.md
trevan: I've renamed it to "AA Combat Phase" and added a new line of "Targeted Attack Combat Phase" that states it is a another name for "AA Combat Phase" and when it is used instead of "AA".

docs/glossary.md
trevan: I renamed it to "General Combat Phase".

docs/glossary.md
trevan: I've renamed this phase to "Evasion Phase" and added a similar term "Submerge Subs Phase" as another name for this phase.

> Submerged units usually don't retreat.
If `Submersible Subs` is not set, then submarines can retreat during this phase.  That's what I was trying to say with "retreat or submerge (determined by `Submersible Subs` property)".  I'm open on suggestions to how to reword that to be more clear.

> Also we have rulesets where the decision to submerge is made before rolling dice - and not during the Press Attack/Retreat step of a battle.
There is a line below this that says "`Sub Retreat Before Battle` defines if this phase occurs before the First Strike Phase or at the end of the Battle Round."  Is that clear enough?

DanVanAtta:
@trevan 

re:
> Is this glossary for developers (including map makers) and players? I was under the impression this is more for developers.

The glossary is a practice inspired by "Domain Driven Design" (DDD). In short one starts by defining a set of bounded contexts, or domains. These are areas that are self contained and have their own set of terminologies. For example, Map XML is a bounded context and the game player staging screen is another bounded context. The word "player" is an example of a term that has one meaning in one, another in the other. When asked, "how many players are in your game", you would answer 2 or 3 perhaps. If you ask, "how many players are defined in the XML", the answer is the number of 'factions.'

The other key concept is of a 'ubiquitous language'. The idea is for each bounded context there are domain experts, those that deal and interact within those domains. For example that would be the game players, the map makers. They have a language they use and terms are well defined for them. In developing a 'ubiquitous language' the ides is to learn what those terms are and use the same terms in code. This then creates a more natural and fitting data model and avoids a translation layer where the domain experts will understand the terms used by the code and developers will use the same terms in code and with the domain experts all with the same meaning.

TripleA has a lot of examples where this was not done, instead the developers invented their own terms and tried to keep everything unique even across different bounded contexts. For example, when XMLs were first made, it was natural to call each 'faction' a 'player'. But when it came time to talk about players joining games over network, these entities were instead called "Nodes". The term "node" in this context is not a natural fit, it creates this constant translation that must happen.

In summary, the language developers use and the domain experts, players and map makers, should match. Development, the code, should *not* seek to invent unique terminology that only means something to the code, the terms should match what people would use in natural language. 

For more, check out: 
- https://martinfowler.com/bliki/UbiquitousLanguage.html
- https://dev.to/microtica/the-concept-of-domain-driven-design-explained-1ccn




 In short, the concepts of a bounded context and a ubiquitous language come into play and arguably are the most important concepts. The idea is that you define 'domains' of your system that then constitute bounded contexts. The terms within the bounded context have well defined meanings and are relative to that context. For example, a banking system, a "transaction" within the context of money transfers can refer to sending money from one person to another, but in the context of an account ledge it could mean a purchase or receipt of funds.

The idea is that once you define the bounded context, you then define a set of terms, called the ubiquitous language


 discusses this topic and calls it the "ubiquitous language". The idea is that developers should not invent their own terminology. One can see a number of examples of this in TripleA.

docs/glossary.md
```diff
@@ -16,14 +16,17 @@
...
+- **General Combat Phase**: This phase is the default phase for units to fire. The attacker fires first and then the defender retaliates. Any casualties that the attacker hits can retaliate. At the end of the phase, all casualties are removed from the game or transformed into other units which do not participate in the battle.
+- **First Strike Combat Phase**: This phase is for units to fire and remove casualties before the General Phase. Historically, this was used for submarines to sink sea based units as a "sneak attack". Now, any unit can participate during this phase if they have the `isFirstStrike` attribute. Units that fire during this phase will not participate during the General Phase. At the end of the phase, all casualties are removed from the game or transformed into other units which do not participate in the battle.
```
DanVanAtta, line 20: This perhaps could just be called "First Strike Combat". 

I think the test for this is to consider the sentence "That occurs during first strike combat". Does that have well defined meaning, or does it need to be qualified to know what is being referenced, eg: "That occurs during the first strike combat phase".

docs/glossary.md
DanVanAtta: My 2 cents, 'combat phase' without qualifier is the 'general combat phase'.

I'd opt for either:
- combat phase
or
- general combat

The latter being a phase of combat I think is implied by the qualifier of 'combat', a battle is the whole, combats are typically phases within a battle (or phases within a combat)

DanVanAtta:
@trevan some bike shedding is worthwhile here as names are very sticky. Once set, it's difficult to rename things and often instead it creates aliases rather than true renames. With some apologies I do think we need to spend some proper time here to really think through what the right language would be so we can do this once and un-ambiguously

DanVanAtta:
@trevan I feel like I should more concisely answer your question. Yes, this glossary is for developers, but the terms should be those used by domain experts (eg: players, map makers) and it is so developers can use _those_ terms and not invent their own.

trevan:
I just came across some UI code that uses "power" for the dice strength and "targeted" for the aa roll and strength.  See https://github.com/triplea-game/triplea/blob/master/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java#L3353.

DanVanAtta:
@trevan good find, that makes me think that perhaps one of strength or power were a legacy term and then the distinction is between total power and power.. It's really unclear to me right now. It sounds like we need a small project to survey the current usage, decide what to go with, and then update as much as we can to be consistnet.

trevan:
Here's some other terms that need to be hashed out:

* There is the AA dice roll which only is rolled during the AA phase.  But then there this the "normal" dice roll that can be rolled for bombardment, general, and first strike.  The code has places that call this "normal" (see `TotalPowerAndTotalRolls` before my changes) but it isn't very descriptive.  What would be a better term than "normal"?  I can see separating "bombardment" out of the group but the dice settings for first strike and general are identical: `attack`/`defense` and `attackRolls`/`defenseRolls`.

* Remove casualties can both mean "remove behind the casualty line but the units can still fire back" and "remove from the game completely".  When AA/Targeted attack happens, the UI shows "Remove casualties" after each AA/Targeted attack but the units can still fire back.  But during the first strike or general combat, the "Remove casualties" is only shown at the end of the entire phase and the units are removed completely.  But in the code, the AA/Targeted and first strike /general combat actually have a "remove casualties" step which is the same as what the UI is showing for AA/Targeted and a "clear casualties" which is what the UI is showing for the first strike/general combat.  Is "clear casualties" a good enough difference from "remove casualties"?  Or should the two parts have different names?

docs/glossary.md
panther2: Sounds good to me.

DanVanAtta:
@trevan let's pick up the hashing out of other terms in an issue: https://github.com/triplea-game/triplea/issues/7892

docs/glossary.md
```diff
@@ -11,3 +11,22 @@
 - **Game**: (AKA Game XML) refers to a specific XML within a map
 - **Map**: collection of files, eg: images, polygons text, properties, and at least one game XML
 
+## Battle Terms:
+- **Roll**: How many dice can the unit roll.
+- **Strength**: The number on the dice that determines whether a hit was made. Any number less than or equal to the number is a hit.
+- **Round**: A battle round includes both sides firing at each other. A battle can consist of multiple rounds. During and at the end of each round, casualties are removed from the game and can not participate anymore.
+- **Phase**: A battle phase is subpart of a battle round where specific rules are in play that determine what the units can do. Some phases allow the units to retreat. Other phases allow the units to fire at each other.
```
DanVanAtta, line 18: nit, perhaps fully qualify this as a 'Battle Phase', to disambiguate from a 'turn phase'. In other words, a 'phase' can be multiple things, so I think we are being specific to a 'battle phase' here.

docs/glossary.md
```diff
@@ -11,3 +11,22 @@
...
+  - If an opposing side has an `isDestroyer` unit, this phase is skipped.
+  - WW2V2 always has this phase. The casualties will not be removed at the end of the phase if an `isDestroyer` unit is present on the opposing side.
+  - `Defending Subs Sneak Attack` game property controls whether the defender gets to fire in the phase.
+- **Submarine Combat Phase**: Another name for **First Strike Phase** and generally used when only submarines have first strike capability.
+- **AA Combat Phase**: This phase is one of the first battle phases. Units determine their strength and roll based on their AA attributes (such as `attackAA` and `offensiveAttackAA`). If a unit has AA attributes, they can fire at this time. The `typeAA` and `targetsAA` attributes determine what units are targeted during this attack. At the end of the phase, all casualties are removed from the game or transformed into other units which do not participate in the battle. Units that fire during this phase can fire during later phases if they have normal strength and roll attributes.
```
DanVanAtta, line 25: > " This phase is one of the first battle phases."

Is that always true, or just typically true due to typical battle phase ordering? Perhaps the mention of ordering can be removed and focus on defining what happens in the phase.

DanVanAtta (commented):
Just a couple quick more comments, this seems almost there!

docs/glossary.md
trevan: I removed the ordering.  But AA is currently the first battle phase.

DanVanAtta: I can't remember if we make it configurable at all. Sounds like not :)
But in general, always best to create loosely coupled documentation, anything to keep it more up-to-date for longer.

PR #4215, opened 2018-10-24
Prompt user to load newer save game on EDT

ssoloff:
## Overview

Fixes #4208.

## Functional Changes

* Ensure the prompt to load a newer save game runs on the EDT.

## Refactoring Changes

* Extracted the code to prompt the user with the out-of-date engine message dialog to a new method.

## Manual Testing Performed

Ran through the scenario described in #4208: play a peer-to-peer network game with a client from this branch and a server from the custom 1.9.0.3050 release.  Verified that the out-of-date engine message dialog is displayed and the user can interact with it.

* When answering **Yes**, verified the game starts for both players, as expected.  (Note that other errors occur once the game starts due to this branch no longer being compatible with the 1.9.0.3050 build because of recent incompatible Checkstyle violation fixes.)
* When answering **No**, verified the client game does not start and that the server game returns to the host network game screen once the client disconnects (e.g. by returning to the main menu or quitting the application).

PR #7871, opened 2020-10-10
Re-add warning about save from a newer compatible version

trevan:
This was removed in 030cb1f because
of a swing deadlock issue and was also considered to be redundant.

It is needed because newer compatible versions might use new saved data
to add or change functionality. When an older version opens the saved
game, the new saved data will be removed and the functionality might end
up broken.

## Testing

I changed the version of my build to 2.1 and then tried to load a saved game.  The warning showed up.  Clicking "Yes" allowed the game to be loaded.  Clicking "No" stopped the loading of the game.

## Screens Shots

Screen Shot 2020-10-10 at 10 35 05 AM

## Release Note

CHANGE|Opening a save game from a newer version will warn that the versions are mismatched

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/GameDataManager.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/GameDataManager.java
@@ -95,27 +95,7 @@ public final class GameDataManager {
                 UrlConstants.DOWNLOAD_WEBSITE));
       } else if (!HeadlessGameServer.headless()
           && ((Version) version).isGreaterThan(ClientContext.engineVersion())) {
-        // we can still load it because our engine is compatible, however this save was made by a
-        // newer engine, so prompt the user to upgrade
-        final String messageString =
-            "Your TripleA engine is OUT OF DATE.  This save was made by a newer version of TripleA."
-                + "\nHowever, because the first version number is the same as your current version, we can "
-                + "still open the savegame."
-                + "\n\nThis TripleA engine is version "
-                + ClientContext.engineVersion().toString()
-                + " and you are trying to open a savegame made with version "
-                + ((Version) version).toString()
-                + "\n\nTo download the latest version of TripleA, Please visit "
-                + UrlConstants.DOWNLOAD_WEBSITE
-                + "\n\nIt is recommended that you upgrade to the latest version of TripleA before playing this "
-                + "savegame."
-                + "\n\nDo you wish to continue and open this save with your current 'old' version?";
-        final int answer =
-            JOptionPane.showConfirmDialog(
-                null, messageString, "Open Newer Save Game?", JOptionPane.YES_NO_OPTION);
-        if (answer != JOptionPane.YES_OPTION) {
-          throw new IOException("Loading the save game was aborted");
-        }
+        promptToLoadNewerSaveGame((Version) version);
       }
 
       final GameData data = (GameData) input.readObject();
@@ -124,6 +104,35 @@ public final class GameDataManager {
       return data;
     } catch (final ClassNotFoundException cnfe) {
       throw new IOException(cnfe.getMessage());
+    }
+  }
+
+  private static void promptToLoadNewerSaveGame(final Version version) throws IOException {
+    // we can still load it because our engine is compatible, however this save was made by a
+    // newer engine, so prompt the user to upgrade
+    // this is needed because a newer client might depend on variables that are part of the new
+    // save but will be stripped by the old client. When the old client re-saves the data, the
+    // new client will load the save game and be in odd state.
+    final String messageString =
+        "This save was made by a newer version of TripleA."
+            + "\nHowever, because the first version number is the same as your current version, "
+            + "we can still open the savegame. There is a possibility that if you save this game "
+            + "and then open it up in the version that created it, the game will be in an "
+            + "incorrect state."
+            + "\n\nThis TripleA engine is version "
+            + ClientContext.engineVersion().toString()
+            + " and you are trying to open a save game made with version "
+            + version.toString()
+            + "\n\nTo download the latest version of TripleA, Please visit "
+            + UrlConstants.DOWNLOAD_WEBSITE
+            + "\n\nIt is recommended that you upgrade to the latest version of TripleA before "
+            + "playing this save game."
+            + "\n\nDo you wish to continue and open this save with your current 'old' version?";
+    final int answer =
+        JOptionPane.showConfirmDialog(
+            null, messageString, "Open Newer Save Game?", JOptionPane.YES_NO_OPTION);
+    if (answer != JOptionPane.YES_OPTION) {
+      throw new IOException("Loading the save game was aborted");
     }
   }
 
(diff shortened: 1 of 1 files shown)


trevan:
The PR that removed this warning is https://github.com/triplea-game/triplea/pull/5211

It mentions a deadlock but I haven't seen the deadlock.  That could be that the code that was removed was using swing while the original code that I copied this PR from is just using `JOptionPane.showConfirmDialog`.

trevan:
I tracked down when it moved to swing - https://github.com/triplea-game/triplea/pull/4215.  That also talks about a deadlock when connecting to a server.

So, I'm not sure how to fix the dead lock issue.

DanVanAtta:
I think the deadlock was due to the background process runner spinner dialog. That dialog blocks the EDT and meanwhile this warning tries to render. The background runner really does have some liabilities, if it is ever waiting on a process that needs to render using the EDT then we'll have a deadlock.

DanVanAtta:
To avoid a deadlock due to the background process runner, you'll need to track down the code execution paths leading to this code and ensure that we stop the background process runner before showing this dialog.

game-core/src/main/java/games/strategy/engine/framework/GameDataManager.java
```diff
@@ -102,6 +107,35 @@ public static GameData loadGame(final InputStream is) throws IOException {
     }
   }
 
+  private static void promptToLoadNewerSaveGame(final Version version) throws IOException {
+    // we can still load it because our engine is compatible, however this save was made by a
```
DanVanAtta, line 111: This comment does not seem to belong in this method. The comment is describing why we can load a newer save game, but the method is all about prompting and warning the user.

game-core/src/main/java/games/strategy/engine/framework/GameDataManager.java
```diff
@@ -102,6 +107,35 @@ public static GameData loadGame(final InputStream is) throws IOException {
...
+    // newer engine, so prompt the user to upgrade
+    // this is needed because a newer client might depend on variables that are part of the new
+    // save but will be stripped by the old client. When the old client re-saves the data, the
+    // new client will load the save game and be in odd state.
+    final String messageString =
+        "This save was made by a newer version of TripleA."
+            + "\nHowever, because the first version number is the same as your current version, "
```
DanVanAtta, line 118: > nHowever, because the first version number is the same as your current version, "
            + "we can still open the savegame. 

This is a level of detail that we can probably omit. The fact that the game is offering to load the save at all already implies this level of compatibility.

game-core/src/main/java/games/strategy/engine/framework/GameDataManager.java
```diff
@@ -102,6 +107,35 @@ public static GameData loadGame(final InputStream is) throws IOException {
...
+            + "we can still open the savegame. There is a possibility that if you save this game "
```
DanVanAtta, line 119: > There is a possibility that if you save this game "
            + "and then open it up in the version that created it, the game will be in an "
            + "incorrect state.

We can probably state this more simply. Perhaps something like:

```
Playing newer saves with an older engine can lead to unpredictable problems. It is recommended that you upgrade to the latest version of TripleA before playing this save game. To download the latest version of TripleA, Please visit UrlConstants.DOWNLOAD_WEBSITE"
+ \n\nDo you wish to continue and open this save with your current 'old' version?";
```

I think there might be a usability problem here as well because the URL cannot be copy/pasted and a user has to type it in manually. We can make the link clickable, I suspect that would be the best UX. @trevan is that something you would be willing to do?

"JEditorPaneWithClickableLinks" is the class that enables messages with clickable links.

game-core/src/main/java/games/strategy/engine/framework/GameDataManager.java
```diff
@@ -102,6 +107,35 @@ public static GameData loadGame(final InputStream is) throws IOException {
...
+            + "\n\nTo download the latest version of TripleA, Please visit "
+            + UrlConstants.DOWNLOAD_WEBSITE
+            + "\n\nIt is recommended that you upgrade to the latest version of TripleA before "
+            + "playing this save game."
+            + "\n\nDo you wish to continue and open this save with your current 'old' version?";
+    final int answer =
+        JOptionPane.showConfirmDialog(
+            null, messageString, "Open Newer Save Game?", JOptionPane.YES_NO_OPTION);
+    if (answer != JOptionPane.YES_OPTION) {
+      throw new IOException("Loading the save game was aborted");
```
DanVanAtta, line 135: It's sad this is how we abort save games, control-flow-by-exception handling. We are also masking other, legitimate causes of IOException as well, for example a permission denied trying to read the save game file. I"m not sure if there is much to do about this problem now, but it's certainly not ideal.

DanVanAtta (commented):
I think we can/should improve the wording and it would be a better UX to have a clickable download link.

@trevan please give your feedback on that and the comments left, hopefully those updates would be straight forward. I think we also need to dig into the deadlock situation, ensure there are not progress runner dialogs active when we show this warning message.

trevan:
I can make the text change, but I don't have much experience with java ui so it will take a while for me to figure out the deadlock and the html link.  It would probably be better if there was someone with more experience in this area or code.

tvleavitt:
"Please" should not be capitalized.

DanVanAtta:
@trevan I can assist, I'll push a few updates to this branch.

game-core/src/main/java/games/strategy/engine/framework/GameDataManager.java
trevan: I've updated the text with what you wrote.  I haven't made the link clickable, though.

game-core/src/main/java/games/strategy/engine/framework/GameDataManager.java
trevan: I moved the comment to the caller.

game-core/src/main/java/games/strategy/engine/framework/GameDataManager.java
```diff
@@ -102,6 +109,26 @@ public static GameData loadGame(final InputStream is) throws IOException {
     }
   }
 
+  private static void promptToLoadNewerSaveGame(final Version version) throws IOException {
```
DanVanAtta, line 112: Codacy found an issue: [Avoid unused method parameters such as 'version'.](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6317906)

DanVanAtta:
@trevan   I pushed a fix for the PMD violation. Let's merge after the build passes and I'll follow up with some further changes.

I'm now thinking that probably we should enforce a new version to be installed. I think that is probably what players would actually prefer really. It is not a fun pill to swallow, but if someone proceeds at their own risk, and then the rules don't work properly, or they get other problems, they'll then wonder "why did you let me play a new save with an old engine?"

Furthermore, it's very important when two players are playing against each other and have different engine versions. That is where rule differences are going to be critical. Less important is a use-case where someone is loading save games to "check out" the save game, in which case this feature is a convenience but a real cost for a corrupted game. Ensuring games are played accurately is one of our highest priorities, far more than a convenience feature.

DanVanAtta:
@trevan  of some interest, the UI window that handles logger notifications already uses the editor pane that renders clickable links. All we need to do is supply html formatted content and the links should be clickable provided we use the logger mechanism.

PR #7876, opened 2020-10-11
Update: Require newer save loads to upgrade engine

DanVanAtta:
## Overview
(1) Return an optional in GameDataManager load and do not throw exceptions. Of note, when we return an optional it means that any errors have been handled, the calling code only needs to deal with the empty optional (and does not need to message the user).

(2) Change the yes/no confirmation of loading a newer save game to instead be a warning message that user should upgrade their game engine to play the save game

## Commits

commit 12e854966a3bff255267a4dee510fd9d51fec68c

    Remove unnecessary null check, new input stream will thrown NPE on the next line

commit 0e134e518b5b564951d135ec4fb6c5ad897ab179

    Return an optional instead of throwing an exception on game load
    
    Rather than control-flow-by-exception, we return an optional and have
    the lower layers take care of error handling and notifications.

commit 9c0967c0c31f15ceb519001b76928ddd4e9c5c2f

    Finish converting game loading to return an optional and handle thrown errors

commit 64a8c1d435b3cba6e1d0c3a94dfe936c93c896ae

    Do not load future save game versions, only prompt user to upgrade their engine
    
    There are subtle risks with playing a *maybe* forward compatible version of TripleA
    with a new save. To help ensure rule integrity, this update removes a yes/no prompt
    for users to continue in such situations and instead instructs them to download
    a newer game engine.

commit 98ef72ef32de73df9026c9bc51b30c5b5c400a6c

    Fix html formatting of load newer save game message

commit 0b5dd59d7da6b5236e5bfbb5658d9e3963a24308

    Remove unused @Log annotation

## Testing

- generated artificial errors and loggings when loading a client game to verify interplay with game loading spinners
- manually adjusted product version to verify newer save game message

## Release Note

UPDATE|Loading saves from newer game engine versions will require you to upgrade your game engine version. This is to avoid subtle errors and other rule problems that could occur.


game-core/src/main/java/games/strategy/engine/framework/GameDataManager.java
```diff
@@ -93,39 +94,28 @@ public static GameData loadGame(final InputStream is) throws IOException {
...
-            + "\n\nDo you wish to continue and open this save with your current 'old' version?;";
-    final int answer =
-        JOptionPane.showConfirmDialog(
-            null, messageString, "Open Newer Save Game?", JOptionPane.YES_NO_OPTION);
-    if (answer != JOptionPane.YES_OPTION) {
-      throw new IOException("Loading the save game was aborted");
+      log.error("Error loading game data", cnfe);
+      return Optional.empty();
+    } catch (final IOException e) {
+      return Optional.empty();
```
trevan, line 118: Should a `log.error` be added here as well?

trevan (commented):
I don't know most of the top callers so I'm not sure if returning vs throwing an exception could cause a problem with them, but other than the one comment I made it looks fine.

game-core/src/main/java/games/strategy/engine/framework/GameDataManager.java
DanVanAtta: Yes, an error of omission on my part. That was something I had intended to come back to, updated in: c7ae2cac3

DanVanAtta:
> I don't know most of the top callers so I'm not sure if returning vs throwing an exception could cause a problem with them,

FWIW, on the plus side, by converting from a return type of `GameData` to `Optional<GameData>`, the type change causes compilation errors and requires us to update all of the callers.

tvleavitt:
@DanVanAtta Have you considered that this will prevent folks from being able to load a savegame from a pre-release into the latest release to test whether an encountered problem is reproducible in the release version? Or does this only prompt for an upgrade to a release?

DanVanAtta:
@tvleavitt , no I had not. It is an interesting consideration. In part the argument for requiring an upgrade is that forward compatibility is not guaranteed, therefore testing would not necessarily be guaranteed either. 

While nerfing test convenience should not be done lightly, having this play out correctly for the main case is more important IMO. Perhaps if this becomes a common problem for testing we can build in a feature to bypass the check?

tvleavitt:
> @tvleavitt , no I had not. It is an interesting consideration. In part the argument for requiring an upgrade is that forward compatibility is not guaranteed, therefore testing would not necessarily be guaranteed either.
> 
> While nerfing test convenience should not be done lightly, having this play out correctly for the main case is more important IMO. Perhaps if this becomes a common problem for testing we can build in a feature to bypass the check?

I think an option to bypass would make sense in this particular case, put it on the "Testing" tab.

DanVanAtta:
Bypass option: https://github.com/triplea-game/triplea/pull/7894

PR #7877, opened 2020-10-11
Roll dice step

trevan:
This is a part of #7823. 

During the firing part of the battle, there are three steps that are performed (you can see the originals in `Fire` and `FireAa`).  This is the step that rolls the dice.  There are two different ways to roll dice: AA and Normal.  So, I moved that logic into `BiFunction`s that are passed into this object.  `RollNormal` is the normal version.  And `AaFireAndCasualtyStep.RollAa` is the Aa version.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
@@ -75,6 +75,9 @@ public interface BattleStep extends IExecutable {
     OFFENSIVE_GENERAL_RETREAT,
     STALEMATE_BATTLE_END_CHECK,
     SUB_DEFENSIVE_RETREAT_AFTER_BATTLE,
+
+    // Fire Round steps
+    ROLL_DICE,
   }
 
   /** @return a list of names that will be shown in {@link games.strategy.triplea.ui.BattlePanel} */
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
@@ -76,8 +76,7 @@ public interface BattleStep extends IExecutable {
     STALEMATE_BATTLE_END_CHECK,
     SUB_DEFENSIVE_RETREAT_AFTER_BATTLE,
 
-    // Fire Round steps
-    ROLL_DICE,
+    FIRE_ROUND_ROLL_DICE,
   }
 
   /** @return a list of names that will be shown in {@link games.strategy.triplea.ui.BattlePanel} */
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/RollDice.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/RollDice.java
@@ -38,20 +38,24 @@ public class RollDice implements BattleStep {
 
   private String getName() {
     return battleState.getPlayer(side).getName()
-        // displaying UNITS makes the text feel redundant so hide it if that is the group name
+        // If the firing group's name is the default "UNITS", then the displayed string will be
+        // "Germans units fire". In that case, it would be better to say "Germans fire".
+        // If the firing group's name is something else, then the displayed string will be
+        // something like "Germans submarines fire" which is ok.
         + (firingGroup.getDisplayName().equals(UNITS) ? "" : " " + firingGroup.getDisplayName())
         + FIRE_SUFFIX;
   }
 
   @Override
   public Order getOrder() {
-    return Order.ROLL_DICE;
+    return Order.FIRE_ROUND_ROLL_DICE;
   }
 
   @Override
   public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
 
-    // remove any target unit that was hit by other units
+    // retain the targets that are still alive since the targets might have been shot
+    // in an earlier firing round
     firingGroup.retainAliveTargets(battleState.filterUnits(ALIVE, side.getOpposite()));
 
     final DiceRoll dice = rollDice.apply(bridge, this);
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/RollDice.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/RollDice.java
@@ -1,0 +1,61 @@
+package games.strategy.triplea.delegate.battle.steps.fire;
+
+import static games.strategy.triplea.delegate.battle.BattleState.UnitBattleFilter.ALIVE;
+import static games.strategy.triplea.delegate.battle.BattleStepStrings.FIRE_SUFFIX;
+import static games.strategy.triplea.delegate.battle.BattleStepStrings.UNITS;
+
+import games.strategy.engine.delegate.IDelegateBridge;
+import games.strategy.triplea.delegate.DiceRoll;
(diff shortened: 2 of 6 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
```diff
@@ -75,6 +75,9 @@
     OFFENSIVE_GENERAL_RETREAT,
     STALEMATE_BATTLE_END_CHECK,
     SUB_DEFENSIVE_RETREAT_AFTER_BATTLE,
+
+    // Fire Round steps
```
DanVanAtta, line 79: Is the comment used to group the enums? 
A bit nitpicky, I'd recommend to avoid it. As a casual reader here, it does not make a ton of sense what the comment is delimiting. It's also a bit of an issue for maintenance, does every line break indicate a new group? Are the groups all up to date?

If needing to mark certain enum values, there should be other ways to do it. For example, one can add name patterns, like a prefix, add flags, or use additional enum types, or convert the enums to strings.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/RollDice.java
```diff
@@ -0,0 +1,61 @@
...
+  private final BiFunction<IDelegateBridge, RollDice, DiceRoll> rollDice;
+
+  @Override
+  public List<String> getNames() {
+    return List.of(getName());
+  }
+
+  private String getName() {
+    return battleState.getPlayer(side).getName()
+        // displaying UNITS makes the text feel redundant so hide it if that is the group name
```
DanVanAtta, line 41: The uncertainty in this comment is interesting  "makes the text feel redundant"

Yoda principle: "do or do not, there is no try"

I recommend making this comment more direct, the code is deterministic, the comment should be as well.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/RollDice.java
```diff
@@ -0,0 +1,61 @@
...
+        + (firingGroup.getDisplayName().equals(UNITS) ? "" : " " + firingGroup.getDisplayName())
```
DanVanAtta, line 42: Not for this PR, but you considered making the display name an object? I wonder if there could be some brittleness with the comparison of strings here. It also seems like we are violating SRP, the display string is serving double duty to both indicate a type and also used for display purpose.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/RollDice.java
```diff
@@ -0,0 +1,61 @@
...
+  @Override
+  public Order getOrder() {
+    return Order.ROLL_DICE;
+  }
+
+  @Override
+  public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
+
+    // remove any target unit that was hit by other units
+    firingGroup.retainAliveTargets(battleState.filterUnits(ALIVE, side.getOpposite()));
```
DanVanAtta, line 55: Could we remove the comment by updating the name of the method? EG:
```
firingGroup.removeUnitsThatWereHit(....);
```

We might want to consider more than that. Offhand I'm not sure just reading this line makes sense. We are getting the alive units of the other side and keeping only those units from within the current firing group. The fact we are getting the other side and are operating on the current firing group seems confusing. WDYT?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/RollNormal.java
```diff
@@ -0,0 +1,30 @@
...
+
+import static games.strategy.triplea.delegate.battle.BattleState.Side.DEFENSE;
+import static games.strategy.triplea.delegate.battle.BattleState.UnitBattleFilter.ACTIVE;
+
+import games.strategy.engine.delegate.IDelegateBridge;
+import games.strategy.triplea.delegate.DiceRoll;
+import java.util.function.BiFunction;
+
+/** Rolls dice for normal (basically, anything that isn't AA) dice requests */
+public class RollNormal implements BiFunction<IDelegateBridge, RollDice, DiceRoll> {
```
DanVanAtta, line 11: Is RollNormal an equivalent thing to "RollGeneralCombat" ?
I wonder if we are mixing "normal" and "general" as synonymous terms.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/RollNormal.java
```diff
@@ -0,0 +1,30 @@
...
+
+  @Override
+  public DiceRoll apply(final IDelegateBridge bridge, final RollDice step) {
+    return DiceRoll.rollDice(
```
DanVanAtta, line 15: side-question, any plans to refactor the rollDice API in the future? Seemingly we should be able to pass a step and bridge to it and little else. It's a bit surprising there is enough data between the 2 or 3 variables to supply the nearly dozen parameters.

DanVanAtta (commented):
Couple comments, questions, mostly around naming and improving comments. Looks good otherwise.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/RollDice.java
trevan: I've made the comment longer.  The name of the method is ok, though.  This is retaining targets which are the units that are being shot at.  So the targets are from the opposite side.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/RollNormal.java
trevan: No, normal and general are not synonymous.  Though I'm open to new names.  General is for a phase in the combat.  But normal is for the type of unit.  You have AA/Targeted dice requests which are rolled during the AA phase.  And then you have normal dice requests which rolled in all the other phases (bombard, first strike, and general).  The use of normal comes from `TotalPowerAndTotalRolls`.  It uses "normal" and "aa" there in that manner and I've been following it here since I didn't think a class called `Roll` would be ok.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/RollNormal.java
trevan: Yes, I do plan to refactor it.  I'd actually like to just have one `rollDice` api instead of a "normal" and an "aa" roll dice api.

The refactoring for `TotalPowerAndTotalRolls` is the start of that work.  The list of parameters should shrink as I group them into more related objects.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/RollNormal.java
DanVanAtta: Perhaps then `MainBattleRoll` as a more descriptive name?

In general, "general", "normal", and "standard" are context dependent descriptors and maintainers and other developers simply won't have that context without going through some work. Saving any amount of work to make code more understandable leads to more maintainable code.

Other suggestions: `MainRoll`, `StandardRoll`

trevan: `Main` sounds good.  I renamed it to `RollMainDice`.

DanVanAtta (approved):
:+1:

PR #7878, opened 2020-10-11
Add mark casualties battle step

trevan:
This is a part of #7823.

During the firing part of the battle, there are three steps that are performed (you can see the originals in Fire and FireAa). This is the step that marks the casualties (after the user has selected them which is another step).


game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleActions.java
```diff
@@ -59,4 +64,12 @@ Territory querySubmergeTerritory(
       GamePlayer retreatingPlayer,
       Collection<Territory> availableTerritories,
       String text);
+
+  default Player getRemotePlayer(final GamePlayer player, final IDelegateBridge bridge) {
+    // if its the null player, return a do nothing proxy
+    if (player.isNull()) {
+      return new WeakAi(player.getName());
```
DanVanAtta, line 71: side-note, this dependency on `WeakAi` is something we really would like to get rid of:
- allows for all of weak ai to be deleted
- kinda poor design, the core game engine should not depend on AI code

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
```diff
@@ -107,4 +109,8 @@ public boolean isFirstRound() {
   Collection<Unit> getTransportDependents(Collection<Unit> units);
 
   Collection<IBattle> getDependentBattles();
+
+  @RemoveOnNextMajorRelease("Only used to handle step string changes")
```
DanVanAtta, line 113: nit, I'm not sure the annotation comment string here makes a lot of sense. Would we be just as well off if it were removed? Otherwise, can it be made more direct, is it indicating that we have to do something else beyond upgrading to a non-compatible release to remove this method?

DanVanAtta (commented):
Looks like largely existing code, just one very minor comment to look at.

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleActions.java
trevan: I grabbed this code from `AbstractBattle#getRemotePlayer`.  I'm not exactly sure why it is needed.

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
trevan: I've removed the annotation comment string.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/MarkCasualties.java
```diff
@@ -0,0 +1,190 @@
...
+      notifyCasualties(bridge);
+    }
+
+    battleActions.removeCasualties(
+        fireRoundState.getCasualties().getKilled(),
+        returnFire,
+        side.getOpposite() == DEFENSE,
+        bridge);
+
+    if (firingGroup.isSuicideOnHit()) {
```
trevan, line 84: This is where the isSuicideOnHit check is being used.  You had wondered about that in the FiringGroup PR.

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleActions.java
DanVanAtta: I do recognize this code, this is one of the few and biggest dependencies on AI code from core-code IIRC. In essence we do this hack to supply a functioning player object and if my vague memory on this really holds for a default OOL algorithm. Using weakAi is a very ugly hack (which we've had for a long time; we can outright delete all of weak AI once we remove this dependency, which would be a significant simplification and removal of dead code).

DanVanAtta (approved):
Looks good, I'm assuming most of the logic is ported from an existing location.

A few merge conflicts came up, should be good to merge once resolved.

trevan:
> Looks good, I'm assuming most of the logic is ported from an existing location.
> 
> A few merge conflicts came up, should be good to merge once resolved.

Yes, most of the logic is ported from `Fire` and `FireAa`.  I did refactor the logic in the "select casualties" step.  I just noticed that I had missed adding my test to this PR so I've added.

I also fixed the merge conflicts.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/SelectMainBattleCasualtiesTest.java
```diff
@@ -0,0 +1,526 @@
...
+  }
+
+  private DiceRoll givenDiceRollWithHits(final int hits) {
+    final DiceRoll diceRoll = mock(DiceRoll.class);
+    when(diceRoll.getHits()).thenReturn(hits);
+    return diceRoll;
+  }
+
+  @Test
+  void isEditMode() {
```
DanVanAtta, line 59: nit, test method feels to be missing the implication of what happens when edit mode. A `@DisplayName` perhaps should become routine for any test method over 5 lines and particularly if there is more than one assert (think of it as a way to replace a javadoc on each test method, and each one should have a summary either in a direct title or display name to describe what we are testing).

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/SelectMainBattleCasualtiesTest.java
```diff
@@ -0,0 +1,526 @@
...
+        mock(SelectMainBattleCasualties.Select.class);
+    final CasualtyDetails expected = new CasualtyDetails(targetUnits, List.of(), false);
+    when(selectFunction.apply(any(), any(), anyCollection(), anyInt())).thenReturn(expected);
+
+    final CasualtyDetails details =
+        new SelectMainBattleCasualties(selectFunction).apply(delegateBridge, selectCasualties);
+
+    assertThat(details.getKilled().toArray(), is(targetUnits.toArray()));
+    // edit mode always sets auto calculated to true
+    assertThat(details.getAutoCalculated(), is(true));
```
DanVanAtta, line 88: nit, the comment on the assertion is essentially what the string arg for assertions is for : )

EG:
```
  assertThat("edit mode sets auto calcuated to true" , details.getAutoCalculated(), is(true));
```

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/SelectMainBattleCasualtiesTest.java
```diff
@@ -0,0 +1,526 @@
...
+          new SelectMainBattleCasualties(selectFunction).apply(delegateBridge, selectCasualties);
+
+      assertThat(details, is(expected));
+
+      verify(selectFunction)
+          .apply(
+              eq(delegateBridge),
+              eq(selectCasualties),
+              (Collection<Unit>) argThat(containsInAnyOrder(targetUnits.toArray())),
+              eq(3));
```
DanVanAtta, line 140: side-note, it does look like we are doing a pretty heavy lift in this test by verifying further implementation calls. If we can have the code under test instead return a result object, then have the caller invoke the select function, it would be more testable, functional, and likely easier to understand when reading the prod code (fewer side effects)

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/SelectMainBattleCasualtiesTest.java
```diff
@@ -0,0 +1,526 @@
...
+
+    @Test
+    void moreHitsThanHitPoints() {
+      final List<Unit> targetUnits = List.of(givenAnyUnit(), givenAnyUnit());
+
+      targetUnits.forEach(
+          unit -> {
+            final UnitAttachment unitAttachment =
+                (UnitAttachment) unit.getType().getAttachment(UNIT_ATTACHMENT_NAME);
+            when(unitAttachment.getHitPoints()).thenReturn(2);
```
DanVanAtta, line 151: It looks like there are more hit points here than hits, is the name of the test correct? "moreHitsThanHitPoints"

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/SelectMainBattleCasualtiesTest.java
```diff
@@ -0,0 +1,526 @@
...
+            when(unit.getHits()).thenReturn(1);
+          });
+
+      final FiringGroup firingGroup = new FiringGroup("", "", List.of(), targetUnits);
```
DanVanAtta, line 155: Any thoughts to create a constructor that removes the no-op arguments? EG:
```
new FiringGroup(targetUnits);
```

DanVanAtta (commented):
Few comments on the tests, please have a look

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/SelectMainBattleCasualtiesTest.java
trevan: The unit has a total of 2 hit points but has already taken one hit.  So the unit has 1 hit point left.  The amount of hits in the test is 3 and there are two units.  So there are more hits than hit points.

I think this is another terminology problem.  Is "hit points" the amount the unit has ignoring any hits they've taken or is it the amount of the user has including the hits they've taken?

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/SelectMainBattleCasualtiesTest.java
trevan: I've added a display name to all of the tests

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/SelectMainBattleCasualtiesTest.java
trevan: Since this constructor is only for tests, I'd rather not.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/SelectMainBattleCasualtiesTest.java
DanVanAtta: Pretty sure - hit points is total hit points regardless of damage. Hit points left or HP remaining is hit points minus damage.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/SelectMainBattleCasualtiesTest.java
DanVanAtta: Fair enough, it's a bit disturbing though that invalid values can be injected. It would seem to put the FiringGroup into an invalid state. It makes one wonder if there is a subset of behavior that could instead be a strategy object and we could test the strategy object directly. Just my 2 cents on seeing this code in test here, I'm not sure if it's practical to fix it or not.

trevan: I've updated the tests to "correctly" create a FiringGroup.  The FiringGroup constructor is now private and the only way to create them is through `FiringGroup.groupBySuicideOnHit`.

DanVanAtta:
A checkstyle cropped up:

> [ant:checkstyle] [WARN] /home/travis/build/triplea-game/triplea/game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/SelectMainBattleCasualtiesTest.java:381: Line is longer than 100 characters (found 103). [LineLength]

DanVanAtta (approved):

PR #7891, opened 2020-10-12
Refactor getUnitPowerAndRollsForNormalBattles/getAaUnitPowerAndRollsForNormalBattles

trevan:
These two methods in `TotalPowerAndTotalRolls` are the two methods that generate `TotalPowerAndTotalRolls` objects.  I refactored both of them so that they now have the same logic inside of `TotalPowerAndTotalRolls`.  The differences between AA and "normal" has been moved into 4 helper classes: `AaDefenseCalculator`, `AaOffenseCalculator`, `NormalDefenseCalculator`, and `NormalOffenseCalculator`.  Each of them contain the logic to calculate the roll and strength for those 4 conditions.  I envision that I can pull the creation of those calculators up the caller chain so that they are passed around instead of the data needed for them (such as `allEnemyUnitsAliveOrWaitingToDie` and `allFriendlyUnitsAliveOrWaitingToDie`).

I also combined `AvailableSupportCalculator`, `SupportBonusCalculator`, and `SupportCalculationResult` into the class `AvailableSupportTracker`.  I named it "Tracker" because it tracks the available support.  As the `TotalPowerAndTotalRolls` objects are built, it uses up support and the `AvailableSupportTracker` handles that logic.

After this is merged, I plan to work on two PRs: one to remove all of the redundant tests in `TotalPowerAndTotalRolls` and the other to push the calculators further up the call chain.  I didn't do the first here because I wanted to make as few changes to the tests as possible to ensure that nothing was broken during the refactoring.  And I didn't do the second because it would add too many changes to this PR.

## Testing

Ran a Hard AI game for a few rounds.


trevan:
A huge portion of this PR is the setup code in the TotalPowerAndTotalRoll tests.  The way things are created has changed in this PR and so I had to change every test in there.  But none of the actual assert lines should have changed.

game-core/src/main/java/games/strategy/triplea/attachments/TerritoryEffectAttachment.java
```diff
@@ -48,8 +49,10 @@ private void setCombatDefenseEffect(final String combatDefenseEffect) throws Gam
     setCombatEffect(combatDefenseEffect, true);
   }
 
-  private void setCombatDefenseEffect(final IntegerMap<UnitType> value) {
+  @VisibleForTesting
```
DanVanAtta, line 52: It does not look like this is yet called from tests, did you mean to commit changes to this file as part of this PR?

game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
```diff
@@ -60,7 +63,8 @@
     private static final long serialVersionUID = -7445551357956238314L;
 
     @Nonnull String name;
-    @Nonnull Integer count;
+
+    @EqualsAndHashCode.Exclude @Nonnull Integer count;
```
DanVanAtta, line 67: What's the rationale for excluding 'count' from equals and hash code?

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaDefenseCalculator.java
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Calculates defense strength and roll for AA/Targeted dice
+ *
+ * <p>This takes into account friendly support, and enemy support
+ */
+@Builder
+@Value
+@Getter(AccessLevel.NONE)
+class AaDefenseCalculator implements OffenseOrDefenseCalculator {
```
DanVanAtta, line 21: We may want to think about the interface naming. 'OffenseOrDefenseCalculator' seems to be a bit too vague, not necessarily single purpose.

IF you saw for example an instance of the interface passed, `OffenseOrDefenseCalculator`, would its API usage and methods be relatively obvious?

DanVanAtta: One suggestion, we may want to avoid the 'calculator' suffix. That tends to implies relatively stateless objects that are functional. Calculators tend to receive in put values and compute output values, they're essentially functions. It looks like the instances each have a pretty decent amount of state behind them.

Perhaps this class could be called `AaDefenseCombatValue` and the interface `CombatValue`

DanVanAtta (commented):
Few comments on class naming

game-core/src/main/java/games/strategy/triplea/attachments/UnitSupportAttachment.java
trevan: The original code used the name as the equality.  And, per pact of steel, "Any support of the same bonusType MUST have the same count.".

Since I'm changing the code to use the object as the hash key instead of just the name of the bonus, I excluded the count to prevent an accidental bug where a map has different counts for the same bonusType.

I checked the maps and there is one map that has different counts for the same bonusType: warcraft war heroes.  The "Wall-Archer" bonusType has a count of 1 and a count of 5.  The game is currently picking the first one in the list, which might be 1 or it might be 5.

game-core/src/main/java/games/strategy/triplea/attachments/TerritoryEffectAttachment.java
trevan: It is being used at https://github.com/triplea-game/triplea/pull/7891/files#diff-bf155c528a8a462543b7b54186929016c76294d8a0d9640f3f2357cf7d9da6b9R68.

game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPurchaseOption.java
```diff
@@ -282,17 +281,14 @@ private double calculateSupportFactor(
     final List<Unit> units = new ArrayList<>(ownedLocalUnits);
     units.addAll(unitsToPlace);
     units.addAll(unitType.create(1, player, true));
-    final SupportCalculationResult supportCalculationResult =
-        AvailableSupportCalculator.getSupport(
+    final AvailableSupportTracker supportCalculationResult =
```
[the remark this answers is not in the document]
trevan: @DanVanAtta, I can do the name change but I'm not sure what you mean about "moving the static functionality back to the calculator class"?  You mean the static constructor back to the calculator class?  Or do you mean some other functionality?  If the former, I don't know why you would want to split the static constructor from its implementation.  If the later, can you clarify?

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AvailableSupports.java
```diff
@@ -0,0 +1,198 @@
...
+    return supportCalculationResult;
+  }
+
+  /**
+   * Constructs an AvailableSupportTracker to track the possible support given by the units
+   *
+   * @param defence are the receiving units defending?
+   * @param allies are the receiving units allied to the giving units?
+   */
+  public static AvailableSupports getSupport(
```
DanVanAtta, line 69: Codacy found an issue: [The method 'getSupport(Collection, Collection, boolean, boolean)' has an NPath complexity of 1203, current threshold is 200](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6326866)

game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPurchaseOption.java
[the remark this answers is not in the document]
trevan: Maybe what you meant was to have a value object that is used by `ProPurchaseOption` and then a "manager" class that takes the value class in its constructor and manages the available support as it is used up?

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/SupportCalculator.java
```diff
@@ -0,0 +1,85 @@
...
+  @Getter Map<UnitSupportAttachment.BonusType, List<UnitSupportAttachment>> supportRules;
+  Map<UnitSupportAttachment, IntegerMap<Unit>> supportUnits;
+  boolean defence;
+  boolean allies;
+
+  /**
+   * @param defence are the receiving units defending?
+   * @param allies are the receiving units allied to the giving units?
+   */
+  public SupportCalculator(
```
DanVanAtta, line 36: Codacy found an issue: [The constructor 'SupportCalculator(Collection, Collection, boolean, boolean)' has an NPath complexity of 1203, current threshold is 200](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6326866)

game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPurchaseOption.java
```diff
@@ -282,17 +281,13 @@ private double calculateSupportFactor(
...
-        supportCalculationResult.getSupportRules();
-    final IntegerMap<UnitSupportAttachment> supportLeft = supportCalculationResult.getSupportLeft();
+    final SupportCalculator availableSupports =
+        new SupportCalculator(units, data.getUnitTypeList().getSupportRules(), defense, true);
 
     double totalSupportFactor = 0;
     for (final UnitSupportAttachment usa : unitSupportAttachments) {
-      for (final List<UnitSupportAttachment> bonusType : supportsAvailable) {
+      for (final List<UnitSupportAttachment> bonusType :
+          availableSupports.getSupportRules().values()) {
```
DanVanAtta, line 290: ` availableSupports.getSupportRules().values()` looks like a violation of the law of delimiter.

What would you think of replacing the above with a method like:
` availableSupports.getUnitSupportAttachments()`

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaDefenseCombatValue.java
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * <p>This takes into account friendly support, and enemy support
+ */
+@Builder
+@Value
+@Getter(AccessLevel.NONE)
+class AaDefenseCombatValue implements CombatValue {
+
+  @NonNull GameData data;
+  @NonNull AvailableSupports friendlySupportTracker;
```
DanVanAtta, line 24: Is 'friendlySupportTracker' a legacy name when the 'supports' was a tracker? WDYT to rename these to 'friendlySupports' and 'enemySupports'

Also worth considering renaming 'enemySupports' to 'enemyNegativeBuffs'. 'enemySupports' sounds like we are tracking both defense and offense and the respective supports for each side. My understanding is the enemySupport here is an "anti-support" being imposed by enemy units (hence, a negative buff).

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaDefenseCombatValue.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  @Override
+  public StrengthOrRollCalculator getStrength() {
+    return new AaDefenseStrength(data, friendlySupportTracker, enemySupportTracker);
+  }
+
+  @Override
+  public boolean isDefending() {
+    return true;
```
DanVanAtta, line 39: It's a bit odd this class is not generic for either offense or defnese. Could you double check if it's possible to make the defense flag, a flag within the implementation.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaDefenseCombatValue.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  @Override
+  public GameData getGameData() {
+    return data;
```
DanVanAtta, line 44: nit, any thoughts to rename `data` to `gameData`, then you can use a lombok `@Getter` to perhaps clean up the code a little bit.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaDefenseCombatValue.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  static class AaDefenseRoll extends StrengthOrRollCalculator {
```
DanVanAtta, line 47: `StrengthOrRollCalculator` IMO is arguably a confusing concept. When is it is a strength calculator, when it is it a roll calculator. If we see an instance of it, do we know what it is or do we have to look at its state and trace back any interactions with the object to understand it. I think this hints to a SRP violation.

If we can break out "StrengthOrRollCalculator" into two interfaces, I believe that would resolve the above questions. Any thoughts/comments on options here @trevan ?

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaDefenseCombatValue.java
```diff
@@ -0,0 +1,95 @@
...
+
+    AaDefenseRoll(final AvailableSupports friendlySupport, final AvailableSupports enemySupport) {
+      super(friendlySupport, enemySupport);
+    }
+
+    @Override
+    public int getValue(final Unit unit) {
+      final RollValue rollValue =
```
DanVanAtta, line 55: nit, unnecessary intermediate variable, can just return with a single line.

It is a bit interesting we don't return a `RollValue` directly in this method, that would make for stronger typing. Any thoughts about doing that?

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaDefenseCombatValue.java
```diff
@@ -0,0 +1,95 @@
...
+          RollValue.of(unit.getUnitAttachment().getMaxAaAttacks())
+              .add(addSupport(unit, friendlySupportTracker))
+              .add(addSupport(unit, enemySupportTracker));
+      return rollValue.getValue();
+    }
+
+    @Override
+    protected Predicate<UnitSupportAttachment> getRuleFilter() {
```
DanVanAtta, line 63: I don't know if a 'rule filter' is cohesive to an API for defense combat value. This seems like it probably should belong to another interface, or a different class. Alternatively, the filter perhaps should accept a collection and then return a filtered result so that the caller is not aware that a predicate is being used. But again, offhand, filtering units seems to be a slightly different operation type compared to combat value calculation. Thoughts on any options around this?

DanVanAtta (commented):
Left some comments for you to consider @trevan , about 30% of the way through reviewing. Will move on to other PRs for now in case any of those can be completed/merged.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaDefenseCombatValue.java
trevan: I think I can get rid of the rule filter completely.  The initial place where unit support attachments is grabbed uses a filter similar to this.  I plan to look into it more once I've done more restructuring of this code.  So, I think this should stay for now and then later rework how the attachments are grabbed initially to not need this filter here.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaDefenseCombatValue.java
trevan: My initial work returned the `RollValue`.  But the only place it is used needs an int and I didn't see the need to have the caller call `getValue` when it can be done here in this `getValue`.  Otherwise, the code would look like:
```
strengthCalculator.getValue(unit).getValue();
```

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaDefenseCombatValue.java
trevan: `StrengthOrRollCalculator` is an abstract class that shares common code that both the roll and strength calculators need.  I could duplicate the class but then both classes would be identical except for their names and that's a different problem.

I could create a `StrengthCalculator` interface and a `RollCalculator` interface and give each of the child classes the corresponding interface but I don't think that will solve your issues with `StrengthOrRollCalculator` since it would still be around.

So, I'm not sure how to resolve your concerns.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaDefenseCombatValue.java
trevan: The name of this class is Aa**Defense**CombatValue.  It can't be generic since it is for defense.  So I'm not sure what I can do.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaDefenseCombatValue.java
trevan: I renamed it to `supportFromFriends` and `supportFromEnemies`.  I didn't want to use `enemyNegativeBuffs` because it might not be a negative buff.  I haven't seen a case where an enemy support rule improved a unit but I have seen a case where a friend support rule hurt a unit.

game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPurchaseOption.java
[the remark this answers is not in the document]
DanVanAtta: To clarify, I was implying to split the class into two. One that has all of the stateful methods, another, "a calculator" that has the static methods. For example, methods like "public static AvailableSupportTracker getSupport" would be in the static utility class. By doing this, each method in the class would use the state of the class, the utility methods on another component. Some judgement needs to be used when methods are related enough to belong together, but that was the suggestion to consider if it helped here.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaDefenseCombatValue.java
DanVanAtta: Good point, I like the rename, it's more clear.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaDefenseCombatValue.java
DanVanAtta: Perhaps rename `StrengthOrRollCalculator` to `StrengthAndRollCalculator`? Or let's make it a more generic term, like `CombatStrengthCalculator`

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaDefenseCombatValue.java
DanVanAtta: If internalizing the flag, then the class would presumably would be merged with the offense combat value class. I somewhat like that as it keeps the AA combat value concerns together, seemingly they are very similar classes.

trevan: This flag is/will be used to generate the `TotalPowerAndTotalStrength` objects.  I might be able to flip the construction around so that `TotalPowerAndTotalStrength` has the defense flag instead of needing to grab it from `CombatValue` but that will have to be a future change.

trevan:
@DanVanAtta, I've pushed renames and removal of intermediate variables.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaDefenseCombatValue.java
trevan: And `AaOffenseStrength` and `AaDefenseStrength` are different.  I'll look at creating a parent class that shares the common roll calculator and just separates the strength calculators.

DanVanAtta: I'm not sure a parent class will be the way to go. Have you considered injecting a strategy object that instead contains the common functionality?

- https://codeburst.io/inheritance-is-evil-stop-using-it-6c4f1caf5117

- https://softwareengineering.stackexchange.com/questions/260343/why-is-inheritance-generally-viewed-as-a-bad-thing-by-oop-proponents

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaCombatValue.java
```diff
@@ -0,0 +1,50 @@
...
+import games.strategy.engine.data.GameData;
+import games.strategy.engine.data.Unit;
+import games.strategy.triplea.attachments.UnitSupportAttachment;
+import java.util.function.Predicate;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NonNull;
+
+/**
+ * Calculates defense strength and roll for AA/Targeted dice
```
DanVanAtta, line 12: Is this still just defense strength? The javadoc and class name do not seem to agree

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaDefenseCombatValue.java
```diff
@@ -0,0 +1,67 @@
...
+    AaDefenseStrength(
+        final GameData gameData,
+        final AvailableSupports friendlySupport,
+        final AvailableSupports enemySupport) {
+      super(friendlySupport, enemySupport);
+      this.gameData = gameData;
+    }
+
+    @Override
+    public int getValue(final Unit unit) {
```
DanVanAtta, line 54: For future consideration, a more specific name than 'getValue' could be of benefit. Specifically to consider, what kind of value are we returning? Is this the roll value, the total strength, or just the die-strength?

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AaOffenseCombatValue.java
```diff
@@ -0,0 +1,68 @@
...
+  AaOffenseCombatValue(
+      final GameData gameData,
+      final AvailableSupports supportFromFriends,
+      final AvailableSupports supportFromEnemies) {
+    super(gameData, supportFromFriends, supportFromEnemies);
+  }
+
+  @Override
+  public StrengthAndRollCalculator getStrength() {
+    return new AaOffenseStrength(gameData, supportFromFriends, supportFromEnemies);
```
DanVanAtta, line 33: There does not seem to be a lot going on in this class implementation. It looks like we could make `AaCombatValue` a concrete class by adding a boolean parameter for offense, defense, and adding a `Supplier<StrengthAndRollCalculator>` and then you could instantiate the base class and not need any inheritance.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AvailableSupports.java
```diff
@@ -0,0 +1,143 @@
...
+    return builder()
+        .supportRules(supportCalculator.getSupportRules())
+        .supportUnits(supportCalculator.getSupportUnits())
+        .build();
+  }
+
+  /** Constructs a filtered version of this */
+  AvailableSupports filter(final Predicate<UnitSupportAttachment> ruleFilter) {
+
+    final Map<UnitSupportAttachment.BonusType, List<UnitSupportAttachment>> supportRules =
```
DanVanAtta, line 63: Any thoughts to use a guava multimap here? That datastructure is equivalent to a `Map<X, Collection<Y>>`

After instantiating the map, you should be able to simplify and write:

```
final var supportRules =     HashMultimap.create();
entry.getValue().stream().filter(ruleFilter).forEach(supportRules::put)
```

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/CombatValue.java
```diff
@@ -0,0 +1,98 @@
...
+
+  StrengthAndRollCalculator getRoll();
+
+  StrengthAndRollCalculator getStrength();
+
+  boolean isDefending();
+
+  GameData getGameData();
+
+  static CombatValue buildMain(
```
DanVanAtta, line 20: `buildMain` is a bit unclear here. It seems implied this is `buildMainCombat`, in which case I think that should be explicitly named to keep it easier on the reader.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/CombatValue.java
```diff
@@ -0,0 +1,98 @@
...
+      final Collection<Unit> allEnemyUnitsAliveOrWaitingToDie,
+      final Collection<Unit> allFriendlyUnitsAliveOrWaitingToDie,
```
DanVanAtta, line 22: side-note, 'waiting to die' is not very meaningful. We should consider renaming that to 'selectedAsCasualties' or simply 'casualties'. In the game, a 'casualty' is relatively well defined to be the units selected as casualties and able to fire back. The removal of casualties is a distinct phase, offense of course removes casualties immediately. 'waiting to die' I think is adding new terminology that is just more stuff to know/remember and is just inventing new domain terminology.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/CombatValue.java
```diff
@@ -0,0 +1,98 @@
...
+      final boolean defending,
```
DanVanAtta, line 23: boolean parameters are often best avoided. I suggest moving these factor methods to a factory class. In that case you can have an internal factory method (private) that has a boolean paramter, but the public methods would be `buildMainOffensiveCombatValue` and `buildMainDefensiveCombatValue` and `buildAaDefensiveCombatValue`

If those methods are in a `CombatValueFactory`, the suffix probably could be dropped, ie:
`buildMainOffensive`, etc..

WDYT?

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/CombatValue.java
```diff
@@ -0,0 +1,98 @@
...
+      final GameData data,
+      final Territory location,
+      final Collection<TerritoryEffect> territoryEffects) {
+
+    // Get all friendly supports
+    final AvailableSupports supportFromFriends =
+        AvailableSupports.getSortedSupport(
+            new SupportCalculator(
+                allFriendlyUnitsAliveOrWaitingToDie,
+                data.getUnitTypeList().getSupportRules(),
```
DanVanAtta, line 33: If we are just using the support rules from the `GameData`, what do you think about passing those support rules directly to the method API? That should help create a more testable interface and drops a dependency on GameData.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/MainDefenseCombatValue.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.Collection;
+import java.util.function.Predicate;
+import lombok.AccessLevel;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.Value;
+
+/**
+ * Calculates defense strength and roll for normal dice
```
DanVanAtta, line 20: Is there a low luck variant of this class? "normal dice" makes me think there is a second defense combat value used for LL.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/MainDefenseCombatValue.java
```diff
@@ -0,0 +1,138 @@
...
+    private boolean isFirstTurnLimitedRoll(final GamePlayer player) {
+      // If player is null, Round > 1, or player has negate rule set: return false
+      return !player.isNull()
+          && gameData.getSequence().getRound() == 1
+          && !isNegateDominatingFirstRoundAttack(player)
+          && isDominatingFirstRoundAttack(gameData.getSequence().getStep().getPlayerId());
+    }
+
+    private boolean isNegateDominatingFirstRoundAttack(final GamePlayer player) {
+      final RulesAttachment ra =
```
DanVanAtta, line 119: nit, please prefer to spell out variable names.

`ra => rulesAttachment`

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/RollValue.java
```diff
@@ -0,0 +1,31 @@
+package games.strategy.triplea.delegate.power.calculator;
+
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.Value;
+
+/**
+ * Holds the value of a dice roll.
+ *
+ * <p>Handles infinite values and correctly limits the value
```
DanVanAtta, line 10: More explanation about the expected behavior here would help. It's not clear what is meant by an infinite roll value, and what 'correctly limiting a value' would mean. EG: what is correct? What would be incorrect? What does it mean to 'limit' a roll value?

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/RollValue.java
```diff
@@ -0,0 +1,31 @@
...
+class RollValue {
+
+  int value;
+  boolean isInfinite;
+
+  static RollValue of(final int value) {
+    return RollValue.of(value, value == -1);
+  }
+
+  RollValue add(final int extraValue) {
```
DanVanAtta, line 23: It looks like we get a no-op when the roll value is infinite. Could that be a silent bug? Why would a caller want to invoke a method for a no-op, it seems that the caller did not realize it was dealing with an infinite roll value. 

I think this leads to some bigger questions. Perhaps `RollValue` is *not* just a simple integer and should not be exposed as that. We have a `getValue()` method as well that returns a magic number to indicate if the roll value is infinite (which is a SRP violation, the number is then context dependent and either it means the roll value, or if it is a magic number then it tell us the roll value is infinite). 

I think this is probably solved by handling the infinite case at the point of the caller. For example, looking at:
```
      return RollValue.of(unit.getUnitAttachment().getDefenseRolls(unit.getOwner()))
          .add(addSupport(unit, friendlySupportTracker))
          .add(addSupport(unit, enemySupportTracker))
          .getValue();
```

If the roll value is infinite:
- are we wasting support?
- is the roll value being returned a -1, or some other number?

It's not really clear for the above two questions.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/StrengthAndRollCalculator.java
```diff
@@ -0,0 +1,43 @@
...
+  // Keep track of the units that are providing support and whom they are providing it to
+  @Getter private final Map<Unit, IntegerMap<Unit>> supportGiven = new HashMap<>();
+
+  StrengthAndRollCalculator(
+      final AvailableSupports friendlySupport, final AvailableSupports enemySupport) {
+    this.friendlySupportTracker = friendlySupport.filter(getRuleFilter());
+    this.enemySupportTracker = enemySupport.filter(getRuleFilter());
+  }
+
+  abstract int getValue(Unit unit);
```
DanVanAtta, line 25: It does not seem to be clear whether this is roll value, or strength value. Would it make sense to have two abstract methods, one for each? This way any caller can explicitly ask for bonus rolls, or bonus strength without having to infer the value returning based on the instance of the object.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/StrengthAndRollCalculator.java
```diff
@@ -0,0 +1,43 @@
...
+
+  protected abstract Predicate<UnitSupportAttachment> getRuleFilter();
+
+  protected int addSupport(final Unit unit, final AvailableSupports supportTracker) {
```
DanVanAtta, line 29: It's a bit difficult to understand the API of this method and what is being returned exactly. Is this a functional method, or are we modifying state. If modifying state, what does the return value represent?

I think it would be easier for future maintainers to have this explicitly stated rather than being left to only reverse engineer the code.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/StrengthValue.java
```diff
@@ -0,0 +1,26 @@
+package games.strategy.triplea.delegate.power.calculator;
+
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.Value;
+
+/**
+ * Holds the value of a dice strength
```
DanVanAtta, line 8: nit, a die does not have strength. 
Rephrasing suggestion: ` Holds the value of units dice roll strength`

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/StrengthValue.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * <p>Correctly limits the value
```
DanVanAtta, line 10: Could you clarify what is meant by:  "Correctly limits the value"

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/SupportCalculator.java
```diff
@@ -0,0 +1,89 @@
...
+import lombok.Getter;
+import lombok.Value;
+import org.triplea.java.collections.CollectionUtils;
+import org.triplea.java.collections.IntegerMap;
+
+/** Calculates how much support units can give */
+@Value
+@Getter(value = AccessLevel.PACKAGE)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class SupportCalculator {
```
DanVanAtta, line 25: Is the code in this class all just moved code? Anything new or modified here?

DanVanAtta (commented):
Finished reviewing the non-test code, will need to come back to review the test code before a first pass review is complete.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AvailableSupports.java
trevan: I think this method might go away similar to the `Predicate` method that you commented on earlier.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/CombatValue.java
trevan: `WaitingToDie` is the original name.  That terminology in a lot of places in the battle code.  So I'd rather not rename it just here.  It would need to be renamed in all the places as a separate PR.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/CombatValue.java
trevan: I'd love to replace all the usage of the boolean `defending` with the BattleState.Side enums.  It is used in a lot of places.  If you are ok with that, I could do that as a separate PR.

As for the factory change, let's wait on that.  I have a branch dependent on this where the construction of the `CombatValue` objects is moved up the call tree.  I think that might be a better PR to discuss how they are constructed.  It might make it more obvious on their construction.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/CombatValue.java
trevan: Yup, that would be part of the `Predicate` change you commented on earlier.  The `getSupportRules` and `getAaSupportRules` uses a similar `Predicate` inside of them and I'd like to unify the logic in some manner and simplify it.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/MainDefenseCombatValue.java
trevan: roll and strength are the same for LL and non-LL.  So, no there isn't a low luck variant.  This is for "non-AA" dice but I didn't want to say that.  Any ideas on a better name?

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/MainDefenseCombatValue.java
trevan: Ok.  This was moved code so I didn't make any changes to it.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/RollValue.java
trevan: "-1" is how a map designer designates that a roll is infinite.  And if the map designer gives an AAroll bonus to a unit with infinite roll, than the bonus will be ignored since it doesn't make any sense.

And the caller shouldn't care if it is infinite or not infinite.  It is taking the map support rules as designed by the map designer and adding it to the roll.  Now, maybe it should throw an exception so that the map designer knows that they are trying to give an AAroll bonus to the unit but the game engine doesn't have a way to notify just the map designer.

As for passing around the RollValue instead of a magic number "-1", that will be a fairly large change since all the callers and usages are expecting ints.  It also shares the same int structures with non `RollValues`.  I'd rather do that as a separate PR.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/StrengthValue.java
trevan: "roll" has another meaning already that means "how many rolls a unit is allowed".  I wouldn't want to use "roll" here since it is confusing terminology.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/SupportCalculator.java
trevan: The only change is supportRules is now `Map<UnitSupportAttachment.BonusType, List<UnitSupportAttachments>>` instead of `List<List<UnitSupportAttachments>>`

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/CombatValue.java
trevan: `CombatValue.buildMainCombatValue()` feels a little redundant so that's why I just left it as `CombatValue.buildMain` but I'll change the name.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/StrengthAndRollCalculator.java
trevan: But calling `getRolls` from a strength calculator is incorrect.  So the caller would still need to know what the object instance is for them to know which method to call.

How about I switch `StrengthAndRollCalculator` to be composed inside of `StrengthCalculator` and `RollCalculator`?  Then each interface has its own clear method but the logic is still in one place.

trevan:
@DanVanAtta , I reworked the `StrengthAndRollCalculator`.  There is now a `StrengthCalculator` interface and a `RollCalculator` interface.

I also reworked the `AaCombatValue` to instead of being a parent abstract class, the shared roll code is moved to a common class `AaRoll` and the `AaOffenseCombatValue` and `AaDefenseCombatValue` use that class instead of their own versions.

And I've updated names and comments.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/SupportCalculator.java
```diff
@@ -0,0 +1,89 @@
...
+  Map<UnitSupportAttachment.BonusType, List<UnitSupportAttachment>> supportRules;
+  Map<UnitSupportAttachment, IntegerMap<Unit>> supportUnits;
+  boolean defence;
+  boolean allies;
+
+  /**
+   * @param defence are the receiving units defending?
+   * @param allies are the receiving units allied to the giving units?
+   */
+  public SupportCalculator(
```
DanVanAtta, line 36: Codacy found an issue: [The constructor 'SupportCalculator(Collection, Collection, boolean, boolean)' has an NPath complexity of 1203, current threshold is 200](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6326866)

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/StrengthAndRollCalculator.java
DanVanAtta: Can you explain that option a bit more? I don't think I quite follow the suggestion.

A method like `StrengthAndRollCalculator.getRolls()` seems to make some sense, I would also expect a `StrengthAndRollCAlculator.getStrength()` method to be present too. Splitting the two could make sense, though the data & concept are somewhat coupled and were combined from what was previously just a `Tuple<Int, Int>`. 

If you're willing to follow up, we can extract this to an issue and follow-up post-merge if that would make review and forward progress easier.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/StrengthValue.java
DanVanAtta: Good point. We should fix "dice strength", a die does not have strength.

Perhaps a "units dice strength" (?)

Or call it the "ToHitStrength". Or, since we've defined power and strength, we perhaps could just call it a units "strength" and it should then be clear as we have defined strength.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/SupportCalculator.java
DanVanAtta: Thanks :+1: 

Any thoughts to use multimap here?

Guava's multimap is a data structure to represent `Map<X, List<Y>>`

DanVanAtta:
Thank you for the updates @trevan to this branch. I hope to give this PR a last review pass tomorrow and we'll get this merged shortly after.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/SupportCalculator.java
trevan: Good idea.  I didn't want to make too big of a change to this code, though.  I want to get better tests around it and then try and make the code easier to understand.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/StrengthAndRollCalculator.java
trevan: Take a look at the new code.  There is now a `StrengthCalculator` and a `RollCalculator` interface.  The `StrengthCalculator` has `getStrength` and the `RollCalculator` has `getRoll`.  The implementing classes contain a reference to the `StrengthAndRollCalculator` (renamed as `UsedSupportTracker`) and use the public `giveSupport` and `getSupportGiven` methods to get the common logic.

trevan:
@DanVanAtta , I've pushed some comment and rename changes.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/SupportCalculator.java
DanVanAtta, line 36: Codacy found an issue: [The constructor 'SupportCalculator(Collection, Collection, boolean, boolean)' has an NPath complexity of 1203, current threshold is 200](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6326866)

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/MainDefenseCombatValue.java
DanVanAtta: Perhaps: `Calculates defense strength and roll bonuses`

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/StrengthAndRollCalculator.java
```diff
@@ -13,8 +13,18 @@
...
-    final IntegerMap<Unit> supportGivenToUnit = supportTracker.giveSupportToUnit(unit);
+  /**
+   * Gives support to the unit and removes it from the availableSupport
+   *
+   * <p>Each time this is called, the amount of available support will decrease equal to the amount
+   * returned.
+   *
+   * @param unit Unit to have support given to it
+   * @param availableSupport Keeps state of what support is still available
+   * @return the amount of support given
```
DanVanAtta: might be worthwhile clarifying if the support is a strength bonus or a roll bonus. Is there a way to know which one will be returned?

DanVanAtta:
@trevan looking quite good, a checkstyle cropped up.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/StrengthAndRollCalculator.java
trevan: There isn't a way to determine if it is a strength or roll bonus.  That is controlled by the list of support rules in AvailableSupports.

I just removed this file, though.  The logic that it had looked like it fit better in AvailableSupports.  So now the `supportGiven` map is managed in `AvailableSupports` and the `CombatValue` classes will merge the two `supportGiven` maps from the friendly and enemy supports.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/MainDefenseCombatValue.java
trevan: Only problem I have with that is that these classes don't just calculate the bonus.  They calculate the final strength and roll value.  That includes adding the bonuses as well as limiting the value within the dice sides.  I've changed the text to say `Calculates defense strength and roll for non-AA dice` for now.

trevan:
@DanVanAtta , I fixed the checkstyle.

PR #7899, opened 2020-10-14
Allow AA to target all infrastructure

### LINKED ISSUE #7833: Bring back the ability to target infrastructures with AA
**Can you please bring back the ability to target infrastructures with AA?** That is a change that I guess happened accidentaly in a previous refactor like this one.

For example, an air unit getting shot in a territory with an aaGun, no matter if the air unit is an infrastructure.

_Originally posted by @Cernelius in https://github.com/triplea-game/triplea/pull/7823#issuecomment-703037942_

trevan:
Fixes #7833

This modifies `Matches.unitTypeCanBeInBattle` to accept the list of attackers and checks if any of those attackers can hit the unit using `targetsAa`.  I've also added tests around the infrastructure part of that method.

This should allow any infrastructure units that are listed in `targetsAa` to be part of the battle.  They'll show up in the Battle UI and can be taken as casualties.

## Testing

I used the save https://github.com/triplea-game/triplea/pull/7823#issuecomment-703055893 to verify that the air_transports would be taken as casualties.

I also ran a hard ai for a few rounds.

## Release Note

CHANGE|Removed from "infrastructure" units the ability to be immune to targeted (AA-like) fire and made infrastructures targetable also beyond the first round of combat (in accordance to the fact that TripleA allows infrastructures to remain in combat beyond the first round of it).

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
@@ -653,7 +653,7 @@ public final class Matches {
   }
 
   /** Checks if the unit type can be hit with AA fire by one of the firingUnits */
-  private static Predicate<UnitType> unitTypeCanByHitByAaFire(
+  private static Predicate<UnitType> unitTypeCanBeHitByAaFire(
       final Collection<UnitType> firingUnits, final GameData gameData, final int battleRound) {
     // make sure the aa firing units are valid for combat and during this round
     final Collection<UnitType> aaFiringUnits =
@@ -2345,15 +2345,14 @@ public final class Matches {
       final boolean doNotIncludeBombardingSeaUnits,
       final Collection<UnitType> firingUnits) {
 
+    // remove infrastructure units unless it can support or fight
+    // or it is AA that can fire this round
+    // or it can be shot at by AA
     final PredicateBuilder<UnitType> canBeInBattleBuilder =
-        // remove infrastructure units
         PredicateBuilder.of(unitTypeIsInfrastructure().negate())
-            // unless it can support or fight
             .or(unitTypeIsSupporterOrHasCombatAbility(attack, player))
-            // or it is AA that can fire this round
             .or(unitTypeIsAaForCombatOnly().and(unitTypeIsAaThatCanFireOnRound(battleRound)))
-            // or it can be shot at by AA
-            .or(unitTypeCanByHitByAaFire(firingUnits, player.getData(), battleRound));
+            .or(unitTypeCanBeHitByAaFire(firingUnits, player.getData(), battleRound));
 
     if (attack) {
       if (!includeAttackersThatCanNotMove) {
--- a/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
@@ -652,6 +652,23 @@ public final class Matches {
     };
   }
 
+  /** Checks if the unit type can be hit with AA fire by one of the firingUnits */
+  private static Predicate<UnitType> unitTypeCanByHitByAaFire(
+      final Collection<UnitType> firingUnits, final GameData gameData, final int battleRound) {
+    // make sure the aa firing units are valid for combat and during this round
+    final Collection<UnitType> aaFiringUnits =
+        CollectionUtils.getMatches(
+            firingUnits,
+            unitTypeIsAaForCombatOnly().and(unitTypeIsAaThatCanFireOnRound(battleRound)));
+    return unitType ->
+        aaFiringUnits.stream()
+            .anyMatch(
+                type -> {
+                  final UnitAttachment attachment = UnitAttachment.get(type);
+                  return attachment.getTargetsAa(gameData).contains(unitType);
+                });
+  }
+
   public static Predicate<Unit> unitIsAaOfTypeAa(final String typeAa) {
     return obj -> UnitAttachment.get(obj.getType()).getTypeAa().matches(typeAa);
   }
@@ -2287,7 +2304,17 @@ public final class Matches {
       final int battleRound,
       final boolean doNotIncludeBombardingSeaUnits) {
     return unitCanBeInBattle(
-        attack, isLandBattle, battleRound, true, doNotIncludeBombardingSeaUnits);
+        attack, isLandBattle, battleRound, doNotIncludeBombardingSeaUnits, List.of());
+  }
+
+  public static Predicate<Unit> unitCanBeInBattle(
+      final boolean attack,
+      final boolean isLandBattle,
+      final int battleRound,
+      final boolean doNotIncludeBombardingSeaUnits,
+      final Collection<UnitType> firingUnits) {
+    return unitCanBeInBattle(
+        attack, isLandBattle, battleRound, true, doNotIncludeBombardingSeaUnits, firingUnits);
   }
 
   public static Predicate<Unit> unitCanBeInBattle(
@@ -2295,7 +2322,8 @@ public final class Matches {
       final boolean isLandBattle,
       final int battleRound,
       final boolean includeAttackersThatCanNotMove,
-      final boolean doNotIncludeBombardingSeaUnits) {
+      final boolean doNotIncludeBombardingSeaUnits,
+      final Collection<UnitType> firingUnits) {
     return unit ->
         unitTypeCanBeInBattle(
                 attack,
@@ -2303,7 +2331,8 @@ public final class Matches {
                 unit.getOwner(),
                 battleRound,
                 includeAttackersThatCanNotMove,
-                doNotIncludeBombardingSeaUnits)
+                doNotIncludeBombardingSeaUnits,
+                firingUnits)
             .test(unit.getType());
   }
 
@@ -2313,14 +2342,18 @@ public final class Matches {
       final GamePlayer player,
       final int battleRound,
       final boolean includeAttackersThatCanNotMove,
-      final boolean doNotIncludeBombardingSeaUnits) {
+      final boolean doNotIncludeBombardingSeaUnits,
+      final Collection<UnitType> firingUnits) {
 
-    // Filter out anything like factories, or units that have no combat ability AND cannot be taken
-    // casualty
     final PredicateBuilder<UnitType> canBeInBattleBuilder =
+        // remove infrastructure units
         PredicateBuilder.of(unitTypeIsInfrastructure().negate())
+            // unless it can support or fight
             .or(unitTypeIsSupporterOrHasCombatAbility(attack, player))
-            .or(unitTypeIsAaForCombatOnly().and(unitTypeIsAaThatCanFireOnRound(battleRound)));
+            // or it is AA that can fire this round
+            .or(unitTypeIsAaForCombatOnly().and(unitTypeIsAaThatCanFireOnRound(battleRound)))
+            // or it can be shot at by AA
+            .or(unitTypeCanByHitByAaFire(firingUnits, player.getData(), battleRound));
 
     if (attack) {
       if (!includeAttackersThatCanNotMove) {
--- a/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
@@ -1021,7 +1021,7 @@ class BattleCalculatorPanel extends JPanel {
                   Matches.unitIsOwnedBy(getDefender())
                       .and(
                           Matches.unitCanBeInBattle(
-                              true, isLand(), 1, hasMaxRounds(isLand(), data), true)));
+                              true, isLand(), 1, hasMaxRounds(isLand(), data), true, List.of())));
           final List<Unit> newDefenders =
               CollectionUtils.getMatches(
                   attackingUnitsPanel.getUnits(),
@@ -1336,7 +1336,8 @@ class BattleCalculatorPanel extends JPanel {
         getAttacker(),
         CollectionUtils.getMatches(
             units,
-            Matches.unitCanBeInBattle(true, isLand(), 1, hasMaxRounds(isLand(), data), false)),
+            Matches.unitCanBeInBattle(
+                true, isLand(), 1, hasMaxRounds(isLand(), data), false, List.of())),
         isLand());
   }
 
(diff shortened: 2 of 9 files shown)


game-core/src/test/java/games/strategy/triplea/delegate/MatchesTest.java
```diff
@@ -254,4 +267,180 @@ void shouldNotMatchWhenSeaTerritoryIsUnownedAndDoesNotHaveTerritoryAttachment()
...
+
+    @Test
+    void infrastructureThatGivesAnyTypeOfSupportCanBeInBattle() {
+      final UnitType unitType = new UnitType("infrastructure", gameData);
+      final UnitAttachment unitAttachment =
+          new UnitAttachment("infrastructure", unitType, gameData);
+      unitAttachment.setIsInfrastructure(true);
+      final UnitSupportAttachment unitSupportAttachment =
+          new UnitSupportAttachment(
+              Constants.SUPPORT_ATTACHMENT_PREFIX + "support", unitType, gameData);
```
DanVanAtta, line 375: Codacy found an issue: [Unnecessary use of fully qualified name 'Constants.SUPPORT_ATTACHMENT_PREFIX' due to existing static import 'games.strategy.triplea.Constants.SUPPORT_ATTACHMENT_PREFIX'](https://app.codacy.com/gh/triplea-game/triplea/pullRequest?prid=6335347)

DanVanAtta:
The release note could use some more detail, maybe an example would clarify it.

trevan:
@Cernelius , what do you think the release note should be for this?

game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
```diff
@@ -652,6 +652,23 @@ private Matches() {}
     };
   }
 
+  /** Checks if the unit type can be hit with AA fire by one of the firingUnits */
+  private static Predicate<UnitType> unitTypeCanByHitByAaFire(
```
DanVanAtta, line 656: Typo: `unitTypeCanByHitByAaFire` -> "BeHitBy.."

game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
```diff
@@ -652,6 +652,23 @@ private Matches() {}
...
+      final Collection<UnitType> firingUnits, final GameData gameData, final int battleRound) {
```
DanVanAtta, line 657: WDYT to pass the data from `GameData` to this method rather than passing `GameData` to it?

`GameData` has a problem where it's passed everywhere and used by everything. That makes decoupling quite difficult, for example, testing involving with a game data almost always means doing a full blown integration test and loading an XML (which is extremely limiting as we may want to test perhaps 12 permutations of one element of game data for just one set of tests).

Anyways, the less dependencies we can have on game data for now-  the better.

game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
```diff
@@ -2313,14 +2342,18 @@ private Matches() {}
...
+      final boolean doNotIncludeBombardingSeaUnits,
+      final Collection<UnitType> firingUnits) {
 
-    // Filter out anything like factories, or units that have no combat ability AND cannot be taken
-    // casualty
     final PredicateBuilder<UnitType> canBeInBattleBuilder =
+        // remove infrastructure units
         PredicateBuilder.of(unitTypeIsInfrastructure().negate())
+            // unless it can support or fight
             .or(unitTypeIsSupporterOrHasCombatAbility(attack, player))
```
DanVanAtta, line 2352: Intermediate variables can perhaps help reduce the commentary. For example:
```
var isAaThatCanFireThisRound = unitType -> unitTypeIsAaForCombatOnly().and(unitTypeIsAaThatCanFireOnRound(battleRound);
```

WDYT of using that technique to avoid the inline commentary? I think that is only needed for the more complex conditions, the simpler ones, the code might be able to just speak for itself.

DanVanAtta: For a long series of statements, having one comment to describe the general outcome and what is generally happening can be a good thing. Make it easier for a reader to know what to expect and makes it easier to see how each statement fits in to the overall outcome. Commenting each line though is more questionable and can speak to inarticulate code (which can be fixed by improving naming, intermediate variables, private helper methods, re-organizing methods).

game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
```diff
@@ -1336,7 +1336,8 @@ private void setAttackingUnits(final List<Unit> initialUnits) {
         getAttacker(),
         CollectionUtils.getMatches(
             units,
-            Matches.unitCanBeInBattle(true, isLand(), 1, hasMaxRounds(isLand(), data), false)),
+            Matches.unitCanBeInBattle(
+                true, isLand(), 1, hasMaxRounds(isLand(), data), false, List.of())),
```
DanVanAtta, line 1340: side-note, not for thsi PR - this API could use work. From the looks of it, most of the parameters are not used. From a caller perspective, it's kind meaningless of an API call, the 'trues', 'falses', redundant call to `isLand()` are certainly minor problems.

I comment on this as we should be careful to avoid making such problems worse when we can, but here I think the horse has already left the barn.

game-core/src/test/java/games/strategy/triplea/delegate/MatchesTest.java
```diff
@@ -254,4 +266,179 @@ void shouldNotMatchWhenSeaTerritoryIsUnownedAndDoesNotHaveTerritoryAttachment()
...
+  @Nested
+  @ExtendWith(MockitoExtension.class)
+  final class UnitCanBeInBattle {
+
+    @Mock GamePlayer player;
+    GameData gameData;
+
+    @BeforeEach
+    void setupGameData() {
+      gameData = givenGameData().build();
```
DanVanAtta, line 279: `givenGameData()` is not as descriptive as one would hope. What is given about this game data? Can we say or require anything to be true about this game data? Though, I suspect the instantiation of any game data is something of a hack for game data being *very* difficult to mock (God class code-smell, we use it everywhere for everything.

I'd recommend renaming this method to `givenAnyGameData()` perhaps to indicate that we require nothing other than an instantiated game data, otherwise any requirements of the data would either be specified as parameters or baked into the game data name. For example, `givenGameDataWithUnits(...)`

DanVanAtta: Looking at this further, perhaps calling it `givenMockGameData()`.

Though, ideally we would not need the object and any data from the game data would individually be passed to this class. In other words, game data is acting as a tuple with other objects. Rather than passing the tuple and grabbing the objects from the tuple, instead we can pass the individual objects from the tuple to where they are needed.

game-core/src/test/java/games/strategy/triplea/delegate/MatchesTest.java
```diff
@@ -254,4 +266,179 @@ void shouldNotMatchWhenSeaTerritoryIsUnownedAndDoesNotHaveTerritoryAttachment()
...
+    }
+
+    @Test
+    void infrastructureShouldNormallyNotBeInBattle() {
+      when(gameData.getDiceSides()).thenReturn(6);
```
DanVanAtta, line 284: Does this method need the other mocked elements of `gameData`, or do we mostly only need for the dice sides?

I'd recommend to try and fully mock objects within the test method where reasonable, at least within a test so the test is more self contained. For example if another test required slightly different mocking of a game data, we'd suddenly have a conflict between that test and any other that depended on the mock.

More fundamentally it probably is better to break up the data needed from game data, use only that data, then you can pass those data objects as test parameters. For example, rather than passing a game data to get dice sides to another method, instead pass the dice sides value directly.

game-core/src/test/java/games/strategy/triplea/delegate/MatchesTest.java
```diff
@@ -254,4 +266,179 @@ void shouldNotMatchWhenSeaTerritoryIsUnownedAndDoesNotHaveTerritoryAttachment()
...
+      final UnitType unitType = new UnitType("infrastructure", gameData);
+      final UnitAttachment unitAttachment =
+          new UnitAttachment("infrastructure", unitType, gameData);
+      unitAttachment.setIsInfrastructure(true);
+      unitType.addAttachment(UNIT_ATTACHMENT_NAME, unitAttachment);
+      final Unit unit = unitType.create(1, player, true).get(0);
+
+      assertThat(
+          "An infrastructure unit normally can not be in battle",
+          Matches.unitCanBeInBattle(true, true, 1, false, List.of()).test(unit),
```
DanVanAtta, line 294: A parameter object I think would help a lot here, this parameter list is not very meaningful:
```true, true, 1, false, List.of()```

I wonder a bit as well why the parameter list is not `Matches.unitTypeCanBeInBattleForRound( < UnitType > , <battleRound:Int> )`

File: game-core/src/test/java/games/strategy/triplea/delegate/MatchesTest.java [hunk 9]
```diff
@@ -254,4 +266,179 @@ void shouldNotMatchWhenSeaTerritoryIsUnownedAndDoesNotHaveTerritoryAttachment()
...
+      final UnitType unitType = new UnitType("infrastructure", gameData);
+      final UnitAttachment unitAttachment =
+          new UnitAttachment("infrastructure", unitType, gameData);
+      unitAttachment.setIsInfrastructure(true);
+      unitAttachment.setDefense(1);
+      unitType.addAttachment(UNIT_ATTACHMENT_NAME, unitAttachment);
+      final Unit unit = unitType.create(1, player, true).get(0);
+
+      assertThat(
+          "An infrastructure unit with defense can not be in battle when it is attacking",
```
DanVanAtta, line 361: The assert comments are good and pretty reasonable setup  :+1:

Cernelius:
@trevan Release note for regular users and map-makers? Maybe:

Removed from "infrastructure" units the ability to be immune to targeted (AA-like) fire and made infrastructure targetable also beyond the first round of combat (in accordance to the fact that TripleA allows infrastructures to remain in combat beyond the first round of it).

---

I don't know (I haven't tested) if infrastructure units have been still targetable by fly-over AA fire. In case, you may want to clarify here we are talking only about AA fire in battle, not AA fly-overs.

game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
trevan: That would require changing `UnitAttachment#getTargetsAa` since this method just passes `gameData` to that method.  So for this PR, I don't think that is ok.

game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
trevan: I personally don't see the intermediate variables as being helpful in explaining what is happening.  I'll just move all the comments to above the definition of the `canBeInBattleBuilder` variable.

game-core/src/test/java/games/strategy/triplea/delegate/MatchesTest.java
trevan: I guess `mockGameDataBuilder()` would be a better name for it since that is what it does.  It creates a mock of `GameData` with some basic setup and then allows you to add functionality using a fluent syntax.  It isn't quite a "builder".

I don't think the rename should be part of this PR, though.  It is used in quite a few of my unittests now.  It is how I create mocked GameData objects.

game-core/src/test/java/games/strategy/triplea/delegate/MatchesTest.java
trevan: To run, the test needs other aspects of gameData.  That's because `Matches#unitTypeCanBeInBattle` will call `Matches#unitTypeIsSupporterOrHasCombatAbility` which in turn calls `UnitAttachment#getAttack` which will then call `TechAbilityAttachment#getAttackBonus` with the gameData that is on the UnitAttachment.  And it is this gameData that it will then look for tech abilities as well as check the dice sides.

game-core/src/test/java/games/strategy/triplea/delegate/MatchesTest.java
trevan: Agreed.  `Matches.unitCanBeInBattle` is a complicated matcher that is used in a lot of places.  You'll notice there are 4 overloaded versions of it with different number of parameters.  I first thought about trying to isolate just what I want but it changed too many things that I wasn't sure of.

game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
DanVanAtta: Fair enough. When it comes to "what is the code doing" commentary, reducing that can help avoid coupling (meaning the code can be changed and the comments do not become out of date) and sometimes such commentary is being used as a bandaid for inexpressive code. There is a right balance though, some "what is the code doing"  commentary can really help set expectations of what a person is about to read and makes it easier to understand (easier to understand => faster to read => quicker to modify => and more comprehensive leads to fewer bugs)

game-core/src/test/java/games/strategy/triplea/delegate/MatchesTest.java
DanVanAtta: Okay, I think we are then already started down a path where we have a 'god mock' of a 'god class'. Let's watch out for this going forward. I think the thing for us to do more of us avoid passing game data and instead pass the data that is needed instead. For example, if units did not have a reference to game data, it would be far easier to instantiate them and use in testing.

DanVanAtta (approved):

game-core/src/test/java/games/strategy/triplea/delegate/MatchesTest.java
trevan: I did a quick test where I removed the `extend` in `Unit` so that it doesn't have a `GameData`.  There were not as many compile errors as I would expect.

Do you have a problem with giving `Unit` a reference to the `GameProperties`?  That is what most of the errors are related to.

DanVanAtta: That is on the right path. Where I'd like to get is that we have one object called a "MapData" or something like that, that is just the data parsed from a map. "MapData" would then be an initial state. A "GameData" would be the final state of the game. A "GameData" would be a "MapData" plus N changes, if a game is new, the number of changes is zero and "GameData" and "MapData" would be equivalent.

With that, we need to think a bit about how the core data model is going to be structured.

In the short term, removing GameData reference for GameProperties sounds like the right path.

Though, does that work for serialization? IIRC the game data reference is to mark/enable the object for serialization into save games.

PR #7900, opened 2020-10-14
Add tests for FiringGroup

trevan:
This is part of #7823

I noticed that I had missed this test when I created the FiringGroup PR.  It also helped find a bug that snuck in during some of the reworking of the code during its PR, so that is fixed as part of this.

### DIFF
--- a/game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroupTest.java
+++ b/game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroupTest.java
@@ -1,0 +1,187 @@
+package games.strategy.triplea.delegate.battle.steps.fire;
+
+import static games.strategy.triplea.Constants.UNIT_ATTACHMENT_NAME;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.is;
+import static org.mockito.Mockito.mock;
+
+import games.strategy.engine.data.GameData;
+import games.strategy.engine.data.GamePlayer;
+import games.strategy.engine.data.Unit;
+import games.strategy.engine.data.UnitType;
+import games.strategy.triplea.attachments.UnitAttachment;
+import java.util.Comparator;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class FiringGroupTest {
+
+  @Mock GameData gameData;
+  @Mock GamePlayer player;
+
+  @Test
+  public void onlyOneGroupIfNoSuicideOnHit() {
+    final List<Unit> units =
+        List.of(
+            givenUnitWithSuicideOnHit("type1", false), givenUnitWithSuicideOnHit("type2", false));
+    final List<FiringGroup> groups =
+        FiringGroup.groupBySuicideOnHit("test", units, List.of(mock(Unit.class)));
+    assertThat("There should only be one group", groups, hasSize(1));
+    assertThat("The group name should have no prefix", groups.get(0).getDisplayName(), is("test"));
+    assertThat(
+        "The group should have all of the units",
+        groups.get(0).getFiringUnits().toArray(),
+        is(units.toArray()));
+    assertThat("The group should not be suicide on hit", groups.get(0).isSuicideOnHit(), is(false));
+  }
+
+  private Unit givenUnitWithSuicideOnHit(final String typeName, final boolean suicideOnHit) {
+    return givenUnitWithSuicideOnHit(givenUnitTypeWithSuicideOnHit(typeName, suicideOnHit));
+  }
+
+  private Unit givenUnitWithSuicideOnHit(final UnitType unitType) {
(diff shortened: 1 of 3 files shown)


game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroupTest.java
```diff
@@ -0,0 +1,187 @@
...
+  @Mock GamePlayer player;
+
+  @Test
+  public void onlyOneGroupIfNoSuicideOnHit() {
+    final List<Unit> units =
+        List.of(
+            givenUnitWithSuicideOnHit("type1", false), givenUnitWithSuicideOnHit("type2", false));
+    final List<FiringGroup> groups =
+        FiringGroup.groupBySuicideOnHit("test", units, List.of(mock(Unit.class)));
+    assertThat("There should only be one group", groups, hasSize(1));
```
DanVanAtta, line 34: nit, the assertion `hasSize(1)` already says there should be one group, what is not said though is why?

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroupTest.java
```diff
@@ -0,0 +1,187 @@
...
+class FiringGroupTest {
+
+  @Mock GameData gameData;
+  @Mock GamePlayer player;
+
+  @Test
+  public void onlyOneGroupIfNoSuicideOnHit() {
+    final List<Unit> units =
+        List.of(
+            givenUnitWithSuicideOnHit("type1", false), givenUnitWithSuicideOnHit("type2", false));
```
DanVanAtta, line 31: Nit, the method name " givenUnitWithSuicideOnHit("type1", false)" is a bit akward with those parameters. I would have expected it to always return a unit with suicide on hit. Boolean parameters are generally evil, so one thing we can do is to create two methods for both cases and that would help with the naming problem. Then it would be:
```
givenUnitWithTypeNameAndSuicideOnHit("type1")
givenUnitWithTypeNameWithoutSuicideOnHit("type2")
```

Though, TBH we would do well if we could create a value object out of this and avoid mocks. That makes me think that the API we are testing should accept the property values rather than the actual unit type object.

If we keep the boolean parameter, then perhaps a more descriptive method name would be:
`givenUnitWithTypeAndHasSuicideOnHit(...)`

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroupTest.java
```diff
@@ -0,0 +1,187 @@
...
+    assertThat("There should only be one group", groups, hasSize(1));
+    assertThat("The group name should have no prefix", groups.get(0).getDisplayName(), is("test"));
+    assertThat(
+        "The group should have all of the units",
+        groups.get(0).getFiringUnits().toArray(),
+        is(units.toArray()));
+    assertThat("The group should not be suicide on hit", groups.get(0).isSuicideOnHit(), is(false));
+  }
+
+  private Unit givenUnitWithSuicideOnHit(final String typeName, final boolean suicideOnHit) {
```
DanVanAtta, line 43: These two lines are pretty overwhelming to read, lots of "givenUnitWithSuicideOnHit", IMO the method names blur together and it becomes hard to tell exactly which API overload is being invoked where.

An intermediate variable could help, eg:

```
final var unitType = givenUnitTypeWithSuicideOnHit(typeName, suicideOnHit);
return givenUnitWithSuicideOnHit(unitType);
```

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroupTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+  @Test
+  public void onlyOneGroupIfSameTypeAndSuicide() {
+    final UnitType unitType = givenUnitTypeWithSuicideOnHit("type", true);
+    final List<Unit> units =
+        List.of(givenUnitWithSuicideOnHit(unitType), givenUnitWithSuicideOnHit(unitType));
+    final List<FiringGroup> groups =
+        FiringGroup.groupBySuicideOnHit("test", units, List.of(mock(Unit.class)));
+    assertThat("There should only be one group", groups, hasSize(1));
+    assertThat("The group name should have no prefix", groups.get(0).getDisplayName(), is("test"));
```
DanVanAtta, line 68: Where does the "test" value come from? Can we add that as a parameter to the data setup within this method so that it can be easily correlated when someone is looking at this test? If not, a reference to the value rather than hardcoding would be another way to lead a reader to where the value is coming from.

trevan: I made "test" as a static constant in the class.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/FiringGroupTest.java
trevan: The `givenUnit()` methods aren't returning mocks.  They are returning an actual `Unit` and/or `UnitType` object.  I've changed them to just return a default `UnitType` and/or `Unit` and then set the `isSuicideOnHit` in the test method itself.

DanVanAtta:
@trevan , please '@' mention me when ready for re-review and unassign yourself if you would.

DanVanAtta (approved):

PR #7901, opened 2020-10-14
Add builder for the fire round steps

trevan:
This is part of #7823

This basically replaces `Fire` and `FireAa`.  The four fire steps (AA, FirstStrike, General, and Bombard) will use this builder to build the three fire round steps.

The next PR after this will be the final integration.  I don't think I can split it up any more.  All of these classes will finally be plumbed through the system.  There will be a lot of test changes to deal with the string changes.


game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsBuilder.java
```diff
@@ -0,0 +1,71 @@
...
+    @NonNull final BattleState battleState;
+    @NonNull final BattleActions battleActions;
+    @NonNull final Function<BattleState, List<FiringGroup>> firingGroupSplitter;
+    @NonNull final BattleState.Side side;
+    @NonNull final MustFightBattle.ReturnFire returnFire;
+    @NonNull final BiFunction<IDelegateBridge, RollDice, DiceRoll> roll;
+    @NonNull final BiFunction<IDelegateBridge, SelectCasualties, CasualtyDetails> selectCasualties;
+  }
+
+  public static List<BattleStep> buildSteps(final Parameters parameters) {
```
DanVanAtta, line 34: I think you could fit the builder pattern more exactly by removing the parameter object, adding `@Builder` to `FireRoundStepsBuilder`, and then convert this 'buildSteps' method to be a no-arg.

Even doing that, I think this class would still be more of a factory. The following flow is a bit odd:

```
*Builder.builder()
     .
     .
     .build()
     .buildSteps(...)
```

If we call this a factory though, I think it works out better and is arguably a better fit to the pattern (a builder tends to return the thing that is prefixed on the builder, EG: a stringBuilder returns a string, a "gameBuilder' returns a 'game')

Calling this a factory, we get something like:

```
FireRoundStepsFactory.builder()
             .
             .
             .build()
             .createSteps();
```

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsBuilder.java
```diff
@@ -0,0 +1,71 @@
...
+
+  @Builder
+  public static class Parameters {
+    @NonNull final BattleState battleState;
+    @NonNull final BattleActions battleActions;
+    @NonNull final Function<BattleState, List<FiringGroup>> firingGroupSplitter;
+    @NonNull final BattleState.Side side;
+    @NonNull final MustFightBattle.ReturnFire returnFire;
+    @NonNull final BiFunction<IDelegateBridge, RollDice, DiceRoll> roll;
+    @NonNull final BiFunction<IDelegateBridge, SelectCasualties, CasualtyDetails> selectCasualties;
```
DanVanAtta, line 31: nit, as a function name, what do you think of chaning 'selectCasualties' to 'casualtySelector'. As a verb I think it fits a bit better, eg:

```
CasualtyDetails casualtyDetails = selectCasualties.apply(...)
// vs
asualtyDetails casualtyDetails =  casualtySelector.apply( ...)
```

DanVanAtta: Ditto for 'roll', perhaps better as 'diceRoller'?

`firingGroupSplitter` already has a verb kind of name, updating these last two bi-functions I think would make that naming more consistent in this class overall.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsBuilder.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  public static List<BattleStep> buildSteps(final Parameters parameters) {
+    final List<BattleStep> steps = new ArrayList<>();
+
+    final List<FiringGroup> firingGroups =
+        new ArrayList<>(parameters.firingGroupSplitter.apply(parameters.battleState));
```
DanVanAtta, line 38: Maybe worth considering using the streaming API here? Seeing an array list copy raises flags for me, but that could just be me.

As a stream, we'd have soemthing like:

```
 final List<FiringGroup> firingGroups =
   parameters.firingGroupSplitter.apply(parameters.battleState).stream().sorted(...).collect(Collectors.toList());
 ```

File: game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsBuilderTest.java [hunk 4]
```diff
@@ -0,0 +1,147 @@
...
+                            .defender(defender)
+                            .build())
+                    .build()));
+
+    final List<String> expected =
+        List.of(
+            "attacker" + FIRE_SUFFIX,
+            "defender" + SELECT_PREFIX + CASUALTIES_WITHOUT_SPACE_SUFFIX,
+            "defender" + REMOVE_PREFIX + CASUALTIES_WITHOUT_SPACE_SUFFIX);
+    assertThat(names, is(expected));
```
DanVanAtta, line 145: Thoughts to use the iterable matcher here? You could avoid created the intermediate list and hopefully the failure message would be more user-friendly and tell you exactly how the lists mismatched, EG:

```
assertThat(names, IsIterableContainingInOrder.contains(
    "attacker" + FIRE_SUFFIX,
            "defender" + SELECT_PREFIX + CASUALTIES_WITHOUT_SPACE_SUFFIX,
            "defender" + REMOVE_PREFIX + CASUALTIES_WITHOUT_SPACE_SUFFIX));

```

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsBuilderTest.java
```diff
@@ -0,0 +1,147 @@
...
+            "defender" + REMOVE_PREFIX + "army" + CASUALTIES_SUFFIX);
+    assertThat(names, is(expected));
+  }
+
+  private List<String> getStepNames(final List<BattleStep> steps) {
+    return steps.stream().flatMap(step -> step.getNames().stream()).collect(Collectors.toList());
+  }
+
+  @Test
+  @DisplayName("With two groups, there should be two sets of fire round strings")
```
DanVanAtta, line 71: Good use of `@DisplayName`  :+1:

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsBuilderTest.java
trevan: I tried the `IsIterableContainingInOrder` and I think the failure message from `is` is actually better.  With `is`, it shows the entire `expected` and `actual` array.  But with `IsIterableContainingInOrder`, it only shows the complete `expected` array and the first failure item in the `actual` array.  I think seeing both complete arrays is useful.

trevan:
@DanVanAtta , I've made it a factory object that uses the `@Builder`.  I also did the other name changes.  I created test helpers in `BattleStepsTest` for the step name creation.  They will be used when I do the final PR and will replace the use of the other `generalFightStepStrings` methods.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsFactory.java
```diff
@@ -0,0 +1,53 @@
...
+import lombok.Builder;
+import lombok.NonNull;
+
+/** Build the steps for the fire round (roll dice, select casualties, and mark casualties) */
+@Builder
+public class FireRoundStepsFactory {
+
+  @NonNull final BattleState battleState;
+  @NonNull final BattleActions battleActions;
+  @NonNull final Function<BattleState, List<FiringGroup>> firingGroupSplitter;
```
DanVanAtta, line 25: What is the implied ordering of the output list from firingGroupSplitter? What contract would instances of this local variable need to satisfy to be consistent with the expected usages within this class?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsFactory.java
```diff
@@ -0,0 +1,53 @@
...
+  @NonNull final BattleState.Side side;
+  @NonNull final MustFightBattle.ReturnFire returnFire;
+  @NonNull final BiFunction<IDelegateBridge, RollDice, DiceRoll> diceRoller;
```
DanVanAtta, line 28: Side note, we may want to rename 'RollDice'. 'RollDice' and 'diceRoll' being inverses of each other is somewhat confusing, the distinction seems to be very weak, another maintainer might see no real difference between the two terms.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsFactory.java
```diff
@@ -0,0 +1,53 @@
...
+
+  public List<BattleStep> createSteps() {
+    final List<BattleStep> steps = new ArrayList<>();
+
+    final List<FiringGroup> firingGroups =
+        firingGroupSplitter.apply(battleState).stream()
+            .sorted(
+                Comparator.comparing(FiringGroup::getDisplayName)
+                    .thenComparing(FiringGroup::isSuicideOnHit))
+            .collect(Collectors.toList());
```
DanVanAtta, line 39: nit, this seemingly could be simpler as a map + flatmap operation.

EG:

```
return   firingGroupSplitter.apply(battleState).stream()
            .sorted(
                Comparator.comparing(FiringGroup::getDisplayName)
                    .thenComparing(FiringGroup::isSuicideOnHit))
             .map(firingGroup -> List.of(new RollDice(battleState, side, firingGroup, fireRoundState, diceRoller),
                           new SelectCasualties(battleState, side, firingGroup, fireRoundState, casualtySelector),
                           new MarkCasualties(
              battleState, battleActions, side, firingGroup, fireRoundState, returnFire))
            .flatMap(Collection::stream)

File: game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsFactoryTest.java [hunk 9]
```diff
@@ -0,0 +1,131 @@
...
+    assertThat(
+        names,
+        is(
+            mergeSteps(
+                generalFightStepStrings(attacker, defender, "army"),
+                generalFightStepStrings(attacker, defender, "spies"))));
+  }
+
+  @Test
+  @DisplayName("With the default name UNITS, the step names should be readable")
```
DanVanAtta, line 103: What is meant by 'readable'?

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsFactoryTest.java
```diff
@@ -0,0 +1,131 @@
...
+  void getNamesWithDefaultUnitName() {
+    final GamePlayer attacker = mock(GamePlayer.class);
+    when(attacker.getName()).thenReturn("attacker");
+    final GamePlayer defender = mock(GamePlayer.class);
+    when(defender.getName()).thenReturn("defender");
+    final List<String> names =
```
DanVanAtta, line 109: nit, I notice there is no new line between the test setup and the invocation of the unit under test. Having the setup, invocation of the test unit, and the assertions be separated by a new line helps delimit them and helps the reader easily see what is being tested vs the setup. (Arrange, Act, Assert (AAA) testing pattern). Any objection to inserting such new lines?

For example, 
```
testMethod() {
    setupVar = setup...
    player = ... setup
    otherState = ... setup

    resultValue = methodUnderTest.invoke(player, otherState ...)

    assertions(....) 
}
```

### Side note on AAA

I've even some styles where developers use boiler plate templates with comment blocks to delimit these sections, eg:
```
testMethod() {
   // arrange

   // act

   // assert
}
```

I"ve also seen those comments be called "given, when, then". IMO the comment blocks add clutter and the test method should be simple enough that simple new lines are delimiter enough to mark the different parts of a test. The latter should be possible almost all of the time, particularly if any complicated arrangements are extracted to helper methods that give a high level overview of what we are setting up with parameters for the key values that would control the result value of the unit under test.

DanVanAtta (commented):
Few comments for you to look @trevan . As always thank you for your contributions.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsFactory.java
trevan: I'll change it to `Collection`.  This class doesn't care about the order.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsFactory.java
trevan: Any ideas on renaming it?  `RollDice` is a battle step that sets up the call to `DiceRoll`.  The 2 implementations of `diceRoller` are called `RollAaDice` and `RollNormalDice`.  `RollNormalDice` just calls `DiceRoll` while `RollAaDice` will play a sound and call `DiceRoll`.

One thought is that the other two steps are called `SelectCasualties` and `MarkCasualties` so maybe this should be similar?  Something like `HitCasualties` or `CalculateCasualties`?  Since in LL, dice isn't guaranteed to be rolled.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsFactoryTest.java
trevan: The "step names" is shown to the user in the Battle UI.  So "readable" would be understandable to the user.  When I first looked at the resulting strings in the UI, it sounded odd to me so I worked it a little to make it more "readable".  But it is very subjective.

trevan:
@DanVanAtta , this should be ready for another review.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsFactoryTest.java
DanVanAtta: Ah, qualifying 'readable' as 'human readable' should make that clear. 'readable' can be understood in different ways.

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsFactoryTest.java
```diff
@@ -0,0 +1,137 @@
...
+                .battleState(
+                    FakeBattleState.givenBattleStateBuilder()
+                        .attacker(attacker)
+                        .defender(defender)
+                        .build())
+                .build()
+                .createSteps());
+
+    assertThat(
+        "units should not be in the name",
```
DanVanAtta, line 133: Why should units not be in the name?

I think this would be a stronger statement if it were something like:
"A battle with offense and all suicide units should be labelled as 'XYZ'"

With the above, I wonder why are we not checking for an exact name?

I suspect that would be more obvious if the checked values were not behind a method. I also notice there is logic for a magic value of emptry string in `generalFightStepStrings`. I think it might make for stronger tests to inline the function everywhere in tests to make the assertions explicit and clear.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsFactory.java
DanVanAtta: Maybe call it ` RollDiceStep`? or ` RollDiceBattleStep`

> "Since in LL, dice isn't guaranteed to be rolled."

I suppose it's a pedantic argument whether rolling 0 dice is rolling dice. If you took that view, then having a LL roll dice step would still make sense.

DanVanAtta (commented):
Couple new comments @trevan

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/FireRoundStepsFactoryTest.java
trevan: I had created the method in a separate PR for the purpose to reduce the size of the final PR.  It changes all of these step names in the same way and would make that PR very large.  Once that is in, I can inline the method in a later PR.

As for why "units" should not be in the name, that is the "human readable" subjective view.  The step name is either:
```
Attacker fires
Defender selects casualties
Defender removes casualties
```
vs 
```
Attacker units fire
Defender selects units casualties
Defender removes units casualties
```
I find the former looks better in this case.

But the map designer can give the group a name in the XML.  An example would be "AA", "Boomstick", or "AntiTankGun".  In that case, the name needs to be in the string:
```
Attacker AntiTankGun fire
Defender selects AntiTankGun casualties
Defender removes AntiTankGun casualties
```

trevan:
@DanVanAtta , I've renamed `RollDice` to `RollDiceStep` and `RollMainDice` and `RollAaDice` are now `MainDiceRoller` and `AaDiceRoller`.

PR #7909, opened 2020-10-15
Update triplea_maps.yaml

Cernelius:
This should be merged or closed the same way and at the same time as:
https://github.com/triplea-maps/big_world/pull/7


DanVanAtta:
Thank you for fixing this @Cernelius , this was an important fix, good to see it done

PR #7910, opened 2020-10-15
Add unzipped maps to the map list again

### LINKED ISSUE #7840: v2.3.21895 cannot see my new map projects when I Select a Map, unless I zip them
Im converting my map making projects from v1.9 to 2.3, but TripleA 2.3.21895 cannot see them when I Select a Map, unless I zip them.

Ideally I want to keep the new maps outside of a Zip file, so in normal Windows file structure, like 1.9 used to, while I develop it, so how do I that?

Trevan said
_I did a quick check of a few pre-releases that I've previously downloaded. 2.3.21326 loads a non-zipped map but 2.3.21445 doesn't. So it regressed somewhere in between those two releases. I bet it is related to @LaFayette's work with the map loading._
here
https://forums.triplea-game.org/topic/2319/triplea-2-3-cannot-see-my-new-map-projects-when-i-select-a-map-unless-i-zip-them

trevan:
The AvailableGamesFileSystemReader was added back in 86eafe8b82c84e34177c61311848f3a4f2371f13 and it was looking for xml files by using "Path#endsWith".  But "Path#endsWith" matches against the root component completely and not parts of it.

This converts the Path to a string and then uses "String#endsWith".

Fixes #7840

## Testing
I verified that my unzipped maps would show up in the map selector.

## Release Note
I'm not adding a release note as this is a pre-release regression.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/AvailableGamesFileSystemReader.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/AvailableGamesFileSystemReader.java
@@ -82,7 +82,7 @@ public class AvailableGamesFileSystemReader {
               mapDir.toPath(),
               8,
               (path, basicAttributes) ->
-                  basicAttributes.isRegularFile() && path.getFileName().endsWith(".xml"))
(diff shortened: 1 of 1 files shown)


DanVanAtta:
:+1:

PR #7913, opened 2020-10-16
Make sure the default game is there or clear it if it isn't

### LINKED ISSUE #7839: Leaving an unsaved game and then restarting the same map opens up to the unsaved spot (with errors)
1. Open up pre-release TripleA of at least 2.3.21447
2. Select a map (I usually pick 1940 global but I've done other maps)
3. Start a local game
4. Do at least one step in the round (in 1940, that would be production). You just have to press Done.  You don't actually have to do anything
5. Leave the game
6. Restart the game by just pressing "Play".  Don't press "Select Map" or "Open Saved Game" as that will not trigger this issue.
7. The game will start from where you left.  You will also generally get a missing image error.

I was able to figure out that this was introduced by @DanVanAtta in 2.3.21447 which is the PR #7611.  I undid that one line change in the latest master and was able to fix this issue.

I looked into why the missing image errors and it is because there are some event listeners from the old game session that are still active.  When the game plays, it will trigger those event listeners and they'll try to render icons.  But their resource loader has closed and so they can't find any images and the error is thrown.

trevan:
If the default game is not on the system, it wasn't being cleared out of
the game.  This would leave the game in an odd state where it is
partially loaded but has unused objects listening for events.

By calling resetDefaultGame when the default game is missing, the game
can be reset to an initial state where no map has been selected.

In addition, the checks to see if the file exists had issues when the
uri had special prefixes and suffixes.  So those are stripped before the
checks.

Fixes #7839

## Testing

To test non-zip maps, I had to cherry-pick the code from #7910.  I don't have that code in this actual PR.
I tested on OSX and the prefix stripping was needed for that operating system.  I tested with both zip and unzip maps.  I also tested with deleting the map and starting the game to ensure that it didn't throw an error.

## Release Note
This is a pre-release regression, so no release notes.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
@@ -12,8 +12,14 @@ import games.strategy.engine.framework.startup.mc.GameSelector;
 import games.strategy.triplea.ai.pro.ProAi;
 import games.strategy.triplea.settings.ClientSetting;
 import java.io.File;
+import java.io.IOException;
 import java.net.URI;
+import java.nio.file.FileSystemNotFoundException;
+import java.nio.file.FileSystems;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.List;
+import java.util.Map;
 import java.util.Observable;
 import java.util.Optional;
 import java.util.function.Function;
@@ -83,22 +89,30 @@ public class GameSelectorModel extends Observable implements GameSelector {
 
   public void load(final URI uri) {
     fileName = null;
-    this.gameData = parseAndValidate(uri);
-    if (gameData == null || gameData.getGameName() == null || uri == null) {
-      this.resetDefaultGame();
+    GameData gameData = null;
+    if (uri != null) {
+      gameData = parseAndValidate(uri);
+      if (gameData != null && gameData.getGameName() == null) {
+        gameData = null;
+      }
+    }
+    this.setDefaultGame(uri, gameData);
+  }
+
+  private void setDefaultGame(@Nullable final URI uri, @Nullable final GameData gameData) {
+    setGameData(gameData);
+    if (gameData == null || uri == null) {
+      ClientSetting.defaultGameName.resetValue();
+      ClientSetting.defaultGameUri.resetValue();
     } else {
-      setGameData(gameData);
       ClientSetting.defaultGameName.setValue(gameData.getGameName());
       ClientSetting.defaultGameUri.setValue(uri.toString());
-      ClientSetting.flush();
     }
+    ClientSetting.flush();
   }
 
   private void resetDefaultGame() {
-    setGameData(null);
-    ClientSetting.defaultGameName.resetValue();
-    ClientSetting.defaultGameUri.resetValue();
-    ClientSetting.flush();
+    setDefaultGame(null, null);
   }
 
   @Nullable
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
```diff
@@ -186,10 +189,28 @@ public void loadDefaultGameSameThread() {
         // can load. (ie: if a previous version of triplea was using running a game within its
         // root folder, we shouldn't open it)
         .filter(
-            defaultGame ->
-                defaultGame.contains(ClientFileSystemHelper.getUserRootFolder().toURI().toString()))
-        .filter(defaultGame -> new File(defaultGame).exists())
+            defaultGame -> {
+              String rootFolder = ClientFileSystemHelper.getUserRootFolder().toURI().toString();
```
DanVanAtta, line 193: There is a good bit going on in this filter method. I worry that this method has a mixture of abstraction levels. From what I've seen, it's generally best to keep most logic out of lambdas and try to convert them to be method references by extracting that logic to a helper method.

WDYT of extracting these filter predicates to helper methods? In that case the filter chain would stay quite high level, eg:

```
ClientSetting.defaultGameUri
   .getValue()
   .filter(GameSelectorModel::defaultGameIsInUserRootFolder)
   .filter(GameSelectorModel::defaultGameExists)
   .map(...)
     
```

DanVanAtta, line 193: I wonder if the check for a given parent folder would not be easier to use the `Path` API and  do a recursive check of parents or use the `iterator()` method to check folders. The special casing by OS is hard to grok, I could envision someone testing this and thinking it is unnecessary code (also we really want the code to be as cross-platform as possible, if this needs modification, most of us have no access to a MAC and would SOL).

We have a `FileUtils` class, as a suggestion, what do you think of creating a helper method there to check if a given parent folder exists on a given path, that can be tested readily, and then use that utility method here?

game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
```diff
@@ -186,10 +189,28 @@ public void loadDefaultGameSameThread() {
...
+        .filter(
+            defaultGame -> {
+              if (defaultGame.startsWith("jar:file:")) {
+                // when the game is from a zip file, it starts with jar:file: and uses a '!'
+                // to separate the local file from the zip content
+                defaultGame = defaultGame.substring("jar:file:".length(), defaultGame.indexOf('!'));
+              } else if (defaultGame.startsWith("file://")) {
+                // when the game is from a local file, it starts with file://
+                defaultGame = defaultGame.substring("file://".length());
+              }
```
DanVanAtta, line 210: The lack of an else implies that the default game can be a third case, that would be valid. If not valid, then we keep processing rather than flagging an error here, which I think the "fail early" principle would suggest that it would be better to throw an exception or some other error handling. Even better probably would be to ensure that we cannot have an invalid value set in default game, and if so to remove it, all before we get to this point. WDYT?

DanVanAtta: Second thought, can we use any of the 'isFile' APIs or otherwise to check if we are dealing with a zip file?


This SO seems to have one suggestion we could use:

```
String contentType = Files.probeContentType(Paths.get(pathToFile);

if (!contentType.equals("application/zip")) {

    throw new NotZipFileException();
}
```

https://stackoverflow.com/questions/33934178/how-to-identify-a-zip-file-in-java

I suspect this is another candidate method to be added to `FileUtils` so we can remove this logic from `GameSelectorModel`. The game selector model certainly has a number of issues, but it knowing how to do file system processing and detection seems to be adding to SRP violation, so I am very much in favor of removing complexity from it to a helper to simplify the file system ops.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
```diff
@@ -85,16 +85,19 @@ public void load(final URI uri) {
...
     } else {
       setGameData(gameData);
       ClientSetting.defaultGameName.setValue(gameData.getGameName());
       ClientSetting.defaultGameUri.setValue(uri.toString());
+      ClientSetting.flush();
     }
+  }
+
+  private void resetDefaultGame() {
+    setGameData(null);
```
DanVanAtta, line 98: This set game data seems to be an unnecessary store coming from line 88. It's also not consistent as the other lines sets the game data directly and this uses a setter method. Generally, AFAIK it's better to avoid setter methods within a class (lest someone overrides your setter method and then suddenly they've changed a lot more code then expected).

2nd, the abstraction of the method above seems to be kinda mixed. It's basically:
```
low level checks
low level setting
low level checks

if( low level condition)
    do medium level
else
   do low level
```

I recommend to accept a `GameData` parameter to this method and then you can re-use. Arguably it's a method so that the code can be re-used, but we still have the same calls on lines 91 and 92 (it's not the same, but the data we set is the same, that would be the idea behind DRY, the information of which values to set is repeated in 2 places).

As a method, it would be:
```
private void setDefaultGame(@Nullable final URI fileUri, @Nullable final GAmeData gamedata) {
  if(gameData == null) {
    ClientSetting.defaultGameName.resetValue();
    ClientSetting.defaultGameUri.resetValue();  
  } else {
      ClientSetting.defaultGameName.setValue(gameData.getGameName());
      ClientSetting.defaultGameUri.setValue(uri.toString());
  }
   ClientSetting.flush();
}

```

Now, looking at the above, the fact we have a decoupled file uri and game name is really not good too. That looks very much like a data clump code smell and would do well to be unified behind a single object or a single data type. I think that is probably best resolved as a follow up where we stop storing the default game name, just the default uri, and lookup the game name from that URI when it is needed.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
trevan: `Path#startsWith` does the checking so I switched to that once I've created a Path object.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
trevan: I got `Path.of` to work but it wasn't as easy as just calling `Path.of`.  If the uri is "jar:file:/location/to/file.zip!/location/to/map.xml", then `Path.of` will return a path pointing to the xml file.  But that file is inside of the zip file and isn't what we want.  But the first file store of the Path object is the "zipfs" store which has a name of the actual zip file.  And that can be used to create a Path.

Once I did that, the code simplified to just using Path objects.  I believe this should be cross-platform.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
trevan: I made this change.

trevan:
@DanVanAtta , this is ready for additional review

DanVanAtta:
Thanks @trevan, I hope to be able to take a deeper look later today, in about 8-10 hours or so.

PR #7934, opened 2020-10-19
Incorporating some best practices that I have learned about to previous commit.  

CharlieAtlas:
DanVanAtta told me some best practices on my previous commit of documentation, so I incorporated them.


CharlieAtlas:
I think i am doing something wrong.  It should still work, but it is showing my old commits, and the commits that i made when updating my fork from the main fork.

DanVanAtta:
@CharlieAtlas , run:

```
git pull --rebase upstream master
git push -f
```

The '-f' is because you will have rebased and so it 'destroys' the history. You can see the changes with `git log`. For PRs that have begun review, you generally should avoid force pushing (in which case you do a merge commit, `git merge upstream/master`)

CharlieAtlas:
So is that list of commits i see just the history then? That makes more sense then.  I could then just rebase my master, and then make a new branch for my next set of changes?  Then it would only show my commits in that branch, even though there would be more in my master as i keep updating it from the upstream master?

DanVanAtta:
> So is that list of commits i see just the history then? That makes more sense then. I could then just rebase my master, and then make a new branch for my next set of changes? 

Yes & yes

>  Then it would only show my commits in that branch, even though there would be more in my master as i keep updating it from the upstream master?

Sort of. Part of the trick is that you keep your master to "track" the remote "master" and try to keep the two in sync. Then when you create a new branch from your master, it has all of the latest changes.

For examples, let's call the upstream master branch "UM", your master branch "LM". Let's say "UM" has commits {1, 2, 3}, and that commit '3' is very new and your "LM" has commits {1, 2}. When you rebase "LM" onto "UM", "LM" will then have all of the commits from "UM" and both branches would have commits {1, 2, 3}.

Another way to think of rebase is it's a stack operation. Your current branch is put to the side and git imports the branch that you are rebasing to. Then git does the rebase step where the 'current branch' has each commit added one by one onto the rebased branch.

For example, let's say you have a feature branch with commits: {1, 2, 3} and we have a remote master (UM) with commits { 1, 2 }. Here you can see you created a feature branch and added one new commit. Let's say then someone else pushes a change to UM and it now has commits { 1, 2, N }. If you rebase your feature branch onto UM, first git will "import" the remote branch and you'll have {1 ,2 , N}. Then git will replace the unique commits on your feature branch onto that base, in this case commit "3" is the only new one, and you'll then have {1, 2, N, 3}. If UM gets another commit, say "X", and you rebase again, your rebased branch will have commits {1, 2, N, X, 3}

Going back to LM and UM, the idea is that you keep LM "clean" and do not add any commits to it. In that case rebasing LM onto UM  will always result in LM and UM being "even" (having the same commits). If you do this frequently, and cut feature branches from LM, then you'll generally have a situation that looks like this:
```
LM: { 1, 2, 3 }
UM: { 1, 2, 3}
Feature branch: { 1, 2, 3, X}
```

With the above, you generally always have the latest code, this reduces the chances of another change interacting badly with your code. For example, let's say you did not do the above and you have:
```
LM: { 1, 2 }
UM: {1, 2,3 4, 5, 6 }
Feature Branch: {1, 2, X}
```

Post merge of Feature branch to UM (when we merge to UM, in github, we do 'squash rebase' we would have:
```
UM: {1, 2, 3, 4, 5, 6, X}
```

The problem with the above is that you only tested: `{1 ,2 , X}`, who is to say if change "X" works with changes "4, 5, 6". The process of keeping your branch up to the latest and always pulling the latest changes is called continuous integration. Another tenet is we try to keep changes relatively small and rapid. This means everyone is pulling frequently getting in a small set of changes and pushing frequently a small set of changes. This avoids a "change the world" scenario where you do a pull and then suddenly lots and lots of code is all different.

PR #7944, opened 2020-10-20
Remove TLS 1.1 from forums Nginx

DanVanAtta:
Leaves just TLS 1.2, 1.1 is considered weak.


RoiEXLab (approved):
Can we already use TLS 1.3?

DanVanAtta:
@RoiEXLab likely, I think we only need to add it after some quick testing.

Issue #7848, opened 2020-10-06
Infinite loop in battle calculator


simon33-2:
### How can the problem be recreated?

Unsure to be honest. We believe that the properties for subs are corrupted in the save file and this later causes issues for attacking subs with defending planes but no defending destroyers. Probably vice versa too.

If you download the below file, Open the battle calculator for SZ112, swap sides, add in SZ111 defenders and some planes, then run the calculator, you will get the problem.

### Attach a Save Game

https://www.axisandallies.org/forums/assets/uploads/files/1600889975703-triplea_35795_1rus.tsvg

### If playing a prerelease, which version are you using?

Game Version:
2.2.20868

### If playing a prerelease, does this happen on the latest release?

- [] yes
- [] no

### Is there anything else we should know?
Hopefully, this is enough. Many players are noticing this problem. Also, subs lose their stealth properties at certain points; this is likely the same problem.

trevan:
@simon33-2 , do you know what version of the game created that save file?  And has it been saved in newer versions of the game?

I'm stepping through the code and I see that the submarines are not flagged as `isSub` or as `canEvade`.  In a battle, a sub will automatically submerge when there are only airplanes, but that is because of the `canEvade` flag.  And since these subs don't have it, the battle calculator will spin forever since both sides should be able to fire but they actually can't.

I know there was a change to how subs worked and so that's why I'm wondering when the save was made.  If it was saved before the change and then resaved afterwards, that might be a clue to the cause.

simon33-2:
It was created with 2.2.20868 it seems, although the version which played the second turn is unknown.

I'll pass that on.

trevan:
Is the original save available?  Or any earlier saves?  It would be interesting to see when the data changed.

simon33-2:
Have a look here: https://www.axisandallies.org/forums/topic/35795/l20-bm3-freh-vs-dawgoneit-allies-18

panther2:
@trevan Chances are that this is related: Users report similar sub-related problem situations for example in
https://www.axisandallies.org/forums/topic/35832/2-2-bugs-forum-autosaves-and-sub-rules-issues

trevan:
@panther2 , yes the sub movement issue there is related.

The problem is that in 2.2, the way canEvade and other sub related attributes was tracked changed to fix some bugs.  The PR for that change is in https://github.com/triplea-game/triplea/pull/7227.

[...]

PR #7948, opened 2020-10-21
If no units can hit each other, then trigger a stalemate

trevan:
This uses the new FiringGroup splitters to figure out if either side can hit each other.  This should fix the infinite loop in #7848 as well as future maps that start to utilize `canNotTarget` and `canNotBeTargetedBy`.

## Testing

Ran a few Hard AI to make sure battles were resolved.  Played a few battles manually in 1940 global, iron war, and domination 1914 (with conscripts to see if that would cause any issues).

## Release Note

CHANGE|A battle that only contains units that can not hit each other because of canNotTarget and canNotBeTargetedBy will result in a stalemate.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
+++ b/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
@@ -708,8 +708,10 @@ public class UnitAttachment extends DefaultAttachment {
     }
   }
 
-  private void setCanNotTarget(final Set<UnitType> value) {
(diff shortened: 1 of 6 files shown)


sneakingcoward:
canNotTarget and canNotBeTargetedBy

where can we get the syntax ?
not in pos2 ??

DanVanAtta:
@sneakingcoward please open an issue to ask the question or post to the forums. Conversations in pull requests should be about the code that has changed in that pull request.

PR #7955, opened 2020-10-23
Add game version definitions

trevan:
I've seen these versions used in the forums, code, and github but I haven't seen a place where they are defined.


trevan:
@panther2 , I took what you sent me and put it here in the glossary.  Can you make sure that I'm linking to the correct rule sets?

Also, can someone make sure I'm linking to the correct "official" map?

trevan:
Also, is the `WW2V2` and `WW2V3` properties supposed to map to `v2` and `v3` or do they map to different versions?

Cernelius:
I see the usefulness of something like this especially for new develpers and I'm sorry to be strongly negative here, but I don't think it's a good idea to map this way, especially by referring to any sort of site external to TripleA, especially if commercial, especially by using terminology like "anniversary", that I don't recall having anything to do with TripleA, for never having seen it mentioned anywhere (I suppose that is not the anniversary of TripleA).

Besides, rules are rules and games are games (using maps to visualize themselves). Rules sets are not even map based, but GAME based, since, for example, the "Revised" map has two games, each following a different rules set.

Side note, TripleA rules sets nomenclature is sort of a sequence of numbers with scarce, if any, sense, other than telling what came first, since you have "v2" for two vastly different rules sets, while you go from "v3" to "v4" for two virtually identical rules sets, while v6 is really just the same rules set as v5, just using less elements of it in the game (almost the same thing can be said for v4 being a reduced version of v3).

They are just labels.

Generally speaking, I don't think TripleA should ever refer to anything outside TripleA, and preferably not even any TripleA map or game, either.

I advise not to merge.

Cernelius:
> Also, is the `WW2V2` and `WW2V3` properties supposed to map to `v2` and `v3` or do they map to different versions?

I'm as sure as I can be that, for example, "v1" is just another (shorter) way to say "WW2V1". I think, but I'm not sure about it, that "WW2V1" is a legacy way to name it (back when all TripleA games were supposed to be "WW2").

trevan:
> I don't think it's a good idea to map this way, especially by referring to any sort of site external to TripleA (regardless whether or not anything referenced has anything to do with anything, as I'm not aware of TripleA being related to anything), especially if commercial, especially by using terminology like "anniversary", that I don't recall having anything to do with TripleA, for never having seen it mentioned anywhere (I suppose that is not the anniversary of TripleA).
> 
> Besides, rules are rules and games are games (using maps to visualize themselves). Rules sets are not even map based, but GAME based, since, for example, the "Revised" map has two games, each following a different rules set.
> 
> Side note, TripleA rules sets nomenclature is sort of a sequence of numbers with scarce, if any, sense, other than telling what came first, since you have "v2" for two vastly different rules sets, while you go from "v3" to "v4" for two virtually identical rules sets, while v6 is really just the same rules set as v5, just using less elements of it in the game (almost the same thing can be said for v4 being a reduced version of v3).
> 
> I advise not to merge.

Then how would you map the v1, v2, v3, etc to something that new developers, players, people, etc can understand?  I still don't really know what v1, v2, v3, etc mean but people use them as if they mean something and expect everyone to know.  They are defined somehow.  So please define them for me so we can get it in this glossary so everyone knows what they mean.

trevan:
> > Also, is the `WW2V2` and `WW2V3` properties supposed to map to `v2` and `v3` or do they map to different versions?
> 
> I'm as sure as I can be that, for example, "v1" is just another (shorter) way to say "WW2V1". I think, but I'm not sure about it, that "WW2V1" is a legacy way to name it (back when all TripleA games were supposed to be "WW2").

There is no property `WW2V1`.  I'm talking about the properties in the xml file: `WW2V2` and `WW2V3`.

Cernelius:
> > I don't think it's a good idea to map this way, especially by referring to any sort of site external to TripleA (regardless whether or not anything referenced has anything to do with anything, as I'm not aware of TripleA being related to anything), especially if commercial, especially by using terminology like "anniversary", that I don't recall having anything to do with TripleA, for never having seen it mentioned anywhere (I suppose that is not the anniversary of TripleA).
> > Besides, rules are rules and games are games (using maps to visualize themselves). Rules sets are not even map based, but GAME based, since, for example, the "Revised" map has two games, each following a different rules set.
> > Side note, TripleA rules sets nomenclature is sort of a sequence of numbers with scarce, if any, sense, other than telling what came first, since you have "v2" for two vastly different rules sets, while you go from "v3" to "v4" for two virtually identical rules sets, while v6 is really just the same rules set as v5, just using less elements of it in the game (almost the same thing can be said for v4 being a reduced version of v3).
> > I advise not to merge.
> 
> Then how would you map the v1, v2, v3, etc to something that new developers, players, people, etc can understand? I still don't really know what v1, v2, v3, etc mean but people use them as if they mean something and expect everyone to know. They are defined somehow. So please define them for me so we can get it in this glossary so everyone knows what they mean.

First of all, just to be sure, are you aware of the existence of this sheet?
https://github.com/triplea-game/triplea/blob/master/docs/game-rule-sets.md

I'm asking because I'm not sure if you just want a list of the "basic" TripleA games. What more do you need? Is it feasible to obtain it by building up that document?

---

If you want something that would allow anyone to know everything, I'm sorry: It's not that I don't want to help you. It's that I don't think there is an easy and sound way to map such things, at this point.

First of all, I believe it is not clear if v1 and so are referring to rules sets or to maps or to games (or what else), that is whether they are merely map names or define something at the program level, as I believe they are somewhat spuriously used for both maps and rules sets. It appears to me that they are chiefly merely map names, since, for example, you have v2 for a map having two games with vastly different rules sets, while you have v5 for a map (having two games), that shares the same rules set with several other maps and many other games (the v5 map is, rules wise, a reduced version of the global map, and v6 is the same for v5). However, they are indeed used in the engine (for properties), thus the confusion whether "v1" (and so on) is a map name or a rules set name or what else.

On top of this, also the "property" rules set related usage is highly inconsistent, as only v2 has it made in a partially decent way, while the v3 property is largely pointless and I believe nothing exists thereafter. Thus such things confirming themselves as being merely "some maps" names from v4 onwards. The "v2" property might even be a legacy from back when TripleA was highly hardcoded, in a one-to-one rules set and map relationship, but I'm not sure.

If anyone would ever make any sense of all this, I don't know how I could satisfy your request of "define them" beside writing down a TripleA rulebook fully and correctly covering all rules, divided by rules sets (and this would be no small task, that I don't intend to do, because of the huge effort that I believe it would require (I believe we are talking of hundreds of hours)).

---

To clarify, when I said "WW2V1", I meant "WW2V*", with "*" meaning any number.

trevan:
> > > I don't think it's a good idea to map this way, especially by referring to any sort of site external to TripleA (regardless whether or not anything referenced has anything to do with anything, as I'm not aware of TripleA being related to anything), especially if commercial, especially by using terminology like "anniversary", that I don't recall having anything to do with TripleA, for never having seen it mentioned anywhere (I suppose that is not the anniversary of TripleA).
> > > Besides, rules are rules and games are games (using maps to visualize themselves). Rules sets are not even map based, but GAME based, since, for example, the "Revised" map has two games, each following a different rules set.
> > > Side note, TripleA rules sets nomenclature is sort of a sequence of numbers with scarce, if any, sense, other than telling what came first, since you have "v2" for two vastly different rules sets, while you go from "v3" to "v4" for two virtually identical rules sets, while v6 is really just the same rules set as v5, just using less elements of it in the game (almost the same thing can be said for v4 being a reduced version of v3).
> > > I advise not to merge.
> > 
> > 
> > Then how would you map the v1, v2, v3, etc to something that new developers, players, people, etc can understand? I still don't really know what v1, v2, v3, etc mean but people use them as if they mean something and expect everyone to know. They are defined somehow. So please define them for me so we can get it in this glossary so everyone knows what they mean.
> 
> First of all, just to be sure, are you aware of the existence of this sheet?
> https://github.com/triplea-game/triplea/blob/master/docs/game-rule-sets.md
> 
> I'm asking because I'm not sure if you just want a list of the "basic" TripleA games. What more do you need? Is it feasible to obtain it by building up that document?

Thanks for pointing out that document.  I wasn't aware of it.  But it doesn't exactly help the problem of defining what v1, v2, etc mean.  That just points to example games.  But what is the definition of v1.III or v5?  Take the https://github.com/triplea-game/triplea/issues/4133 issue.  @panther2 goes through various versions and says that they don't match the rules.  Where are these rules defined?

That's what I want.  A mapping of version numbers that people use and the actual rule sets that are behind that number.  I assumed that they were based on the commercial versions of Axis & Allies.  But you apparently are saying that my assumption is incorrect.  So where are the rule books that define v1, v2, etc?  If there aren't rule books, then how can @panther2 say that the game is not following the rules?

Cernelius:
I'm not aware TripleA is related to anything else, commercial or not. If that is the case, then I think it is simply consequential that it should not reference to something it is not related to. I've not affirmed nor denied that any assumption of similarities between anything within and without TripleA happens to be correct.

Of course, there ought to be intended rules behind the program's behaviour and the program's behaviour may be not doing what it was intended to do, in the first place.

As you've probably realized, TripleA is just vastly missing good and correct documentation about what its intended rules are.

trevan:
@Cernelius , in https://github.com/triplea-game/triplea/pull/7227, you said "What that property appears doing is turning the "submarines" behaviour from v1 (Classic) to v2 (Revised) rules (problems aside)" (see comment https://github.com/triplea-game/triplea/pull/7227#issuecomment-667597547).  Where are you getting the ruleset for v1 and v2?

panther2:
@trevan Not being a developer either, this is my understanding:

The homepage of TripleA and maybe the Wikipedia article might indicate what TripleA is, or at least what the authors maybe think TripleA is.

The homepage for example says that TripleA is
"The World's Foremost Open Source Grand Strategy Game
Community built maps, Axis and Allies game rules:"

Now v1-v5, among hundreds of other scenarios, are community built maps, too.

Because of this it does not seem to be appropriate when the project (at whatever place) links whatever scenario to sources of commericial products, IMHO.

Cernelius:
@trevan While I would still close this issue, it is nice that you are endeavouring to improve the usability of the document I linked, but some of those links you are giving are wrong, but I'm not going to tell which ones are, because I'm opposed on a more general level (as I think I've already clarified).

Just to give some hints, to help you:
Europe, Europe 1940.I and Europe 1940.II are three different games.
Pacific, Pacific 1940.I and Pacific 1940.II are three different games.
Europe and Europe 1940 are two different maps.
Pacific and Pacific 1940 are two different maps.
Europe 1940.I and Pacific 1940.I can be considered sharing the same Global 1940.I rules set, just using only part of it.
Europe 1940.II and Pacific 1940.II can be considered sharing the same Global 1940.II rules set, just using only part of it.

The only thing I'm going clearly to say is that the fact that you have linked "Pact of Steel" to
https://github.com/triplea-maps/the_pact_of_steel/blob/master/map/games/pact_of_steel_2.xml
is wrong, because that is "Pact of Steel 2", not "Pact of Steel".

I also disagree with you adding a "v6" rules set. The so called "v6" game merely uses the same rules set as "v5" (with less elements). So, I don't believe it exists a "v6" rules set. If that makes for a different rules sets, then I guess also "World War II v3 1941" and "Big World : 1942 v3 Rules" use different rules sets, since one has a few units and rules the other one is missing. Going on with this concept, we would have almost 1 rules set per game, then (if you have a new rules set every time you use less than 100% of it). However, on this matter, one could argue that all of v6, v5, Europe 1940.II and Pacific 1940.II are the same rules set as Global 1940 II.

Also "v3" and "v4" would be almost the same rules set, with "v4" being a reduced version of "v3", but, in that case there, would be at least one instance in which they behave differently (unplaced units handling), even though this is not captured by TripleA (in TripleA, "v3" and "v4" are indeed exactly the same rules set, even though they shouldn't be).

@Panther Would you list the v6, v5, Europe 1940.II, Pacific 1940.II and Global 1940.II games under the same rules set? If so, would you call it "v5" or "1940.II"? How about listing "v4", "Europe 1940.I", "Pacific 1940.I" and "Global 1940.I" as the same rules set?

trevan:
> @trevan While I would still close this issue, it is nice that you are endeavouring to improve the usability of the document I linked, but some of those links you are giving are wrong, but I'm not going to tell which ones are, because I'm opposed on a more general level (as I think I've already clarified).
> 
> Just to give some hints, to help you:
> Europe, Europe 1940.I and Europe 1940.II are three different games.
> Pacific, Pacific 1940.I and Pacific 1940.II are three different games.
> Europe and Europe 1940 are two different maps.
> Pacific and Pacific 1940 are two different maps.
> Europe 1940.I and Pacific 1940.I can be considered sharing the same Global 1940.I rules set, just using only part of it.
> Europe 1940.II and Pacific 1940.II can be considered sharing the same Global 1940.II rules set, just using only part of it.

I'm sorry, but none of what you wrote makes any sense.  What is "Europe", "Europe 1940.I", "Europe 1940.II", "Pacific", "Pacific 1940.I", "Pacific 1940.II", and all the other things you listed?  And what is the "Global 1940.I" rules sets and "Global 1940.II" rules sets?

And why won't you help me and tell me what maps/games go to which lines in that file?  I don't think you've clarified anything.

trevan:
> @trevan Not being a developer either, this is my understanding:
> 
> The homepage of TripleA and maybe the Wikipedia article might indicate what TripleA is, or at least what the authors maybe think TripleA is.
> 
> The homepage for example says that TripleA is
> "The World's Foremost Open Source Grand Strategy Game
> Community built maps, Axis and Allies game rules:"
> 
> Now v1-v5, among hundreds of other scenarios, are community built maps, too.
> 
> Because of this to me it does not seem to be appropriate when the project (at whatever place) links whatever scenario to sources of commercial products, IMHO.
> 
> When I comment on any rules questions I always comment form a boardgame's point of view as this is the lowest common denominator, because boardgames have written rules while the countless maps of TripleA don't. Also the engine has not every rule of every boardgame it supports. In the end it is the engine developer or the map maker or the tournament director or the players themselves or whoever who decide which rules they intend to follow.

I understand that they are community built maps.  But they are trying to build those maps to match some sort of rule set.  People file bugs saying that they don't match the rule sets.  But no one seems to define what those rule sets are.  It feels like only you (@panther2) know what the rule sets are.  Can you write up what the rule sets are for v1-v5 so that it can be more widely shared?  That way, when people say that v1, v2, etc is not following the rules, people can point to the rules and say why it isn't following the rules and how it should actually behave?

Cernelius:
@trevan I'm doing what I can to clarify and help you. You cannot expect me to help you going in a direction I advise against.

Europe:
https://github.com/triplea-maps/europe/blob/master/map/games/europe_incomplete.xml
Pacific:
https://github.com/triplea-maps/pacific/blob/master/map/games/pacific.xml
Europe 1940.I:
https://github.com/triplea-maps/world_war_ii_europe/blob/master/map/games/ww2eur40.xml
Pacific 1940.I:
https://github.com/triplea-maps/world_war_ii_pacific/blob/master/map/games/ww2pac40.xml
Global 1940.I:
https://github.com/triplea-maps/world_war_ii_global/blob/master/map/games/ww2global40.xml
Europe 1940.II:
https://github.com/triplea-maps/world_war_ii_europe/blob/master/map/games/ww2eur40_2nd_edition.xml
Pacific 1940.II:
https://github.com/triplea-maps/world_war_ii_pacific/blob/master/map/games/ww2pac40_2nd_edition.xml
Global 1940.II:
https://github.com/triplea-maps/world_war_ii_global/blob/master/map/games/ww2global40_2nd_edition.xml
Global 1942.II:
https://github.com/triplea-maps/world_war_ii_global/blob/master/map/games/ww2global42_2nd_edition.xml

Global 1942.II uses the 1940.II rules set (so, there is not a Global 1942 rules set, to be clear).

panther2:
@Cernelius While I agree that v3-v4 (maybe others) as well as v5-1940x2nd-v6 (maybe others) might be subsummarized under a respective "ruleset" each, following the same status of rules development, I would rather see "a ruleset" on a per-game perspective.
So for example for a boardgame personally I would prefer a "ruleset" to be: Rulebook+FAQ-sheet+Official Clarification.

@trevan I am afraid I cannot resolve this dilemma. I have already talked to you which boargames might have been inspriation for some TripleA-maps. But there is nothing generic, that I am aware of, which would be suitable to be more widely shared in a TripleA project context.

Cernelius:
> @Cernelius While I agree that v3-v4 (maybe others) as well as v5-1940x2nd-v6 (maybe others) might be subsummarized under a respective "ruleset" each, following the same status of rules development, I would rather see "a ruleset" on a per-game perspective.
> So for example for a boardgame personally I would prefer a "ruleset" to be: Rulebook+FAQ-sheet+Official Clarification.

Thanks for clarifying your perspective.

Then, in that case, in the document I linked, "Big World : 1942" should certainly be removed from the "V2 Basic" row and "Big World : 1942 v3 Rules" should certainly be removed from the "V3" row.

"Big World : 1942 v3 Rules" is certainly farther away from "V3" than "World War II v6 1941" is from "V5", rules wise.

I'm not sure if you mean that, for example, also "World War II Europe 1940 Original" is a different rules set from "World War II Pacific 1940 Original" (that is not what the current document is saying, as they are currently listed as games of the same "Europe/Pacific/Global 1940.I" "rule variant").

Then, in this case, let's just completely delete the document I linked, making just a list of basic games, instead (the developers will have to decide which ones amongst the many games that TripleA offers are the "basic" ones, that is the ones having a set of "intended" rules behind them, constituting the basis of the TripleA program (which doesn't really make sense, as I'm not aware TripleA offers any per-game rulebook and I'm still not seeing how it can make any sense somehow searching for it outside TripleA)). I cannot see any point having a "Rule Variant" (whatever that actually means) column beside an "Example Games" column in the moment the "rules set" equals to the "game rules" (so you would have a 1:1 correspondece between the first and the secon column in all or almost all cases).

I reiterate that the root of the problem is that only for "V1" (the default) and "V2" (the only "rules set" having a mostly working general property) there is something that can be seen as a "rules set", upon which you can build whatever number of different maps and games, using the ruleset, while from "V3" onwards you have either a mostly useless property or no property at all. Therefore, anything from "V3" onwards is, as far as TripleA only is concerned, hardly anything more than a game-only label. "World War II v6 1941" is called as such, but maybe it could have been called "Mario" or "Luigi". That is just the game name. What's the point of all this?

In my opinion, here there is an attempt of making sense of something that cannot be make sense of, because it is not sensible at the roots. So, I'm not seeing how any such attempt may succeed. It's merely a legacy issue that the "v-series" nomenclature has evolved in a spurious and inconsistent way (and also arguably leaving out the "1940" series, unless you decide to see those as part of the "V5" rules games, which is yet another undefined matter, as far as I know).

DanVanAtta:
Linking to rule books I think is an intermediate step and is the research needed to get to the document that we want.

@trevan is correct we need a meaning of what our terms mean, like "V1" or "V3". If we do not define that anywhere, then when someone says, the "V3 rules are incorrect in TripleA for this game", then we would respond "V3 has no meaning, therefore it's a meaningless statement to say we're not compliant with a term without meaning." Though, we do know "V3" means:
- subs cannot be hit by air
- naval bombard casualties fire back
- bombing damage is done to factories and are not subtracted from a players PU total.


I think what we are then looking for is a list of rules, for example "naval bombard casualties fire back" and a set of columns indicating if those options are a "yes" or a "no" for each rule set, or a "N/A" where the rule set makes no distinction about it. Thus what I envision is we'd have one row for each rule and columns for each rule set. This way you can look down a column for example at "v3" and see which rule options would be true if you have a "v3" game. Also, you can look at a row to see when a specific rule option should be "on" or "off" depending on the rule set.

Hence, why the term is a 'rule-set", it is a set of rules. The rules mentions above are individual, but something like "v3" implies a set of those to be true, a set of those to be false, and the remainder it would be 'n/a' and would be optional. For example, LHTR bombers, AFAIK is an optional rule for any rule set.

Cernelius:
> I think what we are then looking for is a list of rules, for example "naval bombard casualties fire back" and a set of columns indicating if those options are a "yes" or a "no" for each rule set, or a "N/A" where the rule set makes no distinction about it. Thus what I envision is we'd have one row for each rule and columns for each rule set. This way you can look down a column for example at "v3" and see which rule options would be true if you have a "v3" game. Also, you can look at a row to see when a specific rule option should be "on" or "off" depending on the rule set.
> 
> Hence, why the term is a 'rule-set", it is a set of rules. The rules mentions above are individual, but something like "v3" implies a set of those to be true, a set of those to be false, and the remainder it would be 'n/a' and would be optional.

Under this (partial) definition of what a rules set is, I can confirm that the "v3" and "v4" games have the same rules set and the "v5" and "v6" games have the same rules set (regarding "v3" and "v4", there would actually be a very small difference, making them different rules sets, which is not currently covered by TripleA, however). Therefore, as I said, past "v3", the "v-series" would fail to define rules sets (because same rules-set games have been named with increasing numerals). Moreover, the "v-series" numeration fails in the opposite direction when categorizing "v1" and "v2", because there are two different rules set both called "v1" (the second and third edition, according what is written in the game names, which would imply there is also a first edition that it is not currently supported, but it is meant to be (if it is not the case, then the second edition of "v1" should be renamed as first edition and the third edition should be renamed as second edition)) and two different rules sets both called "v2" (the first one is not even defined, while the second one is called "LHTR" ("LHTR" is so different from the original "v2" that it is about in the middle between "v2" and "v3" actually).

Moreover, I tend to think that the 1940.I series are all games of the "v3" rules set with some additional special rules. Same for the 1940.II series (comprising 1942.II) with respect to "v5".

>For example, LHTR bombers, AFAIK is an optional rule for any rule set.

As far as I know, this is a non-optional rule for the latest "v2" game (only) and not necessarily an option for any other games (of course, I'm sure many games have it as an option). Using this option as the "v3" default is a common mistake.

DanVanAtta:
@Cernelius , we need to put pedantry to the side. Here is the goal, when starting each game, we supply a list of all rules with checkboxes. We also supply a list of rule sets that can be selected, eg: "v1", "v2", etc. The question is, which rules should be selected if someone selects "v2". Any existing lack of definition frankly be damned, we are seeking to *make* the definition here for going forward. To save work and avoid confusion, knowing how best to describe the existing is ideal so we fitting the defined definition to the working definition.

Why is this important? Because it is code, we have to tell the computer what to do, and to be exact. We can't be vague and say, "oh, well V3 sometimes means this, sometimes it means that." If that is just 100% the case, then we need to throw out the "v1" terms completely. I don't think that is the case though, if you say, this is a "V3" version of this map, it implies certain rules. The question is then, what are those rules.

DanVanAtta:
> we supply a list of all rules with checkboxes. We also supply a list of rule sets that can be selected, eg: "v1", "v2", etc. The question is, which rules should be selected if someone selects "v2".

To clarify, this may, or may not be exposed to the user. To the code, it is important as the XML does have a "v3" attribute on it and that is essentially the same thing. The engine needs to know under the covers which rules maps to which rule set.

DanVanAtta:
I think we're I'm landing on this, having the references and examples is a good first table. A second table, the one I think we really want, would be the table mapping specific rules to which variants imply those rules. Hence, the existing table and modification to have reference rule books is good background information.

PR #7967, opened 2020-10-26
Fix AA support calculation

trevan:
All of the friendly and enemy units need to be used to calculate the
support during casualty selection.

When getting the total power and total strength values for the
defendingAa, the "defending" value needs to be flipped since the
"defending" value is the side of the targeted planes and not the side
of the AA.  And the enemy/friendly units need to be flipped as well.

## Testing

The only map with AA unit support is Warcraft War Heroes so I ran several Hard AI games and also used breakpoints to go through the code and make sure things were correct.

I came across this while working on refactoring the TotalPowerAndTotalRolls code.  The AA unit supports is correctly calculated for the dice roll step but it wasn't being correctly calculated for the casualty selection step.


DanVanAtta:
nice work @trevan  - thank you for fixing this  :+1:

PR #7979, opened 2020-10-28
Throw exception when 10,000 rounds are hit in a battle

### LINKED ISSUE #7848: Infinite loop in battle calculator

trevan:
Fixes #7848 

I think this is the last thing to fix for #7848.

## Testing

Ran world at war with Hard AI to turn 13 just to see if I'd hit this exception in normal game play.  The exception did not occur.

## Release Note

CHANGE|An exception will be thrown if a battle ever takes more than 10,000 rounds to help figure out how that happened.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
@@ -1706,7 +1706,6 @@ public class MustFightBattle extends DependentBattle
                 throw new IllegalStateException(
                     "Round 10,000 reached in a battle. Something must be wrong."
                         + " Please report this to TripleA."
-
                         + " Attacking unit types: "
                         + attackingUnits.stream()
                             .map(Unit::getType)
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
@@ -106,6 +106,8 @@ public class MustFightBattle extends DependentBattle
 
   private static final long serialVersionUID = 5879502298361231540L;
 
+  private static final long MAX_INFINITE_ROUNDS = 10000;
+
   private final Collection<Unit> attackingWaitingToDie = new ArrayList<>();
 
   private final Collection<Unit> defendingWaitingToDie = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


Cernelius:
I think 1,000 would be more than enough...

I assume this is a temporary patch, until that never actually happens.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -106,6 +106,8 @@
 
   private static final long serialVersionUID = 5879502298361231540L;
 
+  private static final long MAX_INFINITE_ROUNDS = 10000;
```
DanVanAtta, line 109: What do you think of naming this `MAX_ROUNDS` instead?
"max infinite" does not quite grok well for me.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1699,6 +1701,25 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
           public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
             if (!isOver) {
               round++;
+              if (round > MAX_INFINITE_ROUNDS) {
+                // the battle appears to be in an infinite loop
+                throw new IllegalStateException(
+                    "Round 10,000 reached in a battle. Something must be wrong."
```
DanVanAtta, line 1707: What do you think of adding to the message "Please report this to TripleA"?
A line break between this first line and the unit listing could make it an easier message to read as well.

DanVanAtta (commented):
@trevan couple comments for you to consider.

trevan:
@DanVanAtta , I incorporated your feedback.

DanVanAtta:
@Cernelius I suspect you're right with 1k, at 10k we are quite confident though we do not have a false positive.

I don't expect this to be temporary, like *many* error conditions we have to handle when coding, this is one we never expect to hit but need to add handling if somehow we do (otherwise the game freezes).

DanVanAtta:
@trevan would you mind temporarily & locally setting the max round to a low number (1 or 0) and verifying that the output format looks good?

trevan:
Screen Shot 2020-10-28 at 10 22 54 PM

Here's what it looks like in the error dialog.  And I submitted it to github to see what it looks like here.  That issue is #7984.

DanVanAtta:
Thank you for double checking @trevan  - looks good :+1:

runspired:
If we're hitting this fairly consistently what info would best help y'all track it down?

trevan:
@runspired , if you can give me a save right before it is happening, that would be extremely helpful.  If not, if you could click the "Report to TripleA" and then find the issue that was added to github and link to it from here.

runspired:
@trevan I don't have a save but I'll start saving every turn so that I do for you on the next one, this seems to end 4 of 5 games for me at the moment.  I suspect it's because I make fairly copious use of edit-mode for the scenario I play to keep the sides balanced (the lotr map starts so unbalanced against the "evil" side that even with giving them sizable resource modifiers they lose pretty quick, so I usually bring Rhun back from the dead or some such).

PR #8022, opened 2020-10-31
Pass around CombatValue instead of its pieces

trevan:
This simplifies several of the apis of DiceRoll and CasualtySelector by using CombatValue as a parameter.

## Testing

Ran World at War with Hard AI starting in a round 10+.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/StrategicBombingRaidBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/StrategicBombingRaidBattle.java
@@ -32,6 +32,7 @@ import games.strategy.triplea.delegate.battle.casualty.CasualtySelector;
 import games.strategy.triplea.delegate.battle.casualty.CasualtySortingUtil;
 import games.strategy.triplea.delegate.data.BattleRecord;
 import games.strategy.triplea.delegate.data.CasualtyDetails;
+import games.strategy.triplea.delegate.power.calculator.CombatValue;
 import games.strategy.triplea.formatter.MyFormatter;
 import games.strategy.triplea.util.TuvUtils;
 import java.util.ArrayList;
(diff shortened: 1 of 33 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/CombatValue.java
```diff
@@ -15,8 +15,14 @@
 
   boolean isDefending();
 
+  Collection<TerritoryEffect> getTerritoryEffects();
```
trevan, line 18: I added this api method because `territoryEffects` was being passed around solely to be used in a "simplified" combat value calculator (see `UnitBattleComparator#getUnitPowerForSorting`).  I think I'll be able to replace that completely with the new combat values but I didn't want to include that change in this PR.  So, by adding this `getTerritoryEffects` api, it allows me to remove the need to pass it around manually and then once I rework `UnitBattleComparator#getUnitPowerForSorting`, I'll remove this api completely.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/CombatValue.java
```diff
@@ -15,8 +15,14 @@
...
+
   GameData getGameData();
 
+  Collection<Unit> getFriendUnits();
```
trevan, line 22: `getFriendUnits` and `getEnemyUnits` is used solely to send the friend and enemy units across the wire in `Player#selectCasualties`.  And the only implementation of `Player#selectCasualties` that uses the enemy and friend units is `AbstractProAi#selectCasualties`.  I've thought about putting a `@ChangeOnNextRelease` tag here, but I'm not exactly sure if there is something that could be done to remove the need for these two methods.

DanVanAtta:
@trevan , test failure:
```
games.strategy.triplea.delegate.power.calculator.TotalPowerAndTotalRollsTest > GetUnitPowerAndRollsForNormalBattles > attackShouldNotHaveFirstTurnLimiting() FAILED

    java.lang.NullPointerException: friendUnits is marked non-null but is null
```

game-core/src/main/java/games/strategy/triplea/delegate/battle/StrategicBombingRaidBattle.java
```diff
@@ -587,33 +588,40 @@ private CasualtyDetails calculateCasualties(
...
           battleSite,
-          territoryEffects,
           bridge,
-          text, /* dice */
-          null,
-          /* defending */ false,
-          battleId, /* head-less */
-          false,
+          text,
+          null, /* dice */
```
DanVanAtta, line 601: this kind of line rally 'screams' for a builder.

DanVanAtta (approved):

Issue #1033, opened 2016-08-04
Map XML Changes/Deletions Request


Cernelius:
Note: This post has been edited as per summing up everything relevant at the following discussion. This first post is supposed to be self-sufficient; meaning that you don't need to read any part of the following discussion to integrate what at this first post; but I advise you to read the whole discussion, anyway, to have more details.

This post consists mainly of two lists. Both lists are self sufficient and have the same items.
The first list aims at making simple the most to know where and how the single items' changes are to be made.
The second list groups the items and provides some explanations.
Practically, the second listing aims at making the items fairly understendeable, and the first listing aims at directing the changes as much clearly as possible.
All the items sharing the same number are supposed to be interrelated: it would make no sense making only part of them.

Also, the first list is for tracking what has been already done [done].

In addition to the already made:
Attatch->Attach
attatch->attach

I wish to propose the following changes and deletions, to be included in the 1.9 engine (the numbers at the first listing refer primarily to the second listing):

XML related engine reductions:
_Means that these ones should be value restricted._
7A)**victoryCity": stop supporting boolean** only for this option
8B)**"giveUnitControl": stop supporting boolean** only for this option

XML related engine deletions (if not already done):
_Means that these ones should be just fully deleted in the engine only, but not in any maps' xml._
8A)**"takeUnitControl"**
9A)**"occupiedTerrOf"**
10A)**"isParatroop"**
11A)**"isMechanized"**
12A)**"isTwoHit"**
13A)**option name="conditionType" value="XOR"**
14A)**option name="trigger"**
15A)**"SBR Affects Unit Production"**

XML related engine only changes:
_Means that these ones should be changed in the engine only, but not in any maps' xml._
1A)**Naval Bombard Casualties Return Fire->Naval Bombard Casualties Do Not Return Fire**

XML related engine changes and XML changes:
_Means that these ones should be name-changed both in the engine and in any maps' xml having them._
[done]2A)**isImpassible->isImpassable**
3A)**offence->offensive**
3B)**defence->defensive**
4)**Units repair at end of round->Units Repair Hits End Turn**
5)**Units repair at beginning of round->Units Repair Hits Start Turn**
6)**option name="turns"->option name="rounds"**

XML only deletions:
_Means that these ones should be removed in any maps' xml having them, but not in the engine._
8C)**option name="takeUnitControl" value="true"**
8C)**option name="takeUnitControl" value="false"**
8C)**option name="giveUnitControl" value="true"**
[...]

PR #8023, opened 2020-10-31
Only skip giveUnitControl if the value is false or true

### LINKED ISSUE #7990: Units placed in the India territory not recognised.
When playing the 'Triple A - World War II Global Balanced Mod 3 with Combat Move First' map against AI, after units are placed in the India territory (to existing factory as British player), they are unusable in future rounds. Territory appears empty when running cursor over.

trevan:
Fixes #7990

In #1033, @Cernelius said to remove `giveUnitControl` if its values was a boolean.  But he said to leave it alone if the value wasn't a boolean.  In #7631, it was just skipped without looking at the value.  This changes it to check the value and only skip it if it is a boolean.

I think this breaks all of the world_war_ii_global games as well as several others.

## Testing
Played World War 2 Global 1940 2nd to UK_Pacific Place round, placed some units in India, finished the place round and checked that the new units are now British units.

## Release Note

FIX|World War 2 Global will correctly switch unit ownership of units placed in India from UK_Pacific to British.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
+++ b/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
@@ -895,9 +895,6 @@ public final class GameParser {
       }
       // decapitalize the property name for backwards compatibility
       final String name = LegacyPropertyMapper.mapLegacyOptionName(decapitalize(option.getName()));
-      if (LegacyPropertyMapper.ignoreOptionName(name)) {
-        continue;
-      }
 
       if (name.isEmpty()) {
         throw new GameParseException(
@@ -905,6 +902,9 @@ public final class GameParser {
       }
       final String value = option.getValue();
       final String count = option.getCount();
+      if (LegacyPropertyMapper.ignoreOptionName(name, value)) {
+        continue;
+      }
       final String countAndValue = (count != null && !count.isEmpty() ? count + ":" : "") + value;
       if (containsEmptyForeachVariable(countAndValue, foreach)) {
         continue; // Skip adding option if contains empty foreach variable
--- a/game-core/src/main/java/games/strategy/engine/data/gameparser/LegacyPropertyMapper.java
+++ b/game-core/src/main/java/games/strategy/engine/data/gameparser/LegacyPropertyMapper.java
@@ -55,7 +55,9 @@ class LegacyPropertyMapper {
     return propertyName;
   }
 
-  static boolean ignoreOptionName(final String name) {
-    return name.equalsIgnoreCase("takeUnitControl") || name.equalsIgnoreCase("giveUnitControl");
+  static boolean ignoreOptionName(final String name, final String value) {
+    return name.equalsIgnoreCase("takeUnitControl")
+        || (name.equalsIgnoreCase("giveUnitControl")
+            && (value.equals("false") || value.equals("true")));
   }
 }
--- a/game-core/src/main/java/games/strategy/engine/data/gameparser/LegacyPropertyMapper.java
+++ b/game-core/src/main/java/games/strategy/engine/data/gameparser/LegacyPropertyMapper.java
@@ -58,6 +58,6 @@ class LegacyPropertyMapper {
   static boolean ignoreOptionName(final String name, final String value) {
     return name.equalsIgnoreCase("takeUnitControl")
         || (name.equalsIgnoreCase("giveUnitControl")
-            && (value.equals("false") || value.equals("true")));
+            && (value == null || value.equals("false") || value.equals("true")));
   }
 }


Cernelius:
If TripleA 2.3 breaks all of the world_war_ii_global games, I would say 2.4 needs to be released as soon as possible. I guess we don't have many pre-release playtesters on these very popular maps (I never play any of them).

There is also the problem that any documentation on the matter is missing.
https://github.com/triplea-game/triplea/blob/master/docs/map-making/map-features-change-log.md

It used to be present in some old versions (so, this could be fixed by reverting to one of those).
https://github.com/triplea-game/triplea/blob/0f2851209251e7573ed8bcf5c87cc05f11e6ddd9/docs/map-making/upgrading-1-8-maps.md

It was even specifically documented in the bottom line of what I linked (as I foresaw this could have been an easy source of confusion for developers and mapmakers).
>While all options named "takeUnitControl" must be deleted, any option named "giveUnitControl" must be deleted only if its value is "true" or "false".

DanVanAtta:
@trevan  by

> I think this breaks all of the world_war_ii_global games as well as several others.

Is this, this PR, or are you referring to the 1.8 maps PR: #7631

game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
```diff
@@ -895,16 +895,16 @@ private Attachable findAttachment(
       }
       // decapitalize the property name for backwards compatibility
       final String name = LegacyPropertyMapper.mapLegacyOptionName(decapitalize(option.getName()));
-      if (LegacyPropertyMapper.ignoreOptionName(name)) {
```
DanVanAtta, line 898: Would you mind explaining why this if block is moved down @trevan ?

If we are going to ignore an option, why should we execute `option.getValue()` or `option.getCount()`, wouldn't those just be dead stores?

game-core/src/main/java/games/strategy/engine/data/gameparser/LegacyPropertyMapper.java
```diff
@@ -55,7 +55,9 @@ public String mapPropertyName(final String propertyName) {
     return propertyName;
   }
 
-  static boolean ignoreOptionName(final String name) {
-    return name.equalsIgnoreCase("takeUnitControl") || name.equalsIgnoreCase("giveUnitControl");
+  static boolean ignoreOptionName(final String name, final String value) {
+    return name.equalsIgnoreCase("takeUnitControl")
+        || (name.equalsIgnoreCase("giveUnitControl")
+            && (value.equals("false") || value.equals("true")));
```
DanVanAtta, line 61: Are you positive that `value` could never be null here? I don't remember if we translate empty values to empty strings or null values.

trevan:
> @trevan by
> 
> > I think this breaks all of the world_war_ii_global games as well as several others.
> 
> Is this, this PR, or are you referring to the 1.8 maps PR: #7631

The 1.8 maps PR.

game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
trevan: I moved it because I needed access to `option.getValue()` and I thought it would make more sense to use the local variable `value` that was going to be created vs passing in `option.getValue()`.  But I can move it back up and pass in `option.getValue()`.

game-core/src/main/java/games/strategy/engine/data/gameparser/LegacyPropertyMapper.java
trevan: `value` is never checked for null in the caller before using (see line 908 in GameParser) so I don't think it would ever be null.

trevan:
@DanVanAtta , I responded to your comments

game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
DanVanAtta: The local variable IMO is legacy, it's an unnecessary intermediate variable. This class in general followed a C++ pattern (if not older actually, maybe a C pattern or even Pascal) to declar all local variables at the beginning of a method. IMO all usages of 'value' probably should just get inlined.

I did not spot you were using the `value` variable. That answers my question.

game-core/src/main/java/games/strategy/engine/data/gameparser/LegacyPropertyMapper.java
DanVanAtta: Considering we just had a NPE in `GameParser`, I would not presume it to be already correct. Another consideration, the 'value' is very context dependent, a DTD previously would have enforced values being provided, that is not necessarily guaranteed now. 

I wonder if there is any type of test coverage around this at all that could/should be updated.

DanVanAtta: Hmm, I may be wrong about the DTD previously requiring a specific value, I'm not sure if it can constrain attribute values vs just requiring them being there. Regardless, it is context dependent if we can expect a value.

trevan: Good point on the NPE.  It will now skip `giveUnitControl` if the value is `null` as well as `true` or `false`.

There is some test coverage but it requires creating a brand new xml with these attributes in it.  I'm looking at if there is a way to be able to create the xml in memory instead of committing the xml as a resource.

DanVanAtta: If we can get away from XMLs for testing, that would be excellent.

The decoupling of XML parsing from the semantic data parsing into GameData was meant to help with this.

One should be able to set up a `org.triplea.map.data.elements.Game` object and then pass that to GameParser and avoid having to set up an XML. The former is well tested for a variety of XML snippets, so we can be pretty sure how that works. It should have some existing tests that might be interesting to know if we null out empty values. Beyond that, sending `GameParser` a defined `Game` object hopefully can do the trick. `GameParser` is still very rough, so it might take some plumbing work to accept a `Game` object here. Let me know if that turns out to be the case, for now a manual verification is enough or you can go the extra mile if you feel it's worthwhile given the context acquisition cost you will have payed.

DanVanAtta (approved):

PR #8048, opened 2020-11-01
Give the launched process the same xss settings

trevan:
This gives the sub process the same xss setting as the main client.  This should allow the sub process to open large saved games and map xml files.

## Testing

I used the saved game from #8034.  I created a headless client and a headed client.  I joined the headless client, loaded the saved game, marked all of the players as me, and then let the game start.  Without this change, the game would not start and get stuck.  With this change, the game loads and I can play.

## Release Note

FIX|Playing large save/maps in the lobby will no longer hang because of a stack overflow error.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/GameDataManager.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/GameDataManager.java
@@ -74,7 +74,7 @@ public final class GameDataManager {
       } else {
         return Optional.empty();
       }
-    } catch (final ClassNotFoundException | IOException e) {
+    } catch (final Throwable e) {
       log.error("Error loading game data", e);
       return Optional.empty();
     }
--- a/game-core/src/main/java/games/strategy/engine/framework/ProcessRunnerUtil.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/ProcessRunnerUtil.java
@@ -28,6 +28,11 @@ final class ProcessRunnerUtil {
             .map(s -> s.substring(4))
             .findFirst();
     maxMemory.ifPresent(max -> commands.add("-Xmx" + max));
+    final Optional<String> maxStackSize =
+        ManagementFactory.getRuntimeMXBean().getInputArguments().stream()
+            .filter(s -> s.toLowerCase().startsWith("-xss"))
+            .findFirst();
+    maxStackSize.ifPresent(commands::add);
     if (SystemProperties.isMac()) {
       commands.add("-Dapple.laf.useScreenMenuBar=true");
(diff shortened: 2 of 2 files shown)


DanVanAtta:
@trevan are you sure you can consistently get the game started?

I'm still seeing the frozen 'loading game, please wait...' dialog with this branch.

I wonder if there is some other dependency that is at play. For example, the first launch is potentially okay, or maybe after the first launch, but later no.

I think this update is still probably the right thing to do, but I can't consistently repro a fix here though.

trevan:
@DanVanAtta , I can open/close/open/close without any problems.  Maybe your launcher doesn't have an -Xss setting in it?  I checked the HeadedGameRunner in IntelliJ for me and it has -Xss1250K as a vm option.  If your launcher doesn't have the -Xss setting, then this change wouldn't help.  Does the main package add a -Xss setting?

trevan:
Maybe add a log to `ProcessRunnerUtil#populateBasicJavaArgs` that prints out the commands and see if it is adding the '-Xss' parameter?

DanVanAtta:
The launcher I believe is the same (I suspect I was the one that checked it in!).

One issue, once I have the bot host using '-xss' I do not see any stack overflow errors at all and do still see a frozen 'loading game' progress indicator. It does appear it's stuck on the client side reading the game data input stream.

trevan:
@DanVanAtta , on my system, it never was stuck on reading the data stream.  Add a log line at the start of `ClientModel#startGameInNewThread`.  This method is called after all of the data stream has been sent to the client but before it parses the data stream.  On my system, the log line will print showing that the data has been transferred.

I just pushed a change that will catch all `Throwable`s when loading a game.  This should show the stack overflow error that I'm seeing when I don't have the "xss" change.

trevan:
Also, the headless client never showed any errors for me.  I didn't need to change the "xss" for the headless client.  It was always the headed client that was problematic for me.

DanVanAtta:
'xss' was added to headless client pretty recently. If you remove that, do you see a new stack overflow?

trevan:
> 'xss' was added to headless client pretty recently. If you remove that, do you see a new stack overflow?

If I remove 'xss' from the headless client, I get a stack overflow error in the headless client logs.  It happens immediately after I load the game.  I don't see a loading screen but I can't pick my side.

If I remove the `catch` that I added in the second commit of this PR, then I see an error in the UI complaining about an exception occurring on the remote side.

DanVanAtta:
I'm not sure why the catch on the exception block would make a different, with that I'm no longer able to repro the freeze at 'starting game, please wait...'

:+1:

Issue #8015, opened 2020-10-31
2.4.22192: GameSelectorModel#getDefaultGameRealPath:224 - java.lang.NullPointerException


tripleabuilderbot:
## TripleA Version
2.4.22192

## Java Version
11.0.6

## Operating System
Windows 10

## Memory
Heap utilization statistics [MB]
Used Memory: 24
Free memory: 1954
Total memory: 1979
Max memory: 1979

## Stack Trace
```
Exception: java.lang.NullPointerException
java.lang.Exception
	at java.base/java.nio.file.FileSystems.getFileSystem(FileSystems.java:230)
	at games.strategy.engine.framework.startup.ui.panels.main.game.selector.GameSelectorModel.getDefaultGameRealPath(GameSelectorModel.java:224)
	at games.strategy.engine.framework.startup.ui.panels.main.game.selector.GameSelectorModel.lambda$loadDefaultGameSameThread$1(GameSelectorModel.java:209)
	at java.base/java.util.Optional.filter(Optional.java:223)
	at games.strategy.engine.framework.startup.ui.panels.main.game.selector.GameSelectorModel.loadDefaultGameSameThread(GameSelectorModel.java:207)
	at games.strategy.engine.framework.GameRunner.showMainFrame(GameRunner.java:80)
	at java.base/java.lang.Thread.run(Thread.java:834)


```

beelee1:
closing v 2.4

PR #8058, opened 2020-11-02
Fix, handle invalid default game URIs

DanVanAtta:
Catches NPE reported in: https://github.com/triplea-game/triplea/issues/8015

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import java.util.Observable;
 import java.util.Optional;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import javax.annotation.Nullable;
 import lombok.Getter;
 import lombok.Setter;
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
```diff
@@ -226,7 +226,7 @@ public void loadDefaultGameSameThread() {
   private static Path getDefaultGameRealPath(final URI defaultGame) {
     try {
```
DanVanAtta, line 227: @trevan  , could you remind me why we have the try/catch blocks in this method? It looks like they can be removed entirely. Is there an undocumented side-effect that we are relying on for this method to work properly?

trevan: If the path is a zip file, then the `getFileSystem` will initially through `FileSystemNotFoundException` so you'll have to call `newFileSystem` to create it.  The second time, `getFileSystem` will not have any problem.  For non zip, `getFileSystem` doesn't through anything.  This is documented at https://docs.oracle.com/javase/7/docs/api/java/nio/file/FileSystems.html#getFileSystem(java.net.URI).

I don't know why it would throw null pointer.  Maybe that happens if `defaultGame` is a null uri?

DanVanAtta: Could we call 'newFileSystem' first always and then call 'getFileSystem'?

It looks like if the URI is missing a scheme, then we'll get a NPE in file systems.

DanVanAtta: I force pushed over this PR and drastically changed the solution, just catching a NPE is a bad fix.

I think we probably could use some documentation to explain the needed side effects from the FileSystem calls, I was certainly confused by it and wondered why it was needed. 

FWIW, I think we've made good progress here, seems we have much more robust code here that doesn't break.

Issue #8062, opened 2020-11-02
2.5.22285: GameSelectorModel#gameUriExistsOnFileSystem:217 - java.nio.file.FileSystemNotFoundException


tripleabuilderbot:
## TripleA Version
2.5.22285

## Java Version
11.0.6

## Operating System
Windows 10

## Memory
Heap utilization statistics [MB]
Used Memory: 24
Free memory: 1955
Total memory: 1979
Max memory: 1979

## Stack Trace
```
Exception: java.nio.file.FileSystemNotFoundException
java.lang.Exception
	at jdk.zipfs/jdk.nio.zipfs.ZipFileSystemProvider.getFileSystem(ZipFileSystemProvider.java:169)
	at jdk.zipfs/jdk.nio.zipfs.ZipFileSystemProvider.getPath(ZipFileSystemProvider.java:155)
	at java.base/java.nio.file.Path.of(Path.java:208)
	at games.strategy.engine.framework.startup.ui.panels.main.game.selector.GameSelectorModel.gameUriExistsOnFileSystem(GameSelectorModel.java:217)
	at java.base/java.util.Optional.filter(Optional.java:223)
	at games.strategy.engine.framework.startup.ui.panels.main.game.selector.GameSelectorModel.loadDefaultGameSameThread(GameSelectorModel.java:209)
	at games.strategy.engine.framework.GameRunner.showMainFrame(GameRunner.java:80)
	at java.base/java.lang.Thread.run(Thread.java:834)


```

beelee1:
closing v 2.5

PR #8065, opened 2020-11-02
Handle FileSystemNotFoundException when check if the URI is invalid

trevan:
@DanVanAtta , your fix in #8058 now causes an exception to be thrown when the default game is from a zip file.  That's because `Path.of()` will throw a `FileSystemNotFoundException` when the uri is a zip file and the file system hasn't been created.

## Testing

In a build without #8058, load a zipped map.  Quit, rebuild with #8058, then start.  You'll get an error message saying `FileSystemNotFoundException`.  This PR prevents that exception from happening.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
@@ -237,6 +237,9 @@ public class GameSelectorModel extends Observable implements GameSelector {
    * method will find the location of the zip file itself.
    */
   private static Path getDefaultGameRealPath(final URI defaultGame) {
+    // The file system of the URI needs to be created before Path.of can be called.
+    // So, first see if the file system is already created and if that throws
+    // FileSystemNotFoundException, then try and create it.
     try {
       FileSystems.getFileSystem(defaultGame);
     } catch (final FileSystemNotFoundException notFoundException) {


trevan:
Fixes #8062

DanVanAtta:
An update is already applied to release/2.5, I'll mark that as a latest once it finishes building. So the pressure is off to patch master.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
```diff
@@ -218,6 +218,8 @@ private static boolean gameUriExistsOnFileSystem(final String gameUri) {
     } catch (final IllegalArgumentException ignored) {
       // thrown if the URI is invalid (EG: missing URI scheme)
       return false;
+    } catch (final FileSystemNotFoundException ignored) {
+      // the file system will be created inside of getDefaultGameRealPath
```
DanVanAtta, line 222: Forward referencing another method in a comment is an odd form of coupling.

I think the fix in: https://github.com/triplea-game/triplea/pull/8066 is a bit more direct.

I will admit I am disappointed that Path.of breaks if it references a path inside of a zip file via URI.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
```diff
@@ -237,6 +239,9 @@ private static boolean gameUriExistsOnFileSystem(final String gameUri) {
    * method will find the location of the zip file itself.
    */
   private static Path getDefaultGameRealPath(final URI defaultGame) {
+    // The file system of the URI needs to be created before Path.of can be called.
+    // So, first see if the file system is already created and if that throws
+    // FileSystemNotFoundException, then try and create it.
```
DanVanAtta, line 244: Is there a significant cost to creating the file system?

game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
trevan: Yes, that is better.  I'll change this PR to just include the new comments.

DanVanAtta:
Thanks for jumping on this @trevan 

What do you think of folidng these updates into: #8066? Feel free to modify that branch directly. A patch should be out in a few minutes, the 2.5 release branch is almost done building installers.

trevan:
> Thanks for jumping on this @trevan
> 
> What do you think of folidng these updates into: #8066? Feel free to modify that branch directly. A patch should be out in a few minutes, the 2.5 release branch is almost done building installers.

@DanVanAtta, since my updates would just be comments, I don't think it needs to go into 2.5.  It can stay in just master.

trevan:
> > Thanks for jumping on this @trevan
> > What do you think of folidng these updates into: #8066? Feel free to modify that branch directly. A patch should be out in a few minutes, the 2.5 release branch is almost done building installers.
> 
> @DanVanAtta, since my updates would just be comments, I don't think it needs to go into 2.5. It can stay in just master.

Oh, I just realized that #8066 is for master.  Yeah, I'll push the comments there.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
trevan: I don't think there is a significant cost.  I can't find anything online easily.  But I believe the code that later opens the zip also has to create the filesystem internally.

PR #8087, opened 2020-11-03
Use new roll dice, select casualties, and remove casualties steps

trevan:
The Fire and FireAa classes are no longer used in building the 3 fire
sub steps.

The AirVsNonSubs step is removed because that firing group is now named
and will be displayed in the battle ui when it fires, selects, and
removes.

This is the last part of #7823

This will cause the step names in the battle UI to change because the new step classes generate slightly different names.  Here's an example of the old names:

BombardmentPre

Here's what it looks now:

BombardmentPost

Notice that there is a "fire", "select", and "notify" line for all of the firing groups.

Another change is how "air vs subs" is displayed in the Battle UI.  Previously, a single line was shown "Air defend non subs" or "Air attack non subs".  Here's an example:

AirVsSubsPre

Now, that line is no longer going to be shown and instead each of the firing groups get their own set of "fire", "select", and "remove".  I've added special coding to detect the air vs subs to give it a better name.  Here's what it looks like now:

AirVsNonSubs

In addition to the new names, the Battle UI will actually highlight them.  So when there is different firing groups, each time the dice is rolled, it will highlight which firing group is doing it.  It previously worked this way only for AA firing groups.  If there were multiple other groups, it would just highlight the same "<Player> fire" line or "<Player> select casualties" until all the groups were done.

## Testing

I've run a Hard AI run of WWII 1940 world for 12 rounds.  No errors were found.

I ran multiple of my old saves to verify compatibility.  There is potentially some issues if some of the new step names don't match the old step names.  I've fixed most of them but there might be some really rare cases.  If these issues do happen, the UI will just say that "<step name> wasn't found" and continue.  And this will only affect the current round of the save game.  So it should be very minor.

I ran AA flyover battles, strategic bombing battles, amphibious, land, sea, and sub battles.

## Release Note

CHANGE|Battle step strings in Battle UI have been updated to indicate all of the possible offensive and defensive groups that are participating. Instead of one "British fire", there could be multiple "British units fire" depending on canNotTarget, canNotBeTargetedBy, subs, typeAA, etc

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/FireAa.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/FireAa.java
@@ -1,5 +1,8 @@
 package games.strategy.triplea.delegate.battle;
 
+import static games.strategy.triplea.delegate.battle.BattleState.Side.DEFENSE;
+import static games.strategy.triplea.delegate.battle.BattleState.Side.OFFENSE;
+
 import games.strategy.engine.data.GamePlayer;
 import games.strategy.engine.data.Territory;
 import games.strategy.engine.data.TerritoryEffect;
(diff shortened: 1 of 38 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/battle/FireAa.java
```diff
@@ -182,77 +196,33 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
...
-        hitPlayer,
-        battleId,
-        battleSite);
-  }
+  /**
+   * Breaks list of units into groups of non suicide on hit units and each type of suicide on hit
+   * units since each type of suicide on hit units need to roll separately to know which ones get
+   * hits.
+   */
+  static List<Collection<Unit>> newFiringUnitGroups(final Collection<Unit> units) {
```
trevan, line 204: This method was moved from `MustFightBattle` without any changes.

DanVanAtta (approved):

PR #8088, opened 2020-11-04
Allow power to go over diceSides in AirBattle

trevan:
If an air unit had 2 rolls with a strength equal to the dice sides, that
unit should hit 2 targets in low luck. But with the power limited to
dice sides, the unit would only be able to hit 1 target. The dice sides
limit is not used in any other power calculation, either.

It looks like Battle for Arda and Greyhawk would be affected by this change.  They both have a "dragon" unit that has multiple rolls and the power will be higher after this change.  But they don't have low luck enabled by default so not sure how big of a change this will be to players.

## Release Note

FIX|Air Battle units with multiple rolls in Low Luck battles will be able to use all their power instead of being limited to the dice sides of the game.  Battle of Arda and Greyhawk are two such maps that will be affected in Low Luck.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
@@ -342,8 +342,7 @@ public class DiceRoll implements Externalizable {
     for (final Unit current : units) {
       final UnitAttachment ua = UnitAttachment.get(current.getType());
       final int rolls = AirBattle.getAirBattleRolls(current, defending);
-      int totalStrength = 0;
-      final int strength =
+      int unitStrength =
           Math.min(
               data.getDiceSides(),
               Math.max(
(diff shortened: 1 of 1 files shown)


trevan:
I plan to do a larger refactoring of this code but I wanted to get this change in separately as it isn't just a refactoring/restructuring.

Cernelius:
This does not apply to the case in which you get only 1 hit maximum, right (like in the LHTR offending heavy bombers)?

trevan:
@Cernelius , this code is not really LHTR related.  Technically if the game had dice sides of 24 and the unit had an `attack` of 21, then this change would allow the power to be 25 instead of 24.  But this change also makes the power calculation for air battles to match the power calculation of non air battles.  And I expect that the power calculation is supposed to be the same.

trevan:
The power calculation for non air battle can be found at https://github.com/triplea-game/triplea/blob/master/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/TotalPowerAndTotalRolls.java#L364.  You can see that it doesn't do the dice side limit for the power.  It does do the dice side limit for the strength after support but that is before it does the strength * roll calculation to get the power.

trevan:
I redid this change.  Instead of patching this code, I copied the power code from TotalStrengthAndTotalRolls.  That way, both the power calculations are identical.  In a later PR, I'll work on removing the duplicated code.

PR #8098, opened 2020-11-05
Pass in a TotalPowerAndTotalRolls instance to UnitBattleComparator

trevan:
This removes the duplicate power calculation from UnitBattleComparator.

Even though the comment says that it should have the same calculation as what is in TotalPowerAndTotalRolls, it is slightly different.  The two differences that I've found is that the marine bonus is given to defenders and that if the territory effect is a negative, then the power can be in the negative.

So this change will affect the OOL but it should make it more accurate.  I had to fix one test because of the marine bonus issue.  It ranked marines better than artillery because the marine was incorrectly being given the marine bonus during its defense.

## Testing

I added some code to compare the result of the new power calculator against the old power calculator and ran both World at War and Total World War.  The only differences that I saw were because of the two items that I mentioned above.


game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/CombatValue.java
```diff
@@ -17,19 +18,21 @@
...
-
   GameData getGameData();
 
   Collection<Unit> getFriendUnits();
 
   Collection<Unit> getEnemyUnits();
 
+  CombatValue buildWithNoUnitSupports();
+
+  CombatValue buildOppositeCombatValue();
```
trevan, line 29: The names for these two methods (`buildWithNoUnitSupports` and `buildOppositeCombatValue`) could probably be better.  And maybe there's a way to not need them but I couldn't figure one out.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/CombatValue.java
```diff
@@ -91,24 +94,49 @@ static CombatValue buildAaCombatValue(
...
-            .gameData(data)
+            .gameData(gameData)
             .supportFromFriends(supportFromFriends)
             .supportFromEnemies(supportFromEnemies)
             .friendUnits(allFriendlyUnitsAliveOrWaitingToDie)
             .enemyUnits(allEnemyUnitsAliveOrWaitingToDie)
             .build();
   }
+
+  static CombatValue buildNoSupportCombatValue(
```
trevan, line 118: I added this because the `buildMainCombatValue` requires a territory and some places this is called doesn't have a territory.  And I didn't want to pass in null to `buildMainCombatValue` for territory.  With the bombard combat value change (#8097) removes the territory parameter so I could wait for that to merge and then redo this piece.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/PowerStrengthAndRolls.java
```diff
@@ -124,4 +124,15 @@ public int getStrength(final Unit unit) {
...
+
+  @Override
+  public int calculatePower(final Unit unit) {
+    return totalStrengthAndTotalRollsByUnit.get(unit).calculatePower();
+  }
+
+  @Override
+  public TotalPowerAndTotalRolls buildOpposite() {
+    return PowerStrengthAndRolls.build(
+        totalStrengthAndTotalRollsByUnit.keySet(), calculator.buildOppositeCombatValue());
```
trevan, line 136: Since the unit is ordered differently, this new PowerStrengthAndRolls will not be as optimized.  But the only use of `buildOpposite` is when the combat value has no unit support so it won't actually be affected.  Should I add a comment about that?  Or maybe change this method to always create a non-supported combat value?

DanVanAtta:
Few merge conflicts popped up @trevan

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/PowerCalculator.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.function.Function;
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.Value;
+
+@Value
+@Getter(AccessLevel.NONE)
+@AllArgsConstructor
+public class PowerCalculator {
```
DanVanAtta, line 14: Any objection to java-doc'ing this class to give an overview of what it does?

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/CasualtyOrderOfLossesTestOnGlobal.java
```diff
@@ -367,9 +360,9 @@ void improvedArtillery() {
         CasualtyOrderOfLosses.sortUnitsForCasualtiesWithSupport(amphibAssault(attackingUnits));
```
DanVanAtta, line 360: Drilling into this method, I see: `.defending(false)` , which implies this is offense. @trevan can you comment? Am I missing something?

trevan: @DanVanAtta , in `UnitBattleComparator#compare`, it uses both the offense power and the defense power for comparing.  It will first compare the wanted direction (in this case `offense`) and if they are equal, it will then compare the other direction.  The marine has an offense of 2 (1 + marine bonus) and the artillery has an offense of 2 so they are equal.  It then compares the defense and before this change, the marine would have a defense of 3 (2 + marine bonus) and the artillery would only have a defense of 2.  The artillery would be considered worse and so would be the first in the list.  Now, they are considered identical (`compare` returns 0) so I *think* the behavior is that they get sorted according to how they previously exist in the list.  But it is possible the result could be non-deterministic.

I think maybe who ever wrote this test expects the artillery to support the marines but they only support infantry and mech_infantry.  So the marines get no support from them.

trevan:
@DanVanAtta , I've resolved the conflicts, added the java-doc, and replaced `buildNoSupportCombatValue` with `buildMainCombatValue`

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/CasualtyOrderOfLossesTestOnGlobal.java
trevan: Yeah, it looks like it is based on initial order of the list.  If I add the marines to `attackingUnits` before the artillery, then the resulting order is different.

DanVanAtta: This ordering *should* be deterministic. Attack power, when on attack, should be the overriding attribute above any.

If we have a marine, on attack, with amphib, it has an attack power of 3. The artillery though has an attack power of 2. If choosing between marine (attack 3) and artillery (attack 2), the choice should *always* be artillery first. The ordering previously asserted in the test looks consistent with that.

trevan: A marine on attack with amphib is only 2.  Why do you think it should be 3?

DanVanAtta: Marine has a base of '2'.
The amphib assault bonus gives '+1', -> 3

DanVanAtta: Does global have marines at a power of '1' on attack as base?

trevan: Marines only have strength of 1 in global.

DanVanAtta: Ah, yup, global has marines at '1', I was thinking big world v3 stats.

PR #8119, opened 2020-11-07
Create AirBattleCombatValue objects and use them

trevan:
Move all of the air battle strength/roll/power logic to one place
similar to the AA and main battle combat values work.

## Testing

Did a hard AI run and manually played an air battle with break points to make sure strength/rolls/power was calculated correctly.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleDefenseCombatValue.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleDefenseCombatValue.java
@@ -1,0 +1,93 @@
+package games.strategy.triplea.delegate.power.calculator;
+
+import games.strategy.engine.data.GameData;
+import games.strategy.engine.data.TerritoryEffect;
+import games.strategy.engine.data.Unit;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.Value;
+import org.triplea.java.collections.IntegerMap;
+
+/** Calculates offense strength and roll for Air Battle dice */
+@Builder
+@Value
+@Getter(AccessLevel.NONE)
+class AirBattleDefenseCombatValue implements CombatValue {
+
+  @Getter(onMethod = @__({@Override}))
+  @NonNull
+  GameData gameData;
+
+  @Getter(onMethod = @__({@Override}))
+  @NonNull
+  @Builder.Default
+  Collection<Unit> friendUnits = List.of();
+
+  @Getter(onMethod = @__({@Override}))
+  @NonNull
+  @Builder.Default
+  Collection<Unit> enemyUnits = List.of();
+
+  @Getter(onMethod = @__({@Override}))
+  @NonNull
+  @Builder.Default
+  Collection<TerritoryEffect> territoryEffects = List.of();
+
+  @Override
+  public RollCalculator getRoll() {
+    return new AirBattleDefenseRoll();
+  }
+
+  @Override
+  public StrengthCalculator getStrength() {
+    return new AirBattleDefenseStrength(gameData.getDiceSides());
+  }
+
+  @Override
+  public boolean isDefending() {
+    return false;
+  }
+
+  @Override
+  public int getDiceSides(final Unit unit) {
+    return gameData.getDiceSides();
+  }
+
+  @Value
+  @AllArgsConstructor
+  static class AirBattleDefenseRoll implements RollCalculator {
+
+    @Override
+    public RollValue getRoll(final Unit unit) {
+      return RollValue.of(unit.getUnitAttachment().getDefenseRolls(unit.getOwner()));
+    }
+
+    @Override
+    public Map<Unit, IntegerMap<Unit>> getSupportGiven() {
+      return Map.of();
+    }
+  }
+
+  @Value
+  @AllArgsConstructor
+  static class AirBattleDefenseStrength implements StrengthCalculator {
+
+    int diceSides;
+
+    @Override
+    public StrengthValue getStrength(final Unit unit) {
+      return StrengthValue.of(diceSides, unit.getUnitAttachment().getAirDefense(unit.getOwner()));
+    }
+
+    @Override
+    public Map<Unit, IntegerMap<Unit>> getSupportGiven() {
+      return Map.of();
+    }
+  }
+}
--- a/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleDefenseCombatValue.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleDefenseCombatValue.java
@@ -1,8 +1,8 @@
 package games.strategy.triplea.delegate.power.calculator;
 
 import games.strategy.engine.data.GameData;
-import games.strategy.engine.data.TerritoryEffect;
 import games.strategy.engine.data.Unit;
+import games.strategy.triplea.Properties;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
@@ -14,7 +14,12 @@ import lombok.NonNull;
 import lombok.Value;
 import org.triplea.java.collections.IntegerMap;
 
-/** Calculates offense strength and roll for Air Battle dice */
+/**
+ * Calculates defense strength and roll for Air Battle dice
+ *
+ * <p>Air Battles don't have support and are not affected by territory. The unit's defense strength
+ * is determined by the airDefense attribute.
+ */
 @Builder
 @Value
 @Getter(AccessLevel.NONE)
@@ -34,11 +39,6 @@ class AirBattleDefenseCombatValue implements CombatValue {
   @Builder.Default
   Collection<Unit> enemyUnits = List.of();
 
-  @Getter(onMethod = @__({@Override}))
-  @NonNull
-  @Builder.Default
-  Collection<TerritoryEffect> territoryEffects = List.of();
-
   @Override
   public RollCalculator getRoll() {
     return new AirBattleDefenseRoll();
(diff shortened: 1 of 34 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleDefenseCombatValue.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.Map;
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.Value;
+import org.triplea.java.collections.IntegerMap;
+
+/** Calculates offense strength and roll for Air Battle dice */
```
DanVanAtta, line 17: Is this comment correct? "offense" vs BattleDefense

What is involved when calculating air battle combat value? How is this different from land battles? Answering those questions could perhaps make this a stronger javadoc to give a maintainer a good sense of what and how this class accomplishes it's goal.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleDefenseCombatValue.java
```diff
@@ -0,0 +1,93 @@
...
+@Builder
+@Value
+@Getter(AccessLevel.NONE)
+class AirBattleDefenseCombatValue implements CombatValue {
+
+  @Getter(onMethod = @__({@Override}))
```
DanVanAtta, line 23: minor, if you have '@Getter' on every field, you can add it to the class to have it apply to all fields.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleDefenseCombatValue.java
```diff
@@ -0,0 +1,93 @@
...
+  @NonNull
+  @Builder.Default
+  Collection<Unit> friendUnits = List.of();
+
+  @Getter(onMethod = @__({@Override}))
+  @NonNull
+  @Builder.Default
+  Collection<Unit> enemyUnits = List.of();
+
+  @Getter(onMethod = @__({@Override}))
```
DanVanAtta: minor, it can be a nicer syntax to use the 'underscore' variant, eg: `@Getter(onMethod_ = @Override)`

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleDefenseCombatValue.java
```diff
@@ -0,0 +1,93 @@
...
+    return false;
+  }
+
+  @Override
+  public int getDiceSides(final Unit unit) {
+    return gameData.getDiceSides();
+  }
+
+  @Value
+  @AllArgsConstructor
```
DanVanAtta, line 78: `@Value` already includes `@AllArgsConstructor`, the latter should not be needed.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleDefenseCombatValue.java
```diff
@@ -0,0 +1,93 @@
...
+  static class AirBattleDefenseRoll implements RollCalculator {
+
+    @Override
+    public RollValue getRoll(final Unit unit) {
+      return RollValue.of(unit.getUnitAttachment().getDefenseRolls(unit.getOwner()));
```
DanVanAtta, line 83: IIRC, 'getOwner' can return null for neutrals. Will that cause an issue here? I'd recommend adding a test case to verify this.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleDefenseCombatValue.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @Override
+    public Map<Unit, IntegerMap<Unit>> getSupportGiven() {
+      return Map.of();
```
DanVanAtta, line 88: side-note, a no-op return value is code-smell (indicates that the API is overly-broad and that we implement a sub-set of the API only. It hints that the interface is actually mutiple interfaces and/or the methods on the interface have low cohesion). Any thoughts?

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleDefenseCombatValue.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }
+
+  @Value
+  @AllArgsConstructor
+  static class AirBattleDefenseStrength implements StrengthCalculator {
```
DanVanAtta, line 94: side-note, I really wonder about our OO design. Generally an object should take care of its concerns and we could have a single method like "unit.getAirDefenseStrength". Keeping data and its access consistent is very important on the other hand so we probably should not mix it up until we can fix the majority of it in one go.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleDefenseCombatValue.java
```diff
@@ -0,0 +1,93 @@
...
+
+    int diceSides;
+
+    @Override
+    public StrengthValue getStrength(final Unit unit) {
+      return StrengthValue.of(diceSides, unit.getUnitAttachment().getAirDefense(unit.getOwner()));
```
DanVanAtta, line 100: This API is bothersome:
```
 unit.getUnitAttachment().getAirDefense(unit.getOwner()));
```

A 'unit' should already know who it's owner is. The encapsulation around data is very weak, we have knowledge here of attachments, unit owners, the air defense type.. Any thoughts if this can be fixed?

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleOffenseCombatValue.java
```diff
@@ -0,0 +1,93 @@
...
+
+/** Calculates offense strength and roll for Air Battle dice */
+@Builder
+@Value
+@Getter(AccessLevel.NONE)
+class AirBattleOffenseCombatValue implements CombatValue {
+
+  @Getter(onMethod = @__({@Override}))
+  @NonNull
+  GameData gameData;
```
DanVanAtta, line 30: Having a reference to a game data is a bit of a step backwards. It's a god-class and the fix for it is to avoid adding references to it. 

What if instead of passing the 'gameData' reference to the builder of this class, we push the responsibility getting the dice sides to the caller? That would also make this more of a value class as well where it really just holds values.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleOffenseCombatValue.java
```diff
@@ -0,0 +1,93 @@
...
+    return new AirBattleOffenseStrength(gameData.getDiceSides());
+  }
+
+  @Override
+  public boolean isDefending() {
+    return false;
+  }
+
+  @Override
+  public int getDiceSides(final Unit unit) {
```
DanVanAtta, line 58: It's really odd that `getDiceSides` has `Unit` as a parameter. AFAIK we do not support per-unit dice sides, it's a global config for a game. Why is the `unit` parameter here? Do we ever user `Unit` to compute the dice sides value?

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleDefenseCombatValue.java
trevan: `getOwner` never returns null.  If the player is `null` it sets `owner` to `GamePlayer.NULL_PLAYERID`.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleDefenseCombatValue.java
trevan: These methods are only used by the OOL.  I plan to look into your attempt to rewrite the OOL (#6376).  Hopefully, that will remove the need to grab the support given and I can remove all of the code around the support given.  But if this `CombatValue` is passed into the `CasualtyOrderOfLosses` at this moment, it needs to have this method.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleDefenseCombatValue.java
trevan: We could start adding methods to the `Unit` class that encapsulates this.  So something like:
```
public int getAirDefense() {
  this.getUnitAttachment().getAirDefense(this.getOwner());
}
```

It'll make the `Unit` class a lot bigger with all of these methods since it will probably end up with all of the existing methods in `UnitType` and `UnitAttachment`.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleOffenseCombatValue.java
trevan: I'll work on that after this PR.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleOffenseCombatValue.java
trevan: AA units have per-unit type dice sides.  Look at `AaDefenseCombatValue#getDiceSides` and `AaOffenseCombatValue#getDiceSides`

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AirBattleDefenseCombatValue.java
DanVanAtta: That could make a lot of sense now. The 'roots' of TripleA were to be completely generic, a unit could have been something like a chess piece. We've removed that level of generality to focus on the grand strategy style games.

We probably should plan such a change and design it out so we don't get ourselves caught into a corner. For example, how do we handle methods that do not apply? For example, a 'getAirDefense' method for a land unit? Do we return zero, do we throw an exception? If we return zero, can we distinguish between units that do not participate in air defense vs the ones that do but actually have zero power (IE: fodder).

It's also worth planning for how many methods and attributes we will be accumulating. Perhaps the right object model is to have strategy and/or value objects for the different aspects of a Unit. For example, a 'purchasing', a 'combatStrength', a 'movementStrategy', etc.. rather than having *all* implemenation be on Unit, instead delegate its major sub-components to additional objects.

Issue #7710, opened 2020-09-19
Serializing a change list - Changing the way save games are done


DanVanAtta:
I'm opening this issue to:
- walk through potential technical changes for save game serialization
- think through any problems we might have to solve
- review the design of changing save game serialization
- check if there can be help with this effort

We've long had an idea kicking around to change from java serialization for save games to something that is text based. At some point in the last few months an idea has been formed that perhaps we could serialize (to JSON) the change list of a game data. Essentially this data structure:

```
@Builder
public class SaveGameModel {
  @Nonnull private final String gameName;
  @Nonnull private final String mapName;
  @Nonnull private final List<Change> changeList;
}
```

The idea is that the game data would be treated as three entities. 
1) an immutable core of 'map data' that is derived from loading the map.
2) a list of change objects
3) a final 'game data' that is the map data with all change objects applied.

This would allow us to compute a 'game data' object. In this manner, if we store enough information to know which map, we can load that, then given a list of changes we can re-compute the 'game data' (the game state).

Storing this set of changes is actually quite easy, in GameData we only need a few modifications:

```
  // add this variable
  private final transient List<Change> gameChanges = new ArrayList<>();

  // update 'performChange' to add the change object
  /** Executes a change and notifies listeners. */
  public void performChange(final Change change) {
    gameChanges.add(change);
```

Then in `GameDataWriter` we can change the output quite readily:
```
  public static void writeToFile(
      final GameData gameData,
      final DelegateExecutionManager delegateExecutionManager,
      final File file) {

    // this is new, we convert here to a save game model object
    final var saveGameModel = SaveGameModel.builder()
        .gameName(gameData.getGameName())
        .mapName(gameData.getProperties().get(Constants.MAP_NAME, null))
        .changeList(gameData.getGameChanges())
        .build();
    
    try {
      // instead of using an object writer here, instead we convert the game model to a JSON string and write that.
      Files.writeString(file.toPath(), gson.toJson(saveGameModel));
    } catch (final IOException e) {
      log.log(Level.SEVERE, "Failed to save game to file: " + file.getAbsolutePath(), e);
    }
  }
```

Then when loading, we'd read the save game model, we could then find the right map, we'd parse that to get an initial game state, then we'd run "addChange" of game data to apply all of the saved changes and we would then be back to the latest game state.

Doing this, we run into this error on save:

> SEVERE: jdk.internal.ref.PhantomCleanable<?> declares multiple JSON fields named next
[...]

PR #8155, opened 2020-11-12
Create a new Unit instance instead of the instance that was sent across the wire

trevan:
Currently, if the Unit doesn't exist in the local GameData, it copies the Unit that is being deserialized.  With the idea in #7710, this needs to be changed so that the unit isn't copied.

I've put this change behind the beta flag because I'm not exactly sure what possible issues this could cause.  It is possible that there is some behavior of the game that depends on new Units being copied with their current state.  I played a simple networked game and only saw new units being transferred as part of unit purchase but I'm fairly certain there are other situations (such as units switching sides, units transforming, etc).

Once this is released, we need to ask people to turn on the beta flag and then test out network games on maps.

## Release Note

CHANGE|Unit objects are no longer copied between networked game instances. This is behind the beta flag as there might be situations where the game assumes that the unit's state (such as amphibious, submerged, damage, etc) will be copied without going through change objects.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/GameObjectInputStream.java
+++ b/game-core/src/main/java/games/strategy/engine/data/GameObjectInputStream.java
@@ -1,6 +1,7 @@
 package games.strategy.engine.data;
 
 import games.strategy.engine.framework.GameObjectStreamFactory;
+import games.strategy.triplea.settings.ClientSetting;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.ObjectInputStream;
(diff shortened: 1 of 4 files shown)


trevan:
Here's the places that look like there might be a problem:

Units that change on being captured (see `BattleTracker#captureOrDestroyUnits`)
Units that change when damaged (see `MustFightBattle#damagedChangeInto`)
Units that change when repaired (see `MoveDelegate#repairedChangeInto`)
Infrastructure units that are placed by triggers (see `TriggerAttachments#placeUnits`)

In each of those, there is a change request done to the unit before the add unit change request.  So I'm not sure how that will be affected by this change.

DanVanAtta:
I think we need to move the maps server download code to its own feature flag. Players will probably run into problems, I wonder if we should do a similar dedicated feature flag for serialization changes. That would leave the beta flag available for other features where we have higher confidence that they are less experimental and more early access.

trevan:
Do you want me to wait on you building a new feature flag before I make a new feature flag?

DanVanAtta:
@trevan go ahead, I think both feature sets need their own feature flags.

trevan:
@DanVanAtta , I've added a new feature flag.

PR #8219, opened 2020-11-22
Move the unit sorting into PowerStrengthAndRolls for unification

trevan:
Instead of preceeding each call to PowerStrengthAndRolls.build() with
their own unique sorting logic, the sorting logic is now inside of
build() so that it is uniform.  There is a buildWithPreSortedUnits() to
allow the rare case where the sort is either not needed or requires
special logic.

## Testing

Played a few battles with Hard AI


DanVanAtta (approved):
Good update, much better encapsulation :+1:

PR #8222, opened 2020-11-22
Reword "units to place" text -> "purchased units"

DanVanAtta:
"Units to place" is confusing as the verb tense makes it
sound like it is an immediate action occuring during the
current phase. Considering the 'placement units panel'
appears in all phases except the placements, it's confusing.

Even worse, the 'units to place' text appears next to
'units to move', the latter is an immediate action that
is occurring during the current phase, the former is a
future action (confusing).

"Purchased Units" is direct and is usually semantically correct
on most games (arguably in others, the purchase price was zero
for some units, hence still semantically correct).

## Screens Shots

Screenshot from 2020-11-22 00-46-37

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/PlacementUnitsCollapsiblePanel.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/PlacementUnitsCollapsiblePanel.java
@@ -23,7 +23,7 @@ class PlacementUnitsCollapsiblePanel {
     unitsToPlacePanel =
         new SimpleUnitPanel(
             uiContext, SimpleUnitPanel.Style.SMALL_ICONS_WRAPPED_WITH_LABEL_WHEN_EMPTY);
-    panel = new CollapsiblePanel(unitsToPlacePanel, "Units To Place");
+    panel = new CollapsiblePanel(unitsToPlacePanel, "Purchased Units");
     panel.setVisible(false);
     gameData.addGameDataEventListener(GameDataEvent.GAME_STEP_CHANGED, this::updateStep);
   }


Cernelius:
I disagree.

Beside the obvious fact that units that you didn't buy during a "purchase" phase are certainly not purchased (you can also start the game with units in inventory) a unit doesn't stop being purchased when you place it somewhere. The "purchased units" are, logically, all units that you purchased. For example, an infantry I buy an place on round 1 is still a purchased unit on round 20 if it has never been removed. If you play a game that starts with no units on the board and you can add units only by purchasing them, then all units (on or off the board) are purchased, always.

What distinguished those units is not the fact that they are purchased. It is the fact that they are in the game but not on the map (gameboard).

This change is calling something the wrong way, so I'm against it.

Cernelius:
If "units to place" cannot be left alone, here are some suggestions:

-unplaced units
-units off the map
-units waiting for placement
-units held in inventory

---

Personally, however, I would rather let "units to place" be, and change "units to move" to something else. While all rules oblige you to place as many units as you can (though this is not supported by TripleA), you are normally not obliged to move any of them, thus the label "units to move" is, in my opinion, a poor wording, as it seems to imply that such units need to be moved.

I would change "units to move" to something like "moveable units" and let "units to place" be or change it to one of my suggestions.

---

Finally, when you are in the actual phase, during which you are supposed to place units on the map, it should rather say "units left to place", instead of only "units to place", as long as such units are displayed therein only pending placement. This is already the case, so I'm fine with the current state also on this regard.

DanVanAtta:
> This change is calling something the wrong way, so I'm against it.

I really do think you are digging into some pedantic arguments @Cernelius . The "Place" phase is short for "Place purchased units", that is what you are placing. Are you against the term placements too since that is everything you have ever placed?

What it comes down to is the current text is bad. It makes the panel sound like you are placing those units now, not in the future. The verb tense is *wrong*.


> Beside the obvious fact that units that you didn't buy during a "purchase" phase are certainly not purchased (you can also start the game with units in inventory) 

Obviously you purchased them for zero cost, or they were pre-purchased for you, or they were a bonus-gift purchase. Just like a 'buy one get one free' coupon.

> The "purchased units" are, logically, all units that you purchased. 

Sure, but the units in the purchase panel, or placements panel, or only those that you purchased during the purchase phase and are going to place in the placement phase. If you really want to define 'purchases' this way in your mind, you are welcome to it, I believe that is a stretch and an intentional wrong definition you are using.

---------------


Thank you for the suggestions. Unfortunately this is a difficult one to phrase. I'm okay with a pedantic reasoning for it to be wrong or wrong 5% of the time if we have a better description.

`-unplaced units`

Tense is wrong, that is present tense. The units, when viewed in the 'purchased placements' panel are not unplaced in that round. They are to-be placed (future). Beyond that, unplaced units are those that you fail to place.

`-units off the map`

That sounds like units that you would need to scroll to see.

`-units waiting for placement`

This is not bad, but wordy.

`-units held in inventory`

Inventory is a new term. "Placement" and "purchase" are the phase names and concepts we have existing.

DanVanAtta:
> Finally, when you are in the actual phase, during which you are supposed to place units on the map, it should rather say "units left to place", instead of only "units to place", as long as such units are displayed therein only pending placement. This is already the case, so I'm fine with the current state also on this regard.

The placements panel disappears during the placement phase. Instead we have the old behavior that then calls them "placements", which is accurate. I don't think it's that hard to just accept that you 'place purchases'. It's really a definition thing, it just means whatever we want it to and hopefully the definitions are easy enough that they are meaningful and well understood without having to define terms using 5 words or provide every player a glossary of terms.

DanVanAtta:
> For example, an infantry I buy an place on round 1 is still a purchased unit on round 20 if it has never been removed. 

I think this is a stretch @Cernelius . Your purchase are the unplaced units. Once you place them, they go out of the pool of purchased units and then are placed units. Once placed, nothing distinguishes those units from the 'initial placements'.

The pedantic argument is boring, let's get more suggestions or go with the better verbage. "Units to place" next to another panel called "units to move" highlights how poor of phrasing "units to place" is. As mentioned in the overview, that is an active tense, it makes it sound like you are placing those units, right now.

DanVanAtta:
Having a term that is pedantically incorrect but makes sense is better than one that is confusing. "Units to Place" is just bad because the verb tense is wrong. The description provides insight into this reasoning, for which any disagreements ought to be specific vs a pet-peeve to not call placements purchases. Improvements are welcome, but we've bike-shedded this to death already.

Cernelius:
>Once you place them, they go out of the pool of purchased units and then are placed units.

I believe this is logically wrong. Placing a unit on the map doesn't change its state from "purchased" to "non-purchased".

Cernelius:
> `-unplaced units`
> 
> Tense is wrong, that is present tense. The units, when viewed in the 'purchased placements' panel are not unplaced in that round. They are to-be placed (future). Beyond that, unplaced units are those that you fail to place.

I would agree that "non-placed units" is semantically preferable.

DanVanAtta:
> Placing a unit on the map doesn't change its state from "purchased" to "non-purchased".

If you save a game, then re-load it, nothing will tell you what is original vs purchased. Furthermore, the game engine does not even keep track of units that well, if you mix them in a unit stack they'll be mixed together. If you want to consider a 'purchased' units to be a 'purchased' unit beyond placement, that is up to you. You can also decide that a purchased unit means simply units you have purchased but not yet placed. IE: the unit is purchased and unplaced, then once placed it is simply placed.

> I would agree that "non-placed units" is semantically preferable.

Sure, though it's difficult to get the tense right to make it clear they are units for future phases and not the current. 

I am considering "placements" instead of "purchased units"

PR #8250, opened 2020-11-24
Use a value object to combine the rolls and strength in power calculation

trevan:
Rolls and strength values are intertwined with each other when dealing
with power calculation.  If one is zero, then the other is zero. Instead
of keeping track of these details in the callers, the StrengthAndRolls
value object will ensure it.

Also, pass around the RollValue and StrengthValue as much as possible so
that it can be type checked and the information they store can be kept
through changes.


DanVanAtta:
@trevan , would you mind rebasing to the latest to pick up the github workflow actions (and please confirm that the integration tests still pass via `verify`).

trevan:
I ran `verify` and everything passed.

trevan:
@DanVanAtta , it doesn't seem to be running the github actions.

DanVanAtta (approved):

PR #8262, opened 2020-11-25
Create BombingUnitDamageChange that uses primitives

trevan:
The old BombingUnitDamageChange is marked as deprecated and will stay
around solely for loading old saves.

The new BombingUnitDamageChange also has been modified to accept the
territory that needs to be notified, instead of looping through all the
game territories during the change process.

## Testing

Conducted a strategic bombing run in WW2 Global 1940.  Verified the change was correctly built.  Verified that it "performed" correctly.  Played till the attacked player could repair the damaged units and verified that the change was correctly built.

I created a new Change object and marked the other as deprecated.  I think this might be cleaner since the new change object can have a new name and it doesn't need to hold the non-primitive objects.  This definitely breaks forward compatibility but the compatibility check should prevent users from trying to load saves with this new change object in an older version.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/changefactory/units/BombingUnitDamageChange.java
+++ b/game-core/src/main/java/games/strategy/engine/data/changefactory/units/BombingUnitDamageChange.java
@@ -1,0 +1,71 @@
+package games.strategy.engine.data.changefactory.units;
+
+import games.strategy.engine.data.Change;
+import games.strategy.engine.data.GameData;
+import games.strategy.engine.data.Territory;
+import games.strategy.engine.data.Unit;
+import java.util.Collection;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import org.triplea.java.collections.IntegerMap;
+
+/**
+ * A game data change that captures the damage caused to a collection of units by a bombing attack.
+ */
+public class BombingUnitDamageChange extends Change {
+  private static final long serialVersionUID = 7093658184880237574L;
+  private final IntegerMap<String> hits;
+  private final IntegerMap<String> undoHits;
+  private final Collection<String> territoriesToNotify;
+
+  private BombingUnitDamageChange(
+      final IntegerMap<String> hits,
+      final IntegerMap<String> undoHits,
+      final Collection<String> territoriesToNotify) {
+    this.hits = hits;
+    this.undoHits = undoHits;
+    this.territoriesToNotify = territoriesToNotify;
+  }
+
+  public BombingUnitDamageChange(
+      final IntegerMap<Unit> hits, final Collection<Territory> territoriesToNotify) {
+    this.hits = new IntegerMap<>();
+    this.undoHits = new IntegerMap<>();
+    hits.entrySet()
+        .forEach(
+            entry -> {
+              this.hits.put(entry.getKey().getId().toString(), entry.getValue());
+              this.undoHits.put(entry.getKey().getId().toString(), entry.getKey().getUnitDamage());
+            });
+    this.territoriesToNotify =
+        territoriesToNotify.stream().map(Territory::getName).collect(Collectors.toList());
+  }
+
+  @Override
+  protected void perform(final GameData data) {
+    hits.keySet()
+        .forEach(
+            unitId -> {
+              data.getUnits().get(UUID.fromString(unitId)).setUnitDamage(hits.getInt(unitId));
+            });
+    this.territoriesToNotify.forEach(
+        territory -> {
+          data.getMap().getTerritory(territory).notifyChanged();
+        });
+  }
+
+  @Override
+  public Change invert() {
+    return new BombingUnitDamageChange(undoHits, hits, territoriesToNotify);
+  }
+
+  @Override
+  public String toString() {
+    return "Bombing unit damage change. Hits:"
+        + hits
+        + " undoHits:"
+        + undoHits
+        + " territories:"
+        + territoriesToNotify;
+  }
+}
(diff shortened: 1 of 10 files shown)


trevan:
The only difference between the old BombingUnitDamageChange class and the new one is the primitive usage.

game-core/src/main/java/games/strategy/engine/data/changefactory/units/BombingUnitDamageChange.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.stream.Collectors;
+import org.triplea.java.collections.IntegerMap;
+
+/**
+ * A game data change that captures the damage caused to a collection of units by a bombing attack.
+ */
+public class BombingUnitDamageChange extends Change {
+  private static final long serialVersionUID = 7093658184880237574L;
+  private final IntegerMap<String> hits;
+  private final IntegerMap<String> undoHits;
```
DanVanAtta, line 18: What is undo hits representing?

game-core/src/main/java/games/strategy/engine/data/changefactory/units/BombingUnitDamageChange.java
```diff
@@ -0,0 +1,71 @@
...
+  private final Collection<String> territoriesToNotify;
+
+  private BombingUnitDamageChange(
```
DanVanAtta, line 21: Any thoughts to use an `@AllArgsConstructor(visibility = AccessLevel.PRIVATE)` here to avoid the manually written out constructor?

game-core/src/main/java/games/strategy/engine/data/changefactory/units/BombingUnitDamageChange.java
```diff
@@ -0,0 +1,71 @@
...
+        });
+  }
+
+  @Override
+  public Change invert() {
+    return new BombingUnitDamageChange(undoHits, hits, territoriesToNotify);
+  }
+
+  @Override
+  public String toString() {
```
DanVanAtta, line 63: Is the 'toString' required? Is it used anywhere to print text for users? If not, is it helpful?

game-core/src/main/java/games/strategy/engine/data/changefactory/units/BombingUnitDamageChange.java
trevan: undoHits represents the inverse of what is happening in this change.  Each change object has an "undo" version of it.

game-core/src/main/java/games/strategy/engine/data/changefactory/units/BombingUnitDamageChange.java
trevan: I could.  This is just copied code so I didn't make any major changes except the non-primitive -> primitive change.

game-core/src/main/java/games/strategy/engine/data/changefactory/units/BombingUnitDamageChange.java
trevan: No idea.  Several change objects have a toString.  Several don't.

game-core/src/main/java/games/strategy/engine/data/changefactory/units/BombingUnitDamageChange.java
trevan: The change objects call it "invert".  You can see it being used in the `invert` method.

trevan:
@DanVanAtta , I used the `@AllArgsConstructor` and answered your other questions.

trevan:
The build failed because of a failure in downloading a file.

game-core/src/main/java/games/strategy/engine/data/changefactory/units/BombingUnitDamageChange.java
DanVanAtta: Moving code is one thing, in this case we are re-writing and in my 2 cents treat this is as new code.

game-core/src/main/java/games/strategy/engine/data/changefactory/units/BombingUnitDamageChange.java
DanVanAtta: What would you think to rename:
 'hits' -> 'newHits' | 'hitsAfterChange'
'undoHits' -> 'oldHits' | 'hitsBeforeChange' | 'originalHits'

It looks like other change objects use 'new' & 'old' for naming. The 'undoHits' is not the best name, not really clear whether it is the original hits or a delta to compute the original hits.

game-core/src/main/java/games/strategy/engine/data/changefactory/units/BombingUnitDamageChange.java
DanVanAtta: A lot of 'toStrings' are included to help debugging in Eclipse that used to show the 'toString' of an object when in debug. That was before when you could inspect the values of an object and you'd see just the 'toString' value.

If you play a game with a break-point in this method you'll probably be able to determine if this is dead code relatively quickly. IMO we should not be porting dead-code into new code if we can help it. When we do a new release I don't think we'll want to go back and do any clean up beyond carefully deleting that which is marked for deletion.

trevan:
@DanVanAtta , I saw this as just a code move since the only reason the old file still exists is because of save compatibility.  But if you want to see it as re-writing, that's fine.  I've renamed the variables to `oldDamage` and `newDamage`, removed the `toString` (the debugger never stopped in it), and added some comments to explain what the value of `damage` should be.

DanVanAtta:
Thanks for the updates @trevan .  I'm pretty positive when we remove the old code that we're not going to then improve this moved and slightly changed version, which implies to me now is the best time to make it as reasonable as we can (just as would generally be asked for anything that is added).

DanVanAtta (approved):

PR #8274, opened 2020-11-26
Rename collapsible panel titles

DanVanAtta:
'Purchased Units' -> 'Placements': Helps beter match the phase
and what the units are.

'Units to move' -> 'Active Units': Less wordy and better
describes what the units are within the panel.

Screenshot from 2020-11-25 18-55-40


DanVanAtta:
cc: @Cernelius  , you're likely interested in this update.

PR #8275, opened 2020-11-26
Rename 'map options' and 'select map' to 'game options'

DanVanAtta:
'Game' is more fitting in our vernacular to refer to a specific
game XML while a map is a collection of games and graphics.

Screenshot from 2020-11-25 18-58-43

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/network/ui/ChangeGameOptionsClientAction.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/network/ui/ChangeGameOptionsClientAction.java
@@ -46,7 +46,7 @@ public class ChangeGameOptionsClientAction extends AbstractAction {
       final String cancel = "Cancel";
       pane.setOptions(new Object[] {ok, cancel});
       final JDialog window =
-          pane.createDialog(JOptionPane.getFrameForComponent(parent), "Map Options");
+          pane.createDialog(JOptionPane.getFrameForComponent(parent), "Game Options");
       window.setVisible(true);
       final Object buttonPressed = pane.getValue();
       if (buttonPressed != null && !buttonPressed.equals(cancel)) {
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorPanel.java
@@ -65,14 +65,14 @@ public final class GameSelectorPanel extends JPanel implements Observer {
           .build();
   private final JButton loadNewGame =
       new JButtonBuilder()
-          .title("Select Map")
+          .title("Select Game")
           .toolTip(
               "<html>Select a game from all the maps/games that come with TripleA, "
                   + "<br>and the ones you have downloaded.</html>")
           .build();
   private final JButton mapOptions =
       new JButtonBuilder()
-          .title("Map Options")
+          .title("Game Options")
           .toolTip(
               "<html>Set options for the currently selected game, <br>such as enabling/disabling "
                   + "Low Luck, or Technology, etc.</html>")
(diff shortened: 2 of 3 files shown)


Cernelius:
This was unexpected and, of course, I agree (if, at any point in the future, the terminology will be changed (changing "map" or "game", or both, to some other terms), everything can change, but, until then, I think it is normal practice to keep everything consistent).

On this regard, if wanting to be omnicomprensive, also the information above should have the changes:

"Map Name"->"Game Name"
"Map Version"->"Game Version"

In theory, you could also have a field for the "Map Name", getting the name of the map folder (and maybe substituting underscores with spaces for display), but that is probably something not clearly valuable to anyone but experienced mapmakers, that won't really have much need for it, anyways.

---

Not necessary, but I would also put the program's version immediately under the TripleA logo, as that is the version of TripleA. So, instead of having:

"LOGO"
Java Version: 11.0.9.1
Engine Version: 2.6.0

I would have:

"LOGO"
Program Version: 2.6.0
Program Language Name: Java
Program Language Version: 11.0.9.1

---

For reference, this is TripleA 1.8.0.9:

DanVanAtta:
Good idea to also change the 'map' -> 'game' on the front screen. 

@Cernelius you open a can of worms with the other changes that need some consideration.

> Program Version: 2.6.0
Program Language Name: Java
Program Language Version: 11.0.9.1

Program is ambiguous, 'Engine' is more specific. I'd also say the programming language is a detail, players opening the game do not care, they want to find the button that says "start a game", we should remove all extraneous information that detracts from that goal and any other goal a player would have after launching the application. To that extent, I'm really having thoughts about moving the java version and engine version into an about dialog or removing altogether.

DanVanAtta:
@Cernelius the other changes noted and discussed are independent updates to the main thrust of this update. We want code updates to be single purpose and instead have a number of code updates. It helps for a number of reasons that I won't go into. If you really recommend we change up things like where the java version is reported, we should move that conversation to an issue or forums.

FrostionAAA:
I can’t see the logic in changing the phrase “Select Map” into “Select Game”?

In my mind, when one is navigating Triple A (or any other game program) and then talks about “a new game”, “a saved game” and “load a game”, one is referring to a specific session or play through, not a map.

When I see “Select Game” I think of a list of ongoing or soon-starting multiplayer game sessions. I would argue that it is a universal truth, that everyone knows what “Select Map” means, just as people know what “Download Maps” and “Map Options” mean. Because of this, I think the buttons should be kept as “Select Map” and “Map Options”.

The only place I can see information about this coming “phrase”-change is right here? How come stuff like this is not discussed on the forum and presented to more than a select few people scoping around Github, before being implemented in a new stable release? Maybe I have just not been hanging around the forum on the exact day and therefore missed a post.

DanVanAtta:
I'm starting to think the split between forum and github communication is more and more unsustainable. A fundamental problem is that this a volunteer effort. Feedback needs to arrive within hours or days, getting feedback a month later is not practical. Development is fast and small iteration work, when we have time, we have it and context. It's like leaving your shop in the garage for three months while you ask everyone what kind of tires should be put on.

> I can’t see the logic in changing the phrase “Select Map” into “Select Game”?

This is the root of the most pedantic argument I have ever conceived of. When TripleA was originally made the developers decided to call a "map" the sum contents of a map folder. An XML file was then called a "game". This made more sense when there were multiple game-types available. This terminology was baked in and then exposed to and picked up by map makers.

That terminology DOES NOT match what players think of as games and maps, and to players a 'map' is a what map-maker would call a 'game', and a 'game' is a 'game-in-progress'. This mismatch in terminology is a problem. There is extended discussion on this topic here: https://github.com/triplea-game/triplea/issues/6994

The crux of the problem I think is that it's hard to convince map makers that the 'game' and 'map' terminology depend on their context and that players do not make (nor even care about) this distinction other than that you 'play a map' and 'continue a game'.

This change was made to at least use the 'game' and 'map' terms consistently on the same screen. The rest of the game-client still uses 'game' to refer to XML files rather than to 'save-games'.

DanVanAtta:
Adding a bit, the real difficulty is that players don't know that maps and games are one-to-many. To the average player, the fact that map graphics are shared is transparent, to them a game and map are 1:1. This is in part why the terminology is so painful, the game engine is exposing pretty low-level deatils in an unnecessary way to map makers and users.

Perhaps the original terms of 'map' and 'scenario' would have been better to use and then recognized that those are effectively the same thing to the average player (a 1:1 relationship). When (one fine day) maps and XML files are no longer in the same folder module, it will be a good time to redo this terminology and respect that players think of 'maps' as the XML files and not just as the map graphics.

Issue #8266, opened 2020-11-25
Hard/Easy AI problem with restricting canal attachments


FrostionAAA:
**The Issue:**
The Hard / Easy AI cannot handle and make use of territory connection with canal attachments that only allows certain unit types traverse the canal.

Canals like this one from Warcraft War Heroes, that only allows a group of air units to cross:
```
        <attachment name="canalAttachmentCliffspring - Big Foot Mountains" attachTo="Cliffspring" javaClass="games.strategy.triplea.attachments.CanalAttachment" type="territory">
          <option name="canalName" value="a bridge between Cliffspring and Big Foot Mountains"/>
          <option name="landTerritories" value="Hero Bar"/>
          <option name="excludedUnits" value="$AllAirwayCrossingUnits$"/>
        </attachment>
        <attachment name="canalAttachmentCliffspring - Big Foot Mountains" attachTo="Big Foot Mountains" javaClass="games.strategy.triplea.attachments.CanalAttachment" type="territory">
          <option name="canalName" value="a bridge between Cliffspring and Big Foot Mountains"/>
          <option name="landTerritories" value="Hero Bar"/>
          <option name="excludedUnits" value="$AllAirwayCrossingUnits$"/>
        </attachment>
```

I have here a save game made in the current and newest v0.1.6 Warcraft Warheroes map. It has AI stacking up at all four Horde capitals, as they all have territory connections with restricing canals / blue arrows:

[MAPv0.1.6-AI-Airways.zip](https://github.com/triplea-game/triplea/files/5597791/MAPv0.1.6-AI-Airways.zip)

As this screenshot shows, the AI starts stacking up units at several locations, notably at the Horde players’ capitals, as these places all have “Airways” / such restricting canals, visually marked with blue arrows.

![1](https://user-images.githubusercontent.com/15268177/100251265-db615900-2f3e-11eb-9a64-408ffc160828.png)

**Why is it stacking up units?:**
My theory (without being able to decipher savegames and cryptic code) is that the AI actually wants, tries and fails to move the units via connections with canals … fails as only Air may cross … and then just leaves the units standing there. The AI can have units standing still at the canals for many rounds, but sometimes after many rounds the map situation makes the AI finally make another decision and moves the same units via another route. 

**What should happen instead?:**
I guess that AI should determine what kinds of units can / cannot make use of that individual canal while figuring out how to proceed with CM and NCM.
[...]

PR #8289, opened 2020-11-26
Allow validating canals when finding neighbors or distances

trevan:
The ProAI calculates the value of territories based on neighboring
territories and distances to enemy factories/capitals. If a neighboring
territory had a canal to it, the ProAI would ignore the canal and assume
that the units could travel to it. This could cause unit stacking
because of the higher value of the territory.

Now, when the ProAI calculates the neighboring territories, it checks if
units can travel between between the two territories. This will cause
some neighboring territories to have a higher distance and so the value
of the territory will go down.

## Testing

Played the save from #8266 and verified that the stacked units moved.  Played 7 rounds of a Hard AI warcraft heroes game and ensured there were no stacked units or errors.  Played 3 rounds of a Hard AI WW2 Global 1940 game to ensure no errors.

## Release Note

FIX|AI incorporates canal restrictions when calculating the value of its territories.  This value is used by the AI to determine where units should be moved.  Before this change, if an unopened canal separated two territories and one territory was owned by the AI an the other territory was a high value territory (such as enemy capital), then it would stack up units in that territory since the units couldn't traverse the canal but the value of the territory was really high.  AI should handle now Warcraft War Heroes air routes better.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/GameMap.java
+++ b/game-core/src/main/java/games/strategy/engine/data/GameMap.java
@@ -16,6 +16,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.BiPredicate;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
@@ -107,11 +108,15 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
    * Returns all adjacent neighbors of the starting territory that match the condition. Does NOT
    * include the original/starting territory in the returned Set.
    */
-  public Set<Territory> getNeighbors(final Territory territory, final Predicate<Territory> cond) {
+  public Set<Territory> getNeighbors(
+      final Territory territory,
+      @Nullable final Predicate<Territory> cond,
+      final BiPredicate<Territory, Territory> routeCond) {
     if (cond == null) {
       return getNeighbors(territory);
     }
     return connections.getOrDefault(territory, Set.of()).parallelStream()
+        .filter(n -> routeCond.test(territory, n))
         .filter(cond)
         .collect(Collectors.toSet());
   }
@@ -121,25 +126,34 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
    * condition. Does NOT include the original/starting territory in the returned Set.
    */
   public Set<Territory> getNeighbors(final Territory territory, final int distance) {
-    return getNeighbors(territory, distance, Matches.always());
+    return getNeighbors(territory, distance, Matches.always(), Matches.alwaysBi());
   }
 
   /**
    * Returns all neighbors within a certain distance of the starting territory that match the
    * condition. Does NOT include the original/starting territory in the returned Set.
+   *
+   * @param territory starting territory
+   * @param distance distance to check
+   * @param cond the neighbor territory must match this condition
+   * @param routeCond the route between the starting territory and the neighbor territory must match
+   *     this condition
    */
   public Set<Territory> getNeighbors(
-      final Territory territory, final int distance, final Predicate<Territory> cond) {
+      final Territory territory,
+      final int distance,
+      @Nullable final Predicate<Territory> cond,
+      final BiPredicate<Territory, Territory> routeCond) {
     checkArgument(distance >= 0, "Distance must be non-negative: " + distance);
     if (distance == 0) {
       return Set.of();
     }
-    final Set<Territory> neighbors = getNeighbors(territory, cond);
+    final Set<Territory> neighbors = getNeighbors(territory, cond, routeCond);
     if (distance == 1) {
       return neighbors;
     }
     final Set<Territory> result =
-        getNeighbors(neighbors, new HashSet<>(neighbors), distance - 1, cond);
+        getNeighbors(neighbors, new HashSet<>(neighbors), distance - 1, cond, routeCond);
     result.remove(territory);
     return result;
   }
@@ -152,7 +166,7 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
   public Set<Territory> getNeighbors(
       final Set<Territory> frontier, final int distance, final Predicate<Territory> cond) {
     final Set<Territory> neighbors =
-        getNeighbors(frontier, new HashSet<>(frontier), distance, cond);
+        getNeighbors(frontier, new HashSet<>(frontier), distance, cond, Matches.alwaysBi());
     neighbors.removeAll(frontier);
     return neighbors;
   }
@@ -161,18 +175,19 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
       final Set<Territory> frontier,
       final Set<Territory> searched,
       final int distance,
-      @Nullable final Predicate<Territory> cond) {
+      @Nullable final Predicate<Territory> cond,
+      final BiPredicate<Territory, Territory> routeCound) {
     if (distance == 0 || frontier.isEmpty()) {
       return searched;
     }
     final Set<Territory> newFrontier =
         frontier.stream()
-            .map(t -> getNeighbors(t, cond))
+            .map(t -> getNeighbors(t, cond, routeCound))
             .flatMap(Collection::stream)
             .filter(t -> !searched.contains(t))
             .collect(Collectors.toSet());
     searched.addAll(newFrontier);
-    return getNeighbors(newFrontier, searched, distance - 1, cond);
+    return getNeighbors(newFrontier, searched, distance - 1, cond, Matches.alwaysBi());
   }
 
   /**
@@ -190,7 +205,7 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
     if (distance == 1) {
       return neighbors;
     }
-    final Set<Territory> start = getNeighbors(territory, cond);
+    final Set<Territory> start = getNeighbors(territory, cond, Matches.alwaysBi());
     for (int i = 2; i <= distance; i++) {
       neighbors.addAll(getNeighborsIgnoreEnd(start, new HashSet<>(start), i - 1, cond));
     }
@@ -209,7 +224,7 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
     final Predicate<Territory> neighborCond = (distance == 1) ? Predicates.alwaysTrue() : cond;
     final Set<Territory> newFrontier =
         frontier.stream()
-            .map(t -> getNeighbors(t, neighborCond))
+            .map(t -> getNeighbors(t, neighborCond, Matches.alwaysBi()))
             .flatMap(Collection::stream)
             .filter(t -> !searched.contains(t))
             .collect(Collectors.toSet());
@@ -235,12 +250,17 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
     if (movementLeft.compareTo(BigDecimal.ZERO) == 0) {
       return Set.of();
     }
-    final Set<Territory> neighbors = getNeighbors(territory, cond);
+    final Set<Territory> neighbors = getNeighbors(territory, cond, Matches.alwaysBi());
     if (movementLeft.compareTo(BigDecimal.ONE) <= 0) {
       return neighbors;
     }
     final Set<Territory> result =
-        getNeighbors(neighbors, new HashSet<>(neighbors), movementLeft.intValue() - 1, cond);
+        getNeighbors(
+            neighbors,
+            new HashSet<>(neighbors),
+            movementLeft.intValue() - 1,
+            cond,
+            Matches.alwaysBi());
     result.remove(territory);
     return result;
   }
@@ -305,7 +325,7 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
    * @param t2 end territory of the route
    */
   public int getDistance(final Territory t1, final Territory t2) {
-    return getDistance(t1, t2, Matches.always());
+    return getDistance(t1, t2, Matches.always(), Matches.alwaysBi());
   }
 
   /**
@@ -315,12 +335,17 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
    * @param t1 start territory of the route
    * @param t2 end territory of the route
    * @param cond condition that covered territories of the route must match
+   * @param routeCond condition that routes between neighboring territories must match
    */
-  public int getDistance(final Territory t1, final Territory t2, final Predicate<Territory> cond) {
+  public int getDistance(
+      final Territory t1,
+      final Territory t2,
+      final Predicate<Territory> cond,
+      final BiPredicate<Territory, Territory> routeCond) {
     if (t1.equals(t2)) {
       return 0;
     }
-    return getDistance(0, new HashSet<>(), Set.of(t1), t2, cond);
+    return getDistance(0, new HashSet<>(), Set.of(t1), t2, cond, routeCond);
   }
 
   /**
@@ -333,16 +358,16 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
       final Set<Territory> searched,
       final Set<Territory> frontier,
       final Territory target,
-      @Nullable final Predicate<Territory> cond) {
+      final Predicate<Territory> cond,
+      final BiPredicate<Territory, Territory> routeCond) {
     // add the frontier to the searched
     searched.addAll(frontier);
     // find the new frontier
 
     final Set<Territory> newFrontier =
         frontier.stream()
-            .map(connections::get)
-            .flatMap(Collection::stream)
-            .filter(f -> cond == null || cond.test(f))
+            .flatMap(f -> connections.get(f).stream().filter(t -> routeCond.test(f, t)))
+            .filter(cond)
             .collect(Collectors.toSet());
     if (newFrontier.contains(target)) {
       return distance + 1;
@@ -351,7 +376,7 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
     if (newFrontier.isEmpty()) {
       return -1;
     }
-    return getDistance(distance + 1, searched, newFrontier, target, cond);
+    return getDistance(distance + 1, searched, newFrontier, target, cond, routeCond);
   }
 
   public IntegerMap<Territory> getDistance(
@@ -363,7 +388,7 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
       return distances;
     }
     for (final Territory t : territories) {
-      distances.put(t, getDistance(target, t, condition));
+      distances.put(t, getDistance(target, t, condition, Matches.alwaysBi()));
     }
     return distances;
   }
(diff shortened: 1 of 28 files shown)


DanVanAtta:
Thanks for digging into this @trevan . Have you tested how AI behaves on other WWII games w.r.t canals?

Could an AI be tricked now that it is willing to cross canals? For example, AI approaches canal but does not realize it can't hold both ends of the canal. A player keeps a token force on one side to then capture the canal. Would the AI then attempt to go the long way around the canal or would it then start stacking? If the updated behavior would still be better than the existing, that is all good. Though, I would not under-estimate the possibility for unintended side-effects here. I think we should really hammer this one with the testing to validate the updated strategy is with certainty a net benefit on all maps and doesn't lead to catastrophically bad play on others.

trevan:
@DanVanAtta , do you know of other maps that have land canals?  I only know of maps that have sea canals and this wouldn't affect those maps.

As for your concerns, I think you have the behavior backwards.  Currently, the AI ignores canal restrictions when it calculates the value of territories.  So, currently the AI would have the issue that you just outlined.  It wouldn't recognize that it can't cross the canal and would stack up units if that territory had a better value.  The change that I did makes it actually see the canal restrictions when it calculates the value of the territory.  If it can't cross the canal, then the value of the territory is lower and so it won't want to send units there.

DanVanAtta:
Not aware of other maps with land canals. Thanks for clarification on
behavior.
Do you know if AI puts any weight in trying to own a canal?

trevan:
@DanVanAtta , I can't see any code that tries to own canals.  The method `CanalAttachment#getLandTerritories` is how to find out which territories control the canal and the ai code never calls that method.

DanVanAtta (commented):
Few comments, let's see if we can improve this update a bit more. I think we should avoid injecting the bi-predicate into the neighbor function. There are a few notable items about the bi-predicate:
- it's not clear that 'alwaysBi' is short for 'alwaysTrueBiPredicate'
- while passing in a predicate to a method makes it more flexible, I think it breaks the usability and modularity of the function. If anyone wants to get territory neighbors, they have to wonder what the bi-predicate is and then analyze that it's a filtering condition. At that point the person trying to use the method is reading the source code of the method to understand how it works (rather than it just being clear how it would work from it's name, signature, and javadoc). 

There are a few options to potentially fix this:
1. Create an overloaded method signature so that most callers do not have to know that you can pass in a bi-predicate
2. (preferred) have the caller that needs to do the filtering, do the filtering
3. Extract functions to get neighbor territories to a new object, living with the data (similar to ProTerritory), and then create a first class method that encapsulates the needed predicates.

trevan:
> Few comments, let's see if we can improve this update a bit more. I think we should avoid injecting the bi-predicate into the neighbor function. There are a few notable items about the bi-predicate:
> 
>     * it's not clear that 'alwaysBi' is short for 'alwaysTrueBiPredicate'
> 
>     * while passing in a predicate to a method makes it more flexible, I think it breaks the usability and modularity of the function. If anyone wants to get territory neighbors, they have to wonder what the bi-predicate is and then analyze that it's a filtering condition. At that point the person trying to use the method is reading the source code of the method to understand how it works (rather than it just being clear how it would work from it's name, signature, and javadoc).
> 
> 
> There are a few options to potentially fix this:
> 
>     1. Create an overloaded method signature so that most callers do not have to know that you can pass in a bi-predicate
> 
>     2. (preferred) have the caller that needs to do the filtering, do the filtering
> 
>     3. Extract functions to get neighbor territories to a new object, living with the data (similar to ProTerritory), and then create a first class method that encapsulates the needed predicates.

@DanVanAtta , an overloaded method for `getNeighbor` already exists that removes the need to pass in a condition.  It's signature is just `public Set<Territory> getNeighbors(final Territory territory, final int distance)`.  The `getNeighbors` method that takes in the condition allows you to be more targeted.  There are two types of conditions that can be passed in: one that checks each territory and the new one that allows you to check the connection between each territory.

So, I don't think we should make any changes.  And as for your potential fixes, I don't think 2 or 3 are worth it.  I tried to implement something like that originally but in the end, you need the logic that is contained in `getNeighbors`.

DanVanAtta:
@trevan I see the method invocation looks like this almost everywhere now:
```
       getNeighbors(frontier, new HashSet<>(frontier), distance, cond, Matches.alwaysBi());
```

This is what I'm concerned about,  the API is now unwiedly, it's not clear what this does by reading it. It also looks like we are wiring in no-op (always true) functionality everywhere rather than pursuing another technique. 


If we are almost always doing a 'Matches.alwaysBi()' everywhere, and we only need the bi-predicte in one or two places, it's time to duplicate the getNeighbors method. It also looks like getNeighbors has exceedingly little code in it, it's a very shallow class & abstraction (which is a bad thing, methods should be deep and do a lot, not require the caller to read the implementation to then configure a strategy object that does what is needed).

trevan:
@DanVanAtta I've added two more additional overloaded versions of getNeighbors.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
```diff
@@ -104,6 +104,11 @@ public Territory getTerritory(final String s) {
     return neighbors;
   }
 
+  public Set<Territory> getNeighbors(
+      final Territory territory, @Nullable final Predicate<Territory> cond) {
+    return getNeighbors(territory, cond, Matches.alwaysBi());
```
DanVanAtta, line 109: Please rename `Matches.alwaysBi()`. I had a lot of trouble reading that at first, it was not clear what it meant (even the type being passed was not clear, so it looked like 'Bi' was an abbreviation for something').

The implementation for an always true `bi-predicate` is kinda trivial, I'm not sure we really need a matches for it and can just inline it. I don't think we should be encouraging usages of tautology functions by having it be in `Matches`. A tautology funtion makes methods look a lot more complex then they actually are.

trevan:
I've also added an overloaded method for getDistance

DanVanAtta (commented):
Looks better. Please have a look at renaming or inlining the `alwaysBi` matcher

game-core/src/main/java/games/strategy/engine/data/GameMap.java
trevan: @DanVanAtta, do you also want to get rid of `Matches.always()`.  I thought that it would be better to have something similar for a `bi-predicate`.

game-core/src/main/java/games/strategy/triplea/delegate/move/validation/MoveValidator.java
```diff
@@ -278,13 +279,13 @@ private MoveValidationResult validateCanal(
...
   public String validateCanal(
-      final Route route, final Collection<Unit> units, final GamePlayer player) {
+      final Route route, @Nullable final Collection<Unit> units, final GamePlayer player) {
     return validateCanal(route, units, new HashMap<>(), player);
   }
 
   String validateCanal(
       final Route route,
-      final Collection<Unit> units,
+      @Nullable final Collection<Unit> units,
```
DanVanAtta, line 288: Just noticed, when is the `units` value null? Generally an empty collection is preferred rather than a `null`.

trevan: I don't know.  But one of the callers to this method has it flagged as `@Nullable` so I flagged this one as well since the caller can pass a null value.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
DanVanAtta: Thanks for asking - `Matches.always()` might not be best to inline just yet. Eventually perhaps. Hiding the always true predicate in the implementation class I think is the thing to do. If we can, also hide additional predicates in that class so the callers are just invoking a function and not passing in a predicate. The predicate object is a bit too low level. Consider for example using a method that iterates over territories and requires a predicate, you really have to look at the implementation details to understand what the predicate is going to do exactly. Then if we change that implementation all of the expectations of the callers are broken. I think it's an example of relatively tight coupling and lack of information hiding.

TL;DR - simple inlining of `Matches.always()` could do more harm than good if it is used pervasively. Instead we should seek to first hide as much of the predicate logic in the implementation code to the point where the predicate is an implementation detail, at that time inlining it is likely appropriate.

game-core/src/main/java/games/strategy/triplea/delegate/move/validation/MoveValidator.java
DanVanAtta: Fair enough, it is better safe than sorry.  :+1:

game-core/src/main/java/games/strategy/engine/data/GameMap.java
```diff
@@ -103,15 +104,24 @@ public Territory getTerritory(final String s) {
...
+  }
+
   /**
    * Returns all adjacent neighbors of the starting territory that match the condition. Does NOT
    * include the original/starting territory in the returned Set.
    */
-  public Set<Territory> getNeighbors(final Territory territory, final Predicate<Territory> cond) {
+  public Set<Territory> getNeighbors(
+      final Territory territory,
+      @Nullable final Predicate<Territory> cond,
```
DanVanAtta, line 118: Nit - This parameter is now the 'territoryCondition', correct? Would you have any objection to rename it?

Now that there are two parameters, making the names distinct is more significant as one wonders how the 'cond' and 'routeCond' are different. Secondly, 'cond' was not the best name to begin with as it was cryptic and did not describe what the condition represented or was used for.

DanVanAtta: It looks like this method is only called internally to `GameMap` now as well, if I have that right - we can reduce the visibility to `private`. Would you mind doing so?

game-core/src/main/java/games/strategy/engine/data/GameMap.java
```diff
@@ -127,19 +137,41 @@ public Territory getTerritory(final String s) {
   /**
    * Returns all neighbors within a certain distance of the starting territory that match the
    * condition. Does NOT include the original/starting territory in the returned Set.
+   *
+   * @param territory starting territory
+   * @param distance distance to check
```
DanVanAtta, line 142: This javadoc perhaps could be better. What kind of check are we doing? What is the signifcance of the check?

Suggestion: reword to something like:  "distance All returned territories will be within the max distance from the starting territory. 0 represents the starting territory and no adjacencies, 1 is all immediately adjacent territories."

game-core/src/main/java/games/strategy/engine/data/GameMap.java
```diff
@@ -127,19 +137,41 @@ public Territory getTerritory(final String s) {
...
+   * @param cond the neighbor territory must match this condition
    */
```
DanVanAtta, line 144: Side-note, the getNeighbors code is largely only used by AI. It's something for us to consider as every logical graph operation involving the map does not scale if it's all found within `GameMap`. My thoughts are a `Territory` object having a collection of neighbors would help avoid the need to get all data from `GameMap`, and then also the AI code could have the path finding algorithms localized in the AI packages.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
```diff
@@ -127,19 +137,41 @@ public Territory getTerritory(final String s) {
...
+   * @param distance distance to check
+   * @param cond the neighbor territory must match this condition
+   * @param routeCond the route between the starting territory and the neighbor territory must match
+   *     this condition
+   */
+  public Set<Territory> getNeighbors(
+      final Territory territory,
+      final int distance,
+      @Nullable final Predicate<Territory> cond,
+      final BiPredicate<Territory, Territory> routeCond) {
```
DanVanAtta, line 164: Thank you for the previous cleanups, it's now clear that either `cond` is used with an always true bi-predicate (route condition), or if the bi-predicate is used then it is always with these two conditions:
```
                ProMatches.territoryCanPotentiallyMoveLandUnits(player, data.getProperties()),
                ProMatches.noCanalsBetweenTerritories(player, data));
```

I think there are some issues here:
1. DRY violation, the idea of how to calculate land-passable territories without canal is derived in two different places (happily the matching logic is in one spot, but still it's indirect to go from a matcher to a method call to actually getting the neighbors).
2. Makes for a difficult API to use. I don't think it's very fluent to look up a matcher and seems confusing to keep straight the difference between a route and territory condition. We also have a number of overloads which makes a person wonder how they should get neighbors if that is what they are trying to do.

What do you think we created an API that could hide the two predicates from callers?

The signature would be something like:
```
Set<Territory> getLandTraversableNeighbors(
      final Territory territory,
      final int distance);
```

Then in the javadoc we could define what is meant by land-traversable. Or we could call the method `getNonCanalLandTraversableNeighbors` if we feel the canal aspect is distinct enough to need to be in the method name. 

I would agree that having `ProMatches` in `GameMap` is bad, though I think we are already there indirectly by having entire methods (like this and similar) be in `GameMap` only to support AI. As noted earlier, the entire method likely ought to be picked up and moved to AI specific code.

Otherwise, WDYT?

Such a move should hopefully make the method overloads easier to use, reduce the API surface, reduce the number of parameters (which makes it more testable!), and just easier to understand how to use these methods.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
```diff
@@ -317,10 +355,27 @@ public int getDistance(final Territory t1, final Territory t2) {
...
+  /**
+   * Returns the distance between two territories where the covered territories of the route satisfy
+   * the condition or -1 if they are not connected.
+   *
+   * @param t1 start territory of the route
+   * @param t2 end territory of the route
+   * @param cond condition that covered territories of the route must match
+   * @param routeCond condition that routes between neighboring territories must match
+   */
+  public int getDistance(
```
DanVanAtta, line 370: Similar comment to this method, it's either invoked with a no-op arg or the arg is always:
```
                  ProMatches.territoryCanPotentiallyMoveLandUnits(player, data.getProperties()),
                  ProMatches.noCanalsBetweenTerritories(player, data));
```

That hints that we could use another method rather than it need to be parameterized as part of the public API.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
trevan: The idea sounds good but I'd rather not do it as part of this PR.  I think these types of changes need to be done after looking into all of the usages of `getNeighbor`.  I also think the canals check that I added probably needs to be used in more places than just these few places that I found.  It might turn out that it needs to be in all places that call `getNeighbor` with a territoryCondition.  But I'm not positive about it because of the size and scope of the AI code and my limited knowledge of its inner workings.

trevan:
@DanVanAtta , I've updated some comments, changed some variable names, and moved a method to be private.

game-core/src/main/java/games/strategy/engine/data/GameMap.java
DanVanAtta: I don't understand something. The checks  you added, the bi-predict, you have injected them from what is clearly AI code. What I'm suggesting to do is instead change the method the AI code calls and have the bi-predicates live in `GameMap`. That should not change the surface area of what you have updated and if there are other places that need to invoke the same logic it would also in turn invoke the new method on `GameMap` and would not duplicate the same Bi-predicates.

trevan: I think it is premature to start trying to define the API in a different manner than how it is right now.  Yes, the existing API isn't that great but adding additional API points that are completely different will just make it even more confusing.  The API currently accepts conditions on the territories so adding a new parameter that is also a condition check isn't much different.

DanVanAtta: I think we can do better than adding a parameter to an existing API. I disagree it is premature at all. We are setting a pattern. If another developer were to look at what is done here, "I need more functionality in this method - okay, add a parameter that only one part of the code uses and the rest will just pass a no-op value." Rinse-wash-repeat, and then eventually the concern of having too many params arise and the response is "the code was already like that, fixing all usages is out of scope, it is consistent to just add another parameter." Hence, I would say it's premature to add a parameter that make the method prematurely generic.

I don't see what it is a big change to encapsulate the match condition with the `getNeighbor` method. As-is, I think we violate DRY (which is about expressing an idea in one place, less about removing all duplicate patterns). Currently the neighbor logic is spread between match conditions only used for getNeighbor, the caller that invokes the method with the right parameter bindings, and then the logic in `GameMap`. Which is where the pattern is not a good one to follow, what if someone were to use this new public method API and pass in another match condition? It'd be an even worse situation, rather than the pattern to encapsulate the `getNeighbor` logic to one place.

Though again, I think we can do better than having a 4 param method and just add another method. We are debating so far between adding a 4 arg public method vs a 2 arg public method. The latter is readily tested, the former would likely involve heavy mocking and likely would not test very much. The heuristic of "can this be reasonably tested" I think fails when we consider almost any 4 arg method, this is an example I believe.

For more consideration, a predicate can be expressed as a bi-predicate, so we could keep the predicate logic all in the caller and add just one new API. EG:

```
public void predicateMethod(Predicate<Territory> territoryPredicate) {
   biPredicateMethod((start, end) -> territoryPredicate.apply(end));
}

public void biPredicateMethod(BiPredicate<Territory, Territory> routePredicate) {
   ///   
}
```

trevan: I'll close this then since it doesn't appear to be the correct fix for the issue.

trevan: If you want to work on it, go ahead.  I'm not going to do much else with it.

DanVanAtta: Given the testing statement there is a fix here and that is very valuable. 

"Correct" design is in the eye of the beholder and depends on future events that nobody can know, we can only implement the best design that we see for current and expected future circumstances given a favorable ROI.

Are you going to pursue this fix any further @trevan ?

trevan: @DanVanAtta , at this point, I'm not going to pursue any more here.  I'd love to get a more in-depth understanding of the AI code but the battle code is plenty big at this point in time.

DanVanAtta: @trevan , just saw your last comment. That answers my question.

It appears you have lost patience for this update & process. I'm disappointed we have reached that point.

Let's be direct when we can to avoid any misconception. It's okay to say "no, those design changes are difficult and involved. I'm not sure the updates suggested can cleanly land, feel free to take a stab by adding a commit to this branch."

Already at a cursory glance, we are dealing with a mess:

Screenshot from 2020-11-30 21-23-44

DanVanAtta:
@trevan I added three commits to this update, please have a look and let me know what you think.

For comparison, these were the method overloads before:
Screenshot from 2020-11-30 22-02-28

Then after combining any calls of `(predicate, bipredicate)` to instead be a call to `(bipredicate)` by expressing the predicate & bipredicate combo as `(t1, t2) -> predicate.test(t2) && bipredicate(t1, t2)`, these are the resulting method signatures:
Screenshot from 2020-11-30 22-02-56

game-core/src/main/java/games/strategy/engine/data/GameMap.java
DanVanAtta: The AI code is an interesting beast!  I think that code belongs in a sub-project, and then ideally we could push a ton of supporting logic into that sub-project as well. I think that would clean up the core game-engine (game-core) considerably.

For example, WeakAI would be removed but it is injected as a dummy player into the battle logic. Keep an eye out for that, if you can sever the WeakAI from the OOL & battle logic & battle calculator, then we could drop that AI completely.

trevan:
@DanVanAtta , your changes look fine.

PR #8291, opened 2020-11-27
Unify low luck and normal roll code in DiceRoll

trevan:
The AA low luck and normal rolls have the same logic as the non-AA low
luck and normal rolls. The only difference is handling dice sides that
are different from the game state.  So add a getDiceSides method to
TotalPowerAndTotalRolls and unify the code for low luck and normal
rolls.

## Testing

Played HardAI of Global 40 and WAW to test both normal and low luck games.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
@@ -1,6 +1,5 @@
 package games.strategy.triplea.delegate;
 
-import games.strategy.engine.data.GameData;
 import games.strategy.engine.data.GamePlayer;
 import games.strategy.engine.data.Territory;
 import games.strategy.engine.data.Unit;
@@ -14,6 +13,7 @@ import games.strategy.triplea.delegate.power.calculator.AaPowerStrengthAndRolls;
 import games.strategy.triplea.delegate.power.calculator.CombatValue;
 import games.strategy.triplea.delegate.power.calculator.CombatValueBuilder;
 import games.strategy.triplea.delegate.power.calculator.PowerStrengthAndRolls;
+import games.strategy.triplea.delegate.power.calculator.TotalPowerAndTotalRolls;
 import games.strategy.triplea.formatter.MyFormatter;
 import java.io.Externalizable;
 import java.io.IOException;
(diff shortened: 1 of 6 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
```diff
@@ -110,74 +110,26 @@ public static DiceRoll rollAa(
       final Territory location,
       final CombatValue combatValueCalculator) {
 
-    final GameData data = bridge.getData();
+    final String typeAa = UnitAttachment.get(aaUnits.iterator().next().getType()).getTypeAa();
+    final GamePlayer player = aaUnits.iterator().next().getOwner();
```
DanVanAtta, line 114: Would you mind updating the documentation that 'aaUnits' must have at least one element in it?

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/TotalPowerAndTotalRolls.java
```diff
@@ -13,6 +13,8 @@
 
   List<Die> getDiceHits(int[] dice);
 
+  int getDiceSides();
+
```
DanVanAtta, line 17: Side-note, this interface I think has lost its cohesion, it is more than total power and total rolls. It's also odds as `TotalPowerAndTotalRolls` would seem to be the name for a value object rather than an interface. This is something that if we can would be good t come back and rename. I almost think this would be an even better interface for a 'unitGroup', or have that just be a class and get rid of the sub-classing altogether and have the `unitGroup` use this for an API that would 'just do the right thing'.

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
```diff
@@ -219,37 +171,40 @@ public static DiceRoll rollDice(
...
+    bridge
+        .getHistoryWriter()
+        .addChildToEvent(annotation + " : " + MyFormatter.asDice(diceRoll), diceRoll);
+    return diceRoll;
   }
 
   /** Roll dice for units using low luck rules. */
   private static DiceRoll rollDiceLowLuck(
-      final Collection<Unit> unitsList,
+      final TotalPowerAndTotalRolls unitPowerAndRollsMap,
```
DanVanAtta, line 191: `unitPowerAndRollsMap` is an unexpected name for this parameter as the type is not a map. Any objections to renaming it?

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
```diff
@@ -219,37 +171,40 @@ public static DiceRoll rollDice(
...
-        PowerStrengthAndRolls.build(unitsList, combatValueCalculator).calculateTotalPower();
+    final int power = unitPowerAndRollsMap.calculateTotalPower();
     if (power == 0) {
       return new DiceRoll(List.of(), 0, 0);
     }
 
     // Roll dice for the fractional part of the dice
-    int hitCount = power / data.getDiceSides();
+    final int diceSides = unitPowerAndRollsMap.getDiceSides();
+    int hitCount = power / diceSides;
```
DanVanAtta, line 203: It seems like 'hitCount' should be part of the 'totalPowerAndRolls' interface. Getting data from an object and then using that data to compute a value, as is the case here, appears to be a 'feature-envy' code-smell.

It might be possible that if the API is `getHitCount`, then we won't need the `getDiceSides` method and the dice sides would instead turn into a constructor arg for implementations of `totalPowerAndRolls`

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
DanVanAtta: Side-note, perhaps worth doing here as we've re-written the entire method almost, most of the javadocs on the method are not very valuable.

EG:
- " aaUnits - AA units that could potentially be rolling" < does not really add much information. 
- "@param location - battle territory" < seems like the parameter should just be renamed
- "* @param bridge - delegate bridge" < ditto, parameter could just be renamed.

DanVanAtta (commented):
I like seeing the cleanup. A couple comments for you to consider.

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
trevan: I've updated the documentation.  The requirement that 'aaUnits' have at least on element isn't new after this refactor.  It has always been a requirement.  In a future PR, I plan to rework the parameters to this method so that requirement is no longer needed.

game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
trevan: In a future PR, I'm going to be reworking this so that it correctly handles units with different dice sides.  At that point, `getDiceSides` should no longer be needed.  I don't think a `getHitCount` method will fit in that rework but I'll look into it when I work on the PR.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/TotalPowerAndTotalRolls.java
trevan: This interface and `UnitPowerStrengthAndRolls` are two things I'm thinking of renaming/reworking for the casualty selection and support calculation.  I'm thinking of turning `UnitPowerStrengthAndRolls` (with a better name) into a delegate of `Unit`.  Then instead of passing around Units and calling `TotalPowerAndTotalRolls` to generate the `UnitPowerStrengthAndRolls` objects, the battle logic will first wrap all of the units in a `UnitPowerStrengthAndRolls` delegate and then those will be passed around.

I'm also planning on moving the dice calculation methods to `DiceRoll`.  The only reason they are on `TotalPowerAndTotalRolls` is legacy now.

trevan:
@DanVanAtta , I've pushed an update and responded to your comments.

game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/TotalPowerAndTotalRolls.java
DanVanAtta: Cool. One note here, `DiceRoll` seems like it ought to be a value object, the number and types of methods on it do not seem appropriate. Something for you to consider, seemingly there should be of a strategy object that is more part of battles for determining roll results.

DanVanAtta (approved):

PR #8307, opened 2020-11-28
Inline special sbr/fly over dice roll method

trevan:

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/AaInMoveUtil.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/AaInMoveUtil.java
@@ -116,7 +116,8 @@ class AaInMoveUtil implements Serializable {
               // get rid of units already killed, so we don't target them twice
               validTargetedUnitsForThisRoll.removeAll(casualties);
               if (!validTargetedUnitsForThisRoll.isEmpty()) {
-                // Fly over AA currently doesn't take into account support
+                // Fly over AA currently doesn't take into account support so don't pass in
+                // the enemyUnits or friendlyUnits
                 dice.set(
                     DiceRoll.rollAa(
                         validTargetedUnitsForThisRoll,
--- a/game-core/src/main/java/games/strategy/triplea/delegate/AaInMoveUtil.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/AaInMoveUtil.java
@@ -116,13 +116,20 @@ class AaInMoveUtil implements Serializable {
               // get rid of units already killed, so we don't target them twice
               validTargetedUnitsForThisRoll.removeAll(casualties);
               if (!validTargetedUnitsForThisRoll.isEmpty()) {
+                // Fly over AA currently doesn't take into account support
                 dice.set(
-                    DiceRoll.rollSbrOrFlyOverAa(
+                    DiceRoll.rollAa(
                         validTargetedUnitsForThisRoll,
                         currentPossibleAa,
                         AaInMoveUtil.this.bridge,
                         territory,
-                        BattleState.Side.DEFENSE));
+                        CombatValueBuilder.aaCombatValue()
+                            .enemyUnits(List.of())
+                            .friendlyUnits(List.of())
+                            .side(BattleState.Side.DEFENSE)
+                            .supportAttachments(
+                                bridge.getData().getUnitTypeList().getSupportAaRules())
+                            .build()));
               }
             }
           };
--- a/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/DiceRoll.java
@@ -9,10 +9,8 @@ import games.strategy.engine.random.IRandomStats.DiceType;
 import games.strategy.triplea.Properties;
 import games.strategy.triplea.attachments.UnitAttachment;
 import games.strategy.triplea.delegate.Die.DieType;
-import games.strategy.triplea.delegate.battle.BattleState;
 import games.strategy.triplea.delegate.power.calculator.AaPowerStrengthAndRolls;
 import games.strategy.triplea.delegate.power.calculator.CombatValue;
-import games.strategy.triplea.delegate.power.calculator.CombatValueBuilder;
 import games.strategy.triplea.delegate.power.calculator.PowerStrengthAndRolls;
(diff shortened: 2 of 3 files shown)


trevan:
The Smoke Tests failed for something that doesn't appear to be related to this PR.

game-core/src/main/java/games/strategy/triplea/delegate/AaInMoveUtil.java
```diff
@@ -116,13 +116,20 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
...
                 dice.set(
-                    DiceRoll.rollSbrOrFlyOverAa(
+                    DiceRoll.rollAa(
                         validTargetedUnitsForThisRoll,
                         currentPossibleAa,
                         AaInMoveUtil.this.bridge,
                         territory,
-                        BattleState.Side.DEFENSE));
+                        CombatValueBuilder.aaCombatValue()
+                            .enemyUnits(List.of())
```
DanVanAtta, line 127: nit, you can make the API nicer here by defaulting enemyUnits to an empty list, then you won't need to specify it in the builder.

DanVanAtta (commented):
Would you mind explaining this update in a bit more detail @trevan?

trevan:
@DanVanAtta , I'm not sure what more you want.  This is just inlining a simple method.

game-core/src/main/java/games/strategy/triplea/delegate/AaInMoveUtil.java
trevan: I prefer that it is explicit.  Generally, the enemy list shouldn't be empty.  The only reason it is empty here is because support for AA guns during sbr or fly over isn't implemented (or maybe it isn't wanted).

DanVanAtta: Okay.
side-note:  in that case take caution for a potential interface-segregation-principle violation here.

game-core/src/main/java/games/strategy/triplea/delegate/AaInMoveUtil.java
```diff
@@ -116,13 +116,20 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
               // get rid of units already killed, so we don't target them twice
               validTargetedUnitsForThisRoll.removeAll(casualties);
               if (!validTargetedUnitsForThisRoll.isEmpty()) {
+                // Fly over AA currently doesn't take into account support
```
DanVanAtta, line 119: Is this comment calling out a bug? A TODO item - Should the comment be on the method itself?  I'm not sure how knowing that rollAa does not take into support really helps to understand the method call. To another extent, it is confusing as the support rules are passed to the next method, which raises questions (are they used?)

DanVanAtta (commented):
Okay, knowing we are inlining is helpful. What threw me off is the title of the PR said 'remove'

game-core/src/main/java/games/strategy/triplea/delegate/AaInMoveUtil.java
trevan: It was in the original function so it got inlined here.  `rollA` does take into account support but you have to give it support knowledge.  This method doesn't pass in support knowledge to `rollAa` so it doesn't take into account support.  As for why fly-overs don't include support, I don't know.  I think it might be because it was just never done but there might have been a reason why it was excluded.  I can't find any evidence of why in the git history.

DanVanAtta: I suspect because the fly-over shot is "mis-implemented" and should occur at the end of the move phase instead of immediate. For example, if you move units one by one, they'll get AA shots one-by-one when in actuality the rules treat movements as all happening concurrently at the end of the move phase. Hence, if you move a second unit along the same path, the two units shared path.

Now, we can probably split hairs here as there are implications for capturing unoccuppied territories and the fly-over that is allowed because of that. I can't recall if that is technically a mis-implementation or a special case of the rules somehow, or if the interpretation of concurrent moves is otherwise flawed.

IMO, we should remove this comment or move it somewhere else. A technical debt tracker somewhere else might make more sense. I'm concerned a bit about the value of the comment, it raises more questions than it does add clarity

trevan: I think the comment fits here.  Someone looking at this code would wonder why the `friendlyUnits` and the `enemyUnits` are set to `List.of()`.  This comment explains why.

DanVanAtta: I don't think I agree. If you have a lot of context to the implementation of the invoked method, then yes, but that is after the fact. The `friendlyUnits` and `enemyUnits` being empty is not seemingly directly related to supports. To determine that, you need to look at the implementation.

trevan: Maybe I need to rename `friendlyUnits` and `enemyUnits` to make it obvious that they are directly related to supports.  But I don't want remove the comment.  It is helpful to me in understanding why the support is not being provided.

DanVanAtta: Really getting into the details on this one. I'm sorry to be nitpicking over a comment, there are lots of similar comments in the code that don't really help or were made were excessive context, I think this is one of them as it did not make sense to me (and a reviewer I think is just someone simulating how the code reads when you have less context, perhaps because you yourself are updating this a few months from now).

Okay with that apology out of the way, I think we can do one or two things:
1. Clarify the comment, perhaps just say in the comment "The attacker and defender lists are empty because we are not yet calculating support for fly-over AA shots"
2. Create a special `CombatValueBuilder` that makes the lack of support clear (and perhaps removes the empty list args and secondly maybe makes the comment redundant), eg: `CombatValueBuilder.aaCombatValueWithoutSupports()`


(2) is maybe tricky to name right, but the general gist is to clarify the comment to add that missing context and/or update the code to bake that context into the method calls.

trevan:
@DanVanAtta , I've updated the comment.

DanVanAtta:
Interesting the comment is duplicated, didn't notice that until you updated.
Thanks for the making that update :+1:

PR #8308, opened 2020-11-28
Create dice.calculator and dice.roller to store DiceRoll strategies

trevan:
Code in TotalPowerAndTotalRolls that was related to dice rolling has
been moved to dice.calculator.  Static methods in DiceRoll to generate
dice rolls has been moved to dice.roller.


game-core/src/main/java/games/strategy/triplea/delegate/dice/roller/RollNDice.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.List;
+import lombok.experimental.UtilityClass;
+
+/**
+ * Roll n-sided dice.
+ *
+ * <p>0 based, add 1 to get actual die roll
+ */
+@UtilityClass
+public class RollNDice {
```
DanVanAtta, line 18: RollNDice could be confused with "rollinDice" and "roll and dice". It is also confusing if we are rolling a N-sided die or are rolling n-dice.

This class looks like a dice-roller. I would opt to name it that. Looking at the one method in this class, I notice the API is not the prettiest. Of note, we have two int parameters side by side that could be swapped or would be hard to distinguish.

One fix would be to convert the parameters of the method to private variables and then use a builder to construct and have a no-arg 'rollDice' method. That should clean up the usage. Though, a lot of the parameters are simply to feed the call to "bridge.getRandom(sides, rollCount, playerRolling, diceType, annotation);". I think this method might be trying to do too much an dperhaps should take a `random[]` as a parameter and then it would have far fewer methods. Probably even better would be to have `DiceRoll` to have a constructor that accepts the `random[]` computation and it can then do most of the rest of the logic. 

TL;DR, this method looks like it should instead be a constructor on DiceRoll.

game-core/src/main/java/games/strategy/triplea/delegate/dice/roller/RollBattleDice.java
```diff
@@ -0,0 +1,41 @@
...
+    final DiceRoll diceRoll;
+    final PowerStrengthAndRolls unitPowerAndRollsMap =
+        PowerStrengthAndRolls.build(units, combatValueCalculator);
+    if (Properties.getLowLuck(bridge.getData().getProperties())) {
+      diceRoll = LowLuckDice.calculate(unitPowerAndRollsMap, player, bridge::getRandom, annotation);
+    } else {
+      diceRoll = RolledDice.calculate(unitPowerAndRollsMap, player, bridge::getRandom, annotation);
+    }
+
+    bridge
```
DanVanAtta, line 36: I don't know if the bridge call belongs here. I would personally opt to duplicate this bridge call wherever we use the dice roller. The reasoning is largely just SRP, we are doing a lot of things here and the bridge call is a side effect.

game-core/src/main/java/games/strategy/triplea/delegate/dice/roller/RollBattleDice.java
```diff
@@ -0,0 +1,41 @@
...
+public class RollBattleDice {
+
+  public static DiceRoll rollDice(
+      final Collection<Unit> units,
+      final GamePlayer player,
+      final IDelegateBridge bridge,
+      final String annotation,
+      final CombatValue combatValueCalculator) {
+
+    final DiceRoll diceRoll;
```
DanVanAtta, line 27: nit, diceRoll should be defined preferably as close to where it is assigned. This is a good looking case to use a ternary even and would reduce complexity a little bit.

game-core/src/main/java/games/strategy/triplea/delegate/dice/roller/RollBattleDice.java
DanVanAtta: Looking at what remains, I'm not really convinced this is a useful class. Can this be combined with the RollNDice class? It seems like this should be a stateful 'diceRoller' class and then it handles whether it rolls LL or normal dice as an implementation detail.

DanVanAtta: I am concerned this is going to get confusing, when does one use `RollBattleDice` and when do you use `RollNDice`? I think combining and making these two classes potentially stateful would be a good way to go. I would seriously lean to remove the bridge code as that is side-effect of reporting the result and is not intrinsic to rolling dice.

trevan: The bridge call is unfortunately intrinsic to dice.  When you save a game and then reload it, it will calculate the dice probability of the saved game using these history items.  If they are missing or not written correctly, then the dice probability of the saved game will be invalid.

I don't agree with the design but I don't see how to easily change that.  And since every call of these methods needs to also write the history, I think it is best to leave the call in here so that it isn't forgotten.  I'll add a comment to the code so that it explains why this history is required.

trevan: These three classes (`RollAaDice`, `RollNDice`, and `RollBattleDice`) are basically strategy objects that define what type of dice you are getting.  I could see creating a builder class that you pass in an enum but I'm not sure how that is any better than just calling the static method itself.

game-core/src/main/java/games/strategy/triplea/delegate/dice/roller/RollNDice.java
trevan: This class is a builder of `DiceRoll`.  You shouldn't create your own `DiceRoll` but instead call this method if you need to roll `N` dice.  This also creates special `IGNORED` dice.  You don't want to use this for battle situations.  This is currently used by war bonds but I want to expand it to the other places that call `bridge.getRandom` so that we have a unified dice logic.

I thought of calling this `RollIgnoredDice` but that would be really strange.  It was originally called `rollNDice` and I think it fits.

game-core/src/main/java/games/strategy/triplea/delegate/dice/roller/RollBattleDice.java
trevan: I have been thinking of changing the annotation from a `String` to a `DiceAnnotation` (or similar).  Then the parameter to these methods would require a `DiceAnnotation` instead of a `String` and it would ensure that the annotation is the correct format (since the history string is parsed to build the probability history).  I could make `DiceAnnotation` accept a bridge so that it also writes the history and then change these classes to call a method on the `DiceAnnotation` object that will write the history.  But that would be another PR.

trevan (commented):
@DanVanAtta , I switched to the ternary object and responded to your comments.

game-core/src/main/java/games/strategy/triplea/delegate/dice/roller/RollNDice.java
DanVanAtta: ### RE: rollNDice

When I first read 'rollNDice', my first thought it _was_ 'rollAndDice' and it was similar to 'StengthAndRolls'

I'd say then the original name was clever, too short and required more than naive context to understand (which means it takes longer to understand the code, the more context you have to acquire, because context acquisition takes time, ergo the more context needed implies more time is needed to read and understand code).

TBH I'm still not clear if `n` is the number of dice or the number of dice sides. I think the latter, but maybe it's even both??

So please have some thoughts about the naming for `rollNDice` and consider spelling it out to make it easier on a reader.


### Re: Factory methods

Okay, it's good to know this is supposed to be a factory method. I think the cohesion heuristic is potentially low as we have a factory class, with a factory method, that are in one package, meanwhile `DiceRoll` is in: `package games.strategy.triplea.delegate;`

My concern about the 3 factory classes is that they are each relatively shallow, it's not offhand clear they are factories either explicitly or from the javadocs, and it's potentially not clear when to use and not the other. 

To alleviate these concerns I can suggest we combine the factory methods into one class and then document when you would use one each one. Seeing the series of methods in one class file I think would draw context between when each one would be used. I think ideally the factory class would be 'close' by to the 'DiceRoll' class, though I can understand not having them co-located just yet.

WDYT?

game-core/src/main/java/games/strategy/triplea/delegate/dice/roller/RollBattleDice.java
DanVanAtta: My 2 cents, if anyone is using the dice roll code then they'll be copy/pasting from an existing usage and the bridge will be there. I will admit there is potential for error. 

On the other hand, if you want to get 'just a dice roll', it's tightly bound to the bridge call and it's a side-effect that someone would not want. I could imagine a duct-tape-programmer putting in a boolean parameter to avoid this side-effect.

With that said, leaving it for now is okay. I like your ideas for further improvements.

[resolve]

DanVanAtta:
Thanks for following up @trevan with the updates and commentary. Just one last thing to work through to make it clear & documented the distinct dice roll factory methods and when each one would be used. My recommendation is largely to combine to a single factory class and simply document when to use each.

game-core/src/main/java/games/strategy/triplea/delegate/dice/roller/RollNDice.java
trevan: @DanVanAtta , I've renamed `RollNDice` to `RollNSidedDiceXTimes`.  I hope that is more specific.

As for the changes to the factory methods, I want to move `DiceRoll` but I can't because of saved game issue.  I've already added `@ChangeOnNextMajorRelease` to indicate that `DiceRoll` should be moved.  Once it is moved, it should make more sense.  I understand your desire to merge them into one class with 3 static methods but I hope to change them to `@Builder`s and I plan to try and get more of the `getRandom` usages added as new factory classes.  So I'd rather leave them separate for now.

trevan: I also thought the javadocs explain when you would use one or the other.  I'm not sure what to add to them to make them clearer.

DanVanAtta: `RollNSidedDiceXTimes` is wordy but certainly more clear!

My thinking was with the factory classes merged into one, the existing javadocs would contrast against one another more readily and make it obvious that the factory methods were related. That is an off-the-cuff suggestion, I'll take a deeper look soon in case there is a better suggestion that comes to mind or if my initial impression was just off-base.

trevan:
@DanVanAtta I've moved them all to a single factory class.  I think that I might be able to combine both the AA dice and the battle dice methods but I'll need to do the annotation change first.  I'll do those as separate PRs.

DanVanAtta (approved):

PR #8309, opened 2020-11-28
Add tests for AaCasualtySelector random and individually rolled

trevan:

### DIFF
--- a/game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelectorTest.java
+++ b/game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelectorTest.java
@@ -48,23 +48,7 @@ class AaCasualtySelectorTest {
   private UnitType planeMultiHpUnitType;
   private GameData gameData;
 
-  private List<Unit> givenAaUnits(final int quantity) {
-    return aaUnitType.create(quantity, aaPlayer, true);
-  }
-
-  private List<Unit> givenDamageableAaUnits(final int quantity) {
-    return damageableAaUnitType.create(quantity, aaPlayer, true);
-  }
-
-  private List<Unit> givenPlaneUnits(final int quantity) {
-    return planeUnitType.create(quantity, hitPlayer, true);
-  }
-
-  private List<Unit> givenMultiHpPlaneUnits(final int quantity) {
-    return planeMultiHpUnitType.create(quantity, hitPlayer, true);
-  }
-
-  private DiceRoll givenDiceRoll(final boolean... shots) {
+  private DiceRoll givenDiceRollWithHitSequence(final boolean... shots) {
     final int[] diceRolls = new int[shots.length];
     int hits = 0;
     for (int i = 0; i < shots.length; i++) {
@@ -131,11 +115,11 @@ class AaCasualtySelectorTest {
       final CasualtyDetails details =
           AaCasualtySelector.getAaCasualties(
               List.of(mock(Unit.class)),
-              givenAaUnits(1),
+              aaUnitType.create(1, aaPlayer, true),
               mock(CombatValue.class),
               mock(CombatValue.class),
               "text",
-              givenDiceRoll(),
+              givenDiceRollWithHitSequence(),
               bridge,
               hitPlayer,
               UUID.randomUUID(),
@@ -149,12 +133,12 @@ class AaCasualtySelectorTest {
 
       final CasualtyDetails details =
           AaCasualtySelector.getAaCasualties(
-              givenPlaneUnits(1),
-              givenAaUnits(1),
+              planeUnitType.create(1, hitPlayer, true),
+              aaUnitType.create(1, aaPlayer, true),
               mock(CombatValue.class),
               mock(CombatValue.class),
               "text",
-              givenDiceRoll(true),
+              givenDiceRollWithHitSequence(true),
               bridge,
               hitPlayer,
               UUID.randomUUID(),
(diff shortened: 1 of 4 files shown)


game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelectorTest.java
```diff
@@ -0,0 +1,432 @@
...
+              mock(CombatValue.class),
+              mock(CombatValue.class),
+              "text",
+              givenDiceRoll(true, true, true),
+              bridge,
+              hitPlayer,
+              UUID.randomUUID(),
+              mock(Territory.class));
+
+      assertThat("1 Plane is killed", details.getKilled(), hasSize(1));
```
DanVanAtta, line 318: Why is 1 plane killed? The assertion here is not really checking that it's a plane that is killed either.

I ask that question as that is the question a maintainer would have after reading "1 plane is killed". IMO it paraphrases the asserted state rather than describing why we expect to be in that state. Perhaps if we said something like "Rolling a '0' on the random bridge is a hit"

That could probably be made even more clear as well by assigning the dice rolls to variables, eg:
```
int aaHit = 0;
int aaMiss = 2

whenGetRandom(bridge).thenAnswer(withValues(aaHit, aaMiss, aaMiss));
```


```

File: game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelectorTest.java [hunk 2]
```diff
@@ -0,0 +1,432 @@
...
+      final List<Unit> planes = givenPlaneUnits(5);
+
+      final CasualtyDetails details =
+          AaCasualtySelector.getAaCasualties(
+              planes,
+              givenAaUnits(1),
+              mock(CombatValue.class),
+              givenAaCombatValue(),
+              "text",
+              givenDiceRoll(true, false, true, false, true),
```
DanVanAtta, line 420: This 'given' method is not easy to understand. A javadoc on the given method would help avoid having to reverse engineer the code. The name of the method probably ought to describe what kind of dice roll is being returned. EG: `givenDiceRollsWithHitSequence`, or something like that.

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelectorTest.java
```diff
@@ -0,0 +1,432 @@
...
+
+  private List<Unit> givenAaUnits(final int quantity) {
+    return aaUnitType.create(quantity, aaPlayer, true);
+  }
+
+  private List<Unit> givenDamageableAaUnits(final int quantity) {
+    return damageableAaUnitType.create(quantity, aaPlayer, true);
+  }
+
+  private List<Unit> givenPlaneUnits(final int quantity) {
```
DanVanAtta, line 59: Consider inlining this method, potentially needless indirection. The one liner is more self explanatory than the method name. Seeing `givenPlaneUnits(10)`, it's not clear what the 10 is. The boolean parameter on the `create` API is needless and we should create an overload on the `create` method that has a default 'true' on it so the API can then make a lot of sense and be just "create units, count, player".

game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/MockGameData.java
```diff
@@ -155,4 +161,34 @@ public MockGameData withCaptureUnitsOnEnteringTerritory(final boolean value) {
     when(gameProperties.get(CAPTURE_UNITS_ON_ENTERING_TERRITORY, false)).thenReturn(value);
     return this;
   }
+
+  public MockGameData withEditMode(final boolean value) {
```
DanVanAtta, line 165: Side-note, I don't know if this pattern will scale well. Ideally we'd just use a real object here. To do that, extracting the game properties to it's own class and then passing that properties class around would perhaps be the thing to do. To that extent, we could have the battle logic take in that properties object and not a full game data as well for further benefits.

The reason I don't think this will scale is because eventually with more test coverage we'll have re-implemented the full game properties object and have a near duplicate dummy implementation.

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelectorTest.java
trevan: The method name explains why exactly one of the planes are hit.  But I can move that to this message as well.

As for the dice rolls, they aren't related to hits.  They are used to randomly select the units that get the hits.

trevan (commented):
@DanVanAtta , I've inlined the methods, changed the name of givenDiceRolls, and modified the descriptions to be more helpful.

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelectorTest.java
DanVanAtta: Then perhaps:
```
  public List<Unit> createTemp(final int quantity, final GamePlayer owner) {
    return create(quantity, owner);
  }
```

With that it's clear you are creating a 'temp' unit. (What a temp unit is, is another question altogether, but at least the magic boolean is removed).

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelectorTest.java
DanVanAtta: An assert message and the test method should not be redundant to each other. If so, then the assert message is not needed. Generally the method name is what you are testing, what aspect of a feature, the assert method message is why the condition is expected to be true. Often when there is one assert in a method the reasoning is clear. Again, the goal is to help other developers know why the test is supposed to pass (not to check a box or to pass a style guide, but if someone is developing, breaks this test, they need to know why the test was correct to begin with and why the assertions were expected to pass)

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelectorTest.java
trevan: Let me do that in another PR.  There's a lot of places that use the temp = true and it would be nice to keep them all unified.

DanVanAtta:
Thank you for the updates @trevan - good work :+1:

PR #8318, opened 2020-11-30
Add tests for low luck aa casualties

trevan:

### DIFF
--- a/game-core/src/test/java/games/strategy/triplea/delegate/MockDelegateBridge.java
+++ b/game-core/src/test/java/games/strategy/triplea/delegate/MockDelegateBridge.java
@@ -1,6 +1,7 @@
 package games.strategy.triplea.delegate;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.ArgumentMatchers.anyString;
@@ -58,7 +59,7 @@ public final class MockDelegateBridge {
   public static Answer<int[]> withValues(final int... values) {
     return invocation -> {
       final int count = invocation.getArgument(1);
-      assertEquals(values.length, count, "count of requested random values does not match");
+      assertThat("count of requested random values does not match", values.length, is(count));
       return values;
     };
   }
--- a/game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelectorTest.java
+++ b/game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelectorTest.java
@@ -488,12 +488,12 @@ class AaCasualtySelectorTest {
     @Test
     void oneTypeOfPlaneWithAmountEqualToDiceSides() {
 
-      final List<Unit> planes = planeUnitType.create(6, hitPlayer, true);
+      final List<Unit> planes = planeUnitType.createTemp(6, hitPlayer);
 
       final CasualtyDetails details =
           AaCasualtySelector.getAaCasualties(
               planes,
-              aaUnitType.create(1, aaPlayer, true),
+              aaUnitType.createTemp(1, aaPlayer),
               mock(CombatValue.class),
               givenAaCombatValue(),
               "text",
@@ -514,13 +514,13 @@ class AaCasualtySelectorTest {
     @Test
     void twoTypesOfPlanesAndBothHaveAmountEqualToDiceSides() {
 
-      final List<Unit> planes = planeUnitType.create(6, hitPlayer, true);
-      planes.addAll(otherPlaneUnitType.create(6, hitPlayer, true));
+      final List<Unit> planes = planeUnitType.createTemp(6, hitPlayer);
+      planes.addAll(otherPlaneUnitType.createTemp(6, hitPlayer));
 
       final CasualtyDetails details =
           AaCasualtySelector.getAaCasualties(
               planes,
-              aaUnitType.create(1, aaPlayer, true),
+              aaUnitType.createTemp(1, aaPlayer),
               mock(CombatValue.class),
               givenAaCombatValue(),
               "text",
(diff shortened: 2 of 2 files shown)


game-core/src/test/java/games/strategy/triplea/delegate/MockDelegateBridge.java
```diff
@@ -58,7 +59,7 @@ public static IDelegateBridge newDelegateBridge(final GamePlayer gamePlayer) {
   public static Answer<int[]> withValues(final int... values) {
     return invocation -> {
       final int count = invocation.getArgument(1);
-      assertEquals(values.length, count, "count of requested random values does not match");
+      assertThat("count of requested random values does not match", values.length, is(count));
```
trevan, line 62: I modified this because the order of elements was wrong.  It is `(expected, actual, description)` and it is being used as `(actual, expected, description)`.  I switched to assertThat while modifying this.

trevan:
Is there any documentation on how Low Luck AA casualties are supposed to work?  Because while building out these tests, I came across several instances where things didn't work as I expected.  So, I'd like to know how it is supposed to work so that when I rework this code, I can make it work better.

game-core/src/test/java/games/strategy/triplea/delegate/MockDelegateBridge.java
DanVanAtta: Minor, there is a `IsArrayWithSize.arrayWithSize()` matcher that can be used here.

Side-note/observation: It's really odd for there to be an assertion in mock code to begin with :thinking:

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelectorTest.java
```diff
@@ -417,4 +419,606 @@ void hitsLessThanPlanesKillsAccordingToTheRolledDice() {
           is(List.of(planes.get(0), planes.get(2), planes.get(4))));
     }
   }
+
+  @Nested
+  class LowLuckCasualties {
```
DanVanAtta, line 424: I want to say it's pretty amazing there was no coverage on this before, then again with 27% test coverage, maybe not so much : )

Thanks for adding coverage!

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelectorTest.java
```diff
@@ -417,4 +419,606 @@ void hitsLessThanPlanesKillsAccordingToTheRolledDice() {
...
+          givenGameData()
+              .withDiceSides(6)
+              .withEditMode(false)
+              .withChooseAaCasualties(false)
+              .withLowLuck(true)
+              .build();
+      bridge = mock(IDelegateBridge.class);
+      when(bridge.getData()).thenReturn(gameData);
+
+      aaUnitType = new UnitType("aaUnitType", gameData);
```
DanVanAtta, line 447: observation, the setup is very beefy, lots of test-state is established. If we add a few more tests like this it will be more worthwhile to invest in easier setup. Ideally a test case can have relatively minimal fixture code around it. Just noting that I'm not sure this pattern will scale super gracefully with a lot more tests.

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelectorTest.java
```diff
@@ -417,4 +419,606 @@ void hitsLessThanPlanesKillsAccordingToTheRolledDice() {
...
+
+      assertThat(details.getKilled(), hasSize(1));
+      assertThat(
+          "The otherPlaneUnitType is killed because it's unitType is sorted before planeUnitType "
+              + "(it sorts by unit type name) and there are 3 of them and the random selection was "
+              + "for the second unit in the list.",
+          details.getKilled().stream().map(Unit::getType).collect(Collectors.toList()),
+          containsInAnyOrder(otherPlaneUnitType));
+
+      assertThat(
```
DanVanAtta, line 577: This has been asserted a few times, is there any way that exactly one of these tests would have this assertion fail and not the others, or is it more the case that if one fails they all would fail?

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelectorTest.java
```diff
@@ -417,4 +419,606 @@ void hitsLessThanPlanesKillsAccordingToTheRolledDice() {
...
+
+      final CasualtyDetails details =
+          AaCasualtySelector.getAaCasualties(
+              planes,
+              damageableAaUnitType.createTemp(1, aaPlayer),
+              mock(CombatValue.class),
+              givenAaCombatValue(),
+              "text",
+              // low luck would have 1 hit because 6 power / 6 dice side
+              givenDiceRollWithHitSequence(true),
```
DanVanAtta, line 598: WDYT to rename this method "givenSingleDiceRollWithOneHit", then it might not have to be commented? Or something like `givenLowLuckGuaranteedHit`

I'm wondering about ways we can make this clear without needing the additional comment.

DanVanAtta (commented):
Few comments & observations - generally looks good

game-core/src/test/java/games/strategy/triplea/delegate/MockDelegateBridge.java
trevan: That doesn't work with non-object types.  I'd have to convert `values` from an `int[]` to an `Integer[]`.

trevan (commented):
@DanVanAtta , I've added a new `givenDiceRollWithUnits` that calculates the correct hits.  I had initially not wanted to do this because I worried that the unittypes and units would have to be updated more.  But they worked fine.

I've removed the comments, but I still think they are useful as they explain how the number of hits is calculated.

DanVanAtta:
re: comments, useful comments are useful! One should be careful though that they have enough context. You will know a lot more about what you are commenting on than you think. A comment should convey the thought process of the developer, the reasons why things are they are, tell a story about the code and what it's doing and why - but be careful as someone else reading it might be missing some key pieces of context which make the comments less than useful.

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelectorTest.java
```diff
@@ -511,21 +512,60 @@ void oneTypeOfPlaneWithAmountEqualToDiceSides() {
           is(empty()));
     }
 
+    private DiceRoll givenLowLuckDiceRoll(final List<Unit> aaUnits, final List<Unit> planes) {
```
DanVanAtta, line 515: IMHO - a comment on _this_ method would go a long way, what is this method doing?

DanVanAtta (approved):

game-core/src/test/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelectorTest.java
DanVanAtta: Okay, a heuristic for good tests is that they are independent. This means if you make a change then you wouldn't break all tests, just the specific ones regarding what was changed (which should be relatively targeted).

PR #8321, opened 2020-11-30
Update serverModel to use 'headless' check consistently

DanVanAtta:
Instead of using either the headless instance being null,
or the 'ui' being null to check for headless, this update
changes ServerMode to use HeadlessServer.headless() instead.


game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
```diff
@@ -137,7 +137,7 @@ public void releasePlayer(final INode who, final String playerName) {
 
         @Override
         public void disablePlayer(final String playerName) {
-          if (ui != null) {
+          if (!HeadlessGameServer.headless()) {
```
RoiEXLab, line 140: A bit late to the party, but I want to provide some context to this.
IIRC I introduced the ui != null check consistently as a first step to abstract the UI into an interace eventually, so that disablePlayer for example would just call an interface, which does nothing for headless servers, but does all of the UI stuff for headed servers.

PR #8323, opened 2020-11-30
Add units in the same change set when their damage is set

trevan:
If there are unitsToAdd, then those units have been created based on
other units and damage from those other units are transferred to the new
units. The damage change set is currently being sent before the add
units change set. For now, ensure both change sets are sent together.

In the future, need to ensure that the add units change set is sent
first.

## Testing

I noticed this issue when playing a saved game in the middle of a battle that had units changing into other units when they are damaged.  I tested the same saved game with this change and the error didn't show up anymore.

This is a regression from the change to UnitDamageReceivedChange but that isn't in a release yet so a release note is not being added.


DanVanAtta:
Good catch @trevan :+1:

PR #8326, opened 2020-11-30
Rename ChangeAttachmentChange and use primitives 

trevan:
ChangeAttachmentChange was renamed to AttachmentPropertyChange.

ChangeAttachmentChange stores an Attachable object. This is an
interface and didn't have an easy way to turn it into a primitive. A new
method getPrimitiveForm has been added that will generate an
PrimitiveAttachable. The PrimitiveAttachable can be easily serialized
and can deserialize with the help of the GameData object.

The PrimitiveAttachable is based off of GameObjectStreamData.

## Testing

I played Fast AI of Wacraft Heroes and TWW.  I believe they use lots of triggers and that is where a lot of ChangeAttachmentChanges are used.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/Attachable.java
+++ b/game-core/src/main/java/games/strategy/engine/data/Attachable.java
@@ -1,5 +1,6 @@
 package games.strategy.engine.data;
 
+import games.strategy.engine.data.changefactory.serializers.PrimitiveNamedAttachable;
 import java.util.Map;
 
 /**
@@ -16,4 +17,6 @@ public interface Attachable {
   IAttachment getAttachment(String key);
 
   Map<String, IAttachment> getAttachments();
+
+  PrimitiveNamedAttachable getPrimitiveForm();
 }
--- a/game-core/src/main/java/games/strategy/engine/data/NamedAttachable.java
+++ b/game-core/src/main/java/games/strategy/engine/data/NamedAttachable.java
@@ -1,5 +1,6 @@
 package games.strategy.engine.data;
 
+import games.strategy.engine.data.changefactory.serializers.PrimitiveNamedAttachable;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
@@ -33,5 +34,10 @@ public class NamedAttachable extends DefaultNamed implements Attachable {
   @Override
   public void removeAttachment(final String keyString) {
     attachments.remove(keyString);
+  }
+
+  @Override
+  public PrimitiveNamedAttachable getPrimitiveForm() {
+    return new PrimitiveNamedAttachable(this);
   }
 }
--- a/game-core/src/main/java/games/strategy/engine/data/changefactory/attachments/AttachmentPropertyChange.java
+++ b/game-core/src/main/java/games/strategy/engine/data/changefactory/attachments/AttachmentPropertyChange.java
@@ -1,14 +1,17 @@
-package games.strategy.engine.data;
+package games.strategy.engine.data.changefactory.attachments;
 
+import games.strategy.engine.data.Attachable;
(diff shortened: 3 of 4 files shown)


trevan:
This is a draft because it depends on #8323 so that commit is also in this branch.  Once that PR is merged, I'll be able to update this PR so that it only has the one commit.

I'm also not sure if this is the best way to make `Attachable`s primitive.

game-core/src/main/java/games/strategy/engine/data/changefactory/serializers/PrimitiveAttachable.java
```diff
@@ -0,0 +1,85 @@
...
+import games.strategy.engine.data.TerritoryEffect;
+import games.strategy.engine.data.UnitType;
+import games.strategy.triplea.delegate.TechAdvance;
+import java.io.Serializable;
+
+/**
+ * Serializes NamedAttachments into a primitive form that can be deserialized with access to
+ * GameData
+ */
+public class PrimitiveAttachable implements Serializable {
```
DanVanAtta, line 21: PrimitiveAttachable is just a primitive version of NamedAttachments?

trevan: Basically, though I mistyped the class name.  It isn't `NamedAttachment`s but `NamedAttachable`s.

It appears that all NamedAttachable are constant and based off of the original game xml.  That is how `GameObjectStreamData` treats them.  When they are serialized by`GameObjectOutputStream` will convert these objects into a `GameObjectStreamData`.  Then, when they are unserialized by `GameObjectInputStream`, it will use the name to look up the object in the local `GameData`.

trevan:
This is ready for review.  I think the biggest thing to work out is how to "serialize" all the `Attachable`s (or `NamedAttachable`s as they are the only implementer of `Attachable).

Before this change, they are serialized through special code in `GameObjectInputStream` and `GameObjectOutputStream`.  The "primitive" version that is passed between game instances is a `GameObjectStreamData`.

So my thoughts was to add a new method to `Attachable` that will return a `PrimitiveAttachable`/`PrimitiveNamedAttachable` which does basically what `GameObjectStreamData` already does.

But, the `PrimitiveAttachable` could also be an interface and then each `Attachable` has their own implementation of `PrimitiveAttachable` instead of having the work done in `PrimitiveAttachable`.  And, if there turns out to be `Attachable`s that need to be serialized in the "normal" manner, then it can just implement `PrimitiveAttachable` and return itself.

WDYT?

DanVanAtta:
I'm having trouble grasping the trade-offs and the exact suggestion @trevan - give me some time to consider this deeper. 

What would you say are the biggest benefits and risks to the approach you've outlined? To help evaluate it further, you may want to try implementing a couple more examples to get a better feel if this is going in a good direction.

trevan:
@DanVanAtta , the benefits is that this is known to work since the existing system already does this type of serialization replacement (with `GameObjectStreamData`).  It should also allow the code to be more modular since each `Attachable` can manage their own primitive version (as long as it implements the interface) instead of having a central location.

The risks is that it keeps us in the existing system and there might be a better way to do it.

DanVanAtta:
Would you say this moves us to our goal of having "serialized" change objects (written probably as JSON or some other text format)?

Sorry for the delay on this @trevan , it's been difficult for me to really evaluate this update and approach.

trevan:
@DanVanAtta , yes, this moves us to that goal.  The `Attachable` objects are constant objects in the gameData.  They are defined in the XML and do not change during the course of the game.  So, this makes it so that you store the type and name (which is an enum and a string) of the object instead of the object itself and then use the name to locally lookup the object.

PR #8328, opened 2020-12-01
Simplify the Individual and Random AA casualty selection

trevan:
The logic for both the Individual and Random AA casualty selection is
practically identical.  The only difference is when the number of rolls
is equal to the number of targets. And that difference is just in the
fact that a random generator isn't called.

This combines the two methods but keeps the small difference still
around.


DanVanAtta (approved):
Some nice and very meaty methods to combine, good work :+1:

PR #8334, opened 2020-12-02
Grab all the rolls for AA casualty auto selection

trevan:

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelector.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelector.java
@@ -289,7 +289,7 @@ public class AaCasualtySelector {
       // there is a roll for every target but not enough hits to kill all of the targets
       // so no need to get a random set of units since all units will either have a hit
       // or miss roll
-      final List<Die> rolls = dice.getRolls(unitPowerAndRollsMap.getBestStrength());
+      final List<Die> rolls = dice.getRolls();
       hitTargets = new HashSet<>();
       for (int i = 0; i < rolls.size(); i++) {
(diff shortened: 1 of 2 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelector.java
```diff
@@ -289,7 +289,7 @@ private static CasualtyDetails calculateAaCasualties(
       // there is a roll for every target but not enough hits to kill all of the targets
       // so no need to get a random set of units since all units will either have a hit
       // or miss roll
-      final List<Die> rolls = dice.getRolls(unitPowerAndRollsMap.getBestStrength());
+      final List<Die> rolls = dice.getRolls();
```
DanVanAtta, line 292: @trevan can you provide some more background and context to what this change is for and does? It looks valid, just not clear off the bat, kinda seems like a bug fix. Just curious what the background is.

trevan: It is a change I forgot to make when I combined the code.  Previously, when the code went in the individual roll method, it would check that all of the dice strength was the same.  If it wasn't the same, it would go to the random roll method.  If it was, it would then call `getRolls(unitPowerAndRollsMap.getBestStrength())` which would return all of the rolls since it had already checked that the dice strength was the same.

When I combined them, I removed that check of equal strengths since it doesn't matter but I forgot to remove this `getBestStrength` piece.  So now, if the units have different dice, it will not get all of the rolls.

DanVanAtta: Cool - thanks for the context :+1:

PR #8337, opened 2020-12-02
Make resource discovery more flexible and robust

jdimeo:
For more details, see: https://github.com/triplea-game/triplea/issues/8325

Additionally, change the YAML checker to use `HttpGet` instead of `HttpHead`, since GitHub blocks the download of release assets via `HEAD` requests.

Release assets are a better way of releasing TripleA maps (for those that care ;-)) since it allows people to have other files relating to development (I have a `pom.xml` in mine and an `assembly.xml`) without including those files in the final build. It's also a way to accept pull requests on `master` on a map without having that necessarily deploy to everyone else until it's ready.

## Testing

Ensured that the GitHub Actions checks still passed and in a reasonable amount of time

## Screens Shots

N/A

## Release Note

N/A - too technical


jdimeo:
Ugh. I'm not sure why the YAML check is being denied access to the release asset with a code of 403. I tested it from within the game locally and it downloaded just fine (and from 3 different browsers).

jdimeo:
I found a similar issue: https://github.com/cavaliercoder/grab/issues/43

jdimeo:
OK, the YAML check only took about a minute so I think we can safely switch to `GET`

jdimeo:
Interesting. I did an additional trivial commit (remove a </p> from the description) and it kicked off another YAML check. This time one of the maps failed with:
```
https://github.com/triplea-maps/ww2v3_variants/archive/master.zip: map resource not available (500)
```

Is this because GitHub is not liking a lot of `GET` requests but it allows `HEAD` requests en masse? I'm going to kick it off again...

DanVanAtta:
@jdimeo would you mind splitting up the two updates here?

DanVanAtta:
(splitting -> IE: into two PRs)

jdimeo:
@DanVanAtta I didn't because without that minor fix to the gradle file, my PR to add my map fails the download check and breaks the build. Do you still want me to?

gradle/scripts/yaml.gradle
```diff
@@ -75,7 +75,7 @@ def probeUris(uris) {
         def executor = Executors.newFixedThreadPool(maxInFlightProbes)
         uris.each { uri ->
             probes << new UriProbe(uri, executor.submit({
-                def request = new HttpHead(uri)
+                def request = new HttpGet(uri)
```
DanVanAtta, line 78: As an HTTP 'get', this will download the full file, no? The purpose of this check is only to make sure that the file exists at the given URL. The full set of maps is relatively significant. I would presume running this locally or even on a CI server and we'll see a significant overhead of downloading lots of data. Is my understanding correct, or does this update avoid the full content download somehow?

I notice the download URL is different from what is the general pattern, ie: https://github.com/triplea-maps/over-the-top/archive/master.zip

I realize if we go with the latter URL then there is no control over downloading specific versions and the latest is always downloaded. That is a limitation. Assuming a 'head' avoids content download, and the issue is that the download is coming from github releases, could we specify the download by commit SHA?

That would give a URL like: https://github.com/triplea-maps/over-the-top/archive/662e9966c1d7e6f1a61b6f40cfdae69cc0d4af22.zip

DanVanAtta: It occurred to me, using such a link that is commit specific would need testing. There is an unfortunate reliance on zip file naming for the map to be loaded correctly. If the game engine renames the zip file,we would be okay. Though, I'm not sure if the download actually does that rename. I'm therefore not sure that link would work at all in which case we need to consider options assuming a GET is not desirable.

jdimeo: @DanVanAtta I confirmed that the whole file is not downloaded and the check happens very fast (less than a minute). I think we would have to consume the stream of bytes (which we don't) for the check to be long/slow. You can see from the passing checks on this PR.

The reason why I can't use master for my map is that it includes `pom.xml` and `assembly.xml` which confuses TripleA and it errors thinking those are game files. So the commit SHA fixes the problem of always having the latest but doesn't fix the problem of being able to specify your own assembly/zip. 

I confirmed that this PR works for my map- as long as the zip is named "over-the-top.zip" it can be a specific release tag and is still loaded just fine by the game.

Cernelius: As much as I'm not a fan of it (I prefer to refer to them as "game files"), amongst map-makers, when you simply say "XML", you mean the relevant files within the "games" folder. So, in order to avoid confusion, I believe it would be better no other files in any repository use the "XML" format.

Meaning that, whatever they are, I would rather have those "pom" and "assembly" files not in the "XML" format.

---

Side note. Your game file uses at least one deprecated element (here it is an incomplete list of them: https://github.com/triplea-game/triplea/blob/master/docs/map-making/map-features-change-log.md). That is something I was arguing with @DanVanAtta and I believe demostrates my point that simply adding back support for such elements just assures that new maps will be outated too (in this case, you just accepted a game with the deprecated `isTwoHit` option).

---

Talking about reliance on file naming, I let everyone know that, in the past, it was forbidden to have any dashes in such names, since the dash was reserved for mapskins only. Meaning that, in the past, a map name like `over-the-top` was not acceptable (it could have been changed to `over_the_top`, though back them having spaces in map names was acceptable and common, instead).

@DanVanAtta I can tell you that the rule used to be that, if the map file name doesn't have a dash, then it is a map, if it has one, then it is a mapskin, and nothing should have more than one dash. I assume you can ask veq if you want.

jdimeo: @Cernelius Those files are Maven files that I use to manage the map assembly and testing process. I use Maven to zip up the map and automatically copy it to my TripleA folder for testing. I don't have any choice about what file extension they are- they are not part of my "map". TripleA shouldn't be so rigid as to forbid any non-map/game related files in the repository.

Yes, after much testing and tweaking, I realized that it had to be `over_the_top` and I've changed it. I wish it also wasn't so rigid in that regard as well but that was fine once I figured it out.

When I get a chance I'll remove `isTwoHit` but I'm not sure that's important enough from stopping this from going into the Experimental section right? it works fine on the latest release of the engine...

jdimeo: @Cernelius I updated this PR to use v0.1.1 which removed the deprecated property. YAML checks passed again in <= 1min

DanVanAtta: I would prefer to keep the HEAD request if we can, but I'm not opposed to changing it to a GET if it is needed. The reason for the preference is so that it's more obvious and clear there would not be a full download of files, someone else may wonder and start to investigate a "performance problem" that is not there.

 @jdimeo, does it change anything that you actually should not need to zip your map files? The game engine can read maps in an expanded format, unzipped. The folder format when cloning a repository should work fine. In other words, you can clone a map repository to the `triplea/downloadedMaps` folder and it should read fine. Similarly, you can work on a map in an expanded, unzipped, format and then commit any changes and push that to github without needing to do any zip work. Does that change the utility for having build files present, could they be removed in that case?

I do think it was perhaps not well done to have the engine now read any XML file as if it were a game XML. Changing that though won't help us for the current engine version, so we need to do the best we can.

jdimeo: @DanVanAtta I'm struggling here because I thought I've outlined why it's needed and there is no downside- the GET and the HEAD are exactly the same performance wise (you can look at this PR and all the other PRs- the YAML checks take one minute either way). It's only slow if we were to stream the bytes, but just opening a GET and then closing right away is the same as HEAD. This enables us to support release assets which has a lot of benefits as I've already outlined. As a professional software developer, I am very nervous about having the map a) constantly deploy with every change to `master`- I would rather have it be a deliberate tag and b) have every file be part of the deployment. Another file I have in my map is an XLSX I use to edit the unit attachments- I only want that to be present in the repo, not in the final map. So- many reasons to switch to GET and no downsides.... can we just accept this? I've been waiting for literally years to open this PR! :-)

jdimeo: @DanVanAtta P.S. I would actually say that GET is _preferable_ to HEAD, not just "as good as". The point of these checks is to ensure the game can download the map, and so you want the check to be as realistic to what the game is doing as possible. If for some reason, a server _allowed_ HEAD but blocked GET the check would _pass_ here but the game would fail to download the map.

P.P.S. To add more color as to the extra files, I've developed several map making tools that I'm eager to share with the community. For example:
- Writing the game notes in HTML with a CSS/`<style>` block at the top and then a tool to copy and inject those notes into the game XML, inlining the CSS onto the elements (similar to many marketing email tools) vs. having to maintain the notes inside the CDATA directly
- Editing the unit attributes/attachments in an XLSX, where it's easy to compare and change relative power of the units, as well as the starting placements. Automatically update the game file by "syncing" the XLSX with the game XML using JAXB and Apache POI

Both of these require HTML and XLSX "source" files which are then "compiled" into the final game file that should not be included in the "binary" (final .zip) that is downloaded by players of the map. So it's more than just wanting a couple Maven files, though I do want that too.

DanVanAtta: I ran some numbers and below are the results. To explain, I used a tool called 'iftop' to check the total amount downloaded over network and used the 'gradle' build time for the timing. For the 'clean' validateYaml case I interleaved running builds from 'master' and 'jdimeo/patch1'. For the non-clean case I ran the same build 4 times in a row throwing out the first build and did so on both 'master' and 'jdimeo/patch1'.


```
--- ./gradlew clean validateYaml 
Master1   33s & 497KB downloaded
Master2   13s & 477KB downloaded
Master3   26s & 484KB downloaded


patch1   35s & 9.83MB  
patch2   25s & 9.50MB
patch3   27s & 9.00MB
```

```
--- (non-clean) ./gradlew validateYaml

master1: 8s
master2: 14s
master3: 9s

patch1: 25s
patch2: 25s
patch3: 24s
```

DanVanAtta: @djimeo I want to get this map listing merged and I want this to be as painless as possible. Patience is a virtue, after a year, a few more days is not a significant amount of time. I don't like being 'the bad guy' on this, though we need to consider how changes will impact everyone. We run 'validateYaml' as part of branch verification, it is not only run on PR builds but by all devs locally as well, and (hopefully) quite frequently. Adding time and download weight to that process is considerable particularly considering the efforts put in to try and get it as fast as we can get it. The numbers above are an odd in-between where I don't know if they are deal-breakers, it is not quite parity.

I like the process and tooling you have developed and I believe it is superior compared to what we have, and I would prefer for map downloads to be down via releases instead  of a full zip of the map repository. The reason we use the latter is for simplicity. The amount of resistance to interacting with github was very considerable as it has been baked in culturally for a long time that map-makers do not need to interact with the map hosting, for a very long time it was the job of the maintainers to do that. Now that it is 'self-service', the cultural shift has been difficult. The reason we no longer use releases is just simplicity, to make the process as easy as possible on everyone with as few moving pieces as possible. The automatic 'zip' download of a repository can almost be thought of as a mini CI/CD for maps, a very simple one.

I think there are some really good ideas for simplifying the map making process and a lot of it is very much more difficult than it needs to be. A focus of the 2.4 release was to change how map XML is parsed to remove the need for a DTD and allow us to more easily change the XML structure. I think we are now positioned to help make some of these tasks easier.

I think we have a longer conversation here that should be followed up on, but to address the tooling built:
- XLSX transform: I wonder how much of this is compensating for the current XML structure that could be redone in a way to reduce duplication and generally simplify. Is the solution to have a better XML?
- game notes: I'd like to have the engine read the game notes from an HTML file that is separate from the game-XML file: https://forums.triplea-game.org/topic/2238/move-game-notes-to-its-own-file

The deployment and download to releases is a bit non-standard. Long term I am not sure if that would 'jive' well with the potential plans to automate the 'triplea_maps.yaml' file. Automating the 'triplea_maps.yaml' file would involve a backend server to scan for repositories and then make available the download links via an HTTP request from the game client to that backend server. I have not considered how to handle the case where some map repositories use releases and others do not. I am not sure how to account for that and build that into the automation of the map server. I'm interested in your thoughts and ideas regarding these topics, though it seems like that is a longer conversation that does not belong in this PR.

DanVanAtta: @jdimeo, here is my recommendation, which I do not entirely like:
- extract the map tools you have built and put them in a new repository
- delete the tools from the map repository and add those entries to a .gitignore file
- hold your nose and use 'master' for latest releases

I don't like the last suggestion but it keeps the repo in line with others and resolves the need for a HEAD vs GET request. I share your worry about using master always being released, but if you use feature branches for potential new versions, then effectively whenever you merge to master you have an instantaneous release. Perhaps that perspective helps.

With the above, we have some interesting map process and design questions to answer and I'd really like to talk through that more and get your thoughts and you have help shape the next iteration of map hosting.

Alternatively, we'll need to consider a bit more the implications of using 'releases' for download and whether we want a potential short term solution. The download results are not great, but not a deal breaker, the extra time for non-clean builds is certainly a con and arguably is a deal breaker. We'd want to consider that in more detail as well, perhaps even evaluate options such as removing the URI check altogether (which doesn't seem like the best option, but I'm not sure offhand what good options there).

Please let me know your thoughts here @jdimeo, again, I'm interesting/eager to make the map available as soon as possible, and I think you have used a better and more advanced map making process which solves problems which I would like to see solved as smoothly as possible for everyone and incorporated into the standard map making procedure.

jdimeo: @DanVanAtta Wow, I really opened a can of worms didn't I! A few thoughts:
- Thank you for taking the time to explain all of this to me- I know there are a bunch of forum posts, tickets, PRs, etc. that you are maintaining and I really appreciate the effort you put in here. I definitely lack a lot of the history/context (e.g. resistance to GitHub)
- In general, I'm hugely impressed with TripleA and what it has accomplished with volunteer labor
- But yes, I have a lot of ideas on how to help :-) I actually was thinking *more* structure (e.g. an XSD schema with nice auto-complete in an IDE with JAXB generated code) would be helpful, not less as you alluded to in 2.4, but we can definitely talk about that more later- out of scope here!
- The map tools are already in another repo here: https://github.com/jdimeo/triplea-map-tools  The files that are checked into the map are like "source" files (e.g. the notes HTML) but the code that injects them into the game file is in `triplea-map-tools`. P.S. the CSS inlining is a big win there, since "normal" CSS wouldn't be rendered correctly by the Swing HTML widget right? it would have to be a "real" web browser window like SWT?
- I think I've found a solution that works for me as a "power user" but also works for the current conventions: Create a separate repo with my build files and ancillary files, but _submodule in_ the map into the `/map` directory into the support repo. Then, my build tools can edit the map files "directly" by writing to e.g. `/map/games/over_the_top.xml`, but the commits go to the real map repo which stays "clean". Does that sound good to you?

jdimeo: Ugh, I set up https://github.com/jdimeo/over-the-top-support to look exactly like my repository used to look, using a submodule to bring the map into `/map`. The main map repo is now "flat" with all the map files in the root directory.

This seems to break TripleA. Apparently, you can either have it "flat" in the zip OR you can have `/map_name-master` as the root folder of the zip but then the map files MUST be under a `/map` folder. Is that correct? Is there an easy way to fix this expectation?

DanVanAtta: The problem is that the name of the map found in each XML no longer matches the name of the 'root' folder (over-the-top-support), is that the case? Do I understand correctly the issue?

Removing that dependency has been on my list for some time. It's a current problem though.

jdimeo:
@DanVanAtta I'm not sure you're going to like this change, but I think it's an improvement.
- I reverted the change to `HttpGet` back to `HttpHead`
- I reorganized my repo so that **only** the map files are inside, so it can be downloaded via the master archive .zip
- However, I had to keep my files at the root of my repository. This is because I am using Git submodules to import them into a separate "support" repo for developing the map, and I am pulling it in there under `/map`

See the attached images - note how the map is now "flat" and the support repo shows "map @ 4c0b84a"

This means though that although my .zip's root entry is "over_the_top-master/", the files are directly under the root, and not under a "/map" entry. `ResourceLocationTracker` had a `MASTER_ZIP_MAGIC_PREFIX = "-master/map/";` that assumed this structure that broke my map.

I've improved the resource detection to find the .zip entry for `baseTiles/` anywhere in the zip (regardless of map name, branch name, .zip structure, etc.). Now my map works and the code is much less opinionated/flexible as far as finding the resources in the .zip.

image
image

DanVanAtta:
@jdimeo , if we require an updated game engine to load the map, then those updates won't benefit your new map now. The triplea_maps.yaml file is visible to all game engines, updates to it are live immediately on merge. That would mean that for most users downloading your map would run into errors. 

At this juncture it could make sense to split up the new updates. WDYT on how to proceed?

jdimeo:
@DanVanAtta Good point. I renamed this PR to reflect the contribution to `ResourceLocator`. I'll just force my map repo to be the right layout for now and worry about setting up for future work later so we can get it out there. I'll start a new PR for that.

game-core/src/main/java/games/strategy/triplea/ResourceLocationTracker.java
```diff
@@ -10,42 +15,28 @@
...
-    // map skins will have the full path name as their map name.
-    if (mapName.endsWith("-master.zip")) {
-      return mapName.substring(0, mapName.length() - "-master.zip".length())
-          + MASTER_ZIP_MAGIC_PREFIX;
-    } else {
-      return isUsingMasterZip ? mapName + MASTER_ZIP_MAGIC_PREFIX : "";
-    }
+  	static String getMapPrefix(final URL[] resourcePaths) {
+  		for (val url : resourcePaths) {
+  			try (val zip = new ZipFile(new File(url.toURI()))) {
```
DanVanAtta, line 30: - Have you checked the performance of this method?
- Have you tested for the case where the map is a directory and not a zip?

game-core/src/main/java/games/strategy/triplea/ResourceLocationTracker.java
```diff
@@ -10,42 +15,28 @@
  */
 @UtilityClass
 class ResourceLocationTracker {
+  private static final String REQUIRED_ASSET_FOLDER = "baseTiles/";
```
DanVanAtta, line 18: The variable name is a bit odd as there are other required asset folders, yet the name of the folder is 'baseTiles'. Is this just an example required folder to indicate when we have found the map resources? Does the actual folder matter less so than it is just one of several that we would expect to find?

DanVanAtta:
@jdimeo we use 'google-java-formatter', you should enable it for this project. You can automatically apply formatting as well with `./gradlew spotlessApply`

game-core/src/main/java/games/strategy/triplea/ResourceLocationTracker.java
jdimeo: Yes, precisely. `baseTiles` is necessary, but not sufficient, to specify assets but sufficient to identify the asset path

game-core/src/main/java/games/strategy/triplea/ResourceLocationTracker.java
jdimeo: 1. Only by running locally and not observing any slowdown, but that was just anecdotal
2. I don't think so since I believe I only have .zips locally right now. If it's just a folder and it attempts to open as a .zip, it will throw an IOException and silently return an empty string (which was the previous behavior for non-zips)

game-core/src/main/java/games/strategy/triplea/ResourceLocationTracker.java
jdimeo: @DanVanAtta would you prefer `EXAMPLE_REQUIRED_ASSET_FOLDER`?

jdimeo:
> @jdimeo we use 'google-java-formatter', you should enable it for this project. You can automatically apply formatting as well with `./gradlew spotlessApply`

I noticed that some of the checks claimed that I didn't make a variable `final` but I really did via Lombok `val`. Is there a way to change that check or is it not worth the hassle?

game-core/src/main/java/games/strategy/triplea/ResourceLocationTracker.java
jdimeo: Renamed to `REQUIRED_ASSET_EXAMPLE_FOLDER` and added comment

DanVanAtta:
I do think the checkstyle rules as they currently are, are of mixed value.
Checkstyle regrettably does not understand that the lombok 'val' is the same as `final var`

jdimeo:
@DanVanAtta smoke test failed for reasons that are ether ephemeral or unrelated to this PR, but I've fixed the style issues

game-core/src/main/java/games/strategy/triplea/ResourceLocationTracker.java
```diff
@@ -10,42 +15,30 @@
...
-    // map skins will have the full path name as their map name.
-    if (mapName.endsWith("-master.zip")) {
-      return mapName.substring(0, mapName.length() - "-master.zip".length())
-          + MASTER_ZIP_MAGIC_PREFIX;
-    } else {
-      return isUsingMasterZip ? mapName + MASTER_ZIP_MAGIC_PREFIX : "";
+  static String getMapPrefix(final URL[] resourcePaths) {
+    for (final URL url : resourcePaths) {
+      try (ZipFile zip = new ZipFile(new File(url.toURI()))) {
+        final Optional<? extends ZipEntry> e =
```
DanVanAtta, line 32: nit, the single letter variable names can be hard to decipher. Would you mind replacing it with something more descriptive, like zipFile, or fileInZip, or zipFileEntry?

Similarly for the `$` in the lambda.

game-core/src/main/java/games/strategy/triplea/ResourceLocationTracker.java
```diff
@@ -10,42 +15,30 @@
...
+            zip.stream().filter($ -> $.getName().endsWith(REQUIRED_ASSET_EXAMPLE_FOLDER)).findAny();
+        if (e.isPresent()) {
+          final String path = e.get().getName();
+          return path.substring(0, path.length() - REQUIRED_ASSET_EXAMPLE_FOLDER.length());
```
DanVanAtta, line 36: I notice we are substringing off only the example folder name. Seemingly we want the path to be something like `<map-name>/map`. If the 'baseTiles` is not in that directory, then we might substring too little. For example, let's say `baseTiles` is in a sub-folder of 'map', then the path would no longer be to '/map' but would be to that subfolder. Wouldn't this cause a problem with files being one-folder level higher than expected?

DanVanAtta (commented):
Sorry for the delay. Took a detailed look at this diff, it is nice and bite-sized, thank you @jdimeo .

Left one comment/request for change a question about functionality.

game-core/src/main/java/games/strategy/triplea/ResourceLocationTracker.java
jdimeo: If I understand correctly, this is exactly the problem that this PR is trying to solve. If the map maker for some reason puts the assets in `/map/stuff/baseTiles` this code will use `/map/stuff` as the root path of the resources, not `/map` as was previously hardcoded and assumed. This stops making assumptions about absolute paths for resources and instead makes the assumption that, wherever `baseTiles` is, all the other folders are siblings to that folder.

game-core/src/main/java/games/strategy/triplea/ResourceLocationTracker.java
jdimeo: I like using single character variable names to communicate to the reader it's not an "important" variable- very often in lambdas or for loops. I'm telling you as the reader to "keep going" and focus on other parts of the code. But I'm happy to conform to the current conventions in the code.

jdimeo:
> Sorry for the delay. Took a detailed look at this diff, it is nice and bite-sized, thank you @jdimeo .

No worries and no rush- thank you for the time you spend maintaining this and happy holidays

game-core/src/main/java/games/strategy/triplea/ResourceLocationTracker.java
```diff
@@ -10,42 +15,30 @@
...
+        }
+      } catch (final IOException | URISyntaxException e) {
+        // File is not a zip or can't be opened
```
DanVanAtta, line 39: IOException is highly suspicious. We may want to tell a user about this if we hit it. Perhaps we expect this to be an impossible condition, though if the permissions are wrong then it might be worth notifying the user.

If we first check that a file exists: `new File(url.toURI()).exists()`, and then we get an IOException, it seems like that would be a good place to warn a user. WDYT? @jdimeo 

Beyond this last point, this update looks good.

jdimeo: @DanVanAtta My mindset here is that this is just auto resource location, not actual loading and parsing. We expect an IOException if the file is a folder (not a zip file) or if there are any other reasons it can't be read, but in that case we should just return a default/assumed path to the resources (like `/map`) not actually throw errors. We can catch and throw errors when we go to actually load and parse the map.

DanVanAtta: Cool, I'm wondering & considering if we are doing control-flow-by-exception-handling. We could check in the `new File(..)` that it is a file, ends with '.zip' and if not then avoid processing the file. This would leave IOException for cases that we do not expect and potentially represent error cases. I'm concerned that by no-op'ing the IOException case that we could be silently suppressing an error that we have not expected.

jdimeo: @DanVanAtta What you are describing is exactly what we should do for the actual map loading code. I see this code as very "lazy" - blindly and quickly try to automatically find the resources, purposefully ignoring errors and not being robust. All that you are describing would be appropriate for the actual map parsing code. I wouldn't want a stack trace with a map error to lead a troubleshooter here, since this isn't the "authority" on parsing the map

game-core/src/main/java/games/strategy/triplea/ResourceLocationTracker.java
```diff
@@ -10,42 +15,30 @@
...
+      }
     }
+    return "";
```
jdimeo, line 42: @DanVanAtta should the default be different here? e.g. `/map`? I'm assuming not, since if it's not a zip, it won't be under a `/map` folder typically.

game-core/src/main/java/games/strategy/triplea/ResourceLocationTracker.java
jdimeo: @DanVanAtta I just pushed another commit that removes the single char variable names

game-core/src/main/java/games/strategy/triplea/ResourceLocationTracker.java
DanVanAtta: Keeping existing behavior is prudent unless we are sure it should be changed. I would agree it is suspect and possibly dead-code.

game-core/src/main/java/games/strategy/triplea/ResourceLocationTracker.java
DanVanAtta: I see, I agree that error handling should be handled upstream. It looks like it is not quite though.. 

From ResourceLoader.java:
```
  private ResourceLoader(final String mapName, final String[] paths) {
    final URL[] urls = new URL[paths.length];
    for (int i = 0; i < paths.length; i++) {
      final File f = new File(paths[i]);
      if (!f.exists()) {
        log.error(f + " does not exist");
      }
      if (!f.isDirectory() && !f.getName().endsWith(".zip")) {
        log.error(f + " is not a directory or a zip file");
      }
      try {
        urls[i] = f.toURI().toURL();
      } catch (final MalformedURLException e) {
        throw new IllegalStateException(e);
      }
    }
    mapPrefix = ResourceLocationTracker.getMapPrefix(mapName, urls);
    // Note: URLClassLoader does not always respect the ordering of the search URLs
    // To solve this we will get all matching paths and then filter by what matched
    // the assets folder.
    loader = new URLClassLoader(urls);
    this.mapName = mapName;
  }
```

PR #8338, opened 2020-12-03
Rework the auto low luck casualty selection

trevan:
Rewrite the old comments to fit better with how the game now works. When
this was first worked on, all AA were infinite, the dice sides was
always 6, and AA only had a strength of 1 or 2.

The new logic is a little different from the old logic. The old logic
used a little more randomness than the new logic.  But that randomness
doesn't appear to improve the functionality so it was removed.

## Testing

Ran Hard AI of Warcraft and TWW

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelector.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelector.java
@@ -1,6 +1,6 @@
 package games.strategy.triplea.delegate.battle.casualty;
 
-import com.google.common.collect.Lists;
+import com.google.common.base.Preconditions;
 import games.strategy.engine.data.GameData;
 import games.strategy.engine.data.GamePlayer;
 import games.strategy.engine.data.Territory;
@@ -11,18 +11,13 @@ import games.strategy.triplea.Properties;
 import games.strategy.triplea.attachments.UnitAttachment;
 import games.strategy.triplea.delegate.BaseEditDelegate;
 import games.strategy.triplea.delegate.DiceRoll;
-import games.strategy.triplea.delegate.Die;
 import games.strategy.triplea.delegate.Die.DieType;
 import games.strategy.triplea.delegate.Matches;
 import games.strategy.triplea.delegate.data.CasualtyDetails;
 import games.strategy.triplea.delegate.power.calculator.AaPowerStrengthAndRolls;
 import games.strategy.triplea.delegate.power.calculator.CombatValue;
-import games.strategy.triplea.util.UnitCategory;
-import games.strategy.triplea.util.UnitSeparator;
-import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Deque;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -30,7 +25,7 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.function.Function;
 import java.util.stream.Collectors;
-import lombok.Value;
+import java.util.stream.IntStream;
 import lombok.experimental.UtilityClass;
 
 @UtilityClass
@@ -79,10 +74,12 @@ public class AaCasualtySelector {
         AaPowerStrengthAndRolls.build(defendingAa, planes.size(), aaCombatValueCalculator);
     final List<Unit> availableTargets = calculateAvailableTargets(planes, allowMultipleHitsPerUnit);
 
-    return Properties.getLowLuck(data.getProperties())
-            || Properties.getLowLuckAaOnly(data.getProperties())
-        ? getLowLuckAaCasualties(availableTargets, unitPowerAndRollsMap, dice, bridge)
-        : calculateRolledAaCasualties(availableTargets, unitPowerAndRollsMap, dice, bridge);
+    return buildCasualtyDetails(
+        availableTargets,
+        Properties.getLowLuck(data.getProperties())
+                || Properties.getLowLuckAaOnly(data.getProperties())
+            ? getLowLuckAaCasualties(availableTargets, unitPowerAndRollsMap, dice, bridge)
+            : calculateRolledAaCasualties(availableTargets, unitPowerAndRollsMap, dice, bridge));
   }
 
   /**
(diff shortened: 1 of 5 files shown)


trevan:
The unit test failures don't appear to be related to this PR.

game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelector.java
```diff
@@ -114,265 +112,209 @@ private static CasualtyDetails getLowLuckAaCasualties(
...
-      // so no need to get a random set of units since all units will either have a hit
-      // or miss roll
-      final List<Die> rolls = dice.getRolls();
-      hitTargets = new HashSet<>();
-      for (int i = 0; i < rolls.size(); i++) {
-        if (rolls.get(i).getType() == DieType.HIT) {
-          hitTargets.add(i);
-        }
+    final int guaranteeHitGroupSize =
+        calculateGuaranteeLowLuckHitGroupSize(targets, diceRoll, unitPowerAndRollsMap);
```
DanVanAtta, line 172: Is this a place where we could early terminate and avoid processing on empty lists?

EG:
```
    final int guaranteeHitGroupSize =
        calculateGuaranteeLowLuckHitGroupSize(targets, diceRoll, unitPowerAndRollsMap);
   if(guaranteedHitGroupSize == 0) {
           return LowLuckTargetGroups.NO_GUARANTEED_HITS;
    }
```

Note, this comment was made out of order for other comments below, this could moot some of those suggestions.

DanVanAtta (commented):
I made it about half way through this update. I hope to return to this perhaps tomorrow or Sunday evening to finish up.

Overall my impression is quite positive, things are for sure becoming simpler & more clear. You're certainly brave to be diving into this code @trevan .

game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelector.java
```diff
@@ -114,265 +112,209 @@ private static CasualtyDetails getLowLuckAaCasualties(
...
+   *     size that guarantees a hit
+   */
+  private static int calculateGuaranteeLowLuckHitGroupSize(
+      final Collection<Unit> availableTargets,
+      final DiceRoll diceRoll,
+      final AaPowerStrengthAndRolls unitPowerAndRollsMap) {
+    final int bestStrength = unitPowerAndRollsMap.getBestStrength();
+    final int chosenDiceSize = unitPowerAndRollsMap.getDiceSides();
+
+    final boolean hasOverstackHits =
```
DanVanAtta, line 217: It's not immediately clear what an overstack hit is. Maybe a quick comment explaining the term, or a different variable name would make this clear? I struggled here a bit and then reading a bit more later about the overstack AA hits it made more sense (but I'm still not sure I understand the concept TBH).

game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelector.java
```diff
@@ -114,265 +112,209 @@ private static CasualtyDetails getLowLuckAaCasualties(
...
+    } else if (hits < availableTargets.size()) {
+      // there isn't a roll for every target so need to randomly pick the target for each hit
+      hitTargets = findRandomTargets(availableTargets, bridge, hits);
+    } else {
+      // all targets were hit so add them all
+      hitTargets = availableTargets;
     }
-    return Tuple.of(groupsOfSize, toRoll);
+
+    return buildCasualtyDetails(availableTargets, hitTargets);
```
DanVanAtta, line 305: Perhaps a style preference on my part, to another extent I see we are doing the 'single-return' point design as well. I think some this method block would be relatively clear returning when we can. For example, the last 'else' statement would avoid some extra thought to realize that `availableTargets' is equal to 'hitTargets'.

Personally I've come to find the single-return point design is only really good for longer methods, ones that are complicated and return for different reasons at different places. For short methods, something like the below is just fine IMO:
```
if
   return
else if
   return
else
   return
```

The above example I'd say is more clear than the single return variable, ie:
```
var returnVal;
if
   returnVal = a
else if
   returnVal =b
else
   returnVal =c

return returnVal
```

game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelector.java
```diff
@@ -114,265 +112,209 @@ private static CasualtyDetails getLowLuckAaCasualties(
...
   }
 
-  private static int getTotalHitpointsLeft(final Unit unit) {
-    if (unit == null) {
-      return 0;
+  /** Find the targets that were hit by a dice roll */
+  private static Collection<Unit> findRolledTargets(
+      final List<Unit> availableTargets, final DiceRoll dice) {
+    final List<Die> rolls = dice.getRolls();
```
DanVanAtta, line 311: Would it be too nitpicky to question why there is an intermediate `rolls` variable? `dice.getRolls()` is just as descriptive and it's only used a couple times below. Removing intermediate variables is often a good thing, one less thing for a maintainer to keep on their mental stack when reading code. At this point when reading the method, there are 3 variables in scope, soon to be a 4th one, leaves little room for much else!

game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/AaCasualtySelector.java
```diff
@@ -114,265 +112,209 @@ private static CasualtyDetails getLowLuckAaCasualties(
...
+    final List<Unit> hitTargets = new ArrayList<>();
+    for (int i = 0; i < rolls.size(); i++) {
+      if (rolls.get(i).getType() == DieType.HIT) {
+        hitTargets.add(availableTargets.get(i));
```
DanVanAtta, line 315: It looks like there is an implicit assumption that avaialbleTargets will have the same size as 'dice.getRolls'. This assumption is not necessarily clear here, a `Precondition.checkArgument(availableTarget.size() == dice.getRolls().size())` at the start of the method would help document this. A reader would do a double take here wondering if there could be an indexing problem because `i` is tracking against `rolls`, not `availableTargets`.

DanVanAtta (commented):
Finished looking over this. The updates does make for better code!
I left a number of design suggestions and a few nits, nothing too critical or that I would insist on, all suggestions. Please have a look @trevan and incorporate what makes sense and feel free to discuss further please.

trevan:
@DanVanAtta , I like the suggestion to make the `LowLuckTargetGroups` a more complete object.  I've moved it to its own file, moved the "construction" code to it and made it a true constructor and encapsulated a little more logic in it.

I also changed where the `buildCasualtyDetails` is called so that the methods can return earlier.

DanVanAtta (approved):

PR #8341, opened 2020-12-04
Add player name to the DiceRoll to keep track of players randomness

trevan:
The engine currently requires that the annotation stored with the
DiceRoll in the history has the player name at the beginning. This
change breaks that pairing so that the history annotation can be
different or missing but the players randomness will still be tracked.

## Testing

I've loaded a save game to make sure that it loads still.  I've continue with that game, saved it, and then loaded it to make sure a saved game with both types are loaded correctly.  I stepped through the code as well to make sure that it actually was saving and loading the data correctly.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -94,7 +94,7 @@ subprojects {
         jaxbApiVersion = '2.3.1'
         jaxbCoreVersion = '3.0.0'
         jaxbImplVersion = '3.0.0'
-        jdbiVersion = '3.17.0'
+        jdbiVersion = '3.18.0'
         jlayerVersion = '1.0.1.4'
         junitJupiterVersion = '5.7.0'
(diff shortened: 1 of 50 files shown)


trevan:
With this change, I can remove most of the annotation stuff in DiceRoll and in the factory methods.  That's for a later PR.

This PR will cause a conflict with https://github.com/triplea-game/triplea/pull/8338 and https://github.com/triplea-game/triplea/pull/8308.

DanVanAtta:
This update looks good @trevan . Would you prefer to wait for the preceding PRs to be merged first, or merge this one now and deal with the conflicts in the existing PRs?

trevan:
@DanVanAtta , one of the PRs is already merged so I dealt with those conflicts.  Go ahead and merge this now.  I'm working on the other PR so I can resolve the conflicts during that work.

PR #8349, opened 2020-12-05
Add a clear casualty step for naval bombardment

trevan:
All of the other battle parts (aa, sub, other) use a casualty removal
step to remove casualties that have been hit. This updates the
bombardment to also have its own casualty removal step for the case
where the map doesn't allow naval bombardment casualties to fire back.

## Testing

Played World At War as it doesn't allow casualties of naval bombardment to fire back.  Modified it to allow selection of AA casualties.  I created a save of a battle at the point of selecting the AA casualties.  I then loaded that save in this branch and ensured that the naval bombardment casualties were removed correctly.  I also loaded a save just before the battle started to ensure that the initial battle setup would be correct as well.

I noticed that I had introduced a bug when I did the sub rework.  If a battle has both bombardment and a first strike unit, then the bombardment casualties will always be removed after the first strike fires, even if "Naval Bombard Casualties Return Fire" is true.  I'm not sure if there is a map that would be impacted by that bug.  I'm working on a fix for it but it will probably take a few PRs.

## Release Note

CHANGE|If 'Naval Bombard Casualties Return Fire' is false, then a new step is added to the battle ui to indicate when the casualties are removed from battle.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/ClearBombardmentCasualties.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/ClearBombardmentCasualties.java
@@ -21,7 +21,7 @@ public class ClearBombardmentCasualties implements BattleStep {
 
   @Override
   public List<String> getNames() {
-    return clearCasualties() ? List.of(REMOVE_BOMBARDMENT_CASUALTIES) : List.of();
+    return valid() && clearCasualties() ? List.of(REMOVE_BOMBARDMENT_CASUALTIES) : List.of();
   }
 
   @Override
@@ -31,7 +31,7 @@ public class ClearBombardmentCasualties implements BattleStep {
 
   @Override
   public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
-    if (clearCasualties()) {
+    if (valid() && clearCasualties()) {
       battleActions.clearWaitingToDieAndDamagedChangesInto(bridge);
     }
   }
@@ -39,5 +39,11 @@ public class ClearBombardmentCasualties implements BattleStep {
   private boolean clearCasualties() {
     return !Properties.getNavalBombardCasualtiesReturnFire(
         battleState.getGameData().getProperties());
+  }
+
+  private boolean valid() {
+    return battleState.getStatus().isFirstRound()
+        && !battleState.getBombardingUnits().isEmpty()
+        && !battleState.getBattleSite().isWater();
   }
 }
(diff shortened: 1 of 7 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/ClearBombardmentCasualties.java
```diff
@@ -0,0 +1,49 @@
...
+      battleActions.clearWaitingToDieAndDamagedChangesInto(bridge);
+    }
+  }
+
+  private boolean clearCasualties() {
+    return !Properties.getNavalBombardCasualtiesReturnFire(
+        battleState.getGameData().getProperties());
+  }
+
+  private boolean valid() {
```
DanVanAtta, line 44: nit, valid could be more descriptive. When reading the usage, I wondered "what is valid?"
Some rename suggestions:
-  "isBombardmentRound"
- "bombardmentOccurred"
- "bombardUnitsWillFire"

Side-note, I wonder if it would make sense to create a `BombardmentRules` object so we can centralize logic like this, where we have one source  to detect, when/how/if bombardment should occur.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/MarkCasualties.java
```diff
@@ -45,9 +44,7 @@
 
   private final FireRoundState fireRoundState;
 
-  @ChangeOnNextMajorRelease(
-      "returnFire is ALL for everything except NavalBombardment and old saves."
-          + "Rework so that returnFire isn't needed at all.")
+  @RemoveOnNextMajorRelease("returnFire is ALL for everything except old saves")
```
DanVanAtta, line 47: side comment, just realized the existing comment does not quite make sense.
Specifically "ALL for everything" reads odd, without context it does not really make sense.
If ALL is an enum or a constant defined somewhere, perhaps fully qualifying it would remove any confusion.

DanVanAtta (commented):
Just one comment, looks good otherwise :+1:

trevan:
@DanVanAtta , I've updated the comment to have the enum name `ReturnFire`.  And I've renamed the method to `canBombardmentOccur`.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/ClearBombardmentCasualties.java
trevan: I just wrote my thoughts on how to centralize the logic using the game XML in the forum.  You can see my thoughts at https://forums.triplea-game.org/topic/2481/custom-battle-phases.  That forum post encompasses more than just centralizing bombardment logic.

DanVanAtta (approved):

PR #8352, opened 2020-12-05
Use BattleState.Side instead of a boolean for units being removed

trevan:


game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/fire/firststrike/ClearFirstStrikeCasualtiesTest.java
```diff
@@ -37,134 +36,118 @@
 
   @ParameterizedTest
   @MethodSource
-  void getStep(
-      final List<BattleStateVariation> parameters,
-      final boolean willExecute,
```
trevan, line 42: While updating this test for the boolean -> side change, I noticed that `willExecute` isn't actually needed so I removed it as part of this PR.

DanVanAtta (approved):
:+1:

PR #8368, opened 2020-12-06
Remove duplicate code in RemoveUnits by using BattleState.Side

trevan:
Instead of copying the code for both offense and defense units, use a
helper method that takes a side.


trevan:
The smoke test failure doesn't appear to be related to this PR.

DanVanAtta (approved):

PR #8369, opened 2020-12-06
Move BattleActions#removeCasualties to BattleState#markCasualties

trevan:
Remove the code that depends on the bridge to the caller so that
markCasualties can be in the BattleState. The code that depends on the
bridge is old code that will be removed once save compatibility can be
broken.

Modify MustFightBattle#remove to remove the units from the WaitingToDie
lists so that markCasualties can always move the units to the
WaitingToDie list but the caller can then also remove those units if
needed.

## Testing

Played a save from 2.0 that was a sub battle.  Ensured that the ReturnFire was handled correctly.  Units that should return fire were left in the WaitingToDie lists and units that shouldn't return fire were removed from that list.

Played a brand new sub battle to make sure that the current MarkCasualties and ClearFirstStrikeCasualties steps worked correctly together.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
@@ -92,6 +92,16 @@ public interface BattleState {
 
   Collection<Unit> removeNonCombatants(Side side);
 
+  /**
+   * Mark the units that will be dying
+   *
+   * <p>Units that are only damaged should not be passed into this method.
+   *
+   * @param casualties units that are dying
+   * @param side the side the unit are on
+   */
+  void markCasualties(Collection<Unit> casualties, Side side);
+
   Collection<Unit> getBombardingUnits();
 
   GamePlayer getPlayer(Side side);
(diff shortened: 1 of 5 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
```diff
@@ -92,6 +92,16 @@ public boolean isFirstRound() {
...
 
+  /**
+   * Mark the units that will be dying
+   *
+   * <p>Units that are only damaged should not be passed into this method.
+   *
+   * @param casualties units that are dying
+   * @param side the side the unit are on
+   */
+  void markCasualties(Collection<Unit> casualties, Side side);
```
trevan, line 103: Which would you prefer, the javadoc on the interface, the implementation, or both?

DanVanAtta: IMO interface to explain the general contract that is expected to be provided by any implementation.

DanVanAtta (approved):

PR #8371, opened 2020-12-07
Remove the call to MustFightBattle#remove from damagedChangeInto

trevan:
MustFightBattle#remove does more work than just remove the unit from the
battle. If the unit type is marked to convert to another unit on death,
it will convert it. And it checks for dependent units and kills them.

Neither of those are needed when converting a damaged unit into another
unit, especially the extra conversion check. If a unit type had a
conversion requested when it had 1 hit point left and 0 hit points left,
it would never do the 1 hit point left conversion because of the extra
conversion in remove().

## Testing

Played TWW and Warcraft.  Both of those have units that transform when they take damage.  Warcraft's units generally transform when they die and TWW's transform when they have 1 hp left.  I checked the history to make sure it looked right and checked that the units actually transformed.

Making this change, I realized that dependent units of the transformed unit are currently killed. I don't think that is actually wanted. This change no longer kills them but it also breaks the dependency between the units.  So if there was a transport that had units in it and the transport transforms, the transported units will no longer be transported.

I'm not sure how it should be changed.  Do we want the dependent units to die, should they be connected to the new unit, or something else?

I went through the games in the repo and I couldn't find any game that would currently be affected by it.  So it might be ok to just take this change and then figure out the answer later.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
@@ -552,7 +552,8 @@ public class MustFightBattle extends DependentBattle
             attacker,
             attackingUnits,
             CollectionUtils.getMatches(killedDuringCurrentRound, Matches.unitIsOwnedBy(attacker)),
-            bridge);
+            bridge,
+            side);
       } else {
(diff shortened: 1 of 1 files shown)


trevan:
I'm closing this for now as I work on the dependents.

PR #8388, opened 2020-12-09
"Ensure the player name in the step definition is valid (#8384)"

DanVanAtta:
@trevan I merged the last one too early.

This reverts commit 8008f5456dc3d2a5cc5a906994dd3dc01fe600da.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
+++ b/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
@@ -555,6 +555,16 @@ public final class GameParser {
     for (final GamePlay.Sequence.Step current : stepList) {
       final IDelegate delegate = getDelegate(current.getDelegate());
       final GamePlayer player = getPlayerIdOptional(current.getPlayer()).orElse(null);
+      if (player == null && current.getPlayer() != null && !current.getPlayer().isBlank()) {
+        log.log(
+            Level.SEVERE,
+            "The step "
+                + current.getName()
+                + " wants a player with the name of '"
+                + current.getPlayer()
+                + "' but that player can not be found. "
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
```diff
@@ -555,6 +555,16 @@ private void parseSteps(final List<GamePlay.Sequence.Step> stepList) throws Game
     for (final GamePlay.Sequence.Step current : stepList) {
       final IDelegate delegate = getDelegate(current.getDelegate());
       final GamePlayer player = getPlayerIdOptional(current.getPlayer()).orElse(null);
+      if (player == null && current.getPlayer() != null && !current.getPlayer().isBlank()) {
+        log.log(
```
DanVanAtta, line 559: @trevan this I think should be a game parse exception. You can loop it into the 'orElse' above and change the 'orElse' to be a 'orThrow'.

The game parse exception I believe will trigger messaging to the user. We do not want a logger here as we do not want this as an error report (the problem needs to be fixed by the map maker). Also we do not want processing to continue, the map is going to be busted in such a condition. Finally, the smoke test will verify parsing of almost all existing maps, as an exception we'll see which others fail.

trevan: I created #8391 with the exception instead of the log.  I didn't loop it into the `orElse` because I'm not sure if a step is allowed to be missing the player.  So, the logic is to make sure that a player is actually wanted but that the player couldn't be found.  Do you know if steps always require a player?

DanVanAtta: I think it really should, I don't know if that is enforced though.

PR #8392, opened 2020-12-09
Refactor UnitUtils#translateAttributesToOtherUnits

trevan:

I'll be modifying this function to translate dependent units so this just reworks the code.  One thing I'm worried about is that this could create a lot of CompositeChanges while the original code only created one.


DanVanAtta (approved):
Looks okay @trevan 
Thank you for calling out the concern, how concerned are you? Is there anything we can do to vet it out and make sure it would not be a problem?

trevan:
@DanVanAtta , I'm not that concerned.  One thought I had was to add a method to `CompositeChange` that returns a flattened `CompositeChange`.

trevan:
Just created #8403 with the proposed `flatten` method.

PR #8403, opened 2020-12-11
Add a method to CompositeChange to return a flattened version

trevan:

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/CompositeChange.java
+++ b/game-core/src/main/java/games/strategy/engine/data/CompositeChange.java
@@ -1,7 +1,9 @@
 package games.strategy.engine.data;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
+import java.util.stream.Collectors;
 
 /** A Change made of several changes. */
 public class CompositeChange extends Change {
@@ -20,6 +22,21 @@ public class CompositeChange extends Change {
 
   public CompositeChange(final List<Change> changes) {
     this.changes = new ArrayList<>(changes);
+  }
+
+  public CompositeChange flatten() {
+    return new CompositeChange(
+        changes.stream()
+            .map(
+                change -> {
+                  if (change instanceof CompositeChange) {
+                    return ((CompositeChange) change).flatten().getChanges();
+                  } else {
+                    return List.of(change);
+                  }
+                })
+            .flatMap(Collection::stream)
+            .collect(Collectors.toList()));
   }
 
   public void add(final Change... changes) {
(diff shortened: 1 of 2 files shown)


game-core/src/test/java/games/strategy/engine/data/CompositeChangeTest.java
```diff
@@ -0,0 +1,41 @@
...
+import static org.mockito.Mockito.when;
+
+import games.strategy.engine.data.changefactory.ChangeFactory;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+class CompositeChangeTest {
+
+  @Test
+  void flattenChanges() {
```
DanVanAtta, line 15: Couple more test cases to consider having:
- trivial case where there is a single composite change and the input is the same as output
- deeply nested (more than 2) singleton change

game-core/src/main/java/games/strategy/engine/data/CompositeChange.java
```diff
@@ -22,6 +24,21 @@ public CompositeChange(final List<Change> changes) {
     this.changes = new ArrayList<>(changes);
   }
 
+  public CompositeChange flatten() {
+    return new CompositeChange(
+        changes.stream()
+            .map(
+                change -> {
+                  if (change instanceof CompositeChange) {
```
DanVanAtta, line 32: nit: A ternary would be worth consider having here, would make the lambda syntax a bit nicer.

game-core/src/main/java/games/strategy/engine/data/CompositeChange.java
```diff
@@ -22,6 +24,21 @@ public CompositeChange(final List<Change> changes) {
     this.changes = new ArrayList<>(changes);
   }
 
+  public CompositeChange flatten() {
```
DanVanAtta, line 27: Could use a javadoc IMO, particularly considering the method is recursive.

trevan:
@DanVanAtta , I've added some documentation and added another test case.

game-core/src/test/java/games/strategy/engine/data/CompositeChangeTest.java
trevan: The existing test case already is testing deeply nested changes.  It goes at least 3 deep.

DanVanAtta: Indeed, I'm drawing a slight distinction:
- deeply nested with N+1 changes
- deeply nested with just 1 change

In the latter case one would expect that to be flattened into a single change.

trevan: @DanVanAtta , I see what you mean.  I've added a test for the later case.

DanVanAtta:
:+1:

Issue #7090, opened 2020-07-10
Migrate to SLF4J?


DanVanAtta:
Suggestion, migrate from java.util.Log to SLF4J

**Benefits:**

### 1. Better API

Instead of:
```
log.log(Level.WARNING, "message", e);
```

Can have:
```
log.warning("message", e);
```

### 2. Cleaner log levels

There is no 'fine', 'finer', 'finest', that is just dispensed with

### 3. Built in String.format

SLF4J has built-in wildcard replacements.

Instead of:

```
log.info(String.format("message %s", variable));
``` 

Can do:

```
log.info("message {}", variable);
```

### 4. Unified logging (consistency)

HttpServer is using SLF4J, the rest of the code is not. Decoupling dropwizard from SLF4J would be painful. If we migrate it all to SLF4J then we'd have just the one logging framework.

----

I'm proposing this now as it's a good time to do so. If we migrate all of the logging via bulk changes, we can also see about converting messages from error to warning to remove the error reporting for user-induced errors. This is a migration I've wanted to do for quite some time.

asvitkine:
SGTM

RoiEXLab:
I don't have any strong preferences, but I'd agree that the SLF4J API is nicer

bacrossland:
I support your move to SLF4J. I noticed that you are using Lombok which will make the conversion to SLF4J easier since it has an annotation that creates the log object for you.

PR #8404, opened 2020-12-11
Replace usages of java.util.log with Slf4j

DanVanAtta:


trevan:
So slf4j is the preferred logging method now?

DanVanAtta:
@trevan slow transition into it, now that the major notification problems were fixed - yes. It is preferred notably because the API is on the whole nicer. For background: https://github.com/triplea-game/triplea/issues/7090

Though the console window needs some work before we can release following this update.

DanVanAtta:
Merging sooner, we need to come up with something to fix the game console. We can build an appender that writes to it, though it seems just as well maybe to write to a file. The latter suggestion is up for debate: https://forums.triplea-game.org/topic/2497/replace-console-logging-with-a-file

PR #8415, opened 2020-12-12
Add commons-text as a dependency

trevan:
This brings in StringSubstitutor which allows creating a string template
with replacement strings in it of the format ${variable}.

If there is an existing string substitution library in the project, then I can use that.  But I wasn't able to find a one.
If there is another preferred library for string substitution, I can import that instead.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,6 +74,7 @@ subprojects {
         commonsCodecVersion = '1.15'
         commonsIoVersion = '2.8.0'
         commonsMathVersion = '3.6.1'
+        commonsTextVersion = '1.9'
         databaseRiderVersion = '1.7.3'
         dropwizardVersion = '2.0.16'
         dropwizardWebsocketsVersion = '1.3.14'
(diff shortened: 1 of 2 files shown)


DanVanAtta:
Any example intended usages? I'm interested how StringSubstitutor compares to String.format

trevan:
```
String template = "${units} lost in ${territory}";
final String text =
        new StringSubstitutor(
                Map.of(
                    "units", MyFormatter.unitsToText(killedUnits),
                    "territory", location.getName()))
            .replace(template);
```

It allows named substitution.  So you could create multiple templates that have the variables in different orders or skip some variables.

DanVanAtta:
For comparison, `String.format`:
```
String.format(
  "%s lost in %s",
  MyFormatter.unitsToText(killedUnits),
  location.getName()
);
```
Are there any particular cases you have in mind @trevan  where StringSubstitutor would be particularly useful?

trevan:
@DanVanAtta , yes I'm creating a class that you can pass in different templates.  Some templates will have all of the variables (`${units} lost in ${territory}`), others will have less (`${units} killed by AA fire`), and some will be in different order (`${territory} had ${units} removed from it`).

DanVanAtta:
Are you planning to submit multiple branches in parallel that rely on this as a common update?

trevan:
Currently, it is just one branch.  Do you want me to close this and just add this dependency along with the branch?

DanVanAtta:
Please do, it's a small enough change that seeing it in context would be helpful IMO.  OTOH if it were a common ancestor then submitting it first on its own would make a lot of sense.

trevan:
No problem

PR #8416, opened 2020-12-12
Update triplea_maps.yaml

ZarophChello:


DanVanAtta:
Looks good @ZarophChello . I added some line breaks for nicer formatting.

DanVanAtta:
(note, the content is rendered as HTML, where the line breaks are not significant, so the line breaks are purely for readability of the file.)

PR #8419, opened 2020-12-12
Deprecate AbstractBattle#dependentUnits

trevan:
AbstractBattle#dependentUnits appears to have been created back before
the transportBy property came into being. It was used to keep track of
transported units in the battle territory. It handled allied air craft
that couldn't participate, unload units in amphibious assaults, and
other transported units.

This adds a Unit#getTransporting(Territory) method that allows finding
transported units in the specified territory. Now, instead of tracking
the dependents per battle, the battle can instead ask if there are
transported units in the battle site.

## Testing

* Global40 with allied aircraft. Verified both sinking of the carrier and retreating was handled.  The aircraft were destroyed in the former and retreated in the later.
* Global40 with lone transport with transported units into a scrambled zone.  The transport unloaded its cargo before the battle. Verified that the transport was destroyed and that the unloaded units were killed.
* Iron War with paratroopers.  Verified that if the transporting plane was shot down during the AA round, the paratroopers died with it.

Are there other transporting situations that I should test for?

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/Unit.java
+++ b/game-core/src/main/java/games/strategy/engine/data/Unit.java
@@ -384,7 +384,8 @@ public class Unit extends GameDataComponent implements DynamicallyModifiable {
    * This is a very slow method because it checks all territories on the map. Try not to use this
    * method if possible.
    *
-   * @return Unmodifiable collection of units that the given transport is transporting.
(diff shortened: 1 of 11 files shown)


DanVanAtta (approved):

DanVanAtta:
Looks good, can't think of any additional test cases off the bat.

PR #8438, opened 2020-12-15
Create HistoryChange objects for killing and transforming units

trevan:
The HistoryChange objects will group the calls to add a history child
event and calls to add change objects. This will creating these objects
without the need for the bridge and then returning them to callers where
the bridge is available.

KillUnit object has been used in many of the places in the battle code
that removes units. In most of those places, transformation wasn't being
done, so this fixes that.

KillUnit also makes it explicit that transported and unloaded units are
killed when it is killed. There is still some code (such as
AbstractBattle#unitsLostInPrecedingBattle) which does some of this work,
though.

Rework how units are killed and transformed in MustFightBattle. The
original logic went like:
1. Create a change object to kill a unit
2. Detect if the unit needs to transform when it dies
3. Create a change object to also set its damage to the maximum
 - At this point, there is a change object to kill it and one to set its
 damage
4. Add the units to a list of killed units during the round
5. Execute the change objects
6. Later, during the clearCasualty steps, go through the units hit as
 well as the units that were previously killed in step 4
7. Transform them

If the change object in 3 doesn't happen, then the unit won't transform
because its damage won't match the settings.
If the unit isn't added to the list in 4, then the unit won't transform
because it won't be seen in step 6.

The new logic now has the code to kill units moved to the KillUnits
HistoryChange object. This object will call TransformUnits if needed.
So the new logic looks like this:
1. Create KillUnits object
2. This creates change objects to kill the units
3. This creates a TransformUnits object that creates change objects to
 transform the units.
4. Later, during the clearCasualty step, only the units that have been
hit get passed into a TransformUnits object. The killed units from 1-3
no longer need to be worried about.

Because of save compatibility reasons, the old killedDuringCurrentRound
variable is still around and is still being used in step 4. But it is no
longer being populated so it will only actually be used if the save is
in the middle of a battle and has already populated it.

## Testing

Played TWW and had a battle between battleships.  Tested both having the battleships dying and surviving (either as a retreat or as a win).  I did notice that I could move the damaged battleship after retreating.  It looks like that property (I think it is `wasInCombat` and `alreadyMoved`) isn't being transferred to the new units.  That issue already existed so I didn't include a fix for it.

Played Global40 with transported allied air and infantry.  I also did an assault against an undefended territory and had the sea battle loose (so the infantry should die with the transport), an assault against a scrambled plane (with just a transport and infantry as well as a battleship, transport, and infantry), and an assault that had a sea battle before.

Performed a strategic bombing with an air battle that preceded it in Global40.  Had units on both sides die.

I also tried to get a `FinishedBattle` as a dependent battle but I couldn't figure out a way to do that.  I could trigger a `FinishedBattle` through blitzes and navy movement through kamikaze or convoy spaces.  But I couldn't come up with a way to make it be dependent on another battle.

I thought about creating a smaller interface of `IDelegateBridge` that only has `addChange` and a default method for writing to `IDelegateHistoryWriter`.  This interface would be passed into the `HistoryChange` objects instead of the large `IDelegateBridge` interface.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/history/change/HistoryChange.java
+++ b/game-core/src/main/java/games/strategy/engine/history/change/HistoryChange.java
@@ -1,0 +1,8 @@
+package games.strategy.engine.history.change;
+
+import games.strategy.engine.delegate.IDelegateBridge;
+
+public interface HistoryChange {
+
+  void perform(IDelegateBridge bridge);
+}
--- a/game-core/src/main/java/games/strategy/engine/history/change/HistoryChange.java
+++ b/game-core/src/main/java/games/strategy/engine/history/change/HistoryChange.java
@@ -1,8 +1,11 @@
 package games.strategy.engine.history.change;
 
+import games.strategy.engine.data.Change;
 import games.strategy.engine.delegate.IDelegateBridge;
 
 public interface HistoryChange {
 
   void perform(IDelegateBridge bridge);
+
+  Change invert();
 }
--- a/game-core/src/main/java/games/strategy/engine/history/change/HistoryChangeFactory.java
+++ b/game-core/src/main/java/games/strategy/engine/history/change/HistoryChangeFactory.java
@@ -1,0 +1,28 @@
+package games.strategy.engine.history.change;
+
+import games.strategy.engine.data.Territory;
+import games.strategy.engine.data.Unit;
+import games.strategy.engine.history.change.units.RemoveUnits;
+import games.strategy.engine.history.change.units.TransformDamagedUnits;
+import java.util.Collection;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class HistoryChangeFactory {
+
+  public TransformDamagedUnits transformDamagedUnits(
+      final Territory location, final Collection<Unit> damagedUnits) {
+    return new TransformDamagedUnits(location, damagedUnits);
+  }
+
+  public RemoveUnits removeUnitsFromTerritory(
+      final Territory location, final Collection<Unit> killedUnits) {
+    return new RemoveUnits(location, killedUnits, "${units} lost in ${territory}");
+  }
+
+  public RemoveUnits removeUnitsWithAa(
+      final Territory location, final Collection<Unit> killedUnits, final String aaType) {
+    return new RemoveUnits(
+        location, killedUnits, "${units} killed by " + aaType + " in ${territory}");
+  }
+}
--- a/game-core/src/main/java/games/strategy/engine/history/change/HistoryChangeFactory.java
+++ b/game-core/src/main/java/games/strategy/engine/history/change/HistoryChangeFactory.java
@@ -2,27 +2,27 @@ package games.strategy.engine.history.change;
 
 import games.strategy.engine.data.Territory;
 import games.strategy.engine.data.Unit;
-import games.strategy.engine.history.change.units.RemoveUnits;
-import games.strategy.engine.history.change.units.TransformDamagedUnits;
+import games.strategy.engine.history.change.units.RemoveUnitsHistoryChange;
+import games.strategy.engine.history.change.units.TransformDamagedUnitsHistoryChange;
 import java.util.Collection;
 import lombok.experimental.UtilityClass;
 
 @UtilityClass
 public class HistoryChangeFactory {
 
-  public TransformDamagedUnits transformDamagedUnits(
+  public TransformDamagedUnitsHistoryChange transformDamagedUnits(
       final Territory location, final Collection<Unit> damagedUnits) {
-    return new TransformDamagedUnits(location, damagedUnits);
+    return new TransformDamagedUnitsHistoryChange(location, damagedUnits);
   }
 
-  public RemoveUnits removeUnitsFromTerritory(
+  public RemoveUnitsHistoryChange removeUnitsFromTerritory(
       final Territory location, final Collection<Unit> killedUnits) {
-    return new RemoveUnits(location, killedUnits, "${units} lost in ${territory}");
+    return new RemoveUnitsHistoryChange(location, killedUnits, "${units} lost in ${territory}");
   }
 
-  public RemoveUnits removeUnitsWithAa(
+  public RemoveUnitsHistoryChange removeUnitsWithAa(
       final Territory location, final Collection<Unit> killedUnits, final String aaType) {
-    return new RemoveUnits(
+    return new RemoveUnitsHistoryChange(
         location, killedUnits, "${units} killed by " + aaType + " in ${territory}");
   }
 }
--- a/game-core/src/main/java/games/strategy/engine/history/change/units/KillUnits.java
+++ b/game-core/src/main/java/games/strategy/engine/history/change/units/KillUnits.java
@@ -55,7 +55,7 @@ public class KillUnits implements HistoryChange {
           unit.setHits(unit.getUnitAttachment().getHitPoints());
         });
 
-    transformUnits = new TransformUnits(location, killedUnits, true);
+    transformUnits = new TransformUnits(location, killedUnits);
 
     killedUnits.forEach(
         unit -> {
--- a/game-core/src/main/java/games/strategy/engine/history/change/units/KillUnits.java
+++ b/game-core/src/main/java/games/strategy/engine/history/change/units/KillUnits.java
@@ -1,0 +1,125 @@
+package games.strategy.engine.history.change.units;
+
+import games.strategy.engine.data.CompositeChange;
+import games.strategy.engine.data.Territory;
+import games.strategy.engine.data.Unit;
+import games.strategy.engine.data.changefactory.ChangeFactory;
+import games.strategy.engine.delegate.IDelegateBridge;
+import games.strategy.engine.history.change.HistoryChange;
+import games.strategy.triplea.formatter.MyFormatter;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Collectors;
+import lombok.Value;
+import org.apache.commons.text.StringSubstitutor;
+import org.triplea.java.collections.IntegerMap;
+
+/**
+ * Kills a set of units in a location
+ *
+ * <p>Adds a history event for the killing
+ *
+ * <p>Transforms units to other units if needed. See {@link TransformUnits}
+ */
+@Value
+public class KillUnits implements HistoryChange {
+
+  Territory location;
+  Collection<Unit> killedUnits;
+  Map<Territory, Collection<Unit>> unloadedUnits = new HashMap<>();
+  TransformUnits transformUnits;
+  String messageTemplate;
+  /** Units that were killed */
+  Collection<Unit> oldUnits = new ArrayList<>();
+  /** The units that were created after a transformation */
+  Collection<Unit> newUnits = new ArrayList<>();
+
+  public KillUnits(final Territory location, final Collection<Unit> killedUnits) {
+    this(location, killedUnits, "${units} lost in ${territory}");
+  }
+
+  /** @param messageTemplate ${units} and ${territory} will be replaced in this template */
+  public KillUnits(
+      final Territory location, final Collection<Unit> killedUnits, final String messageTemplate) {
+    this.location = location;
+    this.messageTemplate = messageTemplate;
+
+    // temporarily give the unit maximum damage so that TransformUnits will be able to
+    // recognize units that need to be transformed when they are killed
+    final IntegerMap<Unit> originalHits = new IntegerMap<>();
+    killedUnits.forEach(
+        unit -> {
+          originalHits.add(unit, unit.getHits());
+          unit.setHits(unit.getUnitAttachment().getHitPoints());
+        });
+
+    transformUnits = new TransformUnits(location, killedUnits, true);
+
+    killedUnits.forEach(
+        unit -> {
+          unit.setHits(originalHits.getInt(unit));
+        });
+
+    oldUnits.addAll(killedUnits);
+    // ensure that any units that are being transported are also killed
+    killedUnits.stream()
+        .map(unit -> unit.getTransporting(location))
+        .flatMap(Collection::stream)
+        .forEach(oldUnits::add);
+    // any unit that was unloaded during combat phase needs to be removed but it needs to be removed
+    // from the territory it unloaded to
+    killedUnits.stream()
+        .map(Unit::getUnloaded)
+        .flatMap(Collection::stream)
+        .forEach(
+            unloadedUnit -> {
+              unloadedUnits
+                  .computeIfAbsent(unloadedUnit.getUnloadedTo(), k -> new ArrayList<>())
+                  .add(unloadedUnit);
+              oldUnits.add(unloadedUnit);
+            });
+
+    newUnits.addAll(transformUnits.getNewUnits());
+
+    this.killedUnits = new ArrayList<>(oldUnits);
+    // remove the units that were transformed as TransformUnits will handle them
+    this.killedUnits.removeAll(transformUnits.getOldUnits());
+    // remove the unloaded units as they will be handled separately
+    this.killedUnits.removeAll(
+        unloadedUnits.values().stream().flatMap(Collection::stream).collect(Collectors.toList()));
+  }
+
+  @Override
+  public void perform(final IDelegateBridge bridge) {
+    transformUnits.perform(bridge);
+
+    final Collection<Unit> allKilledUnits = new ArrayList<>();
+
+    final CompositeChange change = new CompositeChange();
+    if (!killedUnits.isEmpty()) {
+      allKilledUnits.addAll(killedUnits);
+      change.add(ChangeFactory.removeUnits(location, killedUnits));
+    }
+    if (!unloadedUnits.isEmpty()) {
+      unloadedUnits.forEach(
+          (territory, units) -> {
+            allKilledUnits.addAll(units);
+            change.add(ChangeFactory.removeUnits(territory, units));
+          });
+    }
+    if (change.isEmpty()) {
+      return;
+    }
+    bridge.addChange(change);
+
+    final String text =
+        new StringSubstitutor(
+                Map.of(
+                    "units", MyFormatter.unitsToText(allKilledUnits),
+                    "territory", location.getName()))
+            .replace(messageTemplate);
+    bridge.getHistoryWriter().addChildToEvent(text, allKilledUnits);
+  }
+}
--- a/game-core/src/main/java/games/strategy/engine/history/change/units/RemoveUnits.java
+++ b/game-core/src/main/java/games/strategy/engine/history/change/units/RemoveUnits.java
@@ -1,5 +1,6 @@
 package games.strategy.engine.history.change.units;
 
+import games.strategy.engine.data.Change;
 import games.strategy.engine.data.CompositeChange;
 import games.strategy.engine.data.Territory;
 import games.strategy.engine.data.Unit;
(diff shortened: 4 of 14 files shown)


trevan:
The Codacy Static Code Analysis is complaining about an import that is only used inside of a javadoc.  Spotless doesn't remove the import and IntelliJ doesn't see it as unused.  Should I remove the import and change the javadoc to not use a `@link` or leave the import in?

game-core/src/main/java/games/strategy/engine/history/change/units/KillUnits.java
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Kills a set of units in a location
+ *
+ * <p>Adds a history event for the killing
+ *
+ * <p>Transforms units to other units if needed. See {@link TransformUnits}
+ */
+@Value
+public class KillUnits implements HistoryChange {
```
DanVanAtta, line 27: Any thoughts if 'kill' might be a confused with battle casualties? In battle, a unit when 'killed' is sent to the casualty bucket before being removed.

Suggestion: Perhaps "RemoveUnits" might be a more flexible and clear name.

game-core/src/main/java/games/strategy/engine/history/change/units/KillUnits.java
```diff
@@ -0,0 +1,125 @@
...
+import org.triplea.java.collections.IntegerMap;
+
+/**
+ * Kills a set of units in a location
+ *
+ * <p>Adds a history event for the killing
+ *
+ * <p>Transforms units to other units if needed. See {@link TransformUnits}
+ */
+@Value
```
DanVanAtta, line 26: Is the 'allArgsConstructor' of '@Value' being used?

As-is, it looks like there are a lot of constructors for this object, enough so it might be hard to reason about how it is instantiated, which makes reasoning about the object state difficult and depend on tracing down call flows.

If we do not use the 'all args constructor', removing that can help the above problem and just replace this with 'getter, setter, and equalsAndHashcode'.

game-core/src/main/java/games/strategy/engine/history/change/units/KillUnits.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Collectors;
+import lombok.Value;
+import org.apache.commons.text.StringSubstitutor;
+import org.triplea.java.collections.IntegerMap;
+
+/**
+ * Kills a set of units in a location
```
DanVanAtta, line 20: Nit, comment can be condensed:
```
Adds a history event for the removing units in a territory
```

game-core/src/main/java/games/strategy/engine/history/change/units/KillUnits.java
```diff
@@ -0,0 +1,125 @@
...
+
+    transformUnits = new TransformUnits(location, killedUnits);
+
+    killedUnits.forEach(
+        unit -> {
+          unit.setHits(originalHits.getInt(unit));
+        });
+
+    oldUnits.addAll(killedUnits);
+    // ensure that any units that are being transported are also killed
```
DanVanAtta, line 66: Does this assume that transported units are always killed? Could we instead make this logic the responsibility of the caller?

Knowing that transported units are killed is a game-rule detail. This history object therefor not only tracks the history transformation, but it also encodes the rule that transported units are killed. If the caller were to handle this detail, and pass the list of transported units into the killed list, then this class would no longer be aware of that rule detail.

game-core/src/main/java/games/strategy/engine/history/change/units/KillUnits.java
```diff
@@ -0,0 +1,125 @@
...
+            unloadedUnit -> {
+              unloadedUnits
+                  .computeIfAbsent(unloadedUnit.getUnloadedTo(), k -> new ArrayList<>())
+                  .add(unloadedUnit);
+              oldUnits.add(unloadedUnit);
+            });
+
+    newUnits.addAll(transformUnits.getNewUnits());
+
+    this.killedUnits = new ArrayList<>(oldUnits);
```
DanVanAtta, line 86: nit, the collection copy and then 'removalAll' can be be avoided, EG:
```
this.killedUnits = oldUnits.stream().filter(oldUnit -> transformedUnits.getOldUnits().contains(oldUnit)).collect(...);
```

game-core/src/main/java/games/strategy/engine/history/change/units/TransformUnits.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.stream.Collectors;
+import lombok.Value;
+import org.triplea.util.Tuple;
+
+/**
+ * Transforms units into other unit types as determined by {@link
+ * UnitAttachment#getWhenHitPointsDamagedChangesInto()}
+ */
+@Value
+public class TransformUnits implements HistoryChange {
```
DanVanAtta, line 27: `TransformUnits` is possibly too generic of a name. Perhaps we can clarify this by getting the 'Damaged' word in there.

Suggestions:
- `DamagedUnitTransformation`
- `ChangeDamagedUnitType`

game-core/src/main/java/games/strategy/engine/history/change/units/TransformUnits.java
```diff
@@ -0,0 +1,100 @@
...
+  Territory location;
+  /** Map of old unit -> new unit */
+  Map<Unit, Unit> transformingUnits = new HashMap<>();
+
+  CompositeChange attributeChanges = new CompositeChange();
+
+  public TransformUnits(final Territory location, final Collection<Unit> transformUnits) {
+    this.location = location;
+
+    for (final Unit unit : transformUnits) {
```
DanVanAtta, line 38: nit, this could use a quick leading comment to summarize what we are doing in this block, help make it easier to understand.

That could potentially be done by assigning the variable `transformingUnits` in this constructor, in which case the helper method would carry that information, eg:
```
transformingUnits = computeUnitTransformations(transformUnits);
```

game-core/src/main/java/games/strategy/engine/history/change/units/TransformUnits.java
```diff
@@ -0,0 +1,100 @@
...
+@Value
+public class TransformUnits implements HistoryChange {
+
+  Territory location;
+  /** Map of old unit -> new unit */
+  Map<Unit, Unit> transformingUnits = new HashMap<>();
+
+  CompositeChange attributeChanges = new CompositeChange();
+
+  public TransformUnits(final Territory location, final Collection<Unit> transformUnits) {
```
DanVanAtta, line 35: It seems that `transformUnits` is conditional and not all the units passed in will be transformed. Named `transformUnits`,it sounds to me as if all the units would be transforming or already have transformed.

Is this the list of damaged units? In that case, for consideration, `damagedUnits` could be a more descriptive name.

game-core/src/main/java/games/strategy/engine/history/change/units/TransformUnits.java
```diff
@@ -0,0 +1,100 @@
...
+
+    bridge.addChange(
+        new CompositeChange(
+            ChangeFactory.addUnits(location, getNewUnits()),
+            ChangeFactory.removeUnits(location, getOldUnits()),
+            attributeChanges));
+
+    // to reduce the amount of history text, group the transforming units by both the original and
+    // new unit type
+    transformingUnits.entrySet().stream()
```
DanVanAtta, line 68: There is a lot going on in this block. Would you mind taking a second look to see if you can make this easier to understand? Let me know if you'd like suggestions or thoughts on how to achieve that, I'll be happy to help.

game-core/src/main/java/games/strategy/triplea/delegate/battle/AirBattle.java
```diff
@@ -910,11 +911,9 @@ private void remove(
       return;
     }
     final Collection<Unit> killed = getUnitsWithDependents(killedUnits);
-    final Change killedChange = ChangeFactory.removeUnits(battleSite, killed);
-    final String transcriptText =
-        MyFormatter.unitsToText(killed) + " lost in " + battleSite.getName();
-    bridge.getHistoryWriter().addChildToEvent(transcriptText, new ArrayList<>(killed));
-    bridge.addChange(killedChange);
+    final KillUnits killUnits = new KillUnits(battleSite, killed);
```
DanVanAtta, line 914: nit, unnecessary intermediate variable. Can be a single line
```
new KillUnits(battleSite, killed).perform(bridge);
```

Side-note: I think the above does reveal a code smell where the bridge should be in control of when the change happens. EG:
```
bridge.addChange(new KillUnits(battleSite, killed);
```

game-core/src/main/java/games/strategy/triplea/delegate/battle/StrategicBombingRaidBattle.java
```diff
@@ -663,18 +658,13 @@ private void removeAaHits(
...
-              MyFormatter.unitsToTextNoOwner(killed) + " killed by " + currentTypeAa,
-              new ArrayList<>(killed));
       final IntegerMap<UnitType> costs = TuvUtils.getCostsForTuv(attacker, gameData);
       final int tuvLostAttacker = TuvUtils.getTuv(killed, attacker, costs, gameData);
       attackerLostTuv += tuvLostAttacker;
       // attackingUnits.removeAll(casualties);
       removeAttackers(killed, false);
-      final Change remove = ChangeFactory.removeUnits(battleSite, killed);
-      bridge.addChange(remove);
+      new KillUnits(battleSite, killed, "${units} killed by " + currentTypeAa + " in ${territory}")
```
DanVanAtta, line 666: The formatting of the message pattern being a responsibility of the caller is concerning to me. This creates a coupling between the data and the formatting strings used internally by 'killUnits' and the caller. The caller must be aware of the string type. It does seem like the formatted string could become an implementation detail of `KillUnits` by adding a third constructor, eg:
```new KillUnits(battleSite, killed, currentTypeAa)```

At that point too I'd recommend to add a `@Builder` on that constructor so that the meaning of the parameters is more clear when reading the calling context. 

WDYT?

File: game-core/src/main/java/games/strategy/engine/history/change/units/KillUnits.java (hunk 4 above)
Comment (trevan) [reply game-core/src/main/java/games/strategy/engine/history/change/units/KillUnits.java:66]: Yes, it assumes that transported units are always killed.  There is currently no way to prevent that from happening in the engine.

I don't want the caller to handle this detail.  Because currently, it handles it incorrectly.  Some parts of the battle code that kills units will correctly remove their transported units.  Other parts of the battle code leaves the units floating in the ocean and luckily some other part of the code much later happens to clear them up.  By hiding this detail from the caller, it makes the code easier to read.

As for allowing the map to control whether this should be killed or not, that is the work I'm doing with the translate object code where transported items can be copied from one unit to another unit.

File: game-core/src/main/java/games/strategy/triplea/delegate/battle/StrategicBombingRaidBattle.java (hunk 11 above)
Comment (trevan) [reply game-core/src/main/java/games/strategy/triplea/delegate/battle/StrategicBombingRaidBattle.java:666]: `KillUnits` has a constructor that is already `Territory, Collection<Unit>, String` so I can't add this third constructor.  I could remove the existing one but there might be another caller that will have a different string to be used and then I'll be back to the same problem.

I like the idea of hiding the string as an implementation detail.  I guess I could create a child class of `KillUnits` that is `KillAaCasualtyUnits` (or something like that) and then it can have this special constructor.  Are you ok with that?

Comment (DanVanAtta) [reply game-core/src/main/java/games/strategy/triplea/delegate/battle/StrategicBombingRaidBattle.java:666]: Multiple constructors with the same signature usually calls for a factory pattern. In that case just create static methods that create the new object and they can be different in their naming. Would that work?

File: game-core/src/main/java/games/strategy/triplea/delegate/battle/AirBattle.java (hunk 10 above)
Comment (trevan) [reply game-core/src/main/java/games/strategy/triplea/delegate/battle/AirBattle.java]: The plan is to return these history change objects to the caller so that they can then pass it to the bridge.  So, there is still a bit of code smell because this is a partial step.

**Comment by trevan** [comment]
@DanVanAtta , I've renamed things and rework some of the code to hopefully be easier to read.

File: game-core/src/main/java/games/strategy/engine/history/change/HistoryChange.java [hunk 12]
```diff
@@ -0,0 +1,8 @@
+package games.strategy.engine.history.change;
+
+import games.strategy.engine.delegate.IDelegateBridge;
+
+public interface HistoryChange {
```
DanVanAtta, line 5: A quick javadoc could perhaps be useful to explain what constitutes a history change.

I wonder though, wouldn't almost any change be eligible for the history? Is there a (ideally) a 1:1 correspondence between history change objects and game state changes? Are we going to have "peer" change objects in this manner?

DanVanAtta (commented):
One new comment. I still need to look through 'removeUnits' in detail, at a glance it and the rest of the PR look good.

game-core/src/main/java/games/strategy/engine/history/change/HistoryChange.java
trevan: When you say "game state changes", I'm assuming you are talking about the `Change` objects.

There will not be a 1:1 correspondence between history change objects and game state changes.  A history change object can comprise of 1 or more game state changes.  Take for example `TransformDamagedUnits`.  It creates at minimum an `AddUnits` and a `RemoveUnits` change object.  It also can create many more depending on if attribute translation is asked for and depending on which attributes need to be translated.

I see the `HistoryChange` as more of a higher level object than the `Change` object.  It maps to what the user sees and to what is listed in the history.  The `Change` object is the low level changes that must be done to keep the game state in sync.

One possible benefit of the `HistoryChange` is that if it was stored in the game data history, then it would actually be possible to undo each history entry because the changes are truly grouped together per history entry.

I do plan to add more `HistoryChange` objects.

DanVanAtta: Since they are related, I wonder if instead there is a way to map the history changes from game state changes. Otherwise it seems we are duplicating information about changes to the game state via another change object mechanism.

trevan: I wouldn't say this is "duplicating information".  I'd say this is grouping information and actually mapping history changes with their underlying game state changes.  Because a history change (and I'm not talking about the history change objects in this PR) might incorporate multiple game state changes.

DanVanAtta: A relationship of many:1 for game state to history changes is interesting. I wonder if that relationship could also be viewed as [0|1] : 1

The reason I am focusing on this is to see if we can simplify how the history is rendered and displayed relative to game state. If we have both a stack of history objects, and a stack of game state objects, how do we easily "point" the game state to an arbitrary location in that game state stack and easily then recreate the history? Or from another point of view, what if the current game is "pointed" to an arbitrary location in history, how do we easily then unwind the game state to that location?

If there is a link (likely bidirectional)  between history objects and state changes, this I think becomes much easier. If such a link is bidirectional, is it potentially actually part of the same data entity? Would it make sense to even just have an API on game state changes like `Optional<HistoryChange> getHistoryElement`

trevan: As I mentioned, if `HistoryChange` was stored in the history, then it could keep track of all of the `Change`s that occur during that history.  And then if you revert a `HistoryChange` (which this PR doesn't include), it would revert all of the `Change` objects that were part of the history event.

The current mechanism that tracks history items with change objects is done in `games.strategy.engine.history.History`.  I believe how that works is that at some point in the game flow, a call to `HistoryWriter.startEvent` will be called.  This adds a new node to `History`.  When `bridge.addChanges` is called, the changes are stored in a big list inside of `History`.  The nodes created by `startEvent` store the index in the changes list when they are added.  When you traverse the history, it looks up the index of the history node, computes a delta of changes from the big list of changes and then undoes those changes.

That logic feels really convoluted.  I'm not entirely sure I understand it completely.  These `HistoryChanges`, though, could simplify that.  Instead of storing indices into a list of changes and assigning them to each node, the history node would actually store the list of changes that it performed.  And each change object would be attached to a history node, where as now, it looks like you can have change objects that aren't attached to history nodes.

trevan:
I just pushed a commit where I added a "invert" method that can undo the changes performed by the HistoryChange.

game-core/src/main/java/games/strategy/engine/history/change/HistoryChangeFactory.java
```diff
@@ -0,0 +1,28 @@
...
+
+import games.strategy.engine.data.Territory;
+import games.strategy.engine.data.Unit;
+import games.strategy.engine.history.change.units.RemoveUnits;
+import games.strategy.engine.history.change.units.TransformDamagedUnits;
+import java.util.Collection;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class HistoryChangeFactory {
```
DanVanAtta, line 11: What other history change objects do you foresee? How many might we have total and will a single factory for them all scale well?

game-core/src/main/java/games/strategy/engine/history/change/units/TransformDamagedUnits.java
```diff
@@ -0,0 +1,127 @@
...
+  CompositeChange attributeChanges = new CompositeChange();
+
+  public TransformDamagedUnits(final Territory location, final Collection<Unit> damagedUnits) {
+    this.location = location;
+
+    // check if each of the damaged units are supposed to change when they take damage
+    // if it is supposed to change, create the new unit and translate attributes from the old unit
+    // to the new unit
+    for (final Unit unit : damagedUnits) {
+      final Map<Integer, Tuple<Boolean, UnitType>> map =
```
DanVanAtta, line 49: side-note, we're trying to kill `Tuple` and replace it with value objects. For example `TotalPowerAndRolls` used to be a `Tuple<Integer,Integer>`,that conversion certainly lead to more legible code. In this case, the `Boolean` is very cryptic, it's kinda hard to know what the represents at first glance.

game-core/src/main/java/games/strategy/engine/history/change/units/RemoveUnits.java
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * Removes a set of units in a location and adds a history event
+ *
+ * <p>Transforms units to other units if needed. See {@link TransformDamagedUnits}
+ */
+@FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
+@Getter
+@EqualsAndHashCode
+public class RemoveUnits implements HistoryChange {
```
DanVanAtta, line 32: I do wonder if these objects should be suffixed with `HistoryChange` so that they are not quite so context dependent. WDYT @trevan ?

DanVanAtta (approved):

game-core/src/main/java/games/strategy/engine/history/change/HistoryChangeFactory.java
trevan: I patterned this off of the existing `ChangeFactory`.  I could imagine one history change class per type of line in the history.

game-core/src/main/java/games/strategy/engine/history/change/units/RemoveUnits.java
trevan: Sounds good.  I've renamed them @DanVanAtta

game-core/src/main/java/games/strategy/engine/history/change/HistoryChangeFactory.java
DanVanAtta: Okay. If it grows to be overwhelming we can probably deal with it.

PR #8440, opened 2020-12-16
Add new AI type

trevan:
I'm horrible with names so I just called this "Field (AI)" and the package name is "flowfield".  I'm open to better names.  This AI will only show up if the beta flag is enabled.  And it won't do anything.  It will do even less than the do-nothing ai.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerType.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerType.java
@@ -3,9 +3,11 @@ package games.strategy.engine.framework.startup.ui;
 import games.strategy.engine.player.Player;
 import games.strategy.triplea.TripleAPlayer;
 import games.strategy.triplea.ai.fast.FastAi;
+import games.strategy.triplea.ai.flowfield.FlowFieldAI;
 import games.strategy.triplea.ai.pro.ProAi;
 import games.strategy.triplea.ai.weak.DoesNothingAi;
 import games.strategy.triplea.ai.weak.WeakAi;
+import games.strategy.triplea.settings.ClientSetting;
 import java.util.Arrays;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
@@ -47,6 +49,13 @@ public enum PlayerType {
     @Override
     public Player newPlayerWithName(final String name) {
       return new ProAi(name);
+    }
+  },
+
+  FLOW_FIELD_AI("Field (AI)") {
+    @Override
+    public Player newPlayerWithName(final String name) {
+      return new FlowFieldAI(name);
     }
   },
 
@@ -97,6 +106,9 @@ public enum PlayerType {
   public static String[] playerTypes() {
     return Arrays.stream(values())
         .filter(PlayerType::isVisible)
+        .filter(
+            value ->
+                ClientSetting.showBetaFeatures.getValue().orElse(false) || value != FLOW_FIELD_AI)
         .map(enumValue -> enumValue.label)
         .toArray(String[]::new);
   }
--- a/game-core/src/main/java/games/strategy/triplea/ai/flowfield/FlowFieldAI.java
+++ b/game-core/src/main/java/games/strategy/triplea/ai/flowfield/FlowFieldAI.java
@@ -1,0 +1,47 @@
+package games.strategy.triplea.ai.flowfield;
+
+import games.strategy.engine.data.GameData;
+import games.strategy.engine.data.GamePlayer;
+import games.strategy.engine.framework.startup.ui.PlayerType;
+import games.strategy.triplea.ai.AbstractAi;
+import games.strategy.triplea.delegate.remote.IAbstractPlaceDelegate;
+import games.strategy.triplea.delegate.remote.IMoveDelegate;
+import games.strategy.triplea.delegate.remote.IPurchaseDelegate;
+import games.strategy.triplea.delegate.remote.ITechDelegate;
+
+public class FlowFieldAI extends AbstractAi {
+  public FlowFieldAI(final String name) {
+    super(name);
+  }
+
+  @Override
+  public PlayerType getPlayerType() {
+    return null;
+  }
+
+  @Override
+  protected void purchase(
+      final boolean purchaseForBid,
+      final int pusToSpend,
+      final IPurchaseDelegate purchaseDelegate,
+      final GameData data,
+      final GamePlayer player) {}
+
+  @Override
+  protected void tech(
+      final ITechDelegate techDelegate, final GameData data, final GamePlayer player) {}
+
+  @Override
+  protected void move(
+      final boolean nonCombat,
+      final IMoveDelegate moveDel,
+      final GameData data,
+      final GamePlayer player) {}
+
+  @Override
+  protected void place(
+      final boolean placeForBid,
+      final IAbstractPlaceDelegate placeDelegate,
+      final GameData data,
+      final GamePlayer player) {}
+}
--- a/game-core/src/test/java/games/strategy/engine/framework/startup/ui/PlayerTypeTest.java
+++ b/game-core/src/test/java/games/strategy/engine/framework/startup/ui/PlayerTypeTest.java
@@ -6,15 +6,18 @@ import static org.hamcrest.core.Is.is;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import games.strategy.engine.player.Player;
+import games.strategy.triplea.settings.ClientSetting;
 import java.util.List;
 import org.hamcrest.Matchers;
 import org.hamcrest.core.IsCollectionContaining;
 import org.junit.jupiter.api.Test;
+import org.sonatype.goodies.prefs.memory.MemoryPreferences;
 
 class PlayerTypeTest {
 
   @Test
   void playerTypes() {
+    ClientSetting.setPreferences(new MemoryPreferences());
     assertThat(
         "Ensure we do not have an example invisible player type in the selection list",
         List.of(PlayerType.playerTypes()),
@@ -32,6 +35,7 @@ class PlayerTypeTest {
 
     stream(PlayerType.values())
         .filter(playerType -> playerType != PlayerType.BATTLE_CALC_DUMMY)
+        .filter(playerType -> playerType != PlayerType.FLOW_FIELD_AI)
         .forEach(
             playerType -> {
               final Player result = playerType.newPlayerWithName(testName);


Cernelius:
Well, how can anyone suggest any name if it is not said what the AI is for.

beelee1:
Is it gonna be added to in the future ? Or just stay the way it is ? If it does absolutely nothing and stays that way, you could call it does absolutely nothing. If it's gonna end up with stuff added, you could call it future.

trevan:
This is the initial class for the AI that I'm working on.  The current idea is that it will calculate a value for territories and then diffuse that value to all of the territories.  Then, a unit just needs to at the territories it can reach and pick the highest value territory.  That's the simplest explanation.  It will have to deal with enemy units, allies, objectives, etc but I believe I can model most of those as territory values and diffuse them as well.

Cernelius:
So harder than hard AI? Call it "New Hard" and, then, once is actually harder than "Hard", rename "Hard" to "Easy", "Easy" to "Very Easy" and "New Hard" to "Hard".

Cernelius:
But why not just updating Hard AI?

trevan:
> But why not just updating Hard AI?

Because there is no guarantee this will actually be better than Hard AI?  And the logic is a lot different from Hard AI so it isn't an update.  It would be a rewrite.

I don't know if you've ever looked at chess AI's but there are dozens of them.  All of them have strength and weaknesses.  It would be really nice if there was a standard API that an AI engine could plug into.  Then you could download an AI engine separate from triplea and run it.  It would also make developing custom AIs or improving AIs easier since people could do it independent of the triplea code.  But that would probably be a fairly large change.

DanVanAtta:
We could come up with a naming convention for the AIs if we'd want. 

It'll be difficult to know how to characterize the AI until it's more developed. Something like WWII allied generals is a possibility. 

The servers got project codenames after WWII planes, we have Spitfire (current lobby) and Stuka (maps server, in development). Though the codenames I think are dropped now that the servers are more baked,  'lobby' and 'maps'  are more generally used.    : )

game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerType.java
```diff
@@ -50,6 +52,13 @@ public Player newPlayerWithName(final String name) {
     }
   },
 
+  FLOW_FIELD_AI("Field (AI)") {
```
DanVanAtta, line 55: If we don't go with some sort of naming convention for AIs, WDYT of  "Flow (AI)"? Field works too : )

game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerType.java
```diff
@@ -97,6 +106,9 @@ public Player newPlayerWithName(final String name) {
   public static String[] playerTypes() {
     return Arrays.stream(values())
         .filter(PlayerType::isVisible)
+        .filter(
+            value ->
+                ClientSetting.showBetaFeatures.getValue().orElse(false) || value != FLOW_FIELD_AI)
```
DanVanAtta, line 111: This kind of condition is concise, but takes some work to grok. Can we hack the 'isVisible' to contain this logic? (and make it a bit simpler?)

game-core/src/main/java/games/strategy/triplea/ai/flowfield/FlowFieldAI.java
```diff
@@ -0,0 +1,47 @@
...
+import games.strategy.engine.data.GameData;
+import games.strategy.engine.data.GamePlayer;
+import games.strategy.engine.framework.startup.ui.PlayerType;
+import games.strategy.triplea.ai.AbstractAi;
+import games.strategy.triplea.delegate.remote.IAbstractPlaceDelegate;
+import games.strategy.triplea.delegate.remote.IMoveDelegate;
+import games.strategy.triplea.delegate.remote.IPurchaseDelegate;
+import games.strategy.triplea.delegate.remote.ITechDelegate;
+
+public class FlowFieldAI extends AbstractAi {
```
DanVanAtta, line 12: 1. It seems like some documentation would be useful to know what/where this AI is going. At least something to start. My initial thought is a README file somewhere or a javadoc to propose an outline of what this AI is. What do you think of either of those two?
 
2. I wonder if we should "start this right" and make this the first AI that lives in an AI sub-project. That would give a hard, compile-time, boundary between the AI code and game-core.

game-core/src/main/java/games/strategy/triplea/ai/flowfield/FlowFieldAI.java
```diff
@@ -0,0 +1,47 @@
...
+  public FlowFieldAI(final String name) {
+    super(name);
+  }
+
+  @Override
+  public PlayerType getPlayerType() {
+    return null;
```
DanVanAtta, line 19: side-note, an override that returns null is a code smell. Indicates an inappropriate abstraction where either there are really two interfaces or something.. Keeping one interface probably make sense, in which case favoring optional return types is preferred compared to nullable return types

game-core/src/test/java/games/strategy/engine/framework/startup/ui/PlayerTypeTest.java
```diff
@@ -6,15 +6,18 @@
...
 import org.hamcrest.Matchers;
 import org.hamcrest.core.IsCollectionContaining;
 import org.junit.jupiter.api.Test;
+import org.sonatype.goodies.prefs.memory.MemoryPreferences;
 
 class PlayerTypeTest {
 
   @Test
   void playerTypes() {
+    ClientSetting.setPreferences(new MemoryPreferences());
```
DanVanAtta, line 20: There is an abstract test that is typically used when the code under test depends on ClientSetting. Did you intentionally skip using that in favor of this setter, if so I'm curious of the benefit of doing a client setting test this way vs extending the base test class.

game-core/src/test/java/games/strategy/engine/framework/startup/ui/PlayerTypeTest.java
```diff
@@ -32,6 +35,7 @@ void newPlayerWithName() {
 
     stream(PlayerType.values())
```
DanVanAtta, line 36: side-note, this test assertion is suspect, contains logic. It might be better to have 4 or 5 relatively duplicated lines that asserts for each expected play type and hardcodes the expected values. It would be easy to understand and extend and avoids having logic in the assertion conditions.

DanVanAtta (commented):
Looks fine, major consideration is whether this should be the start of a new sub-project and where documentation ought to live.

Cernelius:
I would rather call it after the main contributor, followed by an estimate of the strength.

Hard AI->redrum Hard AI
Fast AI->redrum Fast AI
?->trevan Hard AI

---

If there would exist a main AI for TripleA, that can be called as "TripleA Hard AI".

DanVanAtta:
What happens when their are multiple authors, or an existing AI is
overhauled? I'm opposed to renaming an existing and strength of AI is too
relative of a term. Meanwhile it is helpful to set a name early and
difficult to change later.

The choice is up to trevon, field, flow, or pick some naming convention.

Cernelius:
@DanVanAtta The current naming convention quite clearly doesn't do anything like what I understand you have in mind, since the names are Hard, Easy and Fast, that are clearly judgmental. And this has been even done by renaming, since, for example, the Easy one used to be called "Fodder".

Adding something that has a non-particularly descriptive name would leave the system with an incoherent naming convention, as any regular user will see names like Hard, Easy, Fast and Does-Nothing, that are descriptive, and, beside it, another name that appears to be estranged.

If we take that, for example, "Hard" is just a name, chosen by @ron-murhammer, and he could have as well chosen to call it "Strawberry", instead, then I could make an AI that is worse than Easy and call it "Very Hard", since the name is just a name (if we apply this concept to the existing one).

---

If you want to go in the direction I'm understanding, I would suggest to rename Easy back to "Fodder" and have Hard and Fast be renamed to something else less "relative".

DanVanAtta:
There is not really a current naming convention. It was intended to have different levels of AI but that did not work out. We are more than a decade later now and have "does nothing" and "fast" that contradict that naming convention.

So yes, calling "hard" AI "Strawberry" would be just about as meaningful. Does "fast" mean the AI is going to move fast (IE: fast moving units)? It's not clear. It is not even clear what the relatively well named "does nothing AI" does.

- we are not going to rename existing AIs
- the name of the new AI so is not important: https://en.wikipedia.org/wiki/Law_of_triviality
- coming up with a "medium" or "extra hard" name is difficult and does not scale. What happens when there is another "hard" AI, do we call it "slightly less hard?" How do we know how good the AI is until it is done (which is months away, meanwhile a lot of code needs to drop and it will want to have *a* name for the AI). 
- Nothing is to say that the codename for the AI will be public facing. You may never realize that the "strawberry" AI is actually Weak AI. We need a consistent project name on the other hand as we have to call it something. It's computer code, we have to encode in binary for what the computer will do, we can't skip decisions, they *have* to be made. What to call the thing internally is needed and is not pleasant to bulk change later.
- what happens when one AI is better at some things but worse at other things? How do we name that? "Kinda-hard AI on small maps but easy on naval maps?" 

TL:DR
Bottom line, decoupling the name of the AI from the front-end display is an option and can be considered later. The name Trevan wants to use is up tto him. We will not rename existing AI names. If there is a case to be made why that is valuable, that would benefit discussion to justify why we should spend time on that and not other projects (and to justify the user confusion and the extended effort of explaining to confused users what happened).

Cernelius:
Ok, then, all considered, my suggestion to @trevan (since he/she asked) is to call this AI as "**Hard2**" (I would, then, rename "Hard" as "Hard1", but this is not necessary).

---

If someone will make another AI that is likely to be about on the same level as "Hard", then it can be called "Hard3", and so forth.

game-core/src/main/java/games/strategy/triplea/ai/flowfield/FlowFieldAI.java
trevan: For 2, I think the only problem is `PlayerType.java`.  There isn't a way to dynamically add new player types.  They are currently hardcoded and need access to the Player class.  Do you have suggestions on how to make it dynamic?  I can work on that before this is merged and then create a new sub project.

game-core/src/main/java/games/strategy/triplea/ai/flowfield/FlowFieldAI.java
trevan: I missed that I needed to set this value.  It shouldn't return null.

game-core/src/test/java/games/strategy/engine/framework/startup/ui/PlayerTypeTest.java
trevan: I was just trying to get this test to pass.  I don't know about the abstract test you are talking about nor do I know if this test will work using that abstract test.

trevan:
The name displayed in the UI is already decoupled from the name in the code.  I could call the classes "CrazyAIBecauseItIsCrazy" and have the UI show "Common AI".  And changing the name in the UI is really trivial.  The only problem is users' expectation.  By the way, "Hard AI" is actually "pro" in the package name.  So there is already an instance where the UI name and the package name is different.

I don't think this AI should have my name in it, nor do I think it should have "Hard" in it.  The idea of using code names sounds fine.  In Chess AI, you have names like "AlphaZero", "StockFish", "Houdini", "Deep Blue", "Rybka", "Leela Chess Zero", etc.

I guess I'll keep the package name to be "flowfield".  As for what the UI will say, general names could work.  Since this is behind the beta flag, we could come up with the UI name later after it has been fleshed out.

Cernelius:
How about "**pro2**", for the name in the code? To me, I would say that the code names are quite descriptive too, as I believe they are names like "pro" or "weak".

What I'm understanding is that you are making something like "Hard" AI, covering pretty much the same needs, just in a different way (that of course you hope to be better in a majority of cases, I assume).

Cernelius:
Even though the chance of something like this happening is remote, I suggest to check all code names that existed for any AI in TripleA, to avoid taking the same name. For example, if I recall correctly, there was an AI that to the users was given as "Moore..." or "Medium" and that was called "strong" in the code (I'm not sure). I guess it may cause confusion if the same name happens to be retaken (if someone would make a new AI and calle it "strong", not knowing that existed an AI called exactly as such).

As said, I realize that this is remote.

---

Also for all non-developers, I'm thinking it would be good having a document with the full list of all names of every AI that ever existed in TripleA, with the current status of them.

game-core/src/main/java/games/strategy/triplea/ai/flowfield/FlowFieldAI.java
DanVanAtta: You could use the injections class to return the list of `PlayerType`.

For context, game-headed would depend on the AI project  and game-core, and the AI project would depend on just game-core. The game-headed project could grab the list of AI from the AI project and then set them in `Injections.java` which can then be used where needed. 

It's probably not that straight forward though as AI's are specifically referenced and used throughout the engine, so it is quite tightly coupled. We'd probably have to take a look to see how we can start decoupling. 

Ultimately making the `PlayerType` probably no longer an enum is needed. It would probably become just an interface, the game-headed package would instantiate the list of available AIs and inject them into `Injections.java`. In turn we'd have to look at the other locations of specific coupling, see if we can convert those to accept a list instead. For specific references to say 'WeakAI', that is something that we've wanted to dig out for some time.

If we can get at least some partial success here, that could be worth it while we continue to work to decouple AI from game-core. Meaning, even if it's just the one new AI stub in the AI sub-project, it is a step in the right direction.

game-core/src/test/java/games/strategy/engine/framework/startup/ui/PlayerTypeTest.java
DanVanAtta: See `AbstractClientSettingTestCase`

game-core/src/main/java/games/strategy/triplea/ai/flowfield/FlowFieldAI.java
trevan: I'll check out the injections class.  I'm going to close this PR.

PR #8441, opened 2020-12-16
Add interface so that a player object can add debug menu items

trevan:
The Debug menu will have an item for each player that has this
interface. The player can then define the items to be shown in the sub
menu. Useful for AI players to be able to add custom debug actions.

## Testing

I modified AbstractProAI to implement the interface and I checked that the menu items showed up correctly and worked.

## Screens Shots

Here's the Debug menu with all of the Axis in Global40 using Hard AI.

Screen Shot 2020-12-15 at 8 46 44 PM

## Release Note

CHANGE|Allow AI players to add custom menu items to the Debug menu.


game-core/src/main/java/games/strategy/triplea/ui/menubar/DebugMenu.java
```diff
@@ -25,6 +26,16 @@
           .setMnemonic(KeyEvent.VK_X);
     }
 
+    players.stream()
+        .filter(PlayerDebug.class::isInstance)
```
DanVanAtta, line 30: I think we should prefer to avoid using class types to infer properties of an object. It can get odd in terms of nesting interfaces. Second, the menus suffer from a lot of side effect calls with methods like "addMenu(menu)", this makes code a lot harder to follow.

What would you think about adding a method to the `Player` type, something like `List<JMenuItem> getDebugMenutItems()`. Then in this method, when iterating over each player, all you'd need to filter on is that the debug menus items is no non-null, and then you could add the menu times. This would avoid the 'void' return method and instanceOf check.

trevan: I felt like add `getDebugMenuItems()` to the `Player` type is making the type include extra behavior that isn't needed for the `Player`.  That was the purpose of creating a specialized interface `PlayerDebug`.

trevan: I've modified `getDebugMenuItems()` to return a `Collection<JMenuItem>`

DanVanAtta: If we can get away from the `instanceof` check, would be awesome.
I don't think I like how we have such a narrow interface as well. Is there opportunity to consolidate the debug tools with ProAI and the new one? ProAI has an extensive debugging framework built into it, I'm concerned we could be going down a path where each AI builds its very own set of debugging tools.

DanVanAtta: Otherwise, it kinda seems like the interface should be related to AI, 'AiPlayer' and have methods specific for AIs. I think each probably would want a debug menu, or could have some use for it. There are probably other opportunities there as well. That might get us further down the path of having even more instanceof checks to see if a player is an AI.

Perhaps there could be a method like `Optional<PlayerAi> getAI()`, and that would return us the interface that has things like debug menu and other AI specific properties. If the player is human then the method would return empty.

trevan: The `ProAI` debug framework is very tied into the `ProAI` code.  I think it would be better for each AI to have its own debugging tools.  The AI logic will be different enough that you'll need to see different aspects of it.

How about after #8454 is merged, then I add this to the `AbstractAI` class.  And then I can grab the list of `AbstractAI`s from the Injector and check if the current players are one of those `AbstractAI`s.  And if they are, I can then cast it to `AbstractAI` and call `getDebugMenuItems()`.

DanVanAtta: Understood about each AI having its own debug.

It could generally be useful to have a method like 'isAi' on the player object, that would be my preference instead of an instanceof check. Though at the same time, TripleA is designed so that it is transparent whether a player is AI, human, remote or local.

Your idea does not sound bad. For consideration for another option -  What if we did a kind of registration of a debug framework (or menu)? Essentially we make available a static method to add a debug menu, and then if a given AI is initialized, it adds debugging. This would follow the 'open-closed' principle well and would not involve interface API changes nor instance-of checks.

trevan: @DanVanAtta , I've pushed an update where the DebugMenu has a callback registry for adding new debug menus.

DanVanAtta: Thanks! When will this infrastructure code be used @trevan ?

game-core/src/main/java/games/strategy/triplea/ui/menubar/DebugMenu.java
```diff
@@ -4,13 +4,21 @@
...
+import java.util.function.Function;
 import javax.swing.JMenu;
+import javax.swing.JMenuItem;
 import org.triplea.swing.SwingAction;
 
-final class DebugMenu extends JMenu {
+public final class DebugMenu extends JMenu {
   private static final long serialVersionUID = -4876915214715298132L;
 
+  private static final Map<String, Function<TripleAFrame, Collection<JMenuItem>>> menuCallbacks =
```
DanVanAtta, line 19: The type for this callback seems to be a bit overly generic. It's not clear what the semantic would be behind this type. Perhaps this can be resolved by creating an interface for this, eg: `DebugMenuFactory`

Why is the `TripleAFrame` reference required? Shouldn't any newly created debug menu be able to be added to a generic JMenu? Would it make sense for the debug menu factory to simply produce a menu and let this class deal with the detail of which menu it ought to be added to.

trevan: I don't understand your issue for the first paragraph.  You want me to create an interface that is an alias?  You mentioned in https://github.com/triplea-game/triplea/pull/7829#discussion_r499175473 about not using interfaces as aliases so I'm confused on why it is ok to use an alias here and not in that previous discussion?

I believe the `TripleAFrame` is required because to create new dialogs and other swing items, a parent frame is required.  So, if the menu item is going to open a new dialog or do something to the UI, it needs access to the `TripleAFrame`.  In one of my debug menu items, it adds an heatmap overlap and it needs to be able to get a reference to the `MapPanel` which is a member of `TripleAFrame`.

game-core/src/main/java/games/strategy/triplea/ui/menubar/DebugMenu.java
trevan: @DanVanAtta , I'm currently using it in my AI.  I also plan to move the ProAi debug menu code to use this factory instead of its current mechanism.

game-core/src/main/java/games/strategy/triplea/ui/menubar/DebugMenu.java
DanVanAtta: Presumably it would not be more than just an alias considering `Map<String, Function<TripleAFrame, Collection<JMenuItem>>>` is not a SAM. The interface would give more semantic meaning behind the operation, for example give additional meaning to what the first string parameter represents.

Accessing `TripleAFrame`, or the 'parent frame' in a static way is probably okay and better than passing it around through-out every layer of code (which creates a dependency on Swing and UI throughout the code). As perhaps a better option, there could also be an injection of a UI component into any code that needs to access UI. In this latter suggestion any data to be rendered would be passed into the injected code which would then contain the UI logic for rendering that data.

trevan: I wasn't able to get an interface to work.  The `= TreeMap<>();` wouldn't compile anymore since `TreeMap` doesn't implement `DebugMenuFactory`.  So, I've added a comment instead to explain what the map is for.

DanVanAtta: Instead of an interface, how about a 1st class object?  The methods on it could then lend it to be more easily understood and otherwise provide a more intuitive API.

Do we really need a TreeMap as well? If the number of elements is just a few, the efficiency gain is seemingly unlikely to be significant.

game-core/src/main/java/games/strategy/triplea/ui/menubar/DebugMenu.java
DanVanAtta: I would like to see this used at some point. If we have nebulous 3rd party code that is never committed, it's going to be impossible to tell if we have a lot of dead code, if we over-engineered, or if there are dependencies on any of this. It will also be difficult for others to try and use as well as there will not be an existing pattern, or if the pattern needs to change, it won't be easy to know what existing dependencies currently exist.

game-core/src/main/java/games/strategy/triplea/ui/menubar/DebugMenu.java
trevan: It needs a sorted map so the debug menu items are sorted.  There's only two built in implementations of sortedmap and I don't think `ConcurrentSkipListMap` is right for this situation.

I don't see the need for a 1st class object.  The map is literally a map of a string to a SAM.  It is extremely basic and simple.

game-core/src/main/java/games/strategy/triplea/ui/menubar/DebugMenu.java
trevan: I've modified the ProAI debug menu to utilize this.

game-core/src/main/java/games/strategy/triplea/ai/pro/AbstractProAi.java
```diff
@@ -86,24 +90,35 @@ public AbstractProAi(
     storedPurchaseTerritories = null;
     storedPoliticalActions = null;
     storedStrafingTerritories = new ArrayList<>();
+
+    // in case there are multiple Pro AIs running at the same time, unregister the one
+    // created before creating one so that there is only one menu item for all of the ProAI
+    DebugMenu.unregisterMenuCallback("Hard AI");
```
DanVanAtta, line 96: Could `registerMenuCallback` handle this detail for us?

game-core/src/main/java/games/strategy/triplea/ai/pro/AbstractProAi.java
```diff
@@ -86,24 +90,35 @@ public AbstractProAi(
...
+    DebugMenu.registerMenuCallback("Hard AI", this::initialize);
   }
 
   @Override
   public void stopGame() {
     super.stopGame(); // absolutely MUST call super.stopGame() first
     calc.stop();
+
+    DebugMenu.unregisterMenuCallback("Hard AI");
```
DanVanAtta, line 105: If the game frame is going away, presumably the menu will be destroyed and GC'd. Do we need to `unregister`? Wouldn't this happen implicitly by the `TripleAFrame` closing?

game-core/src/main/java/games/strategy/triplea/ai/pro/AbstractProAi.java
```diff
@@ -86,24 +90,35 @@ public AbstractProAi(
     storedPurchaseTerritories = null;
     storedPoliticalActions = null;
     storedStrafingTerritories = new ArrayList<>();
+
+    // in case there are multiple Pro AIs running at the same time, unregister the one
+    // created before creating one so that there is only one menu item for all of the ProAI
+    DebugMenu.unregisterMenuCallback("Hard AI");
+    DebugMenu.registerMenuCallback("Hard AI", this::initialize);
```
DanVanAtta, line 97: The "Hard AI" label is in multiple places, it looks like it is a single thing and we would want to keep it consistent. I think this is a place where extracting it to a constant would be a good move so that the linkage between this label and what players select is explicit and would be the same value.

game-core/src/main/java/games/strategy/triplea/ai/pro/AbstractProAi.java
DanVanAtta: Second, it's problematic that AI now has a dependency on swing. AI should ideally not depend on the UI layer at all.

For example, we saw this as a major problem when trying to replace Swing with JavaFX.

Presumably this could be fixed by passing a `DebugMenu` interface to AbstractProAi which it could use to send data but not be aware of the UI implementation detail. WDYT @trevan ?

game-core/src/main/java/games/strategy/triplea/ai/pro/AbstractProAi.java
```diff
@@ -86,24 +90,35 @@ public AbstractProAi(
...
+
+    DebugMenu.unregisterMenuCallback("Hard AI");
   }
 
   public ProOddsCalculator getCalc() {
     return calc;
   }
 
-  public static void initialize(final TripleAFrame frame) {
+  private Collection<JMenuItem> initialize(final TripleAFrame frame) {
```
DanVanAtta, line 112: The frame dependency appears to only be here due to the `initialize` call. It does not look that is related to the menu item. Could we simply move the initialize call up a level and avoid the `frame` parameter?

game-core/src/main/java/games/strategy/triplea/ui/menubar/DebugMenu.java
DanVanAtta: I don't agree it is simple. Looking at this function:

```Function<TripleAFrame, Collection<JMenuItem>>```

It's not clear how a `TripleAFrame` should transform, or 'map' (functions map one element to another), to a JMenuItem. Why is the frame an input to creating a JMenuItem is a mystery. Beyond that, the type and semantic of the 'string' is not readily understood (stringly typing code-smell).

For example, this could perhaps be instead:
```
List<DebugMenu>
```

Where DebugMenu in an interface with the following API:
```
String getMenuTitle();
JMenu getMenuContents();
```

game-core/src/main/java/games/strategy/triplea/ai/pro/AbstractProAi.java
trevan: It is related to the action of the menu item.  It can't be moved up since each menu item might be doing different things and they need access to the TripleAFrame so they can act on it.

game-core/src/main/java/games/strategy/triplea/ai/pro/AbstractProAi.java
trevan: So which code would own the `ProLogAi`?  That looks like it should be AI code since it is only used by AI players.  And the `ProLogAi` can't be made unaware of the UI since it has buttons, checkboxes, etc.  It is a dialog that it needs to render.

Trying to split the debug menu items and their actions from swing or javafx feels like it will just require us to create a new ui library.

trevan: I would actually expect `registerMenuCallback` to throw an exception if a duplicate string is registered.  The duplicate "Hard AI" is a relic of how Hard AI groups everything together in one logging ui.

game-core/src/main/java/games/strategy/triplea/ui/menubar/DebugMenu.java
trevan: `DebugMenu` would still have a `Function<TripleAFrame, Collection<JMenuItems>>`.  So I think it would be better to just improve the comments.  I understand you are trying to remove the need to use `TripleAFrame` but I don't think that is a possible goal.

I'm fine with changing the `Map<String, Function<TripleAFrame, Collection<JMenuItems>>` to `SortedMap<String, Function<TripleAFrame, Collection<JMenuItems>>` if you are worried that knowing if it is sorted or not is important.

game-core/src/main/java/games/strategy/triplea/ai/pro/AbstractProAi.java
trevan: To walk you through why it is needed as part of the action, when you click on the link, it will call `AbstractProAi::showSettingsWindow`.  And that method calls `ProLogUi.showSettingsWindow` which calls `settingsWindow.setVisible(true)`.  And `settingsWindow' was created in the `initialize` method.  So, without the `TripleAFrame`, the action wouldn't be able to function.  Unless you passed in `ProLogWindow` instead of `TripleAFrame` and now the debug factory is dependent on `ProLogWindow` instead of `TripleFrame` and you are limited to what `ProLogWindow` can do.

trevan:
@DanVanAtta , I've moved the Pro Log debug menu registry code into `ProLogUi`.  It has a `registered` boolean that ensures that the menu is only registered once.  With that change, I've removed the "unregisterMenuCallback" code.

trevan:
If there are multiple games running, then using a static register method will be a problem.  Since there might be multiple debug menus in play.

DanVanAtta:
Given a difficult situation this update is not unreasonable per-say.
With that said, I personally would not consider this API to be stable and could change in the future if we can find and invest in a better way to do this. It will be required to change to some extent when there can exist multiple game windows and games in the same process.

PR #8447, opened 2020-12-17
Pass GameMap instead of GameData for TerritoryAttachment methods

trevan:
The four static methods in TerritoryAttachment that are related with
capitals don't need the entire GameData. They only need the GameMap.

I came across this while looking at the AI code.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/TripleAPlayer.java
+++ b/game-core/src/main/java/games/strategy/triplea/TripleAPlayer.java
@@ -566,7 +566,6 @@ public abstract class TripleAPlayer extends AbstractHumanPlayer {
     if (getPlayerBridge().isGameOver()) {
       return;
     }
-    final GameData data = getGameData();
     // play a sound for this phase
     final IAbstractForumPosterDelegate endTurnDelegate;
     try {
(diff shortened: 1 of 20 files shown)


game-core/src/main/java/games/strategy/triplea/attachments/TerritoryAttachment.java
```diff
@@ -78,19 +80,15 @@ public static boolean doWeHaveEnoughCapitalsToProduce(
...
-      final GamePlayer player, final GameData data) {
+      final GamePlayer player, final GameMap gameMap) {
     final List<Territory> capitals = new ArrayList<>();
     final List<Territory> noNeighborCapitals = new ArrayList<>();
-    for (final Territory current : data.getMap().getTerritories()) {
+    for (final Territory current : gameMap.getTerritories()) {
       final TerritoryAttachment ta = TerritoryAttachment.get(current);
       if (ta != null && ta.getCapital() != null) {
-        final GamePlayer whoseCapital = data.getPlayerList().getPlayerId(ta.getCapital());
-        if (whoseCapital == null) {
```
trevan, line 88: These checks are redundant since `setCapital` doesn't allow unknown player names.  So `ta.getCapital()` will either return a null or will return a String that is a valid player name.

game-core/src/main/java/games/strategy/triplea/TripleAPlayer.java
```diff
@@ -566,7 +566,6 @@ private void endTurn() {
     if (getPlayerBridge().isGameOver()) {
       return;
     }
-    final GameData data = getGameData();
```
DanVanAtta, line 569: It's a bit amusing this is used almost 20 lines later.   : )

"Clean code" was written in 2008, this code almost certainly predates it. Interesting to see the concept of 'good code' evolve.

DanVanAtta (approved):
Looks good. Just please double check that we have not

PR #8454, opened 2020-12-19
Create a "ai" project for ai code

trevan:
Updated the PlayerType so that it is no longer an enum but stores a list
of types.  Added a Service Provider so that the game-ai can register ai
players and game-core will see them and add them to the list of
PlayerType.

Added a stub for a new AI called flowfield that is in the game-ai
project.

## Testing

Tested that the new AI showed up when Beta was enabled and that it didn't show up when Beta was disabled.
Loaded save games to make sure the switch from Enum to Class didn't break any saves.

## Release Note

CHANGE|Add new project "game-ai" to store AI related code.


game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
```diff
@@ -0,0 +1,167 @@
...
+  public PlayerTypes() {
+    playerTypes.add(HUMAN_PLAYER);
+    playerTypes.add(WEAK_AI);
+    playerTypes.add(FAST_AI);
+    playerTypes.add(PRO_AI);
+    playerTypes.add(DOES_NOTHING_AI);
+    playerTypes.add(CLIENT_PLAYER);
+    playerTypes.add(BATTLE_CALC_DUMMY);
+
+    final Injections injection = Injections.getInstance();
```
DanVanAtta, line 99: Injections is mean to be very much a roll-your-own dependency injection framework. A framework like Dagger or Spring or Guice were voted down in favor of not having one some long time ago, yet DI is useful specifically for being able to have top level projects inject configuration into lower level layers. That is the background for how it is intended to be used.

With that context, I think Injections should own the full playerTypes list. As-is, we have an existing list of player types and then we append to that list from injections, which spreads the ownership around. I would envision that injections would have exactly the above `PlayerTypes.add(..)` code and would then be able to include the flow-field AI. Then any usages of injections simply needs to call 'injections.getPlayerTypes()'.

Ideally we would avoid the static dependency and try to keep the imperative shell, any outer-layer code in control of injections and then follow DI to pass the player types object to the method or class that needs it when we instantiate it.

Does that make sense? Any thoughts to changing this towards that goal, any hidden problems with that?

game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
```diff
@@ -0,0 +1,167 @@
...
+    if (injection != null) {
```
DanVanAtta, line 100: injections should always be non-null. While it *can* be null, we should be instantiating as early as possible in any main method.

game-ai/src/main/java/org/triplea/ai/flowfield/FlowFieldAiProvider.java
```diff
@@ -0,0 +1,22 @@
...
+    return new FlowFieldAi(name, playerType);
+  }
+
+  @Override
+  public String getLabel() {
+    return "FlowField (AI)";
+  }
+
+  @Override
+  public boolean isBetaPlayer() {
```
DanVanAtta, line 19: Most AI is not going to be beta, and beta should be a relatively short thing in the long run. After we finish this AI we'll have years and years of no beta players. There is a 'isVisible' field on player type, can we use that to hide beta players?

DanVanAtta (commented):
Couple comments. Also, what are your thoughts to name the sub-project AI vs game-AI?

game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
trevan: `Injections` now stores all of the player types.  It is still passing them into `PlayerTypes` when used.  Do you want me to change those to static methods?  I could do that as well.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
trevan: `Injections` is null during tests.

game-ai/src/main/java/org/triplea/ai/flowfield/FlowFieldAiProvider.java
trevan: The `isBetaPlayer` is converted into the `isVisible` field after checking the beta features flag.  Renaming it to `isVisible` has the same problems as what you mentioned.  It will be "invisible" for a relatively short time and then years and years of being visible.  So I didn't change the name but I can do that if you want.

trevan:
@DanVanAtta , I've pushed changes.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
```diff
@@ -63,7 +64,9 @@ public void launch() {
       final Set<Player> gamePlayers =
           gameData
               .getGameLoader()
-              .newPlayers(playerListing.getLocalPlayerTypeMap(new PlayerTypes()));
+              .newPlayers(
+                  playerListing.getLocalPlayerTypeMap(
+                      new PlayerTypes(Injections.getInstance().getPlayerTypes())));
```
DanVanAtta, line 69: I think we need to try and avoid static coupling. That would mean to inject the 'playerTypes' as a constructor arg. This way only code that is instantiating the `LocalLauncher` is aware of `Injections`, but the code using the injected values is not aware of how it is injected.

I have not researched what this would involve, this is a bit of a quick comment. Let me know what you think and if there are any serious issues with constructor injection.

game-headed/src/main/java/org/triplea/game/client/HeadedGameRunner.java
```diff
@@ -103,10 +106,17 @@ public static void main(final String[] args) {
...
-        .aiProviders(StreamSupport.stream(loader.spliterator(), false).collect(Collectors.toList()))
+        .playerTypes(gatherPlayerTypes())
         .build();
   }
+
+  private static Collection<PlayerTypes.Type> gatherPlayerTypes() {
+    return Stream.concat(
+            StreamSupport.stream(ServiceLoader.load(AiProvider.class).spliterator(), false)
+                .map(PlayerTypes.AiType::new),
+            PlayerTypes.getBuiltInPlayerTypes().stream())
```
DanVanAtta, line 119: I was hoping we could have that explicit list of player types here.

I'm surprised to see 'flow field' is not part of the built-in, which makes me wonder what built-in really means.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
DanVanAtta: This is moot now right?

game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
DanVanAtta: `It is still passing them into PlayerTypes when used.`

How do you mean? I think anything that uses `PlayerTypes` should have it constructor injected. This avoids any tests from having to be aware of `Injections` and avoids any client of injected values from knowing how the values are injected as well.

game-ai/src/main/java/org/triplea/ai/flowfield/FlowFieldAiProvider.java
DanVanAtta: Hmm. Interesting. I've had conversations with people who believe feature flags are an anti-pattern. This situation reminds me of those conversations.

I do think we should avoid having specific APIs that change when a feature flag is removed. Ideally all that happens is an `if` block is removed and the `else` part of it is also deleted as it's no longer used.

If all player types come from when we instantiate 'injections', we should be able to feature flag filter flow-field at that time, right?

DanVanAtta: Though `ClientSetting` is part of game-core. It really should not be though. Without lifting `ClientSettings` into game-core we might have to do something kinda nasty to get the flagging right and have it live in `game-headed`...

game-headed/src/main/java/org/triplea/game/client/HeadedGameRunner.java
trevan: "Built-in" are players that are in "game-core".  They are handled slightly different from ai players in the "ai" project.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
trevan: I looked at passing around Injections and that is a massive change to the code base.  It would touch a lot of places.  I really don't want to do that as part of this PR.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
trevan: `PlayerTypes` is now really a bunch of utility methods.  So it could be changed back to static methods.

game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
trevan: The code has been removed.

game-ai/src/main/java/org/triplea/ai/flowfield/FlowFieldAiProvider.java
trevan: `Injections` is instantiated before `ClientSetting` is setup.  So accessing `ClientSetting` during the `Injections` instantiation fails because it isn't setup.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
DanVanAtta: Injections should not be passed around. I do object to Injections being used in a static context (the javadoc on the class says to avoid using it like that).

When an object uses Injections, it gets any data from Injections constructor injected. It should be no more of a massive change than what is already done. Instead of accessing 'Injections.getInsance().getPlayerTypes()' directly, pass the data to the constructor and it just becomes a constructor arg. Let me know if I'm missing something..

game-headed/src/main/java/org/triplea/game/client/HeadedGameRunner.java
DanVanAtta: Could you explain why and how they are handled slightly different? ProAI AFAIK and DoNothing should be things we can migrate into the ai subproject. The other ones are magic and would be deleted once we've fixed up game-core to not depend on AIs for core logic.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
trevan: `Injections should not be passed around` and `object to Injections being used in a static context` are conflicting ideas to me.  If Injections aren't passed around, then you have to access them through a static context.  Can you explain how you access the Injections without passing them around and without accessing them through a static context?

game-headed/src/main/java/org/triplea/game/client/HeadedGameRunner.java
trevan: They can at some point be migrated to the ai subproject.  But until that is done, they have to be treated differently from the flowfield ai because of how the `PlayerTypes.Type` is built for them.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
DanVanAtta: Example: `new MyObject(Injections.getInstance().getData());`


With:
```
@AllArgsConstructor
class MyObject {
  private final Data data;

    void foo() {
             data.doSomething();
     }
}
```


vs:

```
class MyObject {

    void foo() {
             Injections.getInstance().getData().doSomething();
     }
}
```

Hopefully that clarifies through an example. A thing to note is in the first example the implementation of `MyObject` is not aware of Injections at all. Once we have that pattern, it is only a matter of keeping object creation clean (and paves the way to have the option of a clean bona fide IOC framework being dropped in to cleanly replace `Injections`)

game-headed/src/main/java/org/triplea/game/client/HeadedGameRunner.java
DanVanAtta: Okay, I'll see if I can work on this at some point to improve it further, until now I think we can both see the objective but it's not necessarily feasible to get there right away.

game-ai/src/main/java/org/triplea/ai/flowfield/FlowFieldAiProvider.java
DanVanAtta: AFAIK, client settings can be instantiated first. The init of it does not really depend on anything. Can we swap the order? 

Side-note: I've had thoughts we might want to migrate ClientSettings to be backed by a property file. It could lend to easier code and make the static initialization no longer a concern. It is something to think about how to do that more cleanly (it has scaled pretty well to some extent, though the static access and initialization, particularly in test are painful).

DanVanAtta:
Ball is in your court @trevan :
1. constructor injection vs static access of `Injections`
2. merge conflict showed up
3. investigate failing test (I re-ran it, it could be consistent).

game-ai/src/main/java/org/triplea/ai/flowfield/FlowFieldAiProvider.java
trevan: I just swapped the order and the game loaded.  I've pushed a change to remove beta ai when `Injections` is created.  This means that you have to quit the game after you change the beta flag to see changes to the list of players.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
trevan: That's what I thought you meant.  And my concern is still valid: `I looked at passing around Injections and that is a massive change to the code base.`.  I would still have to pass Injections class to all of the callers before `new PlayerTypes` is called.  That is a large change to the codebase.  There are a lot of callers and I'd be touching a good portion of the code path.  I don't think that fits in this PR.

trevan:
@DanVanAtta , I've fixed the merge conflict and I've remove the `isBetaPlayer` code.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
trevan: To give a brief example, this specific code is in `LocalLauncher#loadGame`.  I would have to modify `LocalLauncher#loadGame` to accept an `Injections` or a collection of `PlayerTypes`.  I would then need to modify `LocalLauncher#launch` to accept one or the other.  Since this is an `Override` method, I'd have to modify the parent method as well and any other `Override` versions.  I would then need to modify `MainPanelBuilder#buildMainPanel` to accept one or the other so that I could pass it into `LocalLaunch#launch`.  `MainPanelBuilder#buildMainPanel` is called from `MainFrame#buildMainFrame` so that would need to be modified.  And then its caller, `GameRunner#start` would also need to be modified.  The caller of `GameRunner#start` is `HeadedGameRunner#main` which is where `Injections` is initialized.

So, just for this one place where `Injections.getPlayerTypes()` is used, I'd need to modify at least 4 different classes as well as one parent class and any other child classes of it.  `Injections.getPlayerTypes()` is used 5 times so that is even more classes I'd have to modify.

trevan:
I can't duplicate the integration test locally.  I've run `./verify` and it passes for me.

DanVanAtta:
```
     ERROR [2020-12-20 18:02:29,506] es.moki.ratelimij.dropwizard.filter.RateLimit429EnforcerFilter: Error occurred checking rate-limit. Assuming under limit
    ! java.lang.NullPointerException: null
```

This is the same issue we've had causing non-determinism. Seems to have become worse here somehow. The ratelimit library that supports dropwizard 2.0 is buggy for the simulated rate limit and will NPE when that threshold is hit. 

That error is unrelated to these updates.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
DanVanAtta: That list is actually not unreasonable  :grin: 
I'll work on this next step to help out.

PR #8466, opened 2020-12-20
Headless logging conversion to logback

DanVanAtta:
commit 048db8b90f16149712f9d2b7b76dc90a1d3678f4

    Add headless logback.xml & convert 'chatHandler' to be a 'chatAppender'
    
    This changes the headless server handling of logging java.util
    logging to chat to instead be slf4j based and to send
    slf4j logging to chat. As part of this change we also configure
    headless logging to have a logback.xml which configures logging.
    
    Functionally this does change the output and there is no longer
    a stack trace. This is generally okay, we have yet to receive
    any bug reports where the output message from a bot into
    chat was reported as a bug (generally these logging events
    are treated as informational and otherwise ignored by players).

commit 810a038210225c0d29e12ed6790d2cd742ebb613

    Rename ChatHandler -> ChatAppender

commit 14d3c825760c14ea3df6e672abe737982943d418

    Simplify Appender and remove thread safety guards
    
    J.U.L handlers are not thread safe, logback 'AppenderBase' is
    thread safe.

## Testing
- ran a headless bot locally 
- modified the 'player has joined' info message to be a 'warn' to then verify message output when joining a bot


game-headless/src/test/java/org/triplea/game/server/debug/ChatHandlerTest.java
```diff
@@ -1,91 +0,0 @@
...
-    }
-
-    private void publish(final LogRecord record) {
-      synchronized (chatHandler) {
-        chatHandler.publish(record, sendChatMessage);
-      }
-    }
-
-    @Test
-    void shouldSendChatMessageWhenRecordIsLoggable() {
```
DanVanAtta, line 46: @trevan FWIW, this test written by Ssoloff I think is something of a good test class. Just wanted to refer you to this example.

PR #8468, opened 2020-12-20
Constructor injection

DanVanAtta:
commit 644876617f057c35921ebf9f76197ed72cba9e6a

    Update Injections.java javadoc
    
    - give more details on what it is used from, basically it is a DIY IOC
    - be more explicit about avoiding singleton "pattern" and static coupling

commit 90d524b37e8af88bb9ac0e0de1394f6235c6316d

    Replace static usage of 'Injections.getInsance().getXXX' with constructor injection

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
@@ -36,19 +36,22 @@ public class LocalLauncher implements ILauncher {
   private final PlayerListing playerListing;
   private final Component parent;
   private final LaunchAction launchAction;
+  private final PlayerTypes playerTypes;
 
   public LocalLauncher(
       final GameSelector gameSelector,
       final IRandomSource randomSource,
       final PlayerListing playerListing,
       final Component parent,
-      final LaunchAction launchAction) {
+      final LaunchAction launchAction,
+      final PlayerTypes playerTypes) {
     this.randomSource = randomSource;
     this.playerListing = playerListing;
     this.gameSelector = gameSelector;
     this.gameData = gameSelector.getGameData();
     this.parent = parent;
     this.launchAction = launchAction;
+    this.playerTypes = playerTypes;
   }
 
   @Override
@@ -62,11 +65,7 @@ public class LocalLauncher implements ILauncher {
       gameData.doPreGameStartDataModifications(playerListing);
       final Messengers messengers = new Messengers(new LocalNoOpMessenger());
       final Set<Player> gamePlayers =
-          gameData
-              .getGameLoader()
-              .newPlayers(
-                  playerListing.getLocalPlayerTypeMap(
-                      new PlayerTypes(Injections.getInstance().getPlayerTypes())));
+          gameData.getGameLoader().newPlayers(playerListing.getLocalPlayerTypeMap(playerTypes));
       final ServerGame game =
           new ServerGame(
               gameData,
@@ -123,13 +122,19 @@ public class LocalLauncher implements ILauncher {
                     PlayerCountrySelection::getPlayerName,
                     PlayerCountrySelection::isPlayerEnabled));
 
-    final PlayerListing pl =
+    final PlayerListing playerListing =
         new PlayerListing(
             playersEnabled,
             playerTypes,
             gameSelectorModel.getGameData().getGameVersion(),
             gameSelectorModel.getGameName(),
             gameSelectorModel.getGameRound());
-    return new LocalLauncher(gameSelectorModel, new PlainRandomSource(), pl, parent, launchAction);
+    return new LocalLauncher(
+        gameSelectorModel,
+        new PlainRandomSource(),
+        playerListing,
+        parent,
+        launchAction,
+        new PlayerTypes(Injections.getInstance().getPlayerTypes()));
   }
 }
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerSelectorRow.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerSelectorRow.java
@@ -17,7 +17,6 @@ import javax.swing.JCheckBox;
 import javax.swing.JComboBox;
 import javax.swing.JComponent;
 import javax.swing.JLabel;
-import org.triplea.injection.Injections;
 
 /**
  * Represents a player selection row worth of data, during initial setup this is a row where a
@@ -48,13 +47,14 @@ public class PlayerSelectorRow implements PlayerCountrySelection {
       final Map<String, Boolean> playersEnablementListing,
       final Collection<String> playerAlliances,
       final SetupPanel parent,
-      final GameProperties gameProperties) {
+      final GameProperties gameProperties,
+      final PlayerTypes playerTypes) {
     this.disableable = disableable;
     this.parent = parent;
     playerName = player.getName();
     this.player = player;
     name = new JLabel(playerName + ":");
-    this.playerTypesProvider = new PlayerTypes(Injections.getInstance().getPlayerTypes());
+    this.playerTypesProvider = playerTypes;
 
     enabledCheckBox = new JCheckBox();
     final ActionListener disablePlayerActionListener =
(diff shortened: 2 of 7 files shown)


game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
```diff
@@ -123,13 +122,19 @@ public static LocalLauncher create(
...
-    final PlayerListing pl =
+    final PlayerListing playerListing =
         new PlayerListing(
             playersEnabled,
             playerTypes,
             gameSelectorModel.getGameData().getGameVersion(),
             gameSelectorModel.getGameName(),
             gameSelectorModel.getGameRound());
-    return new LocalLauncher(gameSelectorModel, new PlainRandomSource(), pl, parent, launchAction);
+    return new LocalLauncher(
```
DanVanAtta, line 132: To point out, 'new' calls are grouped together and object construction is now done in fewer places. This removes two levels of static coupling: 1) invoking 'new' from within a constructor body, and 2) invoking 'Injections.getInstance()' within a constructor body.

trevan: You are still using Injections in a static manner, so I'm not sure how this is any better than what I was doing.

trevan (approved):
The changes look fine.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
DanVanAtta: Yes, but note that the 'new' code is all grouped together, the usage of the data is not in a static manner. If you wish to test 'localLauncher', you can do so quite readily. There is no static coupling to the injected data.

Hence, there is a separation of concerns, one is object creation and data injection vs usage. It is the DI principle, or the "don't call us, we'll call you" principle.

The 'new' of LocalLauncher though is another example of static coupling, so we have moved the problem, but the problem is reduced. Instead of having 'N' layers passing data and invoking new, we have "N-1" and can eventually get that to be '1' where we have one factory instantiating all of the classes.

If you're familiar with Spring or Guice, this kind of pattern would result in no changes to `LocalLauncher` for example if we add in a proper DI framework, the constructor code would become automated for us.

DanVanAtta: Let me really try to clarify, the 'static' manner is all about context. From the point of view of who uses the data, it is not static, it is not statically coupled. The creation of `localLauncher` is still statically coupled, and that is a problem. A 'new' call and a static method call are both static coupling. But at least, `LocalLauncher` is not statically coupled to the data it is using.

DanVanAtta: IE:

```
## Bad
constructor() {
    thisWillBePainfulToTest = staticData();
}
```

vs:
```
## Better

constructor(Data thisCanBeInjected) {
    this.canBeInjected = thisCanBeInjected;
}
```

DanVanAtta:
Let's take another example of static coupling, the `ClientSettings`. Any code that relies on it suddenly needs to create a full blown mock `ClientSettings` framework and deal with the entirety of that, just to use a simple 'boolean' value from `ClientSettings`.

Consider this instead:

```
constructor(Supplier<Boolean> isEnabled) {
   ....
}


vs:

construtor() {
  isEnabled = ClientSettings.isEnabled();
}
```

With the first example, there is no static coupling, you can test the constructor by passing it whatever boolean supplier you wish, there is no need to instantiate the full ClientSettings framework (which also makes for slower tests). For further consideration, in case we were to remove 'ClientSettings', if we go with the first route very few tests will need to change. If we go with the second route, we need to re-do any test that interacts with ClientSetting in any way.

TripleA is just rife with static coupling, it's perhaps one of the weakest design elements of how it was coded.

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
trevan: I understand the problem of statically coupling.  That's not the issue I'm talking about.  I'm trying to understand why one level down in the call tree (where I was doing the static calls) is "bad" while this level is "good".  You are still doing static coupling so why is it better to do it here vs lower in the call tree or higher in the call tree?

When I said it would be too big of a change, I was talking about removing the static coupling completely and moving it all the way up to the top of the call tree.  But you only moved it one level up.  If you had asked me to do that, I wouldn't have had a problem with that size of a change.

DanVanAtta:
FWIW, static coupling is an example of 'tight coupling'. Inheritance is an even stronger form of 'tight coupling'. Tightly coupled code is known to be difficult to work with and change. Using dependency injection helps convert tight coupling into loose coupling. In the examples we've seen, the injection breaks the dependency on how the data is obtained, instead it is just provided. "Don't call us, we'll call you". This has implications for versatility, the way the data can be provided can be changed without changing any APIs that use that data, and it allows for easier testing (notably because the test can inject the data how it wants, which is an example of exactly changing how the data is provided).

game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LocalLauncher.java
DanVanAtta: Agree, pushing up one level does not solve the problem.

Though:
- Injections does not yet have any examples where it is used in a pure static context. The data is injected in a static way, but the data used has no examples of being static (the data is always injected).
- Static coupling is worse the deeper it is in the system. One can consider a number of examples, let's say ClientSettings was initialized at a low level. It would be hard to reason about and appear as a very interesting side effect. It would be very painful to determine the ordering of when something is initialized
- Higher level modules depend on lower level modules, the more static coupling in lower level modules, transitively, the more coupling you have on the higher level modules. 
- Imperative shell, functional core is broken, a low level module with static data coupling is not functional. This can be viewed from the perspective of a dependency graph, if low level modules have dependencies on other modules and not just their higher level modules, it creates a tangle of dependencies.
- Low level modules with tight coupling are harder to re-use
- The testing paradigm shifts and you get the anti-pattern of nearly all integration tests to test anything. Higher level are often tested at a system and integration level, which is fine as you would really want to 'new' up the Injections module. In other words, having a higher level module that has mocks for all of its low level systems is not that valuable of a test (at that point it's just verifying wiring, and a test that just verifies that a series of mock objects have called each other is not necessarily that valuable). Meanwhile, testing a low level module in isolation can be useful and valuable. If you have a tight coupling then either you need to instantiate that framework, which means testing is difficult and dependent on that framework (very painful if the framework changes and can introduce dependencies between tests), or a person is using mock objects to inject mock behavior which is also not great as mocks create implementation dependencies between test and the code under test. Meanwhile if you have a low level object, and you can just pass data to it, and assert return values, then the test is not dependent on implementation and is the most flexible (which is great, because you can then change implementation and *not* have to rewrite the test. No testing is worse than poor testing, but working in an environment where a single change leads to several hours of updating many tests is not great).

DanVanAtta: TL;DR

- I don't want us to start a pattern of using `Injections` for static data.
- low level static coupling is worse than static coupling at higher levels.

If we keep moving the coupling up levels, then eventually we will have most of the initialization code in one place. Likely that will be a good thing. To get there I think we'll need to pivot some of the dependencies so instead of having a dependency tree that is narrow and deep, it would be more shallow and broad.

trevan: So the next time you want something moved up a level, can you be a bit more clear?  Because I misunderstood you and thought you wanted the static coupling removed completely and for Injections to be passed around.  But it looks like you just wanted it moved up one level.

DanVanAtta: Sure : )

trevan:
I've noticed that `GameData` is sometimes used as a form of Dependency Injection.  It not only stores state, history, and a lot of other things, but it is used to get access to various game items (GameMap, UnitCollection, Players, etc).  What do you think about creating a `GameDataInjection` interface that only exposes the methods needed for Dependency Injection and then start moving parts of the code that use `GameData` as a DI to use that interface instead.

DanVanAtta:
I like the brainstorming. There would be multiple game data instances once we stop launching a new process per game.

PR #8472, opened 2020-12-20
Move the DoesNothingAi to the ai package

trevan:
The DoesNothingAi label is defined in PlayerTypes because it is used in
the player selection to pick a default AI for certain types of players.

## Testing

Verified that the Does Nothing AI still shows up in the player selection UI.

The only change besides the code movement is the creation of the `DoesNothingAiProvider` and changing the label of `DoesNothingAi` to be a static String on `PlayerTypes`.  This way, the `PlayerSelectorRow` can check that static string instead of static player type, as the player type is no longer static.

### DIFF
--- a/ai/src/main/java/org/triplea/ai/doesnothing/DoesNothingAi.java
+++ b/ai/src/main/java/org/triplea/ai/doesnothing/DoesNothingAi.java
@@ -1,4 +1,4 @@
-package games.strategy.triplea.ai.weak;
+package org.triplea.ai.doesnothing;
 
 import games.strategy.engine.data.Change;
 import games.strategy.engine.data.GameData;
@@ -7,7 +7,8 @@ import games.strategy.engine.data.ResourceCollection;
 import games.strategy.engine.data.changefactory.ChangeFactory;
 import games.strategy.engine.delegate.IDelegateBridge;
 import games.strategy.engine.framework.startup.ui.PlayerTypes;
-import games.strategy.triplea.ai.AbstractBuiltInAi;
+import games.strategy.triplea.ai.AbstractAi;
+import games.strategy.triplea.ai.weak.WeakAi;
 import games.strategy.triplea.delegate.remote.IAbstractForumPosterDelegate;
 import games.strategy.triplea.delegate.remote.IAbstractPlaceDelegate;
 import games.strategy.triplea.delegate.remote.IMoveDelegate;
@@ -18,15 +19,10 @@ import games.strategy.triplea.delegate.remote.ITechDelegate;
  * An AI implementation that takes no action except to purchase and place units according to very
  * simple rules.
  */
-public class DoesNothingAi extends AbstractBuiltInAi {
+public class DoesNothingAi extends AbstractAi {
 
-  public DoesNothingAi(final String name) {
-    super(name);
-  }
-
-  @Override
-  public PlayerTypes.Type getPlayerType() {
-    return PlayerTypes.DOES_NOTHING_AI;
+  public DoesNothingAi(final String name, final PlayerTypes.AiType playerType) {
+    super(name, playerType);
   }
 
   @Override
--- a/ai/src/main/resources/META-INF/services/games.strategy.triplea.ai.AiProvider
+++ b/ai/src/main/resources/META-INF/services/games.strategy.triplea.ai.AiProvider
@@ -1,2 +1,0 @@
-org.triplea.ai.does.nothing.DoesNothingAiProvider
-org.triplea.ai.flowfield.FlowFieldAiProvider
--- a/ai/src/main/resources/META-INF/services/games.strategy.triplea.ai.AiProvider
+++ b/ai/src/main/resources/META-INF/services/games.strategy.triplea.ai.AiProvider
@@ -1,2 +1,2 @@
-org.triplea.ai.doesnothing.DoesNothingAiProvider
+org.triplea.ai.does.nothing.DoesNothingAiProvider
 org.triplea.ai.flowfield.FlowFieldAiProvider
--- a/ai/src/main/resources/META-INF/services/games.strategy.triplea.ai.AiProvider
+++ b/ai/src/main/resources/META-INF/services/games.strategy.triplea.ai.AiProvider
@@ -1,1 +1,2 @@
+org.triplea.ai.doesnothing.DoesNothingAiProvider
 org.triplea.ai.flowfield.FlowFieldAiProvider
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
@@ -17,7 +17,7 @@ import lombok.Value;
 @Value
 public class PlayerTypes {
 
-  public static final String DOES_NOTHING_PLAYER = "Does Nothing (AI)";
+  public static final String DOES_NOTHING_PLAYER_LABEL = "Does Nothing (AI)";
 
   public static final Type HUMAN_PLAYER =
       new Type("Human") {
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
@@ -5,7 +5,6 @@ import games.strategy.triplea.TripleAPlayer;
 import games.strategy.triplea.ai.AiProvider;
 import games.strategy.triplea.ai.fast.FastAi;
 import games.strategy.triplea.ai.pro.ProAi;
-import games.strategy.triplea.ai.weak.DoesNothingAi;
 import games.strategy.triplea.ai.weak.WeakAi;
 import java.util.Collection;
 import java.util.List;
@@ -17,6 +16,8 @@ import lombok.Value;
 
 @Value
 public class PlayerTypes {
+
+  public static final String DOES_NOTHING_PLAYER = "Does Nothing (AI)";
 
   public static final Type HUMAN_PLAYER =
       new Type("Human") {
@@ -49,13 +50,6 @@ public class PlayerTypes {
         @Override
         public Player newPlayerWithName(final String name) {
           return new ProAi(name);
-        }
-      };
-  public static final Type DOES_NOTHING_AI =
-      new Type("Does Nothing (AI)") {
-        @Override
-        public Player newPlayerWithName(final String name) {
-          return new DoesNothingAi(name);
         }
       };
   /** A hidden player type to represent network connected players. */
@@ -94,7 +88,6 @@ public class PlayerTypes {
         PlayerTypes.WEAK_AI,
         PlayerTypes.FAST_AI,
         PlayerTypes.PRO_AI,
-        PlayerTypes.DOES_NOTHING_AI,
         PlayerTypes.CLIENT_PLAYER,
         PlayerTypes.BATTLE_CALC_DUMMY);
   }
--- a/ai/src/main/java/org/triplea/ai/does/nothing/DoesNothingAi.java
+++ b/ai/src/main/java/org/triplea/ai/does/nothing/DoesNothingAi.java
@@ -1,4 +1,4 @@
-package org.triplea.ai.doesnothing;
+package org.triplea.ai.does.nothing;
 
 import games.strategy.engine.data.Change;
 import games.strategy.engine.data.GameData;
--- a/ai/src/main/java/org/triplea/ai/does/nothing/DoesNothingAiProvider.java
+++ b/ai/src/main/java/org/triplea/ai/does/nothing/DoesNothingAiProvider.java
@@ -1,4 +1,4 @@
-package org.triplea.ai.doesnothing;
+package org.triplea.ai.does.nothing;
 
 import games.strategy.engine.framework.startup.ui.PlayerTypes;
 import games.strategy.triplea.ai.AbstractAi;
@@ -12,6 +12,6 @@ public class DoesNothingAiProvider implements AiProvider {
 
   @Override
   public String getLabel() {
-    return PlayerTypes.DOES_NOTHING_PLAYER;
+    return PlayerTypes.DOES_NOTHING_PLAYER_LABEL;
   }
 }
--- a/ai/src/main/java/org/triplea/ai/doesnothing/DoesNothingAiProvider.java
+++ b/ai/src/main/java/org/triplea/ai/doesnothing/DoesNothingAiProvider.java
@@ -1,0 +1,17 @@
+package org.triplea.ai.doesnothing;
+
+import games.strategy.engine.framework.startup.ui.PlayerTypes;
+import games.strategy.triplea.ai.AbstractAi;
+import games.strategy.triplea.ai.AiProvider;
+
+public class DoesNothingAiProvider implements AiProvider {
+  @Override
+  public AbstractAi create(final String name, final PlayerTypes.AiType playerType) {
+    return new DoesNothingAi(name, playerType);
+  }
+
+  @Override
+  public String getLabel() {
+    return PlayerTypes.DOES_NOTHING_PLAYER;
+  }
+}
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerSelectorRow.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerSelectorRow.java
@@ -254,7 +254,7 @@ public class PlayerSelectorRow implements PlayerCountrySelection {
     if (player.isDefaultTypeAi()) {
       playerTypes.setSelectedItem(PlayerTypes.PRO_AI.getLabel());
     } else if (player.isDefaultTypeDoesNothing()) {
-      playerTypes.setSelectedItem(PlayerTypes.DOES_NOTHING_PLAYER);
+      playerTypes.setSelectedItem(PlayerTypes.DOES_NOTHING_PLAYER_LABEL);
     } else {
       playerTypes.setSelectedItem(PlayerTypes.HUMAN_PLAYER.getLabel());
     }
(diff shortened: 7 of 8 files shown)


ai/src/main/java/org/triplea/ai/doesnothing/DoesNothingAi.java
```diff
@@ -1,4 +1,4 @@
-package games.strategy.triplea.ai.weak;
+package org.triplea.ai.doesnothing;
```
DanVanAtta, line 1: nit, does nothing is two words, canonical convention for the project is to word-split packages with periods.

IE: `org.triplea.ai.does.nothing`

I do like on the other hand that 'doesnothing' represents a singular entity, I think it's then a toss up. In case of a toss-up, the existing word-splitting practice I think wins the day.

ai/src/main/java/org/triplea/ai/doesnothing/DoesNothingAi.java
```diff
@@ -18,15 +19,10 @@
...
 
-  public DoesNothingAi(final String name) {
-    super(name);
-  }
-
-  @Override
-  public PlayerTypes.Type getPlayerType() {
-    return PlayerTypes.DOES_NOTHING_AI;
+  public DoesNothingAi(final String name, final PlayerTypes.AiType playerType) {
+    super(name, playerType);
```
DanVanAtta, line 25: In `AbstractAi`, isn't the player type always (and will always be): `  private final PlayerTypes.AiType playerType;`

In that case, is there a way we can avoid passing the `playerType` parameter if it will always have the same value?

Similarly, we should consider the 'name' value as well. I think the AI implementation ought to own the name and know what it's label would be. In this case it does not make sense to have a `name` parameter passed to it either. I suspect it'll take some work to remove the former parameter given how the game is setup, would you agree that is the right direction in general or is there a major obstacle or reason why the AI instance should have parameters?

ai/src/main/resources/META-INF/services/games.strategy.triplea.ai.AiProvider
```diff
@@ -1 +1,2 @@
+org.triplea.ai.doesnothing.DoesNothingAiProvider
```
DanVanAtta, line 1: I missed this from the other PR, what is the purpose of this file?

game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
```diff
@@ -18,6 +17,8 @@
 @Value
 public class PlayerTypes {
 
+  public static final String DOES_NOTHING_PLAYER = "Does Nothing (AI)";
```
DanVanAtta, line 20: This is now just a label for the DNAI? What do you think of adding a `LABEL` suffix to help clarify that?

ai/src/main/resources/META-INF/services/games.strategy.triplea.ai.AiProvider
trevan: Used by the Java Service Provider to find these classes.

ai/src/main/java/org/triplea/ai/doesnothing/DoesNothingAi.java
trevan: `playerType` is a newly constructed object that is generated in `HeadedGameRunner#gatherPlayerTypes`.  So each AI class will have its own `PlayerTypes.AiType` instance.  It isn't a hardcoded value.

`name` and `label` are two different things.  `name` is something like "Germany", "United States", "Italy", etc.  `label` is "Hard (AI)", "Fast (AI)", "Does Nothing (AI)".  The AbstractAi stores the `name` and the `PlayerTypes.Type` (and its child `PlayerTypes.AiType`) stores the label.  When you are selecting the users, the drop down is built from the `PlayerTypes.Type` list.  Once the game starts, the `PlayerTypes.Type` is used to create the actual player object (see `newPlayerWithName`).

So, these parameters (name and playerType) are required for the `AbstractAi`.

ai/src/main/resources/META-INF/services/games.strategy.triplea.ai.AiProvider
DanVanAtta: Oi, I would really like to get rid of that. The usage of the service provider to find the main classes I think has been really difficult to understand. It makes the new processor instantiation seem to be complete magic. Is there a way we can avoid this? It would be really great to have AIs be a matter of adding classes then adding a line to injections.

trevan: I'd rather keep the service provider.  Because it allows adding AI users without bundling them with the full game.  So people could make their own AI and distribute them outside of triplea.  You just have to put the jar in the class path and it is loaded up.  It also means that `game-headed` doesn't really need to depend on `ai`.  It only depends on it now because I was having difficulties getting it built and packaged without making it a dependency of `game-headed`.

DanVanAtta: >  So people could make their own AI and distribute them outside of triplea. 

What's the use of that exactly? You'd still need to have TripleA specific tie-ins, why not just compile them into TripleA? I think this is encouraging a distributed AI system, this is the first new AI we are considering in years. I don't think there will be more than 1 or 2 new AIs total in the next half decade. 


`game-headed` will likely continue to depend on the 'ai' package so it can gather the list of supported AIs. I think making it plug-and-play is overkill. Eventually I'd like to see the entirety of the UI outside of `game-core` and in game-headed, in which case grabbing the list of supported AIs makes more sense there. `game-core` is in some sense a temporary sub-project, all of the code inside of it is intended to go to another sub-project one day.

Could you walk me through how this is used today to discover AIs?

ai/src/main/java/org/triplea/ai/doesnothing/DoesNothingAi.java
DanVanAtta: Ah, in this case `name` is ambigious. The label is 'node name' and name is 'player name'. The latter is poor naming as it is historically called that simply because the single player game was created first, so a each faction or country was named a 'player' (which ceases to make sense once you are in a multiplayer where a 'player' is a person and their name is their call-handle). 

Regardless, 'name' should probably be clarified to be either `playerName` or `factionName`.

trevan: As a newcomer, neither of those are any more clarifying than `name` :)  So what do you want it renamed to?

ai/src/main/resources/META-INF/services/games.strategy.triplea.ai.AiProvider
trevan: > Could you walk me through how this is used today to discover AIs?

With the Service Provider?  In `HeadedGameRunner#gatherPlayerTypes`, it asks the Service Provider to load all providers of type `AiProvider` (https://github.com/triplea-game/triplea/blob/master/game-headed/src/main/java/org/triplea/game/client/HeadedGameRunner.java#L111).  The Service Provider will look through all jars and check for a file `META-INF/services/games.strategy.triplea.ai.AiProvider `.  If it finds that file, it will then load the classes that are listed in that file.

ai/src/main/java/org/triplea/ai/doesnothing/DoesNothingAi.java
DanVanAtta: `playerName` would work. I hate that refers to the faction (aka country), but it is what it is..

ai/src/main/resources/META-INF/services/games.strategy.triplea.ai.AiProvider
DanVanAtta: Thanks. Would you mind if we keep it dumb and simple for now and just list the AIs? That is very akin to DI as done in spring or guice, it's explicit, compile timem safe, and easy to understand. I really think using a service framework falls under the YAGNI category, we can change this later back to the service provider way if we find there is a need for it.

trevan: I've hardcoded the AIs in HeadedGameRunner now.  But I left the Service Provider code so that an AI can still be loaded.  I think it will make it easier for people to experiment with different AIs.

DanVanAtta: Thank you.


We can take some time to consider it , I'm not demanding we remove the rest now. My only remaining concern beyond unused code being judiciously removed, is that we could be setting a pattern I would not really want repeated. Someone could look at the service code example and extend it to AI behaviors, map parsers, rule engines, dice rollers, etc.. All in all, I'd rather those contributions and items be added to the main code base for everyone to share. We also avoid other problems like it being unused and instances where it is over-engineered and never used.


Though, with that said, the remaining service code IMO is dead code. 
- Who specifically do you think it would help? Could they possibly participate in this conversation and their 2 cents?
- How are we going to be sure we don't break any drop-in modules?
- If someone can build an AI, why not just modify TripleA itself?

DanVanAtta:
One unresolved conversation, but all issues that are related to this change have been addressed, thank you @trevan .

ai/src/main/resources/META-INF/services/games.strategy.triplea.ai.AiProvider
trevan: >     * Who specifically do you think it would help? Could they possibly participate in this conversation and their 2 cents?
It would help me.  Instead of needing to push code to triplea to get people to be able to test the ai, I could publish the jar separately and then they can test it.  It's release cycle can then be independent of triplea.
> How are we going to be sure we don't break any drop-in modules?
The releases of AI will have to be tied to specific versions of triplea.  But that really isn't all that different from saved game compatibility or mod compatibility with other games.
> If someone can build an AI, why not just modify TripleA itself?
Lots of reasons.  Some I can think of: different release cycles, different build requirements, different languages, etc.  Opening up AI development outside of TripleA code base allows more people who are just interested in AI to be involved.  It also allows them to release what they've done without needing to "pollute" the main triplea engine.  Take for example my AI engine.  It is quite likely that it will not work.  Being able to build it, share it, discover its uselessness all without modifying the triplea code is a good thing, in my mind.

DanVanAtta: My first thought is that it is admitting defeat for the current PR process. I'd rather us improve the PR process and/or address any process concerns that slow you down. That makes overall development faster in the long term.  My second thought is it's not out of the question to do post-merge reviews.

The TripleA pre-release cycle is very fast, many times a day. I wonder if this really is not fast enough and if it truly is going to be easier to publish your own Jar. Perhaps any tester you could get them set up on a branch with gradle and they can then pull in changes quite readily. There are options worth exploring I think. Bottom line, I'd certainly rather us improve efficiency to meet needs and solve the problem that way rather than code around what is otherwise a human problem.

trevan: I don't see it as admitting defeat.  If I wanted to create a chess engine, I could build one and have it immediately work with most of the chess UIs out there because there is a standard ai engine interface (https://en.wikipedia.org/wiki/Universal_Chess_Interface).  I don't have to submit it to the chess ui, I don't have even have to share it with anyone.

This is the same thing.  Instead of seeing it as a defeat, see it as allowing people to develop AI engines in their own ways.  The game maps are in their own separate repo.  This allows map designers to create maps at a separate pace than the engine.  And that is a good thing.  So is allowing AI developers to create AI at a separate pace than the engine.

DanVanAtta: Maps living in their own repository has created a contract that we must continue to support them and we in essence cannot change any map property that is every introduced. We must always be able to parse every version of XML that has ever existed.

Similarly, if AI can be developed outside of the main engine, it creates a rigid and potentially fragile contract where we can make changes and unknowingly break code. This locks us in, for better or potentially worse to an existing API.

There is another difference between map development and universal chess interface, the latter is intended to be distributed and is to a wide audience. AI development is for an audience of well, one.

Question 1) Let me ask you this, would all of the release problems go away if you had write @trevan and could merge flow-field AI changes immediately after posting a PR?  Would you still want to distribute your own jar file?

Question 2) If wanting to distribute your own Jar -  are we going to run into problems supporting users in test? Can we partition any bug reports?

Question 3) If the development is done as a big bang, and then is merged in, how is anyone else to know how the code is organized and designed? How are we going to be able to support that code 3 or 5 years down the road? One aspect of PR is there is knowledge sharing for the reviewers and creates a paper trail of changes. The project does not get that if the final merge is mostly a big bang.

Question 4) Will we be creating integration problems for ourselves when months down the road you have around somewhere in the ballpark of 10k to 20k  lines of code ready to submit? How can we reasonably review any of that? (4*)

Question 5) How will you get coding ideas/considerations and new perspectives without doing any kind of review? The code I submit has and does benefit from reviews, from pointing out design flaws, oddities, and or outright bugs - you'll be missing that if we do only a big bang review. 



```
# (4*)  Pro AI is 17k lines (blank lines and comments included)
~/work/triplea/game-core/src/main/java/games/strategy/triplea/ai$ find pro/ -name "*.java" | xargs wc -l | grep "total"
 17307 total
```

DanVanAtta: I am pushing back on this, only because I think this is a major design decision. TripleA is not universal chess, we must keep context in mind. The code is already designed with a pretty clean interface for building new AIs. What if others take advantage of this and build their own AIs? And then they submit those AIs to the code base and we get a repeat of the XML editor merge (which was 8k lines of code that were in horrible need of fixing and contained lots of bugs). Meanwhile we have to maintain the XML editor and it has dependencies to the core game engine that prevents other fixes from coming into place because we have to fix the XML editor first.

So the concern is not only about what you would write @trevan , but if we open it up like this, we have to consider others who would use it with all sorts of different development abilities.  At some point, some AIs, if not many (but I don't think there will be many more than one), would want to be integrated back into the engine. I DO NOT want for there to be a special clique of people that share amongst only themselves a great AI that only they can play because it is not part of the engine. This raise the question that if not for you, but for any other AI, how do ensure that we can get those AIs merged back into the engine and be of reasonable quality and not a major source of maintenance and bugs? 

Phrased another way, all code is debt (features are the return on that debt). Accepting *any* code is going to require effort from us on an ongoing basis as pay interest. We're opening the door then to any code landing, which is for better or worse, but then I see problems when we have a great AI but we don't like the code, it's a source of errors, bugs & overhead, yet it's a great AI so we would like to merge into the engine so all could use it. This creates quite a conflict.

trevan: > Maps living in their own repository has created a contract that we must continue to support them and we in essence cannot change any map property that is every introduced. We must always be able to parse every version of XML that has ever existed.
> 
> Similarly, if AI can be developed outside of the main engine, it creates a rigid and potentially fragile contract where we can make changes and unknowingly break code. This locks us in, for better or potentially worse to an existing API.
Triplea is locked in only as much as saved compatibility is locked in.  If triplea needs to make breaking changes, then it makes the breaking changes.

> 
> There is another difference between map development and universal chess interface, the latter is intended to be distributed and is to a wide audience. AI development is for an audience of well, one.
This statement doesn't make sense to me.  You first talk about map development and then talk about AI development.  "map development" is intended for distribution and wide audience.  That's why people make maps.  Triplea AI development would be done for the same reasons that chess AI development is done.  It would also use the triplea interface just like universal chess interface is used.
> 
> Question 1) Let me ask you this, would all of the release problems go away if you had write @trevan and could merge flow-field AI changes immediately after posting a PR? Would you still want to distribute your own jar file?
I would still want to distribute a jar.  Forcing play testers to run this in preview builds of triplea limits the amount of testers.
> 
> Question 2) If wanting to distribute your own Jar - are we going to run into problems supporting users in test? Can we partition any bug reports?
I haven't looked at the bug reporting framework so I don't know.
> 
> Question 3) If the development is done as a big bang, and then is merged in, how is anyone else to know how the code is organized and designed? How are we going to be able to support that code 3 or 5 years down the road? One aspect of PR is there is knowledge sharing for the reviewers and creates a paper trail of changes. The project does not get that if the final merge is mostly a big bang.
You can require the merge to be done over multiple PRs.  It doesn't have to be a big bang.
> 
> Question 4) Will we be creating integration problems for ourselves when months down the road you have around somewhere in the ballpark of 10k to 20k lines of code ready to submit? How can we reasonably review any of that? (4*)
The PR can be broken up.
> 
> Question 5) How will you get coding ideas/considerations and new perspectives without doing any kind of review? The code I submit has and does benefit from reviews, from pointing out design flaws, oddities, and or outright bugs - you'll be missing that if we do only a big bang review.
The PR can be broken up

> I DO NOT want for there to be a special clique of people that share amongst only themselves a great AI that only they can play because it is not part of the engine. This raise the question that if not for you, but for any other AI, how do ensure that we can get those AIs merged back into the engine and be of reasonable quality and not a major source of maintenance and bugs

Why not treat AIs like maps? You can download maps through the game UI and those maps are not part of the main repo.  At some point, triplea could have an AI chooser where there a different AIs that are available to be found.

DanVanAtta:
> Why not treat AIs like maps? 

- There is a lot of overhead to having a generic plugin framework and supporting it.
- We wind up with coupling that is no longer in our control or visibility that prohibits changes that can be made. TripleA is in a very bad state codewise, the project has almost died due to unmanaged growth in complexity. It's going to continue to be a priority for the next several years to heavily emphasize simplifications and reducing overhead.
- AIs are complex, they are large endeavors, we are not going to see very many usable ones total. Making it be a plugin framework for a population of (1) is overkill with the problems mentioned above.

If there were many AIs, how do we ensure that they could operate in bots properly? How does work if someone is hosting? How do we ensure that there is no malicious code embedded in the AI? TripleA has full access to a users system, even the XML is risky in some ways.

PR #8474, opened 2020-12-21
Update triplea_maps.yaml

axisalliescontango:

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2219,5 +2219,11 @@
   version: 1
   img: https://raw.githubusercontent.com/triplea-maps/g40_alt_universe/master/map/snapshot%20%26%20description/G40_Alt_Universe_Map_Snapshot_Small.png
   description: |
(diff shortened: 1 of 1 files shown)


triplea_maps.yaml
```diff
@@ -2213,3 +2213,11 @@
     <p>Warcraft: Lordaeron Wars is a strategy game based on the (World of) Warcraft setting
     by Blizzard Entertainment. Set in the Eastern Kingdoms of Azeroth, you will control one
     of the numerous races and factions vrying for supremacy of the continent.</p>
+- mapName: G40_Alt_Universe
+  mapCategory: EXPERIMENTAL
+  url: https://github.com/triplea-maps/g40_alt_universe/archive/master.zip
+  version: 1
+  img: https://raw.githubusercontent.com/triplea-maps/g40_alt_universe/master/map/snapshot%20%26%20description/G40_Alt_Universe_Map_Snapshot_Small.png
+  description: |
+    <p style='margin:0cm;font-size:15px;font-family:"Calibri",sans-serif;'><strong>G40 Alt Universe is for those who love Axis and Allies 1940 Global, but are looking for a new strategic experience. &nbsp;The war is the same, but occurs in an alternate universe where the Earth&rsquo;s land formation happened differently. &nbsp;It&rsquo;s a return to the early days of pouring over the map and developing strategies in an unfamiliar setting.</strong></p>
```
DanVanAtta, line 2222: you can do line breaks in this file and they will appear as spaces. It also looks like you have a tab on the following line that is breaking the indentation of the file, the tab should be replaced with spaces.

DanVanAtta:
@axisalliescontango I updated the formatting in this PR.
We need to do a quick check on the map contents and then this will be good merge and your map will be available for download.

PR #8479, opened 2020-12-21
Create GameState interface for getter methods of GameData

trevan:
Use IntelliJ to replace all usages of GameData with GameDataInjections
where possible.

BaseEditDelegate#getEditMode takes GameProperties instead of GameData.

## Testing

Loaded a saved game and played Hard AI for several players.

It is possible that IntelliJ converted more than it should.  I did a quick glance over the code and changed a few back to `GameData` but the majority seemed to be ok.


DanVanAtta:
I'm not sure about this one @trevan . 
- Why is `GameDataInjections` being called an injection? Is it not just the interface to `GameData`?  
- Where do you see this going in the future? How will this update be useful and what does it tie into?

Perhaps an interface on `GameData` could very well be a good thing, but if we pass the interface around, we are not really following 'interface-segregation-principle' any more closely before as all of GameData's APIs and data are still exposed. I think it would be better to remove the references to `GameData` and instead have any API that accepts `GameData` instead accept a subset of that data, what it requires. With that said, we should analyze how it makes sense to break up GameData, whether it should be different modules, or a series of different nested modules. That work I don't think should really be taken lightly given how pervasive `GameData` is. It does have a god-class code-smell, and we see problems related to that in testing (for example, it's a really pain to create a `Unit` object). So, there is no lack of problems to solve. That leaves me with a question of how this update will help us move forward along any of those dimensions and whether this is the right long term solution?

trevan:
>     * Why is `GameDataInjections` being called an injection? Is it not just the interface to `GameData`?

It is extracting out the "injections" portion of `GameData`.  As you say, `GameData` is a god object.  It has a lot of rolls and does a lot of things.  By creating an interface that exposes just the dependency injections pieces of `GameData`, it makes it easier to find code that only needs the injections part.  And then those parts of the code can be more easily refactored since you know that it is only using `GameData` for injections and not for locking, history, listeners, changes, etc.


> I think it would be better to remove the references to `GameData` and instead have any API that accepts `GameData` instead accept a subset of that data, what it requires.

This is an initial step towards that ideal.  Before, you didn't know if the class/method was using `GameData` for dependency injection or if it was using it for other reasons.  With this change, you can now easily see that.  And then you can figure out which parts are being injected and change to use it directly.  I did that already as part of this PR with the `getEditMode`.  As I was going over the changes, I was able to see that `getEditMode` was only using the `getProperties` and so I changed the code to pass in `GameProperties` instead of `GameData`.

Another reason for this is it ensures that we don't have "feature creep" in the injections code paths.  I was a little astonished to see that `Change#perform` could take a `GameDataInjections` instead of a `GameData`.  I thought that `Change#perform` was doing more with `GameData` than just using it as a dependency injection.  Now, as we work on converting the `Change` objects to be serializable, we can ensure that we don't misuse `GameData`.

Another reason is that we can look at the code that is still using `GameData` and try and figure out how to move it away from `GameData`, either to `GameDataInjections` or to another interface/class that does what it needs.

DanVanAtta:
Thank you for the context @trevan . I understand the goal and objective now.

If I understand right, `GameDataInjections` is essentially a `GameDataData` object : )

I don't know if the DI analogy is fully analogous. I think I'd feel better about this change if we could find a different name, or if I understand better exactly how this is analogous to an IOC (inversion of control) container like Guice or Spring. 

For background & context into my perspective, the `Injections` class is, for better or worse, occupying the mental space and analogy that it is our DI container. Having another class suffixed `Injections` will make it seem related, when perhaps it is not necessarily related and potentially used in a different way. Hence, either a different name or understanding how it is related I think would help me.

trevan:
> For background & context into my perspective, the `Injections` class is, for better or worse, occupying the mental space and analogy that it is our DI container.

I see `GameData` (at least part of it) as analogous to a DI container.  If I want to get the `GameMap`, I ask `GameData` for the map.  If I want to get `UnitsList`, I ask `GameData`.  If I want to get the `PlayersList`, I ask `GameData`.  There are other aspects of `GameData` that isn't a DI container and I think that is where our confusion is at.  I don't know the other aspects of `GameData` as well as I understand the DI aspect of it.  Most of the code that I've dealt with so far has been using `GameData` as a DI container.  So it feels natural to peel off of the DI aspect of `GameData` into a separate object/interface/etc.  That's why I'm calling it `GameDataInjections`.  I could easily see modifying the `Injections` class to be a delegate of `GameData` and only expose the methods that `GameDataInjections` exposes (ignoring the problem that GameData is constructed much later than Injections is constructed).

DanVanAtta:
Hmm, the difference in perspective is that I don't see `GameData` as a DI container. It's just a massive data object with lots of mutable data. It could very well have `@Data` on it. It's passed in everywhere because it contains almost all interesting data and has become so large it has its own gravitational pull (to do most operations of interest, you need a reference to `GameData`). 

I think the further difference is that most DI is about injecting dependencies that are often more than just data but behavior. DI at the boundaries of a system is good, DI into the internals of a system at every level can often be quite bad. For example DI is good when you need to inject something that fetches data from either a database, or a from a file system. It's not necessarily good if you have the data, but simply need to pass it to one object to another.

trevan:
> Hmm, the difference in perspective is that I don't see `GameData` as a DI container. It's just a massive data object with lots of mutable data. It could very well have `@Data` on it. It's passed in everywhere because it contains almost all interesting data and has become so large it has its own gravitational pull (to do most operations of interest, you need a reference to `GameData`).

The reason you have to have a reference to `GameData` is not because of `GameData` itself but because of all the interesting data that it gives you access to.  Just like a DI container is.  If a person built a DI container but instead of actually using it as DI, they passed the DI container every where, it would look very similar to what `GameData` is now.  That's how I see `GameData` (at least an aspect of it).  It is a badly used DI container.  It needs to be reworked so that instead of passing `GameData` around, it injects the pieces needed.

For example, `Matches.java` has ~50 methods that were changed from `GameData` to `GameDataInjections`.  We now know that those ~50 methods can be rewritten to have the data it needs injected into it instead of `GameData`.  After those methods are modified, that might allow the callers to be modified, and so on.  Over time, we'll be able to actually make `GameDataInjections` be better container instead.

DanVanAtta:
- DI is usually for injecting behavior and/or data accessors. It is not typically used for runtime data objects.

I think that is where I'm really struggling. For example, `Injections` could be entirely replaced by spring, but I am really not sure if we should replace say for example `getPlayerList()` with Spring. It would have to be a supplier of `getPlayerList()`, and even then you need to know which game you are currently referring to. You could not wire an object at startup to know where it will be obtaining it's player list until a game has started. I think that does not quite align with model.

So, I like the interface, I just think having a different name for it would be better. I don't know how you can really dependency inject runtime data. How do we dependency inject game data when there are potentially 3 games running on the same process? I think it generally will need to be passed by reference. I could envision the game objects being done with a factory object, and that factory object being injected, and the factory object getting a reference to `GameData` and then being responsible to partition the data to the needed sub-components. I can't envision DI though being done for runtime data when typically it is done at startup. I think here we are mixing IOC containers and DI perhaps. `Injection` for me is much more tied to having an IOC container compared to dependency injection.

trevan:
> So, I like the interface, I just think having a different name for it would be better.

I'm fine with changing its name.  What do you want it called?

DanVanAtta:
That is a great question @trevan . It seems `GameDataData` is what it is : )

I'm actually really not sure. `GameState` is so far the best I can think of.

trevan:
@DanVanAtta , I've renamed it to `GameState`

DanVanAtta:
I didn't think about this before, but now we have variable names disagreeing with the type, eg:
```final GameState gameData = getData();```

That would be a *lot* to fix. Is it feasible to come back and fix all this, now or later? If not, are we sure we want to incur that hit to the code quality?

trevan:
Let's deal with it later.  After this is merged, I plan to be submitting smaller PRs where I pass in the needed pieces of `GameData` instead of `GameState`.  That should remove the 'data'/'gameData'/etc variables.

PR #8483, opened 2020-12-21
Update triplea_maps.yaml

Navalland2:

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2164,6 +2164,7 @@
   description: |
    <br>A WWII scenario loosely based on 1941, inspired of AA50 Anniversary and Big World.
    <br>WARNING! The balance hasn't extensively tested yet. Still progressing...
+   <br>WARNING! Last Update: 21.12.2020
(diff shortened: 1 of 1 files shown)


DanVanAtta:
@Navalland , I'm curious why the warning? It is unlikely for anyone already with this map installed to see it. Have you considered to change the version number from '1' to '2'? That will trigger a message to anyone with the map that they can download an update.

Navalland2:
I though I had to edit YAML files for having it avaliable for bots also I didn't know that 0 was representing pre-release so in this update I turned it to 1.

DanVanAtta:
I need to track down when bots get map updates. It's either only once when new maps are added or it's whenever there is an update to the game-engine (merge to master). The version number in this file does not indicate anything special, it's a release number essentially. You increment sequentially and whenever it increases players will be notified of a new map.

DanVanAtta:
@Navalland2, the version number in the YAML file is perhaps what you are looking to change. 
I'm actually not sure if the version number in the map is even used anymore.

PR #8485, opened 2020-12-22
Add the player types to the headless Injections

trevan:
## Testing

Started a headless game bot, started lobby, started headed game.  Connected to the local lobby and connected to the headless game.  Before this change, it would throw errors about unknown player types.  Now it works.

I don't generally play in the lobby but it feels like the headless bots have changed behavior.  Now, when I loaded the bot, it immediately started the game with all players using the AI.  I reverted the code to before the introduction of the "ai" project and it still is acting like that.  I'm going to look back further to see if there is a change in behavior.  But that would be fixed in a different PR, if there is a fix needed.

### DIFF
--- a/game-headless/src/main/java/org/triplea/game/server/HeadlessGameRunner.java
+++ b/game-headless/src/main/java/org/triplea/game/server/HeadlessGameRunner.java
@@ -1,13 +1,10 @@
 package org.triplea.game.server;
 
 import games.strategy.engine.framework.startup.ui.PlayerTypes;
-import games.strategy.triplea.ai.AiProvider;
 import java.util.Collection;
 import java.util.List;
-import java.util.ServiceLoader;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-import java.util.stream.StreamSupport;
 import org.triplea.ai.does.nothing.DoesNothingAiProvider;
 import org.triplea.config.product.ProductVersionReader;
 import org.triplea.injection.Injections;
@@ -35,10 +32,7 @@ public final class HeadlessGameRunner {
   private static Collection<PlayerTypes.Type> gatherPlayerTypes() {
     return Stream.of(
             PlayerTypes.getBuiltInPlayerTypes(),
-            List.of(new PlayerTypes.AiType(new DoesNothingAiProvider())),
-            StreamSupport.stream(ServiceLoader.load(AiProvider.class).spliterator(), false)
-                .map(PlayerTypes.AiType::new)
-                .collect(Collectors.toSet()))
+            List.of(new PlayerTypes.AiType(new DoesNothingAiProvider())))
         .flatMap(Collection::stream)
         .collect(Collectors.toList());
   }
--- a/game-headless/src/main/java/org/triplea/game/server/HeadlessGameRunner.java
+++ b/game-headless/src/main/java/org/triplea/game/server/HeadlessGameRunner.java
@@ -1,5 +1,14 @@
 package org.triplea.game.server;
 
+import games.strategy.engine.framework.startup.ui.PlayerTypes;
+import games.strategy.triplea.ai.AiProvider;
+import java.util.Collection;
+import java.util.List;
+import java.util.ServiceLoader;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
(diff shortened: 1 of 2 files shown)


game-headless/src/main/java/org/triplea/game/server/HeadlessGameRunner.java
```diff
@@ -17,6 +26,20 @@ public static void main(final String[] args) {
...
+    return Injections.builder()
+        .engineVersion(new ProductVersionReader().getVersion())
+        .playerTypes(gatherPlayerTypes())
+        .build();
+  }
+
+  private static Collection<PlayerTypes.Type> gatherPlayerTypes() {
+    return Stream.of(
+            PlayerTypes.getBuiltInPlayerTypes(),
+            List.of(new PlayerTypes.AiType(new DoesNothingAiProvider())),
```
DanVanAtta, line 38: Bots are not able to use AI players, does this make sense to have?

trevan: I don't know.  The `DoesNothingAi` existed in the Headless game before this change so I'm not adding anything new.  I could remove it but I'm not sure how to test to ensure that it isn't actually used.  I do know that `PlayerSelectorRow#setDefaultPlayerType` requires the `DoesNothingAi`.  Is that method ever called in the Headless code paths?

game-headless/src/main/java/org/triplea/game/server/HeadlessGameRunner.java
```diff
@@ -17,6 +26,20 @@ public static void main(final String[] args) {
...
+            StreamSupport.stream(ServiceLoader.load(AiProvider.class).spliterator(), false)
```
DanVanAtta, line 39: I'm quite sure we're not going to deploy custom AIs to bots, for a number of reasons:
- none exist yet
- we keep bots uniform, if something is going to be baked into the deployment it is effectively part of the code
- we're fully responsible for bots running, any drop-in AIs are going to be us managing and debugging, at which point it really may as well be part of the main code base (a distributed code base is inherently more complex)
- AIs don't run on bots.

game-headless/src/main/java/org/triplea/game/server/HeadlessGameRunner.java
DanVanAtta: Let's see if this is really needed before merging.

DanVanAtta: Second question, do bots really need a selection of player types? Isn't this to support the launch screen? If this is needed, are bots broken currently without this?

game-headless/src/main/java/org/triplea/game/server/HeadlessGameRunner.java
trevan: I've removed the service provider code.

game-headless/src/main/java/org/triplea/game/server/HeadlessGameRunner.java
trevan: Bots are currently broken as I mentioned in the testing section.

I removed the does nothing ai line and started a bot.  It looks like it isn't complaining.  I can actually start a game now.

game-headless/src/main/java/org/triplea/game/server/HeadlessGameRunner.java
```diff
@@ -17,6 +19,13 @@ public static void main(final String[] args) {
...
   private static Injections constructInjections() {
-    return Injections.builder().engineVersion(new ProductVersionReader().getVersion()).build();
+    return Injections.builder()
+        .engineVersion(new ProductVersionReader().getVersion())
+        .playerTypes(gatherPlayerTypes())
+        .build();
+  }
+
+  private static Collection<PlayerTypes.Type> gatherPlayerTypes() {
+    return PlayerTypes.getBuiltInPlayerTypes();
```
DanVanAtta, line 29: nit, looks like it would be more concise and consistent (with `  .engineVersion(new ProductVersionReader().getVersion())`) to inline this method

trevan: @DanVanAtta , I just made that change.

PR #8495, opened 2020-12-22
CodeClimate: Increase file line count and cognitive complexity thresholds

DanVanAtta:
This update sets the threshold values for line count and cognitive
complexity to be relatively extreme values. This is to avoid
false positives with either metric such that if the metric fails
we can really pay attention to it and have it flag the cases where
we really would want a fix.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -9,10 +9,13 @@ checks:
       threshold: 5
   method-complexity:
     config:
-      threshold: 7
+      threshold: 25 
   method-lines:
     config:
       threshold: 50
+  file-lines:
+    config:
+      threshold: 1000
   similar-code:
     config:
       threshold: 80


DanVanAtta:
@trevan could you provide a second opinion on this update?

I've been looking through the issues in codeclimate that this issue fixes: https://codeclimate.com/github/triplea-game/triplea/pull/8495

We should see in that list only marginal problems where if they were presented in a PR, we would be okay with that code. We should *not* see in that list anything that is egregious where if it were new code we would be like "whoa, this needs to be simplified".

So far it looks like most cognitive complexities are in the low teens or single digits, but otherwise they are higher or dramatically higher and for blocks of code that really look like it could be simpler. So I think perhaps this has gotten the thresholds about right.

I'd like to get closer to a place where there are fewer false positives in the build checks, that a 'red X' would really be something to fix.

PR #8496, opened 2020-12-22
Move some matches to use relationshiptracker instead of game state

trevan:
This modifies a few of the Matches methods to take just the `RelationshipTracker` instead of the full `GameState`/`GameData`.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
@@ -966,9 +966,9 @@ public final class Matches {
   }
 
   static Predicate<Territory> territoryHasAlliedIsFactoryOrCanProduceUnits(
-      final GameState data, final GamePlayer player) {
+      final RelationshipTracker relationshipTracker, final GamePlayer player) {
     return t ->
-        isTerritoryAllied(player, data.getRelationshipTracker()).test(t)
+        isTerritoryAllied(player, relationshipTracker).test(t)
             && t.getUnitCollection().anyMatch(unitCanProduceUnits());
   }
 
--- a/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
@@ -245,8 +245,9 @@ public final class Matches {
     return unit -> UnitAttachment.get(unit.getType()).getDefense(unit.getOwner()) >= defendValue;
   }
 
-  public static Predicate<Unit> unitIsEnemyOf(final GameState data, final GamePlayer player) {
-    return unit -> data.getRelationshipTracker().isAtWar(unit.getOwner(), player);
+  public static Predicate<Unit> unitIsEnemyOf(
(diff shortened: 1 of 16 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
```diff
@@ -245,8 +245,9 @@ private Matches() {}
     return unit -> UnitAttachment.get(unit.getType()).getDefense(unit.getOwner()) >= defendValue;
   }
 
-  public static Predicate<Unit> unitIsEnemyOf(final GameState data, final GamePlayer player) {
-    return unit -> data.getRelationshipTracker().isAtWar(unit.getOwner(), player);
+  public static Predicate<Unit> unitIsEnemyOf(
```
DanVanAtta, line 248: For consideration, this might not be the right way remove `GameData` from Matches. It could make more sense to instead to inline this match altogether. In that case we'd want the object invoking this to not only have the match inlined, but also for that object to keep a reference to relationship tracker and not `GameData`.

I think any one-line matcher needs to be really questions. The `Matches` API is a DIY java8 Predicate that was created before Java8 came out. We are now re-inventing a java built-in here.

EG:
```
.filter(unitIsEnemyOf(relationshipTracker, player))
```

vs:
```
.filter(unit -> relationshipTracker.isAtWar(unit.getOwner(), player))
```

I think we need to consider how we are going to update Matches, whether it really is a nicer API for such cases. If so, this update makes sense, if not, then the update here is churn.

trevan: I'm fine with inlining Matches, if that is the direction we want to go.  Should I only inline small, one line matches?

And, yes the caller should also not be keeping a reference to `GameData`.  I'm trying to make these changes in pieces so that the PR is fairly small.  I'm first focusing on removing `GameState` from `Matches` as much as possible.  Then I'll work on other parts of the code that use `GameState` and try and replace it with the actual data (relationship tracker, unitlist, etc).

DanVanAtta: I'm very much posing a question, it's an open consideration. I'm not sure of the answer and I'd rather not create a mandate on my opinion alone here.

DanVanAtta: To answer the other question, yes, I think small & one line matches are the most obvious candidates for inlining. The more complex ones, perhaps those are good candidates to move to the class that owns the data about them? For example let's say there were a complex matcher that used `relationshipTracker`, it could make sense to move that matcher to be a method in `relationshipTracker`. (After all, who is to say that `isAtWar` is not a complex matcher itself).

trevan: > The more complex ones, perhaps those are good candidates to move to the class that owns the data about them?

I think that sounds like a good idea.  Though, some of the complex matchers use data from multiple objects.  Let me do more conversions and we'll be able to really see which objects the Matches really do use.

PR #8500, opened 2020-12-22
Add a shutdown registry for registering code to run on shut down

trevan:
## Testing

Started a game with all Hard AI.  Left the game.  Started it again.  Left it.  Using breakpoints, I ensured that the shutdown code was being called and was being cleared up after it was called.

I put the rational behind this in https://github.com/triplea-game/triplea/discussions/8492.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/GameRunner.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/GameRunner.java
@@ -17,7 +17,6 @@ import games.strategy.engine.framework.startup.mc.ServerModel;
 import games.strategy.engine.framework.startup.ui.panels.main.SetupPanelModel;
 import games.strategy.engine.framework.startup.ui.panels.main.game.selector.GameSelectorModel;
 import games.strategy.engine.framework.ui.MainFrame;
-import games.strategy.triplea.ai.pro.ProAi;
 import java.awt.Component;
 import java.awt.Frame;
 import java.io.File;
@@ -74,7 +73,7 @@ public final class GameRunner {
    * lobby/single player game etc..) screen presented to GUI enabled clients.
    */
   public static void showMainFrame() {
-    ProAi.gameOverClearCache();
+    GameShutdownRegistry.runShutdownActions();
 
     if (System.getProperty(TRIPLEA_SERVER, "false").equals("true")) {
       MainFrame.show();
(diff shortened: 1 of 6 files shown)


DanVanAtta:
Is this different from: [ExitStatus#addExitAction](https://github.com/triplea-game/triplea/blob/master/game-core/src/main/java/org/triplea/util/ExitStatus.java#L20)?

DanVanAtta:
Ah, answered my own question - this is a for a game, but not the program itself. Okay : )

DanVanAtta:
@trevan it does not seem like this would work if multiple game instances exist in one process. Could you comment?

DanVanAtta:
There is a doc somewhere that states the design constraints of AI. Notably it said things like an AI should be able to recreate its state from one save to another, possibly even from one game round to another and that it was not allowed to retain state. This draws into question whether we should really support a cache (and it would be worthwhile to check the benefit we are gaining from this cache to confirm it's really significant). We certainly could try to reduce the cost of computing the cached data or even outright remove it if the benefit is marginal.

Next, we likely want this kind of registry to be stateful. I think a `GameInstance` object likely would make sense. The `GameInstance` ideally would have a well designed API, perhaps have as a member variable like `GameState`. Attaching to the `GameInstance` cache data would avoid any need to have a callback as the data would no longer be static and automatically would go out of scope at the same time as the `GameInstance`.

trevan:
@DanVanAtta , this change doesn't actually change any current functionality.  It is only removing the static call into `ProAi`.  So if multiple game instances running at the same time would have problems, then it would have problems with the current code.

I don't plan on using this registry in my AI.  I'm just trying to make the code not depend on `ProAI` so that it can be moved to the `ai` package.

trevan:
I guess I should rephrase.  My AI is not going to have a static cache like ProAi has.  But it will be using the ProLogUI so that to be cleaned up outside of the ProAi code.

DanVanAtta:
It's hard to know whether this is infrastructure we would want to build on. Static data being cleared on game close will be a problem for multi-games per instance. Meanwhile we want the AI cache to be cleared in a stateful way ideally. I don't know if we are just punting on the problem here. On the other hand, removing ProAI dependencies in game-core is very useful for sure. Would it make sense to mark the registration method as deprecated out of the gate so we know not to build on it?

trevan:
@DanVanAtta , that sounds look a fine idea.  I've marked it as deprecated and added a note of using `AbstractBasePlayer#stopGame` instead.  I think that will work in most cases.

PR #8503, opened 2020-12-23
Unify how step names are detected

trevan:
GameStep had most of the steps already so this adds the remainder. Also,
this updates TripleAPlayer and AbstractBuiltInAi to use the GameStep
methods that they weren't using before.

## Testing

Played Global40 Fast AI

There is already a `GameStep` object but it isn't available in the `start` method because the step name is passed between server and clients.  But I think we should change `start` to be a final method and have it turn the step name into a `GameStep` object and then call an abstract protected method (named something like `runStep`, `runPhase`, `startStep`).  Then this new protected method would have a `GameStep` object and it could have an enum part of it to indicate which step it is.  This would probably also allow merging some code in `GameStep` and `GameStepPropertiesHelper`.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/GameStep.java
+++ b/game-core/src/main/java/games/strategy/engine/data/GameStep.java
@@ -165,15 +165,51 @@ public class GameStep extends GameDataComponent {
         || name.endsWith("NonCombatMove");
   }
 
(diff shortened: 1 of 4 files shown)


DanVanAtta (approved):
Data encapsulation FTW  :+1:

PR #8510, opened 2020-12-25
Fix concurrent modification exception when setting player colors

DanVanAtta:
When launching the game, we can get a 'ConcurrentModificationException'
when player colors are assigned.

This fix reverts to using a non-atomic add to the player color map
and removes usage of 'computeIfAbsent' (which is not thread-safe).

Because we set the player color to the same value, but possibly doing
so concurrently, we do not need a thread-safe nor atomic add of the
player color (we just need it to work). By going to a non-atomic
version, we avoid code that would otherwise throw a
'ConcurrentModificationException' on concurrent adds.

## Release Note

FIX|Fix rare game launch error 'ConcurrentModificationException' related to assigning default player colors.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/mapdata/PlayerColors.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/mapdata/PlayerColors.java
@@ -48,7 +48,14 @@ public final class PlayerColors {
         "Illegal player name: %s, use the method 'getImpassableColor()' instead",
         playerName);
 
-    return playerColors.computeIfAbsent(playerName, this::computePlayerColor);
+    // NOTE: we do *not* use computeIfAbsent here to avoid
+    // 'java.util.ConcurrentModificationException'
+    Color playerColor = playerColors.get(playerName);
+    if (playerColor == null) {
+      playerColor = computePlayerColor(playerName);
+      playerColors.put(playerName, playerColor);
+    }
+    return playerColor;
   }
 
   private Color computePlayerColor(final String playerName) {
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -9,7 +9,7 @@ checks:
       threshold: 5
   method-complexity:
     config:
-      threshold: 25 
+      threshold: 25
   method-lines:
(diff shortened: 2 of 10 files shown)


game-core/src/main/java/games/strategy/triplea/ui/mapdata/PlayerColors.java
```diff
@@ -48,7 +48,14 @@ Color getPlayerColor(final String playerName) {
...
 
-    return playerColors.computeIfAbsent(playerName, this::computePlayerColor);
+    // NOTE: we do *not* use computeIfAbsent here to avoid
+    // 'java.util.ConcurrentModificationException'
+    Color playerColor = playerColors.get(playerName);
+    if (playerColor == null) {
+      playerColor = computePlayerColor(playerName);
+      playerColors.put(playerName, playerColor);
+    }
+    return playerColor;
```
RoiEXLab, line 58: Feel free to correct me if I'm wrong but isn't a `ConcurrentModificationException` a more correct behaviour?
If for example 2 Threads try to query the player name at the same time, with this code no exception would be thrown, but it could happen that the code computes 2 different colors, which are both returned, but only one of them is stored in the map.
Unlikely scenario, but we might want to consider using a synchronized/concurrent Map instead

DanVanAtta: This is an example of resolving the race condition by 'let it happen'.
In this case, the update is to the same value, we guaranteed that player color computation is deterministic.

DanVanAtta: Let's also consider if we were to resolve the race condition via exception and that the colors were different. In this case one writer could write a value, the value then read by another reader. The other writer would come in and set the value to something else, meanwhile the other reader has the old value, and that would be incorrect. Having a concurrent exception does not solve the race condition as we need to be sure that any readers are going to get consistent results. So this is properly solved with deterministic player color computation (nonetheless it getting written multiple times is not clean).

PR #8537, opened 2020-12-28
Null player can have no gameData so use unitType to get gameData

trevan:
A null pointer occurs pretty much anytime a neutral player is attacked.

## Testing

I noticed this playing a hard ai game.  I thought about giving the null player a valid gamedata but decided to just use the passed in unit type instead.


DanVanAtta:
This has been the source of several errors, I suspect maybe not the last, but the -1 is very welcome.

The build did not pass @trevan , would you mind having a look?

PR #8542, opened 2020-12-29
Move getters from DelegateFinder to GameData/GameState

trevan:
I'm not sure what the original purpose of `DelegateFinder` is.  It was added back in the first few commits.  But it feels like this abstraction isn't needed anymore.  By adding the functions to `GameData` and `GameState`, I'll be able to split up the use of `GameData` even more.  I'm hoping to be able to change `ProData` to not use `GameData` anymore but the few pieces that it does use (`GameProperties`, `GameMap`, `BattleDelegate`, `RelationshipTracker`, etc).


game-core/src/main/java/games/strategy/engine/data/GameData.java
```diff
@@ -398,6 +402,34 @@ public BattleRecordsList getBattleRecordsList() {
     return battleRecordsList;
   }
 
+  @Override
+  public PoliticsDelegate getPoliticsDelegate() {
```
trevan, line 406: These 5 methods were moved from `DelegateFinder` and modified slightly to not be static methods.  All of the other changes in this PR are edits to use these getters instead of the ones in `DelegateFinder`.

DanVanAtta (approved):

PR #8548, opened 2020-12-30
Add initial work for creating diffusion maps and lanchester calculator

trevan:
Here's my initial work on the ai.  It basically just creates a few of the diffusion maps, implements the lanchester calculator, and hooks up debug so that I can see it in action.  The ai doesn't actually do anything yet.

I'm submitting this so that the hooks into the debug system and the basic setup can be reviewed.

## Testing

I've started a game with the ai to make sure the debug menus work and that things are being built correctly.


ai/build.gradle
```diff
@@ -3,4 +3,7 @@ description = 'TripleA ai implementations'
 dependencies {
     implementation project(":game-core")
     implementation project(":java-extras")
+    implementation project(":swing-lib")
+    testImplementation project(":test-common")
+    testCompile project(':game-core').sourceSets.test.output
```
trevan, line 8: Without this line, the ai tests don't work.  It needs the compiled test files from the game-core.

ai/src/main/java/org/triplea/ai/flowfield/FlowFieldAi.java
```diff
@@ -3,41 +3,174 @@
...
   public FlowFieldAi(final String name, final PlayerTypes.AiType playerType) {
     super(name, playerType);
   }
 
+  @Override
+  public void initialize(final IPlayerBridge playerBridge, final GamePlayer gamePlayer) {
+    super.initialize(playerBridge, gamePlayer);
+    DebugMenu.registerMenuCallback(getName(), this::addDebugMenuItems);
+
+    final DiffusionType enemyCapitals =
```
trevan, line 50: This code of adding the two diffusions is not the final thing.  It is there just to make sure things are building and showing up.

ai/src/main/java/org/triplea/ai/flowfield/FlowFieldAi.java
```diff
@@ -3,41 +3,174 @@
...
       final IAbstractPlaceDelegate placeDelegate,
       final GameState data,
-      final GamePlayer player) {}
+      final GamePlayer player) {
+    if (isNewRound(getGameData().getCurrentRound())) {
+      calculateTurn();
+    }
+  }
+
+  private void calculateTurn() {}
```
trevan, line 175: This is where diffusion maps will be calculated as well as the rest of the turn calculations.  The `@Override` methods from the ai player will just utilize the data that is generated in this method (and its helpers).

ai/src/main/java/org/triplea/ai/flowfield/map/FieldTerritory.java
```diff
@@ -0,0 +1,34 @@
...
+import games.strategy.engine.data.Territory;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+@Data
+@EqualsAndHashCode(exclude = "neighbors")
```
trevan, line 12: `neighbors` needs to be excluded because otherwise lombok will create an equals and hashcode that recurses n itself.

ai/src/main/java/org/triplea/ai/flowfield/map/FieldTerritory.java
```diff
@@ -0,0 +1,34 @@
...
+public class FieldTerritory {
+  private Territory territory;
+  private long value;
+  @ToString.Exclude private final Map<Territory, FieldTerritory> neighbors = new HashMap<>();
```
trevan, line 16: I'd rather this just be a `Set` since `FieldTerritory` has a reference to its `Territory`.  But when I did that, for some reason, I'd get duplicate neighbors in the set.

ai/src/main/java/org/triplea/ai/flowfield/odds/LanchesterDebugUiAction.java
```diff
@@ -0,0 +1,155 @@
...
+
+@RequiredArgsConstructor
+public class LanchesterDebugUiAction extends AbstractAction {
+  private static final long serialVersionUID = -919496373521710039L;
+
+  private final TripleAFrame frame;
+  private final FlowFieldAi ai;
+
+  @Override
+  public void actionPerformed(final ActionEvent e) {
```
trevan, line 42: This debug item only works in Global40.  My hopes are to make this be able to calculate what the lanchenster attrition order is so that map makers could then add that to a file in their map.  It currently just runs a battle simulation and a lanchester calculation and displays the two results.

game-core/src/main/java/games/strategy/triplea/ui/screen/TileManager.java
```diff
@@ -567,6 +567,20 @@ public void setTerritoryOverlay(
...
 
+  public void setTerritoryOverlayForTile(
+      final Territory territory,
+      final Color color,
+      final int alpha,
+      final GameData data,
+      final MapData mapData) {
+    synchronized (mutex) {
+      final IDrawable drawable =
+          new TerritoryOverLayDrawable(color, territory.getName(), alpha, Operation.FILL);
```
trevan, line 578: I thought `setTerritoryOverlay` would be what I wanted but it only colors the border of the territory.  And that isn't very visible.  So, this new method uses the `FILL` operation so that the entire territory is colored.

game-core/src/main/java/games/strategy/triplea/ui/TerritoryDetailPanel.java
```diff
@@ -108,6 +112,14 @@ public void setGameData(final GameData data) {
     territoryChanged(null);
   }
 
+  public void addAdditionalTerritoryDetailsFunction(final Function<Territory, String> method) {
```
trevan, line 115: These two methods (`addAdditionalTerritoryDetailsFunction` and `removeAdditionalTerritoryDetailsFunction`) allow the ai to add text to the territory details panel.  This, with the heat map overlay, makes it really easy to see what the territory valuations are.

trevan:
I couldn't figure out how to use the ww2_g40_balanced.xml resource in game-core for my tests so I've copied it over to the ai package.  That is 15,000 lines in this PR.  If there is a way to not have to copy it to the ai package, I'll gladly do that.

DanVanAtta:
What kinds of things should a reviewer pay particular attention here to be of most value @trevan ?

ai/build.gradle
```diff
@@ -3,4 +3,7 @@ description = 'TripleA ai implementations'
 dependencies {
     implementation project(":game-core")
     implementation project(":java-extras")
+    implementation project(":swing-lib")
```
DanVanAtta, line 6: The swing dependencies in AI would be great to break. It would very much allow for an easier path to JavaFX and perhaps be better general decoupling of AI with UI.

Can we convert the dependencies to be data dependencies instead of actual UI components? What kinds of things do we need UI components for in this AI?

trevan: It is the debug menu and dialogs.  Maybe I should create a "ai-swing-debug" package that is dependent on "swing-lib".  Then "ai" can depend on "ai-swing-debug" but use the new package for all of the debug ui creation stuff.

trevan:
> What kinds of things should a reviewer pay particular attention here to be of most value @trevan ?

At this point, it is mainly the debug pieces.  Most of what I've done has been to ensure that I can visually debug what is happening.

Another thing is the pieces in map.defense and map.offense.  This is where the logic to calculate territory valuations will be.  So, I might be calculating things incorrectly or rewriting parts of the engine without realizing it.

And is there a better way to gain access to the global40.xml file other than copying it?

ai/build.gradle
DanVanAtta: That might just be punting the problem. Perhaps the game can provide an API that AI can use to send debug data and let the game handle the UI elements. Then it'll be up to game-core to bind data to UI.

trevan: "swing-lib" brings in `SwingAction` which is used to handle `JMenuItem` actions.  So, if you don't want "swing-lib", then the ai can't build `JMenuItem`s.  Instead, I guess it will have to send back a map of `TripleaUiAction`s that have the name of the item, any child `TripleaUiAction`s, and whether it is a checkbox, radio button, or regular item.

The other 5 interactions with swing code that the debug uses is
```
TripleAFrame.getTerritoryDetails().addAdditionalTerritoryDetailsFunction(territoryDetailMethod);    TripleAFrame.getTerritoryDetails().removeAdditionalTerritoryDetailsFunction(territoryDetailMethod);
TripleAFrame.getMapPanel().setTerritoryOverlayForTile(territory, color, 100);
TripleAFrame.getMapPanel().clearTerritoryOverlay(territory);
TripleAFrame.getMapPanel().repaint();
```
(see `TerritoryDebugUiAction`)

I was also planning on having `LanchesterDebugUiAction` show the calculated `attributionFactor` or maybe hooking into the `BattleCalculator` somehow.  This isn't done yet.

DanVanAtta: Just a few quick comments:
- `Map<Integer, Integer> = TreeMap<>`, List<Tuple<Integer, Integer>>`, `List<Point>` are all effectively the same data structure, the latter is the most descriptive though. To avoid javascript'esque typing, we can create interfaces that describe the data being passed. IE:
```
interface DebugMenuItem() {
   String getTitle();
   void renderContent(Consumer<Supplier> sendDataToMe);
}
```

We can probably do very similar to hide the 'mapPanel' and 'territory' details as well. EG:

```
Consumer<TerritoryName>  territoryClearFunction =  TripleAFrame.getMapPanel()::clearTerritoryOverlay
```
Is that workable at all? I think there just needs to be an inversion, java 8 functional interfaces potentially being very useful here to bind specific method calls to functions needed by AI. One example where this was used was with LobbyWatcher, instead of having a HeadlessLobbyWatcher and Headed, the only difference needing to be how errors were handled, the two were unified and accepted a `Consumer<String> errorReporter` that was in one case bound to a UI window and in the other was just a Logger call.

trevan: I'm not exactly sure what the first comment is about.

But the `DebugMenuItem` interface won't cut it.  You'd need a class like:
```
@Value
class UiMenuItem {
  enum MenuType {
    CHECKBOX,
    RADIOBUTTON,
    DEFAULT,
    MENU,
  }
  String title;
  MenuType menuType;
  List<UiMenuItem> children;
  Consumer<DebugMenuActions> performAction;
}
```
Then `DebugMenuActions` would look like:
```
interface DebugMenuActions {
  Consumer<Map<Territory, String>> addAdditionalTerritoryDetails;
  Consumer<Map<Territory, String>> removeAdditionalTerritoryDetails;
  Consumer<TerritoryOverlayDetails> addTerritoryOverlay;
  @Value
  class TerritoryOverlayDetails {
    Territory territory;
    Color color; // Color is awt so that might need to change to some other abstraction
    int alpha;
  };
  Consumer<Territory> clearTerritoryOverlay;
  Runnable repaintMap;
}
```
The `DebugMenuActions`, though, will slowly get bigger as other debug requirements are added.

trevan: And that api doesn't come close to encompassing what the ProAi needs for its debug menu and dialog.  It needs to not only create menu items but a dialog with buttons, text areas, checkboxes, etc.  It needs action listeners and probably many other things.

DanVanAtta:
Clicking on 'lanchester' debug option, I got:

```
08:01:50.038 [AWT-EventQueue-0] ERROR o.t.game.client.HeadedGameRunner - null
java.lang.NullPointerException: null
	at games.strategy.triplea.odds.calculator.BattleCalculator.calculate(BattleCalculator.java:69)
	at games.strategy.triplea.odds.calculator.ConcurrentBattleCalculator.lambda$calculate$4(ConcurrentBattleCalculator.java:232)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceTask.doLeaf(ReduceOps.java:952)
	at java.base/java.util.stream.ReduceOps$ReduceTask.doLeaf(ReduceOps.java:926)
	at java.base/java.util.stream.AbstractTask.compute(AbstractTask.java:327)
	at java.base/java.util.concurrent.CountedCompleter.exec(CountedCompleter.java:746)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinTask.doInvoke(ForkJoinTask.java:408)
	at java.base/java.util.concurrent.ForkJoinTask.invoke(ForkJoinTask.java:736)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateParallel(ReduceOps.java:919)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:233)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
	at games.strategy.triplea.odds.calculator.ConcurrentBattleCalculator.calculate(ConcurrentBattleCalculator.java:244)
	at org.triplea.ai.flowfield.odds.LanchesterDebugUiAction.actionPerformed(LanchesterDebugUiAction.java:82)
	at org.triplea.swing.SwingAction$1.actionPerformed(SwingAction.java:60)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1967)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2308)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:405)
	at java.desktop/javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:262)
	at java.desktop/javax.swing.AbstractButton.doClick(AbstractButton.java:369)
	at java.desktop/javax.swing.plaf.basic.BasicMenuItemUI.doClick(BasicMenuItemUI.java:1020)
	at java.desktop/javax.swing.plaf.basic.BasicMenuItemUI$Handler.mouseReleased(BasicMenuItemUI.java:1064)
	at java.desktop/java.awt.Component.processMouseEvent(Component.java:6635)
	at java.desktop/javax.swing.JComponent.processMouseEvent(JComponent.java:3342)
	at java.desktop/java.awt.Component.processEvent(Component.java:6400)
	at java.desktop/java.awt.Container.processEvent(Container.java:2263)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:5011)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2321)
	at java.desktop/java.awt.Component.dispatchEvent(Component.java:4843)
	at java.desktop/java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4918)
	at java.desktop/java.awt.LightweightDispatcher.processMouseEvent(Container.java:4547)
	at java.desktop/java.awt.LightweightDispatcher.dispatchEvent(Container.java:4488)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2307)
	at java.desktop/java.awt.Window.dispatchEventImpl(Window.java:2772)
	at java.desktop/java.awt.Component.dispatchEvent(Component.java:4843)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:772)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:721)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:715)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:85)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:95)
	at java.desktop/java.awt.EventQueue$5.run(EventQueue.java:745)
	at java.desktop/java.awt.EventQueue$5.run(EventQueue.java:743)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
```

trevan:
> Clicking on 'lanchester' debug option, I got:

I mentioned in one of my earlier comments, this debug option only works for Global40 right now.  It doesn't do much at this point besides run a battle simulator and compare the results with the lanchester results.  The plan is for it to try and calculate what the attrition factor should be for the specific map.  I don't know yet if the attrition factor will change per map but I'm thinking it probably will so this will give map makers the chance to auto calculate the number that they can then put in some file in the map folder.

ai/build.gradle
DanVanAtta: The first comment is about unstructured data. Any class with string fields could arguably be represented as a Map. Hence, rather than representing data as a map, more structure can be added. For example, `map.get("x")`  and `Tuple.getFirst()` are both less structured variants of `Point.getX()`

> And that api doesn't come close to encompassing what the ProAi needs for its debug menu and dialog.

It was not meant to be a comprehensive example. I would not worry about ProAi for just now and focus on creating a clean interface for the new debug and avoid adding to any existing problems (particularly at this early phase were patterns are being set and would be continue further consolidating any anti-patterns that we carry-forward). 

`UiMenuItem` seems to still be very coupled to UI. As a suggestion, here is a slightly more fleshed out example for interfaces that game-core could provide:

```
interface AiDebugMenu {
   Map<PlayerType, AiPlayerDebugOption> getPerPlayerDebugOptions;
   Set<AiDebugOption> getAiDebugOptions;
}

interface AiPlayerDebugOption {    
    /** Used by game-core to set the menu item title */
    String getTitle();

    /** This is a callback into the AI code that should tell it to invoke the renderDelegate to do debug */
    void setEnabled(String playerName, boolean enabled);

    /** This will be called by game-core to provide a render delegate to the AI code so it can access drawing routines */
    void setRenderDelegate(DebugMapRenderer renderDelegate);

    /** This set to true converts us from checkbox to radioButton */
    boolean isMutuallyExclusive() ;

    /** This returning a non-null and/or a non-empty list makes the current object a parent of a checkbox or a radio button list */
    List<AiPlayerDebugOption> getSubCategoryDebugOptions();
}

/** interface for providing an API for AI to draw on the map */
interface DebugMapRenderer {
     void drawOnTerritory(String territoryName, String textValue);
     void colorTerritory(String territoryName, Color color, double transparency);
     void clear();
}

/** Interface for debug option that is generic to AI, EG: 'draw lancaster'
interface AiDebugOption {
    String getOptionTitle();
    void setEnabled(boolean enabled);
    void setRenderDelegate(DebugMapRenderer renderDelegate);
    /** This returning a non-empty list will make the current debug option a parent of a checkbox or a radio button list */
    List<AiDebugOption> getSubCategoryDebugOptions();
}
```

In above, for game-core to render game menus, it would take a collection of debug options and check:
- does this have children? If so, it's the parent of a sub-menu
- for all remaining, group them into two:
   - mutually exclusive (radio button group)
   - not mutually exclusive (check boxes)

Example implementation for  'DrawLancasterDebugOption':
```

@AllArgsConstructor
class DrawLancasterDebugOption {
   private final MapData mapData; // or collection of territories, the base data AI is using
   private boolean enabled;
   private DebugMapRenderer debugMapRenderer;

   void setRenderDelegate(DebugMapRenderer renderDelegate) {
      this.debugMapRenderer = debugMapRenderer;
   }

   boolean setEnabled(boolean enabled) {
        this.enabled = enabled;
        if(enabled) {
             drawDebugToMap();
         }
   }
   
   void onUpdate() {
         if(enabled) {
             drawDebugToMap();
         }
    }

    private void drawDebugToMap() {
           Map<String, Color> territoryShadings = ...;
          territoryShadings.entrySet.forEach(entry -> renderDelegate.shadeTerritory(entry.getKey(), entry.getValue());         
    }
}
```

Hopefully that gives a more useful and concrete example. Please follow up with comments & questions for further discussion.

DanVanAtta:
Re: adding `ww2_g40_balanced.xml`

The proliferation of test XMLs is a bad situation, particularly when they are so fully blown and more data than most any data set really needs. The tests that rely on their specific setups are intrinsically complex (for example, requires a maintainer to know that for example SZ4 contains 1 sub and SZ20 contains 2, and that on that map you have a given rule set specified, etc..). 

With that said, dumping in a test XML wholesale is not the first time we have committed that sin, so it's fine for now so long as we are getting unique data not already present and only do so in the future when strictly needed.

ai/src/main/java/org/triplea/ai/flowfield/odds/LanchesterDebugUiAction.java
DanVanAtta: > My hopes are to make this be able to calculate what the lanchenster attrition order is so that map makers could then add that to a file in their map. 

We need to discuss this. It's hell to support run-time dependencies that exist on hard disks. It is something we will forever have to support and potentially retrofit all maps to have such a file. A per-AI, per-map configuration IMO is a last resort, maybe a non-starter.

An example of this has been supporting save-game files, supporting network communication, and supporting variations of Map-XMLs, all have been challenges and each has resulted in bug reports and many hours of additional maintenance effort.

ai/src/main/java/org/triplea/ai/flowfield/map/FieldTerritory.java
DanVanAtta: This is one pitfall of lombok, I would suggest to document this in a test-case, by doing a quick test-case that sets up a minimal `FieldTerritory`, adds data to the list that would recurse and doing an `equals` call. In case someone were to try and be clever and remove this exclude, or somehow trigger it otherwise, we'd have the test case to catch it.

ai/src/main/java/org/triplea/ai/flowfield/FlowFieldAi.java
DanVanAtta: Wouldn't be bad to either `throw new UnsupportedException("TODO")` or add a comment to this method body that the implementation will be landing. I'd recommend creating a Project so that you can have a project number that you can prefix such comments and then find them and clean them up.

ai/build.gradle
DanVanAtta: I'm a bit concerned that we have a dependeny on test data from game-core. I'm interested to learn more about this.

Meanwhile, `testCompile` is legacy and will not be supported in a near future release of gradle. You should be able to swap this for `testImplementation`, does that cause issue doing so?

trevan:
> With that said, dumping in a test XML wholesale is not the first time we have committed that sin, so it's fine for now so long as we are **getting unique data not already present** and only do so in the future when strictly needed.

This file already exists in game-core which is where I copied it from.  So it isn't unique and it is already present.  But I couldn't figure out a way to access the existing file since it is in a different package.  Do you know of a way?

DanVanAtta:
@trevan AFAIK there is no way to bring in the test classpath from one sub-project into another.

trevan:
@DanVanAtta, I just pushed a change to how the debug menu is built.  Take a look at https://github.com/triplea-game/triplea/pull/8548/commits/eadf2190b900adac8f0b58314d82c62f2cfdc2d8.  I still went with an enum instead of `isMutuallyExclusive` since there is three possibilities: checkbox, radio button, and normal.

I also tried to hide the deselect requirement by having the `AiPlayerDebugAction` class remember what was added to the renderer and then undo'ing everything when deselect is called.  And the need to repaint the screen is also hidden from the caller inside of `AiPlayerDebugAction`.

trevan:
I found a way to use the test resource in game-core instead of copying it into ai.  See https://github.com/triplea-game/triplea/pull/8548/commits/1c37c6f91885b674ba23fd40dc1853056d068607 for what I did.

trevan:
@DanVanAtta , take a look at https://github.com/triplea-game/triplea/pull/8548/commits/9ae0f7167317f1cb6bfcc1d8547a2a9a7486836f.  I think that resolves most of your most recent comments.

ai/src/main/java/org/triplea/ai/flowfield/FlowFieldAi.java
trevan: @DanVanAtta , I just tried to create a Project and it doesn't look like I have permissions to do that.  So could you create the project?

DanVanAtta: https://github.com/triplea-game/triplea/projects/22

Let me know if you have any issues managing or adding items to the project

DanVanAtta:
The updates look good, much easier to understand now @trevan , thanks

Issue #8565, opened 2021-01-02
270bc_variants Delete (2020/01/02)


Cernelius:
I want to (and can) delete
https://github.com/triplea-maps/270bc_variants

I'm however asking permission to proceed, since this will be followed by a request to delete the game from download list (which I cannot pull myself).

---

The reason is that
https://github.com/triplea-maps/270bc_wars
is substantially the successor.

Also for this reason, I'm going to ask it to be upgraded to "Good" quality (which is where 270bc_variants is).

Cernelius:
Does the merging of
https://github.com/triplea-game/triplea/pull/8567
implies that I can now proceed with this deletion?

DanVanAtta:
Yes

Cernelius:
Done, thanks.

PR #8566, opened 2021-01-02
Update triplea_maps.yaml

Cernelius:
As per the game's version's first digit being "1", I consider this game complete.

Also, it is the successor of one of the 270bc_variants games, whose map is currently in "good" quality.

Reference:
https://github.com/triplea-game/triplea/issues/8565


DanVanAtta:
The map version number is not what dictates the category FWIW. The XML attribute is actually legacy now and will no longer be reader after 2.5

We don't have a process to really evaluate when a map is migrated up. Let's give it some time to let the map settle in.

Cernelius:
I thought this would have remained open util decided whether or not to move the game to the other section. After how long am I supposed to open an identical PR if I am? If so, why not letting this PR just open for that amount of time?

PR #8567, opened 2021-01-02
Update triplea_maps.yaml

Cernelius:
Merge only after
https://github.com/triplea-maps/270bc_variants
is deleted.

Reference:
https://github.com/triplea-game/triplea/issues/8565


DanVanAtta:
The index should be updated first, otherwise a person can download a game that does not exist. Furthermore we do build verification checks that links are valid, those would fail until the repo were deleted.

PR #8572, opened 2021-01-03
Use Streams instead of Enumeration

RoiEXLab:
I felt like I should code something for TripleA again :)
Haven't tested my changes so far, but will do tomorrow

### DIFF
--- a/java-extras/src/main/java/org/triplea/io/ZipFileUtil.java
+++ b/java-extras/src/main/java/org/triplea/io/ZipFileUtil.java
@@ -6,9 +6,9 @@ import java.net.URI;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


DanVanAtta:
Good to see you back @RoiEXLab

RoiEXLab:
@DanVanAtta Ok it refined my changes and finally got to test them, everything seems to work.
Not sure about the integration test though: I don't think the failure is related to my code changes. Any ideas?

DanVanAtta:
Yeah, the integ test has been an ongoing issue. Seems to have broken a week or two ago. It needs fixing. Should be not be related (it's a bad situation, we won't know if we make things worse, so it is kinda pressing to fix it).

DanVanAtta:
Nice update with the FileSystem API, that simplified quite a bit :+1:

PR #8576, opened 2021-01-04
Convert map download version to be an integer rather than a 'Version'

DanVanAtta:
The map download version is just a single number, this update converts
the data type that we are parse it into from a 'Version' object to
instead be an 'Integer' object.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadFileDescription.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadFileDescription.java
@@ -24,7 +24,7 @@ public final class DownloadFileDescription {
   @EqualsAndHashCode.Include private final String url;
   private final String description;
   private final String mapName;
-  private final Version version;
+  private final Integer version;
   private final DownloadType downloadType;
   private final MapCategory mapCategory;
   private final String img;
@@ -69,7 +69,7 @@ public final class DownloadFileDescription {
         .url(mapDownloadListing.getUrl())
         .description(mapDownloadListing.getDescription())
         .mapName(mapDownloadListing.getMapName())
-        .version(new Version(mapDownloadListing.getVersion()))
+        .version(new Version(mapDownloadListing.getVersion()).getMajor())
         .downloadType(DownloadType.MAP)
         .mapCategory(MapCategory.fromString(mapDownloadListing.getMapCategory()))
         .img(mapDownloadListing.getPreviewImage())
@@ -82,7 +82,7 @@ public final class DownloadFileDescription {
         .url(skin.getUrl())
         .description(skin.getDescription())
         .mapName(skin.getSkinName())
-        .version(new Version(skin.getVersion()))
+        .version(new Version(skin.getVersion()).getMajor())
         .downloadType(DownloadType.MAP_SKIN)
         .mapCategory(MapCategory.fromString(mapCategory))
         .img(skin.getPreviewImageUrl())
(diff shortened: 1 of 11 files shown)


game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadFileDescription.java
```diff
@@ -69,7 +69,7 @@ public static DownloadFileDescription ofMapDownloadListing(
         .url(mapDownloadListing.getUrl())
         .description(mapDownloadListing.getDescription())
         .mapName(mapDownloadListing.getMapName())
-        .version(new Version(mapDownloadListing.getVersion()))
+        .version(new Version(mapDownloadListing.getVersion()).getMajor())
```
RoiEXLab, line 72: What format does the Version String have here to justify the extra step via the version object?

game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadFileDescription.java
```diff
@@ -82,7 +82,7 @@ public static DownloadFileDescription ofMapSkinListing(
         .url(skin.getUrl())
         .description(skin.getDescription())
         .mapName(skin.getSkinName())
-        .version(new Version(skin.getVersion()))
+        .version(new Version(skin.getVersion()).getMajor())
```
RoiEXLab, line 85: Isn't then skin version the same as any map version?

game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadFileProperties.java
```diff
@@ -51,7 +51,9 @@ private static File fromZip(final File zipFile) {
     return new File(zipFile.getAbsolutePath() + ".properties");
   }
 
-  public Optional<Version> getVersion() {
-    return Optional.ofNullable(props.getProperty(VERSION_PROPERTY)).map(Version::new);
+  public Optional<Integer> getVersion() {
+    return Optional.ofNullable(props.getProperty(VERSION_PROPERTY))
```
RoiEXLab, line 55: Doesn't `props.getProperty(VERSION_PROPERTY)` return an integer now?
Or am I just missing something?

RoiEXLab (approved):
Looks good, just a couple of questions

game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadFileDescription.java
DanVanAtta: Not as far as I know, skin versioning is independent.
It should become a moot point as the plan is to merge skins into their skinned maps.

game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadFileDescription.java
DanVanAtta: It is a simple number, this could be converted to just be an integer here as well. One ramification though is the `mapDownloadListing#getVersion` would then return a number which then has further changes to code that I think would be deleted soon.

I think we will be able to simplify this soon enough, but holding off for now as upcoming changes could make it an even easier update.

game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadFileProperties.java
DanVanAtta: It is still a string. It is actually required to store it as a string, storing an Integer throws an IllegalArgumentException.

PR #8577, opened 2021-01-04
Fix test data that is meant to be in the future

DanVanAtta:
Now that it is 2021, test data that assumed Jan-01-2021 is
in the future, is no longer true and tests fails!

This update sets the test data to be at 2200, which will remain
in the future for quite some time yet.


DanVanAtta:
cc: @RoiEXLab 

Turns out the test failure was relatively simple and recent!

RoiEXLab:
The future is now 🤯

PR #8602, opened 2021-01-08
Initial work on how units target other units

trevan:
Currently, a battle assumes that it always has 4 firing phases: AA,
Bombardment, First Strike, and General/Other. At the start of each of
the phases, it uses one of the "FiringGroupSplitter" classes to group
the firing units and their targets into "FiringGroups". It then loops
through all of the "FiringGroups" and creates a set of "RollDiceStep",
"SelectCasualties", and "MarkCasualties" battle steps.

The new logic is to create a list of "BattleGroups" at either the very
beginning of a battle or at the beginning of the BattleDelegate. The
list of BattleGroups is constant for all battles and so it only needs to
be built once.

Each BattleGroup has 1 or more FiringSquadrons. These FiringSquadrons
have predicates that are used to determine what units are firing in it,
what units are targeted in it, and if it should even fire.

In a battle, each BattleGroup will be iterated over and the
active FiringSquadrons will be determined. Then, a set of  battle steps
will be created for each of the FiringSquadrons and the game will
proceed.

This commit only has the work to create the BattleGroups and
FiringSquadrons. They are not being used yet.

This rework will almost replace everything in the
games.strategy.triplea.delegate.battle.steps.fire package.

And end goal is to allow map makers to create their own BattleGroups and
FiringSquadrons. If the map maker creates their own, then the
BattleGroups.createBattleGroups would not be called. It should only be
called for maps that don't have custom BattleGroups/FiringSquadrons.

I can't delete the classes that this replace because they are saved in the game save.  I'll be marking them as deprecated and tagging them to be removed once I've got these new classes being used.

This will also allow grouping together first strike units and non first strike units if the first strike units don't have a sneak attack (an isDestroyer is present).


trevan:
The compatibility test is failing because the 250bc_variants maps have been deleted.  So it can't download that map.

trevan:
The logic that are in these new classes is replacing the logic that can be found in FiringGroupSplitterAA, FiringGroupSplitterBombard, FiringGroupSplitterGeneral, ClearFirstStrikeCasualties, OffensiveFirstStrike, DefensiveFirstStrike, and TargetGroup.  All of those will be marked as deprecated once this is all done.

DanVanAtta:
At first read, sounds like a good simplification.

Re:
>  And end goal is to allow map makers to create their own BattleGroups and
FiringSquadrons. If the map maker creates their own, then the
BattleGroups.createBattleGroups would not be called. It should only be
called for maps that don't have custom BattleGroups/FiringSquadrons.

If map makers are going to customize this, I would like for it to be very carefully designed first. I would certainly like to understand which maps would use custom groups and how that would be defined. I would also want to be sure that it plays nicely with future XML formats so we don't get in situation where we are supporting high number of XML permutations.

--------

Otherwise, I hope to get a chance to review this in the upcoming days. At 1000 lines it will probably take 2 or 3 sessions to review.

trevan:
> If map makers are going to customize this, I would like for it to be very carefully designed first. I would certainly like to understand which maps would use custom groups and how that would be defined. I would also want to be sure that it plays nicely with future XML formats so we don't get in situation where we are supporting high number of XML permutations.

I'm envisioning that whatever is in the XML, they will be converted to the BattleGroups/FiringSquadrons at the beginning of the BattleDelegate.  The BattleGroups/FiringSquadrons are basically a bunch of Predicates.  So, the idea is that the XML would define conditions on what units are firing and who they target.  Then these conditions are transformed into Predicates that the BattleGroups/FiringSquadrons use.

Here's an example of what a Global40 battlegroup/firingsquadrons would look like in a rough xml:
```
<battlegroup name="AA" returnFire="false" type="AA">
  <firingSquadron side="both" round="1">
    <units type="aaGun" />
    <targets type="fighter:tactical_bomber:bomber" />"
  </firingSquadron>
</battlegroup>
<battlegroup name="bombardment" returnFire="true" type="bombardment">
  <firingSquadron name="bombardment" territory="land" side="offense" round="1">
    <units type="cruiser:battleship" />
    <targets type="infantry:artillery:mech_infantry:armour:fighter:tactical_bomber:bomber" />
  </firingSquadron>
</battlegroup>
<battlegroup name="First Strike" returnFire="false">
  <firingSquadron name="submarine sneak attack" notexists="destroyer" side="both">
    <units type="submarine" />
    <targets type="transport:submarine:destroyer:cruiser:carrier:battleship" />
  </firingSquadron>
</battlegroup>
<battlegroup name="General">
  <firingSquadron name="submarine" exists="destroyer" side="both">
    <units type="submarine" />
    <targets type="transport:submarine:destroyer:cruiser:carrier:battleship" />
  </firingSquadron>
  <firingSquadron name="air" side="both">
    <units type="fighter:tactical_bomber:bomber" />
    <targets type="infantry:artillery:mech_infantry:armour:fighter:tactical_bomber:bomber:transport:destroyer:cruiser:carrier:battleship" />
    <targets type="submarine" exists="destroyer"/>
  </firingSquadron>
  <firingSquadron name="land" side="both">
    <units type="infantry:artillery:mech_infantry:armour" />
    <targets type="infantry:artillery:mech_infantry:armour:fighter:tactical_bomber:bomber" />
  </firingSquadron>
  <firingSquadron name="sea" side="both">
    <units type="destroyer:cruiser:carrier:battleship" />
    <targets type="fighter:tactical_bomber:bomber:transport:submarine:destroyer:cruiser:carrier:battleship" />
  </firingSquadron>
</battlegroup>
```

And here is what world_war_ii_revised would look like:
```
<battlegroup name="AA" returnFire="false" type="AA">
  <firingSquadron side="both" round="1">
    <units type="aaGun" />
    <targets type="fighter:bomber" />"
  </firingSquadron>
</battlegroup>
<battlegroup name="bombardment" returnFire="true" type="bombardment">
  <firingSquadron name="bombardment" territory="land" side="offense" round="1">
    <units type="battleship" />
    <targets type="infantry:artillery:armour:fighter:bomber" />
  </firingSquadron>
</battlegroup>
<battlegroup name="First Strike">
  <returnFire side="both">
    <exists type="destroyer" />
  </returnFire>
  <firingSquadron name="submarine" side="both">
    <units type="submarine" />
    <targets type="transport:submarine:destroyer:carrier:battleship" />
  </firingSquadron>
</battlegroup>
<battlegroup name="General">
  <firingSquadron name="air" side="both">
    <units type="fighter:bomber" />
    <targets type="infantry:artillery:armour:fighter:bomber:transport:destroyer:carrier:battleship" />
    <targets type="submarine" exists="destroyer"/>
  </firingSquadron>
  <firingSquadron name="land" side="both">
    <units type="infantry:artillery:armour" />
    <targets type="infantry:artillery:armour:fighter:bomber" />
  </firingSquadron>
  <firingSquadron name="sea" side="both">
    <units type="destroyer:carrier:battleship" />
    <targets type="fighter:bomber:transport:submarine:destroyer:carrier:battleship" />
  </firingSquadron>
</battlegroup>
</game>
```

This would basically deprecate `canBombard`, `isAAforCombatOnly`, `maxRoundsAA`, `isFirstStrike`, `canNotTarget`, `canNotBeTargetedBy`, `Defending Subs Sneak Attack`, and probably some other things.

DanVanAtta:
Are you planning to purely retrofit maps with that XML to make them explicit @trevan ? What is the goal? I think this meta conversation might be beyond the scope of this PR and perhaps should include map-makers. Which maps would actually be using this?

While various properties would be deprecated, we would still need to translate existing maps (without modifications) to the new variants. Seemingly that could be done with or without the new XML.

trevan:
> Are you planning to purely retrofit maps with that XML to make them explicit @trevan ? What is the goal? I think this meta conversation might be beyond the scope of this PR and perhaps should include map-makers. Which maps would actually be using this?
> 
> While various properties would be deprecated, we would still need to translate existing maps (without modifications) to the new variants. Seemingly that could be done with or without the new XML.

I think some of the major maps should probably be retrofitted.  I'd be up for doing that and that could help figure out how the XML should be structured.  I do agree that it should include map-makers.  I had created a forum topics (https://forums.triplea-game.org/topic/2481/custom-battle-phases and https://forums.triplea-game.org/topic/2348/custom-firing-groups-for-units) a while back where I first started presenting my ideas.

As for translating, that is what the `BattleGroups.createBattleGroups` function does and the majority of the code in this PR.  It is building BattleGroups using the existing properties.  The idea is that the engine can be updated to use the BattleGroups/FiringSquadrons without any changes to XML.  Then an XML can be designed that can be mapped to predicates in the BattleGroups/FiringSquadrons.

DanVanAtta:
Okay, let's keep the conversation consolidated and please mention if you are ready to make more progress. I don't think we have designed it sufficiently for fully customized battle phases (it is a major change and has long-term support implications)

trevan:
> please mention if you are ready to make more progress

What do you mean by this?  This PR is ready for review.  I'll have other PRs after this to start using the BattleGroups instead of the old code.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/NormalBattleGroups.java
```diff
@@ -0,0 +1,159 @@
...
+  /**
+   * Creates battle groups for the first strike phase
+   *
+   * <p>If an opposing isDestroyer unit is present, then the phase for that side will not occur (see
+   * {@link #noDestroyerPresent} for both firing squadrons)
+   */
+  static BattleGroup createFirstStrike(
+      final Collection<UnitType> unitTypes, final GameProperties properties) {
+
+    final List<FiringSquadron> firingSquadrons =
```
DanVanAtta, line 39: nit, var name perhaps could be more descriptive, this is the 'first strike' firing squadron?

Is this also the right way to group the units? I wonder if we should instead be going by unit-type to assign them to a phase rather than doing it per-phase. The latter can be complicated as we negate units and then add them later, which leads to logic seemingly in multiple places (ie: if this predicate is false here, it needs to be true in another place).

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/NormalBattleGroups.java
```diff
@@ -0,0 +1,159 @@
...
+                })
+            .collect(Collectors.toList());
+
+    final BattleGroup.BattleGroupBuilder battleGroupBuilder =
+        BattleGroup.builder()
+            .offenseSquadrons(firingSquadrons)
+            .fireType(BattleState.FireType.NORMAL)
+            .casualtiesOnDefenseReturnFire(false);
+
+    if (Properties.getDefendingSubsSneakAttack(properties)) {
```
DanVanAtta, line 56: Can this be inlined with ternaries or defined before hand as variables so that the intermediate builder variable is not needed.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/NormalBattleGroups.java
```diff
@@ -0,0 +1,159 @@
...
+                  return createFiringSquadron(
+                          firingSquadron, unitIsFirstStrike(properties), noDestroyerPresent())
+                      .name(FIRST_STRIKE_UNITS)
+                      .build();
+                })
+            .collect(Collectors.toList());
+
+    final BattleGroup.BattleGroupBuilder battleGroupBuilder =
+        BattleGroup.builder()
+            .offenseSquadrons(firingSquadrons)
```
DanVanAtta, line 52: Could you help clarify why the same firingSquadrons can be both defense and offense?

Noting:
```
 .offenseSquadrons(firingSquadrons)
.defenseSquadrons(firingSquadrons)
```

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/NormalBattleGroups.java
```diff
@@ -0,0 +1,159 @@
...
+                })
+            .collect(Collectors.toList());
+
+    final List<FiringSquadron> offenseSquadrons =
+        Stream.of(firingSquadronsWithFirstStrikeUnits, firingSquadronsWithoutFirstStrikeUnits)
+            .flatMap(Collection::stream)
+            .collect(Collectors.toList());
+
+    final List<FiringSquadron> defenseSquadrons;
+    if (Properties.getDefendingSubsSneakAttack(properties)) {
```
DanVanAtta, line 138: nit, good place for a ternary

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/NormalBattleGroups.java
```diff
@@ -0,0 +1,159 @@
...
+      defenseSquadrons = offenseSquadrons;
+    } else {
+      defenseSquadrons =
+          FiringSquadron.createWithTargetInformation(unitTypes).stream()
+              .map(
+                  firingSquadron -> {
```
DanVanAtta, line 144: nit, this could be written with the simplified lambda syntax?
I wonder as well if more of this logic couldn't be rolled into `FiringSquadron`. There almost seems to be a feature envy code-smell here. Notably as well I would take a look at the `createFiringSquadron` that accepts predicates. I am starting to think that any predicate parameter is an outright code-smell (like a boolean parameter).

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/SuicideAndNonSuicideUnitTypes.java
```diff
@@ -0,0 +1,122 @@
...
+                    .and(
+                        firingUnitFilterData ->
+                            suicide.contains(firingUnitFilterData.getFiringUnit().getType())))
+            .build());
+  }
+
+  private Collection<FiringSquadron> groupMultipleSuicide(final FiringSquadron firingSquadron) {
+    final Collection<FiringSquadron> firingSquadrons =
+        suicide.stream()
+            .map(
```
DanVanAtta, line 95: This gets deeply nested. At a glance, this looks to be pretty dense. Would you mind taking a pass over this and similar blocks and trying to make it all easier and faster to understand?

DanVanAtta (commented):
Initial first pass, left some comments, there is quite a lot left to review.

DanVanAtta:
>>     please mention if you are ready to make more progress

> What do you mean by this? This PR is ready for review. I'll have other PRs after this to start using the BattleGroups instead of the old code.

My understanding the current direction is simplification and then a next step would be to get these same value & properties from XML. I do not know if that latter part is a good idea, I do think it needs to be really thought through. It will take time to reach that consensus. Anything that is added to maps is very consequential as we lose control over it and must support it indefinitely.

On another aspect, a discussion post that has grown old is not necessarily authoritative, so I don't want to see a PR implementing something that did not achieve full consensus, but you could very well say "see, this is an old discussion and there has been time for feedback." On the other hand, I do agree that we need to be able to make those discussion be timely as it's not conducive to getting things done if all design discussions drag out.

trevan:
> My understanding the current direction is simplification and then a next step would be to get these same value & properties from XML.

I think the "end-goal" statement might have thrown you off.  This PR (and subsequent PRs) is currently for simplification of the logic of the engine and to fix the issue that you presented (where subs never roll with other units even though they should be able to).  Another reason for this change is to make it easier for my AI to understand how battles are fought (to know if a sub has first strike abilities requires a bit of logic right now).  I would like, at some point, to be able to allow map makers to define the battle groups, etc so that the map xml can be simplified and easier to read.  But that isn't in the near-term plans because the XML definition still needs to be hashed out.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/FiringSquadron.java
```diff
@@ -0,0 +1,241 @@
...
+  @Builder.Default
+  Predicate<Collection<Unit>> friendlyUnitRequirements = FiringSquadron::defaultUnitsRequirements;
+  /** This FiringSquadron can only fire if the enemy unit requirements are met */
+  @Builder.Default
+  Predicate<Collection<Unit>> enemyUnitRequirements = FiringSquadron::defaultUnitsRequirements;
+  /** This FiringSquadron can only fire if the battle state requirements are met */
+  @Builder.Default
+  Predicate<BattleState> battleStateRequirements = FiringSquadron::defaultBattleStateRequirements;
+
+  /** ErrorProne static analysis doesn't like lambda style @Builder.Default values */
```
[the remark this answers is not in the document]
trevan: I figured out a way to inline it by overriding the default lombok builder.  I did this in https://github.com/triplea-game/triplea/pull/8602/commits/bb56e98c44c6efaec443b627729f2627d646961f.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/FiringSquadron.java
```diff
@@ -0,0 +1,241 @@
...
+  @SuppressWarnings("unused")
+  private static boolean defaultBattleStateRequirements(final BattleState unused) {
+    return true;
+  }
+
+  Predicate<FiringUnitFilterData> getFiringUnits() {
+    // ensure that transported units can not fire
+    return firingUnits.and(
+        firingUnitFilterData ->
+            Matches.unitIsBeingTransported().negate().test(firingUnitFilterData.getFiringUnit()));
```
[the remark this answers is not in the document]
trevan: See https://github.com/triplea-game/triplea/pull/8602/commits/bb56e98c44c6efaec443b627729f2627d646961f

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/NormalBattleGroups.java
trevan: See https://github.com/triplea-game/triplea/pull/8602/commits/bb56e98c44c6efaec443b627729f2627d646961f

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/NormalBattleGroups.java
trevan: See https://github.com/triplea-game/triplea/pull/8602/commits/bb56e98c44c6efaec443b627729f2627d646961f

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/NormalBattleGroups.java
trevan: See https://github.com/triplea-game/triplea/pull/8602/commits/bb56e98c44c6efaec443b627729f2627d646961f

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/SuicideAndNonSuicideUnitTypes.java
trevan: It has been removed in https://github.com/triplea-game/triplea/pull/8602/commits/bb56e98c44c6efaec443b627729f2627d646961f

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/NormalBattleGroups.java
trevan: I think your comment in the forum is similar to this idea you have here.  And it sounds fine except that the `FiringSquadron` needs to be able to handle existing logic as well.  So, the existing logic would have the more complicated logic while what ever is defined through the XML would have a much simpler logic.  But both would be expressed internally through the `FiringSquadron`.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/FiringSquadron.java
```diff
@@ -0,0 +1,241 @@
...
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Value;
+import org.triplea.java.PredicateBuilder;
+
+/** Holds predicates for finding units in a squadron and what they target */
```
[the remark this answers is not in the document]
DanVanAtta: A 'spec' sounds like a config rather than an collection of units. I would suggest these two options:
1. Change the name of 'FiringGroup', do that in another PR and then you'd be free to use that name here. Perhaps we can name 'FiringGroup' to be a bit better, as the 'FiringGroup' is more of a configuration, or more of a computation that says these units can hit these other units, but it is not actually a group of units per say.
2. Document 'FiringSquadron', it is possible we can bake that into the terminology.

Overall this is a naming & vocabulary exercise, not easy, we all know code is much easier to understand when the vocabulary is good and descriptive.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/FiringSquadron.java
```diff
@@ -0,0 +1,241 @@
...
+@Value
```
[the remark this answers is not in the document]
DanVanAtta: Indeed, though the lack of setter or even using `@Immutable` can do so was well. Personally I am less concerned that objects are indicated as immutable rather that they _are_ immutable.

In part, what I'm getting at is that `@Value` is typically used for value objects, which are like data objects but also immutable. So I would expected something like the following to be a value object:
```
@Value
class value {
   String name;
   String address;
}
```

Any methods I would expect to be operations on that data (different views of the data), and generally not to receive inputs.

So in sum, `@Value` indicates to a reader that what follows is an immutable data object, not necessarily just an immutable object (for example and in contrast, a Util object is also immutable).


Second, it's a question whether `toString` and `equals` have really well defined meanings here. Is this class meant to be used for comparison and do we really need equality? If we are only needing just one or two lombok annotations, I would use just those two.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/FiringSquadron.java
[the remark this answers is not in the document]
DanVanAtta: Cool, though sorry in advance here for the churn.

I don't think it's that easy to understand this class. In part because we have a number of predicates, most of them are by default tautologies, it takes a good bit of work to know what is used, when, where and how. I think this class does not do well on the cohesion metrics where most variables are used in most methods.

What do you think of this recommendation: to split the firing squadron data from how it is created. Already there are classes I see here to create firing squadrons, they have some of the information of what is needed while this class has other parts. Having a class that is fully parameterized with generic predicates I suspect is part of the problem and makes this overly abstract (arguably a dry violation as the knowledge of how to create a firing group is split between two class, no one class is independent of the predicates from either using them or defining them).

I think it should be a goal that in this case any predicate we have should be fully encapsulated or activated via a configuration switch, but the class that uses a predicate probably should be the one that defines it. Passing in predicates I would now view as a code smell (unless we are doing a visitor pattern).

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/FiringSquadron.java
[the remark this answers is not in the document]
DanVanAtta: Thanks, I was not intending to suggest that method be added to `Matches`.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/NormalBattleGroups.java
DanVanAtta: The question is whether which of the following two make more sense:
- group units into squadrons by phase
- group units into squadron by unit type

In other words, do we create squadrons first by looking at our unit types and then knowing which phases we would have, or do we start by knowing the full set of phases and then adding units when we get to the logic for each phase.

I do see here an issue where it is assumed that based on some conditions present in this class, that there are related conditions in another class. For example, if a unit is not included here, then we would presume they would be looped in at another phase. This seemingly does create a situation where we could have units not getting assigned to any phase just due to complexity and a logic bug.  I think this can be helped if we consolidate the logic for assigning units to phases, then we won't have a situation where we have different bits of code in different places doing essentially the same thing (DRY).

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/NormalBattleGroups.java
DanVanAtta: Would be nice to see the commits split out a bit more, it's somewhat tough to go through a large diff to look for just a few lines that have changed.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/NormalBattleGroups.java
DanVanAtta: Okay, that helps. I was confused by the terminology, it is actually a configuration or a spec. Are we computing this per-battle, or once for any map?

It could be more clear to have a single variable to hold the valid sides. EG:
```
isValidFor(OFFENSE, DEFENSE)
```

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/BattleGroup.java
```diff
@@ -0,0 +1,107 @@
...
+  private static BattleGroup createGeneralBattleGroups(
+      final Collection<UnitType> unitTypes, final GameProperties properties) {
+    if (Properties.getWW2V2(properties)) {
+      return NormalWW2V2BattleGroups.createGeneral(unitTypes, properties);
+    } else {
+      return NormalBattleGroups.createGeneral(unitTypes, properties);
+    }
+  }
+
+  /**
```
DanVanAtta, line 71: Let's have another look at what this class owns and how it is constructed. It is a code smell if a class can be constructed in too many ways. The presence of predicates as part of configuration I think is core to the problem that makes a class too generic and violates DRY and encapsulation (another class constructing this class knows how the predicates are used, it knows what internal data is filtered with it, it knows about the internal data of the class in order to do that, and the filtering logic is spread between the class that constructed this class and this class).

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/BombardmentBattleGroups.java
```diff
@@ -0,0 +1,63 @@
...
+import games.strategy.triplea.Properties;
+import games.strategy.triplea.delegate.Matches;
+import games.strategy.triplea.delegate.battle.BattleState;
+import java.util.List;
+import java.util.function.Predicate;
+import lombok.experimental.UtilityClass;
+
+/** Constructs battle groups for the bombardment phase of a battle */
+@UtilityClass
+class BombardmentBattleGroups {
```
DanVanAtta, line 15: As an idea, instead of one utility class per unit ability, perhaps we can create an interface to capture the key characteristics of a unit ability. Then battle group would be more generic and would only need a set of unit ability interfaces that it could then use to do any calculation.

DanVanAtta (commented):
A number more comments for you to consider @trevan , largely around class design.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/battlegroup/FiringSquadron.java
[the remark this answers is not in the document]
trevan: We can't change the name of `FiringGroup` because it is part of saves.  Otherwise, I would have renamed it.  And the existing `FiringGroup` is a group of units.

trevan:
@DanVanAtta , I'm worried that a lot of your reviews are based around a misunderstanding of how firing groups/firing squadrons/etc are computed.  You seem to want to be able to something that isn't possible with the current map configuration, unit properties, etc.  The logic is complicated, it is scattered over a lot of places, and it isn't as simple as "unit abilites" or things like that.  The predicates are needed to handle all the odd cases that exist in the current system.

DanVanAtta:
I'm pretty sure I don't understand this properly. If a reviewer does not understand it properly, other people will not, and you yourself will be in that same pool when you lose context into this code after a few weeks or a few months. That indicates things are too complicated. Generally my impression here is that this update is complex. While intrinsically the underlying code is complex, I'm not sure if we have moved far enough in the right direction. That is not saying this is a wrong direction, but that more needs to be done to get us further to where we would want to be.

Overall, a first step I think would be to examine if we can encapsulate more logic and that probably starts by examining the predicates. We might need to change the design and might need some preceding changes to get us to where we want to be.

DanVanAtta:
Another core suggestion is if we can make `FiringSquardron` more of an object rather than a set of data with generic predicates where a lot of the heavy lift is done with static factory classes to construct the `FiringSquardon`. To do this, any input into a factory class would become an input to the `FiringSquardon` and would become the internal data of the `FiringSquardron`. Then based on the right method call, `FiringSquardon` would be able to select the proper predicate (that only `FiringSquardon` would know about, it woudl be an implementation detail).

DanVanAtta:
I do admit I don't fully understand the update here, nor all the details, so this is an open discussion. I've spent 40 minutes with this code, it's 1000 lines, that's not nearly enough to fully understand the details. I do rely on you to provide the additional context to help flesh out the ideas and see if we can get into a better place. On the other hand after reviewing this for that time, I am still not clear on everything that is going on, which does bode poorly for how maintainable and understandable the end result would be. This is an exercise to see if we can get this code to a better place.

trevan:
> Another core suggestion is if we can make `FiringSquardron` more of an object rather than a set of data with generic predicates where a lot of the heavy lift is done with static factory classes to construct the `FiringSquardon`. To do this, any input into a factory class would become an input to the `FiringSquardon` and would become the internal data of the `FiringSquardron`. Then based on the right method call, `FiringSquardon` would be able to select the proper predicate (that only `FiringSquardon` would know about, it woudl be an implementation detail).

I don't understand how that is any different from what this change does.  The static factory classes are constructing the `FiringSquadrron` and are passing in the "internal data".  Then the `FiringSquadron` just uses the predicate that was passed into it.  If you want me to wrap the predicates in other classes, sure, but in the end, it is still predicates and it just is more abstracted and harder to read.

DanVanAtta:
To try and clarify, for example, AABattleGroups takes in data:
```
      final UnitTypeList unitTypeList, final TechnologyFrontier technologyFrontier) {
```

This would be the data for BattleGroups proper. The existing internal data of BattleGroups would then all be computed at the method level. There might be some API changes to pass in a flag to ask for the right kind of data.

Looking at the branch, this is super difficult to know how this is going to be used exactly as BattleGroup is only used by test. I think part of the issue for me is that BattleGroup is just a number of getters, but those getters return functions which seems to me to be operations that BattleGroup could own. It's hard to say without seeing the actual usages. Do you have a branch @trevan where this is more formed and those usages could be observed?

In sum, as-is, it looks like battle group could own more logic and that would be done by avoiding the getters. Instead any usage of a predicate would be baked in, and instead of having battle groups return a predicate to help compute data, it would take on more of that computation.

EG:
```
someStream()
  .filter(battleGroup.predicate())
  .forEach(doOtherStuff);


converted to:->

battleGroups.getSomeData()
   .forEach(doOtherStuff());


Or even:

battleGroups.doOtherStuff(SOME_FLAG);
```

DanVanAtta:
@trevan if possible, please send a link to any branches that use this further. It'll be much easier to understand based on usage and the feedback will be more useful.

Another thought that occurred, perhaps `FiringSquadron` could just be an interface. Then instead of factory classes, we would have implementations. I think defining the right API on `FiringSquardron` is extremely pivotal.

trevan:
No, I don't have a branch with further usage of this yet.  But based on the conversations on the forum, I think redoing this work as unit abilities might be possible.  I'll have to see if I can convert all of the existing logic into unit abilities and then see how that works.  So, I'm closing this for now.

PR #8605, opened 2021-01-09
Add a way to clear the execution stack

trevan:
When a battle is over, the execution stack can be cleared. The remaining
steps don't need to run.

## Testing

Added a breakpoint to battle steps and ensured that it was not executing them.  Also ensured that battles still worked.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/CheckGeneralBattleEnd.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/CheckGeneralBattleEnd.java
@@ -54,12 +54,15 @@ public class CheckGeneralBattleEnd implements BattleStep {
   public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
     if (hasSideLost(OFFENSE)) {
       battleActions.endBattle(IBattle.WhoWon.DEFENDER, bridge);
+      stack.clear();
 
     } else if (hasSideLost(DEFENSE)) {
       battleActions.endBattle(IBattle.WhoWon.ATTACKER, bridge);
+      stack.clear();
 
     } else if (isStalemate() && !canAttackerRetreatInStalemate()) {
       battleActions.endBattle(IBattle.WhoWon.DRAW, bridge);
+      stack.clear();
     }
   }
 
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
@@ -87,11 +87,11 @@ public class OffensiveGeneralRetreat implements BattleStep {
 
   @Override
   public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
-    retreatUnits(bridge);
+    retreatUnits(stack, bridge);
   }
 
   @RemoveOnNextMajorRelease("This doesn't need to be public in the next major release")
-  public void retreatUnits(final IDelegateBridge bridge) {
+  public void retreatUnits(final ExecutionStack stack, final IDelegateBridge bridge) {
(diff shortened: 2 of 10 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/CheckGeneralBattleEnd.java
```diff
@@ -54,12 +54,15 @@ public Order getOrder() {
   public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
     if (hasSideLost(OFFENSE)) {
       battleActions.endBattle(IBattle.WhoWon.DEFENDER, bridge);
+      stack.clear();
```
DanVanAtta, line 57: This class seems misleading in name as the 'general' battle phase is just one phase and perhaps might not be the last phase of combat (it just so happens to be for now). Would this be clarified if this class were simply renamed to `CheckBattleEnd`?


Side-note - At some point too I think we should think about having a proper `Battle` object with a defined API. I'm not suggesting we go back to having a `MustFightBattle` that has everything in it again, but we went from that to its components pieces being broken out, I think re-combining them for a class to use those components pieces could be appropriate. In this way we'd have a lot of package protected strategy objects with very few entry points, the interactions would be mediated by more or less a single object. In other words, instead of many parts of the system having access to battle state and way to change battle state, the ways battle state would be changed would be mediated by more or less a single object.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
```diff
@@ -166,6 +169,7 @@ private void retreat(
 
     if (battleState.filterUnits(ALIVE, OFFENSE).isEmpty()) {
```
DanVanAtta, line 170: It is starting to look like battleActions should own the stack clear logic. That would guarantee that we have not missed any places and could be less wiring code. Instead of wiring both the battle actions and an execution stack, we'd only need the battle actions. Can battle actions be given a reference to the execution stack? It almost seems like they should be combined into the same object.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/CheckGeneralBattleEnd.java
trevan: Renaming isn't possible until we break save compatibility.

I see `BattleState` as the object with a defined API.  All of the component pieces are the strategy objects that interact with the `BattleState` and perform the logic needed.  I hope to start converting the other battle classes (`NonFightingBattle`, `AirBattle`, `DependentBattle`, `StrategicBombingRaidBattle`, `FinishedBattle`) to use `BattleState` and these strategy objects.

As for combining them all in one class, it isn't possible because each of the pieces have to be a unique `IExecutable`. And defining all of the `IExecutables` in one class makes it very hard to make changes since the order of the definition of the `IExecutable` is now important for save compatibility.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveGeneralRetreat.java
trevan: battle actions is just `MightFightBattle` which does have a reference to an execution stack but I'm not sure if it is "the" execution stack during the actual running.  So, I'd rather keep passing around the execution stack so that it can be save compatible.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/change/CheckGeneralBattleEnd.java
DanVanAtta: One class is not necessarily the objective, it's more to reduce the 'fan-in'. Let me know if that makes more sense.

PR #8606, opened 2021-01-09
Put the allied air transportedBy code in the same place

trevan:
Allied air was getting the transportedBy flag in MustFightBattle but
having it cleared in TransportTracker.  This moves the code in
MustFightBattle to TransportTracker so that it is in the same place.

Also, creates a value class for whenCombatDamaged to replace the Tuple
of Tuples.

## Testing

Played a game with allied air on a carrier and checked that it was marked as transported still.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/TransportTracker.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/TransportTracker.java
@@ -6,12 +6,14 @@ import games.strategy.engine.data.CompositeChange;
 import games.strategy.engine.data.GameData;
 import games.strategy.engine.data.GamePlayer;
 import games.strategy.engine.data.GameState;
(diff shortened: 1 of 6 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/TransportTracker.java
```diff
@@ -299,15 +302,47 @@ public static boolean isTransportLoadRestrictedAfterCombat(final Unit transport)
         && transport.getWasInCombat();
   }
 
+  public static AlliedAirTransportChange markTransportedByForAlliedAirOnCarrier(
```
trevan, line 305: This is just moved from the MustFightBattle.  The logic hasn't changed.

DanVanAtta (approved):

PR #8610, opened 2021-01-09
Add a RemoveNonCombatantsBeforeAA step at the beginning of the battle

trevan:
This step removes all units that can't fire or be fired at during the
upcoming round. It doesn't remove AA units that will fire this round,
unlike RemoveNonCombatants. RemoveNonCombatants is after the AA phase so
it removes AA units that just fired and no longer have any shots.

## Testing

Hard AI World at War

Before this change, it appears that AA units could target "captureableOnEnter" units.  I don't think that is expected behavior.

## Release Note

FIX|AA units can no longer accidentally hit units that will be captured after the battle is over.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
@@ -90,7 +90,10 @@ public interface BattleState {
 
   void retreatUnits(Side side, Collection<Unit> units);
 
-  Collection<Unit> removeNonCombatants(Side side);
+  /**
+   * @param removeAaUnitsThatFireThisRound Should AA units that fire this round be removed or not
+   */
+  Collection<Unit> removeNonCombatants(Side side, boolean removeAaUnitsThatFireThisRound);
 
   /**
    * Mark the units that will be dying
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
@@ -655,8 +655,8 @@ public class MustFightBattle extends DependentBattle
           battleId,
           battleSite,
           getBattleTitle(),
-          removeNonCombatants(attackingUnits, defendingUnits, true, false),
-          removeNonCombatants(defendingUnits, attackingUnits, false, false),
+          calculateCombatants(OFFENSE, false),
+          calculateCombatants(DEFENSE, false),
           killed,
           attackingWaitingToDie,
           defendingWaitingToDie,
@@ -691,8 +691,8 @@ public class MustFightBattle extends DependentBattle
         battleId,
         battleSite,
         getBattleTitle(),
-        removeNonCombatants(attackingUnits, defendingUnits, true, false),
-        removeNonCombatants(defendingUnits, attackingUnits, false, false),
+        calculateCombatants(OFFENSE, false),
+        calculateCombatants(DEFENSE, false),
         killed,
         attackingWaitingToDie,
         defendingWaitingToDie,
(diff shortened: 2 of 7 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1248,65 +1248,57 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
...
+   * </ul>
    */
-  private List<Unit> removeNonCombatants(
-      final Collection<Unit> units,
-      final Collection<Unit> enemyUnits,
-      final boolean attacking,
-      final boolean removeForNextRound) {
-    final List<Unit> unitList = new ArrayList<>(units);
-    if (battleSite.isWater()) {
-      unitList.removeAll(CollectionUtils.getMatches(unitList, Matches.unitIsLand()));
```
trevan, line 1282: This match is already being done in `Matches.unitCanBeInBattle` so it isn't needed here.

game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
```diff
@@ -1248,65 +1248,57 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
...
                 .negate()));
-    // remove capturableOnEntering units (veqryn)
+    // remove capturableOnEntering units
     unitList.removeAll(
         CollectionUtils.getMatches(
             unitList,
             Matches.unitCanBeCapturedOnEnteringToInThisTerritory(
                 attacker, battleSite, gameData.getProperties())));
-    // remove any allied air units that are stuck on damaged carriers (veqryn)
-    unitList.removeAll(
```
trevan, line 1303: This logic is already being done in `MustFightBattle.addAttackChange`.

trevan:
This logic should probably be done when the battle starts and then during each of the "clearCasualty" steps.  I plan to look into this again after the BattleGroups stuff is all done.

Cernelius:
Two points:

- How about offending units pointless for combat but that need to be kept able to retreat with the rest (Are you sure this won't re-introduce the problem that in Feudal Japan horses couldn't retreat?).
- How about infrastructures that cannot fire but are possible targets of AA (not necessarily meaning anti-air) units on following combat rounds too.

Cernelius:
> Before this change, it appears that AA units could target "captureableOnEnter" units. I don't think that is expected behavior.

Good catch. I'm fairly positive you can count this change as a problem (bug) fix even though I've never hit it myself nor have I ever been aware of such behaviour.

trevan:
> Two points:
> 
>     * How about offending units pointless for combat but that need to be kept able to retreat with the rest (Are you sure this won't re-introduce the problem that in Feudal Japan horses couldn't retreat?).
> 
>     * How about infrastructures that cannot fire but are possible targets of AA (not necessarily meaning anti-air) units on following combat rounds too.

This uses the existing "removeNonCombatants" code except it now is called before the AA phase and doesn't remove AA units (or their targets) that will be firing this round.  So, unless those issues you mentioned somehow only happen when a battle makes it to a second round, then this shouldn't cause any problems.

I do know that for the first point, the retreat code actually will grab all of the units in the territory and not just attacking units.  This is why submerged subs would incorrectly retreat.  So any unit owned by the attacker in the territory will retreat (except for submerged units).

DanVanAtta:
Nit, the release note is a bit cryptic. Would you be able to make that more concrete @trevan , perhaps give a specific example, keep in mind the audience for the release note.

trevan:
@DanVanAtta , I don't think that there is any map that currently hits this issue.  I also don't know the terminology that fits better with the audience so you'll need to give me better text.

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
```diff
@@ -90,7 +90,10 @@ public boolean isFirstRound() {
 
   void retreatUnits(Side side, Collection<Unit> units);
 
-  Collection<Unit> removeNonCombatants(Side side);
+  /**
+   * @param removeAaUnitsThatFireThisRound Should AA units that fire this round be removed or not
+   */
+  Collection<Unit> removeNonCombatants(Side side, boolean removeAaUnitsThatFireThisRound);
```
DanVanAtta, line 96: Rather than having battle state be told whether to remove AA units, it seems like AA units should know that. Perhaps after a unit fires, there could be a check run on that unit type to determine if the unit should be removed after firing. 

I mention that for a few reasons:
- boolean parameters are best avoided (automatic SRP violations, usually violate encapsulation and DRY)
- In this case it does seem like we have a DRY violation, the 'removeNonCombatants' method is not able to fully own that logic and needs to be told how to properly remove non-combatants, it is unable to determine which units are non-combatants on its own.

Thoughts, comments?

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
```diff
@@ -88,6 +90,7 @@
 
   static List<BattleStep> getAll(final BattleState battleState, final BattleActions battleActions) {
     return List.of(
+        new RemoveNonCombatantsBeforeAA(battleState, battleActions),
```
DanVanAtta, line 93: It would be kinda nice to see 'removeNonCombatants' able to just do its thing and not have a variant that is pre-AA. Having the same battle step listed twice would be fine IMO (ie; this could be a `RemoveNonCombatants`), but the difference is that the object is able to compute at any given time, given a battle state, who the non-combatants are. This would make the object more useful and fully contain the logic of determining non-combatants (I think that would make for a stronger, more cohesive module. I also think the testing for it would be more interesting as well to determine that it computes non-combatants correctly across all cases).

game-core/src/main/java/games/strategy/triplea/delegate/battle/BattleState.java
trevan: Currently, units don't know that they fired or are even dead (even the amount of hits that a unit has doesn't indicate if it is dead or not).  So, the idea is great but it is not easily doable at this time.

game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/BattleStep.java
trevan: This is save compatibility issue.  I can't modify `RemoveNonCombatants` to give it additional logic because then an older save will still run it and do potentially strange things.  I also can't move `RemoveNonCombatents` for the same reason because an older save will have it in the old location.

Also, the `removeNonCombatants` logic inside of `MustFightBattle` is used in other places internally inside of `MustFightBattle` so moving it outside of `MustFightBattle` would either need to make it static method (which I think it can be) or duplicated (which I don't like the idea of).

DanVanAtta:
The save game compatibility issue is a good point to raise that I had not considered. Given the constraints, I could see this being a fine update. Why closed @trevan ?

trevan:
It was a part of the battle group, firing squadron piece.  So, with the change to the unit abilities, this will probably change.

PR #8611, opened 2021-01-10
Syntactic code cleanup

RoiEXLab:
99% of this PR are automated IntelliJ quick fixes applied to the code, all of which I reviewed individually.
There are no semantic changes, everything is purely syntax related.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorPanel.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorPanel.java
@@ -317,14 +317,11 @@ public final class GameSelectorPanel extends JPanel implements Observer {
           loadSavedGame.setEnabled(canSelectGameData || canChangeHostBotGameData);
           loadNewGame.setEnabled(canSelectGameData || canChangeHostBotGameData);
           // Disable game options if there are none.
-          if (canChangeHostBotGameData
-              || (canSelectGameData
-                  && model.getGameData() != null
-                  && !model.getGameData().getProperties().getEditableProperties().isEmpty())) {
-            mapOptions.setEnabled(true);
-          } else {
-            mapOptions.setEnabled(false);
-          }
+          mapOptions.setEnabled(
+              canChangeHostBotGameData
+                  || (canSelectGameData
+                      && model.getGameData() != null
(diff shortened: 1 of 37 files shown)


game-core/src/main/java/games/strategy/triplea/delegate/TechAdvance.java
```diff
@@ -185,7 +179,7 @@ public static void createDefaultTechAdvances(final GameData data) {
    *     instance.
    */
   public static TechAdvance findDefinedAdvanceAndCreateAdvance(
-      final String technologyName, final GameState data) {
```
RoiEXLab, line 188: This is actually an important warning that was pointed out by intelliJ.
The reflection only searches for a constructor that accepts an instance of `GameData`, which is to be fair currently the only implementing class of the `GameState` interface, but if there was another implementation this method would throw an exception

game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorPanel.java
```diff
@@ -317,14 +317,11 @@ private void updateGameData() {
...
           // Disable game options if there are none.
-          if (canChangeHostBotGameData
-              || (canSelectGameData
-                  && model.getGameData() != null
-                  && !model.getGameData().getProperties().getEditableProperties().isEmpty())) {
-            mapOptions.setEnabled(true);
-          } else {
-            mapOptions.setEnabled(false);
-          }
+          mapOptions.setEnabled(
```
DanVanAtta, line 320: To some extent one could argue each automated update should be applied one by one:
1. Can more easily review as all changes are the same type
2. Commit history documentation is better, can copy/paste description of the change type into the commit summary

There is though a running debate whether automated passes are always the right thing. In this code, there is more work to be done IMO to really fix it up. For example, we have one boolean condition extracted to a variable and a second boolean condition that is fully written out. Overall, it's hard to grok. One could argue that if you're going to touch this up, then this location should be fully fixed up to avoid churn.

While I support mass cleaning, there are some aspects of it that do have me on the fence.

DanVanAtta (approved):
Overall cool, thanks :+1: 

I do think I would have liked to have seen the different automated cleanups be run as different commits or PRs with slightly additional documentation of what was being updated in the respective commit messages.

PR #8616, opened 2021-01-10
Add utility to extract zip files

DanVanAtta:
Adds code that can be used to extract a zip file. Future
commits will use this code to scan downloaded maps and
extract zipped files and secondarily to extract newly
downloaded zip files.

## Release Note

Please focus on whether the code is easy to understand and the documentation is clear and concise.


java-extras/src/main/java/org/triplea/io/ZipExtractor.java
```diff
@@ -0,0 +1,161 @@
...
+
+    // iterate over each zip entry and write to a corresponding file
+    try (ZipFile zipFile = new ZipFile(fileZip)) {
+      for (final ZipEntry zipEntry : zipFile.stream().collect(Collectors.toList())) {
+        try {
+          unzipZipEntry(destDir, zipFile, zipEntry);
+        } catch (final IOException e) {
+          throw new FileSystemException(fileZip, e);
+        }
+      }
```
RoiEXLab, line 83: Did you consider using the `FileSystem` API I discovered as part of #8572 ?
So instead of using the `ZipFile` class, it might be way less code to use that and write something along the files of this:
```java
var fs = FileSystems.newFileSystem(fileZip.toPath(), null);
try (Stream<Path> files : Files.walk(fs.getPath("/"))) {
  files.forEach(path -> {
    var filePath = Path.of("path/to/folder").resolve(Path.of(fs.getPath("/").relativize(path).toString()));
    if (Files.isDirectory(path)) {
      Files.createDirectory(filePath);
    } else {
      Files.copy(path, filePath);
    }
  });
}
```
Haven't tested the code, but it sounds shorter I guess?

RoiEXLab (approved):

java-extras/src/test/java/org/triplea/io/ZipExtractorTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+      final File unzippedDirectory = new File(destinationFolder, "directory");
+      assertThat(unzippedDirectory.isDirectory(), is(true));
+      assertThat(unzippedDirectory.exists(), is(true));
+
+      final File unzippedTestFile2 = new File(unzippedDirectory, "test-file2");
+      final String testFile2Contents = Files.readString(unzippedTestFile2.toPath());
+      assertThat(testFile2Contents, is("lupsom osculus\n"));
+    } finally {
+      deleteDirectory(destinationFolder);
```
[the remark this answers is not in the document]
DanVanAtta: Only empty directories can be deleted, so it depends if Files.walk is depth first or breadth first. It does appear to be depth first. In this case I'm happy using a library to avoid the questions (just easier on maintainers). I agree what you posted should do the same thing.

java-extras/src/main/java/org/triplea/io/ZipExtractor.java
DanVanAtta: I don't know if you can copy from a zip path. I did have some issues with URI's in the past being passed from what we got in a zip to Path. There was a Baeldung article that I essentially followed to get an example. I'd only be concerned about any kinda of OS dependency.

FWIW - A large chunk of code is to prevent the 'zip-slip' vulnerability, without that there is quite a bit less. Thoughts? 

To another extent, no objection to the above, but I'm reluctant to test it.

DanVanAtta: This has been nagging me a bit. Bottom line I'm not having the best luck incorporating FileSystems.

Happily there is a test that is implementation agnostic that extracts a simple and known zip file. So long as that worked, and we can confirm it works on multiple OS's, we should be good.

I am having some trouble casting the FileSystem example into something working. There is quite a bit of logic to catch IOException at various points and translate that either into a file system error or as an error reading the zip file itself. Furthermore, we'd have to be sure we have the 'zip-slip' defense logic built with this new API as well. A really large chunk of the overall extraction code is that type of logic.

RoiEXLab: Hmm I honestly didn't put too much effort into this example.
I created a real java file this time (and didn't rely on jshell this time) and wrote an actual testable application.
This code worked well for me, and I believe it should provide the necessary security (I wasn't able to find to find a malicious zip file to test this though):
```java
    var zipFile = Path.of("path/to/file.zip");
    try {
      var zipFs = FileSystems.newFileSystem(zipFile, null);
      // It seems that / will always be the root folder in zip files, however
      // maybe it's safer to use zipFs.getRootDirectories().iterator().next() instead?
      var zipRoot = zipFs.getPath("/");
      try (Stream<Path> files = Files.walk(zipRoot)) {
        // Where the files should be extracted to, this folder doesn't need to exist, but the parent folder has to.
        var targetFolder = zipFile.resolveSibling("target");
        for (Path path : files.collect(Collectors.toList())) {
          // Translate zip location to target path
          var targetFile = targetFolder.resolve(zipRoot.relativize(path).toString());
          // If I understand zip-slip correctly this simple check should prevent it, java does all the parsing for us
          if (!targetFile.startsWith(targetFolder)) {
            throw new IllegalStateException("Zip-slip detected!");
          }
          if (Files.isDirectory(path)) {
            if (!Files.exists(targetFile)) {
              Files.createDirectory(targetFile);
            }
          } else {
            Files.copy(path, targetFile, StandardCopyOption.REPLACE_EXISTING);
          }
        }
      }
    } catch (IOException e) {
      e.printStackTrace();
    }
```

Does that help?

PR #8622, opened 2021-01-12
Store downloaded maps unzipped

DanVanAtta:
Updates code to extract from zip existing and newly downloaded
maps. Various code paths to load maps from zip can then be
removed.

- Ready to get feedback
- Still a draft while some final testing is done
- Most of the submitted code is polished, but could still be rough in a few places

## Release Note

UPDATE|Downloaded maps are now extracted and stored as flat files rather than as zipped files.


RoiEXLab (commented):
I must admit I was sceptical when I started reviewing this PR if we really gain something from this change, but after having a look at the changes I'm convinced that this is a step in the right direction.

game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/FileSystemMapFinder.java
```diff
@@ -47,27 +41,6 @@
         new File(userMapsFolder, normalizedMapName));
```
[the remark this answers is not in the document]
DanVanAtta: Not having to deal with file separator is a win. We likely should be preferring path where we can. Working in existing code is a bit odd as it's mixed now, but I'm certainly on board to convert and use Path in new code.

game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/AvailableGamesFileSystemReader.java
```diff
@@ -27,15 +24,11 @@
...
-  private static final String ZIP_EXTENSION = ".zip";
-
   static synchronized DownloadedMaps parseMapFiles() {
-    final Set<DefaultGameChooserEntry> entries = new HashSet<>();
-    entries.addAll(mapXmlsGameNamesByUri(findAllZippedXmlFiles()));
-    entries.addAll(mapXmlsGameNamesByUri(findAllUnZippedXmlFiles()));
-    entries.forEach(
-        entry -> log.debug("Found game: " + entry.getGameName() + " @ " + entry.getUri()));
-    return new DownloadedMaps(entries);
+    final List<URI> unzippedXmlFiles = findAllUnZippedXmlFiles();
```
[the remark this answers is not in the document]
DanVanAtta: renamed: ce91d278b6  

FWIW, when 'map.yml' lands, this code will change and instead we'll be scanning for the map.yml file and then finding paths based on the contents of the 'map.yml'

game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
```diff
@@ -176,10 +151,8 @@ public boolean hasPath(final String path) {
...
    */
   public @Nullable URL getResource(final String inputPath, final String inputPath2) {
-    final String path = mapPrefix + inputPath;
-    final String path2 = mapPrefix + inputPath2;
-    return findResource(path)
-        .or(() -> findResource(path2))
+    return findResource(inputPath)
+        .or(() -> findResource(inputPath2))
         .or(() -> findResource(inputPath))
         .or(() -> findResource(inputPath2))
```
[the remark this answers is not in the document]
DanVanAtta: Ah, the 'mapPrefix' is now always empty as it is only used for zip files. I simply removed the mapPrefix piece, which did make these paths duplicative. Good eye.

0c627cec2f

game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/AvailableGamesFileSystemReader.java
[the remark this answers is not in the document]
DanVanAtta: cb76162958

game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
```diff
@@ -89,26 +83,6 @@ public ResourceLoader(final String mapName) {
...
-   */
-  private static String getMapPrefix(final File mapZip) {
-    try (ZipFile zip = new ZipFile(mapZip)) {
-      final Optional<? extends ZipEntry> baseTilesEntry =
-          zip.stream()
-              .filter(entry -> entry.getName().endsWith(REQUIRED_ASSET_EXAMPLE_FOLDER))
-              .findAny();
-      if (baseTilesEntry.isPresent()) {
-        final String path = baseTilesEntry.get().getName();
-        return path.substring(0, path.length() - REQUIRED_ASSET_EXAMPLE_FOLDER.length());
```
DanVanAtta, line 104: @jdimeo , this update now has everything being a folder. Does the submodule structure you have still work as directories rather than as expanded zip files?

game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/ZippedMapsExtractor.java
```diff
@@ -0,0 +1,170 @@
...
+      log.error(
+          "Failed to move file: "
+              + mapZip.getAbsolutePath()
+              + ", to: "
+              + badZipFolder.toFile().getAbsolutePath(),
+          e);
+      return Optional.empty();
+    }
+  }
+}
```
RoiEXLab, line 170: Ah one thing I forgot to mention:
The Zipped Maps Extractor is just a migration tool right?
If so, would this be a candidate for a class that should be removed in the next major release or some other point in time?

DanVanAtta:
We'll continue downloading maps as a zip and will want to support manually
installed map zip files. So it'll be around long term.

game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
jdimeo: @DanVanAtta I'm not sure! I think this discovery process may still have to be re implemented for the folders, but I also know that the unzipped maps had less hardcoded assumptions about where things were located.

DanVanAtta: Thanks. It's certainly a bit of a mess that TripleA assumes that a map name is equal to the map folder. This is something I'm trying to remove. Once the 'map.yml' file lands, it might be a lot of what you would want in order to have more customized locations. Regardless, that would probably be the right time FWIW to 'fix' the path discovery.

Cernelius:
> will want to support manually installed map zip files.

That is good.

game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
```diff
@@ -163,8 +137,9 @@ public boolean hasPath(final String path) {
    *     resource. Do not use '\' or File.separator)
    */
   public @Nullable URL getResource(final String inputPath) {
-    final String path = mapPrefix + inputPath;
-    return findResource(path).or(() -> findResource(inputPath)).orElse(null);
+    return findResource(inputPath) //
+        .or(() -> findResource(inputPath))
+        .orElse(null);
```
RoiEXLab, line 142: Looks like I missed this in this PR:
findResource() is called twice with the same parameter

PR #8637, opened 2021-01-17
Simplify link localization code

DanVanAtta:
Main goal of this simplification is to remove 'map name' from resource loader.
One place that is used is in Link Localization. Instead of passing a resource
loader to the link localizer, we can instead pass a file path which in turn
simplifies how link localization can be done.

Additional:
- removed link localization cache
- changed link localizatoin behavior to only localize links that are not already
  absolute. Previously links in single quotes would be localized and links in
  double quotes would be left as-is. The localization of fully qualified links
  would parse the link path, get the file name and localize the file name path.
  This is removed as it is difficult to understand and is much easier if
  we localize any relative links but keep absolute links as they are (regardless
  of quoting).

## Testing
- Downloaded a variety of maps and check that their images still loaded in the game selection window.
- Launched Civil War to do some smoke testing for missing images

## Release Note

UPDATE|Map making - clarified how image links in HTML is made localized. If the link is absolute, it will now always remain absolute (eg: https://path-to-my-image/my-image.png). Otherwise if the image path is relative (eg: "img src=someImage.png"), the path will be made relative to the maps folder 'doc/images/'.

### DIFF
--- a/game-core/src/main/java/org/triplea/util/LocalizeHtml.java
+++ b/game-core/src/main/java/org/triplea/util/LocalizeHtml.java
@@ -1,58 +1,42 @@
 package org.triplea.util;
 
-import com.google.common.annotations.VisibleForTesting;
-import games.strategy.triplea.ResourceLoader;
-import games.strategy.triplea.ui.UiContext;
-import java.net.URL;
-import java.util.HashMap;
-import java.util.Map;
+import java.io.File;
+import java.nio.file.Path;
 import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import lombok.experimental.UtilityClass;
-import lombok.extern.slf4j.Slf4j;
+import org.triplea.io.FileUtils;
 
 /**
  * Provides methods that convert relative links within a game description into absolute links that
- * will work on the local system.
+ * will work on the local system. Links that are already absolute (those beginning with 'http') will
+ * not be changed.
  */
-@Slf4j
 @UtilityClass
 public final class LocalizeHtml {
-  private static final String ASSET_IMAGE_FOLDER = "doc/images/";
-  private static final String ASSET_IMAGE_NOT_FOUND = "notFound.png";
   // Match the <img> src
   private static final Pattern PATTERN_HTML_IMG_SRC_TAG =
       Pattern.compile(
           "(<img[^>]*src\\s*=\\s*)(?:\"([^\"]+)\"|'([^']+)')([^>]*/?>)", Pattern.CASE_INSENSITIVE);
 
   /**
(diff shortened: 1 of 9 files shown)


game-core/src/main/java/org/triplea/util/LocalizeHtml.java
```diff
@@ -62,23 +46,9 @@ static String localizeImgLinksInHtml(final String htmlText, final ResourceLoader
...
-    // remove full parent path
-    final String imageFileName = link.substring(Math.max(link.lastIndexOf("/") + 1, 0));
-
-    // replace when testing with: "REPLACEMENTPATH/" + imageFileName;
-    final String firstOption = ASSET_IMAGE_FOLDER + imageFileName;
-    URL replacementUrl = loader.getResource(firstOption);
-
-    if (replacementUrl == null) {
-      log.error(String.format("Could not find: %s/%s", loader.getMapName(), firstOption));
-      final String secondFallback = ASSET_IMAGE_FOLDER + ASSET_IMAGE_NOT_FOUND;
```
RoiEXLab, line 75: Is the fallback image never used?

game-core/src/main/java/org/triplea/util/LocalizeHtml.java
```diff
@@ -1,58 +1,42 @@
...
-   */
-  public static String localizeImgLinksInHtml(final String htmlText, final String mapNameDir) {
-    return htmlText == null || mapNameDir == null || mapNameDir.isBlank()
-        ? htmlText
-        : localizeImgLinksInHtml(htmlText, new ResourceLoader(mapNameDir));
-  }
-
-  @VisibleForTesting
-  static String localizeImgLinksInHtml(final String htmlText, final ResourceLoader loader) {
+  public static String localizeImgLinksInHtml(final String htmlText, final File mapContentFolder) {
```
RoiEXLab, line 27: So in case the resource does not exist we will no longer notice at this point, right?
And because the HTML is not really handled by us, this will result in a silent error, just like it would in a normal web browser

RoiEXLab (approved):

game-core/src/main/java/org/triplea/util/LocalizeHtml.java
DanVanAtta: Yes, it's never used, the file 'notFound.png' is not part of the game engine. A map could have it, but none do.

game-core/src/main/java/org/triplea/util/LocalizeHtml.java
DanVanAtta: The resource loader was only be used to get the map folder location, it was not actually loading the image or validating that it was there. The previous logic was doing everything as a String operation, so we were not noticing it either way.

PR #8640, opened 2021-01-17
Map version lookup - simplify & move to DownloadedMaps

DanVanAtta:
commit f6b9f52566f46a1bf56155d4160b48936149fa04

    Simplify out of date map version lookup
    
    Previously we would compute the mapping of installed maps to versions,
    available maps to versions, then compute the union of these two maps
    and return any maps that had an installed version less than what
    was available.
    
    This update changes that algorithm to instead to iterate over the
    list of available maps, find that map, find the version for that
    map, and if out of date the map is out of date.
    
    Of note, rather than computing a Map of available map names to
    versions, we directly use 'DownloadFileDescription'.

commit 0ffabd1f9bbc40e17e5532f345c114f909b1bcea

    Move method 'getMapVersionByName' to DownloadedMaps.java

commit 95a653800687587c866e3531c2bd82d6f898c884

    Update test cases for out of date map logic

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/auto/update/UpdatedMapsCheck.java
+++ b/game-core/src/main/java/games/strategy/engine/auto/update/UpdatedMapsCheck.java
@@ -1,22 +1,18 @@
 package games.strategy.engine.auto.update;
 
 import com.google.common.annotations.VisibleForTesting;
-import games.strategy.engine.ClientFileSystemHelper;
 import games.strategy.engine.framework.map.download.DownloadFileDescription;
-import games.strategy.engine.framework.map.download.DownloadFileProperties;
 import games.strategy.engine.framework.map.download.DownloadMapsWindow;
+import games.strategy.engine.framework.map.file.system.loader.DownloadedMaps;
 import games.strategy.engine.framework.map.listing.MapListingFetcher;
 import games.strategy.triplea.settings.ClientSetting;
-import java.io.File;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
-import java.util.Optional;
 import lombok.experimental.UtilityClass;
 import lombok.extern.slf4j.Slf4j;
-import org.triplea.io.FileUtils;
 import org.triplea.swing.SwingComponents;
 
 @UtilityClass
@@ -78,7 +74,7 @@ class UpdatedMapsCheck {
     // Loop over all available maps, check if we have that map present, its version,
     // and remember any whose version is less than what is available.
     for (final DownloadFileDescription availableMap : availableToDownloadMaps) {
-      getMapVersionByName(availableMap.getMapName())
+      DownloadedMaps.getMapVersionByName(availableMap.getMapName())
           .ifPresent(
               installedVersion -> {
                 if (installedVersion < availableMap.getVersion()) {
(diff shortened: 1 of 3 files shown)


game-core/src/main/java/games/strategy/engine/auto/update/UpdatedMapsCheck.java
```diff
@@ -49,111 +44,52 @@ public static void checkDownloadedMapsAreLatest() {
...
+      mapVersionLookup
+          .apply(availableMap.getMapName())
+          .ifPresent(
+              installedVersion -> {
+                if (installedVersion < availableMap.getVersion()) {
+                  outOfDateMapNames.add(availableMap.getMapName());
+                }
+              });
     }
     return outOfDateMapNames;
```
RoiEXLab, line 90:
```java
return availableToDownloadMaps.stream()
  .filter(availableMap -> mapVersionLookup.apply(availableMap.getMapName()).orElse(Integer.MAX_VALUE) < availableMap.getVersion())
  .map(DownloadFileDescription::getMapName)
```

Not sure if that's better tbh. Maybe if the filter is moved to a helper method?

game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/DownloadedMaps.java
```diff
@@ -65,4 +68,51 @@ public boolean hasGame(final String gameName) {
...
+      normalizedName = inputName.substring(0, inputName.indexOf(".zip.properties"));
+    }
+    if (normalizedName.endsWith("-master")) {
+      normalizedName = inputName.substring(0, inputName.indexOf("-master"));
+    }
+
+    normalizedName = normalizedName.replaceAll(" ", "_");
+    normalizedName = normalizedName.toLowerCase();
+
+    return normalizedName;
```
RoiEXLab, line 116: In my opinion using a RegEx simplifies the code by a reasonable amount
```java
return normalizedName.toLowerCase()
  .replaceAll(" ", "_")
  .replaceAll("(?:-master)?(?:.zip.properties)?$", "");
```

RoiEXLab (approved):

game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/DownloadedMaps.java
DanVanAtta: That is more concise, my 2 cents it's kinda perl-esque in that it's a lot shorter, but more difficult to understand. I'd also start to wonder about the runtime as we do this normalization n-squared times.

To another extent, this code I think will go away when we start matching map name from a `map.yml` instead of deriving it from folder name. Let's hold off on changes to how that is done as it could be churn

game-core/src/main/java/games/strategy/engine/auto/update/UpdatedMapsCheck.java
DanVanAtta: Yeah, that filter is overwhelming. A helper method is additional indirection, so not a clear win necessarily. I think I'm on the fence, could argue it both ways.

I think the effort needs to be put into `final Function<String, Optional<Integer>> mapVersionLookup`, I'm not a fan of having such a generic function be passed into a public API, makes the usage of this method very difficult.

DanVanAtta:
Thank you for taking a look over this @RoiEXLab . The code in `UpdatedMapsCheck` should now be relatively stable FWIW.  Though, the implementation of the function used to find the version of an installed map will change (inside of `DownloadedMaps`, most of that class will change).

PR #8642, opened 2021-01-17
Add a YamlUtils to simplify reading & writing YAML

DanVanAtta:
YamlUtils is essentially a SnakeYaml wrapper

### DIFF
--- a/java-extras/src/main/java/org/triplea/yaml/YamlUtils.java
+++ b/java-extras/src/main/java/org/triplea/yaml/YamlUtils.java
@@ -1,0 +1,60 @@
+package org.triplea.yaml;
+
+import com.google.common.base.Strings;
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
+import java.util.Optional;
+import lombok.experimental.UtilityClass;
+import org.snakeyaml.engine.v2.api.Dump;
+import org.snakeyaml.engine.v2.api.DumpSettings;
+import org.snakeyaml.engine.v2.api.Load;
+import org.snakeyaml.engine.v2.api.LoadSettings;
+import org.snakeyaml.engine.v2.exceptions.YamlEngineException;
+
+/**
+ * Methods useful for working reading and writing YAML data. This class largely is a wrapper around
+ * snake yaml.
+ */
+@UtilityClass
+public class YamlUtils {
+  public <T> T readYaml(final String inputString) {
+    final var inputStream =
+        new ByteArrayInputStream(Strings.nullToEmpty(inputString).getBytes(StandardCharsets.UTF_8));
+    return readYaml(inputStream);
+  }
+
+  /**
+   * Reads from input stream a YAML data structure. YAML is assumed to be map at the top level.
+   *
+   * @throws InvalidYamlFormatException Thrown if the YAML is badly formatted.
+   */
+  @SuppressWarnings("unchecked")
+  public <T> T readYaml(final InputStream inputStream) {
+    final Load load = new Load(LoadSettings.builder().build());
+    try {
+      final var result = (T) load.loadFromInputStream(inputStream);
+      return Optional.ofNullable(result)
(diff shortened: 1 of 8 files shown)


java-extras/src/main/java/org/triplea/yaml/YamlUtils.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Methods useful for working reading and writing YAML data. This class largely is a wrapper around
+ * snake yaml.
+ */
+@UtilityClass
+public class YamlUtils {
+  public <T> T readYaml(final String inputString) {
+    final var inputStream =
+        new ByteArrayInputStream(Strings.nullToEmpty(inputString).getBytes(StandardCharsets.UTF_8));
```
RoiEXLab, line 24: Does nullToEmpty really make sense here?
If it was null, wouldn't the parser throw an exception due to missing data?
So a precondition would be a better choice IMO

java-extras/src/main/java/org/triplea/yaml/YamlUtils.java
```diff
@@ -0,0 +1,60 @@
...
+      super(message);
+    }
+
+    public InvalidYamlFormatException(final Exception e) {
+      super("Invalid yaml format: " + e.getMessage(), e);
+    }
+  }
+
+  public String writeToYamlString(final Map<String, Object> data) {
+    return new Dump(DumpSettings.builder().build()).dumpToString(data);
```
RoiEXLab, line 58: Is this method used?
If not, why was it introduced?

java-extras/src/main/java/org/triplea/yaml/YamlUtils.java
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Methods useful for working reading and writing YAML data. This class largely is a wrapper around
+ * snake yaml.
+ */
+@UtilityClass
+public class YamlUtils {
+  public <T> T readYaml(final String inputString) {
+    final var inputStream =
+        new ByteArrayInputStream(Strings.nullToEmpty(inputString).getBytes(StandardCharsets.UTF_8));
+    return readYaml(inputStream);
```
RoiEXLab, line 25: I must admit that this is an elegant way to re-use the code, but I wonder if we could've achieved the same thing by passing either `load.loadFromString` or `load.loadFromInputStream` as a lambda to a common helper function

RoiEXLab (approved):

java-extras/src/main/java/org/triplea/yaml/YamlUtils.java
DanVanAtta: Can you elaborate, I don't quite follow how the common helper function would look.

java-extras/src/main/java/org/triplea/yaml/YamlUtils.java
DanVanAtta: This is forward looking, I should have called out that it's not used yet. It will be soon though.

java-extras/src/main/java/org/triplea/yaml/YamlUtils.java
DanVanAtta: This is lifted from: 

```
@UtilityClass
public class StringToInputStream {

  /** Converts a {@code String} to an equivalent {@code InputStream}. */
  public static InputStream asInputStream(final String inputString) {
    return new ByteArrayInputStream(
        Strings.nullToEmpty(inputString).getBytes(StandardCharsets.UTF_8));
  }
}
```
That is in the test-common package. Does it make sense to make that a java-common utility at this point, to support one non-test code usage? I don't know. Keeping that code similar looking in the meantime has value if we do consolidate it in the future.

With that said, a Precondition would not give a user friendly error message and would be a difficult exception type for calling code to deal with. In part the wrapper here is unifying the error scenarios to throw one exception type. An empty or a null input is potentially a valid empty YAML. 

Is handling a null input a silent failure? It's hard to say at this level of code, it could be, but maybe not - it is a smell.

PR #8646, opened 2021-01-17
Add BattlePhaseList, BattlePhase, and UnitAbilities

trevan:
This adds the objects/classes for handling unit abilities and battle phases.  This doesn't yet add them to the XML nor does it add them to the actual battle logic.

I'll be working on adding them to the battle logic in a later PR.  I'm making them separate PRs because this already has a lot of lines (~1000 if you ignore the tests) and is needed to make the existing maps work.  I plan to do three more PRs: One that adds battle logic but doesn't actually use it, one that finally hooks it in, and one that adds them to the XML parsing.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
+++ b/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
@@ -45,8 +45,8 @@ public class CombatUnitAbility {
   @Builder.Default
   Collection<BattleState.Side> sides = List.of(BattleState.Side.OFFENSE, BattleState.Side.DEFENSE);
 
-  /** How many combat rounds this ability can be used. -1 means unlimited */
-  @Builder.Default int round = -1;
+  /** How many combat rounds this ability can be used */
+  @Builder.Default int round = Integer.MAX_VALUE;
 
   /**
    * Can the casualties return fire after the battle phase that this UnitAbility is active?
--- a/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
+++ b/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
@@ -1,0 +1,106 @@
+package games.strategy.engine.data.unit.ability;
+
+import games.strategy.engine.data.Unit;
+import games.strategy.engine.data.UnitType;
+import games.strategy.triplea.delegate.battle.BattleState;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import lombok.AccessLevel;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NonNull;
+import lombok.experimental.FieldDefaults;
+import lombok.experimental.NonFinal;
+
+/** Unit Ability for combat situations */
+@Data
+@FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
+@Builder(toBuilder = true)
+public class CombatUnitAbility {
+
+  enum DiceType {
+    // Use AA dice (attackAA, offensiveAttackAA)
+    AA,
+    // Use bombard dice (bombard or attack if bombard is empty)
+    BOMBARD,
+    // Use normal dice (attack, defense)
+    NORMAL
+  }
+
+  /** The name of this unit ability that will be shown in the Battle UI */
+  @NonNull String name;
+
+  /** The unit types that have this ability */
+  @NonFinal Collection<UnitType> attachedUnitTypes;
+
+  /** The unit types that can be targeted */
+  @NonNull Collection<UnitType> targets;
+
+  /** The type of dice that will be rolled. */
+  @Builder.Default DiceType diceType = DiceType.NORMAL;
+
+  /** The side(s) that the unit needs to be on to use the ability */
+  @Builder.Default
+  Collection<BattleState.Side> sides = List.of(BattleState.Side.OFFENSE, BattleState.Side.DEFENSE);
+
+  /** How many combat rounds this ability can be used. -1 means unlimited */
+  @Builder.Default int round = -1;
+
+  /**
+   * Can the casualties return fire after the battle phase that this UnitAbility is active?
+   *
+   * <p>Casualties that are created in the battle phase where this UnitAbility is active can always
+   * return fire since they are firing simultaneously
+   */
+  @Builder.Default boolean returnFire = true;
+
+  /** Will these units commit suicide when they have a hit? */
+  @Builder.Default boolean suicideOnHit = false;
+
+  public boolean isTarget(final Unit unit) {
+    return targets.contains(unit.getType());
+  }
+
+  public boolean isSide(final BattleState.Side side) {
+    return sides.contains(side);
+  }
+
+  public boolean isRound(final int round) {
+    return this.round == -1 || this.round <= round;
+  }
+
+  /**
+   * Can the attachedTo from "other" be combined with this?
+   *
+   * <p>This is only used to merge auto created CombatUnitAbilities.
+   *
+   * <p>The name is not important in deciding whether the attachedTo can be combined.
+   *
+   * <p>The attachedTo is only important if this is an suicideOnHit ability. By ensuring that each
+   * CombatUnitAbility is unique per suicideOnHit unitType, this simplifies the logic to determine
+   * which unit should suicide after a hit.
+   */
+  public boolean canMergeAttachedUnitTypes(final CombatUnitAbility other) {
+    // build a new CombatUnitAbility based off of other but with this CombatUnitAbility's name
+    // and attachedUnitTypes (if not suicideOnHit) as they are not relevant for merging.
+    return equals(
+        other.toBuilder()
+            .name(this.name)
+            .attachedUnitTypes(suicideOnHit ? other.attachedUnitTypes : this.attachedUnitTypes)
+            .build());
+  }
+
+  /**
+   * Merges the attachedUnitTypes from the unitAbility with this attachedUnitTypes
+   *
+   * <p>This is only used to merge auto created CombatUnitAbilities.
+   */
+  public void mergeAttachedUnitTypes(final CombatUnitAbility unitAbility) {
+    this.attachedUnitTypes =
+        Stream.of(this.attachedUnitTypes, unitAbility.attachedUnitTypes)
+            .flatMap(Collection::stream)
+            .collect(Collectors.toList());
+  }
+}
--- a/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
+++ b/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
@@ -48,14 +48,12 @@ public class CombatUnitAbility {
 
   public static final CombatUnitAbility EMPTY = CombatUnitAbility.builder().name("Empty").build();
 
+  /** Which {@link games.strategy.triplea.delegate.power.calculator.CombatValue} should be used */
   @Value
-  static class DiceType {
-    // Use AA dice (attackAA, offensiveAttackAA)
-    static final DiceType AA = new DiceType("AA");
-    // Use bombard dice (bombard or attack if bombard is empty)
-    static final DiceType BOMBARD = new DiceType("BOMBARD");
-    // Use normal dice (attack, defense)
-    static final DiceType NORMAL = new DiceType("NORMAL");
+  static class CombatValueType {
+    static final CombatValueType AA = new CombatValueType("AA");
+    static final CombatValueType BOMBARD = new CombatValueType("BOMBARD");
+    static final CombatValueType NORMAL = new CombatValueType("NORMAL");
 
     String type;
   }
@@ -78,7 +76,7 @@ public class CombatUnitAbility {
   @Builder.Default Collection<UnitType> targets = List.of();
 
   /** The type of dice that will be rolled. */
-  @Builder.Default DiceType diceType = DiceType.NORMAL;
+  @Builder.Default CombatValueType combatValueType = CombatValueType.NORMAL;
 
   /** The side(s) that the unit needs to be on to use the ability */
   @Builder.Default
--- a/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
+++ b/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
@@ -14,7 +14,32 @@ import lombok.NonNull;
 import lombok.experimental.FieldDefaults;
 import lombok.experimental.NonFinal;
 
-/** Unit Ability for combat situations */
+/**
+ * Unit Ability for targeting and causing damage to units
+ *
+ * <p>A unit ability indicates what targets a unit type has, what dice it should roll, whether its
+ * casualties can return fire in a later {@link
+ * games.strategy.engine.data.battle.phase.BattlePhase}.
+ *
+ * <p>Some examples of what this does cover are:
+ *
+ * <ul>
+ *   <li>Instant Kill for a Submarine: It would use the NORMAL dice type, its targets would be sea
+ *       units, and casualties would not return fire.
+ *   <li>Land Mines Explosion: It would use the NORMAL dice type, its targets would be land based,
+ *       it would commit suicide, and the casualties would return fire.
+ *   <li>Depth Charges for Destroyer: It would use the AA dice type, its targets would be
+ *       submarines, and the casualties would return fire.
+ * </ul>
+ *
+ * Some examples of what this doesn't cover are:
+ *
+ * <ul>
+ *   <li>The ability to blitz
+ *   <li>The ability to retreat during a battle
+ *   <li>The ability to create new units
+ * </ul>
+ */
 @Data
 @FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
 @Builder(toBuilder = true)
--- a/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
+++ b/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
@@ -7,10 +7,10 @@ import java.util.Collection;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import javax.annotation.Nonnull;
 import lombok.AccessLevel;
 import lombok.Builder;
 import lombok.Data;
-import lombok.NonNull;
 import lombok.experimental.FieldDefaults;
 import lombok.experimental.NonFinal;
 
@@ -57,7 +57,7 @@ public class CombatUnitAbility {
   }
 
   /** The name of this unit ability that will be shown in the Battle UI */
-  @NonNull String name;
+  @Nonnull String name;
 
   /** The unit types that have this ability */
   @Builder.Default @NonFinal Collection<UnitType> attachedUnitTypes = List.of();
--- a/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
+++ b/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
@@ -11,6 +11,7 @@ import javax.annotation.Nonnull;
 import lombok.AccessLevel;
 import lombok.Builder;
 import lombok.Data;
+import lombok.Value;
 import lombok.experimental.FieldDefaults;
 import lombok.experimental.NonFinal;
 
@@ -47,13 +48,16 @@ public class CombatUnitAbility {
 
   public static final CombatUnitAbility EMPTY = CombatUnitAbility.builder().name("Empty").build();
 
-  enum DiceType {
+  @Value
+  static class DiceType {
     // Use AA dice (attackAA, offensiveAttackAA)
-    AA,
+    static final DiceType AA = new DiceType("AA");
     // Use bombard dice (bombard or attack if bombard is empty)
-    BOMBARD,
+    static final DiceType BOMBARD = new DiceType("BOMBARD");
     // Use normal dice (attack, defense)
-    NORMAL
+    static final DiceType NORMAL = new DiceType("NORMAL");
+
+    String type;
   }
 
   enum Suicide {
--- a/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
+++ b/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
@@ -56,8 +56,21 @@ public class CombatUnitAbility {
    */
   @Builder.Default boolean returnFire = true;
 
-  /** Will these units commit suicide when they have a hit? */
-  @Builder.Default boolean suicideOnHit = false;
+  /**
+   * Does this unit commit suicide after it has fired a shot?
+   *
+   * <p>Whether the shot hit or not doesn't matter.
+   *
+   * <p>It can commit suicide as a defensive unit, offensive unit, or both.
+   */
+  @Builder.Default Collection<BattleState.Side> commitSuicide = List.of();
+
+  /**
+   * Does this unit commit suicide after it has successfully hit a target?
+   *
+   * <p>It can commit suicide as a defensive unit, offensive unit, or both.
+   */
+  @Builder.Default Collection<BattleState.Side> commitSuicideAfterSuccessfulHit = List.of();
 
   public boolean isTarget(final Unit unit) {
     return targets.contains(unit.getType());
@@ -76,20 +89,13 @@ public class CombatUnitAbility {
    *
    * <p>This is only used to merge auto created CombatUnitAbilities.
    *
-   * <p>The name is not important in deciding whether the attachedTo can be combined.
-   *
-   * <p>The attachedTo is only important if this is an suicideOnHit ability. By ensuring that each
-   * CombatUnitAbility is unique per suicideOnHit unitType, this simplifies the logic to determine
-   * which unit should suicide after a hit.
+   * <p>CombatUnitAbility with different names and attachedUnitTypes can still be merged together so
+   * the equals need to ignore those two attributes. But equals normally still needs to check the
+   * equality of names and attachedUnitTypes so don't modify the equals method itself.
    */
   public boolean canMergeAttachedUnitTypes(final CombatUnitAbility other) {
-    // build a new CombatUnitAbility based off of other but with this CombatUnitAbility's name
-    // and attachedUnitTypes (if not suicideOnHit) as they are not relevant for merging.
     return equals(
-        other.toBuilder()
-            .name(this.name)
-            .attachedUnitTypes(suicideOnHit ? other.attachedUnitTypes : this.attachedUnitTypes)
-            .build());
+        other.toBuilder().name(this.name).attachedUnitTypes(this.attachedUnitTypes).build());
   }
 
   /**
--- a/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
+++ b/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
@@ -20,6 +20,8 @@ import lombok.experimental.NonFinal;
 @Builder(toBuilder = true)
 public class CombatUnitAbility {
 
+  public static final CombatUnitAbility EMPTY = CombatUnitAbility.builder().name("Empty").build();
+
   enum DiceType {
     // Use AA dice (attackAA, offensiveAttackAA)
     AA,
@@ -33,10 +35,10 @@ public class CombatUnitAbility {
   @NonNull String name;
 
   /** The unit types that have this ability */
-  @NonFinal Collection<UnitType> attachedUnitTypes;
+  @Builder.Default @NonFinal Collection<UnitType> attachedUnitTypes = List.of();
 
   /** The unit types that can be targeted */
-  @NonNull Collection<UnitType> targets;
+  @Builder.Default Collection<UnitType> targets = List.of();
 
   /** The type of dice that will be rolled. */
   @Builder.Default DiceType diceType = DiceType.NORMAL;
--- a/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
+++ b/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
@@ -48,16 +48,6 @@ public class CombatUnitAbility {
 
   public static final CombatUnitAbility EMPTY = CombatUnitAbility.builder().name("Empty").build();
 
-  /** Which {@link games.strategy.triplea.delegate.power.calculator.CombatValue} should be used */
-  @Value
-  static class CombatValueType {
-    static final CombatValueType AA = new CombatValueType("AA");
-    static final CombatValueType BOMBARD = new CombatValueType("BOMBARD");
-    static final CombatValueType NORMAL = new CombatValueType("NORMAL");
-
-    String type;
-  }
-
   enum Suicide {
     NONE,
     /** Commit suicide after the units fire? */
@@ -97,6 +87,16 @@ public class CombatUnitAbility {
   @Builder.Default Suicide suicideOnOffense = Suicide.NONE;
   /** When should this unit commit suicide on defense */
   @Builder.Default Suicide suicideOnDefense = Suicide.NONE;
+
+  /** Which {@link games.strategy.triplea.delegate.power.calculator.CombatValue} should be used */
+  @Value
+  static class CombatValueType {
+    static final CombatValueType AA = new CombatValueType("AA");
+    static final CombatValueType BOMBARD = new CombatValueType("BOMBARD");
+    static final CombatValueType NORMAL = new CombatValueType("NORMAL");
+
+    String type;
+  }
 
   public boolean isTarget(final Unit unit) {
     return targets.contains(unit.getType());
--- a/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
+++ b/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
@@ -62,8 +62,10 @@ public class CombatUnitAbility {
 
   enum Suicide {
     NONE,
-    ALWAYS,
-    ONLY_ON_HIT,
+    /** Commit suicide after the units fire? */
+    AFTER_FIRE,
+    /** Commit suicide after the units actually hit a target? */
+    AFTER_HIT,
   }
 
   /** The name of this unit ability that will be shown in the Battle UI */
--- a/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
+++ b/game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
@@ -56,6 +56,12 @@ public class CombatUnitAbility {
     NORMAL
   }
 
+  enum Suicide {
+    NONE,
+    ALWAYS,
+    ONLY_ON_HIT,
+  }
+
   /** The name of this unit ability that will be shown in the Battle UI */
   @Nonnull String name;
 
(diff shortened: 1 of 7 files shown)


game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
```diff
@@ -0,0 +1,106 @@
...
+
+/** Unit Ability for combat situations */
+@Data
+@FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
+@Builder(toBuilder = true)
+public class CombatUnitAbility {
+
+  enum DiceType {
+    // Use AA dice (attackAA, offensiveAttackAA)
+    AA,
```
DanVanAtta, line 25: Context seems to be missing here by what is meant for 'dice'. Why is the 'dice' comment significant, and what is meant by 'AA dice'?

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
```diff
@@ -0,0 +1,106 @@
...
+import lombok.experimental.FieldDefaults;
+import lombok.experimental.NonFinal;
+
+/** Unit Ability for combat situations */
+@Data
+@FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
+@Builder(toBuilder = true)
+public class CombatUnitAbility {
+
+  enum DiceType {
```
DanVanAtta, line 23: Second point, enums can sometimes have issues when types are added in the future or removed (compatibility issues). In specific one can have issues with the 'valueOf' for types that are unknown. What are your thoughts to the future of this list and implications to compatibility issues that could come up?

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
```diff
@@ -0,0 +1,106 @@
...
+
+  /** The type of dice that will be rolled. */
+  @Builder.Default DiceType diceType = DiceType.NORMAL;
+
+  /** The side(s) that the unit needs to be on to use the ability */
+  @Builder.Default
+  Collection<BattleState.Side> sides = List.of(BattleState.Side.OFFENSE, BattleState.Side.DEFENSE);
+
+  /** How many combat rounds this ability can be used. -1 means unlimited */
+  @Builder.Default int round = -1;
```
DanVanAtta, line 49: Would 'round == 0' be useful at all?

The magic number of '-1' sticks out (SRP issues). Any thoughts to perhaps make this "Integer.MAX" (are we going to see 32 bit vs 64 bit compatibility issues if this value differs for one player or another). If the value were very large, then the at least the special case logic for '-1' would drop away and it would instead be effectively unlimited.

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
```diff
@@ -0,0 +1,106 @@
...
+  /**
+   * Can the casualties return fire after the battle phase that this UnitAbility is active?
+   *
+   * <p>Casualties that are created in the battle phase where this UnitAbility is active can always
+   * return fire since they are firing simultaneously
+   */
+  @Builder.Default boolean returnFire = true;
+
+  /** Will these units commit suicide when they have a hit? */
+  @Builder.Default boolean suicideOnHit = false;
```
DanVanAtta, line 60: Some units are suicide after attack, is that another attribute that would be added, or is it related to this attribute?

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
```diff
@@ -0,0 +1,64 @@
...
+import lombok.experimental.FieldDefaults;
+import lombok.experimental.NonFinal;
+
+/** Unit Ability to convert CombatUnitAbilities into other CombatUnitAbilities or into nothing */
+@Data
+@FieldDefaults(makeFinal = true, level = AccessLevel.PRIVATE)
+@Builder(toBuilder = true)
+public class ConvertUnitAbility {
+
+  enum Faction {
```
DanVanAtta, line 21: 'Faction' could be overloaded or confused. Consider renaming. 'Faction' is the leading contender for renaming what is known as a  'Player'

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
```diff
@@ -0,0 +1,64 @@
...
+
+import games.strategy.engine.data.UnitType;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import lombok.AccessLevel;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NonNull;
```
DanVanAtta, line 11: nit, per previously established convention,  favor using `import javax.annotation.Nonnull;` over lombok 'NonNull'

https://github.com/triplea-game/triplea/issues?q=is%3Aissue+is%3Aclosed+NonNull+

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
```diff
@@ -0,0 +1,64 @@
...
+  @NonFinal Collection<UnitType> attachedUnitTypes;
+
+  /** Does this affect enemy and/or allied units */
+  @Builder.Default Collection<Faction> factions = List.of(Faction.ENEMY);
+
+  /** The unitAbility that will be converted */
+  @NonNull CombatUnitAbility from;
+
+  /** If to is null, then this just removes the "from" UnitAbility */
+  CombatUnitAbility to;
```
DanVanAtta, line 37: Few comments here:
1. IMHO it's really useful to mark fields and parameters as `Nullable`. The IDE hints generated from those annotations really help to catch potential NPEs
2. Would a "no-unit-ability" ability make things simpler and avoid the need for a nullable combat unit ability? I could see that going both ways, I'm curious what your considerations were.
3. Is this potentially a collection, ie: a one-to-many relationship?

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import lombok.AccessLevel;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NonNull;
+import lombok.experimental.FieldDefaults;
+import lombok.experimental.NonFinal;
+
+/** Unit Ability to convert CombatUnitAbilities into other CombatUnitAbilities or into nothing */
```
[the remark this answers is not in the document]
DanVanAtta: Expounding, this topic does make me wonder if the concept of unit abilities is combat specific, generic, and how that relates to support. Are those three concepts, two, or just one concept?

For example, could we one day be asked to support an ability that makes units no-longer air (and hence targetable by say land units that could not otherwise target air?).

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
```diff
@@ -0,0 +1,64 @@
...
+   *
+   * <p>This is only used to merge auto created ConvertUnitAbilities.
+   *
+   * <p>The name and attachedTo is not important in deciding whether the attachedTo can be combined.
+   */
+  public boolean canMergeAttachedUnitTypes(final ConvertUnitAbility other) {
+    // build a new ConvertUnitAbility based off of other but with this ConvertUnitAbility's name
+    // and attachedUnitTypes as they are not relevant for merging.
+    return equals(
+        other.toBuilder().name(this.name).attachedUnitTypes(this.attachedUnitTypes).build());
```
DanVanAtta, line 50: It's not clear why we are using a builder to do an equality check. Does this mean that the two are equal if all fields match *except* the two listed here? 

If that is the case, there is a built in way to do that: `@EqualsAndHashCode(exclude = {})`

Could you elaborate on why this technique with the partial toBuilder copy is used here?

DanVanAtta (commented):
Took a look at most of the code except for the factory and test. Please review the comments, mostly relating to modeling.

One question that came to mind, are we going to wind up with completely generic and XML defined unit abilities? If so, how is AI to be able to handle that? Any thoughts as well to how AIs are going to be maintained in the phase of different abilities for new maps? For example, we would want all existing AIs to reasonably work well on most new maps and not just the latest AI working on one map (which can get into a lava-flow anti-pattern seemingly quite quickly).

Or alternatively are we going to have a predefined list of abilities where XML will instead 'opt-in' to use a certain ability. The latter I can imagine AIs being able to handle more easily, but it is less generic.

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
trevan: The support attachments use `faction` so I copied that term from there.  I couldn't think of another term to describe.  So if you have suggestions, please give them.

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
trevan: I actually went with the `EqualsAndHashCode` code before but then I realized that I actually needed those properties to be part of the equality except when trying to merge the attachedUnitTypes.  I then created a separate equals method and built out the equal checks and realized it was identical to the equal check except for those two fields.  So creating a temp object with an override of those two fields allowed me to simplify the code.

I tried to explain all of that in the comments.  It says that the "name" and "attachedUnitTypes" are not relevant when merging attachedUnitTypes.

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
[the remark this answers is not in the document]
trevan: I didn't go with `AntiUnitAbility` as I explained in the forum.  I need to be able to convert unit abilities from one to another and have that be affected by both units on the same side and on the opposite side.  `Anti` didn't make sense in that regard.

Here's some examples:
* An air unit can not hit a sub unless a destroyer is present.  So the air unit initially has an ability to target all units except the subs.  The destroyer gets a `ConvertUnitAbility` that will swap out the air unit`s normal ability with an ability that allows it to target the subs.  This is a friendly destroyer helping an air unit.
* A sub gets a first strike unless a destroyer is present.  So the sub initially has an ability to target during the first strike phase.  The destroyer gets a `ConvertUnitAbility` that will swap out the subs normal ability with an ability that makes the sub fire during the general phase.
* An AA unit has a `willNotFireIfPresent` attribute that is set to `Tank`.  So if a tank is present, the AA won't fire.  So the AA initially has an ability to fire during the AA phase.  The tank gets a `ConvertUnitAbility` that will remove the AA's ability so that it doesn't do anything.

> So bottom line, it's not fully clear to me how many unit ability concepts we have and which ones would fall into which categories. I was hoping we could spend some time discussing and thinking this through so we have our mental and object models more coherent and thought out.

Do you want this discussion in the forum?  Do you want to start it?

> Expounding, this topic does make me wonder if the concept of unit abilities is combat specific, generic, and how that relates to support. Are those three concepts, two, or just one concept?

I don't know how to answer that since I don't know how you are conceptualizing those concepts.  It looks like you might be envisioning these "unit abilities" in a much larger context than I'm envisioning.

> For example, could we one day be asked to support an ability that makes units no-longer air (and hence targetable by say land units that could not otherwise target air?).

This is already possible using AA `targetsAA` or non-AA `canNotBeTargetedBy` and `canNotTarget`.  This would just unify how those are handled.

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
trevan: > Would a "no-unit-ability" ability make things simpler and avoid the need for a nullable combat unit ability? I could see that going both ways, I'm curious what your considerations were.

That would probably work.  I don't currently see it causing an issue once the battle logic is worked out.  We might need to revisit this as that is worked on.

> Is this potentially a collection, ie: a one-to-many relationship?

For the built-in ConvertUnitAbilities, no.  But there might be a map that could make use of a one-to-many.  Do you want me to make it one-to-many?

trevan:
> One question that came to mind, are we going to wind up with completely generic and XML defined unit abilities? If so, how is AI to be able to handle that? Any thoughts as well to how AIs are going to be maintained in the phase of different abilities for new maps? For example, we would want all existing AIs to reasonably work well on most new maps and not just the latest AI working on one map (which can get into a lava-flow anti-pattern seemingly quite quickly).

I don't think the AI handles the new `canNotBeTargetedBy` and `canNotTarget` (at least I can't find any code).  I also don't think it handles `targetsAA` (such as sending units that can actually target the units).  It does look like it checks for destroyers and subs, though.

But since these unit abilities affect combat and the battle calculator will handle them, I think the AI will do decent.  It will still be able to tell if the battle is won or loss.

The AI could be improved so that it can check for abilities that limit units (such as abilities like `canNotTarget`) or abilities that are helpful (such as abilities like `canNotBeTargetedBy`).  I think the code would probably be simpler with this generic mechanism instead of handling all of the different attributes and how they interact with each other.

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
DanVanAtta: That might be a question for the forum. I could argue it both ways, better to do it now, but also would be a shame to over-complicate it with a one-to-many when it is always one-to-one. When we get to XML, we'll want that spec to land just once. That will be a time when we are far more locked in and the decision there then is more significant.

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
DanVanAtta: Side and team are best i can see from a thesaurus: https://www.thesaurus.com/browse/faction

I agree `faction` fits, though I would like that reserved for the time being as the replacement for `player`. Once that happens, it would be confusing to see it used in a different way.

Perhaps some ideas may come to us if we sleep on it.

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
DanVanAtta: Could you send me a link to refresh my memory? I've my hand in many TripleA pots, lots going on.

I think I'm just confused as those seem like combat abilities, `isSuicideOnAttack` & `isSuicideOnDefense`, and I'm not sure how that is different then from `isSuicideOnHit`, and why only the latter is represented here but no the other two. Can you suggest a way to make this more clear to me and other future maintainers reading through this code?

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
DanVanAtta: I didn't realize you were retrofitting this to existing values. There are a number of original design decisions that would be great to change, but we are stuck with them.

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
DanVanAtta: I like enums, but we have been bitten before. An 'unknown' or 'other' enum type can work as well. I don't have enough context to know which would work better. Typically enum's are good when you know the full set and it will never change (so perhaps argument for a string type, or maybe better yet a value object that wraps a string)

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
trevan: Relationship might work.  Side already has a meaning of "offense" and "defense" so I wouldn't want to use it here.

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
trevan: `isSuicideOnAttack` and `isSuicideOnDefense` means that the unit will have one chance to fire and then will commit suicide whether it hit or not.  `isSuicideOnHit` means that the unit will fire every round and it only commits suicide when it has a successful hit.

Now that you mention it, `isSuicideOnAttack` and `isSuicideOnDefense` should probably be represented as a unit ability.  I`m thinking `Collection<Side> commitSuicide`.  And maybe I should chance `suicideOnHit` to `Collection<Side> commitSuicideAfterSuccessfulHit`.

I also just checked and I can't find a map that is using `isSuicideOnHit` so I'm not sure how it would be used.

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
trevan: I see this work as unifying how unit abilities work between bombard, aa, firststrike, and general phase.  There are several ways to indicate targets and other abilities and so I'd like to unify it both in the engine and in the xml.  So all the existing properties still need to be handled with these new abilities so that the engine only has to care about the new abilities and not the old properties.

But I could probably change the old `-1` to be INTEGER.MAX when converting the property.

trevan: Done in https://github.com/triplea-game/triplea/pull/8646/commits/a6ea214333c993baefa8d1a9a5d95038354190f6

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
trevan: Done in https://github.com/triplea-game/triplea/pull/8646/commits/6b79c49cd64666026b410bb54293f8daa3410537

Also, since `isSuicideOnHit` isn't used in any map, I'm going to change its functionality a little bit.  Currently, the engine splits up all of the `isSuicideOnHit` unit types so that when a hit occurs, it can just kill one of the units since it knows that all of the units are the same unit types.  But I'm going to change it so that`isSuicideOnHit` unit types can be combined and it will just randomly pick out a unit to suicide.  For non-low luck, it will handle the case where a unit type might have different dice sides.  For low luck, it will just randomly select units.

I think the functionality is similar enough that it should be ok.

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
trevan: Done in https://github.com/triplea-game/triplea/pull/8646/commits/3ef07a4f3fe1e9ae6956a97806b33b0201254724

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
trevan: Affiliation is another possible word.

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
trevan: I've added an EMPTY unit ability and set that as the default - https://github.com/triplea-game/triplea/pull/8646/commits/bb8bc0e7b7d2329facd7ddc1f0744040c3ab0cc9

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
DanVanAtta: Couple more suggestions:
`team { friendly | foe }`
`FriendOrFoe { friendly | foe }`

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
```diff
@@ -0,0 +1,114 @@
...
+  @Builder.Default boolean returnFire = true;
+
+  /**
+   * Does this unit commit suicide after it has fired a shot?
+   *
+   * <p>Whether the shot hit or not doesn't matter.
+   *
+   * <p>It can commit suicide as a defensive unit, offensive unit, or both.
+   */
+  @Builder.Default Collection<BattleState.Side> commitSuicide = List.of();
```
DanVanAtta, line 68: I was thinking about this one, the commit suicide on offense, defense, or after a hit or any roll might make for a good enum to flag which cases.

Regardless, the variable name here sounds more like a boolean rather than a list. I'm not sure if there is a better name, but could be worth a second look.

DanVanAtta (commented):
Added a few new comments, just a few things to consider. I'll hopefully spend more time reviewing tomorrow and get at least all the way through the PR : )

Thanks as always for the efforts @trevan

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import lombok.AccessLevel;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NonNull;
+import lombok.experimental.FieldDefaults;
+import lombok.experimental.NonFinal;
+
+/** Unit Ability for combat situations */
```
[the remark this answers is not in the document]
trevan: I added some examples in https://github.com/triplea-game/triplea/pull/8646/commits/38298c2d53c02183b36698e06226855acc539cc3.  Do those work?

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
```diff
@@ -0,0 +1,114 @@
...
+    BOMBARD,
+    // Use normal dice (attack, defense)
+    NORMAL
+  }
+
+  /** The name of this unit ability that will be shown in the Battle UI */
+  @NonNull String name;
+
+  /** The unit types that have this ability */
+  @Builder.Default @NonFinal Collection<UnitType> attachedUnitTypes = List.of();
```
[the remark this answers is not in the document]
trevan: This class will also be serialized as well during battles.  So it isn't because of serialization.

The reason is that each player can have different unit abilities for the same unit type.  But unit types are player agnostic.  So, I either need to make the unit abilities store the player their attached to or store the unit type that they are attached to.  I went with the later because I could put the player <-> unit ability mapping in the battle phase and not have to touch the player class.

I didn't call it `unitTypes` because `targets` is also a collection of `UnitType`s and thought that having one member called `unitTypes` and another called `targets` would be confusing since both are just collection of `UnitType`s.  I wanted to be more explicit in the name to ease identification.

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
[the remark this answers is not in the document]
DanVanAtta: > Do you want this discussion in the forum? Do you want to start it?

I'm not a fan of how we have discussions split between forum and git. So far it's basically the more code focused topics are in git, and conversations that involve map makers and more are in forums (in part because the latter only check forums). IMO it does not matter too much where, I do think the conversation is more code focused as it is more about modelling, how the concepts are modeled out to be mostly transparent. To the extent that model is exposed in XML, that is where the larger community would want to be involved for sure.

I don't quite have enough context and a vision of where these updates are going to start that conversation. Said another way, this is a place where I think good up-front design will be beneficial.

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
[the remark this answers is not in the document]
DanVanAtta: looks good, thank you

game-core/src/main/java/games/strategy/engine/data/unit/ability/UnitAbilityFactory.java
```diff
@@ -0,0 +1,456 @@
...
+   * <p>This map is used to invert that data set at the beginning so that it can be easily looked up
+   * later
+   */
+  private static final Map<UnitType, Collection<UnitType>> unitCanNotTargetWithoutDestroyer =
+      new HashMap<>();
+
+  /**
+   * Create the default unit abilities at the beginning of the battle step
+   *
+   * <p>This needs to be called once at the beginning of each battle step. All the battles in the
```
DanVanAtta, line 63: Noting the use of static state here. I'm very concerned this could get us into trouble. Any thoughts to make this stateful and by-battle?

trevan: When you say "by-battle", you want this logic to be run at the beginning of each battle?  It isn't needed.  This is static information at the beginning of the entire battle phase and re-doing it for each battle would be unnecessary.

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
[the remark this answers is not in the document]
trevan: My vision is really just to unify how units target.  All of the pieces that I've modeled in the unit abilities (and the old firing squadron pr) is only focused on that.  I want to simplify the battle logic instead of having 4 special hardcoded fight phases, instead it can just loop through a list of phases and ask each phase how to handle the combat.

So support and other abilities (such as blitz) are outside the scope in my vision.  Could they be merged into one unified idea?  Maybe.  Same with retreat, withdraw, and paratroop landing (which are part of a battle but not related with targeting).

game-core/src/main/java/games/strategy/engine/data/unit/ability/UnitAbilityFactory.java
DanVanAtta: Didn't notice you responded. My concerns:
1. We're relying on establishing static state, which depends on call ordering,  it is inherently brittle. 
2. Won't work when we have multiple games per process

I'm not sure what the best way to restructure would be to avoid the static state. Perhaps a transient game data member variable could do the trick. Another thought, is the battle phase represented as an object in any way? Perhaps that could have the logic to lazy compute the needed unit ability information.

trevan: @DanVanAtta, no problem with the late reply.

The `BattlePhaseList` will be a member variable of something (probably `GameData`) and it will actually store the state.  This class is just converting the existing attributes into unit abilities.  So it really isn't state.  But there do happen to be two static private members I added to simplify some of the logic and those will cause problems with multiple games per process.  So I'll need to pass that data around instead of creating a private static member.

Also, if/when map makers can create custom unit abilities, then this class will not be used.  The xml parser will need to error out if it finds both custom unit abilities and the deprecated properties.  It could possibly use this class to give helpful hints on how to convert the deprecated properties.

DanVanAtta: Will this comment change perhaps?

> This needs to be called once at the beginning of each battle step

I have also been wondering if even in an intermediate form if that would work. What if for example a battle is persisted at an unexpected time, would we still be able to guarantee all values are instantiated properly?

trevan: No, I don't see that comment changing unless the existing code flow changes even more drastically.  The `BattlePhaseList` will most likely be stored as part of the save game so when the battle starts up again, it will restore it.  And if the code changes enough so that `BattlePhaseList` isn't part of the save game, when the save game is restored, it will be initialized when the Battle Step code is restored.

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
trevan: I've changed `DiceType` to a string value object - https://github.com/triplea-game/triplea/pull/8646/commits/be242593d3e6af671eb33c661fb8590c34264ecd

game-core/src/main/java/games/strategy/engine/data/unit/ability/ConvertUnitAbility.java
trevan: I renamed it to Team - https://github.com/triplea-game/triplea/pull/8646/commits/a75da36aa7e742dd907b21ad0eef9d98e6a3ebc8

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
trevan: I've renamed them and added an enum.  Take a look at https://github.com/triplea-game/triplea/pull/8646/commits/4fe6eda8477965dbf1c05ef4901f35e18dce612d

game-core/src/main/java/games/strategy/engine/data/unit/ability/UnitAbilityFactory.java
trevan: I've removed the static members and now pass the needed state around - https://github.com/triplea-game/triplea/pull/8646/commits/4773376d77361c893b0d5d2490104bcd852c4921

trevan: > Will this comment change perhaps?
> 
> > This needs to be called once at the beginning of each battle step

Also, as long as we have to support old map xmls, this code will need to exist.  If there aren't custom unit abilities, then this code has to be called so that it can convert the old properties into the unit abilities.  It is just adding these unit abilities to the `BattlePhaseList`. If custom unit abilities exist, then the `BattlePhaseList` will already be set up correctly.  Saving and loading of the `BattlePhaseList` is independent of this code since this is just conversion code.

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
trevan: I'm currently not envisioning the unit abilities to specify the dice properties (strength and rolls).  That information would stay on the unit itself.  I'm also not envisioning changing the existing names of the dice properties (`attack`, `defense`, `attackAA`, `offensiveAtackAA`, `bombard`, `airAttack`, `airDefense`).  That's why I have this `DiceType` value class with only three initial items.

But, if we want to make changes to the dice properties, then this class will need to be reworked.  I don't think the dice properties should be on the `CombatUnitAbility`.  That would make it hard to share `CombatUnitAbility` between units.

I also thought about having this `DiceType` be an aspect of the `BattlePhase` instead of the `CombatUnitAbility`.  The existing battle phases and combat unit abilities act like that (all combat unit abilities in the AA phase use the AA dice type).

DanVanAtta: Thank you for the context @trevan . I think I will struggle for some time with the term 'DiceType' as cognitively it does not represent at all in TripleA what I would think it should mean. To be clear, dice type simply refers to the number of sides, or does it also refer to the number of rolls?

DanVanAtta: side-comment, if we can come up with better terminology, it would likely help, I think there is a terminology and model mismatch that makes this confusing.

game-core/src/main/java/games/strategy/engine/data/unit/ability/UnitAbilityFactory.java
DanVanAtta: > Also, as long as we have to support old map xmls, this code will need to exist. 

FWIW, the design of the map-data project is such that it provides a unified data interface to the rest of the game. Any legacy XML attributes (or lack there-of), should be handled within that layer. This makes it transparent to the rest of the engine which XML attributes were present exactly.

DanVanAtta: If this is conversion code, does it actually more properly belong in the parsing layer? There are ways to alias XML properties and/or straight up translate them from one to another. The intent is that map parsing can give a relatively idealized data interface and can "add-in" tags or translate tags from one property to another, or combine multiple tags to set the same data.

game-core/src/main/java/games/strategy/engine/data/unit/ability/UnitAbilityFactory.java
```diff
@@ -0,0 +1,456 @@
...
+  private static CombatUnitAbility addAbility(
+      final BattlePhaseList battlePhaseList,
+      final CombatUnitAbility combatUnitAbility,
+      final GamePlayer player,
+      final String phaseName) {
+    final BattlePhase battlePhase =
+        battlePhaseList
+            .getPhase(phaseName)
+            .orElseThrow(
+                () -> new IllegalArgumentException("The phase " + phaseName + " doesn't exist"));
```
DanVanAtta, line 204: Printing the list of battle phase names in the exception message could make for a better error message, in case there are typos or some small differences, it'll be easier to spot.

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
trevan: > To be clear, dice type simply refers to the number of sides, or does it also refer to the number of rolls?

It refers to the strength and the rolls.

In the current engine, when you have AA guns, the dice properties are `attackAA`, `attackAAmaxDieSides`, `offensiveAttackAA`, `offensiveAttackAAmaxDieSides`, and `maxAAattacks`.  When units bombard, the dice properties are `bombard` or `attack` if `bombard` is missing.  Outside of AA and bombard, units use `attack`, `defense`, `attackRolls`, and `defenseRolls`.

I needed a way to indicate which of those sets of properties should be used in the `CombatUnitAbility`.  So I created 3 groups and called them `AA`, `BOMBARD`, and `NORMAL` and described them as `DiceType`.

DanVanAtta (commented):
The comment about where legacy attribute values are converted is salient, but I'm okay if we come back and follow up with it after you've had time to consider. 'map-data' is designed to entirely handle legacy attributes and convert them up to new so that the game engine only needs to deal with 'new' attributes and a latest XML  spec (and not all permutations of all legacy types).

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
DanVanAtta: Perhaps it could be called 'StrengthAndRolls'? We have that value object and it then seemingly is a more direct name. 

Dice type for me refers either to the number of dice sides or is seemingly a different concept altogether more related to perhaps phases, or even things like technology roll vs a battle roll.

game-core/src/main/java/games/strategy/engine/data/unit/ability/UnitAbilityFactory.java
trevan: I have thought about putting it in the parsing layer but decided not to because of the increase in complexity.  All of the properties that this code depends on can change with triggers.  And some of the properties are technology dependent so players can have different properties depending on what technologies they've researched.  I would have to write code that would run when a trigger happens or a technology is researched.  This code would have to rewrite the unit abilities depending on what changed.  That felt really complex compared to what I have now.

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
trevan: It is more similar to the `CombatValue` value objects.  There are `AaDefenseCombatValue`, `AaOffenseCombatValue`, `BombardmentCombatValue`, `MainDefenseCombatValue`, `MainOffenseCombatValue`.  `DiceType` is used to decide which of the CombatValue classes should be used (AA, Bombardment, or Main).  So maybe call this `CombatValueType`?

game-core/src/main/java/games/strategy/engine/data/unit/ability/UnitAbilityFactory.java
DanVanAtta: Okay, fair enough. You're aware of that design element and it not working out is what it is. :+1:

game-core/src/main/java/games/strategy/engine/data/unit/ability/CombatUnitAbility.java
DanVanAtta: `CombatValueType` or `CombatValue` both seem like improvements to me FWIW

trevan: I've renamed it to `CombatValueType` and added a link to `CombatValue` in the javadoc - https://github.com/triplea-game/triplea/pull/8646/commits/75fb84b36a20a193f5e58c70e2d4adc2f37af77d

DanVanAtta:
@trevan from what i can see remaining:
- https://github.com/triplea-game/triplea/pull/8646#discussion_r563186370
- one code climate violation for a method that is slightly too long

trevan:
@DanVanAtta the last two commits should take care of those things.

trevan:
Apparently I triggered another code climate issue.  I'll see what I can do.

trevan:
@DanVanAtta , I fixed the code climate issue.

DanVanAtta (approved):

PR #8653, opened 2021-01-17
Remove gradle by-default parallel

DanVanAtta:
While we get faster builds with gradle parallel, output can
be hopelessly difficult to look at when tasks are executed
in parallel. This update favors debug-ability over execution
speed. (A dev can still add '--parallel' when they wish to
have a parallel build)


DanVanAtta:
FWIW, if we can fix the test output problem, IMHO this very well could be worth adding back in.

PR #8685, opened 2021-01-24
Update triplea_maps.yaml

kevinkunker:


kevinkunker:
This is to fix the thumbnail image URL

PR #8700, opened 2021-01-25
Mark getMapVersionByName as non-static

DanVanAtta:
Preps for future changes where access to downloaded maps
will be stateful. This update marks getMapVersionByName
as non-static and updates callers in preparation for this.


DanVanAtta:
Resolving, will loop in this update with another change.

Issue #8679, opened 2021-01-23
Map extraction problems (Folder Chaos)


RogerCooper:
The latest version of TripleA(22583) converted the zip files into folders in the downloadedmap folder. Many zip files were not converted and were simply destroyed, not even in the recycle bin. (I have backups). For example, my file AA50-41-Maintenance.zip [link](https://drive.google.com/file/d/1sk85nOG38oDKGElgpxlvYj0bIDGBxr7d/view?usp=sharing) was destroyed.

I understand the advantages to using unzipped folders, but there is no reason that the actual zip files should have deleted. Even worse, the deleted files were sometimes not unzipped before deletion.

I suggest a rollback until this issue is resolved.
[AA50-41-Maintenance.zip](https://github.com/triplea-game/triplea/files/5860377/AA50-41-Maintenance.zip)

DanVanAtta:
Were they moved to a 'bad-zip' folder @RogerCooper ?

DanVanAtta:
The map posted, 'AA50-41-Maintenance.zip', @RogerCooper is not well structured. The first folder in the zip should be a single folder typically matching the map name (EG: AA50-41-Maintenance). The next folder should be called 'map', then under that there would be map files.

What is the background on this map zip file?

DanVanAtta:
@RogerCooper more information and example zips would be really helpful.

RogerCooper:
There is no folder called bad-zip. There is nothing wrong with the file I posted either, I was still using the older file structure which was working fine until the new release. Are double folders now required?

RogerCooper:
I converted AA50-41-Maintenance.zip to the double folder format. It comes up now, with 2 entries in the scenario list for some reason, both of which work. Where are the zip files going? Programs should not delete files from user's computers without permission.

I think it would be better if zip files remained. If the zip file had a more recent date than the folder, the user should be given the option of updating. Consider the problem created if 2 different maps had the same file name. Would the program as it is now, overwrite the existing data without warning and then delete the zip file?

DanVanAtta:
A problem with zip files remaining is the engine will have trouble knowing whether to re-extract it again. That implies the engine would have to examine every zip file every time you select a game.

DanVanAtta:
Extracted zips could perhaps go into an extracted-zips folder so that the engine would know to not look at them. The excessive disk space usage is a concern as the zip files are really not needed post-extraction.

[...]

PR #8712, opened 2021-01-27
Update map zip extraction to extract to a folder

DanVanAtta:
Rather than unzipping map zips directly into the download folder,
this update changes that behavior to create a folder first that
is named after the map zip.

Most map zips have a folder inside of map wherein is all content.
Other zips, notably manually packaged and older zip files instead
have all contents zipped. When extracting these zips their contents
would otherwise be exploded into the downloaded maps folder.

By creating a folder first and extracting there, these latter
type of zips will be more cleanly extracted.

Addresses: https://github.com/triplea-game/triplea/issues/8679

## Testing

- Verified manually extraction of zip files and a .properties file was renamed accordingly.

## Additional Notes

Extracting into a folder lead to the logic being a bit simpler. Rather than having multiple files land in a temp folder and moving files one-by-one from the temp folder into the target location, we instead just move the temp folder to the target location and rename the temp folder.

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/ZippedMapsExtractor.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/ZippedMapsExtractor.java
@@ -1,10 +1,10 @@
 package games.strategy.engine.framework.map.file.system.loader;
 
+import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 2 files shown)


DanVanAtta:
Marking as a draft, the extra sub-folder when a map zip is zipped at the folder level causes problem with resource loader making this not backward compatible..

PR #8721, opened 2021-01-28
Update peview.png for Global 40 House Rules

beelee1:
Not real confident but will give it a shot anyway

## Testing

Reinstalled map and it started.

## Screens Shots

This is what it looks like now
Screenshot from 2021-01-27 23-34-13

Trying to update according to this

https://forums.triplea-game.org/topic/2603/standard-preview-png-image

The Neutral_Navy.png was in the game notes. I left it in the doc/images folder too. Not too sure about that but ...

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2019,7 +2019,7 @@
 - mapName: Global 40 House Rules
   mapCategory: EXPERIMENTAL
   url: https://github.com/triplea-maps/global_40_house_rules/archive/master.zip
-  version: 13
+  version: 14
   img: https://raw.githubusercontent.com/triplea-maps/global_40_house_rules/master/description/TripleA_ww2_global_1940_mini.png
   description:
     <br>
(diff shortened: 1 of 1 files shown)


triplea_maps.yaml
```diff
@@ -2019,7 +2019,7 @@
 - mapName: Global 40 House Rules
   mapCategory: EXPERIMENTAL
   url: https://github.com/triplea-maps/global_40_house_rules/archive/master.zip
-  version: 13
+  version: 14
   img: https://raw.githubusercontent.com/triplea-maps/global_40_house_rules/master/description/TripleA_ww2_global_1940_mini.png
```
DanVanAtta, line 2023: You only need to update the thumbnail image here, the 'img' attribute, specifically to:

`img: https://raw.githubusercontent.com/triplea-maps/global_40_house_rules/master/preview.png`

The version can stay the same, there is no impact on the map. This version value, in this file, controls whether players are notified to re-download the map. They can stay on the same version as we are just updating the preview image displayed in the download window. 

Let me know if you could use any help updating this PR.

beelee1: I'm hitting all sorts of wrong buttons lol. So idk why it's not showing up but just below the closed thing 074c5db the purple thing to the left is the fix. I think it should work. fwiw : )

DanVanAtta:
Good work  :smiley:

PR #8733, opened 2021-01-31
Adding preview.png to Oztea 1942

beelee1:
## Testing

Reinstalled and map works

Adds preview.png to oztea_1942


DanVanAtta:
Thanks for tackling these and helping!

FYI, re testing: Checking that the map works should be unnecessary. If the image is 'moved' from 'doc/images', then you'll want to check that the notes work (and in that case you should copy the image). Otherwise if the image is renamed from the 'description' folder, then you only need to check the download preview in 'download games'. You can also test by checking that the update link in this type of PR still works (ie: you can access it in a web browser and you see the image and not a 404)

PR #8738, opened 2021-02-01
Build Installers with Github Actions

DanVanAtta:
- Use github actions to build installers
- Stop using Travis CI for installer builds
- Updates documentation to describe updated release process

Changes:
 - Notable: To release, instead of marking an existing prerelease
   as a release, instead we create a 'release/<version>' branch.
   Builds from 'release/*' branches will be built as releases.
   Builds from 'master' will be built as a prerelease.

 - Major: There will now be a single prerelease build, this reduces
   the number of tags being pushed and we will no longer
   have a unique tag for every build.

 - Minor: Run number is now appended to the version value in
   product.properties instead of replacing the trailing '.0'.
   This is to make the version value replacement less brittle,
   a person incrementing the version value no longer needs to
   know that it must end with '.0' to work with the build.

 - Minor: '.pre' will now be suffixed to prerelease build versions

## Testing

Verified builds against my fork: https://github.com/DanVanAtta/triplea/releases

- Checked that if a prerelease tag DNE, then deleting it is a no-op
- [x] TODO:  check that a release branch is built as a release 
- [x] TODO:  double check non-release and non-master branches are not built
- [x] TODO: double check PRs do not create releases
 
## Additional

- [x] TODO: add github action secrets so they are available to triplea-game

### DIFF
--- a/.build/build-installer
+++ b/.build/build-installer
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+## Builds TripleA game-installers and build artifacts suitable
+## for distribution.  Build artifacts will be created for each
+## sub-project and copied into: './build/artifacts'
+
+## install install4j
+set -x
+
+function main() {
+  install_install4j
+  build_installers
+  collect_artifacts
+}
+
+# Installs install4j, the license key is injected into install4j during this step.
+function install_install4j() {
+  INSTALL4J_HOME=/tmp/install4j
+  mkdir $INSTALL4J_HOME
+
+  echo "Downloading and installing install4j to '$INSTALL4J_HOME'"
+  wget --no-verbose -O install4j_unix.sh \
+  https://raw.githubusercontent.com/triplea-game/assets/master/install4j/install4j_unix_8_0_7.sh
+  chmod +x install4j_unix.sh
+  ./install4j_unix.sh -q -dir "$INSTALL4J_HOME"
+  "$INSTALL4J_HOME/bin/install4jc" -L "$INSTALL4J_LICENSE"
+}
+
+
+## Runs gradle command that creates the installer executables, uses intall4j
+function build_installers() {
+  JAVA_OPTS=-Xmx4G ./gradlew \
+      --no-daemon \
+      --parallel \
+      -Pinstall4jHomeDir="$INSTALL4J_HOME" \
+      release
+}
+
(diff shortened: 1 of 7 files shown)


RoiEXLab (approved):
Interesting approach to this problem, not necessarily my first choice, but I'm fine with the changes 👍

DanVanAtta:
What would have been your first choice?

Re cleanup, I'm leaning to removing unused travis files in one go. The
deployment actions and auto-issue creation for failed builds are all that
is left.

RoiEXLab:
> What would have been your first choice?

Hard to say what exactly would've been my first choice, but I think it's enough to tell you about the things I wouldn't have considered doing this migration:
I really like the idea of having build artifacts for every iteration. In the past this was an easy way for users to figure out when exactly something broke, which made it easier to nail some issues down. So I would've kept this feature.
Also I think having a simple "release/pre-release" switch makes it convenient to push a release, it's more or less instantly and the only thing left is to update release notes etc.
Creating a release branch obviously isn't hard either, but it's an extra step and something that (AFAIK) can't be done via the web ui.

DanVanAtta:
> I really like the idea of having build artifacts for every iteration. In the past this was an easy way for users to figure out when exactly something broke, which made it easier to nail some issues down.

My recollection is this only happened once or twice. Usually the bisection is down to developers. The intent is that we will release more regularly, the release once a year or two should be considered abnormal and really as a failure.

Meanwhile having a huge multitude of tags is not great, reduces the value of tags and clutters up the branch list. 

> Also I think having a simple "release/pre-release" switch makes it convenient to push a release

Agree, though it is still there to some extent since any release can still be toggled on or off. Arguably it's just more automated now.

> Creating a release branch obviously isn't hard either, but it's an extra step and something that (AFAIK) can't be done via the web ui.

In theory this should be done as part of a release anyways so that we have an easy way to create hotfixes and/or just check out the latest of any specific release.

DanVanAtta:
There was a benefit to having each tag correspond to one sha whereby given a version we were able to know which SHA was used to build that version. I pushed an update: https://github.com/triplea-game/triplea/pull/8738/commits/43556d5a343306dd7418917390658e49e7243eb3, that instead of suffixing prerelease versions with 'pre', we'll suffix with the SHA for that build.

PR #8740, opened 2021-02-01
Adding haveResources to condition statements

WCSumpton:
Changes are only for RulesAttachment other files can be removed. I don't seem to be able to remove them.

Adds haveResources to condition statements, with the ability to sum/add different resources and to include other players resources.

Testing was done using the Classic game and adding checks for techTokens and PUs. This add changes no game data and should be compatible with older xmls.


DanVanAtta:
Could you provide sample XML snippet @WCSumpton using this?

WCSumpton:
@DanVanAtta to check to see if the Russian player has 24 or more PUs:
~~~
<attachment name="conditionAttachmentHaveTechTokens" attachTo="Russians" javaClass="games.strategy.triplea.attachments.RulesAttachment" type="player">
   <option name="haveResources" value="PUs" count="24"/>
</attachment>
~~~
Including the Allies by adding:
~~~
   <option name="players" value="Americans:British"/>
~~~
And to sum together the values:
~~~
   <option name="haveResources" value="Sum:PUs" count="24"/>
~~~
Hope this is helpful.

DanVanAtta:
Yes. Can you comment if we can use more normalized XML? I know it could be a bit to ask though I think it could be worth it. Notably, once the new XML is used in production we will have to continue to supporting it. Hence, if we change it, we'll have multiple formats to support which could make it confusing which one to use and adds complexity in the code.

For players, that would look something like:
```
<option name="players">
  <value>Americans</value>
  <value>British</value>
</option>
```
Arguably the above would be even better if it were a `players` and `player` tag which would help for validation to ensure there are no typos in the player names.


For resource type, there are options, I suppose ideally something like:
```
   <resource type="PUs" quantity="24" operation="Any|Sum"/>
```

WCSumpton:
Removing this for further work

PR #8743, opened 2021-02-02
Migrate release build actions: map push and prerelease deployment

DanVanAtta:
Migrates from Travis to Github actions deployment of latest to
prerelease servers and secondly the update of the map listing
on website.


RoiEXLab (approved):
I didn't expect this migration to work with so few changes

PR #8747, opened 2021-02-02
Fix: update player-in-game status when leaving an in-progress game

### LINKED ISSUE #7804: player info is wrong
when looking at a players info, it showed he was in a bot (201), but there was no one in the bot.  and another player said he noticed that also. 

Game Version:

- [] yes
- [] no

DanVanAtta:
Resolves #7804 where the player info for "player is in games" was incorrect.
The notification to lobby was being short-circuited for in-progress games,
therefore players when joining those games where from then-on listed as
present in that game.

This fix moves the lobby notification that player has left to be before
the short circuit.

## Testing
- tested with a local lobby, connected with two different players and verified menu was updated after joining an in-progress game and then leaving.

## Release Note

FIX|Lobby player info right click menu corrected to properly reflect when players have left an in-progress game

### DIFF
--- a/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
+++ b/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
@@ -619,12 +619,17 @@ public class ServerModel extends Observable implements IConnectionChangeListener
     if (gameToLobbyConnection != null && lobbyWatcherThread != null) {
       lobbyWatcherThread
           .getGameId()
-          .ifPresent(gameId -> connectionAndGameIdAction.accept(gameToLobbyConnection, gameId));
+          .ifPresent(
+              gameId ->
+                  new Thread(() -> connectionAndGameIdAction.accept(gameToLobbyConnection, gameId))
+                      .start());
     }
   }
 
   @Override
   public void connectionRemoved(final INode node) {
+    notifyLobby(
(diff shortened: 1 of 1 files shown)


game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
```diff
@@ -633,8 +638,6 @@ public void connectionRemoved(final INode node) {
       serverLauncher.connectionLost(node);
       return;
```
DanVanAtta, line 639: This return statement was the culprit, for in progress games `serverLauncher != null`

game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
```diff
@@ -619,12 +619,17 @@ private void notifyLobby(
     if (gameToLobbyConnection != null && lobbyWatcherThread != null) {
       lobbyWatcherThread
           .getGameId()
-          .ifPresent(gameId -> connectionAndGameIdAction.accept(gameToLobbyConnection, gameId));
+          .ifPresent(
+              gameId ->
+                  new Thread(() -> connectionAndGameIdAction.accept(gameToLobbyConnection, gameId))
```
DanVanAtta, line 624: Made the lobby notification non-blocking and now runs on a background thread.

RoiEXLab (approved):

PR #8748, opened 2021-02-02
Fix banned username matching

DanVanAtta:
The check for username bans attempted to do a case insensitive match
by lower casing the name to check. But, the names stored where in whatever
casing was originally entered. This implies the username ban only worked
when the original casing was all lower casing.

This update fixes the matching so it is case insensitive, banned usernames
are upper-cased when entered into database and then the name to match is also upper-cased.

## Release Note

This will go out with the next lobby release.

### DIFF
--- a/database/src/main/resources/db/migration/lobby_db/V2.01.00__fix_username_ban_casing.sql
+++ b/database/src/main/resources/db/migration/lobby_db/V2.01.00__fix_username_ban_casing.sql
@@ -1,0 +1,11 @@
+-- delete any extra name bans that are duplicates
+delete from banned_username bu
+where exists (
+  select 1
+  from banned_username t2
+  where bu.date_created < t2.date_created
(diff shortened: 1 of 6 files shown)


database/src/main/resources/db/migration/lobby_db/V2.01.00__fix_username_ban_casing.sql
```diff
@@ -0,0 +1,11 @@
+-- delete any extra name bans that are duplicates
+delete from banned_username bu
+where exists (
+  select 1
+  from banned_username t2
+  where bu.date_created < t2.date_created
+    and bu.username ilike t2.username
```
RoiEXLab, line 7: Didn't know about this `ILIKE` extension from postgres, neat

RoiEXLab (approved):

PR #8780, opened 2021-02-04
Update all preview images to be 'preview.png'

DanVanAtta:
After having downloaded all preview images and uploaded them to each
map repository as 'preview.png' in the top-most folder, this update
changes triplea_map.yaml 'img' attribute to point the preview
image to this file.

With a standard file name we will be able to have the game-client
download the preview.png directly if it knows the URI of the map repo.
This also puts us in a place where map makers can define their preview
image and bundle it with their map.

## Testing
- Did some spot checks of the updated images.
- `gradlew validateYamls` should be covering us, it validates all links are valid.

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -57,7 +57,7 @@
   mapCategory: BEST
   url: https://github.com/triplea-maps/napoleonic_empires/archive/master.zip
   version: 3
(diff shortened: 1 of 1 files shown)


DanVanAtta:
cc: @beelee1 

Any extra spot-checks that the updated links work and look good are appreciated!

PR #8805, opened 2021-02-07
Merge unit and integ tests

DanVanAtta:
commit 93d3dd9397dfdffe758f6f1f190e30f02a33929a

    Remove '@Intregation' and '@ExtendedUnitTest' tags, run integ tests with unit

commit ba7b453478b614806e4da785bb2487b8431a8fab

    Merge unit & integration testing build steps
    
    - Also cleanup upload of codacy coverage reports. We were doing
    so conditionally and were not due to lack of token being defined.


.build/unit-testing/run
```diff
@@ -1,36 +1,22 @@
 #!/bin/bash
 
-set -ux
+set -u
 
 function main() {
-  runTests
+  .build/setup-database
+  ./gradlew --quiet test jacocoTestReport --exclude-task :smoke-testing:test
```
RoiEXLab, line 7: @DanVanAtta Could you explain why you decided to add the --quiet flag here?
When a test fails this flag causes it to not show the test that actually failed, which makes issues really hard to debug that don't happen locally

Issue #8828, opened 2021-02-09
Ready to Add New Map


axisalliescontango:
I have a playable map that is ready to be shared with the TripleA community.  Please send me an invitation to join my own map team at triplea-maps.

https://github.com/axisalliescontango/young_grasshoppers_tournament_edition

DanVanAtta:
Maps to be joined to triplea-maps neither need to be playable nor require a map admin team:
- maps in development are welcome
- github used to have different access policies that made map teams useful, no longer and instead permissions are granted directly without the team.

With that said, I forked the given repo into triplea-maps after a quick content review, please delete the original repo.

PR #8832, opened 2021-02-09
Update triplea_maps.yaml

axisalliescontango:
For completion after Issue #8828 to add the new map.  
I've also put a description.html in the map folder as I believe things are headed in that direction, but not sure where it's at right now.  Hopefully this is ok.
Thanks, 
Contango


beelee1:
@axisalliescontango I think you still need to add the img

img: https://raw.githubusercontent.com/triplea-maps/g40_alt_universe/master/preview.png

goes below version

axisalliescontango:
Going to re-add with the image, thanks for noticing @beelee1.  Not sure how I missed that.

PR #8839, opened 2021-02-11
Update triplea_maps.yaml

axisalliescontango:
Added new pull request as missing the img the first time. 

I don't really understand why it needs the \blob in the address as the others don't seem to, but if I go and open the image from the repo that's the address it uses.  Also doesn't seem successful without it.


axisalliescontango:
@DanVanAtta 

Hi Dan, hate to bug you on a long weekend but if you get a chance can you approve PR 8839?  I know there’s a bunch of people that would like to get access to this map so was hoping to get it live on TripleA this weekend.

Thanks,
Contango

DanVanAtta:
Sorry for the delay, pretty clearly I need to work harder to find some minions that could help with this. Usually the turn-around time is faster...

PR #8858, opened 2021-02-14
Amphibious units without marine bonus should be sorted first

### LINKED ISSUE #8855: Non-Offloaded Units Are Unpredictably Removed Before Same-Named Offloaded Ones (2.5.22294)
Start a game of which the rules set allows retreating non-offloaded units (all basic ones from v3 onwards), make a battle in which there are offloaded and non-offloaded units of the same type and observe that, once only a combination of units which is the set or a sub-set of all offloaded units remain, you will not be prompted to retreat any longer.

For example, start "World War II v3 1941", send in "Baltic States" 2 infantry units from Finland and 1 infantry unit from Poland, and witness how you will have no prompt once only 2 or less infantry units remain offending (enable "Low Luck" to be sure to observe the behaviour).

### Do you have any ideas for an expected fix?
I believe the intended rules in every case it matters (also in case you cannot retreat any land units, yet offloaded units are differentiated due to marine bonuses) are that the player can individually choose to remove an offloaded or a non-offloaded unit for every hit (I'm assuming no multiple hit points units are present, to keep the explanation more concise and since all land units have only 1 hit point each in every basic game.).

For example, if I send 2 offloaded infantries and 1 non-offloaded infantry in a battle and receive 2 hits on the first round of combat, I should be allowed to do only one of the following.
- Remove 2 offloaded infantries.
- Remove 1 offloaded infantry and 1 non-offloaded infantry.

(If the game is v3, taking the second option is idiotic, because you deprive yourself of the ability to retreat and are otherwise in the same condition as taking the first option, yet this is exactly what the program forces you to do.)

### Attach a Save Game
I believe this is easy enough to reproduce that doing so should take less time than downloading and loading any save-game (especially if you have to unzip the save-game).

I'm almost sure in the past offloaded units were always removed before non-offloaded units of the same type, which is just as wrong a behaviour, but at least I believe it is usually preferable if there are no marine bonuses in the game. I think removing non-offloaded units before offloaded ones of the same type is a very bad behaviour and likely the greatest problem which I'm aware the program has.

I think this problem should have the highest priority amongst all problems currently in existence for the release. It is a huge problem for my "270BC Wars" game too.

CC: @panther2 @trevan

trevan:
This partially solves #8855.  It ensures that the amphibious units are taken first when they don't have marine bonuses.

## Testing
I played Global 1940 and did an amphibious assault with 2 infantry as well as 2 non-amphibious infantry.  Verified that the 2 amphibious infantry were taken first

## Release Note

FIX|Amphibious units that do not have a marine bonus will be auto selected as a casualty before the non-amphibious units are.  If the amphibious units have a marine bonus, then they will be auto selected after the non-amphibious units.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtySelector.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtySelector.java
@@ -28,7 +28,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 import java.util.function.Predicate;
-import java.util.stream.Collectors;
 import lombok.experimental.UtilityClass;
 import lombok.extern.slf4j.Slf4j;
 import org.triplea.java.collections.CollectionUtils;
@@ -102,13 +101,6 @@ public class CasualtySelector {
       return new CasualtyDetails(List.of(), List.of(), true);
     }
 
-    if (allTargetsOneTypeOneHitPoint(targetsToPickFrom, dependents)) {
-      final List<Unit> killed =
-          targetsToPickFrom.stream()
-              .limit(Math.min(hitsRemaining, targetsToPickFrom.size()))
-              .collect(Collectors.toList());
-      return new CasualtyDetails(killed, List.of(), true);
-    }
     // Create production cost map, Maybe should do this elsewhere, but in case prices change, we do
     // it here.
     final IntegerMap<UnitType> costs = TuvUtils.getCostsForTuv(player, data);
@@ -135,6 +127,7 @@ public class CasualtySelector {
 
     final CasualtyDetails casualtySelection =
         hitsRemaining >= totalHitpoints
+                || allTargetsOneTypeOneHitPoint(sortedTargetsToPickFrom, dependents)
             ? new CasualtyDetails(defaultCasualties, true)
             : tripleaPlayer.selectCasualties(
                 sortedTargetsToPickFrom,


Cernelius:
I support this (if not implementing the actual intended rules) but I think it should NOT be called a fix. In theory, automatically selecting one way or the other is equally wrong, rules wise, but I agree this is an improvement going in the direction of reducing the impact of the problem. Still, this is not a fix.

DanVanAtta:
Would you mind discussing more of the background of this solution @trevan ? Why should units with amphib bonus be chosen before others? Generally I would think it is because they have a stronger power, in which case is the algorithm to choose casualties not taking into account the amphib attack power? Would updating that latter part be a better fix? Let's say for example if there were a map where somehow the amphib bonus was actually a penalty, this solution would need special casing (which I think indicates that perhaps a better fix is in the attack power evaluation).

Cernelius:
Indeed TripleA supports negative marine bonuses to represent the difficulties related to make sea-borne operations.

trevan:
> Why should units with amphib bonus be chosen before others?

That's the exact opposite of what this fix is doing.  If a unit has amphib bonus, then it should be selected last.

> Generally I would think it is because they have a stronger power, in which case is the algorithm to choose casualties not taking into account the amphib attack power? Would updating that latter part be a better fix?

Oddly enough, if you are only attacking with one unit type, it doesn't use the casualty selection code.  It just selects the first N units as the casualties.  So the order of the units is necessary for this corner case.  If you have more than one unit type (say attack with 2 amphibious infantry, 2 non-amphibious infantry, and 1 tank [amphibious or not, doesn't matter]), then the code will use the casualty selection code which does take into account all the different support components (marine, territory, etc).

So this issue only happens in the situation where only a single unit type is involved in a battle that includes some amphibious and some non-amphibious units.  And this fix is for that.

I'll add in a test to ensure that negative `isMarine` values are handled correctly.

DanVanAtta:
>    Why should units with amphib bonus be chosen before others?

> That's the exact opposite of what this fix is doing. If a unit has amphib bonus, then it should be selected last.

Sorry, that was a typo,  it was meant to read as "chosen last"

DanVanAtta:
> I'll add in a test to ensure that negative isMarine values are handled correctly.

This is precisely the kind of logic I was hoping to avoid. The single unit selection is also wrong in the case of units with differing movement points. 

The lack of logic around "is there just a single unit", okay choose 'N' is a bug. It's a bug of omission, one of the hardest kind to detect since it's code that should be there, but is not.

The algo to choose any N when there is a single unit type assumes all units would have the same characteristics (and that is wrong). The algo for the 'choose n' of same type should instead check unit powers and then unit movement values.

Presumably any terrain modifiers would impact all units of the same type in the same way, I can't think of any other cases that would impact units of the same type in a different way.

DanVanAtta:
TL;DR, I think the fix for this should be sort by power with secondary sort on movement. This avoid special case logic for the amphib attack bonus.

trevan:
So do you want to get rid of the special single unit logic?  That logic was added back in 2004 - fdc15f74f8e334fe5524b291440c048b006a910d.

DanVanAtta:
@trevan no, but I'd suggest that the check for amphib units is where the problem is. That check should not be for amphib units but should instead be checking unit power.

game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtySortingUtil.java
```diff
@@ -10,24 +11,18 @@
...
-  public static Comparator<Unit> compareMarines() {
-    return (u1, u2) -> {
-      final int result = Boolean.compare(u1.getWasAmphibious(), u2.getWasAmphibious());
-      if (result != 0) {
-        return result;
-      }
-
-      final UnitAttachment ua1 = UnitAttachment.get(u1.getType());
-      final UnitAttachment ua2 = UnitAttachment.get(u2.getType());
-      return Integer.compare(ua1.getIsMarine(), ua2.getIsMarine());
```
DanVanAtta, line 30: Problem is here, the fact that we are comparing marines or amphib is red herring and wrong. The fix is not improving this IMO, but changing this to be able to compute unit power.

DanVanAtta: If we can then go on to do a secondary sort, we'll be fixing a second bug where planes with less movement are selected after units with more movement.

trevan:
Just noticed there is also a `killAmphibiousFirst` method that exists in the casualty selection code which was added back in 2009 - a5e328dbf49c70664dc4ea92ce03c69e13aefecd.

So, this commit just gets the functionality back to how it was in the previous releases.  I think either take the PR as it is or figure out a change to the casualty selection menu and then rip out all of this special case code.

game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtySortingUtil.java
trevan: It already has a secondary sort by movement.  So if that bug exists, then this sort won't fix that bug.

DanVanAtta:
@trevan I'm suggesting to extract the comparison to somewhere else where the unit power is calculated (separation of concerns). The amphib attribute will be an input to that calculation. Then the single-unit type casualty selection code only needs to be aware that it is sorting by unit attribute values and now how to compute those values. Getting a fix in here for plane movement would also be useful.

Is the single unit-type selection code also used when a person has selected a casualty count? (EG: I've selected 2 of 3 fighters). If so, then we'll have solved a number of bugs and will not require an updated casualty selection UI to at least get an optimal casualty selection.

trevan:
This sort is unrelated to unit power calculation and I do not want to put unit power comparison in there.  If you want to utilize unit power calculation, than we should just not use the single unit special case and let the casualty order of loss calculation do its work, because it already knows how to handle marine powers and such.  And with that change, this sort won't even be needed anymore since the casualty ool has its own sort.

I'll look into ripping out the single unit stuff and using the casualty ool results instead.

trevan:
Ok, the single unit type target selection code is gone.  It will use the targets that casualty order of loss chooses and what the `killAmphibiousFirst` chooses.

I've force pushed the code to remove the previous two commits.  I've tested this with Global 1940 with 2 amphibious infantry along with 2 non-amphibious infantry.  The 2 amphibious infantry were auto-selected first.

DanVanAtta:
Thanks @trevan. I agree the fix to use an already existing method that takes into account unit power is superior. The checks for amphib or not IMO are hacks, added in a myopic way to get a narrow case working without a good view of the broader picture (which is completely understandable because TripleA combat code has entropied to where nobody can understand it which then begets more similar hacks and further entropy).

 Can you confirm that there is no casualty selection prompt appearing @trevan and that the units are auto-selected without prompt?

Cernelius:
I still strongly suggest not to label this as a "FIX". It should be labelled as something saying that a current wrong behaviour has been changed to another just as much wrong behaviour but being likely less problematic than the former (in this case, because you usually want to take out offloaded first, to conserve the option to retreat). Technically, this change doesn't even make the behavior less wrong.

DanVanAtta:
@Cernelius please keep in mind the context and target audience of the release notes.

Cernelius:
Auto-selecting (whilst always wrong from v3 onwards) is particularly bad if playing a game with both sea-borne partial retreat and positive marine bonuses. In this case, I most likely want to take non-offloaded first if I plan to continue the battle but want to take offloaded first if I plan to retreat on the same round (normally nothing happens between the moment the offender selects casualties and the moment it is prompted with the retreat choice).

If an auto-selection must be, I agree with taking non-offloaded (thus weaker) first for all units having positive marine bonuses, but you almost surely want to do the exact opposite the round in which you want to retreat (a decision you practically make immediately after seeing the hits obtained by the defender, so immediately before selecting your casualties as you will also choose casualties one way or the other depending whether or not you intend to retreat, even though the prompt itself is after the casualties selection is confirmed).

---

If my recollection is correct, until recently the behaviour was always taking offloaded first, no matter any marine bonuses. If I'm right, this implies that the behaviour has flipped to the current behaviour of always taking non-offloaded first for reasons I don't know.

trevan:
> Can you confirm that there is no casualty selection prompt appearing @trevan and that the units are auto-selected without prompt?

The casualty selection prompt is not shown.

trevan:
> If my recollection is correct, until recently the behaviour was always taking offloaded first, no matter any marine bonuses. If I'm right, this implies that the behaviour has flipped to the current behaviour of always taking non-offloaded first for reasons I don't know.

A bug was introduced because of refactoring to handle amphibious units better (the mechanism to remember which unit was amphibious was simplified).  This PR is basically "fixing" that bug.

PR #8920, opened 2021-02-27
Improve unit separator parameters

trevan:
## Testing

Played WW2 Global 1940 and tested that unit choosing still worked.  I tested selecting transports with dependents as well as just regular units.

This is initial work to separate amphibious/offloaded units in the unit chooser dialog.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/util/UnitCategory.java
+++ b/game-core/src/main/java/games/strategy/triplea/util/UnitCategory.java
@@ -37,7 +37,7 @@ public class UnitCategory implements Comparable<UnitCategory> {
   private final BigDecimal movement;
   // movement of the units
   private final int transportCost;
-  private final boolean wasAmphibious;
+  private final boolean amphibious;
   private final GamePlayer owner;
   // the units in the category, may be duplicates.
   private final List<Unit> units = new ArrayList<>();
@@ -51,7 +51,7 @@ public class UnitCategory implements Comparable<UnitCategory> {
     movement = new BigDecimal(-1);
     transportCost = -1;
     this.owner = owner;
-    wasAmphibious = false;
+    amphibious = false;
   }
 
   UnitCategory(
@@ -62,7 +62,7 @@ public class UnitCategory implements Comparable<UnitCategory> {
       final int bombingDamage,
       final boolean disabled,
       final int transportCost,
-      final boolean wasAmphibious) {
+      final boolean amphibious) {
     type = unit.getType();
     this.movement = movement;
     this.transportCost = transportCost;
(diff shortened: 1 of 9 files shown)


game-core/src/main/java/games/strategy/triplea/util/UnitSeparator.java
```diff
@@ -19,11 +19,33 @@
...
+     */
+    @Builder.Default @Nullable final Map<Unit, Collection<Unit>> dependents = null;
+    /** whether to categorize by movement */
+    @Builder.Default final boolean movement = false;
+    /** whether to categorize by transport cost */
+    @Builder.Default final boolean transportCost = false;
+    /** whether to categorize transports by movement */
+    @Builder.Default final boolean transportMovement = false;
+    /**
+     * If true then sort the categories in UnitCategory order; if false, then leave categories in
```
DanVanAtta, line 44: Consider defining 'UnitCategory (sort) order', I'm not 100% clear what it really means. Secondarily the 'sort' variable perhaps could be given some added context (renamed) to help convey more precisely what the sort is. Perhaps the javadoc should also mention when it would be desirable to sort. Third, 'original order' could perhaps be more precise, I would presume that is the order units are added to the separator (but perhaps it is something else, having the documentation be more exact would clarify this).

game-core/src/main/java/games/strategy/triplea/util/UnitCategory.java
```diff
@@ -36,6 +37,7 @@
   private final BigDecimal movement;
   // movement of the units
   private final int transportCost;
+  private final boolean wasAmphibious;
```
DanVanAtta, line 40: Why past tense with 'was'? A unit that is amphibious during combat, is actively an amphib assault, no? I feel like this might make more sense adding the word 'assault' in, as a unit can be amphibious (meaning they can attack from a transport), though 'conductedAmphibiouAssault' I think is the more exact (and verbose) way to describe it. The past tense of 'was' does strike me as a bit odd (for example, a casual reader may wonder when does this value become true, before a combat is initiated, during, or only after?).

DanVanAtta (commented):
Just couple comments where things could be clarified, otherwise looks good.

game-core/src/main/java/games/strategy/triplea/util/UnitSeparator.java
trevan: After going through the commit history, I determined that the sort is no longer needed.  So it is removed in https://github.com/triplea-game/triplea/pull/8920/commits/e0d2a13d3a57c8bc7e8244dc134c99f642bb8706.

game-core/src/main/java/games/strategy/triplea/util/UnitCategory.java
trevan: `wasAmphibious` is the term used in the `UnitAttachment` and so I was trying to keep the same terminology.  But since I used `amphibious` in the UnitSeparator, I've changed this to just `amphibious`.  See https://github.com/triplea-game/triplea/pull/8920/commits/c6950787c6db43e1aeebf8a31ead4830c7a7c4db

PR #8924, opened 2021-02-27
Separate out air units by movement left in the casualty selector

trevan:
## Screens Shots

Screen Shot 2021-02-27 at 4 41 04 PM

## Release Note

NEW|Air units are now separated by their movement left in the battle casualty selection. This allows selecting the air unit with less movement than the other air units.

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtySelector.java
+++ b/game-core/src/main/java/games/strategy/triplea/delegate/battle/casualty/CasualtySelector.java
@@ -329,7 +329,11 @@ public class CasualtySelector {
       final Collection<Unit> targets, final Map<Unit, Collection<Unit>> dependents) {
     final Set<UnitCategory> categorized =
         UnitSeparator.categorize(
-            targets, UnitSeparator.SeparatorCategories.builder().dependents(dependents).build());
+            targets,
+            UnitSeparator.SeparatorCategories.builder()
+                .airMovement(true)
+                .dependents(dependents)
+                .build());
     if (categorized.size() == 1) {
       final UnitCategory unitCategory = categorized.iterator().next();
       return unitCategory.getHitPoints() - unitCategory.getDamaged() <= 1;
--- a/game-core/src/main/java/games/strategy/triplea/ui/UnitChooser.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/UnitChooser.java
@@ -97,7 +97,7 @@ public final class UnitChooser extends JPanel {
     combinedList.addAll(defaultSelections.getKilled());
     createEntries(
         units,
-        UnitSeparator.SeparatorCategories.builder().dependents(dependent).build(),
+        UnitSeparator.SeparatorCategories.builder().dependents(dependent).airMovement(true).build(),
         combinedList);
     layoutEntries();
   }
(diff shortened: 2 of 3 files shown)


DanVanAtta:
How does this look for land units? (EG: tanks with 2 movement, or half tracks from WaW)

trevan:
It only shows movement for air units.  I understand that those are the only ones that can move after a battle so their left over movement is important.  If other units can move after a battle, is there something in the UnitAttachment that indicates that?

trevan:
I just noticed that if only one plane type is in the battle and it has different movements, then it wouldn't ask for a casualty selection.  Pushed a change to fix that.

DanVanAtta:
Good point on it not mattering for land units because they cannot move after battles. AFAIK that is universally true. A simpler UI therefore is useful.

If all planes have the same movement, does the UI still show the movement left?

Another question, have you considered any alternative UI treatments for the movement indicator beyond text? Specifically having the value be a subscript and instead of text use some sort of symbol instead?

DanVanAtta:
> I just noticed that if only one plane type is in the battle and it has different movements, then it wouldn't ask for a casualty selection. Pushed a change to fix that.

Perhaps we should auto-select here the plane with the least movement. The auto-selection is very valuable for game-play speed.

trevan:
> If all planes have the same movement, does the UI still show the movement left?

Yes

> Another question, have you considered any alternative UI treatments for the movement indicator beyond text? Specifically having the value be a subscript and instead of text use some sort of symbol instead?

I'm just using the existing UI mechanism.  The text indicator was already present.

DanVanAtta:
The 2x2, 3x1 can kinda blend together. I think we can get a cleaner UI. I'd suggest we use a smaller font, a subscript and pair that with a movement symbol. If we do that, we do not need to worry as much localization. We also avoid any ambiguity about the abbreviation (particularly helpful for non-native English speakers). Do you agree that is overall a good next move? 

> I'm just using the existing UI mechanism. The text indicator was already present.

Can you clarify, sorry to be confused, how do you mean by existing UI mechanism? AFAIK we did not show plane movement before.

Cernelius:
Maybe the "x4" and so on are better after the thing displaying the number selected and rewritten as "/4". So you would see something like "2/4" (2 out of 4).

Cernelius:
Also "movement" sounds like that is how much they move, so hard to imagine if that means how much the units can still move or how much they have moved already. Maybe "mobility" would be a better term.
https://dictionary.cambridge.org/dictionary/english/mobility

DanVanAtta:
Considering a person is likely going to be familiar with total movements and will have moved the units to begin with, I don't think we need to add distinguishing terminology.

Come to think of it, we are potentially opening this up to misclicks and tedious selection. A player *will* always want the max air movement units to remain. I'm starting to think we may have not thought this through well enough. I don't think we need to even bother with the UI difference if we always select the least movement first. That was the original bug report that was submitted, that sometimes the higher movement units were selected first and that had material impact. To fix that, if we break game flow for a many battles, it's not necessarily the best option.

beelee1:
yea i think auto select least movement and don't have a prompt. i thought I remembered this being talked about a while back but can't really remember

Cernelius:
In practice, the only case I can imagine in which you might want to take a unit with more movement left first is if, this way, you will avoid being obliged moving a carrier to land the unit, but such a situation may realistically never happen. Another situation might be being forced to land the fighter on an allied carrier you would rather not (even less likely).

I'm fairly neutral on whether or not TripleA should go ahead selecting units with lesser movement left first, though the intended rules let you select greater movement if you want.

I would just suggest this being consistent with what TripleA does in any similar cases of non-sensible choices. For example, whether or not allowing selecting non-offloaded marines first when no land units can retreat.

trevan:
> Can you clarify, sorry to be confused, how do you mean by existing UI mechanism? AFAIK we did not show plane movement before.

This showing of the movement already exists in the code.  It is used at some point, but I don't know where.  I'm just turning on the option during casualty selection.

> I don't think we need to even bother with the UI difference if we always select the least movement first.

Ok, I'll close this PR and instead just force the selection.

Issue #8607, opened 2021-01-09
Unit Help Gets Hidden (2.5.22294)


Cernelius:
The "Unit Help" window has this problem that, if you open it in a game in progress, it disappears when the focus goes on the map. This can be especially annoying for new users that are struggling on unknown games.

Example of this recurring problem, as it just happened (my host chatlog):
>(04:40:19) ROLLINGTHUNDER:  the unit help disapears when battle starts
(04:41:04) Cernel:  it goes behind the map
(04:41:24) ROLLINGTHUNDER:  lol
(04:41:32) Cernel:  it would be better if the unit help is in the side bar really
(04:41:34) ROLLINGTHUNDER:  there were about 15 of them
(04:41:40) Cernel:  yeah

I'm not a big fan of the "Unit Help" feature myself (as I believe it is rather better to document anything this feature provides with a well made table in "Notes"), but I guess that it may make sense to add such feature as a table in the right-side bar, since the "Notes" are there too.

Cernelius:
A palliative would be that, at least, every time you click on "Unit Help" again, the already opened one is re-called, instead of opening more and more of them. It makes hardly any sense to open more than one instance of "Unit Help", so I think it would be better for that to be impossible. The same applies to anything else like that, chiefly the "Game Notes".

PR #8925, opened 2021-02-28
Remove game notes action tab

DanVanAtta:
Several discussions have happened in forums, generall agreed that the
game notes in the action tabs is not particularly useful. To keep things
simpler, this update remove the game notes tab.

## Release Notes

REMOVE|Removed game notes 'action' tab, game notes are now only available from the 'help > game notes' menu.


beelee1:
I agree but just want to point out that you can't move the map when accessing the game notes from the help tab. It'd be nice if you could. Not a biggy but still a difference.

Cernelius:
I'm against this change unless it is also assured that the notes always stay above the map view. I'm referring to the following issue, which impacts the game notes as well as the unit help window.
https://github.com/triplea-game/triplea/issues/8607

Even in case that would be fixed, it has so far happened so frequently that windows have issues of various kind that I would suggest keeping this rather useless tab just as a fall-back for future issues with windows.

Besides, I think this is something which should be better discussed in forum (just ask users if they like reading notes in the tabs any better than via the menu).

beelee1:
@Cernelius I think it was talked about in forum. i had meant to comment but forgot about it until seeing it here.

DanVanAtta:
> @Cernelius I think it was talked about in forum. i had meant to comment but forgot about it until seeing it here.

Yes, it was. I'm not sure where the thread is, the consensus between several folks  was that it's not useful.

DanVanAtta:
Fixing the always on top seems easy enough actually, will follow up.

PR #8926, opened 2021-02-28
Set game notes & unit help windows to always be on top

DanVanAtta:
## Release Note

UPDATE|Game notes and unit help windows are now always displayed on top of other windows

### DIFF
--- a/game-core/src/main/java/games/strategy/triplea/ui/menubar/help/InformationDialog.java
+++ b/game-core/src/main/java/games/strategy/triplea/ui/menubar/help/InformationDialog.java
@@ -15,6 +15,7 @@ import org.triplea.swing.SwingAction;
 class InformationDialog {
   static JDialog createDialog(final JComponent component, final String title) {
     final JDialog dialog = new JDialog((JFrame) null, title);
+    dialog.setAlwaysOnTop(true);
     dialog.add(component, BorderLayout.CENTER);
     final JPanel buttons = new JPanel();
     final JButton button =


Cernelius:
This should apply to anything like this all the same (the unit help tab and whatever information window like that).

---

This (game notes always on top) was the behaviour in the past, but it caused the notes to hide prompts (for example, when you were reading the notes and suddenly you received a prompt for whatever reason), so the behaviour was changed to the current. In practice, the games notes should always be above the map view but below everything else (so, if you get a politics prompt or a battle window, that appears above the game notes while the game notes stay below every prompt but above the map).

Rather than being always on top, the game notes should be set being never behind the map view or, even better, the map view itself should be set being always below everything else.

DanVanAtta:
> Rather than being always on top, the game notes should be set being never behind the map view or, even better, the map view itself should be set being always below everything else.

That is not possible unfortunately.


> This (game notes always on top) was the behaviour in the past, but it caused the notes to hide prompts (for example, when you were reading the notes and suddenly you received a prompt for whatever reason), so the behaviour was changed to the current.

DanVanAtta:
> This (game notes always on top) was the behaviour in the past, but it caused the notes to hide prompts (for example, when you were reading the notes and suddenly you received a prompt for whatever reason), so the behaviour was changed to the current. 

Fun stuff : )

This is a good point to bring up. I think the picture might have changed after we set various other dialogs to not be frames and toggled modality. I checked that a modal dialog appearing is not hidden, the prompt does appear above the notes window. Specifically I checked a politics notification. With luck this UX will be universal on all OS's.

Cernelius:
> > This (game notes always on top) was the behaviour in the past, but it caused the notes to hide prompts (for example, when you were reading the notes and suddenly you received a prompt for whatever reason), so the behaviour was changed to the current.
> 
> Fun stuff : )
> 
> This is a good point to bring up. I think the picture might have changed after we set various other dialogs to not be frames and toggled modality. I checked that a modal dialog appearing is not hidden, the prompt does appear above the notes window. Specifically I checked a politics notification. With luck this UX will be universal on all OS's.

I think this migh be the case, because there are a lot of other information windows which already have the always on top behaviour and I don't recall, for example, any report of someone getting stuck while reading the "Politics Panel".

On this matter, as I said
>This should apply to anything like this all the same

I also think the "**Map Options**" (this shows the game options, you may have selected before loading the game), the "**Politics Panel**", the "**Random Stats**", the "**Game Statistics**" and the "**Movement/Selection Help**" should have the same behaviour as the "**Unit Help**" and the "**Game Notes**". I'm not seeing why any information window should behave any differently from any other one (I would make exception only for the "**TripleA Console**", as it is not really a user information widow.).

Off topic, but maybe all windows that are only for giving information should be consolidated within one menu category. Like changing "Help" to "Information" and putting them all there.

DanVanAtta:
I'll have to think about if we want all windows to be always on top. I'd like more feedback on it from others as well. In the meantime this update will fix the problem reported. I do see a use-case where you want the unit help or game-notes to be visible while moving around on the map.

Issue #8163, opened 2020-11-13
2.5.22294: NodeBbTokenGenerator#getToken:149 - java.lang.IllegalStateException


tripleabuilderbot:
## Log Message
Incorrect password or server error.
Return Code: not-authorised
Message from server: A valid login session was not found. Please log in and try again.

## TripleA Version
2.5.22294

## Java Version
11.0.6

## Operating System
Mac OS X

## Memory
Heap utilization statistics [MB]
Used Memory: 39
Free memory: 2008
Total memory: 2048
Max memory: 2048

## Stack Trace
```
Exception: java.lang.IllegalStateExceptionIncorrect password or server error.
Return Code: not-authorised
Message from server: A valid login session was not found. Please log in and try again.
java.lang.Exception
	at games.strategy.engine.posted.game.pbf.NodeBbTokenGenerator.getToken(NodeBbTokenGenerator.java:149)
	at games.strategy.engine.posted.game.pbf.NodeBbTokenGenerator.generateToken(NodeBbTokenGenerator.java:64)
	at games.strategy.engine.framework.startup.ui.posted.game.pbf.ForumPosterEditorViewModel.renewToken(ForumPosterEditorViewModel.java:92)
	at games.strategy.engine.framework.ui.background.BackgroundTaskRunner$1.doInBackground(BackgroundTaskRunner.java:108)
	at java.desktop/javax.swing.SwingWorker$1.call(SwingWorker.java:304)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.desktop/javax.swing.SwingWorker.run(SwingWorker.java:343)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)


```

doctored:
this is the error i get when trying to play by forum with the mac this is with today's java update as well.

DanVanAtta:
@doctored you have access to another OS, and you do not get the same error with the same version of TripleA?

doctored:
i do not but it was working fine until 2.x

DanVanAtta:
Can you see if you have luck on 1.9?

doctored:
it was working fine in 1.9 the problem started after i upgraded to 2.x.

[...]

PR #8942, opened 2021-02-28
Fix authentication issues

RoiEXLab:
Potentially a solution to #8163 
I didn't get the exact same error message, but I was able to succesfully retreive a token after making those changes

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbTokenGenerator.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbTokenGenerator.java
@@ -57,7 +57,8 @@ public class NodeBbTokenGenerator {
     Preconditions.checkNotNull(username);
     Preconditions.checkNotNull(password);
 
-    try (CloseableHttpClient client = HttpClients.custom().disableCookieManagement().build()) {
+    try (CloseableHttpClient client = HttpClients.custom().disableCookieManagement()
+         .disableDefaultUserAgent().build()) {
       final int userId = getUserId(client, username);
       return new TokenInfo(getToken(client, userId, password, otp), userId);
     } catch (final IOException e) {
@@ -72,7 +73,8 @@ public class NodeBbTokenGenerator {
    * @param userId The userId that the token was issued for.
    */
   public void revokeToken(final String token, final int userId) {
-    try (CloseableHttpClient client = HttpClients.custom().disableCookieManagement().build()) {
+    try (CloseableHttpClient client = HttpClients.custom().disableCookieManagement()
+         .disableDefaultUserAgent().build()) {
       deleteToken(client, userId, token);
     } catch (final IOException e) {
       throw new RuntimeException("Failed to revoke login token", e);
@@ -110,9 +112,9 @@ public class NodeBbTokenGenerator {
   private Map<String, Object> queryUserInfo(final CloseableHttpClient client, final String username)
       throws IOException {
     final String encodedUsername = URLEncoder.encode(username, StandardCharsets.UTF_8);
-    final HttpGet post = new HttpGet(forumUrl + "/api/user/username/" + encodedUsername);
-    HttpProxy.addProxy(post);
-    try (CloseableHttpResponse response = client.execute(post)) {
+    final HttpGet get = new HttpGet(forumUrl + "/api/user/username/" + encodedUsername);
+    HttpProxy.addProxy(get);
+    try (CloseableHttpResponse response = client.execute(get)) {
       return YamlReader.readMap(EntityUtils.toString(response.getEntity()));
     }
   }


game-app/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbTokenGenerator.java
```diff
@@ -72,7 +73,8 @@ public TokenInfo generateToken(
    * @param userId The userId that the token was issued for.
    */
   public void revokeToken(final String token, final int userId) {
-    try (CloseableHttpClient client = HttpClients.custom().disableCookieManagement().build()) {
+    try (CloseableHttpClient client = HttpClients.custom().disableCookieManagement()
```
DanVanAtta, line 76: What does 'disableDefaultUserClient' do, should we alwasy use that?
I suspect as well we may want to move this code into a common lib base-layer to make it easier/more generic.

game-app/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbTokenGenerator.java
```diff
@@ -110,9 +112,9 @@ private void checkUser(final Map<?, ?> jsonObject, final String username) {
   private Map<String, Object> queryUserInfo(final CloseableHttpClient client, final String username)
       throws IOException {
     final String encodedUsername = URLEncoder.encode(username, StandardCharsets.UTF_8);
-    final HttpGet post = new HttpGet(forumUrl + "/api/user/username/" + encodedUsername);
-    HttpProxy.addProxy(post);
-    try (CloseableHttpResponse response = client.execute(post)) {
+    final HttpGet get = new HttpGet(forumUrl + "/api/user/username/" + encodedUsername);
```
DanVanAtta, line 115: Is the change from 'get' to 'post' the major fix, or perhaps just part of it?

RoiEXLab: If you look closely you'll notice that I only changed the variable name here, because it was kinda misleading

game-app/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbTokenGenerator.java
RoiEXLab: This seems to be the "fix" for some reason.
During debugging I was reproducing the exact same steps as the code using curl and my webbrowser. Using curl and the browser everything worked as expected only java was having some weird issues.
The only possible difference could be the headers the client sends, so I tried to disable some headers, and after removing the user agent header (which tells the HTTP server which browser is requesting the site) everything worked fine again.

RoiEXLab: > I suspect as well we may want to move this code into a common lib base-layer to make it easier/more generic.

That would be an option of course but I don't think it's worth to invest too much time into this particular code.
Http Components 4 has started to become legacy code (HttpComponents 5 is out and calls the current code scheme the "classic" scheme, see https://hc.apache.org/httpcomponents-client-5.0.x/index.html ). We could consider migrating to the modern async approaches to this problem, or we could convert this code into a feign pattern, or migrate to the "new" HttpClient from the standard library which seems to have a nice API: https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpClient.html

It avoids having to wrap the code in try-with-resource statements and provides a nice builder API, so the new code would looks more like this:
```java
HttpClient client = HttpClients.newBuilder()
  .followRedirects(Redirect.NORMAL)
  .proxy(ProxySelector.of(new InetSocketAddress(ClientSetting.proxyHost.getValueOrThrow(), ClientSetting.proxyPort.getValueOrThrow())))
  .build();

 HttpRequest request = HttpRequest.newBuilder()
    .uri(URI.create("https://foo.com/"))
    .GET()
    .build();
  HttpResponse<String> response = client.send(request, BodyHandlers.ofString());
  System.out.println(response.statusCode());
  System.out.println(response.body());  
```
And instead of `BodyHandlers.ofString()` we could plug in our own yaml/json parser.

This way the API becomes slightly nicer and we could dump a dependency

DanVanAtta: Thanks for the feedback, interesting to see there is already a new API.

IMO, best to keep the tech stack as consistent as we can and use Feign. Feign does not support 'put' though, therefore for file uploads we'll likely use httpcomponents.

doctored:
so is this resolved now?

RoiEXLab:
See https://github.com/triplea-game/triplea/issues/8163#issuecomment-931739762

Issue #8927, opened 2021-02-28
No resource loader set error when joining network games


DanVanAtta:
### How can the problem be recreated?
- host a local game
- launch second game instance
- join the locally hosted game
- observe error message
-  

### If playing a prerelease, does this happen on the latest release?

- [] yes
- [x] no

### Is there anything else we should know?

> java.lang.IllegalStateException: No resource loader has been set
	at org.triplea.sound.ClipPlayer.lambda$play$0(ClipPlayer.java:229)
	at java.base/java.util.Optional.orElseThrow(Optional.java:408)
	at org.triplea.sound.ClipPlayer.play(ClipPlayer.java:229)
	at org.triplea.sound.ClipPlayer.play(ClipPlayer.java:214)
	at games.strategy.engine.chat.ChatMessagePanel.playerJoined(ChatMessagePanel.java:287)
	at games.strategy.engine.chat.Chat.lambda$participantAdded$4(Chat.java:92)
	at java.base/java.util.concurrent.CopyOnWriteArrayList.forEach(CopyOnWriteArrayList.java:807)
	at games.strategy.engine.chat.Chat.participantAdded(Chat.java:91)
	at games.strategy.engine.chat.MessengersChatTransmitter$1.speakerAdded(MessengersChatTransmitter.java:54)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at games.strategy.engine.message.unifiedmessenger.EndPoint.invokeSingle(EndPoint.java:135)
	at games.strategy.engine.message.unifiedmessenger.EndPoint.lambda$invokeMultiple$0(EndPoint.java:119)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
[...]

PR #8943, opened 2021-02-28
Fix resource loader not set error on hosted game chat

DanVanAtta:
This update restores the defaulting of a game-engine only asset loader
when there is no resource loader otherwise set.

Problem introduced in: f3face998ca1ee91eae5ffce1c7580fb9d0ebf0e,

Fixes: https://github.com/triplea-game/triplea/issues/8927


RoiEXLab (approved):

RoiEXLab:
Thanks for fixing this

PR #8985, opened 2021-03-06
Update Oil and Snow with Name Change

beelee1:
Screenshot from 2021-03-05 21-15-27
So i tried removing 1st edition everywhere in the existing map folders an xml and it would show up with the description picture but just go blank when I selected it. I noticed where it was originally forked from it's named Oil and Snow without the 1st Edition. 

At any rate, if the yaml mapName here is the same as the triplea-maps repo one  and it automatically adds underscore to spaces and makes it all lower case, then I think we should be fine. I don't really know how to test until it's added though. I did rename the zip in downloaded maps with 1st_edition and it worked then, so I think it should work.

Just found it here. Was looking at the bottom of the list for it lol :)

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -39,11 +39,11 @@
   version: 2
   description: |
     <br>This is a tutorial for players new to TripleA and its game engine.
-- mapName: WW2 Oil and Snow 
+- mapName: WW2 Oil and Snow 1st Edition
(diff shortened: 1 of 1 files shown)


DanVanAtta:
This was the right fix, I already pushed these changes, didn't notice you had this PR up already.

PR #9010, opened 2021-03-08
Replace 'none' territory selected with an empty string

DanVanAtta:
In the status bar we print the name of the territory as
the mouse cursor changes. When hovering over a territory
boundary the territory text changes to 'none'. Instead
of this behavior, this update prints an empty string
instead of the text 'none'. So we display nothing
rather than the text 'none'.

## Screens Shots

Before:
Screenshot from 2021-03-07 17-27-35

After:
Screenshot from 2021-03-07 17-28-17

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
@@ -228,7 +228,7 @@ public final class TripleAFrame extends JFrame implements QuitHandler {
         void refresh() {
           territoryInfo.removeAll();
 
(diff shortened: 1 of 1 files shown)


Cernelius:
I agree. A mapmaker may well decide to call a territory as "none".

However, what you are calling territory is actually a zone (all territories are zones but not all zones are territories). Also, I suppose by "select" you mean "point at", "focus" or "hover".

When I hover a non-ownable sea zone I'm actually hovering no territory.

Cernelius:
When I make tables I use to add a hyphen where there is no value (instead of nothing) but I'm not sure if that makes really sense.

PR #9012, opened 2021-03-08
Auto-select battle calc sides when using 'ctrl+A' and 'ctrl+D'

DanVanAtta:
This update adds two behaviors for the battle calculator when using
the hotkeys 'ctrl+A' and 'ctrl+D' to add attackers or defenders
to reduce having to switch the current attacker or defender
in the battle calc.

When adding attackers, if there are no matching attackers with
the current side, and there is a defender present in the battle
calc, then the attacking side is automatically set to the player
whith the most units in that territory. For example, if a player
has set up 1 defending UK infantry, and there is a territory
with 1 german tank, and the battle calc has italy set as the
attacker, then using control-a will automatically add the tank
and set the attacker to germany.

For defenders, if there are no units added to the battle calcuator
at all, then if a player uses 'ctrl+d', the defending side will
be set to match any unit that is in the current territory. For
example if a default calculator has germany as the defender, and
a player uses 'ctrl+d' in a russian territory with russian units,
then those units will be added to defenders and the defending
side will be automatically set to russia.

## Release Note

UPDATE|Battle calculator 'ctrl+A' and 'ctrl+D' will automatically set defending and attacking sides if there are no units currently set as the attacker or defender. For example, if there are UK attacking troops in a territory and the calculator has Russia selected as the attacker but no units were added to the calculator, then using 'ctrl+A' on the UK territory will change the attacker from Russia to UK and will add the UK units.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ai/AbstractBuiltInAi.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ai/AbstractBuiltInAi.java
@@ -372,7 +372,7 @@ public abstract class AbstractBuiltInAi extends AbstractBasePlayer {
             } else {
               final IntegerMap<Territory> distanceMap =
                   data.getMap().getDistance(capitals.get(0), doesNotHaveFactoryYet, it -> true);
-              picked = distanceMap.lowestKey();
+              picked = distanceMap.minKey();
             }
           }
         } else {
@@ -387,7 +387,7 @@ public abstract class AbstractBuiltInAi extends AbstractBasePlayer {
               final IntegerMap<Territory> distanceMap =
                   data.getMap().getDistance(test.get(0), territoryChoices, it -> true);
               for (int i = 0; i < maxTerritoriesToPopulate; i++) {
-                final Territory choice = distanceMap.lowestKey();
+                final Territory choice = distanceMap.minKey();
                 distanceMap.removeKey(choice);
                 test.add(choice);
               }
@@ -413,7 +413,7 @@ public abstract class AbstractBuiltInAi extends AbstractBasePlayer {
               }
               final List<Territory> choices = new ArrayList<>();
               for (int i = 0; i < maxTerritoriesToPopulate; i++) {
-                final Territory choice = distanceMap.lowestKey();
+                final Territory choice = distanceMap.minKey();
                 distanceMap.removeKey(choice);
                 choices.add(choice);
               }
@@ -443,7 +443,7 @@ public abstract class AbstractBuiltInAi extends AbstractBasePlayer {
           } else {
             final IntegerMap<Territory> distanceMap =
                 data.getMap().getDistance(capitals.get(0), notOwned, it -> true);
-            picked = distanceMap.lowestKey();
+            picked = distanceMap.minKey();
           }
         }
       }
--- a/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorDialog.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorDialog.java
@@ -1,6 +1,8 @@
 package games.strategy.triplea.odds.calculator;
 
+import games.strategy.engine.data.GamePlayer;
 import games.strategy.engine.data.Territory;
+import games.strategy.engine.data.Unit;
 import games.strategy.engine.history.History;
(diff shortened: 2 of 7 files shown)


Cernelius:
As is, the behaviour for the D one would need related refinements like being able to add multi-player defenses manually. Missing that, I'd skip the D part of this change for now, and do only the A part.

Cernelius:
Ignoring the D part, I suggest the A part (switching when clicking with 0 units of the current selected offender to add) applying only when there are no offenders already in the battlecalculator. Especially in confusing maps, I may be A-clicking on zones around just to make sure I'm not missing anything to add after I already have one or more offenders added in the battle-calculator.

DanVanAtta:
> As is, the behaviour for the D one would need related refinements like being able to add multi-player defenses manually.

Interesting point. That would be a another feature beyond what is here. Getting a good UI for that would be tricky, I think it would need some design work. I wonder though if it's common enough that you can't just "ctrl+D" units in question and perhaps have done with it, whether it's really needed.

The defender part is so that you don't have to manually toggle the defenders if you clear the battle calc and then select new defenders. I believe there is a use for that.

>  I suggest the A part (switching when clicking with 0 units of the current selected offender to add) applying only when there are no offenders already in the battlecalculator.

I believe that is what I have described, the switch only happens if there are no attacking units in the battle calc. So if you select only defenders, then if you use "ctrl+A" and there are any enemies in the territory not matching the current attacker selection, it will switch to them. For example, if the current attacker is Russia with Germany defending, and no Russian units listed, then using "Ctrl+A" with UK units would select them. OTOH if there is a territory with 2 Russian units and 4 UK units, because Russia is already selected, the 2 Russian would be added.

Cernelius:
If by
>When adding attackers, if there are no matching attackers with
the current side, and there is a defender present in the battle
calc, then the attacking side is automatically set to the player
whith the most units in that territory.

you mean "upon clicking Ctrl+A on a zone, the attacking side is automatically set to the player or to one of the players with the greatest number of additionable units in the zone if there are one or more defenders present in the battle-calculator and neither are there matching attackers with the current attacking side in the clicked zone nor are there one or more attackers already present in the battle-calculator", yes, but I don't see a reason for it not working with zero defenders present in the battle-calculator too.

DanVanAtta:
@Cernelius actually I was incorrect to say it depended on defending units being added. This is the condition: https://github.com/triplea-game/triplea/pull/9012/files#diff-e14a5a9de984a7af6b0ed30380538343e4e373f3148db551c3d4cd1cc918ae58R113-R116

The auto-attacker switch has three parts:
- no attacking units added
- current attacker selection does not match any units in the territory
- territory has at least one enemy unit

Cernelius:
> @Cernelius actually I was incorrect to say it depended on defending units being added. This is the condition: https://github.com/triplea-game/triplea/pull/9012/files#diff-e14a5a9de984a7af6b0ed30380538343e4e373f3148db551c3d4cd1cc918ae58R113-R116
> 
> The auto-attacker switch has three parts:
> 
> * no attacking units added
> * current attacker selection does not match any units in the territory
> * territory has at least one enemy unit

Maybe I'm being overly precise and I guess developers are just going to look at the code, but, just to make sure: do you mean:

> The auto-attacker switch happens only if all following conditions apply:
> 
> * no attacking units are already present in the battle-calculator
> * the current attacker selection does not match any units each of which is present in the zone and can be present as an attacker in the battle-calculator (for example, excluding "fort" units in my 270BC Wars game as they can never attack and the battle-calculator correctly does not offer you the option to add them as attackers (so, if I Ctrl+A on a territory with Rome forts and no other Rome units in it, the switch may happen like the forts are not there))
> * the zone has at least one unit which may be added as an attacker to the battle-calculator and belongs to a player which can be matched against the current defending player selected in the battle-calculator

About the last part, a side note is how that is determined. For example, if you add only "enemy", as you say, then you are making me impossible to calculate against current allied or neutral countries in a game with politics in which they may be enemy in the future and maybe on their very coming turn. An example is Germans and Russians being neutral in Global but possibly enemy on the next turn (and there may be custom games in which you go from allied to war at once).

DanVanAtta:
> . For example, if you add only "enemy", as you say, then you are making me impossible to calculate against current allied or neutral countries in a game with politics in which they may be enemy in the future and maybe on their very coming turn.

Hopefully not impossible beyond the limitations of the calc today. One can certainly manually select a given faction and input the units. That scenario perhaps does not change compared to what we have today.

PR #9019, opened 2021-03-09
Partial Restore of Unit scroller Buttons

DanVanAtta:
commit 2314a81ab75208cda010d3f72fe337e04234caa3

    Revert "Simplify unit scroller behavior (#8278)"
    
    This reverts commit 42cfad3a16fcc51c01f9b5b655d99522cc933639.

commit c7c7a142a170e44ec881649cb4ae8c76c8eb3aa3

    Remove unit scroller 'sleep button', assign 's' key to 'skip'

Before:
Screenshot from 2021-03-08 20-12-37

After:
Screenshot from 2021-03-08 20-11-36

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/panel/move/MovePanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/panel/move/MovePanel.java
@@ -1550,9 +1550,7 @@ public class MovePanel extends AbstractMovePanel {
 
   private void registerKeyBindings(final JFrame frame) {
     SwingKeyBinding.addKeyBinding(
-        frame, KeyCode.SPACE, unitScrollerAction(unitScroller::skipCurrentUnits));
-    SwingKeyBinding.addKeyBinding(
-        frame, KeyCode.S, unitScrollerAction(unitScroller::sleepCurrentUnits));
+        frame, KeyCode.S, unitScrollerAction(unitScroller::skipCurrentUnits));
     SwingKeyBinding.addKeyBinding(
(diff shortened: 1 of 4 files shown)


Cernelius:
I would have the order left to right as:
**previous**, **next**, **skip**, **unskip all**.

The reason is that the "skip" is a "next" which also skips (it has the same behaviour as the next plus also skipping the units), so it adds up to the skip function and the unskip all makes any sense only if the skip exists, so it should come thereafter.

It would also be good the button be shown as already pressed when there is no point pressing them (like showing the unskip all button as already pressed if nothing has been skipped yet).

Cernelius:
If "unskip all" has also the behaviour of getting you back to the starting point for the utility (as if clicking on "previous" for as many times you clicked on either of "next" and "skip"), then it could be seen as related to "back", so I would then suggest this order, left to right:
**unskip all**, **previous**, **next**, **skip**.

DanVanAtta:
@Cernelius PRs are not the best places for discussion threads. Closed PRs are usually best not commented on and instead referenced in new issues. Otherwise we will start mixing lots of topics in a single thread which makes it confusing and devalues the item when used as reference (ie: if this gets referenced, and has 30 pages of conversation, it will be lot less clear what is being discussed or decided). Furthermore, PRs are focused on specific changes and the conversation related should usually be directly related.

With all that said, this is probably the best place to discuss further unit-scroller changes: https://forums.triplea-game.org/topic/2458/unit-scroller-issues-rfc-on-changes/18

I'm neutral on the suggestion, I think it could be argued either way. Having the left hand arrow on the far left and right hand arrow on the far right makes it a bit easier in some ways too. There is also the argument to not change it for the sake of changing it as this is an existing 2.5 feature.

PR #9034, opened 2021-03-15
Read game name from map.yml when parsing game files.

DanVanAtta:
- Additionally, change parameter to Parser#load method
  to be a file instead of a URI.

## Testing

- verified default game selected when launching engine, loading save games and various downloaded maps and could start game and save and then load those saves


RoiEXLab (approved):
I'm not sure if this is just me, but I feel like there's an akward amount of toFile() calls in the modified code. I feel like if GameParser#parse would be changed to use Path instead of file a lot of these would go away.
But otherwise the code is fine

Issue #9028, opened 2021-03-14
2.6.89: FileUtils#readContents:184 - java.nio.charset.MalformedInputException


tripleabuilderbot:
## User Description
The notes for this game works fine in 2.5 yet cannot display if their code is pasted inside a blank "270BC_Wars.notes.html"

Here are some lines of the start and end of them:

				<html>
				<body>
				<table width=&quot;0px&quot;border=&quot;0&quot;bgcolor=&quot;776644&quot;style=&quot;text-align:center;font-size:20px&quot;>
				<tr>
				<th><img title=&quot;270BC Wars&quot; src=&quot;Title.png&quot;/>
				</th>
				</tr>
				<tr>
				<td>
				<div style=&quot;text-align:center&quot;>
				<p>
				</p><p><b>Version 1.0.0.0.0.0+</b>
				</p><p>
				</p><p><b>Game by Cernel</b>
				</p><p>for TripleA 2.6
				</p><p>
				</p><p>Modified from &quot;270BC&quot; version 1.7
				</p><p>
				<div style=&quot;font-size:16px;font-style:italic&quot;>Mistress of Italy, Rome, looking over the sea, discovers herself engaged in the inexorable struggle for survival against the might of Carthage, encroaching on Africa and Spain, and the islands nearby, as far as the columns of Hercules, the end of the World.
				</p><p>In Greece, the royal hegemony of Macedonia is bitterly challenged by a warring coalition of free cities and leagues, from Sicily, in the west, to Asia, in the east.
				</p><p>Across what remains of what was taken by Alexander, what had begun as a spate of civil conflicts conflated into a dynastic strife between the new realms of Egypt and Syria.
				</p><p>Beyond the Hellenistic world, the ascendancy of Parthia looms on the horizon, eager to champion the resurgence of Persian supremacy to her former greatness.
				</p><p>Meanwhile, hailing from unknown lands, the tribes of Numidia are swarming out of the wilds of Libya, their savage want for slaves and violent desire for booty, unquenchable.
				</p><p>'Tis a clash of civilizations: for either side, the other side is to be eliminated.
				</div>
				</p><p>
				</p><p>
				</div>

...

				<div style=&quot;text-align:center;text-indent:0px;color:#990000;background-color:000000&quot;>CREDITS
				</p><p>
				</div>
				<div style=&quot;text-align:center;text-indent:0px&quot;>
				</p><p>
				</p><p>Original 270BC map and game by: Doctor Che.
[...]

PR #9036, opened 2021-03-15
Use fallback encodings when reading files

DanVanAtta:
Allows for ISO-8859 files to be read if files
are not encoded as ASCII (UTF-8). ISO-8859 characters
may not render correctly but the file will at least
be read (avoiding relatively obscure error scenarios).

Resolves: https://github.com/triplea-game/triplea/issues/9028

### DIFF
--- a/lib/java-extras/src/main/java/org/triplea/io/FileUtils.java
+++ b/lib/java-extras/src/main/java/org/triplea/io/FileUtils.java
@@ -196,19 +196,13 @@ public final class FileUtils {
       }
 
       try {
-        return Optional.of(Files.readString(fileToRead, Charsets.UTF_8));
-      } catch (final MalformedInputException e) {
-        log.info(
-            "Warning: "
-                + "unable to read file (character encoding problem), will next try ISO-8859_1: {}, {}",
-            fileToRead.toFile().getAbsolutePath(),
-            e.getMessage());
-      }
-
-      try {
         return Optional.of(Files.readString(fileToRead, Charsets.ISO_8859_1));
       } catch (final MalformedInputException e) {
-        throw e;
+        log.warn(
+            "Bad file encoding on file: "
+                + fileToRead.toAbsolutePath().toString()
+                + ", contact the map maker and ask them to save this file as UTF-8");
+        return Optional.empty();
       }
     } catch (final IOException e) {
       log.error(
--- a/lib/java-extras/src/main/java/org/triplea/io/FileUtils.java
+++ b/lib/java-extras/src/main/java/org/triplea/io/FileUtils.java
@@ -2,6 +2,7 @@ package org.triplea.io;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.google.common.base.Charsets;
 import com.google.common.base.Preconditions;
 import java.io.File;
 import java.io.FileInputStream;
@@ -9,6 +10,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.nio.charset.MalformedInputException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Collection;
@@ -180,13 +182,39 @@ public final class FileUtils {
       return Optional.empty();
     }
 
+    // try to read the file with default character encoding, if fails then fallback
+    // to UTF-8 then try ISO-8859_1
     try {
-      return Optional.of(Files.readString(fileToRead));
+      try {
+        return Optional.of(Files.readString(fileToRead));
+      } catch (final MalformedInputException e) {
+        log.info(
+            "Warning: "
+                + "unable to read file (character encoding problem), will next try UTF-8: {}, {}",
+            fileToRead.toFile().getAbsolutePath(),
+            e.getMessage());
+      }
+
+      try {
+        return Optional.of(Files.readString(fileToRead, Charsets.UTF_8));
+      } catch (final MalformedInputException e) {
+        log.info(
+            "Warning: "
+                + "unable to read file (character encoding problem), will next try ISO-8859_1: {}, {}",
+            fileToRead.toFile().getAbsolutePath(),
+            e.getMessage());
+      }
+
+      try {
+        return Optional.of(Files.readString(fileToRead, Charsets.ISO_8859_1));
+      } catch (final MalformedInputException e) {
+        throw e;
+      }
     } catch (final IOException e) {
       log.error(
           "Error reading file: {}, {}", fileToRead.toFile().getAbsolutePath(), e.getMessage(), e);
-      return Optional.empty();
     }
+    return Optional.empty();
   }
 
   public static void writeToFile(final Path fileToWrite, final String contents) {
--- a/lib/java-extras/src/main/java/org/triplea/io/FileUtils.java
+++ b/lib/java-extras/src/main/java/org/triplea/io/FileUtils.java
@@ -175,7 +175,10 @@ public final class FileUtils {
 
   /**
    * Reads and returns the contents of a given file. Returns empty if the file does not exist or if
-   * there were any errors reading the file.
+   * there were any errors reading the file. Character encodings allowed: UTf-8, ISO_8859_1
+   *
+   * @throws MalformedInputException thrown if unrecongized character encodings found
+   * @throws IOException thrown if there are any problems reading the target file
    */
   public static Optional<String> readContents(final Path fileToRead) {
     if (!fileToRead.toFile().exists()) {
@@ -189,8 +192,7 @@ public final class FileUtils {
         return Optional.of(Files.readString(fileToRead));
       } catch (final MalformedInputException e) {
         log.info(
-            "Warning: "
-                + "unable to read file (character encoding problem), will next try UTF-8: {}, {}",
+            "Warning: file was not saved as UTF-8, some characters may not render:  {}, {}",
             fileToRead.toFile().getAbsolutePath(),
             e.getMessage());
       }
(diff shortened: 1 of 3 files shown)


Cernelius:
To be clear, I don't need this: all is working fine as of now. Maybe it's just better making the error more understandable (that the issue is the presence of unsupported characters and the file needs to be re-encoded as ASCII text).

It might be a bad idea to leave TripleA reading something having unnecessarily weird characters as it may be preferable keeping standards.

lib/java-extras/src/main/java/org/triplea/io/FileUtils.java
```diff
@@ -180,13 +182,39 @@ public static URL toUrl(final Path file) {
...
+      } catch (final MalformedInputException e) {
+        log.info(
+            "Warning: "
+                + "unable to read file (character encoding problem), will next try ISO-8859_1: {}, {}",
+            fileToRead.toFile().getAbsolutePath(),
+            e.getMessage());
+      }
+
+      try {
+        return Optional.of(Files.readString(fileToRead, Charsets.ISO_8859_1));
```
RoiEXLab, line 209: I know that Latin-1 is still commonly used in waaay too many places. But I don't think this is the right fallback handling to perform.
If for some arbitrary reason the encoding uses Latin-9 which is very similar except for 8 characters this will probably still work without any issues but the characters will simply be the wrong ones.

Also `Charsets` is a class from the google libraries, the default java way to do it is `StandardCharsets`.

RoiEXLab (changes requested):
I agree with @Cernelius here.
I don't think trying different charsets until no exception is thrown is the right move. Enforcing UTF-8 because it is the most widely used Unicode encoding out there is better in my opinion. Not only does it allow to use all characters of the Unicode standard, but this way we will never end up in a situation where a file is actually saved in a different encoding but happens to parse in a different encoding just with different characters.

DanVanAtta:
@RoiEXLab how do you propose we enforce UTF-8? I don't think we can and we will struggle with map makers who are asking "WTF is a charset? How do I see it? What, I didn't change it? How do I make it UTF-8?". OTOH if we fall back to other charsets, we avoid that. Yeah, it would be better to require only UTF-8, but it introduces problems requiring that. Similarly, requiring only LF line endings in a file modified by a map maker would be good too, but again that introduces more problems than just allowing CRLF. Bottom line, while it might smell not to have UTF-8, the purity requirement adds unnecessary overhead and troubleshooting on ourt part, while not ideal, it can be made to just work.

lib/java-extras/src/main/java/org/triplea/io/FileUtils.java
DanVanAtta: Good point on latin-8 vs latin-9.
- Better to have the content render than crash. 
- From what I've observed, any of the latin specific characters will not render well or at all correctly anyways. 
- I'd rather debug why a character is rendering weirdly than troubleshoot and train someone to use a UTF-8 text editor.

DanVanAtta:
Re:
> Enforcing UTF-8 because it is the most widely used Unicode encoding out there is better in my opinion. 

If we make this a requirement, let me put it this way, then someone must be available to explain and train anyone that runs afoul of the UTF-8 requirement. Simply put, there is no such someone.

RoiEXLab:
@DanVanAtta I think we are agreeing that UTF-8 is the desirable encoding for everything regarding this, right?

So I think we're actually discussing if we think that map makers are able to use UTF-8 by themselves.
I just checked: Every text editor on my machine (notepad.exe, Notepad++, Atom, VS Code, IntelliJ IDEA) uses UTF-8 by default. I actually wasn't sure if the default Windows editor did support proper UTF-8, but it looks like they changed it a while back so it is actually real UTF-8 without BOM.
That being said most if not all map makers shouldn't ever run into encoding issues if they are using popular & modern tooling.
Plus if they ever run into issues, regardless of type (crash, or different character displayed) we will end up with a ticket or a forum post regardless.
This change is only guaranteed to help in case you're using one of the additional 128 characters in ISO 8859-1 and does only prevent a crash, not necessarily fixing the issue using ascii extensions like ISO 8859 2 to 16, which are used even more rarely nowadays.

TL;DR I just don't think it's worth to maintain the extra code for an issue anyone will only ever have every couple years (I don't know if anyone before @Cernelius ever had this issue and reported it to us).

DanVanAtta:
Yeah, we are debating whether to require map makers to use UTF-8 encodings or to be more permissive. 

There is no available support to help map makers debug UTF-8 issues. Assuming issues come up, the energy to do that support at the cost of 8'ish lines of code in a 'lib' module is not a good trade off. If we do require map makers to do UTF-8, then we need to document it, give some handy how-to guides, survey all major text editors, find out more about how this happened in the first place, and provide support whenever it comes up. This lib module is also new, and there are new use-cases for editting more file types than just XML, the fact an early adopter out of a very small number of early adopters (I"m not sure if there are any, so possible the *first* person), is a very bad sign. Even if the number is small, a rare problem is almost worse than a frequent, as it'll take even more time and round-trips to grok what is happening. I don't think this is up for debate, we have no full time staff, and even if we did, supporting *this* problem at the cost of characters rendering not at all, is not worth it.

Cernelius:
I remember that, when I was working on WAW with @ron-murhammer, something like this came up that the game notes could not be read because of unsupported characters and was easily fixed by changing the unsupported characters.

All that is needed is that the message says that the file is not accepted because there are non-supported characters and preferably (but not necessarily) also telling exactly where (the line and such).

Side note, I reported this problem just because this "notes" file was a new thing, so I had no idea of what was working if anything. If it would have happened to anything else (which I know works), I would have easily guessed I must have written something wrong or unreadable without the need of any understandable warning.

In this case, however, I understand the unsupported character is some sort of unsupported white-space that looks like any other white-space to me (Right?), so I would have no clue. However, probably something very rare, as @RoiEXLab says.

And, yeah, chances are that I'm the first one trying to making a "notes" file manually.

I still don't really understand what a UTF-8 is, nor how I can tell my editor to be sure to generate a UTF-8 when I save the file, nor how I can see if an existent file is a UTF-8 or not, but I guess that as long as the file is UTF-8 and I don't add unsupported characters it should remain UTF-8.

The conclusion is that I'm really not seeing how it is a good idea to support substantially unacceptable things as that is burying problems which will accumulate overtime. Better to ensure that everything is correct by making sure that what is not does not work, so map-makers are forced to sort it out. Alternatively, always giving an error while making it work.

Cernelius:
> I just checked: Every text editor on my machine (notepad.exe, Notepad++, Atom, VS Code, IntelliJ IDEA) uses UTF-8 by default. I actually wasn't sure if the default Windows editor did support proper UTF-8, but it looks like they changed it a while back so it is actually real UTF-8 without BOM.

Have you ever taken a look at my favourite text editor?
http://www.jedit.org/

I find it better than Notepad++ and anything else I tried.

RoiEXLab:
@Cernelius I just installed it on my local windows machine and it looks like at least on windows the default charset is CP1252 (also called Windows-1252) which is very similar to ISO 8859-1 (Latin-1), but differs in some aspects. Judging by their "features" page where they advertise to support over 160 encoding and the fact that my java installation currently supports exactly 173 Encodings I would assume that jEdit simply relies on the encodings that come with java by default.
It probably also uses `Charset.defaultCharset()`, which - at least on my windows machine - returns said 'windows-1252' charset.
For you this means that in the settings menu you can simply change Cp1552 to UTF-8:
Screenshot

When using UTF-8 (or UTF-16 or UTF-32 for completeness) every Unicode character can be used, which includes 143,859 characters in Unicode 13 (standardized in 2020) which is probably every possible character you can imagine, including those of languages that are no longer spoken by anyone and of course emojis.
TL;DR: When using most non-unicode encodings you are limited to a set of roughly 255 characters, with Unicode when using UTF-8, UTF-16 or UTF-32 in all of their variants you can encode practically every character out there.

RoiEXLab:
@Cernelius In case you're curious, I found a good comparison table that sums up the problem with character encodings pretty well:
https://en.wikipedia.org/wiki/Western_Latin_character_sets_(computing)#Comparison_table
The "Code Point" column represents the number the unicode standard assigns to this character. All the columns to the right represent the byte representation of said character from 0x00 / 0 to 0xFF / 255 in their respective charset.
There are obviously more than 255 characters worldwide, so some characters have to be skipped in some charsets in order to include all the relevant ones (which is very subjective, the reason why there are so many charsets, unicode tries to be the standard to rule them all).

Regardless I noticed I have been using the terms "charset" and "encoding" synonymously in this issue, which is not always the case.
Many charsets only come with a single encoding, like Latin-1. But the Unicode Standard is a charset that comes with many UTF-X encodings that arrange the unicode code points in different ways for storage each approach with pros and cons

DanVanAtta:
Getting line numbers is not possible, a user will just be faced with the error message you observed. We might be able to say something more about needing to encode in UTF-8, I'm sure that will not be any more helpful though.

Using a different encoding is not wrong per-say, and we if we read the text content in the right encoding nothing is to say it won't render perfectly fine. The fault is arguably with the program for assuming UTF-8 when that is not always the case. UTF-8 is generally preferred, but it is not universally used. By the same token, latin encodings are relatively common, I wouldn't want us to bother going down the rabbit hole of supporting every encoding there is out there or spending even that much more time with it. Having a single fallback to support a very common encoding seems entirely reasonable.

If we do require only UTF-8, then we need a volunteer to improve the error messaging, update documentation, communicate to all users we will have this new requirement, and assist anyone in the future who has this problem. Nobody is able to make that commitment, it's a non-option. Even this dialog here has consumed more time than just allowing the encoding, I don't think there is a clear understanding of the negative impact (arguably none, just a few extra lines of code that someone in the future will have to read and because of where it is located likely never modify). So why exactly should we narrow the constraints the program for arguably no benefit just because UTF-8 "feels" clean? Second, why should we require only UTF-8 when it will cause support work for which there is no staff (and arguably it's a bad use of that staff's time if there were such a staff, even if payed, without evidence of significant negative impact, the application is being too narrowly constrained here).

PR #9039, opened 2021-03-15
Simplify File Path conversions

RoiEXLab:
Just a couple of things I stumbled across.

## Testing
None.


game-app/map-data/src/main/java/org/triplea/map/description/file/MapDescriptionYamlReader.java
```diff
@@ -37,7 +37,8 @@
      * 'map.yml' file at depth 2, eg: 'map_folder-master/map_folder/map/map.yml'.
      */
     final int maxMapYmlSearchDepth = 2;
-    return FileUtils.find(mapFolder, maxMapYmlSearchDepth, MapDescriptionYaml.MAP_YAML_FILE_NAME);
+    return FileUtils.find(
+        mapFolder.toPath(), maxMapYmlSearchDepth, MapDescriptionYaml.MAP_YAML_FILE_NAME);
```
RoiEXLab, line 41: I theory we could propagate the type of mapFolder to be path even further, but it would make the diff rather large so maybe in another PR.

PR #9074, opened 2021-03-27
Migrate from java.io.File to java.nio.Path (Part 2)

RoiEXLab:
This is part 2 of a series where java.io.File APIs get migrated to java.nio.Path

Part 1: #9039

## Testing
None

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/DownloadedMapsListing.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/DownloadedMapsListing.java
@@ -3,6 +3,7 @@ package games.strategy.engine.framework.map.file.system.loader;
 import games.strategy.engine.ClientFileSystemHelper;
 import games.strategy.engine.framework.ui.DefaultGameChooserEntry;
 import java.io.File;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Collection;
 import java.util.Comparator;
@@ -35,8 +36,8 @@ public class DownloadedMapsListing {
 
   private static Collection<DownloadedMap> readMapYamlsAndGenerateMissingMapYamls() {
     // loop over all maps, find and parse a 'map.yml' file, if not found attempt to generate it
-    return FileUtils.listFiles(ClientFileSystemHelper.getUserMapsFolder()).stream()
-        .filter(File::isDirectory)
+    return FileUtils.listFiles(ClientFileSystemHelper.getUserMapsFolder().toPath()).stream()
+        .filter(Files::isDirectory)
         .map(
             mapFolder ->
                 MapDescriptionYaml.fromMap(mapFolder)
--- a/game-app/game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/ZippedMapsExtractor.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/ZippedMapsExtractor.java
@@ -3,7 +3,6 @@ package games.strategy.engine.framework.map.file.system.loader;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import games.strategy.engine.ClientFileSystemHelper;
-import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -41,7 +40,7 @@ public class ZippedMapsExtractor {
    * found, then the progressIndicator is invoked with a callback that will execute the unzip task.
    */
   public void unzipMapFiles() {
-    final Collection<File> zippedMaps = findAllZippedMapFiles();
+    final Collection<Path> zippedMaps = findAllZippedMapFiles();
     if (zippedMaps.isEmpty()) {
       return;
     }
@@ -59,7 +58,7 @@ public class ZippedMapsExtractor {
                             newLocation ->
                                 log.warn(
                                     "Error extracting map zip: "
-                                        + mapZip.getAbsolutePath()
+                                        + mapZip.toAbsolutePath()
                                         + ", zip has been moved to: "
                                         + newLocation.toFile().getAbsolutePath(),
                                     zipReadException));
@@ -72,10 +71,10 @@ public class ZippedMapsExtractor {
                 }));
   }
 
-  private Collection<File> findAllZippedMapFiles() {
-    return FileUtils.listFiles(downloadedMapsFolder.toFile()).stream()
-        .filter(File::isFile)
-        .filter(file -> file.getName().toLowerCase().endsWith(ZIP_EXTENSION))
+  private Collection<Path> findAllZippedMapFiles() {
+    return FileUtils.listFiles(downloadedMapsFolder).stream()
+        .filter(Files::isReadable)
+        .filter(file -> file.getFileName().toString().toLowerCase().endsWith(ZIP_EXTENSION))
         .collect(Collectors.toList());
   }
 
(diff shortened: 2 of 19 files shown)


game-app/game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/DownloadedMapsListing.java
```diff
@@ -35,8 +36,8 @@ public static synchronized DownloadedMapsListing parseMapFiles() {
 
   private static Collection<DownloadedMap> readMapYamlsAndGenerateMissingMapYamls() {
     // loop over all maps, find and parse a 'map.yml' file, if not found attempt to generate it
-    return FileUtils.listFiles(ClientFileSystemHelper.getUserMapsFolder()).stream()
-        .filter(File::isDirectory)
+    return FileUtils.listFiles(ClientFileSystemHelper.getUserMapsFolder().toPath()).stream()
```
DanVanAtta, line 39: Why not have `ClientFileSystemHelper.getUserMapsFolder()` return a path? Is that an upcoming change?

game-app/game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/ZippedMapsExtractor.java
```diff
@@ -72,10 +71,10 @@ public void unzipMapFiles() {
                 }));
   }
 
-  private Collection<File> findAllZippedMapFiles() {
-    return FileUtils.listFiles(downloadedMapsFolder.toFile()).stream()
-        .filter(File::isFile)
-        .filter(file -> file.getName().toLowerCase().endsWith(ZIP_EXTENSION))
+  private Collection<Path> findAllZippedMapFiles() {
+    return FileUtils.listFiles(downloadedMapsFolder).stream()
+        .filter(Files::isReadable)
```
DanVanAtta, line 76: `isReadable` I don't think offhand is the same as 'isFile'. If not, then any directory that has a '.zip' extension would be returned (incorrectly). Please have a look into this to verify behavior.

game-app/game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/ZippedMapsExtractor.java
```diff
@@ -88,22 +87,22 @@ public void unzipMapFiles() {
...
       return Optional.empty();
     }
   }
 
-  private static Optional<File> unzipMapThrowing(final File mapZip) throws IOException {
-    Preconditions.checkArgument(mapZip.isFile(), mapZip.getAbsolutePath());
-    Preconditions.checkArgument(mapZip.exists(), mapZip.getAbsolutePath());
-    Preconditions.checkArgument(mapZip.getName().endsWith(".zip"), mapZip.getAbsolutePath());
+  private static Optional<Path> unzipMapThrowing(final Path mapZip) throws IOException {
+    Preconditions.checkArgument(Files.isReadable(mapZip), mapZip.toAbsolutePath());
```
DanVanAtta, line 100: Ditto on 'isReadable' != 'isFile'.

Unless there is a better API, this might need to be `!Files.isDirectory(mapZip))`
It could be worthwhile to wrap the File APIs in our own FileUtils so we that we will always have a consistent API regardless of implementation.

game-app/game-core/src/main/java/tools/image/DecorationPlacer.java
```diff
@@ -679,7 +680,7 @@ private void loadImageFolder() {
           || !imageFolder.getFile().exists()) {
         currentImageFolderLocation = null;
       } else {
-        currentImageFolderLocation = imageFolder.getFile();
+        currentImageFolderLocation = imageFolder.getFile().toPath();
```
DanVanAtta, line 683: Any reason to not change the 'getFile' method on FileSave to be 'getPath'?

game-app/game-core/src/main/java/tools/image/DecorationPlacer.java
```diff
@@ -779,7 +780,8 @@ private void fillCurrentImagePointsBasedOnImageFolder(
                   ? ImagePointType.SPACE_BETWEEN_NAMES_AND_PUS
                   : 0));
       final List<String> allTerritories = new ArrayList<>(centers.keySet());
-      for (final File file : FileUtils.listFiles(currentImageFolderLocation)) {
+      for (final Path path : FileUtils.listFiles(currentImageFolderLocation)) {
+        final File file = path.toFile();
         if (!file.getPath().endsWith(".png") && !file.getPath().endsWith(".gif")) {
```
DanVanAtta, line 785: This would be a good place to use FileUtils IMO, the conversion from path to file, back to path seems wonky. Do we even need to convert back to file here at all? Immediately we have a path, goes to a file, then goes back to a path again.

game-app/map-data/src/main/java/org/triplea/map/description/file/MapDescriptionYamlGenerator.java
```diff
@@ -79,32 +77,35 @@
...
         MapDescriptionYamlWriter.writeYmlPojoToFile(mapDescriptionYaml);
 
-    if (writtenYmlFile.isPresent() && propsFile.exists()) {
+    if (writtenYmlFile.isPresent() && Files.exists(propsFile)) {
       // clean up the properties file, it is no longer needed
-      propsFile.delete();
+      try {
+        Files.delete(propsFile);
+      } catch (final IOException exception) {
+        log.error("Failed to delete file " + propsFile.toAbsolutePath(), exception);
```
DanVanAtta, line 85: I don't know if we want to get an error report about this. So perhaps this should be a warn.

Second, a user should also know that they need to delete the original props file (where it is, and that it needs to be deleted). There is a behavior as well to move items into a 'bad-map' folder, that could perhaps be even better than instructing the user to delete the props file.

game-app/map-data/src/main/java/org/triplea/map/game/notes/GameNotesMigrator.java
```diff
@@ -41,16 +41,15 @@ public void extractGameNotes() {
...
-          xmlFilesMissingGameNotesFile.forEach(
-              xmlGameFile -> {
-                final String gameNotes = readGameNotesFromXml(xmlGameFile);
-                final String fileNameToWrite = GameNotes.createExpectedNotesFileName(xmlGameFile);
-                final Path fileToWrite = xmlGameFile.resolveSibling(fileNameToWrite);
-                log.info("Writing game notes file: " + fileToWrite.toFile().getAbsolutePath());
-                FileUtils.writeToFile(fileToWrite, gameNotes);
-              });
-        };
+        () ->
```
DanVanAtta, line 44: nit, recommend to keep changes consistent within a PR. This is seemingly a different kind of cleanup being done here.

DanVanAtta (changes requested):

game-app/game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/DownloadedMapsListing.java
RoiEXLab: Yes it is. `ClientFileSystemHelper.getUserMapsFolder()` has more than 10+ usages when I checked and that's simply too much for this PR

game-app/game-core/src/main/java/tools/image/DecorationPlacer.java
RoiEXLab: Will do in a future PR, didn't want to make the diff any larger than it already is

game-app/map-data/src/main/java/org/triplea/map/game/notes/GameNotesMigrator.java
RoiEXLab: Ah yes, wanted to move this to a different branch, but it got here by accident, my bad.

game-app/map-data/src/main/java/org/triplea/map/description/file/MapDescriptionYamlGenerator.java
RoiEXLab: Good point, previously any errors would've been silent if I'm not missing anything here

game-app/game-core/src/main/java/tools/image/DecorationPlacer.java
RoiEXLab: There is no need to convert it back, but note that `getPath()` != `toPath()`, the former returns a string, the latter a `java.nio.Path` object.
After doing some experiments it seems like `File#getPath` is the equivalent of `Path#toString` so I could easily replace that when I'm thinking about it

game-app/game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/ZippedMapsExtractor.java
RoiEXLab: This is a good thing to discuss actually.
There is `Files#isRegularFile`, which I hesitate to use because I ran into some issues with it a while ago.
For example `/dev/stdin` is not a regular file, even though it behaves like one and can be read just fine.
On the other hand if `mapZip` is a a file the current user isn't allowed to access `!Files.isDirectory(mapZip))` would obviously return true but fail somewhere down the line because of that.

So if you ask me checking if we can read the file is a better check for a precondition IMO, but I'm open to changing it

game-app/game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/ZippedMapsExtractor.java
RoiEXLab: You're right. I was so sure that isReadable returned false for directories, but apparently it doesn't.
Good spot 👍🏼

game-app/game-core/src/main/java/games/strategy/engine/framework/map/file/system/loader/ZippedMapsExtractor.java
DanVanAtta: I think this still needs to be resolved. The most likely error and reason for these as precondition checks is to guard against a programming error that we pass to this method a directory, instead of a zip file. The file name check seemingly will help a lot and the only way to have this method process a directory is if that directory has a '.zip' suffix (unlikely). I think there is value in having the check be explicit (and I will admit the original documentation within code was lacking), I think is very valuable.
In essence before we were checking:
- we are processing a file, not a directory
- the file exists
- the file is suffixed with .zip

The three conditions were meant to be a pretty good guarantee that we are in fact processing a zip file. One can actually go further and read the leading bits of a zip file to be really sure, though I did not do that here as it was arguably more complicated than worth it.

TL;DR, please add a simple check that we are processing a directory in addition to the other checks we have here. A check that we can 'read' the directory is a useful addition but not necessarily sufficient.

DanVanAtta: @RoiEXLab if you would do this as a follow-up, would be appreciated. I'll go ahead and merge now considering this is relatively minor (would like to see it be done), but it'll be an easy to review and submit one-liner if so. Meanwhile any 'part 3' would be unblocked :+1:

PR #9091, opened 2021-04-06
Fixed issue 2.6.89 by calling JOptionPane in a Swing event thread

bkelly1984:
Only one commit and one method modified.

## Testing
I was able to verify the problem in my own save game on a World War II Classic map with Java 15.0.2 running on Windows 10.

To reproduce, load [this save game](https://github.com/triplea-game/triplea/files/6262327/2021-4-5-World-War-II-Classic.zip). Then...

Move infantry from Panama to West Panama Sea Zone
Move transport and battleships from West Panama Sea Zone to Mexico Sea Zone
Move infantry from Mexico Sea Zone to Mexico
Press "Done" to end the combat phase

This does not work with the current master branch. It does work in my modification. I also verified that confirmation dialogs from Swing event threads still work, such as requesting to exit the program.

## Screens Shots
Failure
Success

First pull request. Please let me know if anything is incorrect.

## Release Note
FIX|Fixed a bug that caused confirmation dialogs to show no content.

### DIFF
--- a/lib/swing-lib/src/main/java/org/triplea/swing/EventThreadJOptionPane.java
+++ b/lib/swing-lib/src/main/java/org/triplea/swing/EventThreadJOptionPane.java
@@ -211,42 +211,64 @@ public final class EventThreadJOptionPane {
    * @return True if user confirms, false if user closes the confirmation dialog or selects no.
    */
   public static boolean showConfirmDialog(
-      final @Nullable Component parentComponent,
-      final @Nullable Object message,
-      final @Nullable String title,
-      final ConfirmDialogType confirmDialogType) {
+          final @Nullable Component parentComponent,
+          final @Nullable Object message,
+          final @Nullable String title,
+          final ConfirmDialogType confirmDialogType) {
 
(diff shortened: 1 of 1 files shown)


DanVanAtta:
Thank you for submitting a patch, great to see!

@bkelly1984 , did you have issues signing the CLA? It is reporting something like a username mismatch, is the automatic CLA tool broken?

Second, when testing the dialogs, did you double check they are still not-modal? IE: always-on-top but you can still scroll the map?

I re-triggered the build for unit-tests, it looks like it may have failed due to a flaky test.

DanVanAtta (approved):
Congrats on first PR  :100:

bkelly1984:
> @bkelly1984 , did you have issues signing the CLA? It is reporting something like a username mismatch, is the automatic CLA tool broken?

Not that I know of. My account was missing a profile name. I also clicked on the CLA link again and accepted it. When I look again, there is no option. Is it fixed?

> Second, when testing the dialogs, did you double check they are still not-modal? IE: always-on-top but you can still scroll the map?

I did. Thought about removing the setModal(false) line but found one could no longer scroll the map.

PR #9098, opened 2021-04-07
Show the sub retreat step even if a destroyer is present

### LINKED ISSUE #9069: TotalWorldWar error on TripleA
Could not find step name:Germany submerge subs?  If this problem happens frequently and is something you cannot fix, please report it to TripleA

trevan:
If the destroyer is killed during the AA phase, then the subs are
allowed to retreat BEFORE_BATTLE.  And if the destroyer is killed at
anytime during the round, the subs are allwed to retreat AFTER_BATTLE.

Fixes #9069 

## Release Note

FIX|Fix the "Could not find step name" error from showing up in TotalWorldWar.  This happens when  all the destroyers are killed during the Substrike phase.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreat.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreat.java
@@ -34,8 +34,9 @@ public class OffensiveSubsRetreat implements BattleStep {
 
   @Override
   public List<String> getNames() {
-    // do not check for destroyer because it could die before the retreat step
-    // and thus allow the retreat step
+    // even though the #execute method checks for destroyers, we don't do it here
+    // because this is called at the beginning of the round and any destroyer that exists
+    // might die before the #execute is called
     if (isEvaderNotPresent() || isRetreatNotPossible()) {
       return List.of();
     }
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreat.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreat.java
@@ -34,14 +34,9 @@ public class OffensiveSubsRetreat implements BattleStep {
 
   @Override
   public List<String> getNames() {
+    // do not check for destroyer because it could die before the retreat step
+    // and thus allow the retreat step
     if (isEvaderNotPresent() || isRetreatNotPossible()) {
-      return List.of();
-    }
-
-    if (getOrder() == SUB_OFFENSIVE_RETREAT_BEFORE_BATTLE && isDestroyerPresent()) {
-      // only check for destroyers if subs can retreat before battle
-      // because the destroyer could be killed during the battle which would
-      // allow the sub to withdraw at the end of the battle
       return List.of();
     }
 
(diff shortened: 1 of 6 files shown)


game-app/game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreatTest.java
```diff
@@ -107,21 +103,17 @@ void hasNameWhenDestroyerIsOnDefenseAndWithdrawIsAfterBattle() {
...
-                    .withSubmersibleSubs(true)
-                    .withSubRetreatBeforeBattle(true)
-                    .build())
+            .gameData(MockGameData.givenGameData().withSubmersibleSubs(true).build())
             .build();
     final OffensiveSubsRetreat offensiveSubsRetreat =
         new OffensiveSubsRetreat(battleState, battleActions);
 
-    assertThat(offensiveSubsRetreat.getNames(), is(empty()));
+    assertThat(offensiveSubsRetreat.getNames(), hasSize(1));
```
DanVanAtta, line 116: nit, mentioning in the assertThat comment what we expect in the 'offensiveSubsRetreat' would be good context. If this were to fail in the future due to the collection being empty, the coder/maintainer/bug-fixer might wonder why this test expected the collection to be non-emtpy.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreat.java
```diff
@@ -34,17 +34,12 @@
 
   @Override
   public List<String> getNames() {
+    // do not check for destroyer because it could die before the retreat step
```
DanVanAtta, line 37: This comment has me a bit confused. Is it here because of historical context, or is it going to be obvious in some other way that there *should* be a check for a destroyer here that would otherwise make a maintainer want to insert one? Perhaps the comment can be removed? It does seem like it makes sense if you knew what code used to exist here before, but maybe not if you look at the very latest.

DanVanAtta (commented):
Two smaller comments, otherwise looks good :+1:

game-app/game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreatTest.java
trevan: Fixed in https://github.com/triplea-game/triplea/pull/9098/commits/977c2cef1f80931235b9b0a68459f1298df945d3

game-app/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreat.java
trevan: The comment is needed.  In most of the battle steps, the conditions for the "getNames" and for "execute" are almost always identical.  But for this step, that isn't the case.  So the comment explains why the destroyer isn't checked in "getNames" even though it will be checked in "execute".

trevan:
@DanVanAtta , I pushed a commit with fixes and addressed your comments.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreat.java
DanVanAtta: Cool, perhaps a quick prefix to this comment would clarify that, eg:
"In most battle steps we check for destroyers, but here we do not because destroyers may die before the retreat step".

I think a logical extension though is "so what if the destroyers died before?". A second sentence on the comment to explain the implication would make that very clear. Let me know if that ask does not make sense.

game-app/game-core/src/test/java/games/strategy/triplea/delegate/battle/steps/retreat/OffensiveSubsRetreatTest.java
DanVanAtta: :+1:

DanVanAtta:
:+1:

PR #9110, opened 2021-04-09
Avoid usage of deprecated requestBody()

RoiEXLab:
This method was deprecated in openfeign 11

## Testing
None


http-clients/feign-common/src/test/java/org/triplea/http/client/JsonEncoderTest.java
```diff
@@ -35,8 +36,8 @@ void encodeObject() {
     jsonEncoder.encode(sampleObject, SampleObject.class, template);
 
     assertThat(
-        template.requestBody().asString(),
-        is(encodeWithGson(sampleObject).requestBody().asString()));
+        new String(template.body(), StandardCharsets.UTF_8),
+        is(new String(encodeWithGson(sampleObject).body(), StandardCharsets.UTF_8)));
```
RoiEXLab, line 40: Not sure if it's better to do array comparison here

PR #9111, opened 2021-04-09
Format EventThreadJOptionPane correctly

RoiEXLab:
Just a small formatting change, I always get this change when running spotless


DanVanAtta:
Makes me wonder if we are no longer running 'spotlessCheck' somewhere in the build.

RoiEXLab:
> Makes me wonder if we are no longer running 'spotlessCheck' somewhere in the build.

Yeah I was wondering the same thing

PR #9112, opened 2021-04-09
Migrate from java.io.File to java.nio.Path (Part 3)

RoiEXLab:
This is part 3 of a series where java.io.File APIs get migrated to java.nio.Path

Part 1: #9039
Part 2: #9074

## Testing
None

### DIFF
--- a/lib/java-extras/src/main/java/org/triplea/io/ZipExtractor.java
+++ b/lib/java-extras/src/main/java/org/triplea/io/ZipExtractor.java
@@ -29,6 +29,14 @@ public class ZipExtractor {
     }
   }
 
+  public static class ZipSecurityException extends RuntimeException {
+    private static final long serialVersionUID = 4558259887205500763L;
+
+    ZipSecurityException(final Path zipEntry) {
+      super(zipEntry.toString());
+    }
+  }
+
   /**
    * Indicates there was an error accessing or writing to the file system (zip is okay, but
    * extration failed).
@@ -93,11 +101,6 @@ public class ZipExtractor {
       }
     } catch (final IOException e) {
       throw new ZipReadException(fileZip, e);
-    } catch (final ZipSecurityException e) {
-      log.error(
-          "Malicious zip file detected: "
-              + fileZip.toAbsolutePath()
-              + ", please report this to TripleA and delete the zip file");
     }
   }
 
@@ -150,13 +153,5 @@ public class ZipExtractor {
     }
 
     return destFile;
-  }
-
-  private static class ZipSecurityException extends RuntimeException {
-    private static final long serialVersionUID = 4558259887205500763L;
-
-    ZipSecurityException(final Path zipEntry) {
-      throw new IllegalArgumentException("Malicious path: " + zipEntry);
-    }
   }
 }
--- a/lib/java-extras/src/main/java/org/triplea/io/ZipExtractor.java
+++ b/lib/java-extras/src/main/java/org/triplea/io/ZipExtractor.java
@@ -34,7 +34,7 @@ public class ZipExtractor {
     private static final long serialVersionUID = 4558259887205500763L;
 
     ZipSecurityException(final Path zipEntry) {
-      super(zipEntry.toString());
+      super(zipEntry + " is a forbidden path");
     }
   }
 
--- a/lib/java-extras/src/main/java/org/triplea/io/ZipExtractor.java
+++ b/lib/java-extras/src/main/java/org/triplea/io/ZipExtractor.java
@@ -1,12 +1,14 @@
 package org.triplea.io;
 
 import com.google.common.base.Preconditions;
-import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
+import java.nio.file.FileSystem;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import java.util.stream.Collectors;
-import java.util.zip.ZipEntry;
-import java.util.zip.ZipFile;
+import java.util.stream.Stream;
 import lombok.experimental.UtilityClass;
 import lombok.extern.slf4j.Slf4j;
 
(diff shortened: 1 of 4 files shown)


lib/java-extras/src/test/java/org/triplea/io/ZipExtractorTest.java
```diff
@@ -24,43 +48,41 @@
...
+    final Path unzippedTestFile2 = unzippedDirectory.resolve("test-file2");
+    final String testFile2Contents = Files.readString(unzippedTestFile2);
+    assertThat(testFile2Contents, is("lupsom osculus\n"));
   }
 
-  private static File testDataZip() {
-    final File file =
-        new File(ZipExtractorTest.class.getClassLoader().getResource("test-data").getFile());
-    return new File(file, "test-data.zip");
+  @Test
```
RoiEXLab, line 74: In order to add those tests I had to perform some slight refactoring to be able to test for the exception

lib/java-extras/src/main/java/org/triplea/io/ZipExtractor.java
```diff
@@ -105,23 +114,20 @@ public static void unzipFile(final File fileZip, final File destDir)
...
     } else {
       // fix for Windows-created archives
-      final File parent = newFile.getParentFile();
-      if (!parent.isDirectory() && !parent.mkdirs()) {
-        throw new FileSystemException("Failed to create directory " + parent);
-      }
-
-      try (FileOutputStream fos = new FileOutputStream(newFile)) {
-        fos.write(zipFile.getInputStream(zipEntry).readAllBytes());
+      final Path parent = newFile.getParent();
```
RoiEXLab, line 126: @DanVanAtta Do you remember what this fix was about? It seems pretty odd

lib/java-extras/src/main/java/org/triplea/io/ZipExtractor.java
```diff
@@ -61,37 +71,36 @@ public FileSystemException(final String message, final Exception e) {
...
     // iterate over each zip entry and write to a corresponding file
-    try (ZipFile zipFile = new ZipFile(fileZip)) {
-      for (final ZipEntry zipEntry : zipFile.stream().collect(Collectors.toList())) {
-        try {
-          unzipZipEntry(destDir, zipFile, zipEntry);
-        } catch (final IOException e) {
-          throw new FileSystemException(fileZip, e);
+    try (FileSystem zipFileSystem = FileSystems.newFileSystem(fileZip, null)) {
+      final Path zipRoot = zipFileSystem.getPath("/");
+      try (Stream<Path> files = Files.walk(zipRoot)) {
```
RoiEXLab, line 95: I incorporated my proposal from https://github.com/triplea-game/triplea/pull/8616#discussion_r555060868 into this PR.
There is a flaw with `Files#walk` that I noticed when crafting a malicious zip for testing though: If the malicious zip file uses ".." as a folder name, somewhere it is possible that this call walks forever because it recursively searches the same pair of folders forever until the heap runs out of free memory.
Obviously this doesn't happen with regular zip files, but we might want to consider limiting this to a depth of some arbitrary number like 500.

Also another thing to consider: IIRC Files#walk guarantees a fixed order the files are traversed in, so we might be able to drop the checks if the parent directories already exist because we already visited them in the past

RoiEXLab:
@DanVanAtta Ok this is odd, the tests are running fine on my machine.
I don't have a linux dev setup ready at the moment, so is there a chance you can have a look at the failure?
I suspect it has something to do with the zip file I crafted

lib/java-extras/src/test/java/org/triplea/io/ZipExtractorTest.java
```diff
@@ -25,42 +48,42 @@
...
+    final Exception exception =
+        assertThrows(
+            ZipExtractor.ZipSecurityException.class, () -> ZipExtractor.unzipFile(zip, subfolder));
+
+    assertThat(Files.exists(destinationFolder.resolve("matrix.jpg")), is(false));
+    // Make sure file isn't extracted at all
+    assertThat(Files.exists(subfolder.resolve("matrix.jpg")), is(false));
+
+    // Folder creation outside of the extraction directory should be prevented!
+    assertThat(exception.getMessage(), is("/.."));
```
DanVanAtta, line 87: Is "/.." really *the* exception message? That seems like not much for an exception message.
Second comment, testing exact message exceptions is brittle. Generally better to check that an exception contains key pieces of information. IN this case, do we even need to check anything on the exception at all really other than it was thrown?

DanVanAtta:
Just the one comment, otherwise LGTM

lib/java-extras/src/test/java/org/triplea/io/ZipExtractorTest.java
RoiEXLab: It's not that bad when it is printed as a Stacktrace:
```
ZipSecurityException: /..
```

but I get the point you're making

bkelly1984:
@RoiEXLab FYI, I am now getting this error in IntelliJ running on Windows 10 with Java 15.0.2:

```
java: reference to newFileSystem is ambiguous
  both method newFileSystem(java.nio.file.Path,java.lang.ClassLoader) in java.nio.file.FileSystems and method newFileSystem(java.nio.file.Path,java.util.Map<java.lang.String,?>) in java.nio.file.FileSystems match
```

Issue is with lib/java-extras/src/main/org/triplea/io/ZipExtractor.java on line 82. FileSystems.newFileSystem(fileZip, null) is overloaded and the compiler can't determine if "null" is of type Map<String ?> or ClassLoader.

RoiEXLab:
@bkelly1984 You have to use a jdk with version 11.
Newer versions introduced some methods causing this ambiguity

bkelly1984:
@RoiEXLab Understood, is JDK 11 the last officially supported Java release for TripleA?

RoiEXLab:
Yes, we might move to jdk 17 when that's released. However I think you could in theory cast null to the appropriate type and suppress the warning and create a PR if you want

bkelly1984:
Will do on JDK 11. I'll install it so I'm in line with everyone else.

PR #9113, opened 2021-04-09
Migrate from java.io.File to java.nio.Path (Part 4)

RoiEXLab:
This is part 4 of a series where java.io.File APIs get migrated to java.nio.Path

Part 1: #9039
Part 2: #9074
Part 3: #9112

## Testing
None

### DIFF
--- a/game-app/game-core/src/main/java/tools/image/AutoPlacementFinder.java
+++ b/game-app/game-core/src/main/java/tools/image/AutoPlacementFinder.java
@@ -10,12 +10,11 @@ import java.awt.Polygon;
 import java.awt.Rectangle;
 import java.awt.Shape;
 import java.awt.geom.Rectangle2D;
-import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Paths;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
@@ -46,7 +45,7 @@ public final class AutoPlacementFinder {
   private double unitZoomPercent = 1;
   private int unitWidth = UnitImageFactory.DEFAULT_UNIT_ICON_SIZE;
   private int unitHeight = UnitImageFactory.DEFAULT_UNIT_ICON_SIZE;
-  private File mapFolderLocation = null;
+  private Path mapFolderLocation = null;
   private final JTextAreaOptionPane textOptionPane =
       new JTextAreaOptionPane(
           null,
@@ -108,19 +107,19 @@ public final class AutoPlacementFinder {
   private void calculate() {
     // ask user where the map is
     final String mapDir =
-        mapFolderLocation == null ? getMapDirectory() : mapFolderLocation.getName();
+        mapFolderLocation == null ? getMapDirectory() : mapFolderLocation.getFileName().toString();
     if (mapDir == null) {
       log.info("You need to specify a map name for this to work");
       log.info("Shutting down");
       return;
     }
-    final File file = getMapPropertiesFile(mapDir);
-    if (file.exists() && mapFolderLocation == null) {
-      mapFolderLocation = file.getParentFile();
+    final Path file = getMapPropertiesFile(mapDir);
+    if (Files.exists(file) && mapFolderLocation == null) {
+      mapFolderLocation = file.getParent();
     }
     if (!placeDimensionsSet) {
       try {
-        if (file.exists()) {
+        if (Files.exists(file)) {
           double scale = unitZoomPercent;
           int width = unitWidth;
           int height = unitHeight;
@@ -196,7 +195,7 @@ public final class AutoPlacementFinder {
           }
         }
       } catch (final Exception e) {
-        log.error("Failed to initialize from map properties: " + file.getAbsolutePath(), e);
+        log.error("Failed to initialize from map properties: " + file.toAbsolutePath(), e);
       }
     }
     if (!placeDimensionsSet
@@ -269,16 +268,16 @@ public final class AutoPlacementFinder {
     }
     textOptionPane.appendNewLine("\r\nAll Finished!");
     textOptionPane.countDown();
-    final String fileName =
-        new FileSave("Where To Save place.txt ?", "place.txt", mapFolderLocation).getPathString();
+    final Path fileName =
+        new FileSave("Where To Save place.txt ?", "place.txt", mapFolderLocation).getFile();
     if (fileName == null) {
       textOptionPane.appendNewLine("You chose not to save, Shutting down");
       textOptionPane.dispose();
       return;
     }
-    try (OutputStream os = new FileOutputStream(fileName)) {
+    try (OutputStream os = Files.newOutputStream(fileName)) {
       PointFileReaderWriter.writeOneToMany(os, placements);
-      textOptionPane.appendNewLine("Data written to :" + new File(fileName).getCanonicalPath());
+      textOptionPane.appendNewLine("Data written to :" + fileName.normalize().toAbsolutePath());
     } catch (final IOException e) {
       log.error("Failed to write points file: " + fileName, e);
       textOptionPane.dispose();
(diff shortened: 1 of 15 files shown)


game-app/game-core/src/main/java/tools/image/FileHelper.java
```diff
@@ -0,0 +1,21 @@
+package tools.image;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
+public class FileHelper {
+
+  public static Path getFileInMapRoot(
```
DanVanAtta, line 10: Would you mind adding a javadoc to this method? Just to help a maintainer understand this more quickly:
- what is meant by map root?
- what happens if hte file is not found?

We should also consider creating a 'MapFolder' object to own methods like this.

game-app/game-core/src/main/java/tools/image/FileHelper.java
```diff
@@ -0,0 +1,21 @@
...
+      final Path mapFolderLocation, final String mapName, final String fileName) {
+    Path path = null;
+    if (mapFolderLocation != null && Files.exists(mapFolderLocation)) {
+      path = mapFolderLocation.resolve(fileName);
+    }
+    if (path == null || !Files.exists(path)) {
+      path = Path.of(mapName).resolveSibling(fileName);
```
DanVanAtta, line 17: Critical:  A 'mapName' can now be completely arbitrary, it may or may not at all match the map folder or any of the folders within a map. A map name is an arbitrary string in 'map.yml', it now no longer has any required relationship to map repo name or the folder name.

Given that, is this going off of the legacy design where the map name and map folder used to match?

game-app/game-core/src/main/java/tools/image/PolygonGrabber.java
```diff
@@ -77,13 +76,13 @@ private void runInternal() throws IOException {
     mapFolderLocation = MapFolderLocationSystemProperty.read();
     log.info("Select the map");
     final FileOpen mapSelection = new FileOpen("Select The Map", mapFolderLocation, ".gif", ".png");
-    final String mapName = mapSelection.getPathString();
+    final Path mapName = mapSelection.getFile();
```
DanVanAtta, line 79: Ouch,this legacy code is mostly incorrect here to assume that 'mapName' is the folder name  : (
I'm not sure how to fix it or if it matters, this perhaps could be "fixed" simply by renaming variables as this is no longer the map name but is just the map folder.

DanVanAtta (commented):
One significant comment/question about map-name no longer coming from the map folder. Please have a look.

game-app/game-core/src/main/java/tools/image/PolygonGrabber.java
RoiEXLab: Yeah true, the map making tools are mostly technical dept at this point it really feels unsatisfying to fix those

game-app/game-core/src/main/java/tools/image/FileHelper.java
RoiEXLab: > We should also consider creating a 'MapFolder' object to own methods like this.

Yeah, a lot of the code in the different classes is mostly copy-paste which is a bad smell. I'm sure if we were to extract the code to dedicated methods we can save like 50% of the code

RoiEXLab: In case you're not convinced by this statement, just check the usages of this newly created method. Even though I already extracted a lot of common functionality the calling code still looks pretty much the same on the calling end.

If you have the time you might be able to come up with a creative idea to unify a lot of this code, but I'll just focus on this PR series for now

game-app/game-core/src/main/java/tools/image/AutoPlacementFinder.java
```diff
@@ -269,16 +268,16 @@ private void calculate() {
...
+    final Path fileName =
+        new FileSave("Where To Save place.txt ?", "place.txt", mapFolderLocation).getFile();
     if (fileName == null) {
       textOptionPane.appendNewLine("You chose not to save, Shutting down");
       textOptionPane.dispose();
       return;
     }
-    try (OutputStream os = new FileOutputStream(fileName)) {
+    try (OutputStream os = Files.newOutputStream(fileName)) {
       PointFileReaderWriter.writeOneToMany(os, placements);
```
RoiEXLab, line 279: Note to future self: This can be simplified by refactoring the code to pass the path directly to PointFileReader writer

game-app/game-core/src/main/java/tools/image/FileHelper.java
RoiEXLab: Please have a look at the latest changes. It turns out that I could turn some Strings into Path objects and it makes a lot more sense now :)

game-app/game-core/src/main/java/tools/image/FileHelper.java
DanVanAtta: Some thought would need to go into creating an object model that finally encapsulates a lot of these concepts. I think we might have a good chunk of that for example with "downloadedMap", it's perhaps the same object and just needs to be broadened.

The primary ask here though is for java-doc'ing please to clarify this method. It almost seems like it should be in java-extras but then again it uses TripleA specific terms which could use defining. If anything, the javadoc would allow a maintainer to read the docs rather than reverse engineering the code, which is useful for reducing maintenance cost.

game-app/game-core/src/main/java/tools/image/FileHelper.java
DanVanAtta: We might not be on the same page. My point is the statement `Path.of(mapName)` is wrong. The `mapName` is _not_ the name of the map folder, the two are no longer the same. The map name is now decoupled from the file system. Hence, deriving the map name from map folder is no longer correct.

game-app/game-core/src/main/java/tools/image/PolygonGrabber.java
DanVanAtta: Agree, this update we can let slide I suppose. For a more generic util that might get re-used, in that case we should resolve the concern.

game-app/game-core/src/main/java/tools/image/FileHelper.java
RoiEXLab: @DanVanAtta I got that, however I think the at least this part of the code doesn't really care about the name of the map, I realized that this "mapName" that was supplied to `Path#of` was in fact the path to the map converted into a string at some point.
So yeah, you're right that mapName is misleading and I tried to change that to mapFolder whereever I saw it, but I think the latest code fixes some parts of this misleading structure

DanVanAtta: Let's see if we can avoid digging a hole deeper and maybe rename variables if it's only a semantic difference now. I'd suggest we make the class more specific as well, eg: `MapFileFinder`, and then see if we can avoid the 'mapName' name and call it something else. Perhaps ideally see if we can pass in the root folder first. There is actually already a util in `FileUtils` to do exactly that: `FileUtils#find`

RoiEXLab: Renaming the class sounds reasonable, is there anything else that should be done as part of this PR?
`FileUtils#find` is similar but not quite the same, I don't quite see how it can be used to shorten this helper method.

'mapName' isn't in the code anymore AFAIK, so there isn't anything to 'correct' there

DanVanAtta: Thanks, for the latter point, I am confused now.

I see the following:
```
  path = Path.of(mapName).resolveSibling(fileName);
```
Where `mapName` is a parameter. The above is calling something a mapName, which used to be completely synonmous with the folder name, and using that as a path parameter. Is this just as simple as renaming that as `mapFolderPath` or something like that?

RoiEXLab: The code you're seeing is outdated, have a look at the full diff and make sure you're looking at all commits

RoiEXLab:
@DanVanAtta I'm going to merge this now to push some other changes I have in the pipeline.
I will try to address any issues you raise after 👍🏼

Issue #9001, opened 2021-03-07
2.6.73 appear slow


TheDog-GH:
Testing 2.6.73 with my unpublished Settlers: War of the Lance (same as 2.6.39), the one with no enemy within 3-4 Territories.
A local game with 6AI, 2 Factions of 3 aside.

Speed

```
v 2.5.22294 43 40 44 40 45 44 45 mean = 43
v 2.6.73    60 65 63 62
```
Slower than 2.6.35

For reference here is the old version stats
v 2.6.35 55 51 46 49 48 45 47 mean ~ 49

DanVanAtta said
_For game play speed, 2.6.55 is seemingly the only thing that could have made a difference between *35 and the current. If you could check that version and the one preceding, that will narrow it down, but I don't readily see where the difference would come in._

TheDog-GH:
Adding more data and adding values for v2.6.83 gives the following result.

So Im thinking park this lot and maybe look at the speed issues at the end of the publishing of the public 2.6 ?

As I am only doing a very simple test and it does not pin down the actual point of it being slow, but if I had to guess though it would be in the combat phase.
 
```
2.6.83   59 58 56 59 58 =58
2.6.73   60 65 63 62 60 =62
2.6.71   54 56 55 57 59 ~56
2.6.67   60 62 57 56 55 ~58
2.6.60   54 56 56 56 54 ~55
2.6.59   Input ==null
2.6.58   Input ==null
2.6.57   Input ==null
2.6.56   Input ==null
2.6.55   No Code to test
2.6.40   Input ==null

Old data
v 2.6.35    55 51 46 49 48 45 47 mean ~ 49
v 2.5.22294 43 40 44 40 45 44 45 mean = 43

```

TheDog-GH:
Later versions than v2.6.35 & v2.5.22294 are slower, but are acceptable.  Speed will hopefully be addressed in the future.

TheDog-GH:
Downloaded the latest version to compare it to the speed the speed tests before, sadly its still slow compared to 2.5.

2.6.410   63 65 65 65 66 ~65 seconds 

compared to v 2.5.22294 = 43

TheDog-GH:
Some good news for a change.
Note this applies to version 2.6.477

2.6.477 50 50 47 54 47 ~ 50 seconds

**So back to 2.6.35 times.**

This improvement might be due to 2.6.472 if I understand it correctly, though Im not sure.

DanVanAtta:
[...]

PR #9130, opened 2021-04-13
Fix leaking InputStream

RoiEXLab:
2 Unit tests were failing on my machine because of that (JUnit tried to delete the file that was still open)

EDIT: Noticed this method was being misused at a couple more locations which lead to a small refactoring


game-app/game-core/src/main/java/games/strategy/triplea/ui/TooltipProperties.java
```diff
@@ -2,52 +2,23 @@
...
-          log.error("Error reading: " + PROPERTY_FILE, e);
-        }
-      }
-    }
+    super(PROPERTY_FILE);
   }
 
   public static TooltipProperties getInstance() {
-    // cache properties for 5 seconds
-    if (ttp == null || timestamp.plusSeconds(5).isBefore(Instant.now())) {
```
RoiEXLab, line 46: There's a small change in behaviour here, the cache timeout is now 10 seconds, but having a look at the code any timeout seems ridiculous, it's probably to avoid over-using the IO but it's very brittle to be honest

DanVanAtta: We may want/need this cache. You potentially will observe a slow-down on systems with a slower hard disk and during AI play. TheDog was timing AI play on age of tribes and was seeing a noticable slow down that was in part related to this method.

Why it is possibly needed - tooltip is eagerly created for all swing components even if not rendered. That means on any screen draw we're going to be reading hard disk quite a lot for a lot of the on-map entities. Maps like 'age of tribes' that have lots of units and lots of tooltip possibilities I think are where you will see the problem the most.

We'll need to consider this, I generally agree the cache and the way it is done is not ideal. The only way I know to really check this out is to measure AI play on age of tribes over a round or two. You might be able to get some help getting those measurements in the related issue thread: https://github.com/triplea-game/triplea/issues/9001

game-app/game-core/src/main/java/org/triplea/sound/SoundProperties.java
```diff
@@ -1,61 +1,29 @@
...
   private static final String PROPERTY_FILE = "sounds.properties";
   private static final String PROPERTY_DEFAULT_FOLDER = "Sound.Default.Folder";
   private static final String DEFAULT_ERA_FOLDER = "ww2";
-  private static SoundProperties instance = null;
-  private static Instant timestamp = Instant.EPOCH;
-
-  private final Properties properties = new Properties();
 
   SoundProperties(final ResourceLoader loader) {
-    final URL url = loader.getResource(PROPERTY_FILE);
```
DanVanAtta, line 27: Wow, this block is duplicated a full 5 times, nice work consolidating it :+1:

DanVanAtta (commented):
Couple comments to consider, please have a look @RoiEXLab

game-app/game-core/src/main/java/games/strategy/triplea/ui/TooltipProperties.java
DanVanAtta: Thinking about this one, we can use PerfTimer to verify the cost of this method and how often it's executed and exactly how much the cache is saving us. We need to keep in mind slower hard disks could be quite a bit slower executing the IO.

game-app/game-core/src/main/java/org/triplea/sound/PropertiesLoader.java
```diff
@@ -0,0 +1,38 @@
...
+import java.net.URL;
+import java.util.Optional;
+import java.util.Properties;
+import lombok.experimental.UtilityClass;
+import lombok.extern.slf4j.Slf4j;
+import org.triplea.java.UrlStreams;
+
+@UtilityClass
+@Slf4j
+public class PropertiesLoader {
```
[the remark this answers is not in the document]
RoiEXLab: Good idea, this will cause some extra code because suddenly we can no longer fall back to a default ResourceLoader but it fits better

game-app/game-core/src/main/java/games/strategy/triplea/ui/TooltipProperties.java
RoiEXLab: I think intelliJs profiling tools could help here too, I'll try to get some numbers here

RoiEXLab: Hmm I have the suspicion that my system isn't really a good machine to test this on.
I checked how long the constructors took to execute and it took less than 2ms every time, probably because I'm using an SSD. So unless this is called in a loop I doubt that I will measure a significant difference on my machine :/

DanVanAtta: Any luck profiling how long this method takes? Even if your system is fast, I think results that indicate if this a NS or a MS operation are going to be useful. If NS then we are likely fine even if it's 3 magnitudes slower.

DanVanAtta:
@RoiEXLab we need to be sure about the cache removal before doing so, looks like that is the only thing remaining here.

PR #9131, opened 2021-04-13
Make Unit-Test system independent

RoiEXLab:
This test was failing on windows because toURL() adds the drive letter in front of the path so the string suddenly doesn't match anymore

### DIFF
--- a/game-app/game-core/src/test/java/org/triplea/util/LocalizeHtmlTest.java
+++ b/game-app/game-core/src/test/java/org/triplea/util/LocalizeHtmlTest.java
@@ -12,7 +12,7 @@ import org.triplea.io.FileUtils;
 
 class LocalizeHtmlTest {
 
-  private final Path path = Path.of("/usr/local");
+  private final Path path = Path.of("/does/not/exist");
 
   @Test
   void testLocalizeHtml() {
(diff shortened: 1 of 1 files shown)


game-app/game-core/src/test/java/org/triplea/util/LocalizeHtmlTest.java
```diff
@@ -8,8 +8,12 @@
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
+import org.triplea.io.FileUtils;
 
 class LocalizeHtmlTest {
+
+  private final Path path = Path.of("/does/not/exist");
```
DanVanAtta, line 15: nit, i'd recommend to just duplicate this. The 'rules' of how to write clean tests are different from other code. Having a test be more self contained is valuable, particularly so when it means less redirection.

DanVanAtta:
Second nit, I updated the commit comment to match the PR description. Generally the valuable context is in the commit comment. I personally like it when the PR description autofills from the commit comment entirely. To keep in mind, PR is for the reviewer, the comments there-in are throw-away, and the commit comment is the long term and more valuable documentation of what changed and why.

PR #9132, opened 2021-04-13
Use Path instead of passing streams

RoiEXLab:
Using a Path which can also point to a file inside a zip (even though I think that's no longer needed) simplifies the code by a lot and avoids us having to pass streams around everywhere

## Testing
I verified a game could be launched as usual

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
@@ -7,13 +7,12 @@ import games.strategy.engine.framework.map.file.system.loader.DownloadedMapsList
 import games.strategy.engine.framework.startup.launcher.MapNotFoundException;
 import java.awt.Image;
 import java.awt.image.BufferedImage;
-import java.io.ByteArrayInputStream;
 import java.io.Closeable;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
-import java.io.InputStream;
 import java.net.MalformedURLException;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.nio.file.Path;
@@ -25,8 +24,6 @@ import javax.imageio.ImageIO;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.triplea.io.ImageLoader;
-import org.triplea.java.UrlStreams;
-import org.triplea.java.function.ThrowingSupplier;
 import org.triplea.swing.SwingComponents;
 
 /**
@@ -152,9 +149,7 @@ public class ResourceLoader implements Closeable {
    *     resource. Do not use '\' or File.separator)
    */
   public @Nullable URL getResource(final String inputPath) {
-    return findResource(inputPath) //
-        .or(() -> findResource(inputPath))
-        .orElse(null);
+    return findResource(inputPath).orElse(null);
   }
 
   /**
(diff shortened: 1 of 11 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
```diff
@@ -152,9 +149,7 @@ public boolean hasPath(final String path) {
    *     resource. Do not use '\' or File.separator)
    */
   public @Nullable URL getResource(final String inputPath) {
-    return findResource(inputPath) //
-        .or(() -> findResource(inputPath))
-        .orElse(null);
```
RoiEXLab, line 157: This is the regression I was talking about in the other PR, findResource was called twice

game-app/game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
```diff
@@ -179,33 +182,31 @@ public MapData(final String mapNameDir) {
...
-      final ThrowingSupplier<InputStream, IOException> inputStreamFactory,
-      final ThrowingFunction<InputStream, R, IOException> reader)
+  private static <K, V> Map<K, V> readOptionalMap(
+      final ResourceLoader loader,
+      final String path,
+      final ThrowingFunction<Path, Map<K, V>, IOException> mapper)
       throws IOException {
-    try (InputStream is = inputStreamFactory.get()) {
-      return reader.apply(is);
+    @Nullable final Path resourcePath = loader.optionalResource(path).orElse(null);
```
DanVanAtta, line 205: I'm surprised you did not use the functional variant,
```
return loader.optionalResource(path).map(mapper).orElse(Map::of);
```

game-app/game-core/src/main/java/org/triplea/util/PointFileReaderWriter.java
```diff
@@ -198,18 +186,18 @@ public static void writeOneToManyPlacements(
...
   /** Returns a map of the form String -> (points, overflowToLeft). */
-  public static Map<String, Tuple<List<Point>, Boolean>> readOneToManyPlacements(
-      final InputStream stream) throws IOException {
-    checkNotNull(stream);
+  public static Map<String, Tuple<List<Point>, Boolean>> readOneToManyPlacements(final Path input)
+      throws IOException {
+    checkNotNull(input);
 
     final Map<String, List<Point>> mapping = new HashMap<>();
     final Map<String, Tuple<List<Point>, Boolean>> result = new HashMap<>();
```
DanVanAtta, line 198: Would the util method: UrlStreams#openStream help us at all in any of this?

game-app/game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
```diff
@@ -173,33 +168,21 @@ public boolean hasPath(final String path) {
...
-    return UrlStreams.openStream(url)
-        .orElseThrow(() -> new IllegalStateException("Failed to open an input stream to: " + path));
-  }
-
-  public ThrowingSupplier<InputStream, IOException> optionalResource(final String path) {
-    return () ->
-        Optional.ofNullable(getResourceAsStream(path))
-            .orElseGet(() -> new ByteArrayInputStream(new byte[0]));
+  public Optional<Path> optionalResource(final String path) {
+    return findResource(path)
```
DanVanAtta, line 172: A future change could have `findResource` return a `Path`? Yeah, the URL is no longer needed, it was used to be able to index into zip files, we no longer need to do that.

DanVanAtta (commented):
Just a few discussion comments to consider

DanVanAtta:
Fwiw, you should be able to 'approve' code climate checks to turn that green. Totally appropriate when the problems are in existing code and not related to the immediate update:   https://codeclimate.com/github/triplea-game/triplea/pull/9132

game-app/game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
RoiEXLab: Would've liked to but the mapper potentially throws an ioexception that is supposed to be caught up the call tree

game-app/game-core/src/main/java/org/triplea/util/PointFileReaderWriter.java
RoiEXLab: I don't think urlstreams has any functionality regarding paths, but i didn't check

game-app/game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
RoiEXLab: Potentially. I only had a brief look at usages outside of my changes where URLs are used in a lot of places most of them can probably get migrated.
However the resource-loading itself uses UrlClassLoader to load files. It's really convenient because you can pass multiple directories/jar/zip-files as a source which automatically takes care of falling back to our assets we ship with triplea but preferring any assets that come with the map directly. This API only returns URLs though so we're stuck with that for a while unless we want to rewrite the resource loading code.
That being sad I'm sure there's a way to create a "FileSystem" using the nio APIs which has the same effect and creates a safe view of resources in a virtual filesystem

RoiEXLab:
> Fwiw, you should be able to 'approve' code climate checks to turn that green. Totally appropriate when the problems are in existing code and not related to the immediate update: https://codeclimate.com/github/triplea-game/triplea/pull/9132

A thanks for the reminder, forgot about that

game-app/game-core/src/main/java/org/triplea/util/PointFileReaderWriter.java
DanVanAtta: It would be reasonable to add some overloads to it if it's needed, overall though we may be able to avoid the need for paths and files altogeher. UrlStreams#openStreams opens and reads a URL stream and takes a function that process that input stream and returns a result. Hence any kind of intermediate download is not needed and removes a good bit of boiler plate code for closing and reading from streams.

game-app/game-core/src/main/java/games/strategy/triplea/ResourceLoader.java
DanVanAtta: Ok. To not get have the point be lost, I think most of the URL's we can replace with simple Path objects. I don't think that should require too much rewriting, mostly replacing of object types. The URLs do get converted to file system paths eventually, so should be tractable.

PR #9133, opened 2021-04-13
Migrate from java.io.File to java.nio.Path (Part 5)

RoiEXLab:
This is part 5 of a series where java.io.File APIs get migrated to java.nio.Path

Part 1: #9039
Part 2: #9074
Part 3: #9112
Part 4: #9113

## Testing
None


DanVanAtta:
Thank you for tackling these updates :100:  

@RoiEXLab side note, I think we might benefit by using projects more heavily. I somewhat think any thread of work perhaps should have a project to help organize it. If anything, a project would help keep us focused. Converting all 'File' usages to 'Path' seems like a decent example candidate. Something I've been thinking about.

RoiEXLab:
Good point, up until this point this seems like a thing I should've done, but with this PR I'm pretty much done with the changes I had in mind, now I need to search for remaining imports of the File class and go down another rabbit hole

PR #9186, opened 2021-05-02
Adds mod Zombies-World War 2

RogerCooper:
This replaces my earlier AAZ mod.

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2273,3 +2273,12 @@
     Edition.</p>
     <br><p>Key features include combining UK Pacific and ANZAC into one nation, revised
     order of play for nations, victory objectives and 8 rounds of play.</p>
+- mapName: Zombies-World War 2
+  mapCategory: EXPERIMENTAL
+  url: https://github.com/triplhttps://github.com/triplea-maps/zombies-World_War_2/archive/master.zip
+  version: 1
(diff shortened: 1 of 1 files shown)


RoiEXLab:
@RogerCooper There's a typo in the url

triplea_maps.yaml
```diff
@@ -2273,3 +2273,12 @@
     Edition.</p>
     <br><p>Key features include combining UK Pacific and ANZAC into one nation, revised
     order of play for nations, victory objectives and 8 rounds of play.</p>
+- mapName: Zombies-World War 2
+  mapCategory: EXPERIMENTAL
+  url: https://github.com/triplea-maps/zombies-World_War_2/archive/master.zip
+  version: 1
+  img: https://raw.githubusercontent.com/triplea-maps/zombies-World_War_2/master/preview.png
```
DanVanAtta, line 2280: This URL leads to a 404. 
Navigating to the image in the repo, clicking download and copying that link, the URL is: https://github.com/triplea-maps/zombies-World_War_2/raw/master/Preview.png
The difference is the 'P' vs 'p'. A lower case 'p' is preferred, renaming the image would be a preferred fix.

RogerCooper: Fixed

DanVanAtta: Re-ran checks, they passed now  :+1:

PR #9190, opened 2021-05-04
Update triplea_maps.yaml

beelee1:
New Map Upload

G 40 Expansion

## Testing

Have tested locally and map runs correctly.

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2278,6 +2278,6 @@
   url: https://github.com/triplea-maps/g_40_expansion/archive/master.zip
   version: 1
   img: https://raw.githubusercontent.com/triplea-maps/g_40_expansion/master/preview.png
-  description:
-    <p>G 40 House Rules Expansion Created by Captain and Company.</p>
+  description:<br>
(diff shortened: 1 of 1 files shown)


beelee1:
Well I'm out of ideas @DanVanAtta  When you get a chance would you look at this ?

Thanks

RoiEXLab:
@beelee1 The formatting is correct now, however the repository https://github.com/triplea-maps/g_40_expansion does not exist which is why the build fails

beelee1:
@RoiEXLab  thanks for the reply. I didn't realize the "I" was needed. I don't have it in the other map yamls. 

At any rate I just noticed it says "private" next to the repo in triplea-maps. i tried to transfer to triplea-maps and then tried pasting the url as well but the transfer button doesn't light up.

Do you know who i need to transfer it too ?

Thanks

beelee1:
Ok I'm closing this and starting over

RoiEXLab:
@beelee1 Ah if it says "private" this means the repo is only visible to members of triplea-maps, that's why I can't see the repo and the build script can't see it either.
You need to make sure the repository is public

PR #9202, opened 2021-05-08
Update triplea_maps.yaml

beelee1:
Trying this again. i thought maybe the url needed to be re pasted after being made public. Confidence isn't real high about that though but idk what else to try

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2273,3 +2273,12 @@
     Edition.</p>
     <br><p>Key features include combining UK Pacific and ANZAC into one nation, revised
     order of play for nations, victory objectives and 8 rounds of play.</p>
+- mapName: G 40 Expansion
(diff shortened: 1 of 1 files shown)


beelee1:
So this error keeps showing up.

> https://github.com/triplea-maps/g_40_expansion/archive/master.zip: map resource not available (404)

Sadly, I am completely clueless as how to fix it,

DanVanAtta:
What did you use to create the repository @beelee1? The default branch is called 'main' instead of 'master'. Not the first time I've seen this, there is something out there that really like the name 'main' while github (and arguably conventional) it should be called 'master'. You need to rename or create a new branch called 'master' and set it as the default branch.

beelee1:
ahh yea I remember oil snow having that problem now. Yea main is the default branch it makes when you follow the terminal commands.

PR #9205, opened 2021-05-09
Update maps-server backend, store last commit date instead of version

DanVanAtta:
Instead of storing 'version', we will be moving to last commit
date and to know if a map is out of date will look at the last
modified date on map folders.

This update changes the maps server backend to:
1) store last commit date in database instead of map version
2) return last commit date as an epoch milli as part of the
   web-API payload that provides available map information
   replacing 'version'

Next steps:
- update front end to consume map listing data from the maps server instead of triplea_maps.yaml
- add config for dev override of maps server and for front end to know which maps server to use. I'm strongly thinking to avoid this and instead merge the map server into the main lobby server.
-


lib/java-extras/src/main/java/org/triplea/java/StringUtils.java
```diff
@@ -73,4 +77,24 @@ public static String truncateEnding(
...
         : stringToTruncate;
   }
+
+  /**
+   * Reads the entire contents of an inputStream into a string. The inputStream is *not* closed by
+   * this method, the caller is responsible for closing the input stream.
+   */
+  public static String readFully(final InputStream inputStream) {
+    try {
+      return IOUtils.toString(inputStream, StandardCharsets.UTF_8);
```
RoiEXLab, line 87:
```suggestion
      return new String(inputStream.readAllBytes(), StandardCharsets.UTF_8);
```

This avoids the need for commons-io, should be the exact same behaviour

servers/database/src/main/resources/db/migration/maps_db/V1.00.08__maps_repo_store_last_comit_date.sql
```diff
@@ -0,0 +1,12 @@
+/* recreate map_index table with 'last_commit_date' column rather than a 'version' column. */
+drop table map_index;
+create table map_index
+(
+    id           serial primary key,
+    map_name     varchar(256) not null,
+    last_commit_date      timestamp not null check(last_commit_date < now()),
```
RoiEXLab, line 7: I think using `timestamptz` instead of `timestamp` is better here.
1. Because it is more consistent with `date_created` and `date_updated`
2. Because it makes the timestamp region independent (I think it can make a difference with dailight savings time and stuff, not sure though)

servers/database/src/main/resources/db/migration/maps_db/V1.00.08__maps_repo_store_last_comit_date.sql
```diff
@@ -0,0 +1,12 @@
+/* recreate map_index table with 'last_commit_date' column rather than a 'version' column. */
```
RoiEXLab, line 1: I'm actually surprised by this. I would've expected 2 alter table statements to modify the table, but if the table get's recreated every now and then this approach works too I guess :D

RoiEXLab (approved):

servers/database/src/main/resources/db/migration/maps_db/V1.00.08__maps_repo_store_last_comit_date.sql
DanVanAtta: Considering there is no data in the table, this was an okay route. I went this way so the column ordering made more sense, and was relatively easy all in all.

lib/java-extras/src/main/java/org/triplea/java/StringUtils.java
DanVanAtta: Cool, I agree we should avoid libraries when there is a relatively straight forward JDK equivalent. Updated: c81c844

servers/database/src/main/resources/db/migration/maps_db/V1.00.08__maps_repo_store_last_comit_date.sql
DanVanAtta: I was seeing TZ converted to local, but I wonder if that was just a fluke. I found a good article on this topic and it essentially said to have the app code convert to UTC before insert and store the DB column as 'tz', always. To avoid TZ hell, IMO it is best to have the DB be strictly UTC going in and coming out.

In sum, I agree & updated: 16ae75e

PR #9213, opened 2021-05-11
Fix memory leak caused by retained TripleAFrames.

asvitkine:
I noticed my TripleA was taking a lot of memory and took a heap
dump with JVisualVM. Most of the memory was byte arrays which
were backing Tiles in TileManager. But tracing their referenced,
I noticed I had lots of different TripleAframe instances around,
from games that were closed.

Turns out these were leaking. There were two problems:
  - TripleAframe wasn't being disposed on close.
  - ProMetricsUtils held a static that ended up with an eventual
    transitive reference to TripleAFrame.

Fixing both of these made the TripleAFrame instance disappear
from the heap dump once a game was closed. I'm simply removing
the ProMetricsUtils class since it was just being used for some
debug logging.

## Testing

I used the following steps to confirm the fix:

1. Open a game on a map and let AI play a few turns.
2. Close the map window.
3. Open JVisualVM, force a GC and take a heap dump.
4. Search for live instances of TripleAFrame, there should be none.

## Release Note

Fixed leaking memory when a game was closed.


RoiEXLab (approved):
Uhmmm






Do you know how many hours I **_literally wasted_** not finding this memory leak?!?!?!?!
I knew it was there. I knew the references stayed in memory but I just couldn't track it down. 😿 


So good job and thank you for this 👍🏼 
Very appreciated!

asvitkine:
> Uhmmm
> 
> Do you know how many hours I **_literally wasted_** not finding this memory leak?!?!?!?!
> I knew it was there. I knew the references stayed in memory but I just couldn't track it down. 😿
> 
> So good job and thank you for this 👍🏼
> Very appreciated!

Glad to help! Let me see if I can add a doc page on how to debug memory leaks that could be useful to others. JVisualVM is actually has some useful functionality that I used for this, such as finding the GC root of an object.

PR #9227, opened 2021-05-15
Support setting non-PU resources in edit mode.

asvitkine:
Add change resource function to edit mode. Previously, only the PUs resource
was changeable via edit mode. This expands the functionality to support all
resources (other than victory cities).

## Testing
Used the new functionality from edit mode to change a non-PU resource.
Tested that changing PUs on PU-only maps works.

## Screens Shots

Screen Shot 2021-05-31 at 8 21 27 PM

## Release Note

Non-PU resources can now be set in edit mode.


game-app/game-core/src/main/java/games/strategy/engine/framework/map/listing/MapListingFetcher.java
```diff
@@ -31,7 +31,8 @@ private MapListingFetcher() {
     if (ClientSetting.useMapsServerBetaFeature.getValue().orElse(false)) {
       // Get the URI of the maps server (either from override or read it from the servers file) and
       // then send an API call to it requesting the list of maps available for download.
-      return new LiveServersFetcher().getMapsServerUri()
+      return new LiveServersFetcher()
+          .getMapsServerUri()
```
asvitkine, line 35: This got formatted when I ran spotlessJavaApply.

game-app/game-core/src/main/java/games/strategy/triplea/ui/ResourceChooser.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import javax.swing.DefaultListCellRenderer;
+import javax.swing.JDialog;
+import javax.swing.JList;
+import javax.swing.JOptionPane;
+import javax.swing.ListSelectionModel;
+import org.triplea.swing.SwingComponents;
+
+class ResourceChooser extends JOptionPane {
```
asvitkine, line 19: I modelled this after PlayerChooser. We could potentially try to make a common base class, but not sure it's worth the extra complexity since the class is relatively short.

asvitkine:
One question I had is if this is a version-breaking change?

I looked at the comments for `@RemoteActionCode` and it only says changing the signature or method number shouldn't be changed. But nothing about adding new methods.

Will having this new method cause games not to be compatible? Or would it only be a runtime error if someone tries to actually use the new method? I've kept the existing method for changing PUs, so that should still be callable.

DanVanAtta:
Good question about compatibility. If an older client is missing the network method, then the new client and old client would presumably have an issue. You can test this functionality by hosting a local game with this branch and then joining to that same game , at which point you can exercise the updated edit mode method and see how it goes.

asvitkine:
> Good question about compatibility. If an older client is missing the network method, then the new client and old client would presumably have an issue. You can test this functionality by hosting a local game with this branch and then joining to that same game , at which point you can exercise the updated edit mode method and see how it goes.

Thanks! I tried it - host on 2.5 and joiner on 2.6 worked fine, except when using the new functionality - which would give a recoverable error and the game can continue. I changed the functionality to still use the changePUs() function when PUs are edited.

If host is on 2.6 and joiner is on 2.5, the 2.5 client gets an NPE but it's unrelated to my change. If I run from master branch, the same happens.

I've updated the testing information in the PR description.

DanVanAtta:
Is any 2.5 client getting a NPE when joining a 2.6 host? That should not be the case...

I don't think we should be releasing a feature that needs a asterisk about network compatibility. We can require all clients to be on 2.6 to join the same lobby OTOH.

asvitkine:
> Is any 2.5 client getting a NPE when joining a 2.6 host? That should not be the case...
> 
> I don't think we should be releasing a feature that needs a asterisk about network compatibility. We can require all clients to be on 2.6 to join the same lobby OTOH.

What's the way to do that?

Otherwise, I could make the feature only offer to change PUs in a 2.5 game.

For the NPE, I wasn't able to get a different behavior in local testing. Do you want to try? This is happening without my change.

DanVanAtta:
The NPE is seemingly a regression and needs to be fixed.
To require players on 2.6 we'd set up a second lobby and send those players to the 2.6 lobby via: https://github.com/triplea-game/triplea/blob/master/servers.yml

asvitkine:
> The NPE is seemingly a regression and needs to be fixed.
> To require players on 2.6 we'd set up a second lobby and send those players to the 2.6 lobby via: https://github.com/triplea-game/triplea/blob/master/servers.yml

Since the NPE is in the old client, how do we fix that? I guess we have to figure what the new client is doing differently from before? I think this is related to the player list UI before a game starts - probably whoever made changes to the networking code around that should investigate? Who worked on that?

asvitkine:
Is there a way to make this change condition on whenever we break compatibility? Ideally, I'd like to land the improvement but fall back to the previous behavior until we're past the next compatibility boundary.

DanVanAtta:
Given it is network only compatibility, we could launch 1.6 as a stand-alone lobby and migrate users to it. Otherwise check in the network code if there is a version that is shared, I don't think there is.

You could potentially add that though, so that a client shares version with server and vice versa. If the version value is not shared then you can assume 2.5, then if everyone is beyond 2.5 this feature would be enabled. It's a bit convoluted though : |

It might be easier rebuilding this into 3.0

asvitkine:
I updated this to streamline the case where there's only PUs to skip the new dialog.

I think we should just land this. Yes, if someone tries to edit a resource on a map that uses resources on a game that's hosted on 2.5, they'll get an error. But nothing bad happens otherwise.

I think it's acceptable because:
  1. This is an optional "edit mode" feature, which would be used very rarely.
  2. The case where it doesn't work is entirely recoverable with no downside other than an error message shown to the user.

I think given the above, we should just land this without worrying about doing something complicated like creating new lobbies or detecting game versions - or delaying until 3.0 or similar
@DanVanAtta WDYT?

DanVanAtta:
I'm not a fan of releasing incompatible features. That is my first reaction, perhaps that will soften over time if I think about it more.

I see four options:
1. release incompatible feature
2. Disable when clients under 2.6 are present. To do this, add a version value to the network communication. If the version value is not transmitted, then you can assume a 2.5 client.
3. Dedicate a 2.6 lobby
4. hold off on the feature for now

(3) is potentially how we will release 2.6. I think there are users still on 2.2 and earlier that causes issues in the bots. Forcing an upgrade has not happened for some time and is not necessarily a bad thing.
(2) is a good path but perhaps will be a lot of work compared to that being more or less built in to V3.0, I don't think the ROI is that great.

If we go with (1), perhaps if we make the error message very clear that the other user is not using a compatible engine and that the operation did not succeed (and then disable the option going forward). Essentially do a really nice job with the error handling if we are going to let it fail. We'll also want to test it very well with a host/client/observers and a bot.

DanVanAtta:
Working on maps a bit, we will want to force users to upgrade to 2.6 sooner. Until then map makers will need to duplicate their game notes.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/EditDelegate.java
```diff
@@ -172,21 +172,35 @@ public String changeTerritoryOwner(final Territory territory, final GamePlayer p
...
     }
-    final Resource pus = getData().getResourceList().getResource(Constants.PUS);
-    final int oldTotal = player.getResources().getQuantity(pus);
+    final Resource resource = getData().getResourceList().getResource(resourceName);
+    final int oldTotal = player.getResources().getQuantity(resource);
     if (oldTotal == newTotal) {
-      return "New PUs total is unchanged";
+      return "New " + resourceName + " total is unchanged";
     }
     if (newTotal < 0) {
```
DanVanAtta, line 190: Would it make sense to zero this out, or change the error message to say the new total must be zero or greater (rather than simply "invalid")

game-app/game-core/src/main/java/games/strategy/triplea/delegate/EditDelegate.java
```diff
@@ -172,21 +172,35 @@ public String changeTerritoryOwner(final Territory territory, final GamePlayer p
...
-    bridge.addChange(ChangeFactory.changeResourcesChange(player, pus, (newTotal - oldTotal)));
+        "Changing "
+            + resourceName
+            + " for "
+            + player.getName()
+            + " from "
+            + oldTotal
+            + " to "
+            + newTotal,
+        null);
```
DanVanAtta, line 202: Nit, any thoughts to using `String.format` here?

game-app/game-core/src/main/java/games/strategy/triplea/delegate/remote/IEditDelegate.java
```diff
@@ -32,6 +32,9 @@
   @RemoteActionCode(3)
   String changePUs(GamePlayer player, int pus);
```
DanVanAtta, line 33: What's the consideration to keeping the changePU method around?
I'd be in favor to remove it in favor of the new updated and more generic method.

game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
```diff
@@ -481,7 +485,7 @@ public void actionPerformed(final ActionEvent event) {
           }
         };
     changePUsAction =
-        new AbstractAction("Change PUs") {
+        new AbstractAction("Change PUs/Resources") {
```
DanVanAtta, line 488: How do you feel about the "/" ?

IMO this would be better if it were just "Change Resources".
We could alternatively try to compute if there is only one resource and use that for the name. That seems complex and arguably inconsistent. What are your thoughts to simply changing the button title to "Change Resources?"

game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
```diff
@@ -490,33 +494,40 @@ public void actionPerformed(final ActionEvent event) {
             setWidgetActivation();
             final PlayerChooser playerChooser =
                 new PlayerChooser(getData().getPlayerList(), getMap().getUiContext(), false);
+            final Container parent = getTopLevelAncestor();
             final JDialog dialog =
-                playerChooser.createDialog(getTopLevelAncestor(), "Select owner PUs to change");
+                playerChooser.createDialog(parent, "Select Owner PUs/Resources to Change");
```
DanVanAtta, line 499: What would you think if the title of the dialog were a more simple "Change Resources"?

game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
```diff
@@ -490,33 +494,40 @@ public void actionPerformed(final ActionEvent event) {
...
+            if (resources.size() == 1) {
+              resource = resources.get(0);
+            } else {
+              final ResourceChooser chooser =
+                      new ResourceChooser(resources, getMap().getUiContext());
+              resource = chooser.showDialog(parent, "Select resource to change");
+              if (resource == null) {
+                cancelEditAction.actionPerformed(null);
+                return;
+              }
```
DanVanAtta, line 520: We are getting into long method territory. Have you considered extracting the entirety of the edit resources dialog to it's own class? Seemingly edit panel should not be aware of how to implement every single edit function in its entirety.
Another topic, this is getting kinda dense, some guiding comments or breaking this up into helper methods would be very helpful and avoid a maintainer from having to reverse engineer the code to get an idea what is happening.

Another question, why is VPS excluded?

game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
```diff
@@ -528,12 +539,16 @@ public void actionPerformed(final ActionEvent event) {
             }
             int newTotal = oldTotal;
             try {
-              newTotal = Integer.parseInt(pusField.getText());
+              newTotal = Integer.parseInt(totalField.getText());
             } catch (final Exception e) {
               // ignore malformed input
             }
+            final IEditDelegate delegate = EditPanel.this.frame.getEditDelegate();
+            // Use the old changePUs() API since changeResource() was only added in 2.6.
```
DanVanAtta, line 547: I think you can assume 2.6 given the other considerations of how we want players to upgrade to 2.6

game-app/game-core/src/main/java/games/strategy/triplea/delegate/remote/IEditDelegate.java
asvitkine: It was left so that we could be compatible with pre-2.6 clients. If that's not a goal, happy to remove.

game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
asvitkine: I don't have a strong opinion, so changing seems fine. The only reason is discoverability - someone may not realize that PUs are a type of resources. But this is edit mode, which is a pretty advanced feature to begin with, so I'm not too worried.

game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
asvitkine: SG.

game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
asvitkine: OK

game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
asvitkine: I excluded VPS because it's also excluded in other parts of the code - e.g. the economy panel which shows the actual resources.

If you think they should be included, I can keep them.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/EditDelegate.java
asvitkine: Done. Changed to "must not be negative".

asvitkine (commented):
> We are getting into long method territory. Have you considered extracting the entirety of the edit resources dialog to it's own class? Seemingly edit panel should not be aware of how to implement every single edit function in its entirety.

I went with adding some helper methods which I think make the code much easier to follow. Technically it's already its own class, just an anonymous one that subclasses AbstractAction. We could move it to its own file, but I feel if we do that, we should do it consistently to all of them - and so it should be outside the scope of this PR.

OTOH, please take a look at the new code flow - I think it should address your concern and make the code much easier to follow. (I didn't add much comments and instead used method names to help document the flow.)

game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
DanVanAtta: I'm actually not sure what VPS really even is!  = D
(Victory points??)
Hence the exclusion was interesting.

game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
```diff
@@ -480,68 +483,95 @@ public void actionPerformed(final ActionEvent event) {
...
             }
-            getData().acquireReadLock();
-            final Resource pus;
-            try {
-              pus = getData().getResourceList().getResource(Constants.PUS);
-            } finally {
-              getData().releaseReadLock();
+
+            final Optional<Resource> resource = chooseResource();
+            if (resource.isEmpty()) {
```
DanVanAtta, line 502: nit, rather than doing:
```
 final Optional<Resource> resource = chooseResource();
   if (resource.isEmpty()) {
         return;
    }
    resource.get()
```

I found it's easier to do this:
```
 final Resource resource = chooseResource().orElse(null);
   if (resource == null) {
         return;
    }
    resource
```

That avoids the need to do a 'resource.get' later. My rule of thumb for optional is one should always avoid passing it as a parameter and secondarily avoid assignment to an option (but it's okay to return an optional, that also signals that all error handling & notifications have been performed).

DanVanAtta: I think you might be able to use the above pattern to avoid quite a number of `.get()` method calls later, WDYT? Worth updating?

game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
```diff
@@ -480,68 +483,95 @@ public void actionPerformed(final ActionEvent event) {
...
+          }
+
+          private Optional<Resource> chooseResource() {
+            // Ignore VPS resources, since that's what the economy panel does.
+            final List<Resource> resources =
+                getData().getResourceList().getResources().stream()
+                    .filter(r -> !r.getName().equals(Constants.VPS))
+                    .collect(Collectors.toList());
+            if (resources.size() == 1) {
+              return Optional.of(resources.get(0));
```
DanVanAtta, line 542: It's possible for there to be a null element in the resource list? This is surprising and seemingly painful.

game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
```diff
@@ -480,68 +483,95 @@ public void actionPerformed(final ActionEvent event) {
...
-                EditPanel.this.frame.getEditDelegate().changePUs(player, newTotal);
-            if (result != null) {
-              JOptionPane.showMessageDialog(
-                  getTopLevelAncestor(),
-                  result,
-                  "Could not perform edit",
-                  JOptionPane.ERROR_MESSAGE);
-            }
-            cancelEditAction.actionPerformed(null);
+            return Optional.of(newTotal);
```
DanVanAtta, line 574: Rather than returning the old total, why not return an optional empty? (sorry for  not looking this up myself) - Does the empty optional return value trigger a no-op or an error message?

DanVanAtta (commented):
Certainly much easier to follow. Just a few more comments to address, almost there : )

game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
asvitkine: Thanks, made the change. Agree it's nicer.

game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
asvitkine: No. Optional.of() doesn't accept nullable, you're thinking .ofNullable().

game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
asvitkine: Actually, changed to IntTextField which eliminates the need to parse an int.

DanVanAtta:
Thanks for working through this update @asvitkine .
Looks like you addressed all the feedback :+1:

PR #9230, opened 2021-05-15
Clean up code in ProTerritoryManager.java.

asvitkine:
Clean up code in ProTerritoryManager.java.

Moves some repeated logic into a function and extracts common constructs into local variables to make the code more concise. No functional changes.

This should result in some performance improvements since some re-assignment of variables is turned into ternary statements, avoiding doing a useless computation where the result was thrown away before.

## Testing

None.


asvitkine:
The codeclimate warnings are on existing code and should be ignored.

asvitkine:
> The codeclimate warnings are on existing code and should be ignored.

@DanVanAtta is there a way to suppress codeclimate so it doesn't appear as an [X] when the issues are pre-existing?

DanVanAtta:
If a person has write access, then you can click 'details' on the code climate check and there is an 'approve' button that will green-up the check. I just did that.

RoiEXLab (approved):

PR #9236, opened 2021-05-16
AI: Don't leave loaded units on transports at end of turn.

asvitkine:
AI: Don't leave loaded units on transports at end of turn.

I've observed this happening at least on maps where AI had trouble executing its planned amphibuous moves because it
didn't account for fuel.

By adding this additional logic, the AI will at least re-land any loaded units, which is much safer than leaving them on transports.

Only units that weren't loading this turn in combat phase are considered, since unloading units during non-combat that were loaded in the combat phase is not permitted by the rules.

## Testing

Loaded a save game on Iron War where British AI would leave units on ships
and observed these are now landed at end of turn.

## Release Note

AI: Don't leave loaded units on transports at end of turn.


Cernelius:
@asvitkine If you mean that those cargo units are going to be offloaded during a different phase than the phase during which they were loaded, this is obviously an illegal move: as far as I know, in every intended rules-set but V2 Basic (where no unit can), only air units can ever move on more than one phase during the same turn (so, in every intended rules-set, non-air units can never move on more than one phase during the same turn).

---

For the record, I'm against enforcing this limitation unless a property is made not to have it for custom games (so that you can load land units on ships during Combat Move and offload them on the same turn during Non Combat Move (which may be a legal move under the rules of my "270BC Wars" game)). But I guess it is not wanted the AI intentionally to make illegal moves which the engine is allowing.

---

Aside from this, if you keep offloading and re-loading units in a game with fuel costs, you will keep consuming (likely, wasting) such resources. So, a map with fuel costs should rather be the foremost example in which you would rather not offload your units (to avoid having to pay the fuel cost again later to re-load them) unless they have a high chance to be lost.

DanVanAtta:
I'd agree it's better for AI to disembark units than leave them loaded. It's something of a rare situation where you actually want troops to stay loaded. WaW pacific theater has examples, but almost always it's a bad thing.

At the same time, this does seem like a band aid. Shouldn't the AI instead aim to never load those troops in the first place? If the AI tries to load a transport, then leaves the units, it also seems that the AI played suboptimally and wasn't movement by having units shuffle back and forth.

Can you provide any insight about the complexity to try and update the loading logic? I'd suspect if the AI can't load and then unload in the same phase that the AI should instead consider and execute a different move.

asvitkine:
> I'd agree it's better for AI to disembark units than leave them loaded. It's something of a rare situation where you actually want troops to stay loaded. WaW pacific theater has examples, but almost always it's a bad thing.
> 
> At the same time, this does seem like a band aid. Shouldn't the AI instead aim to never load those troops in the first place? If the AI tries to load a transport, then leaves the units, it also seems that the AI played suboptimally and wasn't movement by having units shuffle back and forth.
> 
> Can you provide any insight about the complexity to try and update the loading logic? I'd suspect if the AI can't load and then unload in the same phase that the AI should instead consider and execute a different move.

I'm not sure what are all the reasons an AI would keep units loaded. One is not taking into account fuel and planning moves that it can't perform because it's out of fuel. To fix this, we need it to track how much fuel each move takes and consider those moves as not available if it's out of fuel.

I don't know if the above would be sufficient for the AI to instead choose to land units. For example, the AI loading units in the same turn is one case, but another is if units are already loaded at the start of a turn. (This could happen if previously there were no friendly territories to unload or if you made the AI is taking on a state that was produced by something else, like current AI, a human, a starting position on a map, etc).

I also don't know if other then fuel there would be reasons for units to stay loaded before. Basically, the AI has to maintain its own implementation of the game rules so it can know about things like fuel and so forth and whenever some new option is added, the AI might not understand that and might generate illegal moves that the engine rejects. So this code path is meant to do something a little better in these situations when it comes to transports. Even if the long term solution is to make the AI handle this in its main logic.

asvitkine:
> @asvitkine If you mean that those cargo units are going to be offloaded during a different phase than the phase during which they were loaded, this is obviously an illegal move: as far as I know, in every intended rules-set but V2 Basic (where no unit can), only air units can ever move on more than one phase during the same turn (so, in every intended rules-set, non-air units can never move on more than one phase during the same turn).
> 
> For the record, I'm against enforcing this limitation unless a property is made not to have it for custom games (so that you can load land units on ships during Combat Move and offload them on the same turn during Non Combat Move (which may be a legal move under the rules of my "270BC Wars" game)). But I guess it is not wanted the AI intentionally to make illegal moves which the engine is allowing.
> 
> Aside from this, if you keep offloading and re-loading units in a game with fuel costs, you will keep consuming (likely, wasting) such resources. So, a map with fuel costs should rather be the foremost example in which you would rather not offload your units (to avoid having to pay the fuel cost again later to re-load them) unless they have a high chance to be lost.

asvitkine:
> @asvitkine If you mean that those cargo units are going to be offloaded during a different phase than the phase during which they were loaded, this is obviously an illegal move: as far as I know, in every intended rules-set but V2 Basic (where no unit can), only air units can ever move on more than one phase during the same turn (so, in every intended rules-set, non-air units can never move on more than one phase during the same turn).
> 
> For the record, I'm against enforcing this limitation unless a property is made not to have it for custom games (so that you can load land units on ships during Combat Move and offload them on the same turn during Non Combat Move (which may be a legal move under the rules of my "270BC Wars" game)). But I guess it is not wanted the AI intentionally to make illegal moves which the engine is allowing.
> 
> Aside from this, if you keep offloading and re-loading units in a game with fuel costs, you will keep consuming (likely, wasting) such resources. So, a map with fuel costs should rather be the foremost example in which you would rather not offload your units (to avoid having to pay the fuel cost again later to re-load them) unless they have a high chance to be lost.

I agree that if loading and unloading uses fuel, this doesn't make sense. That's not the case for Iron War (but I should check with tanks), but if it can be the case, this logic should be made to account for it.

Otherwise, I think the rules enforcement belongs in the engine, not in the AI. So if we disallow unloading at a different phase than loading on some maps, the game should reject such moves and AI moves do get validated by the engine and so these moves would also not get performed.

Since this validation is not the case now, I think it's reasonable to let the AI make such moves - after all, humans can too and the AI is just playing by the rules of the engine.

asvitkine:
Looking more, it does seem that the AI will do this already if I artificially make it think transports have 0 movement. That means that it might not be so hard to make it track fuel costs for its moves and adjust the unit range based on the predicted available fuel at the time of the move.

Then, this change shouldn't be necessary if we just want to fix things for maps with fuel. I'll take a look at doing that.

Cernelius:
I'm not seeing how the AI should be coded to do illegal moves. TripleA also allows you to make non-combat movements during Combat Move, but I don't recall any AI has ever done that.

I'll defer to @panther2, but I think it is common knowledge that in every basic rules-set, during Non Combat Move, you can offload land units from sea units only if each of them was either loaded during the same phase or already on board at the start of the turn.

panther2:
> I'm not seeing how the AI should be coded to do illegal moves. TripleA also allows you to make non-combat movements during Combat Move, but I don't recall any AI has ever done that.
> 
> I'll defer to @panther2, but I think it is common knowledge that in every basic rules-set, during Non Combat Move, you can offload land units from sea units only if each of them was either loaded during the same phase or already on board at the start of the turn.

That is correct and I agree!

asvitkine:
I've updated the PR to only consider units that were either not loaded this turn or units that were loaded in non-combat.

(I'm separately exploring if this can be done by just making the AI be smarter and track fuel use during its move planning, but so far haven't had success on that front. My earlier experiment that just hardcoded transport range to 0 doesn't work for fuel tracking since transport range is considered before any moves are determined yet, so fuel might still be "available".)

Cernelius:
I believe the best would be to enforce the limitation (for any player) while also adding a property to allow split-moving non-air units during the same turn (so that games, like my 270BC Wars, in which **you should be able to load units on ships during Combat Move and offload them during Non Combat Move** (normally because you want to use warships both to make battles and to ship units to friendly territories on the same turn), can keep doing it as a supported feature, whereas in all basic games this would be impossible).

Then the AI would have just to play as per the program behaviour.

The property would have to be added manually to every game which is preferred to have the current (wrong) behaviour. For example, if, in World At War, you want to be able to load 1 infantry on a cruiser, move the cruiser to fight in a battle and offload the infantry into a friendly territory after winning the battle, you would have to add the property to the game file, since doing what I described is illegal under every basic rules-set.

---

I would be against enforcing this limitation generally without an opt-out property, because I think it is a dumb rule that custom made game should be allowed not to have (and basic games can have the property in the setting so that users can turn it on).

DanVanAtta:
I'm less concerned about the rule aspect of this than I am the AI making ineffective moves where it loads transports to then offload them again.

My other concern here is that blindly offloading might not always be better. If the units can be captured, it would be bad, or if the units are being landed somewhere where easily killed, also bad.

It's hard to balance these concerns, on the one hand keeping the AI from doing obviously bad moves is an improvement, on the other hand it's hard to judge because the move to begin with (and maybe end with) is just not good either.

DanVanAtta:
I'm bothered that the AI should not move units to a transport to begin with it'll just unload them again. @asvitkine can you comment directly to this concern?

asvitkine:
I have a WIP change that tries to track fuel used so that the AI would be
able to avoid doing moves it can't complete. However, right now it's not
clear that it's sufficient to fix this case.


I plan to still see what can be done there. In the meantime, we can keep
this open until we have a better idea if the other change can work out.

Cernelius:
Did anything change with fuel costing?

I remember fuel costs were charged on loading but not on offloading, so fuel should be irrelevant to determine whether or not something should be offloaded.

---

The matter is that you probably don't want to offload for no reasons because then you will have to pay the fuel cost again to re-load.

DanVanAtta:
@Cernelius the concern is the loading of transports for no reason. Offload makes sense if the unit should not be loaded at end of turn, but then why was the unit loaded to begin with.

@asvitkine let's please re-open this when it is ready, I'd like to try and get the PR queue to only be active items.

PR #9238, opened 2021-05-16
Update triplea_maps.yaml

RogerCooper:
Adds new mod, ww2v4_variants

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2290,3 +2290,11 @@
   description: |
     <br>
     <br>This is Global 40 Expansion. Adds House Rules from The Captain and Company.<br>
+- mapName: ww2v4_variants
+  mapCategory: EXPERIMENTAL
+  url: https://github.com/triplea-maps/ww2v4_variants/archive/master.zip
+  version: 1
+  img: https://raw.githubusercontent.com/triplea-maps/ww2v4_variants/master/preview.png
+  description: |
+    <br>Maintenance-WW2v4: Adds maintenance costs to World World 2 v4
+    


beelee1:
i went to your repo at triplea-maps and it was empty

https://github.com/triplea-maps/ww2v4_variants

Also make sure you have "master" as your default branch and not "main"

RogerCooper:
Try it now.

beelee1:
Yea it looks as if all the stuff is there but it should be in "map". Like this:

Screenshot from 2021-05-18 09-40-33

then when you open map it'll have all the stuff that you have:

Screenshot from 2021-05-18 09-41-29

So, I'd try putting everything in map and make the repo look like the first picture and then resubmit your yaml pr.

triplea_maps.yaml
```diff
@@ -2290,3 +2290,11 @@
   description: |
     <br>
     <br>This is Global 40 Expansion. Adds House Rules from The Captain and Company.<br>
+- mapName: ww2v4_variants
+  mapCategory: EXPERIMENTAL
+  url: https://github.com/triplea-maps/ww2v4_variants/archive/master.zip
+  version: 1
+  img: https://raw.githubusercontent.com/triplea-maps/ww2v4_variants/master/preview.png
```
RoiEXLab, line 2297: The image url is not quite right. It has to be with an uppercase P i.e. 'Preview.png' instead of 'preview.png', or instead you can change the filename in the repository itself

DanVanAtta: Changing the file name is preferred, 'preview.png' will be the standard name and will be required in 2.6

RoiEXLab:
The link still doesn't work, has the repository been renamed?

RogerCooper:
Yes, I ending up posting to ww2v4variants-b. I have not yet figured out how to fix everything so it will be the correct repository.

RogerCooper:
Everything is in the proper repository now. Should it be private or public?

DanVanAtta:
@RogerCooper public

DanVanAtta:
> Yes, I ending up posting to ww2v4variants-b. I have not yet figured out how to fix everything so it will be the correct repository.

Could you clarify? It's more difficult to change repositories after they are launched, it's well worth uploading correctly and getting the map clean before it is launched compared to after.

RogerCooper:
I have renamed the repository to triplea-maps/ww2v4_variants and made it public.

PR #9260, opened 2021-05-22
Migrate from java.io.File to java.nio.file.Path (Part 8)

RoiEXLab:
This is the final part of a series where java.io.File APIs get migrated to java.nio.file.Path

Part 1: #9039
Part 2: #9074
Part 3: #9112
Part 4: #9113
Part 5:  #9133
Part 6: #9216
Part 7: #9256

## Testing
None

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadMapsWindow.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadMapsWindow.java
@@ -395,15 +395,14 @@ public class DownloadMapsWindow extends JFrame {
                         .append(")"));
       }
     } else {
+      sb.append(doubleSpace).append(" (");
       try {
-        sb.append(doubleSpace)
-            .append(" (")
-            .append(newSizeLabel(Files.size(map.getInstallLocation().get())))
-            .append(")");
+        sb.append(newSizeLabel(Files.size(map.getInstallLocation().get())));
       } catch (final IOException e) {
         log.warn("Failed to read file size", e);
+        sb.append("N/A");
       }
-      sb.append("<br>").append(map.getInstallLocation().get().toAbsolutePath());
+      sb.append(")").append("<br>").append(map.getInstallLocation().get().toAbsolutePath());
     }
     sb.append("<br>");
     sb.append("</html>");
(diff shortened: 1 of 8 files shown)


RoiEXLab:
@DanVanAtta This PR also includes the post-merge fix for #9256

RoiEXLab:
@DanVanAtta In theory it would be possible to "migrate" more code, i.e. create the inputstreams ourselves and handle everything in that regard for APIs that don't support java.nio.file.Path but accept Input- and/or OutputStreams.
But as long as we don't really use different file systems like files inside zips I doubt it's worth the effort

DanVanAtta:
@RoiEXLab , I think removing the interfaces that accept URI (in order to deal with Zip file paths) and replace those with Path is potentially worthwhile. I suspect input stream was used in a number of places to avoid handling file paths with URI. Beyond that, probably not worth the effort IMO.

game-app/game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadMapsWindow.java
```diff
@@ -395,15 +395,14 @@ private static String newLabelText(final DownloadFileDescription map) {
...
     } else {
+      sb.append(doubleSpace).append(" (");
       try {
-        sb.append(doubleSpace)
-            .append(" (")
-            .append(newSizeLabel(Files.size(map.getInstallLocation().get())))
-            .append(")");
+        sb.append(newSizeLabel(Files.size(map.getInstallLocation().get())));
       } catch (final IOException e) {
         log.warn("Failed to read file size", e);
```
DanVanAtta, line 402: I was more referring for there to be an update to the logger mesage. The user will be presented with a pretty vague error message here, if downloading multiple maps then it won't be clear what happened or what they should do.

RoiEXLab: Ah my bad. I don't think this will happy very often. I didn't check but unless you change file permission while TripleA is running I don't think we will ever hit this condition. But maybe the best decision is to decrease the logger level so it's only logged but without message to the user?

DanVanAtta: It's tempting to ignore outright, hard to judge best course of action. I think when in doubt slap some additional information in and move on.

Issue #9231, opened 2021-05-15
Feature Request: to help players on the purchase screen with maxBuiltPerPlayer errors


TheDog-GH:
When a xml has maxBuiltPerPlayer it can frustrating for the player when they inadvertently try to purchase more than maxBuiltPerPlayer allows because it resets all their purchases.

One solution would be show how many can be bought, as in the image below;

![image](https://user-images.githubusercontent.com/71022517/118353811-ef862a80-b55f-11eb-9b4b-cbfe5e214593.png)

So a maximum of 2 Wizards and 3 Sorcerers can be can be purchased, the rest are unlimited, to the PU available.
If the red numbers were 0 then, 0 can be purchased.

The discussion is here
https://forums.triplea-game.org/topic/2752/maxbuiltperplayer-error-resets-purchase-ui-was-how-to-limit-the-number-of-units-purchased?_=1621065805932

butterw:
I have hardcoded some strings in ProductionPanel.java Rule.getPanelComponent() to see where best to display per player unitType counts.  IMO the best solution would be improve the max button so that it takes into account maxBuiltPerPlayer (allowedBuild = maxBuilt - currentlyBuilt).

![PerUnit_MaxBuilt_Mod](https://user-images.githubusercontent.com/408915/118719312-a1577c80-b828-11eb-8fe7-b2478b2224d9.png)

butterw:
I hacked the ProductionPanel to display currentlyBuilt (at the end of the tooltip) and added allowedBuild = maxBuilt - currentlyBuilt in the label. currentlyBuilt is not that simple to calculate.
    
![PerUnit_count_maxBuilt_Mod](https://user-images.githubusercontent.com/408915/118874206-eba13200-b8ea-11eb-92a9-3f0c0a70044f.png)

beelee1:
so count:4 is how many are already built ? Even that would be a Huge upgrade. I mentioned this a few years ago, because as we all know, it totally blows when you gotta purchase everything all over again.

butterw:
Yes and the max button is also working, which means the values should now never be reset. 

If maxBuiltPerPlayer isn't used, it would be possible to display the count in the label in place of allowedBuild.
The most difficult part is calculating currentlyBuilt , because the result isn't available directly.

I'm also going to display the max factory production capacity (if applicable).

Cernelius:
I've no idea how something like "Max Built Allowed" can fit in a tool-tip for users.

[...]

PR #9261, opened 2021-05-22
Display "Produce (bid)" in the Production dialog title for bid Production.

butterw:
Issue this aims to solve: when starting a game with a default bid option, the Produce dialog pops up which can be confusing.  
This aims to make things clearer by differentiating the dialog title for bid and regular production.

I have made further changes to the ProductionPanel as discussed in https://github.com/triplea-game/triplea/issues/9231#issuecomment-845466653, but they require more substantial changes. I would appreciate some advice on the preferred approach to move forward.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/ProductionPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/ProductionPanel.java
@@ -93,7 +93,8 @@ class ProductionPanel extends JPanel {
       final GameData data,
       final boolean bid,
       final IntegerMap<ProductionRule> initialPurchase) {
-    dialog = new JDialog(parent, "Produce", true);
+    final private String title = bid ? "Produce (bid)" : "Produce";
+    dialog = new JDialog(parent, title, true);
     dialog.getContentPane().add(this);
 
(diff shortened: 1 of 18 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/ui/ProductionPanel.java
```diff
@@ -93,7 +93,8 @@
       final GameData data,
       final boolean bid,
       final IntegerMap<ProductionRule> initialPurchase) {
-    dialog = new JDialog(parent, "Produce", true);
+    final private String title = bid ? "Produce (bid)" : "Produce";
```
DanVanAtta, line 96: the 'private' modifier is a syntax error, this variable is scoped to the constructor. You could simply omit the 'private' modifier, or (my preference) just inline the ternary expression and avoid the intermediate variable.

DanVanAtta:
Thank you for jumping in @butterw . There is a syntax error to clean up and please sign the CLA. Let us know via an issue or a forum post if you could use more help getting setup. I'll point out there is a './verify' script to do build verifications and would very much recommend you get set up in an IDE like Intellij.

PR #9273, opened 2021-05-23
battleWindow title

butterw:
Objective: Providing more detailed info in the battleWindow Dialog title.
The end-goal is to achieve the following output 
- "Germans attacks British in 110 Sea Zone (Battle Round 1 of 2)" 
the max number of battle rounds will be displayed if >1:   

Which would address the feature request:
https://forums.triplea-game.org/topic/209/gui-battle-interface-window-should-provide-battle-round-info-in-title-germans-attacks-british-in-110-sea-zone-round-1

This initial commit only partially addresses the issue. Expected output:
- Germans attacks Russians in Russia  (Air Raid)
- Germans attacks Russians in Russia  (Bombing Raid)
- Germans attacks British in 110 Sea Zone  (Air Battle)
- Germans attacks British in 110 Sea Zone
- Germans attacks Russians in Russia  

I'm hoping someone can suggest a way to get the current battleround info (it is available in IBattle, how to get it ?) to update the battleWindow title, in which case I'll update my commit. 

## Testing
I've not installed IDEA yet, and am compiling v2.5 code. So I'm relying on the automated tests for now.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/BattlePanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/BattlePanel.java
@@ -263,7 +263,8 @@ public final class BattlePanel extends ActionPanel {
                   defendingWaitingToDieCopy,
                   BattlePanel.this.getMap(),
                   battleType);
-          final String battleStr = battleType == BattleType.NORMAL ? "" : String.format("  (%s)", battleType.toDisplayText());
+          final String battleStr = (battleType == BattleType.NORMAL) ? 
+            "" : String.format("  (%s)", battleType.toDisplayText());
           battleWindow.setTitle(attacker.getName() + " attacks " + defender.getName()
                                 + " in " + location.getName() + battleStr);
           battleWindow.getContentPane().removeAll();
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/BattlePanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/BattlePanel.java
@@ -263,8 +263,10 @@ public final class BattlePanel extends ActionPanel {
                   defendingWaitingToDieCopy,
                   BattlePanel.this.getMap(),
                   battleType);
-          battleWindow.setTitle(
-              attacker.getName() + " attacks " + defender.getName() + " in " + location.getName());
+          String battleStr = battleType.toDisplayText();
+          battleStr = battleStr.equals("Battle") ? "" : String.format("  (%s)", battleStr);
+          battleWindow.setTitle(attacker.getName() + " attacks " + defender.getName() 
+                                + " in " + location.getName() + battleStr);
           battleWindow.getContentPane().removeAll();
           battleWindow.getContentPane().add(battleDisplay);
           final Frame parent = JOptionPane.getFrameForComponent(BattlePanel.this);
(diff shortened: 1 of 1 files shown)


butterw:
- The info needs to be displayed at the latest when the end of round retreat choice is presented (this is the purpose of displaying the battle round).
The battle is driven by Airbattle and MustFightBattle respectively (this is where the battle round gets incremented).
- Is it possible to set a display element from here (ex: battleWindow title) ?

```
triplea/delegate/battle/AirBattle.java
maxRounds = Properties.getAirBattleRounds(data.getProperties());
```

```
triplea/delegate/battle/MustFightBattle.java
maxRounds =
	battleSite.isWater()
		? Properties.getSeaBattleRounds(data.getProperties())
		: Properties.getLandBattleRounds(data.getProperties());
public BattleStatus getStatus()		
private void addRoundResetStep(
```

game-app/game-core/src/main/java/games/strategy/triplea/ui/BattlePanel.java
```diff
@@ -263,8 +263,10 @@ public void showBattle(
                   defendingWaitingToDieCopy,
                   BattlePanel.this.getMap(),
                   battleType);
-          battleWindow.setTitle(
-              attacker.getName() + " attacks " + defender.getName() + " in " + location.getName());
+          String battleStr = battleType.toDisplayText();
+          battleStr = battleStr.equals("Battle") ? "" : String.format("  (%s)", battleStr);
```
RoiEXLab, line 267:
```suggestion
          final String battleStr = battleType == BattleType.NORMAL ? "" : String.format("  (%s)", battleType.toDisplayText());
```

RoiEXLab (approved):
Just a small suggestion here

RoiEXLab:
@butterw You need to clarify a bit, I'm not entirely sure what you want to achieve. In theory a lot is possible using different approaches but most approaches usually don't make a lot of sense.
So yeah the title can be changed, but when should it be changed and why and to what?

RoiEXLab:
If you want to change the title whenever te round is updated then you might want to experiment with the `IDelegateBridge` object a bit.
Inside the "execute" method you could do something like
```java
bridge.getDisplayChannelBroadcaster().incrementRound();
```

and then add a new `incrementRound()` method to `IDisplay` and the implementations. Inside `TripleADisplay` you have access to the battle panel via `ui.getBattlePanel()`.
Note that this approach would be a network incompatible change.

Ideally you could have a look at one of the existing methods in https://github.com/triplea-game/triplea/blob/master/game-app/game-core/src/main/java/games/strategy/engine/display/IDisplay.java and see if one of those methods is called every time the round updates, or even more often. `notifyDice` or `goToBattleStep` sound like potential candidates. I don't know when exactly they are called but I suppose you can play around a little with them to see if something works.

Then you can just modify the related method in BattlePanel accordingly and hope for the best

game-app/game-core/src/main/java/games/strategy/triplea/ui/BattlePanel.java
butterw: I'm assuming there is a 100 char per line length limit ? 

I've installed Intellij but Checkstyle has a problem with suppressions.xml, any idea what the problem is ? 
https://github.com/triplea-game/triplea/issues/9279
 
When the round info is relevant, the goal would to update the battleWindow title each round before the dice are rolled. 
It seems this require non trivial code modification. 
I think it would be best to go ahead with the current PR, and revisit later when v2.6 is out.

RoiEXLab: Ah yes, forgot about that.
When using IntelliJ using the default keybindings you can press the Ctrl Key twice and then enter the command `gradle spotlessJavaApply --parallel` to do the formatting automatically for you

PR #9288, opened 2021-05-28
Fix refresh lag in purchase panel.

asvitkine:
Fix refresh lag in purchase panel. When clicking the up/down and min/max buttons in the purchase panel, the UI may lag behind, with different buttons updating their enabled/disabled state very slowly, sometimes taking seconds for everything to be ready. Calling invalidate() fixes the issue.

## Testing
On the map "Over The Top" (which has lots of units on purchase screen), click Max button on trenches and observe how fast everything updates. Before this change, it can be regularly laggy when trying a few times, whereas after this change it's consistently fast.

## Release Note

Fixed UI lag when interacting with the purchase panel.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/ui/ScrollableTextField.java
+++ b/game-app/game-core/src/main/java/games/strategy/ui/ScrollableTextField.java
@@ -169,6 +169,7 @@ public class ScrollableTextField extends JPanel {
       maxButton.setEnabled(false);
       minButton.setEnabled(false);
     }
+    invalidate();
   }
 
   public int getValue() {


RoiEXLab (approved):
Do you know why calling invalidate fixes the issue?
I tried to find an explanation of this method, but they didn't really explain why this works

asvitkine:
> Do you know why calling invalidate fixes the issue?
> I tried to find an explanation of this method, but they didn't really explain why this works

I'm not an expert, but what it looked like is the different elements weren't really redrawing. I thing invalidate tells Swing that a redraw is needed. (I don't know why changing the state of the elements wasn't triggering that directly.)

I also know it can be a problem if done off the UI thread, but I don't think that was the case here...

PR #9296, opened 2021-05-29
Zoom hotkeys: +(numpad) add, subtract

butterw:
**Adds CTRL with Numpad +/-  support for Zooming the map** 
Currently you can use CTRL with +/- to zoom the map. 
However, the - and +(=) and not the easiest keys to find on some non-us keyboard. 
Using the numpad (if available) is much more practical.

Longer term, it would be desirable to make at least some of the hotkeys user configurable (and bring back removed WASD keys). It would be desirable to first refactor the code to regroup all hotkeys definitions in a single file.  

## Testing
Tested OK with French keyboard.


butterw:
### Streamline Hotkeys

From a development perspective, putting all current shortcuts as constants in a single code file is simple enough to do and has many advantages:
- Developpers would know what all the current hotkeys are (+ their history) and could change them easily.
- Advanced Users could change hotkeys by recompiling with a custom shortcut file.
- The current hotkeys could be displayed automatically in a table, which would guarantee up-to-date info for users.
- Eventually all hotkeys may become user configurable in the GUI or a config file.

I've looked into it a bit, my suggestion is that all keyboard hotkeys should be defined as constants with a meaningfull name in a single file (ex: org.triplea.swing.key.hotkeys). 
Instead of using harcoded values, import the hotkeys file for all menu accelerators/keyboard listeners.  
KeyEvent.VK_RIGHT >> KeyCode.RIGHT >> HotKeys.ScrollRight1  (What would be the correct name and data type to use in Java ?)

ex in TripleAFrame.java: 
if (keyCode == KeyEvent.VK_RIGHT)
would become 	
if (keyCode == HotKeys.ScrollRight1)
In this case we can have multiple hotkeys for the same action (to be able to support WASD or equivalent)
- KeyEvent are integer, a value of -1 would mean the key is disabled

in most cases there would be a single hotkey for an action.
ex: newSaveMenu with a value KeyCode.S
- menu accelerators are used in combination with CTRL or ("meta" if Mac)
- the modification isn't required for menu mnemonics right now.

game-app/game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
```diff
@@ -742,6 +742,20 @@ private void addZoomKeyboardShortcuts() {
         () ->
             mapPanel.setScale(
                 mapPanel.getScale() - (ClientSetting.mapZoomFactor.getValueOrThrow() / 100f)));
+
+    SwingKeyBinding.addKeyBindingWithMetaAndCtrlMasks(
```
DanVanAtta, line 746: Rather than duplicating the above block, what do you think updating  "addKeyBindingWithMetaAndCtrlMasks" to accept a var-arg of key codes?

lib/swing-lib/src/main/java/org/triplea/swing/key/binding/KeyCode.java
```diff
@@ -19,6 +19,8 @@
   COMMA(KeyEvent.VK_COMMA),
   EQUALS(KeyEvent.VK_EQUALS),
   MINUS(KeyEvent.VK_MINUS),
+  ADD(KeyEvent.VK_ADD),
+  SUBTRACT(KeyEvent.VK_SUBTRACT),
```
DanVanAtta, line 23: SUBTRACT and MINUS are essentially synonyms. A javadoc on this entry to mention that this is the keypad only would help clarify that difference for future maintainers. Would you mind adding such a note? Similar for 'ADD'.

DanVanAtta:
Please raise the developer focused conversation presented in: https://github.com/triplea-game/triplea/pull/9296#issuecomment-850992563, as an issue (or as a github discussion). PR chatter is ephemeral and should focus on the changes specifically presented in the PR.

butterw:
if I understand correctly, you are suggesting to add a addKeyBindingWithMetaAndCtrlMasks method that accepts two KeyEvents as input. 
Why does this method add KeyBinding with Meta AND with Ctrl ? I would assume the user is either using a Mac OR Win/Linux and only one binding is necessary ?
ex: KeyStroke.getKeyStroke(KeyEvent.VK_A, Toolkit.getDefaultToolkit().getMenuShortcutKeyMaskEx())
It should also be noted that keyPressed KeyListeners are used in the same TripleAFrame for the arrow hotkeys.

The alternative would be to keep addKeyBindingWithMetaAndCtrlMasks as it is, and create a single zoom-in/zoom-out method. 
The corresponding mouse listener is in ImageScrollerLargeView.

DanVanAtta:
Suggestion is to change:
```
  public static void addKeyBindingWithMetaAndCtrlMasks(
      final JFrame component, final KeyCode key, final Runnable action) {
```
To:
```
  public static void addKeyBindingWithMetaAndCtrlMasks(
      final JFrame component, final Runnable action, final KeyCode key, final KeyCode ... additionalKeys,) {
```
To keep things consistent, a similar change to `addKeyMapping` would be needed.

> Why does this method add KeyBinding with Meta AND with Ctrl ? I would assume the user is either using a Mac OR Win/Linux and only one binding is necessary ?

I can't recall, not sure if I ever knew the context behind this. Yes, one key binding is in theory needed. I would not change it unless you can really dig into the background around this.

> The alternative would be to keep addKeyBindingWithMetaAndCtrlMasks as it is, and create a single zoom-in/zoom-out method.

If there is a need for it, can you clarify why this alternative would be better? IMO the updating of the `addKeyBindingWithMetaAndCtrlMasks` API to accept var-arg is pretty straight forward and could have future uses.

butterw:
> > Why does this method add KeyBinding with Meta AND with Ctrl ? I would assume the user is either using a Mac OR Win/Linux and only one binding is necessary ?
> 
> I can't recall, not sure if I ever knew the context behind this. Yes, one key binding is in theory needed. I would not change it unless you can really dig into the background around this.

Maybe some people using mac wanted to use the same hotkeys as on PC ? 
 
> Suggestion is to change:
> 
> ```
>   public static void addKeyBindingWithMetaAndCtrlMasks(
>       final JFrame component, final KeyCode key, final Runnable action) {
> ```
> 
> To:
> 
> ```
>   public static void addKeyBindingWithMetaAndCtrlMasks(
>       final JFrame component, final Runnable action, final KeyCode key, final KeyCode ... additionalKeys) {
> ```
> 
> To keep things consistent, a similar change to `addKeyMapping` would be needed.
> 
> > The alternative would be to keep addKeyBindingWithMetaAndCtrlMasks as it is, and create a single zoom-in/zoom-out method.
> 
> If there is a need for it, can you clarify why this alternative would be better? IMO the updating of the `addKeyBindingWithMetaAndCtrlMasks` API to accept var-arg is pretty straight forward and could have future uses.

The var-arg approach is likely the right solution, but I think a full refactor of hotkeys is also needed. 
AddKeyBinding is called by a number of classes, and could possibly be generalized for all accelerators/hotkeys.
I've opened a discussion on the subject: https://github.com/triplea-game/triplea/discussions/9315

DanVanAtta:
Consolidating the hotkeys seems like a good move. I'm supportive of work in this direction.
Changing the method API to be a var-arg is quite incremental and avoids having to copy/paste, it's the cleaner way to implement this if we are just wanting to have multiple keys with the same binding. The API in question is only invoked in a half dozen places, it's not that much.

DanVanAtta:
@butterw are you still working on this?

butterw:
I've started early experimentation with a much more in depth hotkey overhaul, but my target for this is v3.0 not v2.6. 
My initial PR could work as is for v2.6.

DanVanAtta:
Okay, please re-open when this is ready for further review.

PR #9324, opened 2021-06-01
Add title to file dialogs

butterw:
adds missing titles to the awt.FileDialogs used for "Open Saved Game" and "Save Games".


butterw:
On Windows the themed swing.JFileChooser, used by ExportMenu is quite poor. The native awt.FileDialog used for Open/Save Game is much nicer. 

I don't know how difficult it is to convert between the two classes, if it is easy, I would suggest using FileDialog instead of JFileChooser in ExportMenu.

DanVanAtta:
One needs to consider the different behavior of the native file chooser on the three main OS types. I recall we used to have an if statement that made using native file chooser OS dependent and I think we removed that because the native was better across the board. I don't rightly recall the full details there, you might have some luck searching the git history for changes between the native file chooser and awt file chooser. If there is nothing 'hidden' and you can get that context, then it would be a good update to make.

PR #9330, opened 2021-06-03
MoveHelpMenu fix

butterw:
a couple of fixes for the new MoveHelpMenu.

Question for devs: 
after running gradlew spotlessApply 
it changed my file to Windows Line Endings which was showing as all diffs. 
Is there are gradlew task to obtain a clean result via commandline ?


DanVanAtta:
Do you have a global .gitattributes set somewhere? Are you actually on windows @butterw ? 
This is probably a better thing to discuss in an issue or github discussion so that conversation can stay to the one topic (PRs discussions are best sticking as closely as they can to just the updated code and related implications)

DanVanAtta (approved):

PR #9338, opened 2021-06-05
Update triplea_maps.yaml

RogerCooper:
Adds ww2v4_variants


DanVanAtta:
Is this a duplicate  @RogerCooper ? Looks like this was already added in another somewhat recent PR.

DanVanAtta:
Please re-open if there is more to follow up with.

RogerCooper:
WW2v4 Variants is now the only mod showing up as installed on my computer (everything else needs an update), so I guess everything is good on it.

PR #9339, opened 2021-06-06
Zoom hotkey display (0. change title of map Zoom dialog)

butterw:
Display zoom info on screen when a zoom hotkey is pressed or zoom is changed via View Menu > Zoom.
See https://forums.triplea-game.org/post/50364

- Zoom only needs to be displayed when a zoom hotkey is used.
- Zoom shall be displayed rounded ex: "38% Map zoom". 
- Zoom should be displayed on screen in 3 cases: 
-1. CTRL+mouse wheel zoom hotkey (mouse wheel listener defined as a lambda in ImageScrollerLargeView, superclass of mapPanel)
-2. keyboard zoom hotkey (defined in TripleAFrame) 
-3. View Menu zoom applied (uses an Integer value, defined in ViewMenu).
- The simplest way to display zoom is in the status JLabel of the TripleAFrame frame. It will stay displayed until blanked or overwritten.

To finalize this PR, I would need some help on how to get zoom to display in only these cases, and where to place the displayMapZoom method (place it in TripleAFrame ?).
**mapPanel.setScale is also called when the window is resized, we don't want to display zoom in this case. How to do this ?**
- Map zoom is set with mapPanel.setScale(scale) where scale is a double in [minZoom, 1].
- the frame is accessible from the mapPanel with SwingUtilities.getRoot(this)

int zoom = Math.round(mapPanel.getScale()*100);
- public void displayMapZoom(int zoom) { 
	status.setText( zoom + "% Map zoom"); 
  }

- Setting zoom in ViewMenu sets it to an integer, which can be slightly different from the current value because of rounding.
- if map scale is 0.596, the rounded value displayed is 60% Zoom.
- However applying 60% zoom in ViewMenu will cause the map to resize to 0.60 scale.


DanVanAtta:
Two concerns with status bar location:
(1) status bar on maps with lots of resources and territory effects may not have enough space to render anything further. Should be verified on civil war map, IIRC it has both.
(2) status bar for transient information is somewhat questionable. Notably that the information is over-written. If zoom percentage were in status bar, I'd expect it to be there always so one can reliably check the zoom level. We then have to consider whether having zoom percentage is important enough to be in the status bar.

In the forum thread I suggested something like the edit mode text. To make that workable the text should be made smaller and put perhaps in a corner of the map. Generally the treatment would be similar, an on-map somewhat opaque, transient text. WDYT @butterw ? 

It might be just as well to land this PR here to update text so that other changes can stay isolated and be more minimal. The more focused PRs are, the better, something to consider.

butterw:
I haven't played the Civil War map, but it looks like it uses 6 ressources and 2 production units: 
ACW

A temporary zoom display in the status bar works fine here.

Status bar display, which is only rarely used in tripleA, is transient by nature.  Permanent displays have a dedicated box in the gameSouthPanel. 
Given that the associated map scale dialog doesn't even have an associated hotkey, map zoom likely isn't critical enough to require a permanent display on screen. Many users probably never change zoom for a given map, but the lack of a visual indicator for zoom would also be a reason for that.

Using a transient (OSD) overlay instead of the status bar could work, the implementation is the same, only the display function in TripleAFrame changes. I'm not sure how practical this will be with Swing though. Graphics2D text antialising isn't currently used for the "EDIT MODE" display.

I'll now change this minor PR to ready. 

I'll go ahead with the proposed changes in a new PR (they will only targets temporary status bar display). This offers a clear improvement to the map zoom feature and should only require minor modifications to the existing code.

As a sidenote, I think map scale preferences are currently getting overwritten when the window size changes, which seems undesirable and the PR would attempt to fix that.

DanVanAtta:
Please try to focus any PR to one change if you can. Status bar changes are going to get a lot of pushback, we do not want players to get used to one thing to then change it later. You also need to consider the territory effects that appear during hover and very long territory names. To boot, a screen might not be quite at max width so there should be some tolerance there if the screen is only 75% of a computers width.

butterw:
I checked out Ice and snow in v2.5, which has full territory effects.  It is constrained by the layout of the gameSouthPanel, rather than the content of the status label. Probably this layout can be improved. The status label currently acts as a kind of spacer. 
Maybe the status text should be centered, rather than left aligned ?

The following screenshot is with 100 dpi (the territory effect icons are nice but the text is too small), 
@125% dpi there is some cut-off (but this map can't handle scaling anyway)  
https://raw.githubusercontent.com/butterw/ui/main/assets/TerritoryEffects_Substance_dpi100pct.png

DanVanAtta:
Is it intended that this PR is to only update the word "scale" to "zoom"? If so, does the value in the dialog box agree with that verbage, ie: it is a percentage? Is there anything further that is intended for this PR @butterw ?

butterw:
scale is a double in the code. 
zoom should designate the user integer Map Scale Percentage *value*. 

I've changed the dialog text to "Choose Map Zoom (%)" which is clearer.

No further changes are required with ViewMenu.

DanVanAtta:
@butterw changing the text is more consistent with a number of other elements.

Please keep in mind ways to minimize review effort. To this extent, I'm a bit confused by the overview commentary and while the detail is interesting and related to requirements. I would prefer for a PR overview to succinctly summarize the changes presented and any useful context to better understand the implications and motivations for the update. Ideally a PR overview is almost completely auto-filled using commit commentary, which reduces the work for everyone.

PR #9346, opened 2021-06-09
Zoom hotkey display 1

butterw:
Displays Zoom in status label of TripleAFrame when zoom hotkey or ViewMenu zoom are used.  
See https://github.com/triplea-game/triplea/pull/9339 for preliminary discussion.

1) Display format "40%"
- "40% zoom" or "40% map zoom" are also possible but longer

2) Impact on the layout of GameSouthPanel
- This modification has no impact on the layout. Status bar currently uses the same space whether it is empty or used. 

3) Horizontal Alignment
- status currently uses the default LEFT alignment. CENTER alignment could be a bit distracting. RIGHT alignment would be possible.

4) the map zoom display stays in status Label until blanked or overwritten
Status gets blanked at phase change. An overwrite occurs when an invalid Move is being considered.
Alternatives:
- a 2 second timer to blank it could be considered. 
- permanent display would also be possible with the compact "40%" format, to do this an additionnal mapZoom Label would be added to stepPanel instead. 

5) A setScaleResize method was added to mapPanel, to avoid displaying zoom when the window is resized (the map zoom doesn't change when this occurs).
From my testing the line uiContext.setScale(scale) of setScaleResize should likely be removed. 
It causes the map scale Preference in UiContext to be overwritten a number of times. No adverse effect was observed from removing it. I've left it only because I don't know why it was put there in the first place.


DanVanAtta:
I do like the user feedback provided from displaying the zoom percentage somewhere.

I have a few concerns though:
1. Seemingly the zoom percentage should be rendered initially on game launch. It seems odd to render only after being changed and then stay persistent. 
2. The percentage value is displayed without context or label and no hover text. It can be a bit odd to have a this 100% floating around on the UI and having to remember this is map zoom and not something else.
3. The information in the status bar becomes less coheisve. Between listing a territory and the current faction being moved, we have map zoom percentage. Looking at the elements in the status bar, a few are not alike, but one is particularly not like the others.

Implementation-wise, this is pretty brittle:
```
final TripleAFrame frame = (TripleAFrame) SwingUtilities.getRoot(this); 
```
If we change the root that is returned by SwingUtilities this would break. This could perhaps be fixed, but we should resolve the UX concerns if possible before attempting to fix the code.

butterw:
> I do like the user feedback provided from displaying the zoom percentage somewhere.
> 
> I have a few concerns though:

The Map zoom could be displayed initially: "80% Map Zoom" in status with default left text alignment.
When the map scale is changed, the display would change to 90% right aligned.

From testing, default tooltips don't work great for elements of the bottom panel. 

The status Label is currently underused. I intend to make more use of it (ex: display relevant info on hotkeys such as the hotkey to exit fullscreen mode).

PR #9368, opened 2021-06-19
Convert the BattlePhaseList into BattlePhaseStep

trevan:
Sorry for taking a long time but here's the next part of the BattlePhase change.  This takes the phases that are created by UnitAbilityFactory (and maybe the XML at a later point) and creates BattlePhaseSteps.

The next part after this will be to flesh out the BattlePhaseSteps so that they actually do stuff.  Then, I'll change the code to use the BattlePhases instead of the existing logic.

## Testing

This code is currently not used in the game yet.  So manual testing isn't possible.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/battle/phase/BattlePhaseList.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/battle/phase/BattlePhaseList.java
@@ -89,36 +89,28 @@ public class BattlePhaseList {
 
     final Collection<ConvertUnitAbility> convertAbilities =
         units.stream()
-            .map(unit -> getConvertUnitAbilities(unit, ConvertUnitAbility.Team.FRIENDLY))
+            .map(unit -> filterConvertUnitAbilities(unit, ConvertUnitAbility.Team.FRIENDLY))
             .flatMap(Collection::stream)
             .collect(Collectors.toList());
 
     convertAbilities.addAll(
         oppositeUnits.stream()
-            .map(unit -> getConvertUnitAbilities(unit, ConvertUnitAbility.Team.FOE))
+            .map(unit -> filterConvertUnitAbilities(unit, ConvertUnitAbility.Team.FOE))
             .flatMap(Collection::stream)
             .collect(Collectors.toList()));
-
-    final Map<CombatUnitAbility, CombatUnitAbility> fromToAbilities = new HashMap<>();
-    final Map<CombatUnitAbility, CombatUnitAbility> toFromAbilities = new HashMap<>();
-    for (final ConvertUnitAbility convertAbility : convertAbilities) {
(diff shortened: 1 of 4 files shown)


trevan:
The unit test failure is in the maps-module which I didn't touch.  So I'm not sure why it is failing.

DanVanAtta:
@trevan 

re: build failures - there was a flaky test, it has been improved by extending a timeout value. Should be more reliable now if you merge latest master. I re-ran the test on this PR and it passed. There is a codacy update that would be good to get fixed as well.

How much of this PR is new code vs moved code?

The placeholder parts where there is a no-op implementation are not obvious that they will be updated in the future. Having a project reference, placeholder comments, something seemingly would be helpful there.

How many iterations would you expect it to require to complete this series of updates?

DanVanAtta:
Another question @trevan , how far along are you on this? Do you have other branches ready to land, or is this the first step and represents your latest progress?

trevan:
@DanVanAtta 

I expect 2-3 more PRs for this.  The next one will be finishing the BattlePhaseStep, the one after that would be to actually use it in battles.  I think it might be worthwhile to put this behind the beta flag in case I missed some edge case in battles so that would be the middle PR.

I don't have other branches ready to land.  This is what I've done so far.  If you want me to, I could wait until I've finished the BattlePhaseStep but I thought a smaller and sooner PR would be better.

This is all new code.  It is not copied from anywhere.

I rebased off of master and added some placeholder comments to the BattlePhaseStep.

trevan:
The test failures look like they are in the game hosting code.  They aren't related to my changes.

DanVanAtta:
@trevan , are you aware of the 3.0 initiative? Can you suggest how valuable this work would be going forward if we were do execute on that initiative and have you considered instead to focus on 3.0? In other words, does this remain the right direction given the potential 3.0 re-structure of most of the code?

trevan:
@DanVanAtta, my understanding of the 3.0 initiative is based on just https://forums.triplea-game.org/topic/2794/triplea-3-0-design-proposal-discussion?page=1.  Based on what it says there, I see this change as still remaining in the right direction.  I assume that the restructuring that I'm doing here will fit in with the restructuring of 3.0.

If there is other places that describe the 3.0 initiative, please let me know.

DanVanAtta:
We need to spend some quality time designing 3.0

I expect we'll change most of the underlying data models and change the orchestration code layer significantly. The implementations are going to be interesting, for battles there could be more rewrite than otherwise.

IMO the priorities right now are to finish 2.6, test it, and then focus on designing 3.0 and try to land that as quickly as feasible.

DanVanAtta:
@trevan I appreciate trying to keep PR diff smaller and more manageable. I'll leave it to your judgement if you want to move forward here or if it would be better to focus on redesign and how this would in 3.0.

I think it would be useful to have at least a first feature land in this so it is not just completely stubbed out.

PR #9392, opened 2021-06-22
Map font size default

butterw:
The goal of the PR is to make the default unit count font size larger. This helps with keeping the map readable on large maps without needing to zoom-in. https://forums.triplea-game.org/post/50518.
Setting a font size different from the rest of the UI, gives new users a hint that this is a preference setting (configurable in the View menu).

It doesn't have any effect if the fontsize has already been set (existing installations) so shouldn't be disruptive.


Cernelius:
Current is 12, right? So, what's the new one?

I've it on 24 (since a long time).

butterw:
the PR is to change to 24.

DanVanAtta:
This needs extensive testing, please describe what testing you have done for this @butterw

butterw:
I've only done limited testing on Windows with 2 different sizes of 1080p monitors. 
I've reset to defaults with vanilla v2.6 and the results are particularly bad, notably for the map unit count font with my system setting (125% dpi):
.
ww2_v6 isn't even a large map and without adjustment the unit counts are displayed way to small  (which means you would be forced to increase map zoom and continually scroll).

This minimal PR will only affects new users and those who click "reset to default" (losing their existing settings in the process).

DanVanAtta:
- extensive testing is part and parcel to TripleA development. Asking what testing you have done is also so we can hypothesize about further cases that perhaps were not considered. At the very least a reviewer can consider whether the testing has been sufficient to be considered complete and thereby reduce risk of support, rollbacks, and release blockers.

- re: testing: have you verified what 24 by default looks like on very crowded territories that have lots of units in them?

- shouldn't this apply to existing users as well? It seems like it might be a bit of an easter egg otherwise.

- this update is seemingly worth a release note, most any type of UI change is notable and deserves inclusion.

butterw:
>     * extensive testing is part and parcel to TripleA development. Asking what testing you have done is also so we can hypothesize about further cases that perhaps were not considered. At the very least a reviewer can consider whether the testing has been sufficient to be considered complete and thereby reduce risk of support, rollbacks, and release blockers.

Current map font default size (12) doesn't work and this is a minimal fix. 

This setting is used for 2 things:
1) Map Unit Count Font >> my preferred size is 24.
2) Map Font-size, only used by a few maps like POS2 which use a font for Territory/Zones Names, Convoy Centers. A size of 18 would be enough here.

So the safest map font size value would be 18, but it would be better to split this into 2 preference values. 

As it potentially depends on map, Screen dpi, OS, L&F, and preferences, this is inherently hard to test. In my view, it should be put into a pre-release as early as possible, then tweaked or reverted if required.
The change isn't game-breaking: if the default map text size isn't quite right for a map, the user can adjust this setting in the View menu dialog.
 
>     * re: testing: have you verified what 24 by default looks like on very crowded territories that have lots of units in them?

Sure (ex: ww2 global). A value higher than 24 displays fine.

>     * shouldn't this apply to existing users as well? It seems like it might be a bit of an easter egg otherwise.
>     * this update is seemingly worth a release note, most any type of UI change is notable and deserves inclusion.

The change should be mentionned in release notes with the new default map font size specified.

Cernelius:
Sadly, there is a problem with non-image territory (actually, zones) names font size.

When (like in "pos2") coordinates for them are defined, they go increasingly out of place as the font size diverges from the one (likely 12) for which they have been made.

This is due to the fact that the coordinates define a corner (not the centre) of the name.

For example, if you open "Pact of Steel 2", at font size 12 you can see "West Balkans", "French West Africa", "French Equatorial Africa" and so on are well centred, whereas at font size 24 those names are missplaced so much to the right that they end up drawn in between of two territories.

The case is made even worse because of wrapping disappearence issues. You can see that, at font size 24, the "Western United States" name is cut to "Western Unite", and (because of the rightwise drift) it seems centred on "Central United States" territory, whose name is cut to "Central" only.

By the way, TripleA here is also inconsistent because it allows you to alter the font size of territory (zone) names if they are not customized but not if they are. For example, changing the font size changes the dimensions of the territory names in "Pact of Steel 2" and "World At War" but not in "270BC".

I don't think it has been a good ide to have a single font size acting on both stack numbers and territory names, and I would say it is even questionable the opportunity of being able to change the font size of territory names at all (especially in the moment you are not able to if the map features images for showing territory names, leaving the user with some maps in which the territory names size can be changed and some other maps in which it cannot).

butterw:
@Cernelius 
In Pact Of Steel "Spain" doesn't scale like the other T/zones names.  
Some text cut-off is present also at size 12.

It may not be perfect but the T/zones display at 18 is still preferable vs the current default display at 12 (unplayable on my screen @100% and @125% dpi). Likely because map text is painted, it doesn't scale with HiDpi.

Good map graphics are probably only possible with images for T/zone names and Pu values. The default map Font size needs to ensure maps are playable.

Cernelius:
> @Cernelius
> In Pact Of Steel "Spain" doesn't scale like the other T/zones names.

As I pointed out,
>TripleA here is also inconsistent because it allows you to alter the font size of territory (zone) names if they are not customized but not if they are.

"Spain" is the only customized territory name of that map, so it is the only one which is completely unaffected by the font size.

So, here you go: now you have found another problem. In cases like this, if you increase the default, you will end up with some names that are the same as before and some other names that are bigger that they were so likely bigger than the unaffected ones.

Cernelius:
Generally speaking, I agree with increasing the default font size to 24, but I believe the font size should be primarily defined at the map level (every mapmaker being able to set the font size for the map).

---

I don't like the fact that the only way of customizing territory names is turning them into images. I would like TripleA supporting customization at the font level (meaning that, instead of customizing the name "Spain", you customize the single letters, for every name having the letter (meaning that, instead of providing an image for "Spain", you provide an image for "S", an image for "p" and so on)).

As it is, however, one may argue that, if you increase from 12 to 18 the font size of non-customized territory names, you should also enlarge by 50% the territory name images for every map having them. Obviously, this would be a very bad idea, because it would alterate the intended view for the map, yet it would address the mentioned problem about "Spain".

butterw:
I've tried the World at War map (at 100% dpi), which also uses text for territories. This map has extreme overcrowding issues in some territories. On my system 18 could work, 24 would not.

butterw:
> > @Cernelius
> > In Pact Of Steel "Spain" doesn't scale like the other T/zones names.
> 
> As I pointed out,
> 
> > TripleA here is also inconsistent because it allows you to alter the font size of territory (zone) names if they are not customized but not if they are.
> 
> "Spain" is the only customized territory name of that map, so it is the only one which is completely unaffected by the font size.

What causes the territory name to be customized in a map ? This scaling issue looks more like a bug than a feature.

DanVanAtta:
Loading this update, this is what I see:
Screenshot from 2021-11-06 16-42-48

This could seem like a major problem for most players if they see this and are unsure of how to 'fix' it. Can we reasonably improve this experience? Perhaps a default that is dependent on DPI?

DanVanAtta:
@butterw , it's been a while since you submitted this issue. If this is good to pick back up, please re-open

PR #9491, opened 2021-08-11
Map Tags & Use Maps Server (triplea_maps.yml now unused)

DanVanAtta:
__Always use Maps Server__
  Stop reading 'triplea_maps.yml'
  Specifically instead of downloading the 'triplea_maps.yml'
  file and parsing it, the front end will instead contact
  the server and receive a JSON payload that contains the
  map download list.

__Introduce and use map tags__
   Map tags are a generic
  way to add additional meta data to maps dynamically.
  For example, instead of having a 'category' value
  specifically enumerated, we instead now have a
  tag whose key is 'category'. This allows us to have
  an arbitrary set of meta information without
  needing to update the front end.

__Update server to send 'preview image url' to the frontend__
  The server computes the preview image URL to make it easier
  and more explicit for the frontend. Having the server
  instead of the client compute the preview image URL also
  allows for it to be updated in database.

## Testing
Tested with a local server
- started database `./spitfire-server/database/start-docker-db`
- started spitfire (lobby) service via IDE
- started a headed client
- updated 'settings' to use a 'local' lobby server
- clicked 'download maps'

## Screens Shots

The local server is using dummy data, otherwise note that a tag with 'rating' was configured
and is now displayed as a column on the download maps page:
Screenshot from 2021-08-10 22-00-24

## Additional Notes / Warnings 

This is going to break the front end for prereleases, specifically download maps might not work.
The deployment of the server needs to happen and even then there is likely more to work out.

## Future & Following Work

- Fix up deployment to server so we get latest code deployed to the  prerelease environment. This will fix up 'download maps' not getting data
- Fix up any download or data quality issues, notably make sure server is doing a good job indexing maps
- Wire in 'map size' to the download payload
- Build an admin toolbox UI to update tag values

## Release Note

UPDATE|Maps information is now stored in database and no longer stored in the file: 'triplea_maps.yaml'. Game clients now download map information from the server. The game client no longer downloads 'triplea_maps.yaml'. 'Map tag' meta data capability added to database, map admins can add tags to maps and those tags are now displayed by the game client in the download maps screen.


beelee1:
Is this for the prerelease or effective now ?

DanVanAtta:
@beelee1 
prerelease clients will become broken for map downloads until we have updated the prerelease server to also have the latest code (that is typically automatically deployed after PR merges, but is currently broken so prerelease server will be incompatible). Once that is done, there will be another step for any additional fixes to automatic map indexing and anything else. But otherwise the server will automatically find any map repository in 'triplea-maps' and store in database without needing a 'triplea_maps.yaml' file at all.

beelee1:
Right on. So does this effect the stable then ?

DanVanAtta:
nope

PR #9500, opened 2021-08-13
Read IP address from X-Forwarded-For

DanVanAtta:
When dropwizard is behind a NGINX (reverse proxy) server, The IP address
return by 'request.getRemoteAddr()' is the IP of the NGINX server and
not the client IP.

To get the correct IP we must read the 'X-Forwarded-For'. This update
makes that change and adds a configuration flag to read 'getRemoteAddr()'
in the development case (where there is no NGINX server).

Note, websocket code that depends on IP address has this same problem
but worse as it has no access to the 'X-Forwarded-For' header. A
future update will need to fix this.

## Testing
Deployed to prerelease manually and added some extra debugging to logs to verify the correct IP address was being used.
Then built a local client and connected to lobby (currently that automatically connects to the 2.6 prerelease server).

### DIFF
--- a/infrastructure/ansible/roles/lobby_server/templates/lobby_server.service.j2
+++ b/infrastructure/ansible/roles/lobby_server/templates/lobby_server.service.j2
@@ -6,6 +6,7 @@ Documentation=
 Environment=ERROR_REPORT_TO_GITHUB_ENABLED=true
 Environment=MAP_INDEXING_ENABLED=true
 Environment=GAME_HOST_CONNECTIVITY_CHECK_ENABLED=true
+Environment=USE_FORWARDED_HEADERS=true
 Environment=DATABASE_USER={{ lobby_server_db_user }}
 Environment=DATABASE_PASSWORD={{ lobby_db_password }}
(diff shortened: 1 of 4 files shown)


DanVanAtta:
I think this is a problem new to DropWizard 2. There is a config to by default true to use forwarded headers. That was disabled in 2.0 (https://github.com/dropwizard/dropwizard/issues/2663)

I was trying to correct this by adding a 'server: [{type: 'http', port: 8080, useForwardedHeaders: true}], though the prerelease server (and not building from local) crashed due to unrecognized http header type and reported there were no known http connector factories (which smells of a dependency issue). 

If we can flip that flag, then this update would be much simpler and we'll have a good chance for the websocket case to also be fixed. We should investigate that (configuration update) further.

DanVanAtta:
Spent some more time on this and found a key config when building 'shadowJar' to merge service files into the build. This fixed being able to use the 'server' tag in configuration.yml and allowed for the 'useForwardedHeaders' config to be enabled. With that, we also needed the proxy config to forward headers from nginx for websocket connections. With those two updates we now are reading the correct IP address when using 'getRemoteAddr()'

PR #9503, opened 2021-08-13
Doc update

emmanuel-perlow:
An minor expansion of the development/tutorial/README.md and docs/development/how-to/local-build-and-testing.md.

Add some notes about how to invoke `./gradlew` and what happens on the
first invocation.

Additionally, add a section about how to run a test.

**Note:** Be aware that I've never used gradle before and the things written down are
found by Internet serach and trial and error.  There might be more quirks which
I'm not aware of.


docs/development/tutorial/README.md
```diff
@@ -3,6 +3,7 @@
 ## Assumptions
 
 - Some familiarity with 'git' is assumed
+- A working Java environment (a JRE or JDK)
```
DanVanAtta, line 6: 'JRE' option perhaps should be removed, a JDK would be needed to compile. A person reading this doc would be interested in checking the code out, building it and running it.

docs/development/tutorial/README.md
```diff
@@ -23,8 +24,14 @@
 
 ## Running the code
 
-- IDE (there are launchers checked in for convenience)
-- CLI: `./gradlew :game-app:game-headed:run`
+To build and run the TripleA program (the one you play the game with) either
+
+- use IDE target `game-headed` (there are launchers checked in for convenience)
```
DanVanAtta, line 29: Is this more specifically the 'gradle target'? If so, the parenthetical comment should be a new bullet point as that is an alternative way to run the code. Using the checked in launched is the preferred way to launch the app (contains all of the parameters and config needed, if we use it then we'll be likely to update it and check it in for everyone else, if anything changes then checking in a config cascades those updates to all)

docs/development/tutorial/README.md
```diff
@@ -23,8 +24,14 @@
...
+- execute the command: `./gradlew :game-app:game-headed:run`  (if you only want to build (but not run) the source use `./gradlew :game-app:game-headed:build` instead)
+
+*Note:* The first time you (or the IDE) run a gradle command, gradle will automatically download all required dependencies and assets.  This step needs a working internet connection.
+
+*Note:* To see a non-reducted version of the output of gradle supply the `-i`/`--info` option to the command.
```
DanVanAtta, line 34: I'm curious what the situation was where you needed the `-i` flag, when did you it come in handy?

docs/development/tutorial/README.md
```diff
@@ -54,3 +61,32 @@ It is also often useful to run code formatting:
 ```
 ./gradew spotlessApply
 ```
+
+
+## Running some tests
```
DanVanAtta, line 66: Looking at this document, it seems that:
(1) it is a how-to, not a tutorial. How-to assumes a person knows the tools, but just not specifically for TripleA. A tutorial strives to teach, eg: let me teach you how to build java code using gradle, vs 'this is how you use gradle with triplea to do a build, step 1, step 2"

Ergo, this document is misplaced to begin with. More importantly, another implication is that the content should all be very terse and additional examples should only be footnotes. A 'how-to' should read very similar to a cooking recipe and give a concise and single way on how to do something. Hence, identifying this document as a how-to helps clarify what should be presented.

Though last, I think it turns out this doc is a duplicate of:  `how-to/local-build-and-testing.md` .   :facepalm: 
Note how both have similar sections and are both the 'if you are dev, here is how you build, test and run triplea'.

To compensate for this latter point, and if you agree that these two docs are basically the same, I'd suggest we remove the changes from this file and merge them in with `how-to/local-build-and-testing.md` . I believe that would have the benefit of removing some redundancy.

As follow-up, we could then merge in the remaining content from this ('tutorial/README') page into the 'how-to' page and remove the 'tutorial/README' page (and will need to take care to update links)

DanVanAtta (commented):
These docs are critical for developers and it's good to see the ownership and desire to keep the campground clean. It helps the project for these docs to be good -  it is really great to see this update = D

Please review my last comment in most detail. While reviewing I realized that the two documentation updates are because the two documents are duplicates! In short, I propose we merge all updates to `how-to/local-build-and-testing.md ` (that would leave just one file modified for this PR). Then as a next step we merge the entire doc, update links and delete the 'tutorial' page. WDYT?

docs/development/tutorial/README.md
emmanuel-perlow: Sounds good. (I wasn't convinced from tutorial/README as it duplicates many things from /how-to), I'll prepare something.

I also don't like the tutorial/pitfalls.md as this is not a tutorial either.

docs/development/tutorial/README.md
emmanuel-perlow: On my system, gradle prints what it is doing ATM but erases that information and replaces it with a "SUCCESS 127ms" message afterwards.  While figuring out how to run tests, I was curious whether Gradle actually runs the tests I specified but the output was saying nothing in particular I could not differentiate between Gradle skipping compiling&running due to nothings changed and code was run but did not fail.

If you supply `-i` (at least I) got a list of targets run and Gradle showed why or why not it ran the target (`UP-DO-DATE`, `FROM-CACHE`, ...)

docs/development/tutorial/README.md
emmanuel-perlow: I intended to give the name of the launcher/target as presented by the IDE (there are AFAIK three), so the developer in spe knows which one to run and doesn't have to guess.  This is not (directly) related to any "gradle target" or a CLI command line.

docs/development/tutorial/README.md
emmanuel-perlow: I was confused, too, but I had only a JRE and `./gradelw` somehow managed to compile it nevertheless.
This [0] stackoverflow answer indicates that Gradle uses a library.  In another answer (which I cannot find anymore), it was said that `javax.tools` is used and I found that jar file in the `~/.gradle/` directory.

[0] https://stackoverflow.com/questions/65817520/how-do-i-know-for-sure-which-version-of-javac-gradle-used-in-a-build

docs/development/tutorial/README.md
DanVanAtta: One note of caution with specific launcher names, it can make the documentation more easily out of date. Hence the mention there are launchers is slightly vague, but would be accurate enough if more are added or any renamed.

I've viewed CLI as a bit of a fallback and something for full verification. Hence the getting started document has been geared towards IDE setup.

docs/development/tutorial/README.md
DanVanAtta: Wow, that is a bit surprising and interesting to learn.

I think we'd want to keep it simpler and just state a JDK (if it's not already implied and part of setting up an IDE), as the IDE would need a JDK. Getting up and running on CLI  is a good step, but not enough to be fully productive. Notably, a person should be able to launch a debugger, go to method declarations to trace through code, etc, in order to be fully productive, which requires an IDE to do efficiently.

emmanuel-perlow:
This is now reworked and the tutorial/README.md is merged into the corresponding how-to/ files.

**Note**: The first commit is the same as in #9505 (does GitHub support pull requests which are based on each other?).  Either merge #9505 before this one (#9503) or close #9505 and only merge #9503.

docs/development/tutorial/README.md
DanVanAtta: It's a good mention about the pitfalls doc, that would seemingly belong somewhere else as well.

emmanuel-perlow: The thing is, I think it does not fit in any of the four categories (background, how-to, reference, tutorial).
Maybe put it directly under `docs/development`?

PR #9517, opened 2021-08-17
MapsListingFeignClient: Correct javadoc

emmanuel-perlow:
Small doc error I stumbled upon.


RoiEXLab (approved):
Good catch, thank you for your contribution

PR #9518, opened 2021-08-17
Make CTRL-A, CTRL-D, CTRL-B work with focused battle calculator dialog

emmanuel-perlow:
Until now CTRL-A, CTRL-D, CTRL-B do only work when the main window has the focus.  This patchset enhances the battle calculator to also recognize these bindings and act the same way.  Thus, it now does not matter any more whether the main window or the battle calculator has the focus.

## Testing
Not sure its sensible to test whether a key binding is active or not.

Note quite happy about duplicating the functions for JFrame and JDialog but SwingKeyBinding already does it in other functions, so yeah.

## Release Note

NEW|Shortcuts (CTRL-A, CTRL-D, CTRL-B) for adding attackers, defenders or opening a new battle calculator dialog now also work when the battle calculator dialog is focused


RoiEXLab (approved):
Nice addition 👍🏼

PR #9532, opened 2021-08-22
Battle Calculator: Allow fighting any player

emmanuel-perlow:
When changing a player, do not set the other player automatically to an
enemy.  This allows fights between any two players.  Moreover, it does
not erase the battle configuration in case of a missclick.

The later point is for me a pain, as I set up the unit configuration for one
party and then set the other party to the other player and from time to
time is miss the tiny entry with the player and select an ally and just
like that is all the preparation gone.

The first point is important for FFA maps as you might still be allied
with someone but an inevitable war is looming, so you want to be
prepared.

Note: It is not always possible to use a substitute, that is a enemy player
which has the same or similar enough units, to do the calculation as
there are quite some maps with "asymmetric" units, e.g. 270BC.

Most of the time (i.e. n-1 out of n) the automatically selected player was
the wrong enemy, so this feature wasn't that useful after all in my opinion.

I hope this is the only place where the setting of the other player is done.

## Release Note

CHANGE|The battle calculator now allows to select any two players to battle against not only enemies

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
@@ -1095,6 +1095,19 @@ class BattleCalculatorPanel extends JPanel {
           }
         }
 
+        // In the above code, if the territory owner and the current player are the same, then
+        // attacker and defender are. Since we cannot infer whether the user wants to defend the
+        // territory or attack with the units in that territory, we cannot refine this setup
+        // further. To not break the user's habits, we solve this by falling back to the legacy
+        // behaviour and force the attacker and defender to be on non-allied teams.
+        if (data.getRelationshipTracker().isAllied(getDefender(), getAttacker())) {
+          if (location.isWater() && players.isEmpty()) {
+            getEnemy(getAttacker()).ifPresent(this::setDefender);
+          } else {
+            getEnemy(getDefender()).ifPresent(this::setAttacker);
+          }
+        }
+
         setAttackingUnits(
             location.getUnitCollection().getMatches(Matches.unitIsOwnedBy(getAttacker())));
         setDefendingUnits(
@@ -1359,6 +1372,27 @@ class BattleCalculatorPanel extends JPanel {
 
   private boolean isLand() {
     return landBattleCheckBox.isSelected();
+  }
+
+  /**
+   * Get a suitable enemy for {@code player}. First checks for players at war, if none found, checks
+   * for non-allied players, if still no foe was found, returns an empty Optional.
+   *
+   * @param player the player to find a foe for
+   * @return Optional with not an ally or an empty Optional
+   */
+  private Optional<GamePlayer> getEnemy(final GamePlayer player) {
+    for (final GamePlayer gamePlayer : data.getPlayerList()) {
+      if (data.getRelationshipTracker().isAtWar(player, gamePlayer)) {
+        return Optional.of(gamePlayer);
+      }
+    }
+    for (final GamePlayer gamePlayer : data.getPlayerList()) {
+      if (!data.getRelationshipTracker().isAllied(player, gamePlayer)) {
+        return Optional.of(gamePlayer);
+      }
+    }
+    return Optional.empty();
   }
 
   private void setResultsToBlank() {
--- a/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
@@ -957,14 +957,6 @@ class BattleCalculatorPanel extends JPanel {
 
     defenderCombo.addActionListener(
         e -> {
-          data.acquireReadLock();
-          try {
-            if (data.getRelationshipTracker().isAllied(getDefender(), getAttacker())) {
-              attackerCombo.setSelectedItem(getEnemy(getDefender()));
-            }
-          } finally {
-            data.releaseReadLock();
-          }
           setDefendingUnits(
               defendingUnitsPanel.getUnits().stream().anyMatch(Matches.unitIsOwnedBy(getDefender()))
                   ? defendingUnitsPanel.getUnits()
@@ -973,14 +965,6 @@ class BattleCalculatorPanel extends JPanel {
         });
     attackerCombo.addActionListener(
         e -> {
-          data.acquireReadLock();
-          try {
-            if (data.getRelationshipTracker().isAllied(getDefender(), getAttacker())) {
-              defenderCombo.setSelectedItem(getEnemy(getAttacker()));
-            }
-          } finally {
-            data.releaseReadLock();
-          }
           setAttackingUnits(null);
           setWidgetActivation();
         });
@@ -1375,21 +1359,6 @@ class BattleCalculatorPanel extends JPanel {
 
   private boolean isLand() {
     return landBattleCheckBox.isSelected();
-  }
-
-  private GamePlayer getEnemy(final GamePlayer player) {
-    for (final GamePlayer gamePlayer : data.getPlayerList()) {
-      if (data.getRelationshipTracker().isAtWar(player, gamePlayer)) {
-        return gamePlayer;
-      }
-    }
-    for (final GamePlayer gamePlayer : data.getPlayerList()) {
-      if (!data.getRelationshipTracker().isAllied(player, gamePlayer)) {
-        return gamePlayer;
-      }
-    }
-    // TODO: do we allow fighting allies in the battle calc?
-    throw new IllegalStateException("No enemies or non-allies for :" + player);
   }
 
   private void setResultsToBlank() {


Cernelius:
It has been a long time since I complained that it was a wrong behaviour for FFA not to allow to calculate battles between allies, as your ally this round can be your enemy subsequently. Despite guessing this being an easy fix, I never found a developer interested in making it (erasing the limitation by which you are not allowed to calculate between allies), so it is nice to see this very old problem being finally solved.

Moreover, another thing that I recall to have reported a long time ago is that (as it can fairly easily happen in some FFA), if it happens that everyone is allied with everyone, the battlecalculator becomes simply unusable and errors out (you cannot even open the battlecalculator when everyone is allied with everyone). So, if this problem is still present, this change should fix it too.

However, the concept of not allowing battlecalculating between players that are never supposed to fight each other is sound. If you go with simply allowing everyone to battlecalculate against everyone, I can easily see a developer in the future seeing, for example, the battlecalculator in Revised allowing calculating Russians fighting Americans as a wrong allowance so reintroducing what is being eliminated with this change.

The point is that, instead of disallowing battlecalculating between players in an allied relationship, the program should disallow battlecalculating within the same alliance.

I mean looking at
https://github.com/triplea-maps/the_pact_of_steel/blob/master/map/games/pact_of_steel_2.xml#L638
instead of looking at
https://github.com/triplea-maps/the_pact_of_steel/blob/master/map/games/pact_of_steel_2.xml#L1618
.

---

As a side note, there may be some weird map that allow fighting between same-alliance players, but I'm not aware there is any such map currently.

---

The conclusion is that I prefer this change over the current problematic behaviour, yet I would rather prefer keeping the feature but substituting checking for allied relationships with looking at shared alliances.

---

Additionally, I don't feel like the issue of mistakenly erasing your selection for selecting a player you are supposed never to have need to select is correcting addressed by allowing calculating for something you should never need to calculate. This may be better addressed by changing the behaviour of the feature rather than by completely removing it. For example, by making the button starting the process unavailable when all defenders are players of the alliance of the attacker.

DanVanAtta:
I was wondering if the calc would work having two allied factions fighting against each other but it appears it does.

Checking this out, I found the initial state is a bit odd where both defender and attacker are populated from the same territory:
Screenshot from 2021-08-28 10-49-49

It's hard to say whether that might be 'better' as most calc's will be of similar units vs similar units, so it might save some clicks. On the other hand, selecting an enemy faction first is now an extra step and the typical player will want it to be two opposing factions. By that reasoning it would make sense to keep the initial behavior where when the battle calc window loads that an opposing faction is automatically selected, but if a player changes one side or the other then we do not automatically update the opposing side. WDYT?

Cernelius:
As it should be clear from what I've already written, I believe it needs to be clarified what exactly are "two opposing factions" before anyone can reliably agree or disagree with any proposal.

Ideally, every game should have a full list of each couple of factions for which there is no need to battlecalculate, or the program should be very smart and read through all the triggers and actions in the game file and sort it out automatically. Realistically, my suggestion is that opposing factions are every couple of so-called players which do not belong to the same alliance (whether or not they are sharing an allied relationship at the start or at any point during the game).

As for the example, I think the "Calculate Odds" button should not be actionable when you have a situation like that.

---

By the way, the "-1" things as to mean infinite in the two retreat options are some terrible UI. The second one doesn't even make sense, because, if "-1" is infinite, then "Retreat When X Units Left" would mean "infinite units left"! There should be a tick mark for each of them (never retreating if deselected) and the ability to have only positive integers only if ticked.

emmanuel-perlow:
> Checking this out, I found the initial state is a bit odd where both defender and attacker are populated from the same territory:

If you select a territory, where you fight against another player that fight is calculated (true for both, if you attack them on their grounds, and if they attack you on your grounds, and even for 3rd partyA vs 3rd partyB).

The scenario you highlighted will only happen when you select your own territory with no fight going on.
When selection your own army in your own territory, there are (in my opinion) two intentions the player could have:

a) you want to calculate a defence on that territory => your units should be the defenders
b) you want to plan an attack from that teritory => your units should be the attackers

It is impossible to know which you want and setting the calculator up for both (i.e. you are attacker and defender) is probably better than assuming an intention and randomly select an opponent (most of the time the opponent select is the wrong one anyway).

DanVanAtta:
@emmanuel-perlow there are some assumptions that are hard to truly make without data. Given that I would be hesitant to change the existing behavior too much. Players are used to it, for better or worse, and changing that expectation will take players out of the game flow and force some relearning (so better to not change some things unless it really is justified).

Battle calc is used very heavily in multiplayer games and rarely is it for a battle where all the units are moved. If anything, because you don't want to give away to your opponent exactly which battles you are considering and secondarily doing all of the movement for the sake of calc is too time consuming. Hence, it's my experience that battle calc almost always starts with 'ctrl-b' on units in one territory and then the other side is manually added or added using 'ctrl-a' or 'ctrl-d'. Also to consider, having the territory come up automatically as attacker or defender makes sense, if it's another players turn then a person is probably more interested in the defensive values and vice versa. Otherwise, there is the 'swap' sides expressly to change things around after the initial 'ctrl-b' press.

>  is probably better than assuming an intention and randomly select an opponent (most of the time the opponent select is the wrong one anyway).

Players are used to this, so FWIW that is a reason to keep it how it was unless something much obviously better is found. Second, usually the opposing faction has similar units across any faction, so whether a calc is done vs 'Germany' and not 'Japan' often does not matter.

IMO we should keep some of the original behavior but still allow for calc's to be done against arbitrary factions (as implemented here)

emmanuel-perlow:
Ok, I included some mimicry code.

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
```diff
@@ -1111,6 +1095,16 @@
           }
         }
 
+        // Mimic the legacy behavior and force the parties to be non-allied modifying either
+        // attacker or defender like the old code did). This legacy behavior was requested in #9532.
```
DanVanAtta, line 1099: A potential maintainer reading this comment might be curious what "legacy behavior" refers to, it's not clear without context of a PR. Requiring a reader to reference a PR is a smell, PRs are ephemeral (and subsequent PRs can change things).

This comment perhaps should focus on explaining what we are doing in this code, what the behavior is, rather than it is simply "legacy" (albeit, still the current behavior).

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
```diff
@@ -1111,6 +1095,16 @@
...
+        if (data.getRelationshipTracker().isAllied(getDefender(), getAttacker())) {
```
DanVanAtta, line 1100: Is this if statement a tautology? When the battle calculator loads initially, I observed the attacker and defender were the same. Hence, checking if an attacker is allied with defender (same player) would always be true.

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
```diff
@@ -1111,6 +1095,16 @@
...
+          if (location.isWater() && players.isEmpty()) {
+            getEnemy(getAttacker()).ifPresent(this::setDefender);
+          } else {
+            getEnemy(getDefender()).ifPresent(this::setAttacker);
```
DanVanAtta, line 1104: Can you clarify why the battle location and players being empty or not would change whether we need to set the 'defender' or 'attacker'?

DanVanAtta (commented):
Just a couple last comments/questions about the latest commit.

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
emmanuel-perlow: As I explained in https://github.com/triplea-game/triplea/pull/9532#issuecomment-908161318, the behaviour you describe happens if the current player selects its own territory.  It does not occur when the battle calculator is initiated on a territory belonging to a foreign player, e.g. current player is Carthage and (without moving pieces) you select a territory of the Greek City states, then Carthage is the attacker and the City States are the defender, in particular they are not allied.  Moreover, we do not want to override the City States with Rom, which would be the first entry in the list of enemies of Carthage.  Hence, this guard.

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
emmanuel-perlow: Sure, that's what the old code did.  If you asking why the old code chose those conditions and why they chose to alter attacker over defender or vice versa, then you have to ask the person who initially wrote that.

DanVanAtta: Walking through this in a debugger, I think the comment no longer agrees with the if/else statement. Essentially the logic is says:
- if we have not found two opposing factions
  - if we are water, then the attacker is correct and find an enemy faction
  - else, defender is correct and find an enemy faction
Basically this seems to be an extension of the previous logic that is assigning an attacker and a defending player. Hence, it's odd to have a comment stating that we are going to choose two opposing sides here to keep with a users expectations when the previous 30 lines of code are also doing exactly that but for different cases. There is an implication that this could never have been removed and/or the preceding logic is dead code.

Does it make sense why I say the comment and this if statement do not jive together?

I think we are also violating the DRY principle where we do the same thing in multiple places. Notably if we assign an attacker then the same assignment earlier would be a dead store. Seemingly this if block needs to be merged with preceding code for it to make more sense.

emmanuel-perlow: > Basically this seems to be an extension of the previous logic that is assigning an attacker and a defending player.

It's not an extension, the same thing was also done by the old code but was well hidden behind a chain of EventListeners and was removed in this PR:

https://github.com/triplea-game/triplea/blob/6fd988dfa42fd7e46e0660276599aa6fa496d81b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java#L959-L961


> Basically this seems to be an extension of the previous logic that is assigning an attacker and a defending player. Hence, it's odd to have a comment stating that we are going to choose two opposing sides here to keep with a users expectations when the previous 30 lines of code are also doing exactly that but for different cases. There is an implication that this could never have been removed and/or the preceding logic is dead code.
>
> Does it make sense why I say the comment and this if statement do not jive together?

Sorry, can't follow you there.  I think maybe the confusion arises from the fact that the code change in the event listener affected this logic?  Since this code is now gone, we have to move parts of the event listener logic here in the setup code.

> I think we are also violating the DRY principle where we do the same thing in multiple places.

Yes, it does but moving this code into the above stacked if-cases would duplicate this if three times (we cannot join the isWater() and isEmpty() case).  I contemplated about it but decided for 2 as it is smaller than 3.  Essentially we have a 3x2 grid guarded by ifs and I don't think we can avoid that.

> Notably if we assign an attacker then the same assignment earlier would be a dead store.

I can understand your hesitation about dead store but

a) it is already dead store, e.g. attacker get assigned (potentially) multiple times:
   - https://github.com/triplea-game/triplea/blob/6fd988dfa42fd7e46e0660276599aa6fa496d81b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java#L1075
   - https://github.com/triplea-game/triplea/blob/6fd988dfa42fd7e46e0660276599aa6fa496d81b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java#L1086  (jepp, that one did in the old code reset the attacker via the event listener)
   - https://github.com/triplea-game/triplea/blob/6fd988dfa42fd7e46e0660276599aa6fa496d81b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java#L1089
   - https://github.com/triplea-game/triplea/blob/6fd988dfa42fd7e46e0660276599aa6fa496d81b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java#L1095  (dito)
   - https://github.com/triplea-game/triplea/blob/6fd988dfa42fd7e46e0660276599aa6fa496d81b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java#L1099  (dito)
   - https://github.com/triplea-game/triplea/blob/6fd988dfa42fd7e46e0660276599aa6fa496d81b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java#L1101

b) I cannot see how to make a single assignment for attacker/defender in all cases, e.g. the land case would become
```java
if (!location.isWater()) {
        defenderCombo.setSelectedItem(location.getOwner());
        boolean setAttacker = false;
        for (final GamePlayer player : players) {
                if (Matches.isAtWar(getDefender(), data.getRelationshipTracker()).test(player)) {
                        attackerCombo.setSelectedItem(player);
                        setAttacker = true;
                        break;
                }
        }
        if (!setAttacker) {
                // Currently player has priority if at war.
                if (currentPlayer.isPresent() && Matches.isAtWar(getDefender(), data.getRelationshipTracker()).test(currentPlayer.get())) {
                                attackerCombo.setSelectedItem(currentPlayer.get());
                } else {
                        // find an enemy if possible
                        Optional<Player> enemy = getEnemy(getDefender());
                        if (enemy.isPresent()) {
                                attackerCombo.setSelectedItem(enemy.get());
                        } else {
                                // Fall back to current player (they are allied but there are no enemies anyway)
                                if (currentPlayer.isPresent()) {
                                        attackerCombo.setSelectedItem(currentPlayer.get());
                                } else {
                                        attackerCombo.setSelectedItem(getFirstPlayer();
                                }
                        }
                }
        }
} else {...}
```

(pseudo code might not compile)

So what should we do?

DanVanAtta: Thank you for the response. To answer your question, we either spend quite a bit more time refactoring to get maybe cleaner code, or we live with where we are. I am concerned by both scope creep and ROI if we spend a lot more time now, on this. So, having been given more context, I think we can move on and call things good enough. I re-read the diff and it's quite reasonable.

PR #9533, opened 2021-08-22
Battle Calculator: Simplify code

emmanuel-perlow:
Do not use a class global JComboBox to store one player temporary when
swapping side, instead use (function) local variables.

We could have used a single temporary variable (as it is essentially a swap)
but using two properly named variables is more readable IMHO.


DanVanAtta:
:+1:

PR #9555, opened 2021-08-31
(Refactor) Rename/Merge 'MapsHttpListingClient' & 'MapsListingClient' -> 'MapsClient'

DanVanAtta:
Merges an interface and class into a single renamed class with a more generic name.


RoiEXLab (approved):
👍🏼

PR #9585, opened 2021-09-09
Battle Simulator: Use library to calculate mean

emmanuel-perlow:
Split out of #9490.


emmanuel-perlow:
The JavaDoc update follows in another PR once this is merged as this is a unrelated change and should not be squashed.
However, currently I can't proceed as this yields in merge conflicts.

Is there a way to submit multiple PR based on each other? (i.e. the end results should be to have different commits in the history not one squashed commit).  Even the first two commits of this PR are unrelated to the mean change and should have been a separate PR, however, that would have delayed this further.

DanVanAtta (approved):

PR #9601, opened 2021-09-12
Battle Calculator: Use GameData instead of History

emmanuel-perlow:
I noticed that both GameData and History passed from external to be used in the Battle Simulator.  I'm not sure whether is intentional or organic growth.  I replaced the occurrences of `history` with `data.getHistory()`, which previously was done on the caller side.

Note that a comment warned that  `taFrame.getGame().getData().getHistory()` can be different from `history`. I guess this is still true for the game data.

I'm not yet 100% sure with the complete codebase, so there might be instances where the stored GameData and the passed history might differ but I'm confident there isn't.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorDialog.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorDialog.java
@@ -1,9 +1,9 @@
 package games.strategy.triplea.odds.calculator;
 
+import games.strategy.engine.data.GameData;
 import games.strategy.engine.data.GamePlayer;
 import games.strategy.engine.data.Territory;
 import games.strategy.engine.data.Unit;
-import games.strategy.engine.history.History;
 import games.strategy.triplea.delegate.Matches;
 import games.strategy.triplea.ui.TripleAFrame;
 import java.awt.BorderLayout;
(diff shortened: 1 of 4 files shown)


DanVanAtta:
The data copy of 'game data' and history is something that we want to avoid and ideally battle calculator would only get a subset of data representing unit data and not the full game data. It turns out even that AI performance time is dominated by the game data copy to do battle calc's, as it performs a lot of battle calc's and the slow part is copying game data.

More context, we're trying to avoid passing 'GameData' as an object and instead should favor using 'GameState' or passing the data needed from 'GameData' rather than 'GameData' itself. This is an effort to avoid having 'GameData' be such a god-object where it subsumes everything and is used everywhere.

With all that said, this is potentially very brittle code and we should exercise caution. @emmanuel-perlow have you done any kind of testing where the game data changes in the background while a battle calc is open? For example, that could be doing a battle calc while the AI is running in the background.

emmanuel-perlow:
> More context, we're trying to avoid passing 'GameData' as an object and instead should favor using 'GameState' or passing the data needed from 'GameData' rather than 'GameData' itself. This is an effort to avoid having 'GameData' be such a god-object where it subsumes everything and is used everywhere.

I see.  I haven't looked into what parts of `GameData` the `BattleCalculatorPanel` is using (and currently don't intend to).  I just noticed that it takes `History` and then retrieves `GameData` on its own (from `TripleAFrame`) and then mixing those sources when choosing the initial players. Moreover, on caller side it was `data.getHistory()`. So I figured it might be simpler passing the complete object directly.

> With all that said, this is potentially very brittle code and we should exercise caution. @emmanuel-perlow have you done any kind of testing where the game data changes in the background while a battle calc is open? For example, that could be doing a battle calc while the AI is running in the background.

Yes. And to be absolutely sure I just redid them. This time I hit some exception while continuously spamming Ctrl-B (which thanks to #9518 is now easier).   The "good news" is: it's unrelated to this PR as it can be triggered on branch `master`, too, bug report follows...

Cernelius:
I'm not sure it is related, but you can use my "270BC Wars" game as an example of very slow loading of data in the battle-calculator (due to the game file being very big).

Cernelius:
It would be very good if the loading of the game data on the battle-calculator while playing "270BC Wars" can be made faster. Having to wait so much before being able actually to start the simulations is a serious playability problem on that game (of mine), and I'm planning to release even heavier games.

emmanuel-perlow:
> It would be very good if the loading of the game data on the battle-calculator while playing "270BC Wars" can be made faster. 

@Cernelius could you please file a new issue for this, so we can track it properly. I guess this requires some more heavy refactoring than moving a function call from A to B as in this PR. Thanks!

DanVanAtta (approved):

PR #9602, opened 2021-09-12
Battle Simulator: Rewrite initial opponents choosing logic

emmanuel-perlow:
Rewrite of the code selecting the attacker and defender for the battle calculator to tackle some if the issues raised in #9532. The new code is (hopefully) stateless and has no dead stores but still has a lot of nested ifs. Some of these ifs are "hidden" – or better "abstracted away" – inside streams and optionals (`filter(...), `ifPresent(...)`) and in concatenating (potentially empty) streams.

The new logic is slightly different than the old one in corner cases:

* The new code does not assume "transitive diplomacy"; by that I mean if A is at war/allied with B, and B is at war/allied with C, then this does not imply that A is at war/allied with C.  This happens regularly on FFA maps (FFA = Free For All).

* Previously, only the strongest or current player was checked for being at war in the territory, now any combat between two parties is detected and the respective players are given priority (still ordered by unit count).

* The new code prioritises fights: If a fight happens then it is used as initial setup.  The old code did this only in some cases.

* Selecting based on units is given priority over selection based on territory owner ship.  This can be seen in the following scenario: A,B allies, and on a territory of A only B has units.
  Previously, A was set as defender, now B is set as defender.
  This might be again useful in FFA cases, e.g. A (current player) at war with B, and at peace with C; B has units on territory of C.
  old: B was selected as defender
  new: A is selected as attacker and C as defender (which probably is what A wants to calculate).

  If requested this can be reversed but fitted better in the (mental) logic flow: first check units/fights, then see what else can be derived.

* I'm pretty sure that if a player has no enemies, or the current player does not exist, the new logic produces slightly different outcomes than the old one.

* and maybe other corner cases I'm not aware of

This code depends on the refactoring done in #9601 and consequently includes that commit.  **THIS COMMIT SHOULD NOT PART OF THIS PR** as it is unrelated to the choosing logic! Sadly, GitHub does not support such PRs and the Squashing policy would mix those commits.

If PR #9601 is rejected, then the `History` object must be passed to the `setupAttackerAndDefender()` function, that's it. The logic is not affected by that.

For reviewing I suggest looking at the results and not at the diff, as the codes have little in common.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
@@ -1062,7 +1062,7 @@ class BattleCalculatorPanel extends JPanel {
 
   private void setupAttackerAndDefender() {
     final OpponentSelector.AttackerAndDefender attAndDef =
-        OpponentSelector.getAttackerAndDefender(location, data);
+        OpponentSelector.with(data).getAttackerAndDefender(location, data);
     attAndDef.getAttacker().ifPresent(this::setAttacker);
     attAndDef.getDefender().ifPresent(this::setDefender);
 
--- a/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
@@ -1044,6 +1044,11 @@ class BattleCalculatorPanel extends JPanel {
     defendingUnitsPanel.addChangeListener(this::setWidgetActivation);
 
     setupAttackerAndDefender();
+    if (location == null) {
+      landBattleCheckBox.setSelected(true);
+    } else {
+      landBattleCheckBox.setSelected(!location.isWater());
+    }
 
     calculator =
         new ConcurrentBattleCalculator(
@@ -1065,22 +1070,8 @@ class BattleCalculatorPanel extends JPanel {
         OpponentSelector.with(data).getAttackerAndDefender(location);
     attAndDef.getAttacker().ifPresent(this::setAttacker);
     attAndDef.getDefender().ifPresent(this::setDefender);
-
-    // Now that the attacker and defender are determined, select their units on this territory.
-    // For the defender this also includes allied units.
-    if (location == null) {
-      landBattleCheckBox.setSelected(true);
-      setDefendingUnits(null);
-      setAttackingUnits(null);
-    } else {
-      landBattleCheckBox.setSelected(!location.isWater());
-      setAttackingUnits(
-          location.getUnitCollection().getMatches(Matches.unitIsOwnedBy(getAttacker())));
-      setDefendingUnits(
-          location
-              .getUnitCollection()
-              .getMatches(Matches.alliedUnit(getDefender(), data.getRelationshipTracker())));
-    }
+    setAttackingUnits(attAndDef.getAttackingUnits());
+    setDefendingUnits(attAndDef.getDefendingUnits());
   }
 
   GamePlayer getAttacker() {
--- a/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
@@ -1062,7 +1062,7 @@ class BattleCalculatorPanel extends JPanel {
 
   private void setupAttackerAndDefender() {
     final OpponentSelector.AttackerAndDefender attAndDef =
-        OpponentSelector.with(data).getAttackerAndDefender(location, data);
+        OpponentSelector.with(data).getAttackerAndDefender(location);
     attAndDef.getAttacker().ifPresent(this::setAttacker);
     attAndDef.getDefender().ifPresent(this::setDefender);
 
--- a/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
@@ -953,12 +953,12 @@ class BattleCalculatorPanel extends JPanel {
           setDefendingUnits(
               defendingUnitsPanel.getUnits().stream().anyMatch(Matches.unitIsOwnedBy(getDefender()))
                   ? defendingUnitsPanel.getUnits()
-                  : null);
+                  : List.of());
           setWidgetActivation();
         });
     attackerCombo.addActionListener(
         e -> {
-          setAttackingUnits(null);
+          setAttackingUnits(List.of());
           setWidgetActivation();
         });
     amphibiousCheckBox.addActionListener(e -> setWidgetActivation());
@@ -966,8 +966,8 @@ class BattleCalculatorPanel extends JPanel {
         e -> {
           attackerOrderOfLosses = null;
           defenderOrderOfLosses = null;
-          setDefendingUnits(null);
-          setAttackingUnits(null);
+          setDefendingUnits(List.of());
+          setAttackingUnits(List.of());
           setWidgetActivation();
         });
     calculateButton.addActionListener(e -> updateStats());
--- a/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
@@ -50,7 +50,6 @@ import javax.swing.JScrollPane;
 import javax.swing.ListSelectionModel;
 import javax.swing.ScrollPaneConstants;
 import javax.swing.SwingUtilities;
-import lombok.Builder;
 import lombok.extern.slf4j.Slf4j;
 import org.triplea.java.ThreadRunner;
 import org.triplea.java.collections.CollectionUtils;
--- a/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
@@ -1073,8 +1073,9 @@ class BattleCalculatorPanel extends JPanel {
               .players(data.getPlayerList().getPlayers())
               .currentPlayer(data.getSequence().getStep().getPlayerId())
               .relationshipTracker(data.getRelationshipTracker())
+              .territory(location)
               .build()
-              .getAttackerAndDefender(location);
+              .getAttackerAndDefender();
 
       attAndDef.getAttacker().ifPresent(this::setAttacker);
       attAndDef.getDefender().ifPresent(this::setDefender);
--- a/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
@@ -35,8 +35,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import javax.swing.BorderFactory;
 import javax.swing.BoxLayout;
 import javax.swing.DefaultListCellRenderer;
@@ -57,6 +55,7 @@ import org.triplea.java.ThreadRunner;
 import org.triplea.java.collections.CollectionUtils;
 import org.triplea.swing.IntTextField;
 import org.triplea.swing.SwingComponents;
+import org.triplea.util.Tuple;
 
 @Slf4j
 class BattleCalculatorPanel extends JPanel {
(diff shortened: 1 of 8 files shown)


emmanuel-perlow:
If the "climate checks" should be satisfied, I can try to split up the function further, though I'm not sure that would make it more readable.

DanVanAtta:
The code climate check probably should boost the max method length to 75.

There are not very many good correlations between code quality metrics and bug, except code length. The longer the methods, the longer the classes, and the more lines total, the more bugs. Hence avoiding too long methods is something that ought to be respected.

When it comes to modifying existing code, if there is a code climate warning we can likely ignore it as the problem was not introduced. Though, new code should be clean of any such warnings.

In terms of this PR I think we can make some good gains by breaking things up. There are pretty lengthy bodies in 'if/else' blocks which would be good candidates for extraction. IMO even better would be to put all this logic into a new class named something like "DefaultPlayerSelector". With that, it should be very possible to then test it as well to verify that the logic is good. Let me know how that sounds and if agreeable if any further pointers in that direction would be helpful. I think overall I'm most interested in seeing the logic extracted (so it does not live in a UI class), and second tested.

emmanuel-perlow:
> IMO even better would be to put all this logic into a new class named something like "DefaultPlayerSelector".

I've extract and split up the logic into its own class I don't like the term 'default' that much as its not "the default" but a elaborated logic – in fact so much that it needs its own class ;-) – so it ended up with `OpponentSelector` better naming suggestions are welcome.

> With that, it should be very possible to then test it as well to verify that the logic is good.

I've not yet added test, as we should first agree on whether this logic is acceptable as it deviates in corner cases from the old logic.

emmanuel-perlow:
Memo to self:
Replace `Tuple` with public inner class `AttackerAndDefender` (of `OpponentSelector`) with public variables `attacker` and `defender`.

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
```diff
@@ -1062,185 +1061,27 @@
     revalidate();
   }
 
-  /**
-   * Set initial attacker and defender.
-   *
-   * <p>Please read the source code for the order of the players and conditions involved.
-   */
   private void setupAttackerAndDefender() {
+    final Tuple<Optional<GamePlayer>, Optional<GamePlayer>> t =
```
emmanuel-perlow, line 1065: Replace `Tuple` with public inner class `AttackerAndDefender` (of `OpponentSelector`) with public variables `attacker` and `defender`.

DanVanAtta:
Quick comment, public variables can be avoided and favor immutability.
Check put the lombok @Value annotation and that can be paired with @Builder
for a more fluent constructor (particularly handy when the constructor args
are the same type)

DanVanAtta:
#rant #discussion

> I've not yet added test, as we should first agree on whether this logic is acceptable as it deviates in corner cases from the old logic.

Responding to this, in some ways test cases make it easier to evaluate what the new logic is. If we decide to change the logic, the test case updates should be pretty straight forward to update expected outcomes. When test cases are included, it facilitates review in a lot of ways as I tend to focus on whether the test cases are complete and reasonably easy enough to understand. If that is in place, then the logic check is just checking that things seem easy to understand. Bottom line, I think writing test cases early is rarely going to hurt (unless we go the route of heavy mocking, in which case there is a lot of rewrite).

One strategy to consider, an idea that came to me the other day regarding this, would be to design the extracted class to receive the value object you want. So write a value object which gives an idealized data representation needed to do whatever data computation. This makes things very easy to test because you can new up the "calculator" object and create some readily made sample data. Then to integrate this, there is a needed extraction from GameData to the value object (which can be done perhaps via a constructor of the value object), then it's a matter of invoking the new extracted object.

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorDialog.java
```diff
@@ -46,14 +46,11 @@
...
-    // GameData
-    // gets cloned when showing history with a different History instance that doesn't correspond to
-    // what's
-    // shown.
-    final BattleCalculatorPanel panel =
-        new BattleCalculatorPanel(taFrame.getGame().getData(), history, taFrame.getUiContext(), t);
+  public static void show(final TripleAFrame taFrame, final Territory t, final GameData data) {
+    // Note: The data param may not be the same as taFrame.getGame().getData() as GameData gets
+    // cloned when showing history with a different History instance that doesn't correspond to
+    // what's shown.
```
DanVanAtta, line 52: I think this comment probably could be removed. I'd opt to put a mention in the javadoc that we use the 'data' param to init the battle calculator. Then it's clear we are not using the game data from 'taFrame'

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/OpponentSelector.java
```diff
@@ -0,0 +1,260 @@
...
+
+/**
+ * Utility class with methods to pick players for the battle calculator.
+ *
+ * <p>Throughout the class the term "neutral player" is used, this denotes a player who is neither
+ * at war nor allied with another player.
+ */
+public class OpponentSelector {
+
+  /** Utility class, no instances please. */
```
DanVanAtta, line 21: Minor, can replace the constructor and comment with class level 'lombok' annotation: `@UtilityClass`

DanVanAtta:
Interesting updates @emmanuel-perlow 

It's better to have a predictable algorithm than a smart and slightly unpredictable algorithm. Also, making it very simple makes a user action more intuitive, require less thought, overall the simpler and less thought power the more automatic and easy something is to use. In other words, instead of wondering which territory to select to get the right attacker/defender combo, our players will just ctrl+b to get the right attacker/defender combo and if not they know and expect to change it and will be quick to do so. My 2 cents, hopefully that makes sense as a goal: predictable > smart

I like the ideas presented and think we should incorporate it all. WDYT if we keep with attacker is always current player. Reasons:
- considering it in detail, I think that is the first part of the current algo
- that decision simplifies the selection quite a bit, then the only question is who to select as the defender

If we go with attacker is current player, then defender selection becomes:
- all at-war players in selected territory. If none, then things get interesting. I think the previous algorithm defaulted to any at-war player.

From another perspective, the battle calculator "ctrl+b" is answering the question of "what if the current player attacked X", or said another way, "what if the current player fights X battle"

Thoughts?

emmanuel-perlow:
> I like the ideas presented and think we should incorporate it all. WDYT if we keep with attacker is always current player. Reasons:
>
> *   considering it in detail, I think that is the first part of the current algo
> *   that decision simplifies the selection quite a bit, then the only question is who to select as the defender

Not quite sure this is what we want as the current player also often wants to calculate a defense, e.g. when select one of their own territories. Thus, something like
```
if (territory.getOwner() == currentPlayer)
    defender = currentPlayer
else
    attacker = currentPlayer
```
might be more appropriate.
[... thinking about sea zones...]
On a second though, we have the swap sides button and also it might be that the player wants to select those units in the territory for an attack. So, I'm for attacker = current player.

> If we go with attacker is current player, then defender selection becomes:
>
> *   all at-war players in selected territory. If none, then things get interesting. I think the previous algorithm defaulted to any at-war player.

I'm for it the new code does this:

1. players at war in selected territory
2. players at war
3. players neutral in selected territory
4. players neutral
5. any player

(that would be the proposed `getOpponentWithPriorityList(playersInTerritory)`).
Though, I don't think that neutrals are that important and rather a corner case so leaving them out would simplify things even further.

Remark: I noticed that the diplomacy seems to be configurable by the map as sometimes it's "neutral" and other times it's "neutral_friendly" (I believe communists in Domination: No Man's Land) or so.  Furthermore, neutral is not a thing than rather a not-thing (it's not at war and not allied). What I want to say is that including neutral would make the algorithm sensitive to a more fine grained relationship ordering which is too flexible and potentially not what we currently think about when thinking in terms of the standard map, so better leave it out?

emmanuel-perlow:
> > I've not yet added test, as we should first agree on whether this logic is acceptable as it deviates in corner cases from the old logic.
>
> Responding to this, in some ways test cases make it easier to evaluate what the new logic is. If we decide to change the logic, the test case updates should be pretty straight forward to update expected outcomes. When test cases are included, it facilitates review in a lot of ways as I tend to focus on whether the test cases are complete and reasonably easy enough to understand. If that is in place, then the logic check is just checking that things seem easy to understand. Bottom line, I think writing test cases early is rarely going to hurt (unless we go the route of heavy mocking, in which case there is a lot of rewrite).

I understand and agree with your points and that writing tests usually doesn't hurt. However, I (as probably any other person, too) have time constraints, a day is only so long.  Writing thorough tests takes time and if there are potentially significant changes or rejects, then a lot of time could be lost because we didn't agree beforehand on what we want or – from my perspective – what the project is willing to accept.  For example in our discussion of this PR we are thinking about making the attacker the current player, which would obsolete about 3/4 of the logic and consequently about 3/4 of the tests would be wasted time, my time.  At some point I have to make a call what I'm willing to give to the project in advance and what to hand-in later when the dust settled. I hope you can understand that.

emmanuel-perlow:
> One strategy to consider [...] would be to design the extracted class to receive the value object you want. So write a value object which gives an idealized data representation needed to do whatever data computation.

Do I understand correctly, that by "value object" you mean an interface, something like
```java
interface ICurrentlyCantThinkOfASuitableName {
	GamePlayer getCurrentPlayer();
	RelationshipTracker getRelationshipTracker();
}
```
and then pass an object (maybe GameData, maybe a wrapper object, maybe a mocked object) implementing this interface to the method `getAttackerAndDefender()` as a replacement of `GameData`?

Or do you really mean an *object*
```java
@Value
class StillDontHaveASuitableName {
	GamePlayer currentPlayer;
	RelationshipTracker relationshipTracker;
}
```

My gut feeling tells me that an interface is probably the better choice.

Other languages have the feature to let you combine multiple interfaces in other words require a parameter to implement two (or more) given interfaces.  This would be helpful as we would do something like
```java
interface HasCurrentPlayer { GamePlayer getCurrentPlayer(); }
interface HasRelationshipTracker { RelationshipTracker getRelationshipTracker(); }

// [...]
void getAttackerAndDefender(HasCurrentPlayer + HasRelationshipTracker data) {...}
```
Sadly, I don't think Java has this kind of feature and thus interfaces must be combined into a new interface:
```java
interface HasCurrentPlayerAndRelationshipTracker extends HasCurrentPlayer, HasRelationshipTracker {}
```

DanVanAtta:
@emmanuel-perlow , a Java class can implement multiple interfaces.

By value object, I did mean an object annotated with `@Value`. 

I was thinking to go further and try and decouple more entirely from the existing TripleA data model.

EG:
```
OpponentSelectionParameters  {
     String currentPlayerName;
     Map<String, String> atWarRelationships;
}
```

Although, the weak types there I think are a problem. So that suggestion turns out to not have been very good.

Though, I think you are actually correct about interfaces here.  EG:
```
OpponentSelectionParameters {
  boolean arePlayersAtWar(Player player1, Player2);
  Player getCurrentPlayer();
```

A test class could create a simple object that implements the interface, meanwhile in the game code proper we would mind an implementation to 'GameData' and it's objects.

emmanuel-perlow:
> a Java class can implement multiple interfaces.
> [...]
> Though, I think you are actually correct about interfaces here. EG:
>
>```
>OpponentSelectionParameters {
>  boolean arePlayersAtWar(Player player1, Player2);
>  Player getCurrentPlayer();
>```
>
>A test class could create a simple object that implements the interface, meanwhile in the game code proper we would mind an implementation to 'GameData' and it's objects.

I'm aware that classes can implement multiple interfaces.  I was talking about a function *requiring* multiple (unrelated) interfaces (e.g. `Cloneable` and `Serializable`; example see above).  This is AFAIK not possible in Java.  Even if we combine interface `A` and interface `B` into an interface `AB` (empty interface which just extends `A` and `B` but adds nothing new) and request that in the function signature, then the objects/classes still must implement `AB`, implementing only the interfaces `A` and `B` is not sufficient.

The background for requesting multiple interfaces being that we can define "properties" (potentially very fine grind), e.g. "HasCurrentPlayer", "HasRelationShips", "HasTerritories", ... and the component/algorithm then requests the combination of the properties it needs. Then classes which only provide parts of the data, e.g. only copy a port of `GameData`, implement only those interfaces which they can satisfy and are then eligable for any algorithm or component which only needs the properties provided.

This would allow a very modular design as a data-store class implements all the interfaces it can provide and is then suitable to be passed to any function which only requires those properties. (This is a common theme in language in the sphere of functional programming)

The drawback of your proposed solution (with new interface and object for a specific use case) is that every algorithm and component now defines its own (local) interface with the mix of properties it needs (and potentially a local class implementing that interface). For the player-choosing-algorithm discussed here, this would be CurrentPlayer+Relationship, for another part it would be Relationship+Territory, a third component wants History+Territory, and yet another wants History+Territory+Relationship.
That alone would make 4 interfaces and 4 classes just to satisfy those cases and they are all not composable.  Applying that to the current situation with GameData having all of it, this would mean that `GameData` needs to implements 4 new interfaces just to "satisfy" those needs (alternatively we create 4 wrapper classes around GameDate which implement those interfaces but the argument stays the same).

I'm not sure that this is the way we want to go. It might be better and start a complete redesign of `GameData` from scratch and split it up into distinct concepts: History, WorldMap, Relations, ... Then define for each concept an interface (and possible simple mocks for testing) and use those interfaces throughout the source code. Obviously, this would be a huge refactor and probably needs lots of rewrites of existing code as `GameData` as crept into lot of places. Not to say, this would be out of scope for this PR.

To summarise, I think Interfaces to capture the parameters are better than value-objects.  And I think instead of partially and one by one replace GameData with such interfaces, it would be better to make a clean cut: invest some weeks/months of work to define proper interfaces and then rewrite the code base to fit the new model.  I think I remember you said that that would be on the agenda for after the next release?

So, to come back to this PR.  If you like I can implement the `OpponentSelectionParameters` interface and rewrite the algorithm to do its work against it, however, as said above I don't think this is viable a long term solution and should only be seen as intermezzo.

DanVanAtta:
>  Applying that to the current situation with GameData having all of it, this would mean that GameData needs to implements 4 new interfaces just to "satisfy" those needs (alternatively we create 4 wrapper classes around GameDate which implement those interfaces but the argument stays the same).


I was thinking the relationship between game data and the wrapper class would be reversed. EG:

```

class PlayerRelationships {
  enum type {
     NEUTRAL, ALLIED, AT_WAR
  }

  private Map<String, type> atWar = ...
  
  PlayerRelationships(GameData data) {
     atWar = data.getRelationshipTracker()....
  }  
}
```

There is a lot left out of that example, but wanted to show the new data class would be derived from game data. If we find some specific cases that are common then we could attach the new data class to game-data as a transient field (or do so when we no longer need to worry about serialization).

> I'm not sure that this is the way we want to go. It might be better and start a complete redesign of GameData from scratch and split it up into distinct concepts: History, WorldMap, Relations, ... Then define for each concept an interface (and possible simple mocks for testing) and use those interfaces throughout the source code. Obviously, this would be a huge refactor and probably needs lots of rewrites of existing code as GameData as crept into lot of places. 

That is roughly for the idea for 3.0, to go there.


-------------


With all that being put to the side. It sounds like the ball is with me on this PR to finish the review.

emmanuel-perlow:
> With all that being put to the side. It sounds like the ball is with me on this PR to finish the review.

Not quiet.  We discussed the change in the logic for the current player to be always the attacker.  I've implemented that now.
I also added a test for the new logic. *However* these fail, because I couldn't figure out how to initialise the History, so there is a current player.  Any idea how to do that?

DanVanAtta:
@emmanuel-perlow it's a bit difficult to explain, so I submitted a PR into this branch that could avoid the problem:
https://github.com/emmanuel-perlow/triplea/pull/1

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
```diff
@@ -1131,12 +1060,27 @@
...
+      setDefendingUnits(null);
+      setAttackingUnits(null);
+    } else {
+      landBattleCheckBox.setSelected(!location.isWater());
+      setAttackingUnits(
+          location.getUnitCollection().getMatches(Matches.unitIsOwnedBy(getAttacker())));
+      setDefendingUnits(
+          location
+              .getUnitCollection()
+              .getMatches(Matches.alliedUnit(getDefender(), data.getRelationshipTracker())));
```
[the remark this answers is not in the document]
DanVanAtta: Thanks, the logic certainly seems more consistent now. IMO the method is now a candidate for being inlined, something ti consider.

DanVanAtta:
Re: failing test

There being a difference between an inline variable and not is odd. It does have an impact on which method is called though. There are three method overloads:

```
public static <T> org.hamcrest.Matcher<java.lang.Iterable<? extends T>> containsInAnyOrder(org.hamcrest.Matcher<? super T>... itemMatchers)
public static <T> org.hamcrest.Matcher<java.lang.Iterable<? extends T>> containsInAnyOrder(T... items)
public static <T> org.hamcrest.Matcher<java.lang.Iterable<? extends T>> containsInAnyOrder(java.util.Collection<org.hamcrest.Matcher<? super T>> itemMatchers)
```

Currently with an explicit variable, the second method is being called (`T... items`). Therefore, the `List` being passed is matching to `T... items`. I suspect that is because of 'type erasure' and Java generic types are erased after initial compilation. That means the compiler is actually seeing:

`List<Object> expectedUnits = ...`

That would then match the `T... items` method overload..

OTOH, when the variable is inline then the compiler cannot be sure what the type is at compilation. As a guess, I think that makes the compiler then evaluate at runtime the exact type it sees and which method should be invoked. So at runtime the JVM sees we have a collection of matchers, and uses that method overload.



Either of these two alternatives would work as expected and avoid this problem:

```
    final Matcher<Unit>[] expectedUnits =
        japan.getUnitCollection().getMatches(Matches.unitIsOwnedBy(japanese)).stream()
            .map(GameDataTestUtil.IsEquivalentUnit::equivalentTo)
            .toArray(BaseMatcher[]::new);


    final Unit[] expectedUnits = japan.getUnitCollection().toArray(Unit[]::new);
```

Comparing exact `Unit` objects is done against UUID. It's not ideal, but I don't necessarily see a major problem with that assertion for now.

DanVanAtta:
I found two problems testing on World at War.


(1) Using 'ctrl+b' for a sea territory with German navy, the neutral player was selected as an opponent. I'd expect the neutral player to be the least preferred selection and would have expected one of the allied players to have been chosen (preferably Russia as they are the next to move). Thinking about the priority for choosing opponent, I would somewhat expect this to be the order:

1. In turn order, the next player that is at war with current player
2. If none for (1), the next player (in turn order) that is not allied with the current player
3. If none (2), the next player (in turn order) that is not the neutral player
4. Fall back to neutral player, presumably in this case we only have the current player and the neutral player.


WDYT?

Here is a screenshot showing the neutral player opponent:

Screenshot from 2021-10-30 10-02-20


(2) 'ctrl+b' for any land battle brings up sea units:

Screenshot from 2021-10-30 10-03-13


I recommend we focus on getting the logic right first. Once that is good I could help here by submitting a PR into this branch to deal with the last set of style and coding standard issues.

I think the next goal is fixing the system design so that the tests which were written can pick up either problem (1) or (2). I think that could be done if the `setupAttackerAndDefender` method were inlined and then all of the unit selection logic from the `BattlePanel` (UI class) were moved into `OpponentSelector`. This would also fix a SRP (single responsibility principle) code-smell, currently the unit selection logic is split between the selector and the battle panel rather than being all in the selector. Once the logic is fully contained in the selector, presumably the test cases could be updated to detect either problem (1) or (2). With those tests in place, it would then only be a matter of fixing up the OpponentSelector logic and then validating with some manual testing that it's all good. WDYT?

emmanuel-perlow:
> (1) Using 'ctrl+b' for a sea territory with German navy, the neutral player was selected as an opponent. I'd expect the neutral player to be the least preferred selection and would have expected one of the Allies to have been chosen (preferably Russia as they are the next to move). Thinking about the priority for choosing opponent, I would somewhat expect this to be the order:
> 
> *   In turn order, the next player that is at war with current player
> *   If none for (1), the next player (in turn order) that is not allied with the current player
> *   If none (2), the next player (in turn order) that is not the neutral player
> *   Fall back to neutral player, presumably in this case we only have the current player and the neutral player.
> 
> WDYT?

No bug in the algo here. It works as intended.  What happened?
The algo preferences players with
- have units on the territory
- own the territory

The sea is owned by "Neutral", thus it gets the preferred treatment.
(There is even a test testing for that property).

Digging deeper, the problem was that territory owner was tested against `null` (like the old algo did), a while ago this was changed to be `NULL_PLAYERID` but was not updated there, so I copied the old code.
This is now fixed.


---


> [...] if the setupAttackerAndDefender method were inlined [...]

I personal think this is bad style.  The constructor is long enough as it is and already mixes UI setup with state setup. Not inlining is IMHO a good idea.

> This would also fix a SRP (single responsibility principle) code-smell, currently the unit selection logic is split between the selector and the battle panel rather than being all in the selector.

I have the feeling you didn't look at my code. The Opponent selector already chooses the units and the only thing the BattleCalculatorPanel does is `setAttacker(OpponentSelector.getAttacker())` (and similar for Defender and units). Otherwise I have no clue what you are talking about, could you please be more specific.

emmanuel-perlow:
> Re: failing test [...some analysis...]
> Either of these two alternatives would work as expected and avoid this problem: [...a solution...]

Thanks for finding a solution. I still find it strange that variable substitution makes a difference but at least now the tests work. :)

DanVanAtta:
> I personal think this is bad style. The constructor is long enough as it is and already mixes UI setup with state setup. 

To some extent I agee. Though.. The extracted code is a void method call. Reading the constructor, you see the method and then wonder what side effects the method call would have, which ordering dependencies might exist, you then have to interrupt the flow of reading to look at that method and then return back to the constructor. If the method were functional it would serve to lift the abstraction, as is, it is more scattering than it is information hiding.


Re: neutral unit for sea battle

Considering the neutral unit as the owner of a sea territory is arbitrary. The 'neutral' player is in some ways like any other player. More to the point, it's not useful to have the neutral player pre-selected for sea battles. 2.5 will for this scenario choose 'russian' and 'german', which is a very useful pre-selection.

DanVanAtta:
>  The Opponent selector already chooses the units and the only thing the BattleCalculatorPanel does is setAttacker(OpponentSelector.getAttacker()) (and similar for Defender and units). O

`setAttacker` is the point. When it comes to choosing units, there is seemingly more logic in `setAttacker` than there is in the `OpponentSelector`

```
BattleCalculatorPanel.java:1270

  private void setAttackingUnits(final List<Unit> initialUnits) {
    final List<Unit> units = Optional.ofNullable(initialUnits).orElseGet(List::of);
    attackingUnitsPanel.init(
        getAttacker(),
        CollectionUtils.getMatches(
            units,
            Matches.unitCanBeInBattle(
                true, isLand(), 1, hasMaxRounds(isLand(), data), false, List.of())),
        isLand());
  }
```

I was stating that the above would ideally be lifted into the opponent selector.


There is more unit selection logic hanging out in the dialog class as well:

```
BattleCalculatorDialog:134
      if (gamePlayer != null) {
        currentPanel.setAttacker(gamePlayer);
      }
    }
    currentPanel.addAttackingUnits(
        t.getUnitCollection().getMatches(Matches.unitIsOwnedBy(currentPanel.getAttacker())));
```

---------------------------

In this case I recommend inlining a lot of stuff mostly so it can be split back up in a more sensible way. Step one in procedural code repair is often to inline all the void methods. Then re-write the methods so that state is only kept at a high level and the helper methods are functional. Breaking methods up by adding void methods is of questionable value, from a complexity point of the view a reader needs to read them to know the side effects, in which case they are not hiding information and in some ways would be better off being inlined even so a reader does not have to jump around methods.

EG:
Bad:
```
constructor() {
  initFirst();
  initSecond();
  initLast();
}
``` 

Better:
```
constructor() {
  attackerPlayer = computeFirst();
  defenderPlayer = computeSecond(attackerPlayer);
  territory = computeLast(attackerPlayer, defenderPlayer);
}
```

DanVanAtta:
@emmanuel-perlow , why didn't the unit tests pick up that sea units were being selected for land battle calcs? What I was driving at was, because the tests did not find that, there is more logic that is not contained in the opponent selector that is modifying the unit selection. Hence, the tests are not as useful as they could be as they do not guarantee we'll have the right unit selection. Further, this then implies that the unit selection logic is in multiple places because the test is not covering it all.

emmanuel-perlow:
> Considering the neutral unit as the owner of a sea territory is arbitrary. The 'neutral' player is in some ways like any other player. More to the point, it's not useful to have the neutral player pre-selected for sea battles. 2.5 will for this scenario choose 'russian' and 'german', which is a very useful pre-selection.

I think we misunderstand each other. The player (in the new version) which is excluded from the territory owner check is the `NULL_PLAYERID`, which by looking at the `Territory` class is its way of saying "i do not belong to anybody" (Please correct me if I'm wrong here). No player is excluded when it comes to selecting a player based on units on a territory. In fact there are maps/scenarios which have neutral ships on sea zones. Thus, that player should be selected for a battle on that sea territory.

I believe the issue you are chasing is already fixed. Maybe test the (new) code and see if your complain is still valid. I tested it with World At War at an empty sea zone and got Russian as opponent.

-----

> `setAttacker` is the point. When it comes to choosing units, there is seemingly more logic in `setAttacker` than there is in the `OpponentSelector` [...]

I think you're mixing unrelated things. The methods you mention (`addAttacker()`, etc) are needed for the key bindings "Add units of territory to attacker/defender" (CTRL-A, CTRL-D), in particular they are unrelated to choosing the initial state of the player logic. This PR, however, is about rewriting the *initial* player choosing logic, it was later expanded to also including the initial units. Now you want to rewrite the behaviour/reaction/user-interaction logic of the battle calculator. This is out of scope of this PR. I concur, however, that these could also see some new paint (as many things do).

-----

> @emmanuel-perlow , why didn't the unit tests pick up that sea units were being selected for land battle calcs?

I guess because the `OpponentSelector` is for giving the initial selection of opponents and units (which the tests check) and not for selecting the checkbox "land battle" (that's obviously UI stuff). The latter triggers a reset of the selected units, which is why I had to switch the order in https://github.com/triplea-game/triplea/pull/9602/commits/d7d085bd8cab169f60560bfe0bfc6cb2061b4f18.

DanVanAtta:
> I guess because the OpponentSelector is for giving the initial selection of opponents and units (which the tests check) and not for selecting the checkbox "land battle" (that's obviously UI stuff). The latter triggers a reset of the selected units, which is why I had to switch the order in d7d085b.

Thank you for this answer. My point was that the interesting test here is one that would show the dependency on the checkbox being selected.  A concern with this overall situation is the number of side effects and places where units and players are selected. IE: in the construction we do some work, we then invoke the opponent selector for more calculations, then we ignore most of that if the checkbox is checked which then fires a trigger to update the units.

To unwind that, perhaps the opponent selector should be returning whether the battle is land or sea, and the set of units that should be displayed as well as their own. After the constructor sets that initial state, it then would add the checkbox listener that would then update units (possibly invoking 'opponent selector' to switch units).

Does it make sense why it seems to me that we're in somewhat of a half state now? Things were a complete mess before hand, I'm not discounting that, it's more that we haven't quite finished the job of extracting the unit/opponent selection entirely and it's still split up. Do you agree, and if so do you have any thoughts if we should finish that?

emmanuel-perlow:
> To unwind that, perhaps the opponent selector should be returning whether the battle is land or sea, and the set of units that should be displayed as well as their own. After the constructor sets that initial state, it then would add the checkbox listener that would then update units (possibly invoking 'opponent selector' to switch units).

I don't think it helps much letting the `OpponentSelector` return the land-or-sea flag as the Battle Calculator UI still has to set it on its own and there we still need to preserve the order due to the listeners.
What you are referring here to, is IMHO to (start to) refactor how the Battle Calculator as a whole works. How does the UI interact with its state, how does it get setup, what part is responsible for the "interactive behaviour", ...

> Does it make sense why it seems to me that we're in somewhat of a half state now?
> Things were a complete mess before hand, I'm not discounting that, it's more that we haven't quite finished the job of extracting the unit/opponent selection entirely and it's still split up.

The question is half state of what?
Half state of refactoring the initial opponent choosing algo? No, I think we are finished there.
Half state of refactoring the Battle Calc? No, there we have just started.
Also I think your handling of coming up with one small point after another (which are related but not part of the initial goal) does not serve the problem well and wastes our both time and frankly nearly two months in it starts to become an nuisance to me (maybe also because my free time became scarce lately).

IMHO it is better to take a wide and long look on how the different battle calc classes are currently structured, which part does what, how they interact with each other, and then figure out a design. E.g. What you might not have noticed yet is that `PlayerUnitsPanel` also tampers with the units you see. Earlier you pointed out the two `addAttacker()` calls, those could probably be combined, maybe even split out the complete internal state of the battle calc and its logic into its own class (think MVC)? Maybe `OpponentSelector` shouldn't be there but be part of the internal state logic? I don't know I haven't look en detail on what is currently done where and which listeners are called when, nor did I think about how to combine them, and I have the feeling neither did you.

I don't like ending this on a sad tone. I liked you well-founded feedback and help. Thank you for that. I'm happy to fix any remaining issue *regarding* the topic of this PR, namely – and I quote the title of this PR – rewriting the initial opponents choosing logic; if you have some, please point them out and I'll have a look. In a previous comment you said you wanted to file a PR fixing some minor style issues, happy to incorporate them.

Maybe I take another (greater) shot at this sometime in the future but not now.

DanVanAtta:
There's a lot to say in response, I'll try to keep it short. First, this is good work in the PR and thank you. I can take over the remaining efforts here on this branch, so we are not at all ending on a sad note.

Ultimately the contributor onboarding experience I think needs more investment. Clearer expectations and more automation perhaps could have made this a smoother process.

PR #9640, opened 2021-09-18
Added World War I 1914

RogerCooper:

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2355,3 +2355,12 @@
   img: https://raw.githubusercontent.com/triplea-maps/Cold_War_1950_4th_edition/master/preview.png
   description: |
     <br>A hypothetical World War 3 starting in 1950 using the Pact of Steel map
+- mapName: World War I 1914
+  mapCategory: EXPERIMENTAL
+  url: https://github.com/triplea-maps/world_war_i_1914/archive/master.zip
+  version: 1
+  img: https://raw.githubusercontent.com/triplea-maps/world_war_i_1914/master/preview.png
+  description: |
+    <br>An adaption of the Axis & Allies: 1914 game. Many rules of the boardgame are not implemented.
+    
+    


DanVanAtta:
@RogerCooper , I'm sorry to say I don't know if we can have this map in an 'official' TripleA map repository. It's unclear exactly what the rules are that we need to follow. In that respect, until we get better guidance, any new maps need to be 100% original creations.

RogerCooper:
We already have maps that are exact duplicates of games published by Avalon-Hill. Why is this approximate version a problem?

Some years ago, there was a letter from Hasbro. Rebranding the game as TripleA was sufficient for Hasbro. 

There are many versions of A&A on the internet. The popular Tabletop Simulator program has Axis & Allies mods and Hasbro has expressed no objection. Most boardgame publishers view fan mods as advertising not competition.

In any case, the law requires has Hasbro to actively request the removal of the mods. There is no liability unless such a request is ignored. It is also not clear whether there is a copyright violation, see this article from the American Bar Association [ABA Journal Mar/Apr 2015](https://www.americanbar.org/groups/intellectual_property_law/publications/landslide/2014-15/march-april/not-playing-around-board-games-intellectual-property-law/)

DanVanAtta:
The existing maps that are exact duplicates have an uncertain future. We are actively debating what to do with them.

Do you know if you can find a copy of that letter anywhere? My recollection of the DMCA takedown notice they sent some years ago around 2011 was more than rebranding. EG: IPC renamed to PU, armour renamed, and other specific names that they claimed as part of the art & form of their game (beyond mechanics). TripleA also stopped distributing all maps from an official repository and instead required maps to come from community hosting.

If Hasbro were to request the removal of mods, they are likely to send a DMCA takedown notice to github to remove the game engine and make no distinction between mods and the core engine. We'll then at that point be taken completely offline (as well as personal liability implications even).

We do have a possible avenue to take for fan content exemption: https://company.wizards.com/en/legal/fancontentpolicy

Regrettably I think this section of the ABA goes against your point though @RogerCooper :
> Therefore, the systems or processes that make up the core of a game—generally referred to as the “game mechanics”—are not subject to copyright, even though the written rules, game board, card artwork, and other elements—often referred to as the “theme” of the game—may be. 

@RogerCooper , I'm highly interested in that letter if you're able to find it. Thinking back to that, I think the letter they sent was just a DMCA take down and nothing further. TripleA responded by unbundling all maps and additional changes and Hasbro never followed up. That is my understanding, if there is such a letter though that we just need to rebrand, it's a game changer. I wonder if the 'rebranding' was a condition on top of unbundling all maps.

RogerCooper:
I don't recall anything more than you do about the DMCA takedown notice. 

Game mechanics are not covered by copyright. The maps are, but our maps are not generally identical to the printed mapboard. 

In any event, Hasbro seems to be ignoring this. There is good A&A mod for Tabletop Simulator.

This discussion should be moved to the main forums,

DanVanAtta:
@RogerCooper , I think we'll need to keep this map in our back pocket for the time being.

The tabletop simulator example is notable, but not clear cut they are in the clear: https://boardgamegeek.com/thread/1754341/tabletop-simulator-and-copyright  (do you happen to know if the A&A mod has explicit permission?) I was able to find what looked like their official page on steam (https://steamcommunity.com/sharedfiles/filedetails/?id=446693914), though no mention about licensing.

Agree that we need to have this overall convo in the main forums, and quite a bit more action beyond as well.

RogerCooper:
There about 20 different A&A mods on TTS. Hasbro has clearly decided not to take any action.

Back in 2011, I considered reformatting the A&A maps to put them on real world geography. That may provide an alternative solution.

DanVanAtta:
> Back in 2011, I considered reformatting the A&A maps to put them on real world geography. That may provide an alternative solution.

That sounds really intriguing.

Sorry to close this PR for now, we need more clarity on the A&A maps issue until we can accept potentially new ones.

PR #9665, opened 2021-09-27
Add additional lobby URI settings  [introduce: UI event queue]

DanVanAtta:
Functional change: update the lobby selection preference to include a new radio button for the 'prerelease' lobby. 

Most important a new reference UI event queue is implemented. The 'UI event queue' is meant to be a reactive style pub/sub for view-model events between a view and its controller. This is meant to provide the infrastructure code for future [view-model-view-model](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93viewmodel) implementations. 

Looking for feedback on the view-model event queue and how that looks. 

The event queue has a few interesting design decisions behind it. Rather than being a true pub-sub model the events are bound to their handlers. This is to ensure that we do not have unhandled events.

Good things about the event queue:
- UI event handling is automatically dispatched to a new thread (off of the Swing EDT thread)
- Model event handling is automatically dispatched to the Swing EDT thread
- Communication via events seemingly lead to a pretty clean implementation overall.

Bad things about the event queue:
- involves more classes than I would like. Any UI component with a backing mode and view-model would require 5 class files.
- lot of marker interfaces, would be nice if there could be fewer of these (perhaps annotations can help wire things up automatically for us?)

TODO:
- [x] build in test support to make controller logic testing easy
- [x] feedback regarding the new queue reference implementation
- [x] squash commits
- [x] thread affinity
- [x] build checks


ssoloff:
@DanVanAtta I haven't been in the weeds of this stuff for several years, so I doubt I'll be able to provide any useful feedback on the design. However, I'll take a look over the next 24-48 hours and drop a review if anything jumps out at me.

lib/swing-lib/src/main/java/org/triplea/ui/events/queue/ViewControllerEventQueue.java
```diff
@@ -0,0 +1,53 @@
...
+ * Adapted for Swing!
+ *
+ * <p>Queue to pass events between a Swing UI and a view-model (AKA presentation layer).
+ *
+ * <p>UIs publish UI events and Model publish model events. UI listen for model events and models
+ * listen for UI events. The model event callbacks (UI) are executed on the Swing EDT thread, and UI
+ * event callbacks are executed in a new thread (off of the EDT).
+ */
+@Setter
+public class ViewControllerEventQueue<
```
ssoloff, line 17: Maybe I misunderstood the purpose of this class, but I expected to find `publishUiEvent` invoking the callback on a new thread and `publishModelEvent` invoking the callback on the EDT; as it stands, they both invoke their respective callbacks on the current thread. Is this simply an oversight, or is the thread affinity code someplace else and I just missed it?

ssoloff (commented):
Looks fine to me. I agree the number of types per UI component is a bit busy, but I don't really have any ideas for reducing that at the moment.

lib/swing-lib/src/main/java/org/triplea/ui/events/queue/ViewControllerEventQueue.java
DanVanAtta: You're right, and following along nicely   = D
The thread affinity code was built in a first revision of this class and this class is a rewrite and simply hasn't gotten back to having the affinity code. It turns out it is not that difficult, and then blissful to know in the handlers that everything is on the right thread.

I'll comment and point out when that logic is added back.

game-app/game-core/src/main/java/games/strategy/triplea/settings/lobby/LobbySelectionView.java
```diff
@@ -0,0 +1,145 @@
...
+            .boxLayoutVertical()
+            .add(new JPanelBuilder().addAll(new ArrayList<>(radioButtons)).build())
+            .add(uriField)
+            .build();
+    updateUiToState(new LobbySelectionViewData());
+  }
+
+  @SuppressWarnings("SwitchStatementWithTooFewBranches")
+  @ControllerListener
+  public void updateUiToState(
```
DanVanAtta, line 76: Note the usage of @ControllerListener to automatically have the event queue invoke this method.

DanVanAtta:
@ssoloff thanks for taking a crack at this!
I added a commit that re-does how events are bound to their handlers. Notably an annotation was added to view and controller with args {eventType, data}. I was hoping to design it so that specific methods would be called when specific events happen, but have not figured out a good way to do that.

Previously the handlers were bound to events in the event enum. Forgetting to add event handlers is an easy bug, and so I really want it so that it is nearly required to add handling logic when an event is added. As a compromise that binding logic was moved out of the event enum and to an annotated method that then has a switch statement. The hope is that the switch statement would be easy to find and update.

There's still quite a bit to do on this PR.. It's *very* rough still in a good many places.

ssoloff:
> I added a commit that re-does how events are bound to their handlers. Notably an annotation was added to view and controller with args {eventType, data}.

@DanVanAtta :+1: Definitely smells better with respect to reducing type proliferation. The only thing I'd suggest is, since you're now dropping the compile-time check for view/controller type compatibility, you might want to add a runtime check in the publication methods. That is, if you don't find a suitably-annotated method on a view/controller, you should treat it as a programmer error and fail with an unchecked exception. Right now, those methods silently ignore a missing annotation. You might even consider failing fast by adding such a runtime check when the view/controller is added to the event queue rather than waiting until event publication.

lib/swing-lib/src/main/java/org/triplea/ui/events/queue/ViewControllerEventQueue.java
```diff
@@ -0,0 +1,100 @@
...
+   * UI event callbacks are executed in a new thread (off of the EDT).
+   */
+  public void publishModelEvent(
+      final ModelEventTypeT modelEvent, final UnaryOperator<ViewDataT> uiDataChange) {
+    Preconditions.checkState(!views.isEmpty());
+
+    views.forEach(
+        view -> {
+          for (final Method method : view.getClass().getMethods()) {
+            if (method.getAnnotation(ControllerListener.class) != null) {
```
RoiEXLab, line 84: I think a compile-time alternative might be to define an interface that provides a method like this:
```java
List<BiConsumer<ModelEventTypeT, UnaryOperator<? extends ViewData>>> controllerListeners() {
  return List.of(
    this::method1,
    this::method2
  );
}
```

Not as elegant and with considerable overhead, but we get the compile-time checks.
Although I might be missing something

game-app/game-core/src/main/java/games/strategy/triplea/settings/lobby/LobbySelectionView.java
```diff
@@ -0,0 +1,145 @@
...
+      final LobbySelectionEvents.ControllerEvent controllerEvent,
+      final UnaryOperator<LobbySelectionViewData> dataChangeFunction) {
+
+    switch (controllerEvent) {
+      case UPDATE_UI_DATA:
+        final var updatedState = dataChangeFunction.apply(readDataState());
+        updateUiToState(updatedState);
+        break;
+    }
+  }
```
RoiEXLab, line 86: Slightly off-topic, but I find it fascinating to see reccuring programming pattern regardless of the language being used.
This is basically reactive programming and reminds me of my job where currently most of my time goes into JavaScript programming using react, and this is basically how data is handled there: You define how data should be rendered and then you only update this data source using different approaches (the react-redux framework does the exact same, by "publishing"/"dispatching" global events to modify the data store, whereas the default react framework prefers a more direct approach where there isn't as much abstraction).

DanVanAtta:
Last update managed to get rid of the reflection by using a view-class and view-class-controller interface. Managed to remove the 'events' class by folding the events enums into the respective view and view-controller.

I would like to continue removing more classes... The factory class could be removed. Checkstyle is getting in the way here. Factory methods (static) are jarring if they are below a constructor (particularly a lengthy constructor with helper methods). It is perhaps argument to remove the checkstyle that requires constructors to be first.

It would be nice if the binding between view/queue/view-controller were more automatic as well. I would like to really optimize for the developer-experience here to make this a very useful/easy pattern.

I've more work to do still on this PR:
(1) thread affinity code
(2) working on a test harness to make testing controller code easy
(3) general PR polishing

game-app/game-core/src/main/java/games/strategy/triplea/settings/lobby/LobbySelectionView.java
DanVanAtta: The design here is very much meant to be a reactive-style.

I was thinking a nice future modification could be to have an automatic progress dialog appear for events that are slow to be completed. We could have an 'acknowledge' event method to remove an event off a queue. A dialog watcher could check the for queue for items that are over 500ms old and show a dialog (or a status message).
 
Stepping back a bit, I think a reactive event-queue style with view-controller-view-model is a good combination. The before and after seem like pretty good improvements. Adding another radio button turned out to be surprisingly complex, in this updated style it seemed straight forward (switch statements and the simplified data model combine to make it pretty easy to write out the logic cases).

@RoiEXLab I don't know much about React. I'm curious about the 'more direct approach', is there an easy example of how this update could be changed to have that direct approach? Would it be a good thing?

DanVanAtta:
@ssoloff good suggestion, was thinking similar. Looking at the code with a fresh set of eyes, it turns out just having a type is the way to compile-time check that a given method is available  =D 

It removes the reflection as well, which I no longer have to worry about benchmarking and potentially caching annotated methods.

------------

I wonder, does this pattern look easy and useful enough to be worth adopting/using more broadly? If it's something that looks appealing, is there anything that could be done to help increase it's uptake?

I'm planning to rework the PBF/PBEM UIs with this updated pattern once done here. The way view-controller was done for those two UIs is problematic (too many events fired back and forth) and that'll serve as another example to help vet out this pattern.

game-app/game-core/src/main/java/games/strategy/triplea/settings/lobby/LobbySelectionView.java
RoiEXLab: I'll try my best to explain it without going too much into the details of react:
So basically almost everything you code with react is a "Component" and using so-called hooks the code looks something like this:
```js
function MyComponent() {
  return <div>
      <span>My Text here</span>
      <button>Some Button</button>
    </div>;
}
```

which would render pretty much exactly the HTML you can see there in the browser. If you want to have some dynamic data in there using react only without any additional frameworks you would do it with a so called "state" that can be modified by your code.

```js
function MyComponent() {
  const [ text, setText ] = React.useState('My Text here');
  const handleClick = event => setText('Updated Text');

  return <div>
      <span>{text}</span>
      <button onClick={handleClick}>Some Button</button>
    </div>;
}
```

So whenever `text` is modified via `setText` react will notice that and re-render the components that did change (so in this case `MyComponent`, `div` and `span` will be re-rendered IIRC, the `button` will be kept around and simply re-attached to the DOM).
And this is pretty much what I meant with a more direct approach to data storage. Basically you define that you need to store a single value (can be more complex but you need to keep in mind to keep it simple enough to not re-render unneccesarily often, just because some single property changed there). If you wanted to modify this value elsewhere, you'd have to define it in the common parent, and pass the setter to the child methods that need access, there is a way to have some sort of global storage, but it rarely makes sense to use it.

Using redux, the framework I mentioned previously the data storage becomes global and the component would looks something like this:
```js
// heavily simplified for clarity
function MyComponentContainer() {
  return <MyComponent  text={store.myComponentText} setText={text => dispatch(SET_TEXT_EVENT, text)}/>;
}

const myComponentReducer(event, currentState) {
  switch(event.action) {
    case SET_TEXT_EVENT:
      return {
        ...currentState,
        myComponentText: event.payload,
      };
  }
}

function MyComponent(props) {
  const { text, setText } = props;
  const handleClick = event => setText('Updated Text');

  return <div>
      <span>{text}</span>
      <button onClick={handleClick}>Some Button</button>
    </div>;
}
```

Which means that all instances of `MyComponent` would share the same underlying storage which can be desirable in some cases but in other cases this is not what you want.

So I don't know if a more direct approach suits our problem in java better, it would require us to implement a generic data wrapper that notifies all listeners whenever it is updated, but it will probably not reach the same level of syntactic sugar as in react.

If you have any questions or I explained something badly feel free to ask.

game-app/game-core/src/main/java/games/strategy/triplea/settings/lobby/LobbySelectionViewData.java
```diff
@@ -0,0 +1,82 @@
...
+    static LobbyChoice fromUri(final URI uri) {
+      return Arrays.stream(values())
+          .filter(lobbyChoice -> URI.create(lobbyChoice.uri.get()).equals(uri))
+          .findAny()
+          .orElse(OTHER);
+    }
+
+    static LobbyChoice fromDisplayText(final String selectedRadioButton) {
+      return Arrays.stream(values())
+          .filter(choice -> choice.getDisplayString().equals(selectedRadioButton))
```
RoiEXLab, line 52: I'm surprised the Radio Buttons don't provide a mechanism to store actual objects and just render some string represantation like you can do with comboboxes

lib/java-extras/src/main/java/org/triplea/java/exception/UnhandledSwitchCaseException.java
```diff
@@ -0,0 +1,24 @@
...
+ *  switch(someValue) {
+ *    case ...:
+ *       break;
+ *    default:
+ *       throw new UnhandledSwitchCaseException(someValue);
+ *  }
+ *   </pre>
+ * </code>
+ */
+public class UnhandledSwitchCaseException extends RuntimeException {
```
RoiEXLab, line 18: Didn't we use `AssertionError` for this in the past?

DanVanAtta: Looks like. I don't think `AssertionError` quite fits the bill for that usage, having a more direct exception seems like an improvement - if anything just so we don't have to repeatedly write the boiler plate `"Unhandled case value: " + value`.

RoiEXLab: IIRC the rationale behind it was that it essentially represented an impossible condition that if it would ever be reached would indicate an error in the jvm.
So throwing an error instead of an exception makes sense there.
However I think you could also argue that this kind of exception would indicate that the developer forgot to handle the case after altering the enum type, where a dedicated runtime exception with a good message would also make sense.
Depends on your standpoint though I think

DanVanAtta: Well, both standpoints can be true : )
`AssertionError` -> true at the time of writing
`UnhandledSwitchCaseException` -> true after anyone modifies the enum

lib/swing-lib-test-support/src/main/java/org/triplea/ui/events/queue/test/support/EventQueueAssertions.java
```diff
@@ -0,0 +1,47 @@
...
+public class EventQueueAssertions {
+
+  /**
+   * Use to assert that the most recently received controller event had a specific event type. EG:
+   * <code>
+   *   <pre>
+   * assertThat(eventQueueTestSupport, lastControllerEventTypeIs(ControllerEvents.EVENT_TYPE));
+   *   </pre>
+   * </code>
+   */
```
DanVanAtta, line 23: TODO: out of date comment

(keep in mind the test-support class has most of what would need to be here)

lib/swing-lib-test-support/src/main/java/org/triplea/ui/events/queue/test/support/EventQueueAssertions.java
```diff
@@ -0,0 +1,47 @@
...
+
+    Preconditions.checkNotNull(eventType);
+    return CustomMatcher.<ControllerEvent<ControllerEventTypeT, ViewDataT>>builder()
+        .debug(
+            controllerEvent ->
+                String.format(
+                    "%s, data change operation (input -> output):\n%s\n->\n%s",
+                    controllerEvent.getEventType(),
+                    inputData,
+                    controllerEvent.getViewDataUpdate().apply(outputData)))
```
DanVanAtta, line 38: TODO: could use some commentary to make the custom matcher easier to grok

lib/swing-lib-test-support/src/main/java/org/triplea/ui/events/queue/test/support/EventQueueTestSupport.java
```diff
@@ -0,0 +1,76 @@
...
+        .when(eventQueue)
+        .publishControllerEvent(Mockito.any(), Mockito.any());
+  }
+
+  public ControllerEvent<ControllerEventTypeT, ViewDataT> popFirstControllerEvent() {
+    Awaitility.waitAtMost(2, TimeUnit.SECONDS)
+        .pollInterval(1, TimeUnit.MILLISECONDS)
+        .pollDelay(1, TimeUnit.MICROSECONDS)
+        .until(() -> !controllerEvents.isEmpty());
+
```
DanVanAtta, line 73: TODO: decide on contract for no event, whether to return an empty or to fail. Then update this so the failure message makes no event received clear.

lib/swing-lib/src/main/java/org/triplea/ui/events/queue/ViewControllerSwingEventQueue.java
```diff
@@ -0,0 +1,84 @@
...
+    views.add(view);
+  }
+
+  @Override
+  public void addController(final ControllerClassT controller) {
+    controllers.add(controller);
+  }
+
+  @Override
+  public void publishUiEvent(final ViewEventTypeT uiEvent, final ViewDataT uiData) {
```
DanVanAtta, line 57: Thread affinity code is now in this method and the `publishControllerEvent` methods (cc: @ssoloff )

lib/swing-lib/src/main/java/org/triplea/ui/events/queue/ViewData.java
```diff
@@ -0,0 +1,23 @@
+package org.triplea.ui.events.queue;
+
+/**
+ * Marker interface for view-model value classes that represent data presented in a UI. See:
+ * https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93viewmodel
+ *
```
DanVanAtta, line 6: TODO: revisit this comment block, see if can be shortened

lib/swing-lib/src/test/java/org/triplea/ui/events/queue/ViewControllerEventQueueTest.java
```diff
@@ -0,0 +1,64 @@
...
+  final ViewClassControllerSample controllerClassSample = new ViewClassControllerSample();
+  final ViewClassSample viewClassSample = new ViewClassSample();
+
+  @BeforeEach
+  void setup() {
+    exampleEventQueue.addController(controllerClassSample);
+    exampleEventQueue.addView(viewClassSample);
+  }
+
+  @Test
```
DanVanAtta, line 28: nit, add display names or javadoc the test methods to describe the test

verify
```diff
@@ -17,4 +17,5 @@ normal="\e[0m"
   :spitfire-server:dropwizard-server:test \
 && .build/code-convention-checks/run \
 && .build/find-unused-or-unmatched-tests/run \
+&& .build/check-links-and-yaml-syntax/run \
```
DanVanAtta, line 20: nit, put in a different PR.

game-app/game-core/src/test/java/org/triplea/live/servers/FetchingCacheTest.java
```diff
@@ -50,13 +51,15 @@ void setUp() {
             .build();
   }
 
+  @Disabled
```
DanVanAtta, line 54: There is more to do.. we now get the latest version from 'ClientSettings' with a default of 'prod.triplea-game.org'

We'll want to send a header with a triplea-version. Later we can configure nginx to forward old versions to an old lobby.

lib/swing-lib/src/main/java/org/triplea/ui/events/queue/ViewControllerSwingEventQueue.java
```diff
@@ -0,0 +1,84 @@
...
+    Preconditions.checkState(!controllers.isEmpty());
+
+    final Runnable publishTask =
+        () ->
+            controllers.stream()
```
ssoloff, line 62: Note that the mutable collection `controllers` may now be accessed on multiple threads, but it's not currently a thread-safe implementation. To avoid having to make it thread-safe, you may want to simply make a copy to use within `publishTask`.

lib/swing-lib/src/main/java/org/triplea/ui/events/queue/ViewControllerSwingEventQueue.java
```diff
@@ -0,0 +1,84 @@
...
+
+  @Override
+  public void publishControllerEvent(
+      final ControllerEventTypeT controllerEvent, final UnaryOperator<ViewDataT> uiDataChange) {
+    Preconditions.checkState(!views.isEmpty());
+    // controller events we expect to always be emitted from a non-EDT thread.
+    Preconditions.checkState(!SwingUtilities.isEventDispatchThread());
+
+    SwingUtilities.invokeLater(
+        () -> views.forEach(view -> view.handleEvent(controllerEvent, uiDataChange)));
```
ssoloff, line 82: Ditto for `views`.

ssoloff (commented):
:+1:

lib/swing-lib/src/main/java/org/triplea/ui/events/queue/ViewControllerSwingEventQueue.java
DanVanAtta: Was a bit torn on this. On the one hand the `views` and `controllers` were originally constructor args and immutable. Because we have a circular relationship between queue and listener, we have the setters. So it's intended that we call the setters just at construction time. So writing the class to be full blown thread safe is odd because if it is used as intended then it does not need to be..

To resolve, I think adding a comment about this would help, and it also does not really hurt to change the `controllers` and `views` lists to be `CopyOnWriteArrayList`

game-app/game-core/src/main/java/games/strategy/triplea/settings/lobby/LobbySelectionView.java
DanVanAtta: Interesting, thanks @RoiEXLab .
It sounds like the approach taken for the PBF screen was more of the direct approach (maybe?). In that design every element had an event listener attached to it that then updated the view-model data state. Then any action clicks were a no-arg request to the view-model to execute the action.

Just for contrast, the event-queue design presented here has the view emitting an immutable view-model object with an event type all in one go and only when 'action clicks' occur. The PBF model did seem to run into performance problems and IMO the thread-hopping was not as clean.

I'm thinking that this event queue model can be extended further and for more contexts. It certainly is 'reactive' programming inspired. To that extent I was thinking the next addition could be for automatic progress dialogs. To do that a queue event would have an optional 'completed callback' attached to it. Our event queue would then check for events that are still in queue (not completed) and if still in progress after 300ms then we show a progress dialog (though we also really need to make our current progress dialog to be closeable)

PR #9670, opened 2021-09-29
Added Lord of the Rings Middle Earth - Fords

RogerCooper:


Cernelius:
A map modification called "Fords" makes me think that it adds fords to the map (rather than removing some of them).

RogerCooper:
You have a point there. I didn't make or name the map. "Fewer Fords" would also look odd.

beelee1:
@RogerCooper Do want to add it as is or rename it ?

PR #9682, opened 2021-10-05
Use current prod2-lobby as default

DanVanAtta:
prod.triplea-game.org is not fully set up.
Update lobby prod host URL to the current DNS alias
in the meantime.


DanVanAtta:
The DNS alias for prod.* is new and not configured quite right. I think we need to update nginx to have a server name lock and return a new cert for prod.*
Recall the previous lobby was at 'lobby.*'
I would like for us to use 'prod.*' indefinitely going forward and use nginx to redirect versions based on header values.

PR #9705, opened 2021-10-13
Added Great War - The Last Stand, March 1918

RogerCooper:


beelee1:
@RogerCooper right on so the repo is good Idk if you need to redo the PR as it's still showing the 1 unsuccessful. 

You might double check the yaml part as I think that's where the error is. i don't see anything unless maybe the dashes "-" in the name ?

PR #9731, opened 2021-10-21
Added 2020

RogerCooper:


DanVanAtta:
FWIW, if you have a map name that is just a number, it needs to be surrounded with double quotes.

DanVanAtta:
Did you want this merged @RogerCooper ? I added a commit that fixed the quoting issue to pass the build, it seems good to go now.

PR #9755, opened 2021-11-08
Update checkstyle to latest version (8.45) & Prune Validations

DanVanAtta:
First, we get a brand new checkstyle.xml file from google-java.
This new file is compatible with the latest checkstyle 8.43.

Then rules were removed to make the current code pass checkstyle.
Further, more (whitespace) formatting rules were removed, we will rely on
'spotlessCheck' to verify formatting.

Next, the suppression file seemingly has no effect. It was removed.
The few exceptions we cannot avoid were simply added to the
'MemberName' regex white list.

Resolves: #7788

-----------

## Additional Notes

We added a custom checkstyle for 'variable may be final'. It is a significant and intentional update to not add it back in.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -64,7 +64,7 @@ subprojects {
         awaitilityVersion = '4.1.1'
         bcryptVersion = '0.9.0'
         caffeineVersion = '3.0.4'
-        checkstyleVersion = '8.23'
+        checkstyleVersion = '8.45'
         commonsCliVersion = '1.5.0'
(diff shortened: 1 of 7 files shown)


DanVanAtta:
FYI, checkstyle version can now be set to the latest

cc: @trevan , @emmanuel-perlow , @RoiEXLab 

There is also a major style change to no longer require optional 'final' modifier.

RoiEXLab:
Thank you for tackling this issue

PR #9860, opened 2021-12-05
Remove the auto-generated game title on the 'select game' screen

DanVanAtta:
Most games have the title built into the game notes. Displaying the game
name and path in a different style compared to the rest of the game notes
does not look good and is almost always redundant (notably the game
notes very typically include the name of the game).

### Before
Screenshot from 2021-12-04 18-42-15

### After
Screenshot from 2021-12-04 18-39-48

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/GameData.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/GameData.java
@@ -546,7 +546,7 @@ public class GameData implements Serializable, GameState {
     // From the game-xml file name, we can find the game-notes file.
     return findMapDescriptionYaml()
         .flatMap(yaml -> yaml.getGameXmlPathByGameName(getGameName()))
-        .map(gameFilePath -> GameNotes.loadGameNotes(gameFilePath, getGameName()))
+        .map(GameNotes::loadGameNotes)
         .orElse("");
   }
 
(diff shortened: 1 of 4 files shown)


Cernelius:
It is helpful for map-makers (and power-users doing manual installations) to be sure where the game file is, especially since the problem of multiple same-named games has never been fixed (as far as I recall).

I suggest having a toggle in engine preferences for displaying information which is useless to casual users (default false).

DanVanAtta:
> It is helpful for map-makers (and power-users doing manual installations) to be sure where the game file is, especially since the problem of multiple same-named games has never been fixed (as far as I recall).

Indeed. Though the number of people and times where this is helpful can be counted on one hand. This compared to thousands of other users and the select game screen looking bad for all of them, all the time.. That makes this an easy decision.

The problem of same-named game is being left as a data problem. There is no really good automatic solution to this. It is up to the map makers and map admins to make sure that map names are given unique names. 

> I suggest having a toggle in engine preferences for displaying information which is useless to casual users (default false).

I appreciate the suggestion. I don't think a toggle would be that easy to find and would be so rarely used as not to be worth it. I think we need other solutions here, perhaps more process based. Nothing can stop a map maker from creating duplicate map names locally. To prevent that the game engine could refuse to load the maps, though that is harder to troubleshoot and less obvious of a problem. A map maker that is creating names and has a duplicate name is in a pretty good position to know what happened and how to fix it.

PR #9882, opened 2021-12-11
Simplify wording of 'update maps' prompt

DanVanAtta:


Cernelius:
I think saying "the following maps have updates available" is rather vague and it sounds like each of the maps has one or more possibly optional updates of some kind available.
I would rather say: `Amongst the maps you already have, the following ones are out-dated: for each of them, a more and most recent version can be downloaded in substitution of the one you have. However, be warned that compatibility with save-games taken with older versions is not assured (so downloading any one of these maps may make you irreversibly unable properly to continue saved games which you already have of the map).`.

DanVanAtta:
- "Updates" vs "version" terminology I think may depend on perspective. Strictly speaking, there is no 'version'. If one considers that one or more files have changed, then 'updates' makes sense.

- Is it truly unclear that 'update' will replace the existing map?

I'm very reluctant to add a very scary warning message about save game compatibility. Map makers in theory should be taking care to submit compatible patches, and most likely bigger changes likely are going to be a fork of a map instead of a major revision.

Though, it would be pretty sweet if the engine could scan existing save games for compatibility issues before installing the map fully. It's hard to say how much that would be worth, depends on how often save-games go because of map updates (my understanding is that by convention we avoid that and it is rare)

Cernelius:
Since I understand the program's lexicon is a bit in flux, the first line may be reworded without using "up-date", "out-date" and "version" as terms, also changing the "can" to a more encouraging "are ready to", as "`Amongst the maps you already have, each of the following ones has been modified after you downloaded it and is ready to be downloaded in substitution of the one you have`".

---

I think "update" tends to sound like it is compatible.

Cernelius:
> Though, it would be pretty sweet if the engine could scan existing save games for compatibility issues before installing the map fully. It's hard to say how much that would be worth, depends on how often save-games go because of map updates (my understanding is that by convention we avoid that and it is rare)

I would rather have it done when the map is up-dated and only with respect to the map (the repository) immediately before the update. This may trigger a first digit unitary up-grade (and a second digit zeroing) if the map is seen as potentially incompatible and a second digit unitary up-grade otherwise. Of course, this would imply having a map version somewhere. Obviously, the mapmakers could handle the project themselves, but I think in practice it would fail to work consistently, or at all, oftentimes.

Let's also not forget that the game running doesn't imply that the game is well playable. A simple example may be that I redraw the (original skin of the) map and add and remove connections only: if, then, someone updates and load an old save-game, the game will run but the visual will be inconsistent.

I doubt any of the above is really worth the time though.

Cernelius:
> Since I understand the program's lexicon is a bit in flux, the first line may be reworded without using "up-date", "out-date" and "version" as terms, also changing the "can" to a more encouraging "is ready to", as "`Amongst the maps you already have, each of the following ones has been modified after you downloaded it and is ready to be downloaded in substitution of the one you have`".

We can also use "state" instead of "version": "`Amongst the maps you already have, each of the following ones has been modified after you downloaded it and is ready to be downloaded at its most recent state in substitution of the one you have`".

DanVanAtta:
> Amongst the maps you already have

I don't like this phrase:
- it seems very wordy
- I think it is already implied. If the map is available for update, it means it exists (otherwise it would be new). Plus, the player would have already explicitly downloaded the map, they should have some recollection that they have installed the map already.

> each of the following ones has been modified after you downloaded it 

This is a more accurate way to describe the situation. But would it not be just as well to say "There are updates available for these maps?". The phrase you have given seems too exact, a person would have to think about it for a solid moment to understand (and that is assuming that the user even reads the text, most computer users read an extremely small fraction of any text presented to them).  Hence, having more than a sentence, while more accurate, is more likely to lead to none of it being read at all. This is why I'm concerned about wordy phrases.


> at its most recent state in substitution of the one you have".

I wonder if we are leaking implementation details. You and I know that the map update is a full replace, but it so happens that usually most content that is replaced will be replaced with the same content. From the perspective of a user, all that they downloaded was perhaps one updated image, even though they actually downloaded and installed the entire map from scratch. Though the emphasis here is on the appearance of how this worked rather than how it actually worked. From a user perspective the map received updates, they do not realize it was fully replaced. 

So I'm debating a bit whether we should mention that the existing map will be replaced, the idea of an update is that the existing will be changed already, so I don't think we need to expose users to this level of detail of how it actual works.

DanVanAtta:
@Cernelius I'd like to move forward for now. Thank you for the discussion, while I have not taken any the recommendations  now, the topics raised deserved some thought. Please note that you can suggest edits to the code from the github web UI, so you can follow up on this if you would like with a suggested text change (to this location or others). There are plenty of texts that are overly wordy, "walls of text", that I strongly suspect nobody is reading and is just confusing. It is a goal to reduce the amount of text that TripleA, too much is said in 3 sentences that could instead be said in 3 words.

PR #9883, opened 2021-12-12
Update triplea_maps.yaml

Navalland2:


Navalland2:
Update triplea_maps.yaml

PR #9938, opened 2021-12-25
Fixes to errorprone errors (incl. code smells)

frigoref:
The other commits seem to be some version problem, but the final file changes are matching the changes made.

- UnnecessaryLambda (also turn off completely in game-core/build.gradle)
- SameNameButDifferent (HttpClient.java)

Fix in PropertyFile.java with Duration for
- JavaTimeDefaultTimeZone

CodeSmell fixes
- ClientSetupPanel.java: empty method without explanation
- EditPanel.java:
- HttpClient.java: %n produces platform-specific line separator instead of \n
- SaveGameFileChooser.java: replaceAll is the same as replace when no regex is used, but consumes more performance
- ServerModel.java:


frigoref:
Errors need fixing first ...

PR #9939, opened 2021-12-25
Fixing some gradle.build warnings on Windows

frigoref:
The other commits seem to be some version problem, but the final file changes are matching the changes made.

UnnecessaryLambda (also turn off completely in game-core/build.gradle)
SameNameButDifferent (HttpClient.java)
Fix in PropertyFile.java with Duration for

JavaTimeDefaultTimeZone
CodeSmell fixes

ClientSetupPanel.java: empty method without explanation
EditPanel.java:
HttpClient.java: %n produces platform-specific line separator instead of \n
SaveGameFileChooser.java: replaceAll is the same as replace when no regex is used, but consumes more performance
ServerModel.java:


game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
```diff
@@ -223,6 +226,7 @@ private void selectUnitsToRemove(
         }
       };
 
+  @SuppressWarnings("UnnecessaryLambda")
```
DanVanAtta, line 229: There are a number of cases where we really want a lambda. Notably when we want to inject or change behavior from a test context and update a lambda variable to do that (grant-it, that is still a code-smell, but nonetheless can be an important tactic for this legacy code base).

With that said, I think this error prone might be accurate. Generally the 'unnecessary lambda' error prone suggests that a lambda be replaced by a method. EG:

```
  private final MouseOverUnitListener mouseOverUnitListener = this::setMouseOverUnitListener;

  private void setMouseOverUnitListener(List<Unit> units, Territory territory) {
    if (!isActive() || currentAction != null) {
      return;
    }
    getMap().setUnitHighlight(Set.of(Collections.unmodifiableList(units)));
  }
```
The method has a few advantages, notably the error handling is easier to reason about (because lambda's can be weird about exceptions).

Next, with the above update made, the 'lambda' variable is not super useful and we can inline any references to it, leading to the following:


```
    if (frame.getEditDelegate() == null) {
      // current turn belongs to remote player or AI player
      getMap().removeMapSelectionListener(mapSelectionListener);
      getMap().removeUnitSelectionListener(unitSelectionListener);
      getMap().removeMouseOverUnitListener(this::setMouseOverUnitListener);
      setWidgetActivation();
    } else if (!this.active && active) {
      getMap().addMapSelectionListener(mapSelectionListener);
      getMap().addUnitSelectionListener(unitSelectionListener);
      getMap().addMouseOverUnitListener(this::setMouseOverUnitListener);
      setWidgetActivation();
    } else if (!active && this.active) {
      getMap().removeMapSelectionListener(mapSelectionListener);
      getMap().removeUnitSelectionListener(unitSelectionListener);
      getMap().removeMouseOverUnitListener(this::setMouseOverUnitListener);
      cancelEditAction.actionPerformed(null);
    }
```

I _think_ this might be the better route here, this lambda does strike me as unnecessary.

DanVanAtta: side-note, to keep a consistent structure in the class, we'd want to make more and similar updates within this class. It might be easier to back out the 'suppress warnings' change here for now and then follow up as desired with further updates.

http-clients/feign-common/src/main/java/org/triplea/http/client/HttpClient.java
```diff
@@ -30,7 +30,7 @@
  * @param <ClientTypeT> The feign client interface, should be an interface type that has feign
  *     annotations on it.
  */
-@Slf4j
+@Log
```
DanVanAtta, line 33: Why java.log over slf4j?

http-clients/feign-common/src/main/java/org/triplea/http/client/HttpClient.java
```diff
@@ -71,21 +71,16 @@ protected Response logAndRebufferResponse(
                     : super.logAndRebufferResponse(configKey, logLevel, response, elapsedTime);
               }
 
+              @FormatMethod
               @Override
               protected void log(
                   final String configKey, final String format, final Object... args) {
                 final String logMessage = String.format(format, args);
-                log.trace(configKey + ": " + logMessage);
+                log.info(configKey + ": " + logMessage);
```
DanVanAtta, line 79: Why elevate this to `info`?

http-clients/feign-common/src/main/java/org/triplea/http/client/HttpClient.java
```diff
@@ -71,21 +71,16 @@ protected Response logAndRebufferResponse(
...
             })
         .logLevel(Logger.Level.BASIC)
-        .options(
-            new Request.Options(
-                DEFAULT_CONNECT_TIMEOUT_MS,
-                TimeUnit.MILLISECONDS,
-                DEFAULT_READ_TIME_OUT_MS,
-                TimeUnit.MILLISECONDS,
-                true))
+        .options(new Request.Options(DEFAULT_CONNECT_TIMEOUT_MS, DEFAULT_READ_TIME_OUT_MS))
```
DanVanAtta, line 83: This `Request.Options` constructor is deprecated. 

```
Deprecated
please use Options(long, TimeUnit, long, TimeUnit, boolean)
```

lib/java-extras/src/main/java/org/triplea/java/Retryable.java
```diff
@@ -3,85 +3,82 @@
...
-public class Retryable<T> {
+public class Retryable {
   private static final Consumer<Duration> DEFAULT_THREAD_SLEEP =
       duration -> Interruptibles.sleep(duration.toMillis());
 
   private final Consumer<Duration> threadSleeper;
   private final int maxAttempts;
   private final Duration fixedBackOff;
-  private final Supplier<Optional<T>> taskRunner;
+  private final BooleanSupplier taskRunner;
```
DanVanAtta, line 23: Is this update to take into account that `T` is so far only a boolean?

This `Retryable` class was meant to be generic, but I don't think it turned out that way very nicely. There might be a better solution to this class, namely deleting it or rewriting it so it is more usable as a generic retry mechanism.

DanVanAtta (commented):
Overall, thank you for the updates. A few comments here and there to consider and respond to.


Side-note, this PR is well sized (thank you!) though contains several independent changes. When possible try to submit these as independent PRs. Review is a huge bottleneck for TripleA, anything that makes review easier is a win. Second, some of these updates could very well be merged as-is, but a PR is all-or-none, so some of it must wait until everything is taken care of. Hence it can be in everyones interest to break things up. TripleA does tend to make a person very good at version control.

game-app/game-core/src/main/java/games/strategy/engine/framework/ui/SaveGameFileChooser.java
```diff
@@ -45,12 +45,13 @@ private static String getSaveGameName(final GameData gameData) {
...
 
   private static String formatGameName(final String gameName) {
+    @SuppressWarnings("JavaTimeDefaultTimeZone")
     final ZonedDateTime now = ZonedDateTime.now();
     return new StringJoiner("-")
         .add(String.valueOf(now.getYear()))
         .add(String.valueOf(now.getMonthValue()))
         .add(String.valueOf(now.getDayOfMonth()))
-        .add(gameName.replaceAll(" ", "-") + GameDataFileUtils.getExtension())
+        .add(gameName.replace(" ", "-") + GameDataFileUtils.getExtension())
```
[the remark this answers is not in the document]
frigoref: As stated in the pull request "replaceAll is the same as replace when no regex is used, but consumes more performance"

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/ui/ClientSetupPanel.java
```diff
@@ -41,7 +41,9 @@ public ClientSetupPanel(final ClientModel model) {
     clientModel.setRemoteModelListener(
         new IRemoteModelListener() {
           @Override
-          public void playersTakenChanged() {}
+          public void playersTakenChanged() {
+            // nothing to do, but implementation required
```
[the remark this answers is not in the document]
frigoref: removed ", but ..."

game-app/game-core/src/main/java/games/strategy/engine/framework/ui/SaveGameFileChooser.java
```diff
@@ -45,12 +45,13 @@ private static String getSaveGameName(final GameData gameData) {
   }
 
   private static String formatGameName(final String gameName) {
+    @SuppressWarnings("JavaTimeDefaultTimeZone")
```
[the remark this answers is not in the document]
frigoref: K, adapted it.
The reason why I added it was just to try to be able to finally compile everything successfully. I spent several hours to fix code just to be able to compile all code once. Therefore, I did not check each warning I needed to suppress, but I also thought about running a check later with disabled suppress warnings to find them back and really fix them.

game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
frigoref: Adjustment made as suggested.
However, I did not see similar updates possible as the other listeners where having multiple method implementations.
Are you sure that a similar update would be possible? If so, where and how?

http-clients/feign-common/src/main/java/org/triplea/http/client/HttpClient.java
frigoref: Fixed.

http-clients/feign-common/src/main/java/org/triplea/http/client/HttpClient.java
frigoref: Fixed.

http-clients/feign-common/src/main/java/org/triplea/http/client/HttpClient.java
frigoref: Fixed.
Good question, I did not notice that IntelliJ was replacing this.

frigoref:
@DanVanAtta 
I did solve the code in parallel and just pushed a new commit.
Do you get individual mails for each "conversation" here? (If so, I am sorry for all the spam!)
Could we have a chat/call to talk? I would like to align with you to 
- Speed up my work (commit slicing, gradle tasks to run, etc.)
- Speed up our interaction (What do you need? When and how can we align?)
- Enhance the documentation for the whole process of submitting a change (fork, branch, commit, review process).
Currently I am on EET time (will be back on CET time later this week).

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/ui/ClientSetupPanel.java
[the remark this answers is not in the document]
DanVanAtta: > An implementation being required is generally already conveyed by the @Override

It is still redundant to the language syntax to say an implementation is required. What more, a person might think this comment was put in here for a strong reason to contradict the '@Override', why mention it if this were not special somehow? But then, the method has no implementation here, it is empty, and then we say the  implementation is required.

The original text was actually more clear. That or a simple `// no-op` comment IMO are both better.

game-app/game-core/src/main/java/games/strategy/engine/framework/ui/SaveGameFileChooser.java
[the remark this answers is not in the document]
DanVanAtta: What command are you running to compile?

This is odd to see.
1. The build does pass, this should not be OS specific.
2. We removed error prone, you should not be encountering these. Which branch & git SHA are you building?

error prone was removed in: commit 337274a8d82c7ae8ec405fb65f1b780d5a6f9df9 (tag: 2.6.372)

game-app/game-core/src/main/java/games/strategy/engine/framework/ui/SaveGameFileChooser.java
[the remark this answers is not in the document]
DanVanAtta: Ah okay, I missed that. The overview formatting could use some touch-ups, it's hard to grok.

game-app/game-core/src/main/java/games/strategy/triplea/ui/EditPanel.java
DanVanAtta: > possible as the other listeners where having multiple method implementations and a specific interfaces implemented.

You're right. 

A person _could_ create some wrapper code to replace those implementations with lambdas. It's not worth it though, it would be better to rewrite this UI class to better standards and remove a litany of other problems while at it.

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/ui/ClientSetupPanel.java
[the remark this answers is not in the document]
DanVanAtta: I think the original text was stronger. I'm now slightly concerned that this text is very contradictory:
````
 // nothing to do, implementation required
```

1. An implementation is required, yet there is no implementation, this is an empty method
2. If there is nothing to do, and we require an implementation, this is contradictory
3. The `@Override` does convey we need to add this method, noting that an implementation is required seems to be commenting on something beyond the fact there is an  `@Override` on the method. That there is something else required. Bsaically, the `@Override` already says "implementation is required". If we feel a strong need to re-state that, presumably it is for a reason. I think we might be starting to raise more questions for the reader than we are answering.

DanVanAtta:
> Do you get individual mails for each "conversation" here? (If so, I am sorry for all the spam!)

They are individual emails, though Gmail de-dupes them into a single message thread. The 'review' feature in github does help to reduce emails, so for the initial review at least it can be a single email. Beyond that, the 'spam' is seemingly a necessary evil, it's okay!

> Could we have a chat/call to talk? I would like to align with you to

Sure. I did not find your username in the forums. Message me there and we can exchange phone numbers.

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/ui/ClientSetupPanel.java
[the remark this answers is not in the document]
frigoref: Please note that this is outdated and the current comment is
`// nothing to do`
which I think should be fine.

lib/java-extras/src/main/java/org/triplea/java/Retryable.java
frigoref: Redone in commit [94b829759c377f9411cf58033c9ae750e3443db8](https://github.com/triplea-game/triplea/pull/9939/commits/94b829759c377f9411cf58033c9ae750e3443db8)

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/ui/ClientSetupPanel.java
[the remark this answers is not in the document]
DanVanAtta: Ah, my misunderstanding.   :+1:

game-app/game-core/src/main/java/games/strategy/engine/framework/ui/SaveGameFileChooser.java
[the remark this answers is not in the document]
frigoref: Check the current failed runs that happened after I just changed a comment.
Before all checks passed successfully.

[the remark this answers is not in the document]
DanVanAtta: It turns out the PR build is rebasing PRs against latest master.

Following the commands in the 'checkout' sequence, I was able to replicate the build errors.

Notice for example these lines:

> /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +4ee61ab2062420c10cc2f1faf8b938b4084711e7:refs/remotes/pull/9939/merge
      
> * [new ref]         4ee61ab2062420c10cc2f1faf8b938b4084711e7 -> pull/9939/merge
       
> /usr/bin/git checkout --progress --force refs/remotes/pull/9939/merge
   Note: switching to 'refs/remotes/pull/9939/merge'.
  
> HEAD is now at 4ee61ab Merge fc03b60f7e2e02c7402c9f1afad530434dafaef6 into c61e13cd2d87b288446a8c00aa9feedab7cbaf4b
  
That last one is really key, fc03b6 is the head of the PR branch, c61e13cd is the head of master. After getting to that point a `./gradlew compileJava` replicates the errors.

----

If you merge latest master into your branch you should see the same errors. Though, a merge of latest master will result in merge conflicts, you'll need to resolve those.

  

What partially clued me in is that `HttpInteractionException` is removed and so is the custom `FeignDecoder` is as well: 
- https://github.com/triplea-game/triplea/pull/9933/files#diff-a66634c3205f49af440b11dbf069687dc307f6fd1e2b0ddb1a3077031d523fb7L107
- https://github.com/triplea-game/triplea/pull/9933/files#diff-854e28d85dbe0561b51afd0075f01be7451ea6113e1fe0c84649413d4c25cea9

Though, those were recent updates - it was odd to see an error message relating to that on a seemingly older branch that would not have had those changes in it.

The fact we merge PRs onto latest master is actually quite good. In this case had the build passed we would have merged to then later see an error in that build. Instead we see it now, though it's confusing where those errors come from. Of note, it looks like the build in your fork worked fine on this step (notably becuase it was rebasing master (frigoref/master) onto itself, and is not using the upstream version 'triplea-game/master').

Side-note: when working with forks it's generally easiest to avoid using 'master'. 'master' essentially kept in sync with upstream master, eg: `git checkout master; git fetch upstream; git pull --rebase upstream/master; git push origin master` Then feature branches are cut from 'master' with `git checkout master; git checkout -b branch-name/some-branch-name`.

PR #9940, opened 2021-12-25
Header redirect/lobby variable port number

DanVanAtta:

commit f68abbeb399b7298106b4b88537db9c0ca0bbad3

    Nginx config to redirect to differing lobby depending on header version
    
    Adds an NGINX config that checks the 'triplea-version' header value,
    and based on that value redirects. To support running multiple servers
    on the same host, we need to parameterize the lobby server to be able
    to inject port number. The end goal is we can have multiple lobbies
    running on the same host, different port numbers, and clients will
    be redirected to the correct lobby based on version number in header.

commit 860ed083909beaf1fb39ffe9903cc9454012523a (header-redirect/wip)

    Nginx Config: remove deprecated ssl property
    
    'sudo nginx -t' was generating this warning:
    
    > nginx: [warn] the "ssl" directive is deprecated, use the "listen ... ssl" directive instead in /etc/nginx/sites-enabled/default:21

### DIFF
--- a/infrastructure/ansible/roles/nginx/defaults/main.yml
+++ b/infrastructure/ansible/roles/nginx/defaults/main.yml
@@ -9,3 +9,19 @@ dhparams_pem_file: /etc/nginx/dhparam.pem
 nginx_allowed_ports:
   - 443
   - 80
+
+# 'lobby_proxy_pass' captures common selection logic for redirecting clients based on their
+# TripleA Version.
+#
+# Note: header names automatically have dashes converted to underscores.
+# Headers containing underscores are ignored entirely!
+lobby_proxy_pass: |
+    if ($http_triplea_version = "") {
+      proxy_pass          http://localhost:{{ lobby_25_port }};
+    }
+    if ($http_triplea_version = "2.5") {
+      proxy_pass          http://localhost:{{ lobby_25_port }};
+    }
+    if ($http_triplea_version = "2.6") {
+      proxy_pass           http://localhost:{{ lobby_26_port }};
+    }
--- a/infrastructure/ansible/roles/nginx/templates/etc_nginx_sites_enabled_default.j2
+++ b/infrastructure/ansible/roles/nginx/templates/etc_nginx_sites_enabled_default.j2
@@ -6,7 +6,7 @@ server {
 
 # warning: be sure that we can run certbot role and then re-run the nginx
 # role without this file changing. If this file is changed (white-space included)
-# then nginx will be restarting, disconnecting all active connections.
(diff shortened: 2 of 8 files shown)


infrastructure/ansible/roles/nginx/templates/etc_nginx_sites_enabled_default.j2
```diff
@@ -34,23 +33,17 @@ server {
...
       proxy_set_header        Host $host;
       proxy_set_header        X-Real-IP $remote_addr;
       proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
       proxy_set_header        X-Forwarded-Proto $scheme;
-
-      # Fix the “It appears that your reverse proxy set up is broken" error.
-      proxy_pass          http://localhost:8080;
       proxy_read_timeout  90;
-
-      proxy_redirect      https://localhost:443 https://localhost:8080;
```
DanVanAtta, line 47: This 'proxy_redirect' line is seemingly unneeded. While testing, this line did not seem to matter.

infrastructure/ansible/roles/nginx/templates/etc_nginx_sites_enabled_default.j2
```diff
@@ -15,7 +15,6 @@ server {
     ssl_certificate {{ cert }} 
     ssl_certificate_key {{ cert_key }}
 
-    ssl on;
```
DanVanAtta, line 18: This 'ssl on' directive is deprecated. This line `listen 443 ssl http2;` is enough to enable SSL

infrastructure/ansible/roles/nginx/defaults/main.yml
```diff
@@ -9,3 +9,19 @@ dhparams_pem_file: /etc/nginx/dhparam.pem
...
   - 443
   - 80
+
+# 'lobby_proxy_pass' captures common selection logic for redirecting clients based on their
+# TripleA Version.
+#
+# Note: header names automatically have dashes converted to underscores.
+# Headers containing underscores are ignored entirely!
+lobby_proxy_pass: |
+    if ($http_triplea_version = "") {
```
DanVanAtta, line 19: 'if' is evil in NGINX apparently: https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/

I tried a solution that used `map` to createa `$lobbby_server` variable - though that did not work at all for the websocket `proxy_pass` configurations.

RoiEXLab (approved):
Looking good, the progress you made there is very impressive

Issue #9936, opened 2021-12-25
2.6 beta: Image display in 270BC


RaiNova:
### Reproduction Steps

1. Start 240BC current head of TripleA 2.6 on Windows 10
2. Select View -> Show Map Blends (don't know if this is relevant)
2. Play around a while
3. Navigate via the mini map
4. The screenshot below shows the image display issue hereby eported
5. The second screenshot shows the same issue in the battle display 

### Attach a Save Game

![image](https://user-images.githubusercontent.com/2886662/147383676-9a529072-9923-433f-80f4-b76abf33d516.png)

![image](https://user-images.githubusercontent.com/2886662/147383736-f896181b-271b-4266-bfeb-1927d0983ad9.png)

### Exception
16:34:57.690 [AWT-EventQueue-0] INFO  g.s.e.m.UnifiedInvocationHandler - Blocking network operation performed from EDT
java.lang.Exception: null
	at games.strategy.engine.message.UnifiedInvocationHandler.invoke(UnifiedInvocationHandler.java:52)
	at com.sun.proxy.$Proxy21.getPlaceableUnits(Unknown Source)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:564)
	at games.strategy.engine.player.DefaultPlayerBridge$GameOverInvocationHandler.invoke(DefaultPlayerBridge.java:151)
	at com.sun.proxy.$Proxy21.getPlaceableUnits(Unknown Source)
	at games.strategy.triplea.ui.PlacePanel.getUnitsToPlace(PlacePanel.java:249)
	at games.strategy.triplea.ui.PlacePanel$1.territorySelected(PlacePanel.java:58)
	at games.strategy.triplea.ui.panels.map.MapPanel.notifyTerritorySelected(MapPanel.java:490)
	at games.strategy.triplea.ui.panels.map.MapPanel$3.mouseReleased(MapPanel.java:213)
	at java.desktop/java.awt.AWTEventMulticaster.mouseReleased(AWTEventMulticaster.java:298)
[...]

PR #9944, opened 2021-12-27
Remove image cache from TileImageFactory

DanVanAtta:
1. The performance gain of the cache does not seem to be there, map renders
   quite well without.

2. Cache revolves around complex logic and is complex itself by storing
   a 'SoftReference' rather than a hard-reference.


RoiEXLab:
@DanVanAtta Did you check with a zoom setting other than 100%?
I remeber stumbling over this in #4921 but I don't know anymore why I didn't remove it back then

RaiNova:
Just for information: I ran the first round of 270BC through a profiler with the following results (total time results from user interaction and is therfore meaningless):

Before (with cache and image display issue):
image

After (no cache, image display correct):
image

frigoref:
@DanVanAtta Will you withdraw this PR?
The measurements of @RaiNova shows an 11% decrease of performance after the commit (no cache, image display correct).

RaiNova:
> @DanVanAtta Will you withdraw this PR? The measurements of @RaiNova shows an 11% decrease of performance after the commit (no cache, image display correct).

I appreciate this quick fix. With current hardware: Is it still necessary to use tiles?

DanVanAtta:
> The measurements of @RaiNova shows an 11% decrease of performance after the commit (no cache, image display correct).

If the image display is not correct, then the performance measurement I do not think is valid. If the code is not correct, it's not correct (it does not matter if it runs faster if it can make the game unplayable). That is my perspective, largely rooted in "it's better to optimize correct code, than it is to correct optimized code." Seemingly we are in the latter case here.

Is the suggestion that we go with 10% better performance at the cost of correctness? Are there other corrections that can be done without removing this cache? (also, 10% does not necessarily seem like all that much, what are the absolute differences? 1s over 15s vs 14s - could that even be just measurement error? Are those measurements accurate to within 10%?)

frigoref:
@DanVanAtta The commit message and your initial PR text 
"The performance gain of the cache does not seem to be there, map renders
quite well without.

Cache revolves around complex logic and is complex itself by storing
a 'SoftReference' rather than a hard-reference."
indicated that this PR was just meant to be a cleanup, i.e. removing of assumed "unnecessary" caching. 
@RaiNova just tested the performance and a drop of performance indicates that the caching is not unnecessary.

If you are saying this PR was about fixing a real bug (correctness), then I would ask you to explain which bug exactly this is trying to fix.

frigoref:
I see 2 points from @RaiNova :
1. Introducing an IDE profiler for usage of the developers. - > I guess a discussion topic in the forum would make sense. @RaiNova could you open one? 
2. Checking whether removing the engine's tile management would result in better performance. - > I guess a feature request would be most suited for this. @RaiNova could you check whether we already have one and if not, own one?

DanVanAtta:
@frigoref I see the confusion, yes the commit mentions nothing about a commit message. That message should (and will after squash merge) will state that this is a fix for: https://github.com/triplea-game/triplea/issues/9936

The rest of the message is basically to say: 10% is not a lot, this caching was not worth a lot, was done in a complicated way, even without the problem fix one could argue for its removal. Without the context of the problem fix, then it's a debatable move (and for 1s from 15s to 14s that is perhaps within the error bounds of measurements, it's very debatable)

frigoref:
@DanVanAtta understood. I cannot say weather the issue #9936 was fixed with this (and closed before this PR was merged), so I recommend to merge this PR. If you are certain this PR fixes an issue, I would vote for merging the PR even though the performance might suffer.
@RaiNova What is you thought on this?

RaiNova:
The issue was fixed. Merge it 👍

PR #4055, opened 2018-09-17
Decouple headless game server from assets folder

ssoloff:
## Overview

Decouples the headless game server from the _assets_ folder so that the headless game server artifact does not have to include this content.  Removal of the _assets_ folder from the headless game server artifact reduces its size from 17.1 MiB to 9.84 MiB (-42%).

This PR also finally gets rid of the smell of having to run the `game-headed` and `game-headless` projects from the `game-core` folder in a development environment.

## Functional Changes

* Moved the _assets_ and _dice_servers_ folders to the `game-headed` project; moved the _.triplea-root_ marker file to both the `game-headed` and `game-headless` projects.
* Updated the Eclipse launch configurations, the Gradle `run` tasks, and the Gradle `integTest` task to run from the corresponding project directory rather than the `game-core` project directory.
* Introduced a new service (`ApplicationContext`) so that code in `game-core` can get access to the entry point class used to start the application (see item below about `ClientFileSystemHelper`).
    * If this turns out to work well, I'll probably follow-up in a separate PR to replace the recently-added `GameRunner$Context` class with the use of services like this.  This will result in many fine-grained interfaces rather than the single coarse-grained `GameRunner$Context` type, which would seem to follow the _Interface segregation principle_.
* Modified `ClientFileSystemHelper#getCodeSourceFolder()` to use `ApplicationContext#getMainClass()` as the code source reference point rather than a hard-coded class from the `game-core` project.  This change allows this method to find the appropriate code source folder in a development environment (i.e. directory based rather than jar based) regardless of whether the game is running from the `game-headed` project folder or the `game-headless` project folder.
* Modified `ResourceLoader#getMapResourceLoader()` to make the presence of the _assets_ folder optional.  If it's not found, it won't be added to the list of locations searched for resources.
    * Even if, for some reason, a headed client was missing its _assets_ folder, a request to load a resource that would normally be found in that folder will still fail in downstream code.

## Refactoring Changes

* Moved the `HeadedGameRunner` and `HeadlessGameRunner` classes to more appropriate packages.

## Manual Testing Performed

* Verified I could run the headless game server in the following configurations:
    * Eclipse launch configuration
    * Gradle `:game-headless:run` task
    * Portable build
* Verified I could run the headed game client in the following configurations:
    * Eclipse launch configuration
    * Gradle `:game-headed:run` task
    * Portable build
    * Unix installer
    * Windows x86_64 installer
* Tested connecting one of the above clients to one of the above servers to verify I could play the following maps:
    * Classic
[...]

PR #9945, opened 2021-12-27
Fix ClientFileSystemHelper#getRootFolder to actually return a folder

DanVanAtta:
The method is improperly returning the path of the '.triplea-root' touch
file rather than the parent folder of that file.


RoiEXLab (approved):
Wait, how wasn't this noticed until now? Isn't this code a couple years old by now?

DanVanAtta:
Diffing on the file, this problem was introduced IIRC a year ago. It's been a very long time since we had a release.

Though, given we have not noticed even on prerelease versions - I wonder if somehow this is dead-code.

frigoref:
@DanVanAtta, @RoiEXLab 
I have created some test cases for the class ClientFileSystemHelper and would ask for introducing them here.
https://github.com/frigoref/triplea/tree/fix/get-root-folder

One was a bit odd, so I would ask to check this carefully. Is the "service" here a system service that is only relevant for a running game and cannot be "properly" covered in a test case?
`@Test
  void getRootFolder() {
    assertThrows(ServiceNotAvailableException.class, () -> ClientFileSystemHelper.getRootFolder());
  }`

RoiEXLab:
@frigoref IIRC this "service" is a mechanism that allows to abstract away some implementation details that differ for different environments. In this particular example there are 2 different service implementations for `ApplicationContext`: There's the `HeadedApplicationContext` which would be loaded for client environments and the `HeadlessApllicationContext` which would be loaded in server environments mostly our 'bots' as they're called.

#4055 Introduced this change, @ssoloff can probably provide a better explanation of how it works than I can. And I honestly have no Idea (or I did know in the past and simply forgot) how to register those services in the first place so they can be used by tests. I think you'll end up having to implement a dummy service yourself and somehow reference it in a special file inside the META-INF folder. Not sure if that's enough to make it work though.

ssoloff:
@frigoref @RoiEXLab There was some discussion of tests interacting with `ServiceLoader` a few months ago on the [forum](https://forums.triplea-game.org/topic/2937/how-test-ui-elements-get-serviceloader-working).

PR #9946, opened 2021-12-27
Draft/Discussion: Add services files to fix 'unable to find service'

DanVanAtta:
Posting for discussion.

The below solves finding an application context from a test instance. Essentially we register an 'application context' implementation so that 'codeSource' then becomes `triplea/game-core/test/out` (where `out` is the output directory for builds)

We then recursively search up for the file '.triplea-root', which is in 'triplea/game-headed/' and not 'triplea/game-core',  so we will not find it. It's difficult to think how to structure this differently for that search to succeed. Ultimately we are looking for the 'assets' folder, and to find that we have to recursively go up directories to then come back down. The real problem is that we have sub-projects that do not match how the file system would look in a deployed context, so we are contorting to try and get this to work.

Possible solutions for how to deal with this situation in test:

(1) Perhaps we can add a 'ROOT_FOLDER_PATH' environment variable that can be used to override the 'root folder'. We can then set that from within test code as needed and/or include runners that set the variable as appropriate.

(2) An alternative solution is for us to use the custom 'Injections' class to get instances of `ClientFileSystemHelper` wherever needed. In this case we can override the injection for test.

(3) We could replace usages of `ClientFileSystemHelper` with equivalent lambdas and then inject those lambda implementations from test. EG:  `ClientFileSystemHelper.getRootFolder` vs `Supplier<Path> rootFolderFinder`, and the `Supplier` becomes a constructor arg to the class.

### DIFF
--- a/game-app/game-core/src/test/java/games/strategy/engine/ClientFileSystemHelperTest.java
+++ b/game-app/game-core/src/test/java/games/strategy/engine/ClientFileSystemHelperTest.java
@@ -2,14 +2,28 @@ package games.strategy.engine;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
+import static org.hamcrest.core.IsNull.notNullValue;
 
 import games.strategy.triplea.settings.AbstractClientSettingTestCase;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 
 @SuppressWarnings("InnerClassMayBeStatic")
 final class ClientFileSystemHelperTest {
+
+  @Test
+  void getRootFolder() {
+    Path rootFolder = ClientFileSystemHelper.getRootFolder();
+
+    assertThat(rootFolder, is(notNullValue()));
+    assertThat("should be a folder", Files.isDirectory(rootFolder));
+    assertThat(
+        "should contain the '.triplea-root' touch file",
+        Files.exists(rootFolder.resolve(".triplea-root")));
+  }
+
   @Nested
   final class GetUserMapsFolderTest extends AbstractClientSettingTestCase {
     @Test
--- a/game-app/game-core/src/test/java/games/strategy/engine/TestApplicationContext.java
+++ b/game-app/game-core/src/test/java/games/strategy/engine/TestApplicationContext.java
@@ -1,0 +1,10 @@
+package games.strategy.engine;
+
+import org.triplea.game.ApplicationContext;
+
+public class TestApplicationContext implements ApplicationContext {
+  @Override
+  public Class<?> getMainClass() {
+    return TestApplicationContext.class;
+  }
+}
--- a/game-app/game-core/src/test/resources/META-INF/services/org.triplea.game.ApplicationContext
+++ b/game-app/game-core/src/test/resources/META-INF/services/org.triplea.game.ApplicationContext
@@ -1,0 +1,1 @@
+games.strategy.engine.TestApplicationContext


RaiNova:
I tried the same approach a while ago, and it didn't work - in the same way as this PR in its current state: `TestApplicationContext` is not in in 'triplea/game-headed/'. 

How about solution (0) (not 4, because it is older than this PR):
_Inject_ the correct path by calling the new method `ClientFileSystemHelper.setCodeSourceFolder`. Disadvantage: It would only be called from test code and is dead code in production. Is that a difference to solutions (1)-(3)?

DanVanAtta:
> Would it be possible to move the assets folder up one level?

I think it is needed where it is for packaging.


> Inject the correct path by calling the new method ClientFileSystemHelper.setCodeSourceFolder. Disadvantage: It would only be called from test code and is dead code in production. 

This raises the question, what is the correct path? The correct path is different depending on the build output folder or if the game is installed vs run from source. Due to the split of the sub-project, I suspect there is no such correct path at all. For example, if 'game-core' knew about 'game-headed' that breaks the architecture where we do not want 'game-core' to know anything about the layers above it. As soon as it does, we are back to 'game-core' being so massive it has a gravitational pull and everything once again must reside in that sub-project (and there was never much enforcement of boundaries within the TripleA code).

Another disadvantage is the test code is going to be for its own sake and will not actually test the production configuration.

> Is that a difference to solutions (1)-(3)?

Yes, it is different.

(1) is the most similar and instead of doing an override we use an env variable to do so. This allows it to be configured per developer, which is what is needed as it's completely unknown which path 'code source' will have (unless we are installing from installer, in which case we can be pretty sure that path is to the jar file)

But even more we should be override 'getRootFolder' and not hacking where it gets it data. That would be assuming a number of implementation details, can be brittle and hard to understand.

(2) Is very different, instead of code doing this:

```

baz() {
  foo();
}

foo() {
 String impossibleToInject = ClientFileSystem.bar();
}
```

We would do:

```
bas() {
  foo(Injections.getInstance().clientFileSystemHelper());
}

foo(ClientFileSystemHelper helper) {
 String impossibleToInject = helper.bar();
}

```

(3) is still yet different:

```
bas() {
  foo(ClientFileSystemHelper::bar);
}

foo(Supplier<String> bar) {
 String impossibleToInject = bar.get();
}
```

(2) and (3) are taking inspiration from the IOC pattern & the "Don't call us, we'll call you" principle.

DanVanAtta:
@RaiNova I think (2) and (3) are probably the best bets. 'game-core' will never be able to really know where image assets are. The best we can do is to ensure that we create the correct path, but we'll never be able to really do the full image loading in test. Basically anything that uses `ClientFileSystemHelper` should be pushed to the edge of the system, and anything that uses it is replaced with mock implementations.

RaiNova:
(2) and (3) need to provide a class/function that knows where the assets folder is/can be found.
That knowledge can be in the code (solution 0) or maybe in an environment variable (solution 1).
When I start a unit test from IntelliJ IDEA, how can I tell the IDE to use particular environment variables - other than editing each Run/Debug configuration once the IDE has set it up?

DanVanAtta:
> other than editing each Run/Debug configuration once the IDE has set it up?

That would be roughly the route. We could check these in at least to make it easier. It's still a gross solution. Generally we are trying to solve something we cannot. The location of the 'assets' folder is arbitrary when building form source, there is no 'one location'. It depends on everyones individual setup and config. If we mandated that nobody use 'eclipse', we could perhaps make that location standard and would then have a hope.

Basically I think what it comes down is anything using `ClientFileSystemHelper` is untestable. We are best pushing anything that uses it to the system edge and then treat that entire component as an integration component (similar to JDBC, or something that does a network call to an API like github).

DanVanAtta:
> (2) and (3) need to provide a class/function that knows where the assets folder is/can be found.

True, though nothing is to say that path has to be correct. If we pass that to components that simply build a path, we can verify that. Then anything that loads an image from that path, we also inject that and then can mock the image it returns.

That is a larger change, but in part because there is so much static coupling in TripleA.

RaiNova:
> Basically I think what it comes down is anything using ClientFileSystemHelper is untestable.

Im too naive to accept this. And I don't want to manually edit every Run/Debug configuration. Can we assume that the assets folder is in a folder parallel to a folder in the code source path, one example of that being in a folder parallel to 'game-core' - which accidently maybe named 'game-headed'?

RaiNova:
In what sense is passing a callback function/class more IOC than passing a path? Data is passive :wink:

DanVanAtta:
> Can we assume that the assets folder is in a folder parallel to a folder in the code source path

The problem is how do we find the assets folder. What is the simplest way we can achieve this?  It was attempted to do so with a touch file: find the touch file -> find the assets folder. It still think that might be the simplest approach.

Beyond that, our constraints are that this needs to work when:
- running via gradle in CLI (root path is arbitrary in this case)
- running from IDE (IDEA at least)
- running from built installer (the most important, but most difficult to validate)

PR #9947, opened 2021-12-27
NEW: Fixes to errorprone errors (incl. code smells)

frigoref:
-Rename Retryable -> Retriable (and RetryableTest -> RetriableTest)
-FileUtilsTest (path with generic line separator)

- UnnecessaryLambda (also turn off completely in game-core/build.gradle)
- SameNameButDifferent (HttpClient.java)

CodeSmell fixes
- ClientSetupPanel.java: empty method without explanation
- EditPanel.java:
- HttpClient.java: %n produces platform-specific line separator instead of \n
- SaveGameFileChooser.java: replaceAll is the same as replace when no regex is used, but consumes more performance
- ServerModel.java:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,6 +6,7 @@ plugins {
 }
 
 apply from: 'gradle/scripts/yaml.gradle'
+apply plugin: 'eclipse'
 
 ext {
     schemasDir = file('config/triplea/schemas')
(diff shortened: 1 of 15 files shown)


DanVanAtta:
@frigoref where are the warnings coming from?

> SuppressWarnings statements for
    UnnecessaryLambda (also turn off completely in game-core/build.gradle)

AFAIK error-prone is removed, so error prone warning suppression should not be needed (and could be all removed).

frigoref:
> @frigoref where are the warnings coming from?
> 
> > SuppressWarnings statements for
> > UnnecessaryLambda (also turn off completely in game-core/build.gradle)
> 
> AFAIK error-prone is removed, so error prone warning suppression should not be needed (and could be all removed).

The SuppressWarnings statements have been removed. Should be fine now.

DanVanAtta (approved):

DanVanAtta:
@frigoref thanks for the work!

Side-note: When you get the chance, please sign the [CLA](https://github.com/triplea-game/triplea/pull/9947#issuecomment-1001592346)

PR #9960, opened 2021-12-29
PoC i18n (internationalization)

frigoref:
@triplea-game/community 
Could you please have a look so we can discuss based on this Proof of Concept commit how we might be able to gradually introduce internationalization (i18n) into TripleA?

Some info:
- My goal is to get TripleA code to be internationalized which I am sure would bring more people around and maybe also contributers.
- For the moment I would just aim to not introduce none-internationalized code so we can gradually increase the internationalization.
- If we go that way I would also update the docs/developments entries for this.
- With the IntelliJ Plugin ResourceBuilder it is rather easy to extract a String into a .properties file.
- I would recommend not to use to many of those, but place them rather high in the resource folder structure so it is easier to translate them.
- The GameResourceBundle.java is currently just a place holder in case I find some reusable patterns. I tried to find some way to have separate classes with static methods to not have to define the paths in every class, but so far I couldn't find any. I always run into the problem that I would like the same static methods to convert the key into the language dependent string (via the ResourceBundle), but also have the specific bundle each in a separate class. It seems not to work with Java inheritance.

Let me know what you think or let's discuss this our strategy how to go forward here.

DanVanAtta already mentioned what he would want to make sure before we continue:
1. we can drive this to completion
2. we need a clean proof of concept:
- easy to understand
- extremely minimal updates to existing code
- flexible to allow for missing translations
3. Incremental. Ideally we have it so we can add translations a few class files at a time and can submit many PRs in parallel when we are in the mode of 'execution' and are no longer designing how we do this.

PS: The initial translation of a ResourceBundle properties file might be able to be done automatically and a review could be done by the community.


game-app/game-core/src/main/java/games/strategy/triplea/settings/ClientSetting.java
```diff
@@ -221,6 +222,11 @@ protected ClientSetting(final Class<T> type, final String name, final @Nullable
    * result in an {@code IllegalStateException} being thrown by methods of this class.
    */
   public static void initialize() {
+    final String defaultLocaleLang = Locale.getDefault().getLanguage();
+    if (!defaultLocaleLang.equals(Locale.ENGLISH.getLanguage())
+        && !defaultLocaleLang.equals(Locale.GERMAN.getLanguage())) {
+      Locale.setDefault(Locale.GERMANY);
+    }
```
RoiEXLab, line 229:
```suggestion
    final String defaultLocale = Locale.getDefault();
    if (!List.of(Locale.ENGLISH, Locale.GERMAN).contains(defaultLocale)) {
      Locale.setDefault(Locale.GERMANY);
    }
```

game-app/game-core/src/main/resources/i18n/games/strategy/engine/framework/ui.properties
```diff
@@ -0,0 +1,23 @@
...
+startup.SetupPanelModel.button.PlayOnline.HostNetworkGame.Tooltip.Line2=Anyone on a LAN will be able to connect.
+startup.SetupPanelModel.button.PlayOnline.HostNetworkGame.Tooltip.Line3=Anyone from the internet can connect as well, but only if the host has configured port forwarding correctly. 
+startup.SetupPanelModel.button.PlayOnline.Label=Play Online
+startup.SetupPanelModel.button.PlayOnline.Tooltip.Line1=Find Games Online on the Lobby Server.
+startup.SetupPanelModel.button.PlayOnline.Tooltip.Line2=TripleA is MEANT to be played Online against other humans.
+startup.SetupPanelModel.button.PlayOnline.Tooltip.Line3=Any other way is not as fun!
+startup.SetupPanelModel.button.StartLocalGame.Label=Start Local Game
+startup.SetupPanelModel.button.StartLocalGame.Tooltip.Line1=Start a game on this computer.
+startup.SetupPanelModel.button.StartLocalGame.Tooltip.Line2=You can play against a friend sitting besides you (hotseat mode),
+startup.SetupPanelModel.button.StartLocalGame.Tooltip.Line3=or against one of the AIs.
```
RoiEXLab, line 22: I really dislike this separation of lines, it makes it really hard to use it in languages that aren't too similar too english and might have huge differences in sentence length.
Instead I'd suggest making use of `\n` in the `.properties` File and replacing those if `<br>` is required instead. This way the whole system is less dependant on certain language features overall.

I'd recommend having a look at https://projectfluent.org/ they provide some examples of why translating software can be challenging (also IIRC this is the system used in Firefox and other Mozilla Software)

game-app/game-core/src/main/resources/i18n/games/strategy/engine/framework/ui.properties
```diff
@@ -0,0 +1,23 @@
+startup.SetupPanelModel.button.ConnectToNetworkedGame.Label=Connect to Networked Game
+startup.SetupPanelModel.button.EnginePreferences.Label=Engine Preferences
+startup.SetupPanelModel.button.EnginePreferences.Tooltip=Configure certain options related to the engine.
+startup.SetupPanelModel.button.HostNetworkGame.Label=Host Networked Game
+startup.SetupPanelModel.button.MapCreatorTools.Label=Map Creator Tools
+startup.SetupPanelModel.button.PlayByEmail.Label=Play By Email
+startup.SetupPanelModel.button.PlayByEmail.Tooltip=Starts a game which will be emailed back and forth between all players.
+startup.SetupPanelModel.button.PlayByForum.Label=Play By Forum
+startup.SetupPanelModel.button.PlayByForum.Tooltip=Starts a game which will be posted to an online forum or a message board.
+startup.SetupPanelModel.button.PlayOnline.ConnectToNetworkedGame.Tooltip.Line1=Connects to someone's hosted game,
```
RoiEXLab, line 10:
```suggestion
startup.SetupPanelModel.button.PlayOnline.ConnectToNetworkedGame.Tooltip.Line1=Connects to someone''s hosted game,
```
Single quotes need to be escape with another single quote

game-app/game-core/src/main/resources/i18n/games/strategy/engine/framework/ui.properties
```diff
@@ -0,0 +1,23 @@
...
+startup.SetupPanelModel.button.PlayOnline.HostNetworkGame.Tooltip.Line3=Anyone from the internet can connect as well, but only if the host has configured port forwarding correctly. 
+startup.SetupPanelModel.button.PlayOnline.Label=Play Online
+startup.SetupPanelModel.button.PlayOnline.Tooltip.Line1=Find Games Online on the Lobby Server.
+startup.SetupPanelModel.button.PlayOnline.Tooltip.Line2=TripleA is MEANT to be played Online against other humans.
+startup.SetupPanelModel.button.PlayOnline.Tooltip.Line3=Any other way is not as fun!
+startup.SetupPanelModel.button.StartLocalGame.Label=Start Local Game
+startup.SetupPanelModel.button.StartLocalGame.Tooltip.Line1=Start a game on this computer.
+startup.SetupPanelModel.button.StartLocalGame.Tooltip.Line2=You can play against a friend sitting besides you (hotseat mode),
+startup.SetupPanelModel.button.StartLocalGame.Tooltip.Line3=or against one of the AIs.
+startup.SetupPanelModel.button.UserGuideHelp.Label=User Guide & Help
```
RoiEXLab, line 23: Nit: Trailing newlines are a good convention

RoiEXLab (approved):
I believe this is a solid approach to this topic. It obviously won't solve all problems that come with internationalization, but I think it's a very decent start to get down this rabbit hole.

game-app/game-core/src/main/resources/i18n/games/strategy/engine/framework/ui.properties
frigoref: Agree, the separation of lines should be in the properties file.
To avoid replace logic, which is not always necessary, I put the <br/> tag in the properties file.

game-app/game-core/src/main/resources/i18n/games/strategy/engine/framework/ui.properties
frigoref: Hm, it works on my computer (Windows 10 with IntelliJ) and should only lead to a proplem

> Whenever you are using MessageFormat you should be aware that the single quote character (') fulfils a special purpose inside message patterns. The single quote is used to represent a section within the message pattern that will not be formatted. A single quote itself must be escaped by using two single quotes ('').

([source link](https://www.mscharhag.com/java/resource-bundle-single-quote-escaping))

Anyhow, I adjusted it.

game-app/game-core/src/main/resources/i18n/games/strategy/engine/framework/ui.properties
frigoref: Adjusted accordingly.

frigoref:
@DanVanAtta 
Could you please also have a look at this pull request?
So far we have RoiEXLab who seems to be fine now with the changes.
If you also agree, I would start with the documentation page.

game-app/game-core/src/main/java/games/strategy/engine/framework/I18nResourceBundle.java
```diff
@@ -0,0 +1,62 @@
...
+   */
+  public static boolean isSupported(final Locale locale) {
+    final Locale[] availableLocales = Locale.getAvailableLocales();
+    return Arrays.asList(availableLocales).contains(locale);
+  }
+
+  /** @return List of supported languages */
+  public static List<String> getSupportedLanguages() {
+    return Arrays.asList(Locale.ENGLISH.getLanguage(), Locale.GERMAN.getLanguage());
+  }
```
RoiEXLab, line 34: No need to return a List of strings here, I think it's better to return `List<Locale>`, and not call `getLanguage()` at all.
This makes a difference for `en-US` and `en-GB` for example, because those are suddenly two different locales. However this would allow us to use a method like https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Locale.html#lookup(java.util.List,java.util.Collection) instead of running our own selection logic here

DanVanAtta:
> @DanVanAtta
Could you please also have a look at this pull request?

I do think we need to focus on a number of high level questions. The changes here might be fine, but is the direction?

Questions:
- how many strings are there approximately in TripleA? What is our desired percentage to have internationalized? How much work will that be? Are we going to target to have every user-display string translated? If no, which kinds of strings will be english-only?
- how will these configurations scale according to the number of strings? I would suspect the number of strings are going to be low thousands, how many config files is that going to be? Will it be manageable? Are we going to have dozens of additional config files that all need to be moved/renamed if we change classes? (This is an example where we are adding quite a bit of cost to relatively routine operations, instead of moving renaming 1 file, we have to update a test file too, and potentially 'n' translation files') 

- how many languages are we looking to have? Is there any evidence this will be more than a nice to have? There is a cost to this capability, do the returns justify those costs (and particularly at this time?). [Side note, I suspect Italian might actually be the best language to add, it is the second highest download per country after the US, & yet it is still really nice that we have several German speakers making that at least easier to have translations]
- If we go full hog on this update, we're going to be blocked pretty quickly by save game compability. What is your plan to address that?
- will we ever have a feature be held up because we need translations? If yes, how can we avoid that? If no, how do ensure that we reasonably keep up with having translations?
- how will we roll this out? It seems odd that simply booting would cause a smattering of strings to be translated but not quite everything. It feels like that would make for an unreleasable state - how can we avoid that so we can do this work incrementally and still have a releasable product after each increment?

Something of a bottom line, this could become a very expensive feature to build & particularly to maintain. We should be sure the value is there, that this is the right thing to do now, and second that we have thought of how to reduce the cost before we push really far along and then find out when we are knee deep that the path we chose was overly expensive (and we should understand how expensive this is before we sign up for it, IMHO the ROI should be a very calculated decision here).

DanVanAtta:
Another item to consider, I would suggest starting with translations of the map and in-game UI. The starting screens have had a rewrite attempt twice now, I strongly suspect they will be completely re-written within the next year.

Another thing that occurred to me, how about map content? Is that going to be weird if the game has translations but the map descriptions and game notes do not? I think a person can probably rely on 'google translate' to help out in a pinch there (I believe the game notes are pretty easily copy/pasted, if not then that would be painful - but is it okay for that to be a thing, or would we want to build something so even game notes can be translated?)

game-app/game-core/src/test/resources/propertyTest.properties
```diff
@@ -0,0 +1 @@
+test.Text=Test Text: Happy Testing
```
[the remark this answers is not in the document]
frigoref: Why is that? What is the value here?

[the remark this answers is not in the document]
RoiEXLab: I'm just hoping that this will add the trailing newline if you apply it, not sure if it works though

game-app/game-core/src/main/java/games/strategy/engine/framework/I18nResourceBundle.java
frigoref: What is the benefit for this change/the method `lookup` that could then be called? Currently, we only want to ensure that all supported languages (English and German in this case) are listed and I do not see how to not call `getLanguage()`. A change to `Locale` seems counterintuitive when there is a 1:n relationship from language to Locale.
 Could you explain this please a bit more?

game-app/game-core/src/test/resources/propertyTest.properties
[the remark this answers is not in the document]
frigoref: I got that the change would add a trailing newline. My question is why should we have/ensure having trailing newlines?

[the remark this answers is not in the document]
RoiEXLab: It's just a general convention that pretty much all files in the repo follow, and is pretty common in the UNIX-like world

game-app/game-core/src/main/java/games/strategy/engine/framework/I18nResourceBundle.java
RoiEXLab: In the end it's all a matter of how many options you want to keep.
If you're fine with having one english (GB, US, CA etc.) translation for all regions in the world where some dialect is spoken then your approach makes perfect sense.
However when staying at a locale-based level, we could have different translations for different regions/countries, but falling back to a translation using the same language but a different region/country if no 100% correct translation is available.
So TL;DR I'd just like to keep an option for more fine grained variety of certain localisations, even if we don't plan on using them right now

game-app/game-core/src/test/java/games/strategy/engine/framework/I18nResourceBundleTest.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  // @Test
+  void getResourcePath() {
+    final I18nResourceBundle i18nBundle = I18nTestBundle.get();
+    assertEquals(resourcePath, i18nBundle.getResourcePath());
+  }
+
+  // @Test
+  void getText() {
```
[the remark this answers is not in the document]
frigoref: Done.

game-app/game-core/src/test/java/games/strategy/engine/framework/I18nResourceBundleTest.java
```diff
@@ -0,0 +1,56 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.Locale;
+import org.junit.jupiter.api.Test;
+
+class I18nResourceBundleTest {
+  private static final String resourcePath =
+      "propertyTest"; // i18n.games.strategy.engine.framework.ui
```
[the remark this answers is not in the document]
frigoref: Done.

game-app/game-core/src/main/resources/i18n/games/strategy/engine/framework/ui_de.properties
```diff
@@ -0,0 +1,23 @@
...
+<br/>=Auch jeder aus dem Internet kann wird sich verbinden k\u00F6nnen, solange der Host Port-Weiterleitung eingestellt hat.
+startup.SetupPanelModel.btn.PlayOnline.Lbl=Spiel Online
+startup.SetupPanelModel.btn.PlayOnline.Tltp=Finde Spiele online auf dem Lobby-Server.\
+<br/>=TripleA ist daf\u00FCr gedacht, online gegen andere Menschen zu spielen.\
+<br/>=Alles andere macht keinen Spa\u00DF!
+startup.SetupPanelModel.btn.StartLocalGame.Lbl=Starte lokales Spiel
+startup.SetupPanelModel.btn.StartLocalGame.Tltp=Start a game on this computer.\
+<br/>=Du kannst gegen einen Freund spielen, der neben Dir sitzt (Hotseat-Modus),\
+<br/>=oder gegen eine der KIs.
+startup.SetupPanelModel.btn.UserGuideHelp.Lbl=Spielanleitung & Hilfe
```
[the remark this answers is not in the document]
frigoref: Done.

game-app/game-core/src/main/resources/i18n/games/strategy/engine/framework/ui.properties
```diff
@@ -0,0 +1,23 @@
...
+<br/>=Anyone from the internet can connect as well, but only if the host has configured port forwarding correctly.
+startup.SetupPanelModel.btn.PlayOnline.Lbl=Play Online
+startup.SetupPanelModel.btn.PlayOnline.Tltp=Find Games Online on the Lobby Server.\
+<br/>=TripleA is MEANT to be played Online against other humans.\
+<br/>=Any other way is not as fun!
+startup.SetupPanelModel.btn.StartLocalGame.Lbl=Start Local Game
+startup.SetupPanelModel.btn.StartLocalGame.Tltp=Start a game on this computer.\
+<br/>=You can play against a friend sitting besides you (hotseat mode),\
+<br/>=or against one of the AIs.
+startup.SetupPanelModel.btn.UserGuideHelp.Lbl=User Guide & Help
```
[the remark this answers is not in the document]
frigoref: Done.

game-app/game-core/src/main/java/games/strategy/engine/framework/GameResourceBundle.java
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  public static String convToString(final Date date) {
+    return dateFormat.format(date);
+  }
+
+  public static void setLocale(final String language, final String country) {
+    currentLocale = new Locale(language, country);
+    numberFormat = NumberFormat.getInstance(currentLocale);
+    // dateFormat = DateFormat.getInstance(currentLocale);
```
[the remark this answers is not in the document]
frigoref: Removed.

game-app/game-core/src/main/java/games/strategy/engine/framework/I18nResourceBundle.java
frigoref: Done.

frigoref:
[i18n_languages.md](https://github.com/frigoref/triplea/blob/i18n_v1/docs/development/how-to/i18n_languages.md) added for documentation taking over the open questions. Please feel free to add some or confirm when you consider them answered.

frigoref:
It's almost 2 weeks since the last feedback. I know @DanVanAtta wanted to take a break from triplea for a while.
@RoiEXLab do you have further comments to the questions from Dan or my answers to them? Do you know anyone else we should consult here?
Is this pull request gets approved I would continue with further string extraction commits.

RoiEXLab:
@frigoref In general I agree with you, but @DanVanAtta raises some good points. If a translation is not maintained at all we might as well just not have it.
However as mentioned already having a rather complete english and german (maybe sometimes behind a week or two) is pretty easy to manage, i would volunteer to keep the german translation up to date. To introduce different locales we'd need to make sure an active member of the community will be maintaining it

frigoref:
The points raised by @DanVanAtta are gathered and so far from my side sufficiently answered (see [i18n_languages.md](https://github.com/frigoref/triplea/blob/i18n_v1/docs/development/how-to/i18n_languages.md)) to go ahead. If not I would like to understand what exactly is still needed to be answered in what detail in order to approve the pull request? Do you have an opinion about that, @RoiEXLab?

For the first step I would suggest to keep the target on having a redesign so we actually _can_ translate and provide the UI in a different language. In the process I would like to also provide German as a first alternative language, but it would be a great help if I can focus more on the extraction (frist step) and someone else, like you, would take over the actual translation.

RoiEXLab:
@frigoref Usually PRs are not merged until all reviewers are satisfied with the result, however because @DanVanAtta might not be responding soon and I see that you put enough thoughts into this topic, so I'll give you the benefit of the doubt and merge this PR under the condition that if @DanVanAtta has some further concerns in the future those will be discussed as well.

asvitkine:
I've not been following super closely, but one option is to have the engine support an external file for providing a translation and to decouple these translations from the main release process.

Then, community can work on and provide translations for a given release that won't block the release.

RoiEXLab:
@asvitkine Using extra files for it is the plan, but @DanVanAtta was mainly referring to the case where a new feature is implemented which requires new translations, but the translations can't keep up with development.
If this happens often enough eventually a 100% translation becomes a 50% translation becomes a 25% translation and so on

frigoref:
@asvitkine Thanks for your feedback!
As @RoiEXLab wrote, the language files are separate from the rest. Furthermore, I have build a test case which is failing in case not all translations of a language we define as a supported language is present. We can thereby see whether a new language is fully available with this test case before we release it for usage.

asvitkine:
I guess my point is a translation would need to be updated for each release. So if v11 is released and a German translation for v11 is not yet posted, then v11 wouldn't support German until that translation is released. Players can stick to v10 if their translation is not yet ready. We don't have to gate the release on those translations existing, they can be provided after the main release.

RaiNova:
@frigoref @DanVanAtta We agreed I'd have a look at it from the programmer's point of view: Is the overhead acceptable? Given the functionality provided by the classes of this PR it is. I'd prefer an easier naming scheme than indicated by the changes in `MetaSetupPanel.java` though: 

1. The prefix - "startup.SetupPanelModel" in this case - should not be necessary in every getText-call. How about providing it as a configuration parameter to `I18nEngineFramework.get()`?
2. Type information like "btn" should only be required where necessary to distinguish between to similar texts.
3. Usage information like "Lbl" and "Tltp" should only be required where necessary to distinguish between to similar texts.
4. There should also be no requirement when to use of upper case / lower case.
5. What remains is typically the content of the English language string. How about not converting it into a (pseudo?) variable name like "UserGuideHelp" but just keeping it as it was, e.g. "User Guide & Help". That way, it can also be used as a default. Then it should also be quite simple to write an IDE macro that wraps the getText call around the current text. 

This would make it easy enough for my brain and my fingers - i.e. minimal additional burden. We used something very similar to that some 23 years ago (in a C++ development environment) and the additional burden for us programmers was negligible.

frigoref:
Hi @RaiNova,
thank you for looking over it and provide your feedback.
Let me try to reply here and in case I fail to clarify your points completely, I suggest we have a call to discuss these further.

1. Prefix: I like the idea not to use prefixes, putting it into a configuration parameter would break the auto-find feature and makes it also more difficult to lookup the value for a certain key as you also need to check the configuration parameter first.
2. I thought about this and I found myself in the dilemma of either having duplicated strings to translate (when using very context sensitive keys) or context-independent translations that might be harder to translate and cannot be changed easily as they affect other area's with a different meaning. But I agree that I might have leaned to much in the direction of context sensitive and can still drop some more distinguishers. It just made it easier to follow the simple rule to always use them, instead of making a decision about their usage each time.
3. I think it is not true, that the same text should be used in these cases. Basically their sameness is already a code smell as e.g. label and tooltip are supposed to be different as they serve different purposes. Maybe the tooltip could be dropped as the label is self-explantory or the option should be given to other languages (where it is not self-explanatory) to provide a more clarifying text, but having them exactly the same does not add any value, does it?
4. What do you mean exactly with "there should be no requirement" (when to use of upper case / lower case)?
5. I like this idea and tried to follow it, but it does not work for longer texts. I agree that the text content should be "encoded" with the key and not so much the detailed context. However, the context cannot be dropped completely. The IDE macro already runs and it always has to be checked and evaluated. There is no automatism that safely converts the strings in a code file into i18n keys as already the check whether strings represent translatable text or not cannot be determined.

RaiNova:
2./3. If there should be different text because of different type (e.g. button vs. menu entry) or usage (e.g. label vs. tooltip), then type info resp. usage info should be added to the lookup string. Otherwise I'd suggest to avoid it.

Mind you the need to have different text can be language specific. What is the editing process for such a case?

> 4. What do you mean exactly with "there should be no requirement" (when to use of upper case / lower case)?

The decveloper should use upper / lower case in the same way as in variable names. This avoids an additional knowledge burden when to use which case.

>5. I like this idea and tried to follow it, but it does not work for longer texts

How about for longer texts the lookup string is the beginning of the text, maybe followed by something like "..."?

> the check whether strings represent translatable text or not cannot be determined.
I had a macro in mind that developer activates by a hotkey. So the developer determins which strings need to be translated.
She can then have a look at the generated lookup string and manually improve it if neccessary.

RaiNova:
@frigoref Some thoughts after our discussion yesterday:
1. Different languages should be in different files. The English text should be in only one file. So translation should bring up two files. This requires a dedicated UI - preferably an IDEA plugin.

2. If the English text is copied over into the resource file of another language (file2) then any way to automatically determine whether text in file2 that is identical to English has been translated or not requires additional data. In other words: In order to avoid additional data, the English text must not be copied over but must remain missing until it is inserted by an (automatic or human) translator => If a text is in file2, then it has been translated. If a text has not been translated, then it is missing in file2.

3. A way to avoid lines like 
```
final var bundle = I18nTestBundle.get();
```
in order to later be as short as 
```
bundle.getString(key);
```

is to introduce a class like 
```
@UtilityClass
public class I18n {
    private var bundle = null;

    public String getString(String key) {
        if (bundle == null) {
            bundle = I18nTestBundle.get();
        }

        return bundle.getString(key);
    }
}
```

Then you can call 
```
I18n.getString(key);
```

(I just typed that into this editor, so there might be small syntax errors, but in essence this should work.)

frigoref:
@RaiNova thanks for your feedback! It is much more motivating if you know somebody else also thinks about the same problems :)
Here my reply to the points you have raised:
> 1. Different languages should be in different files. The English text should be in only one file. So translation should bring up two files. This requires a dedicated UI - preferably an IDEA plugin.
Not sure if I understand the concern here. Each language has a different file according to the i18n standard.
> 2. If the English text is copied over into the resource file of another language (file2) then any way to automatically determine whether text in file2 that is identical to English has been translated or not requires additional data. In other words: In order to avoid additional data, the English text must not be copied over but must remain missing until it is inserted by an (automatic or human) translator => If a text is in file2, then it has been translated. If a text has not been translated, then it is missing in file2.
I fear this is a standard process happening during extraction.
> 3. A way to avoid lines like
> 
> ```
> final var bundle = I18nTestBundle.get();
> ```
> 
> in order to later be as short as
> 
> ```
> bundle.getString(key);
> ```
> 
> is to introduce a class ...
> Then you can call
> 
> ```
> I18n.getString(key);
> ```
> 
> (I just typed that into this editor, so there might be small syntax errors, but in essence this should work.)

Is your suggestion to add the same utility class to each translatable java file or have a separate class for this? If not, how do I distinguish between the different properties files? I am trying not to have everything in just one file to avoid loading to many unneeded key-value pairs.

RaiNova:
> @frigoref  I am trying not to have everything in just one file to avoid loading to many unneeded key-value pairs.

Having multiple sets of files (each set contains one file per languare) adds complexity - which I would avoid, esp. in the situation TripleA is currently in. It can be added later when required.
What disadvantages does loading all the key-value pairs have in 2022? Are we still talking about some 600 entries?

frigoref:
@RaiNova 600 entries for the core based on which I would recommend to start to publish it even though not all would be translated by then. The total number of strings is around 12k. For the core game it is As the loading takes time and we never need all strings, I though to have multiple files to be a good compromise. 
(see https://github.com/frigoref/triplea/blob/i18n_v1/docs/development/how-to/i18n_languages.md)

DanVanAtta:
Until we get to the tens of thousands, I don't think there is a problem.

A couple key questions to consider:
- As a maintainer, how do I know that all languages have complete translations? How would i find strings with missing translations?
- As a contributor, how would I add new translations of newly added strings?

Multiple files I think could be helpful for organizing the translations and keeping the above consistent.

Here is an idea, what if the google translate is powerful enough to almost always give us good enough translations? What if the translations are good enough and we agree that if we can automate translations 100%, then the ability for rare custom translations is something we can simply live without because everything else is automated and a string only needs to be written once.

To do this we could create a daily job that uses the google translate web API to re-translate all strings and commit any differences. Hopefully we would be able to stay within the volume for free pricing: https://cloud.google.com/translate/pricing (so perhaps we run translations every 3 or 4 days)

RoiEXLab:
@DanVanAtta If structured correctly, for editing you can use the built-in IntelliJ resource bundle editor: https://www.jetbrains.com/help/idea/resource-bundle.html

Basically it provides an overview over all the different languages without having to edit the properties files one by one.

frigoref:
Thanks for your feedback @DanVanAtta!
> * As a maintainer, how do I know that all languages have complete translations? How would i find strings with missing translations?

I am currently checking on how to provide this overview. The check for translation would be something like string length > 8 and property value in English = property value in other language -> check fails.
> * As a contributor, how would I add new translations of newly added strings?

Either directly with your commit (there is a guide on how to extract a normal string into the property files) or via a separate pull request with a commit containing only the property file.
> Here is an idea, what if the google translate is powerful enough to almost always give us good enough translations? What if the translations are good enough and we agree that if we can automate translations 100%, then the ability for rare custom translations is something we can simply live without because everything else is automated and a string only needs to be written once.

I am pretty sure that there is no suitable translation possible 100%. We have too little context for the translation and it is quite specific. I prefer the idea of rather showing not translated texts instead of badly auto-translated ones. This could trigger people to participate and leave a clear view on what needs to be translated and what is done.
My assumption is that we need a diff process to identify the changes to the translations (new keys, changed English base values) and some kind of status handling (not translated, auto-translated, manually-translated). But it's not 100% clear to me yet how we should do that.
Currently working on a plugin enhancement to allow mass editing/upload/download...

RaiNova:
> @frigoref As the loading takes time [...]

To minimize the effect load the file upfront, e.g. when the splash screen is showing.
I don't know how it is done in TripleA, but probably the splash screen is shown until the program is completely loaded.
If this is the case, then you can load at startup with 

```
@UtilityClass
public class I18n {
    private var bundle = I18nTestBundle.get();

    public String getString(String key) {
        return bundle.getString(key);
    }
}
```

DanVanAtta:
Generally google translate has good translations. imo if it is good enough
95% of the time, using that automatically will save a lot of development
time

I understand how to physically add translations, though if a string changes
or is added, then there is a maintenance cost to clear the string do that
someone else can translate. How does a developer add strings for 'N'
languages? Im6yhinking 3 years down the road and perhaps we have italian in
the mix and maybe additional languages. It seems we could be building
towards a large maintenance cost if all strings are manually translated

It would possibly frustrating too for the existing (and often poor texts)
are translated and then changed and repeatedly re-translated, which would
seem to be a lot of churn and probably frustrating for translators

DanVanAtta:
Consider, let's say we get to 4 languages. Many updates change strings.
That implies after many updates, post merge we would need to coordinate
updates from the translators and have some sort of queue that contains
changing work, and good luck for all that to be prompt. It's a goal as well
to toggle the prerelease versions to just be straight releases. Missing or
invalid translations seemingly are bad, so we would be frequently in a
break and fix mode. I think we need automation here, we will never have a
full set of full time translators.

PR #10001, opened 2022-01-09
Added New World Order Variants

RogerCooper:

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2472,3 +2472,13 @@
   img: https://raw.githubusercontent.com/triplea-maps/NWO_Andrews/master/preview.png
   description: |
    <br>1939 on New World Order map with low piece count
+- mapName: New_World_Order_Variants
+  mapCategory: EXPERIMENTAL
+  url: https://github.com/triplea-maps/New_World_Order_Variants/archive/master.zip
+  version: 1
+  img: https://raw.githubusercontent.com/triplea-maps/New_World_Order_Variants/master/preview.png
+  description: |
+   <br>1939 on New World Order map with the neutral nations of Spain, Turkey, Middle East, Skandinavia, Poland, Yugoslavia & Greece. There are 3 scenarios.
+   <br>New World Order Variant - Neutrals Unassigned, the neutral nations can be attacked by all, but they will build and move within their own territories.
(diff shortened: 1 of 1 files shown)


Cernelius:
I understand that what you already added is this repository:
https://github.com/triplea-maps/New_World_Order_Variants
However, New World Order already had a number of variants, in this other repository:
https://github.com/triplea-maps/nwo_variants
Beside other specific variants of the same map, namely:
https://github.com/triplea-maps/new_world_order_lebowski_edition
https://github.com/triplea-maps/new_world_order1915lebowski

So, are you sure you want to have a map called "New_World_Order_Variants" when you already have a map called "nwo_variants".

By the way, what are the current naming conventions for maps? Where are they documented? For example, I seem to remember it was mandatory all being lower case, is it not now?

Cernelius:
Here "nwo" is lower case:
https://github.com/triplea-maps/nwo_variants
and here it is upper case:
https://github.com/triplea-maps/NWO_Andrews

RogerCooper:
I had submitted this as "New_World_Order_Variants". I know the casing is inconsistent, but I am generally posting other people's mods with minimal changes and I usually follow the style used by the mod creator. Feel free to make any changes you need to bring this into the repository.

RogerCooper:
It might be better to just rename this "New_World_Order_Neutrals".

These general "*variants" mods have some issues. Can I an add a scenario into someone's else's "variants" mod?  Just because someone has downloaded a mod with some variants, doesn't mean they want every variant.

Another problem with 2.6 is that I need to copy over all the images, rather than having an XML-only mod that refers back to the original mod and gives an opportunity to download it.

beelee1:
@RogerCooper  it says the yaml formatting failed. I think Cernel is right as the only thing I can find different is the upper case. The map name obviously can be upper case and you don't need the underscores, at least i don't see any other maps done that way.

maybe rename repo and try lower case

RogerCooper:
I had forgotten to publishe the mod before adding to the yaml. That may have caused the failure. It is published now.

Cernelius:
I'm just mostly saying that, in the moment TripleA has a map called "nwo_variants", if one wants to add more variants of New World Order, it seems logical to me to add such variants to the mentioned map, rather than adding a map called "New_World_Order_Variants". If the variants are too special, so they cannot be bundled together, I would say they deserve a more specific name, instead.

Unfortunately, I don't believe there are guide-lines about what makes a map derived from an other map a "variant".

PR #10003, opened 2022-01-10
Stochastic battle calculator

RaiNova:


Cernelius:
I suggest keeping the current battle-calculator beside this new one but renaming the current one as "battle-simulator".

It should have been always named as such, as it never was a calculator (beside making the arithmetic mean of all results of its simulations): just a untility running a specified number of simulations.

RaiNova:
> I suggest keeping the current battle-calculator beside this new one but renaming the current one as "battle-simulator".
> It should have always been named as such, as it never was a calculator (beside making the arithmetic mean of all results of its simulations): just a utility running a specified number of simulations.

I'd like to discuss renaming a few other classes with some developers as well. Do you know when Dan will be back?

RaiNova:
@DanVanAtta I'd like to propose this as the basis for two discussions: 
a) how to do ui related testing (not in the way this PR is currently using, but how)?
b) is Kotlin an acceptable second programming language?

DanVanAtta:
@RaiNova ; I see quite a lot more to this update compared to last time, wow!

Are you still wanting the discussions (a) & (b) above on testing & kotlin?

Were you able to fire up an AI with the calculator to observer game speed differences?

PR #10051, opened 2022-01-28
Extraction of common calls for SwingUtilities.isEventDispatchThread

frigoref:
EDT-checks outsourced into new methods in Util.java:
-ensureEventDispatchThread
-ensureNoneEventDispatchThread

Maybe someone experienced in this thread stuff could write also a test method or we leave them without as they are short.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/ui/Util.java
+++ b/game-app/game-core/src/main/java/games/strategy/ui/Util.java
@@ -139,13 +139,13 @@ public final class Util {
     return translatedPolygon;
   }
 
-  public static void ensureEventDispatchThread() {
+  public static void ensureOnEventDispatchThread() {
     if (!SwingUtilities.isEventDispatchThread()) {
       throw new IllegalStateException("Wrong thread (not Event Dispatch Thread)");
     }
   }
 
-  public static void ensureNoneEventDispatchThread() {
+  public static void ensureNotOnEventDispatchThread() {
     if (SwingUtilities.isEventDispatchThread()) {
       throw new IllegalStateException("This method should not called from Event Dispatch Thread");
     }
--- a/game-app/game-core/src/main/java/games/strategy/ui/Util.java
+++ b/game-app/game-core/src/main/java/games/strategy/ui/Util.java
@@ -16,6 +16,7 @@ import java.awt.Toolkit;
 import java.awt.Window;
 import java.awt.geom.GeneralPath;
 import java.awt.image.BufferedImage;
+import javax.swing.SwingUtilities;
 import lombok.experimental.UtilityClass;
 
 /** A collection of methods useful for rendering the UI. */
@@ -136,5 +137,17 @@ public final class Util {
         new Polygon(polygon.xpoints, polygon.ypoints, polygon.npoints);
     translatedPolygon.translate(deltaX, deltaY);
     return translatedPolygon;
+  }
+
+  public static void ensureEventDispatchThread() {
+    if (!SwingUtilities.isEventDispatchThread()) {
+      throw new IllegalStateException("Wrong thread (not Event Dispatch Thread)");
+    }
+  }
+
+  public static void ensureNoneEventDispatchThread() {
+    if (SwingUtilities.isEventDispatchThread()) {
+      throw new IllegalStateException("This method should not called from Event Dispatch Thread");
+    }
   }
 }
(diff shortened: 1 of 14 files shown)


RoiEXLab (changes requested):

game-app/game-core/src/main/java/games/strategy/ui/Util.java
```diff
@@ -137,4 +138,16 @@ public static Polygon translatePolygon(
...
+    if (!SwingUtilities.isEventDispatchThread()) {
+      throw new IllegalStateException("Wrong thread (not Event Dispatch Thread)");
+    }
+  }
+
+  public static void ensureNoneEventDispatchThread() {
+    if (SwingUtilities.isEventDispatchThread()) {
+      throw new IllegalStateException("This method should not called from Event Dispatch Thread");
+    }
+  }
```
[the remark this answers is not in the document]
frigoref: Test methods added

game-app/game-core/src/main/java/games/strategy/ui/Util.java
```diff
@@ -137,4 +138,16 @@ public static Polygon translatePolygon(
...
     return translatedPolygon;
   }
+
+  public static void ensureEventDispatchThread() {
+    if (!SwingUtilities.isEventDispatchThread()) {
+      throw new IllegalStateException("Wrong thread (not Event Dispatch Thread)");
+    }
+  }
+
+  public static void ensureNoneEventDispatchThread() {
```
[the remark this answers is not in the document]
frigoref: Renamed to ensureOnEventDispatchThread and ensureNotInEventDispatchThread.

game-app/game-core/src/test/java/games/strategy/ui/UtilTest.java
```diff
@@ -2,12 +2,37 @@
...
 import org.junit.jupiter.api.Test;
 
 final class UtilTest {
+  @Test
+  void ensureOnEventDispatchThread() {
+    new Thread(
+            () ->
+                assertAll(
+                    () -> SwingUtilities.invokeAndWait(() -> Util.ensureOnEventDispatchThread())))
+        .start();
```
RoiEXLab, line 20: Wrapping assert statements inside a new thread will cause the test to pass even if the assertion fails.
Just remove the new Thread() part around it and it should work as expected. Same for the other method

game-app/game-core/src/test/java/games/strategy/ui/UtilTest.java
```diff
@@ -2,12 +2,58 @@
...
+    threadNotEventDispatchThread.start();
+    try {
+      threadNotEventDispatchThread.join();
+    } catch (final InterruptedException e) {
+      inThreadException = e;
+    }
+    if (inThreadException != null) {
+      throw inThreadException;
+    }
+  }
```
RoiEXLab, line 41: Ok this definitely works, but I'm curious why you explicitly chose this approach?
Is there any reason
```
void exampleTest() {
  assertDoesNotThrow(() -> SwingUtilities.invokeAndWait(Util::ensureOnEventDispatchThread));
}
```
does not work as expected?

If there is a specific reason, I'd recommend using a local `AtomicReference<Throwable>` variable instead of the `inThreadException` field

frigoref: If you are asking for the reason of using an instance variable inThreadException and separate method that is executing the thread and rethrowing any exception, then I can answer you that it seemed more general so we can specifically check for the expected exception.
Thanks for the hint with the AtomicReference I will see to incorporate this asap.

frigoref: Done.

game-app/game-core/src/test/java/games/strategy/ui/UtilTest.java
```diff
@@ -8,35 +8,37 @@
...
 import java.awt.Polygon;
 import java.lang.reflect.InvocationTargetException;
+import java.util.concurrent.atomic.AtomicReference;
 import javax.swing.SwingUtilities;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 
 final class UtilTest {
-  Throwable inThreadException;
+  AtomicReference<Throwable> inThreadException;
```
RoiEXLab, line 17: Now this can be a local variable instead of a field

game-app/game-core/src/test/java/games/strategy/ui/UtilTest.java
RoiEXLab: I was mainly wondering why you have to spawn a new thread in the first place, to my knowledge you can just call `invokeAndWait` directly?

game-app/game-core/src/test/java/games/strategy/ui/UtilTest.java
frigoref: ... or removed completely when using `invokeAndWait` directly and not in a separate thread :)
I could clean out a lot with this hint. Thanks!

RoiEXLab (approved):
Perfect, thanks a lot for your patience

PR #10112, opened 2022-02-13
Added CTF2

RogerCooper:
Removed the description entirely.

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2481,3 +2481,17 @@
    <br>Scenarios with many unit types on the Big World map
    <br>Hearts of Iron: The Rising Sun 1942, starts in early 1942
    <br>Hearts of Iron: Awakening the Giants, starts in mid 1941
+- mapName: CTF2 (Capture the Flag)
+  mapCategory: EXPERIMENTAL
+  url: https://github.com/triplea-maps/CTF2/archive/master.zip
+  version: 1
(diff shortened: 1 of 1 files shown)


RoiEXLab:
@RogerCooper I have the impression so seem to struggle a little bit with the formatting tool, so here's some advice.
The only thing wrong with your PRs is that the files have too many trailing newlines. There should be a single trailing newline at the end and that's it. To be honest it's a bit hard to spot in the error message and I didn't notice it myself first, but once you know it it becomes really obvious.
Here is how it should look like in the online editor:
Screenshot
Just a single newline after the last line of text.

Also there's no need to create a new PR for every attempt. You can just go to the "Files changed" Tab and edit the files there using the online editor.

DanVanAtta:
@RoiEXLab  thank you . @RogerCooper  , please combine all open map requests and submit as one; That will be the easiest way most likely to get everything submitted  (and avoids merge conflicts)

RogerCooper:
Sorry, I was experimenting on trying to get anything in. Now that I know I need 1 blank line at the end, I should not have further problems.

RogerCooper:
DanVanAtta, any progress on fixing the download issue with 2.6? Right now, I am only working in 2.5

DanVanAtta:
> DanVanAtta, any progress on fixing the download issue with 2.6? Right now, I am only working in 2.5

Yes, some. If you use 2.6 and explicitly use the 'prerelease' server via the 'settings > testing' menu, downloads should work.

PR #10122, opened 2022-02-18
Added 270BC Variants

RogerCooper:

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2489,4 +2489,3 @@
   description: |
    <br>270BC Cernel Mod
    <br>270BC Tertiered - Units and production values divided by 3
-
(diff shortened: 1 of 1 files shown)


RogerCooper:
I left a blank line at the end. It still did not work.

Cernelius:
I actually had a map called "270bc_variants" deleted from the repository some time ago. The map had only two games, both made by me. If I recall correctly, I made that request and @DanVanAtta accepted it. I've not changed my mind.

RogerCooper:
The Tertiered scenario looked interesting. However, my main concern here is the formatting error I received. I don't see what is wrong the way I entered the data.

Cernelius:
> The Tertiered scenario looked interesting. However, my main concern here is the formatting error I received. I don't see what is wrong the way I entered the data.

That is substantially the old (and more aggressively scaled down) version of "270BC 40%", which went through a series of versions too.

I have preferred removing from the repository all my variants of "270BC" but "270BC Wars" in order to let any interested player focus on that.

RogerCooper:
I will remove those files.

DanVanAtta:
Merge conflict needs to be resolved now before this can move forward @RogerCooper 
Let me know if you need help doing so

PR #10123, opened 2022-02-18
Feign Common Test coverage and minor corrections

bacrossland:

- Added test coverage for Feign Common
- Corrected the misspelling of success on a method
- Updated comments of HttpClient.java to define better what the timeouts are for. 
- Removed the int multiplication being done from timeout constants. Moved what it was trying to convey to the comments.


DanVanAtta:
Running: `./gradlew spotlessApply` should fix the formatting

For: `[ant:checkstyle] [WARN] /home/runner/work/triplea/triplea/http-clients/feign-common/src/test/java/org/triplea/http/client/GenericServerResponseTest.java:3:47: Using the '.*' form of import should be avoided - org.junit.jupiter.api.Assertions.*. [AvoidStarImport]`

The clean-up import option should be able to resolve that. If not then IDEA has a config for when it uses start imports.

http-clients/feign-common/src/test/java/org/triplea/http/client/AuthenticationHeadersTest.java
```diff
@@ -0,0 +1,27 @@
...
+import org.junit.jupiter.api.Test;
+import org.triplea.domain.data.ApiKey;
+
+class AuthenticationHeadersTest {
+  final ApiKey apiKey = ApiKey.newKey();
+  final AuthenticationHeaders authenticationHeaders = new AuthenticationHeaders(apiKey);
+
+  @Test
+  void createHeaders() {
+    assertTrue(authenticationHeaders.createHeaders() instanceof HashMap);
```
DanVanAtta, line 15: Why is it important that the return value be a `HashMap` and not say a `Map`? If the other test assertions managed to pass and this one failed, would the code be incorrect?


Nit:
- conventionally we use hamcrest assertions. IIRC there is an 'instanceof' check (If think `IsInstanceOf.instanceOf`) that can be used to make this nicer. A first priority is often to get tests in place, a next prioirty is to make it easy to debug failing tests (this makes the next developers life much easier and avoids the situation where small changes wind up spending hours fixing tests). So.. using the hamcrest match would help becuase it would state the type of instance that was found vs expected, vs here we would have just a simple yes/no and no further information if it were to fail.

http-clients/feign-common/src/test/java/org/triplea/http/client/GenericServerResponseTest.java
```diff
@@ -0,0 +1,33 @@
...
+    assertFalse(genericServerResponse.isSuccess());
+    assertNull(genericServerResponse.getMessage());
+  }
+
+  @Test
+  void isSuccess() {
+    GenericServerResponse genericServerResponseTrue =
+        GenericServerResponse.builder().success(true).build();
+    GenericServerResponse genericServerResponseFalse =
+        GenericServerResponse.builder().success(false).build();
```
DanVanAtta, line 21: Take a second look for if we need the 'failure' test case. We might be dangerously close to testing lombok, though it seems this case is already handled in the `builder` case above. The only difference between this one and the other is seemingly library code and JVM defaults (which we don't really need to test necessarily).

DanVanAtta (commented):
I like the changes! Thank you.
- conventionally hamcrest assertions are used instead of junit assertions. Please let me know if that is a concern or if it is difficult ot migrate

http-clients/feign-common/src/test/java/org/triplea/http/client/AuthenticationHeadersTest.java
bacrossland: Cool. I'll switch that over to hamcrest. 

I mess that up regarding the`HashMap`. I saw the `new HashMap<>()` in the method that was in my mind. I'll switch it to `Map` because that is what the method says it will return.

http-clients/feign-common/src/test/java/org/triplea/http/client/GenericServerResponseTest.java
bacrossland: I was going for coverage but I see your point that the builder would handle it. I'll remove it.

bacrossland:
@DanVanAtta I've made the changes you suggested. It's ready for your review again.

bacrossland:
Not a concern or difficult to migrate to. It's a more declarative test case and I'm loving the more readable errors it produces.

DanVanAtta:
:+1: 

Thank you!

Issue #8168, opened 2020-11-14
2.5.22294: TileImageFactory#isImageLoaded:129 - java.lang.NullPointerException


tripleabuilderbot:
## User Description

I just enabled and then disabled the "Show Map Blend" option from the menu. Thanks and Regards.

## Map
world_war_ii_revised

## TripleA Version
2.5.22294

## Java Version
11.0.6

## Operating System
Mac OS X

## Memory
Heap utilization statistics [MB]
Used Memory: 98
Free memory: 1949
Total memory: 2048
Max memory: 2048

## Stack Trace
```
Exception: java.lang.NullPointerException
java.lang.Exception
	at games.strategy.triplea.image.TileImageFactory.isImageLoaded(TileImageFactory.java:129)
	at games.strategy.triplea.image.TileImageFactory.getImage(TileImageFactory.java:150)
	at games.strategy.triplea.image.TileImageFactory.getBaseTile(TileImageFactory.java:137)
	at games.strategy.triplea.ui.screen.drawable.BaseMapDrawable.getImage(BaseMapDrawable.java:17)
	at games.strategy.triplea.ui.screen.drawable.MapTileDrawable.draw(MapTileDrawable.java:39)
	at games.strategy.triplea.ui.screen.Tile.draw(Tile.java:69)
	at games.strategy.triplea.ui.screen.Tile.drawImage(Tile.java:54)
	at games.strategy.triplea.ui.panels.map.MapPanel.lambda$paint$7(MapPanel.java:742)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)


```

PR #10135, opened 2022-02-23
Fix NPE in TileImageFactory when toggling blended map mode.

asvitkine:

Fixes https://github.com/triplea-game/triplea/issues/8168 by checking whether the SoftReference is null.


RoiEXLab (approved):
Nice, thanks for fixing this

PR #10136, opened 2022-02-23
Fix PBEM To field to not delete spaces when typing two addresses.

asvitkine:
Fix PBEM To field to not delete spaces when typing two addresses.

## Release Note

FIX|PBEM To field issue input of multiple space-separated emails.


RoiEXLab (approved):
👍🏼

PR #10142, opened 2022-02-24
Keep player lists and select save if a game ends due to a disconnect.

asvitkine:
Keep player lists and select save if a game ends due to a disconnect.

With this change, if a networked game ends because a joined player disconnects, the
game lobby dialog that will be re-opened for the server is now improved in these ways:
  - The save game that was produced on disconnect will now be selected.
  - The countries / players set up from the ended game will be kept.
     This includes which players have selected which countries and the AI settings
     for the host. Countries that were played by the disconnected player will be cleared
     and be opened to be chosen.
  - Additionally, this PR fixes some NPEs that could happen on disconnect that I discovered
     in my testing.

This makes it much easier to continue the game when the disconnected player re-connects.

- No UI changes, this just sets up the existing UI on a game disconnect.
- I've tested on Revised, running three copies of TripleA on my computer, with
  different countries being played by different players, including AIs.

## Release Note

CHANGE: If a joined player disconnects, the game lobby will now select the auto-saved game and persist the player list from the previous game.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/ServerLauncher.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/ServerLauncher.java
@@ -255,10 +255,10 @@ public class ServerLauncher implements ILauncher {
    * example, a disconnected participant will cause the game to be stopped, while a disconnected
    * observer will have no effect.
    */
-  public void connectionLost(final INode node) {
+  public void connectionLost(final INode droppedNode) {
     if (isLaunching) {
       // this is expected, we told the observer he couldn't join, so now we lose the connection
-      if (observersThatTriedToJoinDuringStartup.remove(node)) {
+      if (observersThatTriedToJoinDuringStartup.remove(droppedNode)) {
         return;
       }
       // a player has dropped out, abort
@@ -268,12 +268,12 @@ public class ServerLauncher implements ILauncher {
     }
     // if we lose a connection to a player, shut down the game (after saving) and go back to the
     // main screen
-    if (serverGame.getPlayerManager().isPlaying(node)) {
+    if (serverGame.getPlayerManager().isPlaying(droppedNode)) {
       if (serverGame.isGameSequenceRunning()) {
-        saveAndEndGame(node);
-        // Release any countries played by this player. This ensures the slots are freed up in the
-        // game lobby as we keep the other player mappings for resuming from the saved game.
-        for (final String countryName : serverGame.getPlayerManager().getPlayedBy(node)) {
+        saveAndEndGame(droppedNode);
+        // Release any countries played by the player that dropped. The other seating assignments
+        // are preserved for the game lobby that will be shown.
+        for (final String countryName : serverGame.getPlayerManager().getPlayedBy(droppedNode)) {
           serverModel.releasePlayer(countryName);
         }
         serverModel.persistPlayersToNodesMapping();
--- a/game-app/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/ServerLauncher.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/ServerLauncher.java
@@ -271,6 +271,12 @@ public class ServerLauncher implements ILauncher {
     if (serverGame.getPlayerManager().isPlaying(node)) {
       if (serverGame.isGameSequenceRunning()) {
         saveAndEndGame(node);
+        // Release any countries played by this player. This ensures the slots are freed up in the
+        // game lobby as we keep the other player mappings for resuming from the saved game.
+        for (final String countryName : serverGame.getPlayerManager().getPlayedBy(node)) {
+          serverModel.releasePlayer(countryName);
+        }
+        serverModel.persistPlayersToNodesMapping();
       } else {
         stopGame();
       }
@@ -294,6 +300,7 @@ public class ServerLauncher implements ILauncher {
     final Path f = launchAction.getAutoSaveFile();
     try {
       serverGame.saveGame(f);
+      gameSelectorModel.setSaveGameFileToLoad(f);
     } catch (final Exception e) {
       log.error("Failed to save game: " + f.toAbsolutePath(), e);
     }
--- a/game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
@@ -61,6 +61,7 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import javax.swing.JFrame;
 import javax.swing.JOptionPane;
@@ -99,6 +100,7 @@ public class ServerModel extends Observable implements IConnectionChangeListener
   private Messengers messengers;
   private GameData data;
   private Map<String, String> playersToNodeListing = new HashMap<>();
+  private boolean playersToNodesMappingPersisted = false;
   private Map<String, Boolean> playersEnabledListing = new HashMap<>();
   private Collection<String> playersAllowedToBeDisabled = new HashSet<>();
   private Map<String, Collection<String>> playerNamesAndAlliancesInTurnOrder =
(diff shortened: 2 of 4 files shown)


Cernelius:
If I'm understanding this correctly, this is going to be a relief for one-versus-one games and a great aid for multiplayers (3 or more players) and especially for FFA.

I wonder if this is only for normal hosts or also for automated ones (bots), since automated ones already reload the connection lost save on disconnection (whereas normal hosts do not).

Cernelius:
An idea may be that when, in a normal host, the hoster itself closes a game in progress (going back to the waiting for players screen), you get a prompt asking you whether or not you want to have this behaviour (you get the behaviour at this PR if you say yes and get the current behaviour if you say no). This may be useful in multiplayers (and especially FFA) when you reload to substitute one of the player with another one or anyway partially remake the assigniation of the powers in order to continue the game. However, the program should not ask you and auto-select "no" if you achieved a victory condition and chose not to continue the game.

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
```diff
@@ -313,36 +315,58 @@ public synchronized void setLocalPlayerType(final String player, final PlayerTyp
...
 
+  /**
+   * Persists the players mappings to re-use upon a game data change. Used to persist the previous
+   * game's player setting for the game restart if a connection is lost.
+   */
+  public void persistPlayersToNodesMapping() {
+    this.playersToNodesMappingPersisted = true;
+  }
+
+  private void updatePlayersOnGameDataChanged(final GameData data) {
```
DanVanAtta, line 326: Nit: private methods should be defined after they are used rather than before their usage: https://github.com/triplea-game/triplea/blob/master/docs/development/reference/code-conventions/java-code-conventions.md#depth-first-method-ordering

To follow depth-first ordering, this method should be moved to be lower in the class file after it's first usage (but not necessarily to the bottom of the class). FWIW: This convention is to help build a consistent expectation of where code is located and to help aid readers to read the class file in one single top-to-bottom scan.

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
```diff
@@ -313,36 +315,58 @@ public synchronized void setLocalPlayerType(final String player, final PlayerTyp
...
+    // If specified, keep the previous player data.
+    if (playersToNodesMappingPersisted) {
+      playersToNodesMappingPersisted = false;
+      final Set<String> dataPlayers =
+          data.getPlayerList().stream().map(GamePlayer::getName).collect(Collectors.toSet());
+      if (dataPlayers.equals(playersToNodeListing.keySet())) {
+        // List of countries matches, keep the persisted mappings.
```
DanVanAtta, line 333: Could you explain a bit more why the list of countries matching is enough to use the 'cached' values for 'playersToNodeListing' etc? Why is it that we know we can skip the additional initialization steps?

From what my (faulty memory) recalls is that we have a lot of parallel data structures that create some very brittle scenarios where if data is one, it is assumed to be in the other. I'm concerned that there could be some room for error there (as we are working with a very brittle design here)

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/ServerLauncher.java
```diff
@@ -271,6 +271,12 @@ public void connectionLost(final INode node) {
     if (serverGame.getPlayerManager().isPlaying(node)) {
       if (serverGame.isGameSequenceRunning()) {
         saveAndEndGame(node);
+        // Release any countries played by this player. This ensures the slots are freed up in the
+        // game lobby as we keep the other player mappings for resuming from the saved game.
+        for (final String countryName : serverGame.getPlayerManager().getPlayedBy(node)) {
+          serverModel.releasePlayer(countryName);
+        }
```
DanVanAtta, line 278: What are your thoughts if we changed the initial seat allocation code to not occupy seats by the host? Is there a complexity there, are you striving for avoiding changing existing behavior?

I like the idea of simply not allocating those seats if it means we can avoid this for loop entirely. This code relies on an ordering of execution that could be brittle. Thereby if we moved the seat allocation code to execute after this, we would have an error (which also speaks to the overall complexity of this whole scenario, I acknowledge working through that would have been difficult! There is a lot of state a person needs to grok to follow the server launcher logic, it's not easy)

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
asvitkine: The if statement on line 332 is mostly a sanity check.

The expectation is if `playersToNodesMappingPersisted` was set, we're exactly in this "restart" scenario and therefore we should use the persisted data.

The if just guards against some hypothetical code path where we did set that bool, but then the map changed before this code was run. It might actually be impossible - so one option is to make it throw an exception instead.

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/ServerLauncher.java
asvitkine: Sorry, I'm not following this comment.

This logic is not about "seats occupied by the host" - this is looping through seats occupied by the player who just dropped. Basically, they dropped, so clear their seats.

Previously, this wasn't needed because a restart would just re-initialize the seats. I guess maybe you're asking if we could just change that code instead to not do that? That seems like a reasonable behavior change to me, but seems orthogonal to this change.

And more specifically, even if that were the case, then we'd still need to have this loop to clear seats when someone drops.

I guess one thing such a change could do is make it so we don't need to have a special `playersToNodesMappingPersisted` codepath - because we would just always persist mappings and only clear them when a map changes.

I think that could be an option, but would be a bigger change.

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
DanVanAtta: Thanks for the context, I'll want to look at this in more detail. Please allow a few days for more to look at this deeper.

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/ServerLauncher.java
DanVanAtta: Oh, I was confused about what was happening. I thought `node` was the host. In `ServerLauncher` and `ServerModel`, there is a reference to the 'host node' that is typically referred to as just 'the node'. (Which is all kinda crazy naming terminology, but c'est la vie).

It seems we can bring some additional clarity here. This would be a great comment on the for loop: "this is looping through seats occupied by the player who just dropped. Basically, they dropped, so clear their seats."

Second, renaming `node` to something like `nodeThatDropped` or something like. (More ideally we'd call it `playerThatDropped` but unfortunately the code naming is f%#@d up and 'players' are called 'nodes'. So presumably that terminology should stay consistent in the commentary too).

asvitkine: Done, renamed to `droppedNode` (to keep more concise that your suggestion which would have caused extra wrapping) and updated the comment to make it clearer.

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
asvitkine: Thanks, done.

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
asvitkine: Any more thoughts on this? I'm open to us thinking of other approaches long term, but this is still a major improvement over the status quo, so perhaps we can just merge this?

DanVanAtta: Sorry, I have been wanting to take a really deep look at this but have been unable to find the time. 

I wonder, is this code "trivially" correct? (in other words, this is always the case and so these checks are going to be no-ops in all cases) If so, should we remove it? On another side of the coin, is this a good safe guard for data integrity where we should throw an error? On another side, if this is not always the case, is there a nicer way of detecting that we should early exit.

I'm interested in having this PR be merged sooner rather than later. We move forward now, I'm not sure we'll come back. If we could spend some time on these thoughts and try to finish this PR entirely in the next day or two, that would be ideal. I think this is a case where you having write privileges would be ideal @asvitkine  as you could then merge after addressing these comments to the best of your ability.

asvitkine: I don't think it's "trivially" correct, because the code is pretty convoluted - so it's hard to tell if this can be hit or not.

On the other hand, my intuition is that it's likely not possible. But I don't have a way to prove that. One option is we go with the exception approach and see if it sticks (e.g. whether it ever gets hit). If it does, then we update the logic.

If it doesn't, we leave it in as a data integrity check to guard against something changing in the future and breaking the functionality. How does that sound?

DanVanAtta: To recap, the plan would be:
- throw an exception
- see if during further prerelease testing we see the problem and then go from there

I can get on board with that. I'm a bit wary that we do enough testing to see the problem or not. If we can get that testing done, then this seems solid.

Assuming we do not 'hit this code', would we remove this check before cutting a release, or maintain it as an exception? It might be nice to do a silent fallback and send an error report, though the complexity of that does not seem worth it to me.

DanVanAtta (approved):

asvitkine:
@DanVanAtta Can you take another glance? I just added one more commit that fixes some NPEs in ActionButtons.java that I discovered during my testing that may happen on disconnect.

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
asvitkine: I think keeping the exception is helpful because it can help discover a regression if some future refactoring changes the flow such that the assumption no longer holds.

game-app/game-core/src/main/java/games/strategy/triplea/ui/ActionButtons.java
```diff
@@ -255,27 +278,40 @@ public UserActionAttachment waitForUserActionAction(
...
+    }
     return battlePanel.waitForBattleSelection();
   }
 
   public Tuple<Territory, Set<Unit>> waitForPickTerritoryAndUnits(
       final List<Territory> territoryChoices,
       final List<Unit> unitChoices,
       final int unitsPerPick) {
+    if (pickTerritoryAndUnitsPanel == null) {
+      return Tuple.of(null, Set.of());
```
DanVanAtta, line 313: Are the usages of this null safe? Is there a trade-off to just return simply `null`?

PR #10188, opened 2022-03-14
Fix NPE in calculateBombingRoutes().

asvitkine:
The previous logic could result in a move being constructed with a null route. This change updates the logic to not create a move in that case.

Fixes https://github.com/triplea-game/triplea/issues/10183.

## Release Note

FIX|Game can crash resulting from AI bombing logic


ron-murhammer (approved):
Gonna merge this since it should avoid the crash. But begs the question why a null route would ever try to be created here. Probably means there is some other bug in the AI somewhere.

PR #10197, opened 2022-03-18
Switch macOS install4j artifact to app bundle.

asvitkine:
The previous setting was an Installer w/ macOS single bundle, which is not recommended:
https://www.ej-technologies.com/resources/install4j/help/doc/concepts/mediaFiles.html

This should result in changing the TripleA release for macOS to be a .dmg containing the app bundle, rather than containing an installer that installs the app bundle. Since the TripleA app bundle is self contained, this is preferable and also recommended by install4j.

## Release Note

CHANGE|macOS releases no longer use an installer.

### DIFF
--- a/game-app/game-headed/build.install4j
+++ b/game-app/game-headed/build.install4j
@@ -804,9 +804,7 @@ return true;</property>
     <windows name="Windows 64 bit" id="87" mediaFileName="${compiler:sys.fullName}_${compiler:sys.version}_windows-64bit">
       <jreBundle usePack200="false" directDownload="true" />
     </windows>
-    <macos name="Mac OS X" id="35" compressDmg="true" launcherId="33">
-      <jreBundle usePack200="false" directDownload="true" />
-    </macos>
+    <macosArchive name="macOS Single Bundle Archive" id="557" launcherId="33" />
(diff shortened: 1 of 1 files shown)


asvitkine:
Note: I generated this change via the install4j GUI, but I'm not able to test it as I'm getting an unrelated error:

```
Cause: com.exe4j.b.z
       When trying to process '${compiler:sys.version}': recursive replacement of variable 'sys.version'. Replacement stack: sys.version
```

asvitkine:
> Note: I generated this change via the install4j GUI, but I'm not able to test it as I'm getting an unrelated error:
> 
> ```
> Cause: com.exe4j.b.z
>        When trying to process '${compiler:sys.version}': recursive replacement of variable 'sys.version'. Replacement stack: sys.version
> ```

OK, I resolved this and was able to test that this is working correctly. To resolve, I had to manually set the version number to something that's not a variable under General Settings. I'm not sure why it's set the way it is - do we have something else that replaces it there?

Anyway as far as this change is concerned, I confirmed it produced a .dmg file with the TripleA directly in it, as intended.

ron-murhammer (approved):

PR #10209, opened 2022-03-20
Add gradle jvmargs to allow spotlessApply to work with newer Java.

asvitkine:
Add gradle jvmargs to allow spotlessApply to work with newer Java.

Without this, running spotlessApply hits an error.

See https://github.com/diffplug/spotless/issues/834.


RoiEXLab:
Thanks for having a look at this

PR #10220, opened 2022-03-22
Improve code health of PlayerAttachment.

asvitkine:

Improve code health of PlayerAttachment.

- Stop using deprecated Triple class
- Make a helper function for repeated code
- Use local variables to reduce wrapping

Many of these issues are identified by Code Climate:
https://codeclimate.com/github/triplea-game/triplea/game-app/game-core/src/main/java/games/strategy/triplea/attachments/PlayerAttachment.java/source#issue-ce7383fb8c1afa54231b9722856166f2

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/attachments/PlayerAttachment.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/attachments/PlayerAttachment.java
@@ -19,9 +19,9 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Predicate;
 import org.triplea.java.collections.CollectionUtils;
 import org.triplea.java.collections.IntegerMap;
-import org.triplea.util.Triple;
 
 /**
  * An attachment for instances of {@link GamePlayer} that defines properties unrelated to rules (see
@@ -52,14 +52,15 @@ public class PlayerAttachment extends DefaultAttachment {
   private IntegerMap<Resource> suicideAttackResources = new IntegerMap<>();
   // what can be hit by suicide attacks
   private Set<UnitType> suicideAttackTargets = null;
+
   // placement limits on a flexible per player basis
-  private Set<Triple<Integer, String, Set<UnitType>>> placementLimit = new HashSet<>();
+  private Set<StackingLimit> placementLimit = new HashSet<>();
 
   // movement limits on a flexible per player basis
-  private Set<Triple<Integer, String, Set<UnitType>>> movementLimit = new HashSet<>();
+  private Set<StackingLimit> movementLimit = new HashSet<>();
 
   // attacking limits on a flexible per player basis
-  private Set<Triple<Integer, String, Set<UnitType>>> attackingLimit = new HashSet<>();
+  private Set<StackingLimit> attackingLimit = new HashSet<>();
 
   public PlayerAttachment(final String name, final Attachable attachable, final GameData gameData) {
     super(name, attachable, gameData);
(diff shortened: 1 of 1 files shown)


asvitkine:
I just realize that this will make saved games incompatible. @DanVanAtta - is this OK or should we hold off on a change like this?

game-app/game-core/src/main/java/games/strategy/triplea/attachments/PlayerAttachment.java
```diff
@@ -165,19 +114,47 @@ private void resetMovementLimit() {
...
+
+  private Set<StackingLimit> getAttackingLimit() {
+    return attackingLimit;
+  }
+
+  private void resetAttackingLimit() {
+    attackingLimit = new HashSet<>();
+  }
+
+  private static class StackingLimit {
```
DanVanAtta, line 132: (observation) A quick comment on this data type might be helpful to clarify what exactly 'StackingLimit' means. 
I agree it can be induced relatively easily by checking the data types.

Second observation, the 'all args constructor' can be omitted and replaced with lombok `@AllArgsConstructor`. Sometimes 'getters' can be nice to have (for example in function references & lambda expressions), in this case `@Value` would be a good choice. An additional benefit is if we ever chose to extract this class to its own file, there would be no updates needed. All that is to say, this is a good candidate class to have `@Value` on it.

DanVanAtta (approved):

DanVanAtta:
> I just realize that this will make saved games incompatible. @DanVanAtta - is this OK or should we hold off on a change like this?

I just read the above

- Do you know why the 'save-game compatibility test' has not caught this? Could you look into it, if you can.
- We do need to avoid breaking save game compatibility. When we do break compatibility, it would likely be for the 3.0 initiative and a lot more could change and hopefully become far simpler.

DanVanAtta:
@asvitkine if save game compatibility is broken, are there any reasonable to update this PR to maintain save game compat? If no, are there any parts here that can be readily salvaged that you would be willing to salvage? (this does seem like a nice update)

DanVanAtta:
Okay, sorry to have commented and then answer my one questions shortly later.

I reviewed the update and see the type changes would likely break compatibility. The save game test likely does not exercise anything that loads the PlayerAttachment, and hence it is a gap in that test.

Fixing this completely in a backward compatible way is seemingly nasty.  To do a backward compatible fix, seemingly we could remove the `Triple` data type from the method signatures. Overall that is not pretty.

I'm happy you brought up the save game issue, as this would have been a major problem @asvitkine . I am disappointed that such a nice cleanup has run afoul of that constraint.

DanVanAtta (changes requested):
Marking request changes   : (
Don't want an accidental merge until we can get save game compat.

asvitkine:
@DanVanAtta Actually, I'm not sure anymore whether this breaks save games. Maybe it doesn't!

I noticed that changing members of UnitAttachment definitely breaks save games as part of my other PR (so I kept the types the same there). So then I assumed this does also.

But seems maybe it doesn't? I tried loading some save games with this PR loaded and they seemed to load fine.

So that may mean that PlayerAttachment is not in the object graph in the save game, but UnitAttachment is. Do you know if that's plausible and what's the explanation for why they're different?

Unless I'm just not testing on any maps that have player attachments with these fields set (assuming that if the field is null, the serialization logic doesn't touch it and doesn't run into errors).

What's the root object of the save game? Perhaps I can try to do some reachability queries from a heap dump.

asvitkine:
Replying to myself.

This PR does indeed have a problem with saved games. But, in a different way than I expected, I'm still a bit confused...

So you definitely need a map that exercises this functionality, otherwise things seem to be fine. One such map is "Pacific Theater Solo Challenge".

Playing this map with the PR and trying to save a game fails because the new type isn't serializable (but Triple evidently was). Surprising that nothing caught this!

But another surprise is that if I save a game on that map with the current prerelease, I'm able to load it with this PR. I don't understand why - I would have expected it would have resulted in an incompatible type error, because we'd tried to load a Triple but the field is now of a different class. Very strange...

DanVanAtta:
Type erasure might be at play (if I'm quoting that term correctly). At runtime, Java drops all of the templated types and hence things like `Set<T>` are stored executed as type `Set` (implicitly `Set<Object>`). Serialization though is kinda evil, so it may very well be deeper than that, and the fact you have observed a problem @asvitkine  at all indicates that something is at play here.

Having a compatibility test that is able to catch this problem would be very valuable. It's been a pretty big burden historically to verify save game compatibility, the more of that we can automate (with reasonable effort) - the nicer.

> For example, perhaps we can have a smoke test that tries to save a game with every available map - since we have one that downloads them already.

The general idea here sounds good!  I have been thinking to tone down the "download & parse" test to be a bit more selective on the maps (so perhaps instead of downloading around 90 maps, we download around 30'ish). There is a trade off to be had between coverage & test execution speed. Enhanced coverage is greatly valuable. Going beyond 30'ish maps might be diminishing returns though in terms of test execution speed and increased coverage (assuming that those 30'ish are well representative of all XMLs & rules that we have configured in any map)

DanVanAtta:
> Playing this map with the PR and trying to save a game fails because the new type isn't serializable (but Triple evidently was). Surprising that nothing caught this!

Could this be fixed by adding `@Serializable` to the class (and then implementing the implicit contract, of a version ID, hashcode & equals? `@Value` provides those implementations, so perhaps it might be a better of adding a few annotations and a version id)

PR #10230, opened 2022-03-25
Add smoke test that downloads some maps and verifies game saving.

asvitkine:

This test downloads several maps (the list can be easily expanded) and start the game on each and tries doing a game save.

This test is able to catch the problem introduced by the pending PR https://github.com/triplea-game/triplea/pull/10220 which introduces a non-serializable type into the game save object graph. Note: We should probably expand the list of maps being tested so that a variety of features are exercised. This initial PR just adds two maps.

### DIFF
--- a/smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
+++ b/smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
@@ -21,7 +21,7 @@ import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
-import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.CsvSource;
 import org.triplea.config.product.ProductVersionReader;
@@ -30,13 +30,17 @@ import org.triplea.injection.Injections;
 import org.triplea.io.ContentDownloader;
 import org.triplea.io.FileUtils;
 
+/**
+ * Checks that no error is encountered when saving a game on several different maps. This test
+ * downloads the maps in question and starts an all-AI game on each of them, before saving.
+ *
+ * <p>Note: A variety of maps are used to ensure different engine features are exercised since
+ * object serialization may run into errors that depend on the state of the object graph.
+ */
 class GameSaveTest {
 
-  @BeforeEach
-  public void setUp() throws IOException {
-    if (Injections.getInstance() != null) {
-      return;
-    }
+  @BeforeAll
+  public static void setUp() throws IOException {
     Injections.init(
         Injections.builder()
             .engineVersion(new ProductVersionReader().getVersion())
--- a/smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
+++ b/smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
@@ -76,11 +76,7 @@ class GameSaveTest {
     try (ContentDownloader downloader = new ContentDownloader(uri)) {
       Files.copy(downloader.getStream(), targetTempFileToDownloadTo);
     }
-
-    final Path[] mapFolderPath = new Path[1];
-    ZippedMapsExtractor.unzipMap(targetTempFileToDownloadTo)
-        .ifPresent(installedMap -> mapFolderPath[0] = installedMap);
-    return mapFolderPath[0];
+    return ZippedMapsExtractor.unzipMap(targetTempFileToDownloadTo).get();
   }
 
   private static URI getMapDownloadURI(final String mapName) throws URISyntaxException {
--- a/smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
+++ b/smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
@@ -70,7 +70,7 @@ class GameSaveTest {
     final Path saveFile = Files.createTempFile("save", GameDataFileUtils.getExtension());
     game.saveGame(saveFile);
     assertNotEquals(Files.size(saveFile), 0);
-    assertThat(Files.size(saveFile), is(not(0));
+    assertThat(Files.size(saveFile), is(not(0)));
   }
 
   private static Path downloadMap(final URI uri) throws IOException {
@@ -81,10 +81,7 @@ class GameSaveTest {
 
     final Path[] mapFolderPath = new Path[1];
     ZippedMapsExtractor.unzipMap(targetTempFileToDownloadTo)
-        .ifPresent(
-            installedMap -> {
-              mapFolderPath[0] = installedMap;
-            });
+        .ifPresent(installedMap -> mapFolderPath[0] = installedMap);
     return mapFolderPath[0];
   }
 
--- a/smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
+++ b/smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
@@ -21,7 +21,9 @@ import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
-import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
 import org.triplea.config.product.ProductVersionReader;
 import org.triplea.game.server.HeadlessLaunchAction;
 import org.triplea.injection.Injections;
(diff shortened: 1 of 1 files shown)


smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
```diff
@@ -0,0 +1,110 @@
...
+    ClientFileSystemHelper.setCodeSourceFolder(tempFolder);
+  }
+
+  @ParameterizedTest
+  @CsvSource({
+    "world_war_ii_revised,map/games/ww2v2.xml",
+    "pacific_challenge,map/games/Pacific_Theater_Solo_Challenge.xml"
+  })
+  void testSaveGame(final String mapName, final String mapXmlPath) throws Exception {
+    final Path mapFolderPath = downloadMap(getMapDownloadURI(mapName));
```
DanVanAtta, line 62: Side-note, the 'optional' `final` modifier can be omitted now (unless dealing the a class variable where we should prefer to have it). I'm hoping that for new code we can remove the unnecessary final modifier and stop including it on local variables and parameters unless there is a specific benefit or reason for doing so.

smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
```diff
@@ -0,0 +1,110 @@
...
+    final GameSelectorModel gameSelector = new GameSelectorModel();
+    gameSelector.load(mapFolderPath.resolve(mapXmlPath));
+    final ServerGame game = startGameWithAis(gameSelector);
```
DanVanAtta, line 65: Is the game ever stopped Does it need to be stopped?

smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
```diff
@@ -0,0 +1,110 @@
...
+    final Path saveFile = Files.createTempFile("save", GameDataFileUtils.getExtension());
+    game.saveGame(saveFile);
+    assertNotEquals(Files.size(saveFile), 0);
```
DanVanAtta, line 68: Please favor using hamcrest assertions for consistency with other tests, eg:
```
MatcherAssert.assertThat(Files.size(saveFile), is(not(0));
```
We also tend to use static imports as well, the fully qualified `MatcherAssert` above is to clarify where the assertion is coming from.

smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
```diff
@@ -0,0 +1,110 @@
...
+import org.triplea.config.product.ProductVersionReader;
+import org.triplea.game.server.HeadlessLaunchAction;
+import org.triplea.injection.Injections;
+import org.triplea.io.ContentDownloader;
+import org.triplea.io.FileUtils;
+
+/**
+ * Checks that no error is encountered when saving a game on several different maps. This test
+ * downloads the maps in question and starts an all-AI game on each of them, before saving.
+ *
```
DanVanAtta, line 36: It is odd that we would be unable to save games that are loaded from a clean XML and latest code. Could you add more detail to this comment about how exactly saving a game can fail?

smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
```diff
@@ -0,0 +1,110 @@
...
+    Injections.init(
+        Injections.builder()
+            .engineVersion(new ProductVersionReader().getVersion())
+            .playerTypes(PlayerTypes.getBuiltInPlayerTypes())
+            .build());
+    ClientSetting.initialize();
+    final Path tempFolder = FileUtils.newTempFolder();
+    FileUtils.writeToFile(tempFolder.resolve(".triplea-root"), "");
+    Files.createDirectory(tempFolder.resolve("assets"));
+    ClientFileSystemHelper.setCodeSourceFolder(tempFolder);
```
DanVanAtta, line 53: I presume these lines were needed because it did not work out of the box. The `ClientFileSystemHelper` should be able to find the 'assets' folder without needing this.

Are we compensating for a bug here (which I think https://github.com/triplea-game/triplea/pull/9945 would fix)?

smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
asvitkine: I'm actually not sure - the test passes fine as is, so it seems it doesn't need to be stopped. But I'm actually not sure that this starts the main game loop where AIs are making moves, I think it just initializes it to the started state.

smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
asvitkine: Added more details.

smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
asvitkine: This code is creating a temporary directory to be used for the "triplea root", as opposed to using some existing folder. So it needs to set it up to the expected state, which requires the file above and this folder. (Without these, we hit errors.)

I'm not sure it's related to the issue you link.

smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
asvitkine: Ack. I'll keep as-is for the PR, but will keep in mind for the future.

smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
DanVanAtta: The way TripleA is packaged compared to how it is run locally is somewhat complicated. A lot of places in code (for better or worse) want to know the absolute path of the "triplea root". Those locations then assume a specific folder structure from there. The general idea if that folder structure would be the same in development as it would be in the installation folder. For various reasons this concept has become less & less true over time and more difficult to ensure. Notably the flaws like the 'root' folder depending on the current folder are far more exposed.

Long story long, the '.triplea-root' folder is meant to designate that 'some' folder. That mechanism was created so that the algorithm to find the 'root' folder is to check the current folder and then recurse upwards until we find that specific touch file. That then indicates our 'root' folder.

Why does this matter? There are a lot of assumptions about the root folder. It's a problem to require multiple locations to all set that up correctly, rather it's better if we can do it once & then have all other code properly use the same root folder. Hence, the work here to create a temp folder, create an assets folder (etc..) is duplicating other functionalities that already exist. What this could lead to is having many places that are all setting up the 'root' folder. It is perhaps clear as well that simply by exercising additional functionality there could be more requirements added to this folder setup (which would make the problem go larger in a non-trivial way).

Now that #9945 is merged; I don't know if this setup code is necessary. If it is, then I wonder why the above algorithm for finding the correct 'root' folder is not working.

asvitkine: #9945 doesn't change anything. The game logic still requires a directory that has the ".triplea-root" file and an "assets" folder, which is what this is setting up. I rebased just to make sure and confirmed the test fails if I remove any of these three lines.

In terms of such a folder already existing that we can use - I know `game-app/game-headed` has a ".triplea-root" file and I think there's some build step that generates an "assets" folder there. But that build step isn't in the build graph for `:smoke-testing:test`, which I'm sure could be changed via gradle dependencies.

But even that's not enough, because even before that, the error we first get is:
```
No service available of type 'org.triplea.game.ApplicationContext'
org.triplea.util.ServiceNotAvailableException: No service available of type 'org.triplea.game.ApplicationContext'
	at org.triplea.util.Services.lambda$loadAny$0(Services.java:22)
	at java.base/java.util.Optional.orElseThrow(Optional.java:408)
	at org.triplea.util.Services.loadAny(Services.java:22)
	at games.strategy.engine.ClientFileSystemHelper.getCodeSourceFolder(ClientFileSystemHelper.java:64)
	at games.strategy.engine.ClientFileSystemHelper.getRootFolder(ClientFileSystemHelper.java:38)
```

I'm sure we there's a way to set up an application context that will be registered as part of this test and have its location point to the right thing and also to change the gradle build dependencies so that the assets folder will get created as a pre-req before running smoke tests.

But this would result in a much more complicated set up that these three lines of code as part of the test and the necessary changes would be disjoint from the test itself which is not a great design.

DanVanAtta: Needed a few days to come back to this. I am concerned that the three lines in question would become boiler plate and your note about adding a dependency on the task to download assets would be a way to solve. Thank you for posting that 'service' discovery error, that would be hard to resolve and I hope that mechanism can be removed when bots are converted to network relays. I'd agree the better options are therefore not super-feasible.

smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
asvitkine: @DanVanAtta 

So after checking this test in the context of https://github.com/triplea-game/triplea/pull/10220, which should cause a failure, I was surprised that it didn't.

Turns out your suggestion wasnot correct.

`assertNotEquals(Files.size(saveFile), 0);` was fine.

But it's not equivalent to:

`assertThat(Files.size(saveFile), is(not(0));`

The above passes when the result is indeed 0.

Turns out the type is important, and so to fix it, we have to do:

`assertThat(Files.size(saveFile), is(not(0L)));`

Note the `0L`.

I can make that change, but it makes me question why are we recommending a framework that has such foot guns?

smoke-testing/src/test/java/games/strategy/engine/data/GameSaveTest.java
asvitkine: One more thought: If we end up with multiple tests needing this - and therefore this becoming "boilerplate", we could always make a test utility method for this that they all call, so as to keep the logic in just one place. Anyway, I think it's moot for now.

DanVanAtta:
The long vs int type is brittle. There is a longer discussion here and we
coming from a place where the API was just two objects.

The hamcrest answer is that it's an anti-pattern to call a method on the
test object. Instead should be something like `assertThat(file,
is(not(emptyfile()))`, or something like that, which would be type safe and
give details like the file name and size of not true.

asvitkine:
> The long vs int type is brittle. There is a longer discussion here and we coming from a place where the API was just two objects. The hamcrest answer is that it's an anti-pattern to call a method on the test object. Instead should be something like `assertThat(file, is(not(emptyfile()))`, or something like that, which would be type safe and give details like the file name and size of not true.

If we're really settled on hamcrest, is there any checker that could flag this sort of error?

It seems like it should be possible to statically check statement like `assertThat(X, is(not(Y));` to ensure X and Y are of the same type. If they're not, then presumably the check will always be false.

I assume that's the behavior of hamcrest where the statement would end up checking objects for equality and presumably Integer.of(0) is not equal to Long.of(0) ... assuming it operates on boxed objects and doesn't have overloads for primitives.

asvitkine:
I found the issue on hamcrest and commented there:

https://github.com/hamcrest/JavaHamcrest/issues/220#issuecomment-1085586998

The reply was:

> Ouch. Not sure what we can do about that. In meantime, I tend to use the JUnit assertNotEquals() for simple comparisons.

If that's coming from a member of the hamcrest team, maybe we should adopt that here too?

ssoloff:
> The hamcrest answer is that it's an anti-pattern to call a method on the test object. Instead should be something like `assertThat(file, is(not(emptyfile()))`, or something like that, which would be type safe and give details like the file name and size of not true.

Indeed. Since Hamcrest 2.0 there is [`FileMatchers#aFileWithSize()`](http://hamcrest.org/JavaHamcrest/javadoc/2.2/org/hamcrest/io/FileMatchers.html#aFileWithSize-long-).

asvitkine:
> Indeed. Since Hamcrest 2.0 there is [`FileMatchers#aFileWithSize()`](http://hamcrest.org/JavaHamcrest/javadoc/2.2/org/hamcrest/io/FileMatchers.html#aFileWithSize-long-).

Thanks! It's too bad it operates on File objects and not on Paths, so we have to do:

```assertThat(saveFile.toFile(), is(not(aFileWithSize(0))));```

(And thus is still "calling a method on the object"). But switched to that in https://github.com/triplea-game/triplea/pull/10253.

PR #10234, opened 2022-03-26
Rolling multiple dice at once

frigoref:
## Rolling multiple dice at once
[db3967b ](https://github.com/frigoref/triplea/commits/diceRollMulti) diceRollMulti
Drawing x numbers in the range of 1 until base is compressed to use minimal number of drawing random numbers:
The introduced map of base numbers and its base^x results avoids repeated computation of these powers.
Each result represents the limit of a range (1-base^x) of all possible combinations of x numbers between 0 and the base-1 (base range).
Therefore, drawing one random number in this range corresponds to drawing x random numbers
in the base range.

For a die with 6 sides, up to 11 rolls can be calculated by one random integer (maximum 2.147.483.647) number (note logarithm y-scale):
graphDice6sides

@DanVanAtta EndTurnDelegate.java has a comment that I don't fully understand and where I would ask you to elaborate further on what the actual error was that was solved with this. Maybe we can try to avoid the sleep-statement.
commentEndTurnDelegate

PlainRandomSource.java
- getRandom: Allow multi-random-draw with one random number

PlainRandomSourceTest.java
- getRandomMany_ShouldReturnRequestedCountOfValues: test also for count = 41 (odd number and > 1)
- getRandomMany_Performance: Added (fails if slow)
- getRandomMany_MultipleThreads: Added ( fails if synchronization missing)

ObjectiveDummyDelegateBridge.java
- getRandom: Simplification for initial int array


DanVanAtta:
Interesting set of changes   = D
I'm not sure when I will have a chance to review in better detail, hopefully tomorrow night or pretty soon thereafter.

Re: that comment - I don't recall writing that. Particularly the part about getting a ton of random numbers. That strikes me as something that @veqryn might have written (see question in description)

DanVanAtta:
@frigoref what is the impact of this update? Did you notice any performance or memory gains?

frigoref:
@DanVanAtta The pact is a better performance for large battles with a lot of units for with the same hit point.
I measured about 5-40% performance gain depending on the number of dice.
I have not checked the memory, but assume either an improvement or little change.

frigoref:
@DanVanAtta Why did 3 checks fail now after I performed Fetch & Merge from the master again?

game-app/game-core/src/main/java/games/strategy/triplea/ui/ObjectiveDummyDelegateBridge.java
```diff
@@ -85,13 +85,12 @@ public Player getRemotePlayer() {
...
     }
-    final int[] numbers = new int[count];
-    for (int i = 0; i < count; i++) {
-      numbers[i] = getRandom(max, player, diceType, annotation);
-    }
-    return numbers;
+    // Oracle documentation: "Each class variable, instance variable, or array component
+    // is initialized with a default value when it is created (§15.9, §15.10)
+    // ... For type int, the default value is zero, that is, 0."
+    return new int[count];
```
asvitkine, line 93: Why is this returning an array of all 0's instead of random numbers?

game-app/game-core/src/main/java/games/strategy/engine/random/PlainRandomSource.java
```diff
@@ -20,9 +23,37 @@
...
+          final int powerResult = basePowers.getPowerResult(diceIdOffset);
+          final int powerFit = randomInt / powerResult;
+          numbers[diceIdBase + diceIdOffset] = powerFit;
+          randomInt -= powerFit * powerResult;
+        }
+        diceIdBase += powerCur;
+        restCount -= powerCur; // newly remaining random numbers to be drawn
+      }
+    } else {
+      for (int i = 0; i < count; i++) {
```
asvitkine, line 54: If this is in the else, then why do we need a loop? Is it because it could be 0? If so maybe add a comment, as otherwise it seems we can just do `numbers[0] =` without a loop.

game-app/game-core/src/main/java/games/strategy/engine/random/PlainRandomSource.java
```diff
@@ -20,9 +23,37 @@
...
+        // Note: x is diceIdOffset
+        for (int diceIdOffset = powerCur - 1; diceIdOffset >= 0; --diceIdOffset) {
+          final int powerResult = basePowers.getPowerResult(diceIdOffset);
+          final int powerFit = randomInt / powerResult;
+          numbers[diceIdBase + diceIdOffset] = powerFit;
+          randomInt -= powerFit * powerResult;
+        }
+        diceIdBase += powerCur;
+        restCount -= powerCur; // newly remaining random numbers to be drawn
+      }
```
asvitkine, line 52: Is there a benchmark that shows that this is faster? Given there's a bunch of extra logic and complexity...

Or is the main benefit that getRandom() may use the network and we reduce the number of network calls? If so, that can be improved by introducing a getRandomNumbers(count) API that can return a bunch of random numbers...

game-app/game-core/src/main/java/games/strategy/triplea/ui/ObjectiveDummyDelegateBridge.java
asvitkine: Oh, because this is a Dummy class?

Anyway, I'm not sure a comment quoting the Oracle docs is needed, you can just say "Return an array of all 0's."

I think it's reasonable to expect people writing Java code to know what the default values of variables are.

game-app/game-core/src/main/java/games/strategy/engine/random/PlainRandomSource.java
asvitkine: Actually, there's a check on line 24 about > 0, so you should be able to remove the loop.

Maybe you could even just do:

```
if (count == 1) {
  return new int[1] { getRandom(max, annotation) };
}
```

Early in the function, to avoid the other block being indented.

game-app/game-core/src/main/java/games/strategy/triplea/ui/ObjectiveDummyDelegateBridge.java
frigoref: Sure? That was not certain in the past if I remember correctly and you asked yourself.

game-app/game-core/src/main/java/games/strategy/engine/random/PlainRandomSource.java
frigoref: Check out the test that has been added. It is faster even without network benefits

asvitkine: I tried it on my machine (M1 MacBook Air) and I'm not seeing conclusive results.

For example, I did two runs (one with the new code and another with the if changed to always be false) and modified the test to print endTime - startTime and got the following:
37960750
37266916

These seem close enough that it doesn't seem like the extra complexity might be worth it?

asvitkine: I modified the test to run both versions (via a parameter on the function) and added an outer loop to do it 100 times (measuring still the inner loop time). This way, we can get JIT to stabilize. Here's what I'm seeing for the last few runs:

Normal 1966334
Batch 1999584
Normal 2012250
Batch 2044959
Normal 1965208
Batch 2128500
Normal 1959833
Batch 2030709
Normal 2017333
Batch 2052708
Normal 2049250
Batch 2082334
Normal 2134750
Batch 2064209
Normal 2134000

So it's a bit of a toss up.

Here's my changes:
https://pastebin.pl/view/raw/26feb56b

Are your results different? I'm curious if running on x86 shows something else.

asvitkine: I also wonder if some of the inner loop logic can be modified to speed up your new code. For example, using bit operators and avoiding division and multiplication (e.g. just using bit masks / shifts).

DanVanAtta:
@frigoref would you mind giving more data behind the claim:

> @DanVanAtta The pact is a better performance for large battles with a lot of units for with the same hit point.
I measured about 5-40% performance gain depending on the number of dice.

Have you tried this in multiplayer? Is there any change to AI performance as it moves and simulates battles?

When it comes to performance changes, my go-to considerations are:
- it's better to optimize correct code than it is to correct optimized code
- most performance benefits are from system design, not optimizing algorithms
- always measure, find the slowest part, change, then measure again
- this blog post, always keep in mind orders of magnitude of difference: https://blog.codinghorror.com/the-infinite-space-between-words/  , while we might make one thing 90% faster, if that thing is running in nano time and then we do an unnecessary network request, we have not really moved the needle.

game-app/game-core/src/main/java/games/strategy/engine/random/PlainRandomSource.java
frigoref: Fixed in [diceRollMulti #3](https://github.com/frigoref/triplea/commit/1823884632318c08ef434e8627a7bf0a8a38e09c)

game-app/game-core/src/main/java/games/strategy/engine/random/PlainRandomSource.java
frigoref: @asvitkine I did the test measurements outside the method (with parameter switch) from a test method multiple times and mostly got for 5-45% performance improvement, but a few time the old way was faster. My assumption was that the time measurement was affected by other programs running.
I agree with @DanVanAtta , that we should test carefully, but do not know how to do that "correctly". The new Java Profiler might be a good start, but we might have to find an agreement between each other what is a valid test first. It does not make sense if I do tests and everyone else has to rebuild and repeat them. Is there the option of test cases that are not regularly executed with the other test cases, but only for performance tests?

game-app/game-core/src/test/java/games/strategy/engine/random/PlainRandomSourceTest.java
```diff
@@ -41,8 +43,31 @@ void getRandomSingle_ShouldThrowExceptionWhenMaxIsNotPositive() {
...
   void getRandomMany_ShouldReturnRequestedCountOfValues() {
-    assertThat(plainRandomSource.getRandom(MAX, 1, ANNOTATION).length, is(1));
-    assertThat(plainRandomSource.getRandom(MAX, 42, ANNOTATION).length, is(42));
+    for (int count : new int[] {1, 41, 42}) {
+      assertThat(plainRandomSource.getRandom(MAX, count, ANNOTATION).length, is(count));
+    }
+  }
+
+  @Test
+  void getRandomMany_Performance() {
```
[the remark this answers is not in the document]
frigoref: How would you go about this? There is no invalid option for a test to fail as also the case were all dice have the same number is valid. There would have to be multiple runs and I was trying not to introduce too much to the automated test overload.

frigoref:
> @frigoref would you mind giving more data behind the claim:
> 
> > @DanVanAtta The pact is a better performance for large battles with a lot of units for with the same hit point.
> > I measured about 5-40% performance gain depending on the number of dice.
The claim was base on the logical argument that "large battles with a lot of units for with the same hit point" benefit more from the reduce number of calls for a random number in conjunction with the measurement results that showed a higher performance gain for larger numbers.
 
> Have you tried this in multiplayer? Is there any change to AI performance as it moves and simulates battles?
 I did not try multiplayer nor did I test AI performance. I just verified the plausible assumption that the drawing random numbers is more challenging than multiplication and division (division could probably even be replaced by subtraction, but also at a cost of simplicity of the code).

game-app/game-core/src/main/java/games/strategy/engine/random/PlainRandomSource.java
DanVanAtta: There is a PerfTimer class that can be used to time methods or specific lines of code. That and/or good old fashioned stop watch (or wall clock). To check this, set up a large battle and then start checking for time differences.

I like the idea of this change, but we do need to be able to concretely the benefits for the "extra" complexity. For example, the performance test that was written, how does that compare with the old algorithm vs the new?

As far as memory usage, I suspect it is not significant. TripleA keeps a full res image of the entire map in memory just to shrink it down & render the minimap - which is in addition to the memory used to actually render the map. So all that is to say is that we can mostly focus on wall-clock time for performance measurements.

DanVanAtta: >  how does that compare with the old algorithm vs the new?

Sorry, I realize @asvitkine did exactly this. I wonder where the difference is, can we run this test in some sort of setup that shows a strong difference?

DanVanAtta:
@frigoref your response might have been cut-off in the last message. Please check


I like the idea of this change. For any time we add complexity, we need to justify it - so there needs to be a clear and demonstrable benefit. I'm eager to observe (or find) that benefit. I'll see if I can run some benchmarks, but I hate for anyone to wait for me to do anything... In the meantime, the more setup & benchmark results that we can gather and post -  the better.

Re: performance tests

The performance tests are good to have on one hand, but on another no unit test should ever take more than 10ms to run (if there are 2000 unit tests, and if they run up to 100ms even, it's a very slow test suite!). Though, I do think there are some cases, particularly around AI where performance is more sensitive, where having performance tests would be really appropriate. This suggests that there should be a 'performance test suite' that is run outside of the main unit tests (basically in the same way that 'smoke tests' are their own build step, 'performance checks' would be its own build step). So notably developers would by default run just the unit tests, and then could opt-in to run the performance checks (though on every PR and every build, we run the performance checks to verify there is no regression).

RE: overall trade-off of complexity vs performance

If the issue is simplicity vs performance, we can also try to do things to make this code less complex as well. Basically make it so any future maintainer would be very comfortable understanding and modifying the code with less than 1 minute of required comprehension time.

I do think we should have some absolute measurements of the performance time. 50ms of runtime is not worth even 1 minute of additional time for maintainers to comprehend the code. The cost of code is largely in its lifetime of maintainers reading it to understand it, many times over (and that time is extremely limited! and hence hella valuable)

game-app/game-core/src/test/java/games/strategy/engine/random/PlainRandomSourceTest.java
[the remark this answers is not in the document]
frigoref: @asvitkine Do you consider this discussion closed for this PR?

game-app/game-core/src/main/java/games/strategy/engine/random/PlainRandomSource.java
frigoref: 2 Result (New vs. Old): 35 : 15
3 Result (New vs. Old): 48 : 2
4 Result (New vs. Old): 46 : 4
5 Result (New vs. Old): 48 : 2
6 Result (New vs. Old): 49 : 1
7 Result (New vs. Old): 50 : 0
8 Result (New vs. Old): 50 : 0
9 Result (New vs. Old): 50 : 0
10 Result (New vs. Old): 50 : 0
11 Result (New vs. Old): 50 : 0
12 Result (New vs. Old): 50 : 0
13 Result (New vs. Old): 50 : 0
14 Result (New vs. Old): 50 : 0
15 Result (New vs. Old): 50 : 0
16 Result (New vs. Old): 50 : 0
17 Result (New vs. Old): 50 : 0
18 Result (New vs. Old): 50 : 0
19 Result (New vs. Old): 50 : 0

Each line is a test for x (first number) number of random numbers that was run 100 times (to get a better average time) and compared with 50 times between the new enhanced version and the old, being a simple call to the method `public int getRandom(final int max, final String annotation)`. The ratio between the two is shown at the end of each line.
I deduce from the output that I pasted above that the new enhanced version is better on average in all cases and gaining complete dominance over the old version when the number of random numbers is high (increasing dominance the more numbers are calculated).

The overall runtime of this test was 283ms (for ca. 20*100*2*50 = 400.000 random numbers in total) and therefore I doubt that drawing number is a general performance issue. The reason why I started this PR was a theory that the new version should be faster. This is true, but the effect seems to be small. 
I would therefore suggest to close this PR discussion as 

1. we have bigger fishes to fry (performance gain is small)
2. the results show a benefit and the increased complexity is acceptable from my point of view

@DanVanAtta, @asvitkine: What do you guys say?

game-app/game-core/src/main/java/games/strategy/triplea/ui/ObjectiveDummyDelegateBridge.java
frigoref: @asvitkine Is this discussion done for you?

game-app/game-core/src/main/java/games/strategy/engine/random/PlainRandomSource.java
asvitkine: Can you share the code you used to test to produce the above output? I'm not exactly following - for example, I don't understand what a "ratio" of "50 : 0" means.

asvitkine: Thanks! I tried your code and the results are similar for me.

I also did a sum of the two runtimes to see not just how often the new way is faster, but also what's the actual difference.
I'm seeing about a 60% improvement of the new algorithm. Of course, this is over the specific set of dice amounts being measured by the test - which is probably not representative of a real game.

However, the dice rolling is very fast either way. I had to bump the `testRuns` value to 1000 to get even to ~1/10 exec time for each of the algorithms. So with 1000 `testRuns`, we roll 18,900,000 dice total, half using the old algo and half using the new, and the total runtime for the old algo is 150ms vs. 96ms with the new.

So even with the old algo, it's able to roll 126 million dice / sec.

So from the above, my vote is to abandon this change, because dice rolling is already very fast and the extra complexity is not worth the improvement in speed in this area.  (Although it may be interesting to check how many dice are rolled in in practice for an AI-heavy game where there's lots of simulations being done - though that's also multithreaded.)

As an aside, if we did want to pursue this, I would want us to write some robust tests to make sure the new algorithm is producing fair results - per my other comment thread (I haven't replied with a suggestion on how since I wanted this to be resolved. As to how, I could imagine something like rolling dice and doing a tally of the number of times each face comes up and then doing a Chi-Square test to determine if they're fair.)

@DanVanAtta do you agree with my thoughts above?

DanVanAtta:
This pull request is getting on the older side..
@frigoref  are you interested to pick this back up?
I think we just need some more performance measurements to validate there are performance gains to compensate for the complexity trade-off.

PR #10250, opened 2022-03-31
Add ssh key for frigoref to admin user authorized keys

DanVanAtta:
cc: @frigoref


frigoref:
@DanVanAtta Why is the line for me not starting with `ssh-ed25519`?

DanVanAtta:
Your key is seemingly a RSA 2048 key, not an ed25519 (elliptic curve)

PR #10257, opened 2022-04-01
Missing images/return image placeholder

DanVanAtta:

This update changes the behavior where if an image is missing then we throw an error. Far previously the engine behavior was to crash, later and until this update we showed no image at all with error message.

missing_unit_image

Commits follow roughly 3 phases
- convert 'getImage' to return the placeholder image instead of an optional
- various refactors to improve readability before the next set of updates
- cleanup and remove further cases where any error messages are shown regarding missing images

## Release Note

UPDATE|If a map has a missing image, instead of showing an error message and rendering no image at all, there will now instead be a placeholder image.

## Testing

I moved all images from the napoleanic wars 'french' player so that the engine could not find them and then loaded that map:
Screenshot from 2022-03-31 19-55-47


game-app/game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
```diff
@@ -229,7 +232,15 @@ public boolean hasImage(final ImageKey imageKey) {
...
                           return scaledImage;
-                        }));
+                        }))
+        .orElseGet(
+            () -> {
+              final URL missingUnitImageLocation =
+                  resourceLoader.getResource(FILE_NAME_BASE + "/missing_unit_image.png");
+              Image image = Toolkit.getDefaultToolkit().getImage(missingUnitImageLocation);
+              Util.ensureImageLoaded(image);
+              return image;
```
RoiEXLab, line 242: Do you know if [ImageIO#read(java.net.URL)](https://docs.oracle.com/en/java/javase/11/docs/api/java.desktop/javax/imageio/ImageIO.html#read(java.net.URL)) does the same but shorter?

RoiEXLab (approved):

game-app/game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
DanVanAtta: There is another library out there that does load an image and guarantees it to be actually loaded - though I do not recall if that was `ImageIO` or another library.

DanVanAtta:
Pushed an update which I think mostly completes this feature:

- The placeholder image is now updated to be:
missing_unit_image

- The placeholder image is now colorized by player
- We now write the name of the unit on the placeholder image. 


Below is an example with this latest update from the same Napoleonic Empires map with all French units missing:
Screenshot from 2022-04-03 21-25-00

game-app/game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
DanVanAtta: It does seem ImageIO was the library we are looking for, and returns a `BufferedImage` as well which avoids future conversions.

asvitkine:
Not sure if it's intended that it's this subtle, but this is what it looks like for missing "SWITCH" unit for Chinese for Pacific Theater Solo Challenge map:

Capture

asvitkine:
I guess maybe the issue is the tile is tinted with the player's color, but if there's no textured terrain, it ends up being the same color as the territory in which it's in (assuming it's owned by that player) and thus ends up being very subtle...

DanVanAtta:
That is somewhat subtle @asvitkine :eagle: 
I think that placeholder image probably would stand out really strongly if moved into the orange territory :grin: 

The colorization is there primarily so the owners can be differentiated, I had not considered this case with flat colors on flat colors. I do think the gray border is doing it's job to make the image visible. Overall, I'm not sure what a really good fix would be. We could experiment with making the image darker or lighter. At some point I'm concerned a bit about ROI, generally a map that has the placeholder image should be considered as somewhat broken (and hopefully a map maker would be pretty eager to fix the problem).

PR #10271, opened 2022-04-05
Update triplea_maps.yaml

Navalland2:

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2586,3 +2586,10 @@
   img: https://raw.githubusercontent.com/triplea-maps/TerraFirma1939/master/preview.png
   description: |
    <br>World War 2 on a huge map
+- mapName: WWI in 1939
+  mapCategory: EXPERIMENTAL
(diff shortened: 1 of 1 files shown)


frigoref:
@Navalland2 You are adding a new map called 'WWI in 1939'  to the maps list, correct?
Sure you want to use an abbreviation instead of 'World War I in 1939'?

triplea_maps.yaml
```diff
@@ -2586,3 +2586,10 @@
   img: https://raw.githubusercontent.com/triplea-maps/TerraFirma1939/master/preview.png
   description: |
    <br>World War 2 on a huge map
+- mapName: WWI in 1939
```
DanVanAtta, line 2589: There is a good chance this will break as the map name must be very similar to the folder name & repo name. The repo name is 'wwi_1939', so the map could be called 'WWI 1939'

It is pretty easy to rename repo's. If you do that, take care to update the 'xml' map (and/or game name) tags in each XML too if doing so.

Navalland2:
I ve opened a new pull request.

https://github.com/triplea-game/triplea/pull/10280

PR #10275, opened 2022-04-06
Add temporary instrumentation to diagnose a concurrent exception.

asvitkine:
This will allow us to determine which subclass of Change is being modified while we're processing it.

Add temporary instrumentation to diagnose a concurrent exception.

This will allow us to determine which subclass of Change is being modified while we're processing it. Once we know it, we can hopefully fix it and revert this.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/framework/HistorySynchronizer.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/framework/HistorySynchronizer.java
@@ -4,6 +4,7 @@ import games.strategy.engine.data.Change;
 import games.strategy.engine.data.GameData;
 import games.strategy.engine.data.GamePlayer;
 import games.strategy.engine.history.EventChild;
+import java.util.ConcurrentModificationException;
 import javax.swing.SwingUtilities;
 
 /**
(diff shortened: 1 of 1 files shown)


game-app/game-core/src/main/java/games/strategy/engine/framework/HistorySynchronizer.java
```diff
@@ -26,8 +27,16 @@
...
-                gameData.getHistory().getHistoryWriter().addChange(localizedChange);
+                try {
+                  final Change localizedChange = (Change) translateIntoMyData(change);
+                  gameData.getHistory().getHistoryWriter().addChange(localizedChange);
+                } catch (ConcurrentModificationException e) {
+                  // Temporary instrumentation to diagnose root cause of:
+                  // https://github.com/triplea-game/triplea/issues/10274
+                  // Include the Change subclass name in the error.
+                  throw new ConcurrentModificationException(
+                      "Translating " + change.getClass().getName(), e);
```
DanVanAtta, line 38: The exception message is what would be shown to a user in the error prompt. This exception message won't really tell them what is happening. Ideally all error message state what the code was trying to do, what failed, and what the user should do about the error. In this case, the user should do very little other than "please report this to TripleA, error translating change class: " + change.getClass().getName().

But, I do get this is temporary code and is intended to be removed before release (notably because it would be fixed by then hopefully too)

asvitkine: To be fair, without this, there's no message and it's just a "java.util.ConcurrentModificationException".

DanVanAtta:
Agree. Sorry to have been pedantic. There are a lot of error messages that
are delivered from the wrong perspective.

PR #10279, opened 2022-04-07
Fix some battle dependency issues and remove an extra slash in image path.

asvitkine:

Fix some battle dependency issues and remove an extra slash in image path.

Has the following changes:
  - Fix display name of battles in error messages (broken in commit 333f643e)
  - Remove an extra slash in a missing image path (from recent commit 652c081dfbca3ebb3773e3c616f9abdd09c60d30)
  - Don't show battle dependency errors from AI battle logic that are already handled properly.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ai/AbstractBuiltInAi.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ai/AbstractBuiltInAi.java
@@ -16,6 +16,7 @@ import games.strategy.triplea.delegate.DiceRoll;
 import games.strategy.triplea.delegate.GameStepPropertiesHelper;
 import games.strategy.triplea.delegate.Matches;
 import games.strategy.triplea.delegate.PoliticsDelegate;
+import games.strategy.triplea.delegate.battle.BattleDelegate;
 import games.strategy.triplea.delegate.battle.IBattle.BattleType;
 import games.strategy.triplea.delegate.data.BattleListing;
 import games.strategy.triplea.delegate.data.CasualtyDetails;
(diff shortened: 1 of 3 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/ai/AbstractBuiltInAi.java
```diff
@@ -609,7 +610,7 @@ protected void battle(final IBattleDelegate battleDelegate) {
         for (final Territory current : entry.getValue()) {
           final String error =
               battleDelegate.fightBattle(current, entry.getKey().isBombingRun(), entry.getKey());
-          if (error != null) {
+          if (error != null && !BattleDelegate.isBattleDependencyErrorMessage(error)) {
             log.warn(error);
```
DanVanAtta, line 614: This looks a bit like a feature envy code smell. WDYT of moving the 'logging' logic into the method `fightBattle`. If we did that, we would not need to use the return value of the method at all (and perhaps could make it a void method).

Maybe though we do not want to always show the 'warn' message, perhaps we can remove the logging entirely?

asvitkine: I think ideally the API would return a structured type or perhaps an enum and the caller would decide what to do with the result.

For example, the battle dependency return value will be shown to the human player if they try to do a battle out of order.

The problem is we can't do that without changing the network protocol, so this is a workaround for that.

I don't think it makes sense to put the logging into the called function since different call sites want to do different things with this information per above.

DanVanAtta: Yeah, looking at this more, the warning statement is perhaps entirely incorrect and should either be an exception (less desirable) or better yet, a case that is very explicitly and correctly handled in code.

PR #10280, opened 2022-04-07
Update triplea_maps.yaml

Navalland2:


Navalland2:
I don't know how do I fix the YAML formatting.

beelee1:
@Navalland2 not certain but try moving "mapName" over a space so it lines up with everything else.

PR #10281, opened 2022-04-07
Improve list of moves UI, especially on small screens.

asvitkine:

Improve list of moves UI, especially on small screens.

This change includes a number of changes to streamline the undoable moves action panel, so that it is more useable on smaller screens:
  - ~~Makes the placements and active units panels collapsed by default if the screen height is low.~~
  - Reduces padding between moves in the list.
  - Changes the buttons at the top of the move panel to be in a row, instead of in a column
  - Moves the "Undo All" button to the button row, instead of at the end of the move list
  - Moves the "Moves:" label to be outside of the scroll area and adds a horizontal separator.
  - Fixes some redrawing bugs.

Here are the before and after screenshots on a MacBook air with Dock and menu bar visible:

*Before:*
Screen Shot 2022-04-07 at 3 27 13 PM

*Before with bottom panels manually collapsed:*
Screen Shot 2022-04-07 at 3 27 19 PM

*After with bottom panels manually collapsed:*
Screen Shot 2022-04-07 at 3 17 23 PM

Edit: The change originally made the placement & active unit panels start collapsed on a smaller screen, but this was omitted from this PR per discussion.

## Release Note

CHANGE|Improved layout of move list, especially on smaller screens.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/posted/game/AbstractForumPosterPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/posted/game/AbstractForumPosterPanel.java
@@ -11,7 +11,6 @@ import games.strategy.triplea.delegate.remote.IAbstractForumPosterDelegate;
 import games.strategy.triplea.ui.ActionPanel;
 import games.strategy.triplea.ui.TripleAFrame;
 import games.strategy.triplea.ui.panels.map.MapPanel;
-import javax.swing.JLabel;
 import javax.swing.SwingUtilities;
 
 abstract class AbstractForumPosterPanel extends ActionPanel {
@@ -20,12 +19,10 @@ abstract class AbstractForumPosterPanel extends ActionPanel {
   IPlayerBridge playerBridge;
   PbemMessagePoster pbemMessagePoster;
   final ForumPosterComponent forumPosterComponent;
-  private final JLabel actionLabel;
   private TripleAFrame tripleAFrame;
 
   AbstractForumPosterPanel(final GameData data, final MapPanel map) {
     super(data, map);
-    actionLabel = new JLabel();
     forumPosterComponent = new ForumPosterComponent(data, this::performDone, getTitle());
   }
 
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/AbstractMovePanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/AbstractMovePanel.java
@@ -18,7 +18,6 @@ import java.util.Set;
 import javax.swing.Box;
 import javax.swing.BoxLayout;
 import javax.swing.JButton;
-import javax.swing.JLabel;
 import javax.swing.JOptionPane;
 import javax.swing.JPanel;
 import javax.swing.SwingUtilities;
@@ -38,7 +37,6 @@ public abstract class AbstractMovePanel extends ActionPanel {
   protected AbstractUndoableMovesPanel undoableMovesPanel;
   private final TripleAFrame frame;
   private boolean listening = false;
-  private final JLabel actionLabel = new JLabel();
   private MoveDescription moveMessage;
   private List<UndoableMove> undoableMoves;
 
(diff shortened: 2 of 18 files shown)


beelee1:
Nice improvement with "Undo All" being easily accessible

DanVanAtta:
I really like the button moves. Collapsing 'placements' panel makes sense as that is really reference information. Though, between "units to move" and "units that have moved", it's debatable which is more important to see. I think it might be more ideal to leave it up to the player to collapse the 'units to move' panel and to then respect that decision (ie: keep it collapsed by default after that point)

@asvitkine can you give some more background regarding these updates? Were they motivated from playing single player, multi-player, or something that you were made aware of from players?

asvitkine:
> I really like the button moves. Collapsing 'placements' panel makes sense as that is really reference information. Though, between "units to move" and "units that have moved", it's debatable which is more important to see. I think it might be more ideal to leave it up to the player to collapse the 'units to move' panel and to then respect that decision (ie: keep it collapsed by default after that point)
> 
> @asvitkine can you give some more background regarding these updates? Were they motivated from playing single player, multi-player, or something that you were made aware of from players?

Playing single player on a small laptop screen. On bigger screens, things generally look OK but on a smaller screen like the screenshots, the two new panels make the moves list way too cramped.

I believe both panels already remember their state, so this is just setting the initial state. Note, the logic to have them start collapsed is conditional and only happens if the screen size is small.

DanVanAtta:
What I was trying to get at, would a player on a small screen be more concerned with the moves they are about to make, or the moves they have already made? Would it not then be better to have the 'moves to do' win out over the 'moves already done'?

The answer is subjective, I wonder how to answer that objectively. My other concern, if we make the unit scroller minimized by default, maybe players would never really discover it and wind up never using it. I think there is therefore more risk in minimizing by default vs allowing a player to express their minimization preference and then respecting that preference going forward (ie: if a player has minimized it, then it should remain minimized).

Probably the ultimate would be to have the 'undo moves' itself be in a similar collapsible panel. In the meantime, I hestiate to have the unit scroller be minimized by default on smaller displays - it does give me pause. WDYT?

asvitkine:
Since you have reservations about that part of the change, let me separate it from the rest of the improvements to the move panel, since they're not really coupled.

Personally, I've not found the unit scroller very useful, but I guess it's because I approach Axis and Allies like a board game where I look at the board and decide what to move - as opposed to something like Civ where I would rely on that functionality.  I think one main difference is that in TripleA, it's often the case that you'd plan attacks that come from different territories onto a single target, so just going through "unmoved units" isn't helpful - since when you're planning that attack, you want to focus on the specific territories from which it's being done. You can't really make the choice in isolation, going in order over all unmoved units.

I do think there is value in knowing which units haven't moved, but I'm not sure scrolling through all units is the best way to do that, since some units make sense to keep stationary... A visual indication on the map is probably more useful (but then you'd need to not stack units that moved and units that haven't of the same type together).

Anyway, that's a whole separate discussion topic...

asvitkine:
@DanVanAtta I've removed the logic to collapse those panels based on screen size from this PR. Please review the rest.

DanVanAtta:
>  I think one main difference is that in TripleA, it's often the case that you'd plan attacks that come from different territories onto a single target, so just going through "unmoved units" isn't helpful - since when you're planning that attack, you want to focus on the specific territories from which it's being done.

I agree. Though, some maps have many theaters per country. I'm thinking of no-man's land in particularly where I personally would forget to consider the German navy in the pacific, essentially all the time. Part of the idea is that a person makes all the attacks they would make, and then uses the scroller to confirm that they have moved everything. In that case, those boats in the pacific would come up, and a person would then move their units in that theater and repeat until they are sure they have everything moved. Though, how people use it, certainly does vary, and even if players use the unit scroller at all.

>  A visual indication on the map is probably more useful (but then you'd need to not stack units that moved and units that haven't of the same type together).

I agree! I've toyed around with some demo code to highlight unmoved units. There is a problem in 'z-ordering' where the highlighted unit is rendered underneath non-highlighted units, so I have not been able to get that to work very cleanly. I was pretty excited about the prototype and having a clear visual indicator of units that can move is really helpful. I certainly use the 'f' key very frequently in all game modes to see which units are available to move.

DanVanAtta (approved):

PR #10285, opened 2022-04-08
Fix text overlap in air transport load dialog.

asvitkine:

Fix text overlap in air transport load dialog.

Also switch to a builder for grid bag constraints to improve readability.

The problem was the gridX value wasn't being incremented when adding the "cst" label.

UI before:
Screen Shot 2022-04-08 at 10 11 21 AM

UI after:
Screen Shot 2022-04-08 at 10 44 17 AM

## Release Note

FIX|Text overlap in dialog when loading air transports

### DIFF
--- a/lib/swing-lib/src/main/java/org/triplea/swing/jpanel/GridBagConstraintsBuilder.java
+++ b/lib/swing-lib/src/main/java/org/triplea/swing/jpanel/GridBagConstraintsBuilder.java
@@ -26,8 +26,8 @@ import java.awt.Insets;
  * </pre></code>
  */
 public class GridBagConstraintsBuilder {
-  private final int gridX;
-  private final int gridY;
+  private int gridX;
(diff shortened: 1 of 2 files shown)


lib/swing-lib/src/main/java/org/triplea/swing/jpanel/GridBagConstraintsBuilder.java
```diff
@@ -26,8 +26,8 @@
  * </pre></code>
  */
 public class GridBagConstraintsBuilder {
-  private final int gridX;
-  private final int gridY;
+  private int gridX;
+  private int gridY;
```
DanVanAtta, line 30: If these values are no longer required (via constructor); adding a no-args constructor would be consistent with the other builders.

DanVanAtta (approved):
Looks good, thank you @asvitkine

PR #10286, opened 2022-04-08
Clear dependent units map when canceling a pending move.

asvitkine:

Clear dependent units map when canceling a pending move.
This fixes a bug where if you start an air transport move, but then cancel it, the loaded units become unmovable.

## Release Note

FIX|Bug where units couldn't be moved after canceling an air transport move


asvitkine:
Converting to draft since I think this may need different handling...

asvitkine:
> Converting to draft since I think this may need different handling...

Actually, this should still be correct.

I switched to draft to verify that it's correct to clear the list entirely, or if it should persist state from previous moves. I confirmed it's OK to clear it given that we clear it after each move in `waitForMove()` which calls `clearDependencies();`.

As an aside, this class could use some clean up and refactoring, which I might do as a follow up.

But this PR should be good to merge.

PR #10297, opened 2022-04-10
Bump gradle to 7.4.2

RoiEXLab:
Just noticed it hasn't been updated for a year


RoiEXLab:
For some reason the pmdCheck suddenly fails with this update.
The issues look like false positives. Not really sure what to do about it

DanVanAtta:
@RoiEXLab are you able to repro the PMD violation locally?
I would guess these are false positives too, though are we sure?
If they are true false positives, we can add a `@SuppressWarnings` annotation (with the correct arg for this PMD violation) to explicitly suppress the violation.

With the all of the above said, it's quite surprising that gradle version would have any relevance here at all :shrug:

RoiEXLab:
@DanVanAtta Yes, I can reproduce them locally as well, but I don't know why a gradle update causes the problem

RoiEXLab:
@DanVanAtta Suppressed the false positives, merging

PR #10302, opened 2022-04-12
Clean up some code related to battles.

asvitkine:

No functional changes. Clean ups included:
  - Restructure SBR sound logic to be more concise
  - Move battle description formatting to battle display
  - Use helper functions to determine if there's a local player in a battle
  - Move some initialization of the battle window to just happen once when the window is created
  - Remove some redundant statements when creating a battle panel, which are already done as part of the same function

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/BattlePanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/BattlePanel.java
@@ -4,9 +4,7 @@ import games.strategy.engine.data.GameData;
 import games.strategy.engine.data.GamePlayer;
 import games.strategy.engine.data.Territory;
 import games.strategy.engine.data.Unit;
-import games.strategy.engine.player.Player;
 import games.strategy.triplea.Properties;
-import games.strategy.triplea.TripleAPlayer;
 import games.strategy.triplea.delegate.DiceRoll;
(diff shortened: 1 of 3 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/ui/BattlePanel.java
```diff
@@ -67,6 +64,7 @@ public final class BattlePanel extends ActionPanel {
   BattlePanel(final GameData data, final MapPanel map, final JFrame parent) {
     super(data, map);
     battleWindow = new JDialog(parent);
+    battleWindow.setDefaultCloseOperation(WindowConstants.DO_NOTHING_ON_CLOSE);
```
DanVanAtta, line 67: Do you know readily what's the default close operation? Can you give some more update about why the default close operation is set here?

I think this update is potentially a bit more risky as we could have issues with the game not closing when the main window is closed due to hidden dialogs. I think we have had a few examples of that problem in the past (and it's pretty hard to track down).

DanVanAtta: I see, we are setting this earlier than we did previously, so there is no change here.  :+1:

DanVanAtta (approved):

PR #10313, opened 2022-04-15
Fix ConcurrentModificationException in getTerritoryEffectImage().

asvitkine:

Fix ConcurrentModificationException in getTerritoryEffectImage(). It's called from multiple threads which add entries to the map. I encountered this exception once when starting a new game in my testing.

## Release Note

FIX|Fixed ConcurrentModificationException in map drawing code

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
@@ -28,6 +28,7 @@ import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
```diff
@@ -124,7 +124,8 @@ public class MapData {
   private Set<String> undrawnTerritoriesNames;
   private final Map<Image, List<Point>> decorations = new HashMap<>();
   private final Map<String, Image> territoryNameImages = new HashMap<>();
-  private final Map<String, Image> effectImages = new HashMap<>();
+  // Use a synchronized map since getTerritoryEffectImage() is called from multiple threads.
+  private final Map<String, Image> effectImages = Collections.synchronizedMap(new HashMap<>());
```
RoiEXLab, line 128: If the key is always guaranteed to be non-null you might want to consider using `ConcurrentHashMap` instead which yields better performance in most cases

RoiEXLab (approved):

game-app/game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
asvitkine: Done,

PR #10315, opened 2022-04-15
Fix long delay when using "undo all" with lots of moves.

asvitkine:
Fix long delay when using "undo all" with lots of moves.

The delay was caused by slow execution of resource bar updates. This change makes it so we only update it once if we receive lots of updates at once, assuming the swing code runs after the updates (which seems to be the case in the case of undo all).

Also makes the code more concise with some clean up.

## Release Note

FIX|The undo all action has been sped up

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/ResourceBar.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/ResourceBar.java
@@ -84,13 +84,13 @@ public class ResourceBar extends JPanel implements GameDataChangeListener {
               continue;
             }
             final double quantity = resourceStat.getValue(player, gameData, uiContext.getMapData());
-            final StringBuilder text = new StringBuilder(IStat.DECIMAL_FORMAT.format(quantity));
             final int income = resourceIncomes.getInt(resource);
+            final StringBuilder text = new StringBuilder(IStat.DECIMAL_FORMAT.format(quantity));
             text.append(" (").append(income >= 0 ? "+" : "").append(income).append(")");
             final JLabel label =
                 uiContext.getResourceImageFactory().getLabel(resource, text.toString());
             label.setBorder(BorderFactory.createEmptyBorder(0, 10, 0, 10));
-            add(label, new GridBagConstraintsBuilder(count++, 0).weightX(0).weightY(1).build());
+            add(label, new GridBagConstraintsBuilder(count++, 0).weightY(1).build());
           }
         });
   }
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/ResourceBar.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/ResourceBar.java
@@ -66,6 +66,7 @@ public class ResourceBar extends JPanel implements GameDataChangeListener {
           final GamePlayer player;
           final IntegerMap<Resource> resourceIncomes;
           try {
+            needsUpdate = false;
             gameData.acquireReadLock();
             player = gameData.getSequence().getStep().getPlayerId();
             if (player == null) {
@@ -74,7 +75,6 @@ public class ResourceBar extends JPanel implements GameDataChangeListener {
             resourceIncomes = AbstractEndTurnDelegate.findEstimatedIncome(player, gameData);
           } finally {
             gameData.releaseReadLock();
-            needsUpdate = false;
           }
 
           this.removeAll();
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/ResourceBar.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/ResourceBar.java
@@ -56,7 +56,7 @@ public class ResourceBar extends JPanel implements GameDataChangeListener {
   public void gameDataChanged(final Change change) {
     // When there are multiple gameDataChanged() notifications in a row, for example
     // from an "undo all" action, no need to do this repeatedly. This will just run
-    // once if the swing code runs after all the notifications have been received.
+    // once if the async code runs after all the notifications have been received.
     if (updateScheduled) {
       return;
     }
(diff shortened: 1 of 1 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/ui/ResourceBar.java
```diff
@@ -53,51 +53,45 @@ private void setResources() {
...
+      return;
     }
+    updateScheduled = true;
+    SwingUtilities.invokeLater(
+        () -> {
+          updateScheduled = false;
+          final GamePlayer player;
+          final IntegerMap<Resource> resourceIncomes;
+          try {
+            gameData.acquireReadLock();
```
RoiEXLab, line 69: This is bad practice. The event dispatch thread (swing thread) should never have to acquire a lock. If this lock is unavailable for some reason this will lead to a freezing UI and a bad user experience.

I'm actually not sure what's the best solution for this. In general your approach is fine. It's just this particular case where it's not optimal. Actually I think it would be better execute later invocations instead of the first one, so that the last change is actually picked up properly (in case it changes something). So that being said I have something like this in mind:
```java
private final AtomicInteger plannedUpdates = new AtomicInteger();

@Override
public void gameDataChanged(final Change change) {
  plannedUpdates.incrementAndGet();
  SwingUtilities.invokeLater(() -> {
    if (plannedUpdates.decrementAndGet() > 0) {
      return;
    }
    ThreadRunner.runInNewThread(() -> {
      // blocking code
      SwingUtilities.invokeLater(() -> {
        // ui code
      });
    });
  });
}
```

Obviously this seems unecessarily complicated because in the end we're switching to the EDT for the timing only, then to a new thread then back to the EDT. So maybe a solution like this would be better:
```java
private final AtomicInteger plannedUpdates = new AtomicInteger();

@Override
public void gameDataChanged(final Change change) {
  plannedUpdates.incrementAndGet();
  CompletableFuture.runAsync(() -> {
    if (plannedUpdates.decrementAndGet() > 0) {
      return;
    }
    // blocking code with error handling (otherwise any exception is just ignored)
    // or have a look at CompletableFutureUtils#logExceptionWhenComplete
    SwingUtilities.invokeLater(() -> {
      // ui code
    });
  });
}
```

This seems a little bit better, but honestly it seems questionable if the "plannedUpdates" counter is still as effective reducing the amount of calls because CompleteableFuture's Thread pool is most likely less busy than the EDT.

So yeah not really happy with this solution either. It definitely works, but the code is not really elegant.

The ideal solution would be to rethink the interface that is used for this because obviously it doesn't fit this use-case, but I'm pretty sure this wouldn't work without breaking network compatibility.

RoiEXLab (changes requested):

game-app/game-core/src/main/java/games/strategy/triplea/ui/ResourceBar.java
asvitkine: > Actually I think it would be better execute later invocations instead of the first one, so that the last change is actually picked up properly (in case it changes something). So that being said I have something like this in mind:

Just to be clear, if a later invocation runs *after* the swing update, it will schedule another swing update, via the boolean. So there's no issue that the update wouldn't happen.

> So that being said I have something like this in mind:
> 
> ```java
> private final AtomicInteger plannedUpdates = new AtomicInteger();
> 
> @Override
> public void gameDataChanged(final Change change) {
>   plannedUpdates.incrementAndGet();
>   SwingUtilities.invokeLater(() -> {
>     if (plannedUpdates.decrementAndGet() > 0) {
>       return;
>     }
>     ThreadRunner.runInNewThread(() -> {
>       // blocking code
>       SwingUtilities.invokeLater(() -> {
>         // ui code
>       });
>     });
>   });
> }

I don't think we need a counter to increment or the three levels of tasks, we could just do two levels, something like:

> ```java
> @Override
> public void gameDataChanged(final Change change) {
>   if (updateScheduled) { return; }
>   updateScheduled = true;
>   ThreadRunner.runInNewThread(() -> {
>      updateScheduled = false;
>      // grab lock and get info from game state
>      SwingUtilities.invokeLater(() -> {
>        // ui code
>      });
>   } 
> ```
 
This way, we avoid doing the locking operation on the UI thread and we don't schedule the op if there is one scheduled that hasn't run. I think the main downside of the above is the extra overhead of starting a new thread each time.

asvitkine: Looks like we can do: `BackgroundTaskRunner.runInBackground()` instead of always starting a new thread.

asvitkine: Nevermind, it looks like `BackgroundTaskRunner` doesn't do the right thing, it's actually something that shows UI.

asvitkine: I guess AsyncRunner.runAsync() does the job.

asvitkine: Done, please take another look.

Cernelius:
Also, that button should better be aligned with the other "undo" ones (that is, being on the left instead of centred).

game-app/game-core/src/main/java/games/strategy/triplea/ui/ResourceBar.java
DanVanAtta: A drive-by comment here, apologies for not being fully in-context:
- there are a number of ways to solving multi-threaded programming problems:
   - make everything single threaded (the EDT is an example of this)
   - use locks / semaphores
   - use thread-safe data

All immutable data is thread safe. Hence, ideally the UI never has a reference to any game data objects and only has references to immutable data. Essentially, the UI should be told to render X, and it should then render X. The references to GameData that the UI has is a mixture of logic in the UI layer, and with swing that gets kinda bad.. 

So.. I do discourage us using the 'gameData.acquireLock' mechanism and instead favor simply copying data and then using immutable data copies.

RoiEXLab (approved):

PR #10325, opened 2022-04-18
Change map scale maximum from 100% to 200%.

asvitkine:

Change map scale maximum from 100% to 200%.

This makes some maps, like Ultimate World much more usable.
However, the current implementation is such that tiles will simply be upscaled, which will make the quality not ideal. In particular, this makes unit images appear worse than they should. Improving this requires re-working how tiles are rendered (e.g. rendering them at the scaled resolution instead of at 1x and scaling them at draw time).

## Release Note

CHANGE|Map scale can now go above 100%, with 200% as maximum

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/UiContext.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/UiContext.java
@@ -52,6 +52,8 @@ public class UiContext {
   static final String UNIT_SCALE_PREF = "UnitScale";
   static final String MAP_SCALE_PREF = "MapScale";
 
+  public static final double MAP_SCALE_MAX_VALUE = 2.0;
+
   private static final String ORIGINAL_SKIN_NAME = "Original";
 
   private static final String MAP_SKIN_PREF = "MapSkin";
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/menubar/ViewMenu.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/menubar/ViewMenu.java
@@ -114,7 +114,7 @@ final class ViewMenu extends JMenu {
             "Map Zoom",
             e -> {
               final SpinnerNumberModel model = new SpinnerNumberModel();
-              model.setMaximum(100);
+              model.setMaximum(UiContext.MAP_SCALE_MAX_VALUE * 100);
               model.setMinimum((int) Math.ceil(frame.getMapPanel().getMinScale() * 100));
               model.setStepSize(1);
               model.setValue((int) Math.round(frame.getMapPanel().getScale() * 100));
--- a/game-app/game-core/src/main/java/games/strategy/ui/ImageScrollerLargeView.java
+++ b/game-app/game-core/src/main/java/games/strategy/ui/ImageScrollerLargeView.java
@@ -2,6 +2,7 @@ package games.strategy.ui;
 
 import com.google.common.primitives.Doubles;
 import games.strategy.triplea.settings.ClientSetting;
+import games.strategy.triplea.ui.UiContext;
 import java.awt.Dimension;
 import java.awt.Point;
 import java.awt.event.ActionEvent;
@@ -74,7 +75,7 @@ public class ImageScrollerLargeView extends JComponent {
             insideCount++;
             if (insideCount > 6) {
               // Scroll the map when the mouse has hovered inside the scroll zone for long enough
-              SwingUtilities.invokeLater(new Scroller());
+              SwingUtilities.invokeLater(ImageScrollerLargeView.this::scroll);
             }
           }
         }
@@ -317,7 +318,7 @@ public class ImageScrollerLargeView extends JComponent {
   }
 
   private double constrainScale(final double value) {
-    return Doubles.constrainToRange(value, getMinScale(), 1.0);
+    return Doubles.constrainToRange(value, getMinScale(), UiContext.MAP_SCALE_MAX_VALUE);
   }
 
   /** Update will not be seen until update is called. Resets the offscreen image to the original. */
@@ -327,13 +328,6 @@ public class ImageScrollerLargeView extends JComponent {
 
   public int getYOffset() {
     return model.getY();
-  }
-
-  private class Scroller implements Runnable {
-    @Override
-    public void run() {
-      scroll();
-    }
   }
 
   protected double getScaledWidth() {


Cernelius:
I think this is a bad idea. Zooming above 100% with any normal (Is TripleA using cubic interpolation or what? I don't recall.) device will give a bad looking (blurred) result. I suggest keeping the default zoom maximum at 100% but adding to Engine Preferences a way to customize it up to 200% (so that only power-users will ever happen to go above 100%). Of course, blurring is much less of an issue for maps with no details (like Ultimate World). For example, a map with mono-colourization of territories and unit zoom at 50% can be generally zoomed at 200%, looking perfectly fine or almost so.

An other thing: I don't know how it works currently, but be sure that, when there are multiple zoom acting, you zoom only once. For example, if the unit zoom is 75% and the map zoom is 200%, you zoom it units at 150% directly, instead of first zooming them at 75% and then at 200% of what you obtained, which would give very bad results. If zooming multiple times has to be, much preferably always do the greater zoom first (for example, zooming at 200% and then at 75% is much better than zooming at 75% and then at 200%).

asvitkine:
> I think this is a bad idea. Zooming above 100% with any normal (Is TripleA using cubic interpolation or what? I don't recall.) device will give a bad looking (blurred) result. I suggest keeping the default zoom maximum at 100% but adding to Engine Preferences a way to customize it up to 200% (so that only power-users will ever happen to go above 100%).

There's already a setting for the zoom level, why do we need a second, "power-user" setting for the same thing? Also, I don't think it should be an engine preference since it really depends on the map if it's useful.

Is your concern mainly around the mouse wheel function to control the zoom instead of the menu? One option is to have the two have different maxes, but this can be confusing...

> Of course, blurring is much less of an issue for maps with no details (like Ultimate World). For example, a map with mono-colourization of territories and unit zoom at 50% can be generally zoomed at 200%, looking perfectly fine or almost so.

Right, I think it really depends on the map,

> 
> An other thing: I don't know how it works currently, but be sure that, when there are multiple zooms acting, you zoom only once. For example, if the unit zoom is 75% and the map zoom is 200%, you'd better zoom the units at 150% directly, instead of first zooming them at 75% and then at 200% of what you obtained, which would give very bad results. If zooming multiple times has to be, much preferably always do the greater zoom first (For example, zooming at 200% and then at 75% is much better than zooming at 75% and then at 200%.).

Yes, as I mentioned this is currently a problem. On ultimate world, if you set unit zoom at 50% and then map zoom to 200%, units get rendered at 50% quality and upscaled, instead of at 100%. This is something that should be improved, but it requires a more in depth change so I'm not doing it as part of this one yet.

Cernelius:
> > I think this is a bad idea. Zooming above 100% with any normal (Is TripleA using cubic interpolation or what? I don't recall.) device will give a bad looking (blurred) result. I suggest keeping the default zoom maximum at 100% but adding to Engine Preferences a way to customize it up to 200% (so that only power-users will ever happen to go above 100%).
> 
> There's already a setting for the zoom level, why do we need a second, "power-user" setting for the same thing? Also, I don't think it should be an engine preference since it really depends on the map if it's useful.
> 
> Is your concern mainly around the mouse wheel function to control the zoom instead of the menu? One option is to have the two have different maxes, but this can be confusing...

I don't think we should offer regular users the ability of obtaining very bad results. The mouse wheel is clearly the worst case, so that's indeed a major concern. On this matter, does it currently display what is the current zoom when wheel-zooming? Most certainly nobody should ever go above 100% without knowing it is.

---

It would be much preferable offering a basic display at 200%, 300% and so on (only integers), without using any interpolation. No interpolation means that, in the 200% case, you will have quadruplets of pixels being the corners of an ideal square which will all act the same. This would be a basic zoom you can set per map, upon which you will apply whatever other zoom.

Cernelius:
Some visual examples, using "my" 270BC Wars game (which I believe has some of the best graphics you can find in TripleA).

100% zoom (no zoom):


200% zoom with cubic interpolation:


200% zoom with no interpolation:


Do you prefer the second or the third image?

As you can see, zooming at 200% with no interpolation is almost like watching the same image using a screen with half the wideness and highness in pixels: watching an image 200% zoomed with no interpolation on a 4k is about the same as watching the same image non-zoomed on a HD having the same screen dimensions. Not wonderful, but at least nowhere as bad as blurring the image by adding 300% more pixels via interpolation.

DanVanAtta:
@asvitkine are the results catastrophic on some maps? Are they good on other maps?

@Cernelius there are other settings that make the map look really bad, even unusable (I'm thinking of the 'map blends' option, that AFAIK basically always needs to be 'on' for the map to look good, and toggling it on/off can actually cause bad rendering). I think that is a counter-point to allowing users to opt-in to changes that do not look good. Though, if the 200% looks good on some maps, and it is a user setting, then we are enabling users to take advantage of it where it works, and to turn it off where it does not. That seems like a bigger win IMO than it is a loss. There are maps where I knock my screen resolution way down to be able to see the map easily, implying there is a pretty reasonable need to have a map zoom.

Cernelius:
This way, you would keep the current interpolating zoom having the 100% hard maximum (keeping it as a shrink-only zoom), and you would have a basic non-interpolating zoom with values like "x1", "x2", "x3" and so on, for making the map bigger. So the basic zoom would be integer and expand-only and the other zoom we already have we'll keep being percentual and shrink-only. In this case, better renaming the current zoom as "zoom-out" (a name it should already have).

asvitkine:
But any value above 100% could be implemented using the above approach behind the scenes. For example, 225% is just 300% integer zoom combined with 75% zoom down.

So I don't see why we need separate controls.

As I mentioned in the PR, the current implementation isn't ideal and could be improved, but that can be separate from the UI. Users shouldn't need to know the details of the implementation to set a zoom level.

DanVanAtta:
It used to be that people had to crank a cars starter manually before the ignition would work. That is now bundled together and so automated that a new driving does not at all need to be aware of what a starter even is, let alone they are actuating it first when turning their car key before the ignition starts. This is essentially the UX argument for information hiding and avoiding exposing too many and unnecessary details to users. We've had similar problems in the lobby with moderators being confused between user bans and username bans, which was even worse when there were multiple ways to ban users. Instead of just "click this button to ban a person", it was "here the button to do an IP ban", here is another for this other kind of ban. The end result, there was more to explain, and whenever a moderator chose to ban someone, then they had to make many further decisions of how to ban them rather than just telling the computer "ban this person". 

So all that is to say, the system should not expose unnecessary detail unless it really needs to. If our '300%' zoom is a 4x zoom followed by 75% reduction, that is a detail. I don't believe have more 'controls' is going to be more here.

PR #10340, opened 2022-04-21
Fix error if the same "undo" is clicked twice in succession.

asvitkine:

Fix error if the same "undo" is clicked twice in succession.

This change disables the button once it's clicked, so that clicking it again won't do anything. Previously, it would result in an "move index out of bounds" error dialog shown.

## Release Note

FIX|Fixed error that would be shown when an undo move button would be quickly clicked twice

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/AbstractUndoableMovesPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/AbstractUndoableMovesPanel.java
@@ -145,7 +145,8 @@ public abstract class AbstractUndoableMovesPanel extends JPanel {
           scrollBarPreviousValue = scroll.getVerticalScrollBar().getValue();
           final String error = movePanel.undoMove(moveIndex);
           if (error == null) {
-            // Disable the button so it can't be clicked again until the UI is updated.
+            // Disable the button so it can't be clicked again. Note: Undoing will cause a later
+            // setMoves() call on this object, which will re-build the UI for all the moves.
             undoButton.setEnabled(false);
             previousVisibleIndex = Math.max(0, moveIndex - 1);
           } else {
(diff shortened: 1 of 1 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/ui/AbstractUndoableMovesPanel.java
```diff
@@ -137,13 +137,27 @@ private JComponent newComponentForMove(final AbstractUndoableMove move) {
...
+    final JButton undoButton = new JButton("Undo");
+    final int moveIndex = move.getIndex();
+    undoButton.addActionListener(
+        (e) -> {
+          // Record position of scroll bar as percentage.
+          scrollBarPreviousValue = scroll.getVerticalScrollBar().getValue();
+          final String error = movePanel.undoMove(moveIndex);
+          if (error == null) {
+            // Disable the button so it can't be clicked again until the UI is updated.
+            undoButton.setEnabled(false);
```
RoiEXLab, line 149: Just so I understand it correctly:
`movePanel.undoMove` causes the button to disappear somehow (presumably because the backing panel disappears), so when the same panel is shown again the button is newly created so it can be clicked again.
Is this assumption correct?

RoiEXLab (approved):

game-app/game-core/src/main/java/games/strategy/triplea/ui/AbstractUndoableMovesPanel.java
asvitkine: There's a button for each move. So this button will disappear when the move is no longer shown in the list. But that's not synchronous so we need to make the button not clickable in the meantime so it's not pressed a second time.

The button doesn't get recreated because the move doesn't exist. But any other moves still have their own buttons and any new moves will get them too.

RoiEXLab: Okay, but which part does manipulate the list?

RoiEXLab: Basically I'm not sure why the code does what you say it does. For me this looks like the button is disabled and for some reason there's no problem for it to be disabled forever.

So maybe this is just me, but if my confusion is justified I'd appreciate an explaining comment there, or an adjustment to the existing comment

asvitkine: Sure, let me expand the comment. But TL;DR is the whole content will be re-created via setMoves() which will be called after an update.

asvitkine: Improved the comment.

PR #10344, opened 2022-04-22
Extract swing UI from ServerModel

RoiEXLab:

This performs some slight refactoring for the ServerModel class and removes any direct references to JFrame by moving them to their respective interface implementations.
Ideally the headed implementation could be moved to the `:game-app:game-headed` subproject (and the headless implementation to the respective headless subproject, but that's not that important) to have a clear compile-time separation of the logic and the UI.
Currently the GameRunner class has too much dependencies on the UI code to make this a reality but that could be fixed as well.


game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/HeadedLaunchAction.java
```diff
@@ -99,4 +116,71 @@ public void onLaunch(final ServerGame serverGame) {}
...
+  }
+
+  @Override
+  public WatcherThreadMessaging createThreadMessaging() {
+    return new WatcherThreadMessaging.HeadedWatcherThreadMessaging(ui);
+  }
+
+  @Override
+  public Optional<ServerConnectionProps> getFallbackConnection(Runnable cancelAction) {
+
```
asvitkine, line 137: Nit: Remove empty line.

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LaunchAction.java
```diff
@@ -6,11 +6,17 @@
...
 import games.strategy.engine.framework.startup.ui.panels.main.game.selector.GameSelectorModel;
 import games.strategy.engine.player.Player;
+import games.strategy.net.Messengers;
 import java.nio.file.Path;
+import java.util.Optional;
 import java.util.Set;
+import org.triplea.game.chat.ChatModel;
 import org.triplea.sound.ISound;
 
 /**
```
asvitkine, line 22: As an aside (not for this PR), but I find the name of this class pretty confusing. It seems that the class is mostly responsible for handling game lifecycle events. I'm not sure what the best terminology is for this. In other codebases, I've seen "Client" or "Delegate" used. I know in TripleA, the Delegate terminology is already used and it's not quite the same as this. If we use Client terminology, it could be ServerModelClient.

Or perhaps we can just a different name altogether. For example, ServerBehavior. Thinking about names more, I think ServerModel is also not quite the right name for what that class is doing - it's a bit more of a controller than a model, so perhaps ServerController could be more apt?

... stepping back a bit, the reason I have a problem with LaunchAction is that:
  1. It's not a single action - there's lots of stuff there.
  2. It's not just related to launching.

So even something like ServerActions could be a better name, although not everything is "actions" - e.g. getAutoSaveFile().

asvitkine (approved):
LGTM

RoiEXLab:
@asvitkine Actually I noticed that with some small tweaks I can move the headed implementation to the headed subproject. Couldn't do the same with the headless implementation though, that one is tied slightly more into the codebase, mainly because of `HeadlessGameServer::headless` which is used in a lot of places to make decisions if swing UI should be used or not :/

Please have another look

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LaunchAction.java
RoiEXLab: Agreed. I came up with this name back when I started to migrate a couple of methods that were mainly UI elements related to starting the game.
As more and more things piled up the original name wasn't really fitting anymore but here we are. Really open to anything.

RoiEXLab:
@asvitkine As a side note: I noticed there actually were some pretty "isolated" ui classes that could be moved to the headed subproject without any adjustments. I suspect there might be a couple more where this is possible without any problems so I would like to encourage you to keep an eye out for them.
Ideally we'd end up with a codebase where game-core no longer has any references to swing/awt and therefore the UI is completely independent of the game logic and could at least in theory be swapped for any other framework

game-app/game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
```diff
@@ -260,10 +260,13 @@ private TripleAFrame(
       final IGame game,
       final LocalPlayers players,
       final UiContext uiContext,
-      @Nullable final Chat chat) {
+      @Nullable final Chat chat,
+      final Runnable clientLeftGame) {
     super("TripleA - " + game.getData().getGameName());
     setDefaultCloseOperation(DISPOSE_ON_CLOSE);
 
+    this.clientLeftGame = clientLeftGame;
```
RoiEXLab, line 268: `TripleAFrame` is literally interconnected with almost every UI element out there, so moving it to `:game-app:game-headed` was not feasible with reasonable effort.
That's why the method is passed in as a lambda.

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ClientModel.java
```diff
@@ -147,13 +147,20 @@ public void doneSelectingPlayers(final byte[] gameData, final Map<String, INode>
...
-      final LaunchAction launchAction) {
+      final ServerSetupModel typePanelModel,
+      final LaunchAction launchAction,
+      final Runnable showMainFrame,
+      final Runnable clientLeftGame) {
     this.launchAction = launchAction;
     this.typePanelModel = typePanelModel;
     this.gameSelectorModel = gameSelectorModel;
+    this.showMainFrame = showMainFrame;
+    this.clientLeftGame = clientLeftGame;
```
RoiEXLab, line 163: `ClientModel` is a class where I'm not sure if it should be part of the headed subproject.
It clearly is only used by the headed client, but it has an odd relationship with `ServerModel` which seems to suggest there is logic in here that's independent of UI, which might indicate the need for s separate class that should be kept in game-core.
Due to this nature this class is also connected with other parts of the code that would need to be carefully separated before being able to move this class to the headed subproject.

game-app/game-headed/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/HeadedServerSetupModel.java
```diff
@@ -100,7 +101,13 @@ public void onServerMessengerCreated(
    */
   public void showClient() {
     Preconditions.checkState(!SwingUtilities.isEventDispatchThread());
-    final ClientModel model = new ClientModel(gameSelectorModel, this, new HeadedLaunchAction(ui));
+    final ClientModel model =
+        new ClientModel(
+            gameSelectorModel,
+            this,
+            new HeadedLaunchAction(ui),
+            HeadedGameRunner::showMainFrame,
```
asvitkine, line 109: Any reason these are passed separately instead just being methods on the launch action?

asvitkine (approved):
Still LGTM, just one comment.

game-app/game-headed/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/HeadedServerSetupModel.java
RoiEXLab: Mainly because there's no headless counterpart from what I can tell. To be fair many methods are rather close to no-op so that's a rather weak argument. To some degree I see this solution as temporary glue until `ClientModel` can be moved to `game-headed` which would allow us to just remove this argument again

PR #10349, opened 2022-04-22
Move logic supporting legacy (1.8) spelling to setter.

asvitkine:
Move logic supporting legacy (1.8) spelling to setter.

The getter version was showing up in my profiles of unit movement UI code. Also updates some docs to use the non-legacy names.

Note: I checked if we could get rid of this entirely, but there are still maps that still use the old names:
https://github.com/search?q=org%3Atriplea-maps+attatchment&type=code
https://github.com/search?q=org%3Atriplea-maps+attatchto&type=code

### DIFF
--- a/docs/development/background/design/engine-code-overview.md
+++ b/docs/development/background/design/engine-code-overview.md
@@ -81,7 +81,7 @@ will stop the parser, and terminate the program.
 
 getAttachment,
 where key is the name of the attachment.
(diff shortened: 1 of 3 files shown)


RoiEXLab (approved):
> The getter version was showing up in my profiles of unit movement UI code.

You mean in performance profiling, seems very likely. This getter is probably called a lot, even if the footprint of each individual call is probably small, it adds up

PR #10356, opened 2022-04-23
Use an in-memory cache in ImageFactory by string key.

asvitkine:
Use an in-memory cache in ImageFactory by string key.

I noticed this showed up in execution time profiles when looking at the performance of handling mouse events during the move phase, which spent a lot of time in `updateRouteAndMouseShadowUnits()` that eventually call down to this for country small flags.

This prevents excessive resource look up and ImageIO calls.

I tested this by sampling CPU in VisualVM on a given save game, where I shift-click all the units from Japan and move my mouse around between Japan, SZ 36 and SZ 24, with 10 territory transition totals on each run.

Here's the profile before:
before_profile

Here's the one after this change:
after_profile2

In particular, you can see that `setMouseShadowUnits()` is much faster and no longer has `getSmallFlag()` taking a lot of time to execute and total CPU time of the AWT event thread is significantly reduced.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/image/ImageFactory.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/image/ImageFactory.java
@@ -1,5 +1,8 @@
 package games.strategy.triplea.image;
 
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import games.strategy.triplea.ResourceLoader;
 import java.awt.Image;
 import java.io.IOException;
@@ -9,6 +12,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import javax.annotation.Nonnull;
 import javax.imageio.ImageIO;
 
 /**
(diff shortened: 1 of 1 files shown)


RoiEXLab:
Hmm this contradicts the statements in #9823 a little bit. Who's right here?

asvitkine:
> Hmm this contradicts the statements in #9823 a little bit. Who's right here?

Interesting. The docs for ImageIO say:

```
Sets a flag indicating whether a disk-based cache file should be used when creating ImageInputStreams and ImageOutputStreams.
When reading from a standard InputStream>, it may be necessary to save previously read information in a cache since the underlying stream does not allow data to be re-read. Similarly, when writing to a standard OutputStream, a cache may be used to allow a previously written value to be changed before flushing it to the final destination.

The cache may reside in main memory or on disk. Setting this flag to false disallows the use of disk for future streams, which may be advantageous when working with small images, as the overhead of creating and destroying files is removed.
```

Notably, it uses the word "may" which sounds like it doesn't guarantee that there's a memory cache. It's also possible that ImageIO's cache logic is not very good. For example, it may evict entries too aggressively.

For this change, I am seeing real world results that suggests it improves performance. However, I'm currently testing locally with some other non-landed changes that speed up the overall operation (of computing units that can move to a territory) very significantly, so I can't provide specific instructions on how to repro my findings.

One option is we can wait on this PR a bit until I can land all my other improvements and then we can see the results against HEAD. (The reason I suggest to wait for the other improvements is that they're more significant than this one, so they drown out this specific improvement...)

RoiEXLab:
@asvitkine Sounds like a good idea to wait and compare again

asvitkine:
> @asvitkine Sounds like a good idea to wait and compare again

OK, the other performance changes landed and I did some measurements with this change. Turns out it wasn't just image lookup that's the problem, but also the resource loading. So I updated the cache to have the resource key as the key.

I've added some screenshots of the profiling results in JVisualVM, it's pretty clear that this makes a difference. Please take another look.

RoiEXLab:
@asvitkine Thank you for this PR. The performance gains are really nice to see.
A thing that somewhat concerns me though is the fact that IO operations are performed on the IO Thread in the first place. They should be performed in some other thread or thread pool and then invoke the edt thread to perform the updates once the data is ready.

I'm curious if swing will ever adopt the upcoming "[virtual threads](https://openjdk.java.net/jeps/425)" which allows developers to write seemingly synchronous code which is actually running on a thread pool and therefore not blocking the actual EDT when waiting for stuff like IO operations to finish.

PR #10363, opened 2022-04-25
Don't recompute getMovementLeft() excessively during sort operations.

asvitkine:
Don't recompute getMovementLeft() excessively during sort operations.

Use a cache object so that the value for each unit is only computed once during sorting. I noticed this in performance profiles of movement code.

Also add UtilityClass annotation and fix a comment in the PerfTimer class.

### DIFF
--- a/lib/java-extras/src/main/java/org/triplea/performance/PerfTimer.java
+++ b/lib/java-extras/src/main/java/org/triplea/performance/PerfTimer.java
@@ -15,7 +15,7 @@ import org.triplea.java.function.ThrowingSupplier;
  * <p>Example usage with auto-close try block:
  *
  * <pre>{@code
- * try(PerfTimer timer = PerfTimer.startTimer("timer_name_0")) {
+ * try(final PerfTimer timer = PerfTimer.time("timer_name_0")) {
  *   // code to be timed
  * }
  * }</pre>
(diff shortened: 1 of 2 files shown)


lib/java-extras/src/main/java/org/triplea/performance/PerfTimer.java
```diff
@@ -15,7 +15,7 @@
  * <p>Example usage with auto-close try block:
  *
  * <pre>{@code
- * try(PerfTimer timer = PerfTimer.startTimer("timer_name_0")) {
+ * try(final PerfTimer timer = PerfTimer.time("timer_name_0")) {
```
RoiEXLab, line 18: Is this change intended?

asvitkine: Yes, the previous comment was wrong. Just fixing the comment along the way.

RoiEXLab: The final keyword was left out on purpose. It doesn't do anything at all.
```java
try (PerfTimer timer = PerfTimer.time("timer_name_0")) {
  timer = PerfTimer.time("another timer"); // wont compile, timer is final
}
```

asvitkine: Sorry, the main purpose of this change is to use `time()` instead of `startTimer()` since the later doesn't exist. Removed the `final` now.

RoiEXLab: No worries, just a small detail I didn't know myself for a long time :D

RoiEXLab (approved):

Issue #6718, opened 2020-06-18
AI Strategic Bombs Already Max Damage Factories


drockken:
### How can the problem be recreated?

Set up a factory with maximum damage, and an opportunity for the AI to strategic bomb the factory. 

### What is an expected fix?

AI enhancement to not strategic bomb factories that cannot be damaged any further.  An algorithm could be added to deny bombing when expected damage is less than the expected cost of losing the plane.

I think I have seen this issue with tactical bombers and harbors/airfields as well, but I don't have an example.

### Which Engine Version are you using?
2.0.19993  - I have seen this issue in a number of different versions of TripleA.  I attached a save-game, look at Italy's most recent move to see.  The max factory damage in London is 20, but Italy strat-bombs the factory anyway.

Let me know if the Zip doesn't upload.

[sb_max_dmg.zip](https://github.com/triplea-game/triplea/files/4800189/sb_max_dmg.zip)

beelee1:
Maybe AI wants to make sure those factories are obliterated. Might want to ask @ron-murhammer he's the AI guy :)

panther2:
FWIW I just want to add that "overbombing" is not violating the rules, however useless it appears to be.

DanVanAtta:
This is the right place to report. Please do not PM or ask specific people
to look at items.

If only one person handles an issue they are unlikely to create
documentation, build code that is easily understood by others, and they
themselves are stuck as they cannot go on vacation. If they do go manage to
go on vacation, then it leaves the rest of us in a bad place

frigoref:
@panther2 @DanVanAtta @beelee1 
I have no clue what the issue here is. Can anyone summarize this nicely and provide an example that can be used?

> AI enhancement to not strategic bomb factories that cannot be damaged any further. An algorithm could be added to deny bombing when expected damage is less than the expected cost of losing the plane.
> 
> I think I have seen this issue with tactical bombers and harbors/airfields as well, but I don't have an example.
My questions would be:
- Is the rule with the "execpted cost of losing the plane" making any sense to you guys? The damage can be repaired, no? Shouldn't the rule reflect the likelihood of killing the factory unit?
- How about examples for tactical bombers and harbors/airfields? Can anyone provide them as well?

beelee1:
[...]

PR #10367, opened 2022-04-25
AI Strategic Bombs (Issue 6718)

frigoref:

Fixing issue [6718](https://github.com/triplea-game/triplea/issues/6718) AI Strategic Bombs Already Max Damage Factories


game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -885,40 +888,61 @@ private void determineUnitsToAttackWith(
...
               Properties.getRaidsMayBePreceededByAirBattles(data.getProperties())
                   && AirBattle.territoryCouldPossiblyHaveAirBattleDefenders(t, player, data, true);
           if (canBeBombedByThisUnit
               && !canCreateAirBattle
               && canAirSafelyLandAfterAttack(unit, t)) {
-            final int noAaBombingDefense =
-                t.getUnitCollection().anyMatch(Matches.unitIsAaForBombingThisUnitOnly()) ? 0 : 1;
+            // get territory data that is independent of the attack unit
+            final BombedTerritoryData bombedTerritoryData =
+                bombedTerritoryDataMap.computeIfAbsent(
```
asvitkine, line 913: Continuing my discussion from the issue.

I don't think this is correct., because bombedTerritoryDataMap isn't keyed by bomber type or factory type, which are relevant for the purpose of `t.getUnitCollection().getMatches(bombingTargetMatch);`.

So you're computing a `minDamageNeeded` value for the territory as a whole which is not right when different units can bomb different types of factories.

As I said earlier, I think the issue goes away if you get rid of the map. Just compute your logic directly in the body, instead of using bombedTerritoryDataMap.computeIfAbsent().

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -885,40 +888,61 @@ private void determineUnitsToAttackWith(
...
+                    t,
+                    territory -> {
+                      int noAaBombingDefense = 1;
+                      // minimum damage to allow bombing = max. damage a defense unit can take
+                      int minDamageNeeded = 0;
+                      final List<Unit> potentialBombingTargetUnits =
+                          t.getUnitCollection().getMatches(bombingTargetMatch);
+                      for (final Unit targetUnit : potentialBombingTargetUnits) {
+                        minDamageNeeded = Math.max(minDamageNeeded, targetUnit.getUnitDamage());
+                        if (Matches.unitIsAaForBombingThisUnitOnly().test(targetUnit)) {
```
asvitkine, line 923: Do we have logic like this somewhere already in the codebase in a utility class? Ideally, we could reuse logic, so if there are rules updates or other changes, it stays consistent.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
frigoref: As mentioned in the forum I have just committed another commit to introduce `minDamageNeeded vs. maxDamage calculated/aggregated per target unit`.
@asvitkine Please have a look and let me know your thoughts.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1729,21 +1754,66 @@ private Map<Unit, Set<Territory>> tryToAttackTerritories(
...
+    public BombedTerritoryData() {
+      this.noAaBombingDefense = 1;
+    }
+
+    public void addPotentialTargets(List<Unit> targetUnits) {
+      for (Unit targetUnit : targetUnits) {
+        minNeededAndMaxDamageMap.compute(
+            targetUnit,
+            (targetUnitOld, damagePairOld) -> {
+              Integer unitMinDamageNeeded = (targetUnitOld == null ? 0 : damagePairOld.getFirst());
```
asvitkine, line 1786: Nit: I think it's a bit clearer if you check `damagePairOld == null`, even though it's semantically equivalent.

Also, let's rename these to `existingUnitKey` and `existingDamagePair` to make it a bit clearer the semantics of `compute()` since not everyone may be familiar with it.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1729,21 +1754,66 @@ private Map<Unit, Set<Territory>> tryToAttackTerritories(
...
+              unitMinDamageNeeded = Math.max(unitMinDamageNeeded, targetUnitOld.getUnitDamage());
```
asvitkine, line 1787: Doesn't this throw a NPE if `targetUnitOld` is null?

asvitkine (changes requested):
So I have some comments on the current state of the code, but at a high level, I feel like the bomber map is adding a lot of complexity. Can't we just keep the code similar to how it was before the latest updates, but not use a map?

For example, adapting from your previous code:

````
                      int existingDamage = 0;
                      final List<Unit> potentialBombingTargetUnits =
                          t.getUnitCollection().getMatches(bombingTargetMatch);
                      for (final Unit targetUnit : potentialBombingTargetUnits) {
                        existingDamage += targetUnit.getUnitDamage();
                        if (Matches.unitIsAaForBombingThisUnitOnly().test(targetUnit)) {
                          noAaBombingDefense = 0;
                        }
                      }
....

            final int remainingDamagePotential = maxDamage - 3 * numExistingBombers - existingDamage;
```
Shouldn't that be sufficient? If not, what am I missing?

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -37,7 +37,11 @@
 import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
+import lombok.Getter;
+import org.apache.commons.math3.util.Pair;
```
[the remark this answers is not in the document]
frigoref: Isn't the use of Range in this case a misuse? We are not trying to define a range, but we try to find the minimal damage needed and the maximum damage possible. The assumption is that the maximum damage one attacking unit can achieve is less (<) than the minimum damage needed. However, the combined damage of multiple attacking units could revert this to max dmg > min dmg needed.
Shall I just define a value object?

game-app/game-core/src/test/java/games/strategy/triplea/delegate/WW2V3Year41Test.java
```diff
@@ -100,6 +102,10 @@
 class WW2V3Year41Test {
   private final GameData gameData = TestMapGameData.WW2V3_1941.getGameData();
 
+  private Territory getTerritory(final String territoryName) {
```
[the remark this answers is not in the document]
frigoref: Apparently a rework of this test class got into the commit. Would it be okay to keep it in? It is just a simplification based on some SonarLint finding.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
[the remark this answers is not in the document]
asvitkine: You're right, I think Range is not a good fit here.

Before reworking, please see my main comment about whether we need this map at all.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -659,7 +663,7 @@ private List<Unit> moveOneDefenderToLandTerritoriesBorderingEnemy(
       territoriesToAttack.add(patd.getTerritory());
     }
 
-    // Find land territories with no can't move units and adjacent to enemy land units
+    // Find land territories without units and adjacent to enemy land units
```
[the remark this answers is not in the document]
frigoref: The method is called `moveOneDefenderToLandTerritoriesBorderingEnemy` so I understand that the defender tries to defend LandTerritoriesBordering the Enemy by moving OneDefender unit. From this I deduced the comment, but I cannot make sense of "Find land territories without units that can't move and adjacent to enemy land units".

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
frigoref: Fixed in [7e15c7a](https://github.com/frigoref/triplea/commit/7e15c7a7310f379b14fea38cc8836059d9d3e6e1).

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
frigoref: Fixed in [7e15c7a](https://github.com/frigoref/triplea/commit/7e15c7a7310f379b14fea38cc8836059d9d3e6e1).

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1729,21 +1754,66 @@ private Map<Unit, Set<Territory>> tryToAttackTerritories(
...
+    public void addPotentialTargets(List<Unit> targetUnits) {
+      for (Unit targetUnit : targetUnits) {
+        minNeededAndMaxDamageMap.compute(
+            targetUnit,
+            (targetUnitOld, damagePairOld) -> {
+              Integer unitMinDamageNeeded = (targetUnitOld == null ? 0 : damagePairOld.getFirst());
+              unitMinDamageNeeded = Math.max(unitMinDamageNeeded, targetUnitOld.getUnitDamage());
+              if (noAaBombingDefense > 0
+                  && Matches.unitIsAaForBombingThisUnitOnly().test(targetUnit)) {
+                noAaBombingDefense = 0;
```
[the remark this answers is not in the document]
frigoref: Fixed in [7e15c7a](https://github.com/frigoref/triplea/commit/7e15c7a7310f379b14fea38cc8836059d9d3e6e1).

frigoref:
@asvitkine Let me first say, that I am very glad that you are assisting me so patiently so get this one solved!

> So I have some comments on the current state of the code, but at a high level, I feel like the bomber map is adding a lot of complexity. Can't we just keep the code similar to how it was before the latest updates, but not use a map?
> 
> Shouldn't that be sufficient? If not, what am I missing?

The problem described in issue [6718](https://github.com/triplea-game/triplea/issues/6718) leads to the need of comparison between damage potential and minimal damage needed. Adding to this the example you've made with **A bombs B and X bombs Z** (B and Z are different factory types) a comparison between each attacking unit and the potential target units is needed. I do not see a way around this. Am I missing anything?

asvitkine:
> @asvitkine Let me first say, that I am very glad that you are assisting me so patiently so get this one solved!
> 
> > So I have some comments on the current state of the code, but at a high level, I feel like the bomber map is adding a lot of complexity. Can't we just keep the code similar to how it was before the latest updates, but not use a map?
> > Shouldn't that be sufficient? If not, what am I missing?
> 
> The problem described in issue [6718](https://github.com/triplea-game/triplea/issues/6718) leads to the need of comparison between damage potential and minimal damage needed. Adding to this the example you've made with **A bombs B and X bombs Z** (B and Z are different factory types) a comparison between each attacking unit and the potential target units is needed. I do not see a way around this. Am I missing anything?

My suggestion should take that into account since it's within the outer loop that's iterating over bombers and it's look at units filtered by `bombingTargetMatch` which should only return units that the bomber can bomb.

frigoref:
> My suggested code takes that into account since it's within the outer loop that's iterating over bombers and it looks at units filtered by `bombingTargetMatch` which should only return units that the bomber can bomb.

The filter by `bombingTargetMatch` is (attack) unit specific as according to panther's reply in issue [6718](https://github.com/triplea-game/triplea/issues/6718)

> different bombers can attack different targets

My understanding is that this leads to the requirement of the "bomber map" or am I misunderstanding your point completely here?

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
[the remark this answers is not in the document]
frigoref: See conversation below (rework to ValueObject awaits result of conversation).

asvitkine:
> The filter by `bombingTargetMatch` is (attack) unit specific as according to panther's reply in issue [6718](https://github.com/triplea-game/triplea/issues/6718)
> 
> > different bombers can attack different targets
> 
> My understanding is that this leads to the requirement of the "bomber map" or am I misunderstanding your point completely here?

Really sorry for taking a long time to reply. Let me explain my thoughts as to why I think my suggested implementation would be sufficient without needing to keep a map:

So from the issue we're trying to fix https://github.com/triplea-game/triplea/issues/6718:

> AI enhancement to not strategic bomb factories that cannot be damaged any further.

Basically, we want to not send additional bombers to factories that are already at max damage. That means both (`a`) not to send any bombers if factory is fully damaged and (`b`) to not send too many, to account for the damage maximum.

Furthermore, we have the requirement that different factories types may be bombable by different unit types (`c`).

So how does my suggested code in https://github.com/triplea-game/triplea/pull/10367#pullrequestreview-955324471 meet these three requirements?

Recall that we're iterating through bomber units and we're deciding where to send them. For each bomber, we calculate the territory with the highest `maxBombingScore`, and then decide to send the bomber there.

How do we handle the fact that different factories can be bombed by different units (`c`)? Well, we're iterating through each individual bomber unit, and for each one, we generate a `potentialBombingTargetUnits` collection - of units that can actually be bombed by it. I think my code sample was missing this, but we can check if this collection is empty, then no units can be bombed by it - and we can skip that territory.

How do we make sure we don't send bombers to fully bombed factories? Well, when computing `remainingDamagePotential`, we subtract the amount of damage already inflicted (`existingDamage`). So if the damage already matches the max damage, the score will be 0 and that territory won't qualify for `maxBombingScore`. So this handles (`a`).

How do we handle (`b`) (not to send too many bombers)? This is similar to (`a`), but the additional factor is how many bombers we're planning to send to that territory, which the code already accounts for, via the `numExistingBombers` variable, which comes from the `attackMap` which is updated every time we choose to send a bomber to a territory.

Therefore, based on the above, I don't think we need the extra map and a solution similar to the one I proposed should be able to work. Do you disagree? Am I missing something in my explanation?

As an aside, there are a few things I'm not certain about with the existing code, which is worth checking:
1. First, it gets `maxDamage` via `ta.getProduction()`. Is that always correct? I thought when multiple specific units can be bombed, it may not always be based on the territory value - but I'm not an expert on the different ruleset options here. We should check how this is actually calculated during bombing battles to understand if this is correct - and ideally use the same helper function from both places. Also, if there are multiple bombable infrastructure units in that territory, is the `maxDamage` for each unit? If so, that's not being taken into account right now.
2. It assumes that each bomber will do `3` points of damage (via the `3 * numExistingBombers` calculation). My guess is this not correct in the general case - since it depends on the map. Obviously, there is also a dice roll involved, but even ignoring that (and assuming 3 is supposed to be the average value or something), I think different maps can control the number of dice and the effect of those for the damage? Ideally, again, we could use the same helper functions that actual bombing battle code uses.

These two things above are not directly related to this change, but since we're looking at this code, would be good to fix these.

One other thing, the issue also says:
> An algorithm could be added to deny bombing when expected damage is less than the expected cost of losing the plane.

I think this part is a separate orthogonal feature request that can be handled in a different PR (i.e. taking into account the expected cost of losing the plane), so I'm ignoring it for this discussion (I think your approach also wasn't handling that). That can be investigated separately.

frigoref:
Let me try to structure this a bit
**1. Requirement list**
> (`a`) not to send any bombers if factory is fully damaged 
> (`b`) to not send too many, to account for the damage maximum
> (`c`) different factories types may be bombable by different unit types 

**2. Your questions/points and my answers**
> Therefore, based on the above, I don't think we need the extra map and a solution similar to the one I proposed should be able to work. Do you disagree? Am I missing something in my explanation?

Yes, I do disagree and I think you are missing the following:
(`c`) cannot be fulfilled when the logic is just using the number of existing bombers with `final int numExistingBombers = attackMap.get(t).getBombers().size();` without taking into account which unit they can actually bomb or not. (this goes in addition to your point 2 (assumption of `3` damage for each bomber)

> 1. First, it gets `maxDamage` via `ta.getProduction()`. Is that always correct? 

According to the A&A rules for [Strategic Bombing](https://axisandallies.fandom.com/wiki/Strategic_Bombing) the rule is
"This is the ability for a bomber to attack a factory. If the bomber survives [AA Gun](https://axisandallies.fandom.com/wiki/AA_Gun) fire, it will proceed to attack the factory for either cash or unit production."
I am not certain that "cash or unit production" is correctly determined, but otherwise, I would think that the maximum damage possible during SBR is the production of the territory attachment. Therefore, `maxDamage = ta.getProduction();` is correct.

> > An algorithm could be added to deny bombing when expected damage is less than the expected cost of losing the plane.
> 
> I think this part is a separate orthogonal feature request that can be handled in a different PR 

I agree, this should be handled separately.

**3. My questions**
Concerning the case of different types of factories I would assume that each bomber can only bomb one factory type in a territory. So either there is a relation bomber types to factory types they can bomb being n:1, or this relation is ensured by how the factories are placed in the territories.
**3.a** Do you agree to this assumption?

Considering (`c`) I would think that only the view of all bombers combined can result in a good choice for the attacker where to send the bombers. 
**3.b** Do you agree to this view?
Example: In the case where 2 bombers `A` and `B` can either 
(Opt1) be send to territory `1` or 
(Opt2) only `A` can be send to territory `2` 
the individual view of each bomber might result for `A` to be send to `2` and then for `B` the needed damage in `1` is not reached alone.

asvitkine:
Thanks for your detailed reply. I see what you're saying.

We previously decided for a bomber to be sent to territory X, which based on the logic in this code means that there was a valid target there. But then, when we're considering a different bomber bombing territory X, it only knows that the other bomber was sent here, but not that the two are able to target the same units.

So the "3 * bombers" calculation may not be correct. I do think there is a trade off between code complexity and handling all the edge cases. For example, I think we could get a pretty good result by just checking the existing bombers scheduled to this territory being able to target the same set of units as this territory.

So instead of just doing:

`            final int numExistingBombers = attackMap.get(t).getBombers().size();`

We could do something like:

```
  int numExistingBombers = 0;
  for (Unit bomber : attackMap.get(t).getBombers()) {
     var targets = t.getUnitCollection().getMatches(
              Matches.unitCanProduceUnitsAndCanBeDamaged()
              .and(Matches.unitIsLegalBombingTargetBy(bomber)));
    if (CollectionUtils.haveEqualSizeAndEquivalentElements(targets, potentialBombingTargetUnits)) {
      numExistingBombers++;
    }
  }
```

This will ensure we're taking into account bombers that can target the same units for the purpose of the logic. It's not perfect, but I think it will do a good job in general - and I don't think the O(n^2) complexity is a big deal here given there won't be thousands of bombers generally, so we're still dealing with small numbers (e.g. < 20 and usually more like ~5) on most standard maps.

> Considering (`c`) I would think that only the view of all bombers combined can result in a good choice for the attacker where to send the bombers.
> **3.b** Do you agree to this view?
> Example: In the case where 2 bombers `A` and `B` can either
> (Opt1) be send to territory `1` or
> (Opt2) only `A` can be send to territory `2`
> the individual view of each bomber might result for `A` to be send to `2` and then for `B` the needed damage in `1` is not reached alone.

So I think we use a "greedy" algorithm right now, so it's true in general that there are cases where it may not result in an optimal assignment. But let me think aloud about your specific example.

First, if `A` is sent to `2`, we know it was the best place for it to go in isolation. Then, let's say `B` is sent to `1` - it means it's still the best place for it to go given the prior decision that `A` was sent to `2`. What changes if `A` wasn't sent to `2`?
Well, if `B` can reach `2`, it could go there instead. This could be better if `B` has a higher damage potential than `A` (i.e. different unit, different dice rolls) and damage maxes are such that swapping them makes more sense (i.e. `2` can take more damage). This becomes similar to the transport packing problem - you have units of varying sizes and transports of varying capacities and you want to fit the most units. I wonder if we could re-use the same algorithm we do there - although I also am not sure if it's really optimal. I recall this being discussed at some point.

But I also worry about the complexity of a more optimal solution. Right now, we're discussing the fixing of a relatively simple bug where AI just doesn't consider max damage. I think we can solve this pretty well with something like I suggest above and we don't need to necessarily do something complicated that needs to handle all possible hypothetical edge cases that don't appear in practice in 99% of the maps.

I also worry that we likely don't have test coverage for this code - it would be great if we had some specific scenarios we can set up and test the AI makes a good decision in those. Especially in the case where we decide to do a more complex solution.

frigoref:
Hi @asvitkine ,

I agree that not all edge cases should be considered if the complexity is too high, but the question would be what you mean with the complexity being "too high". I took up the idea of "we just want to fix issue [6718](https://github.com/triplea-game/triplea/issues/6718)" and reverted the introduced storage class/object.
To solve the issue I introduced 
`neededDamageUnit+= target.getHowMuchMoreDamageCanThisUnitTake(t);`
Following the KISS principle we stay with the approach to evaluate each bomber individually.
For the purpose of considered the right  the logic for `numExistingBombers` I added the logic you have suggested and the result looks like this now:
```
for (Unit target : targetUnits) {
  neededDamageUnits += target.getHowMuchDamageCanThisUnitTakeTotal(t);
  int numExistingBombers = 0;
  for (Unit existingBomber : existingAttackingBombers) {
    if (Matches.unitCanProduceUnitsAndCanBeDamaged()
        .and(Matches.unitIsLegalBombingTargetBy(existingBomber))
        .test(target)) {
      ++numExistingBombers;
    }
  }
  final int remainingDamagePotential =
      maxDamageProduction - neededDamageUnits - 3 * numExistingBombers;
  final int bombingScore = (1 + 9 * noAaBombingDefense) * remainingDamagePotential;
  if (bombingScore >= maxBombingScore) {
    maxBombingScore = bombingScore;
    maxBombingTerritory = Optional.of(t);
  }
}
```

Last but not least I extracted some methods `determineTerritoriesThatCanBeBombed` and `determineBestBombingAttackForBomber` to encapsulate this better.

My retest was with the game file positive.

BR,
frigoref

PS: Test methods are a general issue with the AI at the moment and I would not want to tackle it in this PR.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1202,6 +1162,86 @@ && canAirSafelyLandAfterAttack(unit, t)) {
...
+        territoryManager.getAttackOptions().getBomberMoveMap();
+    final Set<Unit> bombers = bomberMoveMap.keySet();
+    bombers.removeAll(alreadyAttackedWithUnits);
+    final boolean raidsMayBePrecededByAirBattles =
+        Properties.getRaidsMayBePreceededByAirBattles(data.getProperties());
+    for (final Unit bomber : bombers) {
+      final Set<Territory> bomberTargetTerritories = bomberMoveMap.get(bomber);
+      bomberTargetTerritories.removeIf(
+          terr ->
+              raidsMayBePrecededByAirBattles
```
asvitkine, line 1179: Since this won't remove anything if `raidsMayBePrecededByAirBattles` is false, you can rework this to avoid the work if the property isn't set:

```
if (raidsMayBePrecededByAirBattles) {
  bomberTargetTerritories.removeIf(
          terr -> AirBattle.territoryCouldPossiblyHaveAirBattleDefenders(
                      terr, player, data, true));
}
```

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1202,6 +1162,86 @@ && canAirSafelyLandAfterAttack(unit, t)) {
...
   }
 
+  private void determineTerritoriesThatCanBeBombed(
+      final Map<Territory, ProTerritory> attackMap,
+      final Map<Unit, Set<Territory>> sortedUnitAttackOptions,
+      final Set<Unit> alreadyAttackedWithUnits) {
+    final Map<Unit, Set<Territory>> bomberMoveMap =
+        territoryManager.getAttackOptions().getBomberMoveMap();
+    final Set<Unit> bombers = bomberMoveMap.keySet();
+    bombers.removeAll(alreadyAttackedWithUnits);
```
asvitkine, line 1172: I think this changes the underlying map. Is this really intended?

Honestly, I find the previous formulation more straight forward. It seems OK to just skip keys that are in the `alreadyAttackedWithUnits` - and the performance of that is easier to reason about than removeAll(), especially if it would require us to make a copy.

Plus, if you iterate over entries, you can avoid the bomberMoveMap.get() lookup you have on line 1176.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
asvitkine: Additionally, this will modify the underlying map. I think you want to do `getMatches()` instead.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1202,6 +1162,86 @@ && canAirSafelyLandAfterAttack(unit, t)) {
...
+        int neededDamageUnits = 0;
+        final List<Unit> existingAttackingBombers = attackMap.get(t).getBombers();
+        for (Unit target : targetUnits) {
+          neededDamageUnits += target.getHowMuchDamageCanThisUnitTakeTotal(t);
+          int numExistingBombers = 0;
+          for (Unit existingBomber : existingAttackingBombers) {
+            if (Matches.unitCanProduceUnitsAndCanBeDamaged()
+                .and(Matches.unitIsLegalBombingTargetBy(existingBomber))
+                .test(target)) {
+              ++numExistingBombers;
```
asvitkine, line 1224: Nit: I think we usually put ++ at the end in this codebase.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1202,6 +1162,86 @@ && canAirSafelyLandAfterAttack(unit, t)) {
...
+        Matches.unitCanProduceUnitsAndCanBeDamaged()
+            .and(Matches.unitIsLegalBombingTargetBy(bomber));
+    Optional<Territory> maxBombingTerritory = Optional.empty();
+    int maxBombingScore = MIN_BOMBING_SCORE;
+    for (final Territory t : bomberTargetTerritories) {
+      final List<Unit> targetUnits = t.getUnitCollection().getMatches(bombingTargetMatch);
+      if (!targetUnits.isEmpty() && canAirSafelyLandAfterAttack(bomber, t)) {
+        final int noAaBombingDefense =
+            t.anyUnitsMatch(Matches.unitIsAaForBombingThisUnitOnly()) ? 0 : 1;
+        int maxDamageProduction = 0; // territory damage + target unit damage(s)
```
asvitkine, line 1210: I find the name a bit confusing. What do you think about `maxPossibleDamage` instead?

Also, I don't understand the comment after it, since it's implying it's adding stuff up, but this is just set based on the territory attachment.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1202,6 +1162,86 @@ && canAirSafelyLandAfterAttack(unit, t)) {
...
+        final TerritoryAttachment ta = TerritoryAttachment.get(t);
+        if (ta != null) {
+          maxDamageProduction = ta.getProduction();
+        }
```
asvitkine, line 1214: Looks like you can just call TerritoryAttachment.getProduction(t) which has this logic built in.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1202,6 +1162,86 @@ && canAirSafelyLandAfterAttack(unit, t)) {
...
+        int neededDamageUnits = 0;
+        final List<Unit> existingAttackingBombers = attackMap.get(t).getBombers();
+        for (Unit target : targetUnits) {
+          neededDamageUnits += target.getHowMuchDamageCanThisUnitTakeTotal(t);
+          int numExistingBombers = 0;
+          for (Unit existingBomber : existingAttackingBombers) {
+            if (Matches.unitCanProduceUnitsAndCanBeDamaged()
```
asvitkine, line 1221: Since we're iterating over `targetUnits`, we actually know that `Matches.unitCanProduceUnitsAndCanBeDamaged()` is already true since we checked it via `bombingTargetMatch`, so this can be simplified to just `(Matches.unitIsLegalBombingTargetBy(existingBomber)`.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1202,6 +1162,86 @@ && canAirSafelyLandAfterAttack(unit, t)) {
...
+            t.anyUnitsMatch(Matches.unitIsAaForBombingThisUnitOnly()) ? 0 : 1;
+        int maxDamageProduction = 0; // territory damage + target unit damage(s)
+        final TerritoryAttachment ta = TerritoryAttachment.get(t);
+        if (ta != null) {
+          maxDamageProduction = ta.getProduction();
+        }
+        int neededDamageUnits = 0;
+        final List<Unit> existingAttackingBombers = attackMap.get(t).getBombers();
+        for (Unit target : targetUnits) {
+          neededDamageUnits += target.getHowMuchDamageCanThisUnitTakeTotal(t);
```
asvitkine, line 1218: Interesting. Looking at its implementation, I see it has a check for:

Properties.getDamageFromBombingDoneToUnitsInsteadOfTerritories()

That seems to imply that if the property is set, then we shouldn't be using ta.getProduction().

We should check with a rules expert, but that seems to imply that the max damage should be either:

if (Properties.getDamageFromBombingDoneToUnitsInsteadOfTerritories()) {
  max = sum(target.getHowMuchDamageCanThisUnitTakeTotal(t));
} else {
  max = TerritoryAttachment.getProduction(t)
}

Your logic doesn't seem to match that. But maybe I'm missing something.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1202,6 +1162,86 @@ && canAirSafelyLandAfterAttack(unit, t)) {
...
+          int numExistingBombers = 0;
+          for (Unit existingBomber : existingAttackingBombers) {
+            if (Matches.unitCanProduceUnitsAndCanBeDamaged()
+                .and(Matches.unitIsLegalBombingTargetBy(existingBomber))
+                .test(target)) {
+              ++numExistingBombers;
+            }
+          }
+          final int remainingDamagePotential =
+              maxDamageProduction - neededDamageUnits - 3 * numExistingBombers;
```
asvitkine, line 1228: I still think the `3` is not correct here - wouldn't it depend on the dice count and strength of the bomber? I'm OK with still keeping it as `3` for now, but perhaps you can add a TODO comment about this?

asvitkine (changes requested):
Sorry for the delay. I did a pass and have a few comments. Overall, I like this logic much better, but have still have some concerns, see my individual comments. Thanks again for working on this!

asvitkine:
@frigoref Gentle nudge on this - it would be great to wrap this up and land it!

No worries if you're busy, just wanted to be clear that I think this is a worthwhile change and I'm hoping we can iterate to get it to a good state to land. :)

frigoref:
> No worries if you're busy, just wanted to be clear that I think this is a worthwhile change and I'm hoping we can iterate to get it to a good state to land. :)

Sorry, I must have missed your latest reply. Now I am quite busy for another week at least, but I will get back to this PR afterwards. Thanks for you patience!

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
frigoref: Comment added `// assume each other bomber causes a damage of 3`.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
frigoref: I think we are not 100% clear about the difference of damage to the unit or the underlying production of this unit.
From my understanding the method `getHowMuch**More**DamageCanThisUnitTake` should be used.

However, it seemed to me that there was a mistake and I would like to put my thoughts out on the to-be logic which you can comment about separate from the (modified) code now:
`remainingDamagePotential = 

1: maxDamageProduction (of the territory) 
2: - neededDamageUnits (= damage still possible of all units  that can be bombed) 
3. - 3 * numExistingBombers (assumed to be attacking together with the current bomber)`
Note that 3. leave the potential that not all (co-)bombers assumed to attack together might join based on the individual target assessment run. Furthermore, each co-bomber should be counted only ones.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
frigoref: Done.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
frigoref: Removed the comment.
The name `maxDamageProduction` is more suited from my perspective as it is the maximum damage to the territory production and the word "damage" is otherwise used to refer to units.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
frigoref: We have several cases where we do pre-increment and it is a question of style. I prefer this one as I read it as "increcement" and then the variable. 
However, I had to replace this counter anyway.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
frigoref: Whole section reworked also taking into account the other comment.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
frigoref: Extracted `raidsMayBePrecededByAirBattles` as described.
Whole section reworked also taking into account the other comment.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1197,6 +1156,91 @@ && canAirSafelyLandAfterAttack(unit, t)) {
...
+        int maxDamageProduction = TerritoryAttachment.getProduction(t);
+        int neededDamageUnits = 0;
+        final List<Unit> existingAttackingBombers = attackMap.get(t).getBombers();
+        final List<Unit> sameTargetBombers = new ArrayList<>();
+        for (Unit target : targetUnits) {
+          neededDamageUnits += target.getHowMuchMoreDamageCanThisUnitTake(t);
+          for (Unit existingBomber : existingAttackingBombers) {
+            if (!sameTargetBombers.contains(existingBomber)
+                && Matches.unitIsLegalBombingTargetBy(existingBomber).test(target)) {
+              sameTargetBombers.add(existingBomber);
```
asvitkine, line 1222: Don't we want to only add if *all* targets correspond? Right now, you add if *any* correspond.

You can change to something like:
```java
for (Unit existingBomber : existingAttackingBombers) {
  if (targetUnits.stream().allMatch(u -> Matches.unitIsLegalBombingTargetBy(existingBomber).test(u))) {
    sameTargetBombers.add(existingBomber);
  }
}
```

And have a separate loop to sum up `neededDamageUnits`.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1197,6 +1156,91 @@ && canAirSafelyLandAfterAttack(unit, t)) {
...
+            }
+          }
+        }
+        // assume each other bomber causes a damage of 3
+        final int remainingDamagePotential =
+            maxDamageProduction - neededDamageUnits - 3 * sameTargetBombers.size();
```
asvitkine, line 1228: I don't understand the `maxDamageProduction  - neededDamageUnits` expression.

That sounds to me that the territory production value gets reduced by how much damage units can still take.
That's not right, is it?

I would expect it should either be the sum of the two (e.g. max bombing damage = production + damage that individual factories can take), or choosing one or the other, depending on game rules, or the max or min of the two. 
Subtracting just seems not right to me.

Can we verify with the engine code where damage is actually determined from bombing?

asvitkine (commented):
Thanks, got a chance to do another review pass. A few more comments, but I think it's on the right track.

frigoref:
@asvitkine Fixes made. If you have further points, maybe we should set up a call instead of an extended back-and-forth here.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
frigoref: It should be `maxDamageProduction + neededDamageUnits` and I corrected it.

Concerning the damage logic from bombing you can check StrategicBombingRaidBattle especially variable `bombingRaidDamage` and call of method `getTarget`.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
frigoref: In case a bomber cannot raid against all units, but against one (do not know if this is actually possible) this code should consider the bomber as co-bomber if _any_ common target unit exists. Why should the co-bombers be limited to those being able to bomb _all_ target units?

Considering `neededDamageUnits` the idea is that all target units can consume damage before the production is harmed.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1197,6 +1156,91 @@ && canAirSafelyLandAfterAttack(unit, t)) {
...
+      final Map<Unit, Set<Territory>> sortedUnitAttackOptions,
+      final Set<Unit> alreadyAttackedWithUnits) {
+    final boolean raidsMayBePrecededByAirBattles =
+        Properties.getRaidsMayBePreceededByAirBattles(data.getProperties());
+    territoryManager
+        .getAttackOptions()
+        .getBomberMoveMap()
+        .entrySet()
+        .forEach(
+            bomberEntry -> {
```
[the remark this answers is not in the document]
frigoref: I think this is really not necessary, but I did it anyway.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1197,6 +1156,91 @@ && canAirSafelyLandAfterAttack(unit, t)) {
...
+              final Unit bomber = bomberEntry.getKey();
+              if (alreadyAttackedWithUnits.contains(bomber)) {
+                return; // already attacked bombers cannot move
+              }
+              Set<Territory> bomberTargetTerritories = bomberEntry.getValue();
+              if (raidsMayBePrecededByAirBattles) {
+                bomberTargetTerritories =
+                    new HashSet(
```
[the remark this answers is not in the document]
frigoref: Done.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
asvitkine: Thanks. The relevant code seems to be:

```
      int damageLimit = TerritoryAttachment.getProduction(battleSite);
      final boolean limitDamage =
          Properties.getWW2V2(gameData.getProperties())
              || Properties.getLimitRocketAndSbrDamageToProduction(gameData.getProperties());

        if (limitDamage) {
          costThisUnit = Math.min(costThisUnit, damageLimit);
        }
```

and:

```java
      // Limit PUs lost if we would like to cap PUs lost at territory value
      if (Properties.getPuCap(gameData.getProperties())
          || Properties.getLimitSbrDamagePerTurn(gameData.getProperties())) {
        final int alreadyLost = gameData.getMoveDelegate().pusAlreadyLost(battleSite);
        final int limit = Math.max(0, damageLimit - alreadyLost);
        cost = Math.min(cost, limit);
        if (!targets.isEmpty()) {
          for (final Unit u : bombingRaidDamage.keySet()) {
            if (bombingRaidDamage.getInt(u) > limit) {
              bombingRaidDamage.put(u, limit);
            }
          }
        }
      }
```

So if those properties are in place, then we should consider that `damageLimit` (which is `TerritoryAttachment.getProduction(battleSite)`) should be the max, both for the overall cost and to the cost to each target unit.

Additionally, right after there's code that also enforces the `damageLimit` per unit if `Properties.getDamageFromBombingDoneToUnitsInsteadOfTerritories()` is set.

Importantly, if `Properties.getDamageFromBombingDoneToUnitsInsteadOfTerritories()` is false, `bombingRaidDamage` isn't used.

That means, we should take that account in the AI code you're changing. Specifically, `neededDamageUnits` should be 0 in that case.

Otherwise, the other properties should be used to cap the damage per bomber, so I think instead of `3`, it should be `Math.min(3,  maxDamageProduction)`, if `Properties.getPuCap(gameData.getProperties()) || Properties.getLimitSbrDamagePerTurn(gameData.getProperties()))`.
Note: The 3 is still bogus and would ideally be calculated from the expected value of the bomber's dice, but that can be left as a TODO (3 is close to the expected value of 3.5 for a regular d6). Adding the `Math.min()` seems a simple enough change that I think it's worthwhile to do here.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1197,6 +1156,91 @@ && canAirSafelyLandAfterAttack(unit, t)) {
...
+                        CollectionUtils.getMatches(
+                            bomberTargetTerritories,
+                            terr ->
+                                AirBattle.territoryCouldPossiblyHaveAirBattleDefenders(
+                                    terr, player, data, true)));
```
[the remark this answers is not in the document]
asvitkine: I don't think we can check `RAIDS_MAY_BE_PRECEEDED_BY_AIR_BATTLES` at the beginning of the method, because we checked !(1 && 2), which is the same as (!1 || !2), so in particular, if 2 is false (there are no potential air units for the battle), we would execute the code.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
asvitkine: > @asvitkine Fixes made. If you have further points, maybe we should set up a call instead of an extended back-and-forth here.

I think this the tricky code around bombing requires offline studying to have meaningful discussion, so while I appreciate the offer, I don't think I could provide good feedback in a live discussion on this.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
asvitkine: One other note:

Really what we care about is the application of the limit - and it is pretty peculiar.

if `Properties.getDamageFromBombingDoneToUnitsInsteadOfTerritories()` is set, then the limit is per unit, so it's straight up  `maxDamageProduction * targetUnits.size()`, *if* the limit should be applied (based on properties).

If not, then the target units are not relevant, and it's just the overall limit (either infinity or the production value, depending on properties).

frigoref: I agree to the case where the property is _not_ set (then only the territory is relevant, meaning its production value/infinitive).
But in the case where the property is set, each unit needs to be checked how much damage it can take and this is added to the territory part (again its production value/infinitive). This is not the same as your formula `maxDamageProduction * targetUnits.size()`.
If that is true the only thing I see that could be changed is the target units part if the property is _not_ set.
Did you wanted to express a different part or did I get you right?

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
frigoref: Okay, then we'll leave it?

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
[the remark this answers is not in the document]
frigoref: So we want to avoid air battles: If the property allows preceding air battels, only territories without the potential to have one should be kept, else all territories are relevant.
I made a change and added a comment for this.
Note that I use a not now: `!AirBattle.territoryCouldPossiblyHaveAirBattleDefenders`

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
asvitkine: Yes, sorry you're right. If the property is set, we actually do `damageLimit = current.getHowMuchMoreDamageCanThisUnitTake(battleSite)` - so it's not based on `maxDamageProduction`. My mistake.

frigoref: Then this part is resolved, right?

frigoref:
@asvitkine Did we cover everything now?

asvitkine:
> @asvitkine Did we cover everything now?

My earlier comments about the different properties were meant to suggest that the code in question should be checking those, because `maxDamageProduction + neededDamageUnits` is not correct.

For example, if `Properties.getDamageFromBombingDoneToUnitsInsteadOfTerritories()` is false, then the `neededDamageUnits` shouldn't be added.

Looking closer at `getHowMuchDamageCanThisUnitTakeTotal()`, which your code is using, it does check the above property, but if that property is not set, then it returns `Integer.MAX_VALUE`. So this would be actually a problem for this code, because then you'd be adding a bunch of `Integer.MAX_VALUE`'s together, resulting in overflows, etc.

So you should check the above property yourself.

frigoref:
> My earlier comments about the different properties were meant to suggest that the code in question should be checking those, because `maxDamageProduction + neededDamageUnits` is not correct.
> 
> ...
> 
> So you should check the above property yourself.

Done, please check.

asvitkine:
> > My earlier comments about the different properties were meant to suggest that the code in question should be checking those, because `maxDamageProduction + neededDamageUnits` is not correct.
> > ...
> > So you should check the above property yourself.
> 
> Done, please check.

I'm not seeing the changes, did you push everything? To make sure it's not just the github UI, I even looked at the raw diff here: https://patch-diff.githubusercontent.com/raw/triplea-game/triplea/pull/10367.diff

I couldn't find the string `getDamageFromBombingDoneToUnitsInsteadOfTerritories()`.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1197,6 +1157,89 @@ && canAirSafelyLandAfterAttack(unit, t)) {
...
+    final boolean raidsMayBePrecededByAirBattles =
+        Properties.getRaidsMayBePreceededByAirBattles(data.getProperties());
+    for (final Map.Entry<Unit, Set<Territory>> bomberEntry :
+        territoryManager.getAttackOptions().getBomberMoveMap().entrySet()) {
+      final Unit bomber = bomberEntry.getKey();
+      if (alreadyAttackedWithUnits.contains(bomber)) {
+        return; // already attacked bombers cannot move
+      }
+      // Avoid air battles in the target territory
+      final Collection<Territory> bomberTargetTerritories;
```
asvitkine, line 1173: Suggestion:
```java
      Collection<Territory> bomberTargetTerritories = bomberEntry.getValue();
      if (raidsMayBePrecededByAirBattles) {
        bomberTargetTerritories =
            CollectionUtils.getMatches(
                bomberTargetTerritories,
                t -> !AirBattle.territoryCouldPossiblyHaveAirBattleDefenders(t, player, data, true));
      }
```

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
asvitkine: > In case a bomber cannot raid against all units, but against one (do not know if this is actually possible)
I believe this is possible, since each unit type can specify `getBombingTargets()` separately.

> this code should consider the bomber as co-bomber if any common target unit exists. Why should the co-bombers be limited to those being able to bomb all target units?

I think it gets tricky because you don't know which units something is targeting. For example, you have A that can target X and Y and B that can target Z and X. Let's say X,Y,Z are all in the same territory.

When considering B after A is already scheduled, you don't know if A's damage should reduce B's potential or not. So if you just consider units that both can target, you miss out on the fact that A could bomb just Y and B still has full potential between X and Z.

However, I think the case where we consider "all targets possible" is also not right, so I think neither of these simple approaches handle the above example well. As such, I'm OK with your current code on this - it's not perfect, but for most maps it will work well enough.

In the future, the logic could be expanded to handle complicated cases well, if some map runs into it, but that doesn't seem to be a priority.

Maybe just add a comment about the limitation?

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
frigoref: Done.

frigoref:
> I'm not seeing the changes, did you push everything? To make sure it's not just the github UI, I even looked at the raw diff here: https://patch-diff.githubusercontent.com/raw/triplea-game/triplea/pull/10367.diff

Apparently it didn't go through. Should be now.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1199,6 +1159,94 @@ && canAirSafelyLandAfterAttack(unit, t)) {
...
+        int sameTargetBombersCount = 0;
+        final List<Unit> existingAttackingBombers = attackMap.get(t).getBombers();
+        if (!Properties.getDamageFromBombingDoneToUnitsInsteadOfTerritories(data.getProperties())) {
+          sameTargetBombersCount = existingAttackingBombers.size();
+        } else {
+          final List<Unit> sameTargetBombers = new ArrayList<>();
+          for (final Unit target : targetUnits) {
+            neededDamageUnits += target.getHowMuchMoreDamageCanThisUnitTake(t);
+            for (final Unit existingBomber : existingAttackingBombers) {
+              if (!sameTargetBombers.contains(existingBomber)
```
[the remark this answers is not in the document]
frigoref: Done.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -1199,6 +1159,94 @@ && canAirSafelyLandAfterAttack(unit, t)) {
...
+      final List<Unit> targetUnits = t.getUnitCollection().getMatches(bombingTargetMatch);
+      if (!targetUnits.isEmpty() && canAirSafelyLandAfterAttack(bomber, t)) {
+        final int noAaBombingDefense =
+            t.anyUnitsMatch(Matches.unitIsAaForBombingThisUnitOnly()) ? 0 : 1;
+        int maxDamageProduction = TerritoryAttachment.getProduction(t);
+        // determine damage need for unit (if property allows) and number of same target bombers
+        int neededDamageUnits = 0;
+        int sameTargetBombersCount = 0;
+        final List<Unit> existingAttackingBombers = attackMap.get(t).getBombers();
+        if (!Properties.getDamageFromBombingDoneToUnitsInsteadOfTerritories(data.getProperties())) {
```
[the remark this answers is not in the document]
frigoref: Done. With your below suggestion the switch makes sense, otherwise I would have preferred the one-line-block first.

PR #10372, opened 2022-04-26
Clean up some code related to moving. No functional changes.

asvitkine:

Clean up some code related to moving. No functional changes.

Tested using edit mode to make sure the logic that ensures selected units have the same owner works with the refactoring.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/TransportTracker.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/TransportTracker.java
@@ -221,11 +221,11 @@ public class TransportTracker {
   /** Detects if a unit has unloaded units in a previous game phase. */
   public static boolean hasTransportUnloadedInPreviousPhase(final Unit transport) {
     final Collection<Unit> unloaded = transport.getUnloaded();
+    boolean nonCombat = GameStepPropertiesHelper.isNonCombatMove(transport.getData(), true);
     // See if transport has unloaded anywhere yet
     for (final Unit unit : unloaded) {
       // cannot unload in two different phases
-      if (GameStepPropertiesHelper.isNonCombatMove(transport.getData(), true)
-          && unit.getWasUnloadedInCombatPhase()) {
+      if (nonCombat && unit.getWasUnloadedInCombatPhase()) {
         return true;
       }
     }
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/TransportTracker.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/TransportTracker.java
@@ -220,16 +220,8 @@ public class TransportTracker {
 
   /** Detects if a unit has unloaded units in a previous game phase. */
   public static boolean hasTransportUnloadedInPreviousPhase(final Unit transport) {
-    final Collection<Unit> unloaded = transport.getUnloaded();
-    boolean nonCombat = GameStepPropertiesHelper.isNonCombatMove(transport.getData(), true);
-    // See if transport has unloaded anywhere yet
-    for (final Unit unit : unloaded) {
-      // cannot unload in two different phases
-      if (nonCombat && unit.getWasUnloadedInCombatPhase()) {
-        return true;
-      }
-    }
-    return false;
+    return GameStepPropertiesHelper.isNonCombatMove(transport.getData(), true)
+        && transport.getUnloaded().stream().anyMatch(Unit::getWasUnloadedInCombatPhase);
   }
 
   /**
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/move/validation/MoveValidator.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/move/validation/MoveValidator.java
@@ -1171,10 +1171,7 @@ public class MoveValidator {
       final GamePlayer player,
       final Map<Unit, Unit> unitsToTransports,
       final MoveValidationResult result) {
-    if (!route.hasWater()) {
-      return result;
-    }
-    if (!units.isEmpty() && units.stream().allMatch(Matches.unitIsAir())) {
+    if (!route.hasWater() || (!units.isEmpty() && units.stream().allMatch(Matches.unitIsAir()))) {
       return result;
     }
     // If there are non-sea transports return
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/move/validation/MoveValidator.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/move/validation/MoveValidator.java
@@ -1171,11 +1171,10 @@ public class MoveValidator {
       final GamePlayer player,
       final Map<Unit, Unit> unitsToTransports,
       final MoveValidationResult result) {
-    final boolean isEditMode = getEditMode(data.getProperties());
-    if (!units.isEmpty() && units.stream().allMatch(Matches.unitIsAir())) {
+    if (!route.hasWater()) {
       return result;
     }
-    if (!route.hasWater()) {
+    if (!units.isEmpty() && units.stream().allMatch(Matches.unitIsAir())) {
       return result;
     }
     // If there are non-sea transports return
@@ -1189,6 +1188,7 @@ public class MoveValidator {
     final Territory routeEnd = route.getEnd();
     final Territory routeStart = route.getStart();
     // if unloading make sure length of route is only 1
+    final boolean isEditMode = getEditMode(data.getProperties());
     if (!isEditMode && route.isUnload()) {
       if (route.hasMoreThenOneStep()) {
         return result.setErrorReturnResult("Unloading units must stop where they are unloaded");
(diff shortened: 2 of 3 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/ui/panel/move/MovePanel.java
```diff
@@ -158,63 +158,45 @@ public void unitsSelected(
...
-                      ownerMatch);
-            } else {
-              chooser =
-                  new UnitChooser(
-                      unitsToMove,
-                      selectedUnits,
-                      null,
-                      UnitSeparator.SeparatorCategories.builder().build(),
-                      false,
-                      getMap().getUiContext());
```
asvitkine, line 211: We didn't really need to have two branches. The edit mode branch added a filter that ensured only units of the same player were selected. But this is fine to do in non-edit mode too, though it's a no-op.

game-app/game-core/src/main/java/games/strategy/triplea/ui/panel/move/MovePanel.java
```diff
@@ -158,63 +158,45 @@ public void unitsSelected(
...
-              }
-            }
+          final boolean editMode = BaseEditDelegate.getEditMode(getData().getProperties());
+
+          GamePlayer requiredOwner = null;
+          if (!editMode) {
+            requiredOwner = getCurrentPlayer();
+          } else if (!selectedUnits.isEmpty()) {
+            requiredOwner = selectedUnits.iterator().next().getOwner();
           }
```
asvitkine, line 168: I split out this logic to make it much clearer what's being validated.

game-app/game-core/src/main/java/games/strategy/triplea/ui/panel/move/MovePanel.java
```diff
@@ -1006,8 +991,6 @@ private Predicate<Unit> getMovableMatch(final Route route, final Collection<Unit
       } else {
         movableBuilder.and(enoughMovement);
       }
-    }
-    if (route != null) {
```
asvitkine, line 1010: The if above was the same, so this just combined them.

asvitkine:
For some reason, the diff on github is showing the cherry-pick of 391e395, even though that change is already on master. :\

RoiEXLab:
@asvitkine This is because the git history diverges earlier in the timeline at commit 0dee3ac116c3a5f43736188eb9d93513579acb33
This is the last common commit and therefore git will show the diff until this commit. Merge commits or rebases will fix this issue

game-app/game-core/src/main/java/games/strategy/triplea/delegate/TransportTracker.java
```diff
@@ -221,11 +221,11 @@ public static Collection<Unit> getUnitsLoadedOnAlliedTransportsThisTurn(
   /** Detects if a unit has unloaded units in a previous game phase. */
   public static boolean hasTransportUnloadedInPreviousPhase(final Unit transport) {
     final Collection<Unit> unloaded = transport.getUnloaded();
+    boolean nonCombat = GameStepPropertiesHelper.isNonCombatMove(transport.getData(), true);
     // See if transport has unloaded anywhere yet
     for (final Unit unit : unloaded) {
       // cannot unload in two different phases
-      if (GameStepPropertiesHelper.isNonCombatMove(transport.getData(), true)
-          && unit.getWasUnloadedInCombatPhase()) {
+      if (nonCombat && unit.getWasUnloadedInCombatPhase()) {
```
RoiEXLab, line 228: I would even go one step further. Why loop over the whole unloaded collection when nonCombat is false, when you can just return early.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/move/validation/MoveValidator.java
```diff
@@ -1171,11 +1171,10 @@ private MoveValidationResult validateTransport(
       final GamePlayer player,
       final Map<Unit, Unit> unitsToTransports,
       final MoveValidationResult result) {
-    final boolean isEditMode = getEditMode(data.getProperties());
-    if (!units.isEmpty() && units.stream().allMatch(Matches.unitIsAir())) {
+    if (!route.hasWater()) {
       return result;
     }
-    if (!route.hasWater()) {
+    if (!units.isEmpty() && units.stream().allMatch(Matches.unitIsAir())) {
```
RoiEXLab, line 1177: Is there a specific reason preventing those if cases to be combined into a single one?

game-app/game-core/src/main/java/games/strategy/triplea/ui/panel/move/MovePanel.java
RoiEXLab: If you add an else branch that assigns `null` to `requiredOwner` it can be made final.

Even better in my opinion would be to extract these couple of lines into a small helper method that returns an optional instance of `GamePlayer` (either wrapped in an optional, or annotated as `@Nullable`).

game-app/game-core/src/main/java/games/strategy/triplea/ui/panel/move/MovePanel.java
```diff
@@ -261,14 +236,14 @@ private void selectUnitsToMove(
...
             int addCount = 0;
             for (final Unit unit : unitsToMove) {
               if (!selectedUnits.contains(unit)) {
                 selectedUnits.add(unit);
                 addCount++;
-                if (addCount >= iterCount) {
+                if (addCount >= maxCount) {
                   break;
                 }
               }
```
RoiEXLab, line 249: I think this code can be enhanced by rephrasing it similar to this:
```java
final int maxCount = mouseDetails.isAltDown() ? deselectNumber : 1; // deselectNumber is a constant and should be written as DESELECT_NUMBER
final List<Unit> unitsToMove = units.stream()
  .filter(unitsToMoveMatch)
  .filter(Predicate.not(selectedUnits::contains))
  .sorted(UnitComparator.getHighestToLowestMovementComparator())
  .limit(maxCount)

selectedUnits.addAll(unitsToMove);
```

Keeps the weird counter out of the way and doesn't sort more than `maxCount` items because AFAIK streams are smart enough to combine limit with sorted.

Also suited to extract this into an extra method

game-app/game-core/src/main/java/games/strategy/triplea/ui/panel/move/MovePanel.java
asvitkine: Made a helper called `canSelectUnits()` instead, since that seemed cleaner.

game-app/game-core/src/main/java/games/strategy/triplea/ui/panel/move/MovePanel.java
RoiEXLab: I believe it's fine because in this context, but it kinda violates the functional paradigm that there should be no side effects.
Regardless `.forEach` doesn't guarantee any order, `.forEachOrdered` does. Most likely they have the same implementation in this case, but I'd prefer the explicit variant here to make clear order is important

game-app/game-core/src/main/java/games/strategy/triplea/ui/panel/move/MovePanel.java
```diff
@@ -233,21 +223,13 @@ private void selectUnitsToMove(
           } else { // add one
             // best candidate unit for route is chosen dynamically later
             // check for alt key - add 1/10 of total units (useful for splitting large armies)
-            final List<Unit> unitsToMove = CollectionUtils.getMatches(units, unitsToMoveMatch);
-            unitsToMove.sort(UnitComparator.getHighestToLowestMovementComparator());
-
             final int maxCount = mouseDetails.isAltDown() ? deselectNumber : 1;
```
RoiEXLab, line 226: Nit:
`deselectNumber` should ideally be `DESELECT_NUMBER` (and maybe even a better name at all?)

I haven't checked if this constant is used anywhere else, but in case it isn't maybe there's no need to give it a name at all and just inline the value and adjust the comment accordingly (which is also kinda confusing to me).

Kinda out of scope for this PR, but if it isn't a major inconvenience for you I'd appreciate the change nevertheless.

RoiEXLab (approved):

game-app/game-core/src/main/java/games/strategy/triplea/delegate/TransportTracker.java
RoiEXLab: Even better

game-app/game-core/src/main/java/games/strategy/triplea/ui/panel/move/MovePanel.java
asvitkine: Done.

game-app/game-core/src/main/java/games/strategy/triplea/ui/panel/move/MovePanel.java
asvitkine: Done, changed to MULTI_SELECT_NUMBER. Also updated some comments.

PR #10374, opened 2022-04-26
Reduce Duplicate Code

Attacktive:

Inspired by [the report](https://codeclimate.com/github/triplea-game/triplea/pull/10368#issue_6266e68d1e50e40001000ad6), I decided to try some refactoring.
Those comparators for [ProductionRule](https://github.com/triplea-game/triplea/compare/master...Attacktive:master#diff-72fe51e0a7238aa9b8c3e5345de919940f87102159e713078d75b4eff4ef9903L42) and [RepairRule](https://github.com/triplea-game/triplea/compare/master...Attacktive:master#diff-72fe51e0a7238aa9b8c3e5345de919940f87102159e713078d75b4eff4ef9903L84) are pretty much the same, so here's my two cents on them.

I introduce a new interface [Rule](https://github.com/triplea-game/triplea/compare/master...Attacktive:master#diff-1fc5c20175ae8cf69f7c6fbd1fe7a7a4d97b99fdf082de528070e5eed086635dR5) and a class [RuleComparator](https://github.com/triplea-game/triplea/compare/master...Attacktive:master#diff-2ee2c089e9754d7b603b06b5d4afc3308ac29d71e4e47a0b30c72e60673154dbR8).
Here, the interface `Rule` definitely needs to be renamed along with the comparator because it's too ambiguous,
but I'd like someone else to do for me since I'm new to *TripleA*.


asvitkine:
Nice! Perhaps instead of the interface, they can share a parent class? This way, some of the implementation can be shared as well.

One thing we should check is whether this affects save game compatibility. If rules get serialized in saved games, then it may not be possible to change the class hierarchy until we do the next major update that breaks save compatibility.

Attacktive:
> Perhaps instead of the interface, they can share a parent class? This way, some of the implementation can be shared as well.

Sounds good. I'll give it a shot.

> One thing we should check is whether this affects save game compatibility.

Indeed I haven't even thought about that. ☠
Can I get help with the test?

RoiEXLab:
@asvitkine @Attacktive I'm almost completely certain that adding an interface to a class does not affect java serialization in any way and thus save-game compatibility. That being said I'm not so sure about introducing parent classes. So a good compromise might be to use default-methods in interfaces if no local field is required

game-app/game-core/src/main/java/games/strategy/triplea/ui/SimpleUnitPanel.java
```diff
@@ -41,85 +41,31 @@ public enum Style {
...
-            }
-            return n1.getName().compareTo(n2.getName());
-          }
-          if (o1.getResults().size() > o2.getResults().size()) {
-            return -1;
-          } else if (o1.getResults().size() < o2.getResults().size()) {
-            return 1;
+          var comparison = ruleComparator.compare(o1, o2);
+          if (comparison == 0) {
+            return o1.getName().compareTo(o2.getName());
```
RoiEXLab: If `RuleComparator` was made generic, i.e. the class definition changed to
```java
public class RuleComparator<T extends Rule> implements Comparator<T> {
  @Override
  public int compare(T o1, T o2) {
    // your code here
  }
}
```

there would be a super simple way to shorten this code like this:
```java
private final Comparator<RepairRule> repairRuleComparator = new RuleComparator<RepairRule>().thenComparing(RepairRule::getName);
```

Alternatively you might be able to extend the Rule interface to include `getName()`, not sure if that collides with the `DefaultNamed` class though. But if it works you could merge `productionRuleComparator` and `repairRuleComparator` into a single comparator without any generics, just a simple interface declaration

RoiEXLab: Also if you're able to unify those two classes into a single interface you could probably just move the final comparison into the actual class I think

game-app/game-core/src/main/java/games/strategy/triplea/ui/SimpleUnitPanel.java
```diff
@@ -175,7 +89,7 @@ public void setUnitsFromRepairRuleMap(
     final Set<Unit> entries = units.keySet();
     for (final Unit unit : entries) {
       final IntegerMap<RepairRule> rules = units.get(unit);
-      final TreeSet<RepairRule> repairRules = new TreeSet<>(repairRuleComparator);
+      final TreeSet<RepairRule> repairRules = new TreeSet<>(RepairRule.getComparator());
```
RoiEXLab, line 92: I like your solution to the problem, but I think it's better to have the abstract `Rule` class to implement `Comparable<Rule>` instead of providing a static method, then you can just omit this argument entirely.

RoiEXLab: If the superclass works with compatibility there's no reason to change your code, it's pretty good already.
Just make sure to have a look at my newer comment about implementing the Comparable interface directly

RoiEXLab: This is just about making the method non-static

Attacktive: Thanks. It looks much better. 😁

asvitkine:
To test whether this creates a save game problem, you can save a game (on a complex map that uses lots of features - maybe Pacific Solo Challenge) with TripleA before the changes.

Then, you can try to load it up using the version with the changes.

To do a more detailed test, you can add the following method in the classes:

```
 private void readObject(ObjectInputStream in) throws ClassNotFoundException, IOException {
    in.defaultReadObject();
}
 ```

And add a line there to print the properties. You can do this with the changes and without the changes and see if the all the properties have been read the same way.

Attacktive:
Thank you, @asvitkine. I'll let you know after test.

Attacktive:
It turns out that the changes I've made DOES affect serialization. ☠
I'll close it and create another PR using interfaces because it's a little bit too messy now.

Thank you guys.

Issue #9584, opened 2021-09-09
Axis and Allies "Pacific" (Pacific '00) map bug report


Thegoodcaptain:
**Escorts/Interceptors and Strategic Bombing raids in Pacific:**
How it currently works:
Anti-aircraft fire occurs after the air battle.

![image](https://user-images.githubusercontent.com/79759787/132608739-7f9a4ce0-82f9-4775-9f05-9b8a95607049.png)

 
How it should work:

AA fire should be resolved FIRST against ALL attacking bombers and escorting fighters. If there are any hits. the owning player selects which units (fighters and/or bombers) take which hits. AFTER this, surviving escorts engage in the Air Battle with interceptors (page 19).

**First phase for naval combat sequence for attacking submarines:**

How it currently works:

First phase - “German (Japanese) first strike units fire”

![image](https://user-images.githubusercontent.com/79759787/132608818-e16fa08d-0963-4f32-810b-541cb369577e.png)
 
How it should work:

First phase: “Germans (Japanese) submerge subs?”

*i.e. attacking subs can choose to engage in combat or not regardless of the presence of a destroyer. A subs ability to submerge is never negated by the presence of a destroyer, only its “first strike” (page 33). Yes, this will create a hostile sea zone at the end of the turn. This is intentional. Otherwise subs may not move into enemy occupied sea zones (page 24).

**Defending submarines in combat sequence:**

How it currently works:
 
![image](https://user-images.githubusercontent.com/79759787/132608882-e6546a33-9828-4131-a35a-1ffd4dd0c3ed.png)

How it should work:

FIX #1: First phase: “Russians fire” OR (if subs present “Russians first strike units fire”). Defending submarines do not qualify for first strike (page 33).

FIX #2: After “Germans select casualties” it should read, “Germans submerge subs?”. Weather or not the German sub was selected as a casualty, this option is available to any defending submarines. If the defender wishes to “submerge” the sub, it is removed from the battle and placed back on the map in “submerged” status. It does not get to return fire (page 10). It is “resurfaced” after the place new units phase (page 10). Yes, this can create a hostile sea zone at the end of the turn. This is intentional.

**Multi-player Forces:**

How it currently works:
 
![image](https://user-images.githubusercontent.com/79759787/132608898-ce3d9cd7-bd1b-4fdc-93be-7c8c22e22d15.png)

[...]

PR #10377, opened 2022-04-27
Various clean ups to Matches and related code.

asvitkine:

Various clean ups to Matches and related code.

Improves the code by adopting more concise notation, improving readability and removing some unnecessary null checks and collection copies.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
@@ -1689,8 +1689,7 @@ public final class Matches {
       if (unitIsSea().test(unitWhichWillGetBonus)) {
         final Predicate<Unit> givesBonusUnitLand = givesBonusUnit.and(unitIsLand());
         final Collection<Territory> neighbors = gameMap.getNeighbors(territory, territoryIsLand());
-        return neighbors.stream()
-            .anyMatch(t -> t.anyUnitsMatch(givesBonusUnitLand));
+        return neighbors.stream().anyMatch(t -> t.anyUnitsMatch(givesBonusUnitLand));
       }
       return false;
     };
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
@@ -1152,6 +1152,7 @@ public final class Matches {
       if (ua.getIsSea() && stepName.getDisplayName().equals("Non Combat Move")) {
         // If a zone adjacent to the starting and ending sea zones are allied naval bases, increase
         // the range.
+        // TODO Still need to be able to handle stops on the way
         // (history to get route.getStart()
         for (final Territory terrNext : unit.getData().getMap().getNeighbors(route.getStart(), 1)) {
           final TerritoryAttachment taNeighbor = TerritoryAttachment.get(terrNext);
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
@@ -3,7 +3,6 @@ package games.strategy.triplea.delegate;
 import games.strategy.engine.data.GameData;
 import games.strategy.engine.data.GameMap;
 import games.strategy.engine.data.GamePlayer;
-import games.strategy.engine.data.GameStep;
 import games.strategy.engine.data.RelationshipTracker;
 import games.strategy.engine.data.RelationshipTracker.Relationship;
 import games.strategy.engine.data.RelationshipType;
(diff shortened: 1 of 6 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
```diff
@@ -1146,39 +1138,21 @@ public static Predicate<Unit> unitHasEnoughMovementForRoute(final Route route) {
...
-                  && taEndNeighbor.getNavalBase()
-                  && unit.getData().getRelationshipTracker().isAllied(terrEnd.getOwner(), player)) {
-                left = left.add(BigDecimal.ONE);
-                break;
-              }
-            }
-          }
+        if (hasNeighboringAlliedNavalBase(route.getStart(), unit.getOwner())
+            && hasNeighboringAlliedNavalBase(route.getEnd(), unit.getOwner())) {
+          left = left.add(BigDecimal.ONE);
```
asvitkine, line 1155: Doing a self-review, I wasn't sure if this was quite right. The previous logic looped through neighbours of the start territory to find naval bases and for each, looped through neighbors of an end territory to find naval bases.

Theoretically, it could have increased movement by more than one (under very specific circumstances), but this was actually a bug!

According to https://axisandallies.fandom.com/wiki/Naval_Base, the logic should be:
>  Adds 1 to non-combat sea movement, if the ships starts and ends adjacent to friendly naval bases

Which matches what the new implementation does.


<br class="Apple-interchange-newline">

asvitkine: However, the above link also says that the logic should be different for "AAG40" rule set. I don't see this here so I'm wondering if that's a missing implementation in TripleA?

asvitkine: I guess TripleA doesn't support "Axis & Allies: WWI 1914" - either the map or the rule set?

asvitkine: So just to be clear, the logic change is as follows:

Previously, if a sea unit was at a territory adjacent to multiple naval bases and had a route to a territory next to another naval base, then it's movement would be increased by the number of naval bases it was adjacent to! That doesn't seem to match the rules.

Curiously, looking at the actual rules from 1940 global 2nd edition, it doesn't seem to talk about the route *ending* next to a naval base:

https://www.axisandallies.org/wp-content/uploads/2011/12/Axis-Allies-Europe-1940-Second-Edition.pdf
> 
> Increased Sea Unit Range: All sea units beginning their
> movement from a sea zone serviced by an operative
> friendly naval base gain 1 additional point of movement
> range. Sea units in this position can move 3 spaces
> instead of 2.

So, why do we have different logic here?

asvitkine: So, in-game it doesn't seem to work this way - it seems to match the PDF rules (no need for naval base at destination and no multiple bonuses). Perhaps the "source of truth" logic is elsewhere and not here? I can dig more.

If so, it also means there's a bug in this logic.

asvitkine: So, searching around, I found https://github.com/triplea-game/triplea/issues/9584 which mentions the other version of Naval bases which matches the wiki. It quotes a manual that I found here:
https://kipdf.com/gameplay-manual-players-2-to-3-ages-12-and-up-what-s-different-about-axis-allies_5ae431a97f8b9ab3818b45b5.html

> During the Non-combat Movement phase, naval units moving from a sea zone containing a friendly naval base to another sea zone containing a friendly naval base may move three spaces instead of two. Either, or both, of these naval bases may have been captured on this turn. 

But https://github.com/triplea-game/triplea/issues/9584 seems to suggest the current engine doesn't support this. So this code is quite curious.

asvitkine: OK, so after digging a bit more, here are my findings:
  - This code actually does work (and has the bug I mentioned which this is fixing), when there is actually a "Naval base" present.
  - However, most maps don't use this "Naval base" feature - instead, Naval bases on maps like 1940 Global instead just boost units movement using a different function.
  - On "Global War 2", this feature is used: https://github.com/search?q=org%3Atriplea-maps+navalBase&type=code
  - On that map, we can see the bug in action for the British ships in SZ 8, which has two naval bases neighbouring. We can see that the ships there can move 4 spaces to SZ 5, instead of 3 - and with this fix, it becomes 3 as it should be.
  - Having said that, there is no test of this code at all - just removing this logic doesn't fail anything.

So per the above, I'm going to extract this specific part of the PR into a separate one, so this one is kept as "no functional changes" - and the other one can focus on this improvement along with some tests.

asvitkine:
Committing after self-review.

PR #10378, opened 2022-04-27
Reduce Duplicate Code by Introducing an Interface and a Comparator for It.

Attacktive:

It's an alternative to [this](https://github.com/triplea-game/triplea/pull/10374).
Introducing a common `abstract class` in my case broke the save game compatibility hence an `interface` this time.

As @asvitkine [suggested](https://github.com/triplea-game/triplea/pull/10374#issuecomment-1109796034), I did the save game compatibility tests with the code below and the output wasn't affected by the changes.
```java
private void readObject(java.io.ObjectInputStream in) throws ClassNotFoundException, java.io.IOException {
  in.defaultReadObject();

  System.out.format("%s%ncosts: %s%nresults: %s%n", getClass().getSimpleName(), getCosts(), getResults());
}
```

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/Rule.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/Rule.java
@@ -1,12 +1,18 @@
 package games.strategy.engine.data;
 
+import org.triplea.java.ChangeOnNextMajorRelease;
 import org.triplea.java.collections.CollectionUtils;
 import org.triplea.java.collections.IntegerMap;
 
 /**
  * Superclass for {@link games.strategy.engine.data.RepairRule} and {@link
  * games.strategy.engine.data.ProductionRule}.
+ *
+ * <p>It could become an {@code abstract class} instead in to contain more common code chunks from
+ * its children. That way, probably {@link games.strategy.engine.data.RuleComparator} can be
+ * entirely removed by making the child classes implement {@link java.lang.Comparable}.
  */
+@ChangeOnNextMajorRelease
 public interface Rule {
   String getName();
 
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/Rule.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/Rule.java
@@ -14,7 +14,12 @@ public interface Rule {
 
   IntegerMap<NamedAttachable> getResults();
 
-  /** Benefits must be a resource or a unit. */
+  /**
+   * Benefits must be a resource or a unit.
+   *
+   * <p>It's not going to work if any of {@link Rule#getResults()} implementations returns, for
+   * example, a copy of {@link org.triplea.java.collections.IntegerMap}.
+   */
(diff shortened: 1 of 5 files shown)


game-app/game-core/src/main/java/games/strategy/engine/data/Rule.java
```diff
@@ -0,0 +1,35 @@
+package games.strategy.engine.data;
+
+import org.triplea.java.collections.CollectionUtils;
+import org.triplea.java.collections.IntegerMap;
+
+/**
+ * Superclass for {@link games.strategy.engine.data.RepairRule} and {@link
+ * games.strategy.engine.data.ProductionRule}.
+ */
+public interface Rule {
```
asvitkine, line 10: You can use the `@ChangeOnNextMajorRelease` annotation here and mention that this can become an abstract base class in the future.

game-app/game-core/src/main/java/games/strategy/engine/data/RuleComparator.java
```diff
@@ -0,0 +1,52 @@
+package games.strategy.engine.data;
+
+import games.strategy.triplea.attachments.UnitTypeComparator;
+import java.util.Comparator;
+
+public class RuleComparator<T extends Rule> implements Comparator<T> {
+  private static final UnitTypeComparator unitTypeComparator = new UnitTypeComparator();
```
asvitkine, line 7: I would suggest to not make this static. If the unit type comparator makes use of any caching that assumes the underlying units don't change, then having a static instance can lead to bugs.

game-app/game-core/src/main/java/games/strategy/engine/data/RuleComparator.java
```diff
@@ -4,7 +4,7 @@
 import java.util.Comparator;
 
 public class RuleComparator<T extends Rule> implements Comparator<T> {
-  private static final UnitTypeComparator unitTypeComparator = new UnitTypeComparator();
```
Attacktive: This should have been in another new commit. 😂

game-app/game-core/src/main/java/games/strategy/engine/data/Rule.java
Attacktive: Good to know. [I added the annotation](https://github.com/triplea-game/triplea/pull/10378/files#diff-1fc5c20175ae8cf69f7c6fbd1fe7a7a4d97b99fdf082de528070e5eed086635dR15).

game-app/game-core/src/main/java/games/strategy/engine/data/RuleComparator.java
Attacktive: > I would suggest to not make this static. If the unit type comparator makes use of any caching that assumes the underlying units don't change, then having a static instance can lead to bugs.

Haven't thought about that. I've been learning a lot here. 😁
Thank you.

asvitkine (approved):

Issue #10362, opened 2022-04-24
2.5.22294: TransportUtils#findOptimalTransportToUnloadFrom:308 - java.lang.IndexOutOfBoundsException


tripleabuilderbot:
## User Description
Playing "The Rising Sun" version 1.9.3, Round 3, Yamamoto, Combat move phase

In sz48, I had several tranports, all packed with 2 units.

I had one of my transports unload a single infantry unit to Ceram, but did not unload the artillery that was accopanying that unit in the transport. 

I then tried to unload all of my remaining units onto Sarong. The game did not allow me to (as that violates the rules, 1 transport can't delive to 2 different locations). I then tried moving all of my land units, except 1 artillery, onto Sarong. This is a legal move, but this caused the javascript error.

## Map
the_rising_sun

## Log Message
Index 0 out of bounds for length 0

## TripleA Version
2.5.22294

## Java Version
11.0.14.1

## Operating System
Linux

## Memory
Heap utilization statistics [MB]
Used Memory: 122
Free memory: 1925
Total memory: 2048
Max memory: 2048

## Stack Trace
```
Exception: java.lang.IndexOutOfBoundsExceptionIndex 0 out of bounds for length 0
java.lang.Exception
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:248)
	at java.base/java.util.Objects.checkIndex(Objects.java:372)
	at java.base/java.util.ArrayList.get(ArrayList.java:459)
	at games.strategy.triplea.util.TransportUtils.findOptimalTransportToUnloadFrom(TransportUtils.java:308)
	at games.strategy.triplea.util.TransportUtils.findMinTransportsToUnload(TransportUtils.java:157)
	at games.strategy.triplea.ui.panel.move.MovePanel.getUnitsToUnload(MovePanel.java:847)
	at games.strategy.triplea.ui.panel.move.MovePanel$1.selectEndPoint(MovePanel.java:566)
	at games.strategy.triplea.ui.panel.move.MovePanel$1.unitsSelected(MovePanel.java:150)
	at games.strategy.triplea.ui.panels.map.MapPanel.notifyUnitSelected(MapPanel.java:519)
[...]

PR #10381, opened 2022-04-27
Fix issue preventing units being unloaded from transports.

asvitkine:

This fixes an issue where the move UI could prevent the user from making valid transport unload moves.

The core of the issue is that when you select a singular unit from a sea territory where there may be several units of that type with different restrictions (e.g. on transports that have previously unloaded to different territories), there was no logic to determine which unit should be selected. In fact, when you just click the unit, it's impossible to choose the "right" one at that point, given we don't know where the user will move that unit.

The fix is a new TransportUtils function that can substitute equivalent units based on destination territory, which is then used when validating the pending move as the mouse is moved around (for the purpose of the UI showing whether the move is OK or not).

Additionally, when the move is actually performed (i.e. end point clicked), a couple of early returns are modified to still use the algorithm for selecting which units/transports to use for unloading, as opposed to just using the selected unit. 

With the above fixes, the following now works:
  1. Load the save game attached here: https://github.com/triplea-game/triplea/issues/10362#issuecomment-1108637030
  2. Move 2 arti to Iwo Jima, one at a time.
  3. Undo both moves.
  4. Try to re-do the moves in 2.

Without this change, step 4 fails via the UI disallowing the moves. With this change, the UI allows the moves to be made.

## Release Note

FIX|Some valid transport unloads being prevented by the UI

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/util/TransportUtils.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/util/TransportUtils.java
@@ -18,6 +18,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import org.triplea.java.collections.CollectionUtils;
 import org.triplea.java.collections.IntegerMap;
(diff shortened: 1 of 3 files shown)


RoiEXLab:
@asvitkine I'll try to have a look at your PRs tomorrow, haven't been able to do so in the last couple of days because of some irl stuff. Haven't forgotten about them

asvitkine:
Thanks! No worries, totally understand.

game-app/game-core/src/main/java/games/strategy/triplea/util/TransportUtils.java
```diff
@@ -341,4 +343,46 @@ private static void removeTransportAndLoadedUnits(
...
+        // account previous replacements.
+        route.getStart().getUnits().stream()
+            .filter(u -> u.isEquivalent(origUnit))
+            .filter(Predicate.not(updatedUnits::contains))
+            .filter(u -> canUnload(u, route.getEnd()))
+            // Replace the orig unit with the equivalent one.
+            .findAny()
+            .ifPresent(u -> updatedUnits.set(unitIndex, u));
+      }
+    }
```
RoiEXLab, line 380: What do you think about using a stream map function? Something along the lines of:
```java
units.stream()
  .map(origUnit -> {
    if (!canUnload(origUnit, route.getEnd())) {
      return route.getStart().getUnits().stream()
        .filter(u -> u.isEquivalent(origUnit))
        .filter(Predicate.not(units::contains))
        .filter(u -> canUnload(u, route.getEnd()))
        .findAny()
        .orElse(origUnit);
    }
    return origUnit;
  })
```

This avoids the complexity the index introduces, and I believe should behave the same as your current code (note how I changed `updatedUnits` to `units` in the filter)

RoiEXLab (approved):

game-app/game-core/src/main/java/games/strategy/triplea/util/TransportUtils.java
RoiEXLab: Ah, that's what I was concerned about. Makes sense, but unfortunately it doesn't really make the code simpler to understand :/

Issue #10385, opened 2022-04-28
TripleA 2.6.714 null error


TheDog-GH:
@asvitkine 
You have been **very busy**, thanks for all your work!

On the forums I have posted this, as more people will see it.
https://forums.triplea-game.org/topic/3206/triplea-2-6-714-null-error

TheDog-GH:
@asvitkine 
Quick test, its working.
On another test, my speed test on a simple map up to turn 6, it is the fastest version yet of TripleA.
THANKS!

PR #10386, opened 2022-04-28
Fix NPE due to AI code calling getNeighbors() on null.

asvitkine:
Fixes https://github.com/triplea-game/triplea/issues/10385.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java
@@ -396,7 +396,7 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
 
     final Set<Territory> newFrontier =
         frontier.stream()
-            .flatMap(f -> getNeighbors(f).stream().filter(t -> routeCond.test(f, t)))
+            .flatMap(f -> getNeighbors(f, routeCond).stream())
             .collect(Collectors.toSet());
     if (newFrontier.contains(target)) {
       return distance + 1;
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java
@@ -74,7 +74,7 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
 
   private void setConnection(final Territory from, final Territory to) {
     // preserves the unmodifiable nature of the entries
-    final Set<Territory> current = connections.get(from);
+    final Set<Territory> current = getNeighbors(from);
     final Set<Territory> modified = new HashSet<>(current);
     modified.add(to);
     connections.put(from, Collections.unmodifiableSet(modified));
@@ -115,7 +115,7 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
 
   private Set<Territory> getNeighbors(
       final Territory territory, final BiPredicate<Territory, Territory> routeCondition) {
-    return connections.get(territory).stream()
+    return getNeighbors(territory).stream()
         .filter(n -> routeCondition.test(territory, n))
         .collect(Collectors.toSet());
   }
@@ -396,7 +396,7 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
 
     final Set<Territory> newFrontier =
         frontier.stream()
-            .flatMap(f -> connections.get(f).stream().filter(t -> routeCond.test(f, t)))
+            .flatMap(f -> getNeighbors(f).stream().filter(t -> routeCond.test(f, t)))
             .collect(Collectors.toSet());
     if (newFrontier.contains(target)) {
       return distance + 1;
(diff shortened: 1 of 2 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProNonCombatMoveAi.java
```diff
@@ -102,9 +102,13 @@ Map<Territory, ProTerritory> doNonCombatMove(
...
-    final int enemyDistance =
-        ProUtils.getClosestEnemyLandTerritoryDistance(data, player, proData.getMyCapital());
-    moveUnitsToDefendTerritories(prioritizedTerritories, enemyDistance);
+    final Territory myCapital = proData.getMyCapital();
+    int enemyDistanceToMyCapital = Integer.MAX_VALUE;
+    if (myCapital != null) {
+      enemyDistanceToMyCapital =
+          ProUtils.getClosestEnemyLandTerritoryDistance(data, player, myCapital);
+      moveUnitsToDefendTerritories(prioritizedTerritories, enemyDistanceToMyCapital);
+    }
```
DanVanAtta, line 111: Have you tested out this logic block?

Just wondering how the block below would behave with a large value of `enemyDistanceToMyCapital` would work below with:
```
defenseRange = enemyDistanceToMyCapital - 1;
```

Side-comment: I think it's a shame that `enemyDistanceToMyCapital` variable was used for such a large scope, notably 40 lines later.. quite a bit for a pretty innocuous looking local variable.

asvitkine: I believe the block below is inside a capital is not null check, so wouldn't be hit in this case.

game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java
```diff
@@ -396,7 +396,7 @@ private int getDistance(
 
     final Set<Territory> newFrontier =
         frontier.stream()
-            .flatMap(f -> connections.get(f).stream().filter(t -> routeCond.test(f, t)))
+            .flatMap(f -> getNeighbors(f).stream().filter(t -> routeCond.test(f, t)))
```
DanVanAtta, line 399: `getNeighbors` is 'fancy' with a lot of overrides so it can accept a filtering condition.

Presumably this could be something like `.flatMap(frontier -> getNeighbors(frontier, 1, (start, end) -> routeCond.test(start, end));`

Though... I'm looking at `getNeighbors` and feel like it is eagerly DRY, would be better more explicit. In this case avoiding `getNeighbors` would be a good move. In this case, perhaps it would be better to do some simple java:
``` Optional.ofNullable(connections.get(f)).orElse(Set::of).stream....```

This is not a critical change, just pointing out that it's a bit odd for `getNeighbors` to be super overloaded to do route filtering, and then here use a default empty route filter in `getNeighbors` to then do route filtering (IMO, having the route filtering be here and not in the method is more clear, FWIW).

Comment (asvitkine) [reply game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java:399]: The one-arg version of getNeighbors is just a wrapper over connections.get with a null check that provides a more useful error message. So I swapped all the connections.get calls to it.

Agree that this could be improved further.

Comment (asvitkine) [reply game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java:399]: As to orElse(), I explicitly omitted it since I believe only time it should happen is a coding error - e.g. passing null or a territory from a different game object.

Comment (asvitkine) [reply game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java]: Anyway, in this case you're right that it was pretty trivial to inline the filtering into this call site. Done.

Comment (asvitkine) [reply game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java]: Re-reading I see you actually suggested to not use that for DRY. I can revert back if you prefer.

But expanding on my previous point, the getNeighbors() call was just calling the very base version that's just a wrapper over connections.get().

```
  public Set<Territory> getNeighbors(final Territory territory) {
    // ok since all entries in connections are already unmodifiable
    final Set<Territory> neighbors = connections.get(territory);
    if (neighbors == null) {
      throw new IllegalArgumentException("No neighbors for: " + territory);
    }
    return neighbors;
  }
```

Doing `orElse()` makes the code inconsistent - where some places handle the null and others throw an error. By calling this function and it being the only place that does `connections.get()` we make all the logic consistent with regards to expectations and handling of that case.

As an aside, I'm not very keen on the fact that most of GameMap is filled with different variants of getNeighbors(), getRoute(), getDistance(), etc, that all do mostly the same thing, with a complicated control flow and inefficiency of creating lots of intermediate collections.

IMHO, ideally we could extract everything except the base version above that just returns immediate neighbors - and use a helper class for all the other variants. I think I started towards that at some point with the BFS class, but there's quite a bit work to be done to make it handle the different cases and to change call sites. :\

As for this change, let me know what you prefer. For the NPE fix, we could just revert all the changes in GameMap - they're not strictly needed. But I thought they were worthwhile to make to handling of connections.get() consistent across the different places that call it. Another option is to remove the extra check in the base getNeighbors() impl and keep everyone using connections.get() directly. Downside is the error would just be a plain NPE which wouldn't be as obvious the cause of.

asvitkine:
@DanVanAtta let me know what you think about the above. I wanted to give a chance for you to look again since you made some comments (and I replied).

asvitkine:
> @DanVanAtta let me know what you think about the above. I wanted to give a chance for you to look again since you made some comments (and I replied).

I'm going to go ahead and merge this. The `enemyDistanceToMyCapital` part is safe because we never evaluate the Integer.MAX_VALUE value since when it's used, it would have already been assigned. For `getNeighbors()`, we can discuss more about - but I want to resolve the NPE in the meantime since it makes AIs crash with the prerelease on some maps.

PR #10388, opened 2022-04-28
Clean up some move related code and matches.

asvitkine:

- Remove some unnecessary arrived.isEmpty() checks in a branch where that's already been checked.
- Rename some matchers to make more sense and streamline their implementations.
- Remove unnecessary params.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
@@ -1280,12 +1280,18 @@ public final class Matches {
                 && relationshipTracker.isAtWar(player, t.getOwner()));
   }
 
-  public static Predicate<Territory> isTerritoryEnemyAndNotUnownedWaterOrImpassableOrRestricted(
+  public static Predicate<Territory> isTerritoryEnemyAndNotImpassableOrRestricted(
       final GamePlayer player,
       final GameProperties properties,
       final RelationshipTracker relationshipTracker) {
+    return isTerritoryEnemy(player, relationshipTracker)
+        .and(territoryIsPassableAndNotRestricted(player, properties));
+  }
+
+  public static Predicate<Territory> territoryIsNotUnownedWaterAndCanBeTakenOverBy(
+      final GamePlayer attacker) {
     return t -> {
-      if (t.isOwnedBy(player)) {
+      if (t.isOwnedBy(attacker)) {
         return false;
       }
       // if we look at territory attachments, may have funny results for blockades or other things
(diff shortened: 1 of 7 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
```diff
@@ -1281,21 +1287,24 @@ public static Predicate<Territory> isTerritoryEnemyAndNotUnownedWater(
...
+    final RelationshipTracker relationshipTracker = player.getData().getRelationshipTracker();
+    return territoryNotImpassibleOrRestrictedOrNeutralWaterAndNotOwnedBy(player)
+        .and(t -> relationshipTracker.canTakeOverOwnedTerritory(player, t.getOwner()));
+  }
+
+  private static Predicate<Territory> territoryNotImpassibleOrRestrictedOrNeutralWaterAndNotOwnedBy(
+      final GamePlayer player) {
+    return not(isTerritoryOwnedBy(player))
+        .and(not(territoryIsUnownedWater()))
+        .and(territoryIsPassableAndNotRestricted(player, player.getData().getProperties()));
```
RoiEXLab, line 1307: If you're simplifying the method signature anyways, would it be a benefit to have `territoryIsPassableAndNotRestricted` only accept a `GamePlayer` and derive the properties from that? Or is this just a special case that happens only here?

RoiEXLab (approved):

game-app/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
asvitkine: One danger is if the null player instance is ever passed, that one will not have a null data. I filed an issue about getting rid of a global null player instance and instead having a null instance associated with the game data, but that's a bigger refactoring.

So it's not necessarily safe to do - depends on call sites.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/MovePerformer.java
```diff
@@ -256,9 +254,7 @@ public void execute(final ExecutionStack stack, final IDelegateBridge bridge) {
               // Ignore Trn on Trn forces.
               if (Properties.getIgnoreTransportInMovement(bridge.getData().getProperties())) {
                 final boolean allOwnedTransports =
-                    !arrived.isEmpty()
-                        && arrived.stream()
-                            .allMatch(Matches.unitIsTransportButNotCombatTransport());
```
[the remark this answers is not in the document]
asvitkine: Ack.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
asvitkine: Keeping as-is per the above. Once we can change the engine such that the null player is guaranteed to have a non-null data, I'd be more comfortable refactoring broader in this way.

PR #10394, opened 2022-04-29
Refactor code for querying unit tech for players.

asvitkine:

This change makes TechTracker an actual instance class on GameData and cleans up the APIs to retrieve tech information from UnitAttachment.

This is a pure code re-organizing change, no functional changes.

Besides improving the API, this should pave way for allowing the TechTracker to cache this information and only update it when tech changes, which should speed up the engine quite a bit - since very basic things like getting unit attack / defense / movement currently requires expensive tech look up operations (which shows up in profiles). I plan to send the follow-up PR to implement this caching - preliminary results show that it can speed up battle simulations by AI by ~25%.

Tested loading and saving games.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/TechTracker.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/TechTracker.java
@@ -11,14 +11,12 @@ import games.strategy.triplea.attachments.TechAbilityAttachment;
 import games.strategy.triplea.attachments.TechAttachment;
 import java.util.ArrayList;
 import java.util.Collection;
+import lombok.AllArgsConstructor;
 
 /** A collection of methods for tracking which players have which technology advances. */
+@AllArgsConstructor
 public final class TechTracker {
-  private GameData data;
-
-  public TechTracker(final GameData data) {
-    this.data = data;
-  }
+  private final GameData data;
 
   private Collection<TechAdvance> getCurrentTechAdvances(GamePlayer player) {
     return getCurrentTechAdvances(player, data.getTechnologyFrontier());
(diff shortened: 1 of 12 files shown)


asvitkine:
@DanVanAtta Please take a look. The codeclimate check is about number of methods in the class, which I don't think we can avoid here.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/TechTracker.java
```diff
@@ -1,17 +1,68 @@
...
+  public int getMovementBonus(GamePlayer player, UnitType type) {
+    return TechAbilityAttachment.getMovementBonus(type, getCurrentTechAdvances(player));
+  }
+
+  public int getAttackBonus(GamePlayer player, UnitType type) {
+    return TechAbilityAttachment.getAttackBonus(type, getCurrentTechAdvances(player));
+  }
+
+  public int getAttackRollsBonus(GamePlayer player, UnitType type) {
+    return TechAbilityAttachment.getAttackRollsBonus(type, getCurrentTechAdvances(player));
```
RoiEXLab, line 38: Those static methods from `TechAbilityAttachment` seem really odd. The look like they have nothing to do with the actual class except that they all (?) share the helper method `TechAbilityAttachment#sumIntegerMap`.
I believe the newly created `TechTracker` class might be suited better for a lot of them, i.e. inlining the static methods might make the code cleaner.

I have to admit though, I'm not sure how  tangled up some of the helper methods are so maybe the rabbit hole of refactoring is way too deep for this PR. So please tell me what you think about this idea and if you'd consider it worth addressing

RoiEXLab (approved):

PR #10395, opened 2022-04-30
Clean up code and fix the logic for Pacific-ruleset Naval base move boosting.

asvitkine:

Clean up code and fix the logic for Pacific-ruleset Naval base move boosting. The previous code was messy and inefficient and actually had a bug where a unit starting next to multiple Naval bases could get multiple boosts, which does not match the Pacific rule set, although this situation doesn't appear on the Pacific map (and new Naval bases can't be built there).

There was no actual test coverage of this logic, although there was a test that intended to do this, but used a fighter unit instead of a naval unit in the test, thus not exercising the code. This PR expands this test to provide much more extensive test coverage of this logic.

## Release Note

FIX|Fixed the Pacific ruleset Naval base territory attachment logic

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
@@ -1152,6 +1152,7 @@ public final class Matches {
       if (ua.getIsSea() && stepName.getDisplayName().equals("Non Combat Move")) {
         // If a zone adjacent to the starting and ending sea zones are allied naval bases, increase
         // the range.
+        // TODO Still need to be able to handle stops on the way
         // (history to get route.getStart()
         for (final Territory terrNext : unit.getData().getMap().getNeighbors(route.getStart(), 1)) {
           final TerritoryAttachment taNeighbor = TerritoryAttachment.get(terrNext);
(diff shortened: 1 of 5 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/delegate/Matches.java
```diff
@@ -1148,28 +1146,15 @@ public static Predicate<Unit> unitHasEnoughMovementForRoute(final Route route) {
...
-              && taNeighbor.getNavalBase()
-              && unit.getData().getRelationshipTracker().isAllied(terrNext.getOwner(), player)) {
-            for (final Territory terrEnd :
-                unit.getData().getMap().getNeighbors(route.getEnd(), 1)) {
-              final TerritoryAttachment taEndNeighbor = TerritoryAttachment.get(terrEnd);
-              if (taEndNeighbor != null
-                  && taEndNeighbor.getNavalBase()
-                  && unit.getData().getRelationshipTracker().isAllied(terrEnd.getOwner(), player)) {
-                left = left.add(BigDecimal.ONE);
-                break;
```
RoiEXLab, line 1169: Wait, so the code didn't break the outer loop? That's a big oopsie! Good catch!

RoiEXLab (approved):
Very nice change

PR #10404, opened 2022-05-03
Add a units bar, showing units inside the territory after its name in the bottom bar.

asvitkine:
Add a units bar to the main UI, showing units inside the territory after its name at the bottom of the screen.
This feature makes it easy to see which units exist in the given territory, without having to open the battle calculator or switching to the territory tab (which loses focus on that territory when you move the mouse). This is especially helpful on some maps where there are small territories that can get cluttered with units and it's hard to see exactly what's there.

This change makes it so the units in the territory under the cursor are shown at the bottom of the screen, in the same cell in the bottom bar where the territory name is.

I've taken special care to test that this works well on a variety of maps and results in consistently good looking results.

In particular:
  - The content of the cell continues to be centered.
  - On smaller screens, priority is given to the territory name / resources if not everything can fit.
  - It works well with a variety of image sizes and presence of larger images doesn't cause layout issues.
  - A menu item is added to toggle the display of this bar, so users who don't like it can turn it off.

Tested on maps:
  - Pacific Theatre Solo Challenge
  - Civil War: Eastern Campaigns
    - This map has some units that are slightly taller, so verified they don't cause alignment issues
    - This map has some territories with no resources / units, verified that territory name label has consistent alignment
  - Warcraft
  - Star Trek Dilithium War
  - WW2 Oil and Snow
  - Conquest of the World
  - Another World
  - 1914-COW-Empires
  - Feudal Japan
  - Feudal Japan Warlords
  - 270BC Wars

Tested manually resizing window to smaller and larger sizes to verify layouts on different screen sizes.

Screenshots:

*Pacific Theatre Solo Challenge*, mouse over SZ 46 which has a big fleet:
units_bar_pacific

*Civil War: Eastern Campaigns*:
  - Cut off behavior when units don't fully fit the alloted space (I found the territory with most unit types and resized my window down to see this behavior).
  - OK handling of units that may be a bit taller (they just get cut off without causing other layout issues)
units_bar_civil_war

*Warcraft*:
  - Larger image sizes get cut off in a sensible way
units_bar_warcraft

*Conquest of the World*:
  - This map uses very large unit images. The unit bar doesn't look amazing, but also doesn't break catastrophically either. (Note: This map is not in the repo and must be downloaded manually from github). I've also not seen other maps using such extreme image sizes and every map I've tried otherwise generally looked reasonable.
unit_bar_conquest_of_the_world

## Release Note

NEW|The selected territory's units will now be shown at the bottom of the screen

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/BottomBar.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/BottomBar.java
@@ -160,15 +160,10 @@ public class BottomBar extends JPanel {
       if (production > 0) {
         resources.add(new Resource(Constants.PUS, data), production);
       }
-      final ResourceCollection resourceCollection = ta.getResources();
-      if (resourceCollection != null) {
-        resources.add(resourceCollection.getResourcesCopy());
-      }
+      Optional.ofNullable(ta.getResources()).ifPresent(r -> resources.add(r.getResourcesCopy()));
       for (final Resource resource : resources.keySet()) {
-        final JLabel resourceLabel =
-            uiContext.getResourceImageFactory().getLabel(resource, resources);
         territoryInfo.add(Box.createHorizontalStrut(10));
-        territoryInfo.add(resourceLabel);
+        territoryInfo.add(uiContext.getResourceImageFactory().getLabel(resource, resources));
       }
     }
 
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/BottomBar.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/BottomBar.java
@@ -48,7 +48,6 @@ public class BottomBar extends JPanel {
 
     resourceBar = new ResourceBar(data, uiContext);
 
-
     territoryUnitsPanel = new SimpleUnitPanel(uiContext, SimpleUnitPanel.Style.SMALL_ICONS_ROW);
     territoryUnitsPanel.setBorder(new EtchedBorder(EtchedBorder.RAISED));
 
@@ -73,11 +72,9 @@ public class BottomBar extends JPanel {
         territoryInfo,
         gridBuilder.gridX(1).weightX(1).anchor(GridBagConstraintsAnchor.CENTER).build());
 
-
     centerPanel.add(
         territoryUnitsPanel,
         gridBuilder.gridX(2).weightX(1).anchor(GridBagConstraintsAnchor.EAST).build());
-
 
     statusMessage.setPreferredSize(new Dimension(0, 0));
     statusMessage.setBorder(new EtchedBorder(EtchedBorder.RAISED));
@@ -174,7 +171,6 @@ public class BottomBar extends JPanel {
       territoryInfo.add(resourceLabel, gridBuilder.gridX(count++).build());
     }
     SwingComponents.redraw(territoryInfo);
-
 
     var cats = UnitSeparator.categorize(territory.getUnits());
     territoryUnitsPanel.setUnitsFromCategories(cats);
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/BottomBar.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/BottomBar.java
@@ -3,7 +3,6 @@ package games.strategy.triplea.ui;
 import games.strategy.engine.data.GameData;
 import games.strategy.engine.data.GamePlayer;
 import games.strategy.engine.data.Resource;
-import games.strategy.engine.data.ResourceCollection;
 import games.strategy.engine.data.Territory;
 import games.strategy.engine.data.TerritoryEffect;
 import games.strategy.triplea.Constants;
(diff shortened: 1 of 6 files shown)


Cernelius:
You may try Conquest of the World if you want to test it on a map with very big unit images.

Cernelius:
Rather than having "x X" after the unit image, I would have "X" before the unit image, where "X" is the number of units of the type. The different types may be separated by a "+" sign, but I'm not sure if this is preferable.

asvitkine:
> Rather than having "x X" after the unit image, I would have "X" before the unit image, where "X" is the number of units of the type. The different types may be separated by a "+" sign, but I'm not sure if this is preferable.

The "x X" variant is used elsewhere, so this is re-using that code. However, I see it actually adds quite a bit of space sometimes, so let me look at at least omitting it when it would be "x 1".

beelee1:
_On smaller screens, priority is given to the territory name / resources if not everything can fit._

Is there a scroll bar for this ? It seems a good idea as most maps probably have space for it

asvitkine:
> _On smaller screens, priority is given to the territory name / resources if not everything can fit._
> 
> Is there a scroll bar for this ? It seems a good idea as most maps probably have space for it

No scroll bar. It could be improved in the future, although I think one problem with a scrollbar is it the bar itself will take extra vertical space, leaving less room for the images. Perhaps a different scrolling UI could be used. On the other hand, I'm not sure whether doing it is a good use of effort given that there are other ways to see the list of units and it's not a problem on large screens. Still, this could still be improved further, for sure.

DanVanAtta:
I'll try to play test this when I get the chance (this month is slammed for me, not sure when it will be). 

@ron-murhammer had some strong thoughts on the status bar previously. He would be a good reviewer for the updated UX.

ron-murhammer:
Overall, I like the idea. Testing on a few more complex/large maps like Total World War and World at War would be good. I would put a little more space between the territory's resource production and its units to improve readability. The Civil War example is the one that is a bit unfortunate and not sure the best way to deal with maps that have large number of units.

Cernelius:
> > Rather than having "x X" after the unit image, I would have "X" before the unit image, where "X" is the number of units of the type. The different types may be separated by a "+" sign, but I'm not sure if this is preferable.
> 
> The "x X" variant is used elsewhere, so this is re-using that code. However, I see it actually adds quite a bit of space sometimes, so let me look at at least omitting it when it would be "x 1".

I don't like that convention, but I would make no exception for the "1" number. Can you tell us if that is an actual "multiplication" sign or it is just the "x" letter? If it is the "x" letter, it should be changed to be the proper sign in all places if so wanted. To save space, a centred point is a common multiplication symbol too (looking just like a full stop but placed as high as the centre of the images of the unit it references).

As a matter of being conventional, there is no sign on the battle board, so you may even consider having a stack number over-imposed to the unit just like it is on board, since the territory tab still offers the full visual of the units. That would probably both increase readability and save space. I would say that the best for visualization would be having the numbers under the unit images, but that would make the bottom bar ticker, reducing board view.

Side note, the "territory" tab should be called "zone" tab.

asvitkine:
> I would put a little more space between the territory's resource production and its units to improve readability.

I don't have updated screenshots up yet, but in my latest patch set, I'm using a JSeparator that provides a better spacing between units and the existing territory info, so I think this does the trick.

> The Civil War example is the one that is a bit unfortunate and not sure the best way to deal with maps that have large number of units.

Just to be clear, for that example, I explicitly picked a territory with lots of unit types (Washington) and I also made the window smaller than my current screen resolution explicitly for capturing a screenshot of the behavior of when there are too many units.

In my latest patch set, I did change it so single units don't show "x 1", again no screenshots up yet, which also saves some space.

> Can you tell us if that is an actual "multiplication" sign or it is just the "x" letter? If it is the "x" letter, it should be changed to be the proper sign in all places if so wanted. To save space, a centred point is a common multiplication symbol too (looking just like a full stop but placed as high as the centre of the images of the unit it references).

It is the letter "x", but I suppose it could be changed to the Unicode multiplication "×" sign:
https://www.compart.com/en/unicode/U+00D7

Since this will be beyond just this unit bar, I think we can do that separately and look at all the different places that are affected (with screenshots) to decide on the best visual look.

> As a matter of being conventional, there is no sign on the battle board, so you may even consider having a stack number over-imposed to the unit just like it is on board, since the territory tab still offers the full visual of the units. That would probably both increase readability and save space. I would say that the best for visualization would be having the numbers under the unit images, but that would make the bottom bar ticker, reducing board view.

This is not a bad idea, but it would require quite a bit extra code for - right now we just fetch the image from the map assets and put it into an icon on a JLabel without any transformation - if we wanted to draw custom numbers on top of it, it would require a whole bunch of special code just for that (I don't think we can easily re-use the code that draws on the map). Again, could be explored in a separate change.

Cernelius:
> It is the letter "x", but I suppose it could be changed to the Unicode multiplication "×" sign:
> https://www.compart.com/en/unicode/U+00D7

That's what I assumed, and I hate when people use the "x" to mean "per" even though I do it too when I'm just writing in a forum or something. I don't think that's acceptable in a finished product.
In the International System of Units, it is common to use a centred dot as the multiplication sign, maybe to avoid mixing up such sign with the "x" letter (which is often used in algebra).
https://www.compart.com/en/unicode/U+22C5
There is also a strong tendecy at using the asterisk as such, but I've never experienced this outside of computers (and I first met such usage with Lotus 1-2-3, and I just assumed that was because keyboards didn't have the proper sign, but I don't really know), so I doubt it is correct.
I'd still use nothing at all (juxtaposition) in between of the unit image and the stack number, but I think I would prefer the centred dot if anything.

(In my experience in Italy, only small kids use the "×" sign, thereafter you always use the centred dot, and I remember I asked myself why didn't they make us use the centred dot in the first place then.)

PR #10415, opened 2022-05-06
Fix image colorizing using map.properties.

asvitkine:
Fix image colorizing using map.properties.

This was not working because the image returned by the Toolkit was being changed directly.
Copying the image resolves the problem.

Also some clean up to avoid repeated attachment look ups.

Screenshot before on an unreleased map:
Screen Shot 2022-05-06 at 1 57 10 PM

Screenshot after:
Screen Shot 2022-05-06 at 2 02 47 PM


RoiEXLab (approved):
👍🏼

asvitkine:
Just to follow up, this is fixing something that got broken in 2.6, as 2.5 doesn't have the bug this is fixing.

PR #10419, opened 2022-05-07
Remove some unnecessary sleep invocations & clean up some code.

asvitkine:

This change removes some unnecessary sleep invocations that exist to work around excessive calls to remote random servers like PBEM by wrapping the logic in a PBEM-check block.

Additionally, cleans up comment formatting and makes sleeps calls with value 0 (which can happen per engine settings) to be no-ops. 

Also cleans up some redundant checks that are always true.


asvitkine:
By the way, the original change that introduced these sleeps was this one:
https://github.com/triplea-game/triplea/commit/227da1a18c9f045d66a6fc2e4439c8511479f5d9

No idea if they're even needed anymore.

asvitkine:
I fixed one of the existing codeclimate issues, but there's still another one codeclimate flagged that was pre-existing so I don't plan to fix it in this PR.

DanVanAtta (approved):

PR #10424, opened 2022-05-09
Fix a logic error that was introduced during a recent refactoring.

asvitkine:


asvitkine:
codeclimate issue is pre-existing.

PR #10425, opened 2022-05-09
Support <forEach> in the step list.

asvitkine:
Support <forEach> in the step list.

This introduces new XML syntax on maps to streamline the list of steps, using variables and a `<forEach>` tag.

Example use (assuming a `<variable>` named `"players"` in the `<variableList>`):

```xml
<sequence>
	<step name="gameInitDelegate" delegate="initDelegate" maxRunCount="1"/>
	<step name="setupRandomStartDelegate" delegate="randomStartDelegate" maxRunCount="1"/>
	<forEach variable="$players$">
		<step name="@players@FleetPlace" delegate="fleet_place" player="@players@" maxRunCount="1"/>
	</forEach>
	<forEach variable="$players$">
		<step name="@players@WealthPlace" delegate="wealth_place" player="@players@"/>
		<step name="@players@Purchase" delegate="purchase" player="@players@"/>
		<step name="@players@Place" delegate="place" player="@players@"/>
		<step name="@players@StructuresPlace" delegate="structures_place" player="@players@"/>
		<step name="@players@CombatMove" delegate="move" player="@players@">
			<stepProperty name="resetUnitStateAtStart" value="true"/>
		</step>
		<step name="@players@Battle" delegate="battle" player="@players@"/>
		<step name="@players@EndTurn" delegate="endTurn" player="@players@"/>
	</forEach>
	<step name="endRoundStep" delegate="endRound"/>
</sequence>

```

The above is from an actual map I'm developing where I've confirmed it works. It lets you avoid the repetition of steps for players that should have all the same steps and allows you to still have different sequences of steps (for example, above, there's an initial `FleetPlace` step for each player, before the regular turn sequence begins).

### DIFF
--- a/game-app/game-core/src/main/resources/games/strategy/engine/xml/game.dtd
+++ b/game-app/game-core/src/main/resources/games/strategy/engine/xml/game.dtd
@@ -103,7 +103,7 @@
 	>
 	<!-- if no player given, then delegate is initialized with a null player -->
 	<!-- max run count is the maximum number of times that a delegate can run, if specified must be > 0, if not specified its infinity-->
-	<!ELEMENT sequence (step+)>
+	<!ELEMENT sequence (step+ forEach*)>
 		<!ELEMENT step (stepProperty*)>
 		<!ATTLIST step
 			name ID #REQUIRED
@@ -117,6 +117,15 @@
 				name CDATA #REQUIRED
 				value CDATA #REQUIRED
 			>
+		<!ELEMENT forEach (step*)>
+		<!ATTLIST forEach variable CDATA #REQUIRED>
+			<!ATTLIST step
+					name ID #REQUIRED
+					delegate CDATA #REQUIRED
+					player CDATA #IMPLIED
+					maxRunCount CDATA #IMPLIED
+					display CDATA #IMPLIED
+					>
 	<!ELEMENT offset EMPTY>
 	<!ATTLIST offset
 		round CDATA #REQUIRED
--- a/lib/xml-reader/src/main/java/org/triplea/generic/xml/reader/annotations/XmlOffset.java
+++ b/lib/xml-reader/src/main/java/org/triplea/generic/xml/reader/annotations/XmlOffset.java
@@ -1,0 +1,24 @@
+package org.triplea.generic.xml.reader.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Add this annotation to java fields to indicate their value should be mapped to the offset of the
+ * element in the XML file. This can be useful when a parent element supports different types of
+ * child elements and you need to know their relative order.
+ *
+ * <p>Example:
+ *
+ * <pre>{@code
+ * class Tag {
+ *   @XmlOffset
+ *   private int xmlOffset;
+ * }
+ * }</pre>
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface XmlOffset {}
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameDataVariableParser.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameDataVariableParser.java
@@ -1,16 +1,24 @@
 package games.strategy.engine.data.gameparser;
 
+import com.google.common.base.Splitter;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import javax.annotation.Nullable;
+import lombok.Getter;
 import org.triplea.map.data.elements.VariableList;
 
 class GameDataVariableParser {
+  @Getter private Map<String, List<String>> variables;
 
   Map<String, List<String>> parseVariables(final VariableList root) {
-    return root == null ? Map.of() : parseVariableElement(root.getVariables());
+    variables = root == null ? Map.of() : parseVariableElement(root.getVariables());
+    return variables;
   }
 
   private Map<String, List<String>> parseVariableElement(
@@ -26,7 +34,7 @@ class GameDataVariableParser {
               .collect(Collectors.toList());
       variables.put(name, values);
     }
-    return variables;
+    return Collections.unmodifiableMap(variables);
   }
 
   private Stream<String> findNestedVariables(
@@ -35,5 +43,69 @@ class GameDataVariableParser {
       return Stream.of(value);
     }
     return variables.get(value).stream().flatMap(s -> findNestedVariables(s, variables));
+  }
+
+  public List<Map<String, String>> expandVariableCombinations(final String foreach)
+      throws GameParseException {
+    final List<Map<String, String>> combinations = new ArrayList<>();
+    final List<String> nestedForeach = Splitter.on("^").splitToList(foreach);
+    if (nestedForeach.isEmpty() || nestedForeach.size() > 2) {
+      throw new GameParseException(
+          "Invalid foreach expression, can only use variables, ':', and at most 1 '^': " + foreach);
+    }
+    final List<String> foreachVariables1 = Splitter.on(":").splitToList(nestedForeach.get(0));
+    final List<String> foreachVariables2 =
+        nestedForeach.size() == 2 ? Splitter.on(":").splitToList(nestedForeach.get(1)) : List.of();
+    GameParsingValidation.validateForeachVariables(foreachVariables1, variables, foreach);
+    GameParsingValidation.validateForeachVariables(foreachVariables2, variables, foreach);
+    final int length1 = variables.get(foreachVariables1.get(0)).size();
+    for (int i = 0; i < length1; i++) {
+      final Map<String, String> foreachMap1 =
+          createForeachVariablesMap(foreachVariables1, i, variables);
+      if (foreachVariables2.isEmpty()) {
+        combinations.add(foreachMap1);
+      } else {
+        final int length2 = variables.get(foreachVariables2.get(0)).size();
+        for (int j = 0; j < length2; j++) {
+          final Map<String, String> foreachMap2 =
+              createForeachVariablesMap(foreachVariables2, j, variables);
+          foreachMap2.putAll(foreachMap1);
+          combinations.add(foreachMap2);
+        }
+      }
+    }
+    return combinations;
+  }
+
+  private Map<String, String> createForeachVariablesMap(
+      final List<String> foreachVariables,
+      final int currentIndex,
+      final Map<String, List<String>> variables) {
+    final Map<String, String> foreachMap = new HashMap<>();
+    for (final String foreachVariable : foreachVariables) {
+      final List<String> foreachValue = variables.get(foreachVariable);
+      foreachMap.put("@" + foreachVariable.replace("$", "") + "@", foreachValue.get(currentIndex));
+    }
+    return foreachMap;
+  }
+
+  public String replaceForeachVariables(
+      final @Nullable String s, final Map<String, String> foreach) {
+    if (s == null) {
+      return null;
+    }
+    String result = s;
+    for (final Map.Entry<String, String> entry : foreach.entrySet()) {
+      result = result.replace(entry.getKey(), Optional.ofNullable(entry.getValue()).orElse(""));
+    }
+    return result;
+  }
+
+  public String replaceVariables(final String s) {
+    String result = s;
+    for (final Map.Entry<String, List<String>> entry : variables.entrySet()) {
+      result = result.replace(entry.getKey(), String.join(":", entry.getValue()));
+    }
+    return result;
   }
 }
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
@@ -850,9 +850,7 @@ public final class GameParser {
       if (foreach == null || foreach.isBlank()) {
         parseAttachment(current, Map.of());
       } else {
-        final List<Map<String, String>> combinations =
-            variableParser.expandVariableCombinations(foreach);
-        for (Map<String, String> foreachMap : combinations) {
+        for (Map<String, String> foreachMap : variableParser.expandVariableCombinations(foreach)) {
           parseAttachment(current, foreachMap);
         }
       }
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
@@ -39,8 +39,8 @@ import java.io.InputStream;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -133,7 +133,6 @@ public final class GameParser {
           try {
             return new GameParser(xmlFile.toString(), xmlGameElementMapper, engineVersion)
                 .parse(xmlFile, inputStream);
-
           } catch (final EngineVersionException e) {
             log.warn("Game engine not compatible with: " + xmlFile, e);
             return null;
@@ -147,7 +146,6 @@ public final class GameParser {
   @Nonnull
   private GameData parse(final Path xmlFile, final InputStream stream)
       throws XmlParsingException, GameParseException, EngineVersionException {
-
     final Game game = new XmlMapper(stream).mapXmlToObject(Game.class);
 
     // test minimum engine version first
@@ -182,7 +180,8 @@ public final class GameParser {
 
     parseDelegates(game.getGamePlay().getDelegates());
 
-    parseSteps(game.getGamePlay().getSequence().getSteps());
+    variableParser.parseVariables(game.getVariableList());
+    parseSteps(game.getGamePlay().getSequence());
 
     Optional.ofNullable(game.getGamePlay().getOffset()).ifPresent(this::parseOffset);
 
@@ -202,11 +201,7 @@ public final class GameParser {
       TechAdvance.createDefaultTechAdvances(data);
     }
 
-    if (game.getAttachmentList() != null) {
-      final Map<String, List<String>> variables =
-          variableParser.parseVariables(game.getVariableList());
-      parseAttachments(game.getAttachmentList(), variables);
-    }
+    parseAttachments(game.getAttachmentList());
 
     if (game.getInitialize() != null) {
 
@@ -244,8 +239,7 @@ public final class GameParser {
     data.setDiceSides(diceSides == null ? 6 : diceSides.getValue());
   }
 
-  private boolean isEngineCompatibleWithMap(final Triplea tripleA) throws EngineVersionException {
-
+  private boolean isEngineCompatibleWithMap(final Triplea tripleA) {
     return tripleA == null
         || tripleA.getMinimumVersion().isBlank()
         || engineVersion.isCompatibleWithMapMinimumEngineVersion(
@@ -335,14 +329,10 @@ public final class GameParser {
 
   private void parseTerritories(
       final List<org.triplea.map.data.elements.Map.Territory> territories) {
-    territories.forEach(
-        current ->
-            data.getMap()
-                .addTerritory(
-                    new Territory(
-                        current.getName(),
-                        Optional.ofNullable(current.getWater()).orElse(false),
-                        data)));
+    for (final org.triplea.map.data.elements.Map.Territory current : territories) {
+      boolean isWater = Optional.ofNullable(current.getWater()).orElse(false);
+      data.getMap().addTerritory(new Territory(current.getName(), isWater, data));
+    }
   }
 
   private void parseConnections(
@@ -564,39 +554,80 @@ public final class GameParser {
     }
   }
 
-  private void parseSteps(final List<GamePlay.Sequence.Step> stepList) throws GameParseException {
-    for (final GamePlay.Sequence.Step current : stepList) {
-      final IDelegate delegate = getDelegate(current.getDelegate());
-      final GamePlayer player = getPlayerIdOptional(current.getPlayer()).orElse(null);
-      if (player == null && current.getPlayer() != null && !current.getPlayer().isBlank()) {
-        throw new GameParseException(
-            "The step "
-                + current.getName()
-                + " wants a player with the name of '"
-                + current.getPlayer()
-                + "' but that player can not be found. "
-                + "Make sure the player's name is spelled correctly.");
+  private void parseSteps(final GamePlay.Sequence sequence) throws GameParseException {
+    final Iterator<GamePlay.Sequence.Step> stepIter = sequence.getSteps().iterator();
+    @Nullable GamePlay.Sequence.Step nextStep = stepIter.hasNext() ? stepIter.next() : null;
+
+    for (final GamePlay.Sequence.ForEach forEach : sequence.getForEach()) {
+      if (forEach.getSteps().isEmpty()) {
+        continue;
       }
-      final String name = current.getName();
-      String displayName = null;
-      final Properties stepProperties = parseStepProperties(current.getStepProperties());
-      if (current.getDisplay() == null || !current.getDisplay().isBlank()) {
-        displayName = current.getDisplay();
+
+      // We're generating a list of steps in the correct order. When a <forEach> is encountered,
+      // add any non-forEach steps that come before it, before adding the steps from the <forEach>.
+      final int forEachXmlOffset = forEach.getSteps().get(0).getXmlOffset();
+      while (nextStep != null && nextStep.getXmlOffset() < forEachXmlOffset) {
+        data.getSequence().addStep(parseStep(nextStep, Map.of()));
+        nextStep = stepIter.hasNext() ? stepIter.next() : null;
       }
-      final GameStep step = new GameStep(name, displayName, player, delegate, data, stepProperties);
-      if (current.getMaxRunCount() != null && current.getMaxRunCount() > 0) {
-        step.setMaxRunCount(current.getMaxRunCount());
+
+      // For each combination of variables, repeat all the steps in this <forEach>.
+      // Note: The combinations are ordered according to the variable's order.
+      final List<Map<String, String>> combinations =
+          variableParser.expandVariableCombinations(forEach.getVariable());
+      for (final Map<String, String> forEachMapping : combinations) {
+        for (final GamePlay.Sequence.Step xmlStep : forEach.getSteps()) {
+          data.getSequence().addStep(parseStep(xmlStep, forEachMapping));
+        }
       }
-      data.getSequence().addStep(step);
+    }
+
+    // Add any remaining non-forEach steps. We iterate using `nextStep` since we've already
+    // removed it from the iterator.
+    while (nextStep != null) {
+      data.getSequence().addStep(parseStep(nextStep, Map.of()));
+      nextStep = stepIter.hasNext() ? stepIter.next() : null;
     }
   }
 
-  private static Properties parseStepProperties(
-      final List<GamePlay.Sequence.Step.StepProperty> properties) {
+  private GameStep parseStep(
+      final GamePlay.Sequence.Step xmlStep, final Map<String, String> forEachMapping)
+      throws GameParseException {
+    final IDelegate delegate = getDelegate(xmlStep.getDelegate());
+    String playerName = variableParser.replaceForeachVariables(xmlStep.getPlayer(), forEachMapping);
+    final GamePlayer player = getPlayerIdOptional(playerName).orElse(null);
+    if (player == null && xmlStep.getPlayer() != null && !xmlStep.getPlayer().isBlank()) {
+      throw new GameParseException(
+          "The step "
+              + xmlStep.getName()
+              + " wants a player with the name of '"
+              + xmlStep.getPlayer()
+              + "' but that player can not be found. "
+              + "Make sure the player's name is spelled correctly.");
+    }
+    String displayName = null;
+    if (xmlStep.getDisplay() == null || !xmlStep.getDisplay().isBlank()) {
+      displayName = variableParser.replaceForeachVariables(xmlStep.getDisplay(), forEachMapping);
+    }
+    final String name = variableParser.replaceForeachVariables(xmlStep.getName(), forEachMapping);
+    final Properties stepProperties =
+        parseStepProperties(xmlStep.getStepProperties(), forEachMapping);
+    final GameStep step = new GameStep(name, displayName, player, delegate, data, stepProperties);
+    if (xmlStep.getMaxRunCount() != null && xmlStep.getMaxRunCount() > 0) {
+      step.setMaxRunCount(xmlStep.getMaxRunCount());
+    }
+    return step;
+  }
+
+  private Properties parseStepProperties(
+      final List<GamePlay.Sequence.Step.StepProperty> properties,
+      final Map<String, String> forEachMapping) {
     final Properties stepProperties = new Properties();
-    properties.forEach(
-        stepProperty ->
-            stepProperties.setProperty(stepProperty.getName(), stepProperty.getValue()));
+    for (final GamePlay.Sequence.Step.StepProperty p : properties) {
+      stepProperties.setProperty(
+          variableParser.replaceForeachVariables(p.getName(), forEachMapping),
+          variableParser.replaceForeachVariables(p.getValue(), forEachMapping));
+    }
     return stepProperties;
   }
 
@@ -810,68 +841,31 @@ public final class GameParser {
     }
   }
 
-  private void parseAttachments(
-      final AttachmentList root, final Map<String, List<String>> variables)
-      throws GameParseException {
+  private void parseAttachments(final AttachmentList root) throws GameParseException {
+    if (root == null) {
+      return;
+    }
     for (final AttachmentList.Attachment current : root.getAttachments()) {
       final String foreach = current.getForeach();
       if (foreach == null || foreach.isBlank()) {
-        parseAttachment(current, variables, Map.of());
+        parseAttachment(current, Map.of());
       } else {
-        final List<String> nestedForeach = Splitter.on("^").splitToList(foreach);
-        if (nestedForeach.isEmpty() || nestedForeach.size() > 2) {
-          throw new GameParseException(
-              "Invalid foreach expression, can only use variables, ':', and at most 1 '^': "
-                  + foreach);
-        }
-        final List<String> foreachVariables1 = Splitter.on(":").splitToList(nestedForeach.get(0));
-        final List<String> foreachVariables2 =
-            nestedForeach.size() == 2
-                ? Splitter.on(":").splitToList(nestedForeach.get(1))
-                : List.of();
-        GameParsingValidation.validateForeachVariables(foreachVariables1, variables, foreach);
-        GameParsingValidation.validateForeachVariables(foreachVariables2, variables, foreach);
-        final int length1 = variables.get(foreachVariables1.get(0)).size();
-        for (int i = 0; i < length1; i++) {
-          final Map<String, String> foreachMap1 =
-              createForeachVariablesMap(foreachVariables1, i, variables);
-          if (foreachVariables2.isEmpty()) {
-            parseAttachment(current, variables, foreachMap1);
-          } else {
-            final int length2 = variables.get(foreachVariables2.get(0)).size();
-            for (int j = 0; j < length2; j++) {
-              final Map<String, String> foreachMap2 =
-                  createForeachVariablesMap(foreachVariables2, j, variables);
-              foreachMap2.putAll(foreachMap1);
-              parseAttachment(current, variables, foreachMap2);
-            }
-          }
+        final List<Map<String, String>> combinations =
+            variableParser.expandVariableCombinations(foreach);
+        for (Map<String, String> foreachMap : combinations) {
+          parseAttachment(current, foreachMap);
         }
       }
     }
   }
 
-  private static Map<String, String> createForeachVariablesMap(
-      final List<String> foreachVariables,
-      final int currentIndex,
-      final Map<String, List<String>> variables) {
-    final Map<String, String> foreachMap = new HashMap<>();
-    for (final String foreachVariable : foreachVariables) {
-      final List<String> foreachValue = variables.get(foreachVariable);
-      foreachMap.put("@" + foreachVariable.replace("$", "") + "@", foreachValue.get(currentIndex));
-    }
-    return foreachMap;
-  }
-
   private void parseAttachment(
-      final AttachmentList.Attachment current,
-      final Map<String, List<String>> variables,
-      final Map<String, String> foreach)
+      final AttachmentList.Attachment current, final Map<String, String> foreach)
       throws GameParseException {
     final String className = current.getJavaClass();
     final Attachable attachable =
         findAttachment(current, Optional.ofNullable(current.getType()).orElse("unitType"), foreach);
-    final String name = replaceForeachVariables(current.getName(), foreach);
+    final String name = variableParser.replaceForeachVariables(current.getName(), foreach);
     final IAttachment attachment =
         xmlGameElementMapper
             .newAttachment(className, name, attachable, data)
@@ -883,7 +877,7 @@ public final class GameParser {
     attachable.addAttachment(name.replaceAll("ttatchment", "ttachment"), attachment);
 
     final List<Tuple<String, String>> attachmentOptionValues =
-        setOptions(attachment, current.getOptions(), foreach, variables);
+        setOptions(attachment, current.getOptions(), foreach);
     // keep a list of attachment references in the order they were added
     data.addToAttachmentOrderAndValues(Tuple.of(attachment, attachmentOptionValues));
   }
@@ -891,7 +885,7 @@ public final class GameParser {
   private Attachable findAttachment(
       final AttachmentList.Attachment element, final String type, final Map<String, String> foreach)
       throws GameParseException {
-    final String attachTo = replaceForeachVariables(element.getAttachTo(), foreach);
+    final String attachTo = variableParser.replaceForeachVariables(element.getAttachTo(), foreach);
     switch (type) {
       case "unitType":
         return getUnitType(attachTo);
@@ -915,8 +909,7 @@ public final class GameParser {
   private List<Tuple<String, String>> setOptions(
       final IAttachment attachment,
       final List<AttachmentList.Attachment.Option> options,
-      final Map<String, String> foreach,
-      final Map<String, List<String>> variables)
+      final Map<String, String> foreach)
       throws GameParseException {
     final List<Tuple<String, String>> results = new ArrayList<>();
     for (final AttachmentList.Attachment.Option option : options) {
@@ -939,8 +932,9 @@ public final class GameParser {
       if (containsEmptyForeachVariable(countAndValue, foreach)) {
         continue; // Skip adding option if contains empty foreach variable
       }
-      final String valueWithForeach = replaceForeachVariables(countAndValue, foreach);
-      final String interpolatedValue = replaceVariables(valueWithForeach, variables);
+      final String valueWithForeach =
+          variableParser.replaceForeachVariables(countAndValue, foreach);
+      final String interpolatedValue = variableParser.replaceVariables(valueWithForeach);
       final String finalValue = LegacyPropertyMapper.mapLegacyOptionValue(name, interpolatedValue);
       try {
         attachment
@@ -966,14 +960,6 @@ public final class GameParser {
     return results;
   }
 
-  private String replaceForeachVariables(final String s, final Map<String, String> foreach) {
-    String result = s;
-    for (final Entry<String, String> entry : foreach.entrySet()) {
-      result = result.replace(entry.getKey(), Optional.ofNullable(entry.getValue()).orElse(""));
-    }
-    return result;
-  }
-
   private boolean containsEmptyForeachVariable(final String s, final Map<String, String> foreach) {
     for (final Entry<String, String> entry : foreach.entrySet()) {
       if ((entry.getValue() == null || entry.getValue().isEmpty()) && s.contains(entry.getKey())) {
@@ -981,14 +967,6 @@ public final class GameParser {
       }
     }
     return false;
-  }
-
-  private String replaceVariables(final String s, final Map<String, List<String>> variables) {
-    String result = s;
-    for (final Entry<String, List<String>> entry : variables.entrySet()) {
-      result = result.replace(entry.getKey(), String.join(":", entry.getValue()));
-    }
-    return result;
   }
 
   @VisibleForTesting
@@ -1028,7 +1006,7 @@ public final class GameParser {
       } else {
         owner = getPlayerIdOptional(current.getOwner()).orElse(null);
       }
-      final int quantity = Optional.ofNullable(current.getQuantity()).orElse(0);
+      final int quantity = current.getQuantity();
       territory.getUnitCollection().addAll(type.create(quantity, owner, false, hits, unitDamage));
     }
   }
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/screen/TileManager.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/screen/TileManager.java
@@ -48,6 +48,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import org.triplea.util.Tuple;
 
 /** Orchestrates the rendering of all map tiles. */
@@ -352,13 +353,7 @@ public class TileManager {
               uiContext);
       drawing.add(drawable);
       allUnitDrawables.add(drawable);
-      for (final Tile tile :
-          getTiles(
-              new Rectangle(
-                  lastPlace.x,
-                  lastPlace.y,
-                  uiContext.getUnitImageFactory().getUnitImageWidth(),
-                  uiContext.getUnitImageFactory().getUnitImageHeight()))) {
+      for (final Tile tile : getTiles(getUnitRectangleAt(lastPlace))) {
         tile.addDrawable(drawable);
         drawnOn.add(tile);
       }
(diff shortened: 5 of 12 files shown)


DanVanAtta:
Rather than a for-each, WDYT of using a vector-like declaration to do the same?

Eg:

```
<sequence>
	<step name="gameInitDelegate" delegate="initDelegate" maxRunCount="1"/>
	<step name="setupRandomStartDelegate" delegate="randomStartDelegate" maxRunCount="1"/>

        <playerSteps>
               <players>
                 <player>Player1</player>
                 <player>Player2</player>
               </players>
		<step name="WealthPlace" delegate="wealth_place"/>
		:
		:
        </playerSteps>
</sequence>
```
That would be much more canonical-XML. (side-note: Our map XMLs are very non-canonical..)

game-app/game-core/src/main/resources/games/strategy/engine/xml/game.dtd
```diff
@@ -103,7 +103,7 @@
 	>
 	<!-- if no player given, then delegate is initialized with a null player -->
 	<!-- max run count is the maximum number of times that a delegate can run, if specified must be > 0, if not specified its infinity-->
-	<!ELEMENT sequence (step+)>
+	<!ELEMENT sequence (step+ forEach*)>
```
DanVanAtta, line 106: I'm surprised we still have the `game.dtd` file around even, is this update needed? Does the map have an unnecessary DTD definition on it?

lib/xml-reader/src/main/java/org/triplea/generic/xml/reader/annotations/XmlOffset.java
```diff
@@ -0,0 +1,24 @@
...
+ * Add this annotation to java fields to indicate their value should be mapped to the offset of the
+ * element in the XML file. This can be useful when a parent element supports different types of
+ * child elements and you need to know their relative order.
+ *
+ * <p>Example:
+ *
+ * <pre>{@code
+ * class Tag {
+ *   @XmlOffset
+ *   private int xmlOffset;
```
DanVanAtta, line 18: I don't quite understand the need for the 'xmlOffset', what is meant by an offset?
I wonder if we need this, so I'm curious about what it is used for.

DanVanAtta:
@asvitkine , if the full list of permutations were to be listed out, how long would that list be? Perhaps it would be better for the map that needs this functionality, to just enumerate the steps explicitly (even if it is a long list).

The XML could use an overhaul, it's unlikely that we will update many more maps to have this feature, and any XML parsing feature generally needs to be maintained ad-infinitum (a huge cost over time). I'm leaning that we avoid XML updates in favor of developing a new XML spec that supports 3.0 and is more generally more powerful and generic

lib/xml-reader/src/main/java/org/triplea/generic/xml/reader/annotations/XmlOffset.java
asvitkine: The way our xml parsing code works is it puts all child elements of a given tag into a collection.

So if you have:

```xml
<foo>
<bar/>
<bar/>
<baz/>
<bar/>
<baz/>
</foo>
```

You just get two collections, one of `bar` and one of `baz` and you can't tell their relative order. By using this, each of those elements keeps track of its offset in the XML, so we can reason about their relative order, which is needed here since the list of steps can have a mixture of `step` tags and `forEach` tags and their order is relevant.

asvitkine:
> @asvitkine , if the full list of permutations were to be listed out, how long would that list be? Perhaps it would be better for the map that needs this functionality, to just enumerate the steps explicitly (even if it is a long list).
> 
> The XML could use an overhaul, it's unlikely that we will update many more maps to have this feature, and any XML parsing feature generally needs to be maintained ad-infinitum (a huge cost over time). I'm leaning that we avoid XML updates in favor of developing a new XML spec that supports 3.0 and is more generally more powerful and generic

I agree that we can just write them out right now. But obviously it's messy and error prone (very easy to forget to set the right player in both places or spell everything right).

However, the main motivation for me working on this is a follow up change that I want to make which is to support randomizing the order of players (which combined with random start delegate achieves the behavior I want).

That can be achieved with a very simple syntax of:
```xml
<variable shuffle="true">
```

Which would also be very trivial to implement.

I agree with the goal of the XML format rewrite but I think it's also reasonable for some parts of that being incremental in improving the current format, so that it reduces the delta needed for the rewrite. The easier such a rewrite us the more likely it actually happens.

As to the syntax, I actually think there's very good reasons to piggy back on variables. Let me explain. First, this is something that's available for attachments and I think any new map making use of triggers will already be defining the player list as a variable, because it's needed if you want a trigger to apply to each player (unless you want to duplicate identical trigger code). As a case in point, the map I'm working on makes very extensive use of the players list variable. So given this already exists, it makes perfect sense to re-use it. The only thing I can see improving is making some automatic variables - like the list of players and territories to avoid needy to manually specify them - then you could just use it directly.

game-app/game-core/src/main/resources/games/strategy/engine/xml/game.dtd
asvitkine: I was surprised too - but I noticed when I was adding a test for this with a map XML that IntelliJ was actually validating against that and giving me red underlines, so I updated it since it wasn't too hard. The game itself doesn't seem to use the DTD for validation (OTOH I think if the editor I was using for editing my map did use it, it could have been helpful).

asvitkine:
> Rather than a for-each, WDYT of using a vector-like declaration to do the same?
> 
> Eg:
> 
> ```
> <sequence>
> 	<step name="gameInitDelegate" delegate="initDelegate" maxRunCount="1"/>
> 	<step name="setupRandomStartDelegate" delegate="randomStartDelegate" maxRunCount="1"/>
> 
>         <playerSteps>
>                <players>
>                  <player>Player1</player>
>                  <player>Player2</player>
>                </players>
> 		<step name="WealthPlace" delegate="wealth_place"/>
> 		:
> 		:
>         </playerSteps>
> </sequence>
> ```
> 
> That would be much more canonical-XML. (side-note: Our map XMLs are very non-canonical..)

Your structure requires repeating the player list at each sequence of steps, which is more verbose and thus imho worse than my proposed syntax here.

Notice that in my example, I have two `<forEach>`. Using your syntax, we'd have two `<playerSteps>` tags, repeating `<players>` lists in each. By the way, even the  `minimap` example that I used for a test is making use of the multiple `<forEach>` features since it has some bid steps ahead of the main sequence.

The other downside I see is your syntax hides what the actual step name would be - i.e. it becomes non-obvious that `"WealthPlace"` becomes `"GermanWealthPlace"`. You may wonder, so what? Well, it's pretty important for triggers, because the `when` condition has to spell out the step name exactly and for triggers that will go for all players, it would be frequently something like: `<option name="when" value="before:@players@WealthPlace"/>` - which becomes very convenient with my proposed syntax since it's consistent.

lib/xml-reader/src/main/java/org/triplea/generic/xml/reader/annotations/XmlOffset.java
DanVanAtta: If order is important, it can become an attribute of a tag, eg:
```
<foo>
 <bar order="1"/>
 <baz order="2"/>
</foo>
```
In theory the order of XML tags generally should not matter (but often it does, and some XML specs rely on parsers to provide that guarantee).

DanVanAtta:
Re: XML rewrite

I think it will be pretty large. Large enough that we would specify territory types, eg: "sea", "land" "space", "deep sea", etc, make that generic. Overall I'm envisioning a revisit of most basic assumptions and to redo nearly all of it.

To migrate, we will likely keep the old parser and then shim in default values (very much like how the DTD used to do), and/or create a way to output map XML and then re-encode the maps.

Re: Nesting and repetition

My biggest criticism of the current XML is that it does not use nested lists very well, often uses comma delimited and semi-colon values when instead nested elements would work really well and be canonical. I think 'variables' were a bit of a work around to overcome that, so I don't think it was necessarily a perfect solution.

I'm reminded here of War of Relics though as that XML was largely generated with Excell Macros. I'll agree that having a looping structure is convenient. Though, is there a way we can make that better?

Perhaps we can do something like:

```
<playerList id="allies">
  <player name="abc" order="1" />
<playerList>

<trigger>
   .... playerListId="allies"

</trigger>
```

I feel there should be a more canonical 'xml' way to do this without resorting to a near programming like templating language.

asvitkine:
> Re: XML rewrite
> 
> I think it will be pretty large. Large enough that we would specify territory types, eg: "sea", "land" "space", "deep sea", etc, make that generic. Overall I'm envisioning a revisit of most basic assumptions and to redo nearly all of it.

I still think that can be done incrementally. Having one big rewrite is a recipe for either it never being done or having some parts of it rushed and then regretting some decision needing yet another rewrite after. Incremental changes allow for vetting those in isolation and getting them right without a long drawn out project where the light at the end of the tunnel may be far away.

> 
> To migrate, we will likely keep the old parser and then shim in default values (very much like how the DTD used to do), and/or create a way to output map XML and then re-encode the maps.

Having two different parsers sounds like a nightmare. If we can have a converter and re-encode all the maps, that would be ideal. But a converter does require we support everything we currently do in the new format.

> Re: Nesting and repetition
> 
> My biggest criticism of the current XML is that it does not use nested lists very well, often uses comma delimited and semi-colon values when instead nested elements would work really well and be canonical. I think 'variables' were a bit of a work around to overcome that, so I don't think it was necessarily a perfect solution.

Agree that this can be improved. I also don't see why it can't be incremental. However, the main downside with nesting is you're repeating yourself if multiple things want to refer to the same list. So then you're re-inventing variables or something like them (e.g. a way to reference another element by its id).

> 
> I'm reminded here of War of Relics though as that XML was largely generated with Excell Macros. I'll agree that having a looping structure is convenient. Though, is there a way we can make that better?
> 
> Perhaps we can do something like:
> 
> ```
> <playerList id="allies">
>   <player name="abc" order="1" />
> <playerList>
> 
> <trigger>
>    .... playerListId="allies"
> 
> </trigger>
> ```
> 
> I feel there should be a more canonical 'xml' way to do this without resorting to a near programming like templating language.

One challenge is that you still want something like variables, because the player itself isn't just used for attachedTo, but can be used for reference of other triggers, or conditions. Here's an example I have:

```xml
		<attachment name="enable_fort_production_@players@" attachTo="@players@" foreach="$players$"  javaClass="TriggerAttachment" type="player">
			<option name="conditions" value="has_wealth_3_@players@"/>
			....
```

So here, we're using the player name in the name, in the attachTo and also in reference of a condition, which itself is foreach'd for different players. But it can also use conditions that aren't foreached (e.g. checking the round number). I don't really see a way of supporting it other than with a variable or templating language. I could see an improvement being a similar outer <forEach> block that could group such things together, which could make expressions be more explicit and not specify their formatting strings, but this comes with its own complexity - how do you reference things that are global vs. ones that are in the current block? Perhaps by avoiding name collisions.

Anyway, triggers and variables are very complex so you can see my scepticism about a big rewrite that does everything at once - I'd hate for improvements to syntax for specifying territories and players and steps to be blocked on figuring out the optimal way to do attachments, for example.

To come back to this proposal, I think my point is that variables are here to stay because I think it's extremely hard or impossible to come up with something else that supports the same feature set (modulo getting rid of them and forcing people to write XML generator programs that do the variable thing outside the XML and producing giant XML blobs - which doesn't seem ideal).

I could image a more elegant solution than what's now - but that would a much bigger rewrite than all of the XML - instead, reworking entirely the game architecture to not use the "attachment" model - but that's even more ambitious - basically a rewrite of the engine - which I don't think we should block improvements on.

DanVanAtta:
> Having two different parsers sounds like a nightmare. If we can have a converter and re-encode all the maps, that would be ideal. But a converter does require we support everything we currently do in the new format.


This is worthy of its own discussion. The thinking is that the game objects that the XML will eventually translate into would be the place where it is unified. So there would be an adapter layer to convert old XML into the new objects, and a more native XML format for more fine grained control and a more sane DTD.

Overall, TripleA mostly has to support every XML that was ever created, so introducing a new XML syntax is something we have to support forever. It is a big deal..



Re: Variables

Rather than variables, they probably should be named blocks or "tagged" (ie: an ID) blocks that can be referenced in other sections of XML. Rather than 'macros' it is more references.

------------


w.r.t this update:

- I am concerned by introduced 'foreach' we do have to forever maintain it as we do all other XMLs. This is a big cost.
- I'd prefer a more canonical approach that avoids variables, that is less of a templating language (eg: loops & variables, if/else blocks). Have you considered generating the XML for the new map with a templating language like Jinja2 or Mustache?

asvitkine:
> > Having two different parsers sounds like a nightmare. If we can have a converter and re-encode all the maps, that would be ideal. But a converter does require we support everything we currently do in the new format.
> 
> This is worthy of its own discussion. The thinking is that the game objects that the XML will eventually translate into would be the place where it is unified. So there would be an adapter layer to convert old XML into the new objects, and a more native XML format for more fine grained control and a more sane DTD.

I think one challenge you'll run into is the old XML gives pretty low-level access to the engine internals (i.e. the attachment model is 1-1 with what's in the XML), so either you'd need to abstract it away (making the new format having less functionality than the low-level old format), or re-architect, in which case now the full game logic needs to have parallel paths. But I agree that this is a separate discussion.

> 
> Overall, TripleA mostly has to support every XML that was ever created, so introducing a new XML syntax is something we have to support forever. It is a big deal..

I'm curious if that's really true. Given generally all the maps are hosted on github, couldn't we coordinate update to maps to coincide with engine changes? I mean, there's a separate practical question of whether that's feasible - but for example for the rename of Attatchable -> Attachable, couldn't we just fix all the existing XMLs?

Or is the stance that we want to support even maps that live outside of the core repos? (My intuition is these aren't really played, but curious to know if that intuition isn't correct.)

> 
> Re: Variables
> 
> Rather than variables, they probably should be named blocks or "tagged" (ie: an ID) blocks that can be referenced in other sections of XML. Rather than 'macros' it is more references.

I brought up an example where I don't think that really work, see my quoted `enable_fort_production_@players@` XML snippet above. Basically, the variable often needs to be used in a bunch of places - not just one property like `attachTo`. So I really don't see how named blocks can help here without rewriting the whole attachment / trigger model. 

> 
> w.r.t this update:
> 
> * I am concerned by introduced 'foreach' we do have to forever maintain it as we do all other XMLs. This is a big cost.

I agree that maintainability is a valid concern for any new XML syntax. But given how "similar" this syntax is to one of your proposed alternatives, wouldn't it be trivial to migrate if we really decide on that? It's actually easier to migrate than migrating existing long lists of player steps, no? And this could be an incremental way to get maps to a more canonical form to make them easier to convert.

(Separately, I still think your example syntax is inferior to this for the reasons I mentioned already, so it's actually debatable that a migration is needed.)

Separately, on the issue of maintainability, this change actually improves the existing code in a few ways and refactors and adds tests to existing methods as well as end-to-end tests, so it's actually better maintained than existing code. But obviously each new feature does increase the scope.

> * I'd prefer a more canonical approach that avoids variables, that is less of a templating language (eg: loops & variables, if/else blocks). Have you considered generating the XML for the new map with a templating language like Jinja2 or Mustache?

I mentioned earlier that one goal I have, besides improving the syntax, is to easily support randomized player order - which can be done trivially after this. A separate XML generator approach doesn't help with that.

Even if we ignore that, having to write a program that generates the XML doesn't seem ideal - besides extra complexity (now if someone wants to update the map, there's an extra dependency - and who knows if that dependencies might have upgraded to a new version with incompatible syntax, etc, etc), but also we're shipping bigger blobs of XML to game which have to be downloaded, loaded from disk, parsed, etc.

DanVanAtta:
> I'm curious if that's really true. Given generally all the maps are hosted on github, couldn't we coordinate update to maps to coincide with engine changes? I mean, there's a separate practical question of whether that's feasible - but for example for the rename of Attatchable -> Attachable, couldn't we just fix all the existing XMLs?

There are a number of XMLs that we do not host. For example, RogerCooper keeps a number and was displeased when a legacy map format did not extract correctly and corrupted his downloads folder. We also see a number of error reports coming in for maps that are pre 1.8 (notably the same 'Attatchable' error).

We had dropped support for old maps, but that was done on what was supposed to be a temporary basis and hence 1.9 until 2.5 did not support the 1.8 and prior maps. We recently restored that, so at this point pretty much any map should work.

The main point being that often old maps come up, we don't know where they all are, and even if we update all that we host - it's not all of them.


> Even if we ignore that, having to write a program that generates the XML doesn't seem ideal - besides extra complexity (now if someone wants to update the map, there's an extra dependency - and who knows if that dependencies might have upgraded to a new version with incompatible syntax, etc, etc), but also we're shipping bigger blobs of XML to game which have to be downloaded, loaded from disk, parsed, etc.

Agree not ideal. Though, it's a cost for one map rather than a change to th engine. The latter is much more significant of course. If the one map requires an extra hour or three of work, that is less than the ongoing maintenance cost for the engine update (which as mentioned above, we pretty much have to support forever or risk breaking maps that take advantage of this syntax).


---------


Re: canonical XML

Another idea to make the XML more canonical:

```
<playerGroup id="allies">
   <player name=abc order=123|random />
</playerGroup>


<sequence>
   <step>...</step>
   
   <subSequence forPlayerGroup="allies">
   
       <step>...</step>
       
   </subSequence>
</sequence>


<attachments>
    <playerGroupTriggers forPlayerGroup="allies">
        <attachment .... />
    </playerGroupTrigger>
</attachments>
```

So, the above is more canonical and avoids any kind of looping or control structures embedded in the XML (which is a key difference between say XML and templating). If we migrated to JSON or YAML, it could be very odd to see a 'for' loop in either structure, the XML is akin to either of those data interchange formats.

Mainly I don't want us to move more towards this pattern of a template vs canonical XML. What's to stop from having a 'while' loop or start building 'if/else' statements?

I'm not sure if we are at loggerheads here, perhaps. My opinion is that unless we have at least 3 or 4 maps needing an XML feature or syntax, that it's not necessarily worth the long term engine maintenance cost. For example, if we build this into the engine, there are more tests to update with sample XML snippets to verify parsing, we should make sure that errors (like typo'd variable names) give reasonable error messages, we need sample XMLs; and we'll need to make sure to support this going forward. For example, if we do create a new XML spec that is a super-set of the existing, then that new spec is going to have to keep supporting foreach loops (which is complicating that task).

My opinion here is to keep any cleanups/refactoring that make sense, but pay the maintenance & inconvenience cost when developing the map vs ongoing maintenance of the engine - unless we have many maps that would want to adopt this. If there are many maps, I'd quite strongly prefer we discontinue the pattern of the XML spec becoming more a templating language than XML.

I'd recommend to further think about this for a few days and stay open to alternatives. I'm not a hard no on this update, but I'm not in love with almost anything that changes the XML at this juncture.

asvitkine:
> There are a number of XMLs that we do not host. For example, RogerCooper keeps a number and was displeased when a legacy map format did not extract correctly and corrupted his downloads folder.

Hmm, I'm curious as to what the rationale is. I still think a policy of "we're committed to backward compatibility and/or updating the maps" for any maps in the database to be compatible with future versions, but at the same time say "maps that aren't in the database are responsibility of their authors to keep compatible with new versions". This would be a sensible way to allow evolving the data format imho.

> We also see a number of error reports coming in for maps that are pre 1.8 (notably the same 'Attatchable' error).

I don't think we can conclude things from that given that maps in the DB haven't all been migrated away from this. See:
https://github.com/search?q=org%3Atriplea-maps+attatchTo&type=code

> Agree not ideal. Though, it's a cost for one map rather than a change to th engine. The latter is much more significant of course. If the one map requires an extra hour or three of work, that is less than the ongoing maintenance cost for the engine update (which as mentioned above, we pretty much have to support forever or risk breaking maps that take advantage of this syntax).

I guess if we don't have a policy of making updates to maps on GitHub when needed, then it wouldn't be a maintenance cost to us. But I think being able to update maps on GitHub to allow us to clean up the engine would be a good state to be in (and we can develop some scripts that help with that - i.e. check out all the repositories, run a program on all the XMLs etc). In such a world, maps whose XML is generated by some other systems would be problematic.

> 
> Re: canonical XML
> 
> Another idea to make the XML more canonical:
> 
> ```
> <playerGroup id="allies">
>    <player name=abc order=123|random />
> </playerGroup>
> 
> 
> <sequence>
>    <step>...</step>
>    
>    <subSequence forPlayerGroup="allies">
>    
>        <step>...</step>
>        
>    </subSequence>
> </sequence>
> 
> 
> <attachments>
>     <playerGroupTriggers forPlayerGroup="allies">
>         <attachment .... />
>     </playerGroupTrigger>
> </attachments>
> ```

I think something like that could potentially work, with some caveats.
First, I agree that this would address similar requirements to what this current PR does.
For attachments, I agree that this may be a good formulation, BUT:
  1. It shouldn't be limited to triggers, but also conditions.
  2. It could set some things automatically on the `<attachment>` like `attachTo`
  3. I think we would still need to expose the "player" variable available to the attachment bodies. This is because too many things may still need to reference the player - for example the "when" condition (which sometimes you want to do for that player's step, and sometimes you want it attached to a global step or something else - similar for specifying ownership of territories, etc). But I think this is solvable - we could just expose a $_currentPlayer$ variable inside these.
  4. We still need to disambiguate when these reference conditions - whether those conditions are inside the block (and thus duplicated for each player and you want to reference that player's version - or whether it's referencing one outside). A simple solution could be to disallow name collisions OR to have a scoped search - first try to find the player one and then fall back to the outside one if it exists.
  5. The part related to triggers would be a bigger rework than this change. Are we comfortable writing the building blocks for it and supporting it in the steps list without doing the attachment work yet?
  6. If we're talking reworking the XML format, I'd actually change the `<attachment>` syntax too. Instead of:
```xml
		<attachment name="Foo" javaClass="TriggerAttachment" type="player">
```
I would change it to:
```xml
  <triggerAttachment name="Foo">
```

But of course, that's a bigger change. Some other types of attachments, I'd actually argue would be better to be directly part of the entity they're attaching to - i.e. a UnitAttachment should just be directly part of the block that defines that unit type, and not an attachment at all, since it's always 1-1. Same for territories, for example.

But that's a bigger change...
 
> So, the above is more canonical and avoids any kind of looping or control structures embedded in the XML (which is a key difference between say XML and templating).  If we migrated to JSON or YAML, it could be very odd to see a 'for' loop in either structure, the XML is akin to either of those data interchange formats.

Not sure why those other formats are any different than XML - is it because there's no end tag to a block? Python doesn't have anything to end a block and it's defined by indentation and is able to represent loops and if statement just fine. So I don't think the format type makes a difference here.

> Mainly I don't want us to move more towards this pattern of a template vs canonical XML. What's to stop from having a 'while' loop or start building 'if/else' statements?

Well, to be honest, I don't actually see a problem with if statements, although it's not something I need right now. Let's say you have a turn sequence that's the same for all players, except one player doesn't get a certain step (for example, some WW2 variants have special case for Chinese). It seems more concise to have an if statement than have to spell everything out the "low-level" way in that case.

Ultimately, the map format is meant to allow the mapmaker to set up the game how they want and within those constraints, the more readable, understandable and concise the format is, the better - since this would reduce errors and time wasted for map makers. I don't buy that we should prioritize "ideological" data format vs. what solves real problems in the best way with regard to the above properties I mention.

OTOH, right now using variables for players list is a "necessary evil" that does go against the above properties since you have to repeat yourself. So in that world, something more like what you suggest might be an improvement. But we should agree on the end goals and semantics and what's OK to do incrementally - for example, does just adding the support for this for game steps to support randomization OK and leave triggers for later? Or do you still have the stance of "one big rewrite with everything changing at the same time"?

DanVanAtta:
Re: Map Compatibility SLA

> Hmm, I'm curious as to what the rationale is.

I think there was a case of someone finding maps hosted in source forge (which are our maps) and were old 1.8 maps. 

One also needs to consider as well that the user community hardly upgrades all at the same time. For a long time there were 1.8 users concurrent with 1.9 (and even 1.5 and 1.4 users that were weeded out wondering why maps did not work for them).


re: (4) Most blocks could get a well named ID, and could have multiples of these. Then whenever needed, we could have something like:

```
<attachment name="Foo" javaClass="TriggerAttachment" playerId="playerListIdValue">
<attachment name="Foo2" javaClass="TriggerAttachment" allianceId="playerListIdValue">
```

> UnitAttachment should just be directly part of the block that defines that unit type, and not an attachment at all, since it's always 1-1. 

Yes, agree. There are a number of other cases that could be better nested like this.

> Not sure why those other formats are any different than XML - is it because there's no end tag to a block? 

I suppose they would not be that different. Thinking about the parsing logic, it seemingly would be ugly since you would have to code for an optional 'foreach' object, which means that a given element could exist at multipole leves of indentation.

>  It seems more concise to have an if statement than have to spell everything out the "low-level" way in that case.


As counter-example, this seemingly could be done very cleanly with an XML attributes:

```
<attachment name="Foo" javaClass="TriggerAttachment" ifNotPlayerList="playerListIdValue">
```

A person could have more attributes, like `forEachPlayerInList`, or `forEachPlayerNotInList` (and even combine those two expressions). We could further and have this be a list by using a sub-element:

```
<attachment name="Foo" javaClass="TriggerAttachment">
   <forEachPlayerInList>
       list1
       list2
   </forEachPlayerInList>
</attachment>

```


> Ultimately, the map format is meant to allow the mapmaker to set up the game how they want and within those constraints, the more readable, understandable and concise the format is, the better - since this would reduce errors and time wasted for map makers. 

I generally agree here. I think the ideal is that map makers would not need to deal with XML at all. This might have been a dream of one of the OG authors even (I recall that from something I read, but there is also evidence this was the intent with the presence of the export and map making tools that are built in).

We are clearly not there though, so we must be pragmatic and there is a large interation with the direct XML. Hence having that be less error prone is a benefit - we agree here.

>  I don't buy that we should prioritize "ideological" data format vs. what solves real problems in the best way with regard to the above properties I mention.

The reasons I see that are not ideological are:

1. We do not control when users change their engine version, incompatible formats causes a problem here. This is also to say that we've seen examples of users consistently using ancient versions of the engine and/or finding old maps. More to this point, the experience of doing across-the-board incompatible map upgrades for 1.8 to 1.9 was bad. It generally did not go well, at the end of it I hardly thought it was worth it in the end and somewhat said to myself "never again".


2. There is a significant need to really update how the XML is represented. We generally should support any XML ad-infinitum in one manner or another. I could see reasoning to say at some point we support the official versions only. This probably should be discussed with Roger Cooper, the implications for someone maintaining hundreds of zipped maps that could not shell script the changes would potentially be in a deep pickle. That is to say, avoiding the work of upgrading many maps is worth quite a lot. All this is to say, it's not clear whether strategic fixes right now for the convenience of one or two maps is worth maintaining as part of the 'old speck' vs simply moving to a more dramatic change directly.

3. Given (2), we will largely need to maintain 1.9 parsing in one manner or another to help transition and/or read older maps. The more stuff we add into 1.9, the more legacy code we are going to have in the medium-term future. If that code is only used for one or two maps, then it is seemingly a significant cost.

I wonder a bit if there a way we could 'mark' a map as using unstable XML (meaning the engine might one day stop parsing it and require it be changed).
 

A somewhat ideological rationale for restructuring how this is done:
 - XML should generally not have control flow logic in it. 'jinja2' templates or JSP are examples of templates with control flow blocks. Given we are running counter to the "philosophy" of XML, ie: it's intended use, we could potentially go down a bad path. Eventually we could be building our own version of 'jinja2' or something, and are re-inventing the wheel without quite realizing it. We could very well have a 'jinja2' renderer that would transform (via the engine) into XML which then could be read in as a standard game file. It would not take too much to add a jinj2 renderer and automatically use that if we detect there is a jinja2 map XML template. This would give variables, control flow blocks, and a bit more (jinja2 also has unique & sorting operations, union operators and quite a bit more).


> Or do you still have the stance of "one big rewrite with everything changing at the same time"?

My questions are:
1) Will this current effort complicate that rewrite, given we will have to continue maintaining 1.9 parsing in one form or another (we will want to have some way to automatically convert game files, so the 1.9 parsing code is unlikely to ever be fully removed).

2) Are we stacking up enough changes where we are going to want to do a big rewrite anyways? At what point should we just do the rewrite?


I am actually not sure if we will do a full rewrite. Some things I'm thinking we'll want is to un-hardoce the territory types. EG:


```
<territoryTypes>
   <territoryType>Land</territoryType>
   <territoryType>ShallowSea</territoryType>
   <territoryType>Sea</territoryType>
   <territoryType>Space</territoryType>
   <territoryType>Air</territoryType>
</territoryTypes>

<territory name="France">
   <type>Land</type>
   <type>Air</type>
</territory>
   

<territory name="Sz21">
   <type>ShallowSea</type>
   <type>Air</type>
</territory>


<units>
  <unit name="tank">
     <movementCosts>
       <movementCost territoryType="Land" cost="1">       
     </movementCosts>
  </unit>

  <unit name="Canoe">
     <movementCosts>
       <movementCost territoryType="ShallowSea" cost="1">       
       <movementCost territoryType="Sea" cost="2">       
     </movementCosts>
  </unit>
</units>
```

The above changes, looking at them, are pretty large, so we would have a new parser presumably for it. To really 'fix' up the XML at a deep level, it's hard to think we would do a new super-set format (kind of like how all JSON are valid YAML, it's unlikely that 3.0 XML will be compatible with 1.9 XML once the new and unknown tags are stripped away).

Hence, that is the rationale why I think if we are just doing something to benefit one map, it's perhaps better to just make that one map have ugly XML and then upgrade that one map to have 2.0 XML which will be nicer.  Further, most maps are written once, so I don't think there would be many maps that would be retrofitted with nicer 1.9 syntax, or even much benefit from that exercise since it would unlikely be updated much further again.

In the meantime there is quite a bit we can do to get started on the nicer XML (which we could merge in quite readily and incrementally even if the 3.0 format is read into a 1.9 style `GameData` java object).

In this case, as a suggestion how about we create a way to fork the parsing logic of XML and this new map could be the trailblazer for a nicer XML. We could essentially fork the parsing logic (perhaps by having if statements, perhaps via full blown copy/paste). Then, we gradualy modify the maps XML and the new 3.0 parsing code and evolve into a nicer structure. All of this work would be naturally feature gated since unless a map has the `<syntaxVersion>3.0</syntaxVersion>`, we would use the 1.9 parsing code.

I am strongly considering as well that following the 2.6 launch, any merge to master would result in a new release, and not a new prerelease. I mention this to speak to how I think every change will need to be done in an incremental fashion (I'd hope that no branch lives longer than a week or two), and that releases would be far more frequent. I do _not_ have a mindset where I think releases ought to be large, monolithic and incompatible jumps, I want quite the opposite.

TheDog-GH:
Put me down for helping/testing, even manually upgrading some of the older maps.

For version 3 please consider this; (repeated here in case asvitkine has not seen it)

I would like the current xml file to split into three files.

1) A pure map xml, this could then be shared with other mods & variants making it so much easier to have mods and variants, once written this code would hardly change.
- Territory Definitions
- Territory Connections
- Territory Attachments (Capitals would go into the mod/variant file) (could be in (3))
- Canal Attachments (could be in (3))

.
(2) Game Variables File
- This allows the different WW2 rulesets to be defined and thus called as a well established known.
- Also allows custom game variables

.
(3) The remainder of the code. This file would have the mods & variants code.
- Calls the map file to be used
- Calls the game variables file to be used.
- Units
- Politics
- Territory Effects (although it has map specific definitions it also has unit definitions)
- Victory Conditions (includes Capitals)
- ie. the rest of the code
 
You get the idea.

TheDog-GH:
Could this work?

A new map loader program that detects the best version of TripleA to run the map selected by the player.

So the new TripleA download would have the following, for example;

- Map Loader 
- TripleA ver 1.9
- TripleA ver 2.5.22294
- TripleA ver 3.0+

This way for TripleA version 3 backwards compatibility is just ignored and only maps written for version 3 are run by TripleA version 3.

asvitkine:
> Could this work?
> 
> A new map loader program that detects the best version of TripleA to run the map selected by the player. (Maybe based on triplea minimumVersion and the xml syntax)
> 
> So the new TripleA download would have the following, for example;
> 
> * Map Loader
> * TripleA ver 1.9
> * TripleA ver 2.5.22294
> * TripleA ver 3.0+
> 
> This way for TripleA version 3 backwards compatibility is just ignored and only maps written for version 3 are run by TripleA version 3.
> 
> This way you can build the xml structure you want without any baggage.

The problem with that would be that old vers maps are stuck on old version of TripleA (i.e. not benefiting from any new improvements, including UI changes, new operating system compatibility, etc). I don't think it's a great approach to be honest, I'd rather we just migrate maps going forward.

A variant of the above that I'd support would be to allow for maps to have different versions of the XML for different engine versions. This way, a map could start using 2.6 features, in a new XML, while still providing a version that works with 2.5. This would make migrating maps without impacting 2.5 more feasible, since right now it's a bit of a chicken and egg.

asvitkine:
> re: (4) Most blocks could get a well named ID, and could have multiples of these. Then whenever needed, we could have something like:
> 
> ```
> <attachment name="Foo" javaClass="TriggerAttachment" playerId="playerListIdValue">
> <attachment name="Foo2" javaClass="TriggerAttachment" allianceId="playerListIdValue">
> ```

I think that's too simplistic a view, as for example in my map, I have triggers that are foreach's for players * territories. So if you just have playerId="foo", that's not sufficient. You also need to have ways for other triggers to refer to them.

I think if we want to seriously explore this, we should author a spec document and iterate on it, but it wasn't my intention to rewrite how attachments work with this change, since this is about the turn sequence.

> > It seems more concise to have an if statement than have to spell everything out the "low-level" way in that case.
> 
> As counter-example, this seemingly could be done very cleanly with an XML attributes:
> 
> ```
> <attachment name="Foo" javaClass="TriggerAttachment" ifNotPlayerList="playerListIdValue">
> ```

That's true.

> The reasons I see that are not ideological are:
> 
> 1. We do not control when users change their engine version, incompatible formats causes a problem here. This is also to say that we've seen examples of users consistently using ancient versions of the engine and/or finding old maps. More to this point, the experience of doing across-the-board incompatible map upgrades for 1.8 to 1.9 was bad. It generally did not go well, at the end of it I hardly thought it was worth it in the end and somewhat said to myself "never again".

Right, maybe we should introduce a way for XML to specify which engine versions it's compatible with (assuming this doesn't exist already), but the challenge is that 2.5 and older versions won't understand that.

So we'd need to do some magic on the map download server or something. Our hands are a bit limited since we rely on archive.zip produced by GitHub for the downloads - I assume we don't have a way to transform that payload before it's served to clients. Perhaps the XML for newer versions can exist in a different folder or something such that 2.5 won't find it. And 2.6+ could support an XML tag specifying the version range an XML is compatible with.

This way, we can start using new features in XML without breaking support with old versions.

> 2. There is a significant need to really update how the XML is represented. We generally should support any XML ad-infinitum in one manner or another. I could see reasoning to say at some point we support the official versions only. This probably should be discussed with Roger Cooper, the implications for someone maintaining hundreds of zipped maps that could not shell script the changes would potentially be in a deep pickle. That is to say, avoiding the work of upgrading many maps is worth quite a lot. All this is to say, it's not clear whether strategic fixes right now for the convenience of one or two maps is worth maintaining as part of the 'old speck' vs simply moving to a more dramatic change directly.
> 3. Given (2), we will largely need to maintain 1.9 parsing in one manner or another to help transition and/or read older maps. The more stuff we add into 1.9, the more legacy code we are going to have in the medium-term future. If that code is only used for one or two maps, then it is seemingly a significant cost.

Presumably, we have this problem already for any XML / trigger option that was introduced? i.e. a map using it is not compatible with versions before it was introduced. But I don't know how frequently there have been changes here...

Also, I guess "forward" and "backward" compatibility are two separate topics - but they interact if we're ever discussing converting old maps that people are using today to better syntax...

> 
> I wonder a bit if there a way we could 'mark' a map as using unstable XML (meaning the engine might one day stop parsing it and require it be changed).

I mean, it's certainly possible but I think in practice, if too many maps start using something unstable, it defacto becomes stable. But I do think taking into account the ease of conversion is important - a change like this that streamlines the player turn list is one that I think would look very similar in any other improved XML syntax we choose, so converting should be trivial - particularly because of how concise it is (i.e. no need to change hundreds of lines).

> 
> A somewhat ideological rationale for restructuring how this is done:
> 
> * XML should generally not have control flow logic in it. 'jinja2' templates ...

I think the pros/cons of this are:
  1. Pro: We could definitely simplify the XML format itself.
  2. Con: If we still have to support old syntax, we still have a different system with foreach/variables for doing the same thing. So we're not really simplifying things much, if the result is introducing multiple codepaths.
  3. Con: We're now dependent on jinja2 library and its release philosophy, etc. If they make an incompatible change, we're on the hook for migrating all maps using it, or are stuck on an old version which may have security issues.
  4. It makes some things harder - for example, the pie-in-the-sky model where you "edit the map using Edit Mode and save XML", since the saved XML won't have the concise jinja2 form, unless we spend a lot of time implementing logic for that. It's actually already a problem for variables, though.

So overall, I think of it more of a con than a pro, but perhaps there are some reasons to alleviate some of these concerns (e.g. maybe jinja2 has a commitment to syntax compatibility). We could also choose something very mature like XSLT which likely won't change because of how old it is (but also, not the best syntax).

> 
> > Or do you still have the stance of "one big rewrite with everything changing at the same time"?
> 
> My questions are:
> 
> 1. Will this current effort complicate that rewrite, given we will have to continue maintaining 1.9 parsing in one form or another (we will want to have some way to automatically convert game files, so the 1.9 parsing code is unlikely to ever be fully removed).
> 2. Are we stacking up enough changes where we are going to want to do a big rewrite anyways? At what point should we just do the rewrite?
> 
> I am actually not sure if we will do a full rewrite. Some things I'm thinking we'll want is to un-hardoce the territory types. 

As I mentioned before, I don't think we should think of it as a "big rewrite" and instead incremental changes to support new, better XML syntax for different parts of the engine. In this way, we don't need to "stack up changes" or anything like that, instead we just add improved syntax for different things and maps can start using it. Eventually, we get to a point that we're able to fully specify a map using the new syntax and that becomes effectively "the new 3.0 syntax" or whatever.

In this model, a change like this could be seen as the new version of turn sequence syntax that paves way for other XML improvement. Now, in that context, we can discuss whether the `<forEach>` approach is still the best thing to do here, or something else would be better - as I mentioned, I'm not against your previous sketch example with `playerGroup` and `subSequence`. We can try to fully spec it and see if we like it and try to go with that instead.


> The above changes, looking at them, are pretty large, so we would have a new parser presumably for it. To really 'fix' up the XML at a deep level, it's hard to think we would do a new super-set format (kind of like how all JSON are valid YAML, it's unlikely that 3.0 XML will be compatible with 1.9 XML once the new and unknown tags are stripped away).

I disagree here. I think both:
  1. it is possible to have a superset format if we wanted to.
  2. the eventual "3.0 format" _doesn't_ have to be a superset format, but we could still implement new XML tags in the "old format" and have those be re-used in the new format when the time comes that we want to revisit the top-level XML structure. XML is pretty modular, so if we implement code to handle a new tag, it can be re-used if we decide to change where that tag is nested.

> 
> Hence, that is the rationale why I think if we are just doing something to benefit one map, it's perhaps better to just make that one map have ugly XML and then upgrade that one map to have 2.0 XML which will be nicer. Further, most maps are written once, so I don't think there would be many maps that would be retrofitted with nicer 1.9 syntax, or even much benefit from that exercise since it would unlikely be updated much further again.

I agree that that we shouldn't really be doing something to benefit "one map". But a change like this will literally benefit every new map created. I agree that there's a question of what happens to old maps and whether we keep them "as-is" but that argument can be made about any XML improvement. Like, you can say "3.0 XML syntax" would only benefit "one map" at the time when there's only one map that's using it. Over time, it benefits more and more maps.

For this specific change, I would wager if you take 20 maps, 18 or 19 of them would be able to switch to this syntax and hence would have benefited from it, had it been available at the time they're authored.

Certainly, if you want to play the game of arguing that because there exist 100 maps already, anything new won't benefit them therefore we should stop adding features, you can do that. But is misses the point that improving things can improve the ecosystem going forward.

> In the meantime there is quite a bit we can do to get started on the nicer XML (which we could merge in quite readily and incrementally even if the 3.0 format is read into a 1.9 style `GameData` java object).
> 
> In this case, as a suggestion how about we create a way to fork the parsing logic of XML and this new map could be the trailblazer for a nicer XML. We could essentially fork the parsing logic (perhaps by having if statements, perhaps via full blown copy/paste). Then, we gradualy modify the maps XML and the new 3.0 parsing code and evolve into a nicer structure. All of this work would be naturally feature gated since unless a map has the `<syntaxVersion>3.0</syntaxVersion>`, we would use the 1.9 parsing code.

I don't think we need to fork the logic. I see no reason why we can't introduce new tag types, which if present, would use the new functionality and if absent, things would work as before. Isn't the presence of those tags the same as `<syntaxVersion>`? Why do we need to introduce technical debt by forking code?

I guess maybe you want to avoid a "hybrid model" where you only want to support 1.9 and 3.0 parsing code in the end state, but I really don't see why it needs to be that way. If we already have written code for 3.0 that takes a Foo tag and creates some state in GameData from it, it's not much complexity to have that code be called both from 1.9 and 3.0 code in the presence of a Foo tag.

And in fact, if we want to support 1.9 parsing forever, why introduce more technical debt by forking code? If we just want to declare maps as "3.0 clean" eventually, that can still be done by ifs for any 1.9-feature parsing code, which will be distinct from new code. But it's also not clear it's really needed (what would it matter if a map says it's "3.0 clean" if we never plan to deprecate the old syntax). If we *do* plan to deprecate the old syntax, then sure the argument could be that it makes it easier to delete the old code - but I find even that pretty bogus, it's not that hard to delete code even if it's not cleanly separated, sure a bit more work, but much less than e.g. adding any non-trivial feature.

> 
> I am strongly considering as well that following the 2.6 launch, any merge to master would result in a new release, and not a new prerelease. I mention this to speak to how I think every change will need to be done in an incremental fashion (I'd hope that no branch lives longer than a week or two), and that releases would be far more frequent. I do _not_ have a mindset where I think releases ought to be large, monolithic and incompatible jumps, I want quite the opposite.

This seems like a separate discussion, but I think such a model is essentially making "prereleases" "releases". I think the effect of this is everything will be constantly broken. As things are fixed, new things get broken. Without expanding test coverage significantly, I don't think it would be a good experience for players. We need time to stabilize releases because right now it's too easy to introduce bugs. But anyway, that's a whole other discussion.

....

For this change, to move forward, here's my proposal:
  1. We agree that it's OK to add new incremental features that work towards a 3.0 XML rewrite (i.e. that the syntax that's added is the new 3.0 syntax for those features)
  2. We decide on the syntax for players / sequences we'd like (I think your proposal earlier can work and we can narrow down some details in a spec).
  3. I can change this PR to implement that and we can land that

That's all assuming you're OK with 1. If not, then I don't know what's the way forward, maybe just abandoning this. At the end of the day, randomizing the player list is not so important to me at the end, since that can easily be done by players offline before starting the game. So I can focus on other improvements for my map (I've been working on getting AI to deal with more engine features like consumable units that it so far didn't support, for example).

TheDog-GH:
@DanVanAtta

> I am strongly considering as well that following the 2.6 launch, any merge to master would result in a new release, and not a new prerelease. I mention this to speak to how I think every change will need to be done in an incremental fashion (I'd hope that no branch lives longer than a week or two), and that releases would be far more frequent. I do not have a mindset where I think releases ought to be large, monolithic and incompatible jumps, I want quite the opposite.

Please dont do the above, there will be hundreds of versions on the Players computers, the forum mods will be saying upgrade all the time.

As a player I would prefer to have a **stable yearly regular update**, say in January, and not have to wait years for the next round of goodies/improvements.

Consider a new player, downloads a buggy version and is totally put off, they would be lost to us.

Apologies for digressing, but I feel strongly about this.

DanVanAtta:
@asvitkine 

- For jinja2, we could build the translation from template to XML, into the engine. The engine would only need to detect it sees a '.j2' file and then convert it and then load the XML.
- The jinja2 template syntax is stable

I've been thinking quite a bit whether we should evolve the XML syntax into a new variant, or if we simply create a new version of it. The 'min-version' tag works so long as the engine continues to support all preceding variants of the XML.

This is where I am not sure. If we are going to radically change the XML, would there be a lot of intermediate syntax as we morph into something new? In this case, we have to support all of these variants. Seemingly that would make things like detecting if the XML is logically valid (has all needed elements), very difficult since any variety of XML could be used. We would also need to decide and detect how to handle cases where there is redundant XML (ie: old and new syntax that possibly conflicts mixed together). This makes me unsure whether we want to evolve or have a new iteration of the syntax.

If we go with a new syntax, we can do so iteratively and build it in phases, but without the burden of supporting intermediate formats until we feel the syntax is ready to be used in a more widespread manner.

I think the choice of 'evolve' vs 'iterate to a new version & then evolve' is something of *the* question. As far as having loops, I think we can also decide to support jinja2 templates to allow for easier map making (which would very cleanly support variables, if statements, and loops).

asvitkine:
@DanVanAtta 

For your concerns about intermediate formats, I agree it's hard to make a call at a high level in abstract without diving into the actual details of a specific proposal - i.e. for a given proposed new format, consider whether it expresses concepts in a maintainable and future proof way - and what the actual maintainability impact is of having it incremental (concretely in the context of the new spec being proposed).

For example, I made the argument that XML tag handling can be pretty modular - i.e. if we add a new `<foo>` tag, the code to parse it can be isolated and unit tested and can be plugged into a new XML format pretty easily when time, but in the meantime usable from current XML. If done this way, it would be a minimal maintainability impact - that is, we just have a call site into that code from the old code - so literally a line of code, calling code that will be re-used in new XML syntax.

In fact, I'd argue it would be *more* maintainable than an approach where some code is written but not integrated waiting for other rewrites at all and no maps exercise it,  because then it can easily bitrot as the rest of the engine changes and no one is using it.

....

As for jinga2, I'm not against that, but it doesn't actually address the motivation for this PR, which is to have small follow-up PR to support randomizing player order.

I guess if we really want to abandon this, I can look at a solution that doesn't involve changing anything in the step sequence XML - just adding a new map property. That property can then invoke an algorithm that reconstructs the conceptual original structure underlying the step list - i.e. grouping steps of players together and treating them as blocks it re-arranges. Then, we can stick with the current sucky XML and just let the above algorithm take care of the mess.

Given how much you've been pushing back, perhaps you'd find this more acceptable?

But then you're basically rejecting my offer of investing my time into making progress towards the new XML rewrite using the incremental approach I mentioned - which I think makes that project more pie in the sky and less likely to happen... coupled with you wanting to freeze the current XML, it seems the result would be a stagnation of new feature development in TripleA, which would be a pretty frustrating outcome...

beelee1:
@asvitkine 

sorry to see you go. Pretty obvious why  

Peace

DanVanAtta:
> In fact, I'd argue it would be more maintainable than an approach where some code is written but not integrated waiting for other rewrites at all and no maps exercise it, because then it can easily bitrot as the rest of the engine changes and no one is using it.


To clarify, I think the alternative approach would be:
- freeze the 1.9 syntax
- start building the 3.0 syntax from the ground up with XML snippets written along the way, working towards having a complete map



> As for jinga2, I'm not against that, but it doesn't actually address the motivation for this PR, which is to have small follow-up PR to support randomizing player order.

Sorry to not have the full context of what you are trying to do. Jinja2 would support a 'for-each' loop, and arguably should have been the way we supported variables in the first place. My push back is to avoid the XML from becoming more a template-language rather than being a declarative-language. Hence the concern about 'canonical' ways to do things.

My preference for a 'for-each' loop would be to avoid it by doing some sort of canonical XML as we have discussed earlier, with dedicated tags that create a list and then use optional attributes to reference those lists to create a 'for-each' effect, or alternatively add a jinja2 parser to the engine allowing for jinja2 markup to be added. I am not in favor of an XML `for-each` tag if we can cleanly do one of these alternatives (which I think you are on board with, which means I think we are only discussing whether to do a new version of the XML syntax, or to evolve it).


> But then you're basically rejecting my offer of investing my time into making progress towards the new XML rewrite using the incremental approach I mentioned - which I think makes that project more pie in the sky and less likely to happen... coupled with you wanting to freeze the current XML, it seems the result would be a stagnation of new feature development in TripleA, which would be a pretty frustrating outcome...

I hear this. I am on the fence of how to best approach this. Overall, after observiing our progress for about 5 years, there is no "refactoring" our way out of the mess IMHO. We've been trying to clean up so much of the code for so long, it's better in many places - but I think we can only really make progress by fixing the underlying design.

Check this post that describes the "3.0" initiative: https://forums.triplea-game.org/topic/2794/triplea-3-0-design-proposal-discussion?page=1

Where I'm unsure, if we are going to basically alter the entirety of the XML in a very drastic way, how clean will it be to evolve the XML parsing code into that shape vs having a second XML parsing module dedicated to the new XML shape?


---------------------

re: proposed next steps

> We agree that it's OK to add new incremental features that work towards a 3.0 XML rewrite (i.e. that the syntax that's added is the new 3.0 syntax for those features)

I think we need to evaluate this, it's hard to know which is the better answer, and reversing this decision could be difficult. Unfortunately XML changes are some of the most impactful because we have to support them essentially ad-infinitum. So, IMHO, this is both a difficult and an open question.

DanVanAtta:
## incremental XML changes or version jump (we might need to design the new syntax up front either way)

I wonder if we need to design the new XML spec almost entirely before we can make either choice of how to proceed. A concern I have with incremental is that we might not work towards a unified goal. In other words, some efforts or steps might have to be redone and otherwise hinder further progress.

We probably should document the XML syntax in a more systematic way. I wonder if we can leverage annotations to generate documentation and add some attributes for descriptions. In this manner we could also add an "unstable" tag to denote XML entities that could change in the future. This would give us some more freedom for intermediate syntax.

DanVanAtta:
### Addendum regarding incremental vs jump version, many ways to do one thing vs just one

Let's say we spec out the full and new XML syntax. Will we truly want there to be many ways to do equivalent things, or would we prefer for there to be just one way?  It's hard to say what the long term cost would be for the complexity of many ways to do the same thing.

Further, would having more than one XML variant of the same tags be prohibitive to having a clean XML validation system? If there were a single way to define a territory and its connections, presumably it would be a lot easier to detect that the territory is improperly configured. If we can raise this type of validation to a game data level, then the error message could be clean, but it would seemingly be really hard to give error messages that could include specific line numbers or to reference specific XML snippets in any such validation error messages.

In part some of this is Devils Advocate style arguments, but given this is seemingly a difficult decision to reverse, we need to be careful about it.

DanVanAtta:
@asvitkine I see a lot of pro's by going with an 'evolve' approach vs a completely new spec.

I think we would benefit from an up-front design doc of what a new syntax would look like. Up-front design has its issues, but I think there are benefits to having some fore-thought on where the XML syntax is going to go:
- we can avoid excessive intermediate formats that we wind up having to support unnecessarily
- we can ensure that any updates are working towards a greater whole that is cohesive and works. In other words, we won't have to hack our ways from potentially short sighted decisions
- perhaps easier to support map & engine compatibility
- can help ensure we are working towards something that would better support an updated game engine (specifically saves). Eventually it would be great if we can have save games be nothing more than a 'map id' and a set of 'game changes' that when applied in sequence would recreate the game engine state. So, there are a few future re-design initiatives we should keep in mind as we evolve the XML.

For that latter point regarding many engine versions - if we are increase the rate of change to the map XML syntax, how do we handle the game engine upgrade issue? Specifically we do not want everyone to be required to upgrade their engine every few weeks because we are evolving new features to the XML syntax. I think having a design doc for the new syntax would help for this, as we could create logical 'compatibility versions' or checkpoints while we evolve the design so that we don't have to do a version bump every time we have a new syntax available.

asvitkine:
That's fair. I think this means we should abandon this PR in the meantime.

I'll try to split out the code clean up and added test coverage unrelated to the actual forEach feature into a separate PR, so we can land those without that effort being fully lost.

Then we can collaborate on the spec. Just a heads up that as of next week, my free time is going to be significantly diminished so I won't be able to be quite as active on TripleA development as I have been these past few months.

DanVanAtta:
Sounds like a plan @asvitkine :+1: 

Thank you for your patience hanging in there while we worked through what might have been an unexpectedly important decision.

Also - thank you for cranking out the updates! It's all really appreciated and not easy stuff, it was quite impressive to see that many PRs roll through. Hopefully you'll find time here and there for a few more    :sweat_smile:

asvitkine:
Closing per the discussion above where we decided not to go forward with this.

PR #10429, opened 2022-05-10
Fix history related errors caused by some changes being done twice.

asvitkine:
Fix history related errors caused by some changes being done twice.

When you're not the active player, the game operates on a history view where events are added and changes are applied.
In some circumstances, the previous code would result in the same change being applied twice, which can result in an error - for example, in a custom map I have, I have a trigger that resets a resource to 0, which would sometimes trigger an error like "Ca't remove more than player has of resource:".

The root cause was a logic error in getLastChange(). When the current node was a Step node (and thus a IndexedHistoryNode), it would return its first change, rather than its last change. This would result in the history model "fast-forwarding" to apply its other changes, applying changes that were already applied.

I looked where the logic was first added and it looks like the bug existed in the very first implementation of history, here: https://github.com/triplea-game/triplea/commit/4ddf7a05237ea1e584bef959874c0b3697839988

A few small bits of clean up are included.

## Release Note

FIX|Some errors related to history view

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/ResourceCollection.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/ResourceCollection.java
@@ -50,7 +50,7 @@ public class ResourceCollection extends GameDataComponent {
     final int current = getQuantity(resource);
     if ((current - quantity) < 0) {
       throw new IllegalArgumentException(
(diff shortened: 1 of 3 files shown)


DanVanAtta:
Looks good :+1: 
Release note could use a bit of work to be more meaningful.

PR #10442, opened 2022-05-15
Speed up AI turns and reduce memory and save game sizes.

asvitkine:

Speed up AI turns and reduce memory and save game sizes, by not instantiating empty collections for many different properties of attachments.

Standardize collection getters with some helper functions, which return the immutable empty collection in case of null or else unmodifiable views of the underlying collection. Introduce an unmodifiable empty IntegerMap and the ability to return an unmodifiable view of an IntegerMap, for consistency with other collection methods.

Annotate null fields with `@Nullable` which lets IDEs help identify warnings when used without null checks directly.

Also made a few additional clean ups as a result of codeclimate warnings on the existing code.

Context:
I noticed in performance profiles a lot of time being spent in auto-saving games and cloning the game state for AI battle simulations when testing a map I'm working on with a lot of triggers (via many foreach variables). Turns out that attachments were carrying a lot of additional empty state in the form of empty collections, which impact memory use and game serialization time.

With this change, I observed a pretty sizeable reduction in save game sizes on the same map (e.g. 1.4M to 1.2M), which should translate to memory use in general (especially with AIs doing lots of battle simulations with copied game states), as well as engine execution speed. Also note that saved game files are compressed with GZIP so, the underlying saving is even larger.

Save game compatibility considerations:
1. Old save games should continue to work fine, as loading object that have these collections set to empty still works correctly.
2. New save games will not be compatible with older versions (will cause NPEs), but that's already something that's not supported.

Testing done:
Played several rounds of all-AI games on a variety of different maps, ensuring there are no NPEs.


TheDog-GH:
Count me in you want help testing this, when its ready.

asvitkine:
I addressed most of the codeclimate issues. The remaining two is actually one, where codeclimate thinks there's code duplication between `setTerritoryEffects()` and `setRelationshipTypes()`, but actually the two are different since they work on different types of objects, so I actually don't think that trying to unify them is the right thing to do here.

@RoiEXLab ready for review.

Cernelius:
Not a game of a popular map, but have you tested it on "270BC Wars"? I believe that is the heaviest game in TripleA.

asvitkine:
> Not a game of a popular map, but have you tested it on "270BC Wars"? I believe that is the heaviest game in TripleA.

Just did. No errors. It doesn't look like the map benefits that much from it since save files are only ~50K for it, with only a KB of saving. This has bigger effect when there's lots of triggers (e.g. when triggers are foreach'd for all territories and players).

Cernelius:
> > Not a game of a popular map, but have you tested it on "270BC Wars"? I believe that is the heaviest game in TripleA.
> 
> Just did. No errors. It doesn't look like the map benefits that much from it since save files are only ~50K for it, with only a KB of saving. This has bigger effect when there's lots of triggers (e.g. when triggers are foreach'd for all territories and players).

I believe "270BC Wars" has the biggest amount of triggers of every other TripleA game, even though it is not using the "foreach" code. Are you sue you are using "270BC Wars" (and not "270BC" or something)?
I just took a save-game at the start of the first "Carthage Combat Move" phase using TripleA 2.6.760: it is 1,205,046 bytes (1.14 MB), and the version I have is actually some lighter than the one in the repository.

asvitkine:
> I believe "270BC Wars" has the biggest amount of triggers of every other TripleA game, even though it is not using the "foreach" code. Are you sue you are using "270BC Wars" (and not "270BC" or something)? I just took a save-game at the start of the first "Carthage Combat Move" phase using TripleA 2.6.760: it is 1,205,046 bytes (1.14 MB), and the version I have is actually some lighter than the one in the repository.

Ah, I did use "270BC"! With "270BC Wars", I see it going from ~1.1MB to ~1MB.

game-app/game-core/src/main/java/games/strategy/triplea/attachments/TerritoryAttachment.java
```diff
@@ -264,14 +268,7 @@ private void resetIsImpassable() {
...
   public void setCapital(final String value) throws GameParseException {
-    if (value == null) {
-      capital = null;
-      return;
-    }
-    final GamePlayer p = getData().getPlayerList().getPlayerId(value);
-    if (p == null) {
-      throw new GameParseException("No Player named: " + value + thisErrorMsg());
-    }
+    getPlayerOrThrow(value);
```
RoiEXLab, line 271: Is this a no-op for null values? Otherwise we'd get an exception when passing null here, right?

RoiEXLab (commented):
I briefly looked at this PR and from what I saw the changes look fine. The performance and memory gain is really nice, but I dislike that this step introduces the now mandatory null checks that could be avoided so far.

game-app/game-core/src/main/java/games/strategy/triplea/attachments/TerritoryAttachment.java
asvitkine: Null isn't supposed to be passed to setters on attachments. Some test were doing that, which are fixed.

Otherwise, the setters come from parsing XML attributes, which won't ever be null.

asvitkine:
> I briefly looked at this PR and from what I saw the changes look fine. The performance and memory gain is really nice, but I dislike that this step introduces the now mandatory null checks that could be avoided so far.

We just do the null checkers when checking the private members inside the code of the class itself, getters are made to return non-null values.

asvitkine:
> > I briefly looked at this PR and from what I saw the changes look fine. The performance and memory gain is really nice, but I dislike that this step introduces the now mandatory null checks that could be avoided so far.
> 
> We just do the null checkers when checking the private members inside the code of the class itself, getters are made to return non-null values.

And just to add more, I think it makes sense to constrain this change to the attachment classes only, given their special nature and I'm definitely not advocating for doing this sort of thing more broadly.

asvitkine:
> With "270BC Wars", I see it going from ~1.1MB to ~1MB.

And to add to this, the underlying save file is actually .gz encoded, so if I decompress those two files (which the game engine does when loading) and compare the sizes, it's 23.7MB to 20.7MB, so this is saving 3MB of useless collections on that map!

Cernelius:
> > With "270BC Wars", I see it going from ~1.1MB to ~1MB.
> 
> And to add to this, the underlying save file is actually .gz encoded, so if I decompress those two files (which the game engine does when loading) and compare the sizes, it's 23.7MB to 20.7MB, so this is saving 3MB of useless empty lists on that map!

Is this sizeable? The dimensions are 87% of the original, so I guess the gains won't be humanly perceptible. How about the time that the battle-calculator takes to load the data before and after?

asvitkine:
> Is this sizeable? The dimensions are 87% of the original, so I guess the gains won't be humanly perceptible. How about the time that the battle-calculator takes to load the data before and after?

I think the savings for the battle calculator are similar (so not too perceptible). But it helps.

For the battle calculator delay, I had a look at the code and have a way to speed up even more with a different trick. (Right now, we copy the data N times, but actually copying involves serializing and deserializing, so we can serialize once and deserialize N times, which provides some savings.) I'll do a separate PR with that.

game-app/game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
```diff
@@ -156,4 +151,69 @@ public final boolean equals(final Object obj) {
...
+    return Optional.ofNullable(getData().getMap().getTerritory(name))
+        .orElseThrow(() -> new GameParseException("No territory named: " + name + thisErrorMsg()));
+  }
+
+  protected List<GamePlayer> parsePlayerList(final String value, List<GamePlayer> existingList)
+      throws GameParseException {
+    for (final String name : splitOnColon(value)) {
+      if (existingList == null) {
+        existingList = new ArrayList<>();
+      }
```
RoiEXLab, line 165: I dislike this kind of if check. If the parameter is passed we modify it and return it, otherwise we return a new object and return that. I would appreciate if this method always returns a new list and  the callers call either addAll on an existing collection or assign the new collection to the result. I think this would be cleaner code style. Applies to the other variants of this down below too of course

game-app/game-core/src/main/java/games/strategy/triplea/attachments/AbstractRulesAttachment.java
```diff
@@ -179,10 +177,6 @@ private void resetGameProperty() {
   }
 
   private void setRounds(final String rounds) throws GameParseException {
```
RoiEXLab, line 179: I assume rounds is not ever supposed to be null here, just like the other case where I asked?

game-app/game-core/src/main/java/games/strategy/triplea/attachments/AbstractRulesAttachment.java
```diff
@@ -217,17 +211,17 @@ private void setRounds(final Map<Integer, Integer> value) {
...
   private void resetRounds() {
     turns = null;
   }
 
-  protected boolean checkTurns(final GameState data) {
+  protected boolean checkRounds(final GameState data) {
     final int turn = data.getSequence().getRound();
-    for (final int t : turns.keySet()) {
-      if (turn >= t && turn <= turns.get(t)) {
+    for (final int t : getRounds().keySet()) {
```
RoiEXLab, line 223: Instead of using `keySet()` and then calling `get()` with the key, I'd recommend using `entrySet()` instead

game-app/game-core/src/main/java/games/strategy/triplea/attachments/TriggerAttachment.java
```diff
@@ -550,41 +543,28 @@ private void setAvailableTech(final Map<String, Map<TechAdvance, Boolean>> value
...
-      if (!found) {
-        throw new GameParseException(
-            "Could not find unitSupportAttachment. name:" + s[i] + thisErrorMsg());
+  private void setSupport(final String value) throws GameParseException {
+    for (final String entry : splitOnColon(value)) {
+      final boolean remove = entry.startsWith("-");
+      final String name = remove ? entry.substring(1) : entry;
+      UnitSupportAttachment.get(getData().getUnitTypeList()).stream()
+          .filter(support -> support.getName().equals(name))
+          .findAny()
```
RoiEXLab, line 559: Well that's kinda weird if the value is not actually used.

game-app/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
```diff
@@ -642,13 +625,8 @@ private void setIsSub(final String s) {
   @VisibleForTesting
   public void setIsSub(final Boolean s) {
```
RoiEXLab, line 626: Isn't the parameter kind of redundant now?

RoiEXLab (commented):
Looking pretty, good I have some concerns though

game-app/game-core/src/main/java/games/strategy/triplea/attachments/AbstractRulesAttachment.java
asvitkine: Correct, every setter is the same.

game-app/game-core/src/main/java/games/strategy/triplea/attachments/TriggerAttachment.java
asvitkine: It's meant for validation. Do you prefer noneMatch() and an if statement with the throw in the if?

game-app/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
```diff
@@ -1811,11 +1779,7 @@ private void resetCanDieFromReachingMaxDamage() {
   @VisibleForTesting
   public void setIsSuicide(final Boolean s) {
```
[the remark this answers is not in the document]
asvitkine: Hmm, that's a good point, I think this is a bug?

We should do `isSuicide = s;`.

The existing code doesn't handle `<option name="isSuicide" value="false"/>` correctly, because `isSuicide` is always set to true!

I have a downloadedMaps folder with all available TripleA maps and I searched through the XML files and didn't see any occurrences of the above, but I did see instances of:
`            <option name="unitProperty" value="isSuicide" count="false"/>`

i.e. setting the property to false via a trigger. This is on `total_world_war`. So I guess we need to check with that map whether fixing it has an unintended consequence or whether it will fix a bug that exists on that map due to this not working.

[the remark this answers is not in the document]
asvitkine: Looking more at how this is used in `total_world_war` map, they do two triggers:]

```
       <attachment name="triggerAttachmentjapaneseAdvRock2" attachTo="Japan" javaClass="TriggerAttachment" type="player">
            <option name="trigger" value="conditionAttachmentjapaneseAdvancedRockets"/>
            <option name="unitType" value="japaneseAdvancedRocket"/>
            <option name="unitProperty" value="isAir" count="false"/>
            <option name="unitProperty" value="isStrategicBomber" count="false"/>
            <option name="unitProperty" value="isSuicide" count="false"/>
            <option name="unitProperty" value="isKamikaze" count="false"/>
            <option name="unitProperty" value="isLandTransportable" count="true"/>
            <option name="when" value="before:japanNonCombatMove"/>
        </attachment>
        <attachment name="triggerAttachmentjapaneseAdvRock3" attachTo="Japan" javaClass="TriggerAttachment" type="player">
            <option name="trigger" value="conditionAttachmentjapaneseAdvancedRockets"/>
            <option name="unitType" value="japaneseAdvancedRocket"/>
            <option name="unitProperty" value="isLandTransportable" count="false"/>
            <option name="unitProperty" value="isAir" count="true"/>
            <option name="unitProperty" value="isStrategicBomber" count="true"/>
            <option name="unitProperty" value="isSuicide" count="true"/>
            <option name="unitProperty" value="isKamikaze" count="true"/>
            <option name="when" value="after:japanNonCombatMove"/>
        </attachment>
```

So basically, before NonCombatMove, the rockets are made land transportable and after NonCombatMove, they're returned back to non-land transportable. (And the suicide property is also changed to false then to true.)

So I think fixing isSuicide=false to work correctly shouldn't cause issues for this map and it's the only one setting it to false. I'll go ahead and fix it.

game-app/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
asvitkine: Same as for `setIsSuicide()`, this was a bug. Fixed by setting `isSub` to `s`. I checked all maps and none set this to false, so it should be safe to fix.

game-app/game-core/src/main/java/games/strategy/triplea/attachments/AbstractRulesAttachment.java
asvitkine: Done.

game-app/game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
asvitkine: I agree it's messy, but the point of this helper is to reduce duplicate code. Let's say it were changed to:

```
  protected List<GamePlayer> parsePlayerList(final String value)
      throws GameParseException {
    List<GamePlayer> players = new ArrayList<>();
    for (final String name : splitOnColon(value)) {
      existingList.add(getPlayerOrThrow(name));
    }
    return players;
  }
```

Then, what happens to the callers (we have 12 of them)?

Something like this:

```
  private void setHelpPayTechCost(final String value) throws GameParseException {
    helpPayTechCost = parsePlayerList(value, helpPayTechCost);
  }
```

Becomes:

```
  private void setHelpPayTechCost(final String value) throws GameParseException {
    if (helpPayTechCost == null) {
      helpPayTechCost = new ArrayList<>();
    }
    helpPayTechCost.addAll(parsePlayerList(value));
  }
```

I'd guess that codeclimate would flag it as duplicated code.

There are other ways to formulate it, but all are messy. For example, we could do:

```
helpPayTechCost = Optional.ofNullable(helpPayTechCost).orElse(new ArrayList<>());
helpPayTechCost.addAll(parsePlayerList(value));
```

Or we could do:

```
helpPayTechCost = createIfNull(helpPayTechCost);
helpPayTechCost.addAll(parsePlayerList(value));
```

Or:
```
helpPayTechCost = mergeLists(helpPayTechCost, parsePlayerList(value));
```

with the latter basically wrapping one of the above versions.

They're all kind of messy and the current approach is the one that results in the least code at call sites.

RoiEXLab (approved):

RoiEXLab:
Feel free to merge if you're happy with the state of this PR.

game-app/game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
asvitkine: I'll keep as-is since it's the most concise, but I'm also OK with something like the `mergeLists()` version. We can do that as a follow up later if we like.

PR #10454, opened 2022-05-17
Significantly speed up game data cloning for AI and battle calculator.

asvitkine:
Cloning game data involved creating a byte[] array out of the data and then converting it to a new GameData object.
Previously, both steps would be done for each concurrent worker. With this change, we only create the byte[] array once instead of N times, speeding up the process.

Additionally, we no longer compress/decompress the data when cloning nor create temporary files to write the data to, which speeds up things considerably. (This is still done when saving games, but not for cloning data.)

Also fixes a couple comments.

With this change, opening up the battle calc goes from 7.2s to 2.5s on my machine on the "270BC Wars" map for example.

## Release Note

Change|Major speed up to opening the battle calculator and AIs

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/SerializationProxySupport.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/SerializationProxySupport.java
@@ -12,8 +12,8 @@ package games.strategy.engine.data;
  *   <li>https://youtu.be/V1vQf4qyMXg?t=56m12s
  * </ul>
  *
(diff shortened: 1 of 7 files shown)


asvitkine:
Another big saving comes from need gzipping/ungzipping the data when we're not writing to disk. I'm going to look at changing that too - which brings the battle calc bring up to around 3s!

RoiEXLab (approved):
I really like this change. It simplifies the dreaded `ConcurrentBattleCalculator` by quite a bit. Good job

PR #10458, opened 2022-05-17
Deprecate NULL_PLAYERID in favor of PlayerList.getNullPlayer().

asvitkine:
Deprecate NULL_PLAYERID in favor of PlayerList.getNullPlayer().

This change makes the null player be bound to an actual GameData object and fixes the issue where player.getData() may return null. This allows us to simplify a lot of code in the future by passing around fewer parameters to methods that take a GamePlayer, since the GameData can be retrieved from it at all times.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/GamePlayer.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/GamePlayer.java
@@ -16,6 +16,7 @@ import javax.annotation.Nullable;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.EqualsAndHashCode;
+import org.triplea.java.RemoveOnNextMajorRelease;
 
 /** A game player (nation, power, etc.). */
 public class GamePlayer extends NamedAttachable implements NamedUnitHolder {
@@ -24,7 +25,8 @@ public class GamePlayer extends NamedAttachable implements NamedUnitHolder {
   private static final String DEFAULT_TYPE_AI = "AI";
   private static final String DEFAULT_TYPE_DOES_NOTHING = "DoesNothing";
 
-  public static final GamePlayer NULL_PLAYERID =
+  @RemoveOnNextMajorRelease
+  private static final GamePlayer NULL_PLAYERID =
       new GamePlayer(Constants.PLAYER_NAME_NEUTRAL, true, false, null, false, null) {
         private static final long serialVersionUID = -6596127754502509049L;
 
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/GamePlayer.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/GamePlayer.java
@@ -27,6 +27,7 @@ public class GamePlayer extends NamedAttachable implements NamedUnitHolder {
 
   @RemoveOnNextMajorRelease @Deprecated
   private static final GamePlayer NULL_PLAYERID =
+      // Kept for save game compatibility, or we'll get a class not found error loading neutrals.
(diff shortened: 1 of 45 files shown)


game-app/game-core/src/main/java/games/strategy/engine/data/GamePlayer.java
```diff
@@ -24,7 +25,8 @@ public class GamePlayer extends NamedAttachable implements NamedUnitHolder {
   private static final String DEFAULT_TYPE_AI = "AI";
   private static final String DEFAULT_TYPE_DOES_NOTHING = "DoesNothing";
 
-  public static final GamePlayer NULL_PLAYERID =
+  @RemoveOnNextMajorRelease
+  private static final GamePlayer NULL_PLAYERID =
```
RoiEXLab, line 29: What speaks against straight up removing this field? Static fields don't impact serialization, right?

game-app/game-core/src/main/java/games/strategy/engine/data/PlayerList.java
```diff
@@ -20,9 +22,19 @@ public class PlayerList extends GameDataComponent implements Iterable<GamePlayer
...
     super(data);
+    nullPlayer =
+        new GamePlayer(Constants.PLAYER_NAME_NEUTRAL, true, false, null, false, data) {
+          private static final long serialVersionUID = 1;
+
+          @Override
+          public boolean isNull() {
+            return true;
+          }
+        };
```
RoiEXLab, line 37: Did you consider making this a simple "getter"method without a field that backs it, instead creating a new instance on every call?

asvitkine:
Converting to draft since I need to resolve some failing checks.

game-app/game-core/src/main/java/games/strategy/engine/data/GamePlayer.java
asvitkine: Actually, looking more, I think the reason I kept it is that references to this object will exist in the game and since it's a custom subclass, we'll get a class not found error if we delete, I think?

RoiEXLab: No Idea how deserialization of anonymous classes works. But is the null player ever being serialized, or always just used for comparison?

triplea_maps.yaml
```diff
@@ -2623,3 +2623,10 @@
   img: https://raw.githubusercontent.com/triplea-maps/nwovichy/master/preview.png
   description: |
    <br>World War II in Europe with Vichy France acting as a full member of the Axis. Uses New World Order map.
+- mapName: World War I 1914
+  mapCategory: EXPERIMENTAL
+  url: https://github.com/triplea-maps/world_war_i_1914/archive/master.zip
+  version: 1
+  img: https://raw.githubusercontent.com/triplea-maps/world_war_i_1914/master/preview.png
+  description: |
+   <br>Based upon the Axis & Allies: 1914 boardgamew
```
RoiEXLab, line 2632: This is not supposed to be in  this PR.

asvitkine: Did another rebase, should be good now.

game-app/game-core/src/main/java/games/strategy/engine/data/GamePlayer.java
asvitkine: It does get serialized, since it's the default owner for territories and units there when an explicit player isn't listed in the map XML.

asvitkine:
@RoiEXLab this is now passing all the checks (some unit tests needed updating to use a mock GameData with a mock PlayerList to ensure the new way to access the null player was ok for them).

PR #10470, opened 2022-05-19
Intern attachment-related strings to minimize memory and speed up cloning.

asvitkine:
Intern attachment-related strings to minimize memory and speed up cloning.

Java's String.intern() function returns a canonical String object for the given string,
such that Strings that share content can use the same instance.

This is very relevant for triggers properties, where lots of triggers end up having the same properties set, either via variable expansion or even in older maps, where this was done manually.

This also helps tremendously with GameData cloning, since there's much fewer String objects to clone, since the same String instance is pointed to by multiple objects.

With this change, bringing up the Battle Calculator on "270BC Wars" goes from ~1.5s to ~0.7s on my machine (the 1.5s is already after many other improvements I landed).

Additionally, save game size goes to 860k (from 1.1MB), which is already after compression.
RAM use after GC goes to 129M from 192M.
These two are both after starting an all-human "270BC Wars" game at the start of the first human turn (after all the barbarian combats).

This change also cleans up some unnecessary null checks on attachment setters.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
@@ -1054,13 +1054,14 @@ class BattleCalculatorPanel extends JPanel {
     }
     setupAttackerAndDefender();
 
-    Instant t = Instant.now();
+    final Instant t = Instant.now();
     calculator =
         new ConcurrentBattleCalculator(
             () ->
                 SwingUtilities.invokeLater(
                     () -> {
-                      calculateButton.setText("Calculate Odds " + Duration.between(t, Instant.now()));
+                      log.debug("Battle Calculator ready in " + Duration.between(t, Instant.now()));
+                      calculateButton.setText("Calculate Odds");
                       calculateButton.setEnabled(true);
                       calculateButton.requestFocusInWindow();
                     }));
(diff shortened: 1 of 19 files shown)


asvitkine:
@RoiEXLab thoughts?

Another option is to use our own String set on GameData, instead of using Java's built in `.intern()`. This would be a bit more complex since it will require a new member on GameData and a helper function on DefaultAttachment.

Cernelius:
Oh that's great!

270BC Wars actually has 3 main problems.

1. The battlecalculator takes a lot of time to load the data (actually 6 seconds for me).
2. You cannot assign "barbarians" to "Does Nothing" in automated hosts.
3. Automated hosts will crash and not load the game any longer after it has been played for some rounds (I assume because the save-game becomes too big for them.).

asvitkine:
@RoiEXLab wondering what you think about this?

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
```diff
@@ -1052,11 +1054,13 @@ class BattleCalculatorPanel extends JPanel {
     }
     setupAttackerAndDefender();
 
+    final Instant t = Instant.now();
     calculator =
         new ConcurrentBattleCalculator(
             () ->
                 SwingUtilities.invokeLater(
                     () -> {
+                      log.debug("Battle Calculator ready in " + Duration.between(t, Instant.now()));
```
RoiEXLab, line 1063: Is this change intended or just a remainder of your local debugging?

RoiEXLab (approved):
No strong opinion on `intern()`. Your benchmark results sound promising. Just one minor comment 👍🏼

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
asvitkine: I opted to keep it, since it's just a log line that won't happen often (only when user brings up the battle calc), and let's use measure improvements in a consistent way.

PR #10472, opened 2022-05-19
Fix game history and add Options class specifying what to include.

asvitkine:
Fix game history and add Options class specifying what to include.

My previous change resulted in history being omitted in the game copy being used for history view. This change corrects that by introducing an Options class specifying what to include when cloning data.


asvitkine:
I'm going to submit to fix history on head.

PR #10473, opened 2022-05-20
Remove RelationshipTracker param from many matchers.

asvitkine:
Remove RelationshipTracker param from many matchers.

Now that all players have a GameData set, we can always get the relationship tracker from GamePlayer, thus allowing us to eliminate lots of params. This PR adds some helper methods on GamePlayer for checking common relationship status.

IntelliJ makes this type of refactoring pretty simple since it automatically updates the call sites!


asvitkine:
Looks like some tests likely don't set things up (i.e. having players reference game data that has relationshiptracker) for this to work out of the box. So I'll need to dig into those. Leaving this in Draft state for now.

RoiEXLab:
> IntelliJ makes this type of refactoring pretty simple since it automatically updates the call sites!

Do you know about the structural replace feature? It allows you to do semantic replacements rather than using something error-prone like regular expressions

RoiEXLab:
Ah i thought you just used the normal refactoring menu. AFAIK it allows you to remove parameters without much effort

asvitkine:
OK, all tests are passing now, except some pre-existing codeclimate issues.

PR #10474, opened 2022-05-20
Remove unneeded copies done by CollectionUtils.

asvitkine:

### DIFF
--- a/lib/java-extras/src/main/java/org/triplea/java/collections/CollectionUtils.java
+++ b/lib/java-extras/src/main/java/org/triplea/java/collections/CollectionUtils.java
@@ -69,19 +69,20 @@ public class CollectionUtils {
     checkArgument(max >= 0, "max must not be negative");
     checkNotNull(predicate);
 
-    return collection.stream()
-        .filter(predicate)
-        .limit(max)
-        .collect(Collectors.toList());
+    return collection.stream().filter(predicate).limit(max).collect(Collectors.toList());
   }
 
   /** return a such that a exists in c1 and a exists in c2. always returns a new collection. */
   public static <T> List<T> intersection(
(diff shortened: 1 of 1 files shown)


RoiEXLab (commented):
IIRC those defensive copies were added at some point due to concurrency issues. Not sure if it actually solved those or just made them less likely to occur, but I believe it's best to find the commit that introduced them and check if any linked issue can still occur with this change

asvitkine:
> IIRC those defensive copies were added at some point due to concurrency issues. Not sure if it actually solved those or just made them less likely to occur, but I believe it's best to find the commit that introduced them and check if any linked issue can still occur with this change

Thanks for the context, I will look. Ideally, we can fix the concurrency issues separately - it's unfortunate to require all the code to be inefficient due to some bugs. For example, I noticed this while adding a new call to one of these - I know my call is fine, but then I'd hate for us to have a `difference()` and `differenceWithoutCopies()` function.

asvitkine:
This was introduced here:

https://github.com/triplea-game/triplea/pull/5968

I'm going to try a few all-AI games to see if concurrent exceptions still occur with change. If things look good (and I can test on a number of maps), I think removing this is appropriate.

asvitkine:
I did 9 rounds of Global 1942 2nd ed w/ all Pro AIs and no exceptions. I'll try on a few other maps to make sure.

asvitkine:
> I did 9 rounds of Global 1942 2nd ed w/ all Pro AIs and no exceptions. I'll try on a few other maps to make sure.

Also did 3 rounds on Warcraft (which is a very complex and trigger-heavy map), also no concurrent exceptions. At this point, I'm confident we should make this change.

PR #10476, opened 2022-05-21
Teach ProAI how to buy units that consume other units.

asvitkine:

Teach ProAI how to buy units that consume other units.

This is done by not treating these units as 'special' buy options that are ignored and instead to consider them normally, except:
 - Track units that will be consumed.
 - Don't move such units during move phases.
 - When doing combat simulation that include the placed units, omit the ones that will be consumed.

Tested multiple all-AI rounds on the following maps:
  - Imperialism 1974
  - Civil War: A House Divided

Observed AI purchasing and successfully placing units that require consumption of other units. Quite a lot of these on the Civil War map!

I also checked a few other maps that have such units, such as TWW and "270BC Wars", these ran without errors but I didn't notice AI buying units that consume units. Of note, the AI still uses all its current logic to decide which units it should buy, but now can also consider units that consume others units too in the possible options.

## Release Note

NEW|Hard AI can now purchase and place units that consume other units.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProData.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProData.java
@@ -15,8 +15,10 @@ import games.strategy.triplea.delegate.Matches;
 import games.strategy.triplea.util.TuvUtils;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import javax.annotation.Nullable;
 import lombok.Getter;
 import org.triplea.java.collections.CollectionUtils;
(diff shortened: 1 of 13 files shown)


ron-murhammer:
Another good map to test is Total World War. I'll try to take a closer look at this in the next few days.

Cernelius:
A peculiar case (you may want to check) is "270BC Wars". Specifically, legionary units being upgraded to fort units (only for Rome and Syria). What makes this case peculiar is mainly the fact that a mobile (offensive) unit upgrades into an immobile (inoffensive) one.

asvitkine:
@ron-murhammer Would appreciate if you could take a look, as I want to do additional follow up changes dependent on this (e.g. non-combat moves to move consumable units to factories which is complementary to this logic).

asvitkine:
I've done a bunch of testing with this and am confident this is working well. I'll go ahead and merge, but happy to receive feedback if someone still wants to take a look at the code.

PR #10481, opened 2022-05-21
Introduce a cleaner pattern for locking GameData.

asvitkine:


RoiEXLab (approved):
I like this change. What would be even cleaner if is `Unlocker` class would be required to access `GameData`'s mutable resources in the first place, but obviously there's only so much you can do without breaking compatibility.

asvitkine:
Forgot to mention, one big advantage of this is code using this pattern won't be susceptible to the type of bug that we've seen in many places where something locks a gameData object but then tries to unlock a different one, due to it changing underneath on another thread.

RoiEXLab:
@asvitkine That's a really good point. It's really unsatisfying too see an unused variable though

PR #10483, opened 2022-05-22
Make ChangeAttachmentChanges save a copy of the previous value.

asvitkine:

Make ChangeAttachmentChanges save a copy of the previous value of the property.

The previous value is used when inverting the change (e.g. when going back in history and reverting changes). By copying the properties, this fixes several issues:
  - Ensuring that the inverted change is actually correct, as we don't want to be reverting to a "live view" of the current state.
  - Fixing exceptions caused by my recent change where property getters return unmodifiable collections (we don't want to set the property to an unmodifiable collection, since future changes will fail).

Additionally, switches to improved syntax for some locking in a few related places and some small clean ups.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/MutableProperty.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/MutableProperty.java
@@ -110,21 +110,6 @@ public final class MutableProperty<T> {
     return getter.get();
   }
 
-  @SuppressWarnings({"unchecked", "rawtypes"})
-  public T getValueCopy() {
-    T value = getter.get();
-    if (value instanceof List) {
-      return (T) new ArrayList((List) value);
-    } else if (value instanceof IntegerMap) {
-      return (T) new IntegerMap((IntegerMap) value);
-    } else if (value instanceof Set) {
-      return (T) new HashSet((Set) value);
-    } else if (value instanceof Map) {
-      return (T) new HashMap((Map) value);
-    }
-    return value;
-  }
-
   public void resetValue() {
     resetter.run();
   }
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/MutableProperty.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/MutableProperty.java
@@ -2,8 +2,15 @@ package games.strategy.engine.data;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.function.Supplier;
 import javax.annotation.Nullable;
+import org.triplea.java.collections.IntegerMap;
 import org.triplea.java.function.ThrowingConsumer;
 import org.triplea.java.function.ThrowingFunction;
 
@@ -101,6 +108,21 @@ public final class MutableProperty<T> {
 
   public T getValue() {
     return getter.get();
+  }
+
+  @SuppressWarnings({"unchecked", "rawtypes"})
+  public T getValueCopy() {
+    T value = getter.get();
+    if (value instanceof List) {
+      return (T) new ArrayList((List) value);
+    } else if (value instanceof IntegerMap) {
+      return (T) new IntegerMap((IntegerMap) value);
+    } else if (value instanceof Set) {
+      return (T) new HashSet((Set) value);
+    } else if (value instanceof Map) {
+      return (T) new HashMap((Map) value);
+    }
+    return value;
   }
 
   public void resetValue() {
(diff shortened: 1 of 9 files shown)


game-app/game-core/src/main/java/games/strategy/engine/data/MutableProperty.java
```diff
@@ -103,6 +110,21 @@ public T getValue() {
...
+      return (T) new ArrayList((List) value);
+    } else if (value instanceof IntegerMap) {
+      return (T) new IntegerMap((IntegerMap) value);
+    } else if (value instanceof Set) {
+      return (T) new HashSet((Set) value);
+    } else if (value instanceof Map) {
+      return (T) new HashMap((Map) value);
+    }
+    return value;
+  }
```
asvitkine, line 126: I know this is a bit messy, and technically not full proof.

For example, what if the getter is some other type? We don't know how to arbitrarily copy those.
Or if the contents of the collection is itself a mutable object (I've not seen this to be the case with any attachments I modified, but theoretically possible).

We don't want to do just a clone() because if the value is an unmodifiable collection, the copy will be that too.

Though... this is technically solving two distinct problems:
  1. We don't want to point to a "live" value from a Change. So we definitely need a copy.
  2. We don't want to set the actual property on the attachment to an unmodifiable collection.

I guess maybe we could handle these two separately, via:
  1. We could just call "clone()" without needing to inspect the type.
  2. Have all collection setters copy the params themselves. So we'll need to change all of those, but it's not the first time I'm editing all these properties.

@RoiEXLab WDYT?

asvitkine: Actually, we can't use "clone()" because it's protected on Object and Cloneable interface doesn't define it either!

So no way to clone() without knowing the type.

RoiEXLab: I'm 100% convinced this method should **_not_** be part of `MutableProperty` in any shape or form. This class is supposed to bundle a bunch of setters and getters together nothing else.

Regardless what do I think about this?
Honestly, this might be a legitimate use-case for reflection.
1. Get an object's class
2. check if it has a "copy-constructor" and
3. call that with the object.
4. Return the new object, or throw an exception if any of  the steps fails

Also maybe add a comment to `IntegerMap`'s copy constructor that it might be called by reflection, to prevent accidental removal (or verify this behaviour in a unit test, that works too of course)

asvitkine: I don't think the reflective approach will solve the problem fully, since the problem is if the instance is an immutable collection, we want to construct a non-immutable collection.

Given we have one call site, one option is we just move this to ChangeAttachmentChange or perhaps a static on DefaultAttachment that ChangeAttachmentChange calls (e.g. copyPropertyValue() on the result of the getter).

DefaultAttachment could be a reasonable place for this given it already has helpers for dealing with these types (i.e. which is actually where the immutable views of collections come from, such as getIntegerMapProperty(), getListProperty(), etc). WDYT?

RoiEXLab: Good point about the mutability. I really wonder though: Where is the object cast back to its original type? Could the copy be created there? Your ideas sound good nevertheless 👍

So feel free to trust your gut or play around with it for a bit. I don't think there's a definite wrong answer, just many options to weigh against each other

asvitkine: For collections, the type is the interface type (i.e. List not ArrayList), since that's the type of the property. So it doesn't get cast to ArrayList, we just assign it fine to List and the problem only happens if the property is modified again and we try to add to it (since an unmodifiable List is still a List, just its mutators throw exceptions).

No casting needed on the getter side, since MutableProperty is already generic on the type.

OK, I'll go with the approach of adding a `copyPropertyValue()` to `DefaultAttachment`.

game-app/game-core/src/main/java/games/strategy/engine/framework/HistorySynchronizer.java
```diff
@@ -27,15 +26,9 @@ public class HistorySynchronizer {
         public void gameDataChanged(final Change change) {
           SwingUtilities.invokeLater(
               () -> {
-                try {
+                try (GameData.Unlocker unused = gameData.acquireWriteLock()) {
```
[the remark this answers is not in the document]
asvitkine: I'm going to revert this part of the change. I don't think it's needed.

The reason this is on the UI thread is that the GameData that's copied for use with history is marked as "UI thread only", so presumably it's only allowed to be accessed from that thread. If all the code dealing with that GameData follows that, then I guess we don't need locking on this gameData.

asvitkine:
I'll go ahead and submit to unblock a follow up change I have, since @RoiEXLab said the he was ok with the approach I mentioned in the comment thread.

PR #10500, opened 2022-05-25
Added Medieval Europe

RogerCooper:

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2646,3 +2646,11 @@
    <br>All these scenarios use the WW2v3 map & rules
    <br>WW2v3-1941-Buildcap & WW2v3-1942-Buildcap limit the number of units of each that can be built
    <br>WW2v3-1941-Upkeep & WW2v3-1942-Upkeep multiplies costs & income by 10, and applies a 10%/turn upkeep cost
+- mapName: Medieval Europe
+  mapCategory: EXPERIMENTAL
+  url: https://github.com/triplea-maps/Medieval_Europe/archive/master.zip
+  version: 1
+  img: https://raw.githubusercontent.com/triplea-maps/Medieval_Europe/master/preview.png
+  description: |
+   <br>The kingdoms of medieval Europe, grouped into three alliances
+   


Cernelius:
I think the name of this map is too generic. How about if someone else wants to make a game set in Medieval Europe? How about if someone would make a map called just "World War 2"? I don't think that would be acceptable, as it would be confusing will all other maps set in WW2.

I think the name of the map should be at least canged to "Medieval Europe X", where "X" is one or more words or numbers. For example, you could add the starting year of the map or the name of the map maker.

RogerCooper:
I agree that the name is bland and generic. However, I usually don't change the name of the scenarios I post from my collection unless it similar to an existing name. No one complained about Ancient Times.

Cernelius:
> I agree that the name is bland and generic. However, I usually don't change the name of the scenarios I post from my collection unless it similar to an existing name. No one complained about Ancient Times.

Ancient Times is fine because it is creative. Medieval Times would be fine too.
I'm just referring to usual period classifications, like "World War 2" or "Ancient Europe".
I'm also against names like "Great War" (which are in the repository).

RogerCooper:
I don't like "World War anything" because so many scenarios start that way.

asvitkine:
Both this and your other map PR have formatting issues - and make the "Check Formatting" check fail. Can you fix those?

RogerCooper:
I had resubmitted a corrected version.

asvitkine:
> I had resubmitted a corrected version.

I don't see that?

RogerCooper:
I just tried again, but I could not do the commit, possibly because I had just updated another mod. I will try again soon.

beelee1:
_I just tried again, but I could not do the commit, possibly because I had just updated another mod. I will try again soon._
@RogerCooper yea it looked like both PRs were on the same line in the Yaml. I think line 2656 is the extra line that throws the error. You could try deleting it anyway.

Then "Hot War 1956" has the same line(s) for it, so maybe as you said, that's the conflict.
Screenshot from 2022-05-31 22-20-49

Screenshot from 2022-05-31 22-24-40

PR #10522, opened 2022-05-28
Add an end-to-end test that executes an all-AI game to completion.

asvitkine:
Add an end-to-end test that executes an all-AI (hard AI) game to completion.

Currently, only doing this on the very simple "mapmaking tutorial map", which exercises sea zones and amphib invasions. In my testing, this takes 1 min to run and a game completes in 24 rounds.

Also adds a few more maps to GameSaveTest.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/framework/ServerGame.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/framework/ServerGame.java
@@ -84,6 +84,7 @@ public class ServerGame extends AbstractGame {
   /** Has the delegate signaled that delegate execution should stop. */
   private volatile boolean delegateExecutionStopped = false;
 
+  // If true, setting delegateExecutionStopped to true will stop the game (return from startGame()).
   @Setter private boolean stopGameOnDelegateExecutionStop = false;
 
   public ServerGame(
@@ -424,7 +425,8 @@ public class ServerGame extends AbstractGame {
       return;
     }
     // save after the step has advanced otherwise, the delegate will execute again.
-    if (shouldAutoSaveAfterEnd(currentDelegate) && currentStep.getName().endsWith("Move")) {
+    boolean isMoveStep = GameStep.isMoveStep(currentStep.getName());
+    if (isMoveStep && shouldAutoSaveAfterEnd(currentDelegate)) {
       final String stepName = currentStep.getName();
       // If we are headless we don't want to include the nation in the save game because that would
       // make it too difficult to load later.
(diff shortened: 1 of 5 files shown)


game-app/game-core/src/main/java/games/strategy/engine/framework/ServerGame.java
```diff
@@ -441,11 +441,29 @@ private void runStep(final boolean stepIsRestoredFromSavedGame) {
               ? launchAction.getAutoSaveFileUtils().getEvenRoundAutoSaveFile()
               : launchAction.getAutoSaveFileUtils().getOddRoundAutoSaveFile());
     }
-    if (autoSaveThisDelegate && !currentStep.getName().endsWith("Move")) {
+    if (shouldAutoSaveAfterEnd(currentDelegate) && !currentStep.getName().endsWith("Move")) {
```
RoiEXLab, line 444: Probably out of scope, but didn't we have convenience methods for combat and non-combat moves here?
Could we introduce them for "Move" as well to prevent potential typos?

game-app/game-core/src/main/java/games/strategy/engine/framework/ServerGame.java
```diff
@@ -295,9 +300,13 @@ public void startGame() {
       }
       while (!isGameOver) {
         if (delegateExecutionStopped) {
-          // the delegate has told us to stop stepping through game steps
-          // don't let this method return, as this method returning signals that the game is over.
-          Interruptibles.await(delegateExecutionStoppedLatch);
+          if (stopGameOnDelegateExecutionStop) {
+            stopGame();
```
RoiEXLab, line 304: I don't quite understand this change?
Is this for testing only? If yes, could this be documented?

RoiEXLab (approved):

game-app/game-core/src/main/java/games/strategy/engine/framework/ServerGame.java
asvitkine: The test is the only thing that sets this. It's needed so this loop doesn't execute forever doing `Interruptibles.await(delegateExecutionStoppedLatch);`.

I'm not sure how the regular codepath exits this loop - probably by calling stopGame() from a different thread - but that's not something that can be done by the test, since the test body is single-threaded.

game-app/game-core/src/main/java/games/strategy/engine/framework/ServerGame.java
asvitkine: We do, done!

game-app/game-core/src/main/java/games/strategy/engine/framework/ServerGame.java
asvitkine: Added a comment.

PR #10536, opened 2022-05-30
Move headed code to headed subproject

RoiEXLab:
Some more code moved to the headed subproject.

When reviewing, please have a look at 66681e0437ba84469728a1f17c7213789d8d6af1 .
I'm 99% sure this is safe and haven't been able to prove the opposite, but there might be a small chance where `getService(ApplicationContext.class)` is called by a headless instance  for some weird reason


asvitkine:
> there might be a small chance where `getService(ApplicationContext.class)` is called by a headless instance for some weird reason

I think the relevant call is `Services.loadAny(ApplicationContext.class)`, which there are 3 call sites for. Did you check them all?

In particular, I recall when I was setting up some smoke tests, the one in ClientFileSystemHelper.java was being hit until I made the test code call `ClientFileSystemHelper.setCodeSourceFolder()`.

So, do you know why the above *wouldn't* be called in headless mode? I would assume it does...

RoiEXLab:
> I think the relevant call is `Services.loadAny(ApplicationContext.class)`, which there are 3 call sites for. Did you check them all?

After this PR there's only a single call site left. I observed the call tree a couple layers down and I'm pretty sure all of them are for asset loading, which should not occur in the headless environment

asvitkine:
> > I think the relevant call is `Services.loadAny(ApplicationContext.class)`, which there are 3 call sites for. Did you check them all?
> 
> After this PR there's only a single call site left. I observed the call tree a couple layers down and I'm pretty sure all of them are for asset loading, which should not occur in the headless environment

Are you sure headless doesn't load any assets? If I comment out `ClientFileSystemHelper.setCodeSourceFolder(tempFolder)` in `GameTestUtils.java` and run `GameSaveTest` it fails with:

```
Unable to find game assets folder starting from location: <snip>/triplea-1/game-app/game-headless
There is a problem with the installation, please report this to TripleA and the path where TripleA is installed.
games.strategy.triplea.ResourceLoader$GameAssetsNotFoundException: Unable to find game assets folder starting from location: <snip>/triplea-1/game-app/game-headless
There is a problem with the installation, please report this to TripleA and the path where TripleA is installed.
	at java.base@12.0.2/java.util.Optional.orElseThrow(Optional.java:408)
	at app//games.strategy.triplea.ResourceLoader.<init>(ResourceLoader.java:53)
	at app//games.strategy.triplea.ResourceLoader.<init>(ResourceLoader.java:44)
	at app//games.strategy.triplea.ui.UiContext.setResourceLoader(UiContext.java:97)
	at app//org.triplea.game.server.HeadlessLaunchAction.startGame(HeadlessLaunchAction.java:71)
	at app//games.strategy.triplea.TripleA.startGame(TripleA.java:62)
	at app//games.strategy.engine.data.GameTestUtils.setUpGameWithAis(GameTestUtils.java:91)
	at app//games.strategy.engine.data.GameSaveTest.testSaveGame(GameSaveTest.java:42)
	at java.base@12.0.2/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```

So the above has HeadlessLaunchAction in the call stack, but still hits the ResourceLoader code path that needs this.

RoiEXLab:
@asvitkine Good observation. I'll leave it in in this case

asvitkine (approved):

RoiEXLab:
> Additionally, I wonder if the whole `Services.loadAny(ApplicationContext.class)` mechanism is even needed. I feel like we should have the calling code explicitly configure things like paths, as opposed to inner code get "global" information like this.

It's band aid for places where we can't properly pull the implementation into the game-headed/game-headless projects respectively. So yes, ideally it should be removed but this doesn't work as easily in this case.

PR #10544, opened 2022-05-31
Remove HeadlessGameServer singleton

RoiEXLab:
I finally found out how to register the `ChatAppender` outside of the `logback.xml`. There's no longer a need for a singleton.


asvitkine (approved):

game-app/game-headless/src/main/java/org/triplea/game/server/debug/ChatAppender.java
```diff
@@ -2,37 +2,28 @@
...
  * game server to report its own logs to other game clients via the chat window.
  */
 public final class ChatAppender extends AppenderBase<ILoggingEvent> {
-  private boolean enabled = true;
+  private final Chat chat;
 
-  @Override
-  protected void append(final ILoggingEvent record) {
-    // guard against infinite recursion if sendChatMessage also logs
-    if (enabled) {
```
RoiEXLab, line 19: For reference:
This guard is removed because logback's `AppenderBase` class already does this

PR #10554, opened 2022-06-01
Speed up node selection in history tree.

asvitkine:
Speed up node selection in history tree.

Turns out, most of the time was going to repeatedly redrawing the map. This change batches operations.

With this change, using the save game on https://github.com/triplea-game/triplea/pull/10551, selecting the first turn in history goes from 1.2s execution time to 0.3s.

## Release Note

CHANGE|Selecting nodes in the history tree has been sped up

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
@@ -2,6 +2,7 @@ package games.strategy.triplea.ui.panels.map;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.google.common.collect.Sets;
 import games.strategy.engine.data.Change;
 import games.strategy.engine.data.ChangeAttachmentChange;
 import games.strategy.engine.data.CompositeChange;
@@ -77,6 +78,7 @@ import org.triplea.java.Interruptibles;
 import org.triplea.java.ObjectUtils;
 import org.triplea.java.ThreadRunner;
 import org.triplea.java.collections.CollectionUtils;
+import org.triplea.java.concurrency.AsyncRunner;
 import org.triplea.util.Tuple;
 
 /** Responsible for drawing the large map and keeping it updated. */
@@ -106,26 +108,24 @@ public class MapPanel extends ImageScrollerLargeView {
   @Getter private Collection<Collection<Unit>> highlightedUnits = List.of();
   private Cursor hiddenCursor = null;
   private final MapRouteDrawer routeDrawer;
+  private Set<Territory> countriesToUpdate = new HashSet<>();
 
   private final TerritoryListener territoryListener =
       new TerritoryListener() {
         @Override
         public void unitsChanged(final Territory territory) {
           updateCountries(Set.of(territory));
-          SwingUtilities.invokeLater(MapPanel.this::repaint);
         }
 
         @Override
         public void ownerChanged(final Territory territory) {
           smallMapImageManager.updateTerritoryOwner(territory, gameData, uiContext.getMapData());
           updateCountries(Set.of(territory));
-          SwingUtilities.invokeLater(MapPanel.this::repaint);
         }
 
         @Override
         public void attachmentChanged(final Territory territory) {
           updateCountries(Set.of(territory));
-          SwingUtilities.invokeLater(MapPanel.this::repaint);
         }
       };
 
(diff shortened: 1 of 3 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
```diff
@@ -584,14 +583,36 @@ private boolean unitsChanged(final Tuple<Territory, List<Unit>> newUnits) {
...
+              }
+              tileManager.updateTerritories(toUpdate, gameData, uiContext.getMapData());
+              smallMapImageManager.update(uiContext.getMapData());
+              SwingUtilities.invokeLater(
+                  () -> {
+                    smallView.repaint();
+                    repaint();
+                  });
+            })
+        .exceptionally(Throwable::printStackTrace);
```
RoiEXLab, line 615: Using a logger here would be more appropriate

game-app/game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
```diff
@@ -584,14 +583,36 @@ private boolean unitsChanged(final Tuple<Territory, List<Unit>> newUnits) {
...
+      return; // An update is already scheduled.
+    }
+    AsyncRunner.runAsync(
+            () -> {
+              Collection<Territory> toUpdate;
+              synchronized (countriesToUpdate) {
+                // Note: Don't run updateTerritories() inside countriesToUpdate lock, as this causes
+                // a deadlock due to locking game data inside updateTerritories().
+                toUpdate = List.copyOf(countriesToUpdate);
+                countriesToUpdate.clear();
```
RoiEXLab, line 605: If you lock on some other object, you could just assign countriesToUpdate to the `toUpdate` variable and assign it a new instance of a fresh set implementation. This would avoid a copy operation.

RoiEXLab (approved):

RoiEXLab:
> This change batches operations.

I was kinda confused by this statement.
From what I can tell it simply aggregates lists until the first thread picks the task up and then schedules it onto the EDT which might take some additional time where potential changes could be queued up but not picked up. Ideally it would be the same thread, but holding locks on the EDT is bad practice, might lead to a freezing UI. I wonder if there's a way to design this somehow so the EDT just picks up the latest list without the need for a lock and the "pushing" thread needs to do all the waiting if the EDT is currently processing this item.

game-app/game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
asvitkine: Done.

game-app/game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
asvitkine: Done.

asvitkine:
> > This change batches operations.
> 
> I was kinda confused by this statement. From what I can tell it simply aggregates lists until the first thread picks the task up and then schedules it onto the EDT which might take some additional time where potential changes could be queued up but not picked up. 

I think "batching" is a reasonable way to describe that, although it's true that it also can reduce the total amount of work too (if the same country is updated before the async task got to run). But at the very least, it creates a batch of countries to update from events for individual countries.

Small clarification: I think it doesn't schedule it onto the EDT - the operation will be done on an async thread, but then repaint() will be done on EDT via SwingUtilities.invokeLater().

> Ideally it would be the same thread, but holding locks on the EDT is bad practice, might lead to a freezing UI. I wonder if there's a way to design this somehow so the EDT just picks up the latest list without the need for a lock and the "pushing" thread needs to do all the waiting if the EDT is currently processing this item.

I think there's a more fundamental issue with the history view design - it's set up such that manipulating the GameData copy is all done on the EDT. So even ignoring the locking here (which should have very little contention), all the history game data updates happen on the EDT. That means that if you take a long game and select a node in history far back (e.g. in my example, 30 turns was 16000 changes), all those changes get applied to the GameData on the EDT.

Now, even 16000 changes took only 0.3s after this optimization - and I think a good of that is still another UI update like here, that can be improved. But still, the rest is just computation that we're still doing on EDT. And it's potentially unbounded for a really long game with a lot of changes happening.

So yeah, TL;DR we'd need to redesign the history system first if we want to ensure EDT is very responsive. On the other hand, the current system can be made fast enough with changes like this.

PR #10556, opened 2022-06-01
Fix neutral logic causing AI to ignore safe attacks on Imperialism 1974.

asvitkine:

Fix neutral logic causing AI to ignore safe attacks on Imperialism 1974.

The AI logic for "Removing neutral territory that has attackers that are adjacent to enemies" had a consistency issue where it used two different ways to check for neutrality when considering "is the territory being attacked neutral" and "are there non-neutral adjacent territories?".

The result was that the AI was very hesitant to attack neutrals if multiple neutrals are adjacent (whereas it was meant to be only if enemies are adjacent). This was seen on AI moves on Imperialism 1974, where the AI would take many turns to build up a lot of force to start attacking Neutral surrounding territory, even when no enemies were close.

Tested:
  1. Start a game with two Hard AI players on Imperialism 1974, all other players unchecked.
  2. Assuming they start non-adjacent, observe them making attacks on the first round.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
@@ -37,6 +37,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import org.triplea.java.collections.CollectionUtils;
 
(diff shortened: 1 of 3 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProCombatMoveAi.java
```diff
@@ -613,20 +615,20 @@ private void removeTerritoriesThatArentWorthAttacking(
...
-            attackersHaveEnemyNeighbors = true;
-            attackFromTerritoryWithEnemyNeighbors = attackFromTerritory;
-            break;
-          }
-        }
-        if (attackersHaveEnemyNeighbors) {
+        // Note: Use territoryIsEnemyNotNeutralLand(), not territoryIsEnemyNotPassiveNeutralLand()
+        // so that the neutrality check is consistent with logic for isNeutral.
+        Predicate<Territory> enemyTerritory = ProMatches.territoryIsEnemyNotNeutralLand(player);
+        Territory attackFromTerritoryWithEnemyNeighbors =
```
asvitkine, line 621: This is the same logic as before, just using stream() API. The functional change here is the predicate used above, see Note: comment.

PR #10558, opened 2022-06-01
Adding Medieval Europe

RogerCooper:


asvitkine:
I edited your PR in GitHub to try to fix the issue, let's see if it works.

RogerCooper:
That empty line at the end has been required on my previous uploads. I don't know why it is prohibited now.

asvitkine:
I think in this case there were two empty lines actually.

PR #10564, opened 2022-06-02
Hard AI: Don't send units that can't participate in combat on attacks.

asvitkine:
Hard AI: Don't send units that can't participate in combat on attacks.

This is done by moving the logic for combat participation to a new matcher which is then re-used both from the battle code and by Pro AI when deciding attack options.

With this change, AI playing Imperialism 1974 won't send wealth units along with attacks. This paves way for a future change of mine about moving consumable units towards factories.

One other benefit of this change is reducing the number of collection copies when deciding which units can participate in a battle.


asvitkine:
I am going to land this after doing a self-review, so it can unblock my follow up changes. But happy to discuss if anyone has feedback or concerns about this!

PR #10573, opened 2022-06-03
Stop using hardcoded territory name prefixes to handle islands.

asvitkine:
Stop using hardcoded territory name prefixes to handle islands.

The engine checked for "Sea Zone" as a prefix or suffix in the territory name to determine which territory should be considered "on top" to handle islands. This didn't work for maps where the sea zones didn't following the naming convention and also didn't support land territories fully surrounding other land territories.

This PR updates the logic to be based on the size of the polygon being matched instead, with the smallest polygon taking precedence.

The old logic is moved to mapmaking tool code only and no longer the main engine code.

Tested:
  - Mouse-overing Corsica on "Cold War" map works correctly.

## Release Note

CHANGE|Sea zones no longer need to have special names for islands to work correctly

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/mapdata/IslandTerritoryFinder.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/mapdata/IslandTerritoryFinder.java
@@ -1,6 +1,5 @@
 package games.strategy.triplea.ui.mapdata;
 
-import games.strategy.ui.Util;
 import java.awt.Polygon;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


Cernelius:
I support, but there is the possible issue of maps having used the name-based prioritization in creative (or hacky) ways, for example to hide a zone under another one. However, I cannot think of any cases.

---

The problem that the auto-placement runs over "islands" remains, though.

Cernelius:
An other problem which this change won't fix is when the enclaved zone is not a zone but the absence thereof.
For example, cases like the "Aral sea" of World War II Classic (which has been worked around by having a border from it all the way through the Kazakh S.S.R. territory) will remain insoluble beside actually adding and impassable zone for the non-zone which is within the zone.

asvitkine:
> An other problem which this change won't fix is when the enclaved zone is not a zone but the absence thereof. For example, cases like the "Aral sea" of World War II Classic (which has been worked around by having a border from it all the way through the Kazakh S.S.R. territory) will remain insoluble beside actually adding and impassable zone for the non-zone which is within the zone.

Right, for that you need to support "exclude polygons". The current "Aral sea" approach isn't broken by this, but it still requires the surrounding land territory to have a line going to the inner territory.

PR #10575, opened 2022-06-04
Speed up map parsing.

asvitkine:
Speed up map parsing.

With these changes, loading Imperialism 1974 xml file goes from 1s to 400ms on my machine.


Cernelius:
"War of the Relics" and "270BC Wars" are two other games which may be interesting to test (being slow loading).

I've also noticed that save-games load much faster than game file (xml). Does this change effect loading save-games speed at all?

asvitkine:
> I've also noticed that save-games load much faster than game file (xml). Does this change effect loading save-games speed at all?

This change doesn't affect save games. My other changes that reduced the size of saved games would have improved that (for games saving on the new versions).

PR #10593, opened 2022-06-06
Cleanup: UnitAttachment.get(u.getType()) -> u.getUnitAttachment().

asvitkine:
Additionally, address some codeclimate warnings and change a couple of call sites that call isTranporting() followed by getTransporting(), which do the same work and are very inefficient.


asvitkine:
codeclimate issues are pre-existing.

asvitkine:
Did a self-review, landing.

PR #10607, opened 2022-06-08
Refactoring TechTracker in preparation for caching techs.

asvitkine:
Refactoring TechTracker in preparation for caching techs.

This a precursor PR for allowing TechTracker to cache technology bonuses. When the cache is enabled, I've observed 26% improvement in average runtime of MustFightBattle.fight() on Domination 1914 (across hundreds of thousands of battles done by the AI - the improvement varies per map). The improvement is so drastic because a lot of work was being done summing tech bonuses just for getting the strength of one unit, for example.

This PR does not yet enable the caching, since the important thing to get right is when the cache should be invalidated, which I want to be careful about and do more testing and code inspection. However, it sets everything up such that the caching can be done in a minimal follow-up PR without needing refactoring.

Some code is moved from TechAttachment to TechTracker as well as the corresponding unit tests and various callers are updated to take different params so that they can use an instance of TechTracker to lookup tech bonuses.

No functional changes. Some small perf improvements due to code restructuring (e.g. moving computations outside of loops).

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/RocketsFireHelper.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/RocketsFireHelper.java
@@ -273,7 +273,7 @@ public class RocketsFireHelper implements Serializable {
     final int numberOfAttacks;
     // attackFrom could be null if WW2V1
     if (attackFrom == null) {
-      rockets = new ArrayList<>();
+      rockets = List.of();
       numberOfAttacks = 1;
     } else {
       rockets = CollectionUtils.getMatches(attackFrom.getUnits(), rocketMatch(player));
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/RocketsFireHelper.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/RocketsFireHelper.java
@@ -15,7 +15,6 @@ import games.strategy.engine.player.Player;
 import games.strategy.engine.random.IRandomStats.DiceType;
 import games.strategy.triplea.Constants;
 import games.strategy.triplea.Properties;
-import games.strategy.triplea.attachments.TechAbilityAttachment;
 import games.strategy.triplea.attachments.TerritoryAttachment;
 import games.strategy.triplea.attachments.UnitAttachment;
 import games.strategy.triplea.delegate.battle.BattleTracker;
@@ -102,11 +101,9 @@ public class RocketsFireHelper implements Serializable {
     final GameData data = bridge.getData();
     final GamePlayer player = bridge.getGamePlayer();
     final Map<Territory, Integer> previouslyAttackedTerritories = new LinkedHashMap<>();
+    final int maxAttacks = data.getTechTracker().getRocketNumberPerTerritory(player);
     for (final Territory attackFrom : getTerritoriesWithRockets(data, player)) {
       final Set<Territory> targets = getTargetsWithinRange(attackFrom, data, player);
-      final int maxAttacks =
-          TechAbilityAttachment.getRocketNumberPerTerritory(
-              TechTracker.getCurrentTechAdvances(player, data.getTechnologyFrontier()));
       for (final Territory t : previouslyAttackedTerritories.keySet()) {
         // negative Rocket Number per Territory == unlimited
         if (maxAttacks >= 0 && maxAttacks <= previouslyAttackedTerritories.get(t)) {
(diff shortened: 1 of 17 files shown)


asvitkine:
@RoiEXLab please take a look when you have the chance.

asvitkine:
@RoiEXLab friendly ping to take a look when you have a chance. The codeclimate "error" is a TODO I added that I plan to address in a follow up PR. I can always remove that word from the comment to make it happy, lmk.

RoiEXLab:
@asvitkine Sorry for the delay, I'll have a look now. Thanks for your patience :D

game-app/game-core/src/main/java/games/strategy/triplea/delegate/TechTracker.java
```diff
@@ -11,73 +12,204 @@
...
+    return techAdvances.stream()
+        .map(TechAbilityAttachment::get)
+        .filter(Objects::nonNull)
+        .map(mapper)
+        .mapToInt(m -> m.getInt(ut))
+        .sum();
   }
 
-  private boolean getUnitAbilitiesGained(
+  private int getUnitAbilitiesGained(
```
RoiEXLab, line 187: Why not keep it a boolean? Wouldn't it be nicer to just overload the `getCached` method instead?

game-app/game-core/src/main/java/games/strategy/triplea/delegate/battle/StrategicBombingRaidBattle.java
```diff
@@ -882,14 +882,9 @@ private void findCost(final IDelegateBridge bridge) {
...
-            Math.max(
-                0,
-                (costThisUnit
-                    + TechAbilityAttachment.getBombingBonus(
-                        attacker.getType(),
-                        TechTracker.getCurrentTechAdvances(
-                            attacker.getOwner(), gameData.getTechnologyFrontier()))));
+        final int bonus =
+            gameData.getTechTracker().getBombingBonus(attacker.getOwner(), attacker.getType());
+        costThisUnit = Math.max(0, (costThisUnit + bonus));
```
RoiEXLab, line 887: Ignoring integer overflows, can `costThisUnit` ever be assigned a negative value if the `Math#max` part is omitted?

RoiEXLab (approved):
👍🏼

game-app/game-core/src/main/java/games/strategy/triplea/delegate/TechTracker.java
asvitkine: The map value type is Integer, so we'd either need two maps or have the type be Object and have getCached() cast it. WDYT?

game-app/game-core/src/main/java/games/strategy/triplea/delegate/RocketsFireHelper.java
```diff
@@ -283,13 +276,13 @@ private void fireRocket(
       rockets = new ArrayList<>();
       numberOfAttacks = 1;
     } else {
-      rockets =
-          new ArrayList<>(CollectionUtils.getMatches(attackFrom.getUnits(), rocketMatch(player)));
+      rockets = CollectionUtils.getMatches(attackFrom.getUnits(), rocketMatch(player));
```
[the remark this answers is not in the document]
asvitkine: Yes, the list is never modified.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/battle/StrategicBombingRaidBattle.java
asvitkine: Yes, it can, since the bonuses from tech are just integers so they can be negative.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/RocketsFireHelper.java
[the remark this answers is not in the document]
RoiEXLab: In this case can you use List.of() instead of new arraylist a couple of lines above?

game-app/game-core/src/main/java/games/strategy/triplea/delegate/battle/StrategicBombingRaidBattle.java
RoiEXLab: I just feel like"bonus" sounds like a value that's always positive so I wondered

game-app/game-core/src/main/java/games/strategy/triplea/delegate/RocketsFireHelper.java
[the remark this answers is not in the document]
asvitkine: Done.

PR #10617, opened 2022-06-09
Improve blitz validation & add property for attacking from contested territories.

asvitkine:
Improve blitz validation & add property for attacking from contested territories.

This change makes several improvements to combat move validation, including handling combinations of blitzing and air/land transported units correctly and improving error messages related to moving from contested territories. It also adds a property to allow maps to permit attacking from contested territories unrelated to blitzing.

More specifically, the following improvements are done:
  1. When moving with a combination of blitzing, air transports, land transport units and non-blitzing units, the engine will now correctly determine which units are allowed to move. Previously, it would permit any land transportable units with enough movement to tag along in the presence of just one blitzing land transport and similar for air transports, now this is validated explicitly. The following saved game on Pacific Challenge with infantry modified to have 2 movement can be used to test. Prior to this change, including on 2.5, you can take a land transport and 6 infantry units and blitz with them, for example, which is not correct (and same for air transport).
[blitz_test_inf2.tsvg.zip](https://github.com/triplea-game/triplea/files/8882837/blitz_test_inf2.tsvg.zip)
  2. If units start their turn in a contested land zone (e.g. on a map like Civil War with limited combat rounds or via trigger), a better error message will be displayed for why they can't attack from the territory ("Cannot attack out of a contested territory"). Previously, the error message was about blitzing, which was incorrect.
  3. A new map property is added to allow attacks out of contested territories, defaulting to false, so maps that may want that behavior can set that. Maps like Imperialism 1974 and AA 1914 will want to set this, although AA 1914 may require further refinement with additional properties since the rules about leaving contested territories are more nuanced.

Tested:
  - Various blitz scenarios, including using land and air transports, including using an edited game state on Pacific Challenge map with land and air transports and editing infantry to have 2 movement (see save file above).

## Release Note

FIX|Improved validation of blitzes in the presence of land and air transports and better move error messages


asvitkine:
@Cernelius @panther2 

I'd like to get your thoughts as rule experts if this is the right thing to do. One option is to make it a map property - but that would still result in the wrong error message for current maps.

asvitkine:
Hmm, I took a look at the A&A 1914 rules, which also use 1-turn combat.

> Land units that begin the turn in contested territories can only be moved to territories that are controlled by your power, or to territories that are also contested and already contain units belonging to your power. If they are moved by transport, they may also remain at sea.

So, that's not the same logic as this. That requires checking if the destination territory is contested. Though, implementing A&A 1914 rules is its own topic I suppose, but if we want to implement them, some additional properties would need to be added to control this logic.

Cernelius:
> Hmm, I took a look at the A&A 1914 rules, which also use 1-round combat.
> 
> > Land units that begin the turn in contested territories can only be moved to territories that are controlled by your power, or to territories that are also contested and already contain units belonging to your power. If they are moved by transport, they may also remain at sea.
> 
> So, that's not the same logic as this. That requires checking if the destination territory is contested. Though, implementing A&A 1914 rules is its own topic I suppose, but if we want to implement them, some additional properties would need to be added to control this logic.

Are you saying that, under these rules, you can move directly from a contested territory to an enemy owned territory with enemy units in it as long as you have one or more units in it too, yet you cannot move from a contested territory to an empty enemy owned territory? If so, I would say this makes no sense: it should be easier to move into an empty enemy owned territory than to move into an enemy onwed territory crowded by a million enemy units and one of your units...

How about land units which begin the turn in enemy territories devoid of enemy units (in case there is no taking control of such territories immediately)? Are they treated exactly like those starting in contested territories?

Are there any differences between starting in a contested territory which you own and starting in a contested territory which your enemy owns, or ownership just doesn't matter so long as the territory is contested?

My suggestion here is having a few properties, to define what you can do and keeping the default behaviour as simple and generic as possible (documenting it in "PoS2"). This should allow map-makers to choose what they want and would have the benefit of telling everyone the behaviour of the game, to some extent.
For example, having a property like "Units can move between contested zones". If this property is false, a unit starting in a contested territory can never move into an enemy owned territory, whereas it can if this property is true and the enemy territory is contested.

---

In 2017, I've tested the program and written the following explanations as how it works:

>If you start your turn into a hostile land territory, you are allowed to go out and then into the same (to create retreat ways for your land units) (this is possible only if the land territory you go into and then out is friendly) (only multiple movement units can do it).
>
>If you start your turn into a hostile sea zone, you are allowed to go out and then into the same (to create retreat ways for your sea units) (this is possible only if the sea zone you go into and then out has no enemy sea units in it (it can be an enemy owned convoy zone you capture)), with no need of loading units.
>
>If you start your turn into a hostile land territory, you are allowed to go out of it, of 1 movement only, to an adjacent friendly land territory or sea zone (in this second case, you have to load onto sea units), ending your Combat Move for that land unit at this point (escaping combat), unless you want to perform a normal Combat Move movement. Then, you can fully move or unload from sea units, at any point during Non Combat Move, with any units that made such single movements (to escape Combat), during Combat Move, just like any land units that didn't move during Combat Move, except only that, if a land unit moved into an adjacent friendly land territory, during the Combat Move phase beforehand, it will have 1 movement less (meaning that it will be, either, unable to move, if it is a movement 1 unit, or able to move of 1 land territory only, if it is a movement 2 unit, since it already moved of 1 land territory), and will be unable to load onto sea units, and that if a land unit loaded onto sea units of other players, it will be unable to unload, on that same turn (practically, as long as the Non Combat Move phase is concerned, that single movement made during Combat Move is just like it was made during Non Combat Move; and indeed it is to be regarded as a Non Combat Move movement irreversibily (no edit granted, unless agreed) anticipated during Combat Move).
>
>If you start your turn into a hostile sea zone, you are allowed to go out of it, of 1 movement only, to an adjacent friendly sea zone (and eventually load any land units from any hostile territories), ending your Combat Move for that land unit at this point (escaping combat), unless you want to perform a normal Combat Move movement. Then, you can fully move, as well as load and unload units, at any point during Non Combat Move, with any units that made such single movements (to escape Combat), during Combat Move, just like any sea units that didn't move during Combat Move, except only that, if a sea unit moved into an adjacent friendly sea zone, during the Combat Move phase beforehand, it will have only 1 movement left (practically, as long as the Non Combat Move phase is concerned, that single movement made during Combat Move is just like it was made during Non Combat Move; and indeed it is to be regarded as a Non Combat Move movement irreversibily (no edit granted, unless agreed) anticipated during Combat Move).

This is how the program worked back in 2017, according to my experience. I do know that something changed since then.

Something similar should now be written to detail how the program currently works.

Cernelius:
> @Cernelius @panther2
> 
> I'd like to get your thoughts as rule experts if this is the right thing to do. One option is to make it a map property - but that would still result in the wrong error message for current maps.

Yes, I suggest having contested-to-friendly only as the basis and addining any further movement allowances (like being able to go from contested to already contested and whatever this PR is about) as well-worded properties (you need to have true).

Since now also land units can be unable to block or able to move past other units, if either all enemy units that are contesting the territory against you are unable to block other units or all units you are trying to move are able to move past them, then the case should be handled as if the starting territory was non contested (so that you can move directly even into an uncontested enemy territory with enemy units able to block your units).

Moreover, the matter should be the same as land for convoy zones (sea territories) if you cannot enter enemy convoy zones during Non Combat Move (there is a property for this too).

panther2:
@asvitkine 

> Allow moving land units out of contested territories they start in.

Only from a core-boardgame-rules perspective this feature would be restricted to 1914 and its ruleset only.

So IMHO the current beheaviour should be default, any variation maybe a welcome option for 1914 or any custom mapmaker game.

Otherwise it of course reminds me of the "Sea Units starting in hostile Seazones" - implementation - brought to land territories.

Cernelius:
@panther2 Is it true that, in 1914, my sea units starting in a contested sea zone can move to an adjacent sea zone which is occupied by enemy units and no other units, whereas my land units starting in a contested land zone cannot move to an adjacent land zone which is occupied by enemy units unless the adjacent zone is also already occupied by one or more units each of which is owned by me or one of my allies?

If so, we have a different handlying of the case in 1914, possibly depending on whether or not the zone is land or sea.

panther2:
@Cernelius With "otherwise" I meant the wwII-related games. And the "reminds me of" was just a hint towards a code implementation that meets the "Sea unit starting in hostile Sea Zones"-rules of the WWII games, that maybe could somehow be adapted/altered for use for territories in user made games, too. 

Concerning 1914 I am not familiar with the boardgame ruleset, as I have never been interested in this WW I scenario.
Of course - if really needed - I would take a look into the rulebook, but in case you, @Cernelius , are more familiar with it, I would prefer to focus on other things, right now.

asvitkine:
I guess my problem with the current behavior is the error is talking about blitz which is completely unrelated.

Otherwise, the map that's affected by this is Imperialism 1974 where you can move troops through any territories and can choose or not to attack (either other players or neutrals). The land battles being optional is another triplea property that needs to be added (it exists for sea battles but not land battles).

asvitkine:
@Cernelius Thanks. I've switched to split the blitz and no-blitz cases into separate checks such that the "no blitz" case will now have an error message that doesn't mention blitzing ("Cannot attack out of a contested territory"). I've added a property to allow such moves (in the second, non-blitz category) so existing maps are not affected.

I've revised the description of this change to mention the new logic. Can you read over that and let me know if you agree with that logic?

asvitkine:
Hmm, I'm confused. Is blitz broken in 2.5?

Here's a save game on Pacific Theatre Solo Challenge on 2.5, where I've modified infantry to have two movement (but no blitz ability). The fact that they have 2 mv is encoded in the save game, so you don't need to download a different map or change your local copy of that map.

It seems these infantry are able to blitz even without the blitz ability? What am I missing?
@Cernelius any ideas?

Again, this is on 2.5, so without any of the 2.6 changes.
[blitz_test_inf2.tsvg.zip](https://github.com/triplea-game/triplea/files/8882430/blitz_test_inf2.tsvg.zip)

asvitkine:
While testing the edge cases I found some things the current engine doesn't handle well. Please hold off on looking more for now as I'd like to fix them at the same time.

Cernelius:
> @Cernelius Thanks. I've switched to split the blitz and no-blitz cases into separate checks such that the "no blitz" case will now have an error message that doesn't mention blitzing ("Cannot attack out of a contested territory"). I've added a property to allow such moves (in the second, non-blitz category) so existing maps are not affected.
> 
> I've revised the description of this change to mention the new logic. Can you read over that and let me know if you agree with that logic?

I need some time to get around checking out the intended 1914 rules. Wouldn't be advisable also supporting the 1914 rules (with properties to be set true) as part of this PR?

Can we please have an exact and exhaustive description of how the program currently handles these matters when all properties are false? I think that is very important in order to build upon it.

Cernelius:
I believe it is important to clarify how the program behaves in all such situations (by "Neutral" I mean the relationship archetype neutral basic so not the "Neutral" player (which is not neutral but at war with anything)), assuming we are moving 1 land unit over land of 1 space only:

Starting territory ownership:
1. Own/Allied
2. Neutral
3. Enemy

Starting zone units (beside infrastructures):
1. Non-contested (the unit we are trying to move does not share the zone with any enemy units) (if the unit we are trying to move is in an enemy owned territory with no enemy units (beside infrastructures), the zone counts as non-contested)
2. Contested

Ending zone ownership:
1. Own/Allied
2. Neutral
3. Enemy

Ending zone units (beside infrastructures):
1. Empty
2. The only units present are one or more own/allied units
3. The only units present are one or more enemy units
4. Both one or more own/allied units and one or more enemy units are present

Outcome
Cannot move.
Can move.

@asvitkine If it is easier for you to know the results than for me to do so by testing every single case in a game, may you please compile a list of program behaviours without the property set true? This is the list (other than the ones where the ending zone is Neutral, I filled the obvious ones, where the starting territory is friendly owned and has no enemy units in it, and where the ending zone is friendly owned and has no enemy units in it):

1111=Can move
1112=Can move
1113=Can move
1114=Can move
1121=Cannot move
1122=Cannot move
1123=Cannot move
1124=Cannot move
1131=Can move
1132=Can move
1133=Can move
1134=Can move
1211=Can move
1212=Can move
1213=
1214=
1221=Cannot move
1222=Cannot move
1223=Cannot move
1224=Cannot move
1231=
1232=
1233=
1234=
2111=Can move
2112=Can move
2113=
2114=
2121=Cannot move
2122=Cannot move
2123=Cannot move
2124=Cannot move
2131=
2132=
2133=
2134=
2211=Can move
2212=Can move
2213=
2214=
2221=Cannot move
2222=Cannot move
2223=Cannot move
2224=Cannot move
2231=
2232=
2233=
2234=
3111=Can move
3112=Can move
3113=
3114=
3121=Cannot move
3122=Cannot move
3123=Cannot move
3124=Cannot move
3131=
3132=
3133=
3134=
3211=Can move
3212=Can move
3213=
3214=
3221=Cannot move
3222=Cannot move
3223=Cannot move
3224=Cannot move
3231=
3232=
3233=
3234=

Then list all the cases for which the property changes the behaviour.

Thanks.

Once this is done, I think it would be important to add such documentation within PoS2.

asvitkine:
@Cernelius 

Thanks for thinking about this diligently. I'm afraid I don't have a way to verify the above easier than manually testing.

I also think there's unfortunately a lot of extra conditions that the above wouldn't tell the full story of, such as whether the unit has moved before, whether that move was an attack, the presence of air and land transports, how much movement units have, whether a territory was conquered this turn, etc. Having said that, I don't think enumerating all the possible combinations of these options is feasible - there's just too many factors. So maybe just doing it for your suggested list is OK and the rest can be covered with some more targeted testing.

Cernelius:
> I also think there's unfortunately a lot of extra conditions that the above wouldn't tell the full story of, such as whether the unit has moved before, whether that move was an attack, the presence of air and land transports, how much movement units have, whether a territory was conquered this turn, etc.

The case in question is that of a not already moved land unit moving on its own from a land zone directly into an adjacent land zone without transporting any other land units.

Whether or not the move counts as an attack depends respectively on whether or not the move is a XX3X, a XXX3 or a XXX4, so it is already implicit in the list.

I doesn't matter if the unit has more than 1 movement capacity as only 1 is being used. If the unit can thereupon move on is irrelevant for the matter at hand.

It should not matter whether or not the starting or ending territory is conquered on the same turn and, if it does, it is a wrong program behaviour because all properly executed movements within the same turn happen at the same time. However, for testing purposes, the unit would be the only one moving on the turn.

---

I intend to test and to record all cases and possibly also to list all cases for the intended 1914 rules.

Cernelius:
I think I may have found the culprit of the "blitz" reference. I've tested it using "World War II Global 1940 2nd Edition".

If you have two adjacent enemy owned territories and 1 infantry and 1 armour owned by you in one of them, and neither territory has any other units, the infantry cannot enter the other territory, whereas the armour can and, in doing so, will conquer both territories. The armor can actually even conquer up to three territories this way.

I consider both behaviours to be right and to be maintained.

I have, however, found an other inconsistency and will open an issue, hoping for a fix.

panther2:
@Cernelius 
> If you have two adjacent enemy owned territories and 1 infantry and 1 armour owned by you in one of them, and neither territory has any other units, the infantry cannot enter the other territory, whereas the armour can and, in doing so, will conquer both territories. The armor can actually even conquer up to three territories this way.

panther2:
Sorry, a misclick.

asvitkine:
> I think I may have found the culprit of the "blitz" reference. I've tested it using "World War II Global 1940 2nd Edition".
> 
> If you have two adjacent enemy owned territories and 1 infantry and 1 armour owned by you in one of them, and neither territory has any other units, <...>

How is the above possible on Global 1940? I don't think you can have a unit in an enemy territory with no enemy units on that map, because that territory would become yours (outside of edit mode). Can you clarify?

panther2:
Wouldn't it be useful to test on a map supporting "Contested Territories" ?

Cernelius:
> Wouldn't it be useful to test on a map supporting "Contested Territories" ?

Not if the property `Abandoned Territories May Be Taken Over Immediately` is true: this would apply only to a game having both contested territories and no immediate take-over of abandoned ones.

Cernelius:
> > I think I may have found the culprit of the "blitz" reference. I've tested it using "World War II Global 1940 2nd Edition".
> > If you have two adjacent enemy owned territories and 1 infantry and 1 armour owned by you in one of them, and neither territory has any other units, <...>
> 
> How is the above possible on Global 1940? I don't think you can have a unit in an enemy territory with no enemy units on that map, because that territory would become yours (outside of edit mode). Can you clarify?

I've tested it by editing. This would be possible, however, in a variant of this game in which Russia starts with 1 armour inside a German-owned territory (as long as the property `Abandoned Territories May Be Taken Over Immediately` is false) or in any game in which you can go from Allied to War (not necessarily at once), like Napoleonic Empires (Yes, I could have tested this using that game, instead of editing "World War II Global 1940 2nd Edition".).
https://github.com/triplea-maps/napoleonic_empires/blob/master/map/games/Napoleonic_Empires.xml

Cernelius:
> How is the above possible on Global 1940? I don't think you can have a unit in an enemy territory with no enemy units on that map, because that territory would become yours

Wrong: there is no "Abandoned Territories May Be Taken Over Immediately" rule/property in that game, so the territory would become yours with no movement only during the Conduct Combat phase (and the Combat Move phase is beforehand). However, even this would not be a free lunch as all the units you left there would be thereafter unable to move until the end of the turn (as they would conquer the territory by making a battle in it against nothing).
You are correct that can neve happen in that game, but this is due to the fact that only neutral powers can become enemy (and the neutral relationship does not allow to move into any territories of the other power), not because "that territory would become yours".

Cernelius:
To summarize, there are four situations in which it may happen, all of which assume the game does not have the "Abandoned Territories May Be Taken Over Immediately" rule/property:
1. The game has limited combat rounds.
2. The game has land units able to "submerge".
3. The game has relationships allowing moving units into friendly territories not owned by the owner of the units and politics allowing war between powers sharing any such relationship (so, for example, you ally with a power, move a unit into one of its territories and leave it there and thereafter declare war on the power).
4. The game starts with units which are already inside territories of other powers which are or will be at war with the owner of any such unit.

asvitkine:
Thanks for all the feedback, I'll try to check and respond in detail tomorrow. In the meantime, I've been expanding tests for some of the things this PR is fixing and I've also updated the PR description to reflect the current state.

asvitkine:
> Wouldn't it be useful to test on a map supporting "Contested Territories" ?

Yes, sorry I was actually testing this but didn't mention in the PR description. I've updated the PR description to mention the two cases (one without the property set, which just has the better error message and one with the new property, which allows the movement). Saved games also attached.

asvitkine:
> > > I think I may have found the culprit of the "blitz" reference. I've tested it using "World War II Global 1940 2nd Edition".
> > > If you have two adjacent enemy owned territories and 1 infantry and 1 armour owned by you in one of them, and neither territory has any other units, <...>
> > 
> > 
> > How is the above possible on Global 1940? I don't think you can have a unit in an enemy territory with no enemy units on that map, because that territory would become yours (outside of edit mode). Can you clarify?
> 
> I've tested it by editing. This would be possible, however, in a variant of this game in which Russia starts with 1 armour inside a German-owned territory (as long as the property `Abandoned Territories May Be Taken Over Immediately` is false) or in any game in which you can go from Allied to War (not necessarily at once), like Napoleonic Empires (Yes, I could have tested this using that game, instead of editing "World War II Global 1940 2nd Edition".). https://github.com/triplea-maps/napoleonic_empires/blob/master/map/games/Napoleonic_Empires.xml

Hmm, let me try to understand this. So "Abandoned Territories May Be Taken Over Immediately" is documented as:

> Any territory abandoned may be taken over by the remaining units on the same turn it was abandoned on.
> If a territory is attacked and both all the attackers and all defenders die, then any other player with combat units will take over the territory.

So for the first condition, you still need a starting state where multiple enemy units are in a territory. Let's say, there's both German and Russian units in a German territory, and Germany then moves its units out. Then, it creates the situation you describe (assuming property is false, which is the default). Or as you say, a map where politics allows Allied -> War and there were units together.

But if I understand it right, none of the real life games implemented in TripleA can have this happen - so we're talking about custom maps (or house rules), at which point, there's no "spec" about how this should work, only whatever the mapmaker intended (if they thought about this case).

asvitkine:
Marking this as draft since there's actually a pending discussion on #10643 about one aspect of this PR. I think the overall code and test changes can still be reviewed in the meantime, and if there's a further refinement of this - that part would be a small thing to review beyond what's ready to review already.

asvitkine:
After some back and forth with @Cernelius, I'm abandoning changing the blitzing logic changes this PR was originally proposing and keeping movement out of contested territories working exactly as before when the new property isn't set.

I've updated the tests and the PR description to reflect the changes. @DanVanAtta, can you review?

game-app/game-core/src/main/java/games/strategy/triplea/delegate/move/validation/MoveValidator.java
```diff
@@ -82,6 +84,14 @@ public class MoveValidator {
   public static final String TRANSPORT_CANNOT_LOAD_AFTER_COMBAT =
       "Transport cannot load after being in combat";
   public static final String NOT_ALL_UNITS_CAN_BLITZ = "Not all units can blitz";
+  public static final String CANNOT_ATTACK_OUT_OF_CONTESTED_TERRITORY =
+      "Cannot attack out of a contested territory";
+  public static final String CANNOT_BLITZ_OUT_OF_BATTLE_INTO_ENEMY_TERRITORY =
+      "Cannot blitz out of a battle into enemy territory";
+  public static final String NOT_ALL_UNITS_CAN_BLITZ_OUT_OF_EMPTY_ENEMY_TERRITORY =
+      "Not all units can blitz out of empty enemy territory";
```
DanVanAtta, line 92: Is there any way to make this more specific regarding which units cannot blitz? Can we make this dynamic perhaps and list the unique unit types that cannot?

game-app/game-core/src/main/java/games/strategy/triplea/delegate/move/validation/MoveValidator.java
```diff
@@ -1450,62 +1452,67 @@ private MoveValidationResult validateParatroops(
...
+      if (!Properties.getParatroopersCanAttackDeepIntoEnemyTerritory(data.getProperties())
+          && route.getMiddleSteps().stream()
+              .anyMatch(Matches.territoryIsLand().and(Matches.isTerritoryEnemy(player)))) {
+        return result.setErrorReturnResult("Must stop paratroops in first enemy territory");
       }
     }
     return result;
   }
 
+  private Map<Unit, Unit> dependentsToUnitToTransportsMap(
```
DanVanAtta, line 1497: Is there a typo in the name of this method? Reading it, it sounds like we have 'dependents' -> units' -> 'transports'

Should it be: 'dependentUnitsToTransportMap'?

game-app/game-core/src/main/java/games/strategy/triplea/delegate/move/validation/MoveValidator.java
```diff
@@ -1450,62 +1452,67 @@ private MoveValidationResult validateParatroops(
...
+      final Map<Unit, Collection<Unit>> airTransportDependents, final MoveValidationResult result) {
+    Map<Unit, Unit> unitsToTransport = new HashMap<>();
+    for (Unit transport : airTransportDependents.keySet()) {
+      int capacity = TransportTracker.getAvailableCapacity(transport);
+      for (Unit beingTransported : airTransportDependents.get(transport)) {
+        int cost = beingTransported.getUnitAttachment().getTransportCost();
+        unitsToTransport.put(beingTransported, transport);
+        // Validate capacity, as airTransportDependents is coming from the move we're validating.
+        if (capacity < cost) {
+          result.setError("Not all units could be air transported");
```
DanVanAtta, line 1507: I wonder if we can make this error message more specific and helpful to the user. For example, could we say something like "not all units could be air transported, X out of Y transport capacity required"

Or, if we could say which units could not be transported, eg: "Not enough transport capacity for: X, Y, Z"

game-app/game-core/src/test/java/games/strategy/triplea/delegate/VictoryTest.java
```diff
@@ -175,6 +174,194 @@ void testMotorizedNoBlitzBlitzedTerritory() {
...
+      assertThat(Matches.unitCanBlitz().test(tankUnit.get(0)), is(true));
+      // Add them to angloEgypt.
+      gameData.performChange(ChangeFactory.addUnits(angloEgypt, tankAndInfantry));
+
+      advanceToStep(testBridge, "CombatMove");
+      moveDelegate.setDelegateBridgeAndPlayer(testBridge);
+      moveDelegate.start();
+    }
+
+    @Test
```
DanVanAtta, line 203: Consider adding javadocs to '@Test' to describe the test setup and what we are testing. Those comments describing the test I've personally found to be useful pretty soon after to go back and understand more easily what is being tested.

game-app/game-core/src/test/java/games/strategy/triplea/delegate/VictoryTest.java
```diff
@@ -175,6 +174,194 @@ void testMotorizedNoBlitzBlitzedTerritory() {
...
+      Collection<Unit> enemyUnits = CollectionUtils.getMatches(t.getUnits(), isEnemy);
+      gameData.performChange(ChangeFactory.removeUnits(t, enemyUnits));
+      assertThat(CollectionUtils.countMatches(t.getUnits(), isEnemy), is(0));
+    }
+
+    Matcher<String> isNotAllowedWithContestedAttackError() {
+      return is(MoveValidator.CANNOT_ATTACK_OUT_OF_CONTESTED_TERRITORY);
+    }
+
+    Matcher<Object> wasPerformedSuccessfully() {
```
DanVanAtta, line 360: :+1:  good use of a matcher, reads well.

DanVanAtta (commented):
Overall looks good, no major callouts or issues that I can see. I left a few comments for you to review @asvitkine  to consider places where we might be able to improve user messaging.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/move/validation/MoveValidator.java
asvitkine: It's actually "dependents" to "unitsToTransportMap", but I agree that it's hard to read.

How about I rename it to convertTransportKeyedMapToLoadedUnitKeyedMap()?

game-app/game-core/src/main/java/games/strategy/triplea/delegate/move/validation/MoveValidator.java
asvitkine: I'm not going to change anything with this error message since it's existing behavior (I just moved it to a variable), but actually the result object already contains the units that were invalid, since this message is added thus:

```
        result.addDisallowedUnit(NOT_ALL_UNITS_CAN_BLITZ_OUT_OF_EMPTY_ENEMY_TERRITORY, u);
```

With `u` being the unit that's not OK. So it's up to the code that's using the result (e.g. the UI code) to display that better, which would be a separate improvement.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/move/validation/MoveValidator.java
asvitkine: Let me try using result.addDisallowedUnit() here instead, like it's done for the blitz case.

asvitkine: Actually, addDisallowedUnit() is not an error, just a warning, so it's not the same semantics.

Thinking about this more, I think this is actually OK to leave as-is, because this situation won't happen in practice due to validation done in MovePanel - it doesn't let you load more units than allowed on an air transport. So this is just guarding against coding mistakes (e.g. if move panel changes or if AI submits invalid moves) or malicious players that are sending invalid moves with a custom client. As such, I'll just keep it as-is but add a comment to this extent.

game-app/game-core/src/test/java/games/strategy/triplea/delegate/VictoryTest.java
asvitkine: Done!

Cernelius:
"Not all units can blitz out of empty enemy territory" is not an universally good definition for blitzable territories. For example, if in Revised you can blitz-out an empty enemy territory, in Classic you should also be able to blitz out an enemy territory with a factory and a aa gun in it as the only enemy units (capturing or liberating these units).

In V2 and all following rules-set, you can indeed blitz only "empty" territories, but in V1 (which should be the default for the program) you can also blitz territories in which the only enemy units in them are all infrastructures.

So, the messages should be more something like:
V1: "Not all units can blitz out of an enemy territory who is empty of enemy units beside infrastructures."
V2,V3,V4,V5,1940: "Not all units can blitz out of an enemy territory who is empty of enemy units."

By the way, I'm with Classic here: I never liked the "factories stop tanks" nonsense that is the rule since Revised.

Side note, here I've used the term "infrastructures" even though I believe it is a bad term to define the units (Nobody would consider an AA Gun of WW2 an infrastructure, let alone a Napoleonic general.). I've raised the matter here:
https://forums.triplea-game.org/topic/1644/a-better-word-than-infrastructure
My proposal to rename "infrastructure":
https://en.wikipedia.org/wiki/Appurtenance

PR #10620, opened 2022-06-10
Vagrant: Conditionally generate dhparams file

DanVanAtta:
It takes a significant amount of time to generate dhparams file (around 10-20 minutes).
It is a one-time activity, but on vagrant we wind up doing this often because we
frequently recreate the VM. This update makes the file generation to be conditional
that we are not in a vagrant environment, if deploying to vagrant then we deploy
a pregenerated file to save time.


infrastructure/ansible/roles/nginx/tasks/main.yml
```diff
@@ -3,7 +3,16 @@
...
+    src: vagrant-only.dhparam.pem
+    dest: "{{ dhparams_pem_file }}"
+    owner: root
+    mode: "0600"
+
 - name: generate dhparam.pem
+  when: nginx_generate_dhparam
   shell:
     cmd: openssl dhparam -out {{ dhparams_pem_file }} 4096
     creates: "{{ dhparams_pem_file }}"
```
RoiEXLab, line 18: I'm a bit late to the party, but is the when clause required here, or does the `creates` part prevent overriding this parmeter?

DanVanAtta: You're right that the `creates` is not strictly required.

Though, I think there are perhaps at least 3 relatively solid reasons to keep the `when` clause. To illustrate the thinking, these 3 items are:

(1) expectations, `creates` is a pattern that is part of the shell command to skip a command because the shell command has already been run. In other words, it is typically used to skip a task because that task has already been run, as opposed to some other task was run which makes this task not needed (which is what a `when` is typically used for)

(2) there is a small performance benefit, the `when` clause being false means that ansible will not check the server redundantly for if the file exists

(3) explicit if/else structure between the two tasks marked with a `when`. IE: using a `when: true` and `when: false` blocks creates an if/else structure, using a `when: true` and `creates: file exists` is seemingly mixing structures.

PR #10631, opened 2022-06-11
Rename MoveDescription fields to be clear what they're for.

asvitkine:
Rename MoveDescription fields to be clear what they're for.

Also update variables and function names in related code to be explicit as well (e.g. being clear about matchers that check for sea transports and not any other kind).

Note: I don't believe anything changing here is part of the serialized game state. For example, UndoableMove doesn't use MoveDescription.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/MoveDescription.java
@@ -16,27 +16,29 @@ import lombok.Getter;
 public class MoveDescription extends AbstractMoveDescription {
   private static final long serialVersionUID = 2199608152808948043L;
   @Getter private final Route route;
-  @Getter private final Map<Unit, Unit> unitsToTransports;
-  @Getter private final Map<Unit, Collection<Unit>> dependentUnits;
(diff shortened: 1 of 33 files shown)


asvitkine:
CodeClimate issues are pre-existing.

asvitkine:
For this PR, I'd like someone to double check. I'm pretty sure this is correct - but since this is documenting specific semantics, I'd like someone to look over the diff with another pair of eyes to make sure I didn't miss something.

I've tentatively assigned to @RoiEXLab, but happy if someone else wants to review too. Thanks!

DanVanAtta:
Some nitpicks aside, looks okay to me. If this updated class is not serialized as part of the save game, then I have no concerns.

asvitkine:
I did some more function and variable updates to explicitly say SeaTransport instead of Transport for code that checked for those specifically.

After doing a self review, I found a piece of code that seemed suspicious and appeared like it was passing air transports to code expecting sea transports. However, after more investigation, I found the code had an impossible to be true condition and thus never ran. I sent https://github.com/triplea-game/triplea/pull/10639 to fix that separately (I'll rebase this PR on that before landing, since I'd like this PR to just be renames).

PR #10632, opened 2022-06-11
Some small code clean ups.

asvitkine:


game-app/game-core/src/main/java/games/strategy/triplea/delegate/AbstractUndoableMove.java
```diff
@@ -29,19 +28,7 @@ public AbstractUndoableMove(final CompositeChange change, final Collection<Unit>
     this.units = units;
   }
 
-  public boolean containsAnyOf(final Set<Unit> units) {
-    if (units == null) {
```
asvitkine, line 33: Note: I checked that units are never null for the call site this is being inlined into.

PR #10639, opened 2022-06-12
Remove impossible to hit code.

asvitkine:
Remove impossible to hit code.

This matcher requires a unit to be both an air transport and air transportable, which is impossible:
```
          final Predicate<Unit> paratroopNBombers =
              Matches.unitIsAirTransport().and(Matches.unitIsAirTransportable());
```
So this was always false:
```
final boolean paratroopsLanding = units.stream().anyMatch(paratroopNBombers);
```

To prove that a unit can't be both an air transport and air transportable, we can look at `UnitAttachment.validate()`.

First, an air unit can't be air transportable per lines 2606 and 2615 in UnitAttachment.java.
Second, an air transport can't be a land unit or sea unit (and thus must be an air unit), per lines 2632 and 2636 and lines 2619 and 2628 in UnitAttachment.java. 

So `Matches.unitIsAirTransport().and(Matches.unitIsAirTransportable())` can't match any units and this is just dead code.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/panel/move/MovePanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/panel/move/MovePanel.java
@@ -496,18 +496,14 @@ public class MovePanel extends AbstractMovePanel {
           }
           setSelectedEndpointTerritory(territory);
           Collection<Unit> transports = null;
-          final Predicate<Unit> paratroopNBombers =
-              Matches.unitIsAirTransport().and(Matches.unitIsAirTransportable());
           final var units = new ArrayList<>(unitsThatCanMoveOnRoute);
-          final boolean paratroopsLanding = units.stream().anyMatch(paratroopNBombers);
           if (route.isLoad() && units.stream().anyMatch(Matches.unitIsLand())) {
             transports = getTransportsToLoad(route, units);
             if (transports.isEmpty()) {
               cancelMove();
               return;
             }
-          } else if ((route.isUnload() && units.stream().anyMatch(Matches.unitIsLand()))
-              || paratroopsLanding) {
+          } else if (route.isUnload() && units.stream().anyMatch(Matches.unitIsLand())) {
             units.clear();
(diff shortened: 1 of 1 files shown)


panther2:
@asvtikine
Just an idea: In global1940_game the infantry `isAirTransportable`. But different from other editions, once the paratrooper technology has been acquired, the Infantry (Paratrooper) does not fly loaded on a bomber, but flies "on its own". 
Might this be a case where `Matches.unitIsAirTransport().and(Matches.unitIsAirTransportable())` might occur? 
As the unit is its own "Air Transport" ?

I have no idea how this is coded, but maybe....

asvitkine:
@panther2 

Thanks for mentioning. I'm not sure which map is "global1940_games" (I didn't find a map folder / repo matching that exact name), but I tested on "World War II Global 1940 2nd Edition" and I see there's an Airborne_Forces tech which you might have in mind:

```xml
				<attachment name="techAbilityAttachment" attachTo="Airborne_Forces" javaClass="games.strategy.triplea.attachments.TechAbilityAttachment" type="technology">
						<option name="airborneForces" value="true"/>
						<option name="airborneCapacity" value="2:airfield"/>
						<option name="airborneBases" value="airfield"/>
						<option name="airborneTypes" value="infantry"/>
						<option name="airborneDistance" value="3"/>
						<option name="airborneTargettedByAA" value="AA:infantry"/>
				</attachment>
```

So this one makes airfields be able to launch infantry I think. Is that what you're talking about?

I double checked and infantry isn't marked as an air transport, so I don't believe this affects the code I'm changing. Let me know if there's something else I should be checking.

panther2:
@asvitkine 

Sorry, I will be more precise concerning map-names the next time. I am indeed talking about every map related to "World War II Global 1940 2nd Edition". And yes, exactly, I referred to the Airborne-Forces tech,

I have no idea, whether the listed attachment enables an infantry to "fly on its own".
So you maybe might want to check, what actually enables an infantry to "fly on its own", if it is not `isAirTransport`.

Thank you.

PR #10641, opened 2022-06-12
Reduce calls to GameRunner#headless() (again)

RoiEXLab:
After this PR almost no usages are left.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LaunchAction.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LaunchAction.java
@@ -10,7 +10,6 @@ import games.strategy.engine.framework.startup.WatcherThreadMessaging;
 import games.strategy.engine.framework.startup.mc.IServerStartupRemote;
 import games.strategy.engine.framework.startup.mc.ServerConnectionProps;
 import games.strategy.engine.framework.startup.mc.ServerModel;
-import games.strategy.engine.framework.startup.ui.PlayerTypes;
 import games.strategy.engine.framework.startup.ui.panels.main.game.selector.GameSelectorModel;
 import games.strategy.engine.player.Player;
 import games.strategy.net.Messengers;
@@ -43,7 +42,11 @@ public interface LaunchAction {
 
   ChatModel createChatModel(String chatName, Messengers messengers);
 
-  PlayerTypes.Type getDefaultPlayerType();
+  /**
+   * Controls how sparsely the AI should be used when preparing a game. Headless systems may choose
+   * to avoid AI usage where possible to reduce the load on the system.
+   */
+  boolean isAiFallback();
 
   WatcherThreadMessaging createThreadMessaging();
 
--- a/game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
@@ -257,7 +257,7 @@ public class ServerModel extends Observable implements IConnectionChangeListener
               }
             });
 
-        lobbyWatcherThread.createLobbyWatcher(gameToLobbyConnection);
+        lobbyWatcherThread.createLobbyWatcher(gameToLobbyConnection, !launchAction.isAiFallback());
       } else {
         gameHostingResponse = null;
       }
--- a/game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
@@ -191,9 +191,7 @@ public class ServerModel extends Observable implements IConnectionChangeListener
           playersToNodeListing.put(name, null);
         }
       } else {
-        Optional.ofNullable(messengers)
-            .ifPresent(
-                messenger -> playersToNodeListing.put(name, messenger.getLocalNode().getName()));
+        playersToNodeListing.put(name, messengers.getLocalNode().getName());
       }
       playerNamesAndAlliancesInTurnOrder.put(
           name, data.getAllianceTracker().getAlliancesPlayerIsIn(player));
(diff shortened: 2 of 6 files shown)


game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
```diff
@@ -191,9 +190,7 @@ private void updatePlayersOnGameDataChanged(final GameData data) {
           playersToNodeListing.put(name, null);
         }
       } else {
-        Optional.ofNullable(messengers)
-            .ifPresent(
-                messenger -> playersToNodeListing.put(name, messenger.getLocalNode().getName()));
```
RoiEXLab, line 196: Small consistency change. In practice messengers should always be non-null at this point,  but in case something odd happens it would now crash regardless of if the client is headless or not. Previously it would only cause an NPE on headless clients.

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/launcher/LaunchAction.java
```diff
@@ -43,7 +42,11 @@ public interface LaunchAction {
 
   ChatModel createChatModel(String chatName, Messengers messengers);
 
-  PlayerTypes.Type getDefaultPlayerType();
+  /**
+   * Controls how sparsely the AI should be used when preparing a game. Headless systems may choose
+   * to avoid AI usage where possible to reduce the load on the system.
+   */
+  boolean isAiFallback();
```
RoiEXLab, line 49: I'm not really satisfied with this name, but I hope the javadoc helps here.

asvitkine: I find both the name and the description confusing! Admittedly, I'm  not familiar with the code it's replacing and the motivation behind it.

First, I don't understand what it means for the LaunchAction to be an "Ai Fallback". The two are separate - the action  isn't an AI or a fallback AI. Instead, it sounds like this is trying to control some behavior. So perhaps the verb should be chosen to reflect that more - something like shouldUseAiForFallback().

Although, I still don't understand what "fallback" means here and the comment doesn't help me, unfortunately. "AI Fallback" can mean two different things to me - it could be a fallback for an AI (i.e. instead of using AI, we use X and X is the fallback for AI). Or it could be that instead of some thing Y, we use an AI and the AI is the fallback for Y. Now, neither of these actually make sense to me - I don't understand what X or Y could be - I mean the alternative is human players, or perhaps a "do nothing AI"? Can you clarify?

As for the comment, I'm pretty confused by "Headless systems may choose to avoid AI usage" - I guess, I don't understand what that means. Does it mean that they're not allowed to have AI players? Does it mean that the AIs they have should use less resources (e.g. less battle sim rounds)? Something else? Why should this actually be dependent on headless - I could image some headless systems wanting to do different things (e.g. whether it's a bot hosting a game vs. a headless test of AI logic)...

Sorry to zoom in on this little thing, but just offering feedback here about how I feel this name isn't great...

RoiEXLab: No worries, feedback is welcome here.

> Admittedly, I'm not familiar with the code it's replacing and the motivation behind it.

Well regarding the bigger picture I'm trying to remove static references from the code where global state introduces modularity issues. The last few PRs were able to eliminate the static reference to `HeadlessGameServer` allowing me to move individual classes to the game-headless subproject, decoupling the logic from the implementation a bit removing the code from the binary we distribute with the installer and also decreasing the installer size by a couple of bytes.

Ideally game-core should end up without any references to the `java.desktop` module (this includes swing/awt classes) and all the UI code should be part of the game-headed submodule. This would reduce the size of headless binaries, and if we eventually decide we want to make use of java 9 modules (requires some refactoring due to package name conflicts, but could be done in a weekend with a lot of time patience I think) we could also think about using custom JVMs to ship with the installer to avoid shipping modules that are not called at all. This also puts the code into a spot where the UI becomes more of an implementation detail rather than part of the logic. This would allow us to potentially look into alternative UI frontends maybe even the web (when using WASM for the java backend).
Have a look at the `ClipPlayer` class for example. It allocates a static instance of itself if there is no instance present using a generic resource loader instance and recreating this static instance if the resourceloader became outdated because a new map has been loaded. So why does 1. `ClipPlayer` offer static methods to play clips in the first place if it has to be recreated on every map load, shouldn't it be a non-static member of the game class or something (actually there is a soundchannel interface that calls this class) and 2. why is this code part of game-core if headless devices can't even and shouldn't use this code. I'll try to refactor this in the coming weeks but it is heavily used in a lot of places so I'm not sure how well this will work.

For this change specifically the motivation is rather simple. The `GameRunner#headless()` method currently checks a global property and returns true or false based on that property. This makes testing behaviour confusing when you have to set a random property to be able to properly run your test instead of explicitly passing a value.

So why the confusing name? Basically if a "bot" tries to host a game the "bot" does not want to run the AI because the servers are usually pretty low-spec to cut down on costs. So it will only choose AI if a certain player is disabled (some maps offer optional fractions, AFAIK mostly neutral ones with 3 territories maximum) so for those the "bot" will run a weak ai, they are pretty rare though. In all other cases the "bot" makes connected clients run the AI if they don't want to control the fraction manually. For local game hosts, simulating AI is usually not a problem, so this behaviour does not apply there and AI can be offered as a normal option. At least that's how I understand the comments of this code. Does this clarify the issue a bit? Do you have any suggestions`?

Sorry for the wall of text, but I hope you can understand my struggle with this now.

asvitkine: Thanks for the explanation. So if this controls two things:
  1. The type of AI chosen for disabled players.
  2. How the bot sets up seats (i.e. whether it will ever fill seats itself with AIs based on the map defaults?).

If so, perhaps make two functions, each controlling the above?

Alternatively, you could use a function name that is closer to the comment (e.g. `shouldMinimizeExpensiveAiUse()`).

I think either of those would be clearer than the "fallback" terminology that I mentioned I find confusing.

RoiEXLab:
Thank you @asvitkine for your feedback. I changed the method name according to your feedback and tweaked the javadoc a little

PR #10642, opened 2022-06-12
Update units in bottom bar after clicking a move.

asvitkine:
Update units in bottom bar after clicking a move, or as a result of another change in units in the territory.

Also, rework threading in BottomBar to get game data properties under a lock and update the UI on the EDT.


RoiEXLab (approved):
👍🏼

PR #10651, opened 2022-06-14
Cache technology bonuses to speed up engine and AI code.

asvitkine:
On Domination 1914 No Man's land, on all Hard AI game, this speeds up the mean MustFightBattle.fight() execution time from 1.9ms to 1.4ms on my machine, a 26% speed up.

There are about 165,000 MustFightBattle.fight() calls during the first turn by Hard AI on Domination 1914.

A test is added covering the different ways in which tech can change to ensure the cache gets properly validated and the TechTracker returns the correct results.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/TechTracker.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/TechTracker.java
@@ -38,7 +38,7 @@ public class TechTracker {
 
   private final Map<Key, Object> cache = new ConcurrentHashMap<>();
 
-  public void onTechnologyChanged() {
+  public void clearCache() {
     cache.clear();
   }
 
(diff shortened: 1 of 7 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/delegate/TechTracker.java
```diff
@@ -41,6 +38,10 @@ static class Key {
 
   private final Map<Key, Object> cache = new ConcurrentHashMap<>();
 
+  public void onTechnologyChanged() {
```
DanVanAtta, line 41: Consider renaming this method to 'clearCache' so it is a bit more explicit what it is doing. With the current naming, it feels like the method should accept a parameter for the technology that has changed, or a parameter for the new set of technologies, or perhaps some sort of technology object.

Side-comment: It would be nice to have a more automatic way to notify the cache that it needs to be updated, perhaps via an event-listener style pattern (though, I don't really see a super clean way to do this without blowing up the scope of this PR).

DanVanAtta (commented):
Just the one comment regarding consideration of a method name. Otherwise looks good :+1:

game-app/game-core/src/main/java/games/strategy/triplea/delegate/TechTracker.java
asvitkine: Renamed, thanks! Agree that I would have preferred if there was a clearer way to listen to technology changes, but unfortunately it's all scattered around like this.

PR #10658, opened 2022-06-14
Remove injection class

RoiEXLab:
I'm not sure what exactly this class tried to accomplish in the first place. The 2 fields it had were the engineVersion which was effectively used as a singleton. This is a case where using a singleton is perfectly fine so I just made it its own static field. The other field was a good candidate for the `LaunchAction` interface (which should be called `LaunchContext` the more I think about it)


DanVanAtta:
`Injections` is meant to be a way to have Guice-like inversion-of-control dependency injection. There can be cases for example where a low level submodule needs an implementation that is provided by a different submodule that depends on the low level module. For the most part, everything is crammed into game-core and it is one of the lowest submodules that we have, so we do not see this as a problem. Though, if we continue to break apart game-core, this would be more of a problem. For example, the map parsing code will likely become a very low level submodule, and it needs to know the current engine version. So, to solve this, we can create an interface that the lowest level submodule would depend on, and then bind an implementation to that interface at a high level. This can also be extended throughout the AI code that should also not have as much knowledge about modules above it (but alas, as most things, all AI code is bundled in game-core, and hence also has a lot of static coupling with the rest of the engine)

DanVanAtta:
Another example to consider, instead of using the 'service discovery' we could use the 'injections' class to instead inject a game launcher implementation. I do dislike the 'service discovery' code and find it _extremely_ hard to understand. The explicit injection of a game runner I believe would be considerably simpler.

asvitkine:
My 2¢ is I dislike both the "service discovery" and Injections mechanisms.

In general, I'm not a fan of injections much, since it makes it hard to understand what's really happening in the code, but can concede that they can perhaps be useful for some tests, if it's not possible to just pass down objects directly, which I definitely prefer.

The problem with injections is it's really no different than global variables, which have well known cons. The pros are usually about making it easier to mock things, such as in tests, but I haven't seen much of a need for that in TripleA tests. At least not for gameplay, I guess we don't really test Swing UI for example, so can't speak to whether something would be useful with that.

RoiEXLab:
@DanVanAtta Well this PR moves the "submodule dependent" part of the default player types to the headed implementation of `LaunchAction` along with some other parts of the code that are moved there. So I would consider that an improvement. Better separation of headed vs non headed code via two interface implementations and less global state.
The description of the `Injections` class even stated that it should not be misused as a singleton which in my opinion was exactly what was happening with the engine version thingy. Both headed and non headed code were calling the exact same code to initialize the version. So why not making it an actual singleton? This way we get better error messages and a shorter syntax without duplicated code. After those changes the `Injections` class had no members anymore so I removed it. If there was ever again a need for such a class it could be re-introduced but i believe in most cases defining an interface is the cleaner solution, especially now that more and more code can be found in the game-headed submodule.

So are there any objections to merging?

RoiEXLab:
The codeclimate issues are pre existing

DanVanAtta:
@RoiEXLab  re: codeclimate - if the issues are pre-existing and not appropriate to fix, you can 'approve' the exceptions by following the codeclimate details link. LMK if it does not give you a link

PR #10659, opened 2022-06-14
Refactor some GameParser code.

asvitkine:
Refactor some GameParser code.

Moves various helpers for dealing with variables to GameDataVariableParser and adds test coverage for them. Cleans up some related code.

These changes were the pure refactoring / clean up changes extracted from https://github.com/triplea-game/triplea/pull/10425.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
@@ -39,7 +39,6 @@ import java.io.InputStream;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
```diff
@@ -182,6 +179,7 @@ private GameData parse(final Path xmlFile, final InputStream stream)
 
     parseDelegates(game.getGamePlay().getDelegates());
 
+    variableParser.parseVariables(game.getVariableList());
```
DanVanAtta, line 182: There is a side-effect in this call where `parseVariables` caches the variables that were parsed.

WDYT if we made this more functional and `variableParser` returned those parsed variables and did not have any side effects?

DanVanAtta (commented):
Just one comment regarding concern about side-effects

DanVanAtta:
@asvitkine unrelated question: are you on forums? If so - what's your username there (DM LaFayette on forums).

game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
asvitkine: That's how it was before, see line 206 in the previous diff, but then GameParser had all these functions to operate on these variables. This change moves all that functionality to GameDataVariableParser, so it's all together.

We could refactor things so there's two classes, so there's a wrapper over the returned map. Or another option is we rename GameDataVariableParser to GameDataVariables and have a static parse() function that returns it, this way we don't need two classes.

asvitkine:
> @asvitkine unrelated question: are you on forums? If so - what's your username there (DM LaFayette on forums).

Messaged you on the forums.

game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
asvitkine: OK, I ended up doing my proposed refactoring above. Please take another look.

DanVanAtta (approved):

PR #10666, opened 2022-06-15
Add "Land Battles May Be Ignored" property.

asvitkine:
This works the same as "Sea Battles May Be Ignored", but for land battles.

This implements the rules needed by Imperialism 1974, when set.

Tested:
  - Added the property to Imperialism 1974 and observed that the user is given a prompt about whether they want to fight a given land battle. Tested playing several turns with no issues.


asvitkine:
Given the obvious parallel to "Sea Battles May Be Ignored" and the trivial implementation, I'm going to go ahead and merge this.

Issue #10599, opened 2022-06-06
2.6: Defensive dice roll after attack inflicts casualties doesn't handle calculate support bonuses correctly


asvitkine:
### Reproduction Steps

1. Open the following save game: 
[2.5-correct-support-a.tsvg.zip](https://github.com/triplea-game/triplea/files/8848657/2.5-correct-support-a.tsvg.zip)
2. Attack Northern Italy with all units in the neighboring sea zone
3. Land 2 or more hits and select non-Infantry units to die as Germans

The defensive side should roll with 6 dice, regardless of number and types of units that were killed by the attack roll. This is how it works with 2.5.

In 2.6, the number of dice is reduced if supporting units were killed, which is incorrect.

### Attach a Save Game

See

### Engine Version

Engine Version: 2.6.857

asvitkine:
@beelee1 filed based on your report on the forum. I confirm this looks like a 2.6 regression.

Probably this broke when a lot of the battle code was rewritten in 2.6. I'll try to find the revision.

asvitkine:
I went as far back as 2.6.89 where the issue still repros (I needed to use edit mode from scratch because older 2.6 couldn't handle 2.5 saved games), here's a save that works with earlier 2.6 versions:
[repro-2.6-int.tsvg.zip](https://github.com/triplea-game/triplea/files/8849049/repro-2.6-int.tsvg.zip)

Before that tag, it used different numbering scheme so I couldn't go back in time. But if it's really the 89th PR since 2.6, then it shouldn't be too many to go through.

asvitkine:
Actually, looks like older prereleases used bigger numbers for some reason, like:
https://github.com/triplea-game/triplea/releases/tag/2.6.22512

asvitkine:
OK, getting somewhere.

[2.6.22311](https://github.com/triplea-game/triplea/releases/tag/2.6.22311) doesn't have the problem.
[2.6.22343](https://github.com/triplea-game/triplea/releases/tag/2.6.22343) does.

So somewhere between these releases.

asvitkine:
Ok more narrow range, between [2.6.22324](https://github.com/triplea-game/triplea/releases/tag/2.6.22324) (good and [2.6.22328](https://github.com/triplea-game/triplea/releases/tag/2.6.22328) (bad)."

```
git log 2.6.22324..2.6.22328

commit bf7a5367047e34f0f384f19176654a45d54a0048 (tag: 2.6.22328)
[...]

PR #10667, opened 2022-06-15
Fix 2.6 dice roll count from support attachments when there are casualties.

asvitkine:
Fix 2.6 dice roll count from support attachments when there are casualties.

This attempts to fix https://github.com/triplea-game/triplea/issues/10599, where casualties on the defensive side incorrectly stop giving support to the defensive dice roll.

Tested:
  - Open this save game: 
[2.5-correct-support-az.tsvg.zip](https://github.com/triplea-game/triplea/files/8912471/2.5-correct-support-az.tsvg.zip)
  - Perform the pending combat.

The defense side should have 6 dice rolls, regardless of the number of casualties inflicted by the attackers. The same can be confirmed with the same save game on 2.5. You can try several times if attackers end up not hitting enough times.

Adds a test to verify the behavior. Includes a some small clean ups in related code.

Concerns:
  - No unit tests are exercising this scenario. We should add some. Edit: Added.
  - It's not clear whether this has any other implications. @trevan, if you're around, could you review?

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/MainDiceRoller.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/MainDiceRoller.java
@@ -1,7 +1,6 @@
 package games.strategy.triplea.delegate.battle.steps.fire;
 
 import static games.strategy.triplea.delegate.battle.BattleState.UnitBattleFilter.ACTIVE;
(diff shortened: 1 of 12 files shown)


asvitkine:
@DanVanAtta LMK if you have feedback on this one. I'd like to merge today so we can have a prerelease with the fix for this weekend. Now that I've written a test, I'm more confident with the change.

asvitkine:
I'm going to go ahead and merge since it does fix the 2.6 regression and I've added a test for the behavior.

Issue #10649, opened 2022-06-13
AiGameTest > testAiGame sometimes fails with "No winner after 100 rounds"


asvitkine:
I've not seen this in any local runs of the test, but I have seen this on the CI bots. Example:
https://github.com/triplea-game/triplea/runs/6867290877?check_suite_focus=true

Ideally, we could dig into what's going on. The test run will actually result in autosaves like "autosave_round_even.tsvg" being saved, which can be loaded to inspect what happened in the game (the more granular autosaves are disabled).

So if we could get access to these, we can inspect the turns and figure out what's going wrong. Theoretically, it's possible the dice rolls end up such that the game is naturally a stalemate, but I suspect it's more likely that there's an AI bug that doesn't handle some game states well (e.g. perhaps building undefended transports that get destroyed each turn or whatever).

@DanVanAtta is there any way we can recover the save game from test bots? If not, could we make the test code upload the save file to somewhere in the case of such a failure?

asvitkine:
OK, I got a repro locally. Here's a save.

Russians keep buying bombers and Germans keep buying tanks. No one ever invades the other because no ships built.
[autosave_round_odd.tsvg.zip](https://github.com/triplea-game/triplea/files/8896463/autosave_round_odd.tsvg.zip)

@ron-murhammer Any thoughts on this?

asvitkine:
I think Russians AI is the one not playing right. They don't need so many bombers and given they got the middle island, they should instead build a fleet. At the very least, capturing Corner 1 should be simple.

Maybe it's a case where they can't build a useful fleet in one single turn and they don't know how to save up to build one later? But given Germans just have one bomber, they should be able to build up a fleet in their home sea zone even one turn at a time...

asvitkine:
Interesting that the win% if 180%. That doesn't seem right.

If I boost the score of the cruiser, it does get chosen for simulation, but then the result is:
```
    SZ 1, can't defend TUVSwing=1061.7305011554643, win%=162.73050115546425, tried to placeDefenders=[cruiser owned by Russians], enemyAttackers=[bomber owned by Germans]
```

Even though a manual battle sim results in a 34% win rate for the bomber and both units cost the same amount of PUs to build. That seems suspect, I'll investigate.

(In terms of human strategy, I think building a cruiser is definitely the right call here with the available income, and over time building up an attack fleet.)

asvitkine:
[...]

PR #10675, opened 2022-06-16
Teach Hard AI to save up for a fleet.

asvitkine:
With this change, Hard AI will not use its unused PUs to buy high-value non-sea units (that otherwise aren't needed) if it needs to save up for a fleet. It will do this if there are no enemy territories reachable by land from its factories and it's able to build fleets that can reach enemy land. Additionally, the logic only triggers if we tried and failed (due to not being able to defend) to place sea units and only if we're not already at the resources needed to build the max fleet we can.

This resolves a cause of the AI getting into a stalemate on the Mapmaking Tutorial map, which we use as a test for AI, see: https://github.com/triplea-game/triplea/issues/10649

Adds a test that runs 50 games for the first round only and ensures no errors and no winner on the first run (this validates that the above AI logic fix is effective in ensuring the AI doesn't leave its capital exposed). 

Tested:
  - Taking the 100-turn save game from the above issue and continuing AI turns to check that stalemate is broken: 
[autosave_round_odd.tsvg.zip](https://github.com/triplea-game/triplea/files/8915601/autosave_round_odd.tsvg.zip)
  - Truncating same save game to turn 6 (where stalemate started) and verifying it doesn't happen in the first place: 
[autosave_round_odd_t6.tsvg.zip](https://github.com/triplea-game/triplea/files/8915602/autosave_round_odd_t6.tsvg.zip)
  - Doing all-AI games on a number of maps and observing when this triggers. Observed positive improvement for AIs that exercise this (e.g. delaying the victory of the opposite AI side). Notably, this triggers for Classic 3rd ed for UK and for ANZAC on Pacific 1940. It also sometimes kicks in for Japan on Global 1940 on the first turn, when Japan would otherwise purchase 2 bombers instead of a fleet due to how it did its combat / non combat moves.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
@@ -3,11 +3,13 @@ package games.strategy.engine.data.util;
 import games.strategy.engine.data.GameMap;
 import games.strategy.engine.data.Territory;
 import java.util.ArrayDeque;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.function.Predicate;
 import org.triplea.java.ObjectUtils;
+import org.triplea.java.collections.CollectionUtils;
 
 /**
  * Implements Breadth First Search (BFS) to traverse / find territories. Since the search criteria
@@ -41,19 +43,23 @@ public final class BreadthFirstSearch {
   }
 
   /**
-   * @param startTerritory The territory from where to start the search.
+   * @param startTerritories The territories from where to start the search.
    * @param neighborCondition Condition that neighboring territories must match to be considered
    *     neighbors.
    */
   public BreadthFirstSearch(
-      final Territory startTerritory, final Predicate<Territory> neighborCondition) {
-    this.map = startTerritory.getData().getMap();
-    this.visited = new HashSet<>(List.of(startTerritory));
-    this.territoriesToCheck = new ArrayDeque<>(List.of(startTerritory));
+      Collection<Territory> startTerritories, Predicate<Territory> neighborCondition) {
+    this.map = CollectionUtils.getAny(startTerritories).getData().getMap();
+    this.visited = new HashSet<>(startTerritories);
+    this.territoriesToCheck = new ArrayDeque<>(startTerritories);
     this.neighborCondition = neighborCondition;
   }
 
-  public BreadthFirstSearch(final Territory startTerritory) {
+  public BreadthFirstSearch(Territory startTerritory, Predicate<Territory> neighborCondition) {
+    this(List.of(startTerritory), neighborCondition);
+  }
+
+  public BreadthFirstSearch(Territory startTerritory) {
     this(startTerritory, t -> true);
   }
 
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProAi.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProAi.java
@@ -19,6 +19,7 @@ public class ProAi extends AbstractProAi {
   public void stopGame() {
     super.stopGame(); // absolutely MUST call super.stopGame() first
     concurrentCalc.cancel();
+    concurrentCalc.setGameData(null);
   }
 
   @Override
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
@@ -348,23 +348,24 @@ class ProPurchaseAi {
         prioritizeSeaTerritories(purchaseTerritories);
     purchaseSeaAndAmphibUnits(purchaseTerritories, prioritizedSeaTerritories, purchaseOptions);
 
-    // Try to use any remaining PUs on high value units
-    purchaseUnitsWithRemainingProduction(
-        purchaseTerritories, purchaseOptions.getLandOptions(), purchaseOptions.getAirOptions());
-    upgradeUnitsWithRemainingPUs(purchaseTerritories, purchaseOptions);
+    // Try to use any remaining PUs on high value units, except if we need to save up for a fleet.
+    if (!shouldSaveUpForAFleet(purchaseOptions, purchaseTerritories)) {
+      purchaseUnitsWithRemainingProduction(
+          purchaseTerritories, purchaseOptions.getLandOptions(), purchaseOptions.getAirOptions());
 
-    // Try to purchase land/sea factory with extra PUs
-    purchaseFactory(
-        factoryPurchaseTerritories,
-        purchaseTerritories,
-        prioritizedLandTerritories,
-        purchaseOptions,
-        true);
+      upgradeUnitsWithRemainingPUs(purchaseTerritories, purchaseOptions);
 
-    // Add factory purchase territory to list if not empty
-    if (!factoryPurchaseTerritories.isEmpty()) {
-      purchaseTerritories.putAll(factoryPurchaseTerritories);
+      // Try to purchase land/sea factory with extra PUs
+      purchaseFactory(
+          factoryPurchaseTerritories,
+          purchaseTerritories,
+          prioritizedLandTerritories,
+          purchaseOptions,
+          true);
     }
+
+    // Add factory purchase territory to list
+    purchaseTerritories.putAll(factoryPurchaseTerritories);
 
     // Determine final count of each production rule
     final IntegerMap<ProductionRule> purchaseMap =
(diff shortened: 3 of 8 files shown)


asvitkine:
Hmm, we hit "g.strategy.engine.data.AiGameTest - No winner after 50 rounds" on the CI, which means this doesn't fully address the problem. But maybe there's another separate cause.

asvitkine:
Hmm, had an interesting case on the Tutorial map we use for AiTest with this PR. It's not a normal occurrence, but had a case where Germany was able to win on T1. Russia didn't buy any units (likely partially due to this new logic) and moved out their fleet, allowing Germany to attack and win right away (they do start with a stronger fleet and an invasion force lined up).

However, I think that this new code isn't solely responsible for this. First, even if Russia buys a tank, Germany is still favored to win the invasion (although it may make the opposing AI hesitate on the attack). Russia could instead keep its fleet where it is to block the invasion and reinforce the fleet a bit, or it could gather an infantry to defend. Actually, its t1 move made no sense. It didn't capture a convoy. It didn't land a unit in the middle to get a VC worth 4 that's not defended. It didn't try to block Germany's sea route (I know AI doesn't currently do this and it's not clear if it's the best course of action).

Anyway, save game here: [autosave_round_even_t1_win.tsvg.zip](https://github.com/triplea-game/triplea/files/8931364/autosave_round_even_t1_win.tsvg.zip)

ron-murhammer:
I think you really need to be careful with this and test it across a number of maps. I believe this edge case should pretty rarely occur to fire this logic and if its firing when it shouldn't then it could lead to worse AI results.

asvitkine:
@ron-murhammer 

It's peculiar since it's just the first move of the first turn, so it's interesting the AI is doing different things (but obviously there's randomness in combat simulation and some AI decisions).

For example, this is how Russia moved with the same code another time:
Screen Shot 2022-06-17 at 5 44 24 PM

So here it bought an artillery and split up its ships (so germany can just pick them off...) Not exactly the best move, but I guess prevents an attack on the capital.

asvitkine:
OK, usually the purchase is an infantry (75%) and the plan is to collect the infantry from Corner 2 to bring it to the capital. Not sure what's up with the split fleet.

Not clear what happed the one time it didn't build anything. Possibly it got an unlucky simulation and thought it couldn't defend even with 4 units?

asvitkine:
> I think you really need to be careful with this and test it across a number of maps. I believe this edge case should pretty rarely occur to fire this logic and if its firing when it shouldn't then it could lead to worse AI results.

Agree, let me do more testing. I think one other tweak to the logic that makes sense is to only choose this option when we tried to build a fleet but couldn't due to not being able to defend it. Right now, it doesn't take that into account directly.

asvitkine:
I'm still looking at the rare case where the AI choses not to defend its capital on T1 (and then "saves up" its starting 5 PUs using the new logic). It's definitely not common (something like 1/20 games), so I had to put a loop around the test and detect this condition.

When it does happen, it seems "Prioritize territories to defend with isLand=true" skips Moscow due to `enemyAttackOptions.getMax(t) == null` being true (which is not the case for the other games).

@ron-murhammer do you know offhand what that means and why that would be? Is it thinking for some reason that the enemy won't attack it? Is that based on estimating that the attack might be bad for the enemy (e.g. it got unlucky in battle calc simulation)? Something else?

As I said, the other games where this doesn't happen, it's not null (so Moscow isn't skipped).

asvitkine:
So still looking at why `enemyAttackOptions.getMax(t) == null` sometimes.

Seems when we do `populateEnemyAttackOptions()`, the move maps don't include Moscow on a run when that happens and do include on other runs.

good:
`    moveMaps:[{Moscow=Moscow, SZ 3=SZ 3, SZ 2=SZ 2, Corner 2=Corner 2, SZ 1=SZ 1}]``
bad:
`    moveMaps:[{SZ 3=SZ 3, SZ 2=SZ 2, Corner 2=Corner 2, SZ 1=SZ 1}]`

So need to see where that comes from.

asvitkine:
I think maybe what's happening is if Russia moves destroyers to SZ1 and Transport to SZ3, I think they incorrectly conclude that the enemy can't pass (but they can if it's just a transport there). Interestingly, the enemy AI doesn't make that same mistake and thus is able to attack.

This is what I can tell so far with some logging added to findAmphibMoveOptions().

This seems to be an existing problem that's not directly related to this PR (but that this PR helped expose). Will dig more to narrow down the cause.

asvitkine:
Sent https://github.com/triplea-game/triplea/pull/10719/files for the AI logic for movability through sea zones not matching engine logic.

asvitkine:
After the above fix, I observed another different instance of the bug where the capital isn't defended appropriately.

In this case, it seems the AI simulated enemy attacks as if the fleet was kept next to capital (thus concluding capital was safe, although that itself is dubious since I think it needs to check if they enemy can win the sea battle - because if they can, they can still land units), but then moved its units out of the sea zone (!). Seems like another bug.

asvitkine:
> After the above fix, I observed another different instance of the bug where the capital isn't defended appropriately.

Actually, that last observation may have been a misinterpretation of the logs on my part, or else _extremely_ rare. I've modified the test to run 1000 times, but only the first turn, with this PR and the ProMatches fix. There were a couple rare NPEs I ran into, which I have a fix for. With all the above, it now runs successfully all 1000 times (did it twice to make sure).

Successfully means we're no longer seeing the t1 win behavior (which was caused by the ProMatches logic error fixed by https://github.com/triplea-game/triplea/pull/10719) and no other errors like NPEs.

asvitkine:
Been testing a bit more, finding more things where the AI could do be better.
Screen Shot 2022-06-19 at 9 59 34 PM

AI just decides to abandon its fleet by flying away with its three fighters. Sure, they save the fighters, but a better move could have been to reinforce the fleet - buy a ship or a fighter and move the fleet next to the capital. Another option is to sacrifice two destroyers by moving the rest of the fleet behind them.

[fleet_abandon.tsvg.zip](https://github.com/triplea-game/triplea/files/8936975/fleet_abandon.tsvg.zip)

asvitkine:
> I think you really need to be careful with this and test it across a number of maps. I believe this edge case should pretty rarely occur to fire this logic and if its firing when it shouldn't then it could lead to worse AI results.

I've been testing this on several maps with logging that tells me when the new logic kicks in. It's pretty rare.

But for example on Classic 3rd edition, the only time it triggers is for UK first turn, so they are able to place a fleet 2nd turn. As a result, Allies are able to do a bit better than without the change - it delays Axis victories by several rounds.

Here's a save with the change at the end of an all-AI game: 
[saving_classic_3rd.tsvg.zip](https://github.com/triplea-game/triplea/files/8945315/saving_classic_3rd.tsvg.zip)

Here's a save without the change at the end of an all-AI game:
[saving_classic_3rd_normal.tsvg.zip](https://github.com/triplea-game/triplea/files/8945319/saving_classic_3rd_normal.tsvg.zip)

asvitkine:
On ww2 pacific 1940 2nd edition, the new logic triggers a lot, mostly for ANZAC. On a game with the new logic, Japan holds 6 VCs all around on round 21 (win con):
[saving_1940_2nd_ed_pacific.tsvg.zip](https://github.com/triplea-game/triplea/files/8945475/saving_1940_2nd_ed_pacific.tsvg.zip)

Playing the same map without the logic, Japan achieves 6 Vcs on round 13: 
[saving_1940_2nd_ed_pacific_normal.tsvg.zip](https://github.com/triplea-game/triplea/files/8945479/saving_1940_2nd_ed_pacific_normal.tsvg.zip)

Now, I didn't scrutinize all the moves in detail and it's possible there was other luck-based factors in those games, but so far the new logic seems to be helping the side where it kicks it. (On this map, Japan never had that logic kick in, only allies.)

ron-murhammer:
Makes sense that it will tend to trigger more for nations with lower income and those are the nations that it often makes more sense to save some income to be able to place a larger fleet that won't be easily wiped out.

asvitkine:
Converting to a draft as I investigate Japan triggering this on turn 1 sometimes on 1940 Global - and I am not sure if it's correct yet. This is on WW2 Global 40 Balanced Mod3 with Combat Move First.

Relevant line from the log:

>     6 Sea Zone, can't defend TUVSwing=1232.629880800958, win%=333.62988080095806, tried to placeDefenders=[destroyer owned by Japanese, carrier owned by Japanese], enemyAttackers=[transport owned by Americans, fighter owned by Americans, submarine owned by Americans, destroyer owned by Americans, destroyer owned by Americans, submarine owned by Americans, fighter owned by Americans, cruiser owned by Americans, fighter owned by Americans]
> Saving up for a fleet, since enemy territories are only reachable by sea

Notably, it reasons that US can reach SZ 6 around Japan with a big fleet. That's true, but I believe it's getting fighter movement wrong - as US, I can only move 2 fighters there, not three, and the two that "can" move there, I think are still technically not OK because there's no way they can land.

Aside, that logging line doesn't tell me the full defenders it tried to simulate with - i.e. whether it took into account the other units already in that sea zone. Japan definitely didn't move the existing fleet out, so I assume yes...

So need to look into this more.

asvitkine:
After some more testing on global 1940, some more findings:
  1. This case isn't hit too often actually. Sometimes I need to try ~5 games before it happens.
  2. When it happens, if I disable the new code, but just log when it does, then the normal purchase for Japan ends up being 2 bombers. (This is via the "spend remaining resources on high value units" logic.) I don't think that's a great buy for Japan, so actually saving money and using it more effectively next turn is likely not a worse outcome.

Per above, and per the other positive results on other maps from my testing, I'm pretty confident about this change. Going to land.

PR #10677, opened 2022-06-16
Merge Delegate related implementations together

RoiEXLab:

The discussion in #10656 got me thinking so I had a look at some of the classes in this area. `GameDelegateBridge` was mostly delegating it's calls to an underlying `IDelegateBridge` implementation except for an `GameDelegateHistoryWriter` which itself was delegating to an underlying `IDelegateHistoryWriter` after adding an optional prefix when edit mode is active. I went ahead and merged the history writer classes, under the assumption that all bridges will end up becoming game delegate bridges eventually which made those classes redundant so they could be removed.

I ran an all-ai game for a couple of rounds without problems and edit mode still seems to work as expected.

Also `BaseEditDelegate` had a TODO indicating it could be merged with its only child class.

With that being said there are now only 4 implementations of `IDelegateBridge` remaining: `DefaultDelegateBridge`, `ProDummyDelegateBridge`, `ObjectiveDummyDelegateBridge` and `DummyDelegateBridge`. As @DanVanAtta already pointed out that's a bad indicator for code quality. I'm not quite sure how to best deal with those implementations. They differ in some areas but are very similar in others and there's just no way we need 3 dummy implementations


asvitkine:
Changes LGTM, have you checked that there's no problem with 2.5 saves?

asvitkine:
> @asvitkine I haven't checked explicitly but none of the classes implement the serializable interface so there's no way they land in the savegames

Makes sense. I knew we had code to serialize delegates, but looks like the objects themselves aren't serialized, since the code just does:

```
      out.writeObject(DELEGATE_START);
      // write out the delegate info
      out.writeObject(delegate.getName());
      out.writeObject(delegate.getDisplayName());
      out.writeObject(delegate.getClass().getName());
      out.writeObject(DELEGATE_DATA_NEXT);
      out.writeObject(delegate.saveState());
```

PR #10680, opened 2022-06-16
Refactor TuvUtils to allow caching to speed up AIs & battle calculator.

asvitkine:
Refactor TuvUtils to allow caching. New TuvCostsCalculator class is introduced, that keeps around intermediate calculations allowing repeated calls to re-use them. Refactors call sites to call the new class, although the BattleCalculator call site is the one that benefits the most from this, by re-using the same object for all battle simulations (which is heavily used by AI).

This should be especially impactful for maps with lots of units, for example "another world". I tested a round 7 game on "another world" with all Hard AI players and the mean runtime of MustFightBattle.fight() went from 3.0ms to 2.3ms, a 23% speedup (which was called around ~185,000 times during the course of AI simulations).

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/delegate/IDelegateBridge.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/delegate/IDelegateBridge.java
@@ -3,13 +3,16 @@ package games.strategy.engine.delegate;
 import games.strategy.engine.data.Change;
 import games.strategy.engine.data.GameData;
 import games.strategy.engine.data.GamePlayer;
+import games.strategy.engine.data.UnitType;
 import games.strategy.engine.display.IDisplay;
 import games.strategy.engine.history.IDelegateHistoryWriter;
 import games.strategy.engine.player.Player;
 import games.strategy.engine.random.IRandomStats.DiceType;
 import games.strategy.triplea.ResourceLoader;
+import games.strategy.triplea.util.TuvCostsCalculator;
 import java.util.Properties;
 import org.triplea.http.client.web.socket.messages.WebSocketMessage;
+import org.triplea.java.collections.IntegerMap;
 import org.triplea.sound.ISound;
 
 /**
@@ -97,4 +100,8 @@ public interface IDelegateBridge {
   void sendMessage(WebSocketMessage webSocketMessage);
 
   ResourceLoader getResourceLoader();
+
+  default IntegerMap<UnitType> getCostsForTuv(final GamePlayer player) {
+    return new TuvCostsCalculator().getCostsForTuv(player);
+  }
 }
--- a/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculator.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculator.java
@@ -14,6 +14,7 @@ import games.strategy.engine.framework.GameDataUtils;
 import games.strategy.triplea.delegate.battle.BattleResults;
 import games.strategy.triplea.delegate.battle.BattleTracker;
 import games.strategy.triplea.delegate.battle.MustFightBattle;
+import games.strategy.triplea.util.TuvCostsCalculator;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
@@ -92,6 +93,8 @@ class BattleCalculator implements IBattleCalculator {
       final List<Unit> defenderOrderOfLosses =
           OrderOfLossesInputPanel.getUnitListByOrderOfLoss(
               this.defenderOrderOfLosses, defendingUnits, gameData);
+      // Use a single TuvCostsCalculator so its computations are cached.
+      final TuvCostsCalculator tuvCalculator = new TuvCostsCalculator();
       for (int i = 0; i < runCount && !cancelled; i++) {
         final CompositeChange allChanges = new CompositeChange();
         final DummyDelegateBridge bridge =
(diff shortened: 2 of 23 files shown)


game-app/game-core/src/main/java/games/strategy/engine/delegate/IDelegateBridge.java
```diff
@@ -97,4 +100,8 @@ public interface IDelegateBridge {
   void sendMessage(WebSocketMessage webSocketMessage);
 
   ResourceLoader getResourceLoader();
+
+  default IntegerMap<UnitType> getCostsForTuv(final GamePlayer player) {
+    return new TuvCostsCalculator().getCostsForTuv(player);
+  }
```
RoiEXLab, line 106: Hmm this seems oddly specific tbh. `getData()`, `getHistoryWriter()` all seem so generic. Also is there a reason why a new instance of `TuvCostsCalculator` is created every call to this except for the fact that interfaces can't have any fields?
If this method was changed to something like `getTuvCostsCalculator()` this would allow some callers to store the object in a local variable  and use it a couple of times so the behaviour of `DummyDelegateBridge` and the other implementations would be more consistent. What do you think?

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculator.java
```diff
@@ -25,6 +26,8 @@
 
 class BattleCalculator implements IBattleCalculator {
   @Nonnull private final GameData gameData;
+  // Use a single TuvCostsCalculator so its computations are cached.
+  private final TuvCostsCalculator tuvCalculator = new TuvCostsCalculator();
```
RoiEXLab, line 30: Is this absolutely crucial for the performance improvements. I feel like in a lot of cases it would  be sufficient to have a field like this in the individual classes that require it, but I might be completely wrong here

asvitkine: Yes, it's part of the major performance speed up. The BattleCalculator is re-used to perform different computations with the same overall GameData state.

game-app/game-core/src/main/java/games/strategy/engine/delegate/IDelegateBridge.java
asvitkine: I agree that re-using is better, but the place where this matters is the BattleCalculator case, where we use a IDelegateBridge that overrides this version. Ideally, the other implementations ones can override it too, but those don't really matter for performance reasons. I can add a comment.

Exposing a `getTuvCostsCalculator()` will just make the call sites more verbose, but we can certainly change it if you feel strongly about that.

game-app/game-core/src/main/java/games/strategy/triplea/util/TuvCostsCalculator.java
```diff
@@ -0,0 +1,161 @@
...
+      final int costPerGroup = rule.getCosts().getInt(pus);
+      // we round up the cost
+      final int roundedCostPerSingle =
+          (int) Math.ceil((double) costPerGroup / (double) numberProduced);
+      if (differentCosts.containsKey(ut)) {
+        differentCosts.get(ut).add(roundedCostPerSingle);
+      } else {
+        final List<Integer> listTemp = new ArrayList<>();
+        listTemp.add(roundedCostPerSingle);
+        differentCosts.put(ut, listTemp);
```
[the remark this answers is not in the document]
asvitkine: Good catch!

game-app/game-core/src/main/java/games/strategy/engine/delegate/IDelegateBridge.java
asvitkine: One more follow up cost, I disagree with the following:

> If this method was changed to something like getTuvCostsCalculator() this would allow some callers to store the object in a local variable and use it a couple of times

Having callers having to cache the object just adds a lot of complexity, whereas the ideal implementation is the delegate bridge is what's caching the object. It's just that in practice, it only matters for the BattleCalculator case since that does something like 100k battles for every 20 battles in the real game.

asvitkine: Sorry and one more bit of context - the reason we can't just trivially make it a member of e.g. DefaultDelegateBridge, is that the information could potentially change and we would need to clear the cache at appropriate times. That's a lot of extra complexity that's not needed to fix the main performance problem.

RoiEXLab: Thanks for the context, this definitely helps understanding the issue. So the performance improvements won't work in general in a real "scenario" only for simulation, correct?

Regardless I just had an idea, not sure how feasible it is though: Did you consider moving the tuv calculating code into `GamePlayer`? Could that help with the problems you just mentioned or is this something completely different?

asvitkine: I think if we want to persist on something like Gameplay or GameData, then we need to make sure to reset the cache when certain state changes which is a lot of complexity.

game-app/game-core/src/main/java/games/strategy/triplea/util/TuvCostsCalculator.java
```diff
@@ -0,0 +1,161 @@
...
+   * player. Any production rule that produces multiple units (like artillery in NWO, costs 7 but
+   * makes 2 artillery, meaning effective price is 3.5 each) will have their costs rounded up on a
+   * per unit basis (so NWO artillery will become 4). Therefore, this map should NOT be used for
+   * Purchasing information!
+   *
+   * @param player The player to get costs schedule for
+   * @return a map of unit types to PU cost
+   */
+  public IntegerMap<UnitType> getCostsForTuv(final GamePlayer player) {
+    return costsPerPlayer.computeIfAbsent(player, p -> computeCostsForTuv(p));
```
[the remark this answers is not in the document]
asvitkine: Done

asvitkine:
I'm going to go ahead and submit since I'd like this to make it into a prerelease that can be tested over the weekend. Happy to continue the discussion around this if there are other ideas of how to improve it.

Issue #10673, opened 2022-06-16
AiGameTest > testAiGame sometimes hits "Not in a step, but trying to add event"


asvitkine:
When running on the bots, sometimes AiGameTest > testAiGame sometimes hits "Not in a step, but trying to add event".

Example:
https://github.com/triplea-game/triplea/runs/6910149568?check_suite_focus=true

```
    java.lang.IllegalStateException: Not in a step, but trying to add event: Filler event for child: Russians roll AA dice in Moscow : 1. Current history node is: null
    	at games.strategy.engine.history.HistoryWriter.startEvent(HistoryWriter.java:119)
    	at games.strategy.engine.history.HistoryWriter.addChildToEvent(HistoryWriter.java:150)
    	at games.strategy.engine.framework.ServerGame$1.addChildToEvent(ServerGame.java:135)
    	at jdk.internal.reflect.GeneratedMethodAccessor26.invoke(Unknown Source)
    	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
    	at games.strategy.engine.message.unifiedmessenger.EndPoint.invokeSingle(EndPoint.java:136)
    	at games.strategy.engine.message.unifiedmessenger.EndPoint.lambda$invokeMultiple$0(EndPoint.java:120)
    	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
    	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
    	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
    	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
    	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
    	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
    	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
    	at games.strategy.engine.message.unifiedmessenger.EndPoint.invokeMultiple(EndPoint.java:121)
    	at games.strategy.engine.message.unifiedmessenger.EndPoint.invokeLocal(EndPoint.java:111)
[...]

PR #10689, opened 2022-06-18
Fix "Not in a step" exception w/ battle calc.

asvitkine:
Fix "Not in a step" exception w/ battle calc.

This fixes https://github.com/triplea-game/triplea/issues/10673, which can sometimes happen due to the cloned game state being used for battle calc simulation having a blank history. This change ensures that the cleared history has a round and step started.


asvitkine:
Hmm, not sure what's going on with the unit tests checker.

asvitkine:
Still not sure why this makes Unit Tests hang.

asvitkine:
OK, figured out the test hang issue. The new code in resetHistory() was trying to acquire a write lock, while in a block that was holding a read lock, causing a dead lock. Surprises there's no check in the lock code that verifies this.

Fixed by grabbing a write lock in the outer code. This should have been the case anyway, since state is being changed (e.g. swapping history, etc).

PR #10705, opened 2022-06-18
ClipPlayer refactoring

RoiEXLab:

This removes the static fields in the ClipPlayer class and makes the UI context manage its own instance so individual games have their own instance without replacing the one static instance every time a game is loaded.

I verified sound still works in chat and in-game after this change


game-app/game-core/src/main/java/games/strategy/triplea/TripleA.java
```diff
@@ -59,7 +59,9 @@ public void startGame(
       }
     }
     final LocalPlayers localPlayers = new LocalPlayers(players);
-    game.setDisplay(launchAction.startGame(localPlayers, game, players, chat));
-    game.setSoundChannel(launchAction.getSoundChannel(localPlayers));
+    LaunchAction.UserInterface userInterface =
+        launchAction.startGame(localPlayers, game, players, chat);
+    game.setDisplay(userInterface.getDisplay());
+    game.setSoundChannel(userInterface.getSound());
```
DanVanAtta, line 65: `UserInterface` strikes me as a bit of an odd interface name. It is not bad, so I am a bit torn, but it seems somewhat overly generic for what it is - a tuple object.

I wonder if this block would get simpler if `game` were mutated and the end of the `startGame` method simply sets the game sound and display values. Side-effects are not good, but there are a lot of other side effects happening here that I think we could have a net-gain.

WDYT to move the `game.set` calls into the `startGame` method?

game-app/game-core/src/main/java/games/strategy/triplea/TripleAPlayer.java
```diff
@@ -256,6 +255,14 @@ public boolean acceptAction(
             politics);
   }
 
+  private void playSound(final String soundPath) {
+    if (ui == null) {
+      log.warn("UI not set when trying to play sound");
```
DanVanAtta, line 260: The warning will show a pop up to user, seemingly this will not be very informative to them.
Are we sure that `ui` will possibly be null at this point? If so, perhaps it is null for a valid reasons - in those cases, should we be showing a warning message to the user? 
I wonder a bit as well if `TripleAPlayer` should have a function for playing sounds which we could bind to the uiContext clip player when the uiContext is set.

game-app/game-core/src/main/java/org/triplea/sound/ClipPlayer.java
```diff
@@ -129,14 +126,18 @@ public class ClipPlayer {
     }
   }
 
-  protected final Map<String, List<URL>> sounds = new HashMap<>();
+  private final Map<String, List<URL>> sounds = new ConcurrentHashMap<>();
```
DanVanAtta, line 129: Why `ConcurrentHashMap`? Asked another way, what/where is the concurrency concern?

RoiEXLab: The `play` method caches sound paths in this map. In theory `play` could be called from multiple threads. I doubt that ever happens (because in this case this should have went wrong in production already), so I wanted to make sure this would never happen

game-app/game-core/src/main/java/games/strategy/triplea/TripleAPlayer.java
RoiEXLab: This warning should not happen under normal circumstances. `HeadedLaunchAction#startGame` sets the frame on `TripleAPlayer`. Thinking about it this really should be either an exception that is thrown or an error that gets logged instead of a warning, because ui can only ever be null if the code tries to play a sound before the game is completely initialized. `TripleAPlayer` only ever plays sound based on gameplay events if I understand this correctly.

game-app/game-core/src/main/java/games/strategy/triplea/TripleA.java
RoiEXLab: I haven't thought about that. Mixed feelings on this. As you said side-effects are rarely intuitive, but they happen here anyways and I agree it would simplify the code. I think I'll go with that.

game-app/game-core/src/main/java/org/triplea/sound/ClipPlayer.java
DanVanAtta: Even if called from multiple threads, it wouldn't be a problem unless we mutate the contents of the map, right?

RoiEXLab:
@DanVanAtta Thanks for the quick review. I updated the code according to your comments

DanVanAtta:
Thanks, your call on HashMap vs ConcurrentHashMap

DanVanAtta (approved):

game-app/game-core/src/main/java/org/triplea/sound/ClipPlayer.java
RoiEXLab: Not sure about that to be honest. I changed the code to use `computeIfAbsent`. So it's clear that individual entries are never modified. The map as a whole however is, not sure how robust the implementation is there

RoiEXLab:
`ConcurrentHashMap` feels safer. It could be that the default `HashMap` is completely sufficient already, but I doubt that sounds are played so often the performance hit makes a difference there. Thanks a lot

PR #10707, opened 2022-06-18
Code-Climate Config: exclude markdown (md) files

DanVanAtta:
It does not make a lot of sense to do markdown format linting, so it is
being turned off. We routinely ignore the called out warnings and they
are largely not important. The MD syntax enforces a consistent enough style,
beyond this deviations in consistency of MD files is largely inconsequential.
So, this is just not an important thing for us to be linting on right now.
Perhaps in the future an auto-formatter could do all the formatting for us.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -40,3 +40,4 @@ exclude_patterns:
   - "gradlew"
   - ".github/ISSUE_TEMPLATES/*"
   - "**/*Test.java"
+  - "*.md"


DanVanAtta:
Tangentially, FWIW, gave this config a try in `spotless`. Regrettably, the spotless run had a runtime around a minute on every execution of `./gradlew spotlessApply`:

```
    spotless {
        // https://github.com/diffplug/spotless/tree/main/plugin-gradle#prettier
        format 'styling', {
            target '*.md'
            prettier().config(['tabWidth': 2])
        }

        java {
            googleJavaFormat()
            removeUnusedImports()
        }
    }
```

Issue #10028, opened 2022-01-20
2.5.22294: ClientModel#internalPlayerListingChanged:403 - java.lang.NullPointerException


tripleabuilderbot:
## TripleA Version
2.5.22294

## Java Version
11.0.6

## Operating System
Windows 10

## Memory
Heap utilization statistics [MB]
Used Memory: 29
Free memory: 77
Total memory: 107
Max memory: 2048

## Stack Trace
```
Exception: java.lang.NullPointerException
java.lang.Exception
	at games.strategy.engine.framework.startup.mc.ClientModel.internalPlayerListingChanged(ClientModel.java:403)
	at games.strategy.engine.framework.startup.mc.ClientModel.createClientMessenger(ClientModel.java:280)
	at games.strategy.engine.framework.startup.ui.panels.main.SetupPanelModel.showClient(SetupPanelModel.java:110)
	at games.strategy.engine.framework.GameRunner.showMainFrame(GameRunner.java:98)
	at java.base/java.lang.Thread.run(Thread.java:834)


```

asvitkine:
Repro: Host a networked game on 2.6 and connect to it with a 2.5 client. The error happens on the 2.5 client.

beelee1:
closing v 2.5

PR #10709, opened 2022-06-19
Refactor TripleAPlayer and move to game-headed

RoiEXLab:

This PR refactors the structure of the types implementing the `Player` interface. With those changes it's possible to move `TripleAPlayer` to the game-headed submodule.

**NOTE**: I'm pretty sure the changes are safe because otherwise the logic doesn't make sense. However, I haven't explicitly tested my changes on a local lobby setup due to time constraints. I'll try to do that tomorrow, feel free to review in the meantime.


game-app/game-core/src/main/java/games/strategy/engine/player/Player.java
```diff
@@ -39,8 +38,9 @@ public interface Player extends IRemote {
   @RemoteActionCode(6)
   String getName();
 
-  @RemoteActionCode(8)
-  PlayerTypes.Type getPlayerType();
```
RoiEXLab, line 43: I'm 99% sure this is safe because this type seems to have been changed in the past without problems, and it is not serializable. So serialization over the network would have failed anyways.

game-app/game-core/src/main/java/games/strategy/engine/framework/LocalPlayers.java
```diff
@@ -30,7 +29,6 @@ public boolean playing(final GamePlayer player) {
 
   private static boolean isGamePlayerWithPlayerId(
       final Player player, final GamePlayer gamePlayer) {
-    return player.getGamePlayer().equals(gamePlayer)
-        && AbstractHumanPlayer.class.isAssignableFrom(player.getClass());
+    return player.getGamePlayer().equals(gamePlayer) && !player.isAi();
```
RoiEXLab, line 32: I'm not sure if in a headless environment the code tries to invoke this over the network. If it does I have to re-introduce a dirty check if this class inherits from some abtract class in the type hierarchy

game-app/game-core/src/main/java/games/strategy/engine/player/Player.java
DanVanAtta: A 2.5 client connecting to a 2.6 server might exercise this. If that doesn't, then this method is seemingly never invoked.

asvitkine: I think 2.5 hits an npe connecting to a 2.6 server right now

RoiEXLab: @asvitkine Would you mind sharing a stacktrace of the error?

RoiEXLab: @asvitkine Hmm I can run the code without problems in my test scenario which should cover the same code. However I noticed something that could be mistaken for a NPE.
The `UnifiedInvocationHandler` sometimes logs a stacktrace (for debug reference not because there is an actual exception) when the UI thread calls a blocking network operation:
```java
      log.info("Blocking network operation performed from EDT", new Exception());
```

I noticed that gets logged in lobby chat for example. The cleanest fix would be to run it in a non-ui thread, the dirty fix would be to configure the logger to make an exception for this class

game-app/game-core/src/main/java/games/strategy/engine/framework/LocalPlayers.java
RoiEXLab: Seems safe

RoiEXLab:
@DanVanAtta @asvitkine Checked and seems to run without issues

game-app/game-core/src/main/java/games/strategy/engine/player/Player.java
asvitkine: It's this issue: https://github.com/triplea-game/triplea/issues/10028

Note: It's a bug in 2.5, since the NPE happens for the 2.5 client.

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
```diff
@@ -478,10 +478,7 @@ private Map<String, PlayerTypes.Type> getLocalPlayerTypes() {
 
     final Map<String, PlayerTypes.Type> localPlayerMappings = new HashMap<>();
     // local player default = humans (for bots = weak ai)
-    final PlayerTypes.Type defaultLocalType =
-        launchAction.shouldMinimizeExpensiveAiUse()
-            ? PlayerTypes.WEAK_AI
-            : PlayerTypes.HUMAN_PLAYER;
+    final PlayerTypes.Type defaultLocalType = launchAction.getDefaultLocalType();
```
asvitkine, line 481: maybe getDefaultLocalPlayerType()?

RoiEXLab:
@DanVanAtta Oh and btw. I can't mark codeclimate issues as fixed even though I'm logged in. Might be a permissions thingy.

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
```diff
@@ -83,13 +45,7 @@ public PlayerTypes(final Collection<Type> playerTypes) {
...
 
   public static Collection<Type> getBuiltInPlayerTypes() {
-    return List.of(
-        PlayerTypes.HUMAN_PLAYER,
-        PlayerTypes.WEAK_AI,
-        PlayerTypes.FAST_AI,
-        PlayerTypes.PRO_AI,
-        PlayerTypes.CLIENT_PLAYER,
-        PlayerTypes.BATTLE_CALC_DUMMY);
+    return List.of(PlayerTypes.WEAK_AI, PlayerTypes.FAST_AI, PlayerTypes.PRO_AI);
```
asvitkine, line 48: Should this be called getBuildInAiPlayerTypes()?

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
```diff
@@ -127,7 +83,7 @@ public abstract static class Type {
     @Getter(AccessLevel.PACKAGE)
     private final boolean visible;
 
-    Type(final String label) {
+    public Type(final String label) {
```
asvitkine, line 86: AiType?

asvitkine (approved):

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
RoiEXLab: Not necessarily. This class is used for "human players" and AIs alike.

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
RoiEXLab: built-in is grammatically correct.

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/mc/ServerModel.java
RoiEXLab: Good suggestion

game-app/game-core/src/main/java/games/strategy/engine/framework/startup/ui/PlayerTypes.java
asvitkine: Sorry for the typo, I was actually asking if "Ai" should be in the name, given this only returns ai types now.

RoiEXLab: Ah my bad didn't notice it. Anyways I believe Fast_AI and Pro_AI can be removed from this list as well and moved to game-headed. There's really few code that is actually required for game-headless and therefore game-core

asvitkine: I don't understand why AI should go to the headed codebase. Seems like they should be independent of UI.

I think right now there is a dependency in the ProLogUi code for appending to a swing text box, but I think only that part should be refactored out behind an interface.

I think it would be confusing if some Said were in one project and others in another.

RoiEXLab: @asvitkine You have a point and we might need to introduce another submodule eventually (or just move the AI code to `:game-app:ai`?). Ideally `:game-app:game-core` will only contain code that is shared between `:game-app:game-headless` so that includes essential game logic. There is a point to be discussed: "Is AI essential game-logic?" The headless client only offers weak ai, so why would be binary need to be shipped with all the code necessary for the more sophisticated AI approaches?
In essence it's safe to say that while all UI code should be banned from `:game-app:game-core` this doesn't mean that all non-ui code should remain there.

DanVanAtta:
I'll try to read in more detail soon, forgive any missing context.
Game-core fwiw is somewhat of a bridge, eventually all of its code should
be moved to other, well defined submodules. It is all about creating
"waists" in the code, partitions between the major components of the
application: https://www.oilshell.org/blog/2022/02/diagrams.html

PR #10719, opened 2022-06-19
Fix ProMatches logic around passable sea territories.

asvitkine:
Fix ProMatches logic around passable sea territories to account for transports that don't block movement.


asvitkine:
Interestingly, the fix seems to make the "no winner after 50 turns" error happen more frequently, which itself should be getting fixed by #10675. That might mean we'll have to land the two changes together, rather than separately.

asvitkine:
> Interestingly, the fix seems to make the "no winner after 50 turns" error happen more frequently, which itself should be getting fixed by #10675. That might mean we'll have to land the two changes together, rather than separately.

Actually, looks like the matcher method wasn't appropriate - so this wasn't working correctly and actually introducing an issue. I've instead added the `getIgnoreTransportInMovement()` logic into the existing `territoryHasOnlyIgnoredUnits()` function.

PR #10721, opened 2022-06-19
Clean up code in ProTerritoryManager.

asvitkine:
Clean up code in ProTerritoryManager.

Make it more concise. No functional changes.

Some minor perf improvements by moving some work outside loops and passing in the territory to getTransporting() to avoid scanning the full map.


asvitkine:
codeclimate is about deeply nested blocks, but actually this PR is reducing the nesting, so overall nesting is reduced, but codeclimate is not smart enough to figure that out since it just sees more blocks at a given nesting level N even if we have fewer of them at level N+1. I'm going to ignore these.

PR #10725, opened 2022-06-21
Clean up some matches code.

asvitkine:


asvitkine:
Codeclimate issues are existing.

PR #10730, opened 2022-06-22
Use locks instead of polling to start headless game

RoiEXLab:

I kept stumbling over this comment stating that a latch might be useful. So I finally decided to do something about it.

## Release Note

CHANGE|Headless hosts now start the game immediately once all players are ready instead of waiting up to 8 seconds to the next check.


game-app/game-headless/src/main/java/org/triplea/game/server/HeadlessServerSetup.java
```diff
@@ -5,22 +5,32 @@
...
     this.model = model;
     this.gameSelectorModel = gameSelectorModel;
     this.model.setRemoteModelListener(this);
+    lock = new ReentrantLock();
+    playersUpdated = lock.newCondition();
   }
 
   @Override
   public void cancel() {
+    signalPotentialStart(false);
```
asvitkine, line 33: Is it normal that after cancel() is called, a playerListChanged() will set isRunning back to true?
That seems a bit unintuitive to me, but I guess it depends what `isRunning` is supposed to mean.

Should it instead be `cancelled` (with inverted value)? `isRunning` implies to me that the `waitUntilStart()` loop is active, but that doesn't seem like it would be the case if `cancel` breaks us out of that loop and then the player list changes?

Other than that, logic looks good to me.

asvitkine (approved):

game-app/game-headless/src/main/java/org/triplea/game/server/HeadlessServerSetup.java
RoiEXLab: Good point. I mainly wanted to re-use code but I can see that after cancel we probably don't want the code to un-cancel itself

RoiEXLab: But yeah, the only reason this variable exists is that any waiting threads are properly woken up when the system is shut down.

PR #10731, opened 2022-06-22
Refactor calls to Unit.IsTransporting()

frigoref:
Unit.java
-deprecate getTransporting() called from isTransporting
-add method isTransporting(Territory)
-rename fields and their getters/setters causing blocker java:S1845 (field names should not be the same or differ only by capitalization)

ProTerritory.java
-forEach loop instead of keySet loop

ProCombatMoveAi.java
ProMoveUtils.java
ProNonCombatMoveAi.java
ProTerritory.java
ProTransportUtils.java
TransportTrackerTest.java
-territory/proData.getUnitTerritory usage where possible

MoveDelegate.java
RemoveUnitsHistoryChange.java
SpecialMoveDelegate.java
TechAbilityAttachment.java
TransportTrackerTest
TransportUtilsTest.java
UnitUtils.java
UnitUtilsTest.java
-renamed getters/setters of Unit

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/Unit.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/Unit.java
@@ -62,7 +62,7 @@ public class Unit extends GameDataComponent implements DynamicallyModifiable {
   // the transport that is currently transporting us
   private Unit transportedBy = null;
   // the units we have unloaded this turn
-  private List<Unit> unloaded = List.of();
+  private List<Unit> unloadedUnits = List.of();
   // was this unit loaded this turn?
   private boolean wasLoadedThisTurn = false;
   // the territory this unit was unloaded to this turn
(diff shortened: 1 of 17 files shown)


game-app/game-core/src/main/java/games/strategy/engine/data/Unit.java
```diff
@@ -62,7 +62,7 @@ public class Unit extends GameDataComponent implements DynamicallyModifiable {
   // the transport that is currently transporting us
   private Unit transportedBy = null;
   // the units we have unloaded this turn
-  private List<Unit> unloaded = List.of();
```
asvitkine, line 65: I don't think these can change without breaking save games.

frigoref: @asvitkine Sorry, that wasn't so smart. I have've reverted it again.

asvitkine:
Looks great, thanks!

PR #10736, opened 2022-06-24
Fix an NPE.

asvitkine:


RoiEXLab:
This is odd. chatModel can only be null if the whole thing is cancelled early. So how did this happen?

PR #10738, opened 2022-06-24
Fix locking when cloning data.

asvitkine:
Fix locking when cloning data.

Since cloneGameData() internally acquires a lock, the callers don't need to hold one. This fixes an issue where the caller might hold a read lock when calling into clone, which requires a write lock, causing a self-deadlock.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
@@ -1744,7 +1744,7 @@ public final class TripleAFrame extends JFrame implements QuitHandler {
     inGame.set(false);
     setWidgetActivation();
     final GameData clonedGameData;
-    try (GameData.Unlocker ignored = data.acquireReadLock()) {
+    try (GameData.Unlocker ignored = data.acquireWriteLock()) {
       // we want to use a clone of the data, so we can make changes to it as we walk up and down the
       // history
       final var cloneOptions = GameDataManager.Options.builder().withHistory(true).build();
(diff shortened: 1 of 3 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
```diff
@@ -1744,7 +1744,7 @@ private void showHistory() {
     inGame.set(false);
     setWidgetActivation();
     final GameData clonedGameData;
-    try (GameData.Unlocker ignored = data.acquireReadLock()) {
+    try (GameData.Unlocker ignored = data.acquireWriteLock()) {
```
RoiEXLab, line 1747: Why the write lock here all of the sudden?

game-app/game-core/src/main/java/games/strategy/triplea/ui/menubar/ExportMenu.java
```diff
@@ -162,14 +162,14 @@ private void createAndSaveStats(final boolean showPhaseStats) {
...
       return;
     }
+    final var cloneOptions = GameDataManager.Options.builder().withHistory(true).build();
+    final GameData clone = GameDataUtils.cloneGameData(gameData, cloneOptions).orElse(null);
+    if (clone == null) {
+      return;
+    }
     try (PrintWriter writer =
             new PrintWriter(chooser.getSelectedFile(), StandardCharsets.UTF_8.toString());
         GameData.Unlocker ignored = gameData.acquireReadLock()) {
```
RoiEXLab, line 172: Why is a lock to gameData held here, shouldn't it use the clone?

game-app/game-core/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
asvitkine: We're removing a listener on the data, which is a write operation.

game-app/game-core/src/main/java/games/strategy/triplea/ui/menubar/ExportMenu.java
asvitkine: It still uses gameData though. Agree that it can probably use the clone though.

PR #10744, opened 2022-06-24
Avoid null message when lobby message is not set

RoiEXLab:
Just a minor thing that annoys me when using a local lobby without any data


DanVanAtta (approved):

DanVanAtta:
@RoiEXLab this is a problem for a prod 2.6 as well. Previously the `lobby_message` came from `triplea_maps.yaml`. We now instead have an Http endpoint that is now returning an equivalent JSON based on database data. Hence why the lobby message value is null. I don't think it's been really decided how to resolve that problem one way or another.

PR #10745, opened 2022-06-24
Don't call network code from UI

RoiEXLab:

This changed makes sure code that potentially is run over the network does not block the UI thread.
Also the exceptions no longer pop up in the console after this which is nice.


RoiEXLab:
Whoops wrong branch

PR #10746, opened 2022-06-24
Don't call network code from UI

RoiEXLab:

This changed makes sure code that potentially is run over the network does not block the UI thread.
Also the exceptions no longer pop up in the console after this which is nice.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/framework/network/ui/SetMapClientAction.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/framework/network/ui/SetMapClientAction.java
@@ -1,5 +1,6 @@
 package games.strategy.engine.framework.network.ui;
 
+import com.google.common.base.Preconditions;
 import games.strategy.engine.framework.startup.mc.IServerStartupRemote;
 import java.awt.Component;
 import java.util.Collection;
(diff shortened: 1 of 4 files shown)


game-app/game-core/src/main/java/games/strategy/engine/framework/network/ui/SetMapClientAction.java
```diff
@@ -50,6 +46,7 @@ public void actionPerformed(final ActionEvent e) {
     if (name == null || name.length() <= 1) {
       return;
     }
-    serverStartupRemote.changeServerGameTo(name);
+    // don't block UI thread
```
DanVanAtta, line 49: Rather than a comment - it might  be better documentation to have a 'precondition' check that the current thread is the EDT.

DanVanAtta (approved):
Looks good :+1: 
One comment to consider

game-app/game-core/src/main/java/games/strategy/engine/framework/network/ui/SetMapClientAction.java
RoiEXLab: That's a good idea

Issue #10729, opened 2022-06-22
2.5.22294: UnitsDrawer - MISSING UNIT IMAGE (won't be displayed): harbour12, UnitImageFactory.ImageKey(player=PlayerId named:Chinese,...


tripleabuilderbot:
## Map
global_40_house_rules

## Log Message
MISSING UNIT IMAGE (won't be displayed): harbour12, UnitImageFactory.ImageKey(player=PlayerId named:Chinese, type=UnitType{name=harbour12}, damaged=false, disabled=false), in Kwangtung

## TripleA Version
2.5.22294

## Java Version
11.0.6

## Operating System
Windows 10

## Memory
Heap utilization statistics [MB]
Used Memory: 1117
Free memory: 861
Total memory: 1979
Max memory: 1979

beelee1:
Hmm ... I thought I fixed this. What game version is it ?

beelee1:
well should be fixed. been a while since i did one so idk if you'll get a notification. Think I fckd up the yaml update as usual lol. Guess I'll leave it open for a day in case people can't see it

Issue #9252, opened 2021-05-20
Error trying to use 2.6.149  (Error extracting file)


beelee1:
Error extracting file: /tmp/triplea6560939776671296847/Global_40_House_Rules.zip, /tmp/triplea-unzip2861211619708353438/global_40_house_rules-master, {}  java.nio.file.DirectoryNotEmptyException: /tmp/triplea-unzip2861211619708353438/global_40_house_rules-master  If this problem happens frequently and is something you cannot fix, please report it to TripleA 

Trying to replicate this error #9250 but i can't dl map. I get a bunch of errors on startup trying to unzip all the maps in dl maps. Which maxed out my "root", which i had to spend a couple hours figuring out what and how to purge so it'd work again.

So basically, for me, This is unusable

Edit
I removed the map and tried to dl it and it'd dl but not showup. I then manually unzipped the map and put the folder directly in my dl maps and that didn't work either.

DanVanAtta:
Do you wind up with a folder named something like "Global_40_House_Rules" in your downloaded maps folder?

beelee1:
So, I tried this again and no folder showed up. I removed the zip and no folder showed up. When I choose it in DL maps, it throws the above error. It will then not show up as a DL option even though it's not in my downloaded maps folder.

I close DL maps and then it'll show up as a DL option again. I have a map I'm working on that is in a folder, no zip and it shows up and works correctly.

I just tried unzipping and loading the map manually and it shows up now. Idk y it didn't earlier. I must've screwed something up when I first tried it.

DanVanAtta:
Extraction from zip to folder occurs at startup.
If there is no zip, there is nothing to extract to a folder.

DL listing is based on the folders that are present and specifically what is listed in the 'map.yml' files in those folders.  There *could* be an issue here where at present maps do not have a 'map.yml' file, that file is generated on startup, so you could run into an inconsistent state after downloading a map and it then not having a map.yml file. If you re-open the downloader, it might not think you have the map downloaded anymore as it can't find the corresponding map.yml file (even though you have the have the map extracted), at which point you run into the above error as it'll download a zip and then try to incorrectly overwrite the previous.

To help this from a game engine perspective, we can do the following:
1. We need to finally get map.yml files uploaded to all maps
2. The engine should be checking if the map folder already exists and give a better error message or just go ahead and overwrite it. Probably overwrite it as we want the downloader to update maps and not just install maps.

beelee1:
[...]

PR #10749, opened 2022-06-25
Fixing Global 40 House Rules 2.72 Update

beelee1:
Hopefully lol
Fixes Issue #10729

Edit
Is there a way to change the default so it doesn't automatically update without a PR ? I keep forgetting and hit enter without switching it. Seems the default should be the PR one


beelee1:
@asvitkine while this PR works with the 2.5 stable, it doesn't with 2.6.838 It just DLs to the previous game version even if I manually move it to the DL'd maps. Think it's related to #9252

Let me know if I should reopen #9252

PR #10757, opened 2022-06-25
Replace latches with CompletableFuture

RoiEXLab:

Prefer CompletableFuture over own latch logic. I originally wanted to remove the other usage of `CountDownLatch` as well to be able to get rid of the class completely, but the other instance seems to be a little bit weird. It's basically being used as a mutex over 2 threads, but I couldn't come up with a good way to code that using objects from the standard lib. If you have good ideas feel free to tell me.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
@@ -10,13 +10,14 @@ import games.strategy.engine.framework.GameDataManager;
 import games.strategy.engine.framework.GameDataUtils;
 import java.util.Collection;
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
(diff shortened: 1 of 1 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
```diff
@@ -79,12 +83,15 @@ public void setGameData(final GameData data) {
...
-                throwable -> log.error("Error when trying to create Workers", throwable));
+        // assign the future,
+        // so that we can set the data in a different thread wait for its completion
+        latchWorkerThreadsCreation =
+            CompletableFuture.runAsync(() -> createWorkers(data))
+                .exceptionally(
+                    throwable -> {
+                      log.error("Error when trying to create Workers", throwable);
+                      return null;
+                    });
```
asvitkine, line 94: The exceptionally block is pretty verbose with CompletableFuture. WDYT of adding an overload that doesn't require a return value so it can be as concise as the one with AsyncRunner.runAsync()?

asvitkine (approved):

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/ConcurrentBattleCalculator.java
RoiEXLab: To be honest I almost thing this should be made synchronous. If the callers support it this would simplify the code quite a lot

PR #10761, opened 2022-06-25
Fix regression introduced in #10709

### LINKED ISSUE #10734: Regression: 2.6 sets human players to Weak AI in network host screen.
### Reproduction Steps

1. Start a local game on Imperialism 1974 Board Game. Observe this screen:

image

2. Start a network game on the same map. Observe this screen:

image

Notably, the XML for this map is:

```
		<player name="Blue" optional="true" canBeDisabled="true"/>
		<player name="Red" optional="true" canBeDisabled="true"/>
		<player name="Green" optional="true" canBeDisabled="true"/>
		<player name="Yellow" optional="true" canBeDisabled="true"/>
		<player name="Orange" optional="true" canBeDisabled="true"/>
		<player name="Cyan" optional="true" canBeDisabled="true"/>
		<player name="Purple" optional="true" canBeDisabled="true"/>
		<player name="Brown" optional="true" canBeDisabled="true"/>
		<player name="Neutrals" optional="true" defaultType="DoesNothing" isHidden="true"/>
```
So, the problems on the network screen are:
  1. WeakAI set as the non-Neutrals players, instead of Human.
  2. HardAI instead of "Does Nothing AI" set as the Neutrals player.

### Engine Version
The above screenshots are from 2.6 prerelease.

When using 2.5, the network game screen is as follows instead:
Screen Shot 2022-06-23 at 11 20 27 PM

So the WeakAI change in 2.6 is incorrect, and both 2.5 and 2.6 are setting the `defaultType="DoesNothing"` to "Hard AI" incorrectly.

RoiEXLab:

Closes #10734
The "null" mapping meant that the first item in the list was selected by default and the human player was no longer the first item after this change.

### DIFF
--- a/game-app/game-headed/src/main/java/games/strategy/engine/framework/startup/mc/HeadedPlayerTypes.java
+++ b/game-app/game-headed/src/main/java/games/strategy/engine/framework/startup/mc/HeadedPlayerTypes.java
@@ -58,8 +58,9 @@ public class HeadedPlayerTypes {
 
   public static Collection<PlayerTypes.Type> getPlayerTypes() {
     return Stream.of(
+            List.of(HUMAN_PLAYER),
             PlayerTypes.getBuiltInPlayerTypes(),
-            List.of(HUMAN_PLAYER, CLIENT_PLAYER, getDoesNothingType(), getFlowFieldType()))
+            List.of(CLIENT_PLAYER, getDoesNothingType(), getFlowFieldType()))
         .flatMap(Collection::stream)
         .filter(HeadedPlayerTypes::filterBetaPlayerType)
         .collect(Collectors.toList());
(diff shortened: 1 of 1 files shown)


DanVanAtta (approved):

game-app/game-headed/src/main/java/games/strategy/engine/framework/startup/mc/HeadedPlayerTypes.java
```diff
@@ -58,8 +58,9 @@ public Player newPlayerWithName(String name) {
 
   public static Collection<PlayerTypes.Type> getPlayerTypes() {
     return Stream.of(
+            List.of(HUMAN_PLAYER),
```
asvitkine, line 61: Can you add comments about the importance of the order?

PR #10776, opened 2022-06-26
antiDeprecated - DebugMenu.registerMenuCallback

frigoref:
Could someone check if this break backwarts-compatibility and if so how?

DebugMenu.java
-remove deprecated method registerMenuCallback(name, factory) and member variable menuItemsAndFactories
-add (instead) member variable frameFactories and method registerFrameFactory
-add overloaded method registerDebugOptions(playerName, options)
-enhance method renderDebugOption to allow menu item mnemonic

AiPlayerDebugOption.java
-add member variable mnemonic

ProLogUi.java
-replace call to DebugMenu.registerMenuCallback with calls to registerDebugOptions and registerFrameFactory


DanVanAtta:
Should be good for backward compatibility. It is a concern for save games when modifying private fields of classes that extend `GameDataComponent`. There is backward compatibility concerns for network communication as well which is usually a bit easier to spot.

game-app/game-core/src/main/java/games/strategy/triplea/ui/menubar/DebugMenu.java
```diff
@@ -26,9 +25,7 @@
 public final class DebugMenu extends JMenu {
   private static final long serialVersionUID = -4876915214715298132L;
 
-  /** Maps the debug menu title to the function that will create the sub menu items */
-  private static final Map<String, Function<TripleAFrame, Collection<JMenuItem>>>
-      menuItemsAndFactories = new TreeMap<>();
+  private static final List<Consumer<TripleAFrame>> frameFactories = new ArrayList<>();
```
DanVanAtta, line 28: Consider renaming this variable from `frameFactories` to `frameVisitors`. `factories` really caught my attention as a name - I think I realized it is because we are not really using the factory pattern here but instead are using the 'Visitor' pattern (or at least, I think we are, looks like visitor :smile: :wave: :person_fencing: )

DanVanAtta (approved):
LGTM
One smaller comment for you to consider @frigoref .

game-app/game-core/src/main/java/games/strategy/triplea/ui/menubar/DebugMenu.java
frigoref: Thanks, I also had a problem with that, but didn't want to change the name as I wasn't sure why it was chose in the first place.
It's corrected now.

PR #10778, opened 2022-06-26
Remove getPropertyMap() on attachments in favor of getPropertyOrNull().

asvitkine:
Avoid needing to create temporary immutable maps via getPropertyMap(), which shows up in profiles when loading maps.

The idea is we should remove all implementations of `getPropertyMap()`, instead having attachment classes implement `getPropertyOrNull()`, which doesn't require creating massive temporary maps. As a result, callers won't pay the cost of constructing these maps. Instead, they query the property they need which uses a switch statement instead of a map to look up properties, which is probably OK.

The idea is we could get rid of this cost during map parsing (e.g. 0.5s in this case on Imperialism 1974):
Screen Shot 2022-06-26 at 4 30 18 PM

With this change, the profile looks as follows:
Screen Shot 2022-06-26 at 4 24 54 PM

Note the time used by `getPropertyMap()` goes away.

Additionally, the new version is a reduction in LOC.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/DynamicallyModifiable.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/DynamicallyModifiable.java
@@ -32,7 +32,7 @@ public interface DynamicallyModifiable {
   default Optional<MutableProperty<?>> getProperty(final String name) {
     checkNotNull(name);
 
-    return Optional.ofNullable(((TriggerAttachment) this).getPropertyMapFunction().apply(name));
+    return Optional.ofNullable(getPropertyMapFunction().apply(name));
   }
 
   default Function<String, MutableProperty<?>> getPropertyMapFunction() {
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/DynamicallyModifiable.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/DynamicallyModifiable.java
@@ -2,10 +2,7 @@ package games.strategy.engine.data;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import games.strategy.triplea.attachments.TriggerAttachment;
-import java.util.Map;
 import java.util.Optional;
-import java.util.function.Function;
 
 /**
  * An interface to implement by objects that are dynamically being modified. This will most likely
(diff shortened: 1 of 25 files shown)


asvitkine:
@RoiEXLab WDYT?

game-app/game-core/src/main/java/games/strategy/triplea/attachments/TriggerAttachment.java
```diff
@@ -2745,4 +2746,158 @@ public Map<String, MutableProperty<?>> getPropertyMap() {
                 this::resetTerritoryEffectProperty))
         .build();
   }
+
+  @Override
+  public Function<String, MutableProperty<?>> getPropertyMapFunction() {
```
asvitkine, line 2751: Another benefit is this method body is shorter than getPropertyMap() above it, so if we do the replacement, we also make the code less verbose.

asvitkine:
I also checked that no callsites of getPropertyMap() care about .size() or iterating all the elements, which wouldn't work with this approach. Nothing needs that.

We could even simplify further and instead of having something that returns a mapper function, we can just have the implementation live in `getProperty()`.

asvitkine:
> We could even simplify further and instead of having something that returns a mapper function, we can just have the implementation live in `getProperty()`.

Went with this approach and update the PR description. @RoiEXLab WDYT?

game-app/game-core/src/main/java/games/strategy/engine/data/DynamicallyModifiable.java
```diff
@@ -46,4 +45,9 @@ default MutableProperty<?> getPropertyOrThrow(final String name) {
     return getProperty(name)
         .orElseThrow(() -> new IllegalArgumentException("unknown property named '" + name + "'"));
   }
+
+  default void setPropertyOrThrow(String name, String value)
+      throws MutableProperty.InvalidValueException {
+    getPropertyOrThrow(name).setValue(value);
```
RoiEXLab, line 51: This is only used by test code right? I think it would be better to move this helper method into test code

RoiEXLab (commented):
@asvitkine This approach seems perfectly fine.
To give you some backstory to this: Previously those maps didn't exist and the code solely relied on reflection to find the correct methods. This meant that it was impossible to fix spelling mistakes or perform minor refactoring changes without silently breaking the parser. IIRC I used some RegEx magic to scrape all of the methods from the attachment classes and wrote a script that generated the code to create those giant maps for me. So this basically achieves the same thing without the overhead. Seems like a good approach.

game-app/game-core/src/main/java/games/strategy/engine/data/DynamicallyModifiable.java
asvitkine: Done.

asvitkine:
@RoiEXLab Thanks for the context, I've updated all the attachment classes now.

PR #10780, opened 2022-06-26
Drop code-climate

DanVanAtta:
code-climate is a static analysis tool that automatically adds comments in PRs
to flag static check violations. We can customize things like line counts,
arg counts etc.. that constitute violations.

While such checks are really useful, on the whole it is causing more noise
than is helping. Most updates are to existing code that typically has several
violations in almost any module that is being updated. So, we tend to ignore
any flagged violations, or we are improving the violations but still get flagged
due to being above threshold.

If there were a way to have code climate run for only exactly modified code,
or if we had a new code base, then it would be much more useful. Otherwise
it is important to avoid PR checks that can be ignored, a 'red build' should
generally mean 'not suitable for merge'


RoiEXLab:
Agreed. Code-climate is nice to have, but especially when moving files or refactoring code it complains about code that was only indirectly touched quite often unfortunately.

asvitkine:
It's definitely true that it picks up a lot of existing issues, and some of the findings are dubious as well (e.g. it will complain if you go above 20 methods, if you split up a function that's above its 50 LOC threshold but is still above 50 LOC afterward).

Having said that, I did try to address many of its findings when it flagged them during my refactoring. So it did provide a bit of an incentive to improve things, but the signal to noise ratio definitely wasn't great.

Of all the checks, I find the code duplication ones the most valuable - since these are otherwise not very obvious to a human (since code may be somewhere else in the codebase). The other stuff, like deep nesting, long methods, number of methods, etc are already pretty clear from the immediate code and hard thresholds on these weren't super helpful.

Could we just keep the duplicate code warnings and disable everything else?

DanVanAtta:
@asvitkine would you mind checking if PMD can do the same? If that's not possible, I am sympathetic for the one useful check, though I have pause whether it's really worth the CI/CD complexity (I'm not sure).

asvitkine:
> @asvitkine would you mind checking if PMD can do the same? If that's not possible, I am sympathetic for the one useful check, though I have pause whether it's really worth the CI/CD complexity (I'm not sure).

Looks like PMD has something like this: https://pmd.github.io/latest/pmd_userdocs_cpd.html

However, I'm not enough of a gradle expert (and don't have time for that right now) to figure out how to make it work with our build.

I'm not in love enough with codeclimate's duplicate code finding to push against removing codeclimate altogether, especially if other tools exist.

asvitkine:
Thanks. I think making AI know how to do amphib landings past a sea battle would be a nice improvement, but I guess in this case, we could probably also settle for just a ships' arm race, which would allow the AI with the greater production to push through (over several turns).

PR #10765, opened 2022-06-25
Added Darkest Hour

RogerCooper:

PR #10781, opened 2022-06-26
Disconnect github action to update maps on website

DanVanAtta:
We have some configuration to automatically split the 'triplea_maps.yaml'
file (which indexes avaialable maps) and publishes these map descriptions
to the TripleA website.

This has a few problems:
1. We are quickly moving away from 'triplea_maps.yaml' and will need
a different mechanism if we wish to continue updating maps on website.

2. The website receives exceedingly low traffic, it's likely nobody
has even looked at the maps listing on the website for a few weeks.
The value to complexity ratio is seemingly unfavorable for even building
a replacement, which is all more argument for removing the current legacy
implementation.


RoiEXLab:
@DanVanAtta Honestly no strong feelings about this change, even though I created the script a long time ago, but I'd vote to keep the action there until we have officially moved away from the yaml file. The yaml file is still updated semi-regularly so I think this is nice to keep the descriptions on the website up-to-date until then.

RoiEXLab:
Also I think keeping it just a little longer doesn't hurt. I don't think this particular "feature" had some real maintenance cost in the past few years, right?

DanVanAtta:
> but I'd vote to keep the action there until we have officially moved away from the yaml file

We basically are there. I don't think there will be that many more maps updates that we would miss. Further though,  there is a question what to do on the website side. The traffic is pretty anemic, it's not clear we are getting value from the maps listing page.

All in all, it's all the more reason to go ahead and drop this legacy mechanism.

RoiEXLab:
> We basically are there.

If you say so, I didn't keep track of the progress so far.

> I don't think there will be that many more maps updates that we would miss.

I think we can wait for #10765 to get merged before :D

> Further though, there is a question what to do on the website side. The traffic is pretty anemic, it's not clear we are getting value from the maps listing page.

That's a separate concern in my opinion. The website is supposed to be a show reel for the game right? I don't think we need a whole lot of traffic to justify its existence, but I see your point that the maps are just a minor contribution at best to this.

DanVanAtta:
> That's a separate concern in my opinion. The website is supposed to be a show reel for the game right? I don't think we need a whole lot of traffic to justify its existence, but I see your point that the maps are just a minor contribution at best to this.

Generally agree. My point is that the arguably low value on the website is further reason to remove the integration.

And to be sure, just talking about the 'maps' part of the website  : )

I think I might be personally convinced that as far as a show reel, we could do it more effectively than with 'maps' page, and very often "more is not more".

DanVanAtta:
@RoiEXLab ,so, we can keep the integration for moments longer.. Do we intend to re-create this integration from the live maps-server? It would not be too difficult per-say. I am not inclined though because it does not seem to be clearly worth the complexity. Does that mean this integration will soon be orphaned with no plan to replace it?

RoiEXLab:
@DanVanAtta I mean if you want i can recreate this for the live maps server no problem.
The easiest approach would be to create a GitHub action on the website repo that runs every day or so. This way the code only relies on the HTTP API not on the underlying implementation.
Alternatively we could also implement this directly into the code whenever the database is updated, but I think even though it is more efficient it's the worse choice overall

DanVanAtta:
Let's continue this convo in issue#https://github.com/triplea-game/triplea/issues/10834

Issue #10785, opened 2022-06-26
Another instance of AiGameTest going to 100 turns with no winner


asvitkine:
Another instance of AiGameTest going to 100 turns with no winner.

Here's a save game:
[autosave_round_odd_new100.tsvg.zip](https://github.com/triplea-game/triplea/files/8987525/autosave_round_odd_new100.tsvg.zip)

Notably, it seems that the game state at the 100 turn mark is such that Germany can be defeated if the opponent just decides to load up all its units and invade, sending two fighters to the naval combat that needs to be cleared before landing. However, this does not happen, even if you run the game for 40+ more turns.

Need to figure out why AI is not choosing this option. 
<img width="489" alt="Screen Shot 2022-06-26 at 5 15 04 PM" src="https://user-images.githubusercontent.com/17648/175834084-948e20a0-6250-41bd-b358-83e7547f34e8.png">

asvitkine:
The position really didn't change in the last 50 turns, though the odds for the ground combat increased over time. On turn 55, for example, it was already a 87% win rate for the invasion. Navy didn't change since.

asvitkine:
Additionally, the AI doesn't care to capture the German sea zones, which could have let it get a bigger income advantage.

asvitkine:
So ignoring how the AI got in this situation in the first place and even the sea zones, the question is why does it cause a stalemate? If AI thinks the sea combat is too risky, it should buy more ships or planes. Otherwise, it should see it could just attack through and win.

asvitkine:
So if I make Germany's capital completely, empty, the AI still doesn't attack it.
<img width="491" alt="Screen Shot 2022-06-26 at 9 34 33 PM" src="https://user-images.githubusercontent.com/17648/175844368-2e240d97-e396-4c33-9a73-819b80c42002.png">

If I add 10 battleships to accompany the fleet of 92 transports, then they do attack the German fleet that turn, but don't move transports to land units at the same time.

Could it be that AI doesn't know how to do amphib attacks if there's a dependent sea battle to fight first? @ron-murhammer, do you know?

ron-murhammer:
@asvitkine I don't think the AI does amphib after a sea battle (not 100% sure but at least I know it didn't for a long time). If I remember, the main reason is its complex to make sure you have good enough odds on the sea battle and don't take transports as casualties so the landing is then successful.

asvitkine:
@ron-murhammer 

[...]

PR #10797, opened 2022-06-27
ProAI: Improve territoryHasLocalNavalSuperiority().

asvitkine:
In addition to estimating strength differences, check that we also have a positive TUV to attack the strongest defensive enemy flee that's nearby.
This prevents a situation where the AI does not expand its fleet yet does not attack due to its fleet not being strong enough.

An example of this is on the save here on: https://github.com/triplea-game/triplea/issues/10785, where this situation causes a stalemate. With this fix, I confirmed that the above game gets out of the stalemate with the expected result of the AI with more production eventually winning.

It also gets out of the stalemate if I save the above game at turn 50, winning in less than 10 turns: 
[autosave_round_odd_new100-50.tsvg.zip](https://github.com/triplea-game/triplea/files/8997593/autosave_round_odd_new100-50.tsvg.zip)

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProBattleUtils.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProBattleUtils.java
@@ -377,6 +377,7 @@ public final class ProBattleUtils {
     final List<Unit> enemyAttackers = new ArrayList<>(enemyUnitsInSeaTerritories);
     enemyAttackers.addAll(enemyUnitsInLandTerritories);
     final double defenseStrengthDifference = estimateStrengthDifference(t, enemyAttackers, myUnits);
+    boolean hasSuperiority = (defenseStrengthDifference < 50);
     ProLogger.trace(
         t
             + ", current enemy naval attack strengthDifference="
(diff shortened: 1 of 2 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProBattleUtils.java
```diff
@@ -391,8 +395,26 @@ public static boolean territoryHasLocalNavalSuperiority(
...
+      // To really have naval attack superiority, ensure there's also a positive attack TUV for
+      // all nearby territories. Otherwise, prioritizeAttackOptions() will exclude such attacks,
+      // thus coming short of actually having naval superiority.
+      for (final Territory nearby : nearbyEnemySeaTerritories) {
+        final List<Unit> enemySeaUnits = nearby.getUnitCollection().getMatches(enemyNonLandUnit);
+        if (enemySeaUnits.isEmpty()) {
+          continue;
+        }
+        ProBattleResult result =
+            calc.estimateAttackBattleResults(proData, t, myUnits, enemySeaUnits, List.of());
```
ron-murhammer, line 409: Not sure how often `territoryHasLocalNavalSuperiority` is called and what the size of `nearbyEnemySeaTerritories` can be but you'll want to make sure this doesn't have a negative performance impact as I believe this method can call the battle calc.

asvitkine: Good point. I've now changed it to just to a battle calc against the strongest defensive enemy fleet only, which should be sufficient.

PR #10810, opened 2022-06-30
UI improvements to production dialog, especially with long unit lists.

asvitkine:
UI improvements to production dialog, especially with long unit lists.

This PR makes the following improvements:
  - Caps the number of columns to the available screen space, so there's never a horizontal scroll bar.
  - Lays out units left-to-right horizontally, before spilling over and continuing vertically. Previously, it would fill up columns, top to bottom, which was not an optimal use of space.
  - Does not create multiple components for the same production rule, instead keeping track of them and updating the content area when tabs are switched. This should improve performance as opening the production dialog on maps with a lot of units like Another World can take some time due to building all the swing UI.
  - Moves the legend "Attack | Defense | Movement" legend label to the bottom right corner to make the UI more compact.
  - Some additional clean ups, such as removing null checks for consume units (never null).

Example after screenshots on Another World:
Screen Shot 2022-06-30 at 12 15 00 AM
Screen Shot 2022-06-30 at 12 15 04 AM

Before screenshots on same map & same screen size:
Screen Shot 2022-06-30 at 12 01 41 AM
Screen Shot 2022-06-30 at 12 01 45 AM

## Release Note

CHANGE|Improved layout on the production dialog when there are many units


beelee1:
Big Thumbs Up. I have a couple mods with lots of units. I don't like the sideways scroll

game-app/game-core/src/main/java/games/strategy/triplea/ui/TabbedProductionPanel.java
```diff
@@ -46,95 +53,66 @@ static IntegerMap<ProductionRule> getProduction(
...
-            GridBagConstraints.BOTH,
-            new Insets(8, 8, 8, 8),
-            0,
-            0));
+    add(tabs, BorderLayout.CENTER);
     final ProductionTabsProperties properties =
-        ProductionTabsProperties.getInstance(gamePlayer, rules, uiContext.getResourceLoader());
+        new ProductionTabsProperties(gamePlayer, rules, uiContext.getResourceLoader());
     final List<Tuple<String, List<Rule>>> ruleLists = getRuleLists(properties);
+    ruleLists.removeIf(entry -> entry.getSecond().isEmpty());
```
DanVanAtta, line 62: Consider:
- a comment here on the `removeIf` to explain why we are doing the removal
- or move the `removeIf` to  'getRuleLists' (might not be safe / scope creep risk) 
- and/or using a `stream+filter`

To explain why, a reader may wonder:
- why (L61) `getRuleLists(properties)` is returning data that needs to be mutated on (L62)?
-  A reader may also wonder if this is safe to do (mutating data returned from a 'getter'  is just never safe in TripleA)

To explain what I mean by 'stream+filter', that would look like this:
`getRuleLists(...).stream().filter(entry -> entry.getSecond().isEmpty()).collect(Collectors.toList())`

game-app/game-core/src/main/java/games/strategy/triplea/ui/TabbedProductionPanel.java
```diff
@@ -46,95 +53,66 @@ static IntegerMap<ProductionRule> getProduction(
...
+    cellSize = new Dimension();
+    for (Rule rule : rules) {
+      JPanel component = rule.getPanelComponent();
+      rulesComponents.put(rule, component);
+      cellSize.width = Math.max(cellSize.width, component.getPreferredSize().width);
+      cellSize.height = Math.max(cellSize.height, component.getPreferredSize().height);
+    }
```
DanVanAtta, line 69: This block gets a bit dense, consider adding a guiding comment to help explain why we are picking max sizes. It's not immediately clear why one is better than the other, and what the overall goal is.

game-app/game-core/src/main/java/games/strategy/triplea/ui/TabbedProductionPanel.java
```diff
@@ -46,95 +53,66 @@ static IntegerMap<ProductionRule> getProduction(
...
       if (largestList <= 36) {
         maxColumns = Math.max(8, Math.min(12, divideRoundUp(largestList, 3)));
       } else if (largestList <= 64) {
         maxColumns = Math.max(8, Math.min(16, divideRoundUp(largestList, 4)));
       } else {
         maxColumns = Math.max(8, Math.min(16, divideRoundUp(largestList, 5)));
       }
+      // Limit number of columns to the available width. Reserve 64px for borders & scroll bar.
+      final int availableWidth = Util.getScreenSize(dialog).width - 64;
+      maxColumns = Math.min(maxColumns, availableWidth / cellSize.width);
```
DanVanAtta, line 115: Minor, but I wonder why not use a new variable at L115 rather than mutating an existing value?

Nitpicky - this could be a place where we can favor immutabilty: [avoid mutability convention](https://github.com/triplea-game/triplea/blob/master/docs/development/reference/code-conventions/java-code-conventions.md#avoid-mutability)

Overall more descriptive variable naming might help how the code flows for the reader. Suggestions:

L105 `maxColumns` -> `desiredColumns`
L114 `availableWidth` -> `availableScreenWidth`

The above suggestions/convention nit are certainly pretty trivial concerns compared to the complexity of the existing code that makes a variable renaming exercise frustrating. Specifically, it's really hard to know what the first `maxColumns` really represents, what those calculations are, and so a better name is hard to determine. A better name would describe perhaps something more regarding the output of those calculations. That would be an improvement to the biggest problem here where it is not clear what that `if/else` block is doing..

DanVanAtta (approved):
Few comments to consider, nothing major. Good UX update :+1:

game-app/game-core/src/main/java/games/strategy/triplea/ui/TabbedProductionPanel.java
asvitkine: Done.

game-app/game-core/src/main/java/games/strategy/triplea/ui/TabbedProductionPanel.java
asvitkine: Done. Folded it into `getRuleLists()`.

beelee1:
This is a major improvement

Issue #10823, opened 2022-07-02
AiGameTest fails in GameTestUtils


frigoref:
AiGameTest fails in GameTestUtils

    // Use a temp dir for downloaded maps to not interfere with the real downloadedMaps folder.
    if (tempHome == null) {
      tempHome = FileUtils.newTempFolder();
      System.setProperty("user.home", tempHome.toString());
    }
    ClientSetting.initialize();
    assertTrue(ClientFileSystemHelper.getUserMapsFolder().startsWith(tempHome.toAbsolutePath()));
    ClientSetting.aiMovePauseDuration.setValue(0);
    ClientSetting.aiCombatStepPauseDuration.setValue(0);

The comment seems to contradict the call for assertTrue. I wonder if the assert is actually needed here as the temp folder will always be different as far as I can see.

@asvitkine Any explanation why this could make any sense?

asvitkine:
@frigoref 

The test works on the bots and on both my computers (Mac and Windows). We can try to figure out why it fails for you. What's your set up?

> The comment seems to contradict the call for assertTrue. I wonder if the assert is actually needed here as the temp folder will always be different as far as I can see.

The temp folder is set as "user.home" which is used during `ClientSetting.initialize()` to set the `ClientFileSystemHelper.getUserMapsFolder()`, so it's expected that the user map folder is a subdir of `tempHome`. If you print both paths, what do you get?

frigoref:
`ClientFileSystemHelper::getUserRootFolder` returns the mapFolder path:
![image](https://user-images.githubusercontent.com/10353640/177012914-f6c06577-e63a-4aa4-93da-103c669f4379.png)
Seems that because I have a valid maps folder this one is used instead of the temporary folder.

asvitkine:
The relevant code is:

```java
  public static Path getUserRootFolder() {
    final Path userHome = Path.of(SystemProperties.getUserHome());
    final Path rootDir = userHome.resolve("Documents").resolve(USER_ROOT_FOLDER_NAME);
    return Files.exists(rootDir) ? rootDir : userHome.resolve(USER_ROOT_FOLDER_NAME);
  }
```

And
```java
  public static String getUserHome() {
    return checkNotNull(System.getProperty("user.home"));
  }
```

Can you check why the above isn't working as expected for you? `System.getProperty("user.home")` should be getting set to the temp folder created.

asvitkine:
Possibly you have `ClientSetting.mapFolderOverride` set?

[...]

PR #10824, opened 2022-07-03
Don't use real preferences for AiGameTest.

asvitkine:
Fixes https://github.com/triplea-game/triplea/issues/10823.


smoke-testing/src/test/java/games/strategy/engine/data/GameTestUtils.java
```diff
@@ -46,6 +47,7 @@ public static void setUp() throws IOException {
       System.setProperty("user.home", tempHome.toString());
     }
     ClientSetting.initialize();
```
RoiEXLab, line 49: I think the initialize is redundant if setPreferences is called afterwards, right?

asvitkine: It has some locale logic:
```
    final Locale defaultLocale = Locale.getDefault();
    if (!I18nResourceBundle.getMapSupportedLocales().contains(defaultLocale)) {
      Locale.setDefault(Locale.US);
    }
```

RoiEXLab: Ah I see. Didn't realize this was initialized there (imo it should be initialized independent of this class, does not have much to do with the client settings).
The error message is still treating them as equivalent though:
`ClientSetting framework has not been initialized. Did you forget to call ClientSetting#initialize() in production code or ClientSetting#setPreferences() in test code?`
So either the message should be updated, or the code moved somewhere else to be called explicitly.

asvitkine: Fair enough. Since there were only two other call sites of `initialize()`, I just inlined that locale logic into them.

The test worked fine without it, so I just left it out.

Issue #10836, opened 2022-07-06
2.6+0cce482: ViewMenu#lambda$addZoomMenu$4:122 - java.lang.ClassCastException


tripleabuilderbot:
## User Description
This is the stack trace for #10832

## Map
global_40_expansion

## Log Message
class java.lang.Integer cannot be cast to class java.lang.Double (java.lang.Integer and java.lang.Double are in module java.base of loader 'bootstrap')

## TripleA Version
2.6+0cce482

## Java Version
11.0.15

## Operating System
Linux

## Stack Trace
```
Exception: java.lang.ClassCastException class java.lang.Integer cannot be cast to class java.lang.Double (java.lang.Integer and java.lang.Double are in module java.base of loader 'bootstrap')
java.lang.Exception
	at java.base/java.lang.Double.compareTo(Double.java:49)
	at java.desktop/javax.swing.SpinnerNumberModel.incrValue(SpinnerNumberModel.java:362)
	at java.desktop/javax.swing.SpinnerNumberModel.getPreviousValue(SpinnerNumberModel.java:401)
	at java.desktop/javax.swing.plaf.basic.BasicSpinnerUI.updateEnabledState(BasicSpinnerUI.java:514)
	at java.desktop/javax.swing.plaf.basic.BasicSpinnerUI.updateEnabledState(BasicSpinnerUI.java:494)
	at java.desktop/javax.swing.plaf.basic.BasicSpinnerUI.installUI(BasicSpinnerUI.java:123)
	at org.pushingpixels.substance.internal.ui.SubstanceSpinnerUI.installUI(SubstanceSpinnerUI.java:78)
	at java.desktop/javax.swing.JComponent.setUI(JComponent.java:685)
	at java.desktop/javax.swing.JSpinner.setUI(JSpinner.java:189)
	at java.desktop/javax.swing.JSpinner.updateUI(JSpinner.java:214)
	at java.desktop/javax.swing.JSpinner.<init>(JSpinner.java:159)
	at games.strategy.triplea.ui.menubar.ViewMenu.lambda$addZoomMenu$4(ViewMenu.java:122)
	at org.triplea.swing.SwingAction$1.actionPerformed(SwingAction.java:60)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1967)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2308)
[...]

PR #10846, opened 2022-07-07
Fix map zoom exception by using a double value.

asvitkine:
Fixes: https://github.com/triplea-game/triplea/issues/10836

Not sure if the bug is specific to substance L&F, but the fix seems to work.


asvitkine:
Also filed https://github.com/kirill-grouchnikov/radiance against subtance.

PR #10874, opened 2022-07-13
Revert "Infra: use commit number as build number"

DanVanAtta:
Reverts triplea-game/triplea#10869


asvitkine:
I must have missed the context. Why was this reverted?

I thought it was a very elegant solution!

PR #10877, opened 2022-07-13
Move TripleAFrame to game-headed

RoiEXLab:

This PR moves UI related classes like `TripleAFrame` to game-headed along with some related ckasses


game-app/game-headed/src/main/java/games/strategy/triplea/ui/menubar/DebugMenu.java
```diff
@@ -38,43 +28,27 @@ public final class DebugMenu extends JMenu {
...
-              renderDebugOption(option.getOptions()).forEach(playerDebugMenu::add);
-            });
-    if (frameVisitors.isEmpty() && debugOptions.isEmpty()) {
+    DebugMenuInfo.runForOption(
+        (name, options) -> {
+          final JMenu playerDebugMenu = new JMenu(name);
+          add(playerDebugMenu);
+          renderDebugOption(options).forEach(playerDebugMenu::add);
+        });
+    if (DebugMenuInfo.isEmpty()) {
```
asvitkine, line 39: I don't love that we're now relying on global state in DebugMenuInfo.

Is there a way to structure this so that the DebugMenuInfo is an object with non-static methods on it that's passed to the constructor of DebugMenu? Then, we wouldn't be introducing and relying on global state that has questionable lifetime.

game-app/game-core/src/main/java/games/strategy/triplea/ui/AdditionalTerritoryDetails.java
```diff
@@ -0,0 +1,26 @@
+package games.strategy.triplea.ui;
+
+import games.strategy.engine.data.Territory;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class AdditionalTerritoryDetails {
+  private final List<Function<Territory, String>> additionalTerritoryDetailFunctions =
```
asvitkine, line 10: Since this is already a member of AdditionalTerritoryDetails, I don't think we need to spell out the full class name here again. How about just `detailProviders` and  renaming the two methods to add/remove them accordingly?

game-app/game-headed/src/main/java/games/strategy/triplea/ui/menubar/DebugMenu.java
RoiEXLab: I don't like it either, but this is what the class was already doing before my changes, it just was part of the same class. If you have a better idea I'd be happy to implement it. For now I just kept is as-is and added a separate class so the code compiles

game-app/game-core/src/main/java/games/strategy/triplea/ui/AdditionalTerritoryDetails.java
RoiEXLab: Sounds good

game-app/game-headed/src/main/java/games/strategy/triplea/ui/menubar/DebugMenu.java
asvitkine: I guess my thought is if we're refactoring the code anyway, why propagate the bad pattern?

Looking at how it's called, right now `ProLogUi.registerDebugMenu();` is called from `AbstractProAi()`.

I guess the idea is if there's any ProAI, we get a debug menu. Perhaps this can be done by DebugMenu itself? It already takes a TripleAFrame object, from which we should be able to get access to the game data and see if there's any Pro AI players and call some code from ProLogUI to make the Hard AI menu. I don't know if we even need this "registration" mechanism. We could just create the relevant menus at the time when we create the menu bar, based on relevant conditions ("is there a pro ai")?

RoiEXLab: Good point didn't think about that. I'm worried about merge conflicts though if this stays open for too long so if this is simple to implement we can ship it within this PR otherwise a follow-up PR would also be an option. I believe you know more about the whole environment so it's your call

asvitkine: I agree that merge conflicts could be a problem, BUT:
  1. I personally don't have any planned changes right now to UI
  2. We're not currently getting other contributions to UI code either

So given that, I think it would be good to try to give this a shot as part of this PR. It would be a great clean up over the current state and it seems now is a good time to do it given your refactoring is running into this. Do you want to give it a shot? If it ends up being more involved than we're thinking, we could still reconsider.

RoiEXLab: If you can give me a starting point where the `TripleAFrame` stores information about AI I would give it a shot. Might take me a couple of days until I get to it though.

asvitkine: I think `TripleAFrame` holds `GameData` from which you can get the `PlayerList` and iterate over `GamePlayers`. Then you can call `getPlayerType()` which should could be used. Alternatively, it also has access to `LocalPlayers` which is a collection of `Player` objects, which you can check for `instanceof AbstractProAi`. I think the latter is probably the better way to do it. Hope that helps!

RoiEXLab:
@asvitkine So it turns out this whole thing is a little bit trickier than we thought. Your solution would have worked flawlessly if we only look at `AbstractProAi` registering the debug menu, however `FlowFieldAi` uses it too and it registers it for every single instance of its class to be able to display debug information about what the AI is currently thinking.

There are basically 2 problems that need to be solved:
1. AI logging: Currently the code initialized a static swing dialog that the AI code can append logging messages to at any point in the code. I was able to initialize this frame directly without the listener indirection, but what should really happen instead is that the pro AI is initialized with a "logging interface" that delegates to the current UI implementation to avoid the strict dependency.
2. AI-specific debug menus: AI development is tricky so there needs to be a way to show some debug information for developers and this debug information needs to be AI specific, so this sounds like the AI class (or player interface, not sure what's better here) needs a public method that allows the `DebugMenu` class to create debug menus according to the `LocalPlayers`. However I think AI implementations should not create swing elements, they should work regardless of the UI framework being used.

Those 2 problems somewhat overlap. If the latter problem is solved the former becomes easier to solve I think, but that's the point where I stopped because this feels definitely out of scope for this PR.

asvitkine:
If you'd like to make it work for hard ai, we can temporarily unhook the flow ai one and I can look at making it work for that one as a follow up pr.

RoiEXLab:
@asvitkine This week at least my free time is very limited, so I'd like to call it a day here. The code is still a minor improvement over the current state. If you are convinced you can fix the existing problems with static coupling feel free to try it though.
If you have nothing to add I would go ahead and merge this.

asvitkine:
> @asvitkine This week at least my free time is very limited, so I'd like to call it a day here. The code is still a minor improvement over the current state. If you are convinced you can fix the existing problems with static coupling feel free to try it though. If you have nothing to add I would go ahead and merge this.

Sent https://github.com/triplea-game/triplea/pull/10905 with the clean up that I was suggesting.

PR #10905, opened 2022-07-28
Remove unnecessary complexity with DebugMenu.

asvitkine:
Removes global state and several classes by making the menu items get constructed directly when creating the DebugMenu based on the types of players.

Tested playing a game with Hard AI and FlowFieldAI and checking that the debug menu appears as expected.

No functional changes, just code clean up.


TheDog-GH:
Please add a line 1 to each tab that includes
Map Name,     Faction,     Round/Turn number     and AI type that Faction is using

An export ability would be very nice to have, to save on ctrl+A, ctrl+C and pasting in to Notepad++
Thanks!

RoiEXLab (approved):
Well done, thank you for looking into this 👍🏼

PR #11022, opened 2022-09-01
Infra: do not restart lobby on deployment

DanVanAtta:
Just ensure we start the lobby up. A restart right now would
disconnect users, something we do not want.

### DIFF
--- a/infrastructure/ansible/roles/lobby_server/tasks/main.yml
+++ b/infrastructure/ansible/roles/lobby_server/tasks/main.yml
@@ -78,18 +78,19 @@
     owner: "{{ lobby_user }}"
     group: "{{ lobby_user }}"
 
-- name: install systemd service script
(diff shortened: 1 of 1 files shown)


RoiEXLab (approved):
Okay, but how would the lobby use any new versions then?

DanVanAtta:
@RoiEXLab we'll have to trigger it manually for now. This is unlikely to be any worse than where we are today, we are actually still running an old 2.4 lobby. Our existing process really failed to keep the lobby up to date (I think in considerable part because of downtime requirements)

Ideally we would build a zero-downtime mechanism to be able to restart the lobby without (too greatly) impacting players. I think to do that we would need to move the lobby state to database. This way if it restarts, it can recover state. Games already reconnect automatically, we would need the player connections to do something similar.

PR #11092, opened 2022-09-27
Fix code to handle newer savegame json response

### LINKED ISSUE #11056: Savegame File Creation broken with current Pre-release
### Reproduction Steps
Start a PBF game, do a test post, post a turn summary.
You will see that no file is created but a link to a "Null Page" instead.

### Attach a Save Game
See:
https://www.axisandallies.org/forums/post/1606174
https://www.axisandallies.org/forums/post/1606175

### Engine Version
2.6+14142

The problem occurs on the above mentioned forum (see links), that currently runs on NodeBB 1.19.1.
It works fine on forums.triplea-game.org that currently still runs on NodeBB 1.16.2.

It looks like the issue has already started at least back in March 22 with another 2.6-pre-release. 
See for example:
https://www.axisandallies.org/forums/post/1585098

Here @ron-murhammer posted savegames that lead to a "null" , for example https://www.axisandallies.org/forums/topic/20781/panther-s-triplea-pre-release-test-thread/null

It hasn't been noticed before as almost everybody uses the current "stable" for PBF working around all the related issues.

RoiEXLab:

Fixes #11056

This changes the code to accept the newer JSON format instead of silently printing null when it arrives.

## Release Note

FIX|Links to uploaded savegames are no longer 'null' on newer versions of NodeBB

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbForumPoster.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbForumPoster.java
@@ -1,5 +1,6 @@
 package games.strategy.engine.posted.game.pbf;
 
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableSet;
 import games.strategy.engine.framework.system.HttpProxy;
 import games.strategy.triplea.UrlConstants;
@@ -9,6 +10,7 @@ import java.net.HttpURLConnection;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -109,6 +111,14 @@ public class NodeBbForumPoster {
     }
   }
 
+  private static String appendOptionalSaveGameLink(
+      final String text, @Nullable final String saveGameUrl) {
+    if (saveGameUrl == null) {
+      return text;
+    }
+    return text + "\n[Savegame](" + saveGameUrl + ")";
+  }
+
   private void post(
       final CloseableHttpClient client,
       final String token,
@@ -122,7 +132,8 @@ public class NodeBbForumPoster {
             List.of(
                 new BasicNameValuePair(
                     "content",
-                    text + ((saveGame != null) ? uploadSaveGame(client, token, saveGame) : ""))),
+                    appendOptionalSaveGameLink(
+                        text, saveGame != null ? uploadSaveGame(client, token, saveGame) : null))),
             StandardCharsets.UTF_8));
     HttpProxy.addProxy(post);
     try (CloseableHttpResponse response = client.execute(post)) {
@@ -157,12 +168,18 @@ public class NodeBbForumPoster {
       if (status == HttpURLConnection.HTTP_OK) {
         final String json = EntityUtils.toString(response.getEntity());
         try {
-          final String url = (String) YamlReader.readMap(json).get("url");
-          return "\n[Savegame](" + url + ")";
+          final Map<String, Object> jsonObject = YamlReader.readMap(json);
+          final Map<?, ?> responseObject = (Map<?, ?>) jsonObject.get("response");
+          // This is a temporary hack to handle old versions of nodeBB forum json
+          if (responseObject == null) {
+            return (String) jsonObject.get("url");
+          }
+          final List<?> images = (List<?>) Preconditions.checkNotNull(responseObject.get("images"));
+          final Map<?, ?> imageObject = (Map<?, ?>) images.get(0);
+          return (String) Preconditions.checkNotNull(imageObject.get("url"));
         } catch (final Exception e) {
           // This is a temporary hack to handle old versions of nodeBB forum json
-          final String url = (String) YamlReader.readList(json).get(0).get("url");
-          return "\n[Savegame](" + url + ")";
+          return (String) Preconditions.checkNotNull(YamlReader.readList(json).get(0).get("url"));
         }
       }
       throw new IllegalStateException(
--- a/game-app/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbForumPoster.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbForumPoster.java
@@ -1,6 +1,6 @@
 package games.strategy.engine.posted.game.pbf;
 
-import com.google.common.base.Preconditions;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableSet;
 import games.strategy.engine.framework.system.HttpProxy;
 import games.strategy.triplea.UrlConstants;
@@ -114,17 +114,20 @@ public class NodeBbForumPoster {
   private void post(
       final CloseableHttpClient client,
       final String token,
-      final String text,
+      String turnSummaryText,
       @Nullable final SaveGameParameter saveGame)
       throws IOException {
     final HttpPost post = new HttpPost(forumUrl + "/api/v2/topics/" + topicId);
     addTokenHeader(post, token);
+
+    if (saveGame != null) {
+      final String saveGameUrl = uploadSaveGame(client, token, saveGame);
+      turnSummaryText += "\n[Savegame](" + saveGameUrl + ")";
+    }
+
     post.setEntity(
         new UrlEncodedFormEntity(
-            List.of(
-                new BasicNameValuePair(
-                    "content", appendOptionalSaveGameLink(text, client, token, saveGame))),
-            StandardCharsets.UTF_8));
+            List.of(new BasicNameValuePair("content", turnSummaryText)), StandardCharsets.UTF_8));
     HttpProxy.addProxy(post);
     try (CloseableHttpResponse response = client.execute(post)) {
       final int code = response.getStatusLine().getStatusCode();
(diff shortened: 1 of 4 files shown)


DanVanAtta:
Thanks for the patch @RoiEXLab . What are your thoughts to try and capture this in a unit test? Notably have it so we take a copy of the forums payload and then validate that we process it correctly. This example would serve as good documentation going forward so if the actual payload changes we can see the updates. Further, such a test would avoid the fix from being blind and would help give us confidence we really have this resolved. WDYT?

game-app/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbForumPoster.java
```diff
@@ -109,6 +111,14 @@ public CompletableFuture<String> postTurnSummary(
     }
   }
 
+  private static String appendOptionalSaveGameLink(
```
DanVanAtta, line 114: nitpick, methods should be defined after their first usage.
This gets important at scale for consistency and finding where code is within a file.

DanVanAtta: I think this is probably a moot consideration though. 
I'm concerned from an understanding/maintenance perspective by this line here:
```
 appendOptionalSaveGameLink(
                        text, saveGame != null ? uploadSaveGame(client, token, saveGame) : null))),
                        ```
In that line we have control-flow logic, we execute an action to upload a save game, *and* we take the output from that for formatting!! This is a lot to do in just a few lines. Can you work to refactor this a bit so we are better at doing one thing per line? I think that would go a very long way to making this easier to follow. Otherwise, at this juncture, the original code is complex enough that it's very hard to reason about.

File: game-app/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbForumPoster.java [hunk 2]
```diff
@@ -157,12 +168,18 @@ private String uploadSaveGame(
...
-          return "\n[Savegame](" + url + ")";
+          final Map<String, Object> jsonObject = YamlReader.readMap(json);
+          final Map<?, ?> responseObject = (Map<?, ?>) jsonObject.get("response");
+          // This is a temporary hack to handle old versions of nodeBB forum json
+          if (responseObject == null) {
+            return (String) jsonObject.get("url");
+          }
+          final List<?> images = (List<?>) Preconditions.checkNotNull(responseObject.get("images"));
+          final Map<?, ?> imageObject = (Map<?, ?>) images.get(0);
+          return (String) Preconditions.checkNotNull(imageObject.get("url"));
```
DanVanAtta, line 179: If the `checkNotNull` blows up, we'll get an unhelpful error message. We could help fix this by dumping the YAML contents to the error message. WDYT? Ditto for the check-not-null above (and we may want to consider doing a 'try/catch' with ClassCastException to dump the YAML again in case the format changes).

game-app/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbForumPoster.java
```diff
@@ -157,12 +168,18 @@ private String uploadSaveGame(
...
         } catch (final Exception e) {
           // This is a temporary hack to handle old versions of nodeBB forum json
```
DanVanAtta, line 181: Is this comment still current, any idea? Did this behavior get tested, or is just seemingly the right fix (but not actually validated?) This logic down here also helps add to the idea of adding testing so we can validate that we actually handle old & new formats.

Is the 'old versions' of nodeBB largely just the TripleA forum now?

game-app/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbForumPoster.java
```diff
@@ -157,12 +168,18 @@ private String uploadSaveGame(
...
-          final String url = (String) YamlReader.readList(json).get(0).get("url");
-          return "\n[Savegame](" + url + ")";
+          return (String) Preconditions.checkNotNull(YamlReader.readList(json).get(0).get("url"));
```
DanVanAtta, line 182: We previously had this being returned `return "\n[Savegame](" + url + ")";` - is the format of the return value different for 'old versions', why is the '[savegame]' string no longer returned?

game-app/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbForumPoster.java
RoiEXLab: > Is the 'old versions' of nodeBB largely just the TripleA forum now?

Yes, it needs to be updated, but I haven't gotten around to do so. (Also NodeBB 2.0 has been released and I'm not sure if there are any breaking changes that need any special treatment.)

I verified the code works with the a&a forum and our forum and I believe the branch where I check if the "url" key exists will no longer be executed (haven't checked though)

game-app/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbForumPoster.java
RoiEXLab: I mean ideally the error message should never happen. Any error here would indicate the JSON output format changed again, which should hopefully not happen too often. I figured an unhelpful error message would be better than silently printing "null" as a link, but I guess you're right that including the JSON might make debugging in the future better.

game-app/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbForumPoster.java
RoiEXLab: Hmm what do you think about moving the call to `uploadSaveGame(client, token, saveGame)` inside of `appendOptionalSaveGameLink`?

RoiEXLab:
@DanVanAtta 
> What are your thoughts to try and capture this in a unit test? Notably have it so we take a copy of the forums payload and then validate that we process it correctly. This example would serve as good documentation going forward so if the actual payload changes we can see the updates. Further, such a test would avoid the fix from being blind and would help give us confidence we really have this resolved. WDYT?

I agree that unit tests would be nice, but at the moment the code is tightly coupled to the network layer because of the HTTP components thingy. So I believe we would have to turn it into a feign client first to be able to "swap out" the network layer with a mock. I'm not sure how the feign client implementation we have handles changing JSON formats though. Also unit tests just check if the code works with the JSON at the time of implementing, so whenever the JSON happens to change the code would still just break.

game-app/game-core/src/main/java/games/strategy/engine/posted/game/pbf/NodeBbForumPoster.java
DanVanAtta: Perhaps could be reasonable. Though, I don't know if we need to go that "drastic", but perhaps could work. At a minimum I was thinking just to use some intermediate variables and just try to align better with the "one-statement, one-action" principle.

eg:
```
if(saveGame == null) {
   appendOptionalSaveGameLink(text, null);
} else {
   String saveGameLink = uploadSaveGame(client,token,saveGame);
   appendOptionalSaveGameLink(text, saveGameLink);

}
```
 I still see a lot of code smells with the above (eg: side-effect'ful call of 'appendOptionalSaveGameLink', would be much nicer if that were functional), but at least the above is a start towards "sanity"  : )

DanVanAtta:
@RoiEXLab re testing, swapping out the network layer would be good.. though maybe not necessary. If we can at least split the part that handles the response payload, then we can have something very functional that takes as input the JSON response from NodeBB. I do think there should be some sort of test to give us confidence that this works right and to make it easier to update in the future. Without that test, it's really hard to evaluate if any of this was right, and is still correct after updating. I generally like to see tests in review so I can focus on whether the test is complete. If so, then it's only a matter of looking for design issues in the implementation and not "let me be a human debugger & tester to validate that this is correct".

RoiEXLab:
@DanVanAtta I updated the code and incorporated some of your suggestions. I left out any major refactoring changes out to get this finally working for people that actually play PBF.
Regardless we should discuss a migration away from apache http components to allow for better unit tests. Currently the library is used for PBF, PBEM and for the "ContentDownloader" class and to provides proxy configuration for all of those. In my opinion feign does not make sense for the latter case; we could turn the `ContentDownloader` into an interface or mock the class if required. But for the other cases it might make sense to replace it with a feign client to (hopefully) clean up the code and make it more readable and maintainable. WDYT?

DanVanAtta:
@RoiEXLab to really get this patch out before players, we need to finish releasing 2.6; AFAIK almost nothing left to do for it other than to announce it and set every latest to be a release version.  As part of releasing all updates as releases, we need to be sure they are good and find ways to add automated tests for what we can. Check PR: https://github.com/RoiEXLab/triplea/pull/1 where I captured some sample response payloads and plopped them into a test that validated we return the correct URL. With that test harness in place, it was quite a bit easier to remove dead code and simplify, adding clarity for why and where we expect one payload to be a list and another a map.

I would agree that replacing the http clients with Feign for consistency would be a good move. Refactoring the test added to take in a response object would be pretty easy, or would possibly make the test too trivial to keep. Meanwhile it was not that large of an investment to build the test (it did take a while to repeat the manual testing work to capture JSON response payload examples, which is in part why I really wanted us to have those samples written down)

RoiEXLab:
@DanVanAtta Thank you for the PR. I extended the exception handling to print the json when the format mismatches on an inner level as well.

PR #11101, opened 2022-10-01
fixed extra error message when game host is unreachable

klarubbio:


beelee1:
Are you one of the iowa st guys ?

RoiEXLab (changes requested):
This does not fix anything, and annotating `Optional` objects with `@Nullable` is bad practice!

DanVanAtta:
Most java annotations are stripped after compile time. As part of annotations, they indicate how they into the compilation phase they go (some can be retained even through runtime). `@Nullable` is only a hint for compilers & for developers, it has no impact on how the code operates at all. Hence, adding `@Nullable` annotations will not alter behavior. @RoiEXLab  has a good point that `Optional` values should never be null, their purpose is to avoid null values and to provide an API that is null-safe.

PR #11129, opened 2022-10-12
Websocket Relay: add various websocket alternatives for network communication

DanVanAtta:
Behind a feature flag, add a variety of network connections that send data over websocket rather than the  previous java-sockets mechanism.

Further work is needed to finish more conversions.

Of note, the pattern is:
1. For each 'remote action method', we create a message envelope class that contains all of the same data as the method parameters. Given a message envelope, we can then always invoke the existing network method.
2. Wire a network bridge to places that do network connection
3. Instead of invoking the network method, we create the equivalent message envelope and send that
4. In places that receive network messages, we add a listener to the network bridge that will listen for network messages. When we get the right network envelope, our listener uses that envelope to extract data and call the existing network method

Example:
```
@RemoteAction
void foo(String a);
```
We create envelope:

```
class FooMessage extends WebsocketMessage {
   String a;
   ....
}
```

We replace existing invocations:

```
String value = ...;
if(ClientSettings.useWebsockets) {
  networkBridge.sendMessage(new FooMessage(value));
} else {
  remoteMessenger.foo(value);
}
```

Last, we add a listener to handle the receiving side (previously Java customized RMI-like mechanism invokes the method for us)

```
  networkBridge.addLIstener(fooMessage -> foo(fooMessage.a));


@RemoteAction
void foo(String a) {

}
```

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/chat/ChatPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/chat/ChatPanel.java
@@ -2,6 +2,7 @@ package games.strategy.engine.chat;
 
 import games.strategy.engine.chat.ChatMessagePanel.ChatSoundProfile;
 import games.strategy.net.Messengers;
+import games.strategy.net.websocket.ClientNetworkBridge;
 import java.awt.BorderLayout;
 import java.awt.Dimension;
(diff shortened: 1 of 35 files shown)


RoiEXLab:
Looking promising 👍🏼

PR #11147, opened 2022-10-21
Infra: Add lobby 2.6 config & move Atlanta02 bot server to 2.6

DanVanAtta:
- Changes Atlanta bot from 2.5 lobby to 2.6 lobby
- We will now always update atlanta bot fully with source code updates
- Add a '--limit' config option when running ansible so we can selectively update just the atlanta bot
- Update bot install folder to '/opt'. It is more canonical rather than installing into a home folder. We keep '2.5' lobby bots with their existing home folders for now.

### DIFF
--- a/.github/workflows/deploy-servers.yml
+++ b/.github/workflows/deploy-servers.yml
@@ -23,6 +23,7 @@ jobs:
           cd infrastructure
           echo "$ANSIBLE_VAULT_PASSWORD" > vault_password
           ./run_ansible --environment production --tags common,java,postfix,postgres,flyway
+          ./run_ansible --environment production --limit prod2-bot02.triplea-game.org
         env:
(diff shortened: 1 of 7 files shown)


RoiEXLab (approved):

RoiEXLab:
Feel free to merge when you're ready (not sure if this applies the change immediately)

DanVanAtta:
The line added to the github action build config would deploy some of these changes on every merge.
Though, it is "safe":
- we have not configured anything to restart on change (yet)
- I've ran locally deployments with `--dry-run --check` which indicates what changes would be made and does not actually make them. I tweaked our checked in configuration until it matched what we have on the one updated server. I'll emphasize this tactic a bit, it's not the case that you can only make changes via the checked in configuration, but any final configuration should get checked in. So you can tweak a server, and then run dry-run deployments until no diff is detected, meaning the server configuration is then captured. Going forward it's then easier to make tweaks/config updates via the checked in code to then deploy to servers (or a new server).

PR #11150, opened 2022-10-21
Refactor/github api client and error report

DanVanAtta:

Various refactors in prep for updating error reporting module.

First started with cleaning up the error reporting module & tests and noticed it was a bit convoluted. This refactor started with some inlining and reduction of classes.

The last update changes the github api client to require less configuration to be passed around. We do this by instead of passing 'org' and 'repo' in each method invocation, we inject them into the github api client and then pass the api-client around.


spitfire-server/error-reporting-module/src/main/java/org/triplea/modules/error/reporting/ErrorReportModule.java
```diff
@@ -2,68 +2,53 @@
...
+   */
+  public ErrorReportResponse createErrorReport(final CreateIssueParams createIssueParams) {
+    var errorReportRequest = createIssueParams.getErrorReportRequest();
+
+    var githubCreateIssueResponse =
+        githubApiClient.newIssue(
+            CreateIssueRequest.builder()
+                .title(errorReportRequest.getGameVersion() + ": " + errorReportRequest.getTitle())
+                .body(errorReportRequest.getBody())
+                .labels(new String[] {"Error Report"})
```
RoiEXLab, line 39: I dislike the fact you have to explicitly create an array here. Did you consider using `@Singular` to get a sort-of varargs like experience?

See https://github.com/projectlombok/lombok/issues/1568#issuecomment-363562372 for reference

RoiEXLab (approved):

spitfire-server/error-reporting-module/src/main/java/org/triplea/modules/error/reporting/ErrorReportModule.java
DanVanAtta: I did not consider that. I gave it a shot, generates a compilation errors. I thought '@singular' only worked with collections and not arrays. This could help with test cases (though I'm a bit concerned about ordering issues with an array and checking equality). Arguably converting this to a `Set` collection would be better overall (though hard to know if really worth the effort)

PR #11155, opened 2022-10-22
Build: remove YAML validation

DanVanAtta:
- The validated YAML files are going to be soon deprecated with 2.6 launch and we will no longer need to update the checked YAML files. Until 2.6 is launched, we can check YAML updates manually..
- YAML validation fails somewhat consistently with errors like:
> https://raw.githubusercontent.com/triplea-maps/total_world_war/master/preview.png:
>   error requesting map resource (peer not authenticated)

Builds should be very deterministic and it is important they not fail unless there is an actual problem.


config/triplea/schemas/lobby_server.json
```diff
@@ -1,49 +0,0 @@
...
-      },
-      "error_message": {
-        "description": "The message displayed when a user connects to an unavailable lobby server",
-        "type": ["string", "null"]
-      }
-    },
-    "required": ["version", "host", "port","https_port", "message", "error_message"],
-    "additionalProperties": false
-  }
-}
```
RoiEXLab, line 49: Didn't even know we had this check in place

RoiEXLab:
I see your point, no strong feelings on this one. I agree that having sporadic failures due to network issues does not help build trust in the system. However, I don't think having yaml validation did ever hurt for the lobby_server.yaml. So the tradeoff we make here is that we no longer let CI check this file because it is rarely modified, not used in 2.6+ (thus it's not as important as live servers) and we can get rid of code, right?

DanVanAtta:
Unfortunately I've been seeing far more than sporadic failures. It's about 1 or 2 out of every 3 builds where this fails! I think it has something to do with rate limits, once hit, it takes more than 10 minutes (or something around there) to reset so it is pretty easy to get repeated failures. It "helps" that the build takes about 3 or 4 minutes, which makes rate limiting less frequent. I'm pretty sure the 'peer not authenticated' is actually a rate-limit error.

Both 'lobby_server.yaml' and 'servers.yml' are unused in 2.6! I think the code to read those files is likely already dropped. We'll modify one of those files just one last time to notify users to upgrade to 2.6.

PR #11158, opened 2022-10-22
Bump logback (-> 1.4.4) & dropwizard versions (-> 2.1.3)

DanVanAtta:
The two versions need to go together. logback removed 'JmxConfigurator' and the updated dropwizard version handles this by checking if the class is not found.

Notably dropwizard added this code to avoid a classDefNotFound error:
```
    private void configureJMX() {
        try {
            Class.forName("ch.qos.logback.classic.jmx.JMXConfigurator");
        } catch (ClassNotFoundException e) {
            return;
        }
```


DanVanAtta:
Not a lot of luck here, seeing:

>     java.lang.NoSuchMethodError: ch.qos.logback.core.util.CachingDateFormatter.setTimeZone(Ljava/util/TimeZone;)V
>        at ch.qos.logback.access.pattern.DateConverter.start(DateConverter.java:53)
>       at ch.qos.logback.core.pattern.ConverterUtil.startConverters(ConverterUtil.java:37)
>       at ch.qos.logback.core.pattern.PatternLayoutBase.start(PatternLayoutBase.java:90)
>       at ch.qos.logback.access.PatternLayout.start(PatternLayout.java:210)


The 'logback.access' class (PatternLayout) has an updated version but the above is using a version from logback 1.2.11 

From a dependency tree, we have this for the drop-wizard websocket library:

|    |    |    +--- ch.qos.logback:logback-classic:1.2.11 -> 1.4.4 (*)

It looks like this is also brough in by a slightly dated version of 'io.dropwizard:dropwizard-logging'

Forcing that to be upgraded to the latest does not resolve this issue...

PR #11194, opened 2022-11-06
Lobby config: direct 2.5 users to 'prod.triplea-game.org'

DanVanAtta:
Only the 2.5 game engine and below reads 'servers.yml'. This update directs those game clients to 'prod.triplea-game.org'. We have NGINX running on that URL which will in turn redirect the game client to an appropriate lobby server (NGINX going forward will control which game client versions connect to which lobby).

Merging this would have immediate effect for any 2.5 users logging into the lobby.

### DIFF
--- a/servers.yml
+++ b/servers.yml
@@ -1,7 +1,8 @@
+# This file is read by game clients 2.2 through 2.5
 latest: 2.5.22294
 servers:
   - version: "2.0.20057"
-    lobby_uri: https://prod2-lobby.triplea-game.org
+    lobby_uri: https://prod.triplea-game.org
     message: |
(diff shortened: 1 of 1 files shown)


DanVanAtta:
There are some additional 2.6 bots coming online (I turned them off in the 2.5 lobby).
We can probably merge this somewhat sooner when:
- confirmed that we have at least 12 2.6 bots online (maybe we want more?)
- announced on forums

After merge, we can essentially turn off the 2.5 lobby entirely (keeping in mind we will be disconnecting any users and preventing them from rejoining bot-games. Perhaps best down when the lobby is at 'low-tide' in terms of activity).

Issue #10106, opened 2022-02-11
2.6.564: ImageLoader#getImage:21 - java.lang.IllegalArgumentException


tripleabuilderbot:
## Log Message
File must exist at path: /usr/share/triplea/bin/assets/launch_screens/triplea-logo.png

## TripleA Version
2.6.564

## Java Version
17.0.1

## Operating System
Linux

## Stack Trace
```
Exception: java.lang.IllegalArgumentExceptionFile must exist at path: /usr/share/triplea/bin/assets/launch_screens/triplea-logo.png
java.lang.Exception
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:145)
	at org.triplea.io.ImageLoader.getImage(ImageLoader.java:21)
	at games.strategy.triplea.ResourceLoader.loadImageAsset(ResourceLoader.java:75)
	at games.strategy.engine.framework.startup.ui.panels.main.game.selector.GameSelectorPanel.<init>(GameSelectorPanel.java:91)
	at games.strategy.engine.framework.startup.ui.panels.main.MainPanelBuilder.buildMainPanel(MainPanelBuilder.java:25)
	at games.strategy.engine.framework.ui.MainFrame.<init>(MainFrame.java:45)
	at games.strategy.engine.framework.ui.MainFrame.buildMainFrame(MainFrame.java:54)
	at games.strategy.engine.framework.GameRunner.lambda$start$0(GameRunner.java:66)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:318)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:771)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:722)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:716)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:399)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:86)
	at java.desktop/java.awt.EventQueue.dispatchEvent(EventQueue.java:741)
	at java.desktop/java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:203)
	at java.desktop/java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:124)
[...]

PR #11231, opened 2022-11-19
expanded public static Image getImage(final Path path) errorMessage

AndiLilaj:
Expanded Image getImage(final Path path) errorMessage with:
"You should build with the checked in launcher, or and run 'downloadAssets'." 
to avoid unnecessary issues like #10106   because Typically, the logo image is bundled with the game, and if game bundling is a problem, this is not necessarily the first error you would see in that case (hence it is likely a dev issue).

### DIFF
--- a/lib/java-extras/src/main/java/org/triplea/io/ImageLoader.java
+++ b/lib/java-extras/src/main/java/org/triplea/io/ImageLoader.java
@@ -18,7 +18,8 @@ public class ImageLoader {
    * @param path Location of the file to be read as an image.
    */
   public static Image getImage(final Path path) {
-    checkArgument(Files.exists(path), "File must exist at path: " + path.toAbsolutePath());
+    checkArgument(Files.exists(path), "File must exist at path: " + path.toAbsolutePath()
+        + "You should build with the checked in launcher, or and run 'downloadAssets'.");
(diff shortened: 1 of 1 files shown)


RoiEXLab:
Hi @AndiLilaj 

Thank you for your contribution. Our error messages in general need a better system, but for a dev-only context the change seems reasonable. Before being able to merge the checkstyle errors need to be fixed first: https://github.com/triplea-game/triplea/actions/runs/3505162158/jobs/5876611345
Because you are a first-time contributor to this repo, you'll have to wait until a maintainer approves the CI runs, so you most likely won't see the update immediately after pushing the changes.

@DanVanAtta Since you commented on #10106 I'd appreciate to hear your thoughts on this approach in the context of dev setup.

AndiLilaj:
Did Gradle spotlessApply, it's formatted properly now

RoiEXLab:
Thank you very much

IrwinRapper:
> Expanded Image getImage(final Path path) errorMessage with: "You should build with the checked in launcher, or and run 'downloadAssets'." to avoid unnecessary issues like #10106 because Typically, the logo image is bundled with the game, and if game bundling is a problem, this is not necessarily the first error you would see in that case (hence it is likely a dev issue).

Issue #11258, opened 2022-11-28
Unit Purchase Display Improvements + Block Too Many Purchases


mcclelyd:
Delivering a feature that was requested on the forum here: https://forums.triplea-game.org/topic/3311/student-group-studying-triplea/2

TheDog-GH:
@ArieLevental
https://github.com/triplea-game/triplea/discussions/11242

Just wondering how its progressing?

@beelee1

panther2:
#9231

PR #11273, opened 2022-12-06
Updates to UI to show how many units can be purchased.

jtkenny:
## Release Note

This is the pull request for the changes done by the ISU students. This contains the UI updates requested on the forum for displaying the total number of units that can be bought. The related issue is here: https://github.com/triplea-game/triplea/issues/11258

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/ProductionPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/ProductionPanel.java
@@ -236,9 +236,11 @@ class ProductionPanel extends JPanel {
     return panel;
   }
 
-  private void setLeft(final ResourceCollection resourceCollection, final int totalUnits, final int maxUnits) {
+  private void setLeft(
+      final ResourceCollection resourceCollection, final int totalUnits, final int maxUnits) {
     remainingResources.removeAll();
-    left.setText(String.format("%d/%d total units purchased. Remaining resources: ", totalUnits, maxUnits));
+    left.setText(
+        String.format("%d/%d total units purchased. Remaining resources: ", totalUnits, maxUnits));
     if (resourceCollection != null) {
       remainingResources.add(
           uiContext.getResourceImageFactory().getResourcesPanel(resourceCollection, gamePlayer));
@@ -338,8 +340,18 @@ class ProductionPanel extends JPanel {
           final int movement = attach.getMovement(player);
           final int defense = attach.getDefense(player);
           final Resource currentResource = player.getData().getResourceList().getResource("PUs");
-          info.setText(attack + " | " + defense + " | " + movement + " ≤" + (player.getResources().getQuantity(currentResource) / cost.getInt(currentResource)));
-          if( cost.getInt(currentResource) < lowestCost) { lowestCost = cost.getInt(currentResource); }
+          info.setText(
+              attack
+                  + " | "
+                  + defense
+                  + " | "
+                  + movement
+                  + " ≤"
+                  + (player.getResources().getQuantity(currentResource)
+                      / cost.getInt(currentResource)));
+          if (cost.getInt(currentResource) < lowestCost) {
+            lowestCost = cost.getInt(currentResource);
+          }
           tooltip
               .append(type.getName())
               .append(": ")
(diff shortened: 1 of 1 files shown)


beelee1:
@jtkenny nice :+1:  It's been a few days so lets go ahead and ping @DanVanAtta . He hasn't been active lately but I know he's interested in your work. Will ping @RoiEXLab too :)

TheDog-GH:
@RoiEXLab 
It relates to my (& by others) request here;
https://forums.triplea-game.org/topic/3311/student-group-studying-triplea/2

RoiEXLab:
@TheDog-GH Thanks for clarifying. In this case I'm just asking for the screenshots (as mentioned earlier) for future reference and to fix the formatting. 👍🏼 
Thanks for your time

game-app/game-core/src/main/java/games/strategy/triplea/ui/ProductionPanel.java
```diff
@@ -60,10 +60,13 @@ class ProductionPanel extends JPanel {
 
   protected JDialog dialog;
   private boolean bid;
+
+  private int lowestCost;
   final Action doneAction = SwingAction.of("Done", () -> dialog.setVisible(false));
 
   ProductionPanel(final UiContext uiContext) {
     this.uiContext = uiContext;
+    lowestCost = 100000;
```
RoiEXLab, line 69: One thing I forgot to ask:  Where does this value of 100k come from? Is it just an arbitrarily high number or is there a deeper meaning?

jtkenny: Arbitrary high number. If it was zero, then if the associated if statement was never triggered it would cause divide by zero.

RoiEXLab: Ah okay, in this case is something speaking against using [`Integer.MAX_VALUE`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Integer.html#MAX_VALUE)? Also if you want to can assign the value in the same line where you define the variable. No need to put it into the constructor

DanVanAtta:
Thanks for taking a whack at this update and the work you have put in.

Few questions:

(1) Have you considered any other UI treatments?

I'm not sure we want to go with the UI treatment suggested. What bugs me is how it groups visually in my brain, it looks like "1 | 2" and "1 < 10" to me.

If you are short on ideas for alternatives, there are a few UI designer folks who hang out on the forums who likely would provide more than feedback and options than you could shake a stick at.

(2) Are there any functionality changes to limit the max from being entered? For example, does the "up" arrow toggle become disabled when at the max? If a user enters in a value over the max, is it overwritten with the max?

(3) How extensively have edge cases been considered? Notably I'm thinking of maps where there is placement that is outside of factories (one example is placing allied units on china I think in BWv3 maps). Have you checked additional maps like civil war and others that have some extra variation for unit placement?

----------------

My availability has been pretty poor lately. Apologies in advance for any delayed responses. Otherwise, generally I think we should focus on:
- getting these updates right
- getting the updates to look good
- ensuring code quality/style is reasonable and consistent

Thanks for taking a whack at this update and the work you have put in.

Few questions:

(1) Have you considered any other UI treatments?

I'm not sure we want to go with the UI treatment suggested. What bugs me is how it groups visually in my brain, it looks like "1 | 2" and "1 < 10" to me.

If you are short on ideas for alternatives, there are a few UI designer folks who hang out on the forums who likely would provide more than feedback and options than you could shake a stick at.

(2) Are there any functionality changes to limit the max from being entered? For example, does the "up" arrow toggle become disabled when at the max? If a user enters in a value over the max, is it overwritten with the max?

(3) How extensively have edge cases been considered? Notably I'm thinking of maps where there is placement that is outside of factories (one example is placing allied units on china I think in BWv3 maps). Have you checked additional maps like civil war and others that have some extra variation for unit placement?

----------------

My availability has been pretty poor lately. Apologies in advance for any delayed responses. Otherwise, generally I think we should focus on:
- getting these updates right
- getting the updates to look good
- ensuring code quality/style is reasonable and consistent

DanVanAtta:
Ah, just noticed an interesting thing here is the difference between max able to purchase (due to cost) and the max that could be purchased due to placement. I think the latter is the really interesting number.
Perhaps we can prune the UI updates down to just the update in the lower left corner where it says (0/10 units purchased). That denominator value is most interesting when it is equal to the max production capacity (rather than say the max number that can be purchased before running out of money).

DanVanAtta:
Please re-open @jtkenny when you're ready & consider marking this PR as a draft as well. Given it's a been a month without any update, I'm marking this as closed to remove it from the active queue.

TheDog-GH:
@asvitkine 
This PR by some students might be of use.  
It never got actioned and is very similar to the other one with no code.

Issue #10251, opened 2022-03-31
Extra error message when game host is unreachable


DanVanAtta:
Precondition: the IP address of your computer shall not match "https://www.whatismyip.com/" and port forwarding must not be set up.

Repro steps:
  -  launch game
  -  join lobby
  -  host game
  -  wait for "your computer is not reachable from the internet" message to appear
  -  notice additional error message

Incorrect error message:
![Screenshot from 2022-03-30 21-13-27](https://user-images.githubusercontent.com/12397753/160975154-fa49bb2b-8e14-4ec0-a4d0-f51761bac5b8.png)

The right hand side should not appear, we already have the correct error message on screen on the left hand side.

Presumably this is problem with code contracts: expectation of where we expect exceptions to be thrown vs nullable return values.

DanVanAtta:
Presumably host a game without port forwarding enabled. I found this
screenshot elsewhere and did not repro personally

DanVanAtta:
Confirmed that is the repro case, ie:
- launch game
- join lobby
- host game
- wait for "your computer is not reachable from the internet" message to appear
- notice additional error message

Precondition: the IP address of your computer shall not match "https://www.whatismyip.com/" *and* port forwarding must _not_ be set up.

RaiNova:
Let‘s clarify which parts of this fix are a good first issue and which should be done be people with more experience in the TripleA code. I see the following subtasks to fix this:
a) Reproduce the error and document in this issue how to reproduce it (Good first subtask, probably mainly already done by Dan)
b) Fix code contract
- find any documentation about where we expect exceptions to be thrown vs nullable return values (I guess there‘s none. Total absence is hard to find for a novice, so this is an expert subtask)
- produce a general guideline of where we expect exceptions to be thrown vs nullable return values (expert)
- identify where in the context of the error the guideline is violated (in the assumed current absence of a guideline, this is vast, but a novice could identify a small area that suits himself to work on)
- agree on the area to work on (novice + expert)
- fix the area (novice)

suryakapurothu:
> Let‘s clarify which parts of this fix are a good first issue and which should be done by people with more experience in the TripleA code. I see the following subtasks to fix this: a) Reproduce the error and document in this issue how to reproduce it (Good first subtask, probably mainly already done by Dan) b) Fix the code contract
> 
[...]

PR #11317, opened 2022-12-30
Fix "host is unreachable" causing extra error message window

asperry182:

Fixes [issue #10251](https://github.com/triplea-game/triplea/issues/10251)

When trying to host a lobby game, if host is unreachable from lobby, return out of InGameLobbyWatcher to avoid initialization of keepAliveTimer done later in the method. This will avoid a NullPointerException thrown by javax.validation.nonnull that would occur when creating a LobbyWatcherKeepAliveTask instance since gameId is null in this case. 

Also update InGameLobbyWatcher.shutDown method to only call gameToLobbyConnection.disconnect if gameId is non-null (and thus game has been successfully created already) since there is no need to call lobby to remove a game that was not successfully created in this case and also to avoid an IllegalArgumentException thrown by LobbyWatcherClient.removeGame when invoked later.

## Release Note
FIX|Remove extra error message window when host is unreachable over internet when attempting to host a lobby game

### DIFF
--- a/game-app/game-core/src/test/java/games/strategy/engine/framework/startup/ui/InGameLobbyWatcherTest.java
+++ b/game-app/game-core/src/test/java/games/strategy/engine/framework/startup/ui/InGameLobbyWatcherTest.java
@@ -9,7 +9,6 @@ import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.atLeastOnce;
-import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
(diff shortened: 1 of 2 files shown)


beelee1:
@asperry182 Nice that you're jumping in with some code stuff. Hope you hang around until @DanVanAtta  or @RoiEXLab  get back to you. Your time is appreciated :)

game-app/game-core/src/test/java/games/strategy/engine/framework/startup/ui/InGameLobbyWatcherTest.java
```diff
@@ -1,15 +1,119 @@
...
+
+    // need to set LOBBY_GAME_COMMENTS system property to avoid NPE in SystemPropertyReader.gameComments()
+    System.setProperty(LOBBY_GAME_COMMENTS, "testLobbyGameComments");
+
+    IServerMessenger mockIServerMessenger = mock(IServerMessenger.class);
+    GameToLobbyConnection mockGameToLobbyConnection = mock(GameToLobbyConnection.class);
+    WatcherThreadMessaging mockWatcherThreadMessaging = mock(WatcherThreadMessaging.class);
+    InGameLobbyWatcher mockInGameLobbyWatcher = mock(InGameLobbyWatcher.class);
+    INode mockINode = mock(INode.class);
+    GamePostingResponse mockGamePostingResponse = mock(GamePostingResponse.class);
```
RoiEXLab, line 44: Instead of redefining these mocks for every test, I'd recommend declaring them as fields instead.
There are 2 variants to achieve this:

1. Using `@Mock`
```java
import org.mockito.junit.jupiter.MockitoExtension;

@ExtendWith(MockitoExtension.class)
class InGameLobbyWatcherTest {
  // No need to initialize, mockito does the magic behind the scenes.
  @Mock
  private IServerMessenger mockIServerMessenger;
}
```

2. Using regular fields
```java
class InGameLobbyWatcherTest {
  // JUnit creates a new instance of this class for every test, so this is fine.
  private IServerMessenger mockIServerMessenger = mock(IServerMessenger.class);
}
```

game-app/game-core/src/test/java/games/strategy/engine/framework/startup/ui/InGameLobbyWatcherTest.java
```diff
@@ -1,15 +1,119 @@
...
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
+import org.triplea.http.client.lobby.game.lobby.watcher.GamePostingRequest;
+import org.triplea.http.client.lobby.game.lobby.watcher.GamePostingResponse;
+import org.triplea.http.client.web.socket.client.connections.GameToLobbyConnection;
 
 final class InGameLobbyWatcherTest {
+
+  @Test
+  public void testNewInGameLobbyWatcher_hostNotReachable() throws UnknownHostException {
```
RoiEXLab, line 34: We have a convention to just throw `Exception` on test cases, so we don't have to bother with the specifics of the exception details.

RoiEXLab (changes requested):
Thank you for this PR. I really appreciate your efforts. However there are a couple of things that need fixing before we can merge this.
1. A test is failing that needs fixing.
2. The files don't follow our code style standards. Please run `./gradlew spotlessJavaApply` via the command line to automatically reformat the files.

I'm looking forward to hear back from you 👍🏼

game-app/game-core/src/test/java/games/strategy/engine/framework/startup/ui/InGameLobbyWatcherTest.java
asperry182: updated to use reusable mocks at class level defined with @Mock

game-app/game-core/src/test/java/games/strategy/engine/framework/startup/ui/InGameLobbyWatcherTest.java
asperry182: updated these two tests to just throw base Exception now

RoiEXLab (approved):
Thank you very much

PR #11321, opened 2023-01-01
Updated some message for more clarity purposes

darshanjoshi16:

### DIFF
--- a/game-app/map-data/src/main/java/org/triplea/map/game/notes/GameNotes.java
+++ b/game-app/map-data/src/main/java/org/triplea/map/game/notes/GameNotes.java
@@ -19,10 +19,10 @@ public class GameNotes {
   public static String loadGameNotes(final Path xmlGameFile) {
     Preconditions.checkArgument(
         Files.exists(xmlGameFile),
-        "Error, expected file did not exist: " + xmlGameFile.toAbsolutePath());
+        "File Existantial Error, expected file did not exist: " + xmlGameFile.toAbsolutePath());
     Preconditions.checkArgument(
         !Files.isDirectory(xmlGameFile),
-        "Error, expected file was not a file: " + xmlGameFile.toAbsolutePath());
+        "File Type Error, expected file was not a file: " + xmlGameFile.toAbsolutePath());
 
     final String notesFileName = createExpectedNotesFileName(xmlGameFile);
     final Path notesFile = xmlGameFile.resolveSibling(notesFileName);


TheDog-GH:
@darshanjoshi16 
Can I be cheeky and ask you you to expand this error massage as well 

https://forums.triplea-game.org/topic/3326/1941-global-command-decision/51

Repeated below;
However with latest TripleA-2.6+14246 now I get this error;

f0d1659b-1c8b-4cda-bce9-e024f31eb186-image.png

My triplea.log file with a txt extension
[triplea.log.txt](https://forums.triplea-game.org/assets/uploads/files/1672389903014-triplea.log.txt)

In the log it mentions
329975 08:39:44.852 [Thread-12] ERROR o.t.game.client.HeadedGameRunner - null

Is this related to a networked/lobby game?

TheDog-GH:
@RoiEXLab  @DanVanAtta 
Please check darshanjoshi16 code for approval.

darshanjoshi16:
> @darshanjoshi16 Can I be cheeky and ask you you to expand this error massage as well
> 
> https://forums.triplea-game.org/topic/3326/1941-global-command-decision/51
> 
> **Repeated below;** However with latest TripleA-2.6+14246 now I get this error; image
> 
> **My triplea.log file with a txt extension** [[triplea.log.txt](https://forums.triplea-game.org/assets/uploads/files/1672389903014-triplea.log.txt)](https://forums.triplea-game.org/assets/uploads/files/1672389903014-triplea.log.txt)
> 
> In the log it mentions 329975 08:39:44.852 [Thread-12] ERROR o.t.game.client.HeadedGameRunner - null
> 
> **Is this related to a networked/lobby game?**

As I can see the snippet and issues, It seems like it is also a element existance exception which is showing that the element isn't found by interpreter.

game-app/map-data/src/main/java/org/triplea/map/game/notes/GameNotes.java
```diff
@@ -19,10 +19,10 @@ public class GameNotes {
   public static String loadGameNotes(final Path xmlGameFile) {
     Preconditions.checkArgument(
         Files.exists(xmlGameFile),
-        "Error, expected file did not exist: " + xmlGameFile.toAbsolutePath());
+        "File Existantial Error, expected file did not exist: " + xmlGameFile.toAbsolutePath());
```
RoiEXLab, line 22: Is "Existantial" even a real word? I'd use just "File Error" here, or drop the error prefix entirely so the error message becomes "Expected file did not exist: <filename>"

game-app/map-data/src/main/java/org/triplea/map/game/notes/GameNotes.java
```diff
@@ -19,10 +19,10 @@ public class GameNotes {
...
     Preconditions.checkArgument(
         !Files.isDirectory(xmlGameFile),
-        "Error, expected file was not a file: " + xmlGameFile.toAbsolutePath());
+        "File Type Error, expected file was not a file: " + xmlGameFile.toAbsolutePath());
```
RoiEXLab, line 25: I agree that this error message is not ideal, but I don't think adding "File Type" does help with that.
Instead I'd recommend using something like "Expected '<filename>' to be a file, but was a directory instead!", which is certainly better than claiming that a "file was not a file".

game-app/map-data/src/main/java/org/triplea/map/game/notes/GameNotes.java
DanVanAtta: I love this, an existential error for a file : )

ex·is·ten·tial
/ˌeɡzəˈsten(t)SHəl/

adjective
relating to existence.
"the climate crisis is an existential threat to the world"

PHILOSOPHY
concerned with existence, especially human existence as viewed in the theories of existentialism.
"the existential dilemma is this: because we are free, we are also inherently responsible"

LOGIC
(of a proposition) affirming or implying the existence of a thing.

-----------------

On a more serious note, error messages in TripleA are intended for users. The error message should as best as possible describe:
- what happened, what was the error
- what is the impact of the error. In this case, game notes cannot be loaded. Though, the way the error is handled we are going to crash the whole application. It could be appropriate to return an empty string and display a warning pop-up message to the user that no game notes could be loaded. Since the game crashes right after we show this error, there is not much need to describe impact because the application will have just crashed (impact is pretty obvious)
- last, what a user can do to fix the error, or what can the user check to help resolve the error. In this case simply stating that we expected a game notes file at a given path is pretty sufficient.

So, this error message probably could only be slightly improved to be more like:
```
String.format("Could not find game notes file at path: %s", xmlGAmeFile.toAbsolutePath());
```

If we update the code to return an empty string in this condition instead of throwing an exception, then an error message like this would be better:
```
String.format("Could not find game notes file. Game notes will not appear. Expected game notes file to be at location: %s, but could not find a file there. Contact the map maker by posting on forums to have a game notes file be created for this map.");
```

PR #11362, opened 2023-01-15
Build scripts: Run check links in parallel (faster)

DanVanAtta:
Reduces 'check-links' run time from 3s to 1s

Of note, the filter mechanism is a bit different as well. Do we a regex match of the URLs and use grep to filter out the URLs to be ignored.


DanVanAtta:
Original runtime is 3s to 2s. 
Not the biggest gain here per say.

PR #11423, opened 2023-01-29
Update servers.yml

Cernelius:
https://forums.triplea-game.org/topic/3404/revised-toc19-signup


Cernelius:
Please someone merge this.

I would self-merge it, but I think I cannot.

beelee1:
@Cernelius merging. idk if it'll show up right away or if it's only for a prerelease or what. Guess I'll find out lol

PR #11511, opened 2023-02-27
Fix invalid filename

RoiEXLab:
This should hopefully address the problem with #11413

/cc @DanVanAtta


RoiEXLab:
@DanVanAtta It seems like the invalid character in the file caused it to not be applied at all. Fixing it cause the build to fail, so I just removed the file. Now you should be able to check out the repository on windows again.

DanVanAtta:
Any idea what the invalid character was?
Can we fix this without nuking the file?
Perhaps there is something we can add in .gitattributes?

DanVanAtta:
It looks like the invalid character was simply a trailing space. I see that this is an invalid filename for windows.

Removing that space and the configuration did get picked up and the tests do run in parallel (a _lot_ faster). Though, we do hit single-threadedness issues with the AI battle calculator insisting it run on one thread. This seems like a potential place for a massive performance boost if we can resolve that.

I marked the corresponding AI test to run with a single thread and it hardly buys a faster runtime (8 seconds vs 6.7 seconds, sometimes the same runtim @ 8 seconds).

I tried this config out in 'game-core' and it reduced test runtime down from 40s to 22s, but there were concurrency errors. 

Bottom line, seems like the tests are not yet set up to run in parallel. Fixing that would unlock performance gains when running tests, but presumably large gains for the production code if we could run it in parallel.

PR #11512, opened 2023-02-28
Add missing test runs

RoiEXLab:
The build for #11505 is green even though it shouldn't. This should put back the missing checks that were not executed on the PR build

### DIFF
--- a/game-app/run/check
+++ b/game-app/run/check
@@ -4,9 +4,13 @@ set -e
 
 scriptDir=$(dirname "$0")
 "$scriptDir/../../gradlew" --parallel \
+  :game-app:ai:check \
+  :game-app:domain-data:check \
+  :game-app:game-core:check \
   :game-app:game-headed:check \
   :game-app:game-headless:check \
   :game-app:game-relay-server:check \
+  :game-app:map-data:check \
   :game-app:smoke-testing:check
 
 "$scriptDir/.build/check-links"


RoiEXLab:
@DanVanAtta FYI, I'm not sure if it's desirable to have another place where we have too keep track of submodules. Thoughts?

DanVanAtta:
AFAIK gradle is transitive about how it approaches tasks. Thus, any submodules that are dependencies should be rebuilt and have their tests run. I wonder if one of these are outside of that dependency graph.

Or, is my thinking here about transitive tasks in gradle incorrect?

RoiEXLab:
> Or, is my thinking here about transitive tasks in gradle incorrect?

To be fair I'm not entirely sure on this either. My current theory is that the transitivity does only apply for "submodules" in the sense that `gradle :game-app check` runs `gradle :game-app:some-submodule check` for all "submodules". (I should have tried that, actually)
However to compile a module, gradle has to also build the dependencies, which is unrelated to the task we're trying to run, so compilation is run transitively via the dependency tree.

TL;DR I believe that `:game-app:game-headed:check` invokes `:game-app:game-core:compileJava` but not `:game-app:game-core:check` and we should change it to `:game-app check` (even though I'm not sure how the parallel flag works in that context)

DanVanAtta:
Yeah, my thinking was not right. Transitive dependencies are executed but a task is not run in a transitive manner across dependencies. The `--dry-run` flag is very helpful here, it tells us what tasks would run. Checking the list, we see everything is compiled and then at the very end we only run tests on the `game-headed` module:

```
dan@desk:~/work/triplea$ ./gradlew :game-app:game-headed:check --dry-run
:game-app:domain-data:compileJava SKIPPED
:game-app:domain-data:processResources SKIPPED
:game-app:domain-data:classes SKIPPED
:game-app:domain-data:jar SKIPPED
:
:
:
:game-app:map-data:compileJava SKIPPED
:game-app:map-data:processResources SKIPPED
:game-app:map-data:classes SKIPPED
:game-app:map-data:jar SKIPPED
:
:
:
:game-app:game-headed:compileJava SKIPPED
:game-app:game-headed:processResources SKIPPED
:game-app:game-headed:classes SKIPPED
:game-app:game-headed:checkstyleMain SKIPPED
:lib:test-common:compileJava SKIPPED
:lib:test-common:processResources SKIPPED
:lib:test-common:classes SKIPPED
:lib:test-common:jar SKIPPED
:game-app:game-headed:compileTestJava SKIPPED
:game-app:game-headed:processTestResources SKIPPED
:game-app:game-headed:testClasses SKIPPED
:game-app:game-headed:checkstyleTest SKIPPED
:game-app:game-headed:pmdMain SKIPPED
:game-app:game-headed:pmdTest SKIPPED
:spotlessInternalRegisterDependencies SKIPPED
:game-app:game-headed:spotlessAllFiles SKIPPED
:game-app:game-headed:spotlessAllFilesCheck SKIPPED
:game-app:game-headed:spotlessJava SKIPPED
:game-app:game-headed:spotlessJavaCheck SKIPPED
:game-app:game-headed:spotlessCheck SKIPPED
:game-app:game-headed:test SKIPPED
:game-app:game-headed:check SKIPPED

```

PR #11514, opened 2023-02-28
Builds: Update game-app/run/check to dynamically find sub-projects

DanVanAtta:
This update will cause the following command to be run:

```
./game-app/run/../../gradlew --parallel :game-app:game-core:check :game-app:game-headless:check
:game-app:domain-data:check :game-app:map-data:check :game-app:game-headed:check
:game-app:smoke-testing:check :game-app:game-relay-server:check :game-app:ai:check
```

Previously we had to explicitly list out of all of these subprojects. This update assumes any folders under neath 'game-app' will be a dependency runs tests in all of those folders.

### DIFF
--- a/game-app/run/check
+++ b/game-app/run/check
@@ -2,16 +2,13 @@
 
 set -e
 
+
 scriptDir=$(dirname "$0")
-"$scriptDir/../../gradlew" --parallel \
-  :game-app:ai:check \
-  :game-app:domain-data:check \
-  :game-app:game-core:check \
-  :game-app:game-headed:check \
-  :game-app:game-headless:check \
-  :game-app:game-relay-server:check \
-  :game-app:map-data:check \
-  :game-app:smoke-testing:check
+
+SUB_PROJECTS=$(find $scriptDir/../ -name "build.gradle" -type f | xargs dirname | sed 's|.*\.\./|:game-app:|'   | sed 's/$/:check/' | tr '\n' ' ')
+
+"$scriptDir/../../gradlew" --parallel $SUB_PROJECTS
+
 
 "$scriptDir/.build/check-links"
 


DanVanAtta:
@RoiEXLab WDYT?

I'm really surprised that this behavior of running a task on a sub-project and all of its dependencies is not more readily built in. Searching for an answer there, I've only found stack-overflow answers that seem to chain things together using a recursive 'depends-on'

RoiEXLab:
Seems reasonable. What's the reason you're explicitly avoiding calling `gradle :game-app check`? The parallelism? Does it make a difference?

Also I feel like gradle should have some sort of feature to list all the tasks of a particularly module and filter them. Might be a little nicer (if it exists) than searching for gradle build files

DanVanAtta:
> What's the reason you're explicitly avoiding calling `gradle :game-app check`?


```
dan@desk:~/work/triplea$ ./gradlew :game-app check

* What went wrong:
Task 'game-app' not found in root project 'triplea'.

* Try:
> Run gradlew tasks to get a list of available tasks.
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 341ms
```

> Also I feel like gradle should have some sort of feature to list all the tasks of a particularly module and filter them. Might be a little nicer (if it exists) than searching for gradle build files

Yeah.. I'm still a bit shocked this doesn't have a natively supported solution to begin with.

If we grep & sed instead the '--dry-run'  output, we get a better list:

```
dan@desk:~/work/triplea$ ./gradlew -q :game-app:game-headed:check --dry-run 2> /dev/null | sed 's/^\(.*\):.*/\1:check/' | sort | uniq | sed  '/^$/d' | grep -v "^:check$"
:game-app:ai:check
:game-app:domain-data:check
:game-app:game-core:check
:game-app:game-headed:check
:game-app:game-relay-server:check
:game-app:map-data:check
:http-clients:feign-common:check
:http-clients:lobby-client:check
:lib:java-extras:check
:lib:swing-lib:check
:lib:test-common:check
:lib:websocket-client:check
:lib:websocket-server:check
:lib:xml-reader:check
```

BTW, are you running on Mac/Windows/Linux? If Windows, how are you dealing with all of this bash-fu? Cygwin?

RoiEXLab:
> If Windows, how are you dealing with all of this bash-fu? Cygwin?

That's the neat part, I don't. Jokes aside, if I need bash or any other Linux tooling i just use WSL with a regular Ubuntu distro

DanVanAtta:
Do you have docker installed @RoiEXLab ? I presume that gradle is fine to run natively?

I'm wondering if it is better to officially support:
- instruct developers to use cygwin
- use docker to run commands that need bash or python
- bake this into gradle

DanVanAtta:
In the short term, our run and check scripts are currently bash.

Updating it as described by parsing the dry-run output gives us a better project dependency list without having to maintain that dependency list by hand. So, that's good.

Though, in terms of standard tooling and ensuring we are enabling tools for everyone while we also enable those tools for ourselves, that falls short. Seems like we should do a survey of OS's, how people run bash (if at all), and IDE. With that data it'll be easier to know what we should officially bless as "the way"

RoiEXLab:
I have docker installed inside my WSL. In terms of developer setup convenience I think going with a pure gradle approach is the best solution (least dependencies, only java needs to be installed).
However I don't think groovy is necessarily the best language to maintain code in, so kotlin could be considered instead.
Docker is nice too, but developing code inside a docker image via volumes adds unecessary overhead in my opinion.

DanVanAtta:
@RoiEXLab  some questions for you:

(1) Could you post a reference link for WSL? I'm not familiar with it.

AFAIK docker is easy to install & run on windows. Is that not the case? In other words, why not run that directly on windows? 

(2) To the point of what I'm wondering, does having to switch to a WSL to run docker & "do more" cause a delay in your workflow? Could you be a bit faster if you did not have to switch things around? Using a WSL instead of a native CLI seems like it could potentially inconvenient. Assuming it is somewhat inconvenient - to what degree is it inconvenient?

(3) How can we support running the custom-check scripts (which are currently written in bash?) I don't think rewriting those into another language is going to be that great use of effort. I can see some reason to try and drop them even. Though short of that, is there a good path for those checks?

The custom-check scripts are another large bash dependency. I'm wondering how we can drop that dependency as well if we are going to try and drop the bash dependency altogether.

----------------


### is docker even a good choice for running linux stuff on any OS?

If running something in docker means you'll switch to WSL instead, and you can run bash more directly to WSL, then docker is seemingly not even a big win (unless we "strongly" urge and support docker directly on windows, but even then; is running docker single-command containers from windows that easy to do?)

DanVanAtta:
RE: WSL

I found this: https://www.commandlinewizardry.com/post/running-linux-commands-within-windows-using-wsl

It looks like WSL is just a really nice & integrated way to do cygwin. Seemingly WSL is potentially a good answer and enables bash & docker. It may make sense for WSL to be the official workflow on windows.

RoiEXLab:
WSL (Version 2 to be specific) is basically just a lightweight VM running a Linux distro of your choice where the windows filesystem is mounted as a volume and a special Linux kernel maps all kinds of IO and driver interaction to the windows environment.
So it's more than cygwin, somewhat recently hardware accelerated ui support was even introduced so you can even run native Linux software with GUI like gimp using next to no setup.

Lots of IDEs have integration for it. So IntelliJ Ultimate for example can just compile triplea using a JDK I installed for Ubuntu.

As far as I know the latest docker for windows internally uses WSL as well by default to to the virtualization. So you can still install docker "natively" on windows, but using apt is just more convenient for me. But obviously for tooling it's just an unnecessary step.

The only downside to relying on WSL is that sometimes there are some problems that arise from windows-linux differences like file endings. Git on windows by default tries to be smart about line endings and changes them to CRLF, which causes the scripts to not work. So in the past I had to run dos2unix on all the scripts before being able to run them.

I don't think we necessarily need write lengthy scripts for everything and instead can rely on some rather basic gradle syntax for most things. I just wanted to point out that if we end up having to automate a rather complicated task we might want to consider using kotlin instead of groovy because of the better tooling.

DanVanAtta:
>  Git on windows by default tries to be smart about line endings and changes them to CRLF, which causes the scripts to not work. So in the past I had to run dos2unix on all the scripts before being able to run them.

I think this is perhaps the big use-case behind gitattributes. AFAIK we can tell it to use LF for *.sh files. 

> I don't think we necessarily need write lengthy scripts for everything and instead can rely on some rather basic gradle syntax for most things. I just wanted to point out that if we end up having to automate a rather complicated task we might want to consider using kotlin instead of groovy because of the better tooling.

Agree. There is a cost to custom tooling, and a lot to be said about less and simple. Hence, any lengthy scripts we really have to think about and/or look for simpler ways to do the same.

PR #11566, opened 2023-04-02
Simplify CI - CI always runs verify script

DanVanAtta:
- Always run 'verify' script as part of CI builds
- Path specific builds are now only for merges to master to deploy latest builds.
- Convenience updates to 'verify' script, automatically install docker.


DanVanAtta:
@RoiEXLab  ; FYI, something of a notable update to the build process here. Simplified it quite a bit - just one 'verify' script that is run by CI now and can also be run locally. Path specific deployments will be project specific and only following merge to master. PR builds are no longer path specific, just the one 'verify' check.

PR #11567, opened 2023-04-03
see if this works lol

beelee1:

PR #11568, opened 2023-04-03
Revert "see if this works lol (#11567)"

beelee1:
This reverts commit c1d4dfc9e9279c0510253da9aa294015e4214dd5.

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2800,9 +2800,3 @@
   img: https://raw.githubusercontent.com/triplea-maps/adam_met_de_bois/master/preview.png
   description: |
    <br>Free for all fight the city of Amsterdam
-- mapName UHD World War II Global
-  mapCategory: EXPERIMENTAL
-  url: https://github.com/triplea-maps/uhd_world_war_ii_global/archive/master.zip
-  version: 1
-  img: https://github.com/triplea-maps/uhd_world_war_ii_global/master/preview.png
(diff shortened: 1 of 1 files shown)


beelee1:
hmm didn't realize revert would make a new one. close this too

DanVanAtta:
@beelee1  - git never deletes, you can only add an undo. A 'revert' is an undo.

Also recommend to try and be slightly more descriptive. Looking through the history log, commits like "try again" are going to be really vague. The commit history is used to try and track down bugs and understand why changes were made. Having it be somewhat descriptive is helpful. 

This is how that history appears now:
Screenshot from 2023-04-03 17-58-01

DanVanAtta:
@beelee1  looks okay.

Though, the formatting is a bit off: If you click the red 'x' on the latest build, and scroll down you'll see:

```
  -··description:·
          -···<br>This·is·UHD·World·War·II·Global·using·the·Black·Elk·Mega·Map
          +··description:
          +···<br>This·is·UHD·World·War·II·Global·using·the·Black·Elk·Mega·Map
```
That is saying you need to remove the lines marked with a '-' and add the lines marked with a '+'. Looks like there is a trailing space on 'description: ' that is doing it.

DanVanAtta:
@beelee1 generally, let the build checks complete before merging. There is a small window where you can merge things in before github sees it needs to run build checks. If there were a way to close this easily, we would. In the meantime, do let it do its thing and don't merge until after you get some green checks from the build.

PR #11580, opened 2023-04-11
 I updated the url of the eclipse_cleanup.xml file

JsonVladimir:
I updated the link of the xml file, when trying to enter the current url it generates an error 404, page not found.

image


RoiEXLab (approved):
Thanks, very appreciated 👍🏼

PR #11581, opened 2023-04-11
I updated the url of the "importorder" file

JsonVladimir:
I updated the link of the file, when trying to enter the current url it generates an error 404, page not found.
image


RoiEXLab (approved):
Thank you very much 👍🏼

PR #11622, opened 2023-05-06
Update AirBattleDefenseCombatValue.java

WCSumpton:
I believe that BattleState.Side should be Defense in AirBattleDefenseCombatValue


DanVanAtta:
Good catch  :+1:

Issue #10936, opened 2022-08-06
2.6+93c2e16: ConcurrentBattleCalculator#calculate:196 - java.lang.IllegalStateException


tripleabuilderbot:
## User Description
Game: Global 40 House Rules with Canada 
Engine Version: 2.6+93c2e16
Java Version: 11.0.9.1
Platform:  PC with Windows 11
Game Round: 24
Player: Hard AI - Anzac
Phase: Anzac Purchase Units

## Map
global_40_house_rules

## Log Message
java.lang.IllegalStateException: Round 10,000 reached in a battle. Something must be wrong. Please report this to TripleA.
 Attacking unit types: bomberA0C5, Defending unit types: submarine,transport

## TripleA Version
2.6+93c2e16

## Java Version
11.0.9.1

## Operating System
Windows 10

## Stack Trace
```
Exception: java.lang.IllegalStateException java.lang.IllegalStateException: Round 10,000 reached in a battle. Something must be wrong. Please report this to TripleA.
 Attacking unit types: bomberA0C5, Defending unit types: submarine,transport
java.lang.Exception
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
	at java.base/java.util.concurrent.ForkJoinTask.getThrowableException(ForkJoinTask.java:600)
	at java.base/java.util.concurrent.ForkJoinTask.reportException(ForkJoinTask.java:678)
	at java.base/java.util.concurrent.ForkJoinTask.invoke(ForkJoinTask.java:737)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateParallel(ReduceOps.java:919)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:233)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
	at games.strategy.triplea.odds.calculator.ConcurrentBattleCalculator.calculate(ConcurrentBattleCalculator.java:196)
[...]

PR #11736, opened 2023-07-05
Fix stalemate logic to end 0 att plane vs. transport fights.

asvitkine:

Without this, we would trigger the 10K rounds limit in AI simulation when the battle ends up with a strat bomber (that has no attack or defense) faced off against a transport and/or subs, and neither would be able to damage the other.

Added a unit test. Also renames a couple of existing tests for clarity.

Fixes: https://github.com/triplea-game/triplea/issues/10936

Note: This problem does not exist in 2.5 (tested).

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/battle/MustFightBattle.java
@@ -1301,6 +1301,10 @@ public class MustFightBattle extends DependentBattle
                 throw new IllegalStateException(
                     "Round 10,000 reached in a battle. Something must be wrong."
                         + " Please report this to TripleA.\n"
+                        + " Territory: "
+                        + battleSite
+                        + " Attacker: "
+                        + attacker.getName()
                         + " Attacking unit types: "
                         + attackingUnits.stream()
                             .map(Unit::getType)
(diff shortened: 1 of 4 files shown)


panther2:
@asvitkine
Why can't a strat bomber hit a transport? That's quite unusual.
That must be a spiciality of the specific map?

TheDog-GH:
Its from **Global 40 House Rules with Canada**  

Above is not my map, however I have tried something similar for one of mine.

Here is an xml extract
```
    <!-- BomberA0C5 -->
    <attachment name="unitAttachment" attachTo="bomberA0C5" javaClass="games.strategy.triplea.attachments.UnitAttachment" type="unitType">
      <option name="movement" value="6"/>
      <option name="isAir" value="true"/>
      <option name="isStrategicBomber" value="true"/>
      <option name="bombingMaxDieSides" value="6"/>
      <option name="canBeGivenByTerritoryTo" value="British"/>
      <option name="airAttack" value="0"/>
      <option name="canAirBattle" value="true"/>
      <option name="destroyedWhenCapturedBy" value="Germans:Russians:Americans:British:UK_Pacific:French:Italians:Japanese:Chinese:ANZAC:French:Dutch"/>
    </attachment>
```

asvitkine:
Right, sorry in this case the strat bomber has no att/def, only used to damage factories, etc.

PR #11744, opened 2023-07-07
Reverted the line #2419 return Optional.empty();

vignan8:


asvitkine:
The return value of the function is a boolean.

asvitkine:
I'm closing this PR since it doesn't make sense (trying to return an Optional when a boolean is required).
But feel free to re-open if I'm missing something!

PR #11759, opened 2023-07-10
Install4J updates

asvitkine:

- Switch to Install4J 10.0.6 (latest version) from 9.0.6
- Update bundled JDK version to latest java 11 patch (11.0.19+7)
- Switch Mac build to Universal (x86_64 + arm64) to support M1 macs

The bump to latest JDK11 patchlevel should also pick up a fix for this UI issue on Mac: https://bugs.openjdk.org/browse/JDK-8251377

## Release Note

CHANGE|Add arm64 native arch to Mac app.
FIX|Unreadable tab titles on Mac.
CHANGE|Switch bundled Java to 11.0.19+7.

### DIFF
--- a/game-app/game-headed/build.install4j
+++ b/game-app/game-headed/build.install4j
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<install4j version="9.0.6" transformSequenceNumber="9">
+<install4j version="10.0.6" transformSequenceNumber="10">
   <directoryPresets config=".triplea-root" />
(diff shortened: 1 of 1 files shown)


asvitkine:
This failed the build. I'm going to look at what needs to be done to fix it.

I see we're using some stuff from https://github.com/triplea-game/assets/tree/master/install4j

asvitkine:
Looks like it's not so easy. install4j_linux-x64_10_0_6.sh runs into GitHub's 100MB file size limit and GitHub then recommends using https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-git-large-file-storage for it instead, but I'm not sure how well that will work.

Let me revert this for now and can try to reland in pieces - for example, I'm not sure if we really need install4j 10.

Issue #11617, opened 2023-05-03
Error with "canNotTarget"


WCSumpton:
Using engine version: 2.6.14357
Windows 10

In the unitAttachment "isSea" is false and "isAir" is false and using "canNotTarget" produces the following error: "WARN  g.s.triplea.ui.BattleStepsPanel - Could not find step name:(attacking player) select casualties"
![CannotFindError](https://user-images.githubusercontent.com/17694690/235962482-0953ae5e-0d77-4931-b760-f1feea20a78f.png)
In the following save PoS2 game, 'infantry' unitAttachment was modified: <option name="canNotTarget" value="fighter"/>
(did not accept save game file)
After loading save game, set Russian to Human and play. Select "done" to end combatMove, and select "Ukraine S.S.R." for your battle. After the error message is cleared, Russian casualties can be selected normally, so the error does not prohibit the selection of casualties, nor does it stop game play.
The following <option name="isFirstStrike" value="true"/> was added:
(did not accept save game file)
Do the same as above, not their is no error now.

Cheers...

WCSumpton:
[ProblemWithcanNotTarget.zip](https://github.com/triplea-game/triplea/files/11384382/ProblemWithcanNotTarget.zip)
First file ProblemWithcanNotTarget.tsvg is used to show the error. Next file: ProblemFirstStrike.tsvg is with isFirstStrike added.

SupRemeviber:
Check you code and recode

TheDog-GH:
This Problem could do with a **2.6** label/tag otherwise it might get missed.

TheDog-GH:
@asvitkine 
This is still an issue.  
Having download TripleA-2.6+14388 and running the latest version of 1941 Command Decision v120
https://drive.google.com/file/d/1Q5XcTxMNdUZVAjjTWbz4mw0gL-Bjlw6r/view?usp=sharing

It still errors in about 1 minute.  Set all players to FASTAI.

asvitkine:
I think same error as this one:
https://github.com/triplea-game/triplea/issues/10647

I'll dupe that one over to this one, but note that there was some good discussion on that one too.

asvitkine:
I'm actually going to dupe this in the other direction. closing this as a dupe of https://github.com/triplea-game/triplea/issues/10647 since the other issue already has more investigation done.

asvitkine:
[...]

Issue #10647, opened 2022-06-12
TripleA 2.6.895 gives this error - Could not find step name


TheDog-GH:
This map works without error in 2.5

But 2.6.895 has the error below, by taking the defaults, it does not get past round 1.

With 2.6 it does not give me a proper error message or clue me where to look, please take a look.  The step error should work with uppercase **F**inns and **R**ussians, yes?

![image](https://user-images.githubusercontent.com/71022517/173256371-dbc3c824-1c7b-4e58-b044-2103f6f06c95.png)

Map
[WW2_Russia_Finland_1939.zip](https://github.com/triplea-game/triplea/files/8886942/WW2_Russia_Finland_1939.zip)

asvitkine:
So it's referring to steps in a battle. The list of steps the engine has in that situation is:

- Finns select casualties
- Finns notify casualties
- Finns units Elite fire
- Russians select units Elite casualties
- Russians notify units Elite casualties
- Finns units Forest fire
- Russians select units Forest casualties
- Russians notify units Forest casualties
- Remove casualties
- Russians withdraw?

So "Russians select casualties" is not one of them. I guess there's some customization of the step names somewhere and some code might not expecting for that to be the case and is looking for a normal casualties step?

asvitkine:
The step name comes from `getPossibleOldNameForNotifyingBattleDisplay()` which was added in this 2.6 PR:
https://github.com/triplea-game/triplea/commit/b804d3709224ae642c7e420d7d16cd85b372d225

It does a remapping of a step name, for example:
"Finns select air vs non subs casualties" gets mapped to "Finns select casualties"

Because the original string wasn't in the list of:
```
[Finns air vs non subs fire, Russians select air vs non subs casualties, Russians notify air vs non subs casualties, Finns fire, Russians select casualties, Russians notify casualties, Russians units Artillery fire, Finns select units Artillery casualties, Finns notify units Artillery casualties, Russians units Mountain fire, Finns select units Mountain casualties, Finns notify units Mountain casualties, Remove casualties, Finns withdraw?]
```

I'm really not familiar with this code, so maybe @trevan could chime in or @DanVanAtta who reviewed the above PR?

TheDog-GH:
This is the first time/map I have used **canNotTarget** and if I comment out its use the map runs without error.

As an aside
> Finns units Forest fire
> Russians select units Forest casualties

Forest is an **isInfrastructure** unit with no attack or defence, so should it even be in this list?

[...]

PR #11773, opened 2023-07-12
Fix issue a cause of "Could not find step name" (2.6 bug).

asvitkine:

This was happening because the code to generate step names was not excluding units that would not participate in combat, resulting in infrastructure units getting their own steps (which later did not match what the engine generated once the filtering took place).

Uses the same logic as what's done for the battle to exclude units.

This change required adjusting a bunch of tests that were previously not careful about which had mistakes in setting up mock battles where the units didn't match the territory (in terms of sea vs. land). Also makes the tests to mock game data properties leniently, so that only the ones being set to true need to be specified (removing lots of LOC).

Includes a unit test.

Fixes: https://github.com/triplea-game/triplea/issues/10647
Note: Doesn't fix https://github.com/triplea-game/triplea/issues/11617, as the root cause of that error is different, despite the actual error being the same. I will fix that in a separate PR.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/general/FiringGroupSplitterGeneral.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/general/FiringGroupSplitterGeneral.java
@@ -56,24 +56,34 @@ public class FiringGroupSplitterGeneral
 
   @Override
   public List<FiringGroup> apply(final BattleState battleState) {
-    final Collection<Unit> canFire =
-        CollectionUtils.getMatches(
-            battleState.filterUnits(ACTIVE, side),
-            PredicateBuilder.of(getFiringUnitPredicate(battleState))
-                // Remove offense allied units if allied air can not participate
-                .andIf(
-                    side == OFFENSE
-                        && !Properties.getAlliedAirIndependent(
-                            battleState.getGameData().getProperties()),
-                    Matches.unitIsOwnedBy(battleState.getPlayer(side)))
-                .build());
-
     final Collection<Unit> enemyUnits =
         CollectionUtils.getMatches(
             battleState.filterUnits(ALIVE, side.getOpposite()),
             PredicateBuilder.of(Matches.unitIsNotInfrastructure())
                 .andIf(side == DEFENSE, Matches.unitIsSuicideOnAttack().negate())
                 .andIf(side == OFFENSE, Matches.unitIsSuicideOnDefense().negate())
+                .build());
+
+    // Filter participants (same as is done in MustFightBattle.removeNonCombatants()), so that we
+    // don't end up generating combat step names for units that will be excluded.
+    final Predicate<Unit> canParticipateInCombat =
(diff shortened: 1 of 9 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/delegate/battle/steps/fire/general/FiringGroupSplitterGeneral.java
```diff
@@ -56,10 +56,28 @@ public enum Type {
...
   public List<FiringGroup> apply(final BattleState battleState) {
+    final Collection<Unit> enemyUnits =
+        CollectionUtils.getMatches(
+            battleState.filterUnits(ALIVE, side.getOpposite()),
+            PredicateBuilder.of(Matches.unitIsNotInfrastructure())
+                .andIf(side == DEFENSE, Matches.unitIsSuicideOnAttack().negate())
+                .andIf(side == OFFENSE, Matches.unitIsSuicideOnDefense().negate())
+                .build());
+
+    // Filter participants (same as is done in MustFightBattle.removeNonCombatants()), so that we
```
DanVanAtta, line 67: >    // Filter participants (same as is done in MustFightBattle.removeNonCombatants()), so that we
    // don't end up generating combat step names for units that will be excluded.
    
Is there a reasonable way to unify how combat step names are generated?
Overall though, fixing the two logic paths so that they match again, and so that they call each other out is good. If we don't unify the two pieces of code, we should then add a comment to `MustFightBattle.removeNonCombatants()` to indicate that this location should be updated as well to match.

asvitkine: I added a comment to the other location for now.

It *may* be possible to unify things, but it's tricky. Basically, the splitter code does talk to MustFightBattle and gets the units via the `filterUnits()` API. So that could theoretically be made to return the unit with non-combatants removed. But that API is also called from many different places, so such a change may not be safe. Another option would be to make another function next to `filterUnits()` that would do this and call that instead from here, but it's not really saving much duplication.

Cernelius:
Will these changes keep the ability of infrastructures to participate in combat beside being unable to take hits after the opening combat phase?

For example, will an aircraft infrastructure (an aircraft unit which is also an infrastructure) still risk to be shot down by AA fire when attacking a territory defended by an AA Gun?

For example, will an infrastructure with a positive offence or defence value still be able respectively to attack or to defend so long as at least one unit remaining in battle on its side is not an infrastructure? Same thing if the infrastructure is just providing (positive or negative) support (either normal or AA).

These are abilities which are irrelevant for all basic games but not in TripleA games in general.

Cernelius:
Also, sometimes infrastructure units which are not participating in combat at all need to remain in combat if only to be able to retreat with the rest. For example, the horse units of Feudal Japan.

asvitkine:
@Cernelius I believe so, since I'm using the logic that was already being used by the battle code for this which does look at whether some infra units should participate.

But I wouldn't mind if you can test these with the prerelease to make sure.

PR #11826, opened 2023-07-26
Fix: Apply PlayerAttachment stack limits when placing units.

asvitkine:

With this change, these restrictions are checked before showing the dialog of which units to place in a territory.
This way, invalid placements will be omitted from the dialog already, rather than showing up and causing an error when selected.

Tested on the "1941 Global Command Decision" map, where all the factory types have a stacking limit of 1 per territory. Before the change, if you build them and try to place units in a territory with a factory already, the factory would show up in the dialog. With the change, it will be omitted. (On this map, the presence of the factory in the list of units to place would also show an increased max for how many units can be placed for non-factory units - e.g. being able to select 3 infantry to place in a territory that only allows 2.)

This change refactors the code to move the stacking limits logic to a new UnitStackingLimitFilter class, combining the logic that was previously split between two places behind a better API that allows simplification of all the call sites.
Additional test coverage is added, both for the new class and for the PlaceDelegate, the latter also covering the sequence of the different calls (i.e. that placement restrictions are checked before stacking limit filtering).

Additionally, one call site no longer needed to do the checks as it was already doing the validation indirectly through another function it was calling.

The cleanup removes several hundred lines of code, although many lines of test code are also added,

## Release Note

FIX|Unit placement dialog will now filter out units that can't be placed due to stacking limits coming from the PlayerAttachment.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/UnitCollection.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/UnitCollection.java
@@ -226,19 +226,36 @@ public class UnitCollection extends GameDataComponent implements Collection<Unit
 
   @Override
   public boolean remove(final Object object) {
-    final boolean result = units.remove(object);
-    holder.notifyChanged();
-    return result;
+    final boolean changed = units.remove(object);
+    if (changed) {
+      holder.notifyChanged();
+    }
+    return changed;
+  }
+
+  @Override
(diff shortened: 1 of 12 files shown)


TheDog-GH:
I dont think this working as intended, see below 35 Fighters
image

asvitkine:
@TheDog-GH thanks for the quick testing.

Will take a look.

PR #11831, opened 2023-07-29
Fix validation of move/attack stacking limits from player attachment.

asvitkine:
Fixes issue in my last PR where these were not being applied due to incorrect filtering of the unit list.


asvitkine:
@TheDog-GH
This should fix the issue you discovered. Thanks for testing!

TheDog-GH:
Yes it does, Air, Sea & Land units are being limited as intended.

PR #11853, opened 2023-08-04
Adding "haveResources" and "checkAIPlayer" conditions

WCSumpton:

This change allows map creators to check if a player is AI controlled. "checkAIPlayer" uses isAI() method which is already part of rulesAttachment.

This change allows map creators to check the amount of resources a player has. resources is imported to allow "haveResources" to check the resources.

Cheers...


asvitkine:
Did you mean to close this?

Issue #11851, opened 2023-08-03
ConcurrentBattleCalculator#calculate:196 - java.lang.IllegalStateException


tripleabuilderbot:
## User Description
Never seen this happening or this game having any crashes with the AI using 2.5.

## Map
270bc_wars

## Log Message
java.lang.IllegalStateException: Attackers and defenders collections must be distinct with no duplicates. This helps catch logic errors in AI code that would otherwise be hard to debug.

## TripleA Version
[2.6.14423](https://github.com/triplea-game/triplea/releases/tag/2.6.14423)

## Java Version
11.0.19

## Operating System
Windows 10

## Stack Trace
```
Exception: java.lang.IllegalStateException java.lang.IllegalStateException: Attackers and defenders collections must be distinct with no duplicates. This helps catch logic errors in AI code that would otherwise be hard to debug.
java.lang.Exception
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
	at java.base/java.util.concurrent.ForkJoinTask.getThrowableException(ForkJoinTask.java:600)
	at java.base/java.util.concurrent.ForkJoinTask.reportException(ForkJoinTask.java:678)
	at java.base/java.util.concurrent.ForkJoinTask.invoke(ForkJoinTask.java:737)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateParallel(ReduceOps.java:919)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:233)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
	at games.strategy.triplea.odds.calculator.ConcurrentBattleCalculator.calculate(ConcurrentBattleCalculator.java:196)
	at games.strategy.triplea.ai.pro.util.ProOddsCalculator.callBattleCalc(ProOddsCalculator.java:250)
	at games.strategy.triplea.ai.pro.util.ProOddsCalculator.callBattleCalc(ProOddsCalculator.java:226)
[...]

PR #11854, opened 2023-08-05
Linkify version in TripleA error reports.

asvitkine:

Should produce reports like this one (which I manually edited):
https://github.com/triplea-game/triplea/issues/11851


asvitkine:
It would be nice if we could also linkify the stack traces to produce things like:

	at games.strategy.triplea.ai.pro.ProNonCombatMoveAi.doNonCombatMove([ProNonCombatMoveAi.java:109](https://github.com/triplea-game/triplea/blob/2.6.14423/game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProNonCombatMoveAi.java#L109))

But the above is not easily possible because that link requires the full filepath, while we only have the filename from the stack trace...

PR #11857, opened 2023-08-05
Adding "checkAIPlayer" and "haveResources"

WCSumpton:

Adding "checkAIPlayer" and "haveResources" to RulesAttachment.

checkAIPlayer allows map makers to query isAI() about player.

haveResources allow map makers to query resources.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProOddsCalculator.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/util/ProOddsCalculator.java
@@ -14,7 +14,6 @@ import games.strategy.triplea.delegate.TerritoryEffectHelper;
 import games.strategy.triplea.odds.calculator.AggregateResults;
 import games.strategy.triplea.odds.calculator.IBattleCalculator;
 import games.strategy.triplea.util.TuvUtils;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import org.triplea.java.collections.CollectionUtils;
@@ -253,7 +252,7 @@ public class ProOddsCalculator {
             t,
             attackingUnits,
             defendingUnits,
-            new ArrayList<>(bombardingUnits),
+            bombardingUnits,
             TerritoryEffectHelper.getEffects(t),
             retreatWhenOnlyAirLeft,
             runCount);
--- a/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculator.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculator.java
@@ -14,6 +14,8 @@ import games.strategy.triplea.delegate.battle.BattleResults;
 import games.strategy.triplea.delegate.battle.BattleTracker;
 import games.strategy.triplea.delegate.battle.MustFightBattle;
 import games.strategy.triplea.util.TuvCostsCalculator;
+import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
@@ -68,13 +70,13 @@ class BattleCalculator implements IBattleCalculator {
               : gameData.getPlayerList().getPlayerId(defender.getName());
       final Territory location2 = gameData.getMap().getTerritory(location.getName());
       final Collection<Unit> attackingUnits =
-          GameDataUtils.translateIntoOtherGameData(attacking, gameData);
+          translateCollectionIntoOtherGameData(attacking, gameData);
       final Collection<Unit> defendingUnits =
-          GameDataUtils.translateIntoOtherGameData(defending, gameData);
+          translateCollectionIntoOtherGameData(defending, gameData);
       final Collection<Unit> bombardingUnits =
-          GameDataUtils.translateIntoOtherGameData(bombarding, gameData);
+          translateCollectionIntoOtherGameData(bombarding, gameData);
       final Collection<TerritoryEffect> territoryEffects2 =
-          GameDataUtils.translateIntoOtherGameData(territoryEffects, gameData);
+          translateCollectionIntoOtherGameData(territoryEffects, gameData);
       gameData.performChange(ChangeFactory.removeUnits(location2, location2.getUnits()));
       gameData.performChange(
           ChangeFactory.addUnits(location2, mergeUnitCollections(attackingUnits, defendingUnits)));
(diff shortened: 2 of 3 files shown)


asvitkine:
Could you also add some unit tests?

WCSumpton:
> Could you also add some unit tests?

Now there you got me. Sorry please inform how to preform "unit test". I have tested both conditions and work as expected.

Cheers...

game-app/game-core/src/main/java/games/strategy/triplea/attachments/RulesAttachment.java
```diff
@@ -1001,6 +1072,14 @@ private boolean matchTerritories(
     return numberMet >= getTerritoryCount();
   }
 
+  private boolean checkCheckAIPlayer(final List<GamePlayer> players) {
+    boolean bcheck = true;
+    for (GamePlayer player : players) {
+      bcheck = false ? false : player.isAi();
```
asvitkine, line 1078: This doesn't look right. The "false ? false" part does nothing, so the code is the same as:

```bcheck = player.isAi();```

So that means it will only check the last player in the list.
I assume you want to it either check that it's true for all the players or for at least one. Do you know if there's a precedent for such conditions for which one it should be doing?

File: game-app/game-core/src/main/java/games/strategy/triplea/attachments/RulesAttachment.java [hunk 2]
```diff
@@ -1031,6 +1110,20 @@ private boolean checkTechs(final GamePlayer player, final TechnologyFrontier tec
     return found >= techCount;
   }
 
+  private boolean checkHaveResources(final List<GamePlayer> players) {
+    final boolean toSum =
+        haveResources[1].equalsIgnoreCase("sum") || haveResources[1].equalsIgnoreCase("add");
```
asvitkine, line 1115: This logic looks like it treat `sum` and `add` the same. Is that intentional?
If so, why? Is it for consistency with something else?

game-app/game-core/src/main/java/games/strategy/triplea/attachments/RulesAttachment.java
```diff
@@ -1057,6 +1150,12 @@ public MutableProperty<?> getPropertyOrNull(String propertyName) {
             this::setRelationship,
             this::getRelationship,
             this::resetRelationship);
+      case "checkAIPlayer":
```
asvitkine, line 1153: Why not a simpler name like "isAIPlayer"/

asvitkine:
> > Could you also add some unit tests?
> 
> Now there you got me. Sorry please inform how to preform "unit test". I have tested both conditions and work as expected.
> 
> Cheers...

Unit tests are test cases in code that verify the behavior.
For example, you have a bunch of logic to parse the haveResource property in setHaveResources(). The unit tests would call that function with different values and check if the values are correctly checked or not by the code.

Looks like we don't have a `RulesAttachmentTest.java` file already, but you can check `UnitAttachmentTest.java` for an example. In that file, a test case like `setWhenCapturedChangesIntoWithInvalidLength` is checking that property for some invalid strings.

So I suggest adding a RulesAttachmentTest.java file (next to UnitAttachmentTest.java) that can test the above. If you're using IntelliJ, it will have a green triangle in the UI next to the line of code of a given test to run that test.

game-app/game-core/src/main/java/games/strategy/triplea/attachments/RulesAttachment.java
WCSumpton: The pass-through value is "false". In this case if the map maker wishes to use the "players" option, to check if a 'team', "Americans" and "British" were both AI controlled. If one 'team' member is not, then the "false" value must be maintained. Thus if bcheck is false, it must stay false. 
~~~
<option name="checkAIPlayer" value="false"/>
~~~
Also does nothing.

Hope this explains the logic.

Cheers...

asvitkine: This code isn't checking for the value "false" though.

You're not doing something like `someValue.equals("false")`.

```
bcheck = false ? false : player.isAi();
```
Is the same as:
```
if (false) {
  bcheck = false;
} else {
  bcheck = player.isAi();
}
```

And the above is the same as:
```
bcheck = player.isAi();
```

asvitkine: > to check if a 'team', "Americans" and "British" were both AI controlled

So that sounds like you want to check that *both* players are AI controlled.
Please document that with a comment.

I think your code was supposed to do:
```bcheck = (bcheck == false) ? false : player.isAi();```

Which btw is the same as this shorter version:

```bcheck = bcheck && player.isAi();```

asvitkine (changes requested):

game-app/game-core/src/main/java/games/strategy/triplea/attachments/RulesAttachment.java
WCSumpton: Yes add and sum mean the same thing. Yes, it is intentional. They can be used interchangeably without forcing the map makers to chose.

Cheers...

game-app/game-core/src/main/java/games/strategy/triplea/attachments/RulesAttachment.java
WCSumpton: It could be changed to "isAIPlayer", but that implies that with a value of "false" could be used to check for **not** an AI player. I use "checkAIPlayer" to infer that a value of "false" means not to check, and is the pass-through value.

Cheers...

game-app/game-core/src/main/java/games/strategy/triplea/attachments/RulesAttachment.java
WCSumpton: You are right!  The line should be changed. Would I need to remove this merge and submit a new one to do this correction.

WCSumpton: Line changed to read 
`bcheck = !bcheck ? false : player.isAi()`
because if both bcheck and player.isAi() are false then 
`bcheck = bcheck && player.isAi()` would equal true which is not what is wanted.

asvitkine: If bcheck and player.isAi() are false, then the result will be false.

(false && false) == false

WCSumpton: Really? I always thought false and false equals true. But we are saying "and" here and not equals, so that may be what is confusing to me.

Am creating testing for RulesAttachment. 

Cheers...

PR #11887, opened 2023-08-18
PlacePanel: Limit max based on unit requirements.

asvitkine:
This change caps the max for each unit type in the place panel to the max that can be placed for that type when considering "units which require units" logic (i.e. factory types).

This partially prevents the error "Cannot place more units which require units, than production capacity of territories with the required units" after making the selection in the place panel.

Note: That error can still happen when selecting multiple units of different types where their total surpasses the max.

Tested on 1941 Global Command Decision by purchasing a bunker and some infantry and confirming that the place panel limits number of infantry to be placed to the amount that won't give the error (as opposed to that amount + 1, due to the bunker).

## Release Note

CHANGE|Place dialog will now show the correct limit for a unit type to be placed on maps using canProduceXUnits and requiresUnits properties.


Cernelius:
This can also be tested using my 270BC Wars game.

For testing it, buy 1 metropolis and 2 (or any number greater than 1 of) irregulars with Cartage and select a value 1 (or any number lower than the number of irregulars except 0) territory.

For example, if you buy 1 metropolis and 2 irregulars and then try to place on Melita, clicking "Max" prompts you to place 2 irregulars (and you cannot). For faster testing, de-use (untick) the barbarians.

Cernelius:
I tested, and my "270BC Wars" presents the same issue as I described (nothing of it changed). I thought this would fix that issue but maybe this is not related to what at that game?

PR #11896, opened 2023-08-21
Allow tuv to be set to 0.

WCSumpton:
This change allows tuv to be set to 0 and be used for calculations. Also, a check was added for the proper values.

tuv default setting is -1. There was no check for proper values. And when the tuv is calculated, the check was for greater than '0'. Please see [here](https://forums.triplea-game.org/topic/159/tuv-for-units-that-have-consumesunits-property/37?page=2), for the discussion with @Cernelius.

Note: PoS2 should be updated, showing the proper values.

Cheers... 

CHANGE|Allows a zero (0) value to be valid when using the "tuv" option for unitAttachment.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/attachments/UnitAttachment.java
@@ -2516,7 +2516,12 @@ public class UnitAttachment extends DefaultAttachment {
     return Tuple.of(max, s[1].intern());
   }
 
-  private void setTuv(final String s) {
+  private void setTuv(final String s) throws GameParseException {
+    final int value = getInt(s);
+    if (value < -1) {
+      throw new GameParseException(
+              "tuv must be 0 positive (or -1, default, to calculate) " + thisErrorMsg());
+    }
     tuv = getInt(s);
   }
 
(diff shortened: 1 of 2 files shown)


asvitkine:
Thanks!

Can you add a release note message per the instructions here?
https://github.com/triplea-game/triplea/blob/master/docs/development/reference/pr-release-notes.md

Also, it would be helpful to describe what this effects exactly.
From the linked thread, I guess the goal is to have the engine really treat the TUV as 0 and not try to calculate it like it does when it's not provided? (Does this just affect the player table display for information purposes or something else too?)

Cernelius:
> Also, it would be helpful to describe what this effects exactly. From the linked thread, I guess the goal is to have the engine really treat the TUV as 0 and not try to calculate it like it does when it's not provided? (Does this just affect the player table display for information purposes or something else too?)

Beside the AI (which I assume should be influenced in its choices by seeing the unit as having the specified value, depending on how the AI works), I believe that the only not merely informative application is for testing options requiring a minimum TUV destruction to return true in conditions (or objectives).

WCSumpton:
> Also, it would be helpful to describe what this effects exactly.
> From the linked thread, I guess the goal is to have the engine really treat the TUV as 0 and not try to calculate it like it does when it's not provided? (Does this just affect the player table display for information purposes or something else too?)

That is correct, this will allow a TUV value to be set to 0, and calculation are done when set to less then 0. Now a value of 100, will not be calculated. A value of 0, -1 (default) and -100 will be calculated, but are all valid values for TUV. So this introduces a GameParseException (Error Check) with a message explaining the proper values. This value is also used in destroyedTUV, and is also used by the AI when calculating battles and purchases.

WCSumpton:
CHANGE|Allows a zero (0) value to be valid when using the "tuv" option for unitAttachment.

WCSumpton:
As to the suggestion that a negative value be allowed. That would also be possible, but I feel that further discussion is needed about how this would affect other places that TUV is used.

Issue #11962, opened 2023-09-14
Add a Decimal to the Wins and Draw Results in the Battle-Simulator


Cernelius:
I suggest adding a decimal to the Wins and Draw results in the battle-simulator.

Details and further feature requests:
https://forums.triplea-game.org/topic/2604/less-approximative-battle-calculator

Cernelius:
(This is completed but the feature request in forum is still open, for example for upping the max Run Count to 100,000, which becomes a lot more relevant with added decimals.)

PR #11978, opened 2023-09-21
Show up to 2 decimals for percents in battle calc.

asvitkine:

The extra precision is only shown when significant (which may require a different sim count than the default of 200).

As requested here: https://github.com/triplea-game/triplea/issues/11962

## Release Note

NEW|Up to two decimal digits after the decimal point are now shown for percentages in the battle calculator.


Cernelius:
Cool, but what are exactly the Run Count intervals for having 0, 1 and 2 decimals, and do they differ between dice and LL?

Cernelius:
I tested it out some these new settings, and now I tend to think that 2 decimals are too much and a bit distracting when running at 10,000 or even 20,000 simulations: I would limit to 1 decimal. 2 decimals would be good for running at 100,000 thereabout though.

PR #12039, opened 2023-10-15
Bump wiremock to 2.35.1

asvitkine:
This seems to be the latest security release from the 2.x line.

Also switches to wiremock-jre8 since the updated 2.x version is only published under that name.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -87,7 +87,7 @@ subprojects {
         sonatypeGoodiesPrefsVersion = '2.3.6'
         substanceVersion = '4.5.0'
         wireMockJunit5Version = '1.3.1'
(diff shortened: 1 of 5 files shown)


asvitkine:
Odd, I guess 2.35.1 is not published?
https://repo.maven.apache.org/maven2/com/github/tomakehurst/wiremock/

But it theoretically exists here.

asvitkine:
I filed https://github.com/wiremock/wiremock/issues/2446 with wiremock to see if that's expected.

asvitkine:
Abandoning. Same error as 3.0.1, so I'm just going to look at making 3.0.1 work.

PR #12077, opened 2023-11-06
Verify script: run gradlew with --parallel

DanVanAtta:

### DIFF
--- a/verify
+++ b/verify
@@ -50,6 +50,6 @@ set -eu
 
 checkDependency "docker" || installDocker
 "$scriptDir/.build/setup-database"
-"$scriptDir/gradlew" check $@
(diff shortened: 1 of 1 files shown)


DanVanAtta:
I think --parallel works locally because successful tests will not necessarily be re-run, allowing other tests to pass.

Some options to resolve:
- find a way to have gradle not run tests of two sub-projects at the same time
- run multiple databases and let each test set run against a different database

asvitkine:
How much do we care about making this work? If it's too much trouble, maybe best to just abandon?

DanVanAtta:
I've been experimenting around a bit more, it's been difficult to get major gains with '--parallel'

Issue #10693, opened 2022-06-18
TripleA-2.6.929 show Zoom percentage request


TheDog-GH:
@asvitkine 
Im really liking the new 200%.  👍 

But could we have the percentage of the zoom shown on screen for say 5 seconds and then it disappear?

Perhaps it could go here on the **X** (to left of the current Player, for a short duration?

![Screenshot 2022-06-18 092338](https://user-images.githubusercontent.com/71022517/174429474-f730b131-c261-4d07-a2bf-733abd50f825.png)

Then the player can return to 100% easily and also remember what zoom percentage works best on certain maps.
Thanks in advance.

FedaSkywalker:
Hello, I’d like to try do this feature. Can you assign t to me?

asvitkine:
@FedaSkywalker done

FedaSkywalker:
Good. Thanks !

asvitkine:
I think this was implemented by https://github.com/triplea-game/triplea/pull/12082.

PR #12082, opened 2023-11-06
TripleA-2.6.929 show Zoom percentage request #10693

FedaSkywalker:
A zoom option was added onto a bottom bar. A new listener was added upon zooming as well.

### DIFF
--- a/game-app/game-headed/src/main/java/games/strategy/triplea/ui/menubar/ViewMenu.java
+++ b/game-app/game-headed/src/main/java/games/strategy/triplea/ui/menubar/ViewMenu.java
@@ -326,7 +326,7 @@ final class ViewMenu extends JMenu {
   }
 
   private void addShowZoomMenu() {
-    final JCheckBoxMenuItem showMapZoomBox = new JCheckBoxMenuItem("Show Map Zoom");
+    final JCheckBoxMenuItem showMapZoomBox = new JCheckBoxMenuItem("Show Zoom Percentage");
 
     showMapZoomBox.addActionListener(
         e -> {
--- a/game-app/game-headed/src/main/java/games/strategy/triplea/ui/menubar/ViewMenu.java
+++ b/game-app/game-headed/src/main/java/games/strategy/triplea/ui/menubar/ViewMenu.java
@@ -85,6 +85,7 @@ final class ViewMenu extends JMenu {
     }
     addShowMapDetails();
     addShowMapBlends();
+    addShowZoomMenu();
     addMapFontAndColorEditorMenu();
     addChatTimeMenu();
     addShowCommentLog();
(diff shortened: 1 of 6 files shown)


game-app/game-headed/src/main/java/games/strategy/triplea/ui/menubar/ViewMenu.java
```diff
@@ -52,486 +52,488 @@
...
+                                            null,
+                                            new String[] {"OK", "Cancel"},
+                                            0);
+                            if (result != 0) {
+                                return;
+                            }
+                            final Number value = (Number) model.getValue();
+                            frame.getMapPanel().setScale(value.doubleValue() / 100);
+
+                            frame.getGame().getData().notifyMapZoomChanged(value.intValue());
```
DanVanAtta, line 169: WDYT of setting the value directly on `bottomBar`?

EG, something like:

```
              frame.getBottomBar().setMapZoom(value);
```

I see this as advantageous for two reasons:
-  less 'plumbing' 
-  no need to route a UI concern through `GameData`

DanVanAtta:
@FedaSkywalker  thank you for the PR!

Looks like you probably need to run `./gradlew spotlessApply` to apply formatting, and perhaps install 'google-java-format' in your IDE s well.

Could you attach a screenshot of what this update looks like? Assuming that UX looks good, I reviewed the code diff and have a suggestion for to simplify the submission.

FedaSkywalker:
Hello @DanVanAtta, thx for your review. I tried to do it using ``` frame.getBottomBar().setMapZoom(value);``` but it didn't work quite perfectly. This zoom setter only worked in the ViewMenu map zoom setting. The map zoom value can be loaded from a saved game. You can also change the map zoom using the scroll wheel. So I think using it with the zoom listeners is better. I have moved the map zoom listener to TripleAFrame. Now the zoom label is updated in real time every time setScale() is called.   For UI check screens 
Screenshot 2023-11-15 at 08 47 31
Screenshot 2023-11-15 at 09 45 10

DanVanAtta:
Thanks for the updates @FedaSkywalker  and for working on a PR to begin with!

TBH I'm not sure if the zoom percentage is so salient that someone needs to see it all the time. Are we cluttering the UI? For sure zoom percentage is interesting information & is useful to know when opening a map or changing it. Though, I wonder if someone has been playing for an hour or three, whether they would still want to see zoom %.

I wonder if there is maybe somewhere else on the UI that would be nicer? Perhaps the community on the forums could help give suggestions even.

@FedaSkywalker  As a compromise, WDYT if we put this update behind a flag? Simply add a setting that can be turned on to show this - and let's ask the community for feedback? Adding a game option setting is mostly straight forward. This might be better as an option in the view menu, eg: "show zoom percentage"

FedaSkywalker:
I've update text from "Show Map Zoom" to "Show Zoom Percentage" 
updated text

DanVanAtta (approved):
:+1:

TheDog-GH:
@FedaSkywalker It works, thanks for doing this!

However on my PC it scrolls in 9s that is 99, 89 79, even when I reset it to 100%, I know so ungrateful 🙄 
As an aside when I do View> Map Zoom> and preset Reset I get this none critical error

Issue #11819, opened 2023-07-26
Feature request: Save games should save pending purchase selections


asvitkine:
This is not a regression - I tested on 2.5 and it behaves the same as 2.6.

When you save a game in the middle of a purchase phase (before clicking done), your purchase selections do not get saved.
Ideally, they should - similar to pending moves during a move phase.

zemjak:
Hi, can you please assign the issue to me if it's still available? Thank you in advance!

asvitkine:
@zemjak Dove. Let me know if you have ideas around how to do it that you'd like a second opinion on!

PR #12084, opened 2023-11-07
Save games should save pending purchase selections #11819

zemjak:
Hi, I hope it works, I tried it on several games - I got to the purchase phase and saved the game... reloaded it again from the file and the data was loaded correctly. Then I tried to play the game further and get into another purchase phase to see if saving to file doesn't affect other purchase phases and actually it does not affect them.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/PurchaseDelegate.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/PurchaseDelegate.java
@@ -161,7 +161,7 @@ public class PurchaseDelegate extends BaseTripleADelegate
   }
 
   @Override
-  public String  purchase(final IntegerMap<ProductionRule> productionRules) {
+  public String purchase(final IntegerMap<ProductionRule> productionRules) {
     final IntegerMap<Resource> costs = getCosts(productionRules);
     final IntegerMap<NamedAttachable> results = getResults(productionRules);
     if (!canAfford(costs, player)) {
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/PurchaseDelegate.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/PurchaseDelegate.java
@@ -52,6 +52,7 @@ public class PurchaseDelegate extends BaseTripleADelegate
       Comparator.comparing(o -> (UnitType) o.getAnyResultKey(), new UnitTypeComparator());
 
   private boolean needToInitialize = true;
+  private IntegerMap<ProductionRule> pendingProductionRules;
 
   @Override
   public void start() {
@@ -102,6 +103,7 @@ public class PurchaseDelegate extends BaseTripleADelegate
   @Override
   public void end() {
     super.end();
+    pendingProductionRules = null;
     needToInitialize = true;
   }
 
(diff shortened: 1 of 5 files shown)


asvitkine (commented):
Thanks for working on this. I think this is very promising, but I have a few suggestions.

asvitkine:
Hey, just wanted to ask if you saw my review comments and suggestions? I think it would be great to get this improvement landed so let me know if anything I said is unclear. Once you updated the PR, happy to take another look!

zemjak:
Hello, yes, your recommendations are quite clear to me, I have already managed to implement some of them, I still need to check locally if it will work in all cases. I will try to push the code in 2 days at the latest.

zemjak:
Hi, I'm sorry for waiting.. I tried to implement your suggestions mainly in class PurchaseDelegate.. maybe I should have used GameData.Unlocker as you recommended, but I was mainly inspired by usage of PoliticsDelegate class in AbstractAi class - where is not used that (un)lock. I hope that it is getting closer to the result we wanted.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/PurchaseDelegate.java
```diff
@@ -157,7 +161,7 @@ protected boolean canAfford(final IntegerMap<Resource> costs, final GamePlayer p
   }
 
   @Override
-  public String purchase(final IntegerMap<ProductionRule> productionRules) {
+  public String  purchase(final IntegerMap<ProductionRule> productionRules) {
```
asvitkine, line 164: Revert the extra space.

game-app/game-headed/src/main/java/games/strategy/triplea/ui/PurchasePanel.java
```diff
@@ -53,6 +54,12 @@ public class PurchasePanel extends ActionPanel {
         public void actionPerformed(final ActionEvent e) {
           final GamePlayer player = getCurrentPlayer();
           final GameData data = getData();
+          final PurchaseDelegate purchaseDelegate = data.getPurchaseDelegate();
+
+          // load pending production from file it there is any
+          if (purchaseDelegate.getPendingProductionRules() != null) {
```
asvitkine, line 60: Do we need the null check?

Just thinking this through:
  1. Initially it's already null in this class, so null check is not needed (since we'd just be assigning null to something already null).
  2. If it's not null, that means actionPerformed() already ran, but if it did, then purchaseDelegate.setPendingProductionRules() would have been called, so it should never be null.

game-app/game-headed/src/main/java/games/strategy/triplea/ui/PurchasePanel.java
```diff
@@ -53,6 +54,12 @@ public class PurchasePanel extends ActionPanel {
         public void actionPerformed(final ActionEvent e) {
           final GamePlayer player = getCurrentPlayer();
           final GameData data = getData();
+          final PurchaseDelegate purchaseDelegate = data.getPurchaseDelegate();
+
+          // load pending production from file it there is any
```
asvitkine, line 59: This comment isn't quite right, since we're not loading from file here - it was already loaded from file.

I think it should save something " // Restore pending production that was loaded from the save game."

asvitkine (commented):
Thanks, this looks reasonable, I think.

zemjak:
You are right with that condition, I'm gonna remove it.

PR #12157, opened 2023-11-30
Dev Builds: update nginx config for maps server redirect

DanVanAtta:
Adds mapping for the nginx launched by docker compose to redirect /maps requests to a 'maps-server' (will update docker-compose in a future update to launch that server)


DanVanAtta:
Ah, this cannot be separated from other updates..

PR #12208, opened 2023-12-16
Resize of battle window and rework of panels inside

bblazek98:
Minimal window height was changed to suit bigger monitors.
Handling of panels in Battle window was changed so casualties in bigger maps fit in properly.


TheDog-GH:
@asvitkine Please give this a check

beelee1:
Nice ! Looks like that took a lot of work. Yea, right now I have to use the mouse to make it as wide as possible, so I can see how many of each unit there is for battles with a lot of different units.

asvitkine:
I can look at the logic changes, but I noticed you have a lot of formatting changes that don't match the style rules for the project.

Can you run `gradlew spotlessApply` to format the code per the project style and update the PR? I can then review later when I have a chance.

PR #12246, opened 2024-01-03
Update servers.yml - Critical dontations message

DanVanAtta:


Cernelius:
Should this be reverted now that it is not any longer actual (telling the truth as of now)?

PR #12254, opened 2024-01-06
Update triplea_maps.yaml UHD World War II Global Version 2

beelee1:
See if the description image will show up and if the map is available in the lobby.

Also see if it can be downloaded with 2.6 prerelease


beelee1:
It appears gradle is outdated ?

beelee1:
I just had a firmware update that i've never encountered before. 

Idk if that will make a difference.

I'll let this stew then close and try again

beelee1:
hmm ... well idky but it passed now

PR #12353, opened 2024-02-21
Restore 2.5 map list code; 2.6 code is put behind a setting.

asvitkine:
This restores the 2.5 codepath for downloading the map listing, which fetches the yaml listing file directly from github.
This codepath lacks some functionality that the 2.6 codepath has, such as showing map sizes and being able to have accurate progress bars.

But it unbreaks map downloads on 2.6, since it's not working otherwise right now due to server issues. (And when the server was still up, it had trouble showing new maps.)

The "Use Maps Server (Beta)" setting is re-introduced, which can be toggled to use the new codepath.

It's not a straight up revert of the old codepath, since the old code needed some adjustment to work with the code structure changes in 2.6.


asvitkine:
So I'm thinking we can re-add a setting that switches between this codepath and the current 2.6 codepath. If the 2.6 codepath becomes reliable, we can switch back to it.

WCSumpton:
Will this help with the release of 2.6?

beelee1:
Will i be able to play my new map in the lobby now ?

TheDog-GH:
So it looks like the 2.5 map downloads are working.  I know its 2.5 list as I did **not** add my two maps 
Shogun Advanced and 1941 Global Command Decision 
It even unzips the files and adds a 2.6 map.yml file, so looking very good.

PR #12365, opened 2024-02-25
Update triplea_maps.yaml

TheDog-GH:
Thanks!
Amended UHD Global World War II
Added 1941 Global Command 
Added Shogun Advanced


TheDog-GH (commented):
OK

PR #12392, opened 2024-03-03
Update triplea_maps.yaml

beelee1:
fixing map image not showing up in description


beelee1:
This is for UHD World War II Global

Issue #12376, opened 2024-02-29
2.6.14650 Bid ships cannot be placed in a SZ that you own with your own ships


TheDog-GH:
Map=**1941 Global Command Decision**
Currently there is no error message when trying to place a ship in 113 B Sea Zone near Denmark it just does not work
Placing land unit works.

**To reproduce it**
Load Map
Place a Bid for Germany of say 40
Buy Infantry & Ships
Try and place a Ship in 113 B Sea Zone (as it is owned and has own ships) - it fails
Placing land unit works

An old map like world_war_ii_v3 bidding/placing works in 2.6 :-/
But 1941 GCD uses lots of new features and has 2 0EndTurn & 1EndTurn & isAI amongst others

Attached is a log recording the failed bid placement
[triplea.log](https://github.com/triplea-game/triplea/files/14452440/triplea.log)

Link to the latest 1941 Global Command Decision v161
https://drive.google.com/file/d/1F0Ci5ESw21oV_Aa-dBJted0hwa7caoPQ/view?usp=sharing

For reference this forum post by wc_sumpton
https://forums.triplea-game.org/topic/3326/1941-global-command-decision-official-thread/604

@asvitkine

asvitkine:
The error message in the log is unrelated.

Looks like the issue is happening because the ship requires Industry-Hvy and that doesn't exist in the sea zone being selected...

(But looks like there's an adjacent one, so I guess the expectation is the adjacent one is supposed to be counted? Does it work that way for regular placement, outside of Bid?)

asvitkine:
Non-Bid placement works correctly.

Looks like the issue is that `getAllProducers()` from BidPlaceDelegate doesn't consider adjacent territories. I'm not sure why it worked in 2.5, I guess the code was different somehow.

TheDog-GH:
@asvitkine 
Please reopen and take following into consideration
wc_sumpton and myself think that the check in 2.6 **for factories/production should be removed.**
see here for our conclusions and a few posts following
https://forums.triplea-game.org/topic/3326/1941-global-command-decision-official-thread/614

To repeat Cernel bullet points here
* You can place infinite per territory or sea zone
* You can place in any territory you own (no factory or anything required)
* You can place in any friendly sea zone with ship of the same owner already in it

asvitkine:
Thanks for the feedback!

[...]

PR #12409, opened 2024-03-07
Disallow placing units when a territory effect forbids it.

asvitkine:
Previously, this was only considered for movement.
This makes both regular placement and bid placement forbid it as well.

Fixes another issue reported in: https://github.com/triplea-game/triplea/issues/12376

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/delegate/move/validation/UnitStackingLimitFilter.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/delegate/move/validation/UnitStackingLimitFilter.java
@@ -8,6 +8,7 @@ import games.strategy.engine.data.UnitType;
 import games.strategy.triplea.attachments.PlayerAttachment;
 import games.strategy.triplea.attachments.UnitAttachment;
 import games.strategy.triplea.delegate.Matches;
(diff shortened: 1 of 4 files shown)


Cernelius:
@asvitkine I tend to agree about the bid placement, but I'm not sure that this should be the case for regular placement. I understand that this effect means that the unit cannot enter the zone (which is sort of like saying that it cannot move in it), not that it cannot be produced in there. At most, I would expect that, if the unit is produced in there, it would remain stuck in there, but this would not be the case if the territory effects forbid entering but not exiting the zone.

Again, I'm not saying that I believe that this change is wrong: just not sure. However, it is definitely inconsistent with other similar things in TripleA, like the fact that movement stacking limits do not apply also as placement stacking limits, but such limits are kept separate (as far as I know). It would be safer if movement and placement limits are kept distinct, so that the game-maker itself can decide whether or not both apply.

This seems also a very similar concept to the "is impassable" territory option, which allows placement and it is thus functionally used in at least one game (namely "Jurassic" in "Ikl" for "Tithonian") (and a "A Song of Ice and Fire" has a similar concept too for "Others" in "TheFrostfangs", but there the placement is done with triggers and direct production from units).

PR #12420, opened 2024-03-09
Remove 'xms' and 'xmx' from vmoptions file

DanVanAtta:
Rather than explicitly setting max/min memory settings, we will instead let the JVM handle these settings. Without the explicit constraints, the JVM can more dynamically scale memory usage up and down.

### DIFF
--- a/game-app/game-headed/build.install4j
+++ b/game-app/game-headed/build.install4j
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<install4j version="9.0.6" transformSequenceNumber="9">
+<install4j version="9.0.5" transformSequenceNumber="9">
   <directoryPresets config=".triplea-root" />
   <application name="TripleA" applicationId="5251-3669-9623-1649" mediaDir="build/releases" mediaFilePattern="${compiler:sys.fullName}_${compiler:sys.version}_${compiler:sys.platform}" compression="9" lzmaCompression="true" pack200Compression="true" createChecksums="false" shortName="TripleA" publisher="TripleA Developer Team" publisherWeb="https://triplea-game.org" version="${compiler:sys.version}" allPathsRelative="true" autoSave="true" convertDotsToUnderscores="false" macVolumeId="af9346379363d40e" javaMinVersion="11" jdkMode="jdk">
     <jreBundles jdkProviderId="AdoptOpenJDK" release="11/jdk-11.0.19+7" />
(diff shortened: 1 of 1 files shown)


Cernelius:
Have you made sure that there is not something hardcoded in the program which set such things (and that the "vmoptions" modify)?

Meaning, for example, if I have a game which is so massive which simply cannot work with less than 4G assigned, will it work after this change (as now it works only if I increase the XMX manually)?

How about changing the default in the "vmoptions" to be blank, with blank meaning exactly the same as what this PR is doing, but leaving the option to set it like you can now?

DanVanAtta:
> Have you made sure that there is not something hardcoded in the program which set such things (and that the "vmoptions" modify)?

In `ProcessRunnerUtil.java` there is this block of code:

```
    Optional<String> maxMemory =
        ManagementFactory.getRuntimeMXBean().getInputArguments().stream()
            .filter(s -> s.toLowerCase().startsWith("-xmx"))
            .map(s -> s.substring(4))
            .findFirst();
    maxMemory.ifPresent(max -> commands.add("-Xmx" + max));
    Optional<String> maxStackSize =
        ManagementFactory.getRuntimeMXBean().getInputArguments().stream()
            .filter(s -> s.toLowerCase().startsWith("-xss"))
            .findFirst();
```

For the above, if 'xmx' or 'xms' are set, then that will be carried forward to any new processes that are launched (the fact that TripleA launches processes to begin with is somewhat pathological, but be-what-may..)

Removing the min/max limits would cascade forward and we would have new processes similarly launched without explicit lower/upper bounds. So.. this is fine.

-------------------------


> Meaning, for example, if I have a game which is so massive which simply cannot work with less than 4G assigned, will it work after this change (as now it works only if I increase the XMX manually)?


@Cernelius  Let's say you have a game so massive it will not work with less than 6G assigned. Before this update, the hard upper limit for memory was 4G. After this update, the only hard upper limit is however much the JVM can grab from the OS (which is potentially all of the available memory).

What's more, default management will return unused memory to the OS. If you have 8 game instances open, you need at least 32GB of RAM. With this update, if any of those instances are consistently using less memory, this memory will be returned to the OS which would enable you to run more games before reaching a limit based on availabel RAM.

> How about changing the default in the "vmoptions" to be blank, with blank meaning exactly the same as what this PR is doing, but leaving the option to set it like you can now?

Good suggestions, but I don't even want to encourage anyone to muck with these. Effectively we have min and max at 0% and 100%, which is what you want. Each instance consuming as much memory as it only needs. In reality the limits are not 0% and 100%, JVM memory management is complex & typically very smart.

asvitkine (commented):
Do you want to remove heapSizeMiB from earlier in the file too?

DanVanAtta:
> Do you want to remove heapSizeMiB from earlier in the file too?

Not without actually editing this file with install4j.

Having just done that - there is a list length variable that needs to be updated & was too short for the other option to have any effect previously..

DanVanAtta:
Specifically this:
` <element index="1">--add-opens java.desktop/com.apple.eawt.event=ALL-UNNAMED</element>`

Note that a few lines previous it says the length of the list is 3, and this is element #4

@RoiEXLab  could this have something to do with the MacOS click-to-launch feature not working well? 

@asvitkine  you run a mac?

RoiEXLab:
> @RoiEXLab could this have something to do with the MacOS click-to-launch feature not working well?

I doubt that it has any effect on this feature right now. TripleA is still on JDK 11 right? In this case illegal reflective access without opens just causes a warning in the console which can be suppressed with the `--add-opens` command line flag. I believe in later versions of the JDK the warning turns into an error, making the `--add-opens` mandatory. Regardless I'm pretty sure we are not using reflection for this at all. (Maybe we did in the past?) IIRC before JDK 9, we had this compile-time dependency for the `com.apple.eawt` classes which we could use to write code that adds open listeners and stuff. It worked because these classes existed in the JDK, they were just not publicly accessible, so the third-party dependency just mirrored these classes with public visibility to please the compiler, but wasn't compiled into the classpath so that at runtime the code actually used the private classes from the JDK instead. Fishy stuff tbh. but reflection can do the same without all of the static type safety.
At some point these classes were removed from the JDK (probably v9 or v11, not sure) in favour of an actual public API:
https://docs.oracle.com/en/java/javase/21/docs/api/java.desktop/java/awt/Desktop.html#setOpenFileHandler(java.awt.desktop.OpenFilesHandler)
So we eventually migrated to that, and I'm pretty sure the official API doesn't use any shady reflective access for this so you're probably right to just remove it.

PR #12421, opened 2024-03-09
Maps server integ test

DanVanAtta:
With this update:

- DOCKER WILL BE 100% REQUIRED TO RUN ANY TESTS. Before docker was required to run all tests, now it would be required in order to run any test. I believe this should not be a deal-breaker because docker was already a requirement for developers to have installed, following this update it will be a more important requirement. At the same time, I don't know how we can configure gradle to not have a docker compose dependency on the test task without making integ tests abysmally slow & flaky (with constant startup-&-stop of database).
 
- Paradigm for integ tests is changing. Instead of the Junit-test code (dropwizard-support) being responsible for launching servers, instead we will use docker-compose as a build step to launch both database and servers. 

History of integ testing in TripleA, and where we are landing: 
- A long time ago, TripleA would launch Derby (something of a predecessor to H2 IIRC) for database and the unit tests would generally just run. 
- At some point the lobby was converted to use Postgres instead of Derby. I don't recall whether we migrated to H2 at that time or if we went directly to running a postgres DB on docker. Regardless, the pattern was for developer to independently start a DB and then run all tests. 
- Somewhat recently, the need to start DB was automated. Gradle automatically tied into docker-compose and any test that was annotated with "@RequiresDatabase" would automatically invoke docker-compose to start the DB.
- We are now going a bit further, not only will the DB automatically start, but so will the server application (eg: lobby). To make this even cleaner, we are removing the "@RequiresDatabase" test configuration, which simplifies, all tests will start the database & any tests run out of an 'integ' folder will build & start up the needed application.

### Database & Tests

Database container is now launched first time whenever any test is run. All tests will therefore have access to a running database.

The database is not halted after tests are run. Thus, it is available for subsequent test executions, which are then fast to run. Further, database is not halted and restarted as different sub-projects are tested, the DB is just started exactly once. 

### Test configurations:  Two configs ->  One Config

- instead of there being two test configurations there will now only be one.  For example, previously when running tests from within intellij, it would prompt which test configuration it should use. After this update, it will no longer do that.


DanVanAtta:
I'm not sure if I like that the complexity of building just the game-app requires all of the server considerations.

DanVanAtta:
I'm pursing a different direction compared to this PR. I think a different variant of a mutli-repository is potentially a better solution.

Prevoiusly we attempted a mutli-repo config that looked like this:
image

In order to get the code dependency from client to client-lib, each developer would have had to configure an API key in their `.gradle-properties` file. I feel like that is a bit of a deal-breaker for that config. Though, multi-repo still has some important benefits, particularly for simplicity when building the client-side only.


What I'm now currently trying to pursue, is a different variant of the multi-repo config. if we keep the 'http-client' with the game client, then we only need the extra developer config when doing server work only. In this case, TripleA only needs to publish a JAR for the client-lib & the server would depend on that. IE: 
image

PR #12437, opened 2024-03-15
Fix a race on startup that could clear a loaded save.

asvitkine:
I noticed the race when running the game in debug mode from IntelliJ, where it was consistently being hit for me when I would click the button to open a save right after startup.
This PR fixes it by ensuring the save game will only be loaded after the initial map load.

Also, splits the function to load a map vs. a game and updates call sites.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/framework/startup/ui/panels/main/game/selector/GameSelectorModel.java
@@ -18,6 +18,7 @@ import java.nio.file.Path;
 import java.util.List;
 import java.util.Observable;
 import java.util.Optional;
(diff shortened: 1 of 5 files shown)


Cernelius:
I noticed this too but never reported it. If you have a heavy game which takes time to load (for example, "270BC Wars" or "War of the Relics") and you load a game-save of the same while the game is loading (from the XML), the savegame is apparently much faster to load and will be ready before the game has loaded, so, if you are not fast enough at starting the game, the game which is still loading will eventually load and substitute the game-save as if you never loaded it. This can happen in a fraction of seconds, so that you think that you have started the game-save, and, instead, you start a new game.

@asvitkine You mean that this change fixes (or is meant to fix) also what I described?

Why not just having the loading of the game aborted as soon as you start loading a game-save, instead of waiting for the XML-based game to load just then to nullify this process after completion?

asvitkine:
I think canceling the other load would be better to, but that's a bit more complicated to do.

Issue #12500, opened 2024-04-12
Unit Size Different In Bottom Bar and Windows with Zoom in 2.6


beelee1:
### Reproduction Steps

1.
2.
3.

### Engine Version

Engine Version:

### Is there anything else we should know?

This is didcussed in the forum here https://forums.triplea-game.org/topic/2999/2-6-release-getting-close-need-volunteers-to-help-beta-test-2-6/113

Edit
Added to Problem Tracker under Triage. Probably should be under "Fix Sooner"

Changed to Fix Sooner

Cernelius:
Was this an intended change (like someone thinking that the unit zoom is to be applied generally for some reasons) or not?

Anyway, bad change, in my opinion.

Cernelius:
The only argument in favour of the new behaviour (of which I can think) is for the territory tab showing all units as many non-zoomed units may exceed the vertical space of the territory tab itself especially for less-than-full HD screens. Still, I would have zoom 100% there (as it used to be) or maybe just enough zoom to make them fit vertically, although the fact that they now show in bottom bar too somewhat reduces the importance of such a tab and conversely promotes what would be its unicity in showing the units non-zoomed if such previous behaviour would be restored.

PR #12507, opened 2024-04-14
Image fix

DanVanAtta:
## Notes to Reviewer
Fixes: #12500

Issue is difference between: `getImage(imageKey)` vs `getTransformedImage(imageKey)`
`getImage` returns a scaled image, `getTransformedImage` returns an unscaled image.

Issue was introduced in:  1ab27d4031355305b4bf3ae9d9cf0b500aefa78b, there was a mistake to swap 'getTransformedImage' with 'getImage'. Without the clarification that one returns a scaled image and the other does not, it really does look they ought to return the same result. The multiple layers of caching also makes this confusing. It's very surprising for the 'image icon' cache to store unscaled images while the 'images' cache stores scaled images (one would expect both of them to store scaled images). 

The problem with the purchase screen is that the scaled images were being used, rather than the 100% scale images.

image

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/image/UnitImageFactory.java
@@ -227,7 +227,7 @@ public class UnitImageFactory {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


DanVanAtta:
cc: @asvitkine  bit of a subtle one

Issue #12536, opened 2024-04-22
2.6.147xx tripla.log does not truncate and can grow to 210M+


TheDog-GH:
See here
https://forums.triplea-game.org/topic/2999/2-6-release-getting-close-need-volunteers-to-help-beta-test-2-6/123
and posts to 128

TheDog-GH:
@asvitkine 
Just downloaded this and was hoping that my 14MB triplea.log would be truncated or cut into smaller 1MB units, but its still 14MB
:-/

My log file is here, just in case it can be elsewhere
C:\Users\RYZEN\triplea

![image](https://github.com/triplea-game/triplea/assets/71022517/11478d7a-ab32-4e75-9401-8e5d594cfab8)

PR #12538, opened 2024-04-24
Fix TripleA's rolling log policy.

asvitkine:
It wasn't specifying RollingFileAppender so the rolling and max size options weren't being used. I'm also changing the max size to 1MB and 10MB total.

Fixes: https://github.com/triplea-game/triplea/issues/12536


TheDog-GH:
@asvitkine 
I deleted my triplea.log and started from 0 it is now 1.5MB+ and it still has not truncated :-/
image

Issue #12488, opened 2024-04-08
ConcurrentBattleCalculator#calculate:200 - java.lang.IllegalStateException


tripleabuilderbot:
## Map
world_war_ii_global

## Log Message
java.lang.IllegalStateException: Round 10,000 reached in a battle. Something must be wrong. Please report this to TripleA.
 Territory: 62 Sea Zone Attacker: Japanese Attacking unit types: carrier,battleship,fighter,transport,destroyer,cruiser, Defending unit types: carrier,battleship,submarine,fighter,destroyer,cruiser

## TripleA Version
[2.6.14470](https://github.com/triplea-game/triplea/releases/tag/2.6.14470)

## Java Version
11.0.4

## Operating System
Windows 10

## Stack Trace
```
Exception: java.lang.IllegalStateException java.lang.IllegalStateException: Round 10,000 reached in a battle. Something must be wrong. Please report this to TripleA.
 Territory: 62 Sea Zone Attacker: Japanese Attacking unit types: carrier,battleship,fighter,transport,destroyer,cruiser, Defending unit types: carrier,battleship,submarine,fighter,destroyer,cruiser
java.lang.Exception
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(Unknown Source)
	at java.base/java.lang.reflect.Constructor.newInstance(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinTask.getThrowableException(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinTask.reportException(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinTask.invoke(Unknown Source)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateParallel(Unknown Source)
	at java.base/java.util.stream.AbstractPipeline.evaluate(Unknown Source)
	at java.base/java.util.stream.ReferencePipeline.collect(Unknown Source)
	at games.strategy.triplea.odds.calculator.ConcurrentBattleCalculator.calculate(ConcurrentBattleCalculator.java:200)
	at games.strategy.triplea.odds.calculator.BattleCalculatorPanel.lambda$updateStats$13(BattleCalculatorPanel.java:1173)
[...]

PR #12564, opened 2024-05-04
Fix 10k rounds reached combat error.

asvitkine:
This would happen when edit mode was left on when switching to an AI player. This changes makes such transitions turn off edit mode as well as making all battle simulations ignore edit mode.

FIxes: https://github.com/triplea-game/triplea/issues/12488

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/GameStep.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/GameStep.java
@@ -93,7 +93,7 @@ public class GameStep extends GameDataComponent {
     properties = stepProperties;
   }
 
-  public GamePlayer getPlayerId() {
+  public @Nullable GamePlayer getPlayerId() {
     return player;
   }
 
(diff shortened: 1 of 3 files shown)


DanVanAtta (approved):

DanVanAtta:
This update is a good step forward. Thank you @asvitkine  for tracking it down.

This fix makes me think about the case when someone turns on edit mode during combat with AI. There likely is another fix needed somewhere to resolve that type of issue (though, if that fix would be just too gnarly or not, I do not know)

asvitkine:
> This update is a good step forward. Thank you @asvitkine for tracking it down.
> 
> This fix makes me think about the case when someone turns on edit mode during combat with AI. There likely is another fix needed somewhere to resolve that type of issue (though, if that fix would be just too gnarly or not, I do not know)

I did check that and it's currently not possible to turn on edit mode outside of your own turn.

Cernelius:
Wouldn't it make sense that every AI player automatically turns Edit Mode off if it is on when taking over? Meaning that Edit Mode is set to off whenever the control passes to an AI player?

I actually don't even know what happens when the AI is playing with Edit Mode on, but I assume that the AI does not take advantage of it. I have no idea if there is any user-case for letting Edit Mode on while the AI plays.

asvitkine:
@Cernelius that's what my change does.

Cernelius:
Oh ok.

Cernelius:
Also for Does Nothing AI?

Issue #12360, opened 2024-02-23
ResourceCollection#removeResource:54 - java.lang.IllegalArgumentException


tripleabuilderbot:
## Map
world_war_ii_global

## Log Message
Can't remove more than player has of resource: PUs. current:17 toRemove: 27

## TripleA Version
[2.6.14647](https://github.com/triplea-game/triplea/releases/tag/2.6.14647)

## Java Version
11.0.19

## Operating System
Windows 11

## Stack Trace
```
Exception: java.lang.IllegalArgumentException Can't remove more than player has of resource: PUs. current:17 toRemove: 27
java.lang.Exception
	at games.strategy.engine.data.ResourceCollection.removeResource(ResourceCollection.java:54)
	at games.strategy.engine.data.ResourceCollection.subtract(ResourceCollection.java:110)
	at games.strategy.engine.data.ResourceCollection.subtract(ResourceCollection.java:105)
	at games.strategy.engine.data.ResourceCollection.difference(ResourceCollection.java:100)
	at games.strategy.triplea.ui.ProductionPanel.calculateLimits(ProductionPanel.java:254)
	at games.strategy.triplea.ui.ProductionPanel.show(ProductionPanel.java:104)
	at games.strategy.triplea.ui.TabbedProductionPanel.getProduction(TabbedProductionPanel.java:51)
	at games.strategy.triplea.ui.PurchasePanel$1.actionPerformed(PurchasePanel.java:70)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1967)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2308)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:405)
	at java.desktop/javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:262)
	at java.desktop/javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:279)
	at org.pushingpixels.substance.internal.utils.RolloverButtonListener.mouseReleased(RolloverButtonListener.java:108)
	at java.desktop/java.awt.Component.processMouseEvent(Component.java:6635)
	at java.desktop/javax.swing.JComponent.processMouseEvent(JComponent.java:3342)
	at java.desktop/java.awt.Component.processEvent(Component.java:6400)
[...]

PR #12578, opened 2024-05-11
Fix "Can't remove more than" error due to save game purchase code.

asvitkine:
This would happen for a joined player because end() wasn't being called on the delegate object.

Also some code cleanup.

Fixes: https://github.com/triplea-game/triplea/issues/12360


Cernelius:
I don't know if it is related, but in "270BC Wars" I have triggers removing a huge amount of PUs just because it is not possible to have triggers removing all PUs of a player (so I just set a huge amount to be sure to remove them all). That would be a nice feature to have (a trigger removing all PUs of the attached), but I never made a request for it because the hack appeared to work well enough (and because feature requests are very rarely taken into consideration).

asvitkine:
> I don't know if it is related, but in "270BC Wars" I have triggers removing a huge amount of PUs just because it is not possible to have triggers removing all PUs of a player (so I just set a huge amount to be sure to remove them all). That would be a nice feature to have (a trigger removing all PUs of the attached), but I never made a request for it because the hack appeared to work well enough (and because feature requests are very rarely taken into consideration).

That's unrelated.

Issue #10523, opened 2022-05-29
2.6: Trying to open a save game on a map that you don't have downloaded fails without error message


asvitkine:
### Reproduction Steps

1. Open a save game for a map that doesn't exist.
2. Start the game.

Expected: A dialog box saying the map doesn't exist, would you like to download it? (like in 2.5)
Actual: Game fails silently during load and reverts to game lobby screen.

### Attach a Save Game

I tried it with this 2.5 save on ww2 global expansion, which I didn't have: 
[2.5-example-save.tsvg.zip](https://github.com/triplea-game/triplea/files/8793607/2.5-example-save.tsvg.zip)

### Engine Version

Engine Version: 2.6.811

asvitkine:
The code for this used to be here:
https://github.com/triplea-game/triplea/blob/f516043a9d4ae146de4a6305165ce3a686a92a32/game-core/src/main/java/games/strategy/triplea/ResourceLoader.java#L93

Looks like that code moved around in the file a bit, but its removal was here:
https://github.com/triplea-game/triplea/pull/9775/files

PR #12586, opened 2024-05-16
Restore error message and prompt to download a map when not found.

asvitkine:
Fixes: https://github.com/triplea-game/triplea/issues/10523

Tested by trying to open a save game on a map I don't have.

Also some minor refactoring in related code to make the flow easier to understand.


DanVanAtta:
@asvitkine I recall there being an issue with the "would you like to download" prompt being double prompted. Perhaps worth double checking that (1) joining a bot with a game in progress does not have that issue, and (2) similarly, starting a new bot game without a map also does not lead to a double prompt

PR #12608, opened 2024-06-03
Lobby messaging - updates to notify that 2.6 is imminent

DanVanAtta:


Cernelius:
I'm not seeing it in the lobby (joining with 2.6.14748).

DanVanAtta:
@Cernelius it is expected behavior


image

Cernelius:
Ah, good. However, in this case, I would actually also write down something like "You are using an outdated version of TripleA" (since it's already June 05, so I guess you are going to rewrite it today).

PR #12622, opened 2024-06-06
Update triplea_maps.yaml - v1.4 the_shogun

TheDog-GH:

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2766,7 +2766,7 @@
 - mapName: The Shogun
   mapCategory: EXPERIMENTAL
   url: https://github.com/triplea-maps/the_shogun/archive/master.zip
-  version: 1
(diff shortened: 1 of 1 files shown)


beelee1:
@TheDog-GH I will add it but I thought 2.6 doesn't use the yaml anymore. I haven't tried using it for any map updates. Maybe it auto does it when you update the repo ? Idk :) will have to check it out.

This will work for people on older versions though :)

beelee1:
@TheDog-GH yea as @asvitkine pointed out, only whole numbers work for the version change. I didn't look at it, just saw it was good to go. The only changes are usually just the number. Maybe the description.

Be good to have a catch for that though. I will look at the version number in the future before merging. 

Last thing I want todo is create more work :)

TheDog-GH:
Thanks both!

Issue #12593, opened 2024-05-21
I think this may be a logic problem


WCSumpton:
https://github.com/triplea-game/triplea/blob/7aeff7e35a7c53a3feee535cbc5167c10b42dc60/game-app/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AvailableSupports.java#L167

The comment above getNextAvailableSupporter states that the same unit (not unitType) maybe returned multiple times.

If I'm not mistaken giveSupportToUnit is a unit-to-unit comparison and not unitType. A single unit may support another unit on a one-to-one basis. An Artillery may support An Infantry. Even if number is set to 2. Then that Artillery may support another Infantry only if that second Infantry is presence. If bonusType's count is set to 2, the lone Artillery may still only support one Infantry with the bonus amount and if there are two, both would receive the bonus amount. It takes the presence of a second Artillery before bonusType's count has an effect.

Thoughts...

Cheers...

asvitkine:
> An Artillery may support An Infantry. Even if number is set to 2. Then that Artillery may support another Infantry only if that second Infantry is presence. 

This part of your message says that an artillery can support two infantry. So I think that's what the above is doing - i.e. for each supporting unit, keep track how many units it can support.

Cernelius:
To clarify with a simple example, we can have these rules.

- The artillery can give +1 support to up to 3 infantries.
- The infantry attacks at 1 and can receive support from up to 3 artilleries (attacking at 2 if receiving support from 1, attacking at 3 if receiving support from 2 and attacking at 4 if receiving support from 3).

Under these rules, we can have the situation of 1 artillery and 2 infantries offensively attacking.

We can foresee three different reasonable outcomes, depending on the rules.

1. Both infantries attack at 2 (+1 each).
2. One infantry attacks at 3 (+2) and the other one at 2 (+1).
3. One infantry attacks at 4 (+3) and the other one at 1 (+0).

Notice that in the first case one of the three support abilities of the artillery is non-used whereas in the second and third case them all are used.

@WCSumpton Are you saying that (in your opinion) the program may be expected to behave as per point 3 according to the comment but actually behaves as per point 1, and this actual behaviour is correct? Meaning that the behaviour is correct and the comment is either wrong or unclear, so the comment should be improved to better define the behaviour (and the behaviour should be kept just as it is)?

---

Anyway, the best reference here would be @ron-murhammer if he's available for providing any insight. Did he write the comment which you are referencing? If not, who did?

[...]

PR #12632, opened 2024-06-10
Changing how stacked supports are handled. Considering the supporting unit, number of units supported and bonusType/count.

WCSumpton:
As discussed [here](https://github.com/triplea-game/triplea/issues/12593).

This changes the way AvailableSupports handles giving supports to units.

In AvailableSupports method getSupportsAvailable is renamed getSupporterAvailable and now returns the count of supporting units instead of the number of supports. 
Although method getSupportLeft has not been changed, it is now only used for testing.
Method giveSupportToUnit will now send the count to method getNextAvailableSupporter. There is a check for the number of supporters prior to this call because the number of supporters can be reduced in getNextAvailableSupporter while looping through supporters.
Method getNextAvailableSupporter is changed to allow a count to be passed (getAny changed to getAt) when selecting a supporting unit.

In CollectionUtils a new method, getAt, is designed to loop through a collection and pass back an element instead of selecting the first element.

In AvailableSupportsTest only method twoSupportersInAStackWithSupportNumberOfTwoGiveSupportToTwoUnits was changed to reflect that support and bonusType are divided between units, and not given to one unit by supporting units.

As an added bonus method getNextAvailableSupporter was changed to now allow number to be set to -1 (infinite), which could not be done because AvailableSupports was counting supports and a -1 value would cause the supporter to be removed after giving support the first time.

Cheers...

### DIFF
--- a/servers.yml
+++ b/servers.yml
@@ -11,10 +11,8 @@ servers:
       Welcome to the TripleA lobby!
       Please no politics, stay respectful, be welcoming, have fun.
 
-      2.6 Update: tech team is working on a variety of lobby compatibility
-      issues for 2.6. Please stick to the 2.5 game engine until we have
-      those fixed.
-
+      2.6 Update: The 2.6 release is back into a testing phase!
+      Follow 2.6 status on forums: https://tinyurl.com/triplea26
   - version: "2.0.20057"
     inactive: true
   - version: "0.0"


DanVanAtta:
@WCSumpton  - could you describe how to test this (and perhaps what testing you did)? Specifically, which maps & which supporter scenarios would really show a difference?

WCSumpton:
This was tested using the same test that was used to test that was used, AvailableSupportsTest. Using this process a multi-unit support is possible. Whereas the present process does not allow that kind of support.
Suppose you wanted to create a "WolfPack" support for German Submarines. Submarines attack at 2 defend at 1. Now, if 3 Submarines are together you want them to attack at 3 and defend at 2.
```
<!-- German WolfPack support -->
<!-- Requires 3 Submarines stacked together -->
<attachment name="supportAttachmentSubmarineBuff" attachTo="Submarine" javaClass="UnitSupportAttachment" type="unitType">
	<option name="unitType" value="Submarine"/>
	<option name="side" value="offence:defence"/>
	<option name="bonus" value="1"/>
	<option name="bonusType" value="SubmarineBuff" count="3"/>
	<option name="players" value="German"/>
</attachment>

<attachment name="supportAttachmentSubmarineDebuff" attachTo="Submarine" javaClass="UnitSupportAttachment" type="unitType">
	<option name="unitType" value="Submarine"/>
	<option name="side" value="offence:defence"/>
	<option name="bonus" value="-1"/>
	<option name="number" value="2"/>
	<option name="bonusType" value="SubmarineDebuff" count="2"/>
	<option name="players" value="German"/>
</attachment>
```
This type of support is undoable at present.
Or maybe you want Anti-Tank Guns to support Infantry when an enemy Armor is attacking? Can't do it. But with this change it's possible.

Cheers...

WCSumpton:
@DanVanAtta 

The above works, what I'm having problem with is getting the BattleRound counter to be read inside AvailableSupports so that a maxRounds process can be added. But the above still works as stated.

Cheers...

Cernelius:
> @WCSumpton - could you describe how to test this (and perhaps what testing you did)? Specifically, which maps & which supporter scenarios would really show a difference?

I don't think any existent games are going to change (but I cannot be sure). My understanding is that this is for example to change the answer to the question "if I have an infantry which can be supported by up to two artilleries and an artillery which can support up to two infantries, can the infantry be supported twice by a single artillery?" from "yes" to "no", which I think makes more logical sense and gives more options without substantially removing any (because you can still answer "yes" by exchanging the implied attachment with number equal to 2 with two otherwise identical attachments each with number equal to 1).

Cernelius:
I believe the existent behaviour was (maybe not completely intentionally) defined by @ron-murhammer (if he wants to weight in).

Cernelius:
I want to add that I'm not actually reading this: what I said is based on what I understood @WCSumpton is doing.

For example, I'm just assuming that, if an artillery has two support attachments at count 2 each, this single artillery can support the same infantry twice both before and after these changes, even though the title of this PR seems to state otherwise...

game-app/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AvailableSupports.java
```diff
@@ -164,13 +168,13 @@ int getSupportLeft(final UnitSupportAttachment support) {
    * <p>This may return the same unit multiple times in a row depending on how much support that
    * unit can give.
    */
-  private Unit getNextAvailableSupporter(final UnitSupportAttachment support) {
+  private Unit getNextAvailableSupporter(final UnitSupportAttachment support, final int i) {
     final SupportDetails details = supportUnits.get(support);
     final IntegerMap<Unit> intMap = details.supportUnits;
-    final Unit u = CollectionUtils.getAny(intMap.keySet());
+    final Unit u = CollectionUtils.getAt(intMap.keySet(), i);
```
DanVanAtta, line 174: I'm having trouble understanding the significance of this change.
Questions:
(1) Why do we know that the size of `intMap.keySet()` is less than `i`?
(2) Why is the `i`'th element significant? In other words, why that one element and not the first or last one? I think if I spent more time with this, it might "click" and I would understand, but perhaps you can readily give insight here?

lib/java-extras/src/main/java/org/triplea/java/collections/CollectionUtils.java
```diff
@@ -161,6 +162,15 @@ public static <T> T getAny(final Iterable<T> elements) {
     return elements.iterator().next();
   }
 
+  /** Like getAny, but skip elements if possible */
+  public static <T> T getAt(final Iterable<T> elements, final Integer count) {
```
DanVanAtta, line 166: I don't quite fully understand the context behind why `count` is significant.
With that caveat, the interface strikes me as odd. Typically the ordering of an iterable is not deterministic, yet we are getting the "n'th" element seemingly? 


Ignoring the 'no next element' issue, this example highlights the difference:

```
// deterministic
getAt(new LinkedHashMap().keySet(), 0)

vs

// not deterministic
getAt(new HashMap().keySet(), 0)
```

At a surface level, this can be somewhat fixed by moving this method to the class that uses it,and making it private and more context specific.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AvailableSupports.java
WCSumpton: Element stored in intMap are, to my understanding, linear. getAny will always return the first element until that element was removed. getAt returns the element at a given position, thus element 1 may still have a count but has already been used, so I want to skip to the next available element with a count.

WCSumpton: > (1) Why do we know that the size of `intMap.keySet()` is less than `i`?

The size of `intMap.keySet()` should be either greater than or equal to `i` and is determined prior to calling getNextAvailableSupporter by `numSupportAvailableToApply = getSupportersAvailable(rule)`. And because getNextAvailableSupporter can decrease the size, it is again rechecked with `int count = Math.min(i, getSupportersAvailable(rule))`. 



> (2) Why is the `i`'th element significant? In other words, why that one element and not the first or last one? I think if I spent more time with this, it might "click" and I would understand, but perhaps you can readily give insight here?

This is the **problem**, IMHO, that we are having right now. `CollectionUtils.getAny(intMap.keySet())` only returns the first element for `u`, and does not consider if that element for might have already been use prior in the loop.

As of now giveSupportToUnit uses the sun total of **number** and is restricted by **bonusType/count**, and does not care how many units are available. Such that a unit can give as much support as limited by **bonusType/count** to a given unit on a 1 to x basis as determined by **bonusType/count**.
This changes giveSupportToUnit to use the number of units available to give support then restricts that by **bonusType/count** and then tries to choose different supporting units. In assents, a unit can give support to another unit on a 1 to 1 basis.

Hope this makes things clearer.

Cheers...

lib/java-extras/src/main/java/org/triplea/java/collections/CollectionUtils.java
WCSumpton: > I don't quite fully understand the context behind why `count` is significant.

count represents the number of times `for (int i = 1; i <= numSupportAvailableToApply; i++)` has proceeded through the loop adjusted for the size of `intMap.keySet()`, which can be reduced in getNextAvailableSupporter.

The collection intMap is not being recreated, to my understanding. The same collection is being passed around. 

> At a surface level, this can be somewhat fixed by moving this method to the class that uses it,and making it private and more context specific.

Ok.

Cheers...

DanVanAtta: > count represents the number of times for (int i = 1; i <= numSupportAvailableToApply; i++) has proceeded through the loop adjusted for the size of intMap.keySet(), which can be reduced in getNextAvailableSupporter.

What does it mean for the value to be adjusted by the size of that intmap? It's not clear to me yet what 'count' represents, what is being counted. Is that the count of supporters that have been 'used' up so far?

DanVanAtta:
@WCSumpton my apologies for the drawn out review. (A) this existing code was really hard to understand & this makes it difficult to review. (B) I'm really stretched thin between life and the things needed for 2.6 to be live. That latter is really important since no updates will be seen by anyone until that happens. Once 2.6 is out for real, the release model changes and suddenly updates are going to be live right away.

I don't know when I can finish reviewing... @asvitkine  perhaps you might be able to take over the review?

@WCSumpton  I think one question I would have - how confident are you that the updated test cases comprehensively validate the existing and updated logic? In other words - is the test suite very comprehensive? Do you have very high confidence this works based on manual testing, based on high test coverage from automated tests, or both? 

I think if we are very confident in the automated tests and have extensively manually tested to cover all the variety of map scenarios, then perhaps we can skip further review.  While I hear you when you say "this is correct" - we all always have to try and make sure, and try to make sure that is the case ideally in more than just one way too.

asvitkine:
> As discussed [here](https://github.com/triplea-game/triplea/issues/12593).
> 
> This changes the way AvailableSupports handles giving supports to units.
> 
> In AvailableSupports method getSupportsAvailable is renamed getSupporterAvailable and now returns the count of supporting units instead of the number of supports. Although method getSupportLeft has not been changed, it is now only used for testing. Method giveSupportToUnit will now send the count to method getNextAvailableSupporter. There is a check for the number of supporters prior to this call because the number of supporters can be reduced in getNextAvailableSupporter while looping through supporters. Method getNextAvailableSupporter is changed to allow a count to be passed (getAny changed to getAt) when selecting a supporting unit.
> 
> In CollectionUtils a new method, getAt, is designed to loop through a collection and pass back an element instead of selecting the first element.
> 
> In AvailableSupportsTest only method twoSupportersInAStackWithSupportNumberOfTwoGiveSupportToTwoUnits was changed to reflect that support and bonusType are divided between units, and not given to one unit by supporting units.
> 
> As an added bonus method getNextAvailableSupporter was changed to now allow number to be set to -1 (infinite), which could not be done because AvailableSupports was counting supports and a -1 value would cause the supporter to be removed after giving support the first time.
> 
> Cheers...

I'm going to review the code changes in detail, but first I'd like to ask you to update the PR description to clearly describe the behavior change (i.e. as would be visible to a player or mapmaker) from the description about which functions/code was changed. I can figure out the latter myself by reviewing the code, but the former should be clearly documented without mentioning specific code - i.e. document the spec that this is implementing and how it differs before vs. after. Provide examples if possible!

lib/java-extras/src/main/java/org/triplea/java/collections/CollectionUtils.java
```diff
@@ -161,6 +162,15 @@ public static <T> T getAny(final Iterable<T> elements) {
...
+    Iterator<T> iterator = elements.iterator();
```
asvitkine, line 167: I don't think there's any guarantee that iterator() will return elements in the same order each time. It can, if the collection is a List, for example, but not if it's a Set (which in this case it is).

So for example, you can have a scenario where the container has elements A, B, C and the first time you call iterator(), it will give you A, B, C order, but the next time, order will be B, A, C. So any calling logic that expects getAt() to be consistent won't work correctly.

If you really want these semantics, you should just convert the collection to a List first outside and just use standard .get().

game-app/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AvailableSupports.java
```diff
@@ -149,8 +150,11 @@ int giveSupportToUnit(final Unit unit) {
     return amountOfSupportGiven;
   }
 
-  private int getSupportAvailable(final UnitSupportAttachment support) {
-    return Math.max(0, Math.min(support.getBonusType().getCount(), getSupportLeft(support)));
+  /** change to retrun the number of supporters (units) not supports */
```
asvitkine, line 153: Comments should describe what the code is doing, not what it got changed to do from some other version of the code.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AvailableSupports.java
```diff
@@ -131,9 +131,10 @@ int giveSupportToUnit(final Unit unit) {
           continue;
         }
 
-        final int numSupportAvailableToApply = getSupportAvailable(rule);
-        for (int i = 0; i < numSupportAvailableToApply; i++) {
-          final Unit supporter = getNextAvailableSupporter(rule);
+        final int numSupportAvailableToApply = getSupportersAvailable(rule);
+        for (int i = 1; i <= numSupportAvailableToApply; i++) {
+          int count = Math.min(i, getSupportersAvailable(rule));
```
asvitkine, line 136: I don't understand this logic. Wouldn't `count` always be `i` in this case?

game-app/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AvailableSupports.java
asvitkine: Also, I don't know what the difference is between "supporters" and "supports" - both sound like they're talking about units. Is "supports" meant to mean something other than units? If so, we should explain.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AvailableSupports.java
```diff
@@ -149,8 +150,11 @@ int giveSupportToUnit(final Unit unit) {
     return amountOfSupportGiven;
   }
 
-  private int getSupportAvailable(final UnitSupportAttachment support) {
-    return Math.max(0, Math.min(support.getBonusType().getCount(), getSupportLeft(support)));
+  /** change to retrun the number of supporters (units) not supports */
+  private int getSupportersAvailable(final UnitSupportAttachment support) {
+    final SupportDetails details = supportUnits.get(support);
+    final IntegerMap<Unit> intMap = details.supportUnits;
+    return Math.max(0, Math.min(support.getBonusType().getCount(), intMap.size()));
```
asvitkine, line 157: Previously it was using `getSupportLeft()`, which was using `details.totalSupport`, but now it's using `details.intMap.size()`. It's not clear why the latter is more correct. A comment in the code would be useful.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AvailableSupports.java
```diff
@@ -149,8 +150,11 @@ int giveSupportToUnit(final Unit unit) {
     return amountOfSupportGiven;
   }
 
-  private int getSupportAvailable(final UnitSupportAttachment support) {
-    return Math.max(0, Math.min(support.getBonusType().getCount(), getSupportLeft(support)));
+  /** change to retrun the number of supporters (units) not supports */
+  private int getSupportersAvailable(final UnitSupportAttachment support) {
+    final SupportDetails details = supportUnits.get(support);
+    final IntegerMap<Unit> intMap = details.supportUnits;
```
asvitkine, line 156: Previous code was checking whether `details` was not null. Is it intentional that you're not checking for that anymore - i.e. you're convinced it can't be null? Or an oversight?

asvitkine (changes requested):
I left a number of comments, so marking as "changes requested". Happy to take another look when you've had a chance to address them.

game-app/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AvailableSupports.java
WCSumpton: > I don't understand this logic. Wouldn't `count` always be `i` in this case?

No, The number of supporters can be reduced by getNextAvailableSupporter. Because of this there has to check if how many supporters available with getSupportersAvailable.

Cheers...

game-app/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AvailableSupports.java
Cernelius: > > Also, I don't know what the difference is between "supporters" and "supports" - both sound like they're talking about units. Is "supports" meant to mean something other than units? If so, we should explain.
> 
> supporters are units, supports are the number of times this unit can give support.

Are they? I understood that "supporters" are actually supports (so not necessarily units unless in the game there are not support attachments sharing both the same type and the attached unit). That is, if I have two supports for the same unit (that is two support attachments attached to the same unit), they will be able to stack on the same target just as if they come from two different units, just, if they share the same type, they (or at least one of them if you are supporting different counts for the same type across different attachments) need a count greater than 1 for it to do so. I understood that only the case of a single support would never be applied more than once to the same target (regardless of its number, that defines the number of units which that support can support, and currently allows to apply it multiple times to a single unit if the count for the type is greater than 1). Am I not right?

Mind you that I'm not a developer and I've read none of the coding at this PR.

asvitkine: > supports are the number of times this unit can give support

Ah! I think it's a very subtle definition, so I would actually suggest to spell it out (at least in comments).
i.e. when you want to talk about that concept, in comments, actually write "the number of times the unit can give support".

game-app/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AvailableSupports.java
asvitkine: I see, that's very subtle. From reading the code, it's entirely not obvious that `getSupportersAvailable()` would return different values as we iterate through the loop. At the very least, that should be commented about right above this line.

But perhaps we could also find a way to make it clearer from the code itself - e.g. by renaming that function to give a hint that it's changing (e.g. `howManySupportersAreLeft(rule)` or something).

game-app/game-core/src/test/java/games/strategy/triplea/delegate/power/calculator/AvailableSupportsTest.java
WCSumpton, line 1: @DanVanAtta, @asvitkine, @Cernelius 

This change shows the desired results. Presently, "supportUnit" gives 2 support to "unit", while "supportUnit2" gives 2 support to "unit2". If "supportUnit2" is removed, "supportUnit" still gives 2 support to "unit" while "unit2" receives no support. This change changes "supportUnit" to give 1 support to each "unit" and "unit2", while "supportUnit2" gives the stacking support to both "unit" and "unit2". Removing "supportUnit2" will only remove the stacking support as "supportUnit" will still give the initial 1 support to both units.

Maybe these results could be handled better with a list instead of a collection, I'm not sure. I have created test maps using "neuschwabenland" and also "invasion_usa" and have received no errors.

Cheers...

game-app/game-core/src/main/java/games/strategy/triplea/delegate/power/calculator/AvailableSupports.java
asvitkine: Just revisiting this, but another option that doesn't have the issue with getAt() discussed in the other thread (i.e. order not guaranteed) is to actually keep the state in an Iterator object. That way, you can just advance through it via next() etc.

beelee1:
@WCSumpton so did this get merged ? Part of latest pre ?

Issue #12693, opened 2024-07-05
2.6.14752: ProSortMoveOptionsUtils#sortUnitNeededOptionsThenAttack:118 - java.lang.IllegalArgumentException


tripleabuilderbot:
## Map
a_song_of_ice_and_fire / A Song of Ice and Fire

## Log Message
Comparison method violates its general contract!

## TripleA Version
[2.6.14752](https://github.com/triplea-game/triplea/releases/tag/2.6.14752)

## Java Version
11.0.19

## Operating System
Windows 11

## Heap Size
8138M

## Stack Trace
```
Exception: java.lang.IllegalArgumentException Comparison method violates its general contract!
java.lang.Exception
	at java.base/java.util.TimSort.mergeHi(TimSort.java:903)
	at java.base/java.util.TimSort.mergeAt(TimSort.java:520)
	at java.base/java.util.TimSort.mergeCollapse(TimSort.java:448)
	at java.base/java.util.TimSort.sort(TimSort.java:245)
	at java.base/java.util.Arrays.sort(Arrays.java:1515)
	at java.base/java.util.ArrayList.sort(ArrayList.java:1750)
	at games.strategy.triplea.ai.pro.util.ProSortMoveOptionsUtils.sortUnitNeededOptionsThenAttack(ProSortMoveOptionsUtils.java:118)
	at games.strategy.triplea.ai.pro.ProCombatMoveAi.tryToAttackTerritories(ProCombatMoveAi.java:1344)
	at games.strategy.triplea.ai.pro.ProCombatMoveAi.determineTerritoriesToAttack(ProCombatMoveAi.java:336)
	at games.strategy.triplea.ai.pro.ProCombatMoveAi.doCombatMove(ProCombatMoveAi.java:97)
	at games.strategy.triplea.ai.pro.AbstractProAi.move(AbstractProAi.java:134)
	at games.strategy.triplea.ai.AbstractAi.start(AbstractAi.java:516)
	at games.strategy.engine.framework.ServerGame.waitForPlayerToFinishStep(ServerGame.java:603)
	at games.strategy.engine.framework.ServerGame.runStep(ServerGame.java:452)
	at games.strategy.engine.framework.ServerGame.runNextStep(ServerGame.java:347)
	at games.strategy.engine.framework.ServerGame.startGame(ServerGame.java:314)
	at games.strategy.engine.framework.startup.launcher.LocalLauncher.launchInternal(LocalLauncher.java:92)
[...]

PR #12697, opened 2024-07-05
Fix division by zero and check for isFinite() to fix comparator issue.

asvitkine:
The check was intended to verify whether https://github.com/triplea-game/triplea/issues/12693 was caused by non-finite values.
However, tests caught that this could actually trigger due to a division by zero.
I suspect this should therefore fix https://github.com/triplea-game/triplea/issues/12693


asvitkine:
Aha, looks like this precondition actually failed in testAiGameMovingConsumedUnitsToFactories! So that does show it's probably this that is responsible. Will need to debug.

Issue #12447, opened 2024-03-19
2.6 Combat Move units with a double click - units are truncated & ...


TheDog-GH:
Say using 1941 Global Command Decision.
On Combat move
Double click on Paris you get this, note 5 units are displayed
![image](https://github.com/triplea-game/triplea/assets/71022517/618c54b9-4503-4722-a6aa-f863ab58d0f6)

.
Using the Territory tab on the right panel you get this, , note 5 units are displayed +2 non-moving with a scroll bar
![image](https://github.com/triplea-game/triplea/assets/71022517/741a7a6e-11ac-4ab2-9582-4f7f58362467)
![image](https://github.com/triplea-game/triplea/assets/71022517/b2276efd-f81b-4e24-b66b-a88e33258df1)

.
**Observations - The Double click combat move;** 
* units are truncated vertical and horizontally and the order differs from the other image.  
* It does not have a scroll bar
* Maybe the display code can be reused/shared

.
Here is 113 B Sea Zone (West of Denmark) for comparison, the ships can be about 100px wide
![image](https://github.com/triplea-game/triplea/assets/71022517/a5fa7377-8106-412f-9f97-f24bfff691b1)

frigoref:
Could be related to the fact that the image icons are not squared
![image](https://github.com/triplea-game/triplea/assets/10353640/7dcd9cd7-4330-497d-be1d-e9cbe1322d9c)
in comparison to for example **World War II Global 1940**
![image](https://github.com/triplea-game/triplea/assets/10353640/a11bc770-50ae-4e0d-baac-9387d96c0b4c)

Could someone confirm that it would be a feature request to allow non-squared icons?

Files I found that are related to this:
- [UnitChooser.java](https://github.com/triplea-game/triplea/blob/master/game-app/game-core/src/main/java/games/strategy/triplea/ui/UnitChooser.java) method loadNonWithdrawableImage (UnitChooser added via call from [MovePanel.java](https://github.com/triplea-game/triplea/blob/master/game-app/game-headed/src/main/java/games/strategy/triplea/ui/panel/move/MovePanel.java))
The code works works against height and I think thereby assuming squared icons (same pixels in height and width).
```
private void loadNonWithdrawableImage() {
      final int unitImageHeight = unitImageFactory.getUnitImageHeight();
      final double nonWithdrawableImageHeight = getNonWithdrawableImageHeight(unitImageHeight);
[...]

PR #12722, opened 2024-07-13
Issues 12447 (Standardized Sorting of UnitCategory)

frigoref:
[Issue 12447](https://github.com/triplea-game/triplea/issues/12447) was mentioning the sorting of units being different in different UI areas.
This PR tries to unify the sorting (based on UnitCategory) used before adding UI components to ensure a standardized sorting.
Test was done with UnitChooser, TerritoryPanel, Battle Calculator and the UnitScoller for map [1941 Global Command Decision](https://github.com/triplea-maps/1941_global_command_decision/tree/master):
image

One thing I was not so sure about is the way gameData is selected in `UnitChooser.java` by picking the first entry of a list.
Maybe there is a better way to do that which can be suggested by the reviewer.

The commit comment describes the changes:
```
AbstractUndoableMovesPanel.java
-unitCategory sorting with gameData

AvatarPanelFactory.java
-passing territory
-replace sorting by unitRenderingOrder with new generic sorting with territory and player 

BattleCalculatorPanel.java
-rename method to getUnitCategories

BattleDisplay.java (2x)
BattleModel.java
MapPanel.java
OrderOfLossesInputPanel.java
PlayerUnitsPanel.java
SimpleUnitPanel.java
UnitChooser.java
-unitCategory sorting with gameData

UnitScroller.java
-passing territory

UnitSeparator.java
-core Comparator for sorting by UnitCategory: getComparatorUnitCategories
-several overloaded methods getComparatorUnitCategories
```

PS: I will try next time to chop it into smaller pieces. It just grew somehow.


game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/OrderOfLossesInputPanel.java
```diff
@@ -1,5 +1,7 @@
 package games.strategy.triplea.odds.calculator;
 
+import static games.strategy.triplea.util.UnitSeparator.getComparatorUnitCategories;
```
DanVanAtta, line 3: style nitpick, avoid static method imports unless in test code and importing things like assertions.  The reasoning is that when inspecting the code, the method call looks like a call to a local method or to a super class.

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/PlayerUnitsPanel.java
```diff
@@ -122,7 +124,8 @@ public void init(final GamePlayer gamePlayer, final List<Unit> units, final bool
 
     GamePlayer previousPlayer = null;
     JPanel panel = null;
-    for (final UnitCategory category : categories) {
+    unitCategories.sort(getComparatorUnitCategories(data));
```
DanVanAtta, line 127: There is a sort of `unitCategories` higher up on line 69. Does this new sort call here, on line 127, overwrite the results of that previous sort?

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/OrderOfLossesInputPanel.java
DanVanAtta: It might be even cleaner to have a static method in  `UnitSeparato` to do the sorting, rather than returning the comparator to then let the client code to the sorting. EG: `UnitSeparator.sort(  <unit-type-list> )`

game-app/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
```diff
@@ -775,7 +782,14 @@ void setNotificationShort(
 
     private void categorizeUnits(
         final Iterable<UnitCategory> categoryIter, final boolean damaged, final boolean disabled) {
-      for (final UnitCategory category : categoryIter) {
+      final List<UnitCategory> unitCategories =
+          StreamSupport.stream(categoryIter.spliterator(), false).collect(Collectors.toList());
+      if (unitCategories.isEmpty()) {
+        return;
+      }
+      final GameData gameData = unitCategories.get(0).getUnitAttachment().getData();
```
DanVanAtta, line 790: `BattleDisplay:L95` has a reference to `GameData` - can that be used?

game-app/game-core/src/main/java/games/strategy/triplea/ui/SimpleUnitPanel.java
```diff
@@ -120,7 +125,13 @@ public void setUnitsFromRepairRuleMap(
    */
   public void setUnitsFromCategories(final Collection<UnitCategory> categories) {
     removeAll();
-    for (final UnitCategory category : categories) {
+    List<UnitCategory> unitCategories = new ArrayList(categories);
+    if (unitCategories.isEmpty()) {
```
DanVanAtta, line 129: minor, seems like we can check if `categories` is empty first & return sooner (ie: this line can be moved up by one, and check `categories` instead of `unitCategories`

game-app/game-core/src/main/java/games/strategy/triplea/ui/SimpleUnitPanel.java
```diff
@@ -120,7 +125,13 @@ public void setUnitsFromRepairRuleMap(
...
+      return;
+    }
+    final GameData gameData = unitCategories.get(0).getUnitAttachment().getData();
```
DanVanAtta, line 132: `UiContext` might have a reference to `GameData` that can be used.
I'm not sure if the game data referenced by unit attachment is always non-null (it might be, in which case this is fine)

game-app/game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
```diff
@@ -742,7 +743,9 @@ public void paint(final Graphics g) {
     }
     final var unitImageFactory = uiContext.getUnitImageFactory();
     for (final Collection<Unit> value : highlightedUnits) {
-      for (final UnitCategory category : UnitSeparator.categorize(value)) {
+      List<UnitCategory> unitCategories = new ArrayList(UnitSeparator.categorize(value));
+      unitCategories.sort(getComparatorUnitCategories(gameData));
```
DanVanAtta, line 747: There is a little bit of a pattern here. Perhaps this can be cleaned up to be something like:
```List<unitCategory> unitCategories = UnitSeparator.sortByUnitType(value)```

Which in turn would do the list copy, categorize, and sorting.
FWIW, there is a bit of a feature envy code smell here where UnitSeparator is used to generate a return value, more data is obtained from UnitSeparator to then operate on that same data again. Resolving 'feature envy code smell' is usually done by moving the functionality entirely into `UnitSeparator`

File: game-app/game-core/src/main/java/games/strategy/triplea/ui/unit/scroller/AvatarPanelFactory.java [hunk 7]
```diff
@@ -159,24 +165,6 @@ private static Image createUnitStackImage(
     return combinedImage;
   }
 
-  private static Comparator<UnitCategory> unitRenderingOrder(final GamePlayer currentPlayer) {
```
DanVanAtta, line 162: Nice, it's good to see custom sort logic be consolidated & removed like this.

game-app/game-core/src/main/java/games/strategy/triplea/util/UnitSeparator.java
```diff
@@ -48,36 +48,80 @@ public static class SeparatorCategories {
...
+  }
+
+  /** Returns <code>Comparator</code> for unit categories with current <code>GameData</code> */
+  public static Comparator<UnitCategory> getComparatorUnitCategories(final GameData gameData) {
+    return getComparatorUnitCategories(null, gameData, gameData.getHistory().getCurrentPlayer());
+  }
+
+  /** Returns <code>Comparator</code> for unit categories of a <code>Territory</code> */
+  private static Comparator<UnitCategory> getComparatorUnitCategories(
+      final Territory t, final GameData gameData, final GamePlayer currentPlayer) {
```
DanVanAtta, line 82: I notice methods calling this one get `GameData` from `Territory`, any reason to not do that here and remove the `GameData` parameter?

game-app/game-core/src/main/java/games/strategy/triplea/util/UnitSeparator.java
```diff
@@ -48,36 +48,80 @@ public static class SeparatorCategories {
...
+                .thenComparing(ut -> !Matches.unitTypeIsLand().test(ut)))
+        .thenComparingInt(ut -> ut.getUnitAttachment().getMaxBuiltPerPlayer())
+        .thenComparing(
+            uc ->
+                uc.getUnitAttachment()
+                    .getAttack(
+                        (currentPlayer == null
+                            ? uc.getOwner()
+                            : currentPlayer))) // should be currentPlayer
+        .thenComparing(UnitCategory::getType, Comparator.comparing(xmlUnitTypes::indexOf));
```
DanVanAtta, line 122: The comparator logic is a bit dense. Some inline comments might help clean it up a bit. Fully spelled out variables might help as well (potentially usually logical names, like "attackingUnits". Not sure if those suggestions are going to move the needle. Overall this is a block to double check and re-read in case you see a good way to help clean it up a bit more.

game-app/game-core/src/main/java/games/strategy/triplea/util/UnitSeparator.java
frigoref: I am going to create a fix to change `Territory `to `Optional<Territory>`.

game-app/game-core/src/main/java/games/strategy/triplea/ui/panels/map/MapPanel.java
frigoref: Agree, I create a fix to introduce another method `UnitSeparator.getSortedUnitCategories`.

game-app/game-core/src/main/java/games/strategy/triplea/ui/SimpleUnitPanel.java
frigoref: Agree, I create a fix.

game-app/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
frigoref: It cannot as the changed code is in private static class `CasualtyNotificationPanel` for which I could not find `GameData`.

game-app/game-core/src/main/java/games/strategy/triplea/util/UnitSeparator.java
frigoref: Wasn't too happy with it either, but then I thought as long as the method JavaDoc is explaining the logic I could understand the code as well.
I could offer to create separate Comparator instantiations that are then used in the return, but for me it wouldn't be really much better readable.

game-app/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
DanVanAtta: NP - please double check the nullability of `unitAttachment#data`. I think grabbing gameData like this should be 'safe', but let's be sure.

game-app/game-core/src/main/java/games/strategy/triplea/util/UnitSeparator.java
DanVanAtta: Cool, though - at a style level consider having multiple method signatures instead. `Optional` is a "bad" type for a method parameter.  

Though, maybe we can avoid the dependency on territory? Is the ordering change potentially that significant considering territory? Is that important? Second, it's an interesting question if it's more valuable to have a fully consistent ordering regardless of even territory. Thus, perhaps it's worthwhile to see if we can simplify the comparator to not depend on territory (while at it, maybe player too)

game-app/game-core/src/main/java/games/strategy/triplea/util/UnitSeparator.java
DanVanAtta: Mentioned in another comment - if we can remove territory and player as inputs - this could be a bit simpler perhaps (?)
I think the key would be to tie the blocks together to the javadoc, or to simplify the javadoc and put more details inline. Naming the predicates can be of help too, eg:

```
Predicate<UnitCategory> hasMovement = uc -> ....
```

game-app/game-core/src/main/java/games/strategy/triplea/util/UnitSeparator.java
frigoref: I think we should enable the sorting to allow for some criteria to be optional that should still be considered and territory properties like owner or isWater seem plausble criteria as well as gamePlayer with whether its the same as the unit owner.
For me this comes as a cost of having a general sorting logic that it should incorporate several situations where some criteria might not be present.

game-app/game-core/src/main/java/games/strategy/triplea/util/UnitSeparator.java
frigoref: I am not sure how to rework this. Shall we have a call instead of post-ping-pong here? Send me a pm with some time slots you can offer.

game-app/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
frigoref: What should happen if `unitAttachment = null` ? For the sorting a different gameData source would be needed, e.g. from `unitAttachment#unitType`? If all potential sources could be null, we cannot escape this what-if-question.

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/PlayerUnitsPanel.java
frigoref: Good point. Another place where we can remove a "special" sorting with our generic one. Fix will be provided in commit #6.

game-app/game-core/src/main/java/games/strategy/triplea/util/UnitSeparator.java
DanVanAtta: > properties like owner or isWater seem plausble criteria as well as gamePlayer with whether its the same as the unit owner.

Are these plausible criteria or required? For either case, if you would, can you walk me through some scenarios you considering where player or territory would change the Unit-Category ordering?

Also to consider, we can have a method give a generic ordering of unit category, and then the places that do care about player or territory, they can do their one-off logic locally. That is to say, there is flexibility; just because the top-most level method is not super-flexible, doesn't mean we can't work that flexibility in where needed.

DanVanAtta: Second thought: just please be sure to remove any method signatures that are not used if any.

Minimizing inputs is a way to have a more consistent sort ordering. We probably don't need to draw out the convo any further than there.

game-app/game-core/src/main/java/games/strategy/triplea/util/UnitSeparator.java
DanVanAtta: I like the idea of shifting to a call. Though, I'm not sure of a good way to really fix this either! 

FWIW, I intended this to be the operative part of the review comment:
> Overall this is a block to double check and re-read in case you see a good way to help clean it up a bit more.

We can and should move on now that we've done that double check

game-app/game-core/src/main/java/games/strategy/triplea/ui/BattleDisplay.java
DanVanAtta: I looked into this a bit further. You're good here. I'll post some more followup in a bit.

DanVanAtta: Correction.. tracing this down... Neutral units in 2.5 have null `gameData`

If you save a game that has neutral players with 2.5, then load that up with 2.6, then find a territory with only neutral units - in that case `gameData` would seemingly be null. 

Happily that nullity concern will go away kinda soon as how the neutral player is saved to disk was changed. Once we no longer need to load saves from 2.5, that `gameData` would never be null.

PR #12758, opened 2024-07-24
Automatic Bot Moderators

DanVanAtta:
Updates to give the host of a network game 'mod' powers. If the game is a bot, then also give the 'oldest' player 'mod' powers.

Mod powers allow: 
- disconnect players
- ban players (effective until restart of bot)

The bot case is interesting since the 'oldest' player can change. For example, two players enter a bot (to the staging screen), first is mod  & then leaves. Now the second player needs to become moderator.

To achieve this we add a "moderator promoted" message which is sent by the server to inform players of the new moderator.

### WIP

TODO:
- cleanup/polish 
- remove the extra stuff (perhaps a few too many listeners thrown in)
- bit more testing
- limit ability to see 'disconnect/ban' to the moderator

### DIFF
--- a/game-app/domain-data/src/main/java/org/triplea/domain/data/ChatParticipant.java
+++ b/game-app/domain-data/src/main/java/org/triplea/domain/data/ChatParticipant.java
@@ -26,7 +26,7 @@ public class ChatParticipant implements Serializable {
   @Nonnull private final String playerChatId;
 
   /** True if the player has moderator privileges. */
-  private final boolean isModerator;
+  @Setter private boolean isModerator;
 
   /** Status is custom text set by players, eg: "AFK" or "Looking for a game". */
   @Setter @Nullable private String status;
--- a/game-app/game-core/src/main/java/games/strategy/engine/chat/Chat.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/chat/Chat.java
@@ -2,7 +2,8 @@ package games.strategy.engine.chat;
 
 import com.google.common.collect.EvictingQueue;
 import com.google.common.collect.Sets;
-import games.strategy.engine.framework.startup.mc.ModeratorMessage;
+import games.strategy.engine.framework.startup.mc.messages.ModeratorMessage;
+import games.strategy.net.IMessageListener;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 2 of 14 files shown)


Cernelius:
Will moderators be able to be super-moderators with respect to the user acquiring moderating powers for joining the bot earlier than everyone else in it?

How will this be affected by our rules, specifically
>First come first serve; up to a certain degree. If after some time no opponent has shown up, bot has to be given free to others.

about dealing with "oldest" users who just sit in the bot for a long time without starting any games and disallowing other users to use the bot, at which point a moderator may decide that the bot has to be given up?

DanVanAtta:
@Cernelius  you make me wonder for a moment if we should use a different word than "mod". I also really don't want to introduce an entirely new concept either.

> Will moderators be able to be super-moderators with respect to the user acquiring moderating powers for joining the bot earlier than everyone else in it?

No. Though, it would potentially be a nice follow up to add a white list of players that should be mods, and/or to have bots contact the lobby to upgrade players to mods if they are also mods in the lobby. Those would be follow-up efforts though, out of scope for now.

To another extent, 'bans' do cascade down to bots. Lobby moderators could ban a player from the lobby and that would eject the player from the bot.

 Whether a player is given temporary 'mod' powers in a bot or not, I don't think matters relative to occupying a bot for too long.  Therefore, I don't see the "rule" really being impacted.

Cernelius:
> @Cernelius you make me wonder for a moment if we should use a different word than "mod".

This sounds like a very easy "yes" to me. Call it "controller", "assignee", "elder", "grantee" or whatever but "mod" or "moderator" please.

Imagine someone called "mod" or "moderator" because of this matter abusing users, and then we get complaints for that in forum as if we have abusive actual moderators.

DanVanAtta:
`[mod]` is a status. It just means the person can ban/boot. A person can have it in one context, but not another.

I appreciate the concern as noted. I think in that case, with our without the possible confusion, the person needs to be identified and the situation looked at in any case.  If the confusion becomes a real issue, and that plays out as such, we'll fix it.

Though, perhaps let's keep our naming reservations to ourselves for at least a few weeks. In terms of this feature, the naming of `[mod]` is arguably one of the least interesting. Let us discuss the most interesting matters first as needed; and then revisit any confusion once we've had some time to see how things play out.

PR #12780, opened 2024-07-26
Consolidate Dependencies & Cleanup

DanVanAtta:
(1) Move all non-project dependencies to the top level 'build.gradle'. Now that we essentially only package the game-client, it makes sense to have one place for all dependencies. One place is simpler, easier to manage & avoids redundancy. We also get a more concrete list of everything that we will be including in the final build, rather than having to inspect every sub-project.

(2) Simplify, now that lobby is in its own repository a number of dependencies are no longer needed (EG: JDBI for database connectivity).


TheDog-GH:
@DanVanAtta 
For info this commit added 3.5MB to the Windows 64 download, it jumped from 65.4MB to 68.9MB

DanVanAtta:
Thanks, that implies some new dependencies were added. I can spot a few now that probably can be removed.

DanVanAtta:
@TheDog-GH latest is back to 65MB    :+1:

PR #12781, opened 2024-07-26
Remove moderator right click option "fetch game chat history"

DanVanAtta:
We are no longer uploading chat messages from hosted (bot or player) games to lobby. This update removes the 'fetch game chat history' feature since that data will always be blank going forward.


DanVanAtta:
cc: @Cernelius thanks for pointing out this feature is legacy.

Issue #12789, opened 2024-07-28
2.7 ConcurrentModificationException Game Crash


DanVanAtta:
AI vs AI game crashed with:   `ConcurrentModificationException`

> 15:18:31.877 [AWT-EventQueue-0] ERROR o.t.game.client.HeadedGameRunner - null
java.util.ConcurrentModificationException: null
	at java.base/java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1043)
	at java.base/java.util.ArrayList$Itr.next(ArrayList.java:997)
	at java.base/java.util.Collections$UnmodifiableCollection$1.next(Collections.java:1047)
	at games.strategy.triplea.util.UnitSeparator.categorize(UnitSeparator.java:213)
	at games.strategy.triplea.util.UnitSeparator.categorize(UnitSeparator.java:199)
	at games.strategy.triplea.util.UnitSeparator.getSortedUnitCategories(UnitSeparator.java:96)
	at games.strategy.triplea.ui.SimpleUnitPanel.setUnits(SimpleUnitPanel.java:130)
	at games.strategy.triplea.ui.PlacePanel.lambda$updateStep$0(PlacePanel.java:151)

DanVanAtta:
Potentially fixed now.. with luck.

PR #12791, opened 2024-07-29
ConcurrentModification error: slightly hacky fix

DanVanAtta:
Should be a fix for #12789
This is not an ideal fix, and we likely have this problem in other areas. It is possible this is a non-fix too, we should watch out for further error reports to check carefully if this was a non-fix. 

## Notes to Reviewer

I'm not sure if this was a regression, I don't think it necessarily was. Perhaps a revealed problem after other innocuous changes.

### DIFF
--- a/game-app/game-headed/src/main/java/games/strategy/triplea/ui/PlacePanel.java
+++ b/game-app/game-headed/src/main/java/games/strategy/triplea/ui/PlacePanel.java
@@ -22,8 +22,10 @@ import java.awt.Component;
 import java.awt.Dimension;
 import java.awt.Toolkit;
 import java.awt.event.MouseEvent;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Predicate;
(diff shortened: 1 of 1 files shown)


game-app/game-headed/src/main/java/games/strategy/triplea/ui/PlacePanel.java
```diff
@@ -145,15 +146,23 @@ private void updateStep() {
...
     }
 
-    SwingUtilities.invokeLater(
-        () -> {
-          if (showUnitsToPlace) {
-            unitsToPlacePanel.setUnits(unitsToPlace);
+    if (showUnitsToPlace) {
+      // Small hack: copy the unit list before passing it to a new thread.
+      // This is to prevent ConcurrentModification. If the 'unitsToPlace' list is modified
+      // later in this thread, before "SwingUtilities.invokeLater" can execute and complete,
```
asvitkine, line 152: I don't think this comment is quite right. If it's modified before invokeLater, that's fine.

I think the issue is if it's being modified while the swing thread is iterating over it.

I think this is actually the right fix, so I don't agree with the end of the comment "Ideally we would not modify...". I don't the an API should place a restriction on the caller to not modify something that's passed to it just because its implementation does something async to it. If it does something async, it seems appropriate to make a copy.

DanVanAtta: Regarding time of modification:
- either there is a dirty flag that is set as soon as any modification is done (before we grab an iterator)
- or there is a dirty flag that is set after the iterator is grabbed

I think it is the first. I don't have time to research that right now. But the answer would be helpful.

As far as ownership & brittleness. Once the data is passed to a new thread - it seems like that new thread should then own all access & modifications to that data. What is the original thread doing modifying that data? That will create an inconsistent state between the other thread - it's a terribly brittle design. It would be less brittle if the data were immutable, and then any callbacks were owned by the new thread. 

I am guessing that at some point the entire flow was single threaded. That everything was done on the EDT, at some point that was corrected; but then just this last bit was put back on the EDT. Hence the threading issue was presented while still having this very odd flow of control.

PR #12808, opened 2024-08-01
New property to allow map maker the ability to allow non-air SBU units the ability to continue to the main territory battle.

WCSumpton:
I have created a new map property "Use Non-Air SBR Units In Normal Battle" that when set to true will allow these units to patriciate in the normal territory battle.

Sea units are still blocked from using "isStrategicBomber", so this has no effect on them. If this were to change, then they would also be allowed to join.

Because this is a new property, there should be no effect to existing maps.

Cheers...


frigoref:
@WCSumpton Just skimmed over it, is there any test for the new property? If not, would be good to create some concerning the new property.

TheDog-GH:
I will try this property out in 1941 GCD with Artillery.  It also has uses for siege warfare prior to WW2.

Issue #12777, opened 2024-07-26
Clarify the Difference between "Reset" and "Reset to Default" in "Settings" (2.7.14825)


Cernelius:
I assume that clicking on "Reset to Default" will return the "Settings" to their default values, but I don't know what would happen if I click on "Reset" instead (and how that is different from either "Save" or "Reset to Default").

I don't want to test it because I don't want to lose everything which I've customized in "Settings".

Regardless, whatever the difference is, I'm arguing that me not understanding it may mean that there is a substantial lack of clarity, which is the actual problem at this issue.

DanVanAtta:
Reset IIRC removes any 'unsaved' changes. A better tooltip or removing the button would be good moves.

Cernelius:
> Reset IIRC removes any 'unsaved' changes.

Then I would rename it to "Reset to Saved" or remove the button because you can do the same by closing without saving and then reopening.

> A better tooltip or removing the button would be good moves.

I'm not getting any tooltips. Besides, I don't like having tooltips appearing when hovering buttons (though I can agree with having them if they are needed).

Cernelius:
"Reset to Saved" would actually mean "Reset to the Latest Saved State". Just trying to keep it short.

frigoref:
@Cernelius Should be fixed with PR [12819](https://github.com/triplea-game/triplea/pull/12819) (Fix issues 12777 - Rename settings windows button from "Reset" to "Reset to Saved"). Please recheck and confirm.

Cernelius:
> @Cernelius Should be fixed with PR [12819](https://github.com/triplea-game/triplea/pull/12819) (Fix issues 12777 - Rename settings windows button from "Reset" to "Reset to Saved"). Please recheck and confirm.

I confirm it for the latest (2.7.14871) release.

PR #12918, opened 2024-10-03
Added tooltips for settings buttons. Fixed naming inconsistency.

kevinjin12:
## Notes to Reviewer
My first ever pull request!

Fixes Issue #12777 adding tooltips to clarify meaning. Also renamed engine preferences tab to maintain consistency with button name.

{3E06D535-3813-4629-8249-7DAB32E61528}

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/settings/SettingsWindow.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/settings/SettingsWindow.java
@@ -67,7 +67,7 @@ public enum SettingsWindow {
   public void open(final Frame owner) {
     Preconditions.checkState(SwingUtilities.isEventDispatchThread());
     if (dialog == null) {
-      dialog = new JDialog(owner, "Settings");
+      dialog = new JDialog(owner, "Engine Preferences");
       dialog.setContentPane(newContentPane());
       dialog.setMinimumSize(new Dimension(400, 50));
       dialog.pack();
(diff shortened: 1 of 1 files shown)


Cernelius:
Maybe "Settings" would be better than "Engine Preferences" for both?

frigoref:
I would leave the code as it is. The tooltips are just reiterating the text and are not necessary from my perspective. The popping up of a tooltip is not user friendly when there is not much added value.
Renaming "Settings" to something more specific would be good I think. How about "Engine Settings"?

Cernelius:
> How about "Engine Settings"?

"Program Settings" would feel better to me, but I'm not sure what is the difference between "program" and "engine" and which one is the better term.

I'll point out that the Windows folder where TripleA is installed is called "Programs", not "Engines".

Cernelius:
Also, maybe it should be "Program's Settings" or "Engine's Settings", shouldn't it?

frigoref:
For the button change [PR 12999](https://github.com/triplea-game/triplea/pull/12999) (Update SettingsWindow.java) has been merged.
The tooltips are not adding any value and are "rejected" as change.
Hence, there is nothing more at the moment in this PR to be merged.

@Cernelius, @kevinjin12 Please feel free to open an issue page here on GitHub or a forum thread in the [TripleA Forum](https://forums.triplea-game.org/) on the consistency of naming the different settings level.

PR #12955, opened 2024-10-19
Update Europe Improved

VictoryFirst2:
Updated map name and description


beelee1:
@VictoryFirst2 you might want to contact @RogerCooper to approve the workflow. I'm not exactly sure how that all works and don't want to screw anyting up, although it seems pretty straight forward and your changes are minor.

RogerCooper:
Everything is working correctly, I was to use the in-game downloader.

I will be away for the rest of the month with limited computer access.

PR #12958, opened 2024-10-20
Fixing ordering of GamePlayer in sidePanel

RenXFlo:
Fixed ordering for EconomyPanels in game (tested on 3 maps TWW included)

Oredering

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/GamePlayer.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/GamePlayer.java
@@ -201,7 +201,7 @@ public class GamePlayer extends NamedAttachable implements NamedUnitHolder {
     if (data == null) {
       return currentPlayers;
     }
-    for (final GamePlayer player : data.getPlayerList().getSortedPlayers()) {
+    for (final GamePlayer player : data.getPlayerList().getPlayers()) {
       currentPlayers.put(player.getName(), player.getPlayerType().name);
     }
     return currentPlayers;
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/GamePlayer.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/GamePlayer.java
@@ -201,7 +201,7 @@ public class GamePlayer extends NamedAttachable implements NamedUnitHolder {
     if (data == null) {
       return currentPlayers;
     }
-    for (final GamePlayer player : data.getPlayerList().getPlayers()) {
+    for (final GamePlayer player : data.getPlayerList().getSortedPlayers()) {
       currentPlayers.put(player.getName(), player.getPlayerType().name);
     }
     return currentPlayers;
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/PlayerList.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/PlayerList.java
@@ -83,10 +83,10 @@ public class PlayerList extends GameDataComponent implements Iterable<GamePlayer
   /** an iterator of a new ArrayList copy of the players. */
   @Override
   public Iterator<GamePlayer> iterator() {
-    return getPlayers().iterator();
+    return getSortedPlayers().iterator();
   }
 
-  public Stream<GamePlayer> stream() {
+  private Stream<GamePlayer> stream() {
     return players.values().stream();
   }
 
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/PlayerList.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/PlayerList.java
@@ -86,7 +86,7 @@ public class PlayerList extends GameDataComponent implements Iterable<GamePlayer
     return getSortedPlayers().iterator();
   }
 
-  private Stream<GamePlayer> stream() {
+  public Stream<GamePlayer> stream() {
     return players.values().stream();
   }
 
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/PlayerList.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/PlayerList.java
@@ -83,7 +83,7 @@ public class PlayerList extends GameDataComponent implements Iterable<GamePlayer
   /** an iterator of a new ArrayList copy of the players. */
   @Override
   public Iterator<GamePlayer> iterator() {
-    return getSortedPlayers().iterator();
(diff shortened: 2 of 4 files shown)


RenXFlo:
@frigoref could you look at this pull request and give me feedback

game-app/game-core/src/main/java/games/strategy/engine/data/PlayerList.java
frigoref, line 1: 1. Method `iterator`: Forcing the iterator to work on the sorted list is likely an overkill. 
2. Method `stream`: Did you check that it compiles afterwards? For me I get for example:
`triplea\game-app\game-core\src\main\java\games\strategy\triplea\attachments\AbstractConditionsAttachment.java:74: error: stream() has private access in PlayerList
          getData().getPlayerList().stream()`

game-app/game-core/src/main/java/games/strategy/engine/data/GamePlayer.java
frigoref, line 1: Forcing method `currentPlayers` to work on the sorted list is likely an overkill. We only want it for the UI.

game-app/game-core/src/main/java/games/strategy/triplea/printgenerator/PlayerOrder.java
frigoref, line 1: Fixes like this are good: Precise places where the sorting should be enforced.

game-app/game-core/src/main/java/games/strategy/triplea/util/PlayerOrderComparator.java
frigoref, line 1: Probably something for another fix, but the methods `GameStep.isXxxStep` all work on the game step name while their method names ...Step refer to the GameStep object. Hence, either the methods should be rename to something like ...StepName and/or the parameters passed should be rather of type GameStep.
Also, it seems odd to me to have `null` a a step name.
@DanVanAtta: Can you explain why/if this is really needed?

frigoref (commented):
Thanks for the PR!
Have a look for the comments. You can close the one where nothing is to be done (PlayerOrder) and leave the other one (PlayerOrderComparator) to @DanVanAtta.

game-app/game-core/src/main/java/games/strategy/engine/data/GamePlayer.java
RenXFlo: > Forcing method `currentPlayers` to work on the sorted list is likely an overkill. We only want it for the UI.

I'll revert the change.

game-app/game-core/src/main/java/games/strategy/engine/data/PlayerList.java
RenXFlo: > 1. Method `iterator`: Forcing the iterator to work on the sorted list is likely an overkill.
> 2. Method `stream`: Did you check that it compiles afterwards? For me I get for example:
>    `triplea\game-app\game-core\src\main\java\games\strategy\triplea\attachments\AbstractConditionsAttachment.java:74: error: stream() has private access in PlayerList getData().getPlayerList().stream()`

1. I'll revert the change.
2. Checked it didnt work afther I played with it and forgot to change it back. I'll change it in next commit.

frigoref:
@RenXFlo Did you check whether the ordering is still fixed?

RenXFlo:
> @RenXFlo Did you check whether the ordering is still fixed?

It is on my side aam I some anomaly ? Does it not work for you ?

frigoref:
> > @RenXFlo Did you check whether the ordering is still fixed?
> 
> @frigoref It is on my side aam I some anomaly ? Does it not work for you ?

@RenXFlo No, but I haven't tested it. So I wanted to confirm that after the changes done it is still giving the result from the screenshot of your [initial post in this thread](https://github.com/triplea-game/triplea/pull/12958#issue-2600174469).
When you confirm this I would merge the PR.

RenXFlo:
@frigoref Yes it still does work as shown in the image.

RenXFlo:
@frigoref Yes it still does work for me as shown in the initial post

WCSumpton:
Would there be any way of removing isHidden players from these side menus.

Thank you

Cheers...

WCSumpton:
"isHidden" is an optional setting for players and hides the player from in the player selection screen. Since this is talking about displaying players in the UI side panel. I was wondering if players marked "isHidden" could also be hidden from there.

Cheers...

frigoref:
@WCSumpton 
The available XML option `isHidden` is documented under [https://www.starlords3k.com/XOB.html](https://www.starlords3k.com/XOB.html) as

> Optional switch
> If true, does not show in player selection window.

Hence, it should **not** be used for a display property of a player in the table Politics.

@DanVanAtta Could you please review this PR as requested - meaning answering the question on file `PlayerOrderComparator.java` in section [Files changed](https://github.com/triplea-game/triplea/pull/12958/files)?

game-app/game-core/src/main/java/games/strategy/triplea/util/PlayerOrderComparator.java
DanVanAtta: > Can you explain why/if this is really needed?

Are you referring to: "Also, it seems odd to me to have null a a step name"?

If so, I'm not sure why null is a valid step name. I'm vaguely aware that step names are hardcoded in XML configs, so we do not have liberty to really change them. There could be a related reason why they could be null. Or, the nullity potentially could be some over-eager guarding of null values. The latter is very likely, instead of gating where/when things can be null, the code instead just guards against everything being possibly null.

The static methods on `GameStep` are interesting too. That problem is not super related to this update. Cleaning up to at least use the static utility methods is a good move IMO.

game-app/game-core/src/main/java/games/strategy/triplea/printgenerator/PlayerOrder.java
```diff
@@ -45,6 +46,7 @@ void saveToFile(final PrintGenerationData printData) throws IOException {
         playerSet.add(currentGamePlayer);
       }
     }
+    playerSet.sort(new PlayerOrderComparator(printData.getData()));
```
DanVanAtta, line 49: This seems to be the one really important line. It looks like this sort should be fine. The `PlayerOrderComparator` code is wonky, hard to understand. I don't like how there is an acquisition of a read lock that previously was not there, though that *should* not create any problems.

DanVanAtta (approved):
Seems okay

PR #12985, opened 2024-11-06
RepairRule as subclass to ProductionRule

frigoref:
GameParser.java
- use method replace instead of replaceAll
- remove methods parseRepairCosts and parseRepairResults
- parseRepairRules to reuse methods parseCosts and parseResults

Production.java
- reverted in 2nd commit: let inner class RepairRule extend from inner class ProductionRule

ProductionRule.java
- make private fields protected for visibility after inheritance

RepairRule.java
- inherit from ProductionRule and remove duplicates


frigoref:
@DanVanAtta  Do you see a way to use still let inner class RepairRule extend from inner class ProductionRule (in Production.java)?
When I tried to the first commit only, there were XML parsing errors, that I couldn't get fixed without reverting the changes.

PR #13142, opened 2025-01-25
Update to global 1940 maps

simon33-2:
## Notes to Reviewer
Version 2 of French Rebalance (renamed from Parisian Rebalance)
Version 4 of Balanced Mod for the Move 1st version
Move 1st versions now post from the Purchase phase in PBEM/PBF

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2803,7 +2803,7 @@
 - mapName: UHD World War II Global
   mapCategory: EXPERIMENTAL
   url: https://github.com/triplea-maps/uhd_world_war_ii_global/archive/master.zip
-  version: 6
+  version: 7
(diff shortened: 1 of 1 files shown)


beelee1:
@simon33-2 gotta sign the thing I guess :) I don't mind updating existing maps when the only thing that changed is the version number, as is the case here.

Last couple times I did new ones, it shut the lobby down or prevented map downloads. Don't really want to do those anymore. Was hoping someone more knowledgeable than me would.

I guess I could do the revert but if it doesn't work, not really anyone around that frequently to deal with it.

simon33-2:
Merging is blocked? I can't see the reason.

simon33-2:
What is needed to get this change done?

simon33-2:
Hold on, did I stuff this up? I really just wanted to merge in the change to the global version number. But it is still saying merging is blocked.

PR #13199, opened 2025-02-27
Update triplea_maps.yaml

RogerCooper:
Adding Election USA


beelee1:
@RogerCooper Idk if the mapName can have an underscore or not. I don't see any examples in the yaml. Also your older PR's say there's a conflict and can't be merged.

RogerCooper:
I see other maps with underscores in the map name, for example, G40_Alt_Universe.

The previous issues involving a colon in the map name have been resolved.

PR #13203, opened 2025-03-01
Update triplea_maps.yaml

RogerCooper:
Adding Downfall-July1944

### DIFF
--- a/triplea_maps.yaml
+++ b/triplea_maps.yaml
@@ -2992,3 +2992,10 @@
   img: https://raw.githubusercontent.com/triplea-maps/Election_Usa/blob/master/preview.png
   description: |
     <br>A game inspired by US elections.
+- mapName: Downfall-July1944
+  mapCategory: EXPERIMENTAL
+  url: https://github.com/triplea-maps/Downfall-July1944/archive/master.zip
+  version: 1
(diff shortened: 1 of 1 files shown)


beelee1:
I'm not adding anything that has a - _ or just plain regular wording. There is no one here to fix stuff when it screws up. You just doing this deliberately now ?

Whatever. I tried working with you but I'm done

RogerCooper:
I am sorry that this is causing problems for you. I have just been entering the names used by the mods themselves. There are already mods that have a "-" or "_" in their names, so I can't see how this is the source of problems. I am not using ":" in names or descriptions.

Everything I submit has passed through the automated checking process and has been run on my machine. (Only a few of these are my own mods)

There is not much point in having mods if the users can't download them. If triplea_maps.yaml is unreliable, then it should be deprecated and the users referred to https://triplea-game.org/maps-list/maps/

beelee1:
@RogerCooper 

My apologies. 

I'm concerned about the lobby shutting down again. 

Most maps in the yaml list just have a regular name. Will just merge and hope for the best

RogerCooper:
I am only going to submit one mod a week from now on so as to not overburden you.

Issue #13063, opened 2024-12-19
char limit of 40 in field "To:" under "Automatically send mails" introduced


EmilTophoej2:
### Reproduction Steps

1.  start client
2. press play by mail
3. fill out details 
4. in the "to" field add mails until you cross the char limit 

### Attach a Save Game

N/A

### Engine Version

Engine Version: 2.7.14900

### Is there anything else we should know?
method for sending mails to multiple persons in v 2.5, is to use a blank space as seperator 
mine and my opponents mail combined, are 48 chars - though I could imagine multiple people playing pbem

EmilTophoej2:
![41](https://github.com/user-attachments/assets/b1d05adf-4597-4ad7-87e9-e086a6343f34)
![40](https://github.com/user-attachments/assets/9daf2ba6-20cd-4d6e-b5ee-056fea214572)
screenshots

beelee1:
@EmilTophoej2 

Did 2.5 work differently ? Did it allow more charcters ?

EmilTophoej2:
Yes - dont know if there was a max previously

Did it before and also recently (few days ago before om a fresh
installation of the current stable version per normal download site -
before upgrading to newest as per GitHub)

See recent mail attached here for documentation

BR


søn. 22. dec. 2024 19.13 skrev beelee1 ***@***.***>:

beelee1:
ok I'll add a prerelease tag so hopefully this doesn't get buried

BBlaza:
Hello! How far did you go? Is that possible for me to take this issue?

BBlaza:
I found the code caused the problem. Do we still need a limit? Or removing it will be better?

BBlaza:
> It's hard to determine how many chars a mail are allowed to have in order
> to set a limit - does setting a limit improve  performance (not depending
> on actually amount filled in in the Field) ? If there has to be a limit; 40
> chars are not enough for my gamesetup.
> 
> I'd vote for the limit to be abolished
> 
> Although seldom more than 2 people could pbm or have direct spectators if
> no limit.
> 
> 
> Thank you
> BR
> 
> 
> tors. 20. mar. 2025 03.32 skrev Juncheng Jia ***@***.***>:
> […](#)

I finished fixing and I tries to test the functionality. However, neither the fixed version nor orginal version of the game works in my computer. May I see your screen recording of how you start a game with email?

frigoref:
Solves with PR 13230

Issue #5489, opened 2019-11-16
Proposal for Various Coding Conventions


DanVanAtta:
List of conventions that have come up in various PRs, hoping to discuss here and document so that we may make the call-outs and refer to the convention.

### A) Avoid mutability

We have two conventions set towards this:
1. use final where we can
2. favor immutability

The 'use final' where we can AFAIK is to encourage immutability. But, to another extent, if all we are saying is "use the eclipse setting to auto-apply final", it seems meaningless. We can write the following:
```
int value = 2;
if(condition) {
  value = 3;
}
```

That can be re-written to be immutable:
```
final int value = condition ? 3 : 2;
```

On the other hand, if we write:
```
int value = condition ? 3 : 2;
```

The above is flagged in review, but the first example would not be. It seems we might be missing the point.

I'd suggest we make our convention to be to "really" favor immutability and to take a second look at any mutable data in review and see if we can transform mutable data to be immutable, ie: "avoid mutable data". Either that, or perhaps we should drop the "apply final everywhere" since it's kinda meaningless.

### B) Favor `Collections.emptyList()` over `List.of`

`List.of()` is shorter than `Collections.emptyList()`, but it also seems more indirect as it's the same as `List.of(String[0])`, essentially an unused argument. It strikes me as the same as something like `size() == 0` vs `isEmpty()`, the latter is more direct to intent/meaning.

### C) Import inner classes

Instead of:
```
someClass.innerClass.methodCall();
```

Favor:

```
import someClass.innerClass;

innerClass.methodCall();
```

### D) Testing, favor using helper functions  for "given, when, then" like structures to tests

IE: we want to try and have tests that are as much as possible like:
```
void test() {
  givenConditionsReturningValue("value");
  String result = someObject.doSomething();
  assertThat(result, is("value"));
}
```

Essentially this is saying we should favor relatively short tests.

### E) Use `@DislayName` on tests

To help document what is being tested, use `@DisplayName`

```

@Test
@DisplayName("verify error handling for IOExceptions")
void errroHandling() {
   when(...).doThrow(new IOException());
   // test ...
}
```

### F) Use Hamcrest `assertThat`

Instead of
```
assertEquals(expected, actual)
```
Favor:
```
assertThat(actual, is(expected));

// or

assertThat(actual, equals(expected));
```

### G.1) Favor `Optional.ofNullable`

Instead of:
```
if(someObject != null) {
   someObject.methodCall();
}

// favor

[...]

PR #13230, opened 2025-03-21
Remove the char limit of 40 in field "To:" under "Automatically send"

BBlaza:
This PR aims to fix issue #13063
It should remove the char limit of 40 in field "To:" under "Automatically send".

## Notes to Reviewer
Althought the commit allowed field "To:" to take input the emails with more than 40 char, I couldn't check the functionality. Neither the original code nor the updated code works on my computer. I need a volunteer to test my code by sendung test email under the "Automatically send" field.

Screenshot 2025-03-20 at 9 13 18 PM


panther2:
@BBlaza Your problems might rather be related to the GMAIL issue, not accepting third party software for their mails.

Have you tried it with different mail-setups?

BBlaza:
> @BBlaza Your problems might rather be related to the GMAIL issue, not accepting third party software for their mails.
> 
> Have you tried it with different mail-setups?

Hello!
Yes I tried hotmail. It caused a time out.

beelee1:
@BBlaza idk about hotmail but gmx and protonmail both work

BBlaza:
> @BBlaza idk about hotmail but gmx and protonmail both work

I will try these later.

BBlaza:
> gmx

I tried both emails, neither of them works. They both caused a time out.

More than that, the "Help" window may need some updates. For "gmail" and "hotmail" cases, it said there are tab names "Login" and "Password". I assume that they refer to "Email Username" and "Email Password". For the generic SMTP case, more than the issues I mentioned above, it also mentioned "Host" and "Post". I assume that they refer to "SMTP Server" and "Port".

If my assumption is correct, I should did everything correctly. If I did some thing wrong, may I see how you do these?

beelee1:
@BBlaza yea I don't really know much of anything how the engine works code wise. Hopefully frigeroff jump back in. He's the only dev that's been active recently.

Only thing that occurs to me, is if removing it entirely instead of increasing character length would cause an unseen issue. I wouldn't think so. I guess that's why it needs to be tested though :)

BBlaza:
> @BBlaza yea I don't really know much of anything how the engine works code wise. Hopefully frigeroff jump back in. He's the only dev that's been active recently.
> 
> Only thing that occurs to me, is if removing it entirely instead of increasing character length would cause an unseen issue. I wouldn't think so. I guess that's why it needs to be tested though :)

Thanks for your respond! I I will change the limit to a number large enough instead of removing it.

beelee1:
@BBlaza  heh heh well, Idk if it's needed or not :) @DanVanAtta should be back from his bike ride at some point. He should be able to clue you in on what's needed. :)

frigoref:
@BBlaza, @beelee1, @TheDog-GH 
According to the blog [What Is the Maximum Length of an Email Address?](https://www.directedignorance.com/blog/maximum-length-of-email-address) the maximum email address length is 254. Shouldn't this be the value set now instead of `Integer.MAX_VALUE` (=2.147.483.647)?
If there is an option for a concatenation the field could be extended to `Integer.MAX_VALUE`, but there would have to be check on the actual limit for each email address to make sure each is <= 254.
Could you check and adjust the PR, please?

BBlaza:
> @BBlaza, @beelee1, @TheDog-GH According to the blog [What Is the Maximum Length of an Email Address?](https://www.directedignorance.com/blog/maximum-length-of-email-address) the maximum email address length is 254. Shouldn't this be the value set now instead of `Integer.MAX_VALUE` (=2.147.483.647)? If there is an option for a concatenation the field could be extended to `Integer.MAX_VALUE`, but there would have to be check on the actual limit for each email address to make sure each is <= 254. Could you check and adjust the PR, please?

I see. Thanks for responding. I will fix it as soon as possible.

BBlaza:
@frigoref 
I changed the checking function. It should check each email address individually. If any of them have length greater than 254, the input will not be accepted.

Moreover, I saw that the function validate() is a public function, while it will only be called by isValid() function in the same class. Doesn't it break the convention "Use most restrictive scope possible https://github.com/triplea-game/triplea/issues/5489"?

I saw the convention in docs/development/code-conventions/java-code-conventions.md.

game-app/domain-data/src/main/java/org/triplea/domain/data/PlayerEmailValidation.java
frigoref, line 1: @BBlaza If I understood your code correctly, you would like to count all characters that are not spaces and raise a message if this count is larger than the limit `LobbyConstants.EMAIL_MAX_LENGTH`  (=254), right?
The code above however already ensures (or at least should ensure) that the email is following a certain pattern by checking against a regular expression. Hence, at this place I think a simple check for if `(emailAddress.length() > LobbyConstants.EMAIL_MAX_LENGTH)` should be sufficient, no?

frigoref (commented):
Please check my comments and adjust the PR.

BBlaza (commented):
@frigoref
I though the same thing at the beginning.

However, the emailAddress can be something like "[usr0@gmail.com](mailto:usr0@gmail.com) [usr1@hotmail.com](mailto:usr1@hotmail.com) [usr2@gmx.com](mailto:usr2@gmx.com)". It includes multiple email addresses and they were separated by the space.

My code will set "count" to zero when it found a space, so it actually should count each email individually.

For example, an input like "longname123...(200characters)[...@gmail.com](mailto:...@gmail.com) anotherlongname123...(200characters)[...@gmail.com](mailto:...@gmail.com)" will be valid even the total length is greater than 254 because the count was set to zero when it have read the first email.

BBlaza:
> @BBlaza are you a bot ? You just posted the same three posts in a row lol Makes me skeptical of how serious you are :)

😂 Actaully I just want to do one.

The Github has a weird design that you have to post a comment in the PR to add a comment to code. So I posted two same things. And you saw three because the comment for code is a respond to comment of frigoref, so the comment will show on there too. I used a long time to figure out how does that work.

game-app/domain-data/src/main/java/org/triplea/domain/data/PlayerEmailValidation.java
frigoref: @BBlaza Understood, but then there is more that needs changing:

1. Class `LobbyConstants` should have a constant for the mail field with a name indicating that this constant is for the input field 
2. There should be a constant for the general length limit of a mail address with a proper name
3. The check should not work character by character, but split the content of the input field by `space` and perform the evaluation per mail address.
4. The invalid mail address(es) found should be mentioned in the error message instead of a generic error message. (check regex check above your changes)
5.  A (negative) test method (or more) should be added to class [PlayerEmailValidationTest.java](https://github.com/triplea-game/triplea/blob/master/game-app/domain-data/src/test/java/org/triplea/domain/data/PlayerEmailValidationTest.java)

BBlaza: I see.

How can I check the error message?

frigoref: @BBlaza I do not understand what the questions is about. The meaning of a "check" for an error message is not clear to me and I would ask you provide more information.
Also, it would be helpful for me to learn more about your java developer skillset when I am supposed to guide and assist you in this PR. This way I will be able what level of detail is needed for which areas in order to have an efficient communication.

BBlaza: @frigoref 
Sorry for the confusion.

You said that "The invalid mail address(es) found should be mentioned in the error message instead of a generic error message." I don't know how to see the error message and how to change it. I assumed it is the returned string, but I cannot find such message when I run the project on terminal.

About my java developer skillset, I'm familiar to most of the syntax and data structures. I understand how regex expression works. One major part of the project I don't really understand is the GUI. I'm glad to provide more information and learn more for the project if my skillset is not enough.

BBlaza:
@frigoref 
I modified the function, constant and test cases. I tested some input at the "To" field of "Automatically Send Emails" and it works as I expected. I don't know if I did the correct thing for the error message. Please have a check.

game-app/domain-data/src/test/java/org/triplea/domain/data/PlayerEmailValidationTest.java
frigoref, line 1: @BBlaza Great, just also add a comment to explain what the long strings are for (e.g. `//long, but valid mail` and `//too long mail`

game-app/domain-data/src/main/java/org/triplea/domain/data/PlayerEmailValidation.java
frigoref: @BBlaza Thanks, we are close, I think. There are some small point still though:

1. As regex usually need explanation (they are not easily readable for a human) leave a comment for what you try to achieve how with method `split` . 
2. My understand is that the split is supposed to be done by an `\\s*`. Maybe have a look for how the processing of the input field is done.
3. `StringBuilder` is probably best to help build return message - one for invalid addresses and another for addresses that are too long.
4. The return message should be build last from the explaining text(s) and the respective `StringBuilder` while avoiding to duplicate the explaining text(s) in the code.

If anything is not clear, I would try to find some time finalize this .

BBlaza:
@frigoref 
Glad to hear that we are close!

I updated the code following your guidance. Please checked out.

I tested on the "To" field again. Everything still works properly.

game-app/domain-data/src/main/java/org/triplea/domain/data/PlayerEmailValidation.java
BBlaza: @frigoref 
I can't find a proper way to change the error message to Optional<String> object because it is immutable. I changed it to null to avoid the last if statement.

BBlaza:
@frigoref 
I finished the updates you assigned except adding a limit to input field. Please have a check.

If you think such limit is important, please tell me which file handles the functionality. I will try to add it as soon as possible.

frigoref (approved):
@BBlaza Thanks for your patience and the effort! Highly appreciated!!!
And for sure: Congratulations to your first merged PR :)

PR #13260, opened 2025-03-31
Update triplea_maps.yaml

RogerCooper:
Adding Franco-Prussian War


RogerCooper:
When I tried posting all my mods, I received an error, so I just posted one mod. I will try some more next week.

Issue #13232, opened 2025-03-22
The "Help" for the "Automatically Send Email" needs some updates.


BBlaza:
### Reproduction Steps

1. Open the client
2. Click "Play by Email"
3. Click "Help"

### Engine Version

Engine Version: 2.7.14934

### Is there anything else we should know?
For "gmail" and "hotmail" cases, it said there are tabs name "Login" and "Password". I assume that they refer to "Email Username" and "Email Password".

For the generic SMTP case, more than the issues I mentioned above, it also said tabs name "Host" and "Post". I assume that they refer to "SMTP Server" and "Port".

frigoref:
@BBlaza Can you create a PR for the changes you have suggested?

BBlaza:
@frigoref I will do it when I'm free. Probably next month.

PR #13375, opened 2025-05-12
Text fix

BBlaza:
## Notes to Reviewer
fix issue #13232


frigoref (commented):
@BBlaza Thx for the PR!
Could you please check that we have a consistent writing of SMTP (all upper cases)?

frigoref (approved):

Issue #13481, opened 2025-06-18
TechnologyPanel#getColumnCount:277 - java.lang.ArrayIndexOutOfBoundsException


tripleabuilderbot:
## Map
shogun_advanced / Shogun Advanced

## Log Message
Failed to start game

## TripleA Version
[2.7.15081](https://github.com/triplea-game/triplea/releases/tag/2.7.15081)

## Java Version
21.0.7

## Operating System
Windows 11

## Heap Size
3912M

## Stack Trace
```
Exception: java.lang.ArrayIndexOutOfBoundsException Index 0 out of bounds for length 0
java.lang.Exception
	at games.strategy.triplea.ui.TechnologyPanel$TechTableModel.getColumnCount(TechnologyPanel.java:277)
	at java.desktop/javax.swing.JTable.createDefaultColumnsFromModel(JTable.java:1354)
	at java.desktop/javax.swing.JTable.tableChanged(JTable.java:4454)
	at java.desktop/javax.swing.JTable.setModel(JTable.java:3756)
	at java.desktop/javax.swing.JTable.<init>(JTable.java:705)
	at java.desktop/javax.swing.JTable.<init>(JTable.java:645)
	at games.strategy.triplea.ui.TechnologyPanel.createTechTable(TechnologyPanel.java:60)
	at games.strategy.triplea.ui.TechnologyPanel.initLayout(TechnologyPanel.java:56)
	at games.strategy.triplea.ui.TechnologyPanel.<init>(TechnologyPanel.java:50)
	at games.strategy.triplea.ui.TripleAFrame.<init>(TripleAFrame.java:398)
	at games.strategy.triplea.ui.TripleAFrame.lambda$create$7(TripleAFrame.java:492)
	at org.triplea.swing.SwingAction.lambda$invokeAndWaitResult$2(SwingAction.java:130)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:308)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:773)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:720)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:714)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:400)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:87)
[...]

PR #13483, opened 2025-06-18
fix13481_TechnologyPanel#getColumnCount (ArrayIndexOutOfBoundsException)

frigoref:
TechnologyPanel.java
- remove calls to deprecated addDataChangeListener and use addGameDataEventListener(TECH_ATTACHMENT_CHANGED,runnable)
- method getColumnCount: Use variable colList instead of data (case no data exists)

Fix to issue https://github.com/triplea-game/triplea/issues/13481 introduced by commit https://github.com/triplea-game/triplea/commit/a4e94198c9cc96eda2e106e3bf84c90c90d3b47f#diff-52f92c2a0aaec220571b0e19ec36369548e5b6526a006546a91e4e3487ab80c3

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/TechnologyPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/TechnologyPanel.java
@@ -5,15 +5,16 @@ import static games.strategy.triplea.Constants.SIGN_TECH_NOT_ENABLED;
 
 import games.strategy.engine.data.Change;
 import games.strategy.engine.data.GameData;
+import games.strategy.engine.data.GameDataEvent;
 import games.strategy.engine.data.GamePlayer;
 import games.strategy.engine.data.events.GameDataChangeListener;
-import games.strategy.engine.stats.IStat;
 import games.strategy.triplea.Constants;
 import games.strategy.triplea.delegate.TechAdvance;
 import games.strategy.triplea.delegate.TechTracker;
 import java.awt.Component;
 import java.awt.GridLayout;
 import java.awt.Image;
+import java.text.MessageFormat;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
@@ -36,10 +37,9 @@ import org.jetbrains.annotations.NotNull;
 
 class TechnologyPanel extends JPanel implements GameDataChangeListener {
 
-  IStat[] stats;
   private final Map<GamePlayer, ImageIcon> mapPlayerImage = new HashMap<>();
   protected GameData gameData;
-  private final UiContext uiContext;
+  private final transient UiContext uiContext;
   private final TechTableModel techModel;
 
   TechnologyPanel(final GameData data, final UiContext uiContext) {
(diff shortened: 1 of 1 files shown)


game-app/game-core/src/main/java/games/strategy/triplea/ui/TechnologyPanel.java
```diff
@@ -36,10 +37,9 @@
 
 class TechnologyPanel extends JPanel implements GameDataChangeListener {
 
-  IStat[] stats;
   private final Map<GamePlayer, ImageIcon> mapPlayerImage = new HashMap<>();
   protected GameData gameData;
-  private final UiContext uiContext;
+  private final transient UiContext uiContext;
```
DanVanAtta, line 42: Why transient?

game-app/game-core/src/main/java/games/strategy/triplea/ui/TechnologyPanel.java
```diff
@@ -271,10 +274,9 @@ public Object getValueAt(final int row, final int col) {
       return data[row][col];
     }
 
-    // Trivial implementations of required methods
     @Override
     public int getColumnCount() {
-      return data[0].length; // the number of columns of the 2D-array data
+      return colList.length + 1;
```
DanVanAtta, line 279: Why the +1?

frigoref: @BBlaza Do you want to answer that, I was just reinstating the status from before.

BBlaza: colList is actually a list of player's name. The complete table has the first column listing the name of technologies, so adding one is necessary. 
The previous code doesn't work because the structure of JTable includes column header, which allows empty columns. Therefore, if every column is empty, data[0] doesn't exist.

DanVanAtta: Thanks for the context = D
It might not be super obvious in the code to someone reading through this. A comment perhaps might be of value to call that out (?)

I was wondering why the code was not updated to something like this instead?

```
data.length == 0 ? 1 : data[0].length
```

I imagine that is the same, right? If so, getting the number of columns from the list of columns is very reasonable :+1:

BBlaza: @DanVanAtta 
Not really, 😢
It's possible that there are many empty columns, like this:
```
+----------+----------+----------+
| Name   1   | Name    2  | Name   3  |  
+----------+----------+----------+

+----------+----------+----------+
```
Therefore, your code will return 1 while it's actually 3. I guess the colList is necessary. I totally agree that adding a comment will be helpful, I will create a pr very soon.

BBlaza: @DanVanAtta This is the pr #13496

PR #13502, opened 2025-06-21
Update triplea_maps.yaml

RogerCooper:
Adding Future Wars-Fallen Empire, Battle of Iwo Jima & Spoils of War


RogerCooper:
I will try again with different url formatting.

PR #13508, opened 2025-06-23
GameParseException for getResourceOrThrow/getUnitTypeOrThrow

frigoref:
GameParser.java
- multiple methods: MessageFormat for GameParseException
- methods getUnitType/getUnitTypeOrThrow/getResourceOrThrow: Rethrow IllegalArgumentException as GameParseException
- methods parseProductionRules/parseRepairRules: Reworked to avoid duplicated code
- method parseCosts: Renamed to ProductionRule
- new method parseCostsForRule: Reuse method for ProductionRule and RepairRule

Rule.java /
ProductionRule.java / RepairRule.java
- new method: addCost

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/ProductionRule.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/ProductionRule.java
@@ -33,13 +33,14 @@ public class ProductionRule extends DefaultNamed implements Rule {
     this.costs = costs;
   }
 
-  public void addCost(final Resource resource, final int quantity) {
-    costs.put(resource, quantity);
-  }
-
   @Override
   public IntegerMap<Resource> getCosts() {
     return new IntegerMap<>(costs);
+  }
+
+  @Override
+  public IntegerMap<Resource> getInternalCosts() {
+    return costs;
   }
 
   @Override
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/RepairRule.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/RepairRule.java
@@ -2,11 +2,12 @@ package games.strategy.engine.data;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import java.io.Serial;
 import org.triplea.java.collections.IntegerMap;
 
 /** A repair rule. */
 public class RepairRule extends DefaultNamed implements Rule {
-  private static final long serialVersionUID = -45646671022993959L;
+  @Serial private static final long serialVersionUID = -45646671022993959L;
 
   private final IntegerMap<Resource> costs;
   private final IntegerMap<NamedAttachable> results;
@@ -29,13 +30,14 @@ public class RepairRule extends DefaultNamed implements Rule {
     this.results = new IntegerMap<>(results);
   }
 
-  public void addCost(final Resource resource, final int quantity) {
-    costs.put(resource, quantity);
-  }
-
   @Override
   public IntegerMap<Resource> getCosts() {
     return new IntegerMap<>(costs);
+  }
+
+  @Override
+  public IntegerMap<Resource> getInternalCosts() {
+    return costs;
   }
 
   @Override
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
@@ -264,7 +264,10 @@ public final class GameParser {
 
   private GamePlayer getPlayerId(final String name) throws GameParseException {
     return getPlayerIdOptional(name)
-        .orElseThrow(() -> new GameParseException("Could not find player name:" + name));
+        .orElseThrow(
+            () ->
+                new GameParseException(
+                    MessageFormat.format("Could not find player name: {0}", name)));
   }
 
   private Optional<GamePlayer> getPlayerIdOptional(final String name) {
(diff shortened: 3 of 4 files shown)


game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
```diff
@@ -311,18 +333,27 @@ private TechAdvance getTechnology(final String name) throws GameParseException {
...
   }
 
   /** If must find is true and cannot find the Resource, an exception will be thrown. */
   private Resource getResourceOrThrow(final String name) throws GameParseException {
-    return data.getResourceList().getResourceOrThrow(name);
+    try {
+      return data.getResourceList().getResourceOrThrow(name);
+    } catch (final IllegalArgumentException e) {
+      throw new GameParseException(e.getMessage(), e);
+    }
```
DanVanAtta, line 356: Throwing and catching and rethrowing is a bit gross.
This syntax is a bit more straight forward:

```
getResourceOptional().orElseThrow(() -> new GameParseException("could not find resource name: " + name + ", list of available resources is: " + data.getResourceList().getResources());
```

game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
```diff
@@ -663,7 +703,7 @@ private void parseResults(final Rule dataRule, final Production.Rule mapRule)
     List<Production.Rule.Result> ruleResults = mapRule.getRuleResults();
     if (ruleResults.isEmpty()) {
       throw new GameParseException(
-          MessageFormat.format("No results for rule {0}", dataRule.getName()));
+          MessageFormat.format("No results for production rule: {0}", dataRule.getName()));
```
DanVanAtta: Looks like this might be either 'repair' rule or 'production rule'.
The message " for production rule" might be incorrect (?)

game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
frigoref: I am trying to avoid the same error string to have multiple times. Also the rethrow allows the to see the chain from where we came from.

DanVanAtta: >  Also the rethrow allows the to see the chain from where we came from.

Somewhat, it's one call stack lower.

Looking at this method a bit more - it's really not adding much. I'd call this a "[feature envy](https://refactoring.guru/smells/feature-envy)" code smell. I suspect this private helper method ought to be in-lined and updated so that there is a good error message. These better error messages could tell someone where the missing resource was found within the XML and give more context. The most important audience for these messages are map makers, the user. Essentially zero developers are going to be looking at any stack traces here - the error messaging is for the benefit of the user (a map makers).

Making those types of updates would be better for a follow up PR IMO rather than adding to this current diff.

game-app/game-core/src/main/java/games/strategy/engine/data/RepairRule.java
```diff
@@ -2,11 +2,12 @@
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import java.io.Serial;
 import org.triplea.java.collections.IntegerMap;
 
 /** A repair rule. */
 public class RepairRule extends DefaultNamed implements Rule {
-  private static final long serialVersionUID = -45646671022993959L;
+  @Serial private static final long serialVersionUID = -45646671022993959L;
```
DanVanAtta, line 10: Please check out this article: https://github.com/projectlombok/lombok/wiki/WHY-NOT:-serialVersionUID

I suspect future maintainers might wonder why this one `serialiVersionUID` has a `@Serial` and the others do not. If it's not needed for correctness, let's keep consistency - sound reasonable?

frigoref: @DanVanAtta From my perspective it adds clarity as it does not depend on naming convention to express the usage. Also, I thought that this is more stable in the future for IDEs. It is really a minor thing and I agree it should be consistent everywhere, but I would rather vote to have '@Serial' for each field (the link did not provide me an argument, not to use it).

game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
frigoref: Adjusted.

game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
frigoref: K, then we reuse the error message from the `IllegalArgumentException`?

DanVanAtta: Well, the `IllegalArgumentException` does not have enough context to give a good error message. IMO, that exception should not be thrown and instead handled by returning an empty optional. Then, this code would be responsible for constructing the error message (including context).

I think that is how it would look like in a slightly idealized context. I would just leave it and move on, or follow up with another PR to do really good error messages - which would mean we would return optionals and this parsing code would be the thing to construct the error messages. In that case, there is just less `IllegalArgumentException` to deal with.

game-app/game-core/src/main/java/games/strategy/engine/data/ProductionRule.java
```diff
@@ -33,15 +33,16 @@ public ProductionRule(
...
-    costs.put(resource, quantity);
-  }
-
   @Override
   public IntegerMap<Resource> getCosts() {
     return new IntegerMap<>(costs);
   }
 
+  @Override
+  public IntegerMap<Resource> getInternalCosts() {
```
DanVanAtta, line 42: Question - any reason to not use `getCosts()` defined in 'L:37'? Are we just trying to avoid the data copy?

game-app/game-core/src/main/java/games/strategy/engine/data/RepairRule.java
DanVanAtta: I agree on clarity. It'll just be "interesting" to see it in isolated places and not others. Makes a person wonder if something was blocking the annotation from being added to other locations.

The annotation is good though. I'm in favor of it being bulk added.

DanVanAtta: My primary consideration overall is consistency within a codebase is king.

frigoref: Then l'll prepare another PR for the consistent use of the @Serial annotation, okay?

game-app/game-core/src/main/java/games/strategy/engine/data/ProductionRule.java
frigoref: Yes, getCosts is creating a copy and here we need the original IntegerMap to work on.

DanVanAtta: This PR is to follow up: https://github.com/triplea-game/triplea/pull/13515

An interface method solely to return internal data breaks encapsulation.

PR #13515, opened 2025-06-25
Refactor: better encapsulation in Rule.java around adding costs

DanVanAtta:
Rather than exposing internal data through the interface: Rule.java interface, we add the 'addCost' method directly to Rule.java interface

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/Rule.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/Rule.java
@@ -20,17 +20,7 @@ public interface Rule {
 
   IntegerMap<NamedAttachable> getResults();
 
-  IntegerMap<Resource> getInternalCosts();
-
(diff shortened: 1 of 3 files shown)


game-app/game-core/src/main/java/games/strategy/engine/data/Rule.java
```diff
@@ -20,17 +20,7 @@ public interface Rule {
...
-  /**
-   * Add resource cost to rule.
-   *
-   * @param resource of the cost
-   * @param quantity of the {@code resource}
-   */
-  default void addCost(final Resource resource, final int quantity) {
-    getInternalCosts().put(resource, quantity);
-  }
+  void addCost(final Resource resource, final int quantity);
```
DanVanAtta, line 23: We have one less method on the interface and we no longer require implementations to expose mutable internal data structures

PR #13516, opened 2025-06-25
Calculate size of downloaded maps when selected

frigoref:
- method newMapUrlAndSizeLabelText: Extracted logic to separate methods
- new method appendSizeAndPath: Includes sum up in case multiple downloaded maps are selected
- new method appendSizeAndPathSingle

FileUtils.java
- new method getFolderSize
- new method getByteSizeFromPath

## Notes to Reviewer
I tried to also get the size from the maps selected to be downloaded, but the GitHub-URLs did never return the size and a download just to know the size was pointless.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadMapsWindow.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadMapsWindow.java
@@ -15,7 +15,6 @@ import java.awt.FlowLayout;
 import java.awt.Rectangle;
 import java.io.IOException;
 import java.io.Serial;
-import java.nio.file.Files;
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -48,7 +47,6 @@ import javax.swing.table.TableModel;
 import javax.swing.table.TableRowSorter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.io.FileUtils;
-import org.jetbrains.annotations.NonNls;
 import org.jetbrains.annotations.NotNull;
 import org.triplea.http.client.maps.listing.MapDownloadItem;
 import org.triplea.java.Interruptibles;
@@ -186,7 +184,6 @@ public class DownloadMapsWindow extends JFrame {
     if (selectedMapItems.isEmpty()) {
       return "<html>None selected</html>";
     }
-    final @NonNls String doubleSpace = "&nbsp;&nbsp;";
 
     String mapsString =
         String.join(", ", selectedMapItems.stream().map(MapDownloadItem::getMapName).toList());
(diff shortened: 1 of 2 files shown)


game-app/game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadMapsWindow.java
```diff
@@ -186,49 +184,83 @@ private String newMapUrlAndSizeLabelText(final List<MapDownloadItem> selectedMap
...
-    }
+    appendSizeAndPath(selectedMapItems, sb);
     sb.append("<br>");
     sb.append("</html>");
 
     return sb.toString();
   }
 
+  private void appendSizeAndPath(
+      final List<MapDownloadItem> selectedMapItems, final StringBuilder sb) {
```
DanVanAtta: Do you have any thoughts to rewrite these methods to have no side effects?

DanVanAtta:
What is the performance like of this update for the very large maps? I believe TWW is the largest map at like 200MB or something.

game-app/game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadMapsWindow.java
frigoref: Any ideas? I did not think to rewrite these methods.

DanVanAtta: `private void` methods are arguably evil. Namely because they are implicitly stateful.

These methods are calculating multiple things. To make the methods functional, mark them as static and have them return the things calculated. Since we are calculating multiple things - have multiple methods. Try to keep control-flow logic in the "top most method" as much as possible as well.

frigoref: Please check after new commit https://github.com/triplea-game/triplea/pull/13516/commits/5abf2a1359eb7eb29a280f85b22462fe0076d805

DanVanAtta: Added comments to the commit

lib/java-extras/src/main/java/org/triplea/io/FileUtils.java
```diff
@@ -37,6 +41,47 @@ public static Path newTempFolder() {
...
   }
 
+  /**
+   * Calculates the folder size in bytes.
+   *
+   * @param folderPath assumed folder path
+   * @return sum of bytes of the files contained in the folder
+   * @throws IOException if an I/O error is thrown by a visitor method
+   */
+  private static long getFolderSize(Path folderPath) throws IOException {
```
DanVanAtta: nit, private methods should be defined after their first usage, not before.
The ordering of methods within a class  generally should be organized depth-first

lib/java-extras/src/main/java/org/triplea/io/FileUtils.java
```diff
@@ -37,6 +41,47 @@ public static Path newTempFolder() {
...
+    final long[] size = {0};
+
+    Files.walkFileTree(
+        folderPath,
```
DanVanAtta: Does `walkFileTree` work regardless of whether we have a directory or a file? If so, we don't need the check for `is a file` and this method can be combined into `getByteSizeFromPath`

frigoref: The check is already done in method `getByteSizeFromPath`

DanVanAtta: Yeah, what I'm saying is that check is potentially not needed. You can maybe simplify and just walk the file tree whether the parameter is a file or a directory. 

The javadocs for the method make it sound like the parameter can be a `file`:



> Where the file is a directory....

From: 
https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#walkFileTree-java.nio.file.Path-java.nio.file.FileVisitor-

frigoref:
I hope I have not missed any of the already clarified comments in my new commit https://github.com/triplea-game/triplea/pull/13516/commits/2ca1ac53370f5d3f31f46a870d8465a16fd12bf0.
Please check and let me know.

game-app/game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadMapsWindowModel.java
```diff
@@ -29,6 +32,24 @@ Optional<Path> getInstallLocation(final MapDownloadItem mapDownloadItem) {
     return installedMapsListing.findMapFolderByName(mapDownloadItem.getMapName());
   }
 
+  public Optional<@NonNls String> getByteSizeTextForMap(final MapDownloadItem mapDownloadItem) {
```
DanVanAtta, line 35: Should return a number here. Returning a string is mixing the UI and model logic together.

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
```diff
@@ -633,12 +634,12 @@ private void updateStats() {
     final double avgDefIfDefWon = results.getAverageDefendingUnitsLeftWhenDefenderWon();
     defenderLeftWhenDefenderWon.setText(
         Double.isNaN(avgDefIfDefWon)
-            ? "N/A"
+            ? DownloadMapsWindow.TEXT_ABBREVIATION_NOT_APPLICABLE
```
DanVanAtta, line 649: IMO, leave the string inlined here. Why should `DownloadMapsWindow` own the `N/A` string?

I don't think the constant is really helping very much, it's okay to let things be inlined and duplicated sometimes.

frigoref: It is needed to make it translateble only once. The usage is the same, but I did not want to create an own class for it. At leadt not now when there are no others yet.

game-app/game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadMapsWindowModel.java
frigoref: What with exception case?
Maybe you provide a commit for this to show me what you think is best.

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
DanVanAtta: Sure, but we're not making it translatable just yet. Let's do this work when we do the other thing. Otherwise this is just half done. Let's leave it be unless we do the whole effort.

game-app/game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadMapsWindowModel.java
DanVanAtta: Change: 

```
byteSizeText =
                    org.apache.commons.io.FileUtils.byteCountToDisplaySize(
                        org.triplea.io.FileUtils.getByteSizeFromPath(mapPath));
```

To:

```
   return                    org.apache.commons.io.FileUtils.byteCountToDisplaySize(
                        org.triplea.io.FileUtils.getByteSizeFromPath(mapPath));

```

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
frigoref: We are outsourcing the text and combine 3 places into one. I do not see what benefit it has to undo it in order to redo it again in the future.

game-app/game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadMapsWindowModel.java
frigoref: I think this is resolved now, right?

lib/java-extras/src/main/java/org/triplea/io/FileUtils.java
frigoref: @DanVanAtta I am not clear one what that would help: If the path is not a file what is the expectation when calling method `walkFileTree`? From my perspective the way it is written now it is more clear what is supposed to happen.

DanVanAtta: > If the path is not a file what is the expectation when calling method walkFileTree?

Returns just the file

> From my perspective the way it is written now it is more clear what is supposed to happen.

I think questions are raised for why walkFileTree not being used. It's overly complicated, the `getByteSizeFromPath` seems entirely unnecessary.The questions are specifically things like "what did the author know that I don't that makes this not usable?" It's exactly the same scenario where we were wondering about `ppt.getTerritory()` and why that value was not being used directly - it's a mystery why it's coded that way since a simpler solution seems obvious.

DanVanAtta: If I misunderstand how `walkFileTree` works, then the two methods are reasonable. Let's experiment a bit.

The javadocs sound like we should be getting back the same file:

> This method walks a file tree rooted at a given starting file. The file tree traversal is depth-first with the given [FileVisitor](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/file/FileVisitor.html) invoked for each file encountered.

https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/file/Files.html#walkFileTree(java.nio.file.Path,java.util.Set,int,java.nio.file.FileVisitor)

game-app/game-core/src/main/java/games/strategy/engine/framework/map/download/DownloadMapsWindowModel.java
DanVanAtta: I don't see any updates since my original comments nor any discussion based on the suggestion. I don't see any evidence for how this was resolved - I don't think it was.

game-app/game-core/src/main/java/games/strategy/triplea/odds/calculator/BattleCalculatorPanel.java
DanVanAtta: > I do not see what benefit it has to undo it in order to redo it again in the future.

I'm suggesting that is premature optimization. Who is to say this class will even exist in its current form when we do redo it. The rule of thumb, have the simplest possible code for now, do the other stuff when it is needed. Would you prefer to add a feature to the simplest possible code base, or would you prefer to add a feature that tried to predict all the features that were going to be added?

To another extent, this PR is to "calculate the size of downloaded maps when selected." What does that have to do with internationalizing the string N/A?

PR #13522, opened 2025-06-28
Update triplea_maps.yaml

RogerCooper:
Adding Churchills 1939 Path to Victory, Forty Thousand, Nova Games Edition, Wolfys 1939.


beelee1:
Is it ok to use "Nova Games" in the title ?

RogerCooper:
I don't see why not. The 1981 edition of Axis & Allies is long out of print. Nova Games is still around, but they don't have the rights to A&A.

PR #13579, opened 2025-07-26
No return null (get territory)

frigoref:
**noReturnNull (getTerritory) #5**

GameMap.java
- method getTerritoryOrNull: Replaced by getTerritory
- new method getTerritory: Returning Optional and replacing each place formerly calling getTerritoryOrNull

ProSimulateTurnUtils.java
- methods transferMoveMap/transferUnit/transferLoadedTransport: Use getTerritoryOrThrow instead of getTerritoryOrNull

**noReturnNull (getTerritory) #4**

AbstractPlaceDelegate.java
BidPlaceDelegate.java
IAbstractPlaceDelegate.java
- methods placeUnits/isValidPlacement/canProduce/getAllProducers/checkProduction/getBestProducerComparator: Annotate territory parameter NotNull

ProPlaceTerritory.java
ProPurchaseTerritory.java
- Constructor/attribute territory: Annotate territory NotNull

ProPurchaseAi.java
- method place: Rename data to proGameData and use getTerritoryOrThrow as null is not handled
- method doPlace: Annotate territory parameter NotNull

ProSimulateTurnUtils.java
- method simulateBattles: remove re-read territory after owner change (new data already updated)

**noReturnNull (getTerritory) #3**
triplea\Constants.java
- Importing further file name/path constants

Others:
- Adjust to extraction of file constants

**noReturnNull (getTerritory) #2**
DecorationPlacer.java
- new methods loadPolygonsFile, loadCentersFile, selectAndLoadPolygonsFile, selectAndLoadCentersFile: Helper methods to load respective files

PlacementPicker.java
- Constructor: replace load code for polygons with new method loadPolygonsFile

PolygonGrabber.java
- Constructor: replace load code for centers with new method loadCentersFile

MapData.java
- Extraction of file name constants into Constants.java

triplea\Constants.java
- Importing file name constants

Others:
- Adjust to renaming/moving constants

**noReturnNull (getTerritory)**
*Test*.java
- usage of method getTerritoryOrThrow instead of getTerritory

DefaultAttachment.java
- method getTerritoryOrThrow: replaces with getTerritoryOrThrowGameParseException
- new methods getTerritoryOrThrowGameParseException/getTerritoryOrThrowIllegalStateException: Throwing specific exception when no territory was found
- new method getTerritory: private method returning Optional to avoid returning null

GameMap.java
- method getTerritory: Renamed to getTerritoryOrNull
- method getTerritoryOrThrow: Documentation added
- methods getDistance: Parameter annotation NotNull added

GameSelectorModel.java
- method parseAndValidate: Return Optional instead of null
- method setGameData: Parameter annotation Nullable added

GameTestUtils.java
- method getTerritory: Replaced with getTerritoryOrThrow

UnitTypeList.java
- method getUnitTypes: Documentation updated

## Notes to Reviewer
- Replaces null `return` on `getTerritory`
- Introduces new method `getTerritoryOrThrow` for cases where the territory is expected to be found
- Extracts file name constants into triplea\Constants.java
- DecorationPlacer.java
- new methods loadPolygonsFile, loadCentersFile, selectAndLoadPolygonsFile and selectAndLoadCentersFile in DecorationPlacer.java to encapsulate helper logic to load respective files

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
@@ -20,6 +20,7 @@ import javax.annotation.Nullable;
 import lombok.Getter;
 import lombok.Setter;
 import org.jetbrains.annotations.NonNls;
+import org.jetbrains.annotations.NotNull;
 import org.triplea.java.collections.IntegerMap;
 
 /**
@@ -173,9 +174,25 @@ public abstract class DefaultAttachment extends GameDataComponent implements IAt
         && (Objects.equals(name, other.name) || this.toString().equals(other.toString()));
   }
 
-  protected Territory getTerritoryOrThrow(String name) throws GameParseException {
-    return Optional.ofNullable(getData().getMap().getTerritory(name))
-        .orElseThrow(() -> new GameParseException("No territory named: " + name + thisErrorMsg()));
+  private Optional<Territory> getTerritory(@Nullable String territoryName) {
+    return Optional.ofNullable(getData().getMap().getTerritoryOrNull(territoryName));
+  }
+
+  @NotNull
+  protected Territory getTerritoryOrThrowGameParseException(String territoryName)
+      throws GameParseException {
+    return getTerritory(territoryName)
+        .orElseThrow(
+            () -> new GameParseException("No territory named: " + territoryName + thisErrorMsg()));
+  }
+
+  @NotNull
+  protected Territory getTerritoryOrThrowIllegalStateException(@Nullable String territoryName)
+      throws IllegalStateException {
+    return getTerritory(territoryName)
+        .orElseThrow(
+            () ->
+                new IllegalStateException("No territory named: " + territoryName + thisErrorMsg()));
   }
 
   protected List<GamePlayer> parsePlayerList(final String value, List<GamePlayer> existingList)
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
@@ -174,14 +174,12 @@ public abstract class DefaultAttachment extends GameDataComponent implements IAt
         && (Objects.equals(name, other.name) || this.toString().equals(other.toString()));
   }
 
-  private Optional<Territory> getTerritory(@Nullable String territoryName) {
-    return Optional.ofNullable(getData().getMap().getTerritoryOrNull(territoryName));
-  }
-
   @NotNull
   protected Territory getTerritoryOrThrowGameParseException(String territoryName)
       throws GameParseException {
-    return getTerritory(territoryName)
+    return getDataOrThrow()
+        .getMap()
+        .getTerritory(territoryName)
         .orElseThrow(
             () -> new GameParseException("No territory named: " + territoryName + thisErrorMsg()));
   }
@@ -189,7 +187,9 @@ public abstract class DefaultAttachment extends GameDataComponent implements IAt
   @NotNull
   protected Territory getTerritoryOrThrowIllegalStateException(@Nullable String territoryName)
       throws IllegalStateException {
-    return getTerritory(territoryName)
+    return getDataOrThrow()
+        .getMap()
+        .getTerritory(territoryName)
         .orElseThrow(
             () ->
                 new IllegalStateException("No territory named: " + territoryName + thisErrorMsg()));
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java
@@ -86,28 +86,31 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
   }
 
   /**
-   * Case-sensitive search for {@link Territory} by name. *
+   * Case-sensitive search for {@link Territory} by name. Use {@link #getTerritoryOrThrow} if it is
+   * expected that the territory is found.
    *
    * @param territoryName name of the searched territory (case-sensitive)
-   * @return Territory with the given name or {@code null} if no territory can be found.
+   * @return {@link Optional} of {@link Territory} with the given name.
    */
-  public @Nullable Territory getTerritoryOrNull(final String territoryName) {
-    return territoryLookup.get(territoryName);
+  public Optional<Territory> getTerritory(final @NotNull String territoryName) {
+    return Optional.ofNullable(territoryLookup.get(territoryName));
   }
 
   /**
-   * Case-sensitive search for {@link Territory} by name.
+   * Case-sensitive search for {@link Territory} by name. If it is not found a {@link
+   * IllegalArgumentException} is thrown. Use {@link #getTerritory} if it is expected that the
+   * territory might not be found.
    *
    * @param territoryName name of the searched territory (case-sensitive)
    * @return Territory with the given name or {@link Territory} if no territory can be found.
    */
-  public Territory getTerritoryOrThrow(final String territoryName) {
-    return Optional.ofNullable(getTerritoryOrNull(territoryName))
-        .orElseThrow(
-            () ->
-                new IllegalArgumentException(
-                    MessageFormat.format(
-                        "Territory with name {0} could not be found", territoryName)));
+  public Territory getTerritoryOrThrow(final @NotNull String territoryName) {
+    final Territory t = territoryLookup.get(territoryName);
+    if (t == null) {
+      throw new IllegalArgumentException(
+          MessageFormat.format("Territory with name {0} could not be found", territoryName));
+    }
+    return t;
   }
 
   /**
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java
@@ -23,6 +23,7 @@ import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import org.jetbrains.annotations.NotNull;
 import org.triplea.java.RemoveOnNextMajorRelease;
 import org.triplea.java.collections.IntegerMap;
 
@@ -63,7 +64,7 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
     territoryLookup.put(t1.getName(), t1);
   }
 
-  /** Bi-directional. T1 connects to T2, and T2 connects to T1. */
+  /** Bidirectional. T1 connects to T2, and T2 connects to T1. */
   public void addConnection(final Territory t1, final Territory t2) {
     if (t1.equals(t2)) {
       throw new IllegalArgumentException("Cannot connect a territory to itself: " + t1);
@@ -85,21 +86,28 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
   }
 
   /**
-   * Returns the territory with the given name, or null if no territory can be found (case
-   * sensitive).
+   * Case-sensitive search for {@link Territory} by name. *
    *
-   * @param s name of the searched territory (case sensitive)
+   * @param territoryName name of the searched territory (case-sensitive)
+   * @return Territory with the given name or {@code null} if no territory can be found.
    */
-  public @Nullable Territory getTerritory(final String s) {
-    return territoryLookup.get(s);
+  public @Nullable Territory getTerritoryOrNull(final String territoryName) {
+    return territoryLookup.get(territoryName);
   }
 
-  public Territory getTerritoryOrThrow(final String s) {
-    return Optional.ofNullable(territoryLookup.get(s))
+  /**
+   * Case-sensitive search for {@link Territory} by name.
+   *
+   * @param territoryName name of the searched territory (case-sensitive)
+   * @return Territory with the given name or {@link Territory} if no territory can be found.
+   */
+  public Territory getTerritoryOrThrow(final String territoryName) {
+    return Optional.ofNullable(getTerritoryOrNull(territoryName))
         .orElseThrow(
             () ->
                 new IllegalArgumentException(
-                    MessageFormat.format("Territory with name {0} could not be found", s)));
+                    MessageFormat.format(
+                        "Territory with name {0} could not be found", territoryName)));
   }
 
   /**
@@ -382,7 +390,7 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
    * @param t1 start territory of the route
    * @param t2 end territory of the route
    */
-  public int getDistance(final Territory t1, final Territory t2) {
+  public int getDistance(@NotNull final Territory t1, @NotNull final Territory t2) {
     return getDistance(t1, t2, it -> true);
   }
 
@@ -394,7 +402,8 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
    * @param t2 end territory of the route
    * @param cond condition that covered territories of the route must match
    */
-  public int getDistance(final Territory t1, final Territory t2, final Predicate<Territory> cond) {
+  public int getDistance(
+      @NotNull final Territory t1, @NotNull final Territory t2, final Predicate<Territory> cond) {
     return getDistance(t1, t2, (it, it2) -> cond.test(it2));
   }
 
@@ -407,7 +416,9 @@ public class GameMap extends GameDataComponent implements Iterable<Territory> {
    * @param routeCond condition that neighboring territories along route must match.
    */
   public int getDistance(
-      final Territory t1, final Territory t2, final BiPredicate<Territory, Territory> routeCond) {
+      @NotNull final Territory t1,
+      @NotNull final Territory t2,
+      final BiPredicate<Territory, Territory> routeCond) {
     if (t1.equals(t2)) {
       return 0;
     }
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
@@ -309,7 +309,7 @@ public final class GameParser {
   }
 
   private Territory getTerritory(final String name) throws GameParseException {
-    return Optional.ofNullable(data.getMap().getTerritory(name))
+    return Optional.ofNullable(data.getMap().getTerritoryOrNull(name))
         .orElseThrow(
             () ->
                 new GameParseException(
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
@@ -309,7 +309,8 @@ public final class GameParser {
   }
 
   private Territory getTerritory(final String name) throws GameParseException {
-    return Optional.ofNullable(data.getMap().getTerritoryOrNull(name))
+    return data.getMap()
+        .getTerritory(name)
         .orElseThrow(
             () ->
                 new GameParseException(
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
@@ -10,6 +10,7 @@ import java.util.Set;
 import java.util.function.BiPredicate;
 import java.util.function.Predicate;
 import lombok.Getter;
+import org.jetbrains.annotations.NotNull;
 import org.triplea.java.ObjectUtils;
 import org.triplea.java.collections.CollectionUtils;
 
@@ -45,10 +46,10 @@ public final class BreadthFirstSearch {
    * caller before running the BreadthFirstSearch.
    */
   public static class TerritoryFinder implements Visitor {
-    final Territory destination;
+    @NotNull final Territory destination;
     @Getter int distanceFound = -1;
 
-    public TerritoryFinder(Territory destination) {
+    public TerritoryFinder(@NotNull Territory destination) {
       this.destination = destination;
     }
 
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
@@ -50,6 +50,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Predicate;
+import org.jetbrains.annotations.NotNull;
 import org.triplea.java.collections.CollectionUtils;
 import org.triplea.java.collections.IntegerMap;
 
@@ -58,7 +59,7 @@ class ProPurchaseAi {
 
   private final ProOddsCalculator calc;
   private final ProData proData;
-  private GameData data;
+  private GameData proGameData;
   private GameState startOfTurnData; // Used to count current units on map for maxBuiltPerPlayer
   private GamePlayer player;
   private ProResourceTracker resourceTracker;
@@ -79,7 +80,7 @@ class ProPurchaseAi {
     ProLogger.info("Repairing factories with PUsRemaining=" + pusRemaining);
 
     // Current data at the start of combat move
-    this.data = data;
+    this.proGameData = data;
     this.player = player;
     final Predicate<Unit> ourFactories =
         Matches.unitIsOwnedBy(player)
@@ -146,10 +147,10 @@ class ProPurchaseAi {
   Map<Territory, ProPurchaseTerritory> bid(
       final int pus, final IPurchaseDelegate purchaseDelegate, final GameState startOfTurnData) {
     // Current data fields
-    data = proData.getData();
+    proGameData = proData.getData();
     this.startOfTurnData = startOfTurnData;
     player = proData.getPlayer();
-    resourceTracker = new ProResourceTracker(pus, data);
+    resourceTracker = new ProResourceTracker(pus, proGameData);
     territoryManager = new ProTerritoryManager(calc, proData);
     isBid = true;
     final ProPurchaseOptionMap purchaseOptions = proData.getPurchaseOptions();
@@ -260,7 +261,7 @@ class ProPurchaseAi {
   Map<Territory, ProPurchaseTerritory> purchase(
       final IPurchaseDelegate purchaseDelegate, final GameState startOfTurnData) {
     // Current data fields
-    data = proData.getData();
+    proGameData = proData.getData();
     this.startOfTurnData = startOfTurnData;
     player = proData.getPlayer();
     resourceTracker = new ProResourceTracker(player);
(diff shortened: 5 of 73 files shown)


game-app/game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
```diff
@@ -20,6 +20,7 @@
 import lombok.Getter;
 import lombok.Setter;
 import org.jetbrains.annotations.NonNls;
+import org.jetbrains.annotations.NotNull;
```
DanVanAtta, line 23: There were prior efforts to consolidate the "not null" annotation used. There's a javax variant that requires no extra libraries. Please use:  `import javax.annotation.Nonnull;`

game-app/game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
```diff
@@ -173,9 +174,25 @@ public final boolean equals(final Object obj) {
         && (Objects.equals(name, other.name) || this.toString().equals(other.toString()));
   }
 
-  protected Territory getTerritoryOrThrow(String name) throws GameParseException {
-    return Optional.ofNullable(getData().getMap().getTerritory(name))
-        .orElseThrow(() -> new GameParseException("No territory named: " + name + thisErrorMsg()));
+  @NotNull
```
DanVanAtta, line 177: You can remove @NotNull on methods. By default it's assumed methods will not return null and therefore you don't need `@NotNull`. The codebase violates that assumption all over the place, but at the same time the `@NotNull` does not actually do much whether the assumption holds or is violated.

game-app/game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
```diff
@@ -173,9 +174,25 @@ public final boolean equals(final Object obj) {
...
+      throws GameParseException {
+    return getDataOrThrow()
+        .getMap()
+        .getTerritory(territoryName)
+        .orElseThrow(
+            () -> new GameParseException("No territory named: " + territoryName + thisErrorMsg()));
+  }
+
+  @NotNull
+  protected Territory getTerritoryOrThrowIllegalStateException(@Nullable String territoryName)
```
DanVanAtta, line 188: `@Nullable` parameter is a bit sus. 
The "rule" is "never pass null, never return null". So, we should avoid calling `getTerritory` if we know `territoryName` is null. By that token, we really should not call `getTerritoryOrThrowIllegalStateException` if territoryName is null. Can we move any null checking to the caller?

DanVanAtta: Second comment, adding methods to throw specific exception types is dirty. This looks to be a case where `Optional` should be returned and then the caller deals with that. Or, if there is context that the caller can't get, then throw a specific exception type and have the caller catch that and then convert the exception type to what they need.

The issue is we have low level code that is trying to do everything, it's trying to do the correct error handling for the code that is at higher layers. This code does not have enough information to know what the correct exception type would be, which means it's the caller that will need to do some exception handling.

game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java
```diff
@@ -85,21 +86,31 @@ private void setConnection(final Territory from, final Territory to) {
...
+   * @param territoryName name of the searched territory (case-sensitive)
+   * @return Territory with the given name or {@link Territory} if no territory can be found.
+   */
+  public Territory getTerritoryOrThrow(final @NotNull String territoryName) {
+    final Territory t = territoryLookup.get(territoryName);
+    if (t == null) {
+      throw new IllegalArgumentException(
+          MessageFormat.format("Territory with name {0} could not be found", territoryName));
+    }
+    return t;
```
DanVanAtta, line 113: Slightly simpler:
```
return getTerritory(territoryName)
                 .orElseThrow(() -> new IllegalArgumentException(
                            MessageFormat.format("Territory with name {0} could not be found", territoryName));
```            
Side note, we use `String.format` more frequently. I somewhat think we should prefer consistency and avoid using `MessageFormat` unless we specifically need something in `MessageFormat` that String.format does not provide.


*edit* was missing the `() -> new IllegalArgumentException(` in the example, (second line)

game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java
```diff
@@ -394,7 +405,8 @@ public int getDistance(final Territory t1, final Territory t2) {
    * @param t2 end territory of the route
    * @param cond condition that covered territories of the route must match
    */
-  public int getDistance(final Territory t1, final Territory t2, final Predicate<Territory> cond) {
+  public int getDistance(
+      @NotNull final Territory t1, @NotNull final Territory t2, final Predicate<Territory> cond) {
```
DanVanAtta, line 409: Let's stop decorating method parameters so much. I don't believe the `final` and `@NotNull` help.
The nullity annotations are largely for an IDE to help us. IDE's are smart enough to infer a `@NotNull` constraint by how variables are used. So, you're getting the `NotNull` annotation for free without explicitly declaring it. Meanwhile there is a drawback where  `@NotNull final` is adding noise.

game-app/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
```diff
@@ -45,10 +46,10 @@ public interface Visitor {
    * caller before running the BreadthFirstSearch.
    */
   public static class TerritoryFinder implements Visitor {
-    final Territory destination;
+    @NotNull final Territory destination;
     @Getter int distanceFound = -1;
 
-    public TerritoryFinder(Territory destination) {
+    public TerritoryFinder(@NotNull Territory destination) {
       this.destination = destination;
```
DanVanAtta, line 53: Adding `@NotNull` does not help as much as failing "early" would. An IDE warning that is unlikely to trigger is less helpful than asserting that destination must be non null at the time of creation. EG: `this.destination = Preconditions.checkNotNull(destination)`

The scope of this PR is massive.. This file simply can be reverted. Testing the change to assert not null parameter is not trivial.

game-app/game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
```diff
@@ -115,20 +117,6 @@ public class MapData {
...
-  @NonNls private static final String MAP_PROPERTIES = "map.properties";
-  @NonNls private static final String CAPITAL_MARKERS = "capitols.txt";
-  @NonNls private static final String CONVOY_MARKERS = "convoy.txt";
-  @NonNls private static final String COMMENT_MARKERS = "comments.txt";
-  @NonNls private static final String VC_MARKERS = "vc.txt";
-  @NonNls private static final String BLOCKADE_MARKERS = "blockade.txt";
-  @NonNls private static final String PU_PLACE_FILE = "pu_place.txt";
-  @NonNls private static final String TERRITORY_NAME_PLACE_FILE = "name_place.txt";
-  @NonNls private static final String KAMIKAZE_FILE = "kamikaze_place.txt";
-  @NonNls private static final String DECORATIONS_FILE = "decorations.txt";
```
DanVanAtta, line 130: Let's keep these constants here. `Constants` is an evil class. It's completely generic, has no focus. We may as well put every single constant in the whole application there, right? That's why `Constants` is evil, it kinda just begets procedural code and and reduces code cohesion (AKA locality)

game-app/game-core/src/main/java/games/strategy/triplea/ui/UnitChooser.java
```diff
@@ -755,12 +756,12 @@ private void loadNonWithdrawableImage() {
      *     getNonWithdrawableImageHeight</code>.
      */
     private BufferedImage loadNonWithdrawableImage(final int nonWithdrawableImageHeight) {
-      final BufferedImage imgSmall = loadImage("non-withdrawable_small.png");
+      final BufferedImage imgSmall = loadImage(Constants.FILE_NAME_IMAGE_NON_WITHDRAWABLE_SMALL);
```
DanVanAtta, line 759: *edit* for brevity:

Concern: extracting to a single `Constants.java` is fixing the hard-coded problem too much in the "other" direction. Hard-coding that is not repeated is often easier to manage than a very coupled, and very abstracted system. 

For consideration, we can take an intermediate step, simply extracting the constant to a local variable, or a class constant, or a package scoped class that contains constants..

I think I'd rather see this value hard-coded here rather than moved to `Constants`. Ideally this value would go somewhere that is scoped so that only the `ui` layer could see the value. In `Constants`, the scoping will allow the entire game engine to see the value, when it is only a concern of the UI layer..

game-app/game-core/src/main/java/tools/image/DecorationPlacer.java
```diff
@@ -122,6 +123,130 @@ public static void run() {
     }
   }
 
+  /**
+   * Helper method to load the centers.txt file. It searches for the file inside of {@code
+   * mapFolderLocation}. If no file exists there it tries to search inside a directory specified by
```
DanVanAtta, line 128: Diff here is a bit too much to review with everything else. Can we get this update on its own PR?

game-app/game-core/src/main/java/tools/image/DecorationPlacer.java
```diff
@@ -122,6 +123,130 @@ public static void run() {
...
+      }
+    } else {
+      final Optional<Map<String, List<Polygon>>> optionalPolygons =
+          selectAndLoadPolygonsFile(mapFolderLocation);
+      if (optionalPolygons.isPresent()) {
+        return optionalPolygons.get();
+      } else {
+        log.info("You must specify a Polgyon file.");
+        log.info("Shutting down.");
+        throw new IOException("no polygons file specified");
```
DanVanAtta, line 222: Canonical `Optional` usage is essentially to never assign to an `Optional` variable. In most all cases one should "get rid" of the optional.

In this case the entire code block is reduced to:
```
return 
          selectAndLoadPolygonsFile(mapFolderLocation).orElseThrow(() -> 
                 throw new IOException("no polygons file specified"));
```        
Grant it, the `IOException` type is inappropriate, it is beyond the point - the point is the 7 line method here can be just two.

game-app/game-core/src/main/java/tools/image/DecorationPlacer.java
```diff
@@ -956,7 +1013,7 @@ enum ImagePointType {
             + "CTRL/SHIFT + Right Click = delete currently selected image point</html>"),
 
     blockade(
-        "blockade.txt",
+        Constants.FILE_NAME_BLOCKADE_MARKERS,
```
DanVanAtta, line 1016: I suspect things like `blockade.txt` is used in multiple places and is hardcoded. By that token, extracting to a constant is a good thing. `Constants` is not a great place for it. The `tools` package is almost its own application. We should avoid creating any dependency between the `tools` and the rest of the source code as much as we can.

game-app/game-core/src/main/java/games/strategy/engine/data/gameparser/GameParser.java
```diff
@@ -309,7 +309,8 @@ private RepairRule getRepairRule(final String name) throws GameParseException {
   }
 
   private Territory getTerritory(final String name) throws GameParseException {
-    return Optional.ofNullable(data.getMap().getTerritory(name))
+    return data.getMap()
+        .getTerritory(name)
```
DanVanAtta, line 313: :+1: 

This type of change is a good sign.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
```diff
@@ -58,7 +59,7 @@ class ProPurchaseAi {
 
   private final ProOddsCalculator calc;
   private final ProData proData;
-  private GameData data;
+  private GameData proGameData;
```
DanVanAtta, line 62: Why rename this?

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPlaceTerritory.java
```diff
@@ -7,21 +7,22 @@
...
 import lombok.Setter;
+import org.jetbrains.annotations.NotNull;
 
 /** The result of an AI placement analysis for a single territory. */
 @EqualsAndHashCode(of = "territory")
 @Getter
 @Setter
 public class ProPlaceTerritory {
-  private final Territory territory;
+  private final @NotNull Territory territory;
```
DanVanAtta, line 17: @Nullable is the only annotation that we should really be using.

Either for parameters that can be null, or for class variables. `NotNull` is okay on class variables if we want to really emphasize which variables are nulllabe and which are not. 

The current usage looks very inconsistent and the IDE already is inferring NotNull annotations, so it's unnecessary.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/simulate/ProSimulateTurnUtils.java
```diff
@@ -94,12 +94,8 @@ public static void simulateBattles(
         }
         battleDelegate.getBattleTracker().getConquered().add(t);
         battleDelegate.getBattleTracker().removeBattle(battle, data);
-        final Territory updatedTerritory = data.getMap().getTerritory(t.getName());
-        ProLogger.debug(
-            "after changes owner="
-                + updatedTerritory.getOwner()
-                + ", units="
-                + updatedTerritory.getUnits());
+        // note that the Territory object has already changed
```
DanVanAtta, line 97: This comment could use a bit more context. What changed the territory object and how has it changed??

frigoref: If you just read this line, but if you also read the next one I think it becomes clear, don't you?
```java
ProLogger.debug("after changes owner=" + t.getOwner() + ", units=" + t.getUnits());
```

game-app/game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
frigoref: Adjusted in new partial PR https://github.com/triplea-game/triplea/pull/13584

game-app/game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
frigoref: Adjusted in new partial PR #13584. It was used to differentiate it from method `getTerritoryOrNull`.

game-app/game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
frigoref: This is a helper method to make more clear by name what it does. The usage has not been affected and whether it is used with `null` or not can be addressed afterwards, no?
The current change is about to get rid of returning null on `getTerritory`

game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java
frigoref: k, will provide another PR at some point to replace all the Message Format parts I have done so far (that are quite a few).

game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java
frigoref: I do not see that the IDE is making this automatically and as apparently the Territory can be passed sometimes as `null` I think it is better at the moment to state clearly when this is not the case.
We can remove it later, but at the moment I always have to check whether it might be having value `null` or not.
Concerning annotation `final` I would prefer the IDE to work the other way around with method parameters, i.e., only having to declare when they are not final, but sadly it is not that way and therefore, I would vote to keep it for clarity purposes.

game-app/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
frigoref: Also here, as apparently the Territory can be passed sometimes as `null` I think it is better at the moment to state clearly when this is not the case.
We can remove it later, but at the moment I always have to check whether it might be having value null or not.

game-app/game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
frigoref: I agree and tried to find a different class, but maybe that kind of reorganizing the constants is a different task. When we collect all in here for the moment, we at least have an overview of "what kind" of constants we have and that can help to find the best spot for them (new class or existing class).

game-app/game-core/src/main/java/tools/image/DecorationPlacer.java
frigoref: When extracting I tried to extract the same text from all files in the project and replace it with the constant.
I also thought about package dependency and there might be a need of a "common" package that carries for example shared constants and the multiple packages that might use these constants. What do you think about that?

game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java
DanVanAtta: I had the same discussion with other devs within TripleA regarding adding `@NotNull`, the conclusion was to not add it. 

That you don't see the annotation being added, it is added when the IDE is doing nullity analysis. It doesn't physically add it in anything you would see.

> I think it is better at the moment to state clearly when this is not the case.

And what's the chance it gets out of date? It's still not clear.

Second, we still pay costs for having the noise. A param list that could fit on one line is instead 50% keywords. Furhter, why not add `@NotNull` to `cond`? In this example, we have raised additional questions there. Given there can't be certainty that even `@NotNull` is correct, we've only added questions.

> "We can remove it later, but at the moment I always have to check whether it might be having value null or no"

If adding the annotation is a waste of time to begin with, it's not compelling to suggest that we can simply remove it later.

game-app/game-core/src/main/java/games/strategy/engine/data/util/BreadthFirstSearch.java
DanVanAtta: I recommend you submit a PR to the java-code-convention page and we can bike-shed this discussion there. 

Recommend as well that you try and search for some of the first usages of `javax...NotNull`, and perhaps you can find the lengthy conversation that happened regarding which annotation to prefer.

What's more reason to not add `NotNull`, there are several to choose from. The one chosen here would cause issues. 

I'd also consider PR scoping as well. This PR is about fixing up nullity of a single method, why are we sprinkling in changes elsewhere in the code? Where is the clean PR description that explains the rationale behind which methods have been updated? What's your plan for actually @NotNull annotations consistent? Is your vision that every return type and every parameter be annotated with either `@NotNull` or `@Nullable`?

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/simulate/ProSimulateTurnUtils.java
DanVanAtta: The comment is good for a commit, it makes sense when looking at the removal code next to the code added.

I can see that this line: ` final Territory updatedTerritory = data.getMap().getTerritory(t.getName());` is not needed thanks to the comment. A future maintainer of the code is not going to see this diff, the comment won't make sense to them. That's why I said the comment could use more context. The fact we only log a few things kinda indicates that something has changed. It's therefore confusing why there is a call-out that the territory object has changed, it seems obvious already.

In honesty, the comment ought to be part of a commit message and this update its own PR.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
frigoref: It was done for clarity in cases of where-used searches where the context of the "data" is not clear and to better distinguish it from `ProData proData`.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPlaceTerritory.java
frigoref: This is as stated in an earlier post just for the attributes/parameters of class `Territory` that are not yet cleaned up to always be NotNull. As soon as this is resolved a very simple regex search and replace and remove them again.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/simulate/ProSimulateTurnUtils.java
frigoref: Not sure exactly what is this conversion is aiming at anymore. We started with 

> This comment could use a bit more context. What changed the territory object and how has it changed??

To me my response answers this questions and I do not see what you still see as open. Wouldn't it be better you provide a comment that you see fit to speed up things?

game-app/game-core/src/main/java/tools/image/DecorationPlacer.java
DanVanAtta: A single constants class is likely to create a lot of coupling.

More ideally we could modularize and completely hide these constants. Specifically, today `triplea-game/game-app/map-data` is the only thing that does not depend on `triplea-game/game-app/game-core`. If, for example, we moved this tools utility into its own module that did not depend on `game-core`, we could keep more constants scoped to just within the module. This ideal solution is more difficult, but for sure something that is desired. I really believe TripleA code is best cleaned up by creating a number of "low layer" modules that site *below* game-core. 

Short of that goal though, instead of putting constants into a single top level package, we could also create multiple constant files. If we try to keep the constants as "close" as possible to the files that use it, that would be a relatively good step in the right direction. Though.. in some cases, some files should be the proper owner of a constant and everything else should reference it. If we collapse all common usages together, we might lose that ownership criteria. 

To explain what I mean by "proper owner" - For example, maybe there is a value that should be "owned" by `GameData`.  In this case the value should not be in `Constants.java`. When used, it should not be  `Constants.VALUE`, but it properly should be `GameData.VALUE`.  Hence, whether something is even extracted to a constant depends on usage and the concept of "ownership", how we want to achieve data-hiding (AKA encapsulation).

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/simulate/ProSimulateTurnUtils.java
DanVanAtta: This is what I see as a reviewer:

Screenshot from 2025-09-12 18-42-03
Screenshot from 2025-09-12 18-42-52

I agree that it all makes sense. I can see variable `updatedTerritory`, I can see the variable is removed, I can see the comment explaining why the variable is removed. It makes sense.


--------------


OTOH, after we merge this, this block will become:

Screenshot from 2025-09-12 18-44-25

Now, AFAIK, the comment no longer makes sense. The comment refers to `updatedTerritory`, but that was deleted. So, the comment is referring to deleted code... Otherwise, there's a comment on a `ProLogger.debug` statement, which is kind of an "interesting" in its own right.. 

I still recommend you just strike this comment (or if there is something really subtle, to call *that* out explicitly in the code comment)

game-app/game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
DanVanAtta: I pushed a branch to you with this pattern all fixed up.. Please check that for a pretty working example of one way to fix this.

A classic "java" fix would be to have this lower layer throw a single exception type.  Another approach would be to use the one method that returns an `Optional` and to have the caller do something like: `getTerritory(name).orElseThrow(() -> new IllegalArgumentException`...

The proposed update is very similar to the following (notice how the parameter signature allows the caller of this method to tell this method what to throw):

```
protected Territory getTerritoryOrThrowIllegalStateException(@Nullable String territoryName, Function<String, RuntimeException> exceptionToThrowFactory)
 
  final Territory t = territoryLookup.get(territoryName);
  if (t == null) {
       throw exceptionToThrowFactor.apply(
            MessageFormat.format("Territory with name {0} could not be found", territoryName));
      }
  return t;
}
```

game-app/game-core/src/main/java/games/strategy/engine/data/GameMap.java
DanVanAtta: > k, will provide another PR at some point to replace all the Message Format parts I have done so far (that are quite a few).

The consistency between `MessageFormat` and `String.format` is not a huge deal IMO. Assuming you haven't made changes yet. If you haven't already changed things, it's fine to leave things as they are IMO.

DanVanAtta:
```suggestion
       return getTerritory(territoryName)
                 .orElseThrow(() -> new IllegalArgumentException(
                            MessageFormat.format("Territory with name {0} could not be found", territoryName));
```

This is the main suggestion. The above (needs formatting), but will help re-use existing code. 
Though, this point could be moot. If `getTerritory` is public, then a caller can do this type of code with just a couple lines, and attach the right error message & right error type as well.

game-app/game-core/src/main/java/games/strategy/triplea/ui/mapdata/MapData.java
DanVanAtta: I hear you. Ironically, I also think that the current organization is not that horrible. For this file at least. Merging these constants would go from 'some organization' to less.. I also am not sure if we will have an awesome time splitting constants apart. I do not know if that will happen easily.

OTOH, if these constants seem very arbitrarily placed here, and that the scoping is way over the place - then yes - constansts is good. I think the best way to answer this question is essentially via looking at the required 'import' statements and the average "distance" the constants are to where they are used. In short, I think the right metric is cohesion, things that are related are physically close, and things that are unrelated are physically distant. 

So, any glaring problems, are worth fixing. Short of that, the best way to fix this issue IMO is to carve off "sub-modules". In this case, that would be something like a `game-map` module. For example, picture a `game-map` module in the following:

Screenshot from 2025-09-12 19-34-14


The current `map-data` is meant to be a very XML focused view of the world. There is no real logical interpretation, it's largely an IO layer. `game-map` would be a layer that understands what a TripleA map is. 

At this point, I see there is a missing module between `game-core` and `map-data` that should represent a much richer idea of what a `map` is. `map-data` knows XML, but it does not know how anything is related. It could be the case that some of these constants and functionality would more properly belong in `map-data`. Where-ever the line between `game-map` and `map-data` is drawn, the constants in this file would have no visibility outside of those two packages. Which is to say `game-core` would no longer have any need or any way to know about `convoy.txt`, it wouldn't need to - it can just use `game-map` instead.

That's a long answer to your question. In short, don't worry about, in medium term, common denominator to group things together is good, long term, re-drawing the lines in the system so that there are things "below"  `game-core` (thereby reducing scope, creating building blocks rather than an intertangled monolith)

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/ProPurchaseAi.java
DanVanAtta: sounds good :+1:

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/data/ProPlaceTerritory.java
DanVanAtta: Okay, sounds like you are on top of it. Specifically to _not_ have `@NotNull` on method parameters nor return values for the long term.

game-app/game-core/src/main/java/games/strategy/triplea/ai/pro/simulate/ProSimulateTurnUtils.java
DanVanAtta:
```suggestion
```

game-app/game-core/src/main/java/tools/image/DecorationPlacer.java
DanVanAtta:
```suggestion
return 
          selectAndLoadPolygonsFile(mapFolderLocation).orElseThrow(() -> 
                 throw new IOException("no polygons file specified"));
```

PR #13592, opened 2025-07-31
Update BattleDisplay.java - Fix 13591

frigoref:
Potential NPE fix for issue https://github.com/triplea-game/triplea/issues/13591


DanVanAtta:
Any idea why/how the value is null? 
Would you mind testing this update by hardcoding a null value and then testing this branch? What happens if the value and now the code continues instead?

PR #13598, opened 2025-08-05
NotNullRemoved

frigoref:
Change imports for org.jetbrains.annotations.NotNull to javax.annotation.Nonnull

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/TechnologyPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/TechnologyPanel.java
@@ -20,7 +20,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.swing.ImageIcon;
 import javax.swing.JComponent;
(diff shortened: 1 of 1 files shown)


DanVanAtta:
I think you perhaps could have just removed a number of these annotations. The expectation is methods will return a non-null value. If they do not, they should be marked as `@Nullable` (which also makes them a candidate for further fixes to return a non-null object). 

It would be *super* valuable to annotate any & all methods that return nullable. IDE linters would likely find a *lot* of potential NPEs that we would need to fix.

In short, 'non-null' is the expectation, 'nullable' is the exception and is the thing that should be annotated. Unfortunately TripleA is written well before strong conventions were commonly set around nullity.

frigoref:
> I think you perhaps could have just removed a number of these annotations. The expectation is methods will return a non-null value. If they do not, they should be marked as `@Nullable` (which also makes them a candidate for further fixes to return a non-null object).
> 
> It would be _super_ valuable to annotate any & all methods that return nullable. IDE linters would likely find a _lot_ of potential NPEs that we would need to fix.
> 
> In short, 'non-null' is the expectation, 'nullable' is the exception and is the thing that should be annotated. Unfortunately TripleA is written well before strong conventions were commonly set around nullity.

I do agree, but as we do not yet have resolved the issue of methods returning `null`, I'd suggest we first tackle this (e.g. via the noReturnNull commits/PRs) before "cleaning" the `Nonnull` annotations from the method returns.

PR #13599, opened 2025-08-05
MessageFormat removal

frigoref:
Most cases handled via regex's
MessageFormat(.format\(.*\".*)\{0\}(.*\", \w+\))
and MessageFormat(.format\(.*\n.*\".*)\{0\}(.*\")
to
String$1%s$2

**i18n**
*.properties changed via regex
(\=.*)\{0\}(.*)
to
$1%s$2


frigoref:
@DanVanAtta If we also remove `MessageFormat` from the `ResourceBundle`, we might get into problems with translations as the the sorting order of parameters can be defined with `MessageFormat` independent of the parameter order in the code.
Even though we can be more expressive for formatting using `String`, `MessageFormat` automatically handles types like dates and numbers with locale-awareness that might be nice to.
If you agree, I would got the other way and replace `String.format` with `MessageFormat.format` instead.

game-app/game-core/src/main/resources/i18n/games/strategy/engine/framework/ui_en.properties
```diff
@@ -7,7 +7,7 @@ startup.MainPanel.btn.Play.Tltp.line3=Play Online, or Local Game, or PBEM, or Ho
 startup.MainPanel.btn.Quit.Lbl=Quit
 startup.MainPanel.btn.Quit.Tltp=Close TripleA.
 startup.PbfSetupPanel.tab.ForumPosterEditor.Lbl=Play By Forum
-startup.PlayerSelectorRow.btn.Alliances.Tltp=Click to play {0}
+startup.PlayerSelectorRow.btn.Alliances.Tltp=Click to play %s
```
DanVanAtta, line 10: AFAIK, internationalization is actually a place where `MessageFormat` has a difference. Are we sure this does not break any of the NLS work?

DanVanAtta:
> If we also remove MessageFormat from the ResourceBundle, we might get into problems with translations as the the sorting order of parameters can be defined with MessageFormat independent of the parameter order in the code. Even though we can be more expressive for formatting using String, MessageFormat automatically handles types like dates and numbers with locale-awareness that might be nice to.

I just read this message. My 2 cents is `MessageFormat` should be used with `ResourceBundle` and things NLS because it makes a difference. For me, the usage of `MessageFormat` signals a context where NLS matters. 

Perhaps it might be the case that we get good enough with the NLS that we have majority usage of `MessageFormat`. Until then, the project norm is String.format when it is a choice (the norm is based purely on the current usage counts, just counting how many times one is used over the other, String.format is a majority)

frigoref:
@DanVanAtta would be nice to keep it open until we are aligned.

DanVanAtta:
@frigoref make sure you test this well. It's not feasible to accurately review 2000 lines of changed code, I'm not going to attempt it. Please own this update and make sure there are no errors. If you feel very confident, feel free to merge when you are ready.

PR #13614, opened 2025-08-10
No return null (get territory)#1.1

frigoref:
Test.java

usage of method getTerritoryOrThrow instead of getTerritory
DefaultAttachment.java

method getTerritoryOrThrow: replaces with getTerritoryOrThrowGameParseException
new methods getTerritoryOrThrowGameParseException/getTerritoryOrThrowIllegalStateException: Throwing specific exception when no territory was found
new method getTerritory: private method returning Optional to avoid returning null
GameMap.java

method getTerritory: Renamed to getTerritoryOrNull
method getTerritoryOrThrow: Documentation added
methods getDistance: Parameter annotation NotNull added
GameSelectorModel.java

method parseAndValidate: Return Optional instead of null
method setGameData: Parameter annotation Nullable added
GameTestUtils.java

method getTerritory: Replaced with getTerritoryOrThrow
UnitTypeList.java

method getUnitTypes: Documentation updated
Notes to Reviewer
Follow-up PRs shall get to a state that getTerritory is no longer returning null.
Replacement of PR #13584.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
@@ -20,6 +20,7 @@ import javax.annotation.Nullable;
 import lombok.Getter;
 import lombok.Setter;
 import org.jetbrains.annotations.NonNls;
+import org.jetbrains.annotations.NotNull;
 import org.triplea.java.collections.IntegerMap;
 
 /**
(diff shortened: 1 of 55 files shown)


game-app/game-core/src/main/java/games/strategy/engine/data/DefaultAttachment.java
```diff
@@ -173,9 +174,25 @@ public final boolean equals(final Object obj) {
...
+            () -> new GameParseException("No territory named: " + territoryName + thisErrorMsg()));
+  }
+
+  @Nonnull
+  protected Territory getTerritoryOrThrowIllegalStateException(@Nullable String territoryName)
+      throws IllegalStateException {
+    return getTerritory(territoryName)
+        .orElseThrow(
+            () ->
+                new IllegalStateException("No territory named: " + territoryName + thisErrorMsg()));
```
DanVanAtta, line 195: Same issue here as the other PR that is already reviewed. I don't know if it was cleanly stated... I sent you a branch with suggestions on how to fix this.

Okay.. so these two methods `getTerritoryOrThrowGameParseException` and `getTerritoryOrThrowIllegalStateException` are really showing inappropriate error handling. The two methods are there for the sake of the caller being able to call the right one for the right context to the error handling. The caller knows the context of the error, the caller should be the one attaching the right exception type.

A very old school Java approach would be to have these methods be unified and throw a `TerritoryNotFoundException`. The caller would in turn catch that and throws the right one. If the caller can instead grab all needed data for an error message, then it's considerably cleaner to return an empty optional and have the caller do the error handling. 

Having checked exceptions is very weird and makes coding difficult.. I recommend going with the route in the branch I sent your way..

frigoref:
Most content split up in PRs

- #13703
- #13704

Issue #13608, opened 2025-08-08
History#gotoNode:122 - java.lang.IllegalStateException


tripleabuilderbot:
## User Description
Hello,

I am trying to export Setup charts from a in TripleA, version 2.7+15134 for MacOS to my Desktop.
Unfortunately I only receive the following message: "An unexpected error occurred! IllegalStateException".

This doesn't help in any way to fix my problem as I cannot determine where from it derives.
So please help me.

With kind regards,
A.

## Map
world_war_i_1914_improved / World War I 1914 Improved

## TripleA Version
[2.7.15134](https://github.com/triplea-game/triplea/releases/tag/2.7.15134)

## Java Version
21.0.7

## Operating System
Mac OS X

## Heap Size
8192M

## Stack Trace
```
Exception: java.lang.IllegalStateException 
java.lang.Exception
	at com.google.common.base.Preconditions.checkState(Preconditions.java:497)
	at games.strategy.engine.history.History.gotoNode(History.java:122)
	at games.strategy.triplea.printgenerator.InitialSetup.run(InitialSetup.java:19)
	at games.strategy.triplea.printgenerator.SetupFrame.lambda$new$1(SetupFrame.java:58)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1972)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2314)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:407)
	at java.desktop/javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:262)
	at java.desktop/javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:279)
	at java.desktop/java.awt.Component.processMouseEvent(Component.java:6621)
	at java.desktop/javax.swing.JComponent.processMouseEvent(JComponent.java:3398)
	at java.desktop/java.awt.Component.processEvent(Component.java:6386)
	at java.desktop/java.awt.Container.processEvent(Container.java:2266)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:4996)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2324)
	at java.desktop/java.awt.Component.dispatchEvent(Component.java:4828)
	at java.desktop/java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4948)
[...]

PR #13615, opened 2025-08-10
Fix13608 IllegalStateException(gotoNode)

frigoref:
Fix for issue #13608 by mainly
Extract StatisticsAggregator.cloneGameData to GameDataUtils.java as GameDataUtils.cloneGameDataWithHistory including the option to enableSeeking

ExportMenu.java / StatisticsAggregator.java
- Use new method with enableSeeking = true

TripleAFrame.java
- method showHistory: Use new method with enableSeeking = false as it is called later if needed 

## Notes to Reviewer
It is a bit odd that the field `History.seekingEnabled` is checked and can cause dumps if it has the wrong value. I didn't understand why exactly, so I kept it, but maybe we should change this to call method `enableSeeking` everytime this flag should be `true` in the future.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/framework/GameDataUtils.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/framework/GameDataUtils.java
@@ -71,7 +71,8 @@ public final class GameDataUtils {
     }
   }
 
-  public static Optional<GameData> cloneGameDataWithHistory(GameData gameData, boolean enableSeeking) {
+  public static Optional<GameData> cloneGameDataWithHistory(
+      GameData gameData, boolean enableSeeking) {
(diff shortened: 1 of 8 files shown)


DanVanAtta:
> It is a bit odd that the field History.seekingEnabled is checked and can cause dumps if it has the wrong value. I didn't understand why exactly, so I kept it, but maybe we should change this to call method enableSeeking everytime this flag should be true in the future.

To actually fix this problem, I think we need to dig in here. Otherwise seemingly we are adding patching around something that is not understood. If someone else actually wants to try and fix this, the patching I believe will add yet more confusion.

frigoref:
Merged in separate commits starting with commit https://github.com/frigoref/triplea/commit/6fc10adabe1b1863ed38b6f2215a97396a7edaa8

PR #13616, opened 2025-08-10
nullableMethods_strategyEngine

frigoref:
Add to method annotation Nullable in case there is return null; statement 
For package games.strategy.engine


game-app/game-core/src/main/java/games/strategy/engine/framework/ClientGame.java
```diff
@@ -218,7 +219,7 @@ public void addChange(final Change change) {
   }
 
   @Override
-  public IRandomSource getRandomSource() {
+  public @Nullable IRandomSource getRandomSource() {
     return null;
```
DanVanAtta, line 223: I would guess a "OperationNotSupportedException" might be more appropriate here.

PR #13619, opened 2025-08-13
Fix13608 IllegalStateException(gotoNode) 2

frigoref:
Similar to PR #13615, but including also a rework on the export menu.

## Notes to Reviewer
This is a major rework with several smaller commits.
In case reviewing all together is faster, I submit this PR. It can be sliced in individual PRs based on the commits if preferred.

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/engine/framework/GameDataUtils.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/framework/GameDataUtils.java
@@ -71,7 +71,8 @@ public final class GameDataUtils {
     }
   }
 
-  public static Optional<GameData> cloneGameDataWithHistory(GameData gameData, boolean enableSeeking) {
+  public static Optional<GameData> cloneGameDataWithHistory(
+      GameData gameData, boolean enableSeeking) {
     final var cloneOptions = GameDataManager.Options.builder().withHistory(true).build();
     Optional<GameData> optionalGameDataClone = cloneGameData(gameData, cloneOptions);
     if (enableSeeking) {
--- a/game-app/game-core/src/main/java/games/strategy/engine/framework/GameDataUtils.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/framework/GameDataUtils.java
@@ -70,4 +70,13 @@ public final class GameDataUtils {
       throw new RuntimeException(e);
     }
   }
+
+  public static Optional<GameData> cloneGameDataWithHistory(GameData gameData, boolean enableSeeking) {
+    final var cloneOptions = GameDataManager.Options.builder().withHistory(true).build();
+    Optional<GameData> optionalGameDataClone = cloneGameData(gameData, cloneOptions);
+    if (enableSeeking) {
+      optionalGameDataClone.ifPresent(clone -> clone.getHistory().enableSeeking(null));
+    }
+    return optionalGameDataClone;
+  }
 }
--- a/game-app/game-core/src/main/java/games/strategy/engine/stats/StatisticsAggregator.java
+++ b/game-app/game-core/src/main/java/games/strategy/engine/stats/StatisticsAggregator.java
@@ -4,7 +4,6 @@ import com.google.common.collect.HashBasedTable;
 import games.strategy.engine.data.GameData;
 import games.strategy.engine.data.GamePlayer;
 import games.strategy.engine.data.Resource;
-import games.strategy.engine.framework.GameDataManager;
 import games.strategy.engine.framework.GameDataUtils;
 import games.strategy.engine.history.HistoryNode;
 import games.strategy.engine.history.Round;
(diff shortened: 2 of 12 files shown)


lib/swing-lib/src/main/java/org/triplea/swing/FileChooser.java
```diff
@@ -25,6 +28,25 @@ public class FileChooser {
...
 
+  /**
+   * Wraps the file chooser for a given {@code defaultFileName}, infers its file extension and uses
+   * as starting directory the user directory.
+   *
+   * @param frame parent frame for the file chooser
+   * @param defaultFileName default name of the file
+   * @return path to the chosen file
+   */
+  public static Optional<Path> chooseExportFileWithDefaultName(
```
DanVanAtta, line 39: Keep in mind scopes. Swing-lib should not know anything about TripleA concepts, it should not know what an "export file" is. This method belongs closer to where it's used.

game-app/game-core/src/main/java/games/strategy/engine/framework/GameDataUtils.java
```diff
@@ -70,4 +70,14 @@ public static <T> T translateIntoOtherGameData(final T object, final GameData tr
       throw new RuntimeException(e);
     }
   }
+
+  public static Optional<GameData> cloneGameDataWithHistory(
+      GameData gameData, boolean enableSeeking) {
+    final var cloneOptions = GameDataManager.Options.builder().withHistory(true).build();
+    Optional<GameData> optionalGameDataClone = cloneGameData(gameData, cloneOptions);
```
DanVanAtta, line 77: I don't think you needed to have added another method signature to hide the "parameter type" `Options`

The existing pattern would allow for something like this:
```
GameDataManager.Options.builder().withHistory(true).withSeeking(true).build();
```

Using the `Options` parameter has some strengths. 
- The "boolean" parameter presents a poor API, a usability issue (which we do mention to avoid in the java-conventions)
- If we need a case where there is no history copy but we want seeking, we would have to do contortions to have the right set of parameters and method names.


Nit: methods that provide overloaded signatures should be grouped together. Though, this method probably should not exist, instead enhance the existing pattern and it'll be a better API and hopefully less overall code too.

game-app/game-core/src/main/java/games/strategy/engine/framework/GameDataUtils.java
```diff
@@ -70,4 +70,14 @@ public static <T> T translateIntoOtherGameData(final T object, final GameData tr
...
+    if (enableSeeking) {
+      optionalGameDataClone.ifPresent(clone -> clone.getHistory().enableSeeking(null));
```
DanVanAtta, line 79: After all that work to help with null-safety, and now we're pro-actively passing a null value! 

`enableSeeking` accepting a historyPanel for a glorified set operation makes no sense. I really have no clue what "enableSeeking" means and why the historyPanel could be null. Having `HistoryPanel` in the API doesn't make sense.

game-app/game-core/src/main/java/games/strategy/engine/stats/StatisticsAggregator.java
```diff
@@ -34,17 +33,15 @@ OverTimeStatisticType.PredefinedStatistics.UNITS, new UnitsStat(),
   private final MapData mapData;
 
   public StatisticsAggregator(GameData gameData, MapData map) {
-    this.game = cloneGameData(gameData);
+    this.game =
+        GameDataUtils.cloneGameDataWithHistory(gameData, true)
+            .orElseThrow(
+                () ->
+                    new IllegalStateException(
+                        "Cloning game data for StatisticsAggregator failed."));
```
DanVanAtta, line 41: This is not enough information to present to a user. We need to capture more. Generally the user should be made aware of:
1. what happened
2. what the impact is
3.  and what they can do about it

I think we need to look at the API for cloning game data history objects and how errors are handled in detail. 

The existing code is really bad:
` final GameData clone = GameDataUtils.cloneGameData(gameData, cloneOptions).orElse(null);`

A stack trace that has a static message in it is a slight improvement over `NullPointerExceptoin`.. but not much.

lib/swing-lib/src/main/java/org/triplea/swing/FileChooser.java
frigoref: Maybe the name is too specific for "export file" and it should refer more to the generic description in the documentation.
What do you think about `chooseFileWithDefaultNameFromUserDirectory`?

game-app/game-core/src/main/java/games/strategy/engine/framework/GameDataUtils.java
frigoref: I agree that we should try to avoid passing `null `where possible, but according to the code it is possible and therefore I do not see why this should be an issue here. Could you elaborate?
My idea would be to annotated the parameter as well as the member that is assigned with the parameter as `Nullable`.

Concerning `enableSeeking` I did not want to break anything, but I didn't fully understand its usage either. To me it seemed to be related to `nextChangeIndex` in comment before their declarations:
``` java 
  // Index at which point we are in history. Only valid if seekingEnabled is true.
  private int nextChangeIndex;
  private boolean seekingEnabled = false;
```

game-app/game-core/src/main/java/games/strategy/engine/stats/StatisticsAggregator.java
frigoref: You are right, but we cannot save the world in a big bang. A slight improvement is better than no improvement and to me IllegalStateExceptions are rarely helpful for the user, but for some developer analyzing the issue.

game-app/game-core/src/main/java/games/strategy/engine/framework/GameDataUtils.java
frigoref: That sounds good. Shall I provide another commit for this then?

PR #13680, opened 2025-09-12
fix13679 NPE historyPanel in TripleAFrame

frigoref:
- fix13679_NPEhistoryPanelTripleAFrame #&#8203;1 Annotate Nullable for TripleAFrame attributes
- fix13679_NPEhistoryPanelTripleAFrame #&#8203;2 Make popup for "Save Game at this point (BETA)" save for historyPanel being null 
- fix13679_NPEhistoryPanelTripleAFrame #&#8203;3 Make popup "Show Summary Log" and "Show Detailed Log" save for historyPanel being null
- fix13679_NPEhistoryPanelTripleAFrame #&#8203;4 Make popup "Export Gameboard Picture" save for historyPanel being null

### DIFF
--- a/game-app/game-core/src/main/java/games/strategy/triplea/ui/history/HistoryPanel.java
+++ b/game-app/game-core/src/main/java/games/strategy/triplea/ui/history/HistoryPanel.java
@@ -39,7 +39,6 @@ import javax.swing.tree.DefaultTreeCellRenderer;
 import javax.swing.tree.TreeNode;
 import javax.swing.tree.TreePath;
 import lombok.Getter;
-import lombok.Setter;
 
 /** Shows the history as a tree. */
 public class HistoryPanel extends JPanel {
@@ -48,7 +47,7 @@ public class HistoryPanel extends JPanel {
   private final JTree tree;
   private final HistoryDetailsPanel details;
   @Getter private HistoryNode currentPopupNode;
-  @Setter private JPopupMenu popup;
+  private JPopupMenu popup;
   // remember which paths were expanded
   private final Collection<TreePath> stayExpandedPaths = new ArrayList<>();
   private boolean mouseOverPanel;
@@ -80,20 +79,6 @@ public class HistoryPanel extends JPanel {
     // Register the tree with the tooltip manager to make the tooltips we set work.
     ToolTipManager.sharedInstance().registerComponent(tree);
     tree.expandRow(0);
-    tree.add(this.popup);
-    this.popup.addPopupMenuListener(
-        new PopupMenuListener() {
-          @Override
-          public void popupMenuCanceled(final PopupMenuEvent pme) {
-            currentPopupNode = null;
-          }
-
-          @Override
-          public void popupMenuWillBecomeInvisible(final PopupMenuEvent pme) {}
-
-          @Override
-          public void popupMenuWillBecomeVisible(final PopupMenuEvent pme) {}
-        });
     final HistoryTreeCellRenderer renderer = new HistoryTreeCellRenderer(uiContext);
     renderer.setLeafIcon(null);
     renderer.setClosedIcon(null);
@@ -382,6 +367,24 @@ public class HistoryPanel extends JPanel {
     }
     mouseWasOverPanel = mouseOverPanel;
     lastParent = parent;
+  }
+
+  public void setPopup(JPopupMenu popupNew) {
+    this.popup = popupNew;
+    tree.add(this.popup);
+    this.popup.addPopupMenuListener(
+        new PopupMenuListener() {
+          @Override
+          public void popupMenuCanceled(final PopupMenuEvent pme) {
+            currentPopupNode = null;
+          }
+
+          @Override
+          public void popupMenuWillBecomeInvisible(final PopupMenuEvent pme) {}
+
+          @Override
+          public void popupMenuWillBecomeVisible(final PopupMenuEvent pme) {}
+        });
   }
 
   private static final class HistoryTreeCellRenderer extends DefaultTreeCellRenderer {


game-app/game-headed/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
```diff
@@ -1952,14 +1957,23 @@ public void actionPerformed(final ActionEvent ae) {
...
+    }
     historyLog.setVisible(true);
   }
 
+  private HistoryNode getCurrentPopupNodeOrLastNode() {
+    return (historyPanel != null
+        ? historyPanel.getCurrentPopupNode()
+        : data.getHistory().getLastNode());
+  }
+
```
DanVanAtta: Why is this correct behavior?

DanVanAtta (commented):
This does look NPE safe now, and the approach is methodical. I totally agree with the adding of the `@nullable` annotations and I can see how that helped find the potential NPEs.

Overall, I see there are largely two fixes present.
(1) avoid NPE by skipping over `HistoryPanel` operations when `HistoryPanel` is null
(2) avoid NPE by using the "last node" from the `history` object rather than getting the current node from `historyPanel`

I commented on the lines implementing (2) for why that is correct behavior.

Regarding solution (1). Do we have a repro case for this issue? Do you know why/how `historyPanel` is null?

In essence, the fix is to 'delete' the lines that were made NPE safe. If `historyPanel` is null, those lines of code are going to effectively disappear. That might be okay, but we might be converting a "loud" error into a "silent" error. Instead of a pop-up, a person might see nothing happen - or the wrong thing happen. 

If we know for sure what the desired behavior is when `historyPanel` is null, then we'll have better insight to what a best fix would be. 

The code is built in a way where `historyPanel` really should be non-null, check: https://github.com/triplea-game/triplea/pull/13680/files#diff-916cadaa6875caf2c99436fd7c3649467fe57ad50b7557a4fe5722ac8e04cfb9L1941

The NPE's are coming from actions triggered by objects that are within `historyPanel`. Which is weird. The code is written in a way where `historyPanel` should be non-null, so one fix is to actually write the code that way.

A key line is; https://github.com/triplea-game/triplea/pull/13680/files#diff-916cadaa6875caf2c99436fd7c3649467fe57ad50b7557a4fe5722ac8e04cfb9L1814

In that line, the code spawns a new thread to assemble history actions and create a new `historyPanel` while doing so.

One possible solution would be to rewrite things like:

```
          historyPanel = new HistoryPanel(clonedGameData, historyDetailPanel, uiContext);
          final JPopupMenu popup = new JPopupMenu();
          popup.add(....)
          popup.add(....)
          historyPanel.add(popup);
```
In the above, `historyPanel` really ought to be non-null for any action within `popup`.

We could probably  rewrite that with a final `HistoryPanel`:
```
          historyPanel = new HistoryPanel(clonedGameData, historyDetailPanel, uiContext);
          final var historyPanelCopy = historyPanel;
          final JPopupMenu popup = new JPopupMenu();
          popup.add(.... // use historyPanelCopy here)
          popup.add(....)
          historyPanel.add(popup);
```
I did raise a few questions. If we know the cases where we get NPEs, it'll help us quite a bit. I don't know if we are converting loud-error behavior here into silent-error behavior. Sometimes "quietly do nothing" is exactly right, but I don't know if that is actually the case here.

game-app/game-headed/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
frigoref: @DanVanAtta It was late yesterday and I forgot to ask this in the PR. I am also not sure, but it was a way to get a `HistoryNode`.
I also was not able to understand how the `historyPanel` was able to be null at this point.
On further analysis I was thinking that it might have something to do with method `showHistory` is made accessible through the Getter of Action `showHistoryAction` that is attached to a hot key there and in there JPopupMenu `popup` is generated in `SwingUtilities.invokeLater`, but added afterwards to `historyPanel`.
Could you please have a look and let me know what your opinion on this is?

DanVanAtta: Yeah, I spent a solid hour looking at how `historyPanel` could be null and did not come to any conclusion. The hotkey was interesting, it does not look like a person can use a hotkey after right clicking in the history panel. I also wondered if maybe in the 'save game @' feature if maybe the history panel could be closed while the save game dialog was still open (it can't).

I wonder a bit if there is a race condition. Where the `invokeLater` that sets up the objects is delayed and something else comes in to `null` out historypanel.

The nulls are coming from within the "popup menu". I think if we re-write that to use a final, non-null `historyPanel` - we *should* avoid the null situation.In essence, the code as-is tries to do that. It's a real mystery where the null is coming from.

DanVanAtta: Grabbing the latest node is intriguing, but seems like it could be wrong if the desired save point is somewhere in the middle of the history.

frigoref: Could you please check [fix13679_NPEhistoryPanelTripleAFrame #5](https://github.com/triplea-game/triplea/pull/13680/commits/d542e9d8a6114fc5200a71f8337aaf4ca4241568) for the extraction off EDT in method `showHistory`?

game-app/game-headed/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
```diff
@@ -1811,53 +1829,25 @@ private void showHistory() {
...
-
-                @Override
-                public void actionPerformed(final ActionEvent ae) {
+                      TripleAFrame.this, data, getCurrentPopupNodeOrLastNode());
+                  if (historyPanel != null) {
+                    historyPanel.clearCurrentPopupNode();
+                  }
+                })
+            .add(
+                "\"Save Game at this point (BETA)\"",
```
DanVanAtta: Was the introduction of quotes intentional?
```suggestion
                "Save Game at this point (BETA)",
```

game-app/game-headed/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
```diff
@@ -1018,6 +1017,25 @@ public boolean getStrategicBombingRaid(final Territory location) {
         .orElse(null);
   }
 
+  private class HistoryPanelPopupMenuBuilder {
```
DanVanAtta: inner class may be static; if not static classes implicitly hold a reference to the parent class

```suggestion
  private static class HistoryPanelPopupMenuBuilder {
```

DanVanAtta: 2nd (nitpicky), for depth first method ordering and to reduce the space between method definition and usage, I recommend to move this class definition to below the method that first uses it (to be right above: `public void setEditDelegate`)

game-app/game-headed/src/main/java/games/strategy/triplea/ui/TripleAFrame.java
```diff
@@ -1811,53 +1810,25 @@ private void showHistory() {
...
-                  historyPanel.clearCurrentPopupNode();
-                }
-              });
-          popup.add(
-              new AbstractAction("Save Game at this point (BETA)") {
-                private static final long serialVersionUID = 1430512376199927896L;
-
-                @Override
-                public void actionPerformed(final ActionEvent ae) {
+                      TripleAFrame.this, data, popupHistoryPanel.getCurrentPopupNode());
```
DanVanAtta: @frigoref to double check on, the other methods within popup are using `clonedGameData`, we probably should be using `clonedGameData` everywhere? (Maybe?)

